PR #10870, opened 2025-08-22
refactor: avoid importing all of lodash when we only need groupBy

elevatebart:
When we import lodash natively, all of lodash is imported even if we only use 2 functions... this is unwanted


MilosPaunovic (approved):

MilosPaunovic (changes requested):
Maybe cover these two also

image

MilosPaunovic (approved):

PR #10887, opened 2025-08-25
fix: throw an error when trying to create a flow with a reserved keyw…

### LINKED ISSUE #5832: Cannot execute a flow named `pause`
### Describe the issue

You cannot execute a flow named `pause`.

Try for example to create the following flow and execute it:

``` yaml
id: pause
namespace: company.team
tasks:
  - id: pause
    delay: PT30S
  - id: hello
    message: Hello World! 🚀
```

This is because Micronaut redirect to the pause endpoint of the ExecutionController instead of the create endpoint:

```java
    @Post(uri = "/{namespace}/{id}", consumes = MediaType.MULTIPART_FORM_DATA)
    public Publisher<ExecutionResponse> create()
    }

    @Post(uri = "/{executionId}/pause")
    public void pause()
    }
```

The same issue occurs for all endpoints taken an executionId and using a POST method: pause, resume, kill, unqueue, ...

To fix that we can either:
- use a regex (to be validated): `{executionId:[^/]*}`
- rename those endpoints to use something that cannot be used as a flow identifier, like `_pause` so there are no possible mismatches.

### Environment

- Kestra Version: develop

Skraye:
…ord id

close #5832

### DIFF
--- a/core/src/main/java/io/kestra/core/validations/validator/FlowValidator.java
+++ b/core/src/main/java/io/kestra/core/validations/validator/FlowValidator.java
@@ -36,6 +36,19 @@ import static io.kestra.core.models.Label.SYSTEM_PREFIX;
 @Singleton
 @Introspected
 public class FlowValidator implements ConstraintValidator<FlowValidation, Flow> {
+    public static List<String> RESERVED_FLOW_IDS = List.of(
+        "pause",
+        "resume",
+        "force-run",
+        "change-status",
+        "kill",
+        "executions",
+        "search",
+        "source",
+        "disable",
+        "enable"
+    );
+
     @Inject
     private FlowService flowService;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/kestra/core/services/FlowService.java
```diff
@@ -118,6 +131,9 @@ public List<ValidateConstraintViolation> validate(final String tenantId, final S
                 try {
                     FlowWithSource flow = pluginDefaultService.parseFlowWithVersionDefaults(tenantId, source, true);
                     Integer sentRevision = flow.getRevision();
+                    if (RESERVED_FLOW_IDS.contains(flow.getId())) {
```
AcevedoR: wouldnt it cleaner to put it in FlowValidator ?

AcevedoR (changes requested):

core/src/main/java/io/kestra/core/services/FlowService.java
Skraye: True, forgot about it when I saw the validate method in flowService

AcevedoR (approved):
LGTM

PR #10891, opened 2025-08-25
chore(core): make playground feature enabled by default

MilosPaunovic:
Related to https://github.com/kestra-io/kestra-ee/issues/4555.

### DIFF
--- a/ui/src/components/inputs/EditorButtonsWrapper.vue
+++ b/ui/src/components/inputs/EditorButtonsWrapper.vue
@@ -68,7 +68,9 @@
 
     const {translateError, translateErrorWithKey} = useFlowOutdatedErrors();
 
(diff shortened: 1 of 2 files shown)


Piyush-r-bhaskar (changes requested):
Hey @MilosPaunovic 

When I start the application, I see toggle is on but there is no key (`editorPlayground: true`) in the localStorage to detect Playground is ON which causes `No Playground` toggle in Multipanel.

Image



But if you with current state, turn the toggle off from setting - Save you will have `editorPlayground: false` and then toggle on - Save --- Now you will have `editorPlayground: true` in localstorage and you will have the Playground toggle available in Multipanel.

MilosPaunovic:
You were absolutely right @Piyush-r-bhaskar, thanks for being vigilant, it seems my mind is still on vacation 🌴 I've pushed another commit to address your comment.

Piyush-r-bhaskar (approved):
LGTM 🚀

PR #10893, opened 2025-08-25
Fix/pause

loicmathieu:
The Pause task was previously immediatly termindated without taken into account any errors or finally block.
To allow processing those blocks, we need to store the terminated state in the output, then use it to resolve the final state.

### DIFF
--- a/core/src/main/java/io/kestra/core/services/FlowService.java
+++ b/core/src/main/java/io/kestra/core/services/FlowService.java
@@ -19,6 +19,7 @@ import io.kestra.core.repositories.FlowRepositoryInterface;
 import io.kestra.core.repositories.FlowTopologyRepositoryInterface;
 import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.utils.ListUtils;
+import io.kestra.plugin.core.flow.Pause;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/kestra/core/services/FlowService.java
```diff
@@ -250,7 +251,7 @@ public List<String> warnings(Flow flow, String tenantId) {
         // add warning for runnable properties (timeout, workerGroup, taskCache) when used not in a runnable
         flow.allTasksWithChilds().forEach(task -> {
             if (!(task instanceof RunnableTask<?>)) {
-                if (task.getTimeout() != null) {
+                if (task.getTimeout() != null && !(task instanceof Pause)) {
                     warnings.add("The task '" + task.getId() + "' cannot use the 'timeout' property as it's only relevant for runnable tasks.");
```
nkwiatkowski, line 255: I am not sure that all runnable task are instance of Pause

nkwiatkowski (approved):
LGTM

core/src/main/java/io/kestra/core/services/FlowService.java
loicmathieu: It's not what it says, Pause is a flowable and it supports timeout.
So we emit a warning if timeout is used in a flowable expcet if it's the Pause task

PR #10897, opened 2025-08-25
fix(trigger): prevent scheduler crash on large duration

### LINKED ISSUE #621: S3 Interval Trigger Crashes Scheduler on Large interval — Cannot Recover Without State Cleanup
When using an `S3` trigger with an excessively large `interval` value, the **Kestra scheduler crashes** due to an uncaught SQL exception caused by an invalid `EpochDay`.

Even after correcting the interval to a valid value, **Kestra does not recover**. The only way to fix it is by **manually removing the trigger state** from the database.

---

### Flow Example

```yaml
triggers:
  - id: watch_uploads
    type: io.kestra.plugin.aws.s3.Trigger
    interval: PT99999999898989981M  # Intentionally large value

15:39:08.979 ERROR jdbc-queue-WorkerTriggerResult_0 .c.u.ThreadUncaughtExceptionHandlers Caught an exception in Thread[#61,jdbc-queue-WorkerTriggerResult_0,5,main]. Shutting down.
java.time.DateTimeException: Invalid value for EpochDay (valid values -365243219162 - 365241780471): 69444444394570
```
impact:
  - Kestra scheduler enters a crash loop and stops functioning.
  - Fixing the interval in the flow does not resolve the crash.
  - The system cannot recover automatically.
  - Recovery requires manual cleanup of the trigger state from the database.

### Environment

- **Kestra version:** 0.23.1  
- **Deployment method:** Kubernetes (Helm chart)  
- **Database:** PostgreSQL 14  
- **Cloud provider:** DigitalOcean

Malaydewangan09:
- closes https://github.com/kestra-io/plugin-aws/issues/621

### DIFF
--- a/core/src/main/java/io/kestra/core/models/triggers/PollingTriggerInterface.java
+++ b/core/src/main/java/io/kestra/core/models/triggers/PollingTriggerInterface.java
@@ -3,8 +3,13 @@ package io.kestra.core.models.triggers;
 import io.kestra.core.models.annotations.PluginProperty;
 import io.kestra.core.models.conditions.ConditionContext;
 import io.kestra.core.models.executions.Execution;
+import io.kestra.core.runners.RunContext;
 import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.time.DateTimeException;
 import java.time.Duration;
 import java.time.ZonedDateTime;
 import java.util.Optional;
@@ -30,7 +35,8 @@ public interface PollingTriggerInterface extends WorkerTriggerInterface {
      * Schedulable triggers must override this method.
      */
     default ZonedDateTime nextEvaluationDate(ConditionContext conditionContext, Optional<? extends TriggerContext> last) throws Exception {
-        return ZonedDateTime.now().plus(this.getInterval());
+        RunContext runContext = conditionContext.getRunContext();
+        return computeNextEvaluationDate(runContext);
     }
 
     /**
@@ -38,6 +44,29 @@ public interface PollingTriggerInterface extends WorkerTriggerInterface {
      * Schedulable triggers must override this method as it's used to init them when there is no evaluation date.
      */
     default ZonedDateTime nextEvaluationDate() {
-        return ZonedDateTime.now().plus(this.getInterval());
+        return computeNextEvaluationDate(null);
+    }
+
+    /**
+     * computes the next evaluation date using the configured interval.
+     * Falls back to 60 seconds if the interval causes date overflow.
+     */
+    private ZonedDateTime computeNextEvaluationDate(@Nullable RunContext runContext) {
+        Duration interval = this.getInterval();
+        Logger logger = resolveLogger(runContext);
+
+        try {
+            return ZonedDateTime.now().plus(interval);
+        } catch (DateTimeException | ArithmeticException e) {
+            logger.warn("Trigger interval '{}' for type '{}' causes date overflow: {}. Using 60 seconds fallback.", interval, this.getClass().getName(), e.getMessage());
+            return ZonedDateTime.now().plus(Duration.ofSeconds(60));
+        }
+    }
+
+    private Logger resolveLogger(@Nullable RunContext runContext) {
+        if (runContext != null) {
+            return runContext.logger();
+        }
+        return LoggerFactory.getLogger(this.getClass());
     }
 }
--- a/core/src/main/java/io/kestra/core/models/triggers/PollingTriggerInterface.java
+++ b/core/src/main/java/io/kestra/core/models/triggers/PollingTriggerInterface.java
@@ -1,11 +1,12 @@
 package io.kestra.core.models.triggers;
 
+import io.kestra.core.exceptions.InvalidTriggerConfigurationException;
 import io.kestra.core.models.annotations.PluginProperty;
 import io.kestra.core.models.conditions.ConditionContext;
 import io.kestra.core.models.executions.Execution;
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/kestra/core/models/triggers/PollingTriggerInterface.java
```diff
@@ -30,14 +32,26 @@ public interface PollingTriggerInterface extends WorkerTriggerInterface {
      * Schedulable triggers must override this method.
      */
     default ZonedDateTime nextEvaluationDate(ConditionContext conditionContext, Optional<? extends TriggerContext> last) throws Exception {
-        return ZonedDateTime.now().plus(this.getInterval());
+        try {
+            return ZonedDateTime.now().plus(this.getInterval());
+        } catch (DateTimeException | ArithmeticException e) {
+            RunContext runContext = conditionContext.getRunContext();
+            if (runContext != null) {
```
loicmathieu: It should never be null here.

loicmathieu: I wonder if we should not default to one year instead of 60s.

core/src/main/java/io/kestra/core/models/triggers/PollingTriggerInterface.java
```diff
@@ -30,14 +32,26 @@ public interface PollingTriggerInterface extends WorkerTriggerInterface {
...
     /**
      * Compute the next evaluation date of the trigger: by default, it uses the current date and the interval.
      * Schedulable triggers must override this method as it's used to init them when there is no evaluation date.
      */
     default ZonedDateTime nextEvaluationDate() {
-        return ZonedDateTime.now().plus(this.getInterval());
+        try {
+            return ZonedDateTime.now().plus(this.getInterval());
+        } catch (DateTimeException | ArithmeticException e) {
+            return ZonedDateTime.now().plus(Duration.ofSeconds(60));
```
loicmathieu: I wonder if we should not default to one year, I'll raise that to the plugin tream

core/src/main/java/io/kestra/core/models/triggers/PollingTriggerInterface.java
loicmathieu: Moreover, are you sure an ArithmeticException can be thrown here?

fhussonnois (changes requested):
Please create a single method called by both methods. If `runContext` is `null`, it is best to continue logging the exception (with the trigger type).

Based on the initial problem, it would be good to check that the defined `interval` cannot result in an invalid date with a constraint validator.

fdelbrayelle (approved):

core/src/main/java/io/kestra/core/models/triggers/PollingTriggerInterface.java
```diff
@@ -30,14 +35,38 @@ public interface PollingTriggerInterface extends WorkerTriggerInterface {
...
+            logger.warn("Trigger interval '{}' for type '{}' causes date overflow: {}. Using 60 seconds fallback.", interval, this.getClass().getName(), e.getMessage());
+            return ZonedDateTime.now().plus(Duration.ofSeconds(60));
+        }
+    }
+
+    private Logger resolveLogger(@Nullable RunContext runContext) {
+        if (runContext != null) {
+            return runContext.logger();
+        }
+        return LoggerFactory.getLogger(this.getClass());
```
loicmathieu: Please use a static final logger here to avoid re-creating it each time the method is called

loicmathieu (changes requested):

anna-geller:
@Malaydewangan09 can you address the final feedback and merge? looks like this PR got stuck

core/src/main/java/io/kestra/core/models/triggers/PollingTriggerInterface.java
Malaydewangan09: Yes

Malaydewangan09:
Hey Anna, yes, as decided, I'll update this to disable the trigger if this crash happens. I need to dig up a little on how to do that in core.
I'll close this shortly!

Malaydewangan09:
@loicmathieu, could you please have a look, if this is a correct way to go forward? 🤔

scheduler/src/main/java/io/kestra/scheduler/AbstractScheduler.java
```diff
@@ -537,6 +554,55 @@ private List<FlowWithTriggers> computeSchedulable(List<FlowWithSource> flows, Li
...
+
+            triggerQueue.emit(disabledTrigger);
+
+            schedulableNextDate.remove(disabledTrigger.uid());
+
+            synchronized (this) {
+                schedulable = schedulable.stream()
+                    .filter(f -> !f.getTriggerContext().uid().equals(disabledTrigger.uid()))
+                    .toList();
+            }
```
loicmathieu: You don't need to do that, triggers are always taken from the triggerState (database or kafka store) so no need to mess with the schedulableNextDate and schedulable, those would be updated automatically.

loicmathieu (changes requested):
As this is non trivial it would be great to have a test that would reproduce a failingf trigger and assert it's disabled in the triggerState after that

Malaydewangan09:
Yes, I'll add just wanted to confirm this a correct way to go ahead!

scheduler/src/main/java/io/kestra/scheduler/AbstractScheduler.java
Malaydewangan09: Alright!

fdelbrayelle (approved):

Malaydewangan09:
@loicmathieu, added test for the changes!

loicmathieu (approved):
LGTM

PR #10909, opened 2025-08-26
feat(core): implement improved graph for namespace dependencies view

### LINKED ISSUE #10634: Consider changing a namespace-level Dependencies graph
Currently, the `Dependencies Graph` feature focuses on flows and executions. Should we introduce a new graph visualization specifically for namespace-level dependencies?

I'd say yes to this, but I’m inviting @anna-geller and @Ben8t for discussion and confirmation before starting the actual work. It shouldn't be a big change since the endpoint structure is quite similar to what we already have for flows and executions.

Current looks of namespace dependencies:

Image

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/10634.


Piyush-r-bhaskar (approved):
Looks very good. 🚀

Please take care of the Frontend test failing before merge if possible

PR #10922, opened 2025-08-27
feat(retry): use the retry policy on HttpClient

### LINKED ISSUE #10918: Be able to configure the retry policy under HttpConfiguration and HttpClient
### Feature description

To avoid to implement exponential back-off as a retry policy in the plugins, it could be great to be able to define the policy in `HttpClient` in Kestra core instead of putting the same logic in all plugins.

[Here is an example](https://github.com/kestra-io/plugin-dbt/pull/203) where we had to put it in the plugin DBT.

@loicmathieu What do you think about it? We discussed about it we @tchiotludo this morning.

fdelbrayelle:
closes #10918 

---

---


core/src/main/java/io/kestra/core/http/client/HttpClient.java
```diff
@@ -176,123 +164,103 @@ public List<Proxy> select(URI uri) {
...
     }
 
     private HttpClientContext clientContext(HttpRequest request) {
-        ContextBuilder contextBuilder = ContextBuilder.create();
-
+        var contextBuilder = ContextBuilder.create();
         return contextBuilder.build();
     }
 
+    private static Throwable unwrap(Throwable t) {
```
loicmathieu: This may unwrap too much as there can be an IOEXcepton wrap into an Apache client exception.
I wonder if we shuld not just take the cause...

loicmathieu (approved):
LGTM, I'm not sure about the unwrap for exception, but if you think it's good as is let's ship it!

Malaydewangan09 (approved):

PR #10936, opened 2025-08-27
feat(executions): allow flow trigger on concurrency limit

loicmathieu:
Closes https://github.com/kestra-io/kestra-ee/issues/3270

This allow listening to the QUEUED state on the flow trigger. This also fixes an issue that when concurrency limit is setup, you would not listen to the RUNNING state.


brian-mulier-p (approved):
LGTM

PR #10964, opened 2025-08-28
feat(storage): add purge kv task

### LINKED ISSUE #10210: Expired KVs are not deleted from object storage
### Describe the issue

We have some flows that use short lived KVs (expires in less than 1 minute) and we noticed the KV UI page starting to get too slow to open.

By checking our S3 bucket, I noticed there are thousands of objects there for the expired KVs that were never deleted.

Can Kestra have a garbage collection process to remove expired keys from KV backend automatically?

Meanwhile, is there any routine we can programatically execute (even from kestra itself) to remove only expired keys from S3?

### Environment

- Kestra Version: v0.22.9

nkwiatkowski:
Fix #10210

---

---

### DIFF
--- a/core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
+++ b/core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
@@ -120,16 +120,42 @@ public class InternalKVStore implements KVStore {
      */
     @Override
     public List<KVEntry> list() throws IOException {
-        List<FileAttributes> list;
-        try {
-            list = this.storage.list(this.tenant, this.namespace, this.storageUri(null));
-        } catch (FileNotFoundException e) {
-            return Collections.emptyList();
-        }
+        List<FileAttributes> list = listAllFromStorage();
         return list.stream()
             .map(throwFunction(KVEntry::from))
             .filter(kvEntry -> Optional.ofNullable(kvEntry.expirationDate()).map(expirationDate -> Instant.now().isBefore(expirationDate)).orElse(true))
             .toList();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<KVEntry> listAll() throws IOException {
+        List<FileAttributes> list = listAllFromStorage();
+        return list.stream()
+            .map(throwFunction(KVEntry::from))
+            .toList();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<KVEntry> listExpired() throws IOException {
+        List<FileAttributes> list = listAllFromStorage();
+        return list.stream()
+            .map(throwFunction(KVEntry::from))
+            .filter(kvEntry -> Optional.ofNullable(kvEntry.expirationDate()).map(expirationDate -> Instant.now().isAfter(expirationDate)).orElse(false))
+            .toList();
+    }
+
+    private List<FileAttributes> listAllFromStorage() throws IOException {
+        try {
+            return this.storage.list(this.tenant, this.namespace, this.storageUri(null));
+        } catch (FileNotFoundException e) {
+            return Collections.emptyList();
+        }
     }
 
     /**
--- a/core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
+++ b/core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
@@ -138,18 +138,6 @@ public class InternalKVStore implements KVStore {
             .toList();
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public List<KVEntry> listExpired() throws IOException {
-        List<FileAttributes> list = listAllFromStorage();
-        return list.stream()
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
```diff
@@ -120,18 +120,44 @@ public boolean delete(String key) throws IOException {
...
+        return list.stream()
+            .map(throwFunction(KVEntry::from))
+            .toList();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<KVEntry> listExpired() throws IOException {
```
loicmathieu: I'm not sure we would ever reuse this method so it may be better to use listAll() and filter on expired inside the task

core/src/main/java/io/kestra/plugin/core/kv/PurgeKV.java
```diff
@@ -0,0 +1,186 @@
...
+
+    @Override
+    public Output run(RunContext runContext) throws Exception {
+        List<String> kvNamespaces = findNamespaces(runContext);
+        boolean expired = runContext.render(expiredOnly).as(Boolean.class).orElse(true);
+        String renderedKeyPattern = runContext.render(keyPattern).as(String.class).orElse(null);
+        boolean keyFiltering = StringUtils.isNotBlank(renderedKeyPattern);
+        AtomicLong count = new AtomicLong();
+        for (String ns : kvNamespaces) {
+            KVStore kvStore = runContext.namespaceKv(ns);
```
loicmathieu, line 99: The namespaceKv() is supposed to only allow accessing a single namespace KV.

core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
```diff
@@ -120,18 +120,44 @@ public boolean delete(String key) throws IOException {
...
             .map(throwFunction(KVEntry::from))
             .filter(kvEntry -> Optional.ofNullable(kvEntry.expirationDate()).map(expirationDate -> Instant.now().isBefore(expirationDate)).orElse(true))
             .toList();
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<KVEntry> listAll() throws IOException {
```
loicmathieu, line 134: exposing listAll breaks the current KV implementation which only allow accessing a KV from a single namespace.
@brian-mulier-p @fhussonnois WDYT?

core/src/main/java/io/kestra/plugin/core/kv/PurgeKV.java
```diff
@@ -0,0 +1,186 @@
...
+            .build();
+    }
+
+    @VisibleForTesting
+    protected List<String> findNamespaces(RunContext runContext) throws IllegalVariableEvaluationException {
+        String tenantId = runContext.flowInfo().tenantId();
+        String currentNamespace = runContext.flowInfo().namespace();
+        FlowRepositoryInterface flowRepositoryInterface = ((DefaultRunContext) runContext)
+            .getApplicationContext().getBean(FlowRepositoryInterface.class);
+        List<String> distinctNamespaces = flowRepositoryInterface.findDistinctNamespace(tenantId);
```
loicmathieu, line 136: Accessing a repository for a task is a bad pattern that would soon be prohibited.
Doing so creates technical debt

core/src/main/java/io/kestra/plugin/core/kv/PurgeKV.java
```diff
@@ -0,0 +1,186 @@
...
+        String renderedKeyPattern = runContext.render(keyPattern).as(String.class).orElse(null);
+        boolean keyFiltering = StringUtils.isNotBlank(renderedKeyPattern);
+        AtomicLong count = new AtomicLong();
+        for (String ns : kvNamespaces) {
+            KVStore kvStore = runContext.namespaceKv(ns);
+            List<KVEntry> kvEntries = new ArrayList<>();
+            if (expired){
+                kvEntries.addAll(kvStore.listExpired());
+            } else {
+                kvEntries.addAll(kvStore.listAll());
```
loicmathieu: I don't understand here, if you create one KV per namesapce, why do you need to listAll()? The existing list() would work

loicmathieu (changes requested):
This new task raises several concerns, I'm not sure it's how it should be done.

core/src/main/java/io/kestra/plugin/core/kv/PurgeKV.java
nkwiatkowski: list filter expired key, and if I change that, I break the interface contract

core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
nkwiatkowski: this list only the keys of the current KV store, for the current namespace. It's basically the list() without the expired filtering

PR #10975, opened 2025-08-28
New priority attribute on Plugin annotation

fdelbrayelle:
Part-of: https://github.com/kestra-io/kestra-ee/issues/4841 

This PR introduces support for prioritizing plugin class resolution when multiple plugins share the same FQN (fully qualified name), using the new `@Plugin(priority = PRIMARY)` annotation.

**What changed?**

- `DefaultPluginRegistry#getPluginClassesByIdentifier()` now filters duplicates by selecting:
  - A plugin annotated with `@Plugin(priority = PRIMARY)` if available.
  - Otherwise, it falls back to the first available candidate (`SECONDARY` or no annotation).
- This applies to both standard plugin classes and aliases.

**How to use?**

In your plugin class, annotate it like this:

```java
@Plugin(priority = Plugin.Priority.PRIMARY)
public class MyTask extends Task {
    // preferred implementation
}
```

Or for a lower-priority fallback:

```java
@Plugin(priority = Plugin.Priority.SECONDARY)
public class MyTask extends Task {
    // will be ignored if a PRIMARY exists
}
```

This allows you to load multiple plugins with the same class name (e.g., from different JARs), while keeping only the preferred one active.

For instance, we'll have `io.kestra.plugin.git.TenantSync` in both OSS and EE plugins (with the same FQN = package + task name). In this case the EE version will have the `PRIMARY` priority and the OSS the `SECONDARY` one.

---

---


core/src/main/java/io/kestra/core/plugins/DefaultPluginRegistry.java
```diff
@@ -209,33 +198,69 @@ public void register(final RegisteredPlugin plugin) {
...
         return classes;
     }
 
+    private static Class<? extends Plugin> getHighestPriorityPluginClass(Map.Entry<String, List<Class<? extends Plugin>>> entry) {
+        List<Class<? extends Plugin>> classList = entry.getValue();
+
+        Class<? extends Plugin> selected = null;
+        for (Class<? extends Plugin> cls : classList) {
+            io.kestra.core.models.annotations.Plugin annotation = cls.getAnnotation(io.kestra.core.models.annotations.Plugin.class);
+            if (annotation != null && annotation.priority() == io.kestra.core.models.annotations.Plugin.Priority.PRIMARY) {
```
fhussonnois: Please, create a static method `isPrimary` in the `Plugin` model class for this.

Tips: use `Objects.equals(Priority.PRIMARY, annotation)` instead of checking the `null`

fhussonnois (approved):
LGTM!

model/src/main/java/io/kestra/core/models/annotations/Plugin.java
```diff
@@ -36,6 +36,13 @@
      */
     String[] aliases() default {};
 
+    Priority priority() default Priority.SECONDARY;
```
loicmathieu, line 39: Shouldn,'t it be PRIMARY by default to avoid the cost of searching fo a primary plugin for all plugins?

loicmathieu (changes requested):
I wonder why it's not EE only as only EE would override a plugin class right?

model/src/main/java/io/kestra/core/models/annotations/Plugin.java
fdelbrayelle: No because : 
- legacy plugins don't have the attribute
- new OSS plugins will have SECONDARY (because we could add another EE plugin for the same tech or need in the future)
- new EE plugins will have PRIMARY

fdelbrayelle:
> I wonder why it's not EE only as only EE would override a plugin class right?

Yes @loicmathieu, I think I'll open a PR only on EE to update `EEPluginRegistry` then!

fdelbrayelle:
I'll keep open the PR just for the Plugin annotation change @loicmathieu @fhussonnois

model/src/main/java/io/kestra/core/models/annotations/Plugin.java
loicmathieu: I don't think it's a good idea, it's counter intuitive and seems like all plugins has a primary plugin "somewhere".
In this case makes all plugins have an EE overriding capability and remove the attribute.

And again, it would just makes Kestra starts slower as it would search for an alternative on each of our 900 plugins and you would pay the miss cost as almost none have an alternative.

When you need to override a plugin, switch it from the default of PRIMARY to SECONDARY and override it, it should be rare enought so we could afford it.

Please also include timings comparison before and after the change for the loading of plugins, they are logged .

fhussonnois (approved):

model/src/main/java/io/kestra/core/models/annotations/Plugin.java
fdelbrayelle: Before the fix (scan & register time):

```
11:51:18.331 INFO  standalone   i.kestra.core.plugins.PluginScanner Registered 455 plugins from 82 groups (scan done in 472ms)
11:51:18.192 INFO  standalone   i.k.c.plugins.DefaultPluginRegistry Registered if absent plugins from path /home/francois/dev/plugins in 636 ms
```

After the fix (scan & register time):

```
12:54:27.663 INFO  standalone   i.kestra.core.plugins.PluginScanner Registered 455 plugins from 82 groups (scan done in 478ms)
12:54:27.999 INFO  standalone   i.k.c.plugins.DefaultPluginRegistry Registered if absent plugins from path /home/francois/dev/plugins in 575 ms
```

So we even gain some time in the end after the optim we did with @fhussonnois

loicmathieu: Timings on a single run is usually not accurate so when there are a small difference, it could be a run artefact.
Anyway, I read it as it lead to the same performance characteristic so :+1:

PR #10986, opened 2025-08-29
fix(executions): clear errors/finally/afterExecution branches when ch…

loicmathieu:
…anging the state of a taskrun

As changing the state of a taskrun will restart the flow, if we didn't clear those branches, the flow would not resart properly.

Fixes https://github.com/kestra-io/kestra-ee/issues/3211


brian-mulier-p (approved):
LGTM

PR #10990, opened 2025-08-29
fix(storage): delete the metadata with the key when we delete an KV s…

### LINKED ISSUE #10989: KV metadata isn't deleted when value is deleted
Simple flow to reproduce:

```yaml
id: kv_store_set
namespace: company.team

tasks:
  - id: kv_set
    key: key_to_delete
    value: "my value"
    kvType: STRING
    ttl: PT15M

  - id: kv_delete
    type: io.kestra.plugin.core.kv.Delete
    key: key_to_delete
```
Before the fix, a key_to_delete.metadata file should be present in the KV store after this flow.
After the fix, no key_to_delete file should be left in the KV store

nkwiatkowski:
…tore entry

Fix #10989

---

---

### DIFF
--- a/core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
+++ b/core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
@@ -10,6 +10,7 @@ import jakarta.annotation.Nullable;
 import java.io.ByteArrayInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.net.URI;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
```diff
@@ -112,7 +113,14 @@ public Optional<String> getRawValue(String key) throws IOException, ResourceExpi
     @Override
     public boolean delete(String key) throws IOException {
         KVStore.validateKey(key);
-        return this.storage.delete(this.tenant, this.namespace, this.storageUri(key));
+        URI uri = this.storageUri(key);
+        boolean deleted = this.storage.delete(this.tenant, this.namespace, uri);
+        URI metedataURI = URI.create(uri.getPath() + ".metadata");
```
brian-mulier-p:
```suggestion
        URI metadataURI = URI.create(uri.getPath() + ".metadata");
```

brian-mulier-p (approved):
LGTM

PR #10992, opened 2025-09-01
fix(core): allow removal of block items from no code editor

MilosPaunovic:
Closes https://github.com/kestra-io/kestra-ee/issues/4862.

### DIFF
--- a/ui/src/components/flows/tasks/TaskList.vue
+++ b/ui/src/components/flows/tasks/TaskList.vue
@@ -5,7 +5,7 @@
             :elements="items"
             :section
             :block-schema-path="[blockSchemaPath, 'properties', root, 'items'].join('/')"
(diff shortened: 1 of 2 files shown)


ui/src/components/flows/tasks/TaskList.vue
```diff
@@ -5,7 +5,7 @@
             :elements="items"
             :section
             :block-schema-path="[blockSchemaPath, 'properties', root, 'items'].join('/')"
-            @remove="(yaml) => flowStore.flowYaml = yaml"
+            @remove="removeItems"
```
elevatebart: NIT: Since we only remove item can we call it `removeItem`?

```suggestion
            @remove="removeItem"
```

elevatebart (approved):
one nitpick that I will fix and we're golden

PR #10994, opened 2025-09-01
chore(core): improve dependency graph user experience

### LINKED ISSUE #10925: Dependency graph, minor UX improvements
According to the initial designs, the table rows inside the new `Dependency Graph` view should look like this:

Image

https://www.figma.com/design/ew0uXk0NRXJ2NBBJTNe2n1/01_UI?node-id=8375-32924&m=dev

Top one is for `Flows` (and same is used for `Namespaces`) and the bottom one for `Executions`.

`Executions` one is alright, and we're achiveing that look already, but the `Flow` one is tricky, as we don't have (and can't have, according to [this](https://github.com/kestra-io/kestra/issues/10633#issuecomment-3167326833)) the revision parameter, so the row looks kind of empty on the right side:

Image

Maybe we could tweak the design a bit, potentially add an icon on the right (either for one or for both) that could signify that by clicking on it users could open that particular flow, or something similar.

Also, one more thing is the coloring scheme - we had to do a bit of tweaking of it [here](https://github.com/kestra-io/kestra/issues/10680), so it might be good idea to revisit the scheme and adjust colors where needed, now that the feature is up and running.

Needed colors are this ones:

```js
const VARIABLES = {
    node: {
        // This is the regular node, without any special coloring/selection
        default: {
            background: "--ks-dependencies-node-background",
            border: "--ks-dependencies-node-border",
        },
        // This is used for nodes connected to the selected ones
        faded: {
            background: "--ks-dependencies-node-background-selected-level2",
            border: "--ks-dependencies-node-border-selected-level2",
        },
        selected: {
            background: "--ks-dependencies-node-background-selected",
            border: "--ks-dependencies-node-border-selected",
        },
        hovered: {
            background: "--ks-dependencies-node-background-hover",
            border: "--ks-dependencies-node-border-hover",
        },
    },
    edge: {
        default: "--ks-dependencies-node-border",
        faded: "--ks-dependencies-edge-selected-level2",
        hovered: "--ks-dependencies-edge-hover",
    },
};
```

And maybe we can adjust naming of the variables to match the ones from above, to be a bit simpler to follow during development. It would need a final exported file from Figma as the product of it, but Nico will know about which file I'm talking about.

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/10925.
Closes https://github.com/kestra-io/kestra-ee/issues/4865.


Piyush-r-bhaskar (approved):
LGTM 🚀

PR #10995, opened 2025-09-01
feat(ai): langchain4j implementation of AI Copilot

brian-mulier-p:


webserver/build.gradle
```diff
@@ -22,8 +22,9 @@ dependencies {
     implementation "io.micronaut.cache:micronaut-cache-core"
     implementation "io.micronaut.cache:micronaut-cache-caffeine"
 
-    implementation 'com.google.adk:google-adk:0.2.0'
-
+    implementation("com.posthog.java:posthog:1.2.0")
+    implementation("dev.langchain4j:langchain4j:1.3.0")
```
loicmathieu: 1.4.0 is out now, it would be interesting to upgrade to it.

webserver/src/main/java/io/kestra/webserver/services/ai/AiConfiguration.java
```diff
@@ -0,0 +1,15 @@
...
+
+public interface AiConfiguration {
+    String apiKey();
+    String modelName();
+    default double temperature() {
+        return 0.7;
+    };
+    default int seed() {
+        return 50000;
+    };
```
loicmathieu: Seed is for testing purpose where you want to have more chance to generate the same result on multiple call. I would not allow configuring it here and I have no idea if such default value is correct or not.

loicmathieu: I would add:
- topP
- logRequests/logResponses for debug purpose
- maybe baseUrl for hosted models

webserver/src/main/java/io/kestra/webserver/services/ai/AiService.java
```diff
@@ -0,0 +1,43 @@
...
+    private PluginFinder pluginFinder(AiConfiguration aiConfiguration) {
+        return AiServices.builder(PluginFinder.class)
+            .chatModel(this.chatModel(aiConfiguration.modelName(), aiConfiguration.apiKey(), aiConfiguration.temperature(), aiConfiguration.seed(), 1024))
+            .build();
+    }
+
+    private FlowYamlBuilder flowYamlBuilder(AiConfiguration aiConfiguration) {
+        return AiServices.builder(FlowYamlBuilder.class)
+            .chatModel(this.chatModel(aiConfiguration.modelName(), aiConfiguration.apiKey(), aiConfiguration.temperature(), aiConfiguration.seed(), 50000))
+            .build();
```
loicmathieu: You can use the same ChatModel multiple time.
In fact I would use a single AiService where you inject the ChatModel and have multiple CDI produces (factory), one for each LLM provides.
With that, the AiService would be totally independant on own the ChatModel is built and its parameters

webserver/src/main/java/io/kestra/webserver/services/ai/FlowYamlBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+        - Include AT LEAST ONE trigger if execution should start based on an event or interval.
+        - Triggers expose some variables that can be accessed through `trigger.outputName` in expressions. The only variables available are those defined in the trigger's outputs.
+        - Unless specified by the user, never assume a local port to serve any content, always use a remote URL (like a public HTTP server) to fetch content.
+        - Unless specified by the user, do not use any authenticated API, always use public APIs or those that don't require authentication.
+        - To avoid escaping quotes, use double quotes first and if you need quotes inside, use single ones. Only escape them if you have 3+ level quotes, for example: `message: "Hello {_{inputs.userJson | jq('.name')}_}"`.
+        - A property key is unique within each type.
+        - When fetching data from the JDBC plugin, always use fetchType: STORE.
+        - Manipulating date in pebble expressions can be done through `dateAdd` (`{_{now()|dateAdd(-1,'DAYS')}_}`) and `date` filters (`{_{"July 24, 2001"|date("yyyy-MM-dd",existingFormat="MMMM dd, yyyy")}_}`).
```
loicmathieu: This may be annoying in the LLM didn't obey this rules it may break the copilot.
I think it would be better to either:
- Use a lenient template factory as we did in the plugin-ai
- Use a custom template factory with a different separator than `{{}}`

loicmathieu (changes requested):

webserver/src/main/java/io/kestra/webserver/services/ai/AiService.java
brian-mulier-p: Please have another look, I changed the way I handled it. In fact for Posthog tracing it's painful and we need to reinstantiate a new ChatModel for every request to add a conversation-aware listener to it...
On the version you reviewed, I was not instantiating a new ChatModel every time but one at startup and I was reusing the agents as-is. With posthog I didn't find a way anymore... We can discuss and huddle tomorrow if you want.

webserver/src/main/java/io/kestra/webserver/services/ai/AiConfiguration.java
brian-mulier-p: I remember you said that but tbh I find it better that we fix it so people get most likely results that we "almost have fully tested". Anyway I removed that from global conf as it's not available everywhere...
For baseUrl it would be a per-configuration-implementation parameter :+1: 
I'll add logRequest/Response & topP

webserver/src/main/java/io/kestra/webserver/services/ai/AiService.java
loicmathieu: Thinking about it, to make it more accurate, we would need a memory (by default, it's an in-memory memory with the last 10 messages). But, well, with memory comes sessionId and storage which are more challenging.
I think it would be great to open a follow issue about it.

webserver/src/main/java/io/kestra/webserver/services/ai/MetricChatModelListener.java
```diff
@@ -0,0 +1,87 @@
...
+    @Inject
+    private MeterRegistry meterRegistry;
+
+    @Override
+    public void onResponse(ChatModelResponseContext responseContext) {
+        ChatRequest request = responseContext.chatRequest();
+        ChatResponse response = responseContext.chatResponse();
+
+        String[] tags = new String[]{
+            "messages_count", String.valueOf(request.messages().size()),
```
loicmathieu, line 30:
```suggestion
```

A count should never be a tag as it may create a very big cardinality of metrics

loicmathieu (changes requested):

webserver/src/main/java/io/kestra/webserver/services/ai/AiService.java
brian-mulier-p: I went for that at first but I found out that ChatMemory was only providing a message history so I decided to go for my own...

webserver/src/main/java/io/kestra/webserver/services/ai/FlowYamlBuilder.java
brian-mulier-p: I went for the second one

webserver/src/main/java/io/kestra/webserver/services/ai/AiService.java
brian-mulier-p: https://github.com/kestra-io/kestra-ee/issues/4893

webserver/src/main/java/io/kestra/webserver/services/ai/MetricChatModelListener.java
brian-mulier-p: Seen with @tchiotludo, this specific case will not generate a lot because we're speaking about messages coming from users and will quickly reset

webserver/src/main/java/io/kestra/webserver/services/ai/openai/OpenAiConfiguration.java
```diff
@@ -0,0 +1,23 @@
...
+
+import io.kestra.webserver.services.ai.AiConfiguration;
+import io.micronaut.context.annotation.ConfigurationProperties;
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.bind.annotation.Bindable;
+
+@ConfigurationProperties(value = "kestra.ai.openai")
+public record OpenAiConfiguration(
+    String apiKey,
+    @Bindable(defaultValue = "gpt-5-nano")
```
loicmathieu, line 11: I think nano is not a good fit for code completion, maybe the default should be a bigger model.

loicmathieu (approved):
LGTM

webserver/src/main/java/io/kestra/webserver/services/ai/openai/OpenAiConfiguration.java
brian-mulier-p: from my testing it does the job but... Issue is that OpenAI is already SUPER long so moving to larger models moves from 20 sec to 40 sec or worst... Anyway the time is not comparable to what we have with gemini but...

PR #11011, opened 2025-09-02
Chore/the big refacto

loicmathieu:
Part-of: https://github.com/kestra-io/kestra-ee/issues/4884

Extract the scheduler, the worker and the executor in their own module.
As the executor is currently inside JDBC, only the ExecutorService is extracted.

This allow a clear separation between the components and would allow easier future evalolution.

The StandaloneRunner has been moved to the cli module as it's only there that it is used, a new TestRunner has been added for unit tests. The runner is no more created unconditionally but only when started the standalone.


loicmathieu:
All tests failures are flaky test.
It's ready to merge when approved.

fhussonnois (approved):

PR #11018, opened 2025-09-02
fix(docs): fix trigger docs

aj-emerich:


anna-geller (approved):
I don't know in the end why titles have dots at the end, but pretty much all plugin titles have them, so I kept them in the past, up to you AJ

aj-emerich:
For property titles, they look bad if they do not include a complete sentence (e.g., The file URI.) When it is "Trigger a flow on a schedule." in an example, I think it is okay. This is inconsistent all over the place though, so it might be that way for a while.

PR #11027, opened 2025-09-03
chore(deps): add langchain4j version to platform

brian-mulier-p:


loicmathieu (approved):
LGTM, we would need to update the plugin-ai to take its version from the platform.
For that, we would also need the community BOM inside the platoform so they would be aligned, see https://github.com/kestra-io/plugin-ai/blob/master/build.gradle#L43

Issue #11035, opened 2025-09-03
Multi-panel can open two times the no code


Ben8t:
### Describe the issue

https://github.com/user-attachments/assets/805a7666-e1bf-4c28-98eb-0a9ffb5a34e0

### Environment

- Kestra Version: develop

elevatebart:
closed by [#11092](https://github.com/kestra-io/kestra/pull/11092)

PR #11092, opened 2025-09-04
fix(core): Multipanel layout fixes

elevatebart:
fixes the follwoing bug:

- Open a flow
- resize any of the panels
- Open a new tab using the button on the top
- The tab is super small impossible to see

Also fixes the fact that when closing the current tab, we switch to the first tab instead of the direct next tab on the list.

Also, because of a change in the component we use for resizable panels, make the "size" property of panels required.

I believe it fixes this issue as well
https://github.com/kestra-io/kestra/issues/11035


Piyush-r-bhaskar (approved):
LGTM 🚀

PR #11094, opened 2025-09-04
fix(flows): avoid updating code if no-code has not changed

elevatebart:
fixes the following bug

- create a new flow (do not use no-code)
- add a bunch of comments in it to explain what is happening
- open no-code
- without changing anything in no-code, the comments are gone

https://github.com/user-attachments/assets/d5a561d4-5dfc-4249-8edf-d0bffc3ad803


Piyush-r-bhaskar (approved):
LGTM 🚀

PR #11103, opened 2025-09-04
build(main): add skip-test as an option for workflows

fdelbrayelle:

---

---


fdelbrayelle:
Cannot be done after all because the release needs tests :sweat:

PR #11116, opened 2025-09-05
fix(topology): prevent flowTopology failing on a Flow for bugged rela…

AcevedoR:
…tions


fhussonnois (approved):

loicmathieu (approved):

loicmathieu:
When you log a flow better to log both the ns and the flo id

Issue #11125, opened 2025-09-08
Avoid blinking on flow dependencies


tchiotludo:
We should wait loading before displaying no flow or flow dependencies 

https://github.com/user-attachments/assets/ff009f23-47d6-4cff-95d0-d2948c2c121c

This behavior could be applied on all pages that have a no result : concurrency, trigger, apps, etc ... (less visible since no network load, but still visible)

akshatggupta:
i'm working on this issue

MilosPaunovic:
Alright, we'll wait for a PR! 🚀

shreyash1231:
Hello @MilosPaunovic Can you assign me the issue

tchiotludo:
@shreyash1231, it's already merge and fix in #11155. 
Feel free to pick any first good issues 🎉

MilosPaunovic:
Hello, so, the merged fix from https://github.com/kestra-io/kestra/pull/11155 did some related work, but it did not close the issue fully, so I'm re-opening.

If you're still willing to tackle it @shreyash1231, you're more than welcome to.

shreyash1231:
@MilosPaunovic Yes, I am open to work on this issue

MilosPaunovic:
Hey @shreyash1231, are there any updates on this?

shreyash1231:
hey @MilosPaunovic  I have raise the pr please check it.

MilosPaunovic:
Hi @shreyash1231,

As there hasn't been any updates on this for a while, I closed the PR. However, if you do open another one which will address all the comments from the first one, it will be reviewed, and you'll be reassigned if no one else handles it in the meantime.

Thanks for all the efforts so far! 🍀

prayagtushar:
I am free to take upon this issue @MilosPaunovic

MilosPaunovic:
Absolutely, go for it @Prayag-09! 🚀

MilosPaunovic:
Very good work, thanks so much @Prayag-09! 🚀

prayagtushar:
Thank you for the opportunity, @MilosPaunovic

PR #11155, opened 2025-09-08
fixed blinking issue

akshatggupta:
Relates to https://github.com/kestra-io/kestra/issues/11125.

### Fix blinking on flow dependencies (#11125)

**Description**

The Dependencies component was showing the empty state (<Empty>) immediately when mounted, before the API call to load dependencies completed. This caused a visual "blink" or flash where users would see "There are currently no dependencies" for a brief moment, then see loading, then see the actual content

I think the root cause was code given below
`<Empty v-if="!loading && !getElements().length" :type="`dependencies.${SUBTYPE}`" />
<el-splitter v-else class="dependencies">
    <div v-loading="loading" ref="container" />`
    
**Whats my approach here?**

I made some changes in file kestra/ui/src/components/dependencies/Dependencies.vue

**Style Changes:**

Added: .loading-state and .loading-content CSS classes for consistent loading appearance

Removed v-loading="loading" from graph container, moved to main loading state

Here is an video demonstration of my approach

https://github.com/user-attachments/assets/fedf826d-6759-40ff-afd8-b923e68ffc41


MilosPaunovic:
Just a small note for the future, if you could use **GitHub Keywords** when opening PRs to connect it to the proper issue (to automate the entire process a bit more), it would be super helpful to us. You can read a bit more about it [here](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) (I've added closing keyword for this particular issue).

akshatggupta:
@MilosPaunovic Apology for not mentioning the right keywords i will make sure this wont happen again in future PRs

MilosPaunovic (approved):

MilosPaunovic:
Hey @akshatggupta,

I've made some adjustments to your approach so that we now distinguish between data loading and graph node rendering. However, the underlying issue wasn't fully resolved by your initial change and needs to be addressed differently.

That said, your work was a big help in getting us to this point, so thank you for all your efforts! 🚀

PR #11161, opened 2025-09-09
chore(deps): upgrade to Jackson 2.20.0

loicmathieu:
Jackson annotation now uses a version scheme without micro version so it has been updated to 2.20.

Closes #11069


nkwiatkowski (approved):
LGTM

PR #11166, opened 2025-09-09
fix(flows): avoid failing flow dependencies with dynamic defaults

### LINKED ISSUE #11117: Fix Flow Dependencies computation for flows with dynamic defaults
### Describe the issue

As of 1.0, due to a bug, any flow with dynamic defaults, such as the one below, will be ignored by the Dependencies view:

```yaml
id: infrastructure_workflow
namespace: anna

inputs:
  - id: release_env
    displayName: Release Environment
    type: SELECT
    autoSelectFirst: true
    values:
      - DEV
      - TEST
      - QA
      - PROD

  - id: backup_enabled
    type: BOOL
    defaults: "{{ inputs.release_env == 'PROD' ? true : false }}"
    dependsOn:
      inputs:
        - release_env

tasks:
  - id: log
    message: dynamic default {{ inputs.backup_enabled }}
```

We should find a solution so that Flow dependencies from triggers are computed correctly - example from the original issue https://github.com/kestra-io/kestra/issues/11096:

```yaml
id: flow1
namespace: anna

tasks:
  - id: hello
    message: Hello World! 🚀
```

another flow:

```yaml
id: flow2
namespace: anna

tasks:
  - id: hello
    message: Hello World! 🚀

triggers:
  - id: release
    states:
      - SUCCESS
    preconditions:
      id: flows
      flows:
        - namespace: anna
          flowId: flow1
```

### Environment

- Kestra Version: develop

brian-mulier-p:
closes #11117

The main issue is that input defaults are trying to get resolved but if themselves depend on another inputs it's getting painful.

A better way may be to do the same as for resolve input endpoint or resolve other input defaults while keeping track of failed input resolutions and create a new renderer with the resolved input values and keep iterating till the failure list is no longer reducing or empty but it feels overkill and I don't think handling this case is useful as it's failing only in a "static" context (flow dependencies computing)

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/RunVariables.java
+++ b/core/src/main/java/io/kestra/core/runners/RunVariables.java
@@ -10,6 +10,7 @@ import io.kestra.core.models.flows.FlowInterface;
 import io.kestra.core.models.flows.Input;
 import io.kestra.core.models.flows.State;
 import io.kestra.core.models.flows.input.SecretInput;
+import io.kestra.core.models.property.Property;
 import io.kestra.core.models.property.PropertyContext;
 import io.kestra.core.models.tasks.Task;
 import io.kestra.core.models.triggers.AbstractTrigger;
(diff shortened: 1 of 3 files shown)


loicmathieu (commented):
I'm not sure I understand it, you said that it's normal to not be able to resolve inputs?
This is unfortunate as it would means that we would allow nonn-working flow to be defined...

anna-geller (approved):

PR #11180, opened 2025-09-10
fix(core): exclude current namespace in listKeysWithInheritance

### LINKED ISSUE #10969: Make inherited KVs endpoint return only inherited items
New endpoint `GET /api/v1/{tenant}/namespaces/{namespace}/kv/inheritance` is returning not only inherited `KV` pairs but the ones for the current namespace.

Example - namespace `company` has one `KV` pair, namespace `company.team` has one `KV` pair. The endpoint for the `company.team` will return both of the pairs.

This is a bug, it shouldn't happen - inheritance means give me inherited keys (keys from my parents and grandparents), not the keys I already own, so that should be amended.

Follow up of https://github.com/kestra-io/kestra-ee/issues/2830#issuecomment-3232864277.
Related to https://github.com/kestra-io/kestra-ee/issues/4353.

MTarek165:
closes #10969

description:
- Returns only KVs for ancestor namespaces
- Verified at KVControllerTest
- Handles single-level namespace edge case


MTarek165:
Hi @nkwiatkowski , could you review this PR if you have a moment ?

MTarek165:
Hi @loicmathieu , this PR was opened from a while and wasn't reviewed yet. 
It seems to be missed so could you give a look please?

webserver/src/main/java/io/kestra/webserver/controllers/api/KVController.java
```diff
@@ -51,12 +45,18 @@ public List<KVEntry> listKeys(
 
     @ExecuteOn(TaskExecutors.IO)
     @Get("/inheritance")
-    @Operation(tags = {"KV"}, summary = "List all keys for a namespace and parent namespaces")
-    public List<KVEntry> listKeysWithInheritence(
+    @Operation(tags = {"KV"}, summary = "List all keys for ancestor namespaces")
```
loicmathieu:
```suggestion
    @Operation(tags = {"KV"}, summary = "List all keys for inherited namespaces")
```

webserver/src/main/java/io/kestra/webserver/controllers/api/KVController.java
```diff
@@ -51,12 +45,18 @@ public List<KVEntry> listKeys(
...
     ) throws IOException {
-        List<String> namespaces = NamespaceUtils.asTree(namespace);
-        return getKvEntriesWithInheritance(namespaces);
+        int lastDotIndex = namespace.lastIndexOf('.');
+        if (lastDotIndex == -1) {
+            return Collections.emptyList();
+        }
+        String result = namespace.substring(0, lastDotIndex);
+
+        List<String> ancestorNamespaces = NamespaceUtils.asTree(result);
```
loicmathieu: This seems overly complex.
We usually filter the list of namespaces from the `asTree` methods to exclude the given namespace.

Something like that seems easier to read:
```java
List<String> namespaces = NamespaceUtils.asTree(namespace).stream()
    .filter(ns -> !ns.equals(namespace))
    .toList();
```

loicmathieu (changes requested):

webserver/src/main/java/io/kestra/webserver/controllers/api/KVController.java
MTarek165: Thanks for the note, it is updated now.

loicmathieu (approved):
LGTM, thanks

PR #11201, opened 2025-09-10
fix(core): make CRC32 for plugin JARs lazy

fhussonnois:
Make CRC32 calculation for lazy plugin JAR files
to avoid excessive startup time and performance impact.

Avoid byte buffer reallocation while computing CRC32.


loicmathieu (approved):
LGTM

loicmathieu:
Please backport it to 1.0

PR #11202, opened 2025-09-10
fixsystem): always load netty from the app classloader

loicmathieu:
As Netty is used in core and a lot of plugins, and we already load project reactor from the app classloader that depends in Netty.

Fixes https://github.com/kestra-io/kestra-ee/issues/5038


fhussonnois (approved):
LGTM!

PR #11215, opened 2025-09-10
fix(core): fix CrudEvent model for DELETE operation

fhussonnois:
Refactor XxxRepository class to use new factory methods from the CrudEvent class

Related-to: kestra-io/kestra-ee#4727


Skraye (approved):
LGTM

PR #11217, opened 2025-09-10
fix(ui): Fix vsm active tab color flickering on hover

### LINKED ISSUE #11156: [UI] Sidebar tab color is not consistent when you move away from it
The sidebar tab color is not consistent when you move away from it _(in light mode)_

https://github.com/user-attachments/assets/a2c64284-9360-4c8e-9244-8c07b22fc584

Bisxsh:

`!important` is added onto the active tab's background colour to prevent it being overridden by the normal tab's hover style. 

As the hover selector for the active and regular tabs have the same specificity, hovering over the padded deadzone between the tab and the edge of the menu bar caused the regular hover style to overwrite the active tab's one. In light mode, this caused the text to appear invisible due to a lack of contrast.

It appears as if the dynamic class names and hover zones are managed by `vue-sidebar-menu` so styling was changed instead for the fix.

---

closes #11156


MilosPaunovic (approved):
Loving this - simple yet impactful! Great catch, @Bisxsh 🚀

PR #11226, opened 2025-09-11
feat(ui): added reset functionality for settings

### LINKED ISSUE #10640: Add reset theme preferences option
### Feature description

Would be a nice addition to have an option to reset preferences under theme preferences. Specifically to reset the font sizes and family back to the original.

Image

karthik7129:
closes  #10640

What changes are being made and why?
This PR fixes the reset to default functionality for theme preferences in the BasicSettings component.

Changes Made:
The reset actions (restoreDefaultConfigurations and restoreDefaultPreferences) now automatically persist default values to localStorage by invoking saveAllSettings. Overall, this change improves UX by ensuring that reset actions are immediate and user-friendly.


MilosPaunovic (changes requested):
Hey @17297781Karthik,

Good work so far, but before merging, I'd love to see some changes introduced into this PR:

1. If we're doing the reset for the first two blocks, we can add a reset button for the third one, also.
2. I'm not a big fan of the button placement now. Do you think you could create a small button with this icon https://pictogrammers.com/library/mdi/icon/reload/
on the right side of the section title, with the tooltip that would say some thing like "Reset to defaults"?

image

It would require to change `Block` componenet a bit to allow custom slot, and then pass it from `basicSettings` one

karthik7129:
@MilosPaunovic I will make the mentioned changes and commit shortly...Thank you for suggesting 👍

karthik7129:
@MilosPaunovic , I have extended the functionality to 3rd block as well and refactored button style and position...Please have a look and let me know if further refactoring is needed.
image

ui/src/components/settings/BasicSettings.vue
```diff
@@ -538,14 +663,10 @@
                             localStorage.removeItem(key)
                         break
                     case "envName":
-                        if (this.pendingSettings[key] !== this.miscStore.configs?.environment?.name) {
```
MilosPaunovic: Hey @17297781Karthik, can you help me understand why this `if` statement was removed, along with the one just below? Was it by mistake or was there a reason for it?

MilosPaunovic (changes requested):
All in all, very good work! 🚀 

I've pushed couple of changes which will align the end result more with our product, but you've done the foundation if it very good.

Just one small question to clarify before merging.

ui/src/components/settings/BasicSettings.vue
karthik7129: Actually the envColor and envName were behaving inconsistently  while resetting..But when I removed `If` statement they were consistent.
So I removed them @MilosPaunovic .

MilosPaunovic: Okay, I've reverted those changes to previous state as they're not connected with changes for solving this issue.

MilosPaunovic (approved):

karthik7129:
Thank you so much, for making minor tweaks all alone @MilosPaunovic .

PR #11233, opened 2025-09-11
fix(core): add ability to remap sort keys

brian-mulier-p:
part of kestra-io/kestra-ee#5075


Piyush-r-bhaskar (approved):
Looks good 🚀

Could you check it would be fine if we made the change into `ui/src/composables/useDataTableActions.ts` as well ?

https://github.com/kestra-io/kestra/blob/develop/ui/src/composables/useDataTableActions.ts#L50-L67

brian-mulier-p:
> Looks good 🚀
> 
> Could you check it would be fine if we made the change into `ui/src/composables/useDataTableActions.ts` as well ?
> 
> https://github.com/kestra-io/kestra/blob/develop/ui/src/composables/useDataTableActions.ts#L50-L67

Nice catch, will add it

PR #11254, opened 2025-09-12
feat(LogLine): enhance clipboard visibility on hover

### LINKED ISSUE #11220: Logs: Display copy buttons only on row hover
Currently, the copy buttons are visible on every log line, which makes the `Logs` view look cluttered. This change would show the buttons only on row hover, keeping the interface cleaner.

Image

gantoin:
Closes https://github.com/kestra-io/kestra/issues/11220


MilosPaunovic (approved):
Simple, clean - love it! 🚀

PR #11260, opened 2025-09-12
feat(plugins) allow disabling version support for specific tasks

fhussonnois:
Changes:
* Add param 'type' PluginRegistry isVersioningSupported method
* Remove regex validation on 'version' property

Related-to: kestra-io/kestra-ee#5090


loicmathieu (approved):
LGTM

PR #11276, opened 2025-09-15
feat(flows): save editor panel layout after creation

### LINKED ISSUE #9887: Editor forgets panel layout when saving flow for the first time
### Describe the issue

When you create a flow, any changes made to the panel layout beyond the default (flow code on the left, documentation on the right) will be overwritten with the default when you press save. This only happens if the flow has never been saved before.

https://github.com/user-attachments/assets/d0bbdaa9-0ac7-42a2-8077-beff790b898c

### Environment

- Kestra Version: `develop`

### LINKED ISSUE #10550: Changing default tabs on flow creation
Right now, when we open `Flow` creation page, we're greeted with a screen like this:

Image

It always has two default tabs open: `CODE` and `DOCUMENTATION`.

When we edit the flow, we can change the order and visibility of the tabs as we like, and these preferences are saved in local storage on a per-flow basis. This way, when we open that same flow again, we see the layout we had before.

We should allow customizing the default tabs for flow creation - either through the `Settings` page (which might be a more complex approach), or, more simply, by adding another option below the `Actions` button (next to the `Save` one). This option could be labeled `Save this layout for flow creation` or something similar, and when selected, it would save the chosen layout as the default for all new flows. With it, we could also add the `Reset to default layout` button.

### LINKED ISSUE #11934: When using the "No-Code" Editor, after "Save" it closes the panel
### Describe the issue

Steps to reproduce:

1. Create a new Flow
2. Open the "No-Code" Editor panel
3. Click "Save"
4. **Page reloads but the "No-Code" panel is closed**
5. When creating a new Flow again, the "No-Code" panel is now open

https://github.com/user-attachments/assets/395277a8-0c95-42a5-83dd-390622bb5657

### Environment

- Kestra Version:`v1.0.4`

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/9887.
Closes https://github.com/kestra-io/kestra/issues/10550.
Closes https://github.com/kestra-io/kestra/issues/11934.


elevatebart (approved):
Thank you much !

PR #11291, opened 2025-09-15
fix(executions): truncate the execution_running table as in 0.24 ther…

loicmathieu:
…e was an issue in the purge

This table contains executions for flows that have a concurrency that are currently running. It has been added in 0.24 but in that release there was a bug that may prevent some records to being correctly removed from this table. To fix that, we truncate it once.


AcevedoR (approved):
Change is LGTM.
It seems it will work for all cases, I assume this table should always logically be empty when a flyway migration will be run

loicmathieu:
> It seems it will work for all cases, I assume this table should always logically be empty when a flyway migration will be run

No as you can restart Kestra while executions are running, running task would be resubmitted.
But, as there can be stale exec in it, we must truncate the table at least once (this will be done onces as flyway only run not-already-runned migration).

PR #11311, opened 2025-09-16
fix(core): fix plugin stable version resolution (kestra-io/kestra-ee#5129)

fhussonnois:
Fixes: kestra-io/kestra-ee#5129


core/src/main/java/io/kestra/core/utils/Version.java
```diff
@@ -155,8 +159,8 @@ public static Version getOldest(final Collection<Version> versions) {
     }
 
     private final int majorVersion;
-    private final int minorVersion;
-    private final int incrementalVersion;
+    private final Integer minorVersion;
+    private final Integer incrementalVersion;
```
fdelbrayelle: Maybe rename it `patchVersion` like the official [semver](https://semver.org/) name?

fdelbrayelle (approved):

PR #11312, opened 2025-09-16
ci(plugins): do not release all plugins with Kestra core

fdelbrayelle:

- Decoupled plugin releases from Kestra core (independent lifecycle)
- Updated `release-plugins.sh`: added `detect_bump_type()` (major/minor/patch from commits) and computed `RELEASE_VERSION` / `NEXT_VERSION` for each plugin based on its last tag
- New `changed-plugins-list` to get only plugin repos with changes since last release
- Adjusted workflow: removed `releaseVersion` and `nextVersion` (auto-calculated)

**:warning: Remaining work:**

- Removing the `.plugins` usage and use the API instead of the already indexed plugins
- For new plugins repositories, find a way to list and index them

part-of https://github.com/kestra-io/kestra-ee/issues/5188

---

---

### DIFF
--- a/dev-tools/release-plugins.sh
+++ b/dev-tools/release-plugins.sh
@@ -3,104 +3,47 @@
 # SCRIPT: release-plugins.sh
 #
 # DESCRIPTION:
-#   This script can be used to run a ./gradlew release command on each kestra plugin repository.
-#   By default, if no `GITHUB_PAT` environment variable exist, the script will attempt to clone GitHub repositories using SSH_KEY.
-#
-# USAGE: ./release-plugins.sh [options]
-# OPTIONS:
-#   --release-version <version>  Specify the release version (required)
-#   --next-version    <version>  Specify the next version (required)
-#   --dry-run                    Specify to run in DRY_RUN.
-#   -y, --yes                    Automatically confirm prompts (non-interactive).
-#   -h, --help                   Show the help message and exit
-
-# EXAMPLES:
-# To release all plugins:
-#   ./release-plugins.sh --release-version=0.20.0 --next-version=0.21.0-SNAPSHOT
-# To release a specific plugin:
-#   ./release-plugins.sh --release-version=0.20.0 --next-version=0.21.0-SNAPSHOT plugin-kubernetes
-# To release specific plugins from file:
-#   ./release-plugins.sh --release-version=0.20.0 --plugin-file .plugins
+#   This script releases modified Kestra plugins using the Gradle release plugin.
+#   It supports version bumping, tag creation, and release branch management.
+#   Supports auto-detecting bump type (major, minor, patch) per plugin.
 #===============================================================================
 
-set -e;
+set -e
 
-###############################################################
-# Global vars
-###############################################################
 BASEDIR=$(dirname "$(readlink -f $0)")
-WORKING_DIR=/tmp/kestra-release-plugins-$(date +%s);
-PLUGIN_FILE="$BASEDIR/../.plugins"
-GIT_BRANCH=master
+WORKING_DIR=/tmp/kestra-release-plugins-$(date +%s)
+GIT_BRANCH=main
 
-###############################################################
-# Functions
-###############################################################
-
-# Function to display the help message
-usage() {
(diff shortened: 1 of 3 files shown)


dev-tools/release-plugins.sh
```diff
@@ -3,237 +3,150 @@
...
 PLUGINS_ARGS=()
 AUTO_YES=false
 DRY_RUN=false
-# Get the options
+RELEASE_VERSION=""
+NEXT_VERSION=""
+
 while [[ "$#" -gt 0 ]]; do
     case "$1" in
-        --release-version)
```
fhussonnois, line 70: In general, avoid this type of reformatting, as it introduces too much noise into the PR.

dev-tools/release-plugins.sh
```diff
@@ -3,237 +3,150 @@
...
+    case "$BUMP_TYPE" in
+      major) ((MAJOR++)); MINOR=0; PATCH=0 ;;
+      minor) ((MINOR++)); PATCH=0 ;;
+      patch) ((PATCH++)) ;;
+    esac
+    PLUGIN_RELEASE_VERSION="$MAJOR.$MINOR.$PATCH"
+    PLUGIN_NEXT_VERSION="$MAJOR.$((MINOR + 1)).$PATCH-SNAPSHOT"
   else
-    echo "Clone git repository using GITHUB PAT"
-    git clone https://${GITHUB_PAT}@github.com/kestra-io/$PLUGIN.git
```
fhussonnois, line 173: The CI/CD does use the `GITHUB_PAT`. The script  should be runnable both on developer laptop (using SSH key) and on the CI/CD (using PAT)

fhussonnois (changes requested):
The `release-plugins.sh` script is only used during a Kestra release. Therefore, it should not create new patch versions for plugins, since patch versions are managed incrementally over time. Otherwise, the script might attempt to create a patch version that already exists. So in case of a patch version we should do nothing - the plugin is probably already released with that fix.

AcevedoR:
maybe we could schedule a sync on this @fhussonnois and @fdelbrayelle to be sure we all have the same understanding of the issue

fdelbrayelle:
> The `release-plugins.sh` script is only used during a Kestra release. Therefore, it should not create new patch versions for plugins, since patch versions are managed incrementally over time. Otherwise, the script might attempt to create a patch version that already exists. So in case of a patch version we should do nothing - the plugin is probably already released with that fix.

Yes, I'll remove the patch part then.

After this PR, the script would be used for 2 reasons:
- during a Kestra release (like the Tuesday one)
- to explicitely / manually release a plugin major or minor version

fdelbrayelle:
> maybe we could schedule a sync on this @fhussonnois and @fdelbrayelle to be sure we all have the same understanding of the issue

Report from this meeting [here](https://kestra-io.slack.com/archives/C04HTFM3VL6/p1758123890820739)

.github/actions/plugins-list/action.yml
```diff
@@ -1,5 +1,5 @@
 name: 'Load Kestra Plugin List'
```
AcevedoR: you need to keep the current actions as it is, and instead create a new one with your logic to only find plugins that needs to be released :)

AcevedoR (changes requested):

.github/actions/plugins-list/action.yml
fdelbrayelle: I added a `changed-only` input to filter but indeed I badly reported the existing code in the else branch 😅

AcevedoR: could you as well rename the workflow name with something like: release-recently-changed-plugins or something like this. 
Additionaly, we could put this workflow in the actions repos, WDYT @fhussonnois and @brian-mulier-p ?

fdelbrayelle: With not rename it yes.
For the move out to the actions repository, in another PR with the other remaining action maybe?

fhussonnois (approved):
LGTM!

fdelbrayelle:
This workflow won't be used again so I close the PR

PR #11355, opened 2025-09-17
fix(executions): possible NPE on dynamic taskrun

loicmathieu:
Fixes https://github.com/kestra-io/kestra-ee/issues/5166

### DIFF
--- a/executor/src/main/java/io/kestra/executor/ExecutorService.java
+++ b/executor/src/main/java/io/kestra/executor/ExecutorService.java
@@ -1171,7 +1171,7 @@ public class ExecutorService {
             }
         }
 
-        return taskRuns.size() > execution.getTaskRunList().size() ? execution.withTaskRunList(taskRuns) : null;
+        return taskRuns.size() > ListUtils.emptyOnNull(execution.getTaskRunList()).size() ? execution.withTaskRunList(taskRuns) : null;
     }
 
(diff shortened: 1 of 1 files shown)


executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -1171,7 +1171,7 @@ private Execution addDynamicTaskRun(Execution execution, Supplier<Flow> flow, Wo
             }
         }
 
-        return taskRuns.size() > execution.getTaskRunList().size() ? execution.withTaskRunList(taskRuns) : null;
+        return taskRuns.size() > ListUtils.emptyOnNull(execution.getTaskRunList()).size() ? execution.withTaskRunList(taskRuns) : null;
```
Skraye, line 1174:
```suggestion
var taskRunList = ListUtils.emptyOnNull(execution.getTaskRunList());
        return taskRuns.size() > taskRunList.size() ? taskRunList : null;
```

If taskRuns is empty, and `execution.getTaskRunList()` is null, it will enter the first case and throw NPE
In case taskRuns can be empty

Skraye (changes requested):
maybe a nitpick

Skraye (approved):
nevermind previous comment

LGTM

PR #11359, opened 2025-09-17
fix(executions): the Exit task was not correctly ends parent tasks

loicmathieu:
Fixes https://github.com/kestra-io/kestra-ee/issues/5168


brian-mulier-p (approved):
LGTM

PR #11380, opened 2025-09-18
feat(tests): rework runner utils to not use the queue during testing

nkwiatkowski:
The goal here is to not use the queue anymore to find the execution with runner utils. A new class TestRunnerUtils will be added to fetch and test the execution state from the repository instead of the queue to prevent event to be lost when tests are played in parallel

---

---


core/src/main/java/io/kestra/core/utils/VersionProvider.java
```diff
@@ -35,7 +35,7 @@ public class VersionProvider {
     private Optional<SettingRepositoryInterface> settingRepository; // repositories are not always there on unit tests
 
     @PostConstruct
-    public void start() {
+    public synchronized void start() {
```
loicmathieu: Why are you doing that?
AFAIK postconstruct annotated methods are executed only once by the bean factory to except if you create the bean manually or call the method manually it's not needed

loicmathieu (commented):
You should be able to remove methods from the RunnerUtils that was only used in tests.
Or at least deprecate them if you fear they are used in plugin tests which may be

core/src/main/java/io/kestra/core/utils/VersionProvider.java
nkwiatkowski: The one in the post construct is useless, I'll remove it. I added it on a static method to prevent multiple init in database during tests in parallel

AcevedoR (approved):
LGTM, I would to suggest to either: disable the remaining flaky tests, or try to quickly rewrite them in a simpler way

PR #11393, opened 2025-09-19
fix(core): avoid ClassCastException when doing secret decryption

brian-mulier-p:
closes kestra-io/kestra-ee#5191


nkwiatkowski (approved):
LGTM

loicmathieu:
I would have prefered that we found the offending usage of the type variable to fix it...

PR #11409, opened 2025-09-21
fix(core): fix ForEach plugin taskrun.iteration property 

### LINKED ISSUE #11208: taskrun.iteration does not output the number of the current iteration in a foreach loop
### Describe the issue

```
id: loop
namespace: company.team
tasks:
  -id: foreach
    values: [1, 2]
    tasks:
      -id: log0
        message: "another Task"
      -id: log
        message:
        - "{{ taskrun }}" 
```

Image

Is seams to count the number of tasks executed in the loop.
This is not the expected/documented behaviour. https://kestra.io/plugins/core/flow/io.kestra.plugin.core.flow.foreach

### Environment

- Kestra Version: 0.22.6

MTarek165:
closes #11208

description:
This pr solves the problem of incorrect iteration number displayed for each task shown at ForEach plugin


MTarek165:
hi @Ben8t, @anna-geller. could you review this?

MTarek165:
hi @Ben8t, @anna-geller. 
This bug reported at Kestra Version: 0.22.6
but the behavior still same at the last version , could you review this fix?

MTarek165:
task
the new state

MTarek165:
taskupdate
The new state

MTarek165:
taskupdate
The new state for the same mentioned example.

anna-geller:
Someone from backend will have a look, thanks for contributing!

loicmathieu:
I wonder if the index increment shouldn't be done at the end of the loop and not at the begining.
It would be more natural to start with the index 0 than 1 as we set the initial index value to 0

MTarek165:
Yes, it is more intuitive and seems to be the intended indexing at first regardless of the missing batch increment. should I modify it to be zero indexed?

MTarek165:
Yes, I feel it is more intuitive and seems to be the intended indexing at first regardless of the missing batch increment problem. should I modify it to be zero indexed?

MTarek165:
increment
I was looking in docs or codebase for snippet showing if iteration is zero based or one based 
and It is worth mentioning that in ForEachItem plugin I found this code snippet that refers to the initial iteration always considered 1 , so it would be better if we make it start with 1 to preserve consistency and also it is more user friendly.
so I think now that there is no need to modify the PR.

MTarek165:
increment
   I was looking in docs or codebase for snippet showing if iteration is zero based or one based 
and It is worth mentioning that in ForEachItem plugin I found this code snippet that refers to the initial iteration always considered to be 1 , so it would be better if we make it start with 1 also here to preserve consistency and also it is more user friendly.
so I think now that there is no need to modify the PR, or if we modified we may let index=1 and move the increment to the end of the loop.
so that the final desired behavior
for example :  Tasks=[1, 2]
and Values=[x, y, z] 
taskrun.iteration values for the two tasks over the iterations should be : {1,1 ,2,2, 3,3}

loicmathieu:
Current iteration starts with 0 with ForEach so I would be enclined to start with 0.
Anyway, wether we start with 0 or 1, it is more natural to increase the counter at the end of the loop than the begining.

Please also add a test, for ex this flow clearly show the issue and using an `OutputValues` task you can easily assert on the output of the second task.

```yaml
id: loop
namespace: company.team

tasks:
  - id: foreach
    values: [1, 2]
    tasks:
      - id: first
        values:
          iteration : "{{ taskrun.iteration }}"
      - id: second
        values:
          iteration : "{{ taskrun.iteration }}"
```

Test can be added to the existing [ForEachTest}(https://github.com/kestra-io/kestra/blob/develop/core/src/test/java/io/kestra/plugin/core/flow/ForEachTest.java).

loicmathieu:
Please start iteration at 0 as it was before, I asked internally on what we should align and I'll open a separate issue for it.

MTarek165:
> Please start iteration at 0 as it was before, I asked internally on what we should align and I'll open a separate issue for it.

Done.

MTarek165:
> Current iteration starts with 0 with ForEach so I would be enclined to start with 0. Anyway, wether we start with 0 or 1, it is more natural to increase the counter at the end of the loop than the begining.
> 
> Please also add a test, for ex this flow clearly show the issue and using an `OutputValues` task you can easily assert on the output of the second task.
> 
> ```yaml
> id: loop
> namespace: company.team
> 
> tasks:
>   - id: foreach
>     type: io.kestra.plugin.core.flow.ForEach
>     values: [1, 2]
>     tasks:
>       - id: first
>         type: io.kestra.plugin.core.output.OutputValues
>         values:
>           iteration : "{{ taskrun.iteration }}"
>       - id: second
>         type: io.kestra.plugin.core.output.OutputValues
>         values:
>           iteration : "{{ taskrun.iteration }}"
> ```
> 
> Test can be added to the existing [ForEachTest}(https://github.com/kestra-io/kestra/blob/develop/core/src/test/java/io/kestra/plugin/core/flow/ForEachTest.java).

Done.

loicmathieu (approved):
LGTM, thanks

core/src/test/resources/flows/valids/foreach-iteration.yaml
```diff
@@ -0,0 +1,16 @@
+id: loop
+namespace: company.team
```
loicmathieu:
```suggestion
id: foreach-iteration
namespace: io.kestra.tests
```

PR #11420, opened 2025-09-22
 fix(ui): make Flow -> Triggers refresh button reliably fetch triggers

### LINKED ISSUE #11375: [Flow detail page] Refresh icon doesn't work on the Flow -> Triggers tab
The refresh icon doesn't work on the Triggers tab

Example flow:
```yaml
id: s3_listen
namespace: company.team

tasks:
  - id: each
    values: "{{ trigger.objects | jq('.[].uri') }}"
    tasks:
      - id: return
        format: "{{ taskrun.value }}"

triggers:
  - id: watch
    type: io.kestra.plugin.aws.s3.Trigger
    interval: "PT3S"
    accessKeyId: "{{ secret('AWS_ACCESS_KEY_ID') }}"
    secretKeyId: "{{ secret('AWS_SECRET_KEY_ID') }}"
    region: "eu-central-1"
    bucket: "my-bucket"
    prefix: "sub-dir"
    action: MOVE
    moveTo:
      key: archive
      bucket: "new-bucket"
```

Image

EleftheriaBatsou:
Fixes #11375
This pull request addresses the issue #11375 where the refresh icon did not function properly on the Flow -> Triggers tab.

Changes made include:

Modified the loadData method in FlowTriggers.vue to ensure that if there are no triggers in the flow, any previously loaded data is cleared, preventing errors.
Updated the query size in the API call to match the number of triggers available, ensuring that the displayed data is consistent with the flow's state.
These adjustments ensure that the refresh functionality works as intended and provides correct feedback to the user regarding the available triggers.

---

**Root cause:** 
On the Flow → Triggers tab, the refresh callback called loadData(), which returned early when triggersWithType.length === 0 and also used triggersWithType.length to compute the fetch size. Because triggersWithType is a computed/filtered list, the refresh could be a no-op (no server call) when the list was empty due to filters or timing.

**Fix:** Updated ui/src/components/flows/FlowTriggers.vue so loadData():
Uses the actual number of flow triggers from flowStore.flow.triggers to set the fetch size.
Clears local state when a flow has no triggers.
Always performs a fetch when the flow has at least one trigger.

**How I tested**
Manual:
Created a flow with triggers; navigated to Flow → Triggers.
Clicked Refresh and verified a GET to /api/triggers/{namespace}/{flowId} with size equal to flowStore.flow.triggers.length and that the table updated.
Applied a filter so the table was empty, then clicked Refresh; verified it still fetched and updated.
Toggled trigger state and clicked Refresh to confirm the change was reflected.
Verified that for flows with 0 triggers, no errors were thrown and the local list cleared.


Piyush-r-bhaskar (changes requested):
Hey @EleftheriaBatsou Thanks for this.

But I don't see any effect with this PR..
could you help with how you are testing ?

EleftheriaBatsou:
Hmm, let me think about this again. It might be wrong. Sorry.

PR #11431, opened 2025-09-22
fix(core): Add warning logs for mismatched (Parent-Subflow) inputs 

### LINKED ISSUE #11372: Subflow with wrong number of inputs don't raise any error
### Describe the issue

https://jam.dev/c/2ba4c5fd-9ad3-4f43-8d2f-bdc2099dd301

```yaml
id: parent
namespace: company.team

tasks:
  - id: call_my_subflow
    namespace: company.team
    flowId: child
    inputs:
      input1: value1
      input2: value2
    wait: true
```

```yaml
id: child
namespace: company.team
inputs:
  - id: input1
    type: STRING
tasks:
  - id: hello
    message: "{{ inputs.input1 }}"
```

As shown above, the Parent flow calls a subflow with two inputs. However, the Child flow only defines one input.
Everything still works without errors or logs — the second input is simply ignored.

We should log at least an error in the parent and child executions mentionning that the wrong number of inputs was passed.
Making the execution fails would be a breaking change, so we can first implement a log warning for this

### Environment

- Kestra Version: develop oss

MTarek165:
Closes #11372

### Description
This PR adds warning logs when inputs are passed from a parent workflow but are not declared in the subflow.  
This ensures better visibility for users and improves the overall experience by making it clear which inputs are ignored by the subflow.
subflow

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
+++ b/core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
@@ -164,17 +164,16 @@ public final class ExecutableUtils {
                 Set<String> declaredInputs = flow.getInputs().stream().map(Input::getId).collect(Collectors.toSet());
                 for (var inputKey : inputs.keySet()) {
                     if (!declaredInputs.contains(inputKey)) {
-                        String log = String.format(
-                            "Input %s was provided by parent execution %s for subflow %s.%s but isn't declared at the subflow inputs",
+                        runContext.logger().warn(
+                            "Input {} was provided by parent execution {} for subflow {}.{} but isn't declared at the subflow inputs",
                             inputKey,
                             currentExecution.getId(),
                             currentTask.subflowId().namespace(),
                             currentTask.subflowId().flowId()
                         );
-                        runContext.logger().warn(log);
                     }
                 }
-            List<Label> newLabels = inheritLabels ? new ArrayList<>(filterLabels(currentExecution.getLabels(), flow)) : new ArrayList<>(systemLabels(currentExecution));
+                List<Label> newLabels = inheritLabels ? new ArrayList<>(filterLabels(currentExecution.getLabels(), flow)) : new ArrayList<>(systemLabels(currentExecution));
             if (labels != null) {
                 labels.forEach(throwConsumer(label -> newLabels.add(new Label(runContext.render(label.key()), runContext.render(label.value())))));
             }
--- a/core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
+++ b/core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
@@ -160,19 +160,6 @@ public final class ExecutableUtils {
             if (flow instanceof FlowWithException fwe) {
                 throw new IllegalStateException("Cannot execute an invalid flow: " + fwe.getException());
             }
-
-                Set<String> declaredInputs = flow.getInputs().stream().map(Input::getId).collect(Collectors.toSet());
-                for (var inputKey : inputs.keySet()) {
-                    if (!declaredInputs.contains(inputKey)) {
-                        runContext.logger().warn(
-                            "Input {} was provided by parent execution {} for subflow {}.{} but isn't declared at the subflow inputs",
-                            inputKey,
-                            currentExecution.getId(),
-                            currentTask.subflowId().namespace(),
-                            currentTask.subflowId().flowId()
-                        );
-                    }
-                }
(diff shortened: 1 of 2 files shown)


MTarek165:
@loicmathieu could you review this PR?

core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
```diff
@@ -163,6 +161,19 @@ public static <T extends Task & ExecutableTask<?>> Optional<SubflowExecution<?>>
...
+                for (var inputKey : inputs.keySet()) {
+                    if (!declaredInputs.contains(inputKey)) {
+                        String log = String.format(
+                            "Input %s was provided by parent execution %s for subflow %s.%s but isn't declared at the subflow inputs",
+                            inputKey,
+                            currentExecution.getId(),
+                            currentTask.subflowId().namespace(),
+                            currentTask.subflowId().flowId()
+                        );
+                        runContext.logger().warn(log);
```
loicmathieu: Please use log parameters instead of pre-computing the log string via `String.format()`.
Thanks to that, the final String would only be computed if the log is enabled at this level.

loicmathieu:
It didn't close the issue as we would still have the same issue for flows started from the API.
So please either remove your GitHub keyword usage that set this PR to close https://github.com/kestra-io/kestra/issues/11372 (you can use `Part-of: https://github.com/kestra-io/kestra/issues/11372` instead) or also add the same check inside the API (both for webhook and execution creation).

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
MTarek165: Done, Is it ok now ?

MTarek165:
Should I add same checks  inside the API (both for webhook and execution creation) at this PR ?

loicmathieu:
@MTarek165 it you could do that it would be great.
The API endpoint is from the ExecutionController but I would do it inside the FlowInputOutput class on that line: https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java#L234

This class is used to read inputs and here it resolves the input so you would be able to check if the resolved inputs are part of the defined flow inputs or not.
If it's not the case you must create a runContext (you have the runContextFacrtory and an execution, it should be feasible) and log

MTarek165:
Yes, I wanted to move checks and log to FlowInputOutput yesterday but the barrier was the absence of the RunContext. 
Now I will give a try, thanks for the feedback.

MTarek165:
Hi @loicmathieu I added checks and logging at the required place, now it is centralized and can handle Api executions beside sub-flows (except web-hooks) but there is two important  things:

1- We have two cases one when parent workflow starts child workflows in case of Subflow and ForEachItem cases so we need to show warning at UI and the other one is these that I don't have UI like external calls for Api Execution and Web hook trigger and in all cases system logs shown in console now, but to handle UI I needed to pass a parent run context to log warnings at parent UI level so I created a new variation of flowInputOutput.readExecutionInputs() to pass parent run context apart from the existing one to avoid any breaking changes, otherwise I create new run context and log easily but Is there a better approach?

2- I couldn't find that Web-hooks define inputs or uses the FlowInputOutput as I found that it has trigger.body not the same style of the others and It is not expecting declared inputs like the others do so warning may not that much value here but any way it may requires a special handling.

loicmathieu:
@MTarek165 as I said, you can create a run context only when you found offending inputs using the run context factory.
Creating a run context is costly so we don't want to create on all the time.

At the place in code I pointed to you, you have the flow, the execution as method parameter, and the run context factory as class attribute so you can create a run context using `runContextFactory.of(flow, execution)` and log a warning.

Please also be careful to not include any formatting change to facilitate the review

MTarek165:
@loicmathieu Yes I tried to take care of that by adding check before creating run context once after resolving inputs and even I can reuse the parent run context in case of (parent-child) workflows , I pushed the new changes today.
could you review whenever you have a moment ?

loicmathieu:
@MTarek165 please do as I asked to, don't bother passing the runContext, please revert most of your changes to the FlowInputOutput class and create a runContext from the execution when needed

MTarek165:
@loicmathieu Ok, should I keep also the checks and logs in executableutils.subflowExecution() as in first beside adding in FlowInputOutput as required?

loicmathieu:
@MTarek165 yes keep it

MTarek165:
@loicmathieu I did exactly what was asked, can you give a look now ?

loicmathieu (approved):
LGTM, thanks for your contribution!

PR #11432, opened 2025-09-22
fix(triggers): handle RecoverMissedSchedules on trigger batch update

fhussonnois:
* Fix and clean code in TriggerController
* Remove duplicate code in Trigger class


Skraye (approved):
LGTM

PR #11436, opened 2025-09-22
fix(system): avoid trigger locking after scheduler restart

### LINKED ISSUE #11434: [Scheduler] Trigger being locked due to race on trigger initialization & queue consumption
### Describe the issue

There is a race due to our queue consumption being started before the triggers are properly initialized. As a result, the AbstractScheduler.initializedTriggers can submit a trigger state older than one submitted by the queue consumption (eg. a trigger's executionId reset due to a SUCCESS execution)

### Environment

- Kestra Version: 0.22.17 but probably impacts all follow-up versions

brian-mulier-p:
closes #11434


Skraye (approved):
LGTM

PR #11448, opened 2025-09-23
chore: bundle fonts locally instead of using Google CDN

### LINKED ISSUE #11305: Removing Google Fonts for air-gapped environments
### Feature description

When using Kestra in an environment without internet access the UI responds really slow. Meaning it tries to load Google Fonts in the background and blocks the UI until the request has timed out.

It would be really nice to either ship the fonts with Kestra or fallback to another font if there is no internet access.

best regards

KishoreGhost:

This PR updates the way fonts are handled in Kestra by bundling them locally within the codebase. Previously, fonts were downloaded from the Google CDN during the initial application launch.

Closes: https://github.com/kestra-io/kestra/issues/11305


MilosPaunovic (changes requested):
Hey @KishoreGhost,

This is great initial progress, but if you try to compare the two variants (local and one loaded from Google) it has small differences in sizes. Can you give that a look, as we need a 1-1 change.

KishoreGhost:
Hey @MilosPaunovic, I need some help with this.. I tried many things and came to the conclusion that the locally available font and the font which we get from google CDN differs, they're not the exact same thing (or so i think). Even though the font weight, size, everything is the same the size is varying..

MilosPaunovic:
Have you tried loading different formats of font files locally, like `woff` or `woff2` maybe @KishoreGhost?

KishoreGhost:
I have @MilosPaunovic, but they still didn't seem to work

MilosPaunovic:
Alright, I'll move this PR to `DRAFT` mode, and will have a look when I get the chance.

Thanks for getting it so far! 🚀

MilosPaunovic (approved):
I've changed to file format we use to `woff2` and done some minor tweaks around it, but overall, you've done an awesome foundation for this, so thanks so much @KishoreGhost! 🚀

voydz:
@MilosPaunovic @KishoreGhost Thanks for you hard work on this. But I think the approach might not solve the real issue here. The Fonts are loaded fine in air-gapped now, but the site is still trying to fetch fonts from the google servers thus rendering the UI blank and completely unusable for about 30sec on every request.

A solution could be relying only on local assets or make it configurable.

best regards

MilosPaunovic:
You make a good point here @voydz.

Let me think about the solution for a bit and I'll get back to you.

PR #11451, opened 2025-09-23
fix(ai): avoid moving cursor twice after using AI Copilot

### LINKED ISSUE #11314: Once AI Copilot has been used, removing characters removes an extra character at the same time
### Describe the issue

When editing code after the AI Copilot has modified the flow (even if it already existed), it seems to remove 1 more character than are selected.

A few examples:

https://github.com/user-attachments/assets/fd683da5-5725-4af5-a84c-93e20b66f8ad

https://github.com/user-attachments/assets/e1415db2-0049-41eb-b747-9399ccabbd00

### Environment

- Kestra Version: `develop`

brian-mulier-p:
closes #11314

Essentially using a key will get rid of https://github.com/microsoft/monaco-editor/issues/4370, I believe it's occurring because of Monaco registering the arrow keybind once each time an editor is instantiated and since we were reusing the same editor when adding the AI draft diff it makes the cursor jump twice

### DIFF
--- a/ui/src/components/inputs/Editor.vue
+++ b/ui/src/components/inputs/Editor.vue
@@ -1,6 +1,6 @@
 <template>
     <div class="ks-editor edit-flow-editor">
-        <nav v-if="original === undefined && navbar" class="top-nav">
+        <nav v-if="!isDiff && navbar" class="top-nav">
             <slot name="nav">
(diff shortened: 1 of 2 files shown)


MilosPaunovic (approved):
LGTM!

PR #11468, opened 2025-09-23
chore(executions): always log errors from the executor

loicmathieu:
- Logs errors from the Executor catched execution
- Logs errors from the Scheduler catched execution
- Avoid most places where the warning "unable to change state already..." could occur
- Log using the run context logger flow issues from executable tasks so they appears inside execution logs


nkwiatkowski (approved):
LGTM

PR #11469, opened 2025-09-23
fix(flows): delete flows should work

elevatebart:
after ts refactoring of the flows table, thsi slipped through


Piyush-r-bhaskar (approved):
Looks Good

PR #11470, opened 2025-09-23
fix(triggers): do not resolve recoverMissedSchedule when enabling back a trigger

fhussonnois:
Add some refactoring to allow some methods to be overrided


Skraye (approved):
LGTM

PR #11517, opened 2025-09-24
test: add test on executor

AcevedoR:
This is a POC to try to make tests closer to the tested code in isolation.
The test is about a test case that often fails


loicmathieu (commented):
I think it's too early to do that, you open an internal method to be called outside of the executor for testing purpose :( 

We will soon move to a single executor in 2.0, such testing mechanism would be more easier with a single executor and we should take this into consideration when we design it to facilitate such tests.

PR #11536, opened 2025-09-25
  Fixes KV Store search functionality that was failing when entries had null descriptions.

### LINKED ISSUE #11413: KV Store filtering not working
### Describe the issue

Using the KV Store Key to filter doesn't work.
I have the following entries

Image
Will try to filter using any of these keys.

Image

Results are empty

Reported in Version 0.22
Reproduced in 1.0.0

### Environment

- Kestra Version: 1.0.0

Sandipmandal25:
**Problem:**
fixed #11413 
  The search filter was calling .toLowerCase() on null description values, causing a runtime error that broke filtering
  entirely.

  **Solution:**
  Added null check before accessing description property in the filter logic.

  **Testing:**
  - Can search by key name even when description is null
  - Can search by description when it exists
  - No runtime errors with mixed null/valid descriptions
  - All existing functionality preserved

  One line fix that resolves the "Results are empty" issue shown in the screenshot.


MilosPaunovic (approved):
LGTM! 🚀

PR #11549, opened 2025-09-26
feat(triggers): add inputs property to webhook trigger

fhussonnois:
Add a new `inputs` property to the Webhook trigger, allowing input data to be passed to the triggered flow. If no inputs are defined on the trigger, the flow will not receive any inputs, even if some have default values. This behavior ensures backward compatibility with how the Webhook trigger currently works.

### Use case

Easily define a flow that can be triggered either manually or through a webhook with an expected payload.

### Example

```yaml
id: webhook
namespace: example
inputs:
  - id: first
    type: STRING

  - id: second
    type: STRING
    defaults: "World"

tasks:
  - id: hello
    message: "Hello {{ inputs.first }}, {{ inputs.second }}!"

triggers:
  - id: webhook
    key: my_secret_key
    inputs:
      # if the input failed to be resolved the flow is not triggered
      first: "{{ trigger.body.name }}"
```

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
+++ b/core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
@@ -156,6 +156,13 @@ public class Webhook extends AbstractTrigger implements TriggerOutput<Webhook.Ou
            """
     )
     private Boolean wait = false;
+    
+    
+    @Schema(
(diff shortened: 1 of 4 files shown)


fhussonnois:
FYI: @anna-geller

nkwiatkowski (approved):
LGTM

PR #11552, opened 2025-09-27
  fix(scheduler): prevent race condition between queue consumption and trigger initialization

### LINKED ISSUE #11435: [Scheduler] Find a way for AbstractScheduler.initializedTriggers to be race-safe with queue consumption
### Describe the issue

(see https://github.com/kestra-io/kestra/issues/11434, this is partially solved as there was only a condition that was added but it doesn't cover all cases)

### Environment

- Kestra Version: develop

Sandipmandal25:
Resolves #11435 

## Problem
  There was a race condition where the scheduler's queue consumption could start processing WorkerTriggerResult messages before
  triggers were properly initialized. This could cause trigger state updates from successful executions to be overridden by
  older states from the initialization process, leading to inconsistent trigger states and potential loss of execution tracking.

  ## Root Cause
  - Queue consumption started immediately in the `run()` method
  - Trigger initialization happened asynchronously via `initializedTriggers()`
  - The existing `isReady()` check only protected the scheduled evaluation loop, not the queue consumption

  ## Solution
  Added a readiness check to the `workerTriggerResultQueue.receive()` listener that blocks until trigger initialization is
  complete using `Await.until(this::isReady)`. This ensures:

  - No worker trigger results are processed before triggers are initialized
  - No messages are lost (blocks instead of dropping)
  - Maintains proper initialization order
  - Prevents race condition while preserving data integrity

  ## Testing
  - Compilation verified
  - Follows existing codebase patterns (same approach as in `handle()` method)
  - Aligns with test expectations that already use `scheduler.isReady()` for proper initialization


loicmathieu:
Hi,
Thanks for your contribution.

Even if it seems like a good idea, this class is shared with our EE scheduler, which is based on Kafka, so we must check its implications for our Kafka implementation.

Sandipmandal25:
Thank you for the feedback! @loicmathieu 
You're absolutely right to be concerned about Kafka compatibility.

  I've updated the fix to use a **skip approach** instead of blocking. The previous implementation used `Await.until(this::isReady)`
   which would block the consumer thread—this works for JDBC but would cause issues with Kafka's consumer poll loop (potential lag, rebalancing, and backpressure).

 - **For JDBC**: Messages are reprocessed via transaction rollback after initialization
  - **For Kafka**: Offsets aren't committed and messages are redelivered on next poll
  - **No blocking**: Consumer threads remain responsive
  - **No message loss**: All messages are eventually processed once initialization completes

Sandipmandal25:
@loicmathieu  I realized my "skip" approach has a  issue with JDBC queues - messages would be lost because `updateGroupOffsets()` runs after the consumer callback regardless of early returns.

  The original blocking approach with `Await.until(this::isReady)` is safer as it prevents message loss. However, I understand your concern about Kafka consumer blocking.

  Would you prefer:
  1. Keep the blocking approach (safe, but may impact Kafka)
  2. Implement a queue-specific solution (e.g., buffer messages during initialization)
  3. Another approach you'd recommend?

loicmathieu:
@sandyp025 I'm not comptfortable to skip a trigger result as this would means we would probably not create an execution we should be creating.

I'm afraid this kind of issue is very complex to tackle.

Sandipmandal25:
> @sandyp025 I'm not comptfortable to skip a trigger result as this would means we would probably not create an execution we should be creating.
> 
> I'm afraid this kind of issue is very complex to tackle.

Yeah  right - I missed a critical issue with the skip approach. Messages will be lost because the offset
  gets committed even if we return early. That's worse than the original race condition.

  I see two cleaner options:

  **1. Defer queue consumption**
  Start the `workerTriggerResultQueue.receive()` subscription after `initializedTriggers()` completes. This prevents the race entirely and works for both JDBC and Kafka.

  **2. Buffer messages during init**
  Queue incoming messages in memory until `isReady()`, then process them all. More complex but allows immediate consumption.

  Which direction would you prefer? I can update the PR accordingly.

@loicmathieu

loicmathieu:
1 seems a better idea

PR #11559, opened 2025-09-29
feat(execution): add secure variable renderer factory for debug mode

### LINKED ISSUE #11535: Alternative renderer no longuer used when previewing outputs
### Describe the issue

There is a hook inside the `VariableRenderer` to allow alternative rendering.
Such alternative rendered must be produced by replacing the default bean.

But as debuging execution outputs instanciate its own renderer, this bypass any alternative renderer and deffeat the purpose of alternative rendering on its own.

Related support issue: https://app.usepylon.com/issues?conversationID=cce7ce76-491d-4625-87dd-b7f20a6e8093

### Environment

- Kestra Version: develop

hemanthsavasere:

Introduce SecureVariableRendererFactory to create debug renderer instances that wrap the base renderer while maintaining security by masking sensitive functions. This provides a consistent way to handle variable rendering in debug contexts. Closes #11535

---

- Have written the unit tests to verify the functionality with good coverage
- Docker Builds are working as expected and able to access the UI

---


brian-mulier-p (approved):
LGTM, any reason for this PR to be in draft?

MilosPaunovic:
@hemanthsavasere Is there a reason why this PR is still in `DRAFT` mode?

hemanthsavasere:
Hi @MilosPaunovic  and @brian-mulier-p 
I was trying to verify the functionality by running the execution of a flow with secrets, but since secret management is enterprise only feature was stuck with it

image

MilosPaunovic:
Makes sense, @hemanthsavasere, thanks for clarifying.

@brian-mulier-p When you get a chance, give this another look and if alright, let's merge and close the issue.

loicmathieu (changes requested):
@brian-mulier-p I don't think it would work as in the end the default variable renderer with non-masked function would be used.

I know @fhussonnois also need this so a joint force may be used ;)

For this to work, the application  context should be used as it's the only way to create a bean that would use the correct implementation.
Then a default bean instance would exist, used everywhere.
And in the controller (and the default input resolver) we would at postconstruct create a different bean with masked functions.

Another solution would be to create two beans and use a bean with a different implementation class that mask functions.
In this case it would be a BC so our customer that override the VariableResolver may be aware he also need to override the `MaskedVariableResolver` (name TBD).

@fhussonnois the best would be you take over this for now.

brian-mulier-p:
Yes my bad... It would work but the alternative rendering would bypass the function masking. Essentially we would need to use applicationContext to instantiate a VariableRenderer with the same masked function (createBean) but since we're using it through CDI it will automatically use the right variable renderer except we parameterize it with the masked functions

fhussonnois:
Hey @hemanthsavasere, Thank you very much for your contribution. I will take over your Pull Request, as this issue has become a priority for us and is connected to other ongoing issues.

fhussonnois:
Thanks again for your contribution. I took back your commits part of this PR: https://github.com/kestra-io/kestra/pull/11755

PR #11563, opened 2025-09-29
fix(API): implement the exist for tenant method with no ACL filter

nkwiatkowski:
Fix #https://github.com/kestra-io/kestra-ee/issues/5104

---

---

### DIFF
--- a/core/src/main/java/io/kestra/core/repositories/FlowRepositoryInterface.java
+++ b/core/src/main/java/io/kestra/core/repositories/FlowRepositoryInterface.java
@@ -162,4 +162,6 @@ public interface FlowRepositoryInterface {
     FlowWithSource update(GenericFlow flow, FlowInterface previous) throws ConstraintViolationException;
 
     FlowWithSource delete(FlowInterface flow);
+
+    Boolean existForTenant(String tenantId);
 }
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/kestra/core/repositories/FlowRepositoryInterface.java
```diff
@@ -162,4 +162,6 @@ default List<String> findDistinctNamespace(String tenantId, String prefix) {
     FlowWithSource update(GenericFlow flow, FlowInterface previous) throws ConstraintViolationException;
 
     FlowWithSource delete(FlowInterface flow);
+
+    Boolean existForTenant(String tenantId);
```
loicmathieu:
```suggestion
    Boolean existAnyNoAcl(String tenantId);
```

We always include `NoAcl` in the name of methods that bypasses ACLs

loicmathieu (approved):
LGTM but can you update the name of the method to include `NoAcl` in it before merging?

core/src/main/java/io/kestra/core/repositories/FlowRepositoryInterface.java
nkwiatkowski: make sense, done

PR #11575, opened 2025-09-29
chore(system): move the SkipExecution service to the services package

loicmathieu:
It was there before so it will be easier to backport the change if it moves there.


Skraye (approved):
LGTM

PR #11577, opened 2025-09-29
ci: migrate CI to kestra-io/actions

AcevedoR:


core/src/main/java/io/kestra/core/annotations/Retryable.java
```diff
@@ -12,7 +12,7 @@
 import jakarta.validation.constraints.Digits;
 
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
-
+// FIXME remove trigger commit
```
AcevedoR:
```suggestion

```
undo this, only to showcase the PR CI works

ui/src/stores/ai.ts
```diff
@@ -1,7 +1,7 @@
 import axios from "axios";
 import {defineStore} from "pinia";
 import {apiUrl} from "override/utils/route";
-
+// FIXME remove trigger commit
```
AcevedoR:
```suggestion

```
undo this, only to showcase the PR CI works

AcevedoR (commented):
I will obviously remove these lines before merging

.github/workflows/e2e-scheduling.yml
```diff
@@ -0,0 +1,15 @@
+name: 'E2E tests scheduling'
+# 'New E2E tests implementation started by Roman. Based on playwright in npm UI project, tests Kestra OSS develop docker image. These tests are written from zero, lets make them unflaky from the start!.'
+on:
+  schedule:
+    - cron: "0 * * * *"  # Every hour
```
nkwiatkowski, line 5: if it fails, we will be spam. Once per day is not enough?

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,49 @@
...
+    name: Frontend tests
+    uses: kestra-io/actions/.github/workflows/kestra-oss-frontend-tests.yml@main
+    secrets:
+      GITHUB_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
+  publish-maven:
+    name: Publish Maven
+    needs: [ backend-tests, frontend-tests ]
+    if: "!failure() && !cancelled()"
```
nkwiatkowski, line 33: No checks on the branch? this may be use to release a feature branch

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,49 @@
...
+      SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
+      SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+      SONATYPE_GPG_KEYID: ${{ secrets.SONATYPE_GPG_KEYID }}
+      SONATYPE_GPG_PASSWORD: ${{ secrets.SONATYPE_GPG_PASSWORD }}
+      SONATYPE_GPG_FILE: ${{ secrets.SONATYPE_GPG_FILE }}
+
+  publish-github:
+    name: Github Release
+    needs: [build-artifacts, backend-tests, frontend-tests]
+    if: "!failure() && !cancelled()"
```
nkwiatkowski, line 45: Same as above

.github/workflows/release-docker.yml
```diff
@@ -0,0 +1,34 @@
...
+        description: 'Plugin version'
+        required: false
+        type: string
+        default: "LATEST"
+
+jobs:
+  publish-docker:
+    name: Publish Docker
+    if: startsWith(github.ref, 'refs/tags/v')
+    uses: kestra-io/actions/.github/workflows/kestra-oss-publish-docker.yml@main
```
nkwiatkowski, line 26: The name suggest that publishing OSS docker is not the same as publishing EE docker, is it the case?

.github/workflows/e2e-scheduling.yml
AcevedoR: it's already like that since a few month, I would prefer to keep it like this, or every 2h or something, but it is really fast to run anyway so not much compute spend

.github/workflows/pre-release.yml
AcevedoR: indeed

nkwiatkowski (approved):
LGTM

.github/workflows/release-docker.yml
AcevedoR: yes there is secrets differences, code drift and specific behaviours on EE, we will merged them if we can in a second time/cooldown I think

PR #11579, opened 2025-09-29
feat(tests): introduce JUnit test support for triggers

### LINKED ISSUE #11087: Add Java unit test support for triggers
### Feature description

You can easily create an execution in  a Java unit test via the `@ExecuteFlow` annotation but we didn't have such mehanism to create an execution from a trigger.

Idealy, we should be able to create an executiuon from a trigger with a single annotation, this would greatly facilitate trigger testing as today the scheduler and a worker are started manually for that.

MTarek165:
close #11087
### Description

- Added a new annotation: `@EvaluateTrigger` and the corresponding `TriggerEvaluationExtension` to support trigger evaluation in unit tests.
- Supports **Polling Triggers**.
- Added a unit test to verify the functionality of the new annotation.

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
+++ b/core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
@@ -160,7 +160,8 @@ public final class ExecutableUtils {
             if (flow instanceof FlowWithException fwe) {
                 throw new IllegalStateException("Cannot execute an invalid flow: " + fwe.getException());
             }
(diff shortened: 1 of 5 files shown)


MTarek165:
@loicmathieu could you review this PR ?

loicmathieu (approved):
LGTM, nice addition, thanks a lot.

@nkwiatkowski you may want to review it also.

I think a future enhancement may be to optionally execute the flow but only evaluation is also a good idea as you would just test the trigger thanks to that (not the full flow which is usually not needed for a trigger as you could assert on the execution trigger).

MTarek165:
Yes, Extending it with optional execution of the flow is great and flexible addition. Thanks for these nice proposals too.

nkwiatkowski:
LGTM too, thanks for the new test annotation.

PR #11592, opened 2025-09-30
fix(executions): killing queued exec. didn't respect concurrency limit

### LINKED ISSUE #11574: [Execution] QUEUED execution that are KILLED are being retried
### Describe the issue

[From slack channel](https://kestra-io.slack.com/archives/C03FQKXRK3K/p1758930070371799) : 
> I just had to kill a bunch of queued flows and I noticed Kestra retrying all of them... According to the docs KILLED: This terminal state indicates that the execution has been killed upon request by the user. No more tasks will be able to run, and the execution is considered terminated. so I was assuming that killing a QUEUED execution would simply move it to the KILLED terminal state and disregard any retries.

It seems that Queued execution aren't killed correctly

loicmathieu:
There was two issues here:
- When killing a queued execution, the associated ExecutionQueued record was not deleted
- When terminating a killed execution that has concurrency limit, we poped an execution even if the execution was not running (no associated ExecutionRunning record) which may exceed concurrency limit

Fixes #11574

I also fix the TestRunnerUtils that should test the predicate before returning the last execution not after.

@nkwiatkowski you may want to have a look at my changes on the `TestRunnerUtils`.

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/ExecutionRunning.java
+++ b/core/src/main/java/io/kestra/core/runners/ExecutionRunning.java
@@ -32,5 +32,7 @@ public class ExecutionRunning implements HasUID {
         return IdUtils.fromPartsAndSeparator('|', this.tenantId, this.namespace, this.flowId, this.execution.getId());
     }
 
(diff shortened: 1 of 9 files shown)


Skraye (approved):
LGTM

PR #11600, opened 2025-09-30
Try reenabling should get report test

AcevedoR:


AcevedoR:
need to wait for OSS new CI to be merge so it will be run in that pr

PR #11605, opened 2025-09-30
fix(core): adjust date property of PublicHoliday condition

### LINKED ISSUE #11586: `io.kestra.plugin.core.condition.PublicHoliday` is not working as intended
### Describe the issue

We were running the below flow config on sanity checks. Instead of getting triggered "11 AM (UTC) on Public Holiday in France", it's running every day and failing sometimes too.

```yaml
id: schedule-condition-publicholiday
namespace: company.team

tasks:
  - id: hello
    message: Demo for PublicHoliday condition 

triggers:
  - id: schedule
    cron: "0 11 * * *"
    conditions:
      - type: io.kestra.plugin.core.condition.PublicHoliday
        country: FR
```

Image

### Environment

- Kestra Version: develop

MTarek165:
Closes: #11586

Description:

Problem:
the issue was caused by not passing the variables map when rendering trigger.date, I found that after investigating some similar condition classes.

Solution:
include  conditionContext.getVariables() in conditionContext.getRunContext().render(date)

### DIFF
--- a/core/src/test/java/io/kestra/plugin/core/condition/PublicHolidayTest.java
+++ b/core/src/test/java/io/kestra/plugin/core/condition/PublicHolidayTest.java
@@ -1,9 +1,11 @@
 package io.kestra.plugin.core.condition;
 
 import com.google.common.collect.ImmutableMap;
+import io.kestra.core.models.conditions.ConditionContext;
 import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.flows.Flow;
 import io.kestra.core.models.property.Property;
+import io.kestra.core.runners.RunContextFactory;
 import io.kestra.core.services.ConditionService;
 import io.kestra.core.utils.TestsUtils;
 import io.kestra.core.junit.annotations.KestraTest;
@@ -11,12 +13,17 @@ import jakarta.inject.Inject;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
+import java.util.Collections;
+import java.util.Map;
+
 import static org.assertj.core.api.Assertions.assertThat;
 
 @KestraTest
 class PublicHolidayTest {
     @Inject
     ConditionService conditionService;
+    @Inject
+    private RunContextFactory runContextFactory;
 
     @Test
     void valid() {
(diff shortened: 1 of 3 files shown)


loicmathieu:
Hi,
I wonder if it is the good fix, as a Schedule trigger should always have a `trigger.date` output.
Polling trigger, however, should not.

I need to check the issue and discuss internally to see if it's the good fix as it's a change in behavior.

MTarek165:
@loicmathieu Is there any updates for this PR and its related issue?

loicmathieu:
@MTarek165 `trigger.date` should always be present so the default on the `PublicHoliday` should be OK.
I wonder what's happeneded here.
first step would be to reproduce the issue via a unit test

MTarek165:
Hi @loicmathieu, the issue was caused by not passing the variables map coming from condition context when rendering trigger.date, I found that after investigating some similar condition classes. We missed using conditionContext.getVariables() in conditionContext.getRunContext().render(date). I've fixed it now — you can have a look.

MTarek165:
Hi @loicmathieu, the true issue was caused by not passing the variables map coming from condition context when rendering date property, I found that after investigating some similar condition classes. We missed using conditionContext.getVariables() (which contains the actual value of trigger.date) in conditionContext.getRunContext().render(date). I've fixed it now, It works fine — you can have a look.

loicmathieu (changes requested):
This looks good but could you add a test that reproduce the issue?

MTarek165:
@loicmathieu I added two unit tests , the existing tests in PublicHolidayTest was testing the condition itself on static date . I needed to test dynamic render itself so I included variables map with the condition context and since the default on date is to be rendered at pebble expression so I don't set in tests when reaching test() method it is rendered successfully. By the way the added valid() method is good if we want to test similar cases for other conditions.

core/src/main/java/io/kestra/core/services/ConditionService.java
```diff
@@ -45,7 +47,18 @@ public boolean isValid(Condition condition, FlowInterface flow, @Nullable Execut
     }
 
     public boolean isValid(Condition condition, FlowInterface flow, @Nullable Execution execution) {
-        return this.isValid(condition, flow, execution, null);
+        return this.isValid(condition, flow, execution, (MultipleConditionStorageInterface) null);
+    }
+    @VisibleForTesting
+    public boolean isValid(Condition condition,FlowInterface flow,@Nullable Execution execution,Map<String,Object>variables){
```
loicmathieu: Please remove this method and any changes to this class.
It's not needed as you can use one of the method taht takes a condition context as a parameter and creates the condition context in the test

core/src/test/java/io/kestra/plugin/core/condition/PublicHolidayTest.java
```diff
@@ -54,6 +56,42 @@ void invalid() {
             .build();
         assertThat(conditionService.isValid(publicHoliday, flow, execution)).isFalse();
     }
+    @Test
```
loicmathieu:
```suggestion

    @Test
```

core/src/test/java/io/kestra/plugin/core/condition/PublicHolidayTest.java
```diff
@@ -54,6 +56,42 @@ void invalid() {
...
+        Execution execution = TestsUtils.mockExecution(flow, variables);
+
+        PublicHoliday publicHoliday = PublicHoliday.builder()
+            .country(Property.ofValue("FR"))
+            .build();
+
+
+        assertThat(conditionService.isValid(publicHoliday, flow, execution, variables)).isTrue();
+    }
+    @Test
```
loicmathieu:
```suggestion

    @Test
```

core/src/main/java/io/kestra/core/services/ConditionService.java
loicmathieu: We try to avoid adding test related code in non-test sources

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/services/ConditionService.java
MTarek165: @loicmathieu ok but this would requere me to inject RunContextFactory in the test class . Is it ok ?

loicmathieu (approved):
LGTM, thanks

PR #11645, opened 2025-10-01
fix(core): handled whitespaces in pebble Expressions, this fixes issu…

### LINKED ISSUE #11411: Condition field does not allow pebble expressions due to whitespace
### Describe the issue

The condition field for all tasks should be automatically trimmed. 

For example, the following task : 

```
 - id: fail_if_subflow_error
    condition: "{% set hasFailure = false %}\r

      {% for task in outputs | keys %} \r

      \t{% for subtask in outputs[task] | keys %}\r

      \t\t{% if outputs[task][subtask].executionId is defined and outputs[task][subtask].state is defined %}\r

      \t\t\t{% if outputs[task][subtask].state == \"FAILED\" %}\r

      \         {% set hasFailure = true %}\r

      \      {% endif %}\r

      \t\t{% endif %}\r

      \t{% endfor %}\r

      {% endfor %}\r

      {{hasFailure}}"
```

returns true even if {{hasFailure}} returns false, because due to the pebble expression there is whitespace somewhere and Kestra does not properly evaluate "   false" as false. 
Debugging the output in the output tab with the same pebble expression results in this : 

Image

All condition fields should properly trim the output so that boolean expressions are properly captured.

### Environment

- Kestra Version: 0.24.2

YuvrajKatkar:
Closes  https://github.com/kestra-io/kestra/issues/11411

- Added trim method in isTruthy and isFalsy methods of TruthUtils class so that whitespaces in pebble expression for condition are handled
- Added test file for TruthUtils class

---

### Setup Instructions

No setup information needed

### DIFF
--- a/core/src/main/java/io/kestra/core/utils/TruthUtils.java
+++ b/core/src/main/java/io/kestra/core/utils/TruthUtils.java
@@ -6,10 +6,10 @@ abstract public class TruthUtils {
     private static final List<String> FALSE_VALUES = List.of("false", "0", "-0", "");
 
     public static boolean isTruthy(String condition) {
-        return condition != null && !FALSE_VALUES.contains(condition);
+        return condition != null && !FALSE_VALUES.contains(condition.trim());
     }
 
     public static boolean isFalsy(String condition) {
-        return condition != null && FALSE_VALUES.contains(condition);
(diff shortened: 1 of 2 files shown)


AcevedoR (approved):
LGTM, thanks

Issue #11647, opened 2025-10-01
Change Label Deserializer to be only for List (or even remove it completely)


Skraye:
### Feature description

Json serial/deserial for labels is a custom one because we transitionned for map to list of labels.
Its been more than two years, I dont think map of labels is still a thing and we could simplify by remove useless code.

loicmathieu:
It's not for transitionning, we do want to support the two ways.
Unless someone from the product team said the opposite @Ben8t @anna-geller ?

anna-geller:
I have to say I don't understand the implications here. Can you show some examples of what would no longer be possible if we were to merge that PR?

proddata:
We want to support both map and arrays of kvs for labels, therefore closing this

PR #11648, opened 2025-10-01
fix: set Label schema definition as list of label only, deprecate old…

Skraye:
… serdes for it and add schema definition for label

related to kestra-io/client-sdk#62

Following issue after this change :
https://github.com/kestra-io/kestra/issues/11647

Long story short : Label could be a map in previous version from 2years +, but the current Schema definition is breaking the OpenApi Spec, so with this PR I only changed the schema definition, but later would be great to remove totally the map part of the core

edit: 
as per @loicmathieu 
> we already support labels as maps in some trigger or tasks for ex so for consistency it's better to support it here also

We may not remove the map if its not from previous code but just that we wanted to handle both format

Updated the file change so its works with schema and keep the map

### DIFF
--- a/core/src/main/java/io/kestra/core/models/Label.java
+++ b/core/src/main/java/io/kestra/core/models/Label.java
@@ -1,6 +1,7 @@
 package io.kestra.core.models;
 
 import io.kestra.core.utils.MapUtils;
+import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.annotation.Nullable;
 import jakarta.validation.constraints.NotEmpty;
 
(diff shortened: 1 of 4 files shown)


fhussonnois (approved):
LGTM!

loicmathieu (changes requested):
I'm understand the rational here but I strongly advise to discuss this with the produce team first as it's on their demand it was implemented like that.
Moreover, we already support labels as maps in some trigger or tasks for ex so for consistency it's better to support it here also.

it should be fixed at the client side as there is nothing wrong with supporting multiple schema...

loicmathieu:
This also would be a BC

Skraye:
I'll try to find another way to have a proper schema definition, if nothing found I'll create a meet with you and product team

loicmathieu (approved):
LGTM

Issue #11608, opened 2025-09-30
[UI error] Something went wrong while loading the execution (reboot)


tchiotludo:
user still complain and reboot of this issue: https://github.com/kestra-io/kestra/issues/7623

<img width="1145" height="760" alt="Image" src="https://github.com/user-attachments/assets/64412f17-3f99-4eaf-8f74-6f3743b4e2f1" />

Add some server logs at least to have a way to tackle this issues and a more explicit error on the front end.

brian-mulier-p:
We don't have any way to extract a more meaningful error message on the FE side. I'm eager to know which version it occurs on. I've done a bunch of improvement on that area and went out of idea but since my fix I never experienced it again so I'm a bit surprised...

@AcevedoR if you really need to mitigate that further, my only lasting clue is that we open a SSE connection on both the ExecutionRoot component & within TaskRunDetails component and they may conflict with each other. I'd advise to try to have a single one opened at once but that's not an easy task as ExecutionRoot one is needed for all other components & TaskRunDetails may need to listen for an execution that is not the currently viewed one (subflow for eg.)...

But before all that let's confirm the user version and from there check if my fix is included in it before trying to do further magic

AcevedoR:
I think I had the issue yesterday on kafka-preview, I'll check it out

tchiotludo:
maybe at least capture and display the error from see on the [popup ](https://github.com/kestra-io/kestra/blob/2bd4e82b428cb1d56dfa8b08743ca92b1ee3dc13/ui/src/stores/executions.ts#L372)?

AcevedoR:
- a mitigation was added in https://github.com/kestra-io/kestra/pull/11652
- but we found another bug when investigating the issue (404 sometimes happening and the user has to refresh), this one is still in progress https://github.com/kestra-io/kestra-ee/pull/5334

Ben8t:
@AcevedoR pushed some fix last week; @brian-mulier-p to check on the backport parts

Ben8t:
seems related https://github.com/kestra-io/kestra/issues/11975

brian-mulier-p:
It's backported. If reoccuring after the 1.0.6 release feel free to reopen

caxefaizan:
Reported on `1.0.13` here https://github.com/kestra-io/kestra-ee/issues/5948

Ben8t:
Next step: 
- Rewrite everything that is done by backend as recent changes will bring more async. behavior

[...]

PR #11652, opened 2025-10-01
fix(executions): try to mitigate SSE and debug log SSE errors

AcevedoR:
- advance on https://github.com/kestra-io/kestra/issues/11608


brian-mulier-p (approved):
LGTM

PR #11676, opened 2025-10-02
fix(executions): evaluate multiple conditions in a separate queue

loicmathieu:
By evaluating multiple condition in a separate queue, we serialize their evaluation which avoir races when we compute the outputs for flow triggers. This is because evaluation is a multi step process: first you get the existing condtion, then you evaluate, then you store the result. As this is not guarded by a lock you must not do it concurrently.

The race can still occurs if muiltiple executors run but this is less probable. A re-implementation would be needed probably in 2.0 for that.

Fixes https://github.com/kestra-io/kestra-ee/issues/4602


Skraye (approved):
LGTM but would be great to have a note somewhere explaing why this queue isnt necessary in Kafka and is in JDBC

loicmathieu:
@Skraye in fact this already exist in Kafka, it uses a slightly different mechanism and the Kafka implementation didn't use queues but directly topics. But in Kafka, Multiple conditons are already dispatch to a separate topic and handle by a different Kafka Stream executor/application.

Issue #11067, opened 2025-09-03
Int values on labels are breaking Resume flow operations


danlsgiga:
### Describe the issue

Using the blueprint https://kestra.io/blueprints/request-resources and setting any label with value being an INTEGER (e.g: `year: 2025`) fails at the Resume action with the following error:

```
Internal server error: jakarta.validation.ConstraintViolationException: Illegal flowwithsource source: class java.lang.Integer cannot be cast to class java.lang.String (java.lang.Integer and java.lang.String are in module java.base of loader ‘bootstrap’) (through reference chain: io.kestra.core.models.flows.FlowWithSource[“labels”])
```

I cannot reproduce this issue using only string on labels values.
I can reproduce it even setting the integer label value on double quotes and single quotes (e.g: `year: "2025"` or `year: '2025'`) 

### Environment

- Kestra Version: v0.24.4

MilosPaunovic:
Tested this to see if it's a `UI` bug, but seems like it is not.

Reproduction steps:

1. Execute this simple flow:
```yaml
id: resume_execution
namespace: github.qa

labels:
  year: 2025

tasks:
  - id: wait_for_approval
    type: io.kestra.plugin.core.flow.Pause
    onResume:
      - id: approved
        description: Whether to approve the request
        type: BOOLEAN
        defaults: true
```

2. Open `Execution Overview` tab (if not already opened).
3. Click `Resume` button - it will trigger `/executions/{id}/resume/validate` endpoint which will return the error from description of this issue, listed below:

`Internal server error: jakarta.validation.ConstraintViolationException: Illegal flowwithsource source: class java.lang.Integer cannot be cast to class java.lang.String (java.lang.Integer and java.lang.String are in module java.base of loader 'bootstrap') (through reference chain: io.kestra.core.models.flows.FlowWithSource[\"labels\"])`

Utsav7428:
Hi assign it to me

Utsav7428:
Hi @MilosPaunovic  so there are two ways we can go about it 
1. the easy way is to take the value in labels as "{{ '2025' }}"  as this will safeguard our string 
2. the second way involves chaning the yamlParser and taking labels as key:String and value:Object which might be a bigger change 
 so needed guidance if we can move forward with yaml parser change or using the "{{ '2025' }}" hack for non-String values.
this is the output of the first way 
```json
 {
    "id": "resume_execution2",
    "namespace": "github.qa",
    "inputs": [
        {
            "input": {
[...]

PR #11688, opened 2025-10-02
fix(backend): failing `/resume/validate` endpoint for integer label values

sanket-mundra:

This PR closes Issue: https://github.com/kestra-io/kestra/issues/11067

Integer values in the labels are breaking Resume flow operations. Steps to reproduce this is mentioned here: https://github.com/kestra-io/kestra/issues/11067#issuecomment-3252597083

As discussed in the [thread](https://github.com/kestra-io/kestra/issues/11067#issuecomment-3284202910), labels support key value pairs as string: string and we need to convert the value to string inside the endpoint, so I have done the same here.

---

Used the same steps to reproduce the issue as mentioned above.
Please have a look at this:

https://github.com/user-attachments/assets/4a5c85db-ec83-4224-9528-8944e4e4dd77

---

### Setup Instructions
Used the same setup as mentioned above and ran the backend server locally after the fix as mentioned in the contributions page.


loicmathieu (changes requested):
Hi,
Thanks for your contribution.
The change should be done on the resume/validate endpoint which should convert new labels to avoid saving incorrect labels to the database which is the source of truth

sanket-mundra:
@loicmathieu when I tried reproducing this on my local, this is the code path where the exception was being thrown from when I hit `/resume/validate` API. When I click on the resume button on execution overview page, this the code path being hit. It would be helpful if you can you point me to the flow/class you are talking about. Thanks in advance!

loicmathieu:
Certainly in this method: https://github.com/kestra-io/kestra/blob/develop/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java#L1394

sanket-mundra:
Sorry, this is the flow I followed:
https://github.com/kestra-io/kestra/blob/develop/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java#L1394 ->
https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/services/ExecutionService.java#L560 ->
https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/services/ExecutionService.java#L601 ->
https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/services/PluginDefaultService.java#L311 ->
https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/services/PluginDefaultService.java#L393 (where I have put the fix).

Isn't this the same place you are referring to?

loicmathieu:
@sanket-mundra there is something I don't understand here, why it fail at this point as this method is used in a lot of different places.
I need more time to investiguate and undertstand what's going on.

sanket-mundra:
Sure. Let me know once you figure it out. I was also confused about this as we were literally referring to the same thing

loicmathieu:
@sanket-mundra there is a trick that can be used and would be cost free, change the method `FlowRepositoryInterface.findByExecutionWithoutAcl(Execution)` to use `this.findByIdWitSourceWithoutAcl()` instead of `this.findByIdWithoutAcl()` to avoid parsing the flow again when injection plugin versionning which would avoid the issue. Please also change the return type of the method.

Could you also add a test for that? It should be inside the `ExecutionControllerRunnerTest`.

sanket-mundra:
@loicmathieu Thanks for the trick. I have made the fix. Can you check and let me know?

loicmathieu (approved):
LGTM, thanks

PR #11692, opened 2025-10-02
fix(pagination): separate page size keys for flows, executions and logs routes

### LINKED ISSUE #11568: Pagination size from pagination not working as expected.
### Describe the issue

When you change the pagination size  for one page it takes the same size in another page without explicitly applying for that.
**Expectation** is to keep the pagination size change limited to where it was done.

- Flow page - Change size
- Now go to Executions page you will see the same size.

https://github.com/user-attachments/assets/dfc6bac7-09c7-4a6e-bd8d-a0c37d162e1b

### Environment

- Kestra Version: develop

Striker205:

Fixes an issue where lists (flows, executions, logs) were sharing the same localStorage key for pagination size.
As a result, changing the page size in one list also changed it in the others.

Added constants in storageKeys for each route:

PAGINATION_SIZE_FLOWS
PAGINATION_SIZE_EXECUTIONS
PAGINATION_SIZE_LOGS

Updated Pagination.vue to select the key based on route.name.

This ensures each list remembers its own page size.

Closes #11568 

1. Opened Flows and set page size to 10.
2. Opened Executions and set page size to 50.
3. Opened Logs and set page size to 100.
4. Switching back to Flows still shows 10, Executions 50, Logs 100.


MilosPaunovic (approved):
Hey @Striker205,

I loved your initial approach here, only had to make the change to generalize things a bit more is because we have many more pages using pagination than just `Flows`, `Executions` and `Logs`. 

This way, we've got them all covered.

Thanks so much for the initial work on this one! 🚀

PR #11718, opened 2025-10-03
fix(executions): make state_duration generated on queries

### LINKED ISSUE #11593: Duration sort on executions not arranging executions in correct order
### Describe the issue

If you have a mix of executions varying from durations of seconds to minutes then the duration sort does not work. as an example, see this screenshot of attempting to arrange durations in ascending order on 1.0.2. 

Image

### Environment

- Kestra Version: 1.0.2

AcevedoR:
- fixes https://github.com/kestra-io/kestra/issues/11593

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/State.java
+++ b/core/src/main/java/io/kestra/core/models/flows/State.java
@@ -86,10 +86,11 @@ public class State {
 
     @JsonProperty(access = JsonProperty.Access.READ_ONLY)
     public Duration getDuration() {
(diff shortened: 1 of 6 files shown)


AcevedoR:
It cannot work for mysql as it requires deterministic expressions in GENERATED columns, so, I cannot get the current date in such generated column. I guess I'll have to do it explicitly with Jooq when querying

AcevedoR:
old pr, this was fixed elsewhere

PR #11723, opened 2025-10-03
fix(executions): purge executions by 100 by default

loicmathieu:
As 500 may be too much if executions are huge as the batch will be loaded in memory.


brian-mulier-p (approved):
LGTM

PR #11729, opened 2025-10-03
Revert "fix(core): properly encode filenames with spaces in URI (#115…

loicmathieu:
…99)"

This reverts commit aff8fde79df4e05886a707e00281a6fdf23dc71a.


loicmathieu:
Merging it without approval as it revert a commit I approved by mistake.

PR #11733, opened 2025-10-03
fix(core): obfuscate secrets used as default inputs (#11681)

### LINKED ISSUE #11681: Defaults with secrets are rendered on UI
### Describe the issue

Defaults for inputs can now be resolved dynamically, and they can use a secret. That’s exactly what I need, but I think there might be a catch.
 Currently, the secret in the default is displayed as its actual value in the UI, rather than being masked as ****, unlike in other places (logs, outputs).
 Is this a bug, or intended behavior? Personally, I like the current behavior, and it will be cool if we can still use secrets as default, even if they will be displayed as **** on UI
```
- id: envName
    type: STRING
    defaults: "{{secret ('Environment')}}"
```

Image

Image

### Environment

- Kestra Version: 1.0.3

fhussonnois:
Make sure values return from pebble function are obfuscate when return from the input validation endpoints.

Changes:
* UI: Don't send default input values when creating new execution

Fixes: #11681

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
@@ -1,7 +1,6 @@
 package io.kestra.core.runners;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.annotations.VisibleForTesting;
 import io.kestra.core.encryption.EncryptionService;
 import io.kestra.core.exceptions.IllegalVariableEvaluationException;
 import io.kestra.core.exceptions.KestraRuntimeException;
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/kestra/core/runners/VariableRendererFactory.java
```diff
@@ -0,0 +1,34 @@
...
+    
+    @Inject
+    protected VariableRenderer variableRenderer;
+    
+    public VariableRenderer getDefault() {
+        return variableRenderer;
+    }
+    
+    public VariableRenderer createWithMaskedSecrets() {
+        return new VariableRenderer(applicationContext, variableConfiguration, List.of(SecretFunction.NAME));
```
loicmathieu: This would prevent overriding the default variable renderer which is used by one of our customer.
See https://github.com/kestra-io/kestra/pull/11559 which work around the issue, the same factory should be used here.

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/runners/VariableRendererFactory.java
fhussonnois: Thanks, good catch! I'll wait for the PR to be merged to rebase this one!

loicmathieu (approved):
LGTM

MilosPaunovic (approved):
UI side looks alright! ✅

PR #11737, opened 2025-10-03
fix: kv-error-on-missing

### LINKED ISSUE #11072: Get KV with `errorOnMissing: false` with Pebble Expression doesn't work
### Describe the issue

I want to fetch the key based on the input provided.
The key doesn't exist, hence i am using the property `errorOnMissing: false`
But it doesn't work if the key is provided using pebble expression.
If the same key was hardcoded as string, the `errorOnMissing` works as expected

Sample FLow
```
id: kv_issue
namespace: company.team
variables:
  defaultEnv: dev
inputs:
  - id: env
    type: STRING
    required: false

tasks:
  - id: resolveEnv
    values:
      key: "{{ inputs.env ?? vars.defaultEnv }}"
    description: "Use input env if provided, else use defaultEnv"

  - id: fetch-kv-without-pebble
    type: io.kestra.plugin.core.kv.Get
    key: "dev"
    errorOnMissing: false

  - id: fetch-kv
    type: io.kestra.plugin.core.kv.Get
    key: "{{ kv(outputs.resolveEnv.values.key ) }}"
    errorOnMissing: false
```
Version:0.24.4

### Environment

- Kestra Version: develop

Image

hemantmm:
closes: #11072

✅ Added context detection to the KvFunction to identify when it's called inside a KV task.
✅ Modified the KvFunction to default errorOnMissing to false when inside another KV operation.
✅ Added error handling to prevent bubbling up of exceptions from nested KV calls when not needed.

---

Fix: Updated the KV operation handling to solve nested KV function calls:

- Introduces a context flag to detect when a KV function is called from within a KV task
- Prevents cascading failures when an inner KV call fails but the outer operation has errorOnMissing set to false
- Handles rendering exceptions gracefully when a key can't be resolved but errors are suppressed

This change allows the KV task to work correctly when using template expressions like: key: `{{ kv(outputs.resolveEnv.values.key ) }}` with `errorOnMissing: false`


hemantmm:
@loicmathieu, can you have a look at this PR?

hemantmm:
@loicmathieu, can you have a look at this PR?

hemantmm:

loicmathieu:
@hemantmm in fact everything is working correctly, the flow in the issue was not correct as it forgot to add the `errorOnMissing=false` parameter to the kv function.

Thanks for your contribution anyway.

PR #11742, opened 2025-10-04
feat(tests): add test coverage for cli commands with no repo config

### LINKED ISSUE #10416: Add a test in the `cli` module to launch a command without any repository configured
### Describe the issue

Some commands should be able to run withou any repository configured (without any configuration other than the default configuration provided inside the `cli` module).

For example, the `kestra plugins install` command should not need any repository.

But inside `cli` tests, there is an `application-test.yml` that do contains a repository configuration so we don't assert that.
We had multiple time this issue, the last time was for 0.23.8 release, see https://github.com/kestra-io/kestra-ee/actions/runs/16591586897/job/46929053261 for example.

So we should have a test that assert that we can run a command without any specific configuration set.

### Environment

- Kestra Version: develop

MTarek165:
closes: #10416

### Description
 This PR covers two scenarios of cli commands that don't have configs , both uses ApplicationContext.builder().deduceEnvironment(false).start() to ensure that no configs passed and ensure isolation.
1- Used Namespace Kv command (without params) that don't depend on configs , with no configs passed ,should run successfully as expected.
2- Used Create flow command at the same environment and it depends on configs (DB repository), with no configs passed, should give " No bean of type [io.kestra.core.repositories.FlowRepositoryInterface] exists" and fail as expected.


loicmathieu (changes requested):
As it's test related, please don't create a new command but use an existing command in your test

MTarek165:
@loicmathieu I did as asked to create a dummy command but now do you want me to remove this new command and use existing one that don't depend on config at the first scenario, right?

MTarek165:
@loicmathieu done, I removed the added command and replaced with existing one "kestra namespace kv" that don't depend on configs, now both scenarios tested on real commands.

loicmathieu (approved):
LGTM, thanks a lot

PR #11744, opened 2025-10-04
refactor: resume-to-execution-updatable-task

### LINKED ISSUE #9774: Refactor `io.kestra.plugin.core.execution.Resume` to be an `ExecutionUpdatableTask`
### Describe the issue

The `io.kestra.plugin.core.execution.Resume` is today a runnable task but as it's a task that update the execution, it would benefits to be refactored as an ExecutionUpdatableTask, this would avoid depending on the run context implementation and also avoid any race on execution update as the update of the execution would be done by the executor.

### Environment

- Kestra Version: develop

hemantmm:
closes: #9774

✅ Changed interface from `RunnableTask<VoidOutput>` to`ExecutionUpdatableTask`.
✅ Replaced `run()` method with `update()` method that returns the updated execution.
✅ Added `resolveState()` method to return SUCCESS state for the task.
✅ Added tracking of resumed execution in variables for better observability.

---

Fix: Updated execution handling to improve state management:

- Avoids race conditions by letting the executor handle execution updates.
- Prevents dependency on the RunContext implementation details.
- Maintains tracking of which execution was resumed in the task's execution variables

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/execution/Resume.java
+++ b/core/src/main/java/io/kestra/plugin/core/execution/Resume.java
@@ -6,20 +6,16 @@ import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.flows.FlowInterface;
 import io.kestra.core.models.flows.State;
 import io.kestra.core.models.property.Property;
-import io.kestra.core.models.tasks.RunnableTask;
+import io.kestra.core.models.tasks.ExecutionUpdatableTask;
 import io.kestra.core.models.tasks.Task;
-import io.kestra.core.models.tasks.VoidOutput;
-import io.kestra.core.queues.QueueFactoryInterface;
-import io.kestra.core.queues.QueueInterface;
+import io.kestra.core.models.tasks.runners.PluginUtilsService;
 import io.kestra.core.repositories.ExecutionRepositoryInterface;
-import io.kestra.core.runners.FlowMetaStoreInterface;
 import io.kestra.core.runners.DefaultRunContext;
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/kestra/plugin/core/execution/Resume.java
```diff
@@ -89,17 +85,28 @@ public VoidOutput run(RunContext runContext) throws Exception {
         ExecutionService executionService = applicationContext.getBean(ExecutionService.class);
         ExecutionRepositoryInterface executionRepository = applicationContext.getBean(ExecutionRepositoryInterface.class);
         FlowMetaStoreInterface flowExecutor = applicationContext.getBean(FlowMetaStoreInterface.class);
-        QueueInterface<Execution> executionQueue = applicationContext.getBean(QueueInterface.class, Qualifiers.byName(QueueFactoryInterface.EXECUTION_NAMED));
 
-        Execution execution = executionRepository.findById(executionInfo.tenantId(), executionInfo.id())
+        Execution targetExecution = executionRepository.findById(executionInfo.tenantId(), executionInfo.id())
             .orElseThrow(() -> new IllegalArgumentException("No execution found for execution id " + executionInfo.id()));
```
loicmathieu, line 90: You don't need to load the execution from the execution repository as it's already passed to the update method.

loicmathieu (changes requested):
Sorry, I just noticed that the issue is wrong, this task cannot be refactored as an `ExecutionUpdatableTask` task as it not always update the current execution :( 

I'll close both this PR and the issue, sorry about that and thanks for looking into it.

PR #11746, opened 2025-10-04
Fix: Ensure flowable Switch task executes children in parallel #11745

### LINKED ISSUE #11745: Flowable Switch Executes Tasks Sequentially Instead of in Parallel
### Describe the issue

A flowable Switch runs its child tasks one after another instead of at the same time

```yaml

id: definitive_switch_fix_test
namespace: dev

tasks:
  - id: my_flowable_switch
    allowFailure: true 
    value: "A"
    cases:
      "A":
        - id: task_A
          duration: PT5S

        - id: task_B
          duration: PT5S

```
### Total execution time is 10s instead of 5s

Image

_I have found the root cause and am currently working on this issue, can you please assign this issue to me, the original issue(#10508) is also assigned to me _

### Environment

- Kestra Version: develop

bikashagarwala0:
Closes #11745 

**Summary**

This PR addresses a runtime logic bug in the `flowable Switch` task. When a `Switch` task was made `flowable` (e.g., with `allowFailure: true`), it would incorrectly execute the list of tasks in the chosen `case` sequentially, one after another. This defeated the purpose of a flowable task, which is expected to handle child tasks as independent, parallel operations.

This behavior was often hidden by very fast tasks (like `Log`) but became obvious when using tasks with a duration (like `Sleep`), which would cause the total execution time to be the sum of all task durations instead of the duration of the longest task.

**Solution**

The root cause was in the `resolveNexts` method of `Switch.java`, which was incorrectly using the `FlowableUtils.resolveSequentialNexts` utility.

This has been changed to use `FlowableUtils.resolveParallelNexts`, which correctly schedules the child tasks as concurrent operations. A concurrency of `0` (unlimited) is used, as the Switch task itself doesn't impose a concurrency limit on its children.

**How to Test**

1.  Create and execute the following flow:
    ```yaml
    id: definitive_switch_fix_test
    namespace: dev
    
    tasks:
      - id: my_flowable_switch
        allowFailure: true 
        value: "A"
        cases:
          "A":
            - id: task_A
              duration: PT5S
    
            - id: task_B
              duration: PT5S
    ```
2.  Observe the **Gantt chart** for the execution. The two `Sleep` tasks will start at the same time and run in parallel. The total execution duration for the switch will be **\~5 seconds**, not \~10 seconds.

**Checklist**

  - [x] A new unit test, `flowableSwitch_parallelExecution`, has been added to `SwitchTest.java` to cover this specific case and prevent regressions. The test asserts that the total execution duration is less than the sum of the individual task durations, confirming parallel execution.

### Before
Screenshot from 2025-10-04 19-23-51

### After
Screenshot from 2025-10-04 22-53-11

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/flow/Switch.java
+++ b/core/src/main/java/io/kestra/plugin/core/flow/Switch.java
@@ -191,12 +191,16 @@ public class Switch extends Task implements FlowableTask<Switch.Output> {
 
     @Override
     public List<NextTaskRun> resolveNexts(RunContext runContext, Execution execution, TaskRun parentTaskRun) throws IllegalVariableEvaluationException {
-        return FlowableUtils.resolveSequentialNexts(
+        List<ResolvedTask> childTasks = this.childTasks(runContext, parentTaskRun);
+        Integer concurrency = 0;
+
+        return FlowableUtils.resolveParallelNexts(
             execution,
-            this.childTasks(runContext, parentTaskRun),
+            childTasks,
             FlowableUtils.resolveTasks(this.errors, parentTaskRun),
             FlowableUtils.resolveTasks(this._finally, parentTaskRun),
-            parentTaskRun
+            parentTaskRun,
+            concurrency
         );
     }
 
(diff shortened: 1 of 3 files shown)


loicmathieu:
This is by design.

In Kestra, parallel processing should be specifically set using a `Parallel` flowable task.
Only `ForEach` is capable of also doing parallel processing if configured as it's a common pattern.

I'm closing it as the related issue has already been closed.
Thanks anyway for your contribution, and sorry for your lost of time.

PR #11748, opened 2025-10-05
fix(core): prevent AllowFailure with retry from getting stuck in Running state

### LINKED ISSUE #11731: retry with AllowFailure get stuck in Running state
### Describe the issue

https://jam.dev/c/64e84fbb-202a-4e5d-9107-45e2db738dbb

raised by @xdatanomad; the execution get stuck in Running state while the values are good and should allow to move forward past the retry.

```yaml
id: reliability_task_and_flow
namespace: demo.durability

tasks:
  - id: hello
    message: Hello World! 🚀

  - id: set_kv
    key: "retry_counter_1"
    value: "1"
    kvType: NUMBER
    overwrite: true

  - id: retry_block
    type: io.kestra.plugin.core.flow.AllowFailure
    tasks:
      - id: log_counter
        message: "Current value of retry_counter_1 is: {{ kv(namespace=flow.namespace, key='retry_counter_1') | string }}"

      - id: run_script
        script: |
          num = {{kv(namespace=flow.namespace, key='retry_counter_1')}}
          print(f"current value: {num}")
          if num < 2:
            raise Exception("not yet!")
          exit(0)
    errors:
      - id: incr_counter
        key: retry_counter_1
        value: "{{ kv(namespace=flow.namespace, key='retry_counter_1') + 1 }}"
        overwrite: true
    retry:
      type: constant
      behavior: RETRY_FAILED_TASK
      maxAttempts: 3
      interval: PT1S
      warningOnRetry: true
    finally:
      - id: log_ok
        message: it's all good! 👍

  - id: the_end
    message: Let's go home! 🏡
```

Kestra: develop

### Environment

- Kestra Version: develop

gauravarora:
Fixes #11731

When an `AllowFailure` task had both retry logic and a `finally` block, the execution would get stuck in Running state. This occurred because the logic for determining when to execute the finally block only checked if tasks had failed, not whether they should be retried.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/executions/Execution.java
+++ b/core/src/main/java/io/kestra/core/models/executions/Execution.java
@@ -496,7 +496,7 @@ public class Execution implements DeletedInterface, TenantInterface {
         }
 
         if (resolvedFinally != null && (
-            this.isTerminated(resolvedTasks, parentTaskRun) || this.hasFailed(resolvedTasks, parentTaskRun
+            this.isTerminated(resolvedTasks, parentTaskRun) || this.hasFailedNoRetry(resolvedTasks, parentTaskRun
         ))) {
             return resolvedFinally;
         }
--- a/core/src/test/java/io/kestra/plugin/core/flow/AllowFailureTest.java
+++ b/core/src/test/java/io/kestra/plugin/core/flow/AllowFailureTest.java
@@ -55,18 +55,18 @@ class AllowFailureTest {
     @Test
     @ExecuteFlow("flows/valids/allow-failure-with-retry.yaml")
     void withRetry(Execution execution) {
-        // Verify the execution completes successfully
-        assertThat(execution.getState().getCurrent()).isEqualTo(State.Type.SUCCESS);
+        // Verify the execution completes in warning
+        assertThat(execution.getState().getCurrent()).isEqualTo(State.Type.WARNING);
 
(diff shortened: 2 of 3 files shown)


loicmathieu (approved):
LGMT, nice catch!

loicmathieu:
The new test is `AllowFailureTest` didn't pass as in the Kestra core modele, the Python task is not available.
We usually circumvent that by using a simple task like Log or OutputValues that fail based on attemptsCount.

For ex:
```yaml
- id: failed
  message: "{{ taskrun.attemptsCount == 3 ? 'ok' : ko}}"
  retry:
    type: constant
    interval: PT0.250S
    maxAttempts: 4
    maxDuration: PT15S
    warningOnRetry: true
```

Can yoiu update your test please?

loicmathieu:
@gauravarora the test is still failing, could you take care it runs without issue locally before updating the PR to avoid wasting CI resources?

loicmathieu:
@gauravarora your fix was OK, but the assertions in tests were not ;) I fixed them, it should be good now.
- When a tak is retried it has multiple attempts of the same task run (multiple taskrun is for iterative tasks with different values)
- AllowFailure ends the task in WARNING so the execution ends in WARNING
- AllowFailure didn't change the state of the task that failed: it is still FAILED, but it allows to continue the execution.

gauravarora:
Thank you @loicmathieu 🙏

PR #11780, opened 2025-10-07
feat: add patternField and regexPattern

### LINKED ISSUE #10748: New split plugin to split file based on pattern in file
Given some ion file, provide a way to split into multiple files based on file content, rather that bytes/rows. 

For a specific example we received the following:

> I want to subscribe to an Amazon SQS queue, but the problem is SQS queues don't have specific topics. The database rows that need to be updated will be sent in the form of JSON into this single queue. 
> 
> Kestra needs to trigger a flow to read the messages from the SQS Queue and I need to now split the ion file based on the table name field in the message. Is there a way to split the ion file in one run rather than having to read through the ion file multiple times? 
> 
> Throughput will be extremely important here, this message queue may have hundreds of thousands of messages per minute. Having to scan through the ion file multiple times will be extremely inefficient, as i might have 80 tables which means it'll have to scan through the file 80 times. Furthermore, creating 80 separate SQS queues also adds cost and maintenance overhead.
>

mengnankkkk:
 add new optional properties (patternField and regexPattern) and tests
Close #10748 

---

The patternField property allows you to extract routing keys directly from JSON fields.

Parse field values ​​in JSON files using Pebble expressions.
Supports complex field paths and expressions.
The regexPatern property allows you to match regular expressions.

Use Java Pattern/Matcher for regular expression matching.
Extract matching content from text lines as routing keys.

---


loicmathieu (changes requested):
As I said, I think a single split by regex is enought.
Internally, we use the ION format not JSON so splitting for JSON should not be supported by the core framework.

Also, I struggle to understand the difference between pattern and regexPattern.

core/src/main/java/io/kestra/core/services/StorageService.java
```diff
@@ -117,4 +125,38 @@ private static List<Path> partition(RunContext runContext, String extension, Str
         return files.stream().filter(p -> p.toFile().length() > 0).toList();
     }
 
+    private static List<Path> splitByRegex(RunContext runContext, String extension, String separator, BufferedReader bufferedReader, String regexPattern) throws IOException {
+        List<Path> files = new ArrayList<>();
+        Map<String, RandomAccessFile> writers = new HashMap<>();
+        Map<String, Path> writerPaths = new HashMap<>();
```
loicmathieu: This seems to be not used, you put in it but never get any value

loicmathieu (changes requested):
Overall looks good to be but if you could remove the unecessary variable before merging this would be great.

core/src/main/java/io/kestra/core/services/StorageService.java
mengnankkkk: OK I will do it!

loicmathieu (approved):
LGTM, thanks a lot

PR #11785, opened 2025-10-07
fix: modify annotations to improve openapi spec file generated

Skraye:
Related to https://github.com/kestra-io/kestra-ee/pull/5374


AcevedoR (approved):
LGTM, it could be worth to quickly check that frontend autocompletion and editor still works

PR #11814, opened 2025-10-07
fix: artifact-filename-validation

### LINKED ISSUE #10802: Artifacts with spaces or Unicode in filenames fail with IllegalArgumentException
### Describe the issue

When using spaces or unicode in output filename an IllegalArgumentException is thrown

Reproduction Steps

- Run docker start command
` docker run --pull=always --rm -it -p 8080:8080 --user=root -v /var/run/docker.sock:/var/run/docker.sock -v /tmp:/tmp kestra/kestra:latest server local`
- Navigate to http://localhost:8080/
- Click on Flows
- Click on create
- Enter the following flow
`id: unicode_and_space
namespace: company.team
tasks:
  - id: write_weird_names
    script: |
      echo ok > "with space.txt"
      echo สวัสดี > "สวัสดี.txt"
    outputFiles:
      - "with space.txt"
      - "สวัสดี.txt"`

- Click on Save
- Click on Execute

Expected Behaviour

- Both files appear in Outputs, downloadable as artifacts.
- Storage paths handle spaces and Unicode (escaped or encoded as needed).

Actual Behaviour

- Execution fails with
-`java.lang.IllegalArgumentException: Illegal character in path … with space.txt
`

### Environment

- Kestra Version: develop

hemantmm:
closes: #10802

✅ Fixed handling of output files with spaces and Unicode characters in filenames.
✅ Updated FilesService.`resolveUniqueNameForFile` to generate safe, unique names while preserving original filenames
✅ Ensured output artifacts are downloadable and storage paths handle special characters correctly.

---

Fix:

Updated `/core/src/main/java/io/kestra/core/runners/FilesService.java`:

- Uses only the filename for hashing, preserving spaces and Unicode.
- Generates a unique, URL-safe ID for each output file.
- Ensures output files with special characters are uploaded and retrievable.

Example:

Before:

- Output with "with space.txt" or "สวัสดี.txt" failed with IllegalArgumentException.

After:

- Output files with spaces or Unicode are handled and downloadable as artifacts.


loicmathieu:
@hemantmm test is still failing

hemantmm:
> @hemantmm test is still failing

Will try to fix it

hemantmm:
@loicmathieu, tried it, but still the test fails.

Could you help me figure out the issue?

hemantmm:
> @hemantmm test is still failing

@loicmathieu, I have fixed the failing test.

core/src/main/java/io/kestra/core/runners/FilesService.java
```diff
@@ -82,6 +82,12 @@ public static Map<String, URI> outputFiles(RunContext runContext, List<String> o
     }
 
     private static String resolveUniqueNameForFile(final Path path) {
-        return IdUtils.from(path.toString()) + "-" + path.toFile().getName();
+        String filename = path.getFileName().toString();
+        
+        String uniqueId = IdUtils.create();
```
loicmathieu: Why are you adding a unique ID in front of the path?
Would it be possible to use as before only the path and endode the name part?

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/runners/FilesService.java
hemantmm: Thanks for catching that — I’ve updated the method to only URL-encode the filename and use a deterministic prefix from `IdUtils.from(path.toString())` to preserve uniqueness across directories.

loicmathieu (approved):
LGTM, thanks

PR #11867, opened 2025-10-08
feat(triggers): allowing webhook to send only outputs

tchiotludo:


core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
```diff
@@ -48,6 +48,38 @@
...
 
+        The response body will contain the execution ID if the execution is successfully triggered using the following format:
+        ```json
+        {
+          "tenantId": "your_tenant_id",
+          "namespace": "your_namespace",
+          "flowId": "your_flow_id",
+          "flowRevision": 1,
+          "trigger": {
+            "id": "the_trigger_id",}
```
loicmathieu:
```suggestion
            "id": "the_trigger_id",
```

Not 100% sure but this closing braket is strange.

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
```diff
@@ -48,6 +48,38 @@
         - 200 if the webhook triggers an execution.
         - 204 if the webhook cannot trigger an execution due to a lack of matching event conditions sent by other application.
 
+        The response body will contain the execution ID if the execution is successfully triggered using the following format:
```
loicmathieu, line 51: Addingt the format here is kind of annoying as it may change and the documentation becomes wrong.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -487,7 +487,7 @@ public PagedResults<Execution> searchExecutionsByFlowId(
     @Post(uri = "/webhook/{namespace}/{id}/{key}")
     @Operation(tags = {"Executions"}, summary = "Trigger a new execution by POST webhook trigger")
     @SingleResult
-    public Publisher<HttpResponse<WebhookResponse>> triggerExecutionByPostWebhook(
+    public Publisher<HttpResponse<?>> triggerExecutionByPostWebhook(
```
loicmathieu, line 490: This is a BC as it will change the return type on the SDK.
Could you add the label to your PR and add a section describing the change?

webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionRunnerControllerTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+import static io.kestra.core.tenant.TenantService.MAIN_TENANT;
+import static io.micronaut.http.HttpRequest.GET;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@Slf4j
+@KestraTest(startRunner = true)
+class ExecutionRunnerControllerTest {
```
loicmathieu, line 50: We already have an ExecutionControllerRunnerTest, maybe add this new test in it instead of adding a new test class.

loicmathieu (approved):
Some small feedback but overall LGTM.
But it's a BC for the SDK so worth documenting it as BC.

webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionRunnerControllerTest.java
tchiotludo: `ExecutionControllerRunnerTest` don't run the runner for now, why I've created this additional test case

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
tchiotludo: not sure also, but I think if it's not documented, here, only api spec will have the documentation ... why I think it's better than nothing

loicmathieu: It does: https://github.com/kestra-io/kestra/blob/develop/webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionControllerRunnerTest.java#L88

webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionRunnerControllerTest.java
loicmathieu: It does https://github.com/kestra-io/kestra/blob/develop/webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionControllerRunnerTest.java#L88

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
loicmathieu: your call

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
tchiotludo: fixed :)

tchiotludo:
@anna-geller need your feedback for docs for response + name of this props `outputsAsResponse` please.
if ok, you can merge with this one : https://github.com/kestra-io/kestra-ee/pull/5404

anna-geller:
thanks for the ping! to match verb-form of `wait` boolean, let's perhaps go for `returnOutputs`? a bit shorter and has a verb form to match both booleans

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
anna-geller: @loicmathieu is this a correct way of describing the BC for https://github.com/kestra-io/kestra-ee/issues/5138? please correct if sth is wrong or over-explained


### All Editions: Webhook Execution API Return Type Changed

The return type of the Webhook Execution API endpoint has been updated from a **typed response** to a **generic response** to support broader use cases and improve extensibility.

**What changed:** The API method previously returned a strongly typed `WebhookResponse` object. It now returns a generic HTTP response body (`HttpResponse<?>`).

**Impact:** Any custom integrations, SDK consumers, or extensions that previously relied on the `WebhookResponse` type in the response body will need to adjust their handling logic:
* Direct access to `WebhookResponse` methods or fields will no longer compile.
* You must handle the response body dynamically and verify its type at runtime if necessary.

**Migration:**
* Treat the HTTP response body as a generic object.
* Add runtime type checks before casting if your code depends on specific response fields.
* Update any logic that assumes a fixed `WebhookResponse` structure.

This change ensures greater flexibility in future webhook response handling but requires updates to any consuming code that previously depended on a fixed response type.

anna-geller:
@tchiotludo or @loicmathieu I'll let one of you merge as CI isn't green, so I don't want to disrupt anything. Overall, the naming is adjusted, BC label and BC changelog description are added https://github.com/kestra-io/kestra-ee/issues/5138 so if CI is green, we should be good to merge ✅

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
loicmathieu: Great description!

Pradumnasaraf:
Hey,

Can I get an example flow to do the QA?

anna-geller:
@Pradumnasaraf 

```yaml
id: webhook_outputs
namespace: qa

tasks:
  - id: out
    values:
      status: "ok"

outputs:
  - id: status
    type: STRING
    value: "{{ outputs.out.values.status }}"

triggers:
  - id: webhook
    key: mywebhookKey1234
    wait: true
    returnOutputs: true
```

PR #11870, opened 2025-10-08
feat: add-subtask-autocompletion

### LINKED ISSUE #11079: auto-completion not suggesting `outputs` for subtasks in `WorkingDirectory`
### Describe the issue

The subtask IDs are not suggested in auto-completion when using `outputs` in `io.kestra.plugin.core.flow.WorkingDirectory`. Only the main task ID is visible. I have to explicitly type the subtask ID to see all the output properties (ref image 2) of that subtask.

So, it should show all 3 `clone, `delet_file` and `file_system`

### Steps to Reproduce

Copy and paste the snippet below and try to auto-complete the `conditions:` in the `assert` task by clicking and pressing <kbd>ctrl</kbd> + <kbd>space</kbd> after `outputs.`.

```yaml
id: git_clone
namespace: sanitychecks

tasks:
  - id: file_system
    type: io.kestra.plugin.core.flow.WorkingDirectory
    tasks:
      - id: clone
        type: io.kestra.plugin.git.Clone
        url: https://github.com/dbt-labs/jaffle_shop
        branch: main   

      - id: delete_file # Delete one of the files from the cloned repo
        commands:
        - rm README.md

      - id: assert
        type: io.kestra.plugin.core.execution.Assert
        conditions: 
          - "{{ outputs.}}"
```

### Screenshot for reference

Image
Image

### Environment

- Kestra Version: 0.24.4

hemantmm:
closes: #11079

✅ Added recursive extraction of subtask IDs for WorkingDirectory tasks.
✅ Updated auto-completion logic for `outputs` to suggest all subtask IDs, not just top-level ones.
✅ Improved matching for nested outputs (e.g., `outputs.clone`, `outputs.delete_file`, etc.).

---

Fix:

- Added a new helper method `extractAllTaskIds()` that recursively collects all task IDs, including those nested in WorkingDirectory tasks.
- Updated the `nestedFieldAutoCompletion` method to use this helper for the `outputs` field, so auto-completion now suggests all subtask IDs.
- Improved the regex for matching nested outputs to support deeper nesting.


brian-mulier-p (changes requested):
I think it's not scalable and we have a bunch of properties that are a list of task (just like tasks). If you want I'll take over but if you want to give it a try, a hint could be:

Bring the list of possible task types to flowAutoCompletionProvider.ts, you may need to create a getter for it in the pinia store (only tasks, not triggers or anything else).
You transmit the raw YAML source to your method and you can call FlowYamlUtils.extractFieldFromMaps(source, "id", undefined, (element) => element.type !== undefined && allTypes.includes(element.type)). From there you will have a {id}[] as a return and those are all the ids that should be included in the autocompletion.

Even though your current method would help, there would still be a lot of uncovered cases (look at DAG task that have a then else and a bunch of others that are kind of like tasks for WorkingDirectory.

ui/tests/storybook/components/inputs/InputsForm.stories.jsx
```diff
@@ -56,18 +57,58 @@ export const InputTypes = {
...
+        
+        await userEvent.click(singleSelect);
+        
+        await waitFor(() => {
+            const dropdowns = document.querySelectorAll(".el-select-dropdown");
+            expect(dropdowns.length).to.be.greaterThan(0);
+            const items = document.querySelectorAll(".el-select-dropdown__item");
+            expect(items.length).to.be.greaterThan(0);
+            return items;
+        }, {timeout: 3000});
```
brian-mulier-p, line 70: Can you explain why you had to go for such verbose approach for testing instead of the initial getByText?

brian-mulier-p (changes requested):

ui/tests/storybook/components/inputs/InputsForm.stories.jsx
hemantmm: I switched to the `waitFor`+`querySelectorAll` approach because the dropdown items rendered by Element Plus `(.el-select-dropdown__item)` are not part of the same DOM tree as the story’s `canvasElement`.

When using `getByText`, Storybook queries only within the story’s rendered container — but Element Plus portals dropdowns to the global `document.body`. As a result, `getByText('Second value')` couldn’t find them, even though they were visible in the UI.

To handle this, I used `waitFor` with `document.querySelectorAll('.el-select-dropdown__item')` to wait until the dropdown’s portal content was mounted, ensuring the test doesn’t fail due to async rendering or delayed portal initialization.

brian-mulier-p: What is more surprising is that it used to work before :thinking:

hemantmm:
@brian-mulier-p, I will close this is PR and you can go ahead with this issue.

PR #11875, opened 2025-10-08
chore(webserver): exclude metrics from access log

yuri1969:

Do not log metrics requests, such as:

> 2025-10-08 10:16:09,649 INFO  default-nioEventLoopGroup-1-4 io.kestra.webserver.access 2025-10-08T10:16:09.642Z | GET /metrics/jvm.memory.used?tag=area:heap HTTP/1.1 | status: 200 | ip: x.x.x.x | length: 238

### DIFF
--- a/cli/src/main/resources/application.yml
+++ b/cli/src/main/resources/application.yml
@@ -49,6 +49,8 @@ micronaut:
           - /ui/.+
           - /health
           - /health/.+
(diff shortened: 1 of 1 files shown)


loicmathieu (approved):
LGTM, makes sense as we already exclude the prometheus endpoint

PR #11883, opened 2025-10-08
feat(core): add support for execution kind filter

### LINKED ISSUE #10658: [Playground] Ability to list Executions created from playground
Testing the `Playground` feature got me thinking that it's a bit weird that I have no option to list or filter the executions created using this feature inside the executions list.

Maybe there could be a separate filter or checkbox for this, and those executions should not appear in the list unless specifically requested.

MTarek165:
closes : #10658
### Description
- added new Kind filter provides more flexibility at execution filtration based on kind 
- the filter supports EQUAL and NOT_EQUAL operations

https://github.com/user-attachments/assets/f272e48d-44ec-4a39-a437-89f95e0a09f5

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcExecutionRepository.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcExecutionRepository.java
@@ -54,6 +54,8 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static io.kestra.core.models.QueryFilter.Field.KIND;
+
 public abstract class AbstractJdbcExecutionRepository extends AbstractJdbcRepository implements ExecutionRepositoryInterface, JdbcQueueIndexerInterface<Execution> {
     private static final int FETCH_SIZE = 100;
     private static final Field<String> STATE_CURRENT_FIELD = field("state_current", String.class);
@@ -299,7 +301,7 @@ public abstract class AbstractJdbcExecutionRepository extends AbstractJdbcReposi
             .where(this.defaultFilter(tenantId, false));
 
         boolean hasKindFilter = filters != null && filters.stream()
-            .anyMatch(f -> "kind".equalsIgnoreCase(f.field().name()) );
+            .anyMatch(f -> KIND.value().equalsIgnoreCase(f.field().name()) );
         if (!hasKindFilter) {
             select = select.and(NORMAL_KIND_CONDITION);
         }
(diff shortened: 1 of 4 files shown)


jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcExecutionRepository.java
```diff
@@ -295,9 +295,13 @@ private SelectConditionStep<Record1<Object>> findSelect(
                 field("value")
             )
             .from(this.jdbcRepository.getTable())
-            .where(this.defaultFilter(tenantId, false))
-            .and(NORMAL_KIND_CONDITION);
+            .where(this.defaultFilter(tenantId, false));
 
+        boolean hasKindFilter = filters != null && filters.stream()
+            .anyMatch(f -> "kind".equalsIgnoreCase(f.field().name()) );
```
loicmathieu:
```suggestion
            .anyMatch(f -> KIND.value()..equalsIgnoreCase(f.field().name()) );
```

This to avoid hardcoding the value of the KIND enum item

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
```diff
@@ -309,6 +309,9 @@ protected Condition getConditionOnField(
                 throw new InvalidQueryFiltersException("Label field value must but instance of Map");
             }
         }
+        if(field==QueryFilter.Field.KIND){
```
loicmathieu:
```suggestion
        if (field == QueryFilter.Field.KIND) {
```

Nitpick but as you start contributing regularly would be great if you could follow our coding format ;)

loicmathieu (changes requested):
2 small changes if you can make them, otherwize looks good to me

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcExecutionRepository.java
MTarek165: done

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
MTarek165: done ;)

loicmathieu (approved):
LGTM but I need to do the same change to our Elasticsearch repository in EE so I'll merge this one after the change in EE.

PR #11888, opened 2025-10-08
fix(ui): Prevent all doc-cards in a row from expanding

### LINKED ISSUE #11786: Docs Panel Flow Properties box row opens on selection of just one
### Describe the issue

https://github.com/user-attachments/assets/3bab2745-59cf-4c41-bf0b-7d181e7111fb

Kestra Version: Develop

### Environment

- Kestra Version: develop

santhoshh-kumar:

This PR fixes a visual bug in the `doc-card` component. Previously, when a user expanded a `doc-card`, all other cards in the same row would also expand vertically. This was due to the default alignment settings of the flexbox container.

This change applies `align-items: start` to the `.doc-cards` container in `ui/src/components/layout/EnhancedMarkdown.vue`, ensuring that only the selected card expands, while others in the same row maintain their original size.

fixes #11786

---

This is a general UI fix and can be tested on any page that uses the `doc-card` component.

**Screenshot of the fix:**

image

**To verify the fix:**
1. Navigate to the page with multiple `doc-card` components arranged in a row.
2. Click on one of the cards to expand it.
3. Confirm that only the clicked card expands and the others in the row are unaffected.

---

### DIFF
--- a/ui/src/components/layout/EnhancedMarkdown.vue
+++ b/ui/src/components/layout/EnhancedMarkdown.vue
@@ -741,6 +741,7 @@
         display: grid;
         gap: var(--spacer, 1rem);
         grid-template-columns: repeat(auto-fit, minmax(260px, 1fr));
+        align-items: start;
     }
 
     :deep(.doc-card) {


Piyush-r-bhaskar (approved):
LGTM 🚀

PR #11895, opened 2025-10-08
Fix/jdbc timezone prague midnight

### LINKED ISSUE #10214: H2ExecutionRepositoryTest.dailyStatistics test is timing-dependant
### Describe the issue

The `H2ExecutionRepositoryTest.dailyStatistics` test fails when it gets started past midnight CEST (e.g. from 12:00 AM to 1:00 AM) using the `Europe/Prague` timezone ID.

The `assertThat(result.get(10).getDuration().getAvg().toMillis()).isGreaterThan(0L);` assertion fails since the 10th `result` does not contain any stats (`getAvg()`) - stats are contained in the previous - 9th - `result`.

### Environment

- Kestra Version: develop

PaulPio:
This PR resolves an intermittent issue where the grouping of daily execution statistics was inconsistent, leading to missing data points or failing tests, especially around midnight in specific timezones.

The issue is tracked in closes #10214.

The bug was in AbstractJdbcExecutionRepository.java where the date-group keys for the daily statistics interval were being formatted using the host system's default timezone (ZoneId.systemDefault()). When this host system timezone differed from the timezone specified in the query's startDate, the grouping logic would fail to consistently align the records, particularly when aggregated over midnight.

The Fix:
I've updated the logic (specifically lines 706-711) to use the query-provided timezone (startDate.getZone()) when creating the DateTimeFormatter for the interval keys. This ensures that the grouping intervals are always consistent and dependent only on the user's requested query parameters, not the underlying server's local configuration.

How the changes have been QAed?
This change is a direct fix for a test stability and accuracy issue within the core repository.

The updated logic in jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcExecutionRepository.java ensures the underlying database queries and grouping keys are correctly generated.

The resolution is verified by confirming that the problematic unit test, io.kestra.jdbc.repository.H2ExecutionRepositoryTest.dailyStatistics, now runs and passes reliably regardless of the system's local timezone setting.


package-lock.json
loicmathieu, line 1: Please remove this modification from your commit

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcExecutionRepository.java
```diff
@@ -703,7 +703,10 @@ private static List<DailyExecutionStatistics> fillDate(
     ) {
         List<DailyExecutionStatistics> filledResult = new ArrayList<>();
         ZonedDateTime currentDate = startDate;
-        DateTimeFormatter formatter = DateTimeFormatter.ofPattern(format).withZone(ZoneId.systemDefault());
+    // Use the timezone from the provided startDate to format intervals so grouping is consistent
+    // and not dependent on the host system timezone (fixes issues around midnight in Europe/Prague).
+    ZoneId zone = startDate != null ? startDate.getZone() : ZoneId.systemDefault();
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern(format).withZone(zone);
```
loicmathieu:
```suggestion
        // Use the timezone from the provided startDate to format intervals so grouping is consistent
        // and not dependent on the host system timezone (fixes issues around midnight in Europe/Prague).
        ZoneId zone = startDate != null ? startDate.getZone() : ZoneId.systemDefault();
        DateTimeFormatter formatter = DateTimeFormatter.ofPattern(format).withZone(zone);
```

core/src/test/java/io/kestra/core/repositories/AbstractExecutionRepositoryTest.java
```diff
@@ -498,6 +498,64 @@ protected void dailyStatistics() throws InterruptedException {
...
+            .flowRevision(1)
+            .state(new State(State.Type.SUCCESS, List.of(new State.History(State.Type.SUCCESS, dayAfterMidnight.toInstant()))))
+            .taskRunList(List.of())
+            .build();
+
+        executionRepository.save(before);
+        executionRepository.save(after);
+
+        // Small pause for DB
+        Thread.sleep(200);
```
loicmathieu: Does the test fail with it?
Thread.sleep in a test is a smell.

loicmathieu (changes requested):

core/src/test/java/io/kestra/core/repositories/AbstractExecutionRepositoryTest.java
PaulPio: Thanks for the review,  I replaced the blind Thread.sleep in dailyStatisticsEuropePragueMidnight with an Awaitility-based polling wait so the test waits deterministically for the DB to reflect saved executions instead of sleeping.

What I changed:

Replaced the sleep in AbstractExecutionRepositoryTest.java with:
An Awaitility await() that polls up to 2s (50ms intervals) and returns once the repository query returns dates matching the Prague local dates we created.
Added Awaitility as a test dependency to build.gradle so the concrete H2 test runtime can load Awaitility. (Because tests run in each concrete module classpath, the jdbc-h2 module needs Awaitility at test runtime too.)

PaulPio:
good afternoon, have you good time to take a look at the changes that i put as you requested? @loicmathieu

MilosPaunovic:
There is merge conflict in this PR, that needs to be amended first @PaulPio, before anyone takes a look at it.

core/src/test/java/io/kestra/core/repositories/AbstractExecutionRepositoryTest.java
```diff
@@ -498,6 +500,67 @@ protected void dailyStatistics() throws InterruptedException {
...
+            .build();
+
+        executionRepository.save(before);
+        executionRepository.save(after);
+
+        ZonedDateTime start = dayBeforeMidnight.truncatedTo(ChronoUnit.DAYS);
+        ZonedDateTime end = dayAfterMidnight.plusDays(1).truncatedTo(ChronoUnit.DAYS);
+
+        // Wait up to 2 seconds for DB to reflect saved executions using Awaitility
+        List<DailyExecutionStatistics> result = Awaitility.await()
```
loicmathieu, line 549: Saving in DB is synchronous so it should not be needed.
Does it really pass without it? Something strange should happen in this case

loicmathieu (changes requested):

core/src/test/java/io/kestra/core/repositories/AbstractExecutionRepositoryTest.java
PaulPio: you suggest i should remove that part for the test @loicmathieu?

loicmathieu: Yes please

core/build.gradle
```diff
@@ -87,4 +87,5 @@ dependencies {
     testImplementation "org.bouncycastle:bcpkix-jdk18on"
 
     testImplementation "org.wiremock:wiremock-jetty12"
+    testImplementation "org.awaitility:awaitility:4.3.0"
```
loicmathieu, line 90: we have ou own await utility, please use `io.kestra.core.utils.Await`  instead.

package-lock.json
loicmathieu, line 1: Please remove this, there is no need to change this file

loicmathieu (changes requested):

MilosPaunovic:
Hey @PaulPio, are there any updates on this?

MilosPaunovic:
Hi @PaulPio,

As there hasn't been any progress or updates on this, I'll close this `PR`.

However, if you do decide to work on it again, feel free to raise another pull request which will be reviewed, if no one else handles it in the meantime.

PR #11898, opened 2025-10-09
feat(filter): initial rework of Filter UI (Executions)

Piyush-r-bhaskar:
- Periodic Refresh - functions not done yet
- Drag and drop for table column / filter in popper is not yet available.
- Where Multiselect is not working , it's because we don't have that Operator available currently - for later changes.

For frontend-- Code will be organized way better in final PR, not that much of prop drilling, and many more i have in my mind.

### DIFF
--- a/ui/src/components/executions/Executions.vue
+++ b/ui/src/components/executions/Executions.vue
@@ -20,9 +20,9 @@
                     <li>
                         <TriggerFlow
                             v-if="flowStore.flow"
-                            :disabled="flowStore.flow.disabled || isReadOnly"
-                            :flowId="flowStore.flow.id"
-                            :namespace="flowStore.flow.namespace"
+                            :disabled="flowStore.flow?.disabled || isReadOnly"
+                            :flowId="flowStore.flow?.id"
+                            :namespace="flowStore.flow?.namespace"
                         />
                     </li>
                 </template>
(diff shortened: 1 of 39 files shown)


anna-geller:
Quick feedback:
- [ ] Descriptions are missing in the columns view
image
image

- [ ] wrong design is currently used for "Save applied filters" and "Show data options" - when in doubt between Figma and prototype, use https://filters42.replit.app/. Figma is only for colors (even though they should all match in the prototype)
image

- [ ]  Refresh Data doesn't refresh the table yet

- [ ] on the Saved filters, clicking on the filter will apply it, I think there's no ned for the Play icon (I was in fact wondering whether the Play icon would do something different than clicking on the filter to apply it)
image

- [ ] remove the border
image

- [ ] "Add filters" panel uses outdated design from Figma instead of the prototype, please let's fix it
image

- [ ] Wrong icon with wrong arrow direction used on the "Reset all filters" and wrong text for "Add filters"
image

- [ ] ideally, when adding a filter, the UI should jump to allow configuring it - not urgent one but I believe this is a better UX 
- [ ] State filter needs more love, compare the design to the current state: you should be able to click anywhere in the dropdown to select a given option rather than having to click on the checkbox at the end. Also let's fix the in/not in + spacing (too wide atm, this could be narrower as the states have fixed width)
image

- [ ] The saved filters and "Show data options" should be always in the same position, they shouldn't move to the next row
image

- [ ] in contrast to the above, applied filters should move to the next line instead of overflowing the first row - match what has been done on the prototype https://filters42.replit.app/ 
image

- [ ] Labels dropdown doesn't show available labels, maybe a BE subject but flagging as currently a strange info box shows up instead 
image

- [ ] for Labels, we will need more filter operators but I'm aware maybe you keep this for later once they are added on the BE
image

I think this will do for the initial round, let's address the listed issues first and later I'll do the next round of QA 💪

Piyush-r-bhaskar:
Closing in favour of #12265

PR #11902, opened 2025-10-09
fix(dependencies): Fixed blinking on flow dependencies

### LINKED ISSUE #11125: Avoid blinking on flow dependencies

prayagtushar:

This PR fixes the **loading flicker** observed on the **Dependencies** page and improves the **error handling** logic.

**Problem:**
When navigating to the dependencies page, users saw a brief flash where the dependency table appeared for a split second before showing the “no dependencies” empty state.

**Root Causes:**

* `isLoading` initialized as `true` with `elements = []`, causing an intermediate empty table render.
* Missing error handling could leave `isLoading = true` indefinitely on API failures (infinite spinner).

**Solution:**

* Improved template logic to handle loading, empty, and populated states more predictably:

  ```vue
  <div v-if="isLoading" v-loading="true" />
  <Empty v-else-if="!getElements().length" />
  <el-splitter v-else class="dependencies" />
  ```
* Added `try/catch/finally` around API calls to ensure `isLoading` is always reset:

  ```ts
  try {
    // API calls
  } catch (error) {
    // handle error
  } finally {
    isLoading.value = false;
  }
  ```
* Added handling for missing container edge case.

**Result:**
✅ Smooth transition from loading → empty/data state
✅ No flicker during page render
✅ Prevents infinite spinners on failed API calls
✅ Improved user experience and reliability

---

### Files Changed

* `ui/src/views/flows/Dependencies.vue`
* `ui/src/components/dependencies/composables/useDependencies.ts`

closes #11125
---

https://github.com/user-attachments/assets/4fe31bb5-b298-4a4c-9882-60b06d1eb568

---


MilosPaunovic (changes requested):
Hi @sandyp025,

This is not good, as now we don't show dependencies where we should.

Try to import this `.zip` with flows, and try to see `Dependencies` tab from one of the imported flows and you''ll see what I mean.

[flows.zip](https://github.com/user-attachments/files/22795130/flows.zip)

prayagtushar:
Hi @MilosPaunovic  I understand the issue. I’ll work on it and get back to you soon.

Sandipmandal25:
> Hi @sandyp025,
> 
> This is not good, as now we don't show dependencies where we should.
> 
> Try to import this `.zip` with flows, and try to see `Dependencies` tab from one of the imported flows and you''ll see what I mean.
> 
> [flows.zip](https://github.com/user-attachments/files/22795130/flows.zip)

@MilosPaunovic i think there is a mismatch ! you tagged me here i think there is confusion anyways thanks!

MilosPaunovic:
Ahh, yes, @sandyp025, sorry for that one, I've been reviewing so much of your PRs these days, I got confused 🤣

MilosPaunovic (approved):
Apart from some really minor tweaks I've pushed, this is very good work @Prayag-09! 🚀

prayagtushar:
Thank you for the opportunity, @MilosPaunovic

MilosPaunovic:
Just a small note - I've had to add some more checks for testing to be able to work, as part of another PR:
https://github.com/kestra-io/kestra/commit/edc4abc80eadf9651b4b82f530c1b07b7b00a4ca

See changes in `ui/src/components/dependencies/Dependencies.vue` file.

PR #11907, opened 2025-10-09
  feat(ui): Add multiselect filter for blueprint tags

### LINKED ISSUE #11533: Change the Blueprints Tags filter to be multiselect
### Feature description

After having selected one tag, the user should be able to select another tag to filter for blueprints with all matching tags at the same time. For example, the user could select "Getting Started" and then drill down to find all "Getting Started" blueprints about AI by also selecting the "AI" filter. 

Image

Sandipmandal25:
fixes #11533 

## summary

  - Enables users to select multiple blueprint tags simultaneously for more precise filtering
  - Clicking a tag toggles its selection (blue = selected, gray = unselected)
  - Blueprints are filtered to show only those matching ALL selected tags
  - Selected tags persist in URL for shareable links and browser back/forward navigation

  ## changes

  **Frontend (`BlueprintsBrowser.vue`):**
  - Changed tag selection UI from radio buttons to clickable toggle buttons
  - Updated state from single `selectedTag` to array `selectedTags`
  - Modified query logic to send tag array to backend API
  - Added CSS styling for selected/hover states

  **Backend:**
  No changes needed as API already supports `List<String> tags` parameter


ui/src/components/executions/Metrics.vue
MilosPaunovic, line 1: Please revert all changes in this file, as I'm guessing that's left from another PR, but it's not relevant to this one.

ui/src/components/executions/Vars.vue
MilosPaunovic, line 1: Please revert all changes in this file, as I'm guessing that's left from another PR, but it's not relevant to this one.

MilosPaunovic (changes requested):
Hi @sandyp025,

After reverting changes in two mentioned files, I'll give this a proper review.

ui/src/components/executions/Vars.vue
Sandipmandal25: reverted that files thanks!

MilosPaunovic: When there is a conversation like this from the review, it would be super helpful if, after solving the mentioned problem and adding a comment (which is optional) you click on the `Resolve conversation` button just below this, as that signals maintainers that the issue at hand is resolved.

MilosPaunovic (changes requested):
Hi @sandyp025,

I see no tags at all anymore:

image

I'm I missing something here?

Sandipmandal25:
Hi @MilosPaunovic 

I found the issue - there was a `ready` check in the condition that was preventing the tags from displaying. I've pushed a fix that removes this blocking condition. The tags should now be visible.

MilosPaunovic (changes requested):
Hi @sandyp025,

Okay ,we're getting there.

Now I see the tags, but they are looking differently than before:

image

We want to keep the same look, just enhance the functionality by adding multiple selection to it.

Sandipmandal25:
Hi @MilosPaunovic 

  I've fixed the styling issue! The problem was that I had replaced the radio button group with regular buttons.

  I've now reverted the code and made minimal changes - just switching from `el-radio-group` to `el-checkbox-group` 
(and  `el-radio-button` to `el-checkbox-button`). These Element Plus components look identical visually but enable multiselect functionality.

  The UI now maintains the exact same appearance as before while supporting multiple tag selection. Users can click multiple tags to filter blueprints that match all selected tags.

MilosPaunovic:
Haven't started this, but we now have `ready` again in the code - will it prevent tags from showing?

Sandipmandal25:
the `ready` condition was already in the original code! I didn't add it.

  The original line 8 was:  
 `<el-radio-group v-if="ready && !system && !embed" v-model="selectedTag" class="tags-selection">
`

  I only changed `el-radio-group` to `el-checkbox-group` and `selectedTag` to `selectedTags`. The `ready &&` condition was already there.

  It won't prevent tags from showing - it just ensures tags render after the data is loaded (controlled by the useDataTableActions composable), which prevents flickering during initial load. This was the original behavior.

MilosPaunovic:
So, with that `ready` parameter I still can not see tags when I visit the `/blueprints` page. How are you testing this @sandyp025?

@Piyush-r-bhaskar Is there maybe something that you know that's related to this `ready` property in `useDataTableActions` which is not working as expected?

Piyush-r-bhaskar:
I think I know that...let me check Milos

Piyush-r-bhaskar:
fixed two issue:

- tags not showing 
- tags selection is not immediately showing the filtered list unless hot reload.

Piyush-r-bhaskar (approved):
LGTM 🚀

Good work on this @sandyp025

Issue #11650, opened 2025-10-01
Selecting docs dropdown forces me to save flow first


paulgrainger85:
### Describe the issue

Sometimes, not always, if I edit a flow then try to expand the dropdown in the documentation I get the following popup in the UI. If I select OK I lose any edits I made. It's annoying, I shouldn't be forced to save my flow just for looking at documentation

<img width="3474" height="434" alt="Image" src="https://github.com/user-attachments/assets/448fd2af-046b-459a-a346-b509b9be69d3" />

Related: https://github.com/kestra-io/kestra/issues/9614 https://github.com/kestra-io/kestra/issues/10615

### Environment

- Kestra Version: 1.0

kawpii:
@loicmathieu Can you please assign me this?

MilosPaunovic:
Absolutely, go for it @kaushalyap! 🚀

MilosPaunovic:
Just bare in mind that this change might need some work in our `ui-libs` repo (https://github.com/kestra-io/ui-libs/) as well.

MilosPaunovic:
Hey @kaushalyap, are there any updates on this?

kawpii:
@MilosPaunovic sorry I didn't had time. If it is urgent can you pleas assign it to someone else.

yashgoyal0110:
@MilosPaunovic 
Assign please, I want to work on it.

MilosPaunovic:
Hi @yashgoyal0110,  

You're already assigned to another issue in our repository, and we usually assign only one at a time to new contributors. Let's make sure to wrap that one up before jumping onto another.  🚀

iitzIrFan:
@MilosPaunovic  sir you can assign this issue to me. I would be happy to work on it.

ayuxsh009:
Hi @MilosPaunovic I'd like to fix the issue. Can you Please Assign this to me?

Anshumancanrock:
Hi @MilosPaunovic , Could you please assign this to me?

MilosPaunovic:
@iitzIrFan, go for it! 🚀 

@ayuxsh009 @Anshumancanrock, if you're eager to contribute, may I suggest that you check our [curated list of good first issues](https://go.kestra.io/contributing) to see if there is something you find interesting enough and think you can tackle independently. It can be further filtered by desired labels to narrow down the results (suggesting to filter by either `area/backend` or `area/frontend` ones, based on your preferences).

iitzIrFan:
[...]

PR #11911, opened 2025-10-09
feat(plugins): enhance documentation request handling to prevent unnecessary reloads

iitzIrFan:

Solves: https://github.com/kestra-io/kestra/issues/11650

Original Problem: When you tried to view documentation while having unsaved changes in your flow, you would get an unwanted "unsaved changes" warning. This was happening because:

1. The documentation panel interactions were being treated as flow-related changes
2. The system couldn't distinguish between:
          - Actually modifying the flow content
          - Just viewing documentation about the flow's components
---

1. In [MultiPanelFlowEditorView.vue](vscode-file://vscode-app/c:/Users/IRFAN/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html):
        `function isTabFlowRelated(element: Tab){
    // Skip marking doc tab as dirty to avoid unsaved changes warning
    if (element.value === "doc") {
        return false;
    }
    return ["code", "nocode", "topology"].includes(element.value)
        || element.value.startsWith("nocode-")
}`

This prevents documentation tabs from affecting the flow's dirty state.

2.  In [EditorWrapper.vue](vscode-file://vscode-app/c:/Users/IRFAN/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html):

       `let result = selectedElement ? getElementFromRange(selectedElement) : undefined;
// Flag this as a documentation request
result = {...result, hash: hash.value, forceRefresh: true, isDocumentationRequest: true};
pluginsStore.updateDocumentation(result);`

Now explicitly flags documentation requests.

3.  In [plugins.ts](vscode-file://vscode-app/c:/Users/IRFAN/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html):

       `async updateDocumentation(pluginElement?: ({
    type: string, 
    version?: string, 
    forceRefresh?: boolean, 
    isDocumentationRequest?: boolean
} & Record<string, any>)) {
    // Special handling for documentation requests
    if (isDocumentationRequest && this.editorPlugin?.cls === type && !forceRefresh) {
        return;
    }
    // ... rest of the function
}`

Adds special handling for documentation requests.

---

To Verify the Fix: You can test the following scenarios:

1. Make some changes to a flow but don't save

2. Try to:

     - View documentation for different tasks
     - Switch between documentation and editor panels
     - Browse different sections of documentation

3. You should NOT see any unsaved changes warnings

4. However, if you try to navigate away from the flow editor with actual unsaved changes, you should still get the warning

---


ui/src/components/flows/MultiPanelFlowEditorView.vue
```diff
@@ -45,6 +45,10 @@
     import MultiPanelGenericEditorView from "../MultiPanelGenericEditorView.vue";
 
     function isTabFlowRelated(element: Tab){
+        // Skip marking doc tab as dirty to avoid unsaved changes warning when viewing documentation
+        if (element.value === "doc") {
+            return false;
+        }
```
elevatebart: If `element.value === "doc"` it will not be included in `["code", "nocode", "topology"]` nor start with `nocode-` and will return false anyway.

```suggestion
```

ui/src/stores/plugins.ts
```diff
@@ -291,14 +291,19 @@ export const usePluginsStore = defineStore("plugins", {
...
                 return;
             }
 
-            const {type, version, forceRefresh = false} = pluginElement;
+            const {type, version, forceRefresh = false, isDocumentationRequest = false} = pluginElement;
+            
+            // If this is just a documentation request and we already have the documentation loaded, don't reload
+            if (isDocumentationRequest && this.editorPlugin?.cls === type && !forceRefresh) {
+                return;
+            }
```
elevatebart: What other kinds of requests could happen in an `updateDocumentation` function?

elevatebart (changes requested):
Hello @iitzIrFan 

This problem is very complicated. Do not feel bad if you did not fix it already.
Helping us to find a reproduction is already a big help.

The true problem is that on certain browsers, changing the hash is interpreted as a location change that would trigger a loss of data. Hence the annoying warning. This hash change is initiated by the `Collapsible.vue` component in ui-libs. 

The solution is therefore more simple than what you did here. Simply upgrading the ui-libs to 0.0.255 should do the trick.

ui/src/stores/plugins.ts
iitzIrFan: @elevatebart had added comments in this.

iitzIrFan:
Should I revert the changes made in the following files since I’ve updated ui-libs to version 0.0.255?

- ui/src/components/inputs/EditorWrapper.vue
- ui/src/stores/plugins.ts

and thanks for guiding me there :) @elevatebart

ui/src/stores/plugins.ts
elevatebart: If `pluginUpdates` happen in a `updateDocumentation` it is at least confusing.
I think it would be a "design bug" on the existing code.

If it is an improvement, I don't see how it is connected to the ticket. 
I will rremove it from this PR. 

Can you kindly open another PR explaining a use case of what this change fixes?

elevatebart (approved):

Issue #11913, opened 2025-10-09
Add annotation and modify CI to play flaky, but don't fail on error


nkwiatkowski:
- Add a new annotation to use in tests to flag a test as flaky.
 - play the flaky tests, but ignore the errors
 - display in the build report the flaky tests that passed and those who failed

AbhiR-03:
Hey @nkwiatkowski  can you assign this to me ?

MilosPaunovic:
Hey @AbhiR-03,

There is already a PR for this one, but if you're willing to contribute, may I suggest that you check our [curated list of good first issues](https://go.kestra.io/contributing) to see if there is something you find interesting enough and think you can tackle independently. It can be further filtered by desired labels to narrow down the results (suggesting to filter by either `area/backend` or `area/frontend` ones, based on your preferences).

PR #11914, opened 2025-10-09
feat(tests): add annotation and modify CI to handle flaky tests, play…

nkwiatkowski:
… them but not fail

 - Fix #https://github.com/kestra-io/kestra/issues/11913

---

---


AcevedoR (approved):

tests/src/main/java/io/kestra/core/junit/annotations/FlakyTest.java
```diff
@@ -4,11 +4,18 @@
...
  */
-@Target(ElementType.METHOD)
-@Retention(RetentionPolicy.SOURCE)
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+@Tag("flaky")
 public @interface FlakyTest {
+
+    /**
+     * Use to excplain why the test is flaky
```
brian-mulier-p:
```suggestion
     * Use to explain why the test is flaky
```

tests/src/main/java/io/kestra/core/junit/annotations/FlakyTest.java
```diff
@@ -4,11 +4,18 @@
...
+import org.junit.jupiter.api.Tag;
 
 /**
  * used to document that a test is flaky and needs to be reworked
  */
-@Target(ElementType.METHOD)
-@Retention(RetentionPolicy.SOURCE)
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+@Tag("flaky")
```
brian-mulier-p, line 14: Maybe it should also add a @Test. Or rename the annotation to Flaky and not FlakyTest to avoid confusion idk

brian-mulier-p (changes requested):

tests/src/main/java/io/kestra/core/junit/annotations/FlakyTest.java
nkwiatkowski: I'm not sure about that. It can be use on a ParamatrizedTest or a RetryingTest, so fixing the @Test will prevent that

PR #11917, opened 2025-10-09
fix(core): keep web up during graceful shutdown

### LINKED ISSUE #11817: Keep metrics endpoint up during graceful long shutdown
### Feature description

Hey folks, we run multiple types of flows in Kestra and we have configured Kestra to wait for up to 48 hrs for all threads / flows / processes to exit cleanly before shutting down by setting this in the config:

```
kestra:
  server:
    terminationGracePeriod: 48h
    workerTaskRestartStrategy: IMMEDIATELY
```

This is working as expected, however we noticed that we are getting alerts regarding the `up` metric being down during these shutdowns, which means our scraper keeps trying to scrape metrics from the metrics endpoint but fails.

Can we have the metrics listener (8081 by default) to shutdown *only after* the complete Kestra shutdown?

yuri1969:

The `@PreDestroy` hook is triggered very late in the lifecycle.

Started the worker's graceful cleaning process using the global `ShutdownEvent` which fires way earlier.

I was on the fence with removing the worker's `@PreDestroy` hook.

closes #11817

---

```bash
watch -n 1 'curl http://localhost:8081/prometheus'
```

Running the standalone server, executing the following flow and then quitting the server:

```yaml
id: crane_325035
namespace: company.team

tasks:
  - id: sleep
    script: |
      trap 'echo "nope"' INT
      sleep 60
```


loicmathieu (commented):
Shouldn't the `@PreDestroy` annotation be removed from the Worker?
Anyway close() is safe to be used multiple time but as we move the closing of the Worker here it may be a good idea to make it clear (and also add a comment inside the Worker itself).

@fhussonnois WDYT?

fhussonnois (changes requested):
Hey @yuri1969, thank you for your contribution. I'm OK with the use of an `ApplicationEventListener` to manage this issue. However, the proposed solution is not ideal for me. We should not explicitly reference the worker, because it will only work for standalone. In addition, I think the HTTP server should be keep up for any services. 

So maybe we could just delay the shutdown of the embedded netty server while all services are not closed, with something like this: 

_=> This needs to be tested:_ 
 ```java
@Singleton
@Slf4j
@Order(Ordered.LOWEST_PRECEDENCE)
@Requires(property = "kestra.server-type")
public class GracefulEmbeddedServiceShutdownListener implements ApplicationEventListener<ServerShutdownEvent> {
    
    @Inject
    ServiceRegistry serviceRegistry;
    
    @Override
    public void onApplicationEvent(ServerShutdownEvent event) {
        List<LocalServiceState> states = serviceRegistry.all();
        if (states.isEmpty()) {
            return;
        }
        
        List<CompletableFuture<Void>> futures = states.stream()
            .map(state -> CompletableFuture.runAsync(() -> closeService(state), ForkJoinPool.commonPool()))
            .toList();

        // Wait for all services to close, before shutting down the embbeded server
        CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).join();
    }
    
    private void closeService(LocalServiceState state) {
        final Service service = state.service();
        try {
            service.unwrap().close();
        } catch (Exception e) {
            log.error("[Service id={}, type={}] Unexpected error on close", service.getId(), service.getType(), e);
        }
    }
}
```

Then, it's OK to keep `@PreDestroy` on services because the `close` method must be safe if called twice.

yuri1969:
@fhussonnois Thanks, I've tested the `ServiceRegistry`-based version of yours on both standalone and separated instances.

It does not work as intended when listening to the `ServerShutdownEvent` event (fired when the `EmbeddedServer` shuts down) but it does with the `ShutdownEvent` event (fired prior to starting shutdown sequence).

fhussonnois (approved):
Thank you a lot @yuri1969.

PR #11929, opened 2025-10-10
Fix: openapi tweaks

Skraye:
* added some @ApiResponse annotation 
* added nullable annotation for TaskRun class
* added new prop in openapi.properties

companion: https://github.com/kestra-io/kestra-ee/pull/5421

### DIFF
--- a/core/src/main/java/io/kestra/core/models/dashboards/charts/Chart.java
+++ b/core/src/main/java/io/kestra/core/models/dashboards/charts/Chart.java
@@ -12,6 +12,8 @@ import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.experimental.SuperBuilder;
 
+import static io.kestra.core.utils.RegexPatterns.JAVA_IDENTIFIER_REGEX;
+
 @SuperBuilder(toBuilder = true)
 @Getter
 @NoArgsConstructor
@@ -26,7 +28,7 @@ public abstract class Chart<P extends ChartOption> implements io.kestra.core.mod
 
     @NotNull
     @NotBlank
-    @Pattern(regexp = "^[A-Za-z_$][A-Za-z0-9_$]*(\\.[A-Za-z_$][A-Za-z0-9_$]*)*$")
+    @Pattern(regexp = JAVA_IDENTIFIER_REGEX)
     protected String type;
 
     @Valid
--- a/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
+++ b/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
@@ -42,30 +42,36 @@ public class TaskRun implements TenantInterface {
     @NotNull
     String taskId;
 
+    @Nullable
     String parentTaskRunId;
 
     @With
+    @Nullable
     String value;
 
     @With
+    @Nullable
     List<TaskRunAttempt> attempts;
 
     @With
     @JsonInclude(JsonInclude.Include.ALWAYS)
+    @Nullable
     Variables outputs;
 
     @NotNull
     State state;
 
     @With
+    @Nullable
     Integer iteration;
 
     @With
+    @Nullable
     Boolean dynamic;
 
     // Set it to true to force execution even if the execution is killed
-    @Nullable
     @With
+    @Nullable
     Boolean forceExecution;
 
     @Deprecated
@@ -217,7 +223,7 @@ public class TaskRun implements TenantInterface {
     public boolean isSame(TaskRun taskRun) {
         return this.getId().equals(taskRun.getId()) &&
             ((this.getValue() == null && taskRun.getValue() == null) || (this.getValue() != null && this.getValue().equals(taskRun.getValue()))) &&
-            ((this.getIteration() == null && taskRun.getIteration() == null) || (this.getIteration() != null && this.getIteration().equals(taskRun.getIteration()))) ;
+            ((this.getIteration() == null && taskRun.getIteration() == null) || (this.getIteration() != null && this.getIteration().equals(taskRun.getIteration())));
     }
 
     public String toString(boolean pretty) {
@@ -249,7 +255,7 @@ public class TaskRun implements TenantInterface {
      * This method is used when the retry is apply on a task
      * but the retry type is NEW_EXECUTION
      *
-     * @param retry Contains the retry configuration
+     * @param retry     Contains the retry configuration
      * @param execution Contains the attempt number and original creation date
      * @return The next retry date, null if maxAttempt || maxDuration is reached
      */
@@ -270,6 +276,7 @@ public class TaskRun implements TenantInterface {
 
     /**
      * This method is used when the Retry definition comes from the flow
+     *
      * @param retry The retry configuration
      * @return The next retry date, null if maxAttempt || maxDuration is reached
      */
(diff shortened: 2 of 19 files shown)


core/src/main/java/io/kestra/core/models/dashboards/charts/Chart.java
```diff
@@ -26,7 +26,7 @@ public abstract class Chart<P extends ChartOption> implements io.kestra.core.mod
 
     @NotNull
     @NotBlank
-    @Pattern(regexp = "\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*(\\.\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*)*")
+    @Pattern(regexp = "^[A-Za-z_$][A-Za-z0-9_$]*(\\.[A-Za-z_$][A-Za-z0-9_$]*)*$")
```
brian-mulier-p: Can you create a constant to hold this pattern somewhere?

core/src/main/java/io/kestra/core/models/flows/AbstractFlow.java
```diff
@@ -61,18 +61,30 @@ public abstract class AbstractFlow implements FlowInterface {
...
+            Map.class
+        }
     )
     @Valid
     List<Label> labels;
 
-    @Schema(additionalProperties = Schema.AdditionalPropertiesValue.TRUE)
+    @Schema(
+        type = "object",
+        additionalProperties = Schema.AdditionalPropertiesValue.FALSE
```
brian-mulier-p, line 75: Seems weird, you don't describe any property so it essentially means that the object must remain empty no?

core/src/main/java/io/kestra/core/models/flows/AbstractFlow.java
```diff
@@ -61,18 +61,30 @@ public abstract class AbstractFlow implements FlowInterface {
...
     Map<String, Object> variables;
 
+
+    @Schema(
+        oneOf = {
+            String.class,  // Corresponds to 'type: string' in OpenAPI
+            Map.class      // Corresponds to 'type: object' in OpenAPI
+        }
+    )
+    interface StringOrMapValue {}
```
brian-mulier-p: Extract that out of the AbstractFlow class

core/src/main/java/io/kestra/core/models/flows/Flow.java
```diff
@@ -61,6 +61,11 @@ public boolean hasIgnoreMarker(final AnnotatedMember m) {
             }
         });
 
+
+    @Schema(
+        type = "object",
+        additionalProperties = Schema.AdditionalPropertiesValue.FALSE
```
brian-mulier-p, line 67: same as above

core/src/main/java/io/kestra/core/models/flows/PluginDefault.java
```diff
@@ -21,6 +22,14 @@ public class PluginDefault {
     @Builder.Default
     private final boolean forced = false;
 
+    @Schema(
```
brian-mulier-p, line 25: I doubt it can be a String considering it's a Map typed property

brian-mulier-p (changes requested):
Almost LGTM

core/src/main/java/io/kestra/core/models/flows/AbstractFlow.java
Skraye: Due to the Python generator, we need to consider this as an `object` because `Map<String, Object>` is interpreted as `Dict[Str,Dict[Str,Any]` instead of `Dict[Str, Any]`

core/src/main/java/io/kestra/core/models/flows/Flow.java
Skraye: Due to the Python generator, we need to consider this as an object because Map<String, Object> is interpreted as Dict[Str,Dict[Str,Any] instead of Dict[Str, Any]

core/src/main/java/io/kestra/core/models/executions/TaskRun.java
```diff
@@ -42,30 +42,36 @@ public class TaskRun implements TenantInterface {
     @NotNull
     String taskId;
 
+    @Nullable
```
loicmathieu: Shouldn't it the default?
We have thousands of attributes it would be hard to add this annotation everywhere

core/src/main/java/io/kestra/core/models/flows/AbstractFlow.java
brian-mulier-p: Yes but the additionalProperties = Schema.AdditionalPropertiesValue.FALSE seems to be wrong cause it says it can't have any other property than those defined? I understand for the type = "object" though

core/src/main/java/io/kestra/core/models/executions/TaskRun.java
Skraye: Yes mb over did it, only the outputs need it because of `@JsonInclude(JsonInclude.Include.ALWAYS)` making it required I believe (so I need to allows nullable value)

webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
```diff
@@ -279,21 +270,21 @@ public HttpResponse<FlowWithSource> createFlow(
...
+//    @Operation(tags = {"Flows"}, summary = "Create a flow from json object", deprecated = true)
+//    @Deprecated(forRemoval = true, since = "0.18")
+//    @Hidden // we hide it otherwise this is the one that will be included in the OpenAPI spec instead of the YAML one.
+//    public HttpResponse<Flow> createFlowFromJson(
+//        @RequestBody(description = "The flow") @Body Flow flow
+//    ) throws ConstraintViolationException {
+//        log.warn(WARNING_JSON_FLOW_ENDPOINT);
+//
+//        return HttpResponse.ok(doCreate(parseFlowSource(flow.sourceOrGenerateIfNull())).toFlow());
+//    }
```
brian-mulier-p: I believe you need to uncomment it now that you generated the spec from the Slack discussion right?

webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
```diff
@@ -331,44 +322,44 @@ public List<FlowInterface> updateFlowsInNamespace(
...
+//                } catch (FlowProcessingException e) {
+//                    if (e.getCause() instanceof ConstraintViolationException cve) {
+//                        throw cve;
+//                    } else {
+//                        throw e;
+//                    }
+//                }
+//            }))
+//            .toList();
+//    }
```
brian-mulier-p: same here

webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
```diff
@@ -479,30 +471,30 @@ public HttpResponse<FlowWithSource> updateFlow(
...
+//
+//        Optional<Flow> existingFlow = flowRepository.findById(tenantService.resolveTenant(), namespace, id);
+//        if (existingFlow.isEmpty()) {
+//            return HttpResponse.status(HttpStatus.NOT_FOUND);
+//        }
+//
+//        GenericFlow genericFlow = parseFlowSource(flow.sourceOrGenerateIfNull());
+//
+//        return HttpResponse.ok(updateFlow(genericFlow, existingFlow.get()).toFlow());
+//    }
```
brian-mulier-p: same here

brian-mulier-p (approved):
LGTM

PR #11941, opened 2025-10-10
ci: make sonar fetch coverage on correct path in CI

AcevedoR:


AcevedoR:
report is moved in CI: https://github.com/kestra-io/actions/blob/867c61826ea364ad5fe558cca66b1c310d28e58f/composite/report-java/action.yml#L72-L84

PR #11946, opened 2025-10-10
fix: generate-task-trigger-id

### LINKED ISSUE #8921: Task Id or Trigger should be optional for the end users
The goal was more to aligned with github actions way to work where id are required only for reference. If no reference, id is not required. As user, id as useless without pebble expression, only the backend required one.

But before we should have a proper specs, and we need to analyze the impact, maybe it could be a goal for 2.0.
The impact could be big, some idea where no people don't have id, then filled an id to have a reference:
- Trigger will be considered as a new one
- Task replay will have another id and will not works 
- UI will have to adapt the display of the gantt (where id is the only information visible)
- UI no code probably rely on the id to target the good tasks
- Probably a lot more area impacted. 

Idea of solution, have an `internalId` (not visible for the end user) that will be used by the executor, and the id defined by end users is only used for pebble ?
One side advantage : we don't expose on the `runContext` outputs if there is no id

But since this `internalId` will not be exposed, the front end and probably lots of api endpoint will required adjustement

hemantmm:
closes: #8921

✅ Added automatic ID generation for tasks and triggers in the No-Code editor when the user does not specify an ID.
✅ IDs follow the pattern "task1", "task2", "trigger1", etc., incrementing based on existing elements.
✅ Updated TaskEditor to assign an auto-generated ID immediately when a new task/trigger is created or a type is selected.
✅ Ensured backend allows saving flows with tasks/triggers that have no user-specified ID.

---

Fix:  

Updated UI components and logic so that:

- When a user creates a new task or trigger without specifying an ID, the editor generates a unique sequential ID (e.g., "task3", "trigger2").
- The ID is injected into the YAML before saving or updating the flow.
- Validation gracefully handles missing IDs and allows saving.

This improves UX by ensuring all tasks/triggers have valid IDs, even if the user omits them, and aligns with backend requirements.


hemantmm:
@elevatebart, can you review this PR?

elevatebart:
> @elevatebart, can you review this PR?

I will indeed.

elevatebart (changes requested):
Thank you @hemantmm !

This feature could be very useful indeed.

I tested it and the basics of your intent work: When adding a new task, the id is defaulting to a nice `task1`, `trigger2` etc.

Unfortunately, the issue specified that:

> The ask is to make the backend allow saving a flow with tasks or triggers that have no ID

I also found an issue testing the feature as it is today: 
When I want to specify my own id, I start by manually clearing the input. Then the field is filled again with another default value. It makes it hard to work the normal path.

Once you fix the issue, we will see if we want to add a setting to enable/disable it in the Frontend as I could see users being annoyed by it.

Thank you again!

elevatebart:
This now needs updating the Backend tests. 

I would not have any client side generation of ids but this needs to be clarified anyway.
The only visible change I would think of is to remove the systematic requiring of ids on tasks. Ids would be generated backend. 
One would only add Ids if they want to reference one step in another. 

The best code is code not written at all.

hemantmm:
@elevatebart, I have made the changes can you review the PR?

hemantmm:
@elevatebart, I have made the changes. Can you review the PR?

elevatebart:
The BE tests are still broken.

hemantmm:
> The BE tests are still broken.

Will try to resolve it

hemantmm:
Hi @MilosPaunovic ,

I’m encountering a build failure when trying to compile the project locally (and it also happens in CI).
The build fails during :core:compileJava due to missing Lombok annotations like @EqualsAndHashCode, @ToString, and @SuperBuilder.

> Task :core:compileJava
/home/runner/work/kestra/kestra/core/src/main/java/io/kestra/core/models/tasks/Task.java:30: error: cannot find symbol
@EqualsAndHashCode
 ^
  symbol: class EqualsAndHashCode

/home/runner/work/kestra/kestra/core/src/main/java/io/kestra/core/models/triggers/AbstractTrigger.java:29: error: cannot find symbol
@ToString
 ^
  symbol: class ToString

/home/runner/work/kestra/kestra/core/src/main/java/io/kestra/core/models/triggers/AbstractTrigger.java:30: error: cannot find symbol
@EqualsAndHashCode
 ^
  symbol: class EqualsAndHashCode

Execution failed for task ':core:compileJava'.
> java.lang.IllegalArgumentException: The argument does not represent an annotation type: EqualsAndHashCode

Could you please confirm if:

- Is there a specific Lombok or Java version the project currently expects?
- Any recent changes in the build configuration might affect annotation processing?

MilosPaunovic:
Those are more of a questions for @loicmathieu to help us out with.

loicmathieu:
I'm not fan of it being backend generated as it's a no-code thing and no-code is only frontend.
We'll discuss this internally.

loicmathieu:
Sorry @hemantmm this issue was not enought specified and we think it should be handled differently inside our backend. We need to investiguate a bit more of what we want to do for that and someone from the core team would then work on it.

hemantmm:
Ok, @loicmathieu

PR #11954, opened 2025-10-13
Update V1_45__taskrun_submitted.sql

bluesecret:
Add missing breakpoint type


loicmathieu (changes requested):
Hi,
Thanks for your contribution.
Updating a migration is a breaking change, can you instead delete th V1_45 migration and create a V1_46 migration instead?
We allow deletion of migration so it's cleaner to do this way.

bluesecret:
> Hi, Thanks for your contribution. Updating a migration is a breaking change, can you instead delete th V1_45 migration and create a V1_46 migration instead? We allow deletion of migration so it's cleaner to do this way.

Modified, please review

loicmathieu (approved):
LGTM, thanks a lot

PR #11960, opened 2025-10-13
fix(core): allow null val for optional input with default (#11819)

### LINKED ISSUE #11819: Flow with inputs defaults can be null and will use default
Example flow: 
```yaml 
id: buffalo_224980
namespace: company.team

inputs: 
  - type: STRING
    id: test
    required: false
    defaults: "I'm a default"

tasks:
  - id: hello
    message: Hello {{ inputs.test }}! 🚀
```

After pressing the execute button: 

- Remove the content of the input, it will be readd, no way to remove the defaults value 

https://github.com/user-attachments/assets/dfa619c9-a353-442d-b2c8-fe4e04a3afd4

- Even sending no value, the server will autofill with default value 
Image

### more detail 
2 cases are possible: 

1. `defaults` is applied even if the people explicitly passed null,
 a. then required must be true, and we should add a validation that prevent required to be false
 b. the ui is good filling the default on empty value
2. `defaults` is not applied when people passed null when `required=false`,
 a. then the `defaults` is not applied
 b. the ui is wrong, and should not be autofilled

Advice from @anna-geller, the expected version 2.

fhussonnois:
Handles multipart 'null' string as null (no value) for the input

Fixes: #11819

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
@@ -130,7 +130,7 @@ public class FlowInputOutput {
     private Mono<Map<String, Object>> readData(List<Input<?>> inputs, Execution execution, Publisher<CompletedPart> data, boolean uploadFiles) {
         return Flux.from(data)
             .publishOn(Schedulers.boundedElastic())
-            .<AbstractMap.SimpleEntry<String, String>>handle((input, sink) -> {
+            .<Map.Entry<String, String>>handle((input, sink) -> {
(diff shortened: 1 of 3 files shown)


loicmathieu (commented):
I'm not convinced here as now a String input with the string `"null"` would be considered a null input which seems wrong to me and is clearly a BC.

A null input should not have any part.

anna-geller (approved):

fhussonnois:
Replaced by: https://github.com/kestra-io/kestra/pull/12382

PR #11977, opened 2025-10-14
fix(ui): Make global notifications scrollable

### LINKED ISSUE #11924: Notification height issues
I've imported bunch of flows to `Kestra` using the `Import` function, and on success, I've got a notification like this, which was not scrollable.

Image

We should limit the height of the notification and add vertical scroll to it.

It should be handled at a global level for notification, the import one is just an example which helped me find out about the problem

bikashagarwala0:
This PR fixes an issue where notifications with a large amount of content would overflow the screen instead of becoming scrollable.

I've applied a global CSS rule to the notification component to enforce a `max-height` and add `overflow-y: auto`. This ensures that all notifications are now properly constrained and their content remains accessible.

Closes #11924 

### Screenshots

**Before (Overflowing Notification)**

Screenshot from 2025-10-14 02-01-42

**After (Scrollable Notification)**

Screenshot from 2025-10-14 01-59-06

### DIFF
--- a/ui/src/styles/layout/element-plus-overload.scss
+++ b/ui/src/styles/layout/element-plus-overload.scss
@@ -1010,7 +1010,7 @@ form.ks-horizontal {
 
     .el-notification__content {
         text-align: left;
-        max-height: 250px;
+        max-height: 200px;
         overflow-y: auto;
     }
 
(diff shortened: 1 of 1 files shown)


MilosPaunovic (approved):
LGTM! 🚀

Issue #11761, opened 2025-10-06
Allow file detection triggers to react to both new and updated files


anna-geller:
### Feature description

Epic for all issues

- [x] validate if it helps to bring some abstraction at the top level to avoid duplication across plugins

fdelbrayelle:
>  validate if it helps to bring some abstraction at the top level to avoid duplication across plugins

For now, and after last discussion yesterday with core team, let's not add new dependency on core from plugins here. Ideally, we should have a dedicated plugins lib apart from the core for the plugins here. Maybe a cooldown topic.

Ben8t:
@Malaydewangan09 and @fdelbrayelle moved forward on adding core dependencies

PR #11978, opened 2025-10-14
feat(triggers): introduce stateful behaviour for triggers

Malaydewangan09:
- part-of: https://github.com/kestra-io/kestra/issues/11761


fdelbrayelle (approved):

AcevedoR (changes requested):
needs review by @loicmathieu

loicmathieu (changes requested):
I'm not fan of the current design.

You added both an abstract class and an interface, and add default methods on the interface, you generaly add methods on interface because you cannot add them anywhere else.
I would prefer to have methods on the abstract class.

Generally, if you add an abstract class, it's not to add "utility methods" but to implement a shared method with sometime abstract method that would be overriden by subclasses, I have the feeling here that an utility class would be better.

Last but not least, we usually tackle this kind of concern by only providing an interface (to use the same property definitions including any annotation) and a utility class/servie class.
Look at what has been done for `NamespaceFilesInterface` I think it would be better to reuse the same pattern here.

We don't want to have a too complex trigger hierarchy in core and here you add a new flavor of trigger (which is even not the case as it didn't implement any of the bas trigger interface used by the Scheduler).

TLDR: I think it would be better to only have one interface that defines the attributes and one utility class / service that exposes the methods instead of default methods in the interface.

Malaydewangan09:
Alright, makes sense, thanks @loicmathieu !

Quick question: 
Do we want to remove the `AbstractStatefulTrigger`, and use only interface and service class? and then Triggers could define those properties define in interface?
or 
we could have an `AbstractSateFulTrigger` already defined with those properties, and related triggers could just extend 
`AbstractStatefulTrigger` instead of current `AbstractTrigger`, sicne `AbstractStatefulTrigger` already extends `AbstractTrigger`

loicmathieu:
I would rather go with interface + service.

We may rework the hierarchy of tasks and triggers at some point so I would not add a new one specific to something not handled by the Scheduler now.

Malaydewangan09:
Alright, its now interface + Service.
Related triggers implements the new interface and use the service class!

loicmathieu (approved):
LGTM

PR #11979, opened 2025-10-14
feat: add ts and js workers to monaco editor

elevatebart:
When opening a JS or TS namespace file, we get some weird errors about an unknown label. 
That is becuase those workers are not registered.


brian-mulier-p (approved):
LGTM

PR #11984, opened 2025-10-14
fix: keep required fields from schema

### LINKED ISSUE #11983: [No Code] make the `message` field mandatory in `io.kestra.plugin.core.log.Log`
### Describe the issue

Make the `message` field mandatory in `io.kestra.plugin.core.log.Log` because without that field, the flow will have an error, and we can't execute and save it.

Image

Make it like this:

Image

### Reproducibility

Go to the **Flows** tab, click on **Create**, then select No Code and add a task.

### Environment

- Kestra Version: develop

elevatebart:
Closes https://github.com/kestra-io/kestra/issues/11983.

In no code, a little SNAFu was introduced at the same time as the dashboard no-code:

Required fields were not matching the schema anymore...

OOps


ui/src/components/dashboard/components/selector/Selector.vue
```diff
@@ -12,7 +12,7 @@
                     type="primary"
                     :icon="Plus"
                     tag="router-link"
-                    :to="{name: 'dashboards/create', query}"
```
MilosPaunovic, line 15: I have a feeling that this was used for something, but can't figure out what, do you have ideas?

If unused anymore, then all good.

MilosPaunovic (approved):

ui/src/components/dashboard/components/selector/Selector.vue
elevatebart: I will ask florian, it might be something in EE

elevatebart: Since it has nothing to do with this change anyway I should make another PR

PR #12020, opened 2025-10-15
fix(secrets-page): adjust top margin for empty secrets block

akshaywritescode:
This PR fixes an issue where the secrets page of kestra at landing state have margin issue , which leads to the content starting just after navbar.

### Before

image

### After

image


Skraye (approved):
LGTM

Issue #9283, opened 2025-06-03
Values property for input SELECT should be required and not validated if empty


Ben8t:
### Describe the issue

![Image](https://github.com/user-attachments/assets/7bc6e324-21f5-498c-8b1e-d31674ba458c)

### Environment

- Kestra Version: develop

brian-mulier-p:
Larger subject:
I think backend should have an ability to share an id for multiple properties and that id would be displayed in the nocode with a select. The select would show all properties sharing that id.

This mecanism would allow showing on the frontend that you must fill one of the linked properties (validation). Kind of like a "oneOf" but for multiple properties.

Essentially for the select we would add a `@OneOf("value")` to both expression & values properties to tell in the json schema (probably a new custom property "oneOfProperty" or something) that the user must fill at least expression or values.

For now a quick win to solve part of the issue is to add `@Size(min=2)` to tell that if we input an array in `values` property it will mandate at least 2 elements in it. However adding a validation forcing at least values or expression to be filled will not be in place.

Note that we can already do that in the validation by adding a custom validation for a given type but it will not show up in the frontend while ideally we want the frontend to know that it should show a dropdown to user to select which property he wants to fill up.

If too complex to understand we can have a small meeting 👍 
cc @Ben8t @anna-geller

anna-geller:
what you propose sounds OK to me, go for it! frankly, it's quite technical implementation so no need to consult me or Ben here, just go for the solution you consider best to fix the problem 👍

PR #12028, opened 2025-10-15
fix(core): force at least 2 elements in `values` property of SELECT i…

brian-mulier-p:
…nput

part of #9283


nkwiatkowski (approved):
LGTM

PR #12044, opened 2025-10-15
fix(security): do not load OSS BasicAuthService in EE

AcevedoR:
BasicAuthService was loaded by mistake in EE, and it was changed recently in OSS to persist different data


nkwiatkowski (approved):
LGTM

PR #12047, opened 2025-10-15
fix(flows): allow using OSS CLI to validate EE flows

AcevedoR:
- fixes https://github.com/kestra-io/kestra-ee/issues/5455


Skraye (approved):
LGTM

PR #12052, opened 2025-10-15
fix(core): prevent follow endpoint from sending events for invalid executionId

KarthikReddyMaru:
This PR ensures that the /follow SSE endpoint sends the **start event only when the execution id exists**, and returns an appropriate error otherwise.
This prevents invalid execution IDs from triggering partial or hanging responses.

closes kestra-io/kestra-ee#5461

**What changes are being made and why?**

The followExecution endpoint was sending a **start event even when the provided executionId did not exist**, which caused confusing behaviour:

	•	Browser kept loading with no visible error
	•	Curl showed partial data and then terminated with

`curl: (18) transfer closed with outstanding read data remaining`

This happened because the start event was emitted **before** checking if the execution actually existed.
The change moves the start event emission **after** the existence check to ensure a proper error response when the execution is invalid.

⸻

**How the changes have been QAed?**

**Before (Invalid Execution ID):**

```
curl -N -u <USERNAME>:<PASSWORD> http://localhost:8080/api/v1/main/executions/InvalidExecutionId/follow
id: start
data: {"id":"InvalidExecutionId","labels":[],"originalId":"InvalidExecutionId", ...}
curl: (18) transfer closed with outstanding read data remaining

```
**After (Invalid Execution ID):**

```
curl -N -u <USERNAME>:<PASSWORD> http://localhost:8080/api/v1/main/executions/InvalidExecutionId/follow
Unable to find execution InvalidExecutionId

```
**Valid Execution ID (unchanged):**

```
curl -N -u <USERNAME>:<PASSWORD> http://localhost:8080/api/v1/main/executions/<valid-id>/follow
id: start
data: {...}
id: end
data: {...}

```
Now, invalid execution IDs return proper error messages, and SSE behaviour remains consistent for valid executions.

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
@@ -121,7 +121,7 @@ public class ExecutionController {
     @Nullable
     @Value("${micronaut.server.context-path}")
     protected String basePath;
-    
+
     @Inject
     private FlowRepositoryInterface flowRepository;
 
@@ -168,7 +168,7 @@ public class ExecutionController {
 
     @Inject
     private ApplicationEventPublisher<CrudEvent<Execution>> eventPublisher;
-    
+
     @Inject
     private RunContextFactory runContextFactory;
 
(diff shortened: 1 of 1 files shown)


loicmathieu:
@KarthikReddyMaru, unfortunately, it's a tradeoff here.
If the browser close the tab before it receive the start event, there is a risk of connection leak that's why we send a start event immediatly.
In Elasticsearch (EE), the execution may be available with a slight delay, so we need to send the start event then wait for the execution.

I'm afraid the current implementation is "as good as it can" due to these tradeoff

PR #12066, opened 2025-10-16
Feat/remove deprecated

loicmathieu:
Part-of: https://github.com/kestra-io/kestra-ee/issues/3238


jdbc-h2/src/main/resources/migrations/h2/V3_1__remove_templates.sql
```diff
@@ -0,0 +1,22 @@
+DROP TABLE IF EXISTS templates;
```
fhussonnois, line 1: I think all sql scripts should be prefixed with `V2_` otherwise will have conflicts with flyway if two script exist for version V1_46 for example.

fhussonnois (changes requested):

jdbc-h2/src/main/resources/migrations/h2/V3_1__remove_templates.sql
loicmathieu: Unfortunatly it's not possible as V2 is used by EE.
Anhyway, as long as it's not merged to develop we would still be able to update that if needed

loicmathieu: Or we move to V3 in OSS and V4 in EE

Issue #12087, opened 2025-10-16
Convert SwitchView.vue component to TS


MilosPaunovic:
We need `kestra/ui/src/components/inputs/SwitchView.vue` component converted to `TypeScript` with the use of `Composition API`.

```vue
<script setup lang="ts">
   // Converted code here
</script>
```

Vs7sai:
@MilosPaunovic  please assign

Vs7sai:
@MilosPaunovic  done plz check

MilosPaunovic:
After merging https://github.com/kestra-io/kestra/pull/12134, the `SwitchView.vue` component is removed, so closing this one.

Thanks for the efforts on it @Vs7sai, it's much appreciated! 🚀

PR #12115, opened 2025-10-17
Convert SwitchView.vue component to TS

Vs7sai:
What changes are being made and why?
Convert SwitchView.vue component to TS
✅ Converted to <script setup lang="ts">
✅ Replaced Options API with Composition AP
✅ Verified linting passes with no errors
closs #12087

### DIFF
--- a/ui/src/components/executions/Vars.vue
+++ b/ui/src/components/executions/Vars.vue
@@ -24,7 +24,7 @@
 </template>
 
 <script setup lang="ts">
-    import { computed } from "vue"; 
+    import {computed} from "vue"; 
     import Utils from "../../utils/utils";
     import VarValue from "./VarValue.vue";
     import DateAgo from "../../components/layout/DateAgo.vue";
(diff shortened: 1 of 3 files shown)


ui/src/components/executions/Vars.vue
MilosPaunovic, line 1: Changes in this file need to be reverted, as they're not relevant to the issue at hand.

ui/src/components/flows/FlowRun.vue
MilosPaunovic, line 1: Changes in this file need to be reverted, as they're not relevant to the issue at hand.

MilosPaunovic (changes requested):
Hello @Vs7sai,

Before me taking a proper look at the `SwitchView.vue` component, you need to revert changes from the two unrelated files.

Thanks for the efforts so far! 🍀

MilosPaunovic:
Does closing the PR means you should be unassigned from the issue @Vs7sai?

Vs7sai:
@MilosPaunovic   no iam raiseing new  pr with 10 min

PR #12121, opened 2025-10-17
Fix/validate endpoint fix

Skraye:
* Removed the @NotNull on the `disabled` property of the abstractTrigger as its a primitive it can not be null anyway and it was adding this property in the "required"

* In the validate method, we expect a string which is in fact a valid JSON that we will parsed. So I annotated that what has to be send is an object. However Python SDK was trying to send a String load as JSON.


brian-mulier-p (approved):
LGTM

PR #12122, opened 2025-10-17
feat(system): add an administration page for concurrency limit

### LINKED ISSUE #11250: Improve concurrency limit administration by exposing the info from the `concurrency-limit` table
### Feature description

With the JDBC runner, the concurrency limit functionality maintains a per-flow counter of execution running inside the `concurrency-limit` table.
This table is not shown inside the UI.

The issue is that if, for some reason, the counter is not correct in this table, it will not be shown inside the UI, and there will be no way to detect and act on incorrect counter.
As we have a page that shows triggers with their status and allows us to act when something bad happens (unlock manually), we should have the same for the concurrency limit.

~One solution could be to list records inside the `execution_running` table in the flow's concurrency tab instead of from the `executions` table.
We can then either let the user compare what is really running with what is actually running or fetch the status of all running executions to check if they are really running.~

~An alternate solution would be to run a periodic task (each hour?) that would check for each record in RUNNING in this table if there is a RUNNING execution and, if not, self-repair. This would be the first "self-healing" process that we would have, but it may be a good idea to start implementing such a process, and other areas may benefit from it.~

**EDIT**: as this implementation was just updated, I propose for now to list all records from the new `concurrency-limit` table and display them in a new page inside the `Administration` section of the UI.
Then allow updating the counter (the running field) with a big WARNING "Doing this may correct concurrency limit counter so executions may exceed the allowed limit".

This page should only exist in JDBC not Kafka.

loicmathieu:
Closes  #11250

### DIFF
--- a/core/src/main/java/io/kestra/core/services/ConcurrencyLimitService.java
+++ b/core/src/main/java/io/kestra/core/services/ConcurrencyLimitService.java
@@ -3,15 +3,21 @@ package io.kestra.core.services;
 import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.flows.State;
 import io.kestra.core.queues.QueueException;
+import io.kestra.core.runners.ConcurrencyLimit;
 import jakarta.inject.Singleton;
 
 import java.util.EnumSet;
+import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
-@Singleton
-public class ConcurrencyLimitService {
+/**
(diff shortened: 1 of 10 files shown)


anna-geller:
@elevatebart can help on the frontend side

elevatebart:
For design validation:

https://github.com/user-attachments/assets/f463fc3b-063c-4441-adc5-d98d6d9277ca

Skraye (approved):
LGTM

PR #12141, opened 2025-10-18
fix(ui/sidebar): make nav item hovering snackbar darker for better visibility

akshaywritescode:
I noticed that sidebar navigation menu item snackbar not as visible as it should be while hovering. 

### Before

https://github.com/user-attachments/assets/3e8985e2-1552-4b8a-9e42-8f1ecc840357

### After

https://github.com/user-attachments/assets/7d04c5dd-0678-487b-9831-edce8172cc1e

### DIFF
--- a/ui/src/styles/components/sidebar-menu.scss
+++ b/ui/src/styles/components/sidebar-menu.scss
@@ -54,7 +54,7 @@
             }
 
             &:hover, body &_hover {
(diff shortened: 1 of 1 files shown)


MilosPaunovic:
Thanks for the efforts on this @akshaywritescode, but this is by design.

PR #12143, opened 2025-10-18
fix(ui/sidebar): make plugin icon as outline to match icon harmony

akshaywritescode:
In this pr, i make plugin icon as outline to match icon harmony.

# Why was solid before?
There is no outline version of plugin icon available by default on vue material icon library, so i think you guys choose solid version, which looks odd, it seems like it's a selected item of sidebar menu.

# Screenshots
## Before & After Side by side
Screenshot from 2025-10-18 13-30-13

Screenshot from 2025-10-18 13-31-19

# How i fixed it?
icons are eventually svg, so we can edit them like css.

### DIFF
--- a/ui/src/styles/components/sidebar-menu.scss
+++ b/ui/src/styles/components/sidebar-menu.scss
@@ -15,6 +15,13 @@
             }
         }
 
+        // Make Connection icon appear as outline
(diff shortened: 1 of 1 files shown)


MilosPaunovic (approved):

MilosPaunovic:
Thanks! 🚀

PR #12144, opened 2025-10-18
fix(ui): align label form field in no-code editor

### LINKED ISSUE #12038: Editor (No-code): Align and unify label form fields
### Describe the issue

Image

When adding a label in the no-code editor, two fields appear:
- Key
- An empty form field next to it

Currently, both fields don’t have the same height — they should be uniform.
Additionally, the Delete button isn’t vertically centered relative to the form height.

**Actions**:

- [ ] Ensure both input fields have the same height.
- [ ] Vertically center the Delete button with the form fields.

### Environment

- Kestra Version: develop

FreeSpirit0:

---
closes #12038 
Update css to align the form field in no-code editor

---
Screenshot 2568-10-18 at 16 54 14
Screenshot 2568-10-18 at 16 53 55


Piyush-r-bhaskar (changes requested):
Hey,

I don't see you changes seems to solving the issue. Can you check ?

Image

FreeSpirit0:
Thanks @Piyush-r-bhaskar it seems previously I didn't make change from latest develop that's why it's not working, I fixed it. Should be seeing correct result now
image

ui/src/components/no-code/styles/code.scss
```diff
@@ -37,13 +37,16 @@ $code-font-sm: var(--el-font-size-small);
 
 .delete,
 .reorder {
-    height: 32px;
     cursor: pointer;
     padding-left: 0;
     color: $code-gray;
     text-align: right;
 }
```
Piyush-r-bhaskar, line 46: because you removed the height from delete , we will have misaligned delete icon here. can you check ?

Image

Piyush-r-bhaskar (changes requested):

ui/src/components/no-code/styles/code.scss
FreeSpirit0: I put back the height but it seems to still misalign the same way, same result on `develop` branch so I think it a different issue. I also refactor the css so that `.delete` class still the same.

Piyush-r-bhaskar (approved):

PR #12148, opened 2025-10-18
fix(core): handle integer size in chunk Pebble filter

### LINKED ISSUE #12090: Error with chunk object filter
### Describe the issue

Flow:

```
id: chunk
namespace: demo

variables:
  max_items: 2

tasks:
  - id: log
    message: |
      {% set x = [1,2,3,4,5] | chunk(vars.max_items) %} {# workaround: chunk(vars.max_items) * 1 #}
      {{ x }}
```
Error:

`class java.lang.Integer cannot be cast to class java.lang.Long (java.lang.Integer and java.lang.Long are in module java.base of loader 'bootstrap')`

### Environment

- Kestra Version: 0.23.15

hsachdeva9:

This PR fixes a type-casting issue in the `chunk` Pebble filter that caused a runtime error when the `size` argument was an `Integer`. Previously, the filter expected a `Long`, which failed when an `Integer` was provided.  

Changes made:
- Updated `ChunkFilter.java` to accept any `Number` type for the `size` argument.
- Added a runtime check to ensure the `size` argument is a number.

This improves the filter’s robustness and prevents runtime exceptions in user flows.

---

- Verified with the existing `ChunkFilterTest` unit tests → all tests pass.
- Manually tested the filter with integer and long values for `size` in local flows, confirming correct chunking behavior.

> Note: No additional example flow is provided since the existing unit tests already cover the behavior.

---

### Setup Instructions

No special setup is required for this change. The fix applies to the core Pebble filter and can be tested using the existing Kestra test suite.

closes #12090


loicmathieu:
@hsachdeva9 could you add a test with a reproducer from the original issue?

hsachdeva9:
Hi @MilosPaunovic , I’ve added tests that reproduce the original issue reported. Could you please review it and let me know if you need something else.

hsachdeva9:
@MilosPaunovic Any updates on this?

MilosPaunovic:
@loicmathieu Could you take a look at this PR sometime this week?

core/src/test/java/io/kestra/core/runners/pebble/filters/ChunkFilterTest.java
```diff
@@ -38,4 +38,32 @@ void exception() {
...
+            "max_items", 2
+        );
+
+        String render = variableRenderer.render(
+            "{% set x = [1,2,3,4,5] | chunk(max_items) %}{{ x }}", 
+            vars
+        );
+
+        assertThat(render).isEqualTo("[[1,2],[3,4],[5]]");
+    }
```
loicmathieu: This test is more or less the same as before, please remove it

core/src/test/java/io/kestra/core/runners/pebble/filters/ChunkFilterTest.java
```diff
@@ -38,4 +38,32 @@ void exception() {
             }).get();
         });
     }
+    @Test
+    void chunkWithIntegerVariable() throws IllegalVariableEvaluationException {
+        // Reproducer for issue: Integer variable causing ClassCastException
+        Map<String, Object> vars = Map.of(
+            "max_items", 2,  // This is an Integer, not a Long
```
loicmathieu:
```suggestion
            "max_items", Integer.valueOf(2), 
```

Lilke that we are sure that it's and Integer

loicmathieu (changes requested):

core/src/test/java/io/kestra/core/runners/pebble/filters/ChunkFilterTest.java
hsachdeva9: Sure, I will remove this one.

loicmathieu (approved):
LGTM, thanks

PR #12153, opened 2025-10-19
feat(core): store attempts for flowable tasks 

### LINKED ISSUE #11239: Flowable don't store any attempts
For now, flowable tasks don't store any attempyts: 

Image

Neither success, nor [retry](https://github.com/kestra-io/kestra/issues/326).

It's not required internally for the orchestration part but for the observability, would be nice to have a consistent behavior.

MTarek165:
closes #11239
### Description
- Added attempts for flowable tasks like runnable tasks for better observability


MTarek165:
Foe Example 
- Workflow 
id: loop
namespace: company.team

tasks:
  - id: foreach
    values: [2,4]
    tasks:
      - id: first
        values:
          iteration : "{{ taskrun.iteration }}"
- Api Response :
response

MTarek165:
For Example

Workflow
id: loop
namespace: company.team
tasks:

id: foreach
values: [2,4]
tasks:
id: first
values:
iteration : "{{ taskrun.iteration }}"
Api Response :
[
success
](url)

loicmathieu (changes requested):
@MTarek165 thanks for your contribution. I wonder if it makes sense or not for flowable to differentiate created/running as a flowable task didn't run by itself and there would never been a delay before created/running. I'll discuss this internally.

Meanwhile, can you take care of adding assertions in an existing test for example the tests for the `If` task or the `Sequential` task?

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -444,11 +470,10 @@ private Executor handleNext(Executor executor) {
...
         }
-
-        return executor.withTaskRun(
-            this.saveFlowableOutput(nextTaskRuns, executor),
-            "handleNext"
-        );
+return executor.withTaskRun(
+    this.saveFlowableOutput(nextTaskRuns, executor),
+    "handleNext"
+);
```
loicmathieu: Please revert this change

loicmathieu (changes requested):

MTarek165:
@loicmathieu Ok, I have one point :
1- I want to make sure of something in case of retrys, runnable tasks store attempts with the number of retrys but here in case of flowable only one attempt is stored(even if there is retry) since I couldn't see something says that the flowable task itself is retrying, is it ok here?

loicmathieu:
Flowable tasks didn't support retry so nothing more should be done.
In the tests, you should assert that there was an atemp.

MTarek165:
@loicmathieu  Good, I added unit tests for flowable (If and Sequential) and other flowables should follow the same. You can take a look now.

MTarek165:
I forget to name commit correctly it should be feat(tests) not feat(core) but couldn't ammend

MTarek165:
@MilosPaunovic It is about commit ([feat(core): add unit tests for flowable task attempts( If and Sequential)) naming not PR naming :)

MilosPaunovic:
Ahh, my bad, thanks for reverting it back @MTarek165 :)

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -375,6 +387,20 @@ private List<TaskRun> saveFlowableOutput(
...
                     taskRun = taskRun.withOutputs(variables);
+                    /// flowable attempt creation
+                    List<TaskRunAttempt> attempts = Optional.ofNullable(taskRun.getAttempts())
+                        .map(ArrayList::new)
+                        .orElseGet(ArrayList::new);
+
+                    if (taskRun.getState().isCreated()) {
+                        attempts.add(
+                            TaskRunAttempt.builder()
+                                .state(new State().withState(State.Type.CREATED))
```
loicmathieu:
```suggestion
                                .state(new State())
```

loicmathieu (approved):
LGTM

loicmathieu:
@MTarek165 thanks for your contribution, I created https://github.com/kestra-io/kestra/issues/12177 as a followup as we have the same issue for tasks of type `ExecutionUpdatingTask`.

loicmathieu:
@MTarek165 a lot of tests fail with this PR.
You can check them here: https://github.com/kestra-io/kestra/actions/runs/18653754012/job/53177660773?pr=12153
Click on the following part of the Job you'll see a summary
image

MTarek165:
@loicmathieu I checked carefully before there was one test failing related to attempts because of this PR at markAsEachPara() test at ExecutionServiceTest and I fixed it, but Now checked again I found that the remaining failed tests in  core , jdbc-h2 and jdbc-mysql has no relevance with the PR , specially that changes I did doesn't touch anything in the execution flow of the application its regarding observability only.

MTarek165:
Sorry, there is indirect relevance with changes that cause these failures I couldn't understand but I found it exactly occurs because of the place I chose to create the attempt, saveFlowableOutputs() is not the suitable place to create the attempt failures occurs there  . I removed it and run tests again they passed. This is to determine the issue only I must place attempt creation in a place that doesn't affect the execution . I will check it.

MTarek165:
@loicmathieu I checked carefully there was same three tests fails for jdbc-h2, jdbc-mysql and jdbc-postgres and I found that the failure occurs in saveFlowableOutput() when reaching attempt creation (don't know why may be related to the execution life cycle)
I changed to earlier place of the life cycle at Jdbc Executor (now time of creation is same as running) these tests passed !
This change is more reasonable ( just like runnable tasks) and made tests pass.
The two failed tests now is flakey , I tested locally sometimes pass or fail.
Please give a look on the new change and tests if possible

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -247,7 +247,20 @@ private Optional<WorkerTaskResult> childWorkerTaskResult(Flow flow, Execution ex
...
+                     List<TaskRunAttempt> attempts = Optional.ofNullable(parentTaskRun.getAttempts())
+                         .map(ArrayList::new)
+                         .orElseGet(ArrayList::new);
+                     TaskRunAttempt updated = attempts.getLast().withState(state.get());
+                     attempts.set( attempts.size() - 1, updated);
+                     parentTaskRun = parentTaskRun
+                         .withAttempts(attempts)
+                         .withState(state.get());
+                     execution = execution.withTaskRun(parentTaskRun);
+                 }
```
loicmathieu: This should not be done here as the state may not be the terminal state, for ex for ForEach with 100 iterations we arrives here 100  times.

You need to add the attempts to the WorkerTaskResult created for the flowable line 284.

loicmathieu (changes requested):

executor/src/main/java/io/kestra/executor/ExecutorService.java
MTarek165: done

loicmathieu (approved):
LGTM, thanks

MTarek165:
Is it ready to merge ?

loicmathieu:
@MTarek165 no, one test fail

MTarek165:
@loicmathieu I run this test locally and passed
dd

MTarek165:
@loicmathieu It is perfect now 💯

loicmathieu (approved):
LGMT, thanks for fixing the last bits.

PR #12170, opened 2025-10-20
Fea/lock

loicmathieu:


core/src/main/java/io/kestra/core/lock/Lock.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+import java.time.LocalDateTime;
+
+@Getter
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class Lock {
```
fhussonnois: Why don't use a java record here ?

core/src/main/java/io/kestra/core/lock/Lock.java
```diff
@@ -0,0 +1,24 @@
...
+    private String category;
+    private String id;
+    private String owner;
+    private LocalDateTime date;
```
fhussonnois: `created` or `createdAt` is more explicit than just `date`. Personally, I would use an `Instant` for such use case.

core/src/main/java/io/kestra/core/lock/LockException.java
```diff
@@ -0,0 +1,11 @@
+package io.kestra.core.lock;
+
+public class LockException extends Exception {
```
fhussonnois: Kestra specific exception must extend `KestraException`

core/src/main/java/io/kestra/core/lock/LockService.java
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * @implNote There are no expiry for locks so a service may hold a lock infinitely until the service is restarted as the
+ *           liveness mechanism release all locks when the service is unreachable.
+ *           This may be improved at some point by adding an expiry (for ex 30s) and running a thread that will periodically
+ *           increase the expiry for all exiting locks. This should allow quicker recovery of zombie locks than relying on the liveness mechanism,
+ *           as a service wanted to lock an expired lock would be able to take it over.
+ */
+@Slf4j
+@Singleton
+public class LockService {
```
fhussonnois, line 28: I prefer to have a `isLocked` method exposed by this service than having to do `!lockRepositoryInterface.findById(...).isEmpty()`. As a user I don't want to know that the locked mecanism is based on a repository. Also, it's painful to have to inject the repository just to check if something is locked while only  the LockService is needed. This will reduce the coupling with the `LockRepositoryInterface`.

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcServiceLivenessCoordinator.java
```diff
@@ -145,6 +149,12 @@ protected void handleAllNonRespondingServices(Instant now) {
                 log.info("Trigger task restart for non-responding workers after timeout: {}.", workerIdsHavingTasksToRestart);
                 executor.get().reEmitWorkerJobsForWorkers(configuration, workerIdsHavingTasksToRestart);
             }
+
+            // Eventually release all owned locks
+            nonRespondingServices.forEach(instance -> {
+                int released = lockService.releaseAllLocks(instance.uid());
+                log.info("Released {} locks for non-responding service instance {}", released, instance.uid());
```
fhussonnois: For troobleshooting , I'm really wondering if we don't want to log in debug all remaining locks before releasing them ?

fhussonnois (approved):
Approved, minus my comments

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcServiceLivenessCoordinator.java
loicmathieu: Yeah, I had the same question myself, as it's not perf sensitive doing a find then delete is OK.
This has a slight risk of not being accurate but it's OK I think?.

core/src/main/java/io/kestra/core/lock/Lock.java
loicmathieu: We don't use records in our model yet that's why...

core/src/main/java/io/kestra/core/lock/Lock.java
loicmathieu: Agreed, I used createdAt and switch to instant.
Anyway it's for troubleshooting so not a big deal

core/src/main/java/io/kestra/core/lock/LockService.java
loicmathieu: There are no usage of isLocked now but yes, all lock usage should go thought the service.
I'll add the new method but you already forseen usage that are not yet planned as for now it's only to provide lock for the Elasticsearch execution repository ;)

fhussonnois (approved):
LGTM

PR #12181, opened 2025-10-20
performance(jdbc): add deleted field condition when allowDeleted=true

### LINKED ISSUE #12179: [performance] Missing `deleted` field in SQL When `allowDeleted=true` prevents index usage​
### Describe the issue

https://github.com/kestra-io/kestra/blob/f9f3b004d7b569a24b8054547563dcea30c86425/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java#L61
The `defaultFilter` method generates a SQL WHERE clause that ​​excludes the `deleted` field condition​​ when `allowDeleted` is true. This omission prevents the database optimizer from leveraging a composite index (e.g., (tenant_id, deleted)), forcing inefficient full table scans instead of indexed seeks

```sql
CREATE TABLE `executions` (
  `key` varchar(250) COLLATE utf8mb4_unicode_ci NOT NULL,
  `value` json NOT NULL,
  ......
  PRIMARY KEY (`key`),
  KEY `ix_namespace` (`deleted`,`tenant_id`,`namespace`),
  KEY `ix_flowId` (`deleted`,`tenant_id`,`flow_id`),
  KEY `ix_state_current` (`deleted`,`tenant_id`,`state_current`),
  KEY `ix_start_date` (`deleted`,`tenant_id`,`start_date`),
  KEY `ix_end_date` (`deleted`,`tenant_id`,`end_date`),
  KEY `ix_state_duration` (`deleted`,`tenant_id`,`state_duration`),
  KEY `ix_trigger_execution_id` (`deleted`,`tenant_id`,`trigger_execution_id`),
  FULLTEXT KEY `ix_fulltext` (`namespace`,`flow_id`,`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci
```

I'd like to submit a PR to fix it

### Environment

- Kestra Version: develop

lw-yang:

close #12179 
---

exist ut
---

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
@@ -58,7 +58,9 @@ public abstract class AbstractJdbcRepository {
 
     protected Condition defaultFilter(String tenantId, boolean allowDeleted) {
         var tenant = buildTenantCondition(tenantId);
-        return allowDeleted ? tenant : tenant.and(field("deleted", Boolean.class).eq(false));
+        return allowDeleted ?
+            tenant.and(field("deleted", Boolean.class).in(true, false)) :
+            tenant.and(field("deleted", Boolean.class).eq(false));
     }
 
     protected Condition defaultFilterWithNoACL(String tenantId) {
@@ -67,7 +69,9 @@ public abstract class AbstractJdbcRepository {
 
     protected Condition defaultFilterWithNoACL(String tenantId, boolean deleted) {
         var tenant = buildTenantCondition(tenantId);
-        return deleted ? tenant : tenant.and(field("deleted", Boolean.class).eq(false));
+        return deleted ?
+            tenant.and(field("deleted", Boolean.class).in(true, false)) :
+            tenant.and(field("deleted", Boolean.class).eq(false));
     }
 
     protected Condition buildTenantCondition(String tenantId) {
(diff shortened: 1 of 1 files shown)


loicmathieu:
It's surprising that the optimizer didn't use the index if the field is missing and use it if we provide all possible values of the field. Maybe it's a limitation of the MySQL optimizer.

Can you give some numbers before/after your change?

lw-yang:
sure, below are the test cases, the table contains over 2 million records.

When the SQL query ​​excludes the `deleted` condition​​, it takes ​​63 seconds​​ and performs a full table scan.
When the query ​​includes `deleted`​​, it takes ​​8 ms​​ and uses the composite index efficiently.

```sql
select count(*) from `executions`;
```
image

```sql
select * from `executions` where tenant_id is null and deleted in (true,false) and namespace = 'dt.10007';
```
elapsed time: 8ms​​

```sql
select * from `executions` where tenant_id is null  and namespace = 'dt.10007';
```
elapsed time: 63s​​

```sql
explain select * from `executions` where tenant_id is null and namespace = 'dt.15650';
```
image


```sql
explain select * from `executions` where tenant_id is null and deleted in (true,false) and namespace = 'dt.15650';
```
image

---
the introduction to multiple-column-indexes from the official MySQL documentation.
https://dev.mysql.com/doc/refman/8.0/en/multiple-column-indexes.html
image

loicmathieu (changes requested):
Overall, it's OK for me.

Can you add a comment, something like:

> Always include deleted in the query filters as most database optimizers can only use and index if the leftmost columns are used in the query

This would prevent someone to remove this later ;)

I need to check in Postgres if it didn't cause any issue before merging it.
Can you also tell me if you want it to be backported to a specific version, as this is a small change, I can backport it even if we usually don't backport performance improvements.

lw-yang:
I’ve added the comment, we don't need to backport

loicmathieu (approved):
LGTM, thanks

loicmathieu:
Tested it with Postgres and the explain plan seems little better with the new clause :tada:

PR #12204, opened 2025-10-21
chore(core): add a core Disposable interface

fhussonnois:


core/src/main/java/io/kestra/core/utils/Disposable.java
```diff
@@ -0,0 +1,74 @@
...
+                    action.run();
+                }
+            }
+            
+            @Override
+            public boolean isDisposed() {
+                return disposed.get();
+            }
+        };
+    }
```
loicmathieu: I'm not fan of anonymous classes like that, I think there are some performance implication.
Why not adding a private implementation class inside the interface instead? Or a package private implementation if private is not allowed in interface (don't remember).

loicmathieu (approved):
LGMT, minus the anonymous inner class but I can live with it ;)

PR #12205, opened 2025-10-21
fix: load node types for ts files

elevatebart:
also fix some unwanted autocomplete triggering outside of pebble


ui/src/components/inputs/MonacoEditor.vue
```diff
@@ -25,6 +25,75 @@
...
+                monaco.languages.typescript.typescriptDefaults.addExtraLib(
+                    NodeTypesRawContent,
+                    `file://${path}`
+                );
+            }
+            clearInterval(intervalId);
+        } else if(tries-- <= 0) {
+            // only give it a second to try initialization
+            clearInterval(intervalId);
+        }
```
brian-mulier-p, line 51: you rely on the function to throw if Monaco is not yet initialized that's why you have a concept of tries?

ui/src/utils/utils.ts
```diff
@@ -351,3 +351,72 @@ export function deepEqual(x: any, y: any): boolean {
         ok(x).every(key => deepEqual(x[key], y[key]))
     ) : (x === y);
 }
+
+export function getNumberOfNewLinesBetweenIndices(value: string, start: number, end: number): number {
+    // count the number of lines between endSearchIndex and startSearchIndex
+    return end - start // count the number of characters between the two indices
+        - value.substring(start, end).replace(/\n/g, "").length // count the number of new lines
```
brian-mulier-p: why not a simple `.split("\n").length - 1`?

ui/src/utils/utils.ts
```diff
@@ -351,3 +351,72 @@ export function deepEqual(x: any, y: any): boolean {
...
+}
+
+function getColumnFromIndex(value: string, index: number): number {
+    const lastNewLineIndex = value.lastIndexOf("\n", index);
+    return (index - lastNewLineIndex);
+}
```
brian-mulier-p: I don't understand this function please comment or give it a better name

brian-mulier-p (changes requested):
Add a few comments as I said but otherwise LGTM

ui/src/components/inputs/MonacoEditor.vue
elevatebart: Since it is not critical and I do not want it to loop forever, I limit the loading time to one second. If after a second, `monaco.languages.typescript` is still undefined. I give up. 
Now that I think of it, it feels too short.

elevatebart: I updated the timing to wait each time 100ms longer to end after 10 second

ui/src/utils/utils.ts
elevatebart: this function was supposed to get the colmn index from the absolute index in the string.
Confusing and ultimately useless.

elevatebart:
@brian-mulier-p I removed all the complexities from this PR.
It occured to me that the test would be more simple this way.

brian-mulier-p (approved):
LGTM

PR #12206, opened 2025-10-21
Fix ExecutionUpdatableTask attempt state tracking to record proper duration

### LINKED ISSUE #12177: Correctly compute attemps for ExecutionUpdatableTask
### Describe the issue

Currently, task of type `ExecutionUpdatableTask` compute a single instance with the states CREATED/SUCCESS when the task is done.

We should create one attemps with CREATED/RUNNING/SUCCESS and create the atttempt before calling the task and update the attemp with the terminal state after the task execution to correctly record its duration.

The code is inside `ExecutorService.handleExecutionUpdatingTask()`.

### Environment

- Kestra Version: develop

KarthikReddyMaru:
This PR ensures that ExecutionUpdatableTask attempts properly record state transitions (**CREATED → RUNNING → SUCCESS**) with accurate timestamps, enabling correct duration tracking and complete attempt history visibility in the UI.

closes #12177 

## What changes are being made and why?

The `handleExecutionUpdatingTask` method was creating attempt records with only the CREATED and terminal states, skipping the RUNNING state transition. This caused the **RUNNING state to be missing** from attempt history, resulting in incomplete execution timeline in UI.

The fix updates the same attempt object through complete state transitions:

1. Before task execution: transition attempt from **CREATED → RUNNING**
2. After task execution: transition attempt from **RUNNING → terminal state**

## How the changes have been QAed?

**Before** 

Screenshot 2025-10-21 at 6 15 11 PM

*Missing RUNNING state: Only CREATED and SUCCESS states visible*

**After**

Screenshot 2025-10-21 at 6 13 04 PM

*Complete state progression: CREATED → RUNNING → SUCCESS with distinct timestamps*

**Verification:**

- Complete state progression: CREATED → RUNNING → terminal
- UI displays proper attempt timeline
- No regression for skipped tasks (CREATED → SKIPPED)

### DIFF
--- a/executor/src/main/java/io/kestra/executor/ExecutorService.java
+++ b/executor/src/main/java/io/kestra/executor/ExecutorService.java
@@ -1081,22 +1081,9 @@ public class ExecutorService {
                         return false;
                     }
 
-                    List<TaskRunAttempt> attempts = workerTask.getTaskRun().getAttempts();
-                    if (attempts == null || attempts.isEmpty()) {
-                        attempts = new ArrayList<>();
-                        attempts.add(TaskRunAttempt.builder().state(new State()).build());
-                    } else {
-                        attempts = new ArrayList<>(attempts);
-                    }
-
-                    // Update attempt to RUNNING before task execution
-                    int lastIndex = attempts.size() - 1;
-                    TaskRunAttempt runningAttempt = attempts.getLast().withState(State.Type.RUNNING);
-                    attempts.set(lastIndex, runningAttempt);
-
                     TaskRun runningTaskRun = workerTask
                         .getTaskRun()
-                        .withAttempts(attempts)
+                        .withAttempts(List.of(TaskRunAttempt.builder().state(new State().withState(State.Type.RUNNING)).build()))
                         .withState(State.Type.RUNNING);
 
                     executor.withExecution(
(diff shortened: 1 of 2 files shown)


executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -1083,19 +1081,43 @@ private Executor handleExecutionUpdatingTask(final Executor executor) throws Int
...
+                    // Update attempt to RUNNING before task execution
+                    int lastIndex = attempts.size() - 1;
+                    TaskRunAttempt runningAttempt = attempts.getLast().withState(State.Type.RUNNING);
+                    attempts.set(lastIndex, runningAttempt);
+
+                    TaskRun runningTaskRun = workerTask
+                        .getTaskRun()
+                        .withAttempts(attempts)
+                        .withState(State.Type.RUNNING);
+
```
loicmathieu: At this stage, there cannot be any attemps so I think this could be simplified to

```java
TaskRun runningTaskRun = workerTask
                        .getTaskRun()
                        .withAttempts(List.of(TaskRunAttempt.builder().state(new State().withState(State.Type.RUNNING)).build())
                        .withState(State.Type.RUNNING);
```

loicmathieu (changes requested):
Can you also add assertions into an existing test, for ex the `RuntimeLabelsTest`?

KarthikReddyMaru:
Sure, I’ll add the assertions to RuntimeLabelsTest and update the PR shortly.

loicmathieu (changes requested):
LGTM, thanks

loicmathieu (approved):

PR #12207, opened 2025-10-21
fix(ui): amend blueprint page translation key #12193

### LINKED ISSUE #12193: Amend blueprint page translation key
In `kestra/ui/src/override/components/flows/blueprints/BlueprintsBrowser.vue` component, there is currently a translation key which is missing from language files.

The point is, it's wrong, so we need to replace it with the proper one.

So, the task here is to find the search field and to replace it's current placeholder to use the existing key `search blueprint`.

This issue is open only for the first-time contributors to Kestra, as it's quite small and it will help them get a jump into the whole codebase.

ashwini2k6:

Closes #12193
This PR replaces the current placeholder text in the blueprint search bar with the correct, existing translation key `search.blueprint`, as requested in the issue.

### DIFF
--- a/ui/src/override/components/flows/blueprints/BlueprintsBrowser.vue
+++ b/ui/src/override/components/flows/blueprints/BlueprintsBrowser.vue
@@ -23,7 +23,7 @@
                             <el-col :xs="24">
                                 <el-input
                                     v-model="searchText"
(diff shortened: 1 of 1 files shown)


MilosPaunovic (approved):
LGTM! 🚀

PR #12217, opened 2025-10-22
fix(ui): disable id field when editing/ add id as required in dashboard no code editor

### LINKED ISSUE #12194: [No Code Dashboard] The dashboard ID should not be editable once saved
### Describe the issue

The ID section should not be editable once it's saved, just like no-code for flows, because it can't be updated. Plus, put an * asterisk to indicate that it's a mandatory field.

Image

https://jam.dev/c/2ae4ce47-cfb6-4f38-8c42-de963765868 

### Environment

- Kestra Version: develop

FreeSpirit0:

---

closes #12194
In Dashboard No Code Editor
- Add `id` field as required
- Disable `id` input when editing (dashboard already been saved)

---
Creating Dashboard
image

Editing Dashboard
image


elevatebart (approved):
Thank you! LGTM

Maybe we should think of making the same rollback adjustment as we have for flows ids and namespaces.

When they are edited in the code they are immediately rolled back.

PR #12218, opened 2025-10-22
Feat/value from iteration

### LINKED ISSUE #11206: access the value of a previouse iteration in a foreach loop
### Feature description

There is not why to get the value of a previouse iteration of a loop.
If the List is defined in a previouse step, maybe accessing the List direkt with taskrun.iteration can work.
Unfortunately there seams to be a bug in taskrun.iteration because it does not return the iteration number, it returns the count of tasks in the loop. (I will open another request)
Anyway, it would be halbfull to access values of a previouse iteration by using the loop task or a taskrun property.

```
id: each_nested
namespace: company.team
tasks:
  -id: 1_each
    values: '["s1", "s2", "s3"]'
    tasks:
      -id: get_previouse_iteration_value
        format: "{{ outputs.1_each[taskrun.iteration - 1] }}"
```

Anuragsharma15-dell:
closes #11206

feat: add function to access previous iteration value in Kestra loop

- Implemented a workflow 'each_nested' to demonstrate accessing values from a previous iteration in a ForEach loop.
- Added task 'get_previous_iteration_value' using taskrun.iteration to reference the prior iteration's output.
- Highlighted a potential bug: taskrun.iteration returns the count of tasks in the loop instead of the iteration index. Will create a separate issue for this.
- Useful for workflows where outputs of prior loop iterations need to be referenced.
- Register it in Extension.java .


MilosPaunovic:
This PR changes two `Vue` files which are unrelated to the issue, closing.

PR #12267, opened 2025-10-22
feature -> added bulk delete for logs

### LINKED ISSUE #4574: Extend the Logs page with a bulk-delete option
### Feature description

Extend this page:

image

with a similar bulk-action:

image

and a confirmation dialog: "Are you sure you want to delete the selected logs?"

anityasharma11:
closes #4574

Created one api for to bulk delete logs using logs object and other api for deleting logs based on filter if user wants to delete all pages fetched using search filters
And ui changes to support these apis

---

---

### DIFF
--- a/core/src/main/java/io/kestra/core/repositories/LogRepositoryInterface.java
+++ b/core/src/main/java/io/kestra/core/repositories/LogRepositoryInterface.java
@@ -9,6 +9,7 @@ import jakarta.annotation.Nullable;
 import org.slf4j.event.Level;
 import reactor.core.publisher.Flux;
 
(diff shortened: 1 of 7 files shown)


brian-mulier-p:
Closing as this looks AI-generated, the existing method wasn't reused to be enhanced but recreated from scratch + there is no tests +   this is not the way filters are handled on any repository, having 1 parameter per possible filter is not scalable, instead we are passing filters directly to the repository that should do the handling directly. If you're willing to offer a PR please have a look at how the other repositories are handled and avoid as much as possible duplicate methods.

PR #12275, opened 2025-10-22
(Fix: UI) Fixed the line between dots

### LINKED ISSUE #12209: Line should connect the dot properly
Image

Line should be touching the dots.

D-260:
Closes [#12209](https://github.com/kestra-io/kestra/issues/12209)

### What does this PR do?

I have changed the scss code so that the stepper/progress bar component's line touches the dot of progress as mentioned in the issue. I noticed that the width is subtracted from 100% because the line was overflowing into the dots. But the subtraction was a bit too much and was not connecting the dots together. Hence I reduced it in the PR as shown below. 

I hope this gets merged and the UI bug gets solved.

I am attaching a screenshot for reference.
 
image

### DIFF
--- a/ui/src/components/layout/Timeline.vue
+++ b/ui/src/components/layout/Timeline.vue
@@ -147,7 +147,7 @@
                 top: 50%;
                 left: 50%;
                 transform: translateY(-50%);
-                width: calc(100% - 15px);
+                width: calc(100% - 6px);
                 height: 1px;
             }
         }


Piyush-r-bhaskar:
Hey @D-260 

Since, the person assigned to the issue opened a PR and it's merged. 

Just a note for future: We are trying to follow Open Source Convention where you should get the issue assigned first before any actual work. Thanks for your efforts on this 🚀

May I recommend to check out [this filtered list](https://github.com/orgs/kestra-io/projects/15/views/1?filterQuery=label%3A%22good+first+issue%22+no%3Aassignee) of good first issues and find something you think you can help with.

PR #12292, opened 2025-10-23
fix(core): fix an escape bug in parser

### LINKED ISSUE #11373: Different escape symbol for flow and debug
### Describe the issue

In a flow I need to use "{{ ('fisrt.second.last' | split('\\.')) | first }}" \\ as an escape symbol to have a working flow, but in Outputs tab to debug - only one \ to escape.

Image

Image

Image

### Environment

- Kestra Version: 0.24.4

u7165614:

--- Fixes #11373

I updated `YamlParser.java` to correctly handle backslash (`\`) escape sequences during YAML parsing, ensuring that single backslashes are preserved instead of being lost or misinterpreted.

--- Added unit tests in YamlParserEscapeTest() and all tests passed succesfully via Gradle.


loicmathieu (changes requested):
The YamlParser behaves correctly, we use it in a lot of places and it didn't show any espace sequence issue.

This issue is specifically on the "Debug Expression" view, so something should be done in the Debug Expression functionality not the YamlParser.

loicmathieu:
This PR is now empty, closing it.
@u7165614  feel free to reopen it later with changes that fix the original issue including a test case if you still want to work on it

PR #12319, opened 2025-10-23
refactor(system): extract JdbcQueuePoller class from JdbcQueue

fhussonnois:


fhussonnois:
@loicmathieu maybe this kind of PR should be apply on 1.0 ? 

I also realized that, currently, when we run the runnable to stop a consumer, we only set `AtomicBoolean running = false ` but don’t wait for the poller to actually stop. This means that some remaining events can still be processed immediately after calling `disposable.run() ` if the polling thread is still doing the `receiveFetch`. I'm not very fan of that.

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueuePoller.java
```diff
@@ -0,0 +1,159 @@
...
+                }
+            }
+        } finally {
+            stopped.countDown();
+        }
+        
+    }
+    
+    private void waitIfPaused() throws InterruptedException {
+        pauseLock.lock();
```
loicmathieu, line 124: I don't think a lock is worth it here, their should alway be a single thread executing the poll.
Using a volatile attribute (or an Atomic* as it's anyway a volatile on steroid) is enought as it take into account the fact that pause/stop is done from another thread.
But the waitIfPaused is always called from the one and only one thread doing the poll so the lock is not needed here.

loicmathieu: if what's your after is a semaphore to avoid busy looping while in pause, you optimize for the wrong use case: when it's paused.
for that, you could use a countdownlatch if you want, but I think the current design is enought as it's very rare that it's paused so busy looping in this case is not a big deal.

loicmathieu (changes requested):

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueuePoller.java
fhussonnois: Here, I simply use a `ReentrantLock`/`Condition`  for doing the pause/resume mechanism which is a standard pattern. Yes this avoid the busy looping, but we have also the guarentee that we can do : 

```java
queue.pause()
queue.close()  
//  And, because the queue is paused, we can be sure that
//  the table will not be queried one last time when the close method wakes up the paused thread.
``` 

I'm a little too lazy to change it ^^

loicmathieu: I'm still worrying about the performance cost here; this code is damn perf sensitive.
On each poll loop you would lock, it may be properly eliminated by the JIT but it may not, and it this case, you would have a consequent performance penalty.

So I'll only validate this if you benchmark it and it didn't cause any regression.

fhussonnois: Agree, I should at-least check **paused** before locking and waiting for resume

**Before**

```
Result "io.kestra.jdbc.runner.JdbcQueuePollerBenchmark.testPollOnce":
  0.163 ±(99.9%) 0.015 us/op [Average]
  (min, avg, max) = (0.135, 0.163, 0.189), stdev = 0.020
  CI (99.9%): [0.149, 0.178] (assumes normal distribution)

Benchmark                                              Mode  Cnt   Score   Error  Units
i.k.jdbc.runner.JdbcQueuePollerBenchmark.testPollOnce  avgt   25   0.163 ± 0.015  us/op
```

**After**

```
Result "io.kestra.jdbc.runner.JdbcQueuePollerBenchmark.testPollOnce":
  0.147 ±(99.9%) 0.012 us/op [Average]
  (min, avg, max) = (0.121, 0.147, 0.165), stdev = 0.017
  CI (99.9%): [0.135, 0.159] (assumes normal distribution)

Benchmark                                              Mode  Cnt   Score   Error  Units
i.k.jdbc.runner.JdbcQueuePollerBenchmark.testPollOnce  avgt   25   0.147 ± 0.012  us/op
```

**Difference**: ~0.016 µs per operation.

loicmathieu: I like it better this way ;)
Note that, IMHO, semantically it's a Semaphore that should have been used but I'm OK with a lock as long as we don't use it in the hot path.

loicmathieu (approved):
LGTM

PR #12321, opened 2025-10-23
fix(kv): don't throw in KV function with errorOnMissing=false for exp…

### LINKED ISSUE #12294: kv() Pebble function throws unexpected exception when trying to access expired key
### Describe the issue

Hey, I encountered this weird / unexpected behavior (not sure if bug, maybe not, but unexpected behaviour, for me at least).

```
Caused by: io.pebbletemplates.pebble.error.PebbleException: The requested value has expired ({{ ['some_table'] + (inputs.dq_excluded_tables | default([])) + (kv('dq_excluded_tables') | default([])) }}
```

This Pebble expression
`{{ ['some_table'] + (inputs.dq_excluded_tables | default([])) + (kv('dq_excluded_tables') | default([])) }}`

works as expected when the KV key doesn't exist (it defaults to an empty array, which is what we want).

But the first time the KV key is accessed after it expires, it throws the above exception. (Subsequent accesses work as the key got deleted when the exception was thrown, and so the default value filter is then used.) This is (I assume) the code responsible for it:
https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java#L105

So my issue with this behavior is that from the docs/UI, I would expect the expiration to be a standard behavior = when I set a key to expire after say 7 days, it is correct and expected that it expires and is no longer there. So I don't expect an exception in such case, I expect the `kv()` function to return NULL, same as if the key doesn't exist - which is what I expect, when it expires, it shouldn't exist.

So **proposing to change the behavior to reflect that expectation = when a key expires, it is automatically deleted (possibly soft deleted), no exception is thrown, and NULL is returned because the key no longer exists.** (Or alternatively make the behavior configurable with some parameter)

cc @loicmathieu 

### Example

Example flow that reproducibly demonstrates this issue:

```
id: test-kv-expiration-error
namespace: test.joe
tasks:
  - id: kv_set
    key: test_var
    value: "test_value"
    kvType: STRING
    ttl: PT1S

    # will pause the flow; manually Resume in Kestra UI after 1+ seconds
    # note: Sleep task is better if available in your version of Kestra
  - id: pause

  - id: "log"
    type: "io.kestra.plugin.core.log.Log"
    level: DEBUG
    message: "{{ kv('test_var') }}"

```

### Environment

- Kestra Version: 0.19.10 (but it looks like the behavior is the same also in the most recent version, possibly with setting errorOnMissing=false to account for this change: https://kestra.io/docs/migration-guide/0.22.0/kv-error-on-missing)

brian-mulier-p:
…ired kv

closes #12294

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/pebble/functions/KvFunction.java
+++ b/core/src/main/java/io/kestra/core/runners/pebble/functions/KvFunction.java
@@ -38,7 +38,7 @@ public class KvFunction implements Function {
         String key = getKey(args, self, lineNumber);
         String namespace = (String) args.get(NAMESPACE_ARG);
 
-        Boolean errorOnMissing = Optional.ofNullable((Boolean) args.get(ERROR_ON_MISSING_ARG)).orElse(true);
+        boolean errorOnMissing = Optional.ofNullable((Boolean) args.get(ERROR_ON_MISSING_ARG)).orElse(true);
 
         Map<String, String> flow = (Map<String, String>) context.getVariable("flow");
         String flowNamespace = flow.get(NAMESPACE_ARG);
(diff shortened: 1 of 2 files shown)


nkwiatkowski (approved):
LGTM

PR #12324, opened 2025-10-23
feat(core): add Task Run Attempt Reason for (retry-restart-resubmit) cases

### LINKED ISSUE #8689: Add a reason attribute for a new task attempt
### Feature description

This would allow to distinct between retry caused by failure/replay/resubmit

Original attempt should not have any reason.

MTarek165:
closes #8689

### Description
- added a new reason attribute for task run attempt to enhance observability


core/src/main/java/io/kestra/core/models/executions/Reason.java
```diff
@@ -0,0 +1,7 @@
+package io.kestra.core.models.executions;
+
+public enum Reason {
```
loicmathieu:
```suggestion
public enum AttemptReason {
```

core/src/main/java/io/kestra/core/models/flows/State.java
```diff
@@ -269,6 +269,7 @@ public boolean isBreakpoint() {
         public boolean isRetrying() {
             return this == Type.RETRYING || this == Type.RETRIED;
         }
+        public boolean isRestarted() { return  this == Type.RESTARTED; }
```
loicmathieu:
```suggestion
        public boolean isRestarted() { 
            return  this == Type.RESTARTED;
        }
```

Please respect our coding format rules

core/src/main/java/io/kestra/core/models/executions/TaskRun.java
```diff
@@ -312,4 +311,22 @@ public TaskRun resetAttempts() {
             .build();
     }
 
+    public Optional<Reason> resolveAttemptReason() {
```
loicmathieu: I don't think it's a good idea to do this here.
The Worker is not the one responsible to set the reason, and we create attempts in other places.
I would have thought the reason would be set when we restart/retry/resubmit so when we create new attempt not when the Worker receive and attempt and set it to running.

loicmathieu (changes requested):
Moreover, this new REASON should be displayed inside the UI.

core/src/main/java/io/kestra/core/models/executions/TaskRun.java
MTarek165: The good place to add the reason is the place we create attempt to attach it at time of creation and since the attempt creation place is the same as we move it to running so  I don't see a problem here because I am already doing that ( attempt creation time nearly same as attempt running time) , DefaultWorker is central place but whenever we want to add reason for attempt in other place when creating a new attempt it is add easily  since resolveAttemptReason() is in TaskRun class where it holds the current attempts and the state histories neccessary for our validations.
 If there is more better suggestion than current one it would be nice.

worker/src/main/java/io/kestra/worker/DefaultWorker.java
```diff
@@ -907,9 +907,12 @@ private WorkerTask runAttempt(RunContext runContext, final WorkerTask workerTask
             return workerTask.withTaskRun(taskRun);
         }
 
+      Optional<Reason> attemptReason = workerTask.getTaskRun().resolveAttemptReason();
+
         TaskRunAttempt.TaskRunAttemptBuilder builder = TaskRunAttempt.builder()
             .state(new io.kestra.core.models.flows.State().withState(RUNNING))
```
MTarek165, line 913: @loicmathieu here exactly I mean by the new attempt is created and moved to running directly inside the worker, so no problem I am still attaching it at the attempt creation time as wanted

core/src/main/java/io/kestra/core/models/executions/TaskRun.java
MTarek165: I thought to add them exactly after taskrun state transitions to restart/retry/resubmit directly but the problem i couldn't found adding attempts occur there . for ex retryTask() function is responsible for retrying the task when i tried to add attempt reason i couldn't find attempt creation occur there . It is all centered at DefaultWorker (at least for runnable ignoring flowable ones as no reason)  at the exact place I refered to previously

core/src/main/java/io/kestra/core/models/executions/TaskRun.java
```diff
@@ -312,4 +311,22 @@ public TaskRun resetAttempts() {
             .build();
     }
 
+    public Optional<AttemptReason> resolveAttemptReason() {
```
loicmathieu, line 314: There is too much assumption here.
From the UI, you can change the state of a taskRun, force run or many other things that would lead to this algorithm not being accurate.

As I said, the reason should be set when you restart/retry/resubmit.

All in all, this may be harder than that.

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/models/executions/TaskRun.java
MTarek165: @loicmathieu so to check more is validations here missing cases? or it is ok but I should call exactly at places where restart/retry/resubmit occurs instead of calling it once at attempt creation place in DefaultWorker?

loicmathieu: This should be removed, resolving the reason from the attemps would never work.
When you create a new attemps from a restart/retry/resubmit, you should set the reason.

MTarek165: @loicmathieu Excuse me If I miss something but I want to understand , the confusing part is as I see the new attempt creation itself is not actually done and added to the list at places we change task run state to either restart/retry/resubmit. the first place where it is created is when  the task be running at worker. so It is not possible to add the reason there because the new attempt itself wasn't created yet at the time of taskrun state be restart/retry/resubmit .

   so for ex when retrying occur 
task run state goes to RETRYING -> CREATED -> SUBMITTED -> RUNNING
The new attempt is due to retry is added when task run becomes RUNNING not when it is RETRYING
so I depend on task run state history at resolving also when retry or restart to get the latest  , the only case I depend on old list of attempts at resubmit only.

loicmathieu: > new attempt creation itself is not actually done and added to the list at places we change task run state to either restart/retry/resubmit. the first place where it is created is when the task be running at worker. so It is not possible to add the reason there because the new attempt itself wasn't created yet at the time of taskrun state be restart/retry/resubmit .

Yes, this is why I said this would be harder than that.
Maybe the reason should be added to the taskrun not the attempt, or to the state.

Anyway, I think this is more complex and would need some careful design so it may be better to freeze this PR until we have some time to think about it properly internally, sorry about that.

MTarek165: @loicmathieu No problem sir , thanks for your careful review :)

loicmathieu:
I need to raise this internally and decide what would be the next step.

loicmathieu:
@MTarek165 finally we decided to instead of adding a `reason` attribute to an attempt, add a new `RESUBMITTED` state to taskruns that would be used when a taskrun is resubmitted instead of killing the taskrun which didn't make sense as the taskrun is not killed, it's resubmitted.

PR #12337, opened 2025-10-24
Fix: prevent missing translation warnings in TriggerVars.vue by adding fallback for unknown keys

### LINKED ISSUE #11962: Amend translation strings in `TriggerVars.vue` component
If you have a flow defined like this:

```yaml
id: transltions
namespace: github.qa

tasks:
  - id: hello
    message: This is a problem with translation strings

triggers:
  - id: flow_trigger
    conditions:
      - type: io.kestra.plugin.core.condition.ExecutionFlow
        flowId: system_flow
        namespace: system.prod
    states:
      - RUNNING
```
And you open `Triggers` tab of it, you'll see a warning in the console:

```bash
[intlify] Not found 'states' key in 'en' locale messages.
```
It's due to a fact that we have a `getHumanizeLabel` method in `kestra/ui/src/components/flows/TriggerVars.vue` component that does the mapping - so, we either need to cover all possible variations for triggers or check if the key exists and return raw value if not.

wenhan-u7898909:
This PR fixes [#11962](https://github.com/kestra-io/kestra/issues/11962), which caused repeated [intlify] Not found 'states' key in 'en' locale messages

**Changes**

Added a safe translation helper (tOr) that checks if a translation key exists with te() before calling t().

Provided humanized fallback text for missing translation keys.

Updated column labels and button text to use this helper.

**Result**

No more Vue i18n warnings in console.

Cleaner UI and consistent translations.

**Tested**
fixed-no bug

Verified locally with Docker backend


MilosPaunovic (changes requested):
There are some other console warnings now, can you amend those @wenhan-u7898909?

image

MilosPaunovic:
Any updates on this one @wenhan-u7898909?

MilosPaunovic (approved):

Issue #12341, opened 2025-10-24
Store KV Metadata in repository


brian-mulier-p:
### Feature description

It will avoid calling storage every time just for listing KV + more efficient expired key removal (https://github.com/kestra-io/kestra/issues/12320)

anna-geller:
KV Store metadata stored in DB should enable those features:
- allows pagination when returning KV pairs
- better filtering: search by name
- better performance
- automatically removing expired keys
- namespace files revision history

PR #12342, opened 2025-10-24
Feat/kv metadata

brian-mulier-p:
part of https://github.com/kestra-io/kestra/issues/12341

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/kv/Version.java
+++ b/core/src/main/java/io/kestra/plugin/core/kv/Version.java
@@ -1,0 +1,66 @@
+package io.kestra.plugin.core.kv;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import io.kestra.core.models.FetchVersion;
+import io.kestra.core.models.QueryFilter;
+import io.kestra.core.storages.kv.KVEntry;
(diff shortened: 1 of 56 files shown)


core/src/main/java/io/kestra/plugin/core/kv/Version.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Schema(
+        title = "The date before which versions should be purged."
+    )
+    private String before;
+
+    @Schema(
+        title = "How much versions should at least be kept for each matching KV.",
+        description = "By default, every matching versions will be purged. " +
+            "If together with `before`, it's a guaranteed minimum to keep, it will skip `keepAmount` versions from the matching results then proceed to purge."
```
nkwiatkowski: I may have not understand this well, but I tink you can't have both before and keppAmount

nkwiatkowski (approved):
LGTM

core/src/main/java/io/kestra/plugin/core/kv/Version.java
brian-mulier-p: Ups I refactored this part afterward because it was too clunky but forgot to remove that doc part, ty

PR #12344, opened 2025-10-24
Feature(UI): Added split panel for editor tabs opened from topology

### LINKED ISSUE #12048: No Code - New tab opens in different panel
### Feature description

At the moment, when you press the pencil icon on the topology, it opens a no code tab on the same panel as the topology. I could see a world where the topology is always displayed, and no code editor tabs open around it. Same applies for opening a task tab from the no code menu. Allow that to continue to show, and open it as a new tab in a different panel.

This would be best as an option in editor settings. Curious what others think of this idea - allowing the topology view / No-code main tab to stay present and all the additional tabs for tasks, inputs etc to open to the side. 

Future idea to be discussed internally: This could be expanded further where the topology view displays more of the things inside of the no-code editor too, like inputs, plugin defaults etc as no-code is a very visual type of building. 

Image

Abhyshekbhalaji:

---
This PR introduces support for automatically opening editor tabs from topology (pencil icon) clicks in a new/split panel.

- Adds an optional triggeredBy field to the TabLive type.
- When a tab is opened via topology, the field is set as { triggeredBy: "topology" }.
- The tab view logic detects this and uses the split panel handler for these tabs only.

closes  #12048

---
-Upon testing all other tab workflows remain unchanged and backwards-compatible


Abhyshekbhalaji:
@elevatebart Please look into this PR

MilosPaunovic:
It will be reviewed at our earliest convenience. There is no need to ping directly.

Abhyshekbhalaji:
Sorry @MilosPaunovic

ui/src/utils/multiPanelTypes.ts
```diff
@@ -11,6 +11,7 @@ export interface TabLive extends Tab {
     potential?: boolean
     fromPanel?: boolean
     dirty?: boolean,
+    triggeredBy?: string;
```
elevatebart: If we decide to keep this `triggeredBy`, this should be part of the opener. 

But I feel that a flag saying something along the lines of `target: "_blank" ` would be a better architecture for this change.

elevatebart (changes requested):
Thank you @Abhyshekbhalaji,

I really appreciate you looking into this. 

I would prefer to focus on the action than the originator. Indeed, opening a tab in a new panel could be coming from any sources not only from topology.

I would prefer the option to be called `"panelTarget": "new"` with a default to `"self"`.

I also need design to approve that if there is already 3 panels open (topology + code + nocode) we are OK opening a 4th one instead of reusing the nocode panel.

I believe adding a new tab to the panel where no-code is already if it exists would be more useful.

Abhyshekbhalaji:
Thanks @elevatebart for the feedback!

I'll make these changes:

1. Replace `triggeredBy` with `panelTarget: "new" | "self"` (default "self")
2. Move the logic to the opener component as you suggested
3. For the max panel scenario: I like your idea of adding a new tab to an existing nocode panel if it's already open. 
 I'll update the PR. Thanks please look into this!

Abhyshekbhalaji:
@MilosPaunovic can i work on other issues while this been reviewed sir?

MilosPaunovic:
Hi @Abhyshekbhalaji,  

You're already assigned to an issue in our repository, and we usually assign only one at a time to new contributors. Let's make sure to wrap this one up before jumping onto another.  🚀

Abhyshekbhalaji:
Hi everyone! I’ve opened PR #12409 with the requested panelTarget refactor based on @elevatebart ’s feedback. Should I close this PR and continue the review on the new one? Sry for the inconvenience caused from my side !!

Abhyshekbhalaji:
Closing this in favor of #12409 as advised by @MilosPaunovic —thanks for the feedback

PR #12364, opened 2025-10-25
fix(ui/flows): All search source are giving empty result

### LINKED ISSUE #12362: All search source are giving empty result
https://postgres-oss.preview.dev.kestra.io/ui/main/flows/search?page=1

Image

Affecting all versions (EE/OSS/Kafka)

akshaywritescode:
In this pr, i fixed search issue that was giving 0 results, no matter what query you put or what page are you in.

### Before
flow

### After

image
image

## What was problem?

- Component wasn't watching route query changes
- Expected search result structure but got different data format

## What i fixed?
 - Added Route Query Watcher
 `
watch(
    () => route.query,
    (newQuery, oldQuery) => {
        if (oldQuery && newQuery && oldQuery.q !== newQuery.q) {
            loadData();
        }
    },
    { deep: true }
);`

- Added optional chaining (?.) to handle both search and browse result formats
- Support both search fragments and flow labels
- Use correct API endpoints (`/flows/source` for search, `/flows/search` for browse)
- Properly handle empty vs undefined search queries

fixes #12362

### DIFF
--- a/ui/src/components/flows/FlowsSearch.vue
+++ b/ui/src/components/flows/FlowsSearch.vue
@@ -24,22 +24,37 @@
                 </template>
 
                 <template #table>
-                    <template v-for="(item, i) in flowStore.search" :key="`card-${i}`">
-                        <el-card class="mb-2" shadow="never">
-                            <template #header>
-                                <router-link :to="{path: `/flows/edit/${item.model.namespace}/${item.model.id}/source`}">
-                                    {{ item.model.namespace }}.{{ item.model.id }}
-                                </router-link>
-                            </template>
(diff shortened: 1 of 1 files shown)


Piyush-r-bhaskar (commented):
Thanks for the contribution! @akshaywritescode 

 But logic appears to be AI-generated and not well researched. The overall idea is appreciated, but this is adding too much of redundant code that is not required and which we’d like to avoid. Plus there are few extra issues I observed which we need to tackle here.
I’m closing this one now and will follow up with a new PR with simpler and more complete solution soon 🚀.

PR #12374, opened 2025-10-26
fix(core): use correct formatting for tags in blueprints for list and details view

### LINKED ISSUE #12078: Convert BlueprintDetail.vue component to TS
We need `kestra/ui/src/components/flows/blueprints/BlueprintDetail.vue` component converted to `TypeScript` with the use of `Composition API`.

```vue
<script setup lang="ts">
   // Converted code here
</script>
```

### LINKED ISSUE #12372: Show tags abbreviation in uppercase.
We have tags for each blueprints but currently it's forced to show only `Capitalize` text. We want special handling for abbreviation to show properly like `AI`, `AWS`, `GCP`, `SaaS` and for more.

This needs fixing on both List and Details view.

Image

PratikMurari:
Closes #12372 
Closes https://github.com/kestra-io/kestra/issues/12078.


Piyush-r-bhaskar (changes requested):
Hey @PratikMurari 

Thanks for this 🚀 but harcoding the list is not ideal and expected approach to solve this issue. Can you explore more ?

PratikMurari:
Thanks for the feedback! You're absolutely right that hardcoding isn't ideal.

I initially tried a pure CSS approach (`text-transform: uppercase`), which worked for simple abbreviations like AI, AWS, API, but failed for mixed-case tags like SaaS (became SAAS instead of SaaS). That's what led me to the hardcoded map.

I'd love to refactor this with a better approach. Could you share which direction you'd prefer?

**Option 1: Pattern-based detection**
- Auto-uppercase 2-4 letter tags (handles AI, AWS, API automatically)
- Keep a minimal hardcoded list only for true special cases (SaaS, DevOps, IoT)
- New abbreviations handled automatically

**Option 2: Preserve backend casing**
- If tags are already properly cased in the backend, just remove the CSS transform
- No formatting logic needed

**Option 3: Something else?**

Happy to update the PR based on your guidance! 🚀

Piyush-r-bhaskar (approved):

Issue #12356, opened 2025-10-24
plugin-redis


fdelbrayelle:
- [ ] plugin-redis/src/main/java/io/kestra/plugin/redis/list/ListPush.java: private Object from;
- [ ] plugin-redis/src/main/java/io/kestra/plugin/redis/pubsub/Publish.java: private Object from;

shubhammenroy:
Hi @fdelbrayelle,
I'd like to work on both this issue and the MinIO one (https://github.com/kestra-io/kestra/issues/12348).
Could you please assign them to me and share any additional details, guides, or context you have before I begin?

fdelbrayelle:
Hello @shubhammenroy 
I'll assign this one only first to go until the end properly.
For the infos you can get inspiration from 3 PRs in the [parent issue](https://github.com/kestra-io/kestra/issues/9680).
The idea is to make the task implement `Data.Form` and then use it like in the examples.

naveengowda7:
Hi @fdelbrayelle 
I would like to work on this migration to use the new 'ubiquitous from' pattern in the plugin. 
Could you please assign this issue to me?
Let me know if there are any edge cases i should be aware of before i begin. Thanks!

fdelbrayelle:
Hi @naveengowda7 :wave: This issue was already assigned to @shubhammenroy sorry. Let's go in the other you commented ;-)

shubhammenroy:
Hey @fdelbrayelle, I have raised a draft Pr. But 2 test's are failing might need to modify the Data class or I am might be missing something can you please take a look.

shubhammenroy:
Updated the PR and raised a PR in core to fix issue.

fdelbrayelle:
Closing this one since after investigation with @Malaydewangan09 using `Data.from` is not possible and bring too much complexity here, so better to keep it this way!

PR #12375, opened 2025-10-26
chore: Add a check to skip URI;s for Mono.

shubhammenroy:

feat(core) Modified Data to support the ubiquitous for redis plugin https://github.com/kestra-io/kestra/issues/12356
---

These changes are introduced for ignoring uri's in the Flux<Mono> return statement.

---


loicmathieu:
@shubhammenroy I'm not sure I understand why you have to do that.
Normally, the core is ready for using the new ubuiquitous from and there is nothing you should modified in it, you only need to update the Redis plugin to support it.

loicmathieu (changes requested):
The core should not be updated, only the Redis plugin

MilosPaunovic:
So should this PR be closed if that's the case @loicmathieu?

shubhammenroy:
> @shubhammenroy I'm not sure I understand why you have to do that. Normally, the core is ready for using the new ubuiquitous from and there is nothing you should modified in it, you only need to update the Redis plugin to support it.

Hi @loicmathieu, Yes, that’s correct — I only needed to update the Redis plugin to support ubiquitous.
However, while doing that I ran into an issue that was causing the Redis changes to fail.
Specifically, when the Redis plugin calls:
```
Data.from(from).readAs(runContext, String.class, obj -> obj.toString())
```
the condition
```
if (clazz.isAssignableFrom(from.getClass()))
```
returns true for URIs, which is incorrect in this case.
If I instead use:
```
Data.from(from).read(runContext)
```
it tries to convert a String into a Map, which leads to an error.
That’s why I added this check — as per my understanding, it should be there to handle this scenario correctly.
For more context please refer this redis PR: https://github.com/kestra-io/plugin-redis/pull/122

PR #12381, opened 2025-10-27
fix(core): handle namespace removal applied from setting

### LINKED ISSUE #12380: Ensure that default namespace can be removed from the Namespace filter
### Describe the issue

Currently, when setting a default namespace via Settings, it cannot be removed from filters:

https://github.com/user-attachments/assets/9f91d4e9-59bd-4e32-a7b2-2da68021178a

We should find a way so that:
- default namespace is pre-applied by default
- but it can be removed from the applied filters if the user chooses so.

Acc. to @elevatebart, this bug from the video happens only on Flows page 

### Environment

- Kestra Version: develop

Piyush-r-bhaskar:
Closes #12380

### DIFF
--- a/ui/src/components/flows/Flows.vue
+++ b/ui/src/components/flows/Flows.vue
@@ -586,33 +586,29 @@
     // Lifecycle
     onMounted(() => {
         displayColumns.value = loadDisplayColumns();
-        loadData(() => {
-            ready.value = true;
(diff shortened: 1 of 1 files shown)


elevatebart (approved):
Nice!!!! I have been raging abou this one for a while

PR #12422, opened 2025-10-28
fix(system): add resolveVersions method to PluginCatalogService

fhussonnois:
Related-to: kestra-io/kestra-ee#5171


AcevedoR (approved):
no tests possible here ?

PR #12432, opened 2025-10-28
fix(low-code): open tasks from topo in other panel

### LINKED ISSUE #12048: No Code - New tab opens in different panel

elevatebart:
closes #12048

### DIFF
--- a/ui/src/components/MultiPanelTabs.vue
+++ b/ui/src/components/MultiPanelTabs.vue
@@ -197,6 +197,7 @@
 
     import {trackTabOpen, trackTabClose} from "../utils/tabTracking";
     import {Panel, Tab, TabLive} from "../utils/multiPanelTypes";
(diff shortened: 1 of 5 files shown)


Piyush-r-bhaskar (approved):
Looks Good  🚀

PR #12517, opened 2025-10-29
fix: save tabs state onCreate

### LINKED ISSUE #12513: When saving a flow I would like to keep the state of my tabs/panels
### Feature description

In the multiPanelEditor thing I can change the panels and tabs displayed.
When I hit save for the first time I go from add to edit.

I then loose the status of my tabs and default is restored.

It is awkward to say the least. Might be a bug. 

[EDIT]
Here in the end, the same tabs should be open as I set them up.

https://github.com/user-attachments/assets/1221b550-624a-47fa-8bdf-564c1945deb0

elevatebart:
closes #12513

### DIFF
--- a/ui/src/components/MultiPanelGenericEditorView.vue
+++ b/ui/src/components/MultiPanelGenericEditorView.vue
@@ -63,7 +63,12 @@
         }
     };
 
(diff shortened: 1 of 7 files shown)


Piyush-r-bhaskar (approved):
Looking good.

PR #12521, opened 2025-10-29
fix(ui): debug expression now handles empty string results (fixes #12511)

### LINKED ISSUE #12511: Evaluate expression with empty return value doesn't refresh / show in the result
### Describe the issue

If the evaluate expression returns something, everything works:
Image

However if an evaluate expression returns an empty String, the result doesn't change, it keeps the last result (but it puts it outside of the preview box) + doesn't show the actual result which is an empty string (so we should have an empty preview box):
Image

### Environment

- Kestra Version: develop

rubrx:

Fixes #12511

This PR fixes an issue where debug expressions returning empty strings wouldn't update the UI properly.
**Problem:**
The debug expression evaluation uses a truthy check (`if (response.data.result)`) which fails for:
- Empty strings (`""`)
- Number zero (`0`)
- Boolean false (`false`)

This caused the UI to keep showing the previous result instead of displaying the actual empty result.

**Changes:**
- Changed condition from `if (response.data.result)` to `if (response.data.result !== null && response.data.result !== undefined)`
- Applied fix to both JSON parsing path and non-JSON path
- Now correctly handles all falsy values except `null` and `undefined`

**Root cause:**
JavaScript's truthy evaluation treats empty strings as falsy, so `if (result)` returns false for `""`, preventing the UI update.

---

**Test cases verified:**
1. ✅ Empty string: `{{ "" }}` - Now displays empty result correctly
2. ✅ String with spaces: `{{ "   " }}` - Displays correctly
3. ✅ Number zero: `{{ 0 }}` - Displays correctly  
4. ✅ Boolean false: `{{ false }}` - Displays correctly
5. ✅ Normal strings: `{{ "abc" }}` - Still works as expected
6. ✅ Previous results no longer stick when new empty result is evaluated

---

1. Started Kestra backend frontend development server.
2. Created a test flow with various output types
3. Navigated to Outputs tab
4. Tested debug expression with all test cases above
5. Verified results display correctly in all cases


brian-mulier-p:
@MilosPaunovic if you have some time to have a look at E2E tests as said in the other issue failures are unrelated but we may need to look and maybe introduce some timeout for it to load.

MilosPaunovic:
As time allows, I'll be looking at this `PR` next week.

rubrx:
Hi @MilosPaunovic!
Just checking in on this PR, as all checks have passed and it’s ready for review.
Please let me know if there’s anything else needed from my side or any feedback to address.
Thanks!

MilosPaunovic (approved):

PR #12531, opened 2025-10-30
feat(core): add ui part for scope filter on chart overview page

### LINKED ISSUE #12527: Dashboard should be filter for user flow
Image

On dashboard, we don't have the user scope filter that we used on all others place by default (flows / executions). 
We should use the same filters also on dashboard by default.

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra/issues/12527


Piyush-r-bhaskar:
Lot changed in filter and since Steven is already on it Filter, we will see what we can in recent.

PR #12542, opened 2025-10-30
feat(system): set taskrun attempt to resubmitted when a taskrun is re…

### LINKED ISSUE #12481: Add a new `RESUBMITTED` state to a taskrun in case the worker task has been resubmitted
### Feature description

When a worker is detected as dead, all its running worker task are killed then resubmitted.

To be able to detect taskrun that was resubmitted, instead of setting them to `KILLED` we should set them to a new `RESUBMITTED` state.

loicmathieu:
…submitted to a worker

Closes https://github.com/kestra-io/kestra/issues/12481


loicmathieu:
@Skraye just to be sure does it needs changes on the SDK side?

Skraye:
Current version of the SDK will required an update. Next version shouldn't not have issue with new enum anymore.

AcevedoR (approved):

Issue #172, opened 2020-10-12
Restart executions at specific task, no log from all previous tasks


tchiotludo:
When you restart a task at a specific task, we create a new execution with all the previous tasks. 
The log are dropped from this execution, since it's a new execution. 

Find a better way to handle this, the user will not understand the restarted executions.
First, we need to add a visual representation of task that were not running (opacity 0.5?)  
Second, we need to search log from previous execution on this page.

yuri1969:
This happens in v0.16 when you Restart a failed execution and select a different flow revision than the _current_ one. It creates a new execution with the OK tasks copied.

Spawning a new execution (meaning different execution ID, etc.) is confusing for the end users, since invoking Restart while keeping the _current_ flow revision simply retries the failed task within the existing execution.

---

Side note: executions featuring an _original_/_parent_ execution should be distinguishable in the Executions list somehow. Maybe adding a new column to the set?

loicmathieu:
What if we set all previous tasks state to SKIPED and keep the current behavior (only copy taskruns, not logs and metrics)? This would clearly show that the previous tasks were skipped.
Searching for logs/metrics to attach them back would not be that easy.

anna-geller:
unrealistic in 0.20, moved to cooldown + 0.25

anna-geller:
I believe that the main problem we are trying to solve here will be solved with the Test mode https://github.com/kestra-io/kestra-ee/issues/1547 so Anna-Bot is auto-closing this for now 😉 @tchiotludo feel free to reopen if you feel otherwise

PR #12546, opened 2025-10-30
Add maxObjectNameLength default method to StorageInterface

anglerfishlyy:
This adds a new default method maxObjectNameLength() to StorageInterface to define the maximum allowed length for object names. The default implementation returns Integer.MAX_VALUE for unlimited length.

This method is needed to support object name length protection in the S3 storage plugin, as implemented in [storage-s3 PR #172](https://github.com/kestra-io/storage-s3/pull/172)
.

Adding this default method ensures that all storage implementations can optionally enforce object name length limits in a consistent way without breaking existing implementations.


loicmathieu:
Closing it as it's not what needs to be done, proper guidance has been added to the S3 PR.

PR #12566, opened 2025-10-31
feat(pebble): add new pebble function that supports getting outputs from previous iterations

### LINKED ISSUE #11206: access the value of a previouse iteration in a foreach loop

MTarek165:
closes #11206
## Description
- added new pebble function to access outputs of previous iterations at the current iteration.
- modified Outputs run variable to internally use LinkedHashMap instead of HashMap to preserve order.

## Testing
- added unit test for calculating prefix sum indices outputs using the new function and asserts on the two last odd and even indices outputs.


loicmathieu (changes requested):
I'm afraid adding another map inside the RunVariables is a no go as it would have a high cost, it would double the size of the RunVariables which we want to avoid as they can be hundreds of taskruns in an execution

MTarek165:
It is not added for all task runs it is added only for iterative ones like (ForEach) . Is it still costly ? as we are doing similar thing for run variable "tasks" for as we are iterating over all task runs and makes updates each time.

MTarek165:
There is other alternative is to make Outputs itself a LinkedHashMap but it may make many changes at MapUtils from HasMap to LinkedHashMap which I afraid also to introduce breaking changes.

MTarek165:
If it is safe I can go with the other solution it seems more efficient.

MTarek165:
There is a lighter version of the current solution, we may instead of defining the value as LinkedHashMap of iteration number and value , it is better to be List of values only no need for iteration number since we append values in order that is we matter about.

loicmathieu:
Yes, using a LinkedHashMap may work.

MTarek165:
Ok I know where to make these changes, I would go to the second solution.

MTarek165:
I went to the second solution , now Outputs run variable maintains order and no need for adding new run variable now.

core/src/main/java/io/kestra/core/runners/pebble/functions/OutputFromIterationFunction.java
```diff
@@ -0,0 +1,72 @@
...
+
+        if (index < 0) {
+            throw new PebbleException(null, "Cannot fetch iteration " + index + ": no previous iteration exists.", lineNumber, self.getName());
+        }
+
+        List<?> parents = (List<?>) context.getVariable("parents");
+        if (parents != null && !parents.isEmpty()) {
+            throw new PebbleException(null, "The 'outputFromIteration' function is not supported inside nested loops. ", lineNumber, self.getName()
+            );
+        }
```
loicmathieu: Is there a strong reason for that?
You should be able to look for iterations inside the current nest as outputs would be prefixed by the taskrun value (nested map).

core/src/main/java/io/kestra/core/runners/pebble/functions/OutputFromIterationFunction.java
```diff
@@ -0,0 +1,72 @@
...
+
+        //  output of current or future iterations can't occur in current iteration
+        if (index == taskOutputs.size()) {
+            throw new PebbleException(null, "The provided index (" + index + ") refers to the current iteration, " + "whose outputs are not yet available at runtime.", lineNumber, self.getName());
+        }
+        if (index > taskOutputs.size()) {
+            throw new PebbleException(null, "The provided index (" + index + ") is out of range. " + "It refers to a future iteration whose outputs do not exist yet. " + "Maximum valid index is " + (taskOutputs.size() - 1) + ".", lineNumber, self.getName());
+        }
+
+       List<?>taskValues= taskOutputs.keySet().stream().toList();
```
loicmathieu:
```suggestion
       List<?>taskValues= new ArrayList(taskOutputs.keySet());
```

core/src/main/java/io/kestra/core/utils/MapUtils.java
```diff
@@ -27,7 +27,7 @@ public static Map<String, Object> merge(Map<String, Object> a, Map<String, Objec
             return a;
         }
 
-        Map<String, Object> result = HashMap.newHashMap(Math.max(a.size(), b.size()));
```
loicmathieu, line 30: Is it really needed?
This code is used in multiple places so changing it this way is not a good idea.
If it's really needed a check on the type of the Map should be done to know which map implementation to use.

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/runners/pebble/functions/OutputFromIterationFunction.java
MTarek165: Yes, I know that but there is strong reason to not support calling this function from nested iterations.

1- If we made that it should be Output( idx1, idx2, idx3,.....) to get the output exactly from that specific place with indices which is very high cost where time complexity of this function is O(iteration number) per iteration due to getting List each time from map so adding this cost of deep nesting would make a performance problem.

2- There is another stronger reason is that even if we did that and optimized more it is not practical as Kestra has no way to access parent iterations dynamically, there is no Parent.taskrun.iteration like parent.taskrun.value so we can't pass indices dynamically , so it is not much benefit to add this support. 
It is much effort with low benefit.

core/src/main/java/io/kestra/core/utils/MapUtils.java
MTarek165: it is critical because of this Map<String, Object> merged = MapUtils.merge(taskOutputs, outputs(current, byIds));
at execution.outputs() 
but I would check more if there is possibility to keesp it otherwise i would add check on map type or separate between both.

MTarek165: taskOutputs must be LinkedHashMap and it uses this function that is why i changed

core/src/main/java/io/kestra/core/runners/pebble/functions/OutputFromIterationFunction.java
MTarek165: Oh , I was thinking on generic approach getting output from any nest Output(idx1, idx2, ...) but for Output of same nest Output(idx) I wasn't think about it. I should give a look here if it is really possible or not.

MTarek165: You are right I would update it to get value from current nest and add test for it

MTarek165:
wait @MilosPaunovic there is some other changes I would do once finish I would request the final review.

MilosPaunovic:
Ahh, my bad, sorry for the noise.

MTarek165:
No problem , welcome.

MTarek165:
@loicmathieu Now handled the two points:
1- separating between map implementations in MapUtils.merge()
2- added current nested support for the function (covered in test)
please give a look and tell me if there is missing something.

core/src/main/java/io/kestra/core/utils/MapUtils.java
loicmathieu: OK, I would need to run our benchmarks before merging it so see if it causes performance regression or not.

MTarek165:
@loicmathieu what is benchmarks news here should we go?

loicmathieu:
@MTarek165 please be patient, I didn't had time yet to benchmark this as I have other priorities

MTarek165:
@loicmathieu No problem, I appreciate that, take your time.

loicmathieu (changes requested):
I'll soon perform benchmark and decide whether or not we merge this change so can you please rebase onto develop to have the latest change.

If we switch to LinkedHashMap, we will do it everywhere in MapUtils so please remove the merge/mergeOrder and unconditionally use a LinkedHashMap.
Other methods (deepMerge, flattentToNestedMap, ...) must also be changed to align all methods.

MTarek165:
@loicmathieu 
1- I will remove the conditional check for merge() and make it LinkedHashMap always.
2- We don't need to change it for all MapUtils methods, just places affects task outputs (merge(), deepMerge(),..) flattenToNestedMap() not needed, I will check more.
3- I did initial benchmark today and the results shows slight degradation but tolerable around 8% under good test load but I will make sure again.

MTarek165:
1- MapUtils methods that need change to LinkedHashMap are the ones that interact with task outputs which are (merge() - deepMerge() - deepCloneMap()), others that don't interact with task outputs such as flattenToNestedMap() are kept using HashMap. 
2- I have rebased it on top of develop and squashed them, they are ready for benchmarking, please let me know if something missed.

core/src/main/java/io/kestra/core/utils/MapUtils.java
```diff
@@ -105,7 +104,7 @@ private static Object mergeValues(Object valueA, Object valueB) {
     }
 
     private static Map<String, Object> deepCloneMap(Map<String, Object> original) {
-        Map<String, Object> cloned = new HashMap<>(original.size());
+        Map<String, Object> cloned = new LinkedHashMap<>(original.size());
```
loicmathieu:
```suggestion
        Map<String, Object> cloned = LinkedHashMap.newLinkedHashMap(original.size());
```

Would be better to avoid any map resizing

loicmathieu (commented):
Thanks for the rebase.
I'm going to benchmark this today or tomorrow and then we will be able to decide whether or not we merge it.

loicmathieu:
I checked performance quickly and it didn't seems to incurs noticeable performance impact so it's good to GO on this front.

loicmathieu:
Hi,
We discussed it internally with the product team and decided to name this function `iterationOutputs(index)` and change it to return the map of outputs from the previous iteration (as it was done for `currentEachOutput()` instead of the output of the current task so you would be able, from one task, to access the output of any task of a previous iteration of the ForEach.
Could you make the necessary changes?

MTarek165:
@loicmathieu Nice extension, but it may be more about the implementation.
Returning the Map of all outputs of a ForEach iteration for tasks with same sibling is complicated and has a performance problem,the reason is that:
1- The entry points for outputs map are the tasks ids so to do that we should iterate over all task ids and traverse the parents of task runs to get the outputs for each task sibling and then collect them into map with task ids and corresponding value, this will end up with map of (task ids-output from iteration) for the tasks in same level at ForEach, this would be be a heavy computation, and each time we are using this function it would do the same.


At the current implementations (OuputFromIteration() - CurrentEachOutput())   we avoid that by either getting the current task output by id from task meta data firstly and traversing their outputs only (as in OuputFromIteration()) or passing the output of the required task (as in CurrentEachOutput()) and traversing its outputs only too, the common pattern is determining the target task outputs and then traverse them only.

So IMHO a better suggestion is to change the implementation way by adding another param taskId refering to the sibling we need its output from the previous iteration so that the function would be 
iterationOutputs( targetIteration- targetTaskId) and instead of passing current task id i would do the same but for the target sibling task id generalizing it as intended achieving the same goal.

This is much better for two reasons:
1- Complexity: It requires almost no new code (we just read taskId from arguments instead of metadata as for now).
2- Performance: It allows an immediate O(1)
 lookup (outputs.get(targetTaskId)), maintaining high performance without memory bloat and overkilling traversal for all task outputs to return the ForEach sibling ones.

MTarek165:
Nice it worked as expected.
Here is the summary of the latest changes:
1- added another param taskId for iterationOutputs(iteration, taskId) facilitating retrieval of output of any sibling task for previous ForEach iterations
2- changed the naming to iterationOutputs()
3- handled edge cases such as when no taskId provided we consider the current one and when no Iteration we throw excpetion
4- when no proper taskId sibling is passed it will throw pebble exception
5- added test coverage for retreival of outputs of sibling tasks (see iteration-outputs.yaml, task id : iteration_output_sibling)
let me know if it is ok.

loicmathieu:
@MTarek165 nice, yes adding a `taskId` parameter is exactly what needs to be done.
Can you make the iteration the second parameter so it can be omitted easily? I think that in Pebble when you have multiple parameters and only set a subset, they are set in order.

MTarek165:
The final changes:
1- make iteration param default to previous iteration if not provided (this is safest from current iteration as the default task is the current one and outputs of previous iteration for the current task exists but the output of current iteration for the current task can't be computed since the iteration is proceesing)
2- handle cases when the passed taskId doesn't exist or it exists but not for the current ForEach nest
3- make parents immutable because there is interesting bug in old code allows mutating parents list and appears clearly when the function calls more than one time at the same template since they all evaluated with the same evaluation context (it is the same for currentEachOutput function too, I would open a separate PR modifying it for currentEachOutput) :D

the order of params doesn't matter, if we need to ignore param we can pass it as null and we can either write it as:
1- iterationOutputs() or iterationOutputs(null, null) defaulting to iteration= previous iteration and taskId= current taskId
2- iterationOutputs(taskId, null) defaulting to iteration= previous iteration 
3- iterationOutputs(null, iteration) defaulting to taskId= current taskId

I covered them in tests and It need only a proper documentation about usage and edge cases, I will open another PR documenting this feature with details soon.
I think we have addressed all the necessary points, is there something remaining?

MTarek165:
This new function is a complete generalization as it now gets the iteration outputs from either previous or current iterations for any siblings or current tasks making the current CurrentEachOutput() function a special case from it for the current iteration as currentEachOutput(targetSiblingOutput) is equivalent to iterationOutputs( taskrun.iteration, targetSiblingId), so as you said currentEachOutput() may be deprecated now.

core/src/main/java/io/kestra/core/runners/pebble/functions/IterationOutputsFunction.java
```diff
@@ -68,8 +67,12 @@ public Object execute(Map<String, Object> args,
 
         Map<?,?> targetOutputs = (Map<?, ?>) outputs.get(taskId);
 
-        List<Map<?, ?>> parents = (List<Map<?, ?>>) context.getVariable("parents");
-        if (parents != null && !parents.isEmpty()) {
+        if(targetOutputs == null)
+            throw new PebbleException(null, "The provided task with taskId = " + taskId + " has no execution outputs", lineNumber, self.getName());
```
loicmathieu: Please always wraps with `{}`

core/src/main/java/io/kestra/core/runners/pebble/functions/IterationOutputsFunction.java
```diff
@@ -68,8 +67,12 @@ public Object execute(Map<String, Object> args,
...
+
+        List<Map<?, ?>> immutableParents = (List<Map<?, ?>>) context.getVariable("parents");
```
loicmathieu: This is the parent and it's not yet _immutable_ as you didn't copy the collection yet.
By the way, using a stream().reverse() would avoid the copy

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/runners/pebble/functions/IterationOutputsFunction.java
MTarek165: nice, using reversed() is better

core/src/main/java/io/kestra/core/runners/pebble/functions/IterationOutputsFunction.java
```diff
@@ -0,0 +1,103 @@
...
+public class IterationOutputsFunction implements Function {
+    public static final String NAME = "iterationOutputs";
+
+    private static final String ITERATION_ARG = "iteration";
+
+    private static final String TASK_ID_ARG = "taskId";
+
+    @Override
+    public List<String> getArgumentNames() {
+        return List.of(ITERATION_ARG, TASK_ID_ARG);
```
loicmathieu:
```suggestion
        return List.of(TASK_ID_ARG, ITERATION_ARG);
```

It's more natural and I think as TASK_ID is required you would be able to call the Pebble expression without the parameter name if you omit the iteration

core/src/main/java/io/kestra/core/runners/pebble/functions/IterationOutputsFunction.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+
+        Map<?, ?> outputs = (Map<?, ?>) context.getVariable("outputs");
+
+        Map<?,?> targetOutputs = (Map<?, ?>) outputs.get(taskId);
+
+        if(targetOutputs == null){
+            throw new PebbleException(null, "The provided task with taskId = " + taskId + " has no execution outputs", lineNumber, self.getName());
+        }
```
loicmathieu: In this case I think it would be better to return null so the caller may use a conditional construct

core/src/main/java/io/kestra/core/runners/pebble/functions/IterationOutputsFunction.java
```diff
@@ -0,0 +1,103 @@
...
+        if (parents != null && !parents.isEmpty()) {
+            for (Map<?, ?> parent : parents) {
+                Map<?, ?> taskrun = (Map<?, ?>) parent.get("taskrun");
+                if (taskrun != null) {
+                    if(targetOutputs.get(taskrun.get("value")) != null){
+                        targetOutputs = (Map<?, ?>) targetOutputs.get(taskrun.get("value"));
+                    }
+                    else{
+                        throw new PebbleException(null, "The provided task with taskId = " + taskId + " has no execution outputs for the current iterative task runs path", lineNumber, self.getName());
+                    }
```
loicmathieu: Same here, return null

loicmathieu (changes requested):

MTarek165:
@loicmathieu updated. 
please check the last minor updates in the last commit.

loicmathieu (approved):
LGTM, thanks

PR #12572, opened 2025-10-31
chore(core): enhance github issue templates

MilosPaunovic:
Added `Type` property and uniformed `yaml` a bit, with slight wording enhancements.


anna-geller (approved):
Important to keep in mind we don't use and we don't rely on the `Type` property. We rely exclusively on labels `bug` and `enhancements` internally for automations, filtering and KTLO calculations so this could confuse people that we may rely on Type. The Type is very inefficient at the moment, there's no way to easily filter for that in GitHub projects, not supported in most APIs. I'd prefer to not use it to avoid confusing people

MilosPaunovic:
Alright, removed the `type` property, this just leaves us with uniformed style in `yaml`. Will be merged tomorrow, after the freeze.

anna-geller:
some on, this is just issue template, you can merge it now if you want 😄

wrussell1999 (approved):

PR #12605, opened 2025-11-03
fix(ui): prevent "Invalid date" display in Gantt view for executions …

### LINKED ISSUE #12583: Execution that never started shows `Invalid date` across the gantt view
### Describe the issue

While this makes sense to display as the execution didn't run, it looks a bit messy having it repeated across the UI. Maybe we could just leave it blank? @Ben8t wdyt?

Image

### Environment

- Kestra Version: `develop`

vipinsao:
Closes #12583


Piyush-r-bhaskar:
Hey @vipinsao 

Can you share your reproducer how you tested it ?

vipinsao:
Hey @Piyush-r-bhaskar,
I tried to reproduce the “Invalid date” issue locally by creating multiple flows and saving them with future date. However, I was not able to directly trigger the exact “Invalid date” message in my local testing. After going through the code and logic, I noticed that if the history array length is 0, it leads to this issue. So, I focused on updating the conditional checks in the code to handle that scenario properly, making sure it won’t display the invalid date even if the array is empty.

Piyush-r-bhaskar (approved):
Thanks for this @vipinsao 🚀

Issue #12615, opened 2025-11-03
LoopUntil generates duplicated next tasks


loicmathieu:
### Describe the issue

Try with the following flow:

```yaml
id: waitfor-multiple-tasks
namespace: io.kestra.tests

tasks:
  - id: waitfor
    type: io.kestra.plugin.core.flow.WaitFor
    condition: "{{ outputs.output_values.values.count == '4'}}"
    checkFrequency:
      maxIterations: 3
      interval: PT2S
    tasks:
      - id: echo
        type: io.kestra.core.tasks.log.Log
        message: "looping ..."
      - id: output_values
        type: io.kestra.plugin.core.output.OutputValues
        values:
          count: "{{ outputs.waitfor.iterationCount + 1 }}"
```

When executed, you can see the following WARN log that indicates that the LoopUntil task generates duplicate next tasks which indicates a bug and prevent removing the next task deduplication mechanism.

```
17:49:19.021 WARN  jdbc-execution-executor_2 io.kestra.jdbc.runner.JdbcExecutor Duplicate Nexts on execution '4ll9wa4V3FwoDCw0FN46Cu' with key '2oSmL2gHQ3nfTtYU13Jgz9-echo-null-01'
```

### Environment

- Kestra Version: develop

KishoreGhost:
I would like to work on this issue

MilosPaunovic:
Maybe it would be better to chose one of the issue with `good first issue` label, as this does not seem like an easy fix.

KishoreGhost:
I would like to take on this problem, I want to challenge myself. I will ask clarification questions if I am faced with a blocker or some sort. :)

KishoreGhost:
Hey @MilosPaunovic, I am facing some issue reproducing this bug i am not able to see any warning for creating the flow mentioned above

https://github.com/user-attachments/assets/89d73dec-492e-4cbc-8607-81a2d859de2c

Can you help me with this

MilosPaunovic:
@loicmathieu is the one who might help out here.

loicmathieu:
@KishoreGhost the warning would only be shown using 1.1.0.

This is a very complex issue that requires knowledge of the inside workings of our executor, so I'll unassign you as it would probably need someone from our internal team. Sorry about that.

You can still work on it if you want, I'll be happy to review a PR, but I would not provide any guidance as it would be too complex to explain what happened and why to someone that didn't have such deep understanding of our executor.

loicmathieu:
[...]

PR #12616, opened 2025-11-03
Feat/executor v2

loicmathieu:
- Single executor wich uses state stores provided by the repository
- Queues and repositories are now un-coupled: any queue can be used with any repository
- Single implementation for service lifecycle moved inside the executor module
- Single implementation for the indexer
- Flow topology computation externalized in its own component

TODO:
- [ ] Externalize the indexer in its own module
- [ ] Re-evaluate QueueIndexer
- [ ] Re-enable the Scheduler: will be done after Scheduler v2 would be merged
- [ ] Re-evaluate flow listeners & the Flow metastore
- [ ] Investiguate if flow topology could be computed in another way
- [ ] Evaluate if we need to listen to both the execution queue and the execution event queue
- [ ] Refactor the DefaultServiceLivenessUpdater
- [ ] Reevaluate metastores: they are provided by the queue but the Kafka one is limited to executor so we need to have an Elastic one


fhussonnois (approved):

loicmathieu:
`PauseTest.runEmptyTasks()` appears to be flaky as it pass locally.
The *.waitFor* tests would fail until https://github.com/kestra-io/kestra/issues/12615 would be resolved.

This PR will be merged as soon as its EE counterpart would pass.

PR #12619, opened 2025-11-03
fix(core): ability to specify a commands separator on scripting

brian-mulier-p:
part of https://github.com/kestra-io/plugin-scripts/issues/301


brian-mulier-p:
https://github.com/kestra-io/plugin-scripts/pull/302 this was the issue in the end so no need for this

PR #12638, opened 2025-11-04
feat(TaskEdit): enhance task loading logic with effective source resolution

### LINKED ISSUE #12625: Show task source not working
### Describe the issue

The `show task source` button from task run logs / Gantt doesn't display the source:
https://jam.dev/c/93a18499-0245-419a-9210-5a58359544d7

### Environment

- Kestra Version: develop

iitzIrFan:
Fixes: https://github.com/kestra-io/kestra/issues/12625

---

Video Preview: 

https://github.com/user-attachments/assets/32f01517-c315-43da-ae4c-05e8a8f01720

---


ui/src/components/flows/TaskEdit.vue
```diff
@@ -180,12 +180,40 @@
...
+        // Try direct block extraction first
+        const sectionKey = (typeof props.section === "string" ? props.section : "tasks").toLowerCase();
+        const block = YAML_UTILS.extractBlock({
+            section: sectionKey,
+            source: effectiveSource ?? "",
             key: taskId,
         });
+        if (block && block.length > 0) return block;
+
+        // Fallback: parse whole flow and find task by id then stringify it
```
elevatebart: I hope this is never necessary

ui/src/components/flows/TaskEdit.vue
```diff
@@ -180,12 +180,40 @@
...
+        // 3) Finally, fallback to store flow source
+        if (!effectiveSource) {
+            effectiveSource = flowStore.flow?.source;
         }
-        return YAML_UTILS.extractBlock({
-            section: props.section,
-            source: source.value,
+
+        // Try direct block extraction first
+        const sectionKey = (typeof props.section === "string" ? props.section : "tasks").toLowerCase();
```
elevatebart: Since section is a required prop, we should count on it respecting the contract. this line make the code more convoluted.

ui/src/components/flows/TaskEdit.vue
```diff
@@ -180,12 +180,40 @@
...
+        }
+
+        // 2) Fallback to provided flowSource (e.g., from logs/Gantt)
+        if (!effectiveSource && props.flowSource) {
+            effectiveSource = props.flowSource;
+        }
+
+        // 3) Finally, fallback to store flow source
+        if (!effectiveSource) {
+            effectiveSource = flowStore.flow?.source;
```
elevatebart: You still rely on the flowStore.flow to exist and not be empty. 
The root cause here is it should not be empty.

elevatebart (changes requested):
Add this at the top of the load function and things start to behave better.

```js
    const load = async (taskId: string) => {
        await flowStore.loadFlow({
            namespace: props.namespace,
            id: props.flowId,
            revision: props.revision?.toString(),
        });
```

The two issues here are:
- the flowStore is not initialized before its source gets used
- when instantiating the TaskEdit component, the section prop is `TASKS` while it should be `tasks` lowercase.

This feels a little convoluted a fix for my taste.

iitzIrFan:
Thanks for the review @elevatebart  sir, I have updated the PR as per your suggestions, and yes turns out it could indeed be implemented in a simpler way than my earlier approach. Thanks for pointing that out! :)

elevatebart (approved):

MilosPaunovic:
Are we ready to merge this one @elevatebart?

PR #12650, opened 2025-11-04
fix(ui): exclude default equal query filter from base in case of there is selected one

### LINKED ISSUE #12640: Namespace filter doesn't work on the Namespace page -> KV Store tab
### Describe the issue

Image

Demo showing it's broken (doesn't correctly add KV pairs from selected child namespaces): https://jam.dev/c/358fca4a-d6a4-44d1-a1b1-ced3d2cdf9b4

### Environment

- Kestra Version: develop

MTarek165:
closes #12640
## Problem
- There was default filter EQUAL added with the selected one and sent to backend which makes database filter using both condition1 && Condition2(selected) which creates unexpected results due to this extra one.

## Solution
- Simply prevent that filter to be sent to backend in case of there is selected one otherwise send it.

see demo here after fix
https://drive.google.com/file/d/1a_TF-aItykbg0ekxqLlTYK-_xStbZ9ee/view?usp=sharing


Piyush-r-bhaskar (commented):
Hey @MTarek165 

with your changes , I am able to access KV from other namespace as well which are not the Parent/Child.

Also, I need to discuss if we really want to have the KV from upstream parent to show in the table,  as it's currently namespace specific 
```
                    filters: {
                        namespace: {
                            EQUALS: props.namespace
                        }
                    }
```
and we do have a `Inherited KV Pair	` option in navbar which will give us the Upstream parent namespace KVs, so not sure why not just add the namespace filter we are in from route but keep it readonly.

What do you think about this @MilosPaunovic and what was the table data earlier ?

MilosPaunovic:
I'd say you're right @Piyush-r-bhaskar, we should probably not allow seeing KV pairs from other namespaces.

Piyush-r-bhaskar:
Thank you Milos 

@MTarek165  we will discuss it first internally in cycle and decide wayforward. till then will move your PR on Hold.

MTarek165:
@Piyush-r-bhaskar I thought that we should access KVs of any namespace just like in KVStore page, Can you tell me please what is exactly the expected behavior when filtering KVs by namespaces from inside specific namespace in our case?

MTarek165:
@Piyush-r-bhaskar A worth mentioning note that even before the change , It was allowing showing KVs from other namespaces that are not parent/child. It does same as in KV Store page. So it is not obvious here what was the intended behavior here before the change?
waiting for your decisions.

Piyush-r-bhaskar:
Is it ? Can you share a recording with us how you are seeing/reproducing kv from another namespace under the KV Store tab ?

MTarek165:
@Piyush-r-bhaskar Yes , this change doesn't affect behavior it only removes the obstacle that prevents that behavior to occur (that default filter). Even before theoretically It behaves just like in KV store page filtration from any namespace before fixing the bug. Of course it wouldn't be shown in UI directly because of the bug that is covering it but This record shows debugging the old code at FE and BE sides which confirms what I said that it doesn't do any restriction on namespace filtration. 

You can check here 
https://drive.google.com/file/d/18aD_XEp02qapiElerYyWe5J2NDQTcB_d/view?usp=sharing

MTarek165:
So you would see at recording two filters passed the selected one contains all selected namespaces without any restriction and the default EQUAL with current namespace, making AND together producing KVs for current namespace only ignoring others. after fix I remove that so selected one passed only and gets the correct results revealing the true behavior , check recording at PR description after fix and compare. 

So fixing the bug is half of the solution , the behavior was previously exist but hidden behind bug, not as you thought that my change is the one that changed the behavior :)

 after that now we need to change this behavior to make it only includes parent/child namespaces at this selected filter or whatever it be to get the really intended reults. That is let to you to take that decision and I can continue the PR after that and finish .

Piyush-r-bhaskar:
Closing in favour of https://github.com/kestra-io/kestra/pull/12840

Thanks for your efforts on this 🍀

PR #12703, opened 2025-11-05
fix(core): bring dashboard selector in navbar and also keep the selected dashboard route specific

### LINKED ISSUE #12700: Applying custom dashboard on Main is applying on Flows and Namespace and vice-versa.
### Describe the issue

Applying custom dashboard on Main is applying on Flows and Namespace and vice-versa. It should be route specific. also the storing of dashboard selected should use the storage key of different route.

https://github.com/user-attachments/assets/43c045b4-c061-4a68-a867-d95c6e896680

### Environment

- Kestra Version: develop

Piyush-r-bhaskar:
Closes #12700
Closes https://github.com/kestra-io/kestra-ee/issues/5672

https://github.com/user-attachments/assets/68dba336-e57f-4d58-a11a-2bd7e293c507


ui/src/stores/dashboard.ts
```diff
@@ -28,6 +28,7 @@ export const useDashboardStore = defineStore("dashboard", () => {
         const dashboard = ref<Dashboard>();
         const chartErrors = ref<string[]>([]);
         const isCreating = ref<boolean>(false);
+        const loadDashboard = ref<((value: any) => void) | undefined>();
```
elevatebart: Where do you assign this function/variable?

elevatebart (changes requested):
Thank you @Piyush-r-bhaskar this is a good start

I have one question to start with.
Then I would have loved a little bit of explanation.

I feel that storing the active dashboardid only in localstorage is wrong because opening a new browser will have none. 

I would prefer to have it always as a URL param.

This could be another PR though

ui/src/stores/dashboard.ts
Piyush-r-bhaskar: Ahh, I think i went over engineering way doing function reference in store, I was storing the local `load()`  reference. let me not keep it tightly coupled and refactor it with reactive route (though we are hiding the ID in url for `isFlow` and `isNamespace`).

Piyush-r-bhaskar:
> I feel that storing the active dashboardid only in localstorage is wrong because opening a new browser will have none.
> 
> I would prefer to have it always as a URL param.

I would love little bit more guidance on this 🚀

elevatebart (approved):

PR #12762, opened 2025-11-06
fix(core): bring the usage of restore url

### LINKED ISSUE #12755: Restore url is not working any more
Previously on all the list, we have a system to restore the past url with all the query string saved in session storage

Seems that it was totally remove and it's a critical part of the UX 

- go on flows page 
- search for something 
- see the url that changed to `/ui/main/flows?filters[scope][EQUALS]=USER&filters[q][EQUALS]=test` 
- click on executions 
- go back to flows page 
- see the url that is the homepage `/ui/main/flows?filters[scope][EQUALS]=USER` with no filters 

We need to restore this feature

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra/issues/12755

Extra here is `ui/src/components/logs/LogsWrapper.vue` in compostition API to use composable instead of `beforeRouteEnter` at component level.

### DIFF
--- a/ui/src/composables/useRestoreUrl.ts
+++ b/ui/src/composables/useRestoreUrl.ts
@@ -86,7 +86,7 @@ export default function useRestoreUrl(options: UseRestoreUrlOptions = {}) {
 
     // Automatically call goToRestoreUrl on mount if needed (equivalent to created() hook)
     onMounted(() => {
-        if (Object.keys(route.query).length === 0 && restoreUrl) {
+        if (restoreUrl) {
             loadInit.value = false;
             goToRestoreUrl();
         }
--- a/ui/src/composables/useRestoreUrl.ts
+++ b/ui/src/composables/useRestoreUrl.ts
@@ -1,6 +1,8 @@
 import {computed, nextTick, onMounted, ref} from "vue";
 import {useRoute, useRouter} from "vue-router";
 import {defaultNamespace} from "./useNamespaces";
+import {useDefaultFilter} from "../components/filter/composables/useDefaultFilter";
+import isEqual from "lodash/isEqual";
 
 interface UseRestoreUrlOptions {
     restoreUrl?: boolean;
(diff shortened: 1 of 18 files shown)


ui/src/composables/useRestoreUrl.ts
```diff
@@ -86,7 +86,7 @@ export default function useRestoreUrl(options: UseRestoreUrlOptions = {}) {
 
     // Automatically call goToRestoreUrl on mount if needed (equivalent to created() hook)
     onMounted(() => {
-        if (Object.keys(route.query).length === 0 && restoreUrl) {
+        if (restoreUrl) {
```
elevatebart: This should probably be `route.query !== defaultQuery && restoreUrl` where 		`defaultQuery` is the filter applied when one first comes to the page. 

As of now, if you have been to this page ever, linking to this page with a URL filter is useless since it will always be replaced.

elevatebart (changes requested):
Since all instances of `useDataTableActions()` use the restoreUrl feature, it shoudl be an internal implementation detail of this hook. and not a parameter passed to it.

Also, once again, if you bookmark a URL, the system will always restore the previously opened url. It make bookmarks moot.

Let's talk about it

elevatebart:
I believe a quick refactor of the default filters would be welcome here so we avoid having them recreated everywhere.

ui/src/routes/routes.js
```diff
@@ -62,7 +49,25 @@ export default [
...
+                next({
+                    name: to.name,
+                    params: to.params,
+                    query,
+                });
+                return;
+            }
+
+            next();
+        }
```
elevatebart: Since this is twice the same code, should we make it a function?

elevatebart (approved):
one simple change on the routes but this looks good

ui/src/routes/routes.js
Piyush-r-bhaskar: Good idea 🚀

PR #12765, opened 2025-11-06
fix(kv): all types properly handled and avoid trimming string KV values

brian-mulier-p:
closes https://github.com/kestra-io/kestra-ee/issues/5718

https://jam.dev/c/5c008492-4cd1-4124-a8ab-c826e82708a2

### DIFF
--- a/ui/src/components/kv/KVTable.vue
+++ b/ui/src/components/kv/KVTable.vue
@@ -235,7 +235,7 @@
     import {useI18n} from "vue-i18n";
     import {useRoute} from "vue-router";
     import _groupBy from "lodash/groupBy";
-    import {computed, ref, useTemplateRef, watch} from "vue";
+    import {computed, nextTick, ref, useTemplateRef, watch} from "vue";
 
     import Check from "vue-material-design-icons/Check.vue";
     import Delete from "vue-material-design-icons/Delete.vue";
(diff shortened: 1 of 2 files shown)


ui/src/components/kv/KVTable.vue
```diff
@@ -491,6 +491,8 @@
         kv.value.key = entry.key;
         const {type, value} = await namespacesStore.kv({namespace: entry.namespace, key: entry.key});
         kv.value.type = type;
+        // Force the type reset before setting the value
+        await nextTick();
```
brian-mulier-p, line 495: There was a bug where the type watcher was resetting the boolean value to false AFTER the value being set to true from the retrieved value

ui/src/components/kv/KVTable.vue
```diff
@@ -543,14 +545,16 @@
             const type = kv.value.type;
             let value: any = kv.value.value;
 
-            if (["STRING", "DURATION", "JSON"].includes(type)) {
+            if (type === "STRING") {
+                value = `"${value}"`;
```
elevatebart: We will have problems if the value contains a `"`

Should we prefer the more common?

```suggestion
                value = JSON.stringify(value)
```

elevatebart (changes requested):
One question suggestion if you agree and we are gold

ui/src/components/kv/KVTable.vue
brian-mulier-p: Nice spot my bad, ty, I pushed your suggestion.

elevatebart (approved):

PR #12789, opened 2025-11-07
fix: when removing a queued execution, directly delete instead of fet…

Skraye:
…ching then delete to reduce deadlock


loicmathieu (approved):
LGTM

PR #12800, opened 2025-11-07
chore(flows): amend flow export filename to include namespace and id parameters

### LINKED ISSUE #12790: Name exported flows by flow id
### Feature description

Small feedback nit: I've been doing lots of exports and imports of flows lately, and it's a bit of a hassle always renaming the exported flow from the default name flow.yaml to a different name, so that the downloads directory doesn't fill with multiple anonymous versions of flow.yaml. Couldn't the download action automatically name the download file to `{{ id }}.yaml` at least, i.e., name the file by the contents of the id kv?

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/12790.

### DIFF
--- a/ui/src/components/inputs/EditorButtonsWrapper.vue
+++ b/ui/src/components/inputs/EditorButtonsWrapper.vue
@@ -59,12 +59,12 @@
     const {t} = useI18n();
 
     const exportYaml = () => {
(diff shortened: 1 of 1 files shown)


anna-geller (approved):
LGTM, thx a lot!

PR #12804, opened 2025-11-07
feat(tests): intercept created executions through queue proxy & kill …

brian-mulier-p:
…them if running after test


tests/src/main/java/io/kestra/core/runners/TestRunner.java
```diff
@@ -25,6 +29,8 @@
 @Slf4j
 @Singleton
 public class TestRunner implements Runnable, AutoCloseable {
+    public static final ThreadLocal<List<Execution>> testExecutions = ThreadLocal.withInitial(ArrayList::new);
```
loicmathieu: Why don't you move this thread local inside the TestQueueFactory?
This would avoid setting it so you would directly access it via the factory

tests/src/main/java/io/kestra/core/junit/extensions/KestraTestExtension.java
```diff
@@ -62,5 +73,49 @@ public void afterTestExecution(ExtensionContext context) throws Exception {
...
+
+            TestRunner.testExecutions.get().clear();
+        }
+    }
+
+    private void retryingExecutionKill(ExecutionRepositoryInterface executionRepository, QueueInterface<ExecutionKilled> killQueue, int retriesLeft) throws InterruptedException {
+        try {
+            TestRunner.testExecutions.get().stream()
+                .flatMap(launchedExecution -> executionRepository.findById(launchedExecution.getTenantId(), launchedExecution.getId()).stream())
+                .filter(inRepository -> inRepository.getState().isRunning())
```
loicmathieu:
```suggestion
                .filter(inRepository -> !inRepository.getState().isTerminated())
```

better to also kill paused/queued exec

tests/src/main/java/io/kestra/core/junit/extensions/KestraTestExtension.java
```diff
@@ -62,5 +73,49 @@ public void afterTestExecution(ExtensionContext context) throws Exception {
         super.afterTestExecution(context);
 
         TestsUtils.queueConsumersCleanup();
+
+        KestraTest kestraTest = context.getTestClass()
+            .orElseThrow()
+            .getAnnotation(KestraTest.class);
+        if (kestraTest.startRunner()
+            && applicationContext.containsBean(ExecutionRepositoryInterface.class)
```
loicmathieu, line 93: It's easier to check if there is something in the thread local, in this case you know you can kill exec

tests/src/main/java/io/kestra/core/queues/TestQueueFactory.java
```diff
@@ -0,0 +1,44 @@
...
+    public void setTestExecutionsList(List<Execution> testExecutions) {
+        this.testExecutions = testExecutions;
+    }
+
+    @Singleton
+    @Replaces(named = QueueFactoryInterface.EXECUTION_NAMED)
+    @Named(QueueFactoryInterface.EXECUTION_NAMED)
+    @Bean(preDestroy = "close")
+    public QueueInterface<Execution> execution() {
+        return (QueueInterface<Execution>) Proxy.newProxyInstance(this.getClass().getClassLoader(), new Class[]{QueueInterface.class}, (proxy, method, args) -> {
```
loicmathieu, line 28: I'm not a fan of dynamic proxy usage, not easy to understand such code.
But I guess it's easier than maintaining a delegate....

tests/src/main/java/io/kestra/core/junit/extensions/KestraTestExtension.java
```diff
@@ -62,5 +73,49 @@ public void afterTestExecution(ExtensionContext context) throws Exception {
...
+                            .executionId(inRepository.getId())
+                            .state(ExecutionKilled.State.REQUESTED)
+                            .isOnKillCascade(true)
+                            .build()
+                        );
+                    } catch (QueueException e) {
+                        log.warn("Couldn't kill execution {} after test execution", inRepository.getId(), e);
+                    }
+                });
+        } catch (ConcurrentModificationException e) {
```
loicmathieu, line 129: I'm not sure I understand why you want to retry and what would cause concurrent modifications here

loicmathieu (changes requested):
Overall I would prefer we understand why things are not ended properly but as it seems overly complicated why not going into that direction...

tests/src/main/java/io/kestra/core/runners/TestRunner.java
brian-mulier-p: Because the execution emit can come from everywhere including non-test threads (scheduler for eg.) so it will not be shared and retrievable from the test thread which is what we need

tests/src/main/java/io/kestra/core/junit/extensions/KestraTestExtension.java
brian-mulier-p: Because the test that just finished could still be populating executions which would, from the proxy, modify concurrently the list of executions. This is good for us as we want  to detect that. My retry will avoid failing by retrying till a gap between execution addition is available. 10 retries max so no way to be stuck forever and I log it so we will no there is something wrong and in which test :)

tests/src/main/java/io/kestra/core/queues/TestQueueFactory.java
brian-mulier-p: I know I went for the delegate way at first and then thought about the future us that would have to maintain adding simple methods...

brian-mulier-p:
I replied and fix all your comments + I spotted one last issue in context handling that is now fixed.

loicmathieu (approved):
Still not convinced by what you said about the thread local, if you're not in the test thread, you would not have access to it anyway.

I approved it, but it seems it causes more harm than good as a lot of tests are now failing

loicmathieu:
This kind of PR should only be merged if it proved to brings benefits: for ex when launching multiple time you see none of very few flaky tests

brian-mulier-p:
It just needs some love & polish, trust the process :+1:

brian-mulier-p:
In the end @loicmathieu  I removed the need for ThreadLocal so it's easier to understand + I added a wait after sending the killing message to be 100% sure it doesn't overflow across tests :crossed_fingers:

brian-mulier-p:
Not needed for now

PR #12806, opened 2025-11-07
fix(executions): don't urlencode files as they would already be insid…

loicmathieu:
…e the storage

It reverts a mis-behaving fix for supporting files with special characters.
Tested with both the local storage and MinIO (which fail otherwise)


AcevedoR (commented):
A test could be added on local storage right ?

loicmathieu:
@AcevedoR I update the existing test with a special char in one of the file

AcevedoR (approved):

loicmathieu:
It has been backporting to 1.1 and 1.0

PR #12807, opened 2025-11-07
fix(triggers): send delete triggers parameters properly to API

### LINKED ISSUE #11386: [Administration -> Triggers page] Provide an option to delete a trigger
Often, some triggers are very difficult to troubleshoot and understand, especially triggers where source is not found:
Image

We should provide a way to Delete/Clear such triggers

this would be more effective than adding extra filters like `missingSource` to just hide those https://github.com/kestra-io/kestra/issues/11230

EDIT: technically easy to do; the impact is that if the flow still exists, it will restart processing as this will be treated as a new save, so the trigger may reappear.

brian-mulier-p:
closes #11386

### DIFF
--- a/ui/src/components/admin/Triggers.vue
+++ b/ui/src/components/admin/Triggers.vue
@@ -77,7 +77,7 @@
                                 <el-button @click="deleteBackfills()">
                                     {{ $t("delete backfills") }}
                                 </el-button>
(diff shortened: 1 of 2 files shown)


Piyush-r-bhaskar (approved):
LGTM 🚀

PR #12820, opened 2025-11-10
fix(executions): allow reading from subflow even if we have a parent

### LINKED ISSUE #12629: Nested Subflow read() fails for parent execution when parent is not top-level.
### Describe the issue
Nested Subflow read() fails for parent execution when not parent is not top-level

#### Steps to reproduce:

- Create a flow `simple_write` that writes a file using `io.kestra.plugin.core.storage.Write` and outputs its URI:

```
id: simple_write
namespace: company.test

outputs:
  - id: uri
    type: STRING
    value: "{{ outputs.write.uri }}"

tasks:
  - id: write
    type: io.kestra.plugin.core.storage.Write
    content: "test write"
```

- Create a parent flow `parent_flow` that calls `simple_write` as a subflow and reads the file:

```
id: parent_flow
namespace: company.test

tasks:
  - id: subflow
    flowId: simple_write
    namespace: company.test

  - id: log_parent
    message: "{{ read(outputs.subflow.outputs.uri) }}"

```

- Create a second-level parent flow `parent_flow_level_2` that calls `parent_flow` as a subflow:

```
id: parent_flow_level_2
namespace: company.test

tasks:
  - id: subflow_level_2
    flowId: parent_flow
    namespace: company.test
```

**Observed behavior:**

Running `parent_flow` directly works fine.  `read()` successfully reads the file from `simple_write`.

Running `parent_flow_level_2` → `parent_flow` → `simple_write` fails inside `parent_flow` with the error:

`Unable to read the file ... as it didn't belong to the parent execution`

```
java.lang.IllegalArgumentException: Unable to read the file 'kestra:///company/test/simple-write/executions/4MgtuIkh3cCltijbLHf8G5/tasks/write/2FiZHlLYRAtuTB6N73MmRX/15526676570492818959.tmp' as it didn't belong to the parent execution
```

**Expected behavior:**

The `parent_flow` should be able to read the outputs of its direct subflow `simple_write`, regardless of whether it is a top-level flow or executed as a subflow itself.

### Environment

- Kestra Version: 1.0.8
- Docker deployment with postgres.18

loicmathieu:
This fixes an issue where you cannot read from a Subflow file if the execution has iteself be triggered by another Subflow task. It was caused by the trigger check beeing too aggressive, if it didn't pass the check it fail instead of return false so the other check would not be processed.

Fixes #12629

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/pebble/functions/AbstractFileFunction.java
+++ b/core/src/main/java/io/kestra/core/runners/pebble/functions/AbstractFileFunction.java
@@ -151,10 +151,7 @@ abstract class AbstractFileFunction implements Function {
             // if there is a trigger of type execution, we also allow accessing a file from the parent execution
             Map<String, String> trigger = (Map<String, String>) context.getVariable(TRIGGER);
 
-            if (!isFileUriValid(trigger.get(NAMESPACE), trigger.get("flowId"), trigger.get("executionId"), path)) {
-                throw new IllegalArgumentException("Unable to read the file '" + path + "' as it didn't belong to the parent execution");
-            }
-            return true;
+            return isFileUriValid(trigger.get(NAMESPACE), trigger.get("flowId"), trigger.get("executionId"), path);
         }
         return false;
     }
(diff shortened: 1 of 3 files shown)


Skraye (approved):
LGTM

PR #12887, opened 2025-11-12
test(e2e): make e2e tests pass again with restoreUrl

### LINKED ISSUE #12755: Restore url is not working any more

elevatebart:
- isolated the restoreUrl scheme to the filters and put it before the default filters so it does not conflict with it
- Make sure it only triggers if no query params have been specified

Closes https://github.com/kestra-io/kestra/issues/12755


Piyush-r-bhaskar:
This is more preferable. 🤞 on E2E


and Thanks so much for handling it 🎉

Piyush-r-bhaskar (approved):
Looks Good. Thank for this 🚀

PR #12889, opened 2025-11-12
fix(system): access log configuration

loicmathieu:
Due to a change in the configuration file, access log configuration was in the wrong sub-document.

Fixes https://github.com/kestra-io/kestra-ee/issues/5670


AcevedoR (commented):
LGTM

AcevedoR (approved):

PR #12908, opened 2025-11-12
fix(core): ensure flow labels schema supports both array and map formats

### LINKED ISSUE #12837: Wrong editor warning on labels autocompletion
### Describe the issue

To reproduce, add flow with any labels:

```yaml
id: crane_957994
namespace: company.team
labels:
  team: data
  owner: kestrel
  project: falco
  environment: dev
  country: US
tasks:
  - id: hello
    message: Hello "{{labels}}"
```
Image

### Environment

- Kestra Version: develop

vikramsingh117:
Closes https://github.com/kestra-io/kestra/issues/12837.

addresses issue https://github.com/kestra-io/kestra/issues/12837

AbstractFlow.labels now declares the schema as implementation = Object.class with oneOf = {List.class, Map.class}, matching how the model already accepts both styles in YAML.
JsonSchemaGenerator has a fixLabelsSchema that rewrites the generated JSON schema so the labels property has an anyOf with both "type": "array" and "type": "object". This addresses Monaco-YAML’s validation warning (“Incorrect type. Expected array”).

Manual verification by rebuilding the backend, restarting the UI, and opening a flow YAML using map-style labels: Monaco no longer emits the Expected "array" warning, and autocompletion continues to suggest label keys correctly

## Before 

image

## After
image


loicmathieu:
I know there was a bunch of back and forth lately on our JSON schema due to the SDK so I'll prefere that @Skraye first have a look at this PR.
And I'm not fan of another customization done at the generation side, those are brittle and may explode at some point when we update the JSON schema library.

vikramsingh117:
Makes sense. I agree that post processing on schema generation is a choppy fix, I only did it because I saw other committed issues present in the code (mentioned in the issue discussion). This solves the linting problem for now but definitely not ideal. A better solution might require refactoring but am not experienced with schema validation.

Skraye (changes requested):
Your PR doesnt really solve the issue, by bringing back
`implementation = Object.class,` , this break the SDK (that why we remove it at first)

also, all the code in other files seems to be noise, as they dont change anything, if I remove the mention line, autocomplete is throwing again the warning

If they fix something else, please open a PR for each issue

Thanks!

MilosPaunovic:
@Skraye So should we close this PR?

PR #12911, opened 2025-11-12
fix(flow): don't URLEncode the fileName inside the Download task

loicmathieu:
Also provide a `fileName` property that when set would override any filename from the content disposition in case it causes issues.


brian-mulier-p (approved):
LGTM

PR #12915, opened 2025-11-12
Revert "fix(core): bring the usage of restore url (#12762)"

elevatebart:
this makes e2e tests pass again

This commit introduced a regression that impacts e2e but users as well:
- go to a url with a filter like http://localhost:5173/main/flows?filters[scope][EQUALS]=USER&filters[labels][EQUALS][foo]=bar
- change the filter (for instance remove the label)
- go to the url again, the label is not there anymore
- the system uses the restoreUrl instead of the one passed by the user.


AcevedoR (approved):

elevatebart:
follow-up on https://github.com/kestra-io/kestra/pull/12887

PR #12931, opened 2025-11-13
wip scheduler

fhussonnois:


core/src/main/java/io/kestra/core/models/triggers/TriggerId.java
loicmathieu, line 1: The construct with interface and private default implementation seems a bit overkill, I cannot think of why another implementation may be needed.
But, well, keep it if you want.

core/src/main/java/io/kestra/core/runners/Scheduler.java
```diff
@@ -2,5 +2,35 @@
 
 import io.kestra.core.server.Service;
 
-public interface Scheduler extends Service, Runnable {
+public interface Scheduler extends Service, Runnable{
+    
+    // The default max threads 
+    int DEFAULT_MAX_THREAD = 4;
```
loicmathieu:
```suggestion
    int MIN_MAX_THREAD = 4;
```

core/src/main/java/io/kestra/core/runners/WorkerGroupMetaStore.java
loicmathieu, line 1: This seems unrelated.
Can you revert? Or maybe you're not fastforward?

core/src/main/java/io/kestra/scheduler/vnodes/VNodeConsistentHashRing.java
```diff
@@ -0,0 +1,100 @@
...
+     * @return the assignment map.
+     */
+    public Map<String, Set<Integer>> assignVNodes() {
+        if (ring.isEmpty()) {
+            return Map.of(); // fast path
+        }
+        
+        Map<String, Set<Integer>> assignment = new HashMap<>();
+        for (int vnode = 0; vnode < vNodeCount; vnode++) {
+            SortedMap<Integer, String> tail = ring.tailMap(VNodes.hash("vNode-" + vnode));
```
loicmathieu, line 93: I don't understnad `VNodes.hash("vNode-" + vnode)` as has are computed with `nodeId + "#" + i` shouldn't we compute the tailMap from `VNodes.hash(`nodeId + "#")` instead?

core/src/main/java/io/kestra/core/services/AbstractService.java
```diff
@@ -0,0 +1,86 @@
+package io.kestra.core.services;
```
loicmathieu:
```suggestion
package io.kestra.core.server;
```

This is not a service but an abstract class for servers (that implements a Service interface but we may rename this interface)

core/src/main/java/io/kestra/scheduler/vnodes/VNodeController.java
```diff
@@ -0,0 +1,165 @@
...
+                    ));
+                    
+                    try {
+                        // [2] Wait for 'SchedulerListReply' from schedulers
+                        Duration rebalanceTimeout = schedulerConfiguration.vnodesRebalanceTimeout();
+                        boolean timeout = !countDownLatch.await(rebalanceTimeout.toMillis(), TimeUnit.MILLISECONDS);
+                        if (timeout) {
+                            Set<String> timeoutSchedulers = new HashSet<>(activeSchedulerServices);
+                            timeoutSchedulers.removeAll(replySchedulers);
+                            log.warn("VNode rebalancing in progress: {} scheduler(s) did not respond within the timeout period of {}. Affected schedulers: {}", timeoutSchedulers.size(), rebalanceTimeout, timeoutSchedulers);
```
loicmathieu, line 143: Shouldn't we stop here and not continue with step 3 and 4?

core/src/main/java/io/kestra/scheduler/vnodes/internals/DefaultVNodesAssigner.java
```diff
@@ -0,0 +1,207 @@
...
+    private void handle(SchedulerEvent event, VNodeAssignmentSubscription registration) {
+        
+        final ControllerIdAndEpoch currentControllerIdAndEpoch = registration.controllerIdAndEpoch();
+        
+        if (!(event instanceof SchedulerEvent.VNodesAssignmentEvent vNodesAssignmentEvent) || event instanceof SchedulerEvent.VNodesAssignmentRejected ){
+            return;
+        }
+        
+        // Check whether the received event is stale
+        if (currentControllerIdAndEpoch != null) {
```
loicmathieu, line 94: Why can it be null here?

core/src/main/java/io/kestra/scheduler/internals/DefaultVNodesAssigner.java
```diff
@@ -0,0 +1,207 @@
...
+        }
+        
+        this.subscriptions.put(service, new VNodeAssignmentSubscription(null, service, new LoggerTriggerAssignmentListener(listener, LOG, service)));
+    }
+    
+    private void handle(SchedulerEvent event, VNodeAssignmentSubscription registration) {
+        
+        final ControllerIdAndEpoch currentControllerIdAndEpoch = registration.controllerIdAndEpoch();
+        
+        if (!(event instanceof SchedulerEvent.VNodesAssignmentEvent vNodesAssignmentEvent) || event instanceof SchedulerEvent.VNodesAssignmentRejected ){
```
loicmathieu: Maybe use a switch here, would have been easier to read

core/src/main/java/io/kestra/scheduler/vnodes/internals/StaticVNodesAssigner.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Simple {@link VNodesAssigner} which immediately rebalance VNodes on subscribe.
+ * <p>
+ * Only for testing-purpose.
+ */
+@Singleton
+@Requires(missingBeans = VNodesAssigner.class)
+public class StaticVNodesAssigner implements VNodesAssigner {
```
loicmathieu, line 22: if it's for test, move it to test sources if possible

core/src/main/java/io/kestra/scheduler/model/TriggerState.java
```diff
@@ -0,0 +1,301 @@
...
+import java.time.ZonedDateTime;
+import java.util.List;
+
+/**
+ * Immutable class representing the state of a trigger.
+ */
+@Getter
+@EqualsAndHashCode
+@AllArgsConstructor
+public final class TriggerState implements TriggerId {
```
loicmathieu, line 25: It seems a duplication of the Trigger class.
We already have Trigger/TriggerContext, adding a third class add more confusion IMHO

core/src/test/java/io/kestra/scheduler/utils/InMemorySchedulerEventQueue.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.function.Consumer;
+
+/**
+ * Simple in-memory event queue implementation.
+ * <p>
+ * This class is for testing-purpose only.
+ */
+public class InMemorySchedulerEventQueue implements SchedulerEventQueue {
```
loicmathieu, line 17: It would be better to use H2 if possible to avoid maintening an in-memory impl, we had that before: too much to maintain

jdbc-h2/src/main/resources/migrations/h2/V3_4__trigger_vnodes.sql
```diff
@@ -0,0 +1,30 @@
+-- Add generated columns
+ALTER TABLE triggers
+    ADD COLUMN "vnode" INT GENERATED ALWAYS AS (JQ_INTEGER("value", '.vNode'));
+
+ALTER TABLE triggers
+    ADD COLUMN "locked" BOOLEAN GENERATED ALWAYS AS (JQ_BOOLEAN("value", '.locked'));
+
+-- Indexes
+CREATE INDEX IF NOT EXISTS idx_triggers_vnode ON triggers ("vnode");
+CREATE INDEX IF NOT EXISTS idx_triggers_locked ON triggers ("locked");
```
loicmathieu: A compound index with all the columns used by the scheduler evaluation loop would be more performant.
We alreadyt have one so maybe it shouild be deleted and recreted with the two new columns.

An index with too few cadinility would have less change to be used.

scheduler/src/main/java/io/kestra/scheduler/DefaultScheduler.java
loicmathieu, line 1: It lack support for maintenance mode which shoudl pause the evaluation loop

scheduler/src/main/java/io/kestra/scheduler/TriggerEventHandler.java
loicmathieu, line 1: Would it be possible for two trigger events to be processed concurrrently?
I'm a little concerned here that we potentially may override one event by another.

core/src/main/java/io/kestra/scheduler/events/TriggerCompleted.java
```diff
@@ -0,0 +1,21 @@
...
+
+import io.kestra.core.models.flows.State;
+import io.kestra.core.models.triggers.TriggerId;
+
+import java.time.Instant;
+
+/**
+ * A trigger execution completed.
+ */
+public record TriggerCompleted(
```
loicmathieu:
```suggestion
public record TriggerExecutionTerminated(
```

Be explicit.
We don't use completed but terminated execution

core/src/main/java/io/kestra/scheduler/events/TriggerExecuted.java
```diff
@@ -0,0 +1,22 @@
...
+
+import io.kestra.core.models.executions.Execution;
+import io.kestra.core.models.triggers.TriggerId;
+
+import java.time.Instant;
+
+/**
+ * A trigger was executed.
+ */
+public record TriggerExecuted(
```
loicmathieu:
```suggestion
public record TriggerEvaluated(
```

Or TriggerEvaluation but we use evaluate not execute so it didn't conflict with the execution itself which is not yet started

scheduler/src/main/java/io/kestra/scheduler/stores/CachedFlowMetaStore.java
loicmathieu, line 1: I'm very concerned by duplicating the FlowMetaStore as flows can be big in memory.
I would prefer that you use the current FlowMetaStore and if needed the FlowListners and filter when needed on the flow you need.

I have some ideas to refactor the FlowMetaStore and the FlowListeners later, we need to discuss this...

scheduler/src/main/java/io/kestra/scheduler/stores/DefaultFlowMetaStore.java
loicmathieu, line 1: I think you can use the existing metastore

scheduler/src/test/java/io/kestra/scheduler/utils/InMemoryFlowMetaStore.java
loicmathieu, line 1: It should be moved to test resources.
But as we use H2 in test I would prefer not to have specific in-memory impl, you can always use `@MockBean` when needed to mock the default impl

scheduler/src/test/java/io/kestra/scheduler/utils/InMemoryTriggerStateStore.java
loicmathieu, line 1: Same here, use H2 or at least move to test sources

core/src/main/java/io/kestra/scheduler/SchedulerClock.java
loicmathieu, line 1: Is it really needed? Why can't you just use a `Clock` when needed?
I didn't look at tests as the PR is too big so if it's needed can you just explain it briefly?

scheduler/src/main/java/io/kestra/scheduler/internals/NextEvaluationDate.java
loicmathieu, line 1: This should not be an interface as it didn't expose any _contract_.
it should either be an utility class, or mayube just static method at the use site

scheduler/src/main/java/io/kestra/scheduler/stores/CachedTriggerStateStore.java
loicmathieu, line 1: I didn't look in detail as the PR is huge but why do you bother partitionning the cach by vNode?
Cache should not be aware of vnodes at all, it should be as dumb as possible: it caches stuff, that's all

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
```diff
@@ -0,0 +1,358 @@
...
+        Instant nextScheduleTime = clock.instant();
+        Instant tick = clock.instant();
+        try {
+
+            while (running.get()) {
+                try {
+                    long elapsed = clock.instant().toEpochMilli() - tick.toEpochMilli();
+                    if (elapsed > (SCHEDULE_INTERVAL_MILLIS + (SCHEDULE_INTERVAL_MILLIS / 10))) {
+                        // useful for debugging unexpected schedule delay
+                        LOG.warn("Thread starvation or clock leap detected (elapsed since previous loop {}ms", elapsed);
```
loicmathieu: This may just be too much trigger to evaluate in the loop no?

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
```diff
@@ -0,0 +1,358 @@
...
+                    }
+                    
+                    // Check whether vNodes are available for this event-loop
+                    // The list of vNodes assignments can be empty if:
+                    // 1. This scheduler is starting
+                    // 2. The vNode assignments was cleared/revoked
+                    if (assignments.isEmpty()) {
+                        doOnEndLoop();
+                        if (assignments.isEmpty()) {
+                            Thread.sleep(Duration.ofMillis(50));
```
loicmathieu, line 127: This seems arbitrary, can you explain the rational here?

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
```diff
@@ -0,0 +1,358 @@
...
+                        }
+                        continue;
+                    }
+                    
+                    final Instant now = clock.instant();
+                    
+                    if (!initialized) {
```
loicmathieu: Maybe an AtomicBoolean would be more safe here?

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
loicmathieu, line 1: We previously used a ScheduledExecutor which take care for us of scheduling an low-level threading stuff, why was it not possible here to do the same?

loicmathieu, line 1: The usage of a blocking queue and a lock seems to be overkill here.
Or maybe I didn't understand its purpose as you have a single looping thread so why would you use a blocking queue and a lock which are usually used for inter-proccess communication?

loicmathieu (changes requested):
- I love the vNode implementation (hashing, ring, ...)! Very elegant! I'm sure we will reuse it at some point ;)
- Please separate all vnodes related code to its own package so it may be easily read, and extracted if needed.
- vNodes are computed on flow, maybe they should be computed at the trigger level?
- I ignored any queue related code as I consider it as temporary as it would be replaced by the queue v2 later. But anyway I say it now I'm not a fan of one class per queue message as it would just adds a lot of classes without real added value
- I ignored any worker related code as I consider it as temporary as it would be replaced by the Worker v2 later
- Overall, I think you should have kept using the `Trigger` object and not add a new `TriggerState` or rename `Trigger` to `TriggerState` but everybody is used to `Trigger` so it would be easier for th whole team to keep it

core/src/main/java/io/kestra/core/models/triggers/TriggerId.java
fhussonnois: The `TriggerId` interface follows the same pattern than the existing `FlowId` interface (which also have a minimal default implementation).

The purpose of the default implementation is mostly for serialization to only serialize/deserialize what is really needed. That's why we can do `Trigger.of(abstractTrigger)` like it's already done with `FlowId`.  So it's not a new pattern.

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
fhussonnois: Yes, you're right! It could also be too many triggers, the warning isn't specific enough, I'll update to include that cause. In any case, the idea is to have a WARNING when there is a problem.

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
fhussonnois: Previously, the `ScheduledExecutor` was responsible for periodically triggering the evaluation of triggers (every second). In the new design, however, we now have an event loop that is responsible not only for evaluating triggers but also for processing trigger-related events and updating their state. In addition, to that the event-loop may have to also handle actions when an VNode rebalancing happens;

Using a `ScheduledExecutor` would force us to wait for the next interval (e.g., 1 second) even when new events are already available. With the event-loop model, we can react to incoming events immediately instead of waiting for the next scheduled tick. 

On a Kestra instance with many triggers, the event-loop thread will likely stay busy handling events continuously.

fhussonnois: The `TriggerSchedulingLoop` itself is not responsible for producing trigger events, but only for processing them. A producer thread (i.e., the queue consumer) asynchronously pushes events into the blocking queue.

The `Lock` and `Condition` implement a wait/signal mechanism, allowing the loop to be notified immediately when events are available, avoiding busy-waiting. Using a BlockingQueue ensures thread-safe access when the loop drains all events for processing. So, even with a single looping thread, the BlockingQueue provides safe inter-thread communication between the producer (e.g., JdbcPoller, Kafka Consumer) and the `TriggerSchedulingLoop` thread. So "overkill", I would not say, it’s there for thread safety.

scheduler/src/main/java/io/kestra/scheduler/TriggerEventHandler.java
fhussonnois: No, this is not possible! `Trigger -> VNode --> Scheduler -> SchedulingLoop Thread.` For a trigger, all events are processed by a single thread in FIFO.

scheduler/src/main/java/io/kestra/scheduler/DefaultScheduler.java
fhussonnois: Yes, this is already tracked in the follow-up issue.

core/src/main/java/io/kestra/scheduler/vnodes/VNodeConsistentHashRing.java
fhussonnois: Nop, here, we just iterate over virtual nodes not server node. Then, for each we compute a `hash key` which is placed on the ring and we retrieve the nearest server node. This ring logic is actually materialize by the tailMap (there is actually no real ring data structure).

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
loicmathieu: > Previously, the ScheduledExecutor was responsible for periodically triggering the evaluation of triggers (every second)

This behavior is very important as it kind of guarantee that triggers would be evaluated "on time".
This is pretty important for Schedule trigger when we want them to be trigger at the planned time (of course, best effort).

Here, for example, if there is a lot of events on triggers (for ex a lot of TriggerExecutionTerminated events) it could potentially delay trigger evaluation if they share the same event loop which would be annoying?.

loicmathieu: OK, I need to have a second look at this

core/src/main/java/io/kestra/scheduler/vnodes/internals/DefaultVNodesAssigner.java
fhussonnois: `currentControllerIdAndEpoch` is `null` when no vNode rebalance is in progress. The variable is cleared once the rebalance is completed.

scheduler/src/test/java/io/kestra/scheduler/utils/InMemoryTriggerStateStore.java
fhussonnois: Moved to test package. I preferred the "inmemory" utility class to H2 in order to simplify unit testing of the scheduler.

core/src/test/java/io/kestra/scheduler/utils/InMemorySchedulerEventQueue.java
fhussonnois: Moved to test package. I preferred the "inmemory" utility class to H2 to simplify unit testing of the scheduler and avoid conflict between tests.

core/src/main/java/io/kestra/scheduler/vnodes/internals/StaticVNodesAssigner.java
fhussonnois: Micronaut seems to be lost when this class is in a test package. Could be investigated later

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
fhussonnois: No strong reason, assignment is expected to complete in a few milliseconds so no need to wait more;

scheduler/src/main/java/io/kestra/scheduler/stores/CachedFlowMetaStore.java
fhussonnois: You're right! There is definitely an optimization that could be made to avoid duplicating states, I had already that in mind. But this is only relevant for standalone deployments. So can be done later.

In a distributed/production deployment, a scheduler only needs to load the flows and triggers assigned to its vNodes, which is why the `FlowMetaStore` and `TriggerStateStore` are vNode-aware. TriggerStates and Flows are loaded in parallel when the SchedulingEventLoops are assigned to vNodes. This reduces the time required to initialize a scheduler instance; previously, initialization had to wait for the FlowListener to become ready.

When a vNode is revoked, all entries associated with that vNode are immediately discarded, reducing memory usage.

scheduler/src/main/java/io/kestra/scheduler/stores/CachedTriggerStateStore.java
fhussonnois: Answered above!

core/src/main/java/io/kestra/scheduler/internals/DefaultVNodesAssigner.java
fhussonnois: Cleaned the class for clarity!

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
fhussonnois: You're right, this is indeed the main trade-off of moving to an event-loop model, and it's something I had already identified earlier. This is why, in the follow-up issue, I explicitly mentioned the need to benchmark the scheduler under heavy load to determine the theoretical limits of a single event loop. For example, we may find that beyond a certain volume (e.g., 100k trigger-related events per second), the event-loop timing begins to drift.

Once we know that threshold, we can evaluate mitigation strategies. The simplest option would be to distribute the workload across multiple scheduling loops or scheduler instances so that trigger evaluations are not delayed by spikes in event processing.

loicmathieu (approved):
LGTM

PR #12939, opened 2025-11-13
fix(flow): flow trigger with both conditions and preconditions

### LINKED ISSUE #12905: Multiple triggers containing a mix of preconditions and conditions causing multiple executions
## Describe the issue

For flows that have multiple flow triggers defined, and containing a mix of `preconditions` and `conditions`, when one of trigger conditions are met, the flow will be executed twice. Each triggered execution has the same trigger details.

## Steps to reproduce

1. Define `flow_a` which is executed manually

```yaml
id: flow_a
namespace: dev.general

tasks:
  - id: hello
    message: Hello World! 🚀
```

2. Define `flow_b` which is executed when `flow_a` completes either successfully or unsuccessfully.

```yaml
id: flow_b
namespace: dev.general

tasks:
  - id: hello
    message: Hello World! 🚀

triggers:
  - id: on_completion
    states: [SUCCESS]
    conditions:
      - type: io.kestra.plugin.core.condition.ExecutionFlow
        namespace: dev.general
        flowId: flow_a

  - id: on_failure
    preconditions:
      id: flowsFailure
      flows:
        - namespace: dev.general
          flowId: flow_a
          states: [FAILED]
```

3. Execute `flow_a`, and observe that `flow_b` will be triggered twice, whereby both executions show a trigger id value of `on_completion`.

Obviously, in the above example one could change them to either both use `conditions` or `preconditions` and all would be well. However, in our scenario, we have triggers defined as below:

```yaml
triggers:
  - id: on_orchestration_flow_completion
    states: [RUNNING, SKIPPED, FAILED, WARNING, SUCCESS, KILLED, CANCELLED, RETRIED]
    conditions:
      - type: io.kestra.plugin.core.condition.ExecutionLabels
        labels:
          type: orchestration

  - id: on_dbt_scheduler_flow_completion
    states: [RUNNING, SKIPPED, FAILED, WARNING, SUCCESS, KILLED, CANCELLED, RETRIED]
    preconditions:
      id: test
      where:
        - id: filtering
          filters:
            - field: FLOW_ID
              type: STARTS_WITH
              value: dbt_scheduler
```

From what I understand, I don't believe `preconditions` supports filtering on `labels`, and `conditions` doesn't support flow ID text filtering, so I'm not sure if a conversion would work here. Of course, if you have any advice on how this could be achieved, that would be great.

## Environment

Kestra Version: `0.24.14`

loicmathieu:
When a flow have both a condition and a precondition, the condition was evaluated twice which lead to double execution triggered.

Fixes #12905

### DIFF
--- a/core/src/test/java/io/kestra/core/runners/AbstractRunnerTest.java
+++ b/core/src/test/java/io/kestra/core/runners/AbstractRunnerTest.java
@@ -274,6 +274,12 @@ public abstract class AbstractRunnerTest {
     }
 
     @Test
+    @LoadFlows({"flows/valids/flow-trigger-mixed-conditions-flow-a.yaml", "flows/valids/flow-trigger-mixed-conditions-flow-listen.yaml"})
+    void flowTriggerMixedConditions() throws Exception {
+        multipleConditionTriggerCaseTest.flowTriggerMixedConditions();
+    }
+
+    @Test
     @LoadFlows({"flows/valids/each-null.yaml"})
     void eachWithNull() throws Exception {
         EachSequentialTest.eachNullTest(runnerUtils, logsQueue);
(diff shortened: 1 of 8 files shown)


AcevedoR (approved):

loicmathieu:
It has been backported to 0.24, 1.0 and 1.1

PR #12946, opened 2025-11-14
chore: update the volume path in compose file

Pradumnasaraf:
Update the compose config so that we avoid the path with the version in postgresql. Also, it's recommended by PostgreSQL.

Screenshot 2025-11-14 at 1 42 43 PM

### DIFF
--- a/docker-compose-dind.yml
+++ b/docker-compose-dind.yml
@@ -17,7 +17,7 @@ services:
   postgres:
     image: postgres:18
     volumes:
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
This is strange, I changed it a few days ago and at that time Postgres team advise to use version specific directory.

If it works like that, it's better!

Pradumnasaraf:
Hey @loicmathieu, some of the checks are failing, not due to the chnages. It's fine to merge the PR?

PR #12948, opened 2025-11-14
chore(system): reduce repository code duplication between OSS and EE

loicmathieu:
Part-of: https://github.com/kestra-io/kestra-ee/issues/1684


AcevedoR (approved):
It would be cool to have this backported to 1.0 and 1.1

loicmathieu:
@AcevedoR I worked on more refactoring so it would be too complex to backport, moreover we usually don't backport small refactoring like that.

PR #12973, opened 2025-11-14
docs(return): fix example

aj-emerich:


wrussell1999 (commented):
At the moment, all inputs are required so the conditional expression is kinda redundant. Maybe make the inputs optional?

core/src/main/java/io/kestra/plugin/core/debug/Return.java
```diff
@@ -44,15 +44,33 @@ Handy for troubleshooting (the rendered value is logged) or for generating small
...
-                  Bearer {{ inputs.token }}
-                  {%- elseif inputs.username is not empty and inputs.password is not empty -%}
-                  Basic {{ (inputs.username + ':' + inputs.password) | base64encode }}
-                  {%- endif -%}
+                id: return
+                namespace: company.team
+
+                inputs:
+                  - id: token
+                    type: STRING
```
aj-emerich, line 54:
```suggestion
                    type: STRING
                    required: false
```

aj-emerich: @wrussell1999 like so for optional inputs?

core/src/main/java/io/kestra/plugin/core/debug/Return.java
```diff
@@ -44,15 +44,33 @@ Handy for troubleshooting (the rendered value is logged) or for generating small
...
+                    displayName: "API Token"
+                  
+                  - id: username
+                    type: STRING
+                    displayName: "Username"
```
aj-emerich, line 60:
```suggestion
                    displayName: "Username"
                    required: false
```

core/src/main/java/io/kestra/plugin/core/debug/Return.java
```diff
@@ -44,15 +44,33 @@ Handy for troubleshooting (the rendered value is logged) or for generating small
...
+
+                  - id: password
+                    type: STRING
+                    displayName: "Password"
```
aj-emerich, line 65:
```suggestion
                    displayName: "Password"
                    required: false
```

wrussell1999 (approved):

PR #12976, opened 2025-11-14
fix(ui): Wrap header actions on Flows page for mobile view

### LINKED ISSUE #12748: Ensure that the sidebar collapse/expand toggle is always visible, even on tiny screens (e.g. on mobile)
Image

This happens on mobile because we have too many action buttons on this page. This will be partially addressed in https://github.com/kestra-io/kestra/issues/12747.

However, we should ensure that the sidebar collapse/expand toggle is always visible, even on tiny screens and even when there is more than 1 action button.

MSHIVVANI:
Screenshot from 2025-11-14 20-21-00

Closes #12748 
On the main "Flows" page, when viewed on a small screen (mobile resolution), the top action buttons ("Import", "Source search", "Create") are pushed off-screen and become inaccessible if the main navigation sidebar is opened. This PR fixes this responsive layout bug by adding CSS to the button container. On smaller screens, the buttons will now wrap and stack vertically, ensuring they remain visible and clickable at all times.

---

This is a visual UI fix that is not dependent on any specific flow content. It can be QAed with the following manual steps:

Navigate to the Flows page.

Open the browser's developer tools and switch to a responsive/mobile view (e.g., a width of 600px or less).

Open the main navigation sidebar on the left.

Observe (Before): The "+ Create" button is pushed off-screen and is not visible.

Observe (After): With the changes in this PR, the action buttons now correctly wrap and stack, and all buttons remain visible and clickable.

---


ui/src/components/flows/Flows.vue
```diff
@@ -679,4 +679,16 @@
 :deep(.flows-table) .el-scrollbar__thumb {
     background-color: var(--ks-border-active) !important;
 }
+:deep(.header-actions-list) {
```
Piyush-r-bhaskar: Do you really need `:deep` if the class is component level ?

Piyush-r-bhaskar (changes requested):

MSHIVVANI:
@Piyush-r-bhaskar Thank you for the suggestion! You're right, :deep wasn't necessary here. I've removed it and updated the PR.

ui/src/components/flows/Flows.vue
MilosPaunovic: If you've addressed any of the requested changes from the PR review, @MSHIVVANI, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

MSHIVVANI: Okayy!

Piyush-r-bhaskar (approved):
Thank You for your contribution 🚀

PR #12979, opened 2025-11-14
chore(system): share JDBC repository code in an abstract CRUD repository

loicmathieu:


jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
```diff
@@ -0,0 +1,436 @@
...
+ *
+ * @Override
+ * protected Condition defaultFilter() {
+ *     return DSL.trueCondition();
+ * }
+ * }</pre>
+ *
+ * @param <T> the type of the persisted entity.
+ */
+public abstract class AbstractJdbcCrudRepository<T> extends AbstractJdbcRepository {
```
fhussonnois, line 39: `<T>` could not be  `<T extend HasUUID>` ? This will allow us to have the `exists/delete` base methods

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
```diff
@@ -0,0 +1,436 @@
...
+        this.jdbcRepository = jdbcRepository;
+        this.queueService = queueService;
+    }
+
+    /**
+     * Creates an item: persist it inside the database and return it.
+     * It uses an insert on conflict update to avoid concurrent write issues.
+     */
+    public T create(T item) {
+        Map<Field<Object>, Object> fields = this.jdbcRepository.persistFields(item);
```
fhussonnois, line 56: Depending on if we can have a default `exists` method, it could be nice to throw an exception if the entity already exists.

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcExecutionRepository.java
```diff
@@ -130,48 +131,19 @@ private QueueInterface<Execution> executionQueue() {
...
-                    .select(field("value"))
-                    .from(this.jdbcRepository.getTable())
-                    .where(this.defaultFilter(tenantId, false))
-                    .and(field("namespace").eq(namespace))
-                    .and(field("flow_id").eq(flowId))
-                    .and(statesFilter(states))
-                    .orderBy(field("start_date").desc());
-                return this.jdbcRepository.fetchOne(from);
-            });
+        var condition = field("namespace").eq(namespace)
```
fhussonnois, line 143: When possible, we should use the defined static `Field` objects. Could be nice to always do that for all fields.

fhussonnois (approved):

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
fhussonnois: Here the `create` method could just call `save`

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
loicmathieu: I'll open an issue for that as I think not all our entities extend HasUID but I agree they should!
If they all extend it, we would be able to get rid of the QueueService which would also be a win

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
loicmathieu: Create cannot call save as some implementation override save with create which effectivement create an infinite loop :( 
Overall we should review all create/save/update impl and align them as we didn't always do the same.
I already openned an issue to tackle this late.

Some implementation throws if the entity exist, other not. I choose this current implementation as it's the most used but it is currently overriden very often.

What I think is that instead of throw if exist, use a real insert when we call create as if you query exist then insert, it's not safe, someone may have already inserted it in-between.

So, I agree with you, but it's not that simple and that's why I didn't did it and openned an issue about it.

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
loicmathieu: See https://github.com/kestra-io/kestra/issues/13014

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
loicmathieu: See https://github.com/kestra-io/kestra/issues/13002

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcExecutionRepository.java
loicmathieu: I did it for the Abstract CRUD repo but I'll let the change on all repositories for later, too much work...

PR #12980, opened 2025-11-14
fix(cli): no-args will properly return help

brian-mulier-p:
closing https://github.com/kestra-io/kestra-ee/issues/5842

### DIFF
--- a/cli/src/main/java/io/kestra/cli/App.java
+++ b/cli/src/main/java/io/kestra/cli/App.java
@@ -1,5 +1,6 @@
 package io.kestra.cli;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.kestra.cli.commands.configs.sys.ConfigCommand;
 import io.kestra.cli.commands.flows.FlowCommand;
 import io.kestra.cli.commands.migrations.MigrationCommand;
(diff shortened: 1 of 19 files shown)


AcevedoR (approved):

cli/src/main/java/io/kestra/cli/App.java
```diff
@@ -49,24 +49,50 @@
 @Introspected
 public class App implements Callable<Integer> {
     public static void main(String[] args) {
-        execute(App.class, new String [] { Environment.CLI }, args);
+        System.exit(cliRun(args));
+    }
+
+    public static int cliRun(String[] args, String... extraEnvironments) {
```
loicmathieu: Nitpick but would it be better named `runCommand` or `runCli`?

cli/src/main/java/io/kestra/cli/App.java
```diff
@@ -49,24 +49,50 @@
...
+        return cliRun(App.class, args, extraEnvironments);
+    }
+
+    public static int cliRun(Class<?> cls, String[] args, String... extraEnvironments) {
```
loicmathieu: It seems you never set any extraEnvironments? This parameter seems useless

cli/src/main/java/io/kestra/cli/App.java
```diff
@@ -77,33 +103,41 @@ protected static void execute(Class<?> cls, String[] environments, String... arg
         applicationContext.close();
 
         // exit code
-        System.exit(Objects.requireNonNullElse(exitCode, 0));
+        return Objects.requireNonNullElse(exitCode, 0);
```
loicmathieu:
```suggestion
        return exitCode, 0;
```

Exit code cannot be null here as it's set in all code path and, well, it's an int which cannot be null

loicmathieu (changes requested):

cli/src/main/java/io/kestra/cli/App.java
brian-mulier-p: I do for tests 
image

loicmathieu (approved):
LGTM

PR #12993, opened 2025-11-17
fix(scheduler): convert 'now' to UTC to avoid any offset error on next_execution_date

fhussonnois:
Related-to: kestra-io/kestra-ee#5611


Skraye (approved):
LGTM

PR #13004, opened 2025-11-17
fix(flows): synchronize unsaved popup with flow store

### LINKED ISSUE #12878: [Flows] False unsaved changes warning in Flow Editor
### Describe the issue

Even though there are no unsaved changes when exiting the flow editor, it still shows a warning saying there are unsaved changes, even though everything remains the same.

Interestingly, I’ve noticed this issue several times before, but it was hard to capture. This time, I managed to record it: https://jam.dev/c/2398de17-dfca-4cc2-8654-670c4da99807.

**Please don’t close this issue if it’s not reproducible, as it happens occasionally and is not easy to reproduce. We need to be double sure**

### Environment

- Kestra Version: develop (not version specfic, happening since 1.0)

elevatebart:
closes #12878


Pradumnasaraf (approved):
Looks good to merge! Thank you for the fix.

---

Tested with Docker image above.

MilosPaunovic (approved):

PR #13006, opened 2025-11-17
fix(scheduler): mysql convert 'now' to UTC to avoid any offset error on next_execution_date

fhussonnois:
Fixed a previous commit to only apply the change for MySQL

Related-to: kestra-io/kestra-ee#5611


Skraye (approved):
LGTM

PR #13023, opened 2025-11-18
fix: loop until avoid taskrun duplication +  handle submitted status

### LINKED ISSUE #12615: LoopUntil generates duplicated next tasks

Skraye:
As we currently can not handle a huge amount of taskrun, LoopUntil clean every taskrun on each loop and play again its task. 

Two errors were done : 
* In the `resolveWaitForNext`, if all tasks had been executed, we were sending again the first one, but this is unnecessary as said above when a new loop start again, we clean all previous taskrun so we always restart from the first one.

*  In the `handleChildWorkerTaskResult` from the ExecutorService, we add a case if LoopUntil had still some next taskrun, but we could only enter the if condition if all task of a loop were executed, so its was impossible to enter it, except if we had duplicated tasks.

close #12615

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowableUtils.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowableUtils.java
@@ -157,8 +157,6 @@ public class FlowableUtils {
 
             if (currentTasks.size() > lastIndex + 1) {
                 return Collections.singletonList(currentTasks.get(lastIndex + 1).toNextTaskRunIncrementIteration(execution, parentTaskRun.getIteration()));
-            } else {
-                return Collections.singletonList(currentTasks.getFirst().toNextTaskRunIncrementIteration(execution, parentTaskRun.getIteration()));
             }
         }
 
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
LGTM

PR #13025, opened 2025-11-18
Feat/entity delete

### LINKED ISSUE #13747: Dashboards don't report same values
### Describe the issue

Take this side by side view of the same flow. 

- The overview reports 8 failed executions
- The executions view reports 1

Both views are over the same timeframe. Executions were manually deleted via the Kestra UI so perhaps some data is still cached? 

Image

### Environment

- Kestra Version: 1.1.9

loicmathieu:
- Refactor Soft deletion and harmonize the implementation
- Remove the unused deleted column on metrics and logs

Fixes #13747

### DIFF
--- a/core/src/main/java/io/kestra/core/models/SoftDeletion.java
+++ b/core/src/main/java/io/kestra/core/models/SoftDeletion.java
@@ -1,0 +1,18 @@
+package io.kestra.core.models;
+
+/**
+ * This interface marks entities that implement a soft deletion mechanism.
+ * Soft deletion is based on a <code>deleted</code> field that is set to <code>true</code> when the entity is deleted.
+ * Physical deletion either never occurs or occurs in a dedicated purge mechanism.
+ */
+public interface SoftDeletion<T> {
+    /**
+     * Whether en entity is deleted or not.
+     */
+    boolean isDeleted();
+
+    /**
+     * Delete the current entity: set its <code>deleted</code> field to <code>true</code>.
+     */
+    T toDeleted();
+}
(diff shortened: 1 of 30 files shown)


core/src/main/java/io/kestra/core/models/SoftDeletion.java
```diff
@@ -0,0 +1,18 @@
+package io.kestra.core.models;
+
+/**
+ * This interface marks entities that implement a soft deletion mechanism.
+ * Soft deletion is based on a <code>deleted</code> field that is set to <code>true</code> when the entity is deleted.
+ * Physical deletion either never occurs or occurs in a dedicated purge mechanism.
+ */
+public interface SoftDeletion<T> {
```
fhussonnois: Personally, I would have named this interface `SoftDeletable` that seems more standard.

jdbc-h2/src/main/resources/migrations/h2/V1_53__logs_metrics_deleted.sql
```diff
@@ -0,0 +1,22 @@
+DROP INDEX logs_execution_id;
+DROP INDEX logs_execution_id__task_id;
+DROP INDEX logs_execution_id__taskrun_id;
+DROP INDEX logs_namespace_flow;
+
+ALTER table logs drop column "deleted";
+
+CREATE INDEX IF NOT EXISTS logs_execution_id ON logs ("execution_id");
```
fhussonnois, line 8: We must be careful when adding a new index, as this can impact migration. We must at least validate the existing table with a significant amount of data in order to get a rough idea of how long it might take.

fhussonnois (approved):
Approved, but the index subject must be validated internally

jdbc-h2/src/main/resources/migrations/h2/V1_53__logs_metrics_deleted.sql
loicmathieu: Yes, this is planned, I opened an internal discussion about it already.

core/src/main/java/io/kestra/core/models/SoftDeletion.java
loicmathieu: I hesitate a lot on the naming but I agree yours is best ;) 
I'll do the change

loicmathieu:
Decision has been to pursue with this PR but switch to asynchronous index creation

PR #13026, opened 2025-11-18
feat(ns-files): introduce revision history

brian-mulier-p:


ui/src/components/layout/Revisions.vue
```diff
@@ -0,0 +1,316 @@
...
+
+              if (newQuery.revisionRight !== oldQuery.revisionRight && newQuery.revisionRight) {
+                  const newRightIndex = revisionIndex(newQuery.revisionRight.toString());
+                  if (newRightIndex !== revisionRightIndex.value) {
+                      revisionRightIndex.value = newRightIndex;
+                  }
+              }
+          },
+          {deep: true}
+    );
```
elevatebart: I would wish for watch to be kept to a minimum as they cannot be optimized.
If you cannot keep them within `computed` values, please make sure as much as possible that they do one thing only. It makes them easier optimize and understand.

```suggestion
    watch(() => route.query.revisionLeft,
          (revisionLeft) => {
             
                  const newLeftIndex = revisionIndex(revisionLeft.toString());
                  if (newLeftIndex !== revisionLeftIndex.value) {
                      revisionLeftIndex.value = newLeftIndex;
                  }
              })
              
watch(() => route.query.revisionRight, (revisionRight) => {
                  const newRightIndex = revisionIndex(revisionRight.toString());
                  if (newRightIndex !== revisionRightIndex.value) {
                      revisionRightIndex.value = newRightIndex;
                  }
              
    });
```

ui/src/components/flows/FlowRevisions.vue
elevatebart, line 1: I love this knd of change ❤️

ui/tests/storybook/layout/Revisions.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+        await expect(
+            ([...canvasElement.querySelectorAll(".editor .monaco-editor")] as HTMLElement[])
+            .every(el => el.getAttribute("aria-uri")?.endsWith(`.${args.lang}`))
+        ).toBeTruthy();
+
+        let revisionSelectorsOptions = await selectorOptions(canvasElement);
+
+        await expect(revisionSelectorsOptions.length).toEqual(2);
+
+        let simplifiedOptions = getSimplifiedOptions(revisionSelectorsOptions);
```
elevatebart, line 95: In e2e and browser testing you will have to be very careful about creating functions. 
Some playwright commands auto-wait, some commands do not, but once a value is returned it will never be retried. Some object take a split second to appear and they get missed.

A good architecture would be to do the action or the check within the function. This way you would benefit from the magic auto-wait within the function.

Docs: https://playwright.dev/docs/actionability

elevatebart (changes requested):
Super PR!

Some small changes before I start testing for real

ui/src/components/layout/Revisions.vue
brian-mulier-p: Agree, I splitted them in 2 because there's no way around compute for that one.

core/src/main/java/io/kestra/core/models/namespaces/files/NamespaceFileMetadata.java
```diff
@@ -0,0 +1,133 @@
...
+    @Builder.Default
+    private Instant created = Instant.now();
+
+    @Nullable
+    private Instant updated;
+
+    private boolean deleted;
+
+    @JsonCreator
+    public NamespaceFileMetadata(String tenantId, String namespace, String path, String parentPath, Integer version, boolean last, Long size, Instant created, @Nullable Instant updated, boolean deleted) {
```
nkwiatkowski, line 59: maybe remove parentPath from this constructor, it's not used

core/src/main/java/io/kestra/core/models/namespaces/files/NamespaceFileMetadata.java
```diff
@@ -0,0 +1,133 @@
...
+        return NamespaceFileMetadata.builder()
+            .tenantId(tenantId)
+            .namespace(namespace)
+            .path(path)
+            .created(Instant.ofEpochMilli(fileAttributes.getCreationTime()))
+            .updated(Instant.ofEpochMilli(fileAttributes.getLastModifiedTime()))
+            .size(fileAttributes.getSize())
+            .version(1)
+            .build();
+    }
```
nkwiatkowski, line 112: I think you can add the parent path here, it's created from the path

core/src/main/java/io/kestra/core/models/namespaces/files/NamespaceFileMetadata.java
```diff
@@ -0,0 +1,133 @@
...
+
+    public static NamespaceFileMetadata of(String tenantId, NamespaceFile namespaceFile) {
+        return NamespaceFileMetadata.builder()
+            .tenantId(tenantId)
+            .namespace(namespaceFile.namespace())
+            .path(namespaceFile.path(true).toString())
+            .version(namespaceFile.version())
+            .build();
+    }
```
nkwiatkowski, line 100: I think you can add the parent path here, it's created from the path

core/src/test/java/io/kestra/core/repositories/AbstractNamespaceFileMetadataRepositoryTest.java
```diff
@@ -0,0 +1,216 @@
...
+        assertThat(found.get().getPath()).isEqualTo(path);
+        // Soft delete
+        assertThat(found.get().getVersion()).isEqualTo(1);
+        assertThat(found.get().isLast()).isTrue();
+        assertThat(found.get().isDeleted()).isTrue();
+        assertThat(found.get().getUpdated()).isAfter(beforeDeleteUpdateDate);
+    }
+
+    @Test
+    void findWithFilters() throws IOException {
```
nkwiatkowski: For this kind of test, it's easier to make a parametrized test to validate all working query, and all non working queries

nkwiatkowski (approved):
LGTM

core/src/main/java/io/kestra/core/models/namespaces/files/NamespaceFileMetadata.java
brian-mulier-p: It's already done in the constructor from above, I don't want to bother setting it up everywhere, it's a computed property

core/src/main/java/io/kestra/core/models/namespaces/files/NamespaceFileMetadata.java
brian-mulier-p: Unfortunately, Lombok is not of the same advice :/ 
image

core/src/test/java/io/kestra/core/repositories/AbstractNamespaceFileMetadataRepositoryTest.java
brian-mulier-p: I've added a new ParameterizedTest with a bunch of scenario :+1:

ui/tests/storybook/layout/Revisions.stories.tsx
brian-mulier-p: I'm not sure to understand, I've tested them carefully and I added wait everywhere it's needed imo so idk where is the catch?

brian-mulier-p:
@elevatebart if you want to give it another look round feel free, I've added a comment on one of yours where I'm not sure to understand the trap

ui/tests/storybook/layout/Revisions.stories.tsx
elevatebart: Playwright offers some simplicity in the code so as to avoid having to explicitly wait.

The more functions and methods you create the more you risk short circuiting these behaviors and having to add `wait` in your code. I would rather have less code maintained by us and more code maintained at MS than the opposite. 

Anyway, it is a nitpick and I do not see your tests as flaky as they are.

elevatebart (approved):

PR #12809, opened 2025-11-08
fix: day-week-in-month-date-time-between-failures

hemantmm:
closes: #8230

### What changes are being made and why?

✅ Fixed dynamic property rendering for date in `DayWeekInMonth` and `DateTimeBetween` conditions.
✅ Updated usage to `.render(date).as(String.class, vars).orElseThrow()` for correct value extraction.
✅ Ensured tests pass after migration from `String date` to `Property<String> date`.

---

Fix:

Updated the rendering logic for dynamic date properties in schedule conditions:

- Migrated `date` field from `String` to `Property<String>` in both classes.
- Changed rendering from `runContext.render(date, vars)` to `runContext.render(date).as(String.class, vars).orElseThrow()`.

PR #13027, opened 2025-11-18
Fix/gh 10900

fhussonnois:


loicmathieu (approved):
LGMT.

I wonder if it's not the same cause as https://github.com/kestra-io/kestra/pull/12809 so you may want to have a look at this PR.

PR #13067, opened 2025-11-19
feat(flows): improve human task in bulk method, kill and force run

nkwiatkowski:
- fix #https://github.com/kestra-io/kestra-ee/issues/5655


loicmathieu (approved):
LGTM

PR #13069, opened 2025-11-19
chore(system): extract queue consumers processing into message handlers

loicmathieu:


loicmathieu:
Superceded by https://github.com/kestra-io/kestra/pull/13094

PR #13072, opened 2025-11-19
fix(execution): improve property skip cache

loicmathieu:
When using Property.ofExpression(), the cache should never be used as this is usually used as providing a default value inside a task, which can change from rendering to rendering as it's an expression.

Also retain skipCache in a boolean so it can be rendered more than 2 times ans still skip the cache.

It should prevent future issues like #13027

### DIFF
--- a/core/src/main/java/io/kestra/core/models/property/Property.java
+++ b/core/src/main/java/io/kestra/core/models/property/Property.java
@@ -35,7 +35,6 @@ import static io.kestra.core.utils.Rethrow.throwFunction;
 @JsonDeserialize(using = Property.PropertyDeserializer.class)
 @JsonSerialize(using = Property.PropertySerializer.class)
 @Builder
-@NoArgsConstructor
 @AllArgsConstructor(access = AccessLevel.PACKAGE)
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/kestra/core/models/property/Property.java
```diff
@@ -61,12 +61,23 @@ public class Property<T> {
     // Note: when not used, this constructor would not be deleted but made private so it can only be used by ofExpression(String) and the deserializer
     public Property(String expression) {
         this.expression = expression;
+        this.skipCache = false;
```
fhussonnois: delegate to the new constructor you added:  

```
public Property(String expression) {
  this(expression, false);
}
```

fhussonnois (approved):

core/src/main/java/io/kestra/core/models/property/Property.java
loicmathieu: Agreed, updated

loicmathieu:
Backported to 1.1 and 1.0

PR #13075, opened 2025-11-19
tests(e2e): reproduce and fix flake

elevatebart:
- add the possibility to use a `.env` file to load login pass and use login/pass/baseurl and namespace
- add a `await page.waitForURL("**/flows?filters*");` before clicking on the create button to avoid being hit by redirection


AcevedoR:
Nice! the issue would exist on EE tests too right ?

elevatebart:
> Nice! the issue would exist on EE tests too right ?

Probably but with the fix added, since there is no more setTimeout, less likely to happen.
We should still make sure it's fixed though.

Also it would be a good idea to document the env thing and make it work on EE.

Piyush-r-bhaskar (approved):

PR #13076, opened 2025-11-19
feat(flows): add new checks condition on flow

fhussonnois:
Adds new property 'checks' on flow in order to allow pre-conditions to be evaluated before execution

Fixes: kestra-io/kestra-ee#5759

**QA**

Flow example: 
```yaml
id: monkey_246433
namespace: company.team
inputs:
  - id: name
    type: STRING
checks:
  - message: "Sorry, this flow can only be executed with 'Kestra'"
    condition: "{{ (inputs.name | upper ) == 'KESTRA'}}"
    style: ERROR
    behavior: BLOCK_EXECUTION
tasks:
  - id: hello
    message: Hello World! 🚀
```

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/Flow.java
+++ b/core/src/main/java/io/kestra/core/models/flows/Flow.java
@@ -11,6 +11,7 @@ import com.fasterxml.jackson.databind.introspect.JacksonAnnotationIntrospector;
 import io.kestra.core.exceptions.InternalException;
 import io.kestra.core.models.HasUID;
 import io.kestra.core.models.annotations.PluginProperty;
+import io.kestra.core.models.flows.check.Check;
 import io.kestra.core.models.flows.sla.SLA;
 import io.kestra.core.models.listeners.Listener;
 import io.kestra.core.models.tasks.FlowableTask;
@@ -129,6 +130,14 @@ public class Flow extends AbstractFlow implements HasUID {
     @Valid
     @PluginProperty
     List<SLA> sla;
+    
+    @Schema(
+        title = "Conditions evaluated before the flow is executed.",
+        description = "A list of conditions that are evaluated before the flow is executed.  If no checks are defined, the flow executes normally."
+    )
+    @Valid
+    @PluginProperty
+    List<Check> checks;
 
     public Stream<String> allTypes() {
         return Stream.of(
--- a/core/src/main/java/io/kestra/core/services/FlowService.java
+++ b/core/src/main/java/io/kestra/core/services/FlowService.java
@@ -2,20 +2,26 @@ package io.kestra.core.services;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import io.kestra.core.exceptions.FlowProcessingException;
+import io.kestra.core.exceptions.IllegalVariableEvaluationException;
 import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.flows.*;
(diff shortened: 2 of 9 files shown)


core/src/main/java/io/kestra/core/models/flows/Flow.java
```diff
@@ -129,6 +130,10 @@ public List<PluginDefault> getTaskDefaults() {
     @Valid
     @PluginProperty
     List<SLA> sla;
+
```
AcevedoR, line 133: It would be cool to add your documentation here, so it can be in the json schema, and so later displayed in the UI editor ?

AcevedoR: it would be useful to have one full example of how to use this field, and the information that only inputs are available

core/src/main/java/io/kestra/core/services/FlowService.java
```diff
@@ -84,7 +94,44 @@ private FlowRepositoryInterface repository() {
...
+            RunContext runContext = runContextFactory.get().of(flow, Map.of("inputs", inputs));
+            List<Check> falseConditions = new ArrayList<>();
+            for (Check check : flow.getChecks()) {
+                try {
+                    boolean result = Boolean.TRUE.equals(runContext.renderTyped(check.getCondition()));
+                    if (!result) {
+                        falseConditions.add(check);
+                    }
+                } catch (IllegalVariableEvaluationException e) {
+                    falseConditions.add(Check
```
AcevedoR, line 128: maybe you could add a debug log with exception full stacktrace if one day the exception message is not enough to debug an issue

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -2571,7 +2594,8 @@ public record ApiValidateExecutionInputsResponse(
         @Parameter(description = "The namespace")
         String namespace,
         @Parameter(description = "The flow's inputs")
-        List<ApiInputAndValue> inputs
+        List<ApiInputAndValue> inputs,
+        List<ApiCheckFailure> checks
```
AcevedoR, line 2601: I think this will impact the SDK we need to check if either: 
- it is only a compile issue then no issue > we will update the SDK with this change for the release
- or it break deserialisation for old Kestra instances trying to use the SDK (1.0 or 1.1), then it is a breaking change. In that case we might be able to find a way to make it backward/forward compatible @Skraye, otherwise > breaking change

AcevedoR (changes requested):
Seems good to me besides doc comments!
But we need to check the impact of the API change for the SDK

core/src/main/java/io/kestra/core/models/flows/Flow.java
fhussonnois: Added some docs. 

Actually there is more than `inputs`. But this will be documented by the product team. For the moment, we don't have doc example on the flow classes directly so I'll not start with that PR. That's something that can be done later for all props and part of SDK if it's really necessary.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
AcevedoR: we checked! its good

AcevedoR (approved):
LGTM!

PR #13085, opened 2025-11-20
docs(foreach): improve explanation of concurrencyLimit semantics

proddata:


core/src/main/java/io/kestra/plugin/core/flow/ForEach.java
```diff
@@ -36,23 +36,26 @@
     description = """
         You can control how many task groups are executed concurrently by setting the `concurrencyLimit` property. \
 
-        - If you set the `concurrencyLimit` property to `0`, Kestra will execute all task groups concurrently for all values. \
+        - A `concurrencyLimit` of `0` means no limit — all task groups run in parallel. \
 
-        - If you set the `concurrencyLimit` property to `1`, Kestra will execute each task group one after the other starting with the task group for the first value in the list. \
+        - A `concurrencyLimit` of `1` means full serialization — only one task group runs at a time, in order. \
+
+        - A `concurrencyLimit` greater than `1` allows up to that number of task groups to run in parallel, with the rest queued until capacity becomes available. \
```
loicmathieu: `queued` is a resererved work for an execution state that is restricted to flow-level concurrency when the behavior is QUEUE.
I'm not comptfortable to use it here as taskruns will not be queued, they would just not be created at all.

core/src/main/java/io/kestra/plugin/core/flow/ForEach.java
```diff
@@ -210,12 +213,14 @@ public class ForEach extends Sequential implements FlowableTask<VoidOutput> {
...
-        description = """
-        If you set the `concurrencyLimit` property to 0, Kestra will execute all task groups concurrently for all values (zero limits!). \
+      title = "The number of concurrent task groups for each value in the `values` array",
+      description = """
+        A `concurrencyLimit` of 0 means no limit — all task groups run in parallel.
 
+        A `concurrencyLimit` of 1 means full serialization — only one task group runs at a time, in order.
 
-        If you set the `concurrencyLimit` property to 1, Kestra will execute each task group one after the other starting with the first value in the list (limit concurrency to one task group that can be actively running at any time)."""
+        A `concurrencyLimit` greater than 1 allows up to the specified number of task groups to run in parallel, with the remaining ones queued until capacity is free.
```
loicmathieu: Same here, please don't use queued

loicmathieu (changes requested):

loicmathieu (approved):
LGTM

PR #13094, opened 2025-11-20
chore(system): extract queue consumers processing into message handlers

loicmathieu:
Split the Executor into multiple pieces, each queue consumer is now implemented in a separate message handler.

### DIFF
--- a/executor/src/main/java/io/kestra/executor/handler/ExecutionEventMessageHandler.java
+++ b/executor/src/main/java/io/kestra/executor/handler/ExecutionEventMessageHandler.java
@@ -1,0 +1,278 @@
+package io.kestra.executor.handler;
+
+import io.kestra.core.exceptions.InternalException;
+import io.kestra.core.models.executions.Execution;
+import io.kestra.core.models.executions.LogEntry;
+import io.kestra.core.models.executions.TaskRun;
+import io.kestra.core.models.executions.TaskRunAttempt;
+import io.kestra.core.models.flows.FlowId;
+import io.kestra.core.models.flows.FlowWithSource;
+import io.kestra.core.models.flows.State;
+import io.kestra.core.models.flows.sla.ExecutionMonitoringSLA;
+import io.kestra.core.models.flows.sla.SLA;
+import io.kestra.core.models.flows.sla.SLAMonitor;
+import io.kestra.core.models.tasks.WorkerGroup;
+import io.kestra.core.queues.QueueException;
+import io.kestra.core.queues.QueueFactoryInterface;
+import io.kestra.core.queues.QueueInterface;
+import io.kestra.core.runners.*;
+import io.kestra.core.services.WorkerGroupService;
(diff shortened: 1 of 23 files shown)


executor/src/main/java/io/kestra/executor/handler/ExecutionEventMessageHandler.java
```diff
@@ -0,0 +1,278 @@
...
+
+    private final Tracer tracer;
+
+    @Inject
+    public ExecutionEventMessageHandler(TracerFactory tracerFactory) {
+        this.tracer = tracerFactory.getTracer(DefaultExecutor.class, "EXECUTOR");
+    }
+
+    @Override
+    public Optional<ExecutorContext> handle(ExecutionEvent message) {
```
fhussonnois, line 86: The `handle` method is a bit too large and difficult to read. I think most of the code in each `if` block could have its own dedicated method.

executor/src/main/java/io/kestra/executor/handler/ExecutionEventMessageHandler.java
```diff
@@ -0,0 +1,278 @@
...
+        return executionStateStore.lock(message.executionId(), execution -> tracer.inCurrentContext(
+            execution,
+            FlowId.uidWithoutRevision(execution),
+            () -> {
+                try {
+                    final FlowWithSource flow = flowMetaStore.findByExecutionThenInjectDefaults(execution).orElseThrow();
+                    ExecutorContext executor = new ExecutorContext(execution, flow);
+
+                    // schedule it for later if needed
+                    if (execution.getState().getCurrent() == State.Type.CREATED && execution.getScheduleDate() != null && execution.getScheduleDate().isAfter(Instant.now())) {
```
fhussonnois, line 96: Also, to improve readability and facilitate maintenance, it might be useful to create a `var` with an explicit name for such conditition, e.g.: `isCreatedButDelayedExecution`. Same comment for other conditions below.

fhussonnois (approved):

executor/src/main/java/io/kestra/executor/handler/ExecutionEventMessageHandler.java
loicmathieu: This method has been extracted as is but I agree that it would benefits for more refactoring.

loicmathieu: For now I prefer to keep it "as before" to be able to more easily report changes from 1.0 as I know it will happen so it would be easier to find the corresponding line of code.

executor/src/main/java/io/kestra/executor/handler/ExecutionEventMessageHandler.java
loicmathieu: For now I prefer to keep it "as before" to be able to more easily report changes from 1.0 as I know it will happen so it would be easier to find the corresponding line of code.

PR #13105, opened 2025-11-21
validate executionId before sending response

### LINKED ISSUE #13101: Follow Execution always return a fake empty Execution object in first
### Describe the issue

As you can see in the image below, the start event will send a fake execution with incomplete data.

This was not an issue because it have no side effect on the UI, but people using the SDK/API can have trouble to find why the first object is empty.

Also if the execution ID send for the follow endpoint is wrong/fake, we still have this fake object response, which is error prone.

Image

### Environment

- Kestra Version: develop

coder-shankar:
### ✨ Description

What does this PR change?  
It will validate the executionId first before emitting start event
It will send execution event with actual data

### 🔗 Related Issue

fix: https://github.com/kestra-io/kestra/issues/13101

### 🛠️ Backend Checklist

### 📝 Additional Notes
Screenshot:
Screenshot 2025-11-21 at 11 07 30


loicmathieu (changes requested):
This event must be send as early as possible to avoid potential resource leak (SSE socket and Java thread).

coder-shankar:
Event must be send as early as possible to avoid potential resource leak (SSE socket and Java thread).

PR #13114, opened 2025-11-22
feat(core): Support Ion Binary format and Ion Hash in FileSerde

### LINKED ISSUE #3155: Change ION to binary format to support compression and reduce file size when processing large datasets + add Ion Hash
ION supports compression that we could use when changing to binary format - no impact to the end user other than faster downloads/uploads due to the smaller file size

Additionally, it would be beneficial to store a hash value for each ION file in internal storage to make it easy for the user to perform operations such as checking whether a file has changed compared to the last value in the State Store/KV Store.
https://amazon-ion.github.io/ion-hash/

bikashagarwala0:
This PR addresses **Issue #3155** by updating the core `FileSerde` utility to support Amazon Ion Binary format (for compression/performance) and Ion Hash (for data integrity/deduplication).

### Changes

  * **Refactored `FileSerde`:**
      * Added `writeAllBinary(OutputStream, ...)`: Writes data in Ion Binary format and calculates the SHA-256 Ion Hash on the fly. Returns a `Pair<Count, Hash>`.
      * Updated `readAll` / `reader`: Now accepts `InputStream` instead of `Reader`. This enables Jackson/Ion to **auto-detect** whether the source is Text (legacy) or Binary (new), ensuring full backward compatibility.
      * Marked legacy `writeAll(Writer, ...)` and `reader(BufferedReader)` methods as `@Deprecated`.
  * **Updated Consumers:**
      * Refactored `IonFileRender` (File Preview), `FromIonFunction`,  `URIFetcher`, and `WorkingDirectory`  to use the new `InputStream`-based methods to support reading both text and binary files.
  * **Tests:**
      * Added `writeAllBinary_checkFormatAndHash` to `FileSerdeTest` to verify binary magic cookies and hash generation.
      * Updated `PropertyTest` to verify `writeAllBinary` functionality.

### Note on Integration

This PR implements the low-level serialization capabilities required by the issue. The integration into the KV Store/State Store (to persist and compare the generated hash for deduplication) can be implemented in a follow-up PR using these new methods.

### Closes

Closes #3155


core/src/main/java/io/kestra/core/runners/pebble/functions/FromIonFunction.java
```diff
@@ -38,7 +40,10 @@ public Object execute(Map<String, Object> args, PebbleTemplate self, EvaluationC
             try {
                 String ion = (String) args.get("ion");;
 
-                Flux<Object> flux = FileSerde.readAll(new BufferedReader(new StringReader(ion)));
+                Flux<Object> flux = FileSerde.readAll(
+                    new ByteArrayInputStream(ion.getBytes(StandardCharsets.UTF_8)),
```
loicmathieu: Would be better to wrap it into a `BufferedInputStream` for better performance with a buffer size of `FileSerde.BUFFER_SIZE`.

core/src/main/java/io/kestra/core/serializers/FileSerde.java
```diff
@@ -157,8 +167,67 @@ public static <T> Flux<T> readAll(MappingIterator<T> mappingIterator) throws IOE
     }
 
     /**
-     * For performance, it is advised to wrap the writer inside a BufferedWriter, see {@link #BUFFER_SIZE}.
```
loicmathieu, line 126: I think this still stands no?

core/src/main/java/io/kestra/core/serializers/FileSerde.java
```diff
@@ -157,8 +167,67 @@ public static <T> Flux<T> readAll(MappingIterator<T> mappingIterator) throws IOE
...
+
+    public static <T> Flux<T> readAll(ObjectMapper objectMapper, InputStream input, Class<T> type) throws IOException {
+        return readAll(objectMapper, input, (Object) type);
+    }
+
+    public static <T> Flux<T> readAll(ObjectMapper objectMapper, InputStream input, TypeReference<T> type) throws IOException {
+        return readAll(objectMapper, input, (Object) type);
+    }
+
+    private static <T> Flux<T> readAll(ObjectMapper objectMapper, InputStream input, Object type) throws IOException {
```
loicmathieu: I prefer the previous construct where we create the mapping iterator in the caller method and pass it here.
This avoids the if/else

core/src/main/java/io/kestra/core/serializers/FileSerde.java
```diff
@@ -179,6 +248,60 @@ public static <T> Mono<Long> writeAll(Flux<T> values, SequenceWriter seqWriter)
             .count();
     }
 
+    /**
+     * Writes a Flux of values to an OutputStream in Ion Binary format and returns the count and SHA-256 hash.
+     * * @param output The raw output stream (do not wrap in Writer)
+     * @param values The data to write
+     * @return A Mono containing a Map.Entry where Key = Count (Long) and Value = SHA-256 Hash (String)
+     */
+    public static <T> Mono<AbstractMap.SimpleEntry<Long, String>> writeAllBinary(OutputStream output, Flux<T> values) {
```
loicmathieu: Using a sequence writter here should bring better performance

webserver/src/main/java/io/kestra/webserver/utils/filepreview/IonFileRender.java
```diff
@@ -22,11 +20,9 @@ public class IonFileRender extends FileRender {
     }
 
     private void renderContent(InputStream filestream) throws IOException {
```
loicmathieu, line 24: I didn't check the usage but here we no longuer close the inputstream so if it's not done at use site it would leak

core/src/test/java/io/kestra/core/serializers/FileSerdeTest.java
loicmathieu, line 1: Can you add tests for writting with the old mechanism and reading with the new and the opposite so we're sure it's backward compatible?

loicmathieu (changes requested):
Hi,
Thanks a lot for your contribution.
You added a computation of a hash? As far as I remember this was not asked so why did you take the decision to add that as it's not used anywhere in our code but it complexifies the implementation.

Also, it would be great to have pereformance comparison with the previous methods, I can do them later if you didn't had time.

Last thinbgn I wonder if switching from a reader to an inputstream and a writter to an outputstream may have side effect :thinking:

bikashagarwala0:
Hi @loicmathieu, thanks for the review!

Regarding the Hash: It was actually part of the original issue description (#3155): "Additionally, it would be beneficial to store a hash value for each ION file... to perform operations such as checking whether a file has changed". That is why I included it in the writer.

InputStream vs Reader: Switching to InputStream is mandatory for Ion Binary. Reader and BufferedReader are character-based classes intended for Text. If we pass binary data through a Reader, it attempts to decode bytes into Characters (UTF-8), which corrupts the binary structure of Ion files. InputStream allows us to read raw bytes, and Jackson/Ion automatically detects if those bytes represent Text or Binary.

Performance: I can try to run a small benchmark, but theoretically, Ion Binary is significantly faster to read/write (no text parsing overhead) and smaller in size (compression), which reduces I/O.

I will address the code-specific comments (buffering, closing streams, sequence writer) in a new commit shortly!

core/src/main/java/io/kestra/core/serializers/FileSerde.java
bikashagarwala0: Done

core/src/test/java/io/kestra/core/serializers/FileSerdeTest.java
bikashagarwala0: Done

webserver/src/main/java/io/kestra/webserver/utils/filepreview/IonFileRender.java
bikashagarwala0: Fixed

core/src/main/java/io/kestra/core/serializers/FileSerde.java
bikashagarwala0: Done

core/src/main/java/io/kestra/core/serializers/FileSerde.java
bikashagarwala0: Added this one as well

core/src/main/java/io/kestra/core/runners/pebble/functions/FromIonFunction.java
bikashagarwala0: Done

core/src/main/java/io/kestra/core/serializers/FileSerde.java
MilosPaunovic: If you've addressed any of the requested changes from the PR review, @BikashAgarwala, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

core/src/main/java/io/kestra/core/serializers/FileSerde.java
loicmathieu, line 1: Can you deprecate all existing readAll/writeAll methods with `since="1.2"`?

core/src/main/java/io/kestra/plugin/core/flow/WorkingDirectory.java
```diff
@@ -346,7 +346,7 @@ public Outputs outputs(final RunContext runContext) throws IOException {
             return null;
         }
 
-        try(Reader is = new BufferedReader(new InputStreamReader(runContext.storage().getFile(uri)))) {
+        try (InputStream is = new BufferedInputStream(runContext.storage().getFile(uri))) {
```
loicmathieu: Please use `FileSerde.BUFFER_SIZE` as buffer size

loicmathieu (changes requested):
Can you also remove the hash calculation for now as we didn't store it anywhere so it's useless?
We would be able to add it again later if we decide to support it

core/src/main/java/io/kestra/core/serializers/FileSerde.java
```diff
@@ -106,43 +112,55 @@ private static <T> T convert(String row, Class<T> cls) throws JsonProcessingExce
 
     /**
      * For performance, it is advised to wrap the reader inside a BufferedReader, see {@link #BUFFER_SIZE}.
+     * @deprecated Use {@link #readAll(InputStream, Class)} instead.
```
loicmathieu:
```suggestion
     * @deprecated Use {@link #readAll(InputStream)} instead.
```

loicmathieu (changes requested):
For all read methods, can you add the following comment:

```
     * For performance, it is advised to wrap the input stream inside a BufferedInputStream, see {@link #BUFFER_SIZE}
 ```

File: core/src/main/java/io/kestra/core/serializers/FileSerde.java (hunk 7 above)
Comment (bikashagarwala0) [reply core/src/main/java/io/kestra/core/serializers/FileSerde.java]: Cannot resolve symbol 'readAll(InputStream)'

Comment (loicmathieu) [reply core/src/main/java/io/kestra/core/serializers/FileSerde.java]: It's because you forgot to implement it ;)

It should be more or less

```java
public static <T> Flux<T> readAll(InputStream inputStream) throws IOException {
        return readAll(DEFAULT_OBJECT_MAPPER, inputStream, DEFAULT_TYPE_REFERENCE);
    }
```

core/src/main/java/io/kestra/core/serializers/FileSerde.java
```diff
@@ -179,6 +258,48 @@ public static <T> Mono<Long> writeAll(Flux<T> values, SequenceWriter seqWriter)
             .count();
     }
 
+    /**
+     * Writes a Flux of values to an OutputStream in Ion Binary format.
+     * @param output The raw output stream (do not wrap in Writer)
+     * @param values The data to write
+     * @return A Mono containing the Count (Long)
+     */
+    public static <T> Mono<Long> writeAllBinary(OutputStream output, Flux<T> values) {
```
loicmathieu:
```suggestion
    public static <T> Mono<Long> writeAll(OutputStream output, Flux<T> values) {
```

loicmathieu: Would be more consistent

loicmathieu (changes requested):

loicmathieu (approved):
LGTM, thanks a lot.

I need to test it and verify the performance impact before merging it.

MilosPaunovic:
Hello @bikashagarwala0,

bikashagarwala0:
Sure @MilosPaunovic, I will resolve the merge conflict as soon as I get some time.

bikashagarwala0:
Hi @MilosPaunovic, I had resolved the merge conflicts, but the changes had not been tested from my side after the merge, as there were some version mismatches. 

I might have to set up the project once again from the beginning.

MilosPaunovic:
So should we close this PR and open a new one @bikashagarwala0?

bikashagarwala0:
Yeah @MilosPaunovic, Since this PR haven't been merged from a long time, and there were multiple changes made in the project in the mean time.

It will be better to close this PR as of now and open a new PR to solve this issue, if it's still relevant.

PR #13120, opened 2025-11-24
feat/new queue2

tchiotludo:


queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcSubscriber.java
```diff
@@ -0,0 +1,111 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import java.time.Duration;
+import java.time.ZonedDateTime;
+import java.util.List;
+
+import static io.kestra.core.utils.Rethrow.throwRunnable;
+
+@Slf4j
+public abstract class JdbcSubscriber<T extends GenericEvent> extends AbstractSubscriber<T> {
```
fhussonnois: I'm not fan of the multiple extends: `JdbcSubscriber` > `AbstractSubscriber`  > `AbstractQueue`. This introduces complexity and make hard to isolate tests for testing only the polling, pause, and resume logic.

It would have been preferable to keep the current pattern used by the `JdbcQueuePoller` class, which contains all the logic and can be tested independently of any actual queue implementation.

So a `JdbcSubscriber` would have been composed of a `JdbcQueuePoller`. And eventually we could have an `AbstractQueuePoller` to be used by other queue implementations.

jdbc/src/main/java/io/kestra/jdbc/JdbcQueueItem.java
```diff
@@ -0,0 +1,17 @@
+package io.kestra.jdbc;
+
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+import java.time.Instant;
+
+@NoArgsConstructor
+@Getter
+public class JdbcQueueItem {
```
fhussonnois: A Java `record` would probably have been preferable here.

queue-jdbc/src/main/java/io/kestra/queue/jdbc/AbstractJdbcQueue.java
```diff
@@ -0,0 +1,50 @@
...
+@Slf4j
+public abstract class AbstractJdbcQueue<T extends GenericEvent> extends AbstractQueue<T> {
+    protected final JdbcQueueClient jdbcQueueClient;
+
+    public AbstractJdbcQueue(Class<T> cls, QueueUtils queueUtils, JdbcQueueClient jdbcQueueClient) {
+        super(cls, queueUtils);
+        this.jdbcQueueClient = jdbcQueueClient;
+    }
+
+    public void internalEmit(@Nullable String key, T message) throws QueueException {
```
fhussonnois: It seems strange to name a method `internalSomething`  with public `visibility`.

queue-jdbc/src/main/java/io/kestra/queue/jdbc/JdbcQueueFactory.java
```diff
@@ -0,0 +1,54 @@
...
+    private QueueUtils queueUtils;
+
+    @Inject
+    private ApplicationContext applicationContext;
+
+    @Inject
+    private JdbcQueueClient jdbcQueueClient;
+
+    @PostConstruct
+    void init() {
```
fhussonnois, line 26: Is it not preferable to use the `@EachBean`Micronaut annotation which is designed for such use case ?  Also, when does Micronaut will close the queue (i.e., call the PreDestroy method) ?

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueClient.java
```diff
@@ -0,0 +1,198 @@
...
+                .from(this.jdbcRepository.getTable())
+                .where(io.kestra.jdbc.repository.AbstractJdbcRepository.field("type").eq(queueNameToType(queue)))
+                .fetchAny("max", Long.class);
+        });
+
+        if (initialOffset != null) {
+            maxOffset.set(initialOffset);
+        }
+
+        return maxOffset;
```
fhussonnois: Why are you returning an `AtomicReference` ? Just return the result of the query

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueClient.java
```diff
@@ -0,0 +1,198 @@
...
+
+                    delete.execute();
+                }
+            }
+
+            return queueItems.size();
+        });
+    }
+
+    public AtomicReference<Long> subscribeBroadcastMaxOffset(String queue) {
```
fhussonnois: The method should been named `fetchMaxOffset`

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueClient.java
```diff
@@ -0,0 +1,198 @@
...
+        });
+
+        if (initialOffset != null) {
+            maxOffset.set(initialOffset);
+        }
+
+        return maxOffset;
+    }
+
+    public Integer subscribeBroadcast(String queue, AtomicReference<Long> maxOffset, MessageConsumer<String, Exception> consumer) {
```
fhussonnois: Passing an `AtomicReference` as a method param looks like a code smells :) I didn't check yet how that method is used, bere there is probably something better to be done like returning the max offset of the method return.

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueClient.java
```diff
@@ -0,0 +1,198 @@
...
+    );
+
+    @Inject
+    @Named("queue")
+    private AbstractJdbcRepository<JdbcQueueItem> jdbcRepository;
+
+    @Inject
+    private JooqDSLContextWrapper dslContextWrapper;
+
+    @Inject
```
fhussonnois: Please use constructor dependency injection. First, because it's a best practice. Second, we should be able to properly create a new `JdbcQueueClient` when we want to have a dedicated table for a queue. We have already discussed several times that having all queues in one table is not always a good idea. It impacts performance and makes troubleshooting with customers more difficult.

queue/src/main/java/io/kestra/queue/QueueFactory.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+
+@Factory
+public class QueueFactory {
+    public final static String QUEUE_EXECUTOR = "queueExecutor";
+
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    public static <T extends GenericEvent> List<Class<T>> listAllEvent(ClassLoader classLoader, Class<T> eventClass) {
```
fhussonnois: Honestly, returning a hard-coded list of classes would have been much simpler and more explicit that using the bean introspection stuff of Micronaut.

queue/src/main/java/io/kestra/queue/QueueFactory.java
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+
+        return list;
+    }
+
+    @Named(QUEUE_EXECUTOR)
+    @Singleton
+    @Inject
+    public ExecutorService executorsUtils(ExecutorsUtils executorsUtils, QueueConfiguration queueConfiguration) {
```
fhussonnois: If the goal is only to inject `ExecutorService` into the `QueueUtil` class, simply inject `ExecutorsUtils` into `QueueUtil` to avoid over-fragmenting the code.

queue/src/main/java/io/kestra/queue/QueueUtils.java
```diff
@@ -0,0 +1,107 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.concurrent.ExecutorService;
+
+@Slf4j
+@Singleton
+public class QueueUtils {
```
fhussonnois: Usually `XxxUtils` class are static. So best to name that class `QueueService`

queue/src/main/java/io/kestra/queue/QueueUtils.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final ObjectMapper MAPPER = JacksonMapper.ofJson(false).copy();
+
+    @Inject
+    @Named(QueueFactory.QUEUE_EXECUTOR)
+    @Getter
+    ExecutorService executorService;
```
fhussonnois: please us constructor dependency injection

fhussonnois (changes requested):

jdbc-h2/src/main/resources/migrations/h2/V3_8__queue_v2.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS queue (
```
loicmathieu, line 1: We already have a table `queues`  and now we have a table `queue` !
To avoid any confusion I would have called it `queues_v2` as it's a multi-queue table. If we remove the old one at some point we would rename it and remove the `v2`.

queue/src/main/java/io/kestra/queue/QueueService.java
```diff
@@ -0,0 +1,107 @@
...
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.concurrent.ExecutorService;
+
+@Slf4j
+@Singleton
+public class QueueUtils {
+    private static final ObjectMapper MAPPER = JacksonMapper.ofJson(false).copy();
```
loicmathieu: You don't need to copy it if you didn't modify it

queue/src/main/java/io/kestra/queue/KeyedDispatchEvent.java
```diff
@@ -0,0 +1,4 @@
+package io.kestra.queue;
+
+public interface KeyedDispatchEvent extends DispatchEvent {
```
loicmathieu, line 3: Shouldn't we have a method to get the key?

queue/src/main/java/io/kestra/queue/GenericEvent.java
```diff
@@ -0,0 +1,5 @@
+package io.kestra.queue;
+
+public interface GenericEvent {
```
loicmathieu: I would only nam e it `Event` or `QueueEvent` as it's the parent of all queue events and not a generic vs type event

queue/src/main/java/io/kestra/queue/GenericQueueInterface.java
```diff
@@ -0,0 +1,5 @@
+package io.kestra.queue;
+
+public interface GenericQueueInterface<T extends GenericEvent> {
```
loicmathieu: I would have it named just `Queue`.
Personally I don't like `Interface` suffix but `QueueInterface` is also good for me

queue/src/main/java/io/kestra/queue/AbstractQueue.java
```diff
@@ -0,0 +1,32 @@
+package io.kestra.queue;
+
+import com.google.common.base.CaseFormat;
+import jakarta.annotation.Nullable;
+
+public abstract class AbstractQueue<T extends GenericEvent> {
```
loicmathieu: Shouldn't it impelment the parent queue interface?

queue/src/main/java/io/kestra/queue/AbstractQueue.java
```diff
@@ -0,0 +1,32 @@
...
+    protected final Class<T> cls;
+    protected final QueueUtils queueUtils;
```
loicmathieu: I find it strange to have the QueueUtils here but never us it. It's bad encapsulation should should only defined dependencies you use

queue/src/main/java/io/kestra/queue/AbstractSubscriber.java
```diff
@@ -0,0 +1,146 @@
...
+import io.kestra.core.queues.QueueException;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
+
+@Slf4j
+public abstract class AbstractSubscriber<T extends GenericEvent> extends AbstractQueue<T> implements QueueSubscriber<T> {
```
loicmathieu: A subscriber that extends a Queue this seems strange.

queue/src/main/java/io/kestra/queue/QueueFactory.java
loicmathieu: I agree we should not rely on Micronaut bean introspection too much.
in fact, looking at the code, I don't understand how the queue beans are created!

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcSubscriber.java
```diff
@@ -0,0 +1,111 @@
...
+        JdbcQueueClient jdbcQueueClient,
+        String queueName
+    ) {
+        super(cls, queueUtils);
+
+        this.jdbcQueueClient = jdbcQueueClient;
+        this.queueName = queueName;
+    }
+
+    protected abstract Integer pool(JdbcQueueClient.MessageConsumer<String, Exception> messageConsumer);
```
loicmathieu:
```suggestion
    protected abstract Integer poll(JdbcQueueClient.MessageConsumer<String, Exception> messageConsumer);
```

Or I didn't understand what it does ;)

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueClient.java
loicmathieu: I think we should decide of that, we either use a single tab or separated tab but I don't want to have a shared table + a dedicated table!
Anyway, this can be done as a followup

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueClient.java
```diff
@@ -0,0 +1,198 @@
...
+                List<Long> processedItems = queueItems
+                    .stream()
+                    .map(queueItem -> {
+                        Exception exception = consumer.apply(queueItem.getValue());
+                        return exception == null ? queueItem.getOffset() : null;
+                    })
+                    .filter(Objects::nonNull)
+                    .toList();
+
+                if (!processedItems.isEmpty()) {
```
loicmathieu, line 127: Here, if an exception occur we would not consume the item so it would becomes a poison pill.
Is it really what we want? In the previous JDBC impl it was not the case

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueClient.java
loicmathieu: So maxOffset in an IN/OUT parameter that's why you use an AtomicReference

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueClient.java
```diff
@@ -0,0 +1,198 @@
...
+            }
+
+            List<JdbcQueueItem> queueItems = select
+                .orderBy(io.kestra.jdbc.repository.AbstractJdbcRepository.field("offset").asc())
+                .limit(configuration.pollSize())
+                .forUpdate()
+                .skipLocked()
+                .fetchInto(JdbcQueueItem.class);
+
+            if (!queueItems.isEmpty()) {
```
loicmathieu, line 175: Broadcast are not deleted, so how would it be purged?

loicmathieu (changes requested):

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcSubscriber.java
tchiotludo: - `AbstractSubscriber` is used by all implementation and not managing any polling mechanism , only managing the current state of the subscriber, there is no polling logic inside 
- `AbstractQueue` is only managing queue name 
- the whole logic of polling is inside the `JdbcSubscriber` 
  - I found during the implementation that the best solution should be `QueueSubscriberPoller` that is a generic poll mechanism 
  - But I didn't found any queue that we planned that could benefits from this, the polling is often managed by the underlying drivers. 
  - So to keep it simple stupid, I keep the whole logic where we have only 1 use case, the abstraction is useless for now. 
  - We could reavaluate if we have another queue implementation that could benefit from this polling mechanism

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueClient.java
tchiotludo: the dedicated table per queue is still on the plan, but we need to move forward for now. 
The challenge is more about creating the table in a scalable way, is flyway the best solution? 
Since we are deleting in live inside the table, not even sure that the optimisation is useful

queue/src/main/java/io/kestra/queue/QueueFactory.java
tchiotludo: See `@EachBean` above

jdbc-h2/src/main/resources/migrations/h2/V3_8__queue_v2.sql
tchiotludo: My plan was to drop the `queues` before the release to avoid confusion

queue-jdbc/src/main/java/io/kestra/queue/jdbc/JdbcQueueFactory.java
tchiotludo: Same feeling first, but `@EachBean` require bean already created, mostly only configuration is possible. 
For now, Events are created by code only, no way to use it that I could found.

queue/src/main/java/io/kestra/queue/KeyedDispatchEvent.java
tchiotludo: - inherited from the `GenericEvent`, the key is only the key of the message. 
- The routing key is not stored inside the event
- the queue is managing the key how it's required
- I renamed all the occurence to `routingKey` to avoid confusion

queue/src/main/java/io/kestra/queue/AbstractQueue.java
tchiotludo: No I didn't see, since all method are protected

queue/src/main/java/io/kestra/queue/AbstractQueue.java
tchiotludo: It's used by all the implementation to have access to serialization etc ...

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueClient.java
tchiotludo: the consumer now have the responsibility to manage the [exception](https://github.com/kestra-io/kestra/blob/05a31a5f15269d23f3f6d1ba36f4f65ec47ccb47/queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcSubscriber.java#L61).

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueClient.java
tchiotludo: It should be done, I don't have a strong opinion for now what could be a good time, probably simply deleting all item older than 5 min is good enough?

loicmathieu: No because we could have late consumers.
That's why it should be a bit more complex, at least in JDBC where we still have some high cardinality messages that may be consumed by multiple consumers.
We can at least have a jdbc cleaner as before with one day by default, or just reuse the old one.
At some point we should also implement as before an eager deletion mechanism by key

tchiotludo: For Broadcast, the contract expected is "if you are here only", no more guarantee than that. The broadcast message must not be critical, so deleting after 5 min seems a viable options and really safe.

loicmathieu: We have one use case where we dispatch to a consumer group AND broadcast to all anonymous consumer.
Would it still be possible? Because if we would need to send two messages this would be an issue performance wise

tchiotludo: What is the use case for that? 
But clearly we have defined that it will be 2 messages. A double produce is preferable vs the complexity of double consume the same message

loicmathieu: The executor dispatch an `ExecutionEvent` to itself then broadcast the `ExecutionEvent` message to everybody so it could be listened by the follow endpoint.
I need to check but it also dispatch the ExecutionEvent to the Scheduler, we may change that to only dispatching ann ExecutionEnd message (I think Florian already do the change).

tchiotludo: `ExecutionEvent` will be probably multiple events: 
`WebserverExecutionEvent` for the webserver as broadcast
`ExecutorExecutionEvent` for the executor as dispatch
`SchedulerExecutionEnd` for the scheduler as dispatch or vnode dispatch

loicmathieu: This would have a risk of concerning performance impact

PR #13122, opened 2025-11-24
fix(triggers): migrate to new model and fix api

fhussonnois:
**Done**
* Rewrites TriggerRepositoryInterface tests
* Migrates Trigger model to TriggerState model
* Migrates next_execution_date to next_evaluation_epoch (fix timezone issue)
* Refactors and cleanup trigger REST APIs
* Migrates trigger.Toggle task to use new trigger event queue
* Removes legacy trigger queue
* Adds migration trigger script

**TODOs**
* Create a dedicated DTO to not return `TriggerState` object to UI. Will be done later on another PR.
* Handle UI changes for the `TriggerState`


scheduler/src/main/java/io/kestra/scheduler/TriggerEventHandler.java
```diff
@@ -85,12 +86,13 @@ public void handle(Clock clock, Integer vNode, TriggerEvent event) {
...
             case TriggerExecutionTerminated evt -> onTriggerExecutionTerminated(clock, evt);
             case TriggerEvaluated evt -> onTriggerEvaluated(clock, evt);
+            case TriggerReceived evt -> onTriggerReceived(clock, evt);
             // Commands
             case CreateBackfillTrigger evt -> onCreateBackfill(clock, evt);
             case SetPauseBackfillTrigger evt -> onSetTriggerDisable(clock, evt);
             case SetDisableTrigger evt -> onSetTriggerDisable(clock, evt);
             case ResetTrigger evt -> onResetTrigger(clock, evt);
-            default -> throw new IllegalStateException("Unexpected value: " + event);
+            default -> LOG.warn("Unhandled trigger event: {}", event);
```
loicmathieu: I prefer to throw if we recevie an unprocessable event. In fact it would be the default behavior if we use sealed classes ;)

webserver/src/main/java/io/kestra/webserver/models/api/ApiErrorResponse.java
```diff
@@ -0,0 +1,29 @@
+package io.kestra.webserver.models.api;
+
+import io.micronaut.http.HttpResponse;
+import io.micronaut.http.HttpStatus;
+
+import java.util.Objects;
+
+public class ApiErrorResponse {
```
loicmathieu, line 8: I think there already exist an exception for that: `HttpStatusException`.

core/src/main/java/io/kestra/core/services/TriggerStateService.java
```diff
@@ -0,0 +1,296 @@
...
+     * @param tenant  the tenant identifier.
+     * @param filters the filters to match triggers.
+     * @throws NotFoundException if no trigger can be found.
+     */
+    public int toggleAllTriggersMatching(String tenant, List<QueryFilter> filters, boolean disabled) {
+        return triggerRepository.find(tenant, filters).map(throwFunction(trigger -> {
+                try {
+                    toggleTriggerById(trigger, disabled);
+                    return 1;
+                } catch (NotFoundException | ConflictException e) {
```
loicmathieu: I think previously we fail in case there is a missing trigger

core/src/main/java/io/kestra/core/services/TriggerStateService.java
```diff
@@ -0,0 +1,296 @@
...
+                .builder()
+                .tenantId(triggerId.getTenantId())
+                .namespace(triggerId.getNamespace())
+                .flowId(triggerId.getFlowId())
+                .triggerId(triggerId.getTriggerId())
+                .build()
+            );
+            triggerEventQueue.send(new ResetTrigger(TriggerId.of(triggerId), Instant.now()));
+        } catch (QueueException e) {
+            throw new RuntimeException(e); // TODO
```
loicmathieu: Better to let it bubble up

core/src/main/java/io/kestra/core/services/TriggerStateService.java
```diff
@@ -0,0 +1,296 @@
...
+        Flux<TriggerId> flux = triggerRepository.find(tenant, filters).filter(TriggerState::isLocked).map(TriggerId::of);
+        return unlockAllTriggersByIds(flux);
+    }
+    
+    private int unlockAllTriggersByIds(final Flux<TriggerId> triggers) {
+        return triggers.map(trigger -> {
+                try {
+                    unlockTriggerById(trigger);
+                    return 1;
+                } catch (NotFoundException | ConflictException ignored) {
```
loicmathieu: I think before it throws

core/src/main/java/io/kestra/core/services/TriggerStateService.java
loicmathieu, line 1: I love having a service to avoid too much "business" code in controllers but I wonder if it should not be inside the webserver as it has dependencies with exceptions that was designed to be used in the API like NotFound and Conflict.
Anyway, I'm OK to let it here.

loicmathieu: Also, the PR is big so I may have miss it but do you forget to add a test for it?

loicmathieu (changes requested):
Overall I love the direction it takes, more clearer and with better separation of concern?

However, I'm a bit uneasy with the migration, I think we should move all migration related artefacts (Trigger class, V1 repository methods) into the CLI module maybe inside  a dedicated v1 package as we may have other things to move.

Another solution would be to use maps and jOOQ directly as what was done for some other migration.
Anyway, moving to the CLI module is a must have as this module is not in the classpath of others so there are less risks they would be used for anything else than the migratiopn

core/src/main/java/io/kestra/core/services/TriggerStateService.java
fhussonnois: The class is mainly tested by the controller. I haven't forgotten, but I took the shortest path ;). So, yes the service class should be tested but I will do it another PR.

Regarding the location of the class, let's postpone the decision when all the code base will be stable.

loicmathieu (approved):
LGTM

PR #13124, opened 2025-11-24
fix: orderField was ignored in the findAsync method

Skraye:


loicmathieu (commented):
Are you sure about your change?
Because select.orderBy update the select clause then return it so there should not be any difference here.

I use the same construct in find() and findOne() by the way

Skraye:
@loicmathieu Just tested and you are right! Closing it then as it dont make sens

PR #13127, opened 2025-11-24
feat(system): configure Docker Compose termination grace period to 6m

loicmathieu:
By default, Kestra has a termination grace period of 5m. To be sure all tasks are terminated, we need to configure Docker Compose with a termination grace period of more than that: 6m.

Part-of: https://github.com/kestra-io/kestra-ee/issues/5556


brian-mulier-p (approved):
LGTM

PR #13130, opened 2025-11-24
fix(ui): render level badge and header inline with message

### LINKED ISSUE #12736: Executions > Logs : reduce width of level
Image

The width of labels is taking 1/4 of the screen 

make them inline: 

Image

Raviikumar001:
closes #12736 

Summary: 
Rendered the log level badge, timestamp, and metadata inline with the message in the Execution > Logs view for a more compact and readable layout.

Changes made:

- Moved the log level tag `[.log-level]` into the `[.header]` markup so the level badge is part of the timestamp/meta group.
- Reworked layout from display: flex to inline-level layout for the badge and content:
  [.log-content] -> display: inline-block, vertical-align: middle
  [.log-level] -> display: inline-flex, removed top margin
  [.header] -> display: inline-flex, align-items: center
 

Improved cross-browser wrapping by adding:
  - overflow-wrap: anywhere; (better because it avoids splitting words in odd places).
  - word-break: break-word; (keeps older browser behavior as fallback)
  - min-width: 0; (prevents overflow in flex contexts when the element is allowed to shrink.)
  
  
image

image

### DIFF
--- a/ui/src/components/logs/LogLine.vue
+++ b/ui/src/components/logs/LogLine.vue
@@ -207,7 +207,6 @@ div.line {
         vertical-align: middle;
         /* prevent Firefox word breaks */
         max-width: calc(100% - 6rem);
-        /* allow breaking long words anywhere and avoid overflow across browsers */
         overflow-wrap: anywhere;
         word-break: break-word;
         min-width: 0;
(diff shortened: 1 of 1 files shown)


Piyush-r-bhaskar (approved):
Thanks for the contribution. Appreciate it!

PR #13137, opened 2025-11-25
feat: allows to export flows/triggers/executions as a streamed CSV files

### LINKED ISSUE #6592: Add an Export To Excel / CSV Feature
### Feature description

There are many use cases in which I would like to export Kestra data, for record keeping / audits, or to simply export the data and create a spreadsheet from it for a ToDo list. It's useful for reporting as well, where I can export the data to excel which is accessible by everyone, even if they're not a Kestra user. 

Therefore, it would be great if there was an export to csv / excel feature in the three main tabs : 
Flows, Executions, Administration - Triggers. 

The export feature would simply export the entire dataset (not paginated, Executions could be hundreds of thousands of rows) into a single CSV.

Skraye:
close #6592

companion PR : https://github.com/kestra-io/kestra-ee/pull/5952

### DIFF
--- a/core/src/main/java/io/kestra/core/repositories/ExecutionRepositoryInterface.java
+++ b/core/src/main/java/io/kestra/core/repositories/ExecutionRepositoryInterface.java
@@ -2,7 +2,6 @@ package io.kestra.core.repositories;
 
 import io.kestra.core.models.QueryFilter;
 import io.kestra.core.models.executions.Execution;
-import io.kestra.core.models.executions.TaskRun;
 import io.kestra.core.models.executions.statistics.DailyExecutionStatistics;
 import io.kestra.core.models.executions.statistics.ExecutionCount;
 import io.kestra.core.models.executions.statistics.Flow;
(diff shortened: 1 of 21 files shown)


jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
```diff
@@ -388,6 +390,14 @@ protected Flux<T> findAllAsync(Condition defaultFilter) {
             }), FluxSink.OverflowStrategy.BUFFER);
     }
 
+    public Flux<T> findAsync(String tenantId, @Nullable List<QueryFilter> filters, @Nullable String dateColumn, QueryFilter.Resource resource) {
```
loicmathieu: The CrudRepository was not intended to have too much methods in it, but only generic methods.
I think this is too specific and should not be added their, anyway, it's just a small wrapper over existing methods.

So please, avoid adding new methods in it or it would becomes in a few month an un-readable bag of specific methods.

Anyway, it should work with null conditions or you can create a condition that resolves to `noCondition` when your filter is null, this is how I did for other use cases

loicmathieu (changes requested):

loicmathieu (approved):
LGTM

PR #13147, opened 2025-11-25
Metastores v2

loicmathieu:
## Metastores v2:
All metastores now work with an in-memory cache.
They initialize the cache via the repository.
They consume the queue to update the cache in real-time.

The FlowMetaStore still has a fallback on the repository to be able to fetch old revisions or deleted flows.

## FlowListeners removal:
Remove the FlowListners:
- Replace usage of listen(allFlows) by the FlowMetaStore
- Replace usage of listen(previous, next) by consuming the flowQueue and accessing the previous flow via the repository.

For more information: https://www.notion.so/kestra-io/KIP-35-MetaStores-v2-and-the-FlowListeners-2b236907f7b580ce9b25d59787711d7e
Part-of: https://github.com/kestra-io/kestra-ee/issues/5751


scheduler/src/main/java/io/kestra/scheduler/pubsub/TriggerEventPublisher.java
```diff
@@ -1,53 +1,90 @@
...
     }
-    
-    @PostConstruct
-    @Override
-    public void run() {
-        this.flowListeners.listen((flow, previous) -> {
-            
+
+    // Make it a StartupEvent listener so it starts when Kestra start
+    @EventListener
```
fhussonnois, line 61: Maybe, should we just use `@Context` instead of `@Singleton` to avoid the listener. Currently, there is a hack in the `DefaultScheduler` where we inject the `TriggerEventPublisher` only to force Micronaut to instantiate the service

core/src/main/java/io/kestra/core/runners/DefaultFlowMetaStore.java
```diff
@@ -1,61 +1,93 @@
...
+    @VisibleForTesting
+    void clearCache() {
+        cache.clear();
+    }
+
+    @PostConstruct
+    void start() {
+        // listen to flow updates from the flow queue
+        this.cancellation = this.flowQueue.receive(either -> {
+            if (either.isRight()) {
```
fhussonnois, line 51: I'm not fully convinced that we should publish the entire flow to the queue.

I would prefer emitting simple CRUD events that include only the tenant / namespace / flow / revision identifiers. The `FlowMetaStore` could then fetch the corresponding `FlowSource` from the repository when needed. The repository would remain responsible for injecting defaults during deserialization from the database.

This approach would eliminate the need to duplicate default-injection logic in multiple places and would avoid potential deserialization issues when consuming flows. 

In addition, this avoid each consumer to determine what is the actual CRUD event behind the received flow.

It would also reinforce the repository (not the queue) as the single source of truth.

That said, I won't block the PR on this, as we need to move forward. This point should be discussed further and can be implemented later if we agree on the direction.

cli/src/main/java/io/kestra/cli/services/FileChangedEventListener.java
```diff
@@ -64,37 +66,34 @@ public void startListeningFromConfig() throws IOException, InterruptedException
...
             // Listen for new/updated/deleted flows
-            flowListeners.listen((current, previous) -> {
-                // If deleted
-                if (current.isDeleted()) {
-                    this.flows.stream().filter(flowWithPath -> flowWithPath.uidWithoutRevision().equals(current.uidWithoutRevision())).findFirst()
-                        .ifPresent(flowWithPath -> {
-                            deleteFile(Paths.get(flowWithPath.getPath()));
-                        });
-                    this.flows.removeIf(flowWithPath -> flowWithPath.uidWithoutRevision().equals(current.uidWithoutRevision()));
+            flowQueue.receive(either -> {
```
fhussonnois, line 76: See my comment below, but I think all of this could be simplified if the flow queue only contains CRUD event instead of the flow;

core/src/main/java/io/kestra/core/runners/DefaultFlowMetaStore.java
```diff
@@ -1,61 +1,93 @@
...
+@Slf4j
 public class DefaultFlowMetaStore implements FlowMetaStoreInterface {
     private final FlowRepositoryInterface flowRepository;
     private final PluginDefaultService pluginDefaultService;
+    private final ConcurrentHashMap<String, FlowWithSource> cache = new ConcurrentHashMap<>();
+    private final QueueInterface<FlowInterface> flowQueue;
 
-    @Setter
-    private List<FlowWithSource> allFlows;
+    private Runnable cancellation;
```
fhussonnois: I prefer that we now encapsulate these `Runnable` objects in the new `Disposable` class in order to obtain something more typed, which could be confusing for new developers/contributors (and of course for consistency reasons).

core/src/main/java/io/kestra/core/utils/QueueCache.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A cache backed by a queue.
+ * @param <T> the item of the cache
+ */
+@Slf4j
+public class QueueCache<T extends DeletedInterface & HasUID> implements AutoCloseable {
+    private final Map<String, T> cache = new ConcurrentHashMap<>();
+    private final Runnable cancellation;
```
fhussonnois: Use the `Disposable` class here for consistency

fhussonnois (changes requested):

scheduler/src/main/java/io/kestra/scheduler/pubsub/TriggerEventPublisher.java
loicmathieu: I think the order is a bit different.
Context would be created before regular beans and start eagerly.
Here, the bean would be created with regular beans, and then it would start at application start instead of waiting to be wired up. This is a bit later.
And it's clearer that we want this method to be called at startup, the fact that context beans trigger their postconstruct immediately is because other beans have optimization to not be created at all if not injected somewhere.

I think context beans should be limited to where it is really needed

core/src/main/java/io/kestra/core/runners/DefaultFlowMetaStore.java
loicmathieu: I think this is a job for a followup issue as we would need to change a lot of things, especially in tests for that.
I would move that to queue 2.0 where we would take the opportunity to review each queue and decide how to implement it in 2.0.

cli/src/main/java/io/kestra/cli/services/FileChangedEventListener.java
loicmathieu: Yes, but it's a big rework that I'll let for queue 2.0

core/src/main/java/io/kestra/core/runners/DefaultFlowMetaStore.java
loicmathieu: Changed to Disposable in all 3 places

core/src/main/java/io/kestra/core/utils/QueueCache.java
loicmathieu: Changed to Disposable in all 3 places

fhussonnois (approved):

PR #13177, opened 2025-11-26
Fix/concurrent test best practices

brian-mulier-p:


jdbc/src/test/java/io/kestra/jdbc/runner/AbstractJdbcDeserializationIssuesTest.java
```diff
@@ -35,8 +32,6 @@ public abstract class AbstractJdbcDeserializationIssuesTest {
 
     @BeforeAll
```
nkwiatkowski: You can drop the whole beforeEach here

webserver/src/test/java/io/kestra/webserver/controllers/api/FlowControllerTest.java
```diff
@@ -97,9 +93,6 @@ public static void beforeAll() {
 
     @BeforeEach
     protected void init() throws IOException, URISyntaxException {
-        jdbcTestUtils.drop();
-        jdbcTestUtils.migrate();
-
         TestsUtils.loads(MAIN_TENANT, repositoryLoader);
```
nkwiatkowski, line 105: this can't stay in the before each, you will load all flows for each tests, and it will fail the second time. We should always use @LoadFlows annotation and never use all flows

webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionControllerTest.java
```diff
@@ -68,9 +64,6 @@ class ExecutionControllerTest {
     @SneakyThrows
     @BeforeEach
     protected void setup() {
-        jdbcTestUtils.drop();
-        jdbcTestUtils.migrate();
-
         TestsUtils.loads(MAIN_TENANT, repositoryLoader);
```
nkwiatkowski, line 77: This can't stay here, it will fail the second time

nkwiatkowski (approved):
LGTM

PR #13179, opened 2025-11-26
fix(execution): failed flowable should also have a failed attempt

### LINKED ISSUE #12614: Flowable Task fails but sub tasks state in gantt/log view is still running
### Describe the issue

Even though the task failed (expression invalid), the If task has failed state, but expanded it says Running and continues to keep the execution time going up on the task, despite execution in failed state.

Image

Image

```yaml
id: error_in_flow
namespace: company.team

tasks:
  - id: if
    condition: "{{ inputs.run }}"
    then:
      - id: hello
        message: Hello World! 🚀
```

### Environment

- Kestra Version: `develop` EE

loicmathieu:
When a flowable fail, it should also switch its attempts to FAILED.

Fixes #12614


Skraye (approved):
LGTM

PR #13181, opened 2025-11-26
refactor: unify management of anonymous routes

elevatebart:
Set all routes that should be anonymous as anonymous in the routes.js file instead of in multiple places.

This will avoid mistakes like redirect loops at setup


MilosPaunovic (approved):
Loving this! 🚀

PR #13185, opened 2025-11-26
Fix/retry with flowable

loicmathieu:
- Improve SKIPPED tasks by adding an attempt
- Fix Sequential/Dag/Parallel/... to ends properly when all subtasks are disabled

This also has the effect to fix https://github.com/kestra-io/kestra-ee/issues/5714 as it's due to a race and the new fix for ending when no more enabled subtasks work around that.


AcevedoR (approved):

loicmathieu:
BAckported to 1.0 and 1.1

Issue #13143, opened 2025-11-25
Backfills do not work with SELECT input


wrussell1999:
### Describe the issue

Execute backfill doesn't work:
<img width="850" height="363" alt="Image" src="https://github.com/user-attachments/assets/5e8ad386-a71a-43ad-9ad4-73faf23fdba0" />

If a `defaults` is set for the input, it will let you backfill but will ignore your input and only use the `defaults` value. It also ignores the triggers input.

```yaml
id: backfills_broken
namespace: company.team

inputs:
  - id: taxi
    type: SELECT
    values: [green, yellow]

tasks:
  - id: hello
    type: io.kestra.plugin.core.log.Log
    message: "{{ inputs.taxi }}"

triggers:
  - id: green
    type: io.kestra.plugin.core.trigger.Schedule
    cron: "0 10 1 * *"
    inputs:
      taxi: green
  
  - id: yellow
    type: io.kestra.plugin.core.trigger.Schedule
    cron: "0 9 1 * *"
    inputs:
      taxi: yellow
```

### Environment

- Kestra Version: `v1.1` aka `1.1.4`

wrussell1999:
Also, if you change the input, it now complains that the input is `null`

<img width="891" height="417" alt="Image" src="https://github.com/user-attachments/assets/255ddee4-1839-4b66-8a99-afad501d627c" />

AcevedoR:
the issue is the UI sending the wrong input

wrussell1999:
<img width="1711" height="1316" alt="Image" src="https://github.com/user-attachments/assets/567e751c-c177-4b41-90d4-7dd730478f43" />

Also broken on `v1.0` image

AcevedoR:
1.1 backport done, 1.0 required additional work :https://github.com/kestra-io/kestra/pull/13235

AcevedoR:
- merged to develop,
- backport 1.0
- backport 1.1

PR #13186, opened 2025-11-26
fix(backfills): inputs was always the default one in the ui

AcevedoR:
- fix https://github.com/kestra-io/kestra/issues/13143
- develop equivalent: https://github.com/kestra-io/kestra/pull/13217


elevatebart (approved):
fixed indeed

PR #13215, opened 2025-11-27
fix(core): JDBC Executor will consume executions from queue per execu…

brian-mulier-p:
…tionId in order

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
@@ -297,10 +297,23 @@ public class JdbcExecutor implements ExecutorInterface {
         this.receiveCancellations.addFirst(((JdbcQueue<Execution>) this.executionQueue).receiveBatch(
             Executor.class,
             executions -> {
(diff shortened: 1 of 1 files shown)


loicmathieu (changes requested):
Please move it to DRAFT until we benchmark it.

A few remarks:
- start left mdon'essages before as they are the only one that really count
- for right messages, you can then directly log them here (or call a dedicated method with it) instead of calling the `executionQueue()` method async and switch the executionQueue message method to directly receive an exectution: no need to have async here as it would be counter-performant and makes the code harder to read

loicmathieu:
Overall, performance seems to increase, I have some doubt that it increase in scenario where no concurrent processing exist but on a ForEach loop with 100 iterations and unlimitted concurrency time decrease from 2.9s to 1.9s.

Raw result:
BEFORE:
1k: 143ms
2k: 146ms
3k: 238ms
4k: 858ms
5k: 2.4s
ForEach:  2.9s

AFTER:
1k: 139ms
2k: 143ms
3k: 184ms
4k: 160ms
5k: 1.5s
ForEach:  1.9s

So let's discuss this soon, it seems interesting ;)

loicmathieu (approved):
LGTM

PR #13220, opened 2025-11-27
fix(core): concurrency limit on JDBC was decrementing when using FAIL…

### LINKED ISSUE #13141: Concurrency Limit doesn't work for new executions after first execution reaches its limit
### Describe the issue

Steps to reproduce:
- Run the `getting_started` workflow below 4 times
- The 3rd execution should fail because of the concurrency limit
- The 4th should fail assuming the first and second workflow are still running. However, it runs.

It doesn't work for `FAIL` or `CANCEL` behaviours. It does work as expected for queue.

You can use this workflow to have it run 4 executions back to back so the concurrency limit is hit:
```yaml
id: test
namespace: zoomcamp

tasks:
  - id: foreach
    values: ["Will", "Bob", "Dave", "Sarah"]
    tasks:
      - id: subflow
        namespace: zoomcamp
        flowId: getting_started
        inputs:
          name: "{{ taskrun.value }}"
        transmitFailed: false
        wait: false
```

The concurrency view even shows it's wrong:
Image

Workflow which reproduces the bug:
```yaml
id: getting_started
namespace: zoomcamp

inputs:
  - id: name
    type: STRING
    defaults: Will

concurrency:
  behavior: FAIL
  limit: 2

variables:
  welcome_message: "Hello, {{ inputs.name }}!"

tasks:
  - id: hello_message
    message: "{{ render(vars.welcome_message) }}"

  - id: generate_output
    format: I was generated during this workflow.

  - id: sleep
    duration: PT30S

  - id: log_output
    message: "This is an output: {{ outputs.generate_output.value }}"

  - id: goodbye_message
    message: "Goodbye, {{ inputs.name }}!"

pluginDefaults:
  - type: io.kestra.plugin.core.log.Log
    values:
      level: ERROR

triggers:
  - id: schedule
    cron: "0 10 * * *"
    inputs:
      name: Sarah
    disabled: true
```

### Environment

- Kestra Version: `v1.1` aka `1.1.4`

brian-mulier-p:
… or CANCEL behavior

closes https://github.com/kestra-io/kestra/issues/13141

### DIFF
--- a/core/src/test/resources/flows/valids/flow-concurrency-fail.yml
+++ b/core/src/test/resources/flows/valids/flow-concurrency-fail.yml
@@ -8,4 +8,4 @@ concurrency:
 tasks:
   - id: sleep
     type: io.kestra.plugin.core.flow.Sleep
-    duration: PT2S
+    duration: PT5S
--- a/core/src/test/resources/flows/valids/flow-concurrency-fail.yml
+++ b/core/src/test/resources/flows/valids/flow-concurrency-fail.yml
@@ -8,4 +8,4 @@ concurrency:
 tasks:
   - id: sleep
     type: io.kestra.plugin.core.flow.Sleep
-    duration: PT5S
+    duration: PT10S
--- a/executor/src/main/java/io/kestra/executor/ExecutorService.java
+++ b/executor/src/main/java/io/kestra/executor/ExecutorService.java
@@ -125,7 +125,7 @@ public class ExecutorService {
                 case CANCEL ->
                     executionRunning
                         .withExecution(executionRunning.getExecution().withState(State.Type.CANCELLED))
-                        .withConcurrencyState(ExecutionRunning.ConcurrencyState.RUNNING);
+                        .withConcurrencyState(ExecutionRunning.ConcurrencyState.CANCELLED);
                 case FAIL -> {
                     var failedExecution = executionRunning.getExecution().failedExecutionFromExecutor(new IllegalStateException("Execution is FAILED due to concurrency limit exceeded"));
                     try {
@@ -135,7 +135,7 @@ public class ExecutorService {
                     }
                     yield executionRunning
                         .withExecution(failedExecution.getExecution())
-                        .withConcurrencyState(ExecutionRunning.ConcurrencyState.RUNNING);
+                        .withConcurrencyState(ExecutionRunning.ConcurrencyState.FAILED);
                 }
 
             };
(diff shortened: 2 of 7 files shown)


jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
```diff
@@ -1174,7 +1174,9 @@ private void toExecution(Executor executor, boolean ignoreFailure) {
                     boolean queuedThenKilled = execution.getState().getCurrent() == State.Type.KILLED
                         && execution.getState().getHistories().stream().anyMatch(h -> h.getState().isQueued())
                         && execution.getState().getHistories().stream().noneMatch(h -> h.getState().isRunning());
-                    if (!queuedThenKilled) {
+                    boolean concurrencyShortCircuitState = Concurrency.maybeConcurrencyLimited(execution.getState().getCurrent())
+                        && execution.getState().getHistories().get(execution.getState().getHistories().size() - 2).getState().isCreated();
```
loicmathieu, line 1175: I wonder if `execution.getState().getHistories().stream().noneMatch(h -> h.getState().isRunning());` would not be better / more explicit?

Anyway, I'm fine with that as long as you add a test or update the existing test for that

loicmathieu (changes requested):

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
brian-mulier-p: I went for that at first but I was afraid some executions would not be concurrent-related and yet not have running states in them so I went for a more precise approach

loicmathieu: Yeah, maybe.
I was wondering if we should not check inside the Executor if the previous `from` was from this method exactly.
But let's keep it simple as you did

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -127,7 +127,7 @@ public ExecutionRunning processExecutionRunning(FlowInterface flow, int runningC
                 case CANCEL ->
                     executionRunning
                         .withExecution(executionRunning.getExecution().withState(State.Type.CANCELLED))
-                        .withConcurrencyState(ExecutionRunning.ConcurrencyState.RUNNING);
```
loicmathieu, line 128: So this was needed in the previous impl because it meant send an executionrunning object to be processed later.
But I think in the current impl this field is not needed anymore.
Anyway, didn't hurt

core/src/test/resources/flows/valids/flow-concurrency-fail.yml
```diff
@@ -8,4 +8,4 @@ concurrency:
 tasks:
   - id: sleep
     type: io.kestra.plugin.core.flow.Sleep
-    duration: PT2S
```
loicmathieu, line 11: not fan, tests already too long

loicmathieu (approved):
LGTM

core/src/test/resources/flows/valids/flow-concurrency-fail.yml
brian-mulier-p: In fact I increased it to 10 seconds but I kill it as soon as the assert on concurrency-limited executions are passed because we don't care about it in fact. So we save time :)

https://github.com/kestra-io/kestra/pull/13220/commits/5656abf4e7c1fd33ef455c68320fc88544c045b4

PR #13236, opened 2025-11-28
feat(executions): add support to ACL check inside the run context

loicmathieu:


core/src/main/java/io/kestra/core/runners/RunContext.java
```diff
@@ -193,4 +195,18 @@ public record FlowInfo(String tenantId, String namespace, String id, Integer rev
...
 
     public abstract boolean isInitialized();
+
+    /**
+     * Finds a flow for a taskrun.
+     * This method checks that the flow related to the taskrun is allowed to access the target flow (via ACL check)
+     *
+     * @see #acl() which is used to check the taskrun ACL to the target flow.
+     */
+    public abstract FlowInterface findFlow(String tenantId, String namespace, String flowId, Optional<Integer> revision, TaskRun taskRun);
```
fhussonnois: I would prefer to have a facade class (e.g., `FlowAccessor` ) here to access flow like this: `runContext.flows.find()`, following the DSL style we already have with other resources (i.e, storage, workingDir, etc). Agree, that for the moment we only have a single method. But maybe if need to add one later this will have no impact on the RunContext class.

core/src/main/java/io/kestra/core/runners/AclChecker.java
```diff
@@ -0,0 +1,67 @@
+package io.kestra.core.runners;
+
+import io.kestra.core.services.FlowService;
+
+import java.util.List;
+import java.util.Objects;
+
+public interface AclChecker {
+
+    void checkAllowAllNamespace();
```
fhussonnois: I think, we should create a dedicated un-checked exception for thoses methods, e.g., `ResourceAccessDeniedException` that will be explicilty documented in this interface.

core/src/main/java/io/kestra/core/runners/AclChecker.java
```diff
@@ -0,0 +1,67 @@
+package io.kestra.core.runners;
+
+import io.kestra.core.services.FlowService;
+
+import java.util.List;
+import java.util.Objects;
+
+public interface AclChecker {
```
fhussonnois, line 13: I'm not very fan of the current interface. I think we could provide something more user-friendly with a nice DSL something like: 

```java
public interface AclChecker {

    AllowedResources allowNamespace(String namespace);
    AllowedResources allowNamespaces(List<String> namespaces);
    AllowedResources allowAllNamespaces();

    interface AllowedResources {
        void check();
   }
   
   //...
```

This will allow us to have:

* `runContext.acl().allowNamespace("my-ns").check()`
* `runContext.acl().allowNamespaces(List.of("ns1", "ns2")).check()`
* `runContext.acl().allowAllNamespaces().check()`

Then, in your `AclCheckerImpl` class

```java
    @Override
    public AllowedResources allowNamespace(String namespace) {
        return new AllowNamespace(runContext.flowInfo(), flowService, namespace);  // RunContext or FlowInfo must be injected the AclCheckerImpl constructror
    }
```

core/src/main/java/io/kestra/core/runners/AclChecker.java
```diff
@@ -0,0 +1,67 @@
...
+
+    void checkAllowNamespaces(List<String> namespaces);
+
+    void checkAllowResources(AllowedResources allowedResources);
+
+    interface AllowedResources {
+        void checkAllowed(RunContext.FlowInfo flowInfo);
+    }
+
+    class AllowAllNamespace implements AllowedResources {
```
fhussonnois: I find the class names here slightly confusing with plural vs singular. Maybe could we consider: 

* `SingleNamespace` instead of `AllowNamespace`
* `MultipleNamespaces` instead of `AllowNamespaces`
* `AllNamespaces` instead of `AllowAllNamespace`

`Allow` prefix is redondant for me and could be removed as we use inner classes here.

core/src/main/java/io/kestra/core/runners/AclChecker.java
```diff
@@ -0,0 +1,67 @@
...
+
+public interface AclChecker {
+
+    void checkAllowAllNamespace();
+
+    void checkAllowNamespace(String namespace);
+
+    void checkAllowNamespaces(List<String> namespaces);
+
+    void checkAllowResources(AllowedResources allowedResources);
```
fhussonnois: It seems this method is not required. Am I miss something ?

core/src/main/java/io/kestra/core/runners/AclChecker.java
```diff
@@ -0,0 +1,67 @@
...
+
+    interface AllowedResources {
+        void checkAllowed(RunContext.FlowInfo flowInfo);
```
fhussonnois: With my above proposal this becomes : 

```
interface AllowedResources {
        void check();
}
```

fhussonnois (changes requested):

core/src/main/java/io/kestra/core/runners/RunContext.java
loicmathieu: Yeah maybe

core/src/main/java/io/kestra/core/runners/AclChecker.java
loicmathieu: Agreed, I didn't refactor the underlying code, but it throws `IllegalArgumentException` which is not accurate.
I'll do the change.

core/src/main/java/io/kestra/core/runners/AclChecker.java
loicmathieu: Yeah I was not fan of it neither, that's why I asked you.
I was too focus on AllowedResources that I surface it in the interface at the wrong place, thanks for the idea.

core/src/main/java/io/kestra/core/runners/AclChecker.java
loicmathieu: This is for extension but I'll probably remove it in the next iteration phase

core/src/main/java/io/kestra/core/runners/RunContext.java
loicmathieu: Anyway, I'll move this to another PR

fhussonnois (approved):

PR #13243, opened 2025-11-28
fix(core): prevent log lines from breaking too early in firefox

### LINKED ISSUE #12736: Executions > Logs : reduce width of level

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/12736.

Related to https://github.com/kestra-io/kestra/pull/13130.


Piyush-r-bhaskar (approved):
Thanks for this 🚀 . It seems we can close the issue with this + the one merged earlier as it has what was asked in the issue.

PR #13255, opened 2025-12-01
feat(filters): add default visible filters for state and level

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra-ee/issues/5976

We now have a new prop `visibleByDefault` which if added with in any key in configurations file , will automatically add that filter to the filter list when the page loads as kinda `level in any`, even if no value to filter. Users can remove them, but they will reappear on page refresh.

https://github.com/user-attachments/assets/ef337411-1f8c-41cf-8793-e1a0cea8269a


MilosPaunovic (approved):
Love this approach! 🚀

PR #13267, opened 2025-12-01
chore(system): add `@CheckReturnValue` to AclCheck methods to be sure…

loicmathieu:
… `check()` is called


brian-mulier-p (approved):
LGTM

PR #13279, opened 2025-12-02
fix concurrency limit overload issue for jdbcexecutor

### LINKED ISSUE #13211: Concurrency limit is ignored when an execution is killed and other executions are queued
### Describe the issue

we found a flow concurrency bug. The issue is as follows:
When the workflow concurrency limit is set to 2:
```
id: hello-world
namespace: tutorial
tasks:
  - id: second_task
    commands:
      - sleep 1m
concurrency:
  limit: 2

```
- If we kill a running execution, it will start 2 new execution, resulting in 3 running instances - exceeding the concurrency limit.

Image

- If we kill a queued execution, it will start 1 new instance without validating against the concurrency limit.

Image

### Environment

kestra/kestra:latest

zhengchar:
### ✨ Description

Here are the three key updates to the toExecution method：

Regarding the judgment of the queueThenKilled state, the implementation of the h.getState().onlyRunning() method should only filter executions in the running state（exclude killing state）.
Concerning the concurrency limit, when processing execution instance in the queueThenKilled state, add logic to determine whether the execution's current state is killed and not​ triggered by a workerTaskResult message.
When checking the concurrency behavior, it is necessary to add a check for concurrency limit breaches (out-of-bounds judgment).

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/13211

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
@@ -176,7 +176,7 @@ public class JdbcExecutor implements ExecutorInterface {
 
     @Inject
     private AbstractJdbcWorkerJobRunningRepository workerJobRunningRepository;
-    
+
     @Inject
     private SLAMonitorStorage slaMonitorStorage;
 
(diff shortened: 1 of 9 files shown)


jdbc/src/main/java/io/kestra/jdbc/runner/AbstractJdbcConcurrencyLimitStorage.java
```diff
@@ -135,6 +135,23 @@ private Optional<ConcurrencyLimit> fetchOne(DSLContext dslContext, FlowInterface
         return this.jdbcRepository.fetchOne(select.forUpdate());
     }
 
+    public ConcurrencyLimit fetchOne(FlowInterface flow) {
```
loicmathieu: This would not be safe when called concurrently

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
```diff
@@ -1170,27 +1170,34 @@ private void toExecution(Executor executor, boolean ignoreFailure) {
...
-                            );
+                        var currentFlow = executor.getFlow();
+
+                        if (!(execution.getState().getCurrent() == State.Type.KILLED && executor.getFrom().contains("addWorkerTaskResult"))) {
+                            concurrencyLimitStorage.decrement(executor.getFlow());
+
+                            int currentConcurencyLimit = this.concurrencyLimitStorage.fetchOne(currentFlow).getRunning();
+
+                            if (executor.getFlow().getConcurrency().getBehavior() == Concurrency.Behavior.QUEUE
+                                && currentConcurencyLimit < currentFlow.getConcurrency().getLimit()) {
```
loicmathieu: This is not safe if called concurrently so I don't think it should be added here even if I understand the rational

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
```diff
@@ -1170,27 +1170,34 @@ private void toExecution(Executor executor, boolean ignoreFailure) {
...
-                                    executionQueue.emit(newExecution);
-                                    metricRegistry.counter(MetricRegistry.METRIC_EXECUTOR_EXECUTION_POPPED_COUNT, MetricRegistry.METRIC_EXECUTOR_EXECUTION_POPPED_COUNT_DESCRIPTION, metricRegistry.tags(newExecution)).increment();
-
-                                    // process flow triggers to allow listening on RUNNING state after a QUEUED state
-                                    processFlowTriggers(newExecution);
-                                })
-                            );
+                        var currentFlow = executor.getFlow();
+
+                        if (!(execution.getState().getCurrent() == State.Type.KILLED && executor.getFrom().contains("addWorkerTaskResult"))) {
```
loicmathieu: I'm not fan of using the `from` here, we never use this anywhere like that, it has been added only for debugging purpose I would not rely on it.
The purpose of the boolean just before are to do the same kind of check but based on the states which would be better

loicmathieu (changes requested):

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
loicmathieu: One solution would be to return the limit from the decrement method as it is done inside a FOR UPDATE so it's always safe

loicmathieu (approved):
LGTM, however as I contribute to this PR @brian-mulier-p can you also have a quick look please?

brian-mulier-p (approved):
LGTM

loicmathieu:
@zhengchar I refactor the code to use the original state istead of looking inside the from.
I also added test, if the CI pass I will integrate it inside today's release.

I remove the sagfeguard to only pop a new one if the concurrency level is under the max, I'll keep the idea for later I may add this safeguard in a followup PR: https://github.com/kestra-io/kestra/issues/13301

PR #13309, opened 2025-12-02
fix(executions): ability to skip messages failure in executor

brian-mulier-p:


core/src/main/java/io/kestra/core/utils/Rethrow.java
```diff
@@ -117,4 +120,17 @@ public static <E extends Exception> Runnable throwRunnable(RunnableChecked<E> ru
     private static <E extends Exception, R> R throwException(Exception exception) throws E {
         throw (E) exception;
     }
+
+    public static <T> Consumer<T> failAwareConsumer(Consumer<T> consumer, boolean failSilently) {
```
loicmathieu, line 124: As it's a bad practice to silently fail exception, please don't add it here but instead directly inside the JdbExecutor

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
```diff
@@ -241,7 +243,8 @@ public JdbcExecutor(
         final TracerFactory tracerFactory,
         final ExecutorsUtils executorsUtils,
         final MaintenanceService maintenanceService,
-        @Value("${kestra.jdbc.executor.thread-count:0}") final int threadCount
+        @Value("${kestra.jdbc.executor.thread-count:0}") final int threadCount,
+        @Value("${kestra.jdbc.executor.skip-failure:false}") final boolean skipFailure
```
loicmathieu, line 247: As we named this behavior fail-fast, I would prefer a `fail-fast` property true by default

loicmathieu (changes requested):
I'm still not convinced that we need this, but in case we come to an agreement and implement this I made a first pass of review

brian-mulier-p:
Closing as we will never merge it

PR #13312, opened 2025-12-03
feat(core): modernize Pebble cache

yuri1969:
### ✨ Description

What does this PR change?  

* Modernized the Pebble Template LRU cache from Guava to Caffeine.
* Registered Micrometer metrics of the cache.

The main goal was to be able to monitor the Pebble Template cache hitrate, size, etc.

### 🛠️ Backend Checklist

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/pebble/PebbleLruCache.java
+++ b/core/src/main/java/io/kestra/core/runners/pebble/PebbleLruCache.java
@@ -1,29 +1,30 @@
 package io.kestra.core.runners.pebble;
 
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import io.kestra.core.metrics.MetricRegistry;
+import io.micrometer.core.instrument.binder.cache.CaffeineCacheMetrics;
 import io.pebbletemplates.pebble.cache.PebbleCache;
 import io.pebbletemplates.pebble.template.PebbleTemplate;
-import lombok.extern.slf4j.Slf4j;
 
-import java.util.concurrent.ExecutionException;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/kestra/core/runners/pebble/PebbleLruCache.java
```diff
@@ -1,29 +1,30 @@
...
             .maximumSize(maximumSize)
+            .recordStats()
             .build();
     }
 
     @Override
     public PebbleTemplate computeIfAbsent(Object key, Function<? super Object, ? extends PebbleTemplate> mappingFunction) {
         try {
-            return cache.get(key, () -> mappingFunction.apply(key));
+            return cache.get(key, k -> mappingFunction.apply(key));
```
loicmathieu:
```suggestion
            return cache.get(key, k -> mappingFunction.apply(k));
```

Should avoid a copy-lambda

core/src/main/java/io/kestra/core/runners/pebble/PebbleLruCache.java
```diff
@@ -34,4 +35,13 @@ public PebbleTemplate computeIfAbsent(Object key, Function<? super Object, ? ext
     public void invalidateAll() {
         cache.invalidateAll();
     }
+
+    public void register(MetricRegistry metricRegistry) {
+        CaffeineCacheMetrics<Object, PebbleTemplate, Cache<Object, PebbleTemplate>> metrics = new CaffeineCacheMetrics<>(
+            cache,
+            "pebble.lru.cache",
```
loicmathieu: We usually monitor cache via: 

```
CaffeineCacheMetrics.monitor(meterRegistry, cache, "cache-name");
```

Would be better to do the same here for consistency, with for ex `pebble-template` as cache name

loicmathieu (changes requested):

loicmathieu (approved):
LGTM, thanks a lot

PR #13320, opened 2025-12-03
fix: correct regex when importing flow

Skraye:


core/src/main/java/io/kestra/core/models/HasSource.java
```diff
@@ -65,7 +64,7 @@ static void readSourceFile(final CompletedFileUpload fileUpload, final BiConsume
 
             if (isYAML(fileName)) {
                 byte[] bytes = inputStream.readAllBytes();
-                List<String> sources = List.of(new String(bytes).split("---"));
+                List<String> sources = List.of(new String(bytes).split("^---"));
```
Malaydewangan09: We can make it more safe and use` ^---\s*$`?
Currently it will match e.g `---something` which might not be correct?

Malaydewangan09 (approved):

PR #13355, opened 2025-12-03
Fix/scheduler wip

fhussonnois:


loicmathieu (approved):
I wonder if it would have not bean easier to store the UUID instead of an EventId to simpligy the object model an inline the UUID inside the generated JSON.

But other than that, I think having an UID for each event is a good thing, it may event be added as an abstract method in a parent interface as it would be great to have that for all events (including the future ExecutionEvent).

PR #13358, opened 2025-12-03
fix: unify label deserialization for numeric values in array form

### LINKED ISSUE #13359: Updating Execution State fails when label value is numeric
### Describe the issue

When updating an `Execution` state, deserialization fails if a label contains a numeric value:

###  Root Cause

`labels` are deserialized from **different sources** in different code paths, leading to inconsistent behavior:

| Scenario | Serialization source | Serialized shape | Deserialization branch | Type conversion |
|---------|----------------------|------------------|-------------------------|----------------|
| **Loading Flow (normal case)** | `flow.source`  | `Map<String, Object>` | `START_OBJECT` | ✔ has type conversion |
| **Updating Execution state** | `flow` object → JSON string | `List<Map<String, String>>` | `START_ARRAY` | ❌ no type conversion |

During state update, the label value becomes a non-String (e.g. Integer), but the array branch assumes all values are Strings and fails.

###  Proposed Fix

Apply the same allowed-type checking and conversion logic in the `START_ARRAY` branch as already done in the `START_OBJECT` branch.

I verified that after adding this conversion, updating state with numeric label values works correctly.

### PR:https://github.com/kestra-io/kestra/pull/13358

My PR adds the missing conversion logic to ensure consistent deserialization across both code paths.

### Environment

- Kestra Version: develop

thyw1:
Closes https://github.com/kestra-io/kestra/issues/13359.

## 🐛 Background

When updating an `Execution` state, deserialization fails if a label contains a
numeric value (e.g. `"value": 123`):

This does not happen in normal flow loading, which is why the issue was tricky to locate.

---

## 🔍 Root Cause

Kestra deserializes labels from **two different sources**, and they behave inconsistently:

| Scenario | Serialization source | Serialized form | Deserialization branch | Type conversion |
|---------|----------------------|------------------|-------------------------|----------------|
| **Flow loading** | `flow.source`  | `Map<String, Object>` | `START_OBJECT` | ✔ converts allowed types (String/Number/Boolean → String) |
| **Execution state update** | Flow object → JSON string | `List<Map<String, String>>` | `START_ARRAY` | ❌ assumes all values are Strings |

When updating execution state, numeric values remain as `Integer`, but the `START_ARRAY` branch tries to treat them as `String`, causing a cast failure.

---

## 💡 What This PR Does

This PR **unifies both code paths** by adding the same allowed-type conversion logic
to the `START_ARRAY` branch as used in the `START_OBJECT` branch.

This ensures:

- consistent behavior regardless of how labels are serialized  
- numeric / boolean values no longer cause deserialization failures  
- safer handling of future cases where labels are serialized as a list

---

## ✔️ Verification

- Tested with numeric label values (`Integer`, `Long`, `Double`, `Boolean`)  
- Execution state update no longer throws deserialization errors  
- Existing behavior for flows loaded from `flow.source` remains unchanged

---

## 🙏 Request for Review

Please help confirm that this behavior is desired.  
It makes the two deserialization paths consistent and prevents real production failures in state updates.

### DIFF
--- a/core/src/main/java/io/kestra/core/serializers/ListOrMapOfLabelDeserializer.java
+++ b/core/src/main/java/io/kestra/core/serializers/ListOrMapOfLabelDeserializer.java
@@ -26,7 +26,14 @@ public class ListOrMapOfLabelDeserializer extends JsonDeserializer<List<Label>> 
         else if (p.hasToken(JsonToken.START_ARRAY)) {
             // deserialize as list
             List<Map<String, String>> ret = ctxt.readValue(p, List.class);
-            return ret.stream().map(map -> new Label(map.get("key"), map.get("value"))).toList();
+            return ret.stream().map(map -> {
+                Object value = map.get("value");
+                if (isAllowedType(value)) {
+                    return new Label(map.get("key"), String.valueOf(value));
+                } else {
+                    throw new IllegalArgumentException("Unsupported type for key: " + map.get("key") + ", value: " + value);
+                }
+            }).toList();
         }
         else if (p.hasToken(JsonToken.START_OBJECT)) {
             // deserialize as map
--- a/core/src/test/java/io/kestra/core/serializers/YamlParserTest.java
+++ b/core/src/test/java/io/kestra/core/serializers/YamlParserTest.java
@@ -3,6 +3,7 @@ package io.kestra.core.serializers;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.kestra.core.models.flows.Flow;
+import io.kestra.core.models.flows.FlowWithSource;
 import io.kestra.core.models.flows.Input;
 import io.kestra.core.models.flows.Type;
 import io.kestra.core.models.flows.input.StringInput;
@@ -32,6 +33,7 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 @KestraTest
 class YamlParserTest {
     private static final ObjectMapper MAPPER = JacksonMapper.ofJson();
+    private static final ObjectMapper OBJECT_MAPPER = JacksonMapper.ofYaml().copy();
 
     @Inject
     private ModelValidator modelValidator;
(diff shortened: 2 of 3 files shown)


MilosPaunovic:
Is this solving any particular issue @thyw1?

core/src/main/java/io/kestra/core/serializers/ListOrMapOfLabelDeserializer.java
```diff
@@ -26,7 +26,14 @@ public List<Label> deserialize(JsonParser p, DeserializationContext ctxt) throws
         else if (p.hasToken(JsonToken.START_ARRAY)) {
             // deserialize as list
             List<Map<String, String>> ret = ctxt.readValue(p, List.class);
-            return ret.stream().map(map -> new Label(map.get("key"), map.get("value"))).toList();
+            return ret.stream().map(map -> {
```
loicmathieu, line 29: Could it be replaced by `.map(this::validateAndCreateLabel)`?

Also adding a test would be great

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/serializers/ListOrMapOfLabelDeserializer.java
thyw1: I kept the `START_ARRAY` and `START_OBJECT` branches using different methods intentionally, to **minimize the changes**, for the following reasons:

1. **For `List<Map<String, Object>>` (`START_ARRAY` branch):**  
   - Each map has fixed keys `"key"` and `"value"`.  
   - These two fields correspond to a single `Label`'s key and value.  
   - So it's straightforward to perform type checking and conversion directly on the map.  
   - Forcing it to use `Map.Entry` would add unnecessary complexity.  

2. **For `Map<String, Object>` (`START_OBJECT` branch):**  
   - Each entry corresponds directly to a `Label`'s key and value.  
   - The logic here is already correct and doesn’t require modification.  

While it is theoretically possible to unify both branches into a single method, which is a larger change.  
By keeping the existing structure and adding type conversion only in the `START_ARRAY` branch, we solve the problem while minimizing impact on other logic.

loicmathieu: Yes, sorry, I read the code too quickly

loicmathieu:
@thyw1 please add a test then i'll be happy to merge it

core/src/main/java/io/kestra/core/serializers/ListOrMapOfLabelDeserializer.java
thyw1: I have added a unit test that simulates the scenario I encountered.  
It serializes a `Flow` object to a string (where `labels` becomes a list), then deserializes it using `YamlParser`.  
This ensures that numeric label values are correctly converted to strings and the deserialization works as expected.

thyw1:
> @thyw1 please add a test then i'll be happy to merge it

I have added a unit test that simulates the scenario I encountered.  
It serializes a `Flow` object to a string (where `labels` becomes a list), then deserializes it using `YamlParser`.  
This ensures that numeric label values are correctly converted to strings and the deserialization works as expected.

loicmathieu (commented):
LGTM, thanks a lot

core/src/test/resources/flows/valids/labels-deserialization.yaml
```diff
@@ -0,0 +1,17 @@
+id: full
```
loicmathieu:
```suggestion
id: labels-deserialization
```

loicmathieu (changes requested):
Sorry, I didn't saw that the flow name is not correct, by convention it should be the name of the file without .yaml

loicmathieu:
@thyw1 don't bother to updte your PR for my last comment I just did it.
I'll merge when all tests pass

core/src/test/java/io/kestra/core/serializers/YamlParserTest.java
```diff
@@ -235,6 +237,29 @@ void duplicateKey() {
         assertThat(new ArrayList<>(exception.getConstraintViolations()).getFirst().getMessage()).contains("Duplicate field 'variables.tf'");
     }
 
+    @Test
+    void vaildLabelsParser() throws IOException {
+        Flow flow = parse("flows/valids/labels-deserialization.yaml");
+        // like change execution state api,Serialize flow to YAML/JSON string
+        String s = OBJECT_MAPPER.writeValueAsString(flow);
+        assertThat(s).isEqualTo("id: full\n" +
```
loicmathieu:
```suggestion
        assertThat(s).isEqualTo("id: labels-deserialization\n" +
```

loicmathieu (approved):

PR #13369, opened 2025-12-03
fix(core): Await renaming would break plugins

brian-mulier-p:


AcevedoR:
- "reverting" this commit https://github.com/kestra-io/kestra/commit/9fa94deba933f3afebb4317dc3d36caf994ab66f, and adding deprecation and a new method instead

AcevedoR (approved):
LGTM

PR #13381, opened 2025-12-04
fix(core): safeguard for null flow when trying to reset trigger in Jd…

brian-mulier-p:
…bcExecutor


jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
```diff
@@ -1135,11 +1135,18 @@ private void toExecution(Executor executor, boolean ignoreFailure) {
...
                 if (execution.getTrigger() != null && execution.getState().isFailed() && ListUtils.isEmpty(execution.getTaskRunList())) {
                     FlowWithSource flow = executor.getFlow();
-                    triggerRepository
-                        .findByExecution(execution)
-                        .ifPresent(trigger -> {
-                            this.triggerState.update(executionService.resetExecution(flow, execution, trigger));
-                        });
+
+                    if (flow == null) {
+                        log.warn("Couldn't reset trigger for execution {} as flow {} is missing. Trigger {} might stay stuck.",
```
loicmathieu:
```suggestion
                        log.error("Couldn't reset trigger for execution {} as flow {} is missing. Trigger {} might stay stuck.",
```

loicmathieu (approved):
LGTM, best to log in error

PR #13385, opened 2025-12-04
feat(system): remove FlowTopologyHandler and TriggerEventPublisher

loicmathieu:
And make all flow modification pass throught the FlowService so that downstream consumers are always updated


fhussonnois (approved):
Cool, it will really make flow management easier if `FlowService` becomes the single point of entry for all operations. At some point, we should plan a refactoring of the `FlowController` that still contains too much business logic.

loicmathieu:
Most of the tests that fail here passed locally so I'll merge this one and see if after a rebase which fixes some issues in our CI pipeline the test still fails.

PR #13416, opened 2025-12-05
chore(tests): remove running tests in parallel

loicmathieu:


fhussonnois (approved):
Approved, but could you please just comment the parralel block instead of removing it, thanks

loicmathieu:
@fhussonnois I don't like commenting code out as, well, it's still in the git history! But I did it anyway.

PR #13421, opened 2025-12-05
chore(core): remove unnecessary flowable attempt list copying

MTarek165:
Related to #12153
## Description
- Remove unnecessary list copying for flowable task attempts.
In this place, there can’t be any previous attempts yet, so the check is redundant and could cause confusion.
- No behavior changes, purely a cleanup.


loicmathieu (approved):
LGTM

PR #13425, opened 2025-12-05
chore(test): rework and fix gradle check phase

fhussonnois:
- cleanup and refactor gradle :check Task to have more control over it
- separate integration test from unit test in gradle :check
- By default, all tests annotated with KestraTest are considered to be integration-tests (for the moment).
- fix: gradle :check to handle all test failure cases (this should help for https://github.com/kestra-io/kestra-ee/issues/6066)
- fix core/src/main/java/io/kestra/core/plugins/serdes/PluginDeserializer.java checkState() to work regardless of test ordering (if a @MicronautTest ran before a Junit only test if could lead to a fail where KestraContext was already init and required a Bean)

- advance on https://www.notion.so/kestra-io/Flaky-tests-and-KestraTest-2c636907f7b580fb9e39fb0ca62eb473?source=copy_link#2c636907f7b5805b9564ef4d6ba6f80b


AcevedoR:
currently it is not working yet @fhussonnois, we are missing 300 test (webserver?)

loicmathieu:
I think we should not hurry too much on this and first look at how our CI is working now that parallel is disabled during a few days then think of it from a higher perspective.

As, integration tests are tests running with `@KestraTest(startRunner=true)` so we already know them, no need for another annotation I think.

AcevedoR:
I see your point, we can discuss it in our meet.
But the issue about the gradle test build skipping tests needs to be fixed anyway, for example on a actual EE PR we miss webserver tests (~1000 tests missing)

AcevedoR:
- for the tests not being run we should fix it in a separate issue: https://github.com/kestra-io/kestra-ee/issues/6066

build.gradle
```diff
@@ -263,10 +340,35 @@ subprojects {subProj ->
...
+                systemProperty("junit.jupiter.execution.parallel.mode.classes.default", "concurrent")
+            }
+        }
 
-            finalizedBy(tasks.named('flakyTest'))
+        tasks.named('check') {
+            // this task will aggregate test tasks results, and fail is there was any error
+            dependsOn(tasks.named('test'))// default behaviour
+            dependsOn(tasks.named('integrationTest'))
+            dependsOn(tasks.named('flakyTest'))
```
nkwiatkowski: This will make the build fail when we have flaky test failing. Is it what we want?

nkwiatkowski (approved):
LGTM

build.gradle
AcevedoR: fixed

AcevedoR:
TODO:
- [x] FIX do log into stderr when exit 1

PR #13427, opened 2025-12-05
wip: showcase JDBC executor being stuck on any exception in queue con…

brian-mulier-p:
…sumption

never merge it, it's just to showcase an issue in our JDBC executor pattern


brian-mulier-p:
Closing as it will never be merged

PR #13499, opened 2025-12-07
fix(core) : Ensure that a label key cannot contain spaces, special characters or encoded values (e.g. emojis)

### LINKED ISSUE #11229: Ensure that a label key cannot contain spaces, special characters or encoded values (e.g. emojis)
- [ ] Currently, a label key can contain spaces -- we should change that even if it would mean a BC. 
- [ ] The same is true for special characters/encoded values which should be marked as invalid as label keys -- we can reuse the same validation as we use on flow ID https://github.com/kestra-io/kestra/issues/5832 -- Only hyphens (-), underscores (_), lowercase characters, and numbers are allowed. Keys must start with a lowercase character. International characters are allowed. 
```yaml
labels:
  💩: "This stinks, we shouldn't allow that."
```

Array can be allowed as label value long as we treat it as a JSON string

shivanshsharma13:
Closes https://github.com/kestra-io/kestra/issues/11229.

### ✨ Description

What does this PR change?  

This PR adds validation to label keys to enforce the same naming pattern as flow IDs. Label keys must now start with a lowercase character and can only contain lowercase letters, numbers, hyphens (-), and underscores (_).

### ✅ Valid Label Keys
- `environment` - lowercase letters
- `app-name` - lowercase with hyphens
- `user_id` - lowercase with underscores
- `version2` - lowercase with numbers
- `foo.bar` - contains dot (.)
- `a` - single character
### ❌ Invalid Label Keys (Now Rejected)
- `My Label` - contains space
- `Foo` - starts with uppercase letter
- `💩` - emoji/special character
- `0test` - starts with number
- `foo@bar` - contains at sign (@)
- `foo:bar` - contains colon (:)
- `-foo` - starts with hyphen
- `_foo` - starts with underscore

### 🔗 Related Issue

Closes
- #11229

### 📝What I fixed
Screenshot (21)

Screenshot 2025-12-07 231033

### DIFF
--- a/core/src/main/java/io/kestra/core/models/Label.java
+++ b/core/src/main/java/io/kestra/core/models/Label.java
@@ -4,13 +4,16 @@ import io.kestra.core.utils.MapUtils;
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.annotation.Nullable;
 import jakarta.validation.constraints.NotEmpty;
+import jakarta.validation.constraints.Pattern;
 
 import java.util.*;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 @Schema(description = "A key/value pair that can be attached to a Flow or Execution. Labels are often used to organize and categorize objects.")
-public record Label(@NotEmpty String key, @NotEmpty String value) {
+public record Label(
+        @NotEmpty @Pattern(regexp = "^[\\p{Ll}][\\p{L}0-9._-]*$", message = "Invalid label key. A valid key contains only lowercase letters numbers hyphens (-) underscores (_) or periods (.) and must begin with a lowercase letter.") String key,
+        @NotEmpty String value) {
     public static final String SYSTEM_PREFIX = "system.";
 
     // system labels
--- a/core/src/test/java/io/kestra/core/models/LabelTest.java
+++ b/core/src/test/java/io/kestra/core/models/LabelTest.java
@@ -134,4 +134,47 @@ class LabelTest {
         Optional<ConstraintViolationException> emptyKeyLabelResult = modelValidator.isValid(new Label("", "bar"));
         assertThat(emptyKeyLabelResult.isPresent()).isTrue();
     }
+
+    @Test
+    void shouldValidateValidLabelKeys() {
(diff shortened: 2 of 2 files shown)


loicmathieu (changes requested):
A dot `.` should be allowed, we use it internally.

shivanshsharma13:
> A dot `.` should be allowed, we use it internally.

Sorry, just overlooked my description, the regex `^[\\p{Ll}][\\p{L}0-9._-]*$` handles the period (.) and accepts it. Edited the same!

### Regex Breakdown

```
^[\p{Ll}][\p{L}0-9._-]*$
```

| Component | Meaning |
|----------|---------|
| `^` | Start of the string |
| `\p{Ll}` | The **first character must be a lowercase letter** (`a–z` or any Unicode lowercase letter) |
| `[\p{L}0-9._-]*` | The rest of the key may include:<br>• letters (uppercase or lowercase)<br>• digits (`0–9`)<br>• dot (`.`)<br>• underscore (`_`)<br>• hyphen (`-`) |
| `$` | End of the string |

---

#### Summary

A valid label key **must start with a lowercase letter** and may contain **letters, digits, dots, underscores, or hyphens** Amending that to error message. 

Also do we need to allow `10message` - Currently it doesn't

loicmathieu:
@shivanshsharma13 I think you miss the escape of the dot in the regex.
And your test do assert that dot is not validated.

shivanshsharma13:
> @shivanshsharma13 I think you miss the escape of the dot in the regex. And your test do assert that dot is not validated.

Ohh, thanks for pointing it out - just to clarify, in this case the `.` is inside a character class, so it's treated as a literal dot rather than a wildcard. The `[0-9._-]` part is intentional.

And the confusion is obvious because I added the test cases also but fixed the overlooked test cases to align with the expected behavior.

Attaching a short video walkthrough for clarity on how the dot is interpreted.


https://github.com/user-attachments/assets/b5ca3443-4368-46e6-90ff-6bc81bfab7ed

loicmathieu (approved):
LGTM, thanks

PR #13525, opened 2025-12-08
fix(execution): skip the render cache in flowable for properties used…

loicmathieu:
… to compute next tasks

As when the flowable is itself in a flowable that process tasks concurrently like the ForEach when using a concurrency limit, it can be done multiple time with different values.

This can only occurs if the expression is using `taskRun.value`.

Fixes https://github.com/kestra-io/kestra-ee/issues/6055


nkwiatkowski (approved):
LGTM

PR #13528, opened 2025-12-08
fix(queue): change index to remove a sort

nkwiatkowski:
- fix #https://github.com/kestra-io/kestra-ee/issues/6067


nkwiatkowski:
It remove the sort, but slow the global query

PR #13530, opened 2025-12-08
fix/scheduler

fhussonnois:
* [X] add metrics to trigger scheduling loop
* [X] make sure to participate in rebalancing while stopping
* [X] skip missed schedules when re-enabling a trigger
* [X] handle `DeleteBackfillTrigger` event
* [X] fix EventId serialization

### DIFF
--- a/scheduler/src/main/java/io/kestra/scheduler/DefaultScheduler.java
+++ b/scheduler/src/main/java/io/kestra/scheduler/DefaultScheduler.java
@@ -62,6 +62,8 @@ public class DefaultScheduler extends AbstractService implements Scheduler {
     private final List<Disposable> consumerDisposables = new ArrayList<>();
 
     private final Set<Integer> currentVNodesAssignment = new HashSet<>();
+    
+    private Disposable rebalanceDisposable;
 
     @Inject
     public DefaultScheduler(final TriggerSchedulingLoopFactory schedulerEventLoopFactory,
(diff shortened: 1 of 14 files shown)


scheduler/src/main/java/io/kestra/scheduler/DefaultScheduler.java
```diff
@@ -57,11 +59,14 @@ public class DefaultScheduler extends AbstractService implements Scheduler {
...
     // Services
     private final TriggerWorkerJobResultSubscriber triggerWorkerJobResultSubscriber;
+    private final MetricRegistry metricRegistry;
 
     // Consumers
     private final List<Disposable> consumerDisposables = new ArrayList<>();
 
     private final Set<Integer> currentVNodesAssignment = new HashSet<>();
+    
+    private Disposable rebalanceDisposable;
```
loicmathieu, line 69: Why don't you add it to the list of consumerDisposables?

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
```diff
@@ -225,7 +244,11 @@ private void waitIfPaused() throws InterruptedException {
     /**
      * Gets the current assignment for this event loop.
      *
+<<<<<<< HEAD
      * @return  the assignments.
+=======
+     * @return the assignments.
+>>>>>>> 5d130c954 (fix(scheduler): add metrics to trigger scheduling loop)
```
loicmathieu: Argh!

loicmathieu (approved):
LGTM

scheduler/src/main/java/io/kestra/scheduler/DefaultScheduler.java
fhussonnois: Because consumers are disposed when vnodes are revoked, and this one must only be disposed when stopping the scheduler.

PR #13538, opened 2025-12-09
fix: allow zero-byte file uploads in execution inputs

### LINKED ISSUE #8218: Zero-byte `FILE` input treatment
### Describe the issue

A change of behavior happened between v0.16 and v0.17+ regarding both optional and required `FILE` inputs.

When such input gets passed a zero-byte (empty) file the API rejects such parameter value.

---

Versions prior v0.17 accept the empty file input and present it as a regular 0B file input. Is this behavior intended?

# Example

Flow:

```yaml
id: mule_167053
namespace: company.team

inputs:
  - id: empty_file
    type: FILE
    description: Do we accept zero-byte files or not?
    required: false

tasks:
  - id: hello
    message: Hello World! 🚀
```

Reproduction commands:

```console
touch /tmp/empty && \
curl -v -X POST -H 'Content-Type: multipart/form-data' -F 'files=@/tmp/empty;filename=empty_file' 'http://localhost:8080/api/v1/executions/company.team/mule_167053'
```

Error response body:

> {"message":"Invalid entity: empty_file: Invalid input for `empty_file`, missing required input, but received `null`","logref":null,"path":null,"_links":{"self":{"href":"/api/v1/executions/company.team/mule_167053","templated":false,"profile":null,"deprecation":null,"title":null,"hreflang":null,"type":null,"name":null}},"_embedded":{"errors":[{"message":"Invalid input for `empty_file`, missing required input, but received `null`","logref":null,"path":"empty_file","_links":{},"_embedded":{}}]}}

### Environment

- Kestra Version: develop

bikashagarwala0:
## Description
This PR fixes a regression introduced in v0.17 where zero-byte file uploads were treated as missing inputs (`null`). This caused workflows to fail with validation errors ("missing required input") when a `FILE` input was marked as `required: true` but the uploaded file was empty.

## Changes
- Removed the check in `FlowInputOutput.java` that explicitly rejected input streams with 0 transferred bytes.
- Zero-byte files are now correctly processed, stored, and passed as valid execution inputs (e.g., `kestra:///.../empty_file`).

## Related Issue
Closes #8218

## Checklist
- [x] Logic fixed to allow 0-byte streams.
- [x] Verified via local unit test (`FlowInputOutputTest`).
- [x] Verified via manual reproduction (`curl` upload of an empty file).

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
@@ -158,11 +158,7 @@ public class FlowInputOutput {
                             File tempFile = File.createTempFile(prefix, fileExtension);
                             try (var inputStream = fileUpload.getInputStream();
                                  var outputStream = new FileOutputStream(tempFile)) {
-                                long transferredBytes = inputStream.transferTo(outputStream);
-                                if (transferredBytes == 0) {
-                                    sink.error(new KestraRuntimeException("Can't upload file: " + fileUpload.getFilename()));
-                                    return;
-                                }
+                                inputStream.transferTo(outputStream);
                                 URI from = storageInterface.from(execution, inputId, fileName, tempFile);
                                 sink.next(Map.entry(inputId, from.toString()));
                             } finally {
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
LGTM, thanks

PR #13542, opened 2025-12-09
fix(scheduler): re-implement maintenance mode

fhussonnois:
* Refactor MaintenanceService to be an interface
* Fix VNodeController to also list schedulers in MAINTENANCE state
* Add maintenance mode support for scheduler


scheduler/src/main/java/io/kestra/scheduler/DefaultScheduler.java
```diff
@@ -58,8 +59,12 @@ public class DefaultScheduler extends AbstractService implements Scheduler {
     // Services
     private final TriggerWorkerJobResultSubscriber triggerWorkerJobResultSubscriber;
 
+    private final MaintenanceService maintenanceService;
+
     // Consumers
     private final List<Disposable> consumerDisposables = new ArrayList<>();
+    
+    private Disposable maintenanceListener;
```
loicmathieu, line 70: Why didn't you add it to the list of consumerDisposables?

loicmathieu (approved):
LGTM

PR #13572, opened 2025-12-10
feat(execution): Execution Command

loicmathieu:
Use a command pattern for all changes to the execution for the Executor to process them.

### DIFF
--- a/executor/src/main/java/io/kestra/executor/handler/ExecutionCommandMessageHandler.java
+++ b/executor/src/main/java/io/kestra/executor/handler/ExecutionCommandMessageHandler.java
@@ -1,0 +1,73 @@
+package io.kestra.executor.handler;
+
+import io.kestra.core.exceptions.FlowNotFoundException;
+import io.kestra.core.executor.command.*;
+import io.kestra.core.models.flows.State;
+import io.kestra.core.runners.FlowMetaStoreInterface;
+import io.kestra.core.services.ExecutionService;
+import io.kestra.executor.ExecutionStateStore;
+import io.kestra.executor.ExecutorContext;
+import io.kestra.executor.ExecutorMessageHandler;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.Optional;
+
+@Singleton
+@Slf4j
+public class ExecutionCommandMessageHandler implements ExecutorMessageHandler<ExecutionCommand> {
+    private final ExecutionService executionService;
+    private final ExecutionStateStore executionStateStore;
+    private final FlowMetaStoreInterface flowMetaStore;
+
+    @Inject
+    public ExecutionCommandMessageHandler(ExecutionService executionService,
+                                          ExecutionStateStore executionStateStore,
+                                          FlowMetaStoreInterface flowMetaStore) {
+        this.executionService = executionService;
+        this.executionStateStore = executionStateStore;
+        this.flowMetaStore = flowMetaStore;
+    }
+
+    @Override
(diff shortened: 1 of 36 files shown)


webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -993,7 +997,7 @@ public HttpResponse<FileMetas> getFileMetadatasFromExecution(
     @ExecuteOn(TaskExecutors.IO)
     @Post(uri = "/{executionId}/restart")
     @Operation(tags = {"Executions"}, summary = "Restart a new execution from an old one")
-    public Execution restartExecution(
+    public EventId restartExecution(
```
fhussonnois: The `EventId` will be serialized as a string which is not ideal in the REST response. It cannot evolve without introducing breaking changes and it doesn't follow standard API design practices. It's better to define a dedicated DTO for the response, e.g. `record ApiAsyncEvent(String id)` that we can re-use everywhere

jdbc-h2/src/main/resources/migrations/h2/V3_6__execution_command.sql
```diff
@@ -0,0 +1,23 @@
+ALTER TABLE queues ALTER COLUMN "type" ENUM(
```
fhussonnois, line 1: I think it would be preferable to just update the existing file `V3_3__executor_v2.sql` to avoid having too much migrations

executor/src/main/java/io/kestra/executor/handler/ExecutionCommandMessageHandler.java
```diff
@@ -0,0 +1,71 @@
...
+import io.kestra.executor.ExecutorContext;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.Optional;
+
+@Singleton
+@Slf4j
+public class ExecutionCommandMessageHandler {
```
fhussonnois: This class could be just named `ExecutionCommandHandler` ?

fhussonnois (changes requested):

jdbc-h2/src/main/resources/migrations/h2/V3_6__execution_command.sql
loicmathieu: My plan is to merge all migrations into a single `kestra_v2` before we merge this branch.
For now we can create individual migrations, it's easier as it keep each PR atomic.

executor/src/main/java/io/kestra/executor/handler/ExecutionCommandMessageHandler.java
loicmathieu: Naming is hard.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
loicmathieu: Good call, this pattern may be reused for the TriggerController also so better to design it carefully

loicmathieu: Done

executor/src/main/java/io/kestra/executor/handler/ExecutionCommandMessageHandler.java
loicmathieu: I'll keep it for concistency with the other MessageHandler, I noticed that I forgot to implements the interface, I added it, it would makes more sense now ;)

fhussonnois (approved):

PR #13573, opened 2025-12-10
feat(executions): add a protection mecanism to avoid any potential co…

### LINKED ISSUE #12031: Concurrency doesn't work
### Describe the issue

Concurrency doesn't work if you run the flow manually
```
id: sparrow_298436
namespace: company.team

concurrency:
  behavior: QUEUE
  limit: 1

tasks:
  - id: hello
    type: "io.kestra.plugin.core.flow.Sleep"
    duration: "PT120S"
```

Image

### Environment

- Kestra Version: 0.24.4/1.0.6

### LINKED ISSUE #13301: Add a safeguard for concurrency limit to avoid any potential excess
### Describe the issue

Currently, when a flow ends with a concurrency limit, we pop a new execution whether or not there is a free concurrency slot.
This should be valid, but in case a bug exist (see https://github.com/kestra-io/kestra/pull/13279 for ex) or any other unexpected behavior this may be wrong.

What we can do is when we decrement, return the new limit and avoid poping a new execution.
This of course has the risk of never poping an execution if for ex there was no more running execution so it's a **tradeoff that we need to evaluate before implementing**.

### Environment

- Kestra Version: develop

loicmathieu:
…ncurrency overflow

Concurrency limit is based on a counter that increment and decrement the limit each time a flow is started and terminated.

This count should always be accurate.

But if some unexpected event occurs (bug or user manually do something wrong), the count may not be accurate anymore.

To avoid any potential issue, when we decrement the counter, we chech that concurrency count is bellow the limit before unqueing an execution.

Fixes #12031
Closes  #13301

### DIFF
--- a/jdbc-h2/src/test/java/io/kestra/runner/h2/H2RunnerConcurrencyTest.java
+++ b/jdbc-h2/src/test/java/io/kestra/runner/h2/H2RunnerConcurrencyTest.java
@@ -1,6 +1,6 @@
 package io.kestra.runner.h2;
 
-import io.kestra.core.runners.AbstractRunnerConcurrencyTest;
+import io.kestra.jdbc.runner.JdbcConcurrencyRunnerTest;
 
-public class H2RunnerConcurrencyTest extends AbstractRunnerConcurrencyTest {
+public class H2RunnerConcurrencyTest extends JdbcConcurrencyRunnerTest {
 }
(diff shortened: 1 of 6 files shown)


brian-mulier-p (approved):
LGTM

loicmathieu:
Backported to 1.0 and 1.1

Issue #12426, opened 2025-10-28
Input defaults bug while using subflows


caxefaizan:
### Describe the issue

Passing Inputs to subflow only renders the `defaults` for `destTable` input correctly only in the first task.

Sample Flow
```
id: ticket_938
namespace: company.team

tasks:
  - id: MEETING_SESSION_PROPOSAL
    type: io.kestra.plugin.core.flow.Subflow
    namespace: company.team
    flowId: oracle_copy_table
    inputs:
      srcSchema: "CALENDAR_PLANNING_MANAGER"
      srcTable: "MEETING_SESSION_PROPOSAL"     
 
  - id: MEETING_PROPOSAL
    type: io.kestra.plugin.core.flow.Subflow
    namespace: company.team
    flowId: oracle_copy_table
    inputs:
      srcSchema: "CALENDAR_PLANNING_MANAGER"
      srcTable: "MEETING_PROPOSAL"
```

Subflow
```
id: oracle_copy_table
namespace: company.team

inputs:
  - id: srcSchema
    type: STRING
  - id: srcTable
    type: STRING
  - id: destTable
    type: STRING
    defaults: "B__{{ inputs.srcSchema | upper | split('_MANAGER') | first }}__{{ inputs.srcTable }}"
    dependsOn:
      inputs:
        - srcSchema
        - srcTable
tasks:
  - id: hello
    type: io.kestra.plugin.core.log.Log
    message: Hello {{inputs.destTable}}! 🚀
```
Expected  inputs for first subflow execution
`B__CALENDAR_PLANNING__MEETING_SESSION_PROPOSAL`
Expected  inputs for second subflow execution
`B__CALENDAR_PLANNING__MEETING_PROPOSAL`

### Environment

- Kestra Version: 1.0.7

yummyash:
i want to contribute for this

MilosPaunovic:
Hi @yummyash,  

You're already assigned to another issue in our repository, and we usually assign only one at a time to new contributors. Let's make sure to wrap that one up before jumping onto another.  🚀

caxefaizan:
Hi Team, 
Could we check this out please. It ahs been pending for a while now.

Thank you

nkwiatkowski:
fixed and backported
It was a cache issue, we resolved both expression with the value of the first inputs

PR #13576, opened 2025-12-10
fix(flows): deserialyze expression without cache

nkwiatkowski:
- fix #https://github.com/kestra-io/kestra/issues/12426


core/src/main/java/io/kestra/core/models/property/Property.java
```diff
@@ -317,7 +317,7 @@ public Property<?> deserialize(JsonParser p, DeserializationContext ctxt) throws
             } else {
                 s = p.getValueAsString();
             }
-            return new Property<>(s);
+            return new Property<>(s, true);
```
loicmathieu: This would basically disable the cache for all expressions as the deserialization is used everywhere.
Two things to consider it:
- removing the cache as now it would never  be used
- using skipCache() for reading that particular expression

Overall, I don't really understand how this can cause the issue as two different tasks should have two different Property instance so the cache should not hit us here.

But as I said, if the cache is such a problem, remove it

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/models/property/Property.java
fhussonnois: It seems that the issue is coming from the executor passing the same `Flow` instance to the `Subflow.createSubflowExecutions` method for both tasks. 

The expressions for the input `defaults` are evaluated and cached by the first task 

kestra-input-default-task1

second task:

kestra-input-default-task2

loicmathieu: Yes, we have a cache of flows

fhussonnois: The issue is actually from the use of the `Property` class in `Input`

fhussonnois: The solution is maybe to just `skipCache` here: 
https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java#L380-L386

fhussonnois (approved):

core/src/main/java/io/kestra/core/models/property/Property.java
nkwiatkowski: You are right, adding the skipCache() here solve the issue

loicmathieu: This is what pointed out since the beginig ;) 
Maybe I was not clear enought to where skipCache should be called.

PR #13579, opened 2025-12-10
fix(core): improve exception handling at FlowInputOutput

### LINKED ISSUE #13577: Inconsistent exception messages with some inputs
### Describe the issue

Sometimes error messages becomes inconsistent causing confusion to user with cases like 
- we are using SECRET input without encryption configuation not knowing what is the exact error is it the input or the missing configuration?
- at least in other cases causes duplication occur if used with input URI for example.
Image

Image

### Environment

- Kestra Version: develop

MTarek165:
closes #13577
## Description
- some parts used with generic exception with all situations in Data interface causing either confusion about the correct error  or even duplication
## Solution
- Added InputOutputValidationException to represent Inputs/Outputs
  validation issues and added handler to it in ErrorsController
- Added support for throwing multiple constraint violations for the same
  input
- Added support for throwing multiple constraints at MultiselectInput
- Refactored exception handling at FlowInputOutput
- Added merge() function to combine constraint violation messages and
  added test for it at InputsTest
- Fixed the failed tests

### DIFF
--- a/core/src/main/java/io/kestra/core/exceptions/InputOutputValidationException.java
+++ b/core/src/main/java/io/kestra/core/exceptions/InputOutputValidationException.java
@@ -1,0 +1,37 @@
+package io.kestra.core.exceptions;
+
+import io.kestra.core.models.flows.Data;
+import io.kestra.core.models.flows.Input;
+import io.kestra.core.models.flows.Output;
+
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Exception that can be thrown when Inputs/Outputs have validation problems.
+ */
+public class InputOutputValidationException extends KestraRuntimeException {
+    public InputOutputValidationException(String message) {
+        super(message);
+    }
+    public static InputOutputValidationException of( String message, Input<?> input){
+        String inputMessage = "Invalid value for input" + " `" + input.getId() + "`. Cause: " + message;
+        return new InputOutputValidationException(inputMessage);
+    }
+    public static InputOutputValidationException of( String message, Output output){
+        String outputMessage = "Invalid value for output" + " `" + output.getId() + "`. Cause: " + message;
+        return new InputOutputValidationException(outputMessage);
+    }
+    public static InputOutputValidationException of(String message){
+        return new InputOutputValidationException(message);
+    }
+
+    public static InputOutputValidationException merge(Set<InputOutputValidationException> exceptions){
+        String combinedMessage = exceptions.stream()
+                .map(InputOutputValidationException::getMessage)
+                .collect(Collectors.joining(System.lineSeparator()));
+        throw new InputOutputValidationException(combinedMessage);
+    }
+
+}
--- a/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
@@ -3,6 +3,8 @@ package io.kestra.core.runners;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.kestra.core.encryption.EncryptionService;
 import io.kestra.core.exceptions.IllegalVariableEvaluationException;
+
+import io.kestra.core.exceptions.InputOutputValidationException;
 import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.flows.Data;
 import io.kestra.core.models.flows.DependsOn;
@@ -17,7 +19,6 @@ import io.kestra.core.models.property.Property;
 import io.kestra.core.models.property.PropertyContext;
 import io.kestra.core.models.property.URIFetcher;
 import io.kestra.core.models.tasks.common.EncryptedString;
-import io.kestra.core.models.validations.ManualConstraintViolation;
 import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.storages.StorageContext;
 import io.kestra.core.storages.StorageInterface;
@@ -207,8 +208,8 @@ public class FlowInputOutput {
             .filter(InputAndValue::enabled)
             .map(it -> {
                 //TODO check to return all exception at-once.
-                if (it.exception() != null) {
-                    throw it.exception();
+                if (it.exceptions() != null && !it.exceptions().isEmpty()) {
+                    throw  InputOutputValidationException.merge(it.exceptions());
                 }
                 return new AbstractMap.SimpleEntry<>(it.input().getId(), it.value());
             })
@@ -292,13 +293,9 @@ public class FlowInputOutput {
                 try {
                     isInputEnabled = Boolean.TRUE.equals(runContext.renderTyped(dependsOnCondition.get()));
                 } catch (IllegalVariableEvaluationException e) {
-                    resolvable.resolveWithError(ManualConstraintViolation.toConstraintViolationException(
-                        "Invalid condition: " + e.getMessage(),
-                        input,
-                        (Class<Input>)input.getClass(),
-                        input.getId(),
-                        this
-                    ));
+                    resolvable.resolveWithError(
+                        InputOutputValidationException.of("Invalid condition: " + e.getMessage())
+                    );
                     isInputEnabled = false;
                 }
             }
@@ -331,7 +328,7 @@ public class FlowInputOutput {
             // validate and parse input value
             if (value == null) {
                 if (input.getRequired()) {
-                    resolvable.resolveWithError(input.toConstraintViolationException("missing required input", null));
+                    resolvable.resolveWithError(InputOutputValidationException.of("Missing required input:"  + input.getId()));
                 } else {
                     resolvable.resolveWithValue(null);
                 }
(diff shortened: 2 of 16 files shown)


loicmathieu:
Those messages are used in both input validation and flow output validation so I'm not sure the proposed enhancement is what's best.
As @fhussonnois worked on this are, I'll let him advise what's the best sentences are here.

MTarek165:
Thanks for your response, yes I am aware of that but the current pattern causes many inconsistent messages more than this, although the proposed pattern may not be the best but I am still investigating.
May be distinguishing between the constant pattern for Input/output and others where both uses ConstraintViolationExecption with flag may be the best option.

MTarek165:
I think the best way is to pass the exception to toConstraintViolationException() method and check on the exception type:
-  keep the old pattern for inputs/outputs either if they are invalid or violating constraints ( IllegalArgumentExecption and ConstrainViolationException )
-  otherwise it would be of type Exception and in this case send the embedded message only

 this would solve the problem with keeping the used pattern for checking Illegal or violating Input/Output only and would be more consistent.I have updated the solution to use this idea.

fhussonnois (changes requested):
Hey @MTarek165 thank you very for working on this inconsistency issue. 

After reviewing a bit of our current code, I think the current use of `ConstraintViolationException`  in the input validation chain adds too much complexity. In fact, we only care about having an clean list of error messages for each input to be display in the UI - so I would not go for a strategy to test exception type.

What I would do:
- Create a dedicated `InputValidationException` that extends `KestraRuntimeException`
- Rename the `InputAndValue#exception` to `InputAndValue#exceptions` and change the field to be a `List` of `InputValidationException`
- The `toConstraintViolationException` method should not be used anymore when validating inputs
- When catching a ConstraintValidationException just convert the list of `getConstraintViolations`  to a list of `InputValidationException`
- The `InputValidationException` class could have a static factory method to provide the `Invalid value for input '<input name>. Cause:` prefix

The redesign will be more extensive, but will provide a clearer solution in the long term. Let me know, if you are OK to handle this ? 

Thank you

MTarek165:
@fhussonnois Of course, I would be ok to work on it, once I complete these change I would ping you, thanks too.

MTarek165:
@fhussonnois I have two small points I want to point to only:
1- Sometimes parseData() return IllegalArgumentException() if we sent invalid URI for ex, so I would consider that InputValidationException is either a ConstrainViolationException or IllegalArgumentException when thrown with inputs( or outputs).
2- Note also that we should consider the same for outputs as there is one place uses toConstraintViolationException() method when parsing outputs and since parseData() can throw IllegalArgumentException so we should consider the same approach

That is why I see that we may name it InputOutputValidationException instead of InputValidationException

MTarek165:
1- Summary of current changes

- Introduced InputOutputValidationException, used for both inputs and outputs

- All exceptions related to FlowInputOutput now go through this exception

- When a ConstraintViolationException or IllegalArgumentException is related to inputs/outputs, an InputOutputValidationException is created by calling static factory method used to produce a standardized message

- For other errors (e.g. missing encryption configuration, invalid condition), uses the same exception type but with the original exception message is preserved

- Performed some refactoring around exception handling and message construction in FlowInputOutput

2- Important Questions to wonder about

-  Exception collection in InputAndValue
    - validate() always throws only the first constraint violation per input.
    - The exception collection is never fully used as the first violated constraint is thrown only even if there multiple violations, adding unnecessary complexity.
    - Question: Can we simplify by removing the collection and using a single exception, or is there a reason to keep it?

- Handling non-validation input errors
     - Some errors in resolveInputValue() (e.g., missing encryption configuration, invalid condition) are non-validation errors, but are currently forced into InputOutputValidationException to fit the input error collection.
     - Question: Should we use a more flexible exception type like Exception at InputAndValue and keep the new exception used for validations only and others with their exception types, or should we ignore them in the UI input error list and keep them outside validation list?

core/src/main/java/io/kestra/core/exceptions/InputOutputValidationException.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Exception that can be thrown when Inputs/Outputs have validation problems.
+ */
+public class InputOutputValidationException extends KestraRuntimeException {
+    public InputOutputValidationException(String message) {
+        super(message);
+    }
+
+    public static InputOutputValidationException of( String message, Data data){
```
fhussonnois: Could you create two methods here to avoid the if/insteaceof :
* `of(String message, Input input)`
* `of(String message, Output input)`

core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
```diff
@@ -209,8 +210,11 @@ private Map<String, Object> readExecutionInputs(
             .filter(InputAndValue::enabled)
             .map(it -> {
                 //TODO check to return all exception at-once.
-                if (it.exception() != null) {
-                    throw it.exception();
+                if (it.exceptions() != null && !it.exceptions().isEmpty()) {
+                    String combinedMessage = it.exceptions().stream()
```
fhussonnois: Could you please, create a static helper method for that in the `InputOutputValidationException` (e.g., `merge(List<InputOutputValidationException> exceptions)`` and create a simple unit-test please

core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
```diff
@@ -333,7 +333,7 @@ private InputAndValue resolveInputValue(
             // validate and parse input value
             if (value == null) {
                 if (input.getRequired()) {
-                    resolvable.resolveWithError(input.toConstraintViolationException("missing required input", null));
+                    resolvable.resolveWithError(InputOutputValidationException.of("missing required input"));
```
fhussonnois: I think the message should be `"Missing required input:"  + input.getName()`

core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
```diff
@@ -505,7 +510,7 @@ private Object parseType(Execution execution, Type type, String id, Type element
                     if (matcher.matches()) {
                         yield current.toString();
                     } else {
-                        throw new IllegalArgumentException("Expected `URI` but received `" + current + "`");
+                        throw new IllegalArgumentException("The value isn't a valid uri");
```
fhussonnois: Change the message to `Invalid URI format.`

fhussonnois (changes requested):
Hey thanks again for your PR. For the moment let's keep the list of exceptions. In addition, it's OK to have all IllegalArgumentException to be `InputOutputValidationException`

Also, you need to edit the `ErrorController` class to add an handler for the new exception: 

```java
    @Error(global = true)
    public HttpResponse<JsonError> error(HttpRequest<?> request, InputOutputValidationException e) {
       return jsonError(error, HttpStatus.UNPROCESSABLE_ENTITY, "Invalid entity");
    }
```

Could check the following tests which are failing: 
* `io.kestra.core.runners.InputsTest`
* `io.kestra.webserver.controllers.api.ExecutionControllerRunnerTest`.

core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
MTarek165: Ok, I would use getId() as getName() is deprecated

core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
MTarek165: I have added this method, but as I told before there is no way to produce a multiple constraint violations on the same input and I couldn't see such input doing that, but I can create an input used for testing and can throw multiple constraint violations at validate(), hence we could test our method in a real workflow, what do you think about that?

MTarek165:
@fhussonnois I have updated with all notes and fixed tests, there is a small note about merge() function testing above, please take a look

core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
MTarek165: The MultiSelectInput is a good use case , I can collect the two existing violations and send both in collection and test on it. It must then shows the two messages as the function states!

MTarek165: I have updated MultiSelectInput to collect multiple constraint violations 
-  multi non existing input (now if there is multiple input not existing, it would report all)
- using both values and options together
and return message contains all constraints violations
and created test coverage for merge() function in real workflow now.

MTarek165:
@fhussonnois It is ok now , is there anything remaining?

fhussonnois (commented):
Hey @MTarek165 thank you for all your changes. Could you please rebase, resolve conflicts with the  `develop` branch and squash your commits so I can approved and merge your PR ? Thanks

MTarek165:
Hey @fhussonnois , most likely I did something wrong during rebase including so many file changes from develop, I afraid corrupting something , If this is important I can make a clean PR with one commit?

MTarek165:
@fhussonnois finally I compacted it into one clean and descriptive commit , It is ok now

fhussonnois (approved):

fhussonnois:
Thanks again for your contribution.  I have rebased your PR, which will be merged once the CI is complete.

PR #13582, opened 2025-12-10
fix(executions): don't remove worker task result for killed execution

loicmathieu:
As killing an executuion is asynchronous, it is inherently racy. So when we kill an execution it move to the rerminal state which then will remove any worker task result to purge eagerly the queue table.

But if such worker task result arrives late and was not already processed by the executor, it will be purged before ever been able to be processed so the task would nevert be updated as KILLED.

Note: this may un-flaky some falky unit tests

Fixes https://github.com/kestra-io/kestra-ee/issues/6040


Skraye (approved):
LGTM

loicmathieu:
Tests fail in unrelated areas, merging it.

loicmathieu:
Backported to 1.0 and 1.1

PR #13586, opened 2025-12-10
fix(repositories): unwanted integer to string conversion

nkwiatkowski:


jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
```diff
@@ -346,6 +346,24 @@ protected Condition getConditionOnField(
...
+            || o instanceof Integer
+            || o instanceof Long
+            || o instanceof Float
+            || o instanceof Double
+            || o instanceof Character
+            || o instanceof String) {
+            return o;
+        }
+
+        return Objects.toString(o);
```
loicmathieu: No need to something more complex as you already check for nullability.

```suggestion
        return o.toString();
```

loicmathieu (changes requested):

PR #13587, opened 2025-12-10
fix(scheduler): fix timezone used for trigger evalution context

fhussonnois:
Changes:
* Adds common timezone property to the Schedulable interface


Skraye (approved):
LGTM

PR #13605, opened 2025-12-11
fix(tests): remove test UncaughtExceptionHandler

brian-mulier-p:


core/src/main/java/io/kestra/core/utils/ThreadUncaughtExceptionHandler.java
```diff
@@ -15,15 +15,16 @@ public void uncaughtException(Thread t, Throwable e) {
...
             // If logging fails, e.g. due to missing memory, at least try to log the
             // message and the cause for the failed logging.
             System.err.println(e.getMessage());
             System.err.println(errorInLogging.getMessage());
         } finally {
+            KestraContext.getContext().shutdown();
+
             if (!isTest) {
-                KestraContext.getContext().shutdown();
                 Runtime.getRuntime().exit(1);
```
loicmathieu, line 28: Can you explain why it's still under isTest?

core/src/test/java/io/kestra/core/tasks/test/SleepTrigger.java
```diff
@@ -44,6 +44,6 @@ public Optional<Execution> evaluate(ConditionContext conditionContext, TriggerCo
 
     @Override
     public Duration getInterval() {
-        return null;
+        return Duration.ofSeconds(1);
```
loicmathieu: Can you explain why you are doing that?

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
```diff
@@ -504,9 +512,21 @@ protected Runnable poll(Supplier<Integer> runnable) {
                     throw new RuntimeException(e);
                 }
             }
+
+            running.set(false);
         });
 
-        return () -> running.set(false);
```
loicmathieu, line 487: We always use this pattern on the disposable runnable: we set a flag to false so the main loop ends.
What are the rational to change that?

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
```diff
@@ -336,18 +336,25 @@ public Runnable receive(String consumerGroup, Consumer<Either<T, Deserialization
...
+                    if (!result.isEmpty()) {
+                        maxOffset.set(result.getLast().get("offset", Integer.class));
+                    }
 
-                return result;
-            });
+                    return result;
+                });
+            } catch (Exception e) {
+                log.error("Error while receiving messages from JDBC queue. Thread stacktrace: {}", parentStackTrace, e);
```
loicmathieu: This seems like debug info that you commit.
Is it really needed? It capture a stacktrace and retain it forever which I'm not comptfortable with

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/utils/ThreadUncaughtExceptionHandler.java
brian-mulier-p: Because otherwise the runtime exit will exit entirely the tests. IMO the best of both world is to shutdown Kestra upon failure but keep the JVM running to avoid stopping the entire test suite while still seeing the test failure.

core/src/test/java/io/kestra/core/tasks/test/SleepTrigger.java
brian-mulier-p: It was long time ago but this was silently dropped, the return null was leading to scheduler failure but silently caught by the uncaught exception handler. When I removed it entirely I identified that issue and fixed it (there was a NPE).

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
brian-mulier-p: Because imo we can't afford having a queue in the middle of a transaction and being hard shutdown by Micronaut because our cancel runnable isn't waiting for a true finish.

Here we switch from "set a running = false flag" to "set a queriedToStop = true and wait for acknowledgement and full stop by waiting for running = false"

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
brian-mulier-p: Idk I thought about removing it initially and then I thought it was a good idea for troubleshooting if this ever occurs. Most important point is the parentThreadStackTrace because polling is done in a child thread and because of that the initial stacktrace won't from where the queue listen initiated.

core/src/main/java/io/kestra/core/utils/ThreadUncaughtExceptionHandler.java
loicmathieu: I think Runtime.exit() is not needed at all as if you shutdown the RunContext any server commands will stop as they are doing: `Await.until(() -> !this.applicationContext.isRunning());`.

This only serve to return 1 instead of 0 so the exec ends in error.
And more funny, there is a shutdown hook so that if the JVM exist it shutdown the application context.
So in fact, you should only be able to Runtime.exit(1) and the runContext should be properly closing...

Overall, it seems a bit fishy.

core/src/test/java/io/kestra/core/tasks/test/SleepTrigger.java
loicmathieu: OK, can you lower the duration as for test to not be too long? Maybe 100ms or even 1ms if it just to avoid a NPE

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
loicmathieu: Hum interesting, I would want to have the opinion of @fhussonnois on this one. I too think it may explain our closing issues

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
loicmathieu: I'm a bit skeptical, the stacktrace always contains the link to the runnable (the lamba usually) so I don't think it's necessary and it seems low level so no users would understand it well.

loicmathieu (approved):
LGTM

PR #13629, opened 2025-12-11
fix(scheduler): fix TriggerSchedulerMonitor

fhussonnois:
Adds new `trigger_id` to the executions table with index 
Adds new method findAllByTriggerUid on ExecutionRepository


loicmathieu (approved):
LGTM

PR #13630, opened 2025-12-11
fix(scheduler): add failed execution with logs when scheduled triggers fail during evaluation

### LINKED ISSUE #13307: Error in trigger doen't lead to creation a flow with Failed state
### Feature description

If the trigger contains an error, the flow will not be triggered, and you don't even know that something went wrong.
Perhaps we can initiate execution with a Failed state to trigger some alerts?

Example with a wrong trigger

```
id: sparrow_298436
namespace: company.team

concurrency:
  behavior: QUEUE
  limit: 1

tasks:
  - id: hello
    type: "io.kestra.plugin.core.flow.Sleep"
    duration: "PT120S"

triggers:
  - id: minute
    disabled: false
    cron: "*/1 * * * *"
    inputs:
      mergeMethod: "{{ now().hour == 0 ? 3 : 2 }}"
```

Now we can only see the problem in the Logs tab

Image

But there will be no executions

Image

MTarek165:
closes #13307
## Description
- Now with scheduled trigger when there is issue at trigger evaluation( unsupported expressions, flow inputs resolving issues ...)
it fails silently without creating failed execution, you will now only from logs that there is a problem
## Root Cause
- They all pass through evaluateScheduleTrigger() method to evaluate trigger and hence the rendering and resolving inputs processes happen, so when any exception from there happen it is propagated to this method and caught in the catch block creating logs but ignoring sending failed execution
## Solution
- Whenever an exception caught from this method create a failed execution and attach logs to them 
fail1
fail2


scheduler/src/main/java/io/kestra/scheduler/AbstractScheduler.java
```diff
@@ -712,17 +712,9 @@ private void handle() {
...
                                 if (schedulerExecutionWithTrigger.isPresent()) {
                                     this.handleEvaluateSchedulingTriggerResult(schedule, schedulerExecutionWithTrigger.get(), f.getConditionContext(), scheduleContext);
-                                } else {
-                                    // compute next date and save the trigger to avoid evaluating it each second
-                                    Trigger trigger = Trigger.fromEvaluateFailed(
-                                        f.getTriggerContext(),
-                                        schedule.nextEvaluationDate(f.getConditionContext(), Optional.of(f.getTriggerContext()))
-                                    );
-                                    trigger = trigger.checkBackfill();
-                                    this.triggerState.save(trigger, scheduleContext, "/kestra/services/scheduler/handle/save/on-eval-true/schedule");
```
loicmathieu, line 725: Why did you remove this part?

loicmathieu (changes requested):
It also miss a test

scheduler/src/main/java/io/kestra/scheduler/AbstractScheduler.java
MTarek165: - Because when exception is caught at evaluateScheduleTrigger() it returns Optional.Empty() to handle() entering this section, so instead of calling this logic there I include it at the added function and call from catch block at evaluateScheduleTrigger() before returning the empty object.
- Please take a look on the logic at the added function

MTarek165: If there is an edge case at moving this code at catch block, I can revert back this change and prepare and emit the excution at catch block in evaluateScheduleTrigger() directly

MTarek165:
> It also miss a test

I have added a test now

scheduler/src/main/java/io/kestra/scheduler/AbstractScheduler.java
MTarek165: Anyway, forget about that, I have reverted it back as I found more better way
- I changed it to create failed execution and emit it directly when evaluateScheduleTrigger() catch the exceptions
 this is more cleaner and no changes on old code 
- Added test for it and now CI is green.

scheduler/src/test/java/io/kestra/scheduler/SchedulerScheduleTest.java
```diff
@@ -489,9 +489,9 @@ void stopAfterSchedule() throws Exception {
             Await.until(() -> this.triggerState.findLast(trigger).map(t -> t.getDisabled()).orElse(false).booleanValue(), Duration.ofMillis(100), Duration.ofSeconds(10));
         }
     }
-
+// covers trigger evaluation failures from failed conditions at scheduled triggers
     @Test
-    void failedEvaluationTest() {
+    void failedEvaluation1() {
```
loicmathieu: Please name the test more accurately like `failedEvaluationFromFailedCondition()`.

scheduler/src/test/java/io/kestra/scheduler/SchedulerScheduleTest.java
```diff
@@ -527,6 +527,62 @@ void failedEvaluationTest() {
...
+            receive.blockLast();
+
+            assertThat(queueCount.getCount()).isEqualTo(0L);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+    // covers trigger evaluation failures propagated to evaluateScheduleTrigger() such as invalid expressions, input resolving issues at scheduled triggers...etc
+    @Test
+    void failedEvaluation2() {
```
loicmathieu: Please name the test more accuratly, like `failedExecutionFromInvalidExpression()`.

loicmathieu (changes requested):

MTarek165:
@loicmathieu test names updated

loicmathieu (approved):
LGTM, thanks

PR #13671, opened 2025-12-15
feat(system): save the edition in settings

loicmathieu:
This would allow to detect OSS -> EE migration.

Closes https://github.com/kestra-io/kestra-ee/issues/5106


Skraye (approved):
LGTM

PR #13685, opened 2025-12-16
fix(core): properly sync default namespace filters from settings with default filter

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra-ee/issues/6140

### DIFF
--- a/ui/src/components/filter/composables/useDefaultFilter.ts
+++ b/ui/src/components/filter/composables/useDefaultFilter.ts
@@ -25,22 +25,18 @@ export function applyDefaultFilters(
         includeScope, 
         legacyQuery,
     }: DefaultFilterOptions = {}): { query: LocationQuery, change: boolean } {
-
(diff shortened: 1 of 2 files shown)


Piyush-r-bhaskar:
Hey @Pradumnasaraf 

Could you QA this please in incognito ?

- Add a default namespace from settings -  check the dashboard, Flows, Executions, Logs has that namespace applied
- change the previous namespace selected from setting to another - new namespace will be showing now.
- Remove the default namespace from settings-  No namespace filter will be on pages
- Now manually add the namespace (after doing completing above three points) on Flows/Executions page - let's say you added on Namespace from Add Filter dropdown on Executions, it's added now switch to any other page and come back to the Executions, you will have that namespace preapplied unless manually removed to respect the restore url feature on Flows and Executions page

MilosPaunovic (approved):

Pradumnasaraf (approved):
It is working flawlessly. Thank you for fixing it.

---

Tested with Docker image above!

Issue #13673, opened 2025-12-15
ScheduleOnDates ignores inputs/labels passed with backfill


fhussonnois:
## Error

**Flow**
```
id: schedule_on_date
namespace: company.team
inputs:
  - id: message
    type: STRING
    defaults: "Hello World"

tasks:
  - id: hello
    type: io.kestra.plugin.core.log.Log
    message: "{{ inputs.message }}"

triggers:
  - id: schedule
    type: io.kestra.plugin.core.trigger.ScheduleOnDates
    dates:
      - "2025-12-01T00:00:00+01:00"
      - "2025-12-02T00:00:00+01:00"
      - "2025-12-03T00:00:00+01:00"
```

**Steps to reproduce**
1. Create  backfill with some labels and an input for `message`
2. Scheduled flows displayed the default value for the input.

### Environment

- Kestra Version: develop

fhussonnois:
Closed by: https://github.com/kestra-io/kestra/pull/13688

PR #13688, opened 2025-12-16
fix(trigger): ScheduleOnDates should work with backfill

fhussonnois:
Changes:
* ScheduleOnDates must not be re-scheduled when trigger is updated
* ScheduleOnDates must not be scheduled on previous dates when created
* ScheduleOnDates should properly support backfill

Create new SchedulableExecutionFactory class to hold all methods related
to Schedulable trigger and which are only used by core triggers

Related-to: #13673


core/src/main/java/io/kestra/plugin/core/trigger/SchedulableExecutionFactory.java
loicmathieu, line 1: duplicated

loicmathieu (approved):
Overall LGTM but it seems there is a duplicated service and a test would have been great

PR #13689, opened 2025-12-16
feat(trigger): refactor Schedule to not use the application context

loicmathieu:
Part-of:  https://github.com/kestra-io/kestra-ee/issues/4228


fhussonnois (approved):
LGTM!

PR #13703, opened 2025-12-16
fix(scheduler): emit ExecutionKilled event when real-time trigger is deleted

fhussonnois:


scheduler/src/main/java/io/kestra/scheduler/TriggerEventHandler.java
```diff
@@ -290,7 +300,27 @@ void onTriggerUpdated(Clock clock, TriggerUpdated event) {
...
+                this.executionKilledQueue.emit(ExecutionKilledTrigger
+                    .builder()
+                    .tenantId(state.getTenantId())
+                    .namespace(state.getNamespace())
+                    .flowId(state.getFlowId())
+                    .triggerId(state.getTriggerId())
+                    .build()
+                );
+            } catch (QueueException e) {
+                Logs.logTrigger(event.id(), Level.WARN, "Cannot kill real-time trigger. Cause: {}", e.getMessage(), e);
```
loicmathieu:
```suggestion
                Logs.logTrigger(event.id(), Level.WARN, "Cannot kill a real-time trigger, it will continue processing until Kestra is restarted. Cause: {}", e.getMessage(), e);
```

loicmathieu: Better be explicit as I don't think there are other way in this case

loicmathieu (approved):
LGMT

PR #13729, opened 2025-12-17
fix(executions): query ExecutionDelay to UTC to avoid any offset error

fhussonnois:


loicmathieu:
I think we may suffer the same for the SLA monitor loop, and maybe also for multiplecondition storage :(
Maybe even more

fhussonnois:
@loicmathieu I checked SLA and MultipleCondition and we are good. We already use an `Instant` so no TZ issue.

loicmathieu (approved):
LGTM

PR #13730, opened 2025-12-17
docs(PurgeFiles): update documentation and example

aj-emerich:
I am not 100% on the example, but I tried to include each property. @brian-mulier-p can you adjust if you need to?

### ✨ Description

What does this PR change?  
Replaces title, description, and example for the PurgeFiles core plugin

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/namespace/PurgeFiles.java
+++ b/core/src/main/java/io/kestra/plugin/core/namespace/PurgeFiles.java
@@ -26,25 +26,29 @@ import java.util.concurrent.atomic.AtomicLong;
 @Getter
 @NoArgsConstructor
 @Schema(
-    title = "Delete expired keys globally for a specific namespace.",
-    description = "This task will delete expired keys from the Kestra KV store. By default, it will only delete expired keys, but you can choose to delete all keys by setting `expiredOnly` to false. You can also filter keys by a specific pattern and choose to include child namespaces."
+    title = "Purge namespace files for one or multiple namespaces.",
+    description = "This task purges namespace files (and their versions) stored in Kestra. You can restrict the purge to specific namespaces (or a namespace glob pattern), optionally include child namespaces, and filter files by a glob pattern. The purge strategy is controlled via `behavior` (e.g. keep the last N versions and/or delete versions older than a given date)."
 )
 @Plugin(
     examples = {
         @Example(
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/kestra/plugin/core/namespace/PurgeFiles.java
```diff
@@ -26,25 +26,29 @@
...
+                  - id: purge_files
+                    type: io.kestra.plugin.core.namespace.PurgeFiles
                     namespaces:
                       - company
                     includeChildNamespaces: true
+                    filePattern: "**/*.sql"
+                    behavior:
+                      type: version
+                      keepAmount: 3
+                      before: "2025-01-01T00:00:00Z"
```
brian-mulier-p: There is a validation to forbid using both at the same time (due to being confusing when using both we decided to keep it simple and only allow one of each for now)

brian-mulier-p (changes requested):
Small fix otherwise LGTM, ty, you can see the inspiration, I forgot to rewrite the doc after copy paste sorry :'(

core/src/main/java/io/kestra/plugin/core/namespace/PurgeFiles.java
aj-emerich: Got it!

core/src/main/java/io/kestra/plugin/core/namespace/PurgeFiles.java
```diff
@@ -26,25 +26,29 @@
...
-                    expiredOnly: true
+                  - id: purge_files
+                    type: io.kestra.plugin.core.namespace.PurgeFiles
                     namespaces:
                       - company
                     includeChildNamespaces: true
+                    filePattern: "**/*.sql"
+                    behavior:
+                      type: version
+                      keepAmount: 3
```
aj-emerich:
```suggestion
```

brian-mulier-p (approved):
LGTM

aj-emerich:
@brian-mulier-p Anything to do about backend tests failing? Should we wait for tests to be fixed or what? I assume this change has no backend impact, but I don't want to merge reds 😄

brian-mulier-p:
I rebased, there should be no harm in merging this anyway but for the green sake let's wait a bit more. Then I'll merge and backport it.

PR #13733, opened 2025-12-17
Feat/micronaut 4 10

### LINKED ISSUE #9743: Rare issue in HTTP multi-part parser
A bug exists in the HTTP multi-part parser from Netty (used by Micronaut), where a `\r `character is incorrectly added at the end of all input parts. This affects any input, potentially causing problems for downstream consumers.
- The root cause: https://github.com/netty/netty/issues/12729 Netty issue raised over 3 years ago
- A fix has been proposed and implemented by a Quarkus community contributor, but it has not yet been merged (since 7 months)

Impact:
- Affects all Java applications using Netty for multi-part parsing.
- Trimming all input as a workaround is not ideal since it may cause backward compatibility issues.
- The issue occurs randomly and infrequently (approx. 1 in 10,000 executions), making it difficult to reproduce and debug.

Current Workarounds & Limitations:
- Forking Netty is not a viable option due to the complexity and maintenance overhead.
- Patching Netty within dependencies is not recommended.
- Using a trim function in a Kestra Flow logic is unreasonable as our user would need to rewrite all their flows to handle this across all their inputs.

For now, there's no solution other than to hope https://github.com/netty-contrib/codec-multipart/pull/25 will be merged soon

loicmathieu:
Closes https://github.com/kestra-io/kestra/pull/13713
Fixes #9743

### DIFF
--- a/platform/build.gradle
+++ b/platform/build.gradle
@@ -30,7 +30,7 @@ dependencies {
     // as Jackson is in the Micronaut BOM, to force its version we need to use enforcedPlatform but it didn't really work, see later :(
     api enforcedPlatform("com.fasterxml.jackson:jackson-bom:$jacksonVersion")
     api enforcedPlatform("org.slf4j:slf4j-api:$slf4jVersion")
-    api platform("io.micronaut.platform:micronaut-platform:4.9.4")
+    api platform("io.micronaut.platform:micronaut-platform:4.10.5")
     api platform("io.qameta.allure:allure-bom:2.32.0")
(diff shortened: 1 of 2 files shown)


fhussonnois (approved):

loicmathieu:
@yuri1969 this should fix the multipart content parsing bug. I can try to backport for you, on which version do you need it?

yuri1969:
@loicmathieu Thanks for your effort, but the relevant environment is still at v0.22.

loicmathieu:
@yuri1969 I can try to backport it down to 0.22, I'll have a look and report here if it works or not.

PR #13736, opened 2025-12-17
fix: Make sure parentTaskRun attempts are also set to Killed

Skraye:
Relate to https://github.com/kestra-io/kestra-ee/issues/6069


loicmathieu (approved):
LGTM but modifying an existing test would have been great

Skraye:
Added a test, will merge once ci is done

Issue #12557, opened 2025-10-30
Execution continues after killed subflow


sjoerdsmink:
### Describe the issue

Normally when a task run is killed, it seems to stop further steps. [Documentation about the states](https://kestra.io/docs/workflow-components/states) is not entirely clear on the relation between run states and execution states, but it seems to do that.

```
id: test_killed
namespace: test

tasks:
  - id: killed
    type: io.kestra.plugin.core.execution.Exit
    state: KILLED
  - id: log
    type: io.kestra.plugin.core.log.Log
    message: "Continued"
```

<img width="836" height="281" alt="Image" src="https://github.com/user-attachments/assets/30b2a6b4-8e1b-4b59-9ab1-a97a38866bc1" />

But when a subflow is killed, it seems to continue with the tasks.
```
id: test_subflow
namespace: test

tasks:
  - id: pause
    type: io.kestra.plugin.core.flow.Pause
```

```
id: test_mainflow
namespace: test

tasks:
  - id: createAction
    type: io.kestra.plugin.core.flow.Subflow
    namespace: test
    flowId: test_subflow
  - id: log
    type: io.kestra.plugin.core.log.Log
    message: "Continued"
```

1. Run the main flow
2. Open the subflow
3. Kill the paused subflow
4. I would expect `log` not to be executed, but it is.

<img width="1016" height="579" alt="Image" src="https://github.com/user-attachments/assets/b8f503ef-85cc-435e-bd6c-a0dd84efa45e" />

### Environment

- Kestra Version: v1.0.7

sjoerdsmink:
@anna-geller Just want to make you aware of this issue as I think it's at the core of your product. I completely understand there are many issues that needs your attention, but this concerns the the state machine.

loicmathieu:
The subflow execution is killed ,but not the parent execution.
You kill an execution not a task.

Maybe we need to end the parent's execution `Subflow` task in FAILED and not KILLED when you kill a subflow.
If you want to kill both the parent and the child execution, you should kill the parent as by default it will kill each subflow transitively.

Ben8t:
> Maybe we need to end the parent's execution Subflow task in FAILED and not KILLED when you kill a subflow.
If you want to kill both the parent and the child execution, you should kill the parent as by default it will kill each subflow transitively.

Yes, makes total sense to me

sjoerdsmink:
[...]

PR #13746, opened 2025-12-18
fix(ui): rephrase "kill parents" to "kill currernt"

loicmathieu:
This has always been kill current / kill current and sublow as we never kill parent executions, it's a kill on cascade that didn't go backward.

Part-of: #12557


Ben8t (approved):

loicmathieu:
Backported to 1.0 and 1.1

Issue #13759, opened 2025-12-18
Add new CLI to update all flows from sources


nkwiatkowski:
During LM migration, we found out that some flows where persisted with pluginDefaults. To remove those unwanted fields, we want to have a new CLI that can update all flows from their source.

PR #13760, opened 2025-12-18
feat(CLI): add a new update from flow source CLI

nkwiatkowski:
- fix #https://github.com/kestra-io/kestra/issues/13759


loicmathieu (approved):
LGMT, I would name it a little differently as it's recreate the flow definition from source so maybe something like `FlowSyncFromSource` and not update?

PR #13763, opened 2025-12-18
Feat/assets

brian-mulier-p:
part of https://github.com/kestra-io/kestra-ee/issues/5435
Part of https://github.com/kestra-io/kestra-ee/issues/6173


core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
loicmathieu, line 1: I'm not fan of such customization only for assets.
@fhussonnois WDYT I know you work on the same kind of issues.

Normally JSON subtypes should be used here not such customization

core/src/main/java/io/kestra/core/models/assets/AssetUser.java
```diff
@@ -0,0 +1,17 @@
+package io.kestra.core.models.assets;
+
+import io.kestra.core.models.flows.FlowId;
+import io.kestra.core.utils.IdUtils;
+
+/**
+ * Represents an entity that used an asset
+ */
+public record AssetUser(String tenantId, String namespace, String flowId, Integer flowRevision, String executionId, String taskId, String taskRunId) {
```
loicmathieu: User has a strong "human" connotation, maybe AssetUsage would be a better name?

core/src/main/java/io/kestra/core/models/assets/AssetsDeclaration.java
```diff
@@ -0,0 +1,22 @@
...
+
+@Getter
+public class AssetsDeclaration extends AssetsInOut {
+    private boolean enableAuto;
+
+    @JsonCreator
+    public AssetsDeclaration(Boolean enableAuto, List<AssetIdentifier> inputs, List<Asset> outputs) {
+        super(inputs, outputs);
+
+        this.enableAuto = Optional.ofNullable(enableAuto).orElse(true);
```
loicmathieu: I thought we agree to have `false` as default for two reasons:
- performance
- least surprise

core/src/main/java/io/kestra/core/models/assets/DatasetAsset.java
```diff
@@ -0,0 +1,73 @@
...
+import io.kestra.core.models.annotations.Plugin;
+import lombok.Builder;
+import lombok.NoArgsConstructor;
+
+import java.time.Instant;
+import java.util.Map;
+import java.util.Optional;
+
+@NoArgsConstructor
+@Plugin(aliases = DatasetAsset.ASSET_TYPE)
```
loicmathieu: aliases ater not designed for that, they are design for "replacement" and generates an "info" notice at use site.
`@Plugin.Id` are designed for that.
Here you abuse aliases

core/src/main/java/io/kestra/core/models/assets/Dataset.java
loicmathieu, line 1: @Ben8t we already discuss this multiple time but here our dataset miss so many fields compared to the OpenLineage dataset that I cannot see how to map it to the OpenLineage dataset.
It even miss the basic mandatory fields to identify a dataset: name and namespace.

core/src/main/java/io/kestra/core/models/assets/AssetsDeclaration.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.List;
+import java.util.Optional;
+
+@Getter
+public class AssetsDeclaration extends AssetsInOut {
+    private boolean enableAuto;
+
+    @JsonCreator
+    public AssetsDeclaration(Boolean enableAuto, List<AssetIdentifier> inputs, List<Asset> outputs) {
```
loicmathieu, line 17: I'm not sure I understand why you have asset indentifier as inputs but full assets as outputs.
In both cases assets can either be pre-defined or auto-defined so you should have assets in both places no?

core/src/main/java/io/kestra/core/models/assets/AssetsInOut.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.Optional;
+
+@Getter
+public class AssetsInOut {
+    private List<AssetIdentifier> inputs;
+
+    private List<Asset> outputs;
+
+    @JsonCreator
+    public AssetsInOut(List<AssetIdentifier> inputs, List<Asset> outputs) {
```
loicmathieu, line 17: Same here, no sure to understand why not both are assets

core/src/main/java/io/kestra/core/runners/AclCheckerImpl.java
```diff
@@ -6,15 +6,19 @@
 import java.util.List;
 import java.util.Objects;
 
-class AclCheckerImpl implements AclChecker {
+public class AclCheckerImpl implements AclChecker {
```
loicmathieu: Why are you doing that?
It's per design that implementations are hidden.
If there is no strong reason, please revert

core/src/main/java/io/kestra/core/runners/Assets.java
```diff
@@ -0,0 +1,12 @@
+package io.kestra.core.runners;
+
+import io.kestra.core.models.assets.Asset;
+import io.kestra.core.queues.QueueException;
+
+import java.util.List;
+
+public interface Assets {
```
loicmathieu: As it's for the RunContext, I would prefer to have a more specific name like `AssetEmitter` event if you get it from `runContext.assets()` for simplicity

core/src/main/java/io/kestra/core/runners/RunContext.java
```diff
@@ -190,7 +193,18 @@ public StateStore stateStore() {
      */
     public abstract LocalPath localPath();
 
+    public record ExecutionInfo(String id, String taskRunId) {
```
loicmathieu: This is not executionInfo but TaskRunInfo.
And I think it would be great to also add tenantid, namespace and flowid so it would be easier to use.

loicmathieu: Or maybe as the other id are already present in FlowInfo, having TaskRunInfo(String executionId, String taskId, String taskRunId, Object value)` would be a good idea

core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
```diff
@@ -594,6 +609,19 @@ public AclChecker acl() {
         return new AclCheckerImpl(this.applicationContext, flowInfo());
     }
 
+    @Override
+    public Assets assets() {
+        if (this.assetsManager == null) {
```
loicmathieu: This is not thread safe and may be seen as premature optimization.
I'll prefer if you just build it when needed as we do for `acl()` the other solution would be to construct it at init but as it may not be used very often building it when needed seems enought for me

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -283,6 +298,10 @@ private Optional<WorkerTaskResult> childWorkerTaskResult(FlowWithSource flow, Ex
                     return Optional.of(new WorkerTaskResult(workerTaskResult
                         .getTaskRun()
                         .withOutputs(variables)
+                        .withAssets(new AssetsInOut(
```
loicmathieu: I don't think any flowable task would generate assets.
It may be better to restrict assets to worker task, you can do that by updating the existing validation code that prevent some fields to be used outside runnable tasks.
At some point RunnableTask and FlowableTask would be refactoted to not share anymore such attributes that cannot be used in both

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -240,12 +250,17 @@ private Optional<WorkerTaskResult> childWorkerTaskResult(FlowWithSource flow, Ex
 
         if (parent instanceof FlowableTask<?> flowableParent) {
 
-            RunContext runContext = runContextFactory.of(flow, parent, execution, parentTaskRun);
+            RunContext runContext = runContextInitializer.forExecutor((DefaultRunContext) runContextFactory.of(
```
loicmathieu: This is costly and it should not be needed.

executor/src/main/java/io/kestra/executor/ExecutorService.java
loicmathieu, line 1: I would not support assets outside of runnable tasks, I cannot see any legitimate use case anyway as those tasks are "orchestration tasks" so they should not use any assets.

Thanks to that, the code to "add the asset" to a taskrun would be moved inside the worker which would drastically simplify the implementation.

Then, in the executor, when it receive a worker task that contains assets, it will emit the needed events.

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -1172,6 +1232,45 @@ public void addWorkerTaskResult(Executor executor, Supplier<FlowWithSource> flow
...
+                    taskRun.getTenantId(),
+                    taskRun.getNamespace(),
+                    taskRun.getFlowId(),
+                    newExecution.getFlowRevision(),
+                    taskRun.getExecutionId(),
+                    taskRun.getTaskId(),
+                    taskRun.getId()
+                );
+
+                if (!taskRun.getAssets().getInputs().isEmpty() || !taskRun.getAssets().getOutputs().isEmpty()) {
```
loicmathieu: you can merge this if inside the enclosing if so if no inputs and no outputs the code is not run at all

executor/src/main/java/io/kestra/executor/ExecutorService.java
loicmathieu: Moreover, I would revert any use of the runContext initializer that was not stricly mandatory as it may impact performance

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
```diff
@@ -294,6 +297,8 @@ public void run() {
 
         Await.until(() -> this.allFlows != null, Duration.ofMillis(100), Duration.ofMinutes(5));
 
+        this.receiveCancellations.addAll(assetService.eventListeners(Executor.class));
```
loicmathieu: I don't understand that it seems useless

loicmathieu (changes requested):
I think it would be great to have some insight from @fhussonnois at least on JSON serialization of assets, run context changes and ExecutionService changes

core/build.gradle
```diff
@@ -24,6 +24,9 @@ dependencies {
     // reactor
     api "io.projectreactor:reactor-core"
 
+    // awaitility
+    api "org.awaitility:awaitility"
```
loicmathieu, line 28: Shouldn't it be testImplementation only?

core/src/main/java/io/kestra/core/models/assets/AssetUser.java
loicmathieu: I would not add the revision here, if the assert is not more used by a new revision of a flow we should delete it.
It's what we did for FlowTopology and we should align both to be consistent.

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/models/assets/AssetUser.java
brian-mulier-p: You have no way to know that, dynamic assets will be big part of the asset's future and you can't know if one will still be used every once in a while in a task because... It's dynamic..

core/src/main/java/io/kestra/core/models/assets/DatasetAsset.java
brian-mulier-p: I know but I do what I can in the time I have 😉
Plugin ids are only used in configuration there is no mechanism of overriding an id as of now.

core/src/main/java/io/kestra/core/models/assets/AssetsDeclaration.java
brian-mulier-p: Input assets are expected to be existing (or we will create them as "EXTERNAL" type). A task should not generate its own input assets but rely on existing ones that's why.

core/src/main/java/io/kestra/core/models/assets/AssetsDeclaration.java
brian-mulier-p: idk I don't remember about it 

Image

core/src/main/java/io/kestra/core/models/assets/AssetUser.java
brian-mulier-p: It's already my db-persisted entity but I agree user can be seen as human... However I didn't find a better wording, it's really a user which can be an entity like flow or execution but idk...

core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
brian-mulier-p: It's not optimization it's by design, since things are done in memory before being sent in workertaskresult I need to hold the same instance to get back the emitted assets. However I'll add a synchronized block.

core/build.gradle
brian-mulier-p: I would say no as we want to shift to that pattern for any await?

core/src/main/java/io/kestra/core/models/assets/Dataset.java
Ben8t: Dataset in Kestra is not the same definition as in OpenLineage. We said it several time with @tchiotludo.

For name you can use the asset `id`, for namespace you can use `system`. Taking also inspiration from Airflow [example in OL documentation](https://openlineage.io/docs/spec/facets/custom-facets), this doesn't seem a blocker

core/src/main/java/io/kestra/core/models/assets/DatasetAsset.java
```diff
@@ -12,6 +12,7 @@
 
 @NoArgsConstructor
 @Plugin(aliases = DatasetAsset.ASSET_TYPE)
+@Plugin.Id()
```
loicmathieu: My point was, remove the alias and use `@Plugin.Id(DatasetAsset.ASSET_TYPE)` instead

core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
```diff
@@ -610,16 +613,18 @@ public AclChecker acl() {
...
     @Override
-    public Assets assets() {
-        if (this.assetsManager == null) {
-            this.assetsManager = assetManagerFactory.of(
-                Optional.ofNullable(task).map(Task::getAssets).map(AssetsDeclaration::isEnableAuto)
-                    .or(() -> Optional.ofNullable(trigger).map(AbstractTrigger::getAssets).map(AssetsDeclaration::isEnableAuto))
-                    .orElse(true)
-            );
+    public AssetEmitter assets() {
+        synchronized (this) {
```
loicmathieu: Still not a safe construct, you should use double-checked locking.

loicmathieu: I think a more simple impl like what's done for metric would be enought here.

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/models/assets/Dataset.java
loicmathieu: `id` and `namespace` are what identify a dataset in OpenLineage, as explained a dataset is shared at the enterprise level between all applications, and name and the namespace is defined by the standard so you cannot use `system`.

We'll discuss this next wek together so, as i said, I cannot see how to fit our Dataset object inside OpenLineage.

loicmathieu: @Ben8t the link you provided is for custom facets which allow to send **additional** metadata to OpenLineage, in my early prototype this is used to send Kestra specific task-level metadata.

But you need to fill the other facets, some are mandatory, others not. If you didn't fill the other facets you're defeating the whole purpose of OpenLineage which is a standard in how you represent lineage data.

Let's sync next week t discuss this.

Ben8t: I think it can fit if we treat Kestra “asset” as an OL Dataset identifier + custom facet.
In OL, Dataset uniqueness is (namespace, name) where the only real constraint is “name unique within namespace”. Our asset uniqueness is <tenantId, assetId>, so we can map deterministically:

namespace = `kestra://<tenantId>` (or `kestra://<kestra-host>/<tenantId>` if needed to avoid collisions across deployments)
name = `<assetId>`

Then we put the free-form metadata in a custom facet kestra.asset (tenantId, assetId, plus other metadata).
This keeps OL standard semantics and still preserves Kestra specifics via facets.

Let's discuss yes next week

core/src/main/java/io/kestra/core/models/assets/CustomAsset.java
```diff
@@ -0,0 +1,30 @@
...
+import io.kestra.core.models.annotations.Plugin;
+import lombok.Builder;
+import lombok.NoArgsConstructor;
+
+import java.time.Instant;
+import java.util.Map;
+
+@NoArgsConstructor
+@Plugin
+public class CustomAsset extends Asset {
```
fhussonnois: I don't fully grasp why we have a `CustomAsset` class. It couldn't have been simpler to just make the `Asset` class not abstract ? So it can be use directly as a fallback asset type in the deserializer

core/src/main/java/io/kestra/core/models/assets/DatasetAsset.java
fhussonnois: Just why ? This is clearly a hack of the aliases mechanism which was designed to provide a backward compatibility in case of a plugin being renamed. 

Why don't we use the full qualified class name as usual? Asset is a plugin so it should use the FQCN. This is a design principal of Kestra.

core/src/main/java/io/kestra/core/runners/WorkerTaskResult.java
```diff
@@ -10,6 +11,7 @@
 import java.util.List;
 
 import jakarta.validation.constraints.NotNull;
+import lombok.With;
```
fhussonnois: This import seems to benot used

core/src/main/java/io/kestra/core/services/AssetService.java
```diff
@@ -0,0 +1,32 @@
...
+import io.kestra.core.queues.QueueException;
+import jakarta.annotation.PostConstruct;
+import jakarta.inject.Named;
+import jakarta.inject.Singleton;
+
+import java.util.Collections;
+import java.util.List;
+
+@Singleton
+public class AssetService implements Runnable {
```
fhussonnois: I would have create a package `io.kestra.core.assets`. We start having anything and everything in the `io.kestra.core.service`. 

This class should be changed to an `interface` with a Noop implementation for OSS.

fhussonnois (changes requested):

core/src/main/java/io/kestra/core/models/assets/DatasetAsset.java
loicmathieu: Decision has been to use the FQCN as for all other plugins

core/src/main/java/io/kestra/core/runners/WorkerTaskResult.java
loicmathieu: removed

core/src/main/java/io/kestra/core/services/AssetService.java
loicmathieu: Done

core/src/main/java/io/kestra/core/models/assets/Dataset.java
loicmathieu: Decision is to mandate mapping definition in the asset exporter

core/src/main/java/io/kestra/core/models/assets/CustomAsset.java
loicmathieu: We always use an abstract root plugin class, I think we should follow the same pattern here.

loicmathieu: But it may be a nested type inside the Asset type because it should never be used directly. I wonder if we should not hide it by the way (`@Hidden`)

MilosPaunovic (approved):
Okay, as far as the `UI` side goes.

loicmathieu (approved):

PR #13764, opened 2025-12-19
jdbc: Implement robust retry logic for JDBC execution queue

### LINKED ISSUE #13744: Concurrency limit is not work when execution is killed in bulk
### Describe the issue

I've found that there is still an issue with the concurrency control.

When the workflow concurrency limit is set to 2:
```
id: hello-world
namespace: tutorial
tasks:
  - id: second_task
    commands:
      - sleep 1m
concurrency:
  limit: 2
```
I kill two running executions in bulk, only one execution is started afterward. This results in fewer executions actually running than the concurrency limit.

Image

### Environment

- Kestra Version: develop

vamsi172323:
This commit refactors the pop method in AbstractJdbcExecutionQueuedStorage to better handle database lock contention when multiple workers are polling the queue.The new implementation includes:

- A retry loop with a capped exponential backoff to wait and retry if an item is locked.
- Each retry attempt now uses its own short-lived transaction.
- An additional check is performed to exit the loop early if the queue is empty, avoiding unnecessary waits.
- Retry parameters have been extracted into named constants for clarity.

The new unit test, flowConcurrencyTwoQueue, was added to FlowConcurrencyCaseTest to validate this change. This test failed 4 out of 5 times before the fix and now passes consistently, confirming the improved reliability of the queueing mechanism.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/13744
Related to https://github.com/kestra-io/kestra/issues/13211

### 🛠️ Backend Checklist

### DIFF
--- a/core/src/test/java/io/kestra/core/runners/AbstractRunnerConcurrencyTest.java
+++ b/core/src/test/java/io/kestra/core/runners/AbstractRunnerConcurrencyTest.java
@@ -89,4 +89,10 @@ public abstract class AbstractRunnerConcurrencyTest {
     void flowConcurrencyQueueKilled() throws Exception {
         flowConcurrencyCaseTest.flowConcurrencyQueueKilled("flow-concurrency-queue-killed");
     }
+
+    @Test
+    @LoadFlows(value = {"flows/valids/flow-concurrency-2-queue.yml"}, tenantId = "flow-concurrency-2-queue")
+    void flowConcurrencyTwoQueue() throws Exception {
+        flowConcurrencyCaseTest.flowConcurrencyTwoQueue("flow-concurrency-2-queue");
+    }
 }
--- a/core/src/test/java/io/kestra/core/runners/FlowConcurrencyCaseTest.java
+++ b/core/src/test/java/io/kestra/core/runners/FlowConcurrencyCaseTest.java
@@ -330,6 +330,61 @@ public class FlowConcurrencyCaseTest {
         }
     }
 
+    public void flowConcurrencyTwoQueue(String tenantId) throws QueueException, InterruptedException {
+        Flow flow = flowRepository
+            .findById(tenantId, NAMESPACE, "flow-concurrency-2-queue", Optional.empty())
+            .orElseThrow();
+        Execution execution1 = runnerUtils.runOneUntilRunning(tenantId, NAMESPACE, "flow-concurrency-2-queue", null, null, Duration.ofSeconds(30));
+        Execution execution2 = runnerUtils.runOneUntilRunning(tenantId, NAMESPACE, "flow-concurrency-2-queue", null, null, Duration.ofSeconds(30));
+        Execution execution3 = runnerUtils.emitAndAwaitExecution(e -> e.getState().getCurrent().equals(Type.QUEUED), Execution.newExecution(flow, null, null, Optional.empty()));
+        Execution execution4 = runnerUtils.emitAndAwaitExecution(e -> e.getState().getCurrent().equals(Type.QUEUED), Execution.newExecution(flow, null, null, Optional.empty()));
+
+        try {
+            assertThat(execution1.getState().isRunning()).isTrue();
+            assertThat(execution2.getState().isRunning()).isTrue();
+            assertThat(execution3.getState().getCurrent()).isEqualTo(Type.QUEUED);
+            assertThat(execution4.getState().getCurrent()).isEqualTo(Type.QUEUED);
+
+            // we kill execution 1 & 2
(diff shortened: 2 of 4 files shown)


loicmathieu:
@vamsi172323 thanks for your contribution but I'm a bit skeptic of the solution here, we use forUpdate().skipLocked() with a limit of 1 so normally it should skip the already locked record and pop another one if existing so retrying should not be needed.

loicmathieu:
If it didn't work, we may just remove the skipLock() it should wait at the database level for the other transaction to ends.

vamsi172323:
> If it didn't work, we may just remove the skipLock() it should wait at the database level for the other transaction to ends.

@loicmathieu, SKIP LOCKED almost always works well and is a good design choice because it prevents threads from blocking. Database-as-queue patterns commonly rely on SKIP LOCKED for this reason. Since SKIP LOCKED can legitimately return no rows at times, workers must use a poll-and-wait strategy. This approach preserves performance and throughput while avoiding thread synchronization and lock contention at the database level.

The PR I raised implements this poll-and-wait behavior. It is triggered either when there are no remaining items to delete or when the database temporarily returns no rows in valid scenarios, such as the one described in this issue. In the majority of cases, a row is immediately returned and processing continues without delay. I believe adding this retry mechanism is necessary and correct.

loicmathieu:
@vamsi172323 if the concern is that in some circustmances there are races that makes skip locked return no rows but there are existing ones, I think a safier and more simple implementation would be when there are no rows return to use an exclusive lock: so when nothing is return to redo the same query but without SKIP LOCKED. WDYT?

vamsi172323:
> @vamsi172323 if the concern is that in some circustmances there are races that makes skip locked return no rows but there are existing ones, I think a safier and more simple implementation would be when there are no rows return to use an exclusive lock: so when nothing is return to redo the same query but without SKIP LOCKED. WDYT?

Following are the four cases I can think of 

1. **Valid row available**

   * If a row exists and is not locked, `SKIP LOCKED` returns it.
   * The thread processes the row, deletes it from the table, and returns.
   * This is the normal “happy path.”
   
2. **Empty table**

   * If the table contains no entries, `FOR UPDATE SKIP LOCKED` returns no row.
   * The `COUNT` query confirms the table is empty.
   * The thread safely returns without processing anything.

3. **Row available but `SKIP LOCKED` returns null**

   * In rare cases, a row may be available but `SKIP LOCKED` returns no row like in this issue's scenario
   * The `COUNT` query prevents premature return, so the thread retries.
   * On a subsequent attempt, `SKIP LOCKED` will likely return the row, allowing the thread to process it and continue.
   * This ensures that legitimate rows are not missed.

4. **All rows locked by other threads**

   * If there are rows but all are currently locked, `SKIP LOCKED` returns no row.
   * The `COUNT` query detects that rows exist, so the thread retries.
   * As other threads finish processing and delete their rows, eventually the table becomes empty.
   * The retrying thread sees the table empty and returns.
   * This situation typically occurs when the number of queued executions is less than the concurrency limit, and looks safe to me

Using SELECT FOR UPDATE without SKIP LOCKED as a fallback would introduce unnecessary blocking in scenario 3, whereas a simple retry with SKIP LOCKED is likely to succeed without waiting. Given the current setup, where terminating a flow must trigger another flow’s execution, this retry mechanism is efficient and avoids blocking threads at the database level. I welcome any feedback or suggestions. What are your thoughts?

loicmathieu:
Maybe @fhussonnois has some idea about it, I don't like handing retry and using Thread.sleep() if not absolutely necessary.

vamsi172323:
> I don't like handing retry and using Thread.sleep() if not absolutely necessary

As mentioned [above](https://github.com/kestra-io/kestra/pull/13764#issuecomment-3675006786), in the first two cases, no retry or sleep is needed. In the third case, retrying is necessary; otherwise, the Kestra flow would use less concurrency than configured. The fourth case occurs when the number of queued executions is less than the concurrency limit and all other concurrent executions finish at the same time, this is uncommon and should be safe. I believe this poll-and-wait mechanism helps ensure we meet concurrency expectations under the current implementation model.

vamsi172323:
As mentioned in my comments, this PR is correct under the current implementation model. However, PR https://github.com/kestra-io/kestra/issues/13785 changes the implementation model by making the entire decrement and pop logic single-threaded. In that model (concurrent threads stop at decrement operation of the concurrency limit value itself until current thread closes the transaction), the scenario where a SKIP LOCKED returns a null value due to concurrent threads will no longer occur. As a result, this PR is no longer needed, so it has been closed. This issue will not happen once the other PR goes through.

PR #13768, opened 2025-12-19
doc(task): improve TemplatedTask example

Ben8t:


fdelbrayelle (approved):

fdelbrayelle:
I think this one is safe to merge, what do you think @MilosPaunovic on E2E side?

MilosPaunovic:
I've updated the branch and it will re-run all tests, but even if they fail this is good to merge, as the change is not related to frontend.

MilosPaunovic:
Tests were sorted out in the meantime and now they've passed.

PR #13769, opened 2025-12-19
fix(scheduler): remove mutation methods from TriggerRepositoryInterface

fhussonnois:
Make sure any change made on a trigger state is performed by the scheduler.

Move the TriggerStateStore interface to core, and remove default implementation to use the repository implementation


loicmathieu (approved):
LGTM

PR #13783, opened 2025-12-20
fix(core): os incompitibility file path problem

### LINKED ISSUE #13740: Namespace files on Windows can't be accessed inside of Kestra
### Describe the issue

Namespace files cannot be used in Windows.

### Environment

- Kestra Version: develop
- Windows 11 Standalone

MTarek165:
Closes https://github.com/kestra-io/kestra/issues/13740.

## Description
- UNC path problem occur during file path normalization because Path.of() is OS dependent for ex:
when passing Path.of("/file.txt") to normalize at windows it becomes "\file.txt" entering the second check as it will consider it not starting with "/" returning "/\file.txt" causing UNC problem as it expects a correct UNC  path not interpreted as file path
## Solution
- we must convert windows style to unix style before checks as checks assumes unix style
## Testing  
- added test for both unix and windows styles and assert on their logical paths to ensure they are working on all environments
- it tests normalization by simulating running on both by passing the formats produced from windows and unix styles and should normalize them correctly ( although that is sufficient, for more certainty it also passes locally on my windows machine and on CI (Linux) )
## Notes
 - now this fixes many of failing tests related to files if running on windows since all depends on normalize()
 - there is redundant checks ( third check ) at normalize() as always we consider leadingBackSlash is true so keeping these checks and flag adding complexity only, instead WindowsUtils.windowsToUnixPath(String path) handle the issue with adding leading back slash as it is always the case  
 - on the long term, it may be better to depend on unified logical path using String to avoid these OS issues because Path object depends on the OS which may make problems as the path format isn't unified

### DIFF
--- a/core/src/main/java/io/kestra/core/models/namespaces/files/NamespaceFileMetadata.java
+++ b/core/src/main/java/io/kestra/core/models/namespaces/files/NamespaceFileMetadata.java
@@ -94,7 +94,7 @@ public class NamespaceFileMetadata implements DeletedInterface, TenantInterface,
         return NamespaceFileMetadata.builder()
             .tenantId(tenantId)
             .namespace(namespaceFile.namespace())
-            .path(namespaceFile.path(true).toString())
+            .path(namespaceFile.filePath().toString())
             .version(namespaceFile.version())
             .build();
     }
--- a/core/src/main/java/io/kestra/core/models/namespaces/files/NamespaceFileMetadata.java
+++ b/core/src/main/java/io/kestra/core/models/namespaces/files/NamespaceFileMetadata.java
@@ -94,7 +94,7 @@ public class NamespaceFileMetadata implements DeletedInterface, TenantInterface,
         return NamespaceFileMetadata.builder()
             .tenantId(tenantId)
             .namespace(namespaceFile.namespace())
-            .path(namespaceFile.path().toString())
+            .path(namespaceFile.path(true).toString())
             .version(namespaceFile.version())
             .build();
     }
--- a/core/src/main/java/io/kestra/core/models/namespaces/files/NamespaceFileMetadata.java
+++ b/core/src/main/java/io/kestra/core/models/namespaces/files/NamespaceFileMetadata.java
@@ -94,7 +94,7 @@ public class NamespaceFileMetadata implements DeletedInterface, TenantInterface,
         return NamespaceFileMetadata.builder()
             .tenantId(tenantId)
             .namespace(namespaceFile.namespace())
-            .path(namespaceFile.path(true).toString())
+            .path(namespaceFile.path().toString())
             .version(namespaceFile.version())
             .build();
     }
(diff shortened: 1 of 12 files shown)


loicmathieu (changes requested):
Instead of doing the convertion Windows <-> Unix all the time, I think it would be easier to do it once when we save the namespace file so it's always saved as a Unix path.

MTarek165:
I didn't do that as It is may be a bigger change and I am not fully aware if there is case may depend on differentiating formats using Path, for now we can do conversion before repository save and after some time change this?

MTarek165:
@loicmathieu After some investigation I see that converting path from windows to unix when saving file only has a bad consequences although it appears easier at first but when I tried to make such a change I found more challenges:

- If we save a file path as Unix like /folder/file.txt in putFile() always as asked , for ex: retrieving it with fileByPath() after that must use the exact same value and since fileByPath() and the file operations are using Path, on Windows this would be converted to \folder\file.txt, causing a mismatch (since we always storing it as unix path now) and no result would be returned so we got back to the same problem again for windows.

- This would require changing all Path objects to String across all file operations since they are tightly coupled in order to avoid Path OS dependency, which is bad because we lose the benefits of using Path operations and break existing Path-specific logic (such as in LocalStorage).

- So to keep taking advantages of using Path with achieving better compatibility , just keep file format to be stored depending on its OS type as it is and handling path conversion only at checks and concatenation points  is better and more safer.

- And one place to fix most of the broken file operations on windows is normalize() since it is almost used at all file operations.

MTarek165:
Ping here @loicmathieu to not forget this PR :)
could you please check my justification here for the last asked change?

loicmathieu:
@brian-mulier-p can you give your advice on this PR.

brian-mulier-p (approved):
No strong opinion, this looks safe to merge, however may I ask why not replacing any "/" check with File.separator when we have a path since we know it will use the filesystem separator? Btw @MTarek165 did you test this PR on Windows to see if it solves the original issue?

MTarek165:
@brian-mulier-p 
- In my opinion that File.Separator ignores malicious inputs if for ex we have \dir\folder/file.txt and say we check startWith(File.Separator) and we are running on linux where file separator is / , it would add / to path becoming /\dir\folder/file.txt returning to the same problem but this time on linux, but regular replacement of \\ with / would guarantee consistency for all the path since it would be now /dir/folder/file.txt so of course the best is considering unix as default and conversions at checks and concatenation places only as here 
- Yes, I tested it on windows it is ok 
- Also notice that the added normalization test is itself a proof as it is OS agnostic as we are saying whenever we see format having \\ it should deal with correctly which is the case for windows paths so we are like simulating if it is running on windows and of course on the old version it would fail even it was run on linux

brian-mulier-p:
Merged, ty for your contribution. Due to the nature of that change and how impactful it can be, I'll wait for a week before backporting it to releases (so it will not be in the Tuesday's one but the one after)

PR #13786, opened 2025-12-22
fix(concurrency): make decrement and pop atomic to prevent stuck queued executions

### LINKED ISSUE #13785: Flows with concurrency set, "queue" indefinitely and do not execute
### Describe the issue

Subflow includes:

```
id: backup_influxdb2_bucket
namespace: home.nejmantowicz.com

concurrency:
  limit: 1
```

Parent Flow includes:

```
  - id: iterate_buckets
    disabled: false
    type: io.kestra.plugin.core.flow.ForEachItem
    items: "{{ outputs.json_to_jsonl.uri }}"
    namespace: home.nejmantowicz.com
    flowId: backup_influxdb2_bucket
    wait: true # wait for the subflow execution
    transmitFailed: true # fail the task run if the subflow execution fails
    inputs:
      host: "{{ inputs.host }}"
      username: "{{ inputs.username}}"
      vmid: "{{ inputs.vmid}}"
      folder: "{{ inputs.folder }}"
      bucket: "{{ json(read(taskrun.items)) }}" # special variable that contains the items of the batch
      timestamp: "{{ outputs.utc_timestamp.values.timestamp }}"
```

Image

Everything worked fine prior to version 1.1.10

If I remove the concurrency lines, the flow executes.

### Environment

- Kestra Version: 1.1.10

omairqazi29:
## Summary
- Fixes race condition where executions with concurrency settings get stuck in QUEUED state indefinitely
- Makes the decrement and pop operations atomic within a single transaction

## Problem
The protection mechanism added in v1.1.10 (commit 71f605898f) has a race condition:

1. When an execution terminates, it decrements the concurrency counter (Transaction 1)
2. Between decrement and pop, another execution can start and increment the counter
3. The limit check `newLimit < limit` then fails because the counter was incremented by the new execution
4. The queued execution never gets popped and remains stuck indefinitely

## Solution
Introduce `decrementAndPop()` method that performs:
1. Decrement the counter
2. Check if newLimit < limit  
3. Pop the next queued execution and increment

All within the same database transaction, ensuring:
- The counter value read during limit check is consistent
- No race between decrement and new execution start
- Queued executions are always properly dequeued when capacity is available

## Changes
- `AbstractJdbcConcurrencyLimitStorage.java`: Add `decrementAndPop()` method for atomic operation
- `AbstractJdbcExecutionQueuedStorage.java`: Add `popWithContext()` method to use existing transaction
- `JdbcExecutor.java`: Use the new atomic `decrementAndPop()` for QUEUE behavior

## Test plan
- [ ] Existing concurrency tests should pass
- [ ] `flowConcurrencyQueuedProtection` test continues to work (protection still applies within atomic transaction)
- [ ] ForEachItem with subflow concurrency no longer gets stuck

Fixes #13785

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/runner/AbstractJdbcConcurrencyLimitStorage.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/runner/AbstractJdbcConcurrencyLimitStorage.java
@@ -4,6 +4,7 @@ import io.kestra.core.models.flows.FlowInterface;
 import io.kestra.core.runners.ConcurrencyLimit;
 import io.kestra.core.runners.ExecutionRunning;
 import io.kestra.jdbc.repository.AbstractJdbcRepository;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.tuple.Pair;
 import org.jooq.*;
 import org.jooq.exception.DataAccessException;
@@ -12,8 +13,10 @@ import org.jooq.impl.DSL;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 
+@Slf4j
 public class AbstractJdbcConcurrencyLimitStorage extends AbstractJdbcRepository {
     protected io.kestra.jdbc.AbstractJdbcRepository<ConcurrencyLimit> jdbcRepository;
 
@@ -87,6 +90,50 @@ public class AbstractJdbcConcurrencyLimitStorage extends AbstractJdbcRepository 
                         return newLimit;
                     }
                 ).orElse(0);
+            });
+    }
+
+    /**
+     * Atomically decrement the concurrency limit counter and pop a queued execution if available.
+     * This ensures the decrement, limit check, and pop all happen within the same transaction,
+     * preventing race conditions that could leave executions stuck in queue indefinitely.
+     *
+     * @param flow the flow to decrement the counter for
+     * @param executionQueuedStorage the storage to pop from
+     * @param consumer the consumer to call with the popped execution (only called if pop succeeds and limit allows)
+     */
+    public void decrementAndPop(FlowInterface flow, AbstractJdbcExecutionQueuedStorage executionQueuedStorage,
+                                BiConsumer<DSLContext, io.kestra.core.models.executions.Execution> consumer) {
+        this.jdbcRepository
+            .getDslContextWrapper()
+            .transaction(configuration -> {
+                var dslContext = DSL.using(configuration);
+
+                // Decrement the counter
+                int newLimit = fetchOne(dslContext, flow).map(
+                    concurrencyLimit -> {
+                        int decremented = concurrencyLimit.getRunning() == 0 ? 0 : concurrencyLimit.getRunning() - 1;
+                        update(dslContext, concurrencyLimit.withRunning(decremented));
+                        return decremented;
+                    }
+                ).orElse(0);
+
+                // Only pop if we're below the limit
+                if (newLimit < flow.getConcurrency().getLimit()) {
+                    executionQueuedStorage.pop(
+                        dslContext,
+                        flow.getTenantId(),
+                        flow.getNamespace(),
+                        flow.getId(),
+                        (ctx, queued) -> {
+                            // Increment the counter for the newly running execution
+                            increment(ctx, flow);
+                            // Call the consumer
+                            consumer.accept(ctx, queued);
+                        }
+                    );
+                } else {
+                    log.error("Concurrency limit reached for flow {}.{} after decrementing the execution running count. No new executions will be dequeued.", flow.getNamespace(), flow.getId());                }
             });
     }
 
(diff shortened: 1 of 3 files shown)


jdbc/src/main/java/io/kestra/jdbc/runner/AbstractJdbcConcurrencyLimitStorage.java
```diff
@@ -90,6 +91,49 @@ public int decrement(FlowInterface flow) {
...
+                int newLimit = fetchOne(dslContext, flow).map(
+                    concurrencyLimit -> {
+                        int decremented = concurrencyLimit.getRunning() == 0 ? 0 : concurrencyLimit.getRunning() - 1;
+                        update(dslContext, concurrencyLimit.withRunning(decremented));
+                        return decremented;
+                    }
+                ).orElse(0);
+
+                // Only pop if we're below the limit
+                if (newLimit < flow.getConcurrency().getLimit()) {
```
vamsi172323, line 122: In the earlier code, if this check allows the current thread to proceed incorrectly (for example, because the counter was incremented by a different thread), it could result in more concurrent executions than the configured limit. In that case, instead of a single concurrent job, we might observe two or more running at the same time.

The issue description mentions that all executions remain in a queued state. Could you please share more details about how this behavior / fix was tested or verified?

jdbc/src/main/java/io/kestra/jdbc/runner/AbstractJdbcExecutionQueuedStorage.java
loicmathieu, line 1: `pop` is only used in a single place so please update the existing method instead of creating a new one

jdbc/src/main/java/io/kestra/jdbc/runner/AbstractJdbcConcurrencyLimitStorage.java
loicmathieu, line 1: `decrement` is only used in a single place so please update the existing method instead of creating a new one.

jdbc/src/main/java/io/kestra/jdbc/runner/AbstractJdbcConcurrencyLimitStorage.java
```diff
@@ -90,6 +91,49 @@ public int decrement(FlowInterface flow) {
...
+    /**
+     * Atomically decrement the concurrency limit counter and pop a queued execution if available.
+     * This ensures the decrement, limit check, and pop all happen within the same transaction,
+     * preventing race conditions that could leave executions stuck in queue indefinitely.
+     *
+     * @param flow the flow to decrement the counter for
+     * @param executionQueuedStorage the storage to pop from
+     * @param consumer the consumer to call with the popped execution (only called if pop succeeds and limit allows)
+     */
+    public void decrementAndPop(FlowInterface flow, AbstractJdbcExecutionQueuedStorage executionQueuedStorage,
```
loicmathieu, line 105: I'm no fan of passing the executionQueuedStorage here, the best would be to pass a Consumer<DSLContext> and let the executor pop the queued execution inside the consumer.

loicmathieu (changes requested):
Overall it makes sense, this may also explain #13744 so the other PR may not be necessary

MilosPaunovic:
Hey @loicmathieu,

The original author of the issue this PR is trying to solve said it's already working https://github.com/kestra-io/kestra/issues/13785#issuecomment-3688573633. Should we close this PR or does it still make sense to have it worked on and merged?

loicmathieu:
@MilosPaunovic this PR fixes a potential race so it should be done.

loicmathieu:
@omairqazi29 would you proceed with the changed asked on the review or should I take it  over and do them by myself?

vamsi172323:
@loicmathieu, please assign this to me if needed. I can raise a PR.

loicmathieu:
@vamsi172323 there is already a PR, if the original author cannot finish it or didn't respond I'll take it over.

vamsi172323:
> @vamsi172323 there is already a PR, if the original author cannot finish it or didn't respond I'll take it over.

@loicmathieu , Sure, I understand 👍

Just a heads up: I have a couple of unit tests in a [cancelled PR](https://github.com/kestra-io/kestra/pull/13764/files) that are relevant and should still be useful for this work:

If you take this over, would you mind extracting those tests and including them in the new PR? I think they’d fit well with the changes. Thanks!

loicmathieu:
@vamsi172323 yes, good idea. Can you refer to your PR here so we linked the two?

vamsi172323:
> @vamsi172323 yes, good idea. Can you refer to your PR here so we linked the two?

https://github.com/kestra-io/kestra/pull/13764/files

loicmathieu:
@omairqazi29 I took over this PR and made the necessary adjustments, thanks for your contribution.
@vamsi172323 I added your test also, it passes ;)

loicmathieu (approved):
LGTM, thanks a lot

loicmathieu:
BAckported to 1.0 and 1.2

PR #13802, opened 2025-12-22
feat(core): enable label-based metrics

yuri1969:
### ✨ Description

This PR is more like a PoC/RFC - does such approach make sense, is possible cardinality explosion a problem, etc.

It attempts to solve the problem of monitoring executions of flow which executes data in different contexts conveyed by labels - e.g. `country: Germany`, `country: Italy`.

Therefore users could list label keys which they want to monitor. Execution-related metrics would become tagged by those labels.

### 🛠️ Backend Checklist

Tests including a PAUSED task seems to hang on my machine with both this branch and develop...

### DIFF
--- a/core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
+++ b/core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
@@ -1,5 +1,6 @@
 package io.kestra.core.metrics;
 
+import io.kestra.core.models.Label;
 import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.executions.ExecutionKilled;
 import io.kestra.core.models.tasks.Task;
@@ -13,6 +14,11 @@ import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.ArrayUtils;
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
+
+import java.util.List;
+import java.util.stream.Stream;
 
 @Singleton
 @Slf4j
@@ -145,6 +151,8 @@ public class MetricRegistry {
     public static final String TAG_QUEUE_CONSUMER = "consumer";
     public static final String TAG_QUEUE_CONSUMER_GROUP = "consumer_group";
     public static final String TAG_QUEUE_TYPE = "queue_type";
+    public static final String TAG_LABEL_PREFIX = "label";
+    public static final String TAG_LABEL_PLACEHOLDER = "__none__";
 
     @Inject
     private MeterRegistry meterRegistry;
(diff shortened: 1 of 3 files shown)


loicmathieu (commented):
If I understand it correctly, you add the labels as tags if they are configured. So it should not have any impact if not configured (so no metric cardinality issue).

I totally understand the rational behind this, it can be useful, we'll discuss internaly whether we want to proceed with this PR or not before I'll make a more careful PR.

core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
```diff
@@ -429,5 +441,24 @@ public void bind(MeterBinder meterBinder) {
...
+                        "%s_%s".formatted(TAG_LABEL_PREFIX, labelKey),
+                        labels.stream()
+                            .filter(label -> labelKey.equals(label.key()))
+                            .map(Label::value)
+                            .findFirst()
+                            .orElse(TAG_LABEL_PLACEHOLDER)
+                    )
+                )
+                .toArray(String[]::new);
+    }
```
loicmathieu, line 484: Metrics tags are re-computed very often, so they are in the hot path.
I think it should be improved with performance in mind, even if it would lead to less elegant implementation:
- Avoid String format, replace it with concatenation
- Avoid the replacement placeholder: if no value, no tag should be added

loicmathieu (changes requested):
We discussed internally and agreed to integrate this change.

core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
yuri1969: The method got optimized by using lower-level abstractions. An ad-hoc JMH benchmark of `getLabelTags()` implementations - both configured with 3 keys and an execution having 8 execution labels finished as:

Old

```text
  1.354 ±(99.9%) 0.062 us/op [Average]
  (min, avg, max) = (1.270, 1.354, 1.623), stdev = 0.083
  CI (99.9%): [1.292, 1.416] (assumes normal distribution)
```

New

```text
  0.154 ±(99.9%) 0.005 us/op [Average]
  (min, avg, max) = (0.143, 0.154, 0.167), stdev = 0.007
  CI (99.9%): [0.149, 0.159] (assumes normal distribution)
```

---

As for the placeholder, sadly, this seems to be the way Micrometer handles situations when a tag might not be always present. I've linked the docs.

loicmathieu: OK so basically what it means is that when you configure metricLabel, they will always be present in the metrics. Understood.

loicmathieu (approved):
LGTM

loicmathieu:
@aj-emerich this could be a good thing to add this new config option to our documentation https://kestra.io/docs/configuration#metrics

This will be part of 1.2

aj-emerich:
@loicmathieu @yuri1969 Can you provide a quick example configuration, and I can add it to our release branch

loicmathieu:
```yaml
kestra:
  metrics:
    labels: ["country", "bu"]
```

This will add for each metrics `label_country` and `label_bu` tags with the value of the tags comming from the value of the corresponding labels of the execution. If the execution didn't have that label the value of the tag will be `__none__`.

aj-emerich:
Got it, added

Issue #13837, opened 2025-12-25
[Bug] ForEach uses iteration value as output key instead of index


proddata:
### Describe the issue

Inside a ForEach, task outputs are keyed by the stringified iteration value rather than by an index or stable identifier.
This makes it hard (or impossible) to access outputs in a predictable way when iterating over objects.

```yaml
id: test
namespace: debug


tasks:

  - id: task0
    type: io.kestra.plugin.core.output.OutputValues
    values:
      x: 1
  - id: log_task0
    type: io.kestra.plugin.core.log.Log
    message: "{{outputs.task0}}"

  - id: foreach
    type: io.kestra.plugin.core.flow.ForEach
    values: [{"x":1}]
    tasks:
      
      - id: task1
        type: io.kestra.plugin.core.output.OutputValues
        values:
          x: 3

      - id: log_task1
        type: io.kestra.plugin.core.log.Log
        message: "{{outputs.task1[taskrun.value]}}"
      
      - id: log_task2
        type: io.kestra.plugin.core.log.Log
        message: "{{outputs.task1}}"
```

log_task0 output
```
{"values":{"x":1}}
```

log_task1 output
```
{"values":{"x":1}}
```
This suggests that outputs.task1[taskrun.value] resolves to the original loop value ({"x":1}), not to the result of task1.

log_task2 output
```
{"{\"x\":1}":{"values":{"x":3}}}
```

This shows that:
	- outputs.task1 is a map
	- keyed by the JSON stringified iteration value ("{\"x\":1}")
	- containing the real task output ({"values":{"x":3}})

### Environment

- Kestra Version: 1.1.10

MTarek165:
There is already a PR https://github.com/kestra-io/kestra/pull/12566 around this issue that adds support for using index to access outputs from iterations but waits for performance testing

proddata:
> There is already a PR https://github.com/kestra-io/kestra/pull/12566 around this issue that adds support for using index to access outputs from iterations but waits for performance testing

Thanks, I saw that. However this seems really like a bug, as even within the same iteration there is no ~~reliable~~ sensible way to access a previous tasks values. Having the complete object (i.e. `value`) as key, seems really just wrong, as this can be quite big. There is already a `taskrun.iteration`, which rather should be used.

This also currently breaks in the UI

<img width="1113" height="518" alt="Image" src="https://github.com/user-attachments/assets/139bcc0b-052e-46dd-84c6-e0531dfce4de" />

[...]

PR #13843, opened 2025-12-26
fix(core): foreach duplicate values should be preserved

MTarek165:
Related to https://github.com/kestra-io/kestra/issues/13837.

## Description
this PR preserves duplicates for foreach values


loicmathieu:
@MTarek165 this would be a breaking change so I must validate first internally.

loicmathieu:
@MTarek165 thanks for your contribution but this has other implication so we will probably integrate that in a broader refactoring about ForEach loop.

PR #13846, opened 2025-12-27
refactor(core): use List.getLast() for lastAttempt()

MTarek165:
## Description
- Replace stream-based reduction with Java 21 List.getLast() for clearer and more efficient access to the last attempt.
- This avoids unnecessary iteration by using constant-time access, which is beneficial when many retries are present.
- Add defensive check on attempts list before calling lastAttempt() in nextRetryDate()


loicmathieu (approved):
LGTM

PR #13898, opened 2025-12-30
chore(logs): make start/stop messages clearer

yuri1969:
### ✨ Description

Server start and stop messages were not completely correct or clear.

### 🛠️ Backend Checklist

### DIFF
--- a/cli/src/main/java/io/kestra/cli/AbstractCommand.java
+++ b/cli/src/main/java/io/kestra/cli/AbstractCommand.java
@@ -207,14 +207,13 @@ public abstract class AbstractCommand implements Callable<Integer> {
                     try {
                         endpoint = UriBuilder.of(server.getURL().toURI())
                             .port(this.endpointConfiguration.getPort().get())
-                            .path("/health")
                             .build();
                     } catch (URISyntaxException e) {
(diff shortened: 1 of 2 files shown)


cli/src/main/java/io/kestra/cli/AbstractCommand.java
```diff
@@ -207,14 +207,13 @@ private void startWebserver() {
                     try {
                         endpoint = UriBuilder.of(server.getURL().toURI())
                             .port(this.endpointConfiguration.getPort().get())
-                            .path("/health")
```
loicmathieu, line 210: This was specifically asked by @tchiotludo so he can click on the log from the console and it opens the health endpoint so I think he may be sad if you remove it (but I agree, it's strange).

loicmathieu (changes requested):

cli/src/main/java/io/kestra/cli/AbstractCommand.java
yuri1969: That's understandable. I've separated the health endpoint to a separate log message.

loicmathieu (approved):
LGTM, thanks

Issue #9789, opened 2025-07-01
Allow deleting a given revision or all revisions older than the selected one


anna-geller:
### Feature description

![Image](https://github.com/user-attachments/assets/b58e15f7-d55d-4d21-b716-36a07bf1c7f3)

CRITICAL FIRST STEP TO SHIP: allow deleting a single (botched) revision e.g. after the user has leaked their secret there

loicmathieu:
This may be challending to do that safely as it means running executions may crash the executor, or replaying an execution based on this revision may crash the executor.

It would also not work on Kafka which keeps all revisions in its topic.

It's a design principle to keep all revision of the flow.

anna-geller:
This was a decision made in the meeting that we'll need to add the feature. Keep in mind that nobody asks the user to delete their revisions, it's sth they can do if strictly needed (e.g. 100 bad revisions created when iterating on fixing a bug or even worse: having some revision that contains credentials).

Still, agree we can add a warning in the confirmation box notifying about possible risks when deleting a revision 👍 thx for the hint!

@Nico-Kestra when working on a design screen, let's add a message in the confirmation box: "Are you sure you want to delete XX previous revisions? Some features, such as replaying executions using those revisions, will no longer be feasible."

anna-geller:
Design is ready: https://www.figma.com/design/ew0uXk0NRXJ2NBBJTNe2n1/01_UI?node-id=9150-49639&t=rie9tujMHZNfOt3f-4

<img width="3840" height="2199" alt="Image" src="https://github.com/user-attachments/assets/f1d3001c-d119-4142-8634-0f7db6cfa6b0" />

<img width="3840" height="2198" alt="Image" src="https://github.com/user-attachments/assets/b5b285d2-2f7e-4aaf-b557-afe62504c8f0" />

anna-geller:
The same design from above will need to be applied to https://github.com/kestra-io/kestra-ee/issues/3666

anna-geller:
- [ ] Backend needs to revisit multiple places in the Executor when it can't find the deleted revision
- [ ] On the frontend, we need to adjust endpoint listing revisions as in some places we loop over available revisions -- related to @brian-mulier-p's work with Metadata versioning for KV Store 
[...]

PR #13902, opened 2025-12-30
fix(executor): avoid infinite loop when flow not found by changing ex…

nkwiatkowski:
- fix #https://github.com/kestra-io/kestra/issues/9789


loicmathieu (approved):
LGTM

PR #13926, opened 2025-12-31
fix: fallback on string when facing json exception in set kv value

Skraye:
As the key value can be anything, when it cannot be parsed as JSON just store it as String

close kestra-io/kestra-ee#6074


nkwiatkowski (approved):
LGTM

PR #8522, opened 2025-04-24
feat(ui): enable variable filter refresh interval

yuri1969:
### What changes are being made and why?

Let users pick longer/shorter automatic refresh interval than the default 10 sec.

These new interval values have been selected rather arbitrary.

PR #13937, opened 2025-12-31
chore(ui): Make the Refresh button a proper one, with a dropown for auto-refresh

proddata:
### ✨ Description
The Refresh action is a core interaction for dashboards and data tables, yet it is currently shown as a small tertiary text button, which makes it feel secondary and easy to miss. Auto-refresh is also buried in settings, even though it is closely related.

This change also removes the redundant pencil icon, since the filter itself is already clickable and opens the edit state.

Finally, the refresh icon is replaced with one that is more typical for data and dashboarding tools, reducing the risk of it being confused with a “reset” action.

#### Before

image
image

#### After
image
image

### 🎨 Frontend Checklist

_If this PR does not include any frontend changes, delete this entire section._

In line with e.g.

**Lightdash**

image

**Grafana**

image


Piyush-r-bhaskar:
@proddata I would require @Nico-Kestra look for the design first on this because I don't like the refresh from dropdown just to show a checkbox checked for only one value.

MilosPaunovic:
I agree with Piyush here, it would make sense for @Nico-Kestra to have a look at this one from the design perspective.

Nico-Kestra:
Would we only have a single value? If so, a simple hover tooltip would be enough.
If there are multiple values, the Lightdash version works very well for me. And no checkbox, because that would imply we can select several.

Piyush-r-bhaskar:
Nope @Nico-Kestra 

For context we had the hardcoded values in past versions.

If I understand correctly, @proddata pitched for list of harcoded values same as grafana. But then our concern is we would not be able to define any values for interval of refresh like change of 1s, 2s , and this was one of changes for a customer/ Yuri https://github.com/kestra-io/kestra/pull/8522


like currently defining from settings:
image

proddata:
> Would we only have a single value? If so, a simple hover tooltip would be enough.

We could also support multiple values like [off, "5s", "10s", "30s"].

> And no checkbox, because that would imply we can select several.

For single value imho it is fine (as it acts like a switch), but agree otherwise.

> If I understand correctly, @proddata pitched for list of harcoded values same as grafana. But then our concern is we would not be able to define any values for interval of refresh like change of 1s, 2s , and this was one of changes for a customer/ Yuri https://github.com/kestra-io/kestra/pull/8522

That said, all the examples in the PR (and what Yuri asked for) look like the standard way this kind of refresh control is usually designed:

image
image

I also wasn’t aware that this would live on the user “Settings” page. I hadn’t considered that this would be configurable there. That page already feels like it’s carrying a lot of different concerns, so adding more UI-level behavior there might make it harder to reason about what applies where. It’s called “Settings”, but it’s accessed via the user menu, and it’s not always obvious which settings are user-specific versus more global, especially with things like “Main configuration” in the mix. That can make it harder to understand what a change will actually affect. IMHO we should avoid adding anything there by all means, and rather see how we can get rid of it (cc: @Ben8t)

That said, this could still work as a default refresh interval. Although one could even argue that users shouldn’t need to define a global refresh interval for all tables (or even for the refresh button itself). For views that are mostly static — like flows, secrets, apps, or tests — a more natural UX might be for the server to push updates (for example via webhooks) instead of relying on periodic polling.

MilosPaunovic:
So do we have any decision on how we want to proceed here @proddata?

Do we go with this change, what do we do with the item in Settings page?

proddata:
@MilosPaunovic I still think the change is something to consider. However I will close it for now and reopen when a complete  (and agreed upon) solution is there

PR #13948, opened 2026-01-01
fix(core): add missing rendered labels for worker/schedule trigger run context

### LINKED ISSUE #13807: IllegalVariableEvaluationException thrown when labels are referenced within triggers
### Describe the issue

IllegalVariableEvaluationException is thrown when label is used within trigger

### Steps to reproduce

- Create the flow as following

```
id: label-trigger
namespace: company.team

labels:
  - key: source
    value: flows
  - key: destination
    value: source

inputs:
  - id: upsert_query
    type: STRING
    defaults: SELECT * from flows
tasks:
  - id: fetch_source
    type: io.kestra.plugin.jdbc.mysql.Query
    url: jdbc:mysql://host.docker.internal:3306/kestra?useSSL=false&allowPublicKeyRetrieval=true
    username: "kestra"
    password: "k3str4"
    sql: "SELECT * FROM {{labels.source}}"
    fetchType: STORE 

triggers:
  - id: watch_table
    type: io.kestra.plugin.jdbc.mysql.Trigger
    url: jdbc:mysql://host.docker.internal:3306/kestra?useSSL=false&allowPublicKeyRetrieval=true
    username: "kestra"
    password: "k3str4"
    sql: "SELECT 1 FROM {{labels.source}}"
    fetchType: FETCH
```

- Wait for the trigger to start the flow

### What is observed
The labels is not found and IllegalVariableEvaluationException is thrown

```
2025-12-23 11:34:45.880[watch_table](http://localhost:8080/ui/dev-ee/executions/company.team/label-trigger/watch_table)watch_table
io.kestra.core.exceptions.IllegalVariableEvaluationException: Unable to find `labels` used in the expression `SELECT 1 FROM {{labels.source}}` at line 1
	at io.kestra.core.runners.VariableRenderer.properPebbleException(VariableRenderer.java:48)
	at io.kestra.core.runners.VariableRenderer.renderOnce(VariableRenderer.java:108)
	at io.kestra.core.runners.VariableRenderer.render(VariableRenderer.java:79)
	at io.kestra.core.runners.VariableRenderer.render(VariableRenderer.java:64)
	at io.kestra.core.runners.VariableRenderer.render(VariableRenderer.java:56)
	at io.kestra.core.runners.DefaultRunContext.render(DefaultRunContext.java:260)
	at [io.kestra.core.models.property.Property.as](http://io.kestra.core.models.property.property.as/)(Property.java:176)
	at [io.kestra.core.models.property.Property.as](http://io.kestra.core.models.property.property.as/)(Property.java:164)
	at io.kestra.core.runners.RunContextProperty.lambda$as$0(RunContextProperty.java:75)
	at io.kestra.core.utils.Rethrow.lambda$throwFunction$4(Rethrow.java:89)
	at java.base/java.util.Optional.map(Unknown Source)
	at [io.kestra.core.runners.RunContextProperty.as](http://io.kestra.core.runners.runcontextproperty.as/)(RunContextProperty.java:75)
	at io.kestra.plugin.jdbc.AbstractJdbcTrigger.evaluate(AbstractJdbcTrigger.java:86)
	at io.kestra.worker.WorkerTriggerCallable.doCall(WorkerTriggerCallable.java:27)
	at io.kestra.worker.AbstractWorkerCallable.call(AbstractWorkerCallable.java:64)
	at io.kestra.worker.WorkerSecurityService.callInSecurityContext(WorkerSecurityService.java:10)
	at io.kestra.worker.DefaultWorker.lambda$callJob$19(DefaultWorker.java:980)
	at io.kestra.core.trace.NoopTracer.inCurrentContext(NoopTracer.java:15)
	at io.kestra.worker.DefaultWorker.callJob(DefaultWorker.java:976)
	at io.kestra.worker.DefaultWorker.lambda$handleTrigger$16(DefaultWorker.java:605)
	at io.micrometer.core.instrument.composite.CompositeTimer.record(CompositeTimer.java:141)
	at io.kestra.worker.DefaultWorker.handleTrigger(DefaultWorker.java:583)
	at io.kestra.worker.DefaultWorker.lambda$run$8(DefaultWorker.java:296)
	at io.micrometer.core.instrument.internal.TimedRunnable.run(TimedRunnable.java:49)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
```

### What is expected

labels are correctly resolved to the desired key/value pair and the sql query is executed. 

### Workaround

Use [KV_Store](https://kestra.io/docs/concepts/kv-store) key

```
triggers:
  - id: watch_table
    type: io.kestra.plugin.jdbc.mysql.Trigger
    url: jdbc:mysql://host.docker.internal:3306/kestra?useSSL=false&allowPublicKeyRetrieval=true
    username: "kestra"
    password: "k3str4"
    sql: "SELECT 1 FROM {{ kv('my_key') }}"
    fetchType: FETCH
```

### Environment

- Kestra Version: 1.1.5, 1.1.10

MTarek165:
closes #13807
## Problem
- When trigger starts rendering its properties during evaluation (for ex : ScheduleTrigger renders inputs, HttpTrigger renders responseCondition, AbstractJdbcTrigger renders sql,...) and these properties use labels, evaluation fails because at the time reaching there, trigger run context has no labels variables loaded yet 
## Solution
- Labels must be loaded earlier than evaluation phase to avoid this issue 
- The most suitable place to load them is in RunContextInitializer before evaluation and in this place we attach more variables to run context and this place is centric for all trigger types (scheduled and worker triggers) too, also we can't add labels directly into the first time creating trigger run context because we need to render these labels 
## Testing
There are two tests covering both Worker Trigger and Schedule Trigger:
- Added a test for Http Trigger as it is the same type for AbstractJdbcTrigger (stated in issue), same path for worker triggers and should execution is created with labels
- Added another test covering schedule triggers also as the same problem occurs 
## Notes
-  I found that the same problem occur with other label types such as system labels, trigger labels and backfill labels for the same reason, now the fix covers all labels


MTarek165:
This is a better comprehensive example for the issue, I would add a test tomorrow
`id: label-test
namespace: company.team
inputs:
  - id: user
    type: STRING
labels:
  - key: source
    value: flows
  - key: destination
    value: source
tasks:
  - id: hello
    message: Hello from {{inputs.user}}
triggers:
  - id: schedule
    cron: "*/3 * * * *"
    inputs:
         user: "system labels with trigger: {{labels.source}}, {{labels.destination}}, {{labels.tr_label}}, {{labels.system.from}}, {{labels.system.correlationId}}"
    labels:
      - key: tr_label
        value: tr_value
`

MTarek165:
@loicmathieu I wonder also if we can create new run context of new execution after adding labels and use it for rendering inputs, this is a simpler fix but I feel that it is not clean as there is no way to inject RunContextFactory into ScheduledExecutionFactory without making some undesired changes in evaluate() method

MTarek165:
@loicmathieu I have finished my idea for this PR, could you review it whenever you have some time?

core/src/main/java/io/kestra/core/runners/RunContextInitializer.java
```diff
@@ -206,6 +210,20 @@ public DefaultRunContext forScheduler(final DefaultRunContext runContext,
         final Map<String, Object> variables = new HashMap<>(runContext.getVariables());
         variables.put(RunVariables.SECRET_CONSUMER_VARIABLE_NAME, (Consumer<String>) runContextLogger::usedSecret);
 
+        /**
+         * Adding labels into Run Context earlier is important for rendering properties of worker and scheduled triggers that uses labels during rendering
+         */
+            Map<String, Object> labels = (Map<String, Object>) variables.get("labels");
```
loicmathieu, line 216: This seems dubious, the labels should have been added to the runcontext before it is sent to the Worker so you shoud not have to do it here.

loicmathieu (changes requested):
Overall, this PR seems to have a lot of changes when the original issue point to only a single place where we miss to pass the flow label so I would need some time to check the original issue and if the proposed fix is what should be done or not.

MTarek165:
> Overall, this PR seems to have a lot of changes when the original issue point to only a single place where we miss to pass the flow label so I would need some time to check the original issue and if the proposed fix is what should be done or not.

Yes, I detected the same issue for the other label types, all have the same root cause, you can try it and evaluate the solution carefully.

core/src/main/java/io/kestra/core/runners/RunContextInitializer.java
MTarek165: I referred to that in description on why I choose this place,  rendering very early when creating runcontext for the first time can't may be done as we need to render the labels itself also , so you need at least the fundamental added variables to use at rendering time, label rendering depends on the previous added variables (around 6 variables) added at first time, this is what was happening at first when rendering labels and this is ok, the main change  that the PR care about is loading labels earlier than the other properties used for triggers.

MTarek165: I was doing that at first but as we need to take care of labels rendering for trigger and backfill also I changed that, I have found that RunContextInitializer is the right place after trying these changes before.

MTarek165: Adding labels earlier is not perfect as labels for triggers and backfills cannot be added too early because they require rendering, and rendering itself depends on the RunContext. When the RunContext is created for the first time, label rendering is not possible yet, since it depends on a set of fundamental variables (around 6 variables) that must be added beforehand. We can do that if we actually wasn't rendering labels but I don't want to alter an existing behavior just focusing on the main change.

This behavior already existed and is expected. What this PR changes is only the order in which labels are loaded: labels are now loaded earlier than other trigger-related properties rendering, without changing how or when label rendering actually works.

I initially tried loading labels even earlier as you ask, but once trigger and backfill label rendering was taken into account, that approach didn’t work correctly. After experimenting with those options, RunContextInitializer proved to be the right place, as it guarantees that the required base variables are present before label rendering happens and of course adding labels earlier before rendering the other trigger-related properties, hence successful rendering for those properties when using labels which is the main goal for the PR.

loicmathieu:
I'm still not convinced by the proposed changes and would need more time to investiguate the issue and check if there is a more straighforward path for resolving it.

MTarek165:
- The issue focuses on using flow labels only in AbstractJdbc (worker) triggers.
If that were the only case, labels could be added before sending to the worker, during the initial RunContext creation, although at this phase execution wasn't created yet to add the labels but we can still add them when trigger != null and execution == null.

- However, the same problem occurs with other label types when they are used in triggers, not only flow labels or worker triggers.

- Therefore, I’m proposing a general solution that works for all label types, for both worker and schedule triggers.

- The core constraint is label rendering: trigger and backfill labels cannot be rendered during the initial RunContext creation because the required variables are not available yet. Adding all labels at that stage will not work.

- In this context, RunContextInitializer seems like an acceptable place to handle this, since it already enriches the RunContext with additional variables.
  
- This approach has been validated by testing all scenarios. Anyway, I’m open to better suggestions if there’s a cleaner or more appropriate alternative.

- Another suggestion: I want to say that may be label rendering itself is not that important  because in UI it is not allowed to use pebble expressions with labels, it is assumed to be static ones, but they are rendered at backend, which makes this hassle, if this is not really needed as it is not actually allowed to be exposed to user so why not get rid of that and then we can simplify labels addition to initial run context as you ask here.

MTarek165:
- The issue focuses on using flow labels only in AbstractJdbc (worker) triggers.
If that were the only case, labels could be added before sending to the worker, during the initial RunContext creation, although at this phase execution wasn't created yet to add the labels but we can still add them when trigger != null and execution == null.

- However, the same problem occurs with other label types when they are used in triggers, not only flow labels or worker triggers.

- Therefore, I’m proposing a general solution that works for all label types, for both worker and schedule triggers.

- The core constraint is label rendering: trigger and backfill labels cannot be rendered during the initial RunContext creation because the required variables are not available yet so adding all labels at this stage will not work, that is the reason that prevents me to add labels there.

- In this context, RunContextInitializer seems like an acceptable place to handle this, since it already enriches the RunContext with additional variables and there we can render the labels based on the previously added variables to run context (which was the same behavior for label rendering before the change, I am centralizing it only beside solving the issue).

-  RunContextInitializer acts like in the middle place as we can't add labels too early because of label rendering and also we must add labels before rendering the other trigger properties.
  
- For these reasons, I see that this approach makes sense and also it has been validated by testing all scenarios. Anyway, I’m open to better suggestions if there’s a cleaner or more appropriate alternative.

johnkm516:
Hello, any update to this issue?

MTarek165:
I am waiting for @loicmathieu approval, I am confident of the solution, I have shared more details about the proposal at my last comment addressing the point of inconvenience, could you give a second look when you have some free time? thanks.

PR #14001, opened 2026-01-06
fix(executions): make sure BulkErrorResponse violations are returned

AcevedoR:
this will help to debug failing tests, otherwise we only have invalid bulk restart in the resulting http exception

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/responses/BulkErrorResponse.java
+++ b/webserver/src/main/java/io/kestra/webserver/responses/BulkErrorResponse.java
@@ -1,13 +1,30 @@
 package io.kestra.webserver.responses;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import io.kestra.core.models.validations.ManualConstraintViolation;
+import io.kestra.core.serializers.JacksonMapper;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.experimental.SuperBuilder;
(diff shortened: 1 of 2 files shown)


webserver/src/main/java/io/kestra/webserver/responses/BulkErrorResponse.java
```diff
@@ -1,13 +1,30 @@
...
 @SuperBuilder
 @Getter
 @NoArgsConstructor
+@Slf4j
 public class BulkErrorResponse {
     String message;
-    Object invalids;
+    Set<ManualConstraintViolation<String>> invalids;
+
+    public String getMessage(){
```
loicmathieu, line 21: I'm not fan of overriding getMessage, this is used automatically for logs.
I'm not fan neither to have logs inside a getter...

If the idea is to be able to access the set of violations inside a test, I'll prefer that you add a getter for the invalids with the `@VisibleForTesting` annotation

loicmathieu (changes requested):

webserver/src/main/java/io/kestra/webserver/responses/BulkErrorResponse.java
AcevedoR: Yeah you are right this is ugly, we should just handle it properly with regular serialization.
The goal is also to return violations is the API response for API users, I'll redo it later

PR #14002, opened 2026-01-06
fix: deleting files should work and close editor

### LINKED ISSUE #13788: [Namespaces Files] Deleted files remain open in editor tabs
### Describe the issue

When a file is deleted from the namespace using, for example, a Namespacefile purge (see Jam), while that file is open in the editor, it remains visible as an open tab after deletion. The tab should be removed automatically. Otherwise, it causes confusion, as the file no longer exists but still appears to be open.

Jam: https://jam.dev/c/b61d580f-bfa7-450a-8a20-818e757e5bcc

### Expected Behaviour
When a file is deleted, its editor tab should be automatically removed so only existing files are shown.

### Environment

- Image hash and time: 04246ac `Sat, Dec 20, 2025 12:09 AM`
- Edition: Open Source & Enterprise

elevatebart:
closes #13788


MilosPaunovic (approved):
Apart from this warning showing up in the console when the delete confirmation dialog pops up, all good.

image

PR #14015, opened 2026-01-06
feature(UI): Turn Jump too.. into modal style command palette

### LINKED ISSUE #13869: [UI/UX] "Jump to..." jumps around and is distracting
### Feature description

The “Jump To” menu is a very prominent UI element, but in practice it’s mainly useful when accessed via keyboard shortcuts. Otherwise, it largely duplicates the existing navigation. It also shifts position on every page change because it’s right-aligned, which makes it feel visually unstable.

### Suggestions:

~~Option A: Move the menu to the center and match it to the surrounding background so it blends in more naturally and feels less dominant.~~
e.g. 

Image

✅  Option B: Turn “Jump to…” into a modal-style command palette (similar to GitHub). This fits better with how it’s primarily used (via keyboard) and avoids it competing with the main navigation.

Image

Both options would make the UI feel calmer and more intentional, while still keeping fast navigation easily accessible.

Opinions @Nico-Kestra , @Ben8t , @MilosPaunovic

mahadevaperuka:
### ✨ Description
Covert the GlobalSearchbar.vue to modal style command palette (similar to GitHub).

### 🔗 Related Issue
Closes #13869 

### 🎨 Frontend Checklist

https://github.com/user-attachments/assets/107addb1-709c-4991-913e-01cdfef016e0

image

image

### DIFF
--- a/ui/src/components/layout/GlobalSearch.vue
+++ b/ui/src/components/layout/GlobalSearch.vue
@@ -167,6 +167,10 @@
     }
 
     const navItems = computed(() => {
+        if (!isOpen.value) {
+            return [];
+        }
+
         const root = scopeStack.value.length > 0 ? scopeStack.value[scopeStack.value.length - 1].items : menu.value;
         const order = {value: 0};
         // When scoped, we treat the scope root as depth 0 for ordering.
--- a/ui/src/components/layout/GlobalSearch.vue
+++ b/ui/src/components/layout/GlobalSearch.vue
@@ -142,8 +142,7 @@
         left: 0;
         width: 100vw;
         height: 100vh;
-        background: rgba(0, 0, 0, 0.4);
-        backdrop-filter: blur(3px);
+        background: var(--el-overlay-color-lighter);
         z-index: 10000;
         display: flex;
         justify-content: center;
@@ -167,7 +166,6 @@
                 border-radius: var(--el-input-border-radius);
                 
                 input {
-                    height: 48px;
                     color: var(--ks-content-primary);
                 }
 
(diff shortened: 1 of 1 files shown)


ui/src/components/layout/GlobalSearch.vue
```diff
@@ -92,21 +135,56 @@
...
-            }
+    .search-overlay {
+        position: fixed;
+        top: 0;
+        left: 0;
+        width: 100vw;
+        height: 100vh;
+        background: rgba(0, 0, 0, 0.4);
+        backdrop-filter: blur(3px);
+        z-index: 10000;
```
proddata: Currently we don't use blur filters for modal background, needs to be decided by @Nico-Kestra if we should align with other modals.

proddata:
@mahadevaperuka

Thanks a lot so far — I got some good feedback from Nico.

We should change the modal background to be consistent with other modal backgrounds (for example when executing a flow) and remove the background blur for now.

For example:
image

Nico also suggested unifying the search field and the results into a single container, similar to GitHub’s command palette:

image

If you can take a look at that as well, that would be great — otherwise we can follow up on it later.

---

@elevatebart will look at it from a technical PoV

mahadevaperuka:
I will try to implement the github-based approach, but it would require a lot of changes to the file. Many of the built-in features provided by el-autocomplete and the popper would need to be reimplemented manually to achieve the same behavior.

I’m not completely sure about it at the moment, so I’ll continue experimenting and see if I can make it work. If I manage to get a good solution, I’ll open a PR.

proddata:
@mahadevaperuka I added a commit to implemented the comments. I assumed since you rerequested my review without any changes, you would not implement them.

Current State (aligned with Nico):

https://github.com/user-attachments/assets/0ac8e5bd-3801-43eb-8ee7-3d84ab2394d0

elevatebart (approved):
Me likes !!!! ❤️

PR #14016, opened 2026-01-06
fix(system): MySQL fulltext search didn't work for suffix

loicmathieu:
This was because the search term was changed to `+term*` which mandates  the term to be a prefix, it is now `*term*` which work for prefix also.

But it has the caveats that if multiple terms are used, the terms are no longuer mandatory so the result may be less accurate in this case, this is a tradeoff.


Skraye (approved):
LGTM

PR #14037, opened 2026-01-07
Chore/micronaut 4 10

### LINKED ISSUE #9743: Rare issue in HTTP multi-part parser

loicmathieu:
Fixes #9743 
Closes #13815

### DIFF
--- a/platform/build.gradle
+++ b/platform/build.gradle
@@ -26,11 +26,13 @@ dependencies {
     def jugVersion = "5.2.0"
     def langchain4jVersion = "1.11.0"
     def langchain4jCommunityVersion = "1.11.0-beta19"
+    def jettyVersion = "12.0.31" // Wiremock uses an older version as Micronaut
 
     // as Jackson is in the Micronaut BOM, to force its version we need to use enforcedPlatform but it didn't really work, see later :(
     api enforcedPlatform("com.fasterxml.jackson:jackson-bom:$jacksonVersion")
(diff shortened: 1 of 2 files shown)


loicmathieu:
Blocked by an upstream issue: https://github.com/micronaut-projects/micronaut-core/issues/12312

AcevedoR (changes requested):
LGTM but e2e tests must pass, maybe it was fixed in develop branch

PR #14039, opened 2026-01-07
fix(core): regression on execution delay (#14023)

### LINKED ISSUE #14023: Delayed execution not being executed in times
### Describe the issue

The following flow should iterate until iteration 4; but it get stuck at first iteration
```
id: quetzal_592226
namespace: company.team

tasks:
  - id: loop
    type: io.kestra.plugin.core.flow.LoopUntil
    condition: "{{ outputs.return.value == '4' }}"
    tasks:
      - id: return
        format: "{{ outputs.loop.iterationCount }}"
```

### Environment

- Kestra Version: develop-ee `f580a0f Wed, Dec 31, 2025 1:45 PM`

fhussonnois:
Fixes: #14023


loicmathieu (approved):
LGTM, it's annoying it cannot be reproduced in test (it should have failed before) but it's obviously corrrect to reduce the fix to only where it was an issue: in MySQL.

AcevedoR (changes requested):
I know it's not easy, but would it be possible to add test around this bug, so we can be sure it works for various cases ?

AcevedoR (approved):
lgtm, urgent fix, we can see for tests later today

PR #14044, opened 2026-01-07
chore(flow): deprecate core tasks that have replacements inside the K…

loicmathieu:
…estra plugin

Part-of: https://github.com/kestra-io/kestra-ee/issues/4228

@Malaydewangan09 I think those 3 have been merged inside plugin-kestra and would be available in 1.2 so we can deprecate them here.
I don't think other would have the time to make it.


Malaydewangan09:
Yes, I will release new plugin version with those 3 tasks, `Fetch` can be covered once logs endpoint is avaialable.

Malaydewangan09 (approved):

PR #14061, opened 2026-01-08
fix: focus cursor on key field when adding variable in no-code editor

### LINKED ISSUE #14008: [No-code editor] Adds variable with cursor in value first, causing null in YAML until key is set
### Describe the issue

When adding a variable in the no-code editor, the cursor starts in the Value field instead of the Key field. If I start typing a value, the flow editor shows it as null because the key is not set yet. To make the value show correctly, I have to go back and fill in the key first. The cursor should start in the Key field so users can set the key before typing the value.

Jam: https://jam.dev/c/eb2acfad-3aa8-4b32-9039-952ef2276872

### Expected Behaviour

When adding a variable, the cursor should start in the Key field. The value should show correctly once the key is set, so users don’t end up typing values that appear as null.

### Environment

- Image hash and time: cbc4690 `Tue, Jan 6, 2026 12:48 PM`
- Edition: Open Source

anuj123upadhyay:
---

### ✨ Description

What does this PR change?  
This PR fixes the cursor focus behavior when adding new variables in the no-code editor. Previously, when clicking "Add a new value", the cursor would automatically focus on the Value field instead of the Key field, causing confusion as users would type values that appeared as null because no key was set yet.

After this change, the cursor now correctly starts in the Key field, allowing users to naturally enter the key first before the value.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/14008

### 📝 Additional Notes
Technical Implementation:

Uses nextTick() with a 100ms setTimeout to ensure Element Plus components are fully mounted before focusing
The delay is necessary because Element Plus components have their own internal mounting lifecycle


MilosPaunovic (commented):
@anuj123upadhyay, could you share a quick screen recording showing the current behavior?

anuj123upadhyay:
https://github.com/user-attachments/assets/c10b029a-20e7-4293-b524-820eb1527a5b

MilosPaunovic:
@elevatebart Could you give this a look when you get a chance? 🙏🏻

elevatebart (approved):

Issue #7132, opened 2025-01-31
Improve error messags of the `flow namespace update` CLI command


yuri1969:
### Feature description

The `flow namespace update <namespace> <directory>` CLI command is often used to update tens or even hundreds of flows at once. This means the referenced directory would hold tends or hundreds of YAML files.

When there is an error encountered during an upload it makes the command fail. Such error might be caused by one or more invalid flow definitions being uploaded.

---

Currently there is no detailed information specifying which particular flow(s) caused the error. This could be resolved by showing either the path to the YAML file(s) or flows's `namespace.id`.

karelkryda:
Hi,
unfortunately we repeatedly encounter this problem when investigating the cause of failed flows deployment.
Are there any plans to improve logging for this process?

Thanks in advance

nkwiatkowski:
Hi, it is not easy to add detail to the flow that is not valid, because of the bulk operation.
But we do have a CLI to validate flows from a folder, with good error handling:

flow validate /my/folder

Result:
✓ - tutorial.business-processes
✓ - company.team.ticket_938
✓ - tutorial.microservices-and-apis
✓ - tutorial.hello-world
✓ - tutorial.data-engineering-pipeline
✓ - tutorial.business-automation
✓ - tutorial.infrastructure-automation
✓ - io.kestra.tests.sleep
✓ - tutorial.dwh-and-analytics
✘ - /home/nicolas/Downloads/flows_brokens/tutorial.infrastructure-automation.yaml
	Unable to parse flowwithsources due to the following error:
	- jakarta.validation.ConstraintViolationException: Invalid type: io.kestra.plugin.core.log.Loooog

karelkryda:
Hi @nkwiatkowski,
While testing the command you mentioned, we encountered a bug with incorrect identification of the problematic file. I opened a PR that fixes this bug.

We are currently on Kestra v0.22, would it be possible to backport the PR to this version as well?

Thank you

loicmathieu:
@yuri1969 @karelkryda I merged both https://github.com/kestra-io/kestra/pull/14078 and https://github.com/kestra-io/kestra/pull/14130 so I'm closing this one as they should provide a better way to validate flows with per-file error messages.

I'm OK to backport it down to 1.0 but 0.22 would need more effort and may be technically challenging.

PR #14078, opened 2026-01-08
Replace YAML string body with multipart in namespace update

karelkryda:
### ✨ Description

Hey,

I took a look at the `flow namespace update` CLI command and would like to discuss possible improvements with you.
Previously it read all flow files, concatenated them into one big YAML multi-document string, and sent that as the request body.
It seemed to me that this wasn't exactly the ideal way to provide data to the backend from the CLI.

So, I moved most of the logic from CLI to the backend. Now the CLI just builds a multipart body and sends individual YAML files instead of the concatenated string. The backend has a new multipart endpoint that reads each file separately and can replace namespaces when the override query parameter is passed from CLI - this logic has also been moved from the CLI to the backend, as the CLI no longer manipulates YAML files in any way.
From my point of view, this is also better for cases where someone would like to use only the API without the CLI for their work - now all the logic happens on the backend, which seems like the right approach to me.

Everything else works basically the same, but this opens up possibilities for better validation and error reporting back to the CLI, which is something I'd like to work on next - it is kind of related to #7132.
This is kind of a "showcase PR" so I can discuss with you guys possibilities with this approach and show you how I'm thinking about it.

If you're cool with this approach, I'd like to use the same pattern for `flow validate` and other commands where it makes sense.
I would also like to use these changes as a basis for improving how validation or parsing errors are reported back to the CLI, thus enabling better observability.

**I mention this potential breaking change:**
CLI performed flows expansion using a long-deprecated function (`IncludeHelperExpander`) as one of its steps. I have now completely removed this step because, in my opinion, it is no longer relevant, but it can of course be implemented in the backend.

As for the API, the old YAML string endpoint still works btw.

What do you think about these changes? Do they make sense to you?

### 🔗 Related Issue

Kind of relates to #7132.

### 🛠️ Backend Checklist

### 📝 Additional Notes

I manually tested the changes and everything seems to be working. I simulated scenarios with both valid and invalid flows.

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
@@ -62,7 +62,6 @@ import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 
 import java.io.IOException;
-import java.nio.charset.Charset;
 import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
@@ -58,9 +58,11 @@ import jakarta.validation.constraints.Min;
 import jakarta.validation.constraints.NotEmpty;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
+import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 
 import java.io.IOException;
+import java.nio.charset.Charset;
 import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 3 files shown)


webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
```diff
@@ -329,6 +331,38 @@ public List<FlowInterface> updateFlowsInNamespace(
...
+        @Parameter(description = "If missing flows should be deleted") @QueryValue(defaultValue = "true") Boolean delete
+    ) throws ConstraintViolationException, IOException {
+        List<CompletedFileUpload> flows = Flux.from(flowsPublisher)
+            .collectList()
+            .blockOptional()
+            .orElse(Collections.emptyList());
+
+        List<GenericFlow> genericFlows = new ArrayList<>();
+        for (CompletedFileUpload flow : flows) {
+            String source = new String(flow.getBytes(), Charset.defaultCharset()).trim();
```
loicmathieu:
```suggestion
            String source = new String(flow.getBytes()).trim();
```

We're storing UTF-8 in the database

loicmathieu (approved):
LGMT, I'll merge the proposed changed before mering it.

I'll open a followup issue to check if you should deprecate the old endpoint or not, I don't know if it's used by the UI.

loicmathieu:
@karelkryda can you add a test?

karelkryda:
@loicmathieu sure, will do

karelkryda:
@loicmathieu added cases for:
 - single flow file update
 - multiple flow files update
 - update with namespace different in URL vs in files **- throws exception**
 - update with namespace different in URL vs in files but with override query parameter set **- works without issues**

loicmathieu (approved):
LGTM, thanks

loicmathieu:
Backported to 1.0 and 1.2

PR #14082, opened 2026-01-09
fix(core): worker trigger emits execution for failed evaluation

### LINKED ISSUE #14081: worker trigger fail silently during failed evaluation
### Describe the issue

Image

1- Save this workflow in image, then go Triggers and enable this trigger.
2- Try running it again but this time make repsonse condition as 
responseCondition: "{{ response.statusCode == 300 }}"

## Expected
1- The first should succeed and trigger the flow successfully
2- The second evaluates successfully but wouldn't trigger the flow
## Actual
- Both wouldn't trigger the flow without any execution logs

The first should succeed but it fails because of a bug when using labels with triggers https://github.com/kestra-io/kestra/issues/13807

but regardless of that, the problem is that when worker trigger fails during evaluation, it doesn't emit failed execution which makes confusion for user, the confusion is that this behavior doesn't distinguish between successful evaluation with unsatisfied condition and failed evaluation due to exception occur 

so at the first case the user would be confused as the expression is true but it acts like an evaluation with unsatisfied condition but actually it is a failure!

### Environment

- Kestra Version: develop

MTarek165:
closes #14081
## Description
- we should distinguish between successful evaluation with unsatisfied conditions and evaluation failure for worker triggers
- In DefaultWorker.handleTriggerError(), failOnTriggerError should be true to allow creating non nullable execution and send with worker trigger result, hence it would be shown in execution logs at UI
## Testing
- run workflow with illegal evaluation expression in http trigger and assert on execution creation with failed state

### DIFF
--- a/core/src/main/java/io/kestra/core/models/triggers/AbstractTrigger.java
+++ b/core/src/main/java/io/kestra/core/models/triggers/AbstractTrigger.java
@@ -82,7 +82,7 @@ abstract public class AbstractTrigger implements TriggerInterface {
 
     @Builder.Default
     @PluginProperty(hidden = true, group = PluginProperty.CORE_GROUP)
-    private boolean failOnTriggerError = false;
+    private boolean failOnTriggerError = true;
 
     @PluginProperty(group = PluginProperty.CORE_GROUP)
     @Schema(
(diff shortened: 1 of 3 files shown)


loicmathieu:
Closing it as changing this default is not something we agreed on.

Issue #14092, opened 2026-01-09
MySQL full text do a prefix search only


loicmathieu:
### Describe the issue

MySQL full text do a prefix search only.
The attempt to fix it https://github.com/kestra-io/kestra/pull/14040 shows too much false positive.

### Environment

- Kestra Version: develop

PR #14093, opened 2026-01-09
fix(repositories): use term query instead of prefix for mysql

nkwiatkowski:
- fix #https://github.com/kestra-io/kestra/issues/14092


loicmathieu (changes requested):
Full text query do a little more than just a like.
So maybe the best would be to combine the two, do a MATCH query as before and do an OR with a LIKE query for prefix.

loicmathieu (approved):
LGTM, cool

PR #14094, opened 2026-01-09
chore(system): small code and perf improvements

loicmathieu:


jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
```diff
@@ -45,9 +44,15 @@
 
 @Slf4j
 public abstract class JdbcQueue<T> implements QueueInterface<T> {
-    private static final int MAX_ASYNC_THREADS = Runtime.getRuntime().availableProcessors();
     protected static final ObjectMapper MAPPER = JdbcMapper.of();
 
+    private static final int MAX_ASYNC_THREADS = Runtime.getRuntime().availableProcessors();
+    private static final Field<Object> KEY_FIELD = AbstractJdbcRepository.field("key");
```
fhussonnois, line 51: `KEY_FIELD`, and `VALUE_FIELD` could be in `AbstractJdbcRepository` because they are common to all tables. Also, I'm sure some repo re-define those static fields.

fhussonnois (changes requested):
It seems that a lot of tests failed ?

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
loicmathieu: Yes, it's defined in a lot of places, agreed it would benefit to be in only one.

loicmathieu: I centralized it and use it anywhere.

fhussonnois (approved):

PR #14098, opened 2026-01-10
feat(core): enable strict parsing for flows imported from files

### LINKED ISSUE #14055: Flow import should apply the same validation as UI/editor
### Feature description

When importing flows, task-level validation does not behave the same way as when writing or editing flows in the UI/editor.

Image

Below could be easily imported without any errors:

```yaml
id: eagle_917999
namespace: base

tasks:
  - id: updateTargetTable
    message: "Test"

triggers:
  - id: dailyEvery1Hour
    cron: "10 */1 * * *"
    recoverMissedSchedules: LAST
    mergeMethod: "{{ now().hour == 0 ? 3 : 2 }}" 👈 -------- this is invalid property
```

MTarek165:
closes #14055
## Problem
- Simply, strict parsing is not enabled when importing flow from file , bypassing strict validations (unknown properties/duplicate keys)  not as like as in editor, this allows saving wrong workflows into database and creating failed executions at runtime and scheduling for triggers even if it has wrong added properties  

## Solution
The two main changes are:
- To prevent importing files with invalid flow yaml we should enable strict parsing when validating flow source as in editor
- Since the current message is not descriptive and not showing why the flow is not imported, we need to show error on UI and for that we should enable failOnError property at frontend to be true so that at backend the condition of catching invalid flow files satisfies and reports the failure

other changes are just aspects around both

## Testing
- Added coverage for two examples one as in the issue with wrong trigger property added and the other one for missed needed property for a plugin to work properly

https://github.com/user-attachments/assets/e10d40c3-5c67-45e7-b4a4-52c6a1559487

### DIFF
--- a/core/src/main/java/io/kestra/core/services/ExecutionService.java
+++ b/core/src/main/java/io/kestra/core/services/ExecutionService.java
@@ -373,7 +373,7 @@ public class ExecutionService {
 
         Execution newExecution = execution.withMetadata(execution.getMetadata().nextAttempt());
 
-        final FlowWithSource flowWithSource = pluginDefaultService.injectVersionDefaults(flow, false , false);
+        final FlowWithSource flowWithSource = pluginDefaultService.injectVersionDefaults(flow, false );
 
         for (String s : taskRunToRestart) {
             TaskRun originalTaskRun = newExecution.findTaskRunByTaskRunId(s);
@@ -617,7 +617,7 @@ public class ExecutionService {
     private Mono<Optional<Task>> getFirstPausedTaskOr(Execution execution, FlowInterface flow){
         return Mono.create(sink -> {
             try {
-                final FlowWithSource flowWithSource = pluginDefaultService.injectVersionDefaults(flow, false, false);
+                final FlowWithSource flowWithSource = pluginDefaultService.injectVersionDefaults(flow, false);
                 var runningTaskRun = execution
                     .findFirstByState(State.Type.PAUSED)
                     .map(throwFunction(task -> flowWithSource.findTaskByTaskId(task.getTaskId())));
--- a/core/src/main/java/io/kestra/core/services/ExecutionService.java
+++ b/core/src/main/java/io/kestra/core/services/ExecutionService.java
@@ -373,7 +373,7 @@ public class ExecutionService {
 
         Execution newExecution = execution.withMetadata(execution.getMetadata().nextAttempt());
 
-        final FlowWithSource flowWithSource = pluginDefaultService.injectVersionDefaults(flow, false);
+        final FlowWithSource flowWithSource = pluginDefaultService.injectVersionDefaults(flow, false , false);
 
         for (String s : taskRunToRestart) {
             TaskRun originalTaskRun = newExecution.findTaskRunByTaskRunId(s);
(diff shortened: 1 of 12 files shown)


core/src/main/java/io/kestra/core/services/PluginDefaultService.java
```diff
@@ -287,11 +287,12 @@ public FlowWithSource injectAllDefaults(final FlowInterface flow, final boolean
      *
      * @param flow the flow to be parsed
      * @param safe whether parsing errors should be handled gracefully
+     * @param strictParsing determine strictness of flow source parsing
      * @return a parsed {@link FlowWithSource}, or a {@link FlowWithException} if parsing fails and {@code safe} is {@code true}
      *
      * @throws FlowProcessingException if an error occurred while processing the flow and {@code safe} is {@code false}.
      */
-    public FlowWithSource injectVersionDefaults(final FlowInterface flow, final boolean safe) throws FlowProcessingException {
+    public FlowWithSource injectVersionDefaults(final FlowInterface flow, final boolean safe, final boolean strictParsing) throws FlowProcessingException {
```
loicmathieu, line 295: `safe=false` means strict parseing, we don't need another boolean here

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/services/PluginDefaultService.java
MTarek165: No, there was some confusion: safe has nothing to do with parsing. It only controls how errors are handled, not how the YAML is parsed.

The execution flow is:

importFlow() → injectVersionDefaults() → parseFlowWithAllDefaults() → YamlParser.parse()

The actual parsing behavior is decided only inside YamlParser.parse() using the strictParsing flag, which determines whether the strict or non-strict ObjectMapper is used.

Currently, injectVersionDefaults() always passes strictParsing = false, so the non-strict mapper is used by default.

Since injectVersionDefaults() is also used in other places, we can’t just change its default behavior.

Solution:
Add a strictParsing parameter to injectVersionDefaults().

- Set strictParsing = true when called from importFlow()

- Keep it false for all other callers to preserve existing behavior

MTarek165: There is some confusion here, just to clarify, safe does not control parsing behavior. It only affects error handling.

Parsing is decided exclusively in YamlParser.parse() via the strictParsing flag, which selects the strict vs non-strict ObjectMapper.

injectVersionDefaults() currently always calls parseFlowWithAllDefaults() with strictParsing = false, so parsing is non-strict by default. Since this method is used in multiple places, changing that default isn’t good

For that reason, a separate strictParsing parameter is required. It’s set to true only when called from importFlow(), and remains false elsewhere to preserve existing behavior.

MTarek165:
@loicmathieu any updates here? it should be ok.

core/src/main/java/io/kestra/core/services/PluginDefaultService.java
loicmathieu, line 295: Can you add an overload instead of adding the new field and remove all non-necessary changes?

loicmathieu (changes requested):

MTarek165:
Closed by mistake 😅

MTarek165:
@loicmathieu I added overloaded method simplifies changes, is there any other changes needed?

core/src/main/java/io/kestra/core/services/ExecutionService.java
```diff
@@ -373,7 +373,7 @@ private Execution markAs(final Execution execution, FlowInterface flow, String t
 
         Execution newExecution = execution.withMetadata(execution.getMetadata().nextAttempt());
 
-        final FlowWithSource flowWithSource = pluginDefaultService.injectVersionDefaults(flow, false);
+        final FlowWithSource flowWithSource = pluginDefaultService.injectVersionDefaults(flow, false );
```
loicmathieu:
```suggestion
        final FlowWithSource flowWithSource = pluginDefaultService.injectVersionDefaults(flow, false);
```

loicmathieu (approved):
LGTM, thanks

loicmathieu:
@MilosPaunovic can you validate the frontend changes before I merge this PR?

MilosPaunovic (approved):
`UI` changes are good to go.

PR #14130, opened 2026-01-12
Replace YAML string body with multipart in flows validate

karelkryda:
### ✨ Description

_I am opening this as a follow-up to PR #14078._

In this PR, an endpoint was added for validating flows using multipart body instead of one large YAML string multidocument.
The CLI has been rewritten to use this endpoint. Thanks to these changes, invalid flows are no longer mapped to the file name using the index number, but instead use the file name information from the multipart body.

The presence of the index has been retained for now, until the same approach is used for templates. After that, it can be completely removed.
Since I think a better approach is to use multiple smaller PRs, a follow-up will be opened.

### 🛠️ Backend Checklist

### 📝 Additional Notes

The validation endpoint using YAML body is used to validate flows from the UI. Therefore cannot be deprecated.

Closes #14022.

### DIFF
--- a/core/src/main/java/io/kestra/core/services/FlowService.java
+++ b/core/src/main/java/io/kestra/core/services/FlowService.java
@@ -19,7 +19,6 @@ import io.kestra.core.runners.RunContextFactory;
 import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.utils.ListUtils;
 import io.kestra.plugin.core.flow.Pause;
-import io.micronaut.http.multipart.CompletedFileUpload;
 import jakarta.inject.Inject;
 import jakarta.inject.Provider;
 import jakarta.inject.Singleton;
@@ -28,7 +27,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.ClassUtils;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 
-import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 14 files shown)


loicmathieu:
@karelkryda if I understand it correctly, we need first to merge https://github.com/kestra-io/kestra/pull/14078 right?

karelkryda:
@loicmathieu, there should be no need.
This PR is based on the development branch, so it does not depend on the previous PR in any way.
It's just a continuation of the work I started in the aforementioned PR.

loicmathieu:
OK, I didn't understood that, I'll review it

core/src/main/java/io/kestra/core/services/FlowService.java
```diff
@@ -203,6 +207,67 @@ public List<ValidateConstraintViolation> validate(final String tenantId, final S
...
     }
 
+    /**
+     * Validates the given flow files.
+     *
+     * @param tenantId  The tenant identifier.
+     * @param flowFiles The YAML files to validate.
+     * @return The list of validation constraint violations.
+     */
+    public List<ValidateConstraintViolation> validate(final String tenantId, final List<CompletedFileUpload> flowFiles) {
```
loicmathieu: It's unfortunate to have an HTTP dependency in the FlowService.
I would prefere to resolves the parts inside the controller.

loicmathieu (changes requested):
We also need a test for the new controller endpoint inside `FlowControllerTest`.

karelkryda:
@loicmathieu, added few tests for different combinations of valid and invalid flows.
To resolve the HTTP dependency, I deduplicated the `validate` function in `FlowService` which now consumes list of custom objects called `FlowSource`. It contains (or may not) filename and flow String content.

This is populated in both endpoints - YAML body and multipart body and passed to the validation function.
Existing tests has been modified and new one for filename population has been added.

All tests passing, hope this approach is fine for you, LMK 😉

loicmathieu (approved):
LGTM, thanks a lot

loicmathieu:
Backported to 1.0 and 1.2

PR #14131, opened 2026-01-12
chore(core): use annotation instead of logger field

yuri1969:
### ✨ Description

Normalize `Slf4` logger initialization to the already used Lombok annotation.

### 🛠️ Backend Checklist


loicmathieu (approved):
I personally prefer constant logger but consistency is also important so I'll sponsor this PR anyway.

PR #14135, opened 2026-01-13
fix: Increase job name random suffix to prevent collisions under high concurrency

aulightzxy-web:
✨ Description
This PR increases the length of the random suffix for Kubernetes job names from 5 to 10 characters. This change significantly reduces the probability of job name collisions, which can occur in high-concurrency scenarios and cause tasks to fail.

The issue was identified in a high-throughput environment where a single workflow containing a foreach task with a parallelism of 60 was executed 100 times in rapid succession. This creates a burst of nearly 6,000 tasks (60 * 100), all attempting to generate a K8s job with a similar prefix. The original 5-character random suffix was not sufficient to guarantee uniqueness, leading to name collisions and task failures.

This fix addresses the problem by expanding the random suffix, thus making collisions statistically improbable.

🔗 Related Issue
N/A

🛠️ Backend Checklist
Code compiles successfully and passes all checks
All unit and integration tests pass

📝 Additional Notes
The Math Behind the Collision Problem (The Birthday Problem)
The probability of a name collision can be estimated using the principles of the "Birthday Problem". The approximate formula for the probability of at least one collision is:

P(k) ≈ 1 - e^(-k^2 / 2N)

Where:

k is the number of items being created (e.g., the number of pods in a concurrent burst).
N is the total number of possible unique names.
Before the Change:

The suffix was 5 alphanumeric, case-insensitive characters (36 possibilities: a-z, 0-9).

N = 36^5 = 60,466,176 (≈ 60 million)
If we take a moderately concurrent burst of k = 1,000 pods:

P(1000) ≈ 1 - e^(-1000^2 / (2 * 60466176)) ≈ 0.82%
With a larger burst, like k = 6,000 pods (from the user scenario):

P(6000) ≈ 1 - e^(-6000^2 / (2 * 60466176)) ≈ 25.7%
A ~26% chance of collision is unacceptably high and explains the observed task failures.

This paradox states that when the number of items in a set increases, the probability of a collision (a shared attribute, like a birthday) rises much more dramatically than intuition suggests！

After the Change:

The suffix is now 10 alphanumeric, case-insensitive characters.

N = 36^10 ≈ 3,656,000,000,000,000 (≈ 3.6 quadrillion)
Even with the same k = 6,000 burst:

P(6000) ≈ 1 - e^(-6000^2 / (2 * 3.656e15)) The probability becomes so small that it is effectively zero, ensuring the stability of task execution in high-concurrency environments.

🤖 AI Authors
Why don't cats play poker in the jungle? Too many cheetahs! 🐱

### DIFF
--- a/core/src/main/java/io/kestra/core/models/tasks/runners/ScriptService.java
+++ b/core/src/main/java/io/kestra/core/models/tasks/runners/ScriptService.java
@@ -234,7 +234,7 @@ public final class ScriptService {
     }
 
     /**
-     * Create a job name like {namespace}-{flowId}-{taskId}-{random} with random being 5 alphanumerical characters.
+     * Create a job name like {namespace}-{flowId}-{taskId}-{random} with random being 10 alphanumerical characters.
      * The Job name will be normalized based on the DNS Subdomain Names (RFC 1123) with a limit of 63 characters as used by Kubernetes
      */
     @SuppressWarnings("unchecked")
@@ -249,12 +249,12 @@ public final class ScriptService {
             task.get("id")
         ));
         String normalized = normalizeValue(name, true, true);
-        if (normalized.length() > 58) {
-            normalized = normalized.substring(0, 57);
+        if (normalized.length() > 53) {
+            normalized = normalized.substring(0, 52);
         }
 
-        // we add a suffix of 5 chars, this should be enough as it's the standard k8s way
-        String suffix = RandomStringUtils.secure().nextAlphanumeric(5).toLowerCase();
+        // we add a suffix of 10 chars, this is safer for high-concurrency scenarios
+        String suffix = RandomStringUtils.secure().nextAlphanumeric(10).toLowerCase();
         return normalized + "-" + suffix;
     }
 }
--- a/core/src/main/java/io/kestra/core/models/tasks/runners/ScriptService.java
+++ b/core/src/main/java/io/kestra/core/models/tasks/runners/ScriptService.java
@@ -234,7 +234,7 @@ public final class ScriptService {
     }
 
     /**
-     * Create a job name like {namespace}-{flowId}-{taskId}-{random} with random being 10 alphanumerical characters.
+     * Create a job name like {namespace}-{flowId}-{taskId}-{random} with random being 8 alphanumerical characters.
      * The Job name will be normalized based on the DNS Subdomain Names (RFC 1123) with a limit of 63 characters as used by Kubernetes
      */
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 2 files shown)


loicmathieu:
Is this fixing a real bug or a theoretical one?
Because Kubernetes uses a 5 alpha suffix like we do so if it's OK for Kubernetes it should be OK for Kestra.

aulightzxy-web:
> 这是在修复一个实际存在的漏洞还是一个理论上的漏洞？ 因为 Kubernetes 和我们一样都使用以 5 开头的字母后缀，所以如果 Kubernetes 可以这样做，那么 Kestra 也应该可以。

In our stress testing and production environments, we encountered this vulnerability. Mathematical calculations have also demonstrated that pod name duplication issues can occur under high concurrency conditions.

aulightzxy-web:
> Is this fixing a real bug or a theoretical one? Because Kubernetes uses a 5 alpha suffix like we do so if it's OK for Kubernetes it should be OK for Kestra.

We are the workflow scheduling team at Xiaomi. Our production environment generates hundreds of thousands of instances daily, each containing a large number of foreach tasks. This results in a massive number of pod names with identical prefixes being created simultaneously.

Here is a real-world example: a single workflow contains foreach tasks with a parallelism of 60, and users submitted over 100 workflows at the same time. This led to more than 6,000 foreach tasks being generated concurrently, all sharing the same naming prefix. As a result, certain pod names ended up being identical, which subsequently caused some instances to fail.

loicmathieu:
@aulightzxy-web OK, maybe doubling the suffix size is not needed as each increment of one char doubles the number of possible combination. Maybe a suffix of 8?

aulightzxy-web:
> @aulightzxy-web好吧，或许不需要将后缀长度翻倍，因为每增加一个字符，可能的组合数量就会翻倍。也许后缀长度为 8 就足够了？

Based on experiments, when prefixes are identical, the average number of strings required for a collision with 8‑digit random numbers exceeds around one million. However, in our tests, the minimum number of strings needed for an 8‑digit random number to collide was fifty thousand. Theoretically, an 8‑digit random string length should be sufficient.

Nevertheless, our production environment runs many critical workflows with high concurrency. We want to avoid workflow failures caused by duplicate names as much as possible.

Therefore, we prefer to use 10‑digit random numbers. If you believe 8 digits are adequate, we can also adjust accordingly.

loicmathieu:
@aulightzxy-web this is tradeoff, as in Kubernetes name lenght are limited to 64, each character added to the suffix limit the number of characters of the meaningful part of the name.

I would prefer to start with 8, if it still collides in production, I would agree to increse it.
But in test, if you do test on a single machine with a single source of entropy, it may collide more than in production in multiple machines.

aulightzxy-web:
> @aulightzxy-web这是一个权衡，因为 Kubernetes 名称长度限制为 64 个字符，添加到后缀的每个字符都会限制名称有意义部分的字符数。
> 
> 我倾向于先从 8 开始，如果生产环境中仍然存在冲突，我会同意增加这个值。 但在测试环境中，如果只在一台机器上使用单一熵源进行测试，其冲突发生的可能性可能比在多台机器的生产环境中更高。

My colleagues and I all agree with the proposal to use an eight-digit random number. The previous ten-digit one was indeed quite long. I have now changed the length of the random number to eight digits to handle high-concurrency situations.

loicmathieu (approved):
We have an agreement ;)

LGTM, thanks

loicmathieu:
Backported to 1.0 and 1.2

PR #14137, opened 2026-01-13
fix: Ensure normalize method reliably strips all trailing special chars

aulightzxy-web:
✨ Description
This PR fixes a bug in the ScriptService.normalize() method where it failed to reliably strip all invalid trailing characters (-, ., _).

The previous implementation used three separate, sequential calls to StringUtils.stripEnd(). This created a bug where a string ending in a combination of these characters, like "my-string-_", would be incorrectly processed. The final result would be "my-string-", which violates the RFC 1123 naming conventions that this method is intended to enforce.

The fix consolidates the three calls into a single, robust call: StringUtils.stripEnd(string, "-._"). This ensures that any combination of these invalid characters is correctly stripped from the end of the string.

The corresponding unit test for the normalize method has also been expanded to include assertions for these specific edge cases, verifying the fix and preventing future regressions.

🔗 Related Issue
N/A

🛠️ Backend Checklist
Code compiles successfully and passes all checks
All unit and integration tests pass

📝 Additional Notes
This is a correctness fix for a subtle bug in a core utility method.

🤖 AI Authors
What do you call a cat that gets everything it wants? Purr-suasive. 🐱

### DIFF
--- a/core/src/test/java/io/kestra/core/models/tasks/runners/ScriptServiceTest.java
+++ b/core/src/test/java/io/kestra/core/models/tasks/runners/ScriptServiceTest.java
@@ -195,9 +195,19 @@ class ScriptServiceTest {
 
     @Test
     void normalize() {
+        // existing tests
         assertThat(ScriptService.normalize(null)).isNull();
         assertThat(ScriptService.normalize("a-normal-string")).isEqualTo("a-normal-string");
-        assertThat(ScriptService.normalize("very.very.very.very.very.very.very.very.very.very.very.very.long.namespace")).isEqualTo("very.very.very.very.very.very.very.very.very.very.very.very.lon");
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/kestra/core/models/tasks/runners/ScriptServiceTest.java
```diff
@@ -195,9 +195,19 @@ void jobName() {
 
     @Test
     void normalize() {
+        // existing tests
```
loicmathieu:
```suggestion
```

This comment is useless

loicmathieu (approved):
LGTM, thanks.

I'll push the suggestion before merging.

loicmathieu:
Backported to 1.0, 1.1 and 1.2

PR #14143, opened 2026-01-13
fix(jdbc) : fix issue where delayed instances bypass concurrency limit checks

gluttonweb:

---

### ✨ Description
Instances with a future scheduleDate are handled via delayed scheduling. If the workflow has concurrency limits, this process currently skips the concurrency check. Consequently, when triggered, the instance transitions directly from CREATED to RUNNING (skipping QUEUED ), which prevents the workflow from effectively enforcing concurrency limits.
image

What does this PR change?  

If a CREATED instance is persisted to ExecutionDelay because its scheduleDate is in the future, it retains the CREATED status. Once the scheduled time is reached, the instance is submitted to the executionQueue consumer—still as CREATED —ensuring that concurrency limit validation is performed.

### 🛠️ Backend Checklist

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
@@ -577,7 +577,7 @@ public class JdbcExecutor implements ExecutorInterface {
                             ExecutionDelay executionDelay = ExecutionDelay.builder()
                                 .executionId(executor.getExecution().getId())
                                 .date(execution.getScheduleDate())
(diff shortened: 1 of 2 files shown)


loicmathieu (changes requested):
Can you add a test for it?

gluttonweb:
> Can you add a test for it?

Okay, I have added unit tests for this changes, and they are executing successfully.

loicmathieu (approved):
LGTM, thanks

loicmathieu:
Backported to 1.0 and 1.2

PR #14145, opened 2026-01-13
fix(executions): Subflow and ForEachItem should transmit the kind

### LINKED ISSUE #14141: Subflow execution type is not PLAYGROUND when called from PLAYGROUND
### Describe the issue

In PLAYGROUND mode, when a flow calls a subflow, the subflow execution is not of type PLAYGROUND. It should match the parent execution type. The same issue may apply to ForEachItem and Unit Test.

Jam: https://jam.dev/c/d0bc3739-09bf-4d8f-9214-11d63e5344e2

Create both of the below flows and execute the `parent_flow` in the PLAYGROUND mode.

```yaml
id: parent_flow
namespace: demo

tasks:
  - id: call_subflow
    namespace: demo
    flowId: subflow
    wait: true

  - id: parent_log
    message: "Parent flow finished"

----

id: subflow
namespace: demo

tasks:
  - id: sub_log
    message: "Hello from subflow"
```

### Expected Behaviour
If a subflow is called in PLAYGROUND mode, the subflow execution should also be PLAYGROUND (same for ForEachItem and Unit Test).

### Environment
- Kestra Version: `develop` | `1.3.0-SNAPSHOT`
- Image Hash and Time: 57b6a8c `Mon, Jan 12, 2026 5:17 PM`

loicmathieu:
Fixes https://github.com/kestra-io/kestra/issues/14141

### DIFF
--- a/core/src/main/java/io/kestra/core/models/executions/Execution.java
+++ b/core/src/main/java/io/kestra/core/models/executions/Execution.java
@@ -157,6 +157,25 @@ public class Execution implements DeletedInterface, TenantInterface {
         final BiFunction<FlowInterface, Execution, Map<String, Object>> inputs,
         final List<Label> labels,
         final Optional<ZonedDateTime> scheduleDate) {
+        return newExecution(flow, inputs, labels, scheduleDate, null);
+    }
+
+    /**
+     * Factory method for constructing a new {@link Execution} object for the given {@link Flow} and
+     * inputs.
+     *
+     * @param flow The Flow.
+     * @param inputs The Flow's inputs.
+     * @param labels The Flow labels.
+     * @param kind The ExecutionKind.
+     *
+     * @return a new {@link Execution}.
+     */
+    public static Execution newExecution(final FlowInterface flow,
+                                         final BiFunction<FlowInterface, Execution, Map<String, Object>> inputs,
+                                         final List<Label> labels,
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/kestra/core/models/executions/Execution.java
```diff
@@ -157,6 +157,25 @@ public static Execution newExecution(final FlowInterface flow,
...
+     * @param inputs The Flow's inputs.
+     * @param labels The Flow labels.
+     * @param kind The ExecutionKind.
+     *
+     * @return a new {@link Execution}.
+     */
+    public static Execution newExecution(final FlowInterface flow,
+                                         final BiFunction<FlowInterface, Execution, Map<String, Object>> inputs,
+                                         final List<Label> labels,
+                                         final Optional<ZonedDateTime> scheduleDate,
```
fhussonnois, line 177: I really don't like the use of `Optional` in the method parameters. I know we do it in other places.

fhussonnois (approved):

core/src/main/java/io/kestra/core/models/executions/Execution.java
loicmathieu: Yeah, it was like that before, I didn't change it...

loicmathieu:
Backported to 1.0 and 1.2

PR #14147, opened 2026-01-13
fix(repositories): timezone issue when searching expired timezone

nkwiatkowski:
Fix a timezone issue when searching for expired windowed multiple condition


loicmathieu (commented):
Without a test it's difficult to know if it would be cross-database or not.

Asking for a review by @fhussonnois as he worked on the same area lately and had to do a fix **only** for MySQL and not the other.

fhussonnois (approved):

PR #14163, opened 2026-01-14
fix(test): try to unflaky WorkerTest.killed()

loicmathieu:
It may not fix the flakiness, if so, the next step would be to mark it as flaky...


AcevedoR (approved):
finger crossed 🤞

PR #14166, opened 2026-01-14
Remove redundant edit icon from filter chips

proddata:
Filter chips were already clickable to open the edit popover, so the pencil icon duplicated that affordance and added visual noise.

- Remove the pencil button and anchor popover positioning to the chip.
- Keep chip height consistent and align the close (x) styling with Saved Filters.
- Apply lint-driven string escaping in the related Storybook test.

**approved by Nico ✅** 

was:
image

is:
image


MilosPaunovic (approved):
LGTM! 🚀

PR #14167, opened 2026-01-14
Added drag-and-drop functionality for file operations in the sidebar.

### LINKED ISSUE #9497: Prevent dragging folders/files away from sidebar
Follow up of https://github.com/kestra-io/kestra/issues/8924.

https://github.com/user-attachments/assets/9c41e975-c47c-443b-9718-e91984a662cd

One of the points from the original issue (https://github.com/kestra-io/kestra/issues/8924) has been resolved, but one item still remains.

When the file explorer tab is open and files/folders are draggable for rearrangement, we need to make it clear that they can only be dropped onto the sidebar - not elsewhere.

The best approach would be to change the cursor to `cursor: not-allowed;` when the dragged item leaves the valid drop target area.

adishchowdhury:
Added drag-and-drop functionality for file operations in the sidebar.

✨ Description
This PR improves the Namespace Files sidebar drag-and-drop UX by showing a not-allowed cursor when a dragged file/folder leaves the sidebar drop zone, making it clear items can only be dropped back onto the sidebar.
It adds a document-level dragover guard that toggles cursor state and sets dataTransfer.dropEffect = "none" outside the sidebar area, and cleans up the state on drag end.

🔗 Related Issue
Fixes https://github.com/kestra-io/kestra/issues/9497
​

🎨 Frontend Checklist
Code builds without errors (npm run build)

All existing E2E tests pass (npm run test:e2e)

🛠️ Backend Checklist
Deleted (no backend changes).

📝 Additional Notes
Change is isolated to the UI component handling Namespace Files drag/drop (FileExplorer.vue).

If needed, the global cursor styles can be moved to a shared/global stylesheet later.


adishchowdhury:
hey kindly review the pr @MilosPaunovic @elevatebart

MilosPaunovic:
Hi @adishchowdhury,

ui/src/components/inputs/FileExplorer.vue
```diff
@@ -457,6 +464,53 @@
...
+
+onMounted(() => {
+    document.addEventListener("dragover", onDocDragOver, {capture: true});
+    document.addEventListener("drop", onDocDropOrEnd, {capture: true});
+});
+
+onBeforeUnmount(() => {
+    document.removeEventListener("dragover", onDocDragOver, {capture: true} as any);
+    document.removeEventListener("drop", onDocDropOrEnd, {capture: true} as any);
+});
```
elevatebart: Why do. you use Pure JavaScript here instead of using VueJs event handling?

If you chose to do it anyway, you should make sure that no lingering handler is assigned when the html element is removed from the DOM. It could lead to memory leaks and unwanted behavior.

I really think using the VueJs `@dragover` and consorts is a better idea but if you have a good reason for doing it, so be it. Please cleanup `beforeUnmount` though.

elevatebart (changes requested):
Hey @adishchowdhury,

Thank you for this proposal. I like your initiative and it seems to work.

Without a bit of an explanation, I don't get why we use `document.addEventListener()` instead of `@dragover`. Is the Vue event system not working as you'd expect to?

ui/src/components/inputs/FileExplorer.vue
adishchowdhury: I’m using pure JS (document.addEventListener) because this behavior needs to be global, not scoped to the FileExplorer component’s DOM: @dragover/@drop in Vue only fires when the pointer is over the element that has the directive, but the drag can leave the sidebar and still be dropped somewhere else in the page.

I used document.addEventListener(..., { capture: true }) intentionally because we need a global guard: when users drag from the sidebar, the cursor can leave the FileExplorer DOM subtree, and in that case @dragover/@drop on the component wouldn’t fire, so the browser default drop behavior could still happen. Using document-level listeners lets us intercept dragover/drop anywhere in the page and cancel the default only when draggingFromSidebar is true.

I also added proper cleanup in onBeforeUnmount() (removing the document listeners with the same capture flag, and resetting draggingFromSidebar + cleanupDragClasses()), so no global handlers/classes remain after the component is destroyed.

elevatebart (changes requested):
Nothing as changed the `document.addEventListener()` calls are still in the script on mount. I don't understand. 

Sorry, am I missing something?

MilosPaunovic:
Hi @adishchowdhury,

As there hasn't been any progress or updates on this, I'll close this PR.

PR #14177, opened 2026-01-14
refactor(system): re-implement MaintenanceService api

fhussonnois:
Adds listener pattern for maintenance mode

Changes:
* Adds utility class Disposable


loicmathieu (approved):
LGTM

PR #14178, opened 2026-01-14
chore(core): use record instead of value class

yuri1969:
### ✨ Description

Modernized & possible perf optimization.

### 🛠️ Backend Checklist

- [x] All unit and integration tests pass (all but `ExecutionControllerRunnerTest.restartExecutionWithNewInputs()`)


core/src/main/java/io/kestra/core/http/HttpResponse.java
```diff
@@ -78,11 +78,10 @@ public static <T> HttpResponse<T> from(ClassicHttpResponse httpResponse, T body,
             .build();
     }
 
-    @Value
     @Builder
```
loicmathieu, line 85: i don't think `@Builder` work for records and anyway it should not be needed right?

core/src/main/java/io/kestra/core/models/executions/Execution.java
```diff
@@ -914,19 +914,15 @@ private FailedTaskRunWithLog lastAttemptsTaskRunForFailedExecution(TaskRun taskR
...
-
-        private TaskRun taskRun;
-        private List<LogEntry> logs;
+    public record FailedTaskRunWithLog(
+        TaskRun taskRun,
+        List<LogEntry> logs) {
     }
 
-    @Value
     @Builder
```
loicmathieu: Again, should no be needed

core/src/main/java/io/kestra/core/http/HttpResponse.java
loicmathieu: I'm afraid this may break a lot of plugins...
Can you rollback this change?

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/http/HttpResponse.java
yuri1969: Rolled the `Status` change back.

loicmathieu (approved):
LGTM, I need to check on EE side if it needs modification before merging.

PR #14179, opened 2026-01-15
fix(core): decrypt secret output when used with on resume for pause task

### LINKED ISSUE #14125: Secrets provided in the `onResume` in `Pause` task are not decrypted.
### Describe the issue

💬 [original Slack discussion](https://kestra-io.slack.com/archives/C03FQKXRK3K/p1767361512756929)

Secrets provided in the `onResume` in `Pause` task are not decrypted:

```yaml
id: create_user
namespace: company.team

inputs:
  - id: secret_input
    type: SECRET

tasks:
  - id: get_user_input    
    onResume:
      - id: secret_pause
        type: SECRET
        required: true

  - id: node_uppercase_secrets
    type: io.kestra.plugin.scripts.node.Script
    script: |
      const secretInput = "{{ inputs.secret_input }}";
      const onResumeSecret = "{{ outputs.get_user_input.onResume.secret_pause }}";

      console.log("Uppercased secret_input: " + secretInput.toUpperCase());
      console.log("Uppercased secret_pause: " + onResumeSecret.toUpperCase());
```

```
Uppercased ******_input: SECRET
Uppercased ******_pause: VZOQRRWW3AVCJQPTMXBMG3UWTPKEWR7C0RXXQRCVYNWNQA==
```

### Environment

- Kestra Version: develop

MTarek165:
closes #14125
## Description
- Output decryption for secrets needs encryption type to verify whether this output should be decrypted or not. When using Pause task the onResume inputs are attached to task outputs directly in Pause class and secret input returns encrypted string value only without type, so after that for resumed execution when we try to decrypt execution outputs, onResume outputs deycryption fail  as it doesn't find the needed type at the map for decryption so it stay encrypted finally.
## Solution
- Since the main reason for non decryption is the missing encryption type then we should use EncryptedString object for secret input instead of plain string, this has two benefits:
  1- For pause task we should return EncryptedString object which has the type , so that when trying to decrypt for execution outputs it would see it hence decrypts the secret ouput solving the issue.
  2- Using EncryptedString directly enriches us from explicit mapping to EncryptedString that is done at FlowInputOutput.typedOutputs()
Overall it is more better and looks more natural than using plain string
## Testing
- The secret output for pause task should be decrypted properly.


loicmathieu (approved):
LGTM, thanks

loicmathieu (changes requested):
Unfortunatly, tests didn't pass.

loicmathieu:
Asking review from our security officer @tchiotludo

MTarek165:
> Unfortunatly, tests didn't pass.

There are 2 tests failing, I would fix today

MTarek165:
@loicmathieu Tests are passing now, with a small addition to RunVariables. Some tests were using decryptVariable as false, so no encryption happen and since secret is object now so when extracting value to pass for log masking it would fail with cast from map to string , so we need to handle it only.

core/src/main/java/io/kestra/core/models/flows/input/SecretInput.java
```diff
@@ -14,16 +15,16 @@
...
         title = "Regular expression validating the value."
     )
     @Regex
     String validator;
 
     @Override
-    public void validate(String input) throws ConstraintViolationException {
-        if (validator != null && !Pattern.matches(validator, input)) {
+    public void validate(EncryptedString input) throws ConstraintViolationException {
+        if (validator != null && !Pattern.matches(validator, input.getValue())) {
```
loicmathieu, line 27: This would not work as the value would be encrypted right?
We need to decrypt it to be able to validate it.

core/src/main/java/io/kestra/core/runners/RunVariables.java
```diff
@@ -302,9 +303,18 @@ public Map<String, Object> build(final RunContextLogger logger, final PropertyCo
                     // if a secret input is used, add it to the list of secrets to mask on the logger
                     if (logger != null && !ListUtils.isEmpty(secretInputs)) {
                         for (String secretInput : secretInputs) {
-                            String secret = (String) inputs.get(secretInput);
-                            if (secret != null) {
-                                logger.usedSecret(secret);
+                            String secret;
+                            Object secretValue = inputs.get(secretInput);
+                            if(secretValue != null) {
```
loicmathieu: Can we do better than getting an object and looking at its type?
As you said it's when decrytpVariables is false so do we have this information here?

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/models/flows/input/SecretInput.java
MTarek165: No it works correctly, this is the intended behavior before my change, I am just wrapping it and pass the value, but it works correctly because it resolves to InputAndValue and their assertions works at tests.

MTarek165: They are validating the encrypted pattern

loicmathieu: Well, it works as before didn't it works correctly ;)
Validating the encrypted pattern is I think not what we would expect here, but as it was already the case let's keep it as is.

core/src/main/java/io/kestra/core/runners/RunVariables.java
MTarek165: So I think we should enable decryptVariables to be true here , this is the true bug as if we used it with secret input and tried to show it in output , it wouldn't be decrypted

core/src/main/java/io/kestra/core/models/flows/input/SecretInput.java
MTarek165: No problem, we may fix it in another PR if it is a true bug.

core/src/main/java/io/kestra/core/runners/RunVariables.java
MTarek165: Indeed, I have updated validation to use decryptVariables, could you check the modified version?

core/src/main/java/io/kestra/core/models/flows/input/SecretInput.java
MTarek165: I observed that secret inputs are weak, they can be of any number of characters, most likely because of this, we aren't validating on the decrypted or the original value.

MTarek165:
Hi @loicmathieu, Is it ok here?

loicmathieu (approved):
LGTM, thanks

loicmathieu:
Backported to 1.0 and 1.2

PR #14185, opened 2026-01-15
fix(execution): improve flow trigger processing perf

loicmathieu:
By initializing a condition context later

Fixes https://github.com/kestra-io/kestra-ee/issues/6401


loicmathieu:
Preliminary tests shows a reduction of allocation in a profile from 70% of allocations to 35% of allocations for the method `FlowTriggerService.computeExecutionsFromFlowTriggerPreconditions` other methods and flow trigger usage should improve but it's complex and need times to do the benchmark.

nkwiatkowski (approved):
LGTM

PR #14186, opened 2026-01-15
Feat/queue v2

loicmathieu:
Improvements and fixes to queue v2 implementation:
- Fail fast on queue consumer
- Manage deps in the platform
- emitAsync()
- Move all interfaces to core as they are needed in a lot of different components including core
- Change factory to a static one as otherwise it's not possible to habe an Event be itself an interface (like we do for ExecutionCommand for ex)

First batch of upgrade from queue v1 to queue v2, other updates will ne done on followup PRs

Preliminary performance test shows no big performace impact (too early to be more precise).


queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcSubscriber.java
```diff
@@ -39,70 +36,72 @@ public JdbcSubscriber(
 
     protected abstract void init();
 
-    public QueueSubscriber<T> subscribe(Rethrow.ConsumerChecked<Either<T, DeserializationException>, Exception> consumer) throws QueueException {
-        this.queueService.execute(throwRunnable(() -> {
+    public QueueSubscriber<T> subscribe(Consumer<Either<T, DeserializationException>> consumer) {
+        this.queueService.execute(() -> {
```
fhussonnois, line 40: Should be preferable to reuse the `JdbcQueuePoller` class here. So we could have dedicated unit tests only on the polling mechanism independently of the QueueSubscriber logic.

fhussonnois (approved):

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcSubscriber.java
loicmathieu: :+1, I added this to the list of TODO on queue v2

PR #14200, opened 2026-01-16
chore(system): deprecate RunnerUtils

loicmathieu:
TestRunnerUtils should be used in test.
The only usage has been replaced by polling the repository


nkwiatkowski (approved):
LGTM

PR #14204, opened 2026-01-18
chore(core): remove redundant Boolean.Type.equals checks on non-nullable values

MTarek165:
## Description
Boolean.TRUE.equals(...) is useful only when the value may be null.
Since orElseThrow() in render().as(...).orElseThrow() throws exception if value is null and otherwise return non-nullable value, so this check is redundant, does not affect behavior, and only adds unnecessary noise / IDE warnings.


loicmathieu:
Hi,
Thanks for your contribution but we use this pattern reliantly in plugins and core and, even if it's right that it resolves to the same if we rely on Lombok, I prefer to be explicit so I don't want to change this code if not necessary.

PR #14210, opened 2026-01-19
feat(jdbc): allow minPollInterval and maxPollInterval to be same

### LINKED ISSUE #9608: [JDBC] Allow setting min and max poll interval to the same value.
### Feature description

This would disable the switch to a different poll interval.

hemantmm:
closes: #9608

### ✨ Description

What does this PR change?  

- Permit configuration where `minPollInterval == maxPollInterval`.
- Updated validation to only reject `max < min`.
- Added unit test verifying single-step behavior.

image


loicmathieu (approved):
LGTM, thanks

PR #14215, opened 2026-01-19
test(system): bring back partial parallelization + load flow with tenant annotation

AcevedoR:
- companion PR: https://github.com/kestra-io/kestra-ee/pull/6454


loicmathieu:
Be sure it takes less time and it works at least 3 times in a row before merging...

nkwiatkowski (approved):
LGTM

loicmathieu:
See test result duration, it didn't bring any performance improvement so I think we should not run tests in parallel. It risks bringing again flakiness for now added value.

PR #14218, opened 2026-01-19
fix(trigger): Expression and ExecutionOutputs should not be scheduled…

loicmathieu:
… condition

ScheduledCondition are special types of conditions used when computing the Schedule trigger next date. If no next date can be computed, the Scheduler will try to compute again the next date after 1s. So ScheduleConditions must be conditions that only works on date to limit the number of date a trigger should be scheduled on.

Without that, it creates a kind of inifinte loop.
In my local machine, this uses 5% CPU whereas with the fix the CPU becomes idle. On resource restricted computers this can cost way more.

Fixes https://github.com/kestra-io/kestra-ee/issues/6444


Skraye (approved):
LGTM

loicmathieu:
Backported to 1.0 and 1.2

PR #14222, opened 2026-01-19
docs(storage): fix examples

wrussell1999:
- [x] Concat
- [x] DeduplicateItems
- [x] Delete
- [x] FilterItems
- [x] PurgeCurrentExecutionFiles
- [x] Reverse
- [x] Size
- [x] Split
- [x] Write


aj-emerich (approved):

core/src/main/java/io/kestra/plugin/core/storage/DeduplicateItems.java
```diff
@@ -38,15 +38,36 @@
...
+                      - "my_data.csv"
+                    
+                  - id: csv_to_ion
+                    type: io.kestra.plugin.serdes.csv.CsvToIon
+                    from: "{{ outputs.generate_files.outputFiles['my_data.csv'] }}"
+
+                  - id: dedup
+                    type: io.kestra.plugin.core.storage.DeduplicateItems
+                    from: "{{ outputs.csv_to_ion.uri }}"
+                    expr: "{{ customer_email }}"
```
wrussell1999, line 69: Should `expr` be renamed? Basically you just want to say which of the columns you want to target for duplicates. But maybe you want to remove some stuff that matches an expression. I couldn't work out what.

core/src/main/java/io/kestra/plugin/core/storage/FilterItems.java
```diff
@@ -35,16 +35,27 @@
...
+                    uri: https://huggingface.co/datasets/kestra/datasets/raw/main/csv/orders.csv
+
+                  - id: csv_to_ion
+                    type: io.kestra.plugin.serdes.csv.CsvToIon
+                    from: "{{ outputs.download.uri }}"
+                  
+                  - id: filter
+                    type: io.kestra.plugin.core.storage.FilterItems
+                    from: "{{ outputs.download.uri }}"
+                    filterCondition: "{{ (product_id | number) == 20 }}"
```
wrussell1999, line 57: Perhaps the property description could be clearer about what expressions, and specifically being able to use the headers directly in here like I have.

core/src/main/java/io/kestra/plugin/core/storage/Split.java
```diff
@@ -29,31 +29,78 @@
...
+
+                tasks:
+                  - id: download
+                    type: io.kestra.plugin.core.http.Download
+                    uri: https://huggingface.co/datasets/kestra/datasets/raw/main/csv/orders.csv
+
+                  - id: split
+                    type: io.kestra.plugin.core.storage.Split
+                    from: "{{ outputs.download.uri }}"
+                    bytes: "5KB"
```
wrussell1999: This only works on KB and above. It didn't recognise B so maybe should add in property description.

core/src/main/java/io/kestra/plugin/core/storage/DeduplicateItems.java
aj-emerich: Probably one for the plugin guys

core/src/main/java/io/kestra/plugin/core/storage/FilterItems.java
aj-emerich: Let's be as descriptive as possible so it's not up to the user to work out what a property is asking for.

core/src/main/java/io/kestra/plugin/core/storage/DeduplicateItems.java
wrussell1999: I can't think of a better name so gonna leave this for now.

core/src/main/java/io/kestra/plugin/core/storage/FilterItems.java
```diff
@@ -66,8 +77,8 @@ public class FilterItems extends Task implements RunnableTask<FilterItems.Output
     private Property<String> from;
 
     @Schema(
-        title = "The 'pebble' expression used to match items to be included or excluded",
-        description = "The 'pebble' expression should return a BOOLEAN value (i.e. `true` or `false`). Values `0`, `-0`, and `\"\"` are interpreted as `false`. " +
+        title = "The expression used to match items to be included or excluded",
+        description = "Headers from the file can be referenced directly e.g. `{{ product_id }}` but will be rendered as a string unless combined with a filter e.g. `product_id | number`. The Pebble expression should return a BOOLEAN value (i.e. `true` or `false`). Values `0`, `-0`, and `\"\"` are interpreted as `false`. " +
```
aj-emerich:
```suggestion
        description = "Headers from the file can be referenced directly, e.g., `{{ product_id }}`, but will be rendered as a string unless combined with a filter, e.g., `product_id | number`. The Pebble expression should return a BOOLEAN value (i.e., `true` or `false`). Values `0`, `-0`, and `\"\"` are interpreted as `false`. " +
```

aj-emerich (approved):

PR #14225, opened 2026-01-20
feat: Separate RUNNING and QUEUED counts for ForEach executions

### LINKED ISSUE #7988: ForEach counts
### Feature description

In "ForEach" executions, and with calls limited with concurrencyLimit, all tasks to be executed are indicated under the "In-progress" count. Would it be possible to separate them by status (Running, Queued) as is already the case with "Created". Ex: I launch ForEach of several thousand or tens of thousands of calls with a concurrencyLimit of 8 or 16. I would like to see those actually in "Running" and those in "Queued" separately

hemantmm:
closes: #7988

### ✨ Description

What does this PR change?  

- Separate RUNNING and QUEUED counts in ForEach outputs.
- Update UI component to display both states.
- Update Storybook example.
- Fix countTaskRunsByState bug.


core/src/main/java/io/kestra/plugin/core/flow/ForEachItem.java
```diff
@@ -654,7 +654,7 @@ public static class Batch implements StorageSplitInterface {
     public static class Output implements io.kestra.core.models.tasks.Output {
         @Schema(
             title = "The counter of iterations for each subflow execution state",
-            description = "This output will be updated in real-time based on the state of subflow executions.\n It will contain one counter by subflow execution state."
```
loicmathieu, line 657: Can you revert this to avoid needing to maintain two list here in doc and in the code.
We didn't list them before so why now?

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -113,8 +113,9 @@ public ExecutionRunning processExecutionRunning(FlowInterface flow, int runningC
                     Logs.logExecution(
                         executionRunning.getExecution(),
                         Level.INFO,
-                        "Execution is queued due to concurrency limit exceeded, {} running(s)",
```
loicmathieu, line 124: Please revert, this seems not linked to the issue and didn't provide any value as you add a 0 in all cases...

loicmathieu (changes requested):
Backend changes should be reverted, they didn't provide any value.
I also asked clarification on the issue as it talks about `ForEach` but this specific display is for `ForEachItem` we should wait before merging.

hemantmm:
Thanks for the review @loicmathieu.

I will:
- Revert the changes in ExecutorService.java as they are not directly related to the issue.
- Revert the modification in ForEachItem.Output documentation to keep it consistent with previous behavior and avoid maintaining duplicate lists.

Regarding the confusion between ForEach and ForEachItem, you are right – the current issue description refers to ForEach, while the display I modified is related to ForEachItem.

I agree that we should wait for clarification on the original issue before proceeding further.

I will push an update shortly with the backend changes reverted so that the PR only contains relevant modifications.

Thanks for pointing this out.

core/src/main/java/io/kestra/plugin/core/flow/ForEachItem.java
hemantmm: The reason I added this was to make the output schema more discoverable for users, but I agree it's not worth the duplication risk.

I will remove the output description additions and keep the list only in the code where it's actually maintained.

executor/src/main/java/io/kestra/executor/ExecutorService.java
hemantmm: I thought showing "0" would be useful for clarity, but it's just noise when there are no running executions.

loicmathieu:
@hemantmm @MilosPaunovic the issue author confirmed the issue in for `ForEachItem` so we can proceed.
As there are no more backend changes, @MilosPaunovic I'll let you validate and merge ;)

MilosPaunovic (changes requested):
@hemantmm Can you please provide a flow `yaml` you've used to test this?

hemantmm:
Thanks for the review!

Here is the flow YAML that I used to test this change. This flow YAML creates a ForEach with a mix of subflow statuses: SUCCESS, FAILED, RUNNING, and QUEUED.

```
id: foreach-concurrency-counting
namespace: io.kestra.tests

tasks:
  - id: generate_items
    values: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
    concurrencyLimit: 2
    tasks:
      - id: process_item
        message: "Processing item {{ taskrun.value }}"
      
      - id: sleep
        commands:
          - sleep 1

```


```
id: foreach-item-counting-subflow
namespace: io.kestra.tests

inputs:
  - id: item_batch
    type: FILE

tasks:
  - id: process_batch
    message: "Processing batch from {{ inputs.item_batch }}"
  
  - id: simulate_work
    commands:
      - sleep 2

```

ui/src/utils/executionUtils.ts
```diff
@@ -40,6 +40,15 @@ export function findTaskRunsByState(execution: Execution, state: string)  {
     return execution.taskRunList.filter((taskRun) => taskRun.state.current === state);
 }
 
+export function countTaskRunsByState(execution: Execution): Record<string, number> {
```
MilosPaunovic, line 43: @hemantmm Where is this used?

MilosPaunovic (changes requested):
Can you also provide a screenshot or a screen recording of how this is working now compared to what was not working as expected before?

MilosPaunovic:
Hey @hemantmm, are there any updates on this?

hemantmm:
Hey @MilosPaunovic, I'm currently blocked by an unexpected issue, due to which I am unable to provide a screenshot.

MilosPaunovic:
Okay, so do we close this then @hemantmm?

hemantmm:
Yeah @MilosPaunovic

PR #14250, opened 2026-01-21
test(tracing): unflaky TracesTest.runningAFlowShouldGenerateTraces

AcevedoR:


loicmathieu (commented):
LGTM, the most important is that there are traces so it works

loicmathieu (approved):

AcevedoR:
requesting rereview because I had to isolate it by tenant otherwise it was still flaky

loicmathieu (approved):
LGTM, if that was the issue, we may now go back to equals(7) but I'll let you decide

nkwiatkowski (approved):
LGTM

PR #14254, opened 2026-01-21
fix(system): honor JDK proxies when installing plugins

loicmathieu:
Fixes https://github.com/kestra-io/kestra-ee/issues/6417


AcevedoR (approved):

loicmathieu:
Backported to 1.0 and 1.2

PR #14255, opened 2026-01-21
fix(executions): subflow restart should reflect parent flow changes

### LINKED ISSUE #13641: Subflow restart causes new execution and keeps running
### Describe the issue

There seems to be issues when restarting a subflow. For reproduction purposes I have two flows; one correct and one incorrect.
```
id: test_subflow_correct
namespace: test

tasks:
  - id: log
    message: "Correct log"
```

```
id: test_subflow_incorrect
namespace: test

tasks:
  - id: log
    message: "{{incorrect.reference}}"
```

Now we have a main flow that first references the incorrect flow:
```
id: test_mainflow
namespace: test

tasks:
  - id: log1
    message: "Started mainflow"
  - id: startSubflow
    namespace: test
    flowId: test_subflow_incorrect
  - id: log2
    message: "Finished mainflow"
```

1. Execute the `test_mainflow`. It will fail.
2. Edit the flow, and change `flowId: test_subflow_incorrect` to `flowId: test_subflow_correct`. Save it.
3. Go back to the execution. Note the execution ID. Press `Restart` -> `Restart latest revision`
4. Notice the `startSubflow` is failed, but keeps running. See the screenshot. This seems to be a bug.
Image

6. Go to the Overview tab. Notice the state is Running. Also notice the new execution ID (and the original execution ID in the bottom).
Image

So it seems there are three bugs here:
- According to the [docs](https://kestra.io/docs/workflow-components/retries#retry-vs-restart-vs-replay:~:text=Restarts%20keep%20the%20same%20execution%20ID.), Restarts keep the same execution ID. Which is not the case here.
- The state stays in Running (instead of Success)
- The `startSubflow` is Failed (instead of Success)

### Environment

- Kestra Version: 1.1.9

MTarek165:
closes #13641
## Description
- There is assumption that always restarting parent execution should restart last executed subflow execution, which is causing trigger for last executed subflow even if there is change in flowId or namespace in subflow task that refers to the child flow in parent execution which shows why even after changing subflow flowId referance it still gets failed state (which is the last state for subflow execution) not giving succeed as we changed the child flow id reference.
- And since the followed execution has revision for subflow task different from the executed one, a running state for the task appears even it is finished and in failed state (the version of the tracked is different than the versio of the executed one)
## Solution
- We should check that current subflow task has the same flow id and namespace for the last executed subflow to confirm they are the same, otherwise we should create a new subflow execution as we detected a change in parent revision.
- Flow revision should be attached to restarted execution before getting flow from repository to get the right flow revision.

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
+++ b/core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
@@ -93,7 +93,13 @@ public final class ExecutableUtils {
                 Optional<Execution> existingSubflowExecution = Optional.empty();
                 if (currentTaskRun.getOutputs() != null && currentTaskRun.getOutputs().containsKey("executionId")) {
                     // we know which execution to restart; this should be the case for Subflow tasks
-                    existingSubflowExecution = executionRepository.findById(currentExecution.getTenantId(), (String) currentTaskRun.getOutputs().get("executionId"));
+                    Optional<Execution> subflowExecution = executionRepository.findById(currentExecution.getTenantId(), (String) currentTaskRun.getOutputs().get("executionId"));
+                    if(subflowExecution.isPresent()
+                        && currentTask.subflowId().flowId().equals(subflowExecution.get().getFlowId())
+                        && currentTask.subflowId().namespace().equals(subflowExecution.get().getNamespace())){
+                        existingSubflowExecution = subflowExecution;
+                    }
+
                 }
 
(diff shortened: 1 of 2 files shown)


loicmathieu:
@MTarek165 sorry but this issue is more broad than what it looks, we are thinking more broadly of restart/replay with subflow and the impact of new revision so this issue is not open for contribution yet as we may go in a different direction.

PR #14260, opened 2026-01-21
feat: purge-current-execution-files-scope

### LINKED ISSUE #7985: PurgeCurrentExecutionFiles Task with childs and/or with states
### Feature description

Concerning the Purge Current Execution Files task, would it be possible to add a parameter to indicate whether or not we wish to purge the child executions of the subflows. In fact, we regularly have flows that call dedicated subflows that we cannot purge until the main flow is finished. The idea is to be sure not to leave unpurged "traces" during the main execution. Along the same lines, we could include the optional purging of states generated by the "io.kestra.plugin.core.state.Set" tasks.

hemantmm:
closes: #7985

### ✨ Description

What does this PR change?  

- This PR extends the `PurgeCurrentExecutionFiles` task to give users explicit control over what execution data is purged, while keeping the current behavior safe by default.

- The existing behavior makes it difficult to safely purge execution files during runtime without risking premature cleanup of subflows or leaving residual execution traces.
 
- This change addresses that limitation while preserving backward compatibility.

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/storage/PurgeCurrentExecutionFiles.java
+++ b/core/src/main/java/io/kestra/plugin/core/storage/PurgeCurrentExecutionFiles.java
@@ -2,6 +2,7 @@ package io.kestra.plugin.core.storage;
 
 import io.kestra.core.models.annotations.Example;
 import io.kestra.core.models.annotations.Plugin;
+import io.kestra.core.models.property.Property;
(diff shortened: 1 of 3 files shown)


loicmathieu:
Closing it as it didn't even compile, it contains test that uses flow that didn't exist, ...
Please validate that code is compiling and new tests are passing before openning a PR

PR #14263, opened 2026-01-22
feat: Abstract the implementation of the emitOnly method into the interface as well

gluttonweb:
---

### ✨ Description

Abstract the emitOnly method into the interface.

Benefits:

1. Avoid type casting: It enables queues like executionQueue to use the emitOnly method directly without requiring explicit type casting.
image
2. Bypass persistence for non-critical messages: It allows using emitOnly instead of emit for messages (such as workerTaskResultQueue ) that do not require persistence via jdbcIndexerInterface.save .

3. Optimize custom Queue implementations (e.g., RocketMQ): In scenarios where the Queue implementation is replaced (e.g., with RocketMQ), using emitOnly directly prevents the worker from initiating redundant MySQL transactions before sending workerTaskResultQueue messages. This avoids potential failures caused by attempting to acquire a JDBC connection when the pool is busy, even though the transaction would execute no actual logic.

### 🛠️ Backend Checklist

### 📝 Additional Notes

### DIFF
--- a/core/src/main/java/io/kestra/core/queues/QueueInterface.java
+++ b/core/src/main/java/io/kestra/core/queues/QueueInterface.java
@@ -27,6 +27,13 @@ public interface QueueInterface<T> extends Closeable, Pauseable {
         emitAsync(null, messages);
     }
 
+    default void emitOnly(T message) throws QueueException {
+        emitOnly(null, message);
+    }
+
+    void emitOnly(String consumerGroup, T message) throws QueueException;
+
+
     void emitAsync(String consumerGroup, List<T> messages) throws QueueException;
 
     default void delete(T message) throws QueueException {
--- a/core/src/main/java/io/kestra/core/queues/QueueInterface.java
+++ b/core/src/main/java/io/kestra/core/queues/QueueInterface.java
@@ -31,7 +31,9 @@ public interface QueueInterface<T> extends Closeable, Pauseable {
         emitOnly(null, message);
     }
 
-    void emitOnly(String consumerGroup, T message) throws QueueException;
+    default void emitOnly(String consumerGroup, T message) throws QueueException {
+        throw new UnsupportedOperationException();
+    }
 
 
     void emitAsync(String consumerGroup, List<T> messages) throws QueueException;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/kestra/core/queues/QueueInterface.java
```diff
@@ -27,6 +27,13 @@ default void emitAsync(List<T> messages) throws QueueException {
         emitAsync(null, messages);
     }
 
+    default void emitOnly(T message) throws QueueException {
+        emitOnly(null, message);
+    }
+
+    void emitOnly(String consumerGroup, T message) throws QueueException;
```
loicmathieu: Can you add a default implementation that throws an `UnsupportedOperationException`?

worker/src/main/java/io/kestra/worker/DefaultWorker.java
loicmathieu, line 1: Are those really needed?
As it's also called by our Kafka Backend, I would prefer not to integrate this change

loicmathieu (changes requested):
We have a Kafka backend which queue share the same interface but didn't implement emitOnly as it didnt't need it.

So I would prefer that this change didn't touch the Worker as it's shared with our kafka Backend.

core/src/main/java/io/kestra/core/queues/QueueInterface.java
gluttonweb: Could you clarify the reason for throwing UnsupportedOperationException ? JdbcQueue already has an implementation that throws QueueException , which is consistent with the other interface methods.

worker/src/main/java/io/kestra/worker/DefaultWorker.java
gluttonweb: So for WorkerTaskResultQueue , does this mean the Kafka backend must also use emit instead of emitOnly ?
In that case, I can revert the changes in DefaultWorker.

core/src/main/java/io/kestra/core/queues/QueueInterface.java
loicmathieu: Because our Kafka queue didn't implement it, and our Kafka backend didn't use it.
So by doing this I would not have to update our Kafka backend.

worker/src/main/java/io/kestra/worker/DefaultWorker.java
loicmathieu: The Kafka backend use the same DefaultWorker with a Kafka backed queue so yes, please revert

gluttonweb: Thanks for the clarification, that makes sense. I've updated the code based on your comments, could you please review it again?

core/src/main/java/io/kestra/core/queues/QueueInterface.java
gluttonweb: Get it

loicmathieu (approved):
LGTM

PR #14271, opened 2026-01-22
feat: purge_current_execution_files_scope

### LINKED ISSUE #7985: PurgeCurrentExecutionFiles Task with childs and/or with states

hemantmm:
closes: #7985

### ✨ Description

What does this PR change?  

✅ Added missing assertion to `shouldPurgeIncludingChildExecutionsAndStates()` test method.
✅ Test now properly validates the output URIs size when `includeChildExecutions` and `includeStates` are enabled.

image


loicmathieu (changes requested):
I cannot see how it can work as you add two new properties but never use them

hemantmm:
Thanks for the feedback! I've removed the unused properties since the underlying `Storage.deleteExecutionFiles()` method doesn't currently support these parameters.

loicmathieu (changes requested):
I think what needs to be done is exactly to add those new properties to `runContext.storage().deleteExecutionFiles()` and add support for them.

This may not be easy for subflow executions, I'll discuss with our product team, hold on for now.

loicmathieu:
@hemantmm do you still plan to work on this issue or should I take it over?

hemantmm:
Hi @loicmathieu, yes, I still plan to continue working on this issue.

hemantmm:
Hi @loicmathieu,

The implementation has been modified to pass the new properties `includeChildExecutions` and `includeStates` to `runContext.storage().deleteExecutionFiles(...)`, which are now supported in the storage layer.

hemantmm:
Let me know if this approach is correct or if you would like to implement it differently.

loicmathieu:
State is deprecated and will be removed soon so it's not that important to support its deletion.
As childExecutionFile is not implemented, this PR seems to not do what needs to be done to cover the issue so I'm closing it for now.

PR #14282, opened 2026-01-22
fix(validation): apply validation on secret input properly and prevent introducing it as item type

### LINKED ISSUE #14280: Secret input problems with validations and log masking
### Describe the issue

There is two issues with secret input:
1- secret input validation is not working as expected, no validation happens on the passed input as discussed here in this PR https://github.com/kestra-io/kestra/pull/14179
To reproduce it try any string not matching the defined validator like 'any' with this workflow:
```
 id: secret_input1
 namespace: secrets
 inputs:
   - id: input1
     type: SECRET
     validator: "(?=.{8,})(?=.*[A-Z])(?=.*[0-9]).*"
 tasks:
   - id: hello
     message: "{{inputs.input1}}"
```

2- the second problem is that secret input was neither masked nor decrypted when used as inner type for Array 
To reproduce it try this workflow with any array input and execute the flow:

```
id: secret_input2
namespace: secrets
inputs:
  - id: input1
    type: ARRAY
    itemType: SECRET
tasks:
  - id: hello
    message: "{{inputs.input1}}" 
```
it would return the full EncryptedString object of type and value with no masking for value.

Solving both is related to each other and this screen shows the second problem

Image

### Environment

- Kestra Version: develop

MTarek165:
closes #14280 

## 1. Secret Input Validation

- Validation currently runs on the **encrypted value**.
- This happens because `parseType()` returns an encrypted value that is passed to the generic validator.
- As a result, validation does **not operate on the original secret value**, making it unreliable.

## 2. Secrets Used as `ItemType` Are Not Masked

- `ItemType` does **not have an ID**.
- Only flow inputs with IDs are registered in  
  `RunContextFactory.secretInputsFromFlow()`.
- When secrets are used inside `Array` or `Multiselect`, they are treated as item types and therefore:
  - Are not decoded
  - Are not masked
  - Appear in logs as plain text

---

## Core Rule

> **Not all `Type`s should be allowed as `ItemType`.**

Using any `Type` as an `ItemType` strips it of schema-level semantics.  
While this is acceptable for simple types (e.g. `INT`), it is **unsafe for `SECRET`**, because:

- Secret item types have **no ID** → no masking or decryption
- Secret item types have **no validator** → treated as plain strings
- Only the class name remains, losing all security guarantees

---

### 1. Validate Secrets Before Encryption

- Validate secrets **inside `parseType()`**, before encryption.
- The original value is still available at this stage.
- This avoids decrypting inside `validate()`, which:
  - Has no access to the secret key
  - Adds unnecessary overhead
  - Breaks generic validation design

**Conclusion:**  
Secret validation must happen **before encryption**.

---

### 2. Disallow Secrets as `ItemType`

The masking and decryption issue exists because `ItemType` has no ID.  
Adding IDs to all item types would be intrusive and semantically incorrect.

**Instead:** disallow `SECRET` from being used as `ItemType`.

#### Why

- **Security:** Secret item types cannot be masked or decrypted because an `id` cannot be defined for them.
- **Validation:** Secret item types do not support validators, so they are treated as plain strings.
- **Breaking change:** While adding an `id` to secret item types would solve the issue, it would be a breaking change, unnecessary for other item types, and misaligned with `ItemType` semantics.
---

## Testing

- Added secret validation tests (one passing, one failing)
- Added a test asserting failure when a secret is used as an array `ItemType`

## Screens after fixes
after2

after


MTarek165:
@loicmathieu could you please give a look here?

loicmathieu (commented):
I'm not sure how it prevent using it inside ARRAY, and this would be a BC so it may be better to make it work.

MTarek165:
Nearly I have another idea to make it work without a breaking change, once I finish changes I would ping you

loicmathieu:
@MTarek165 we discussed it and it makes sense, however it should fail flow validation.

MTarek165:
Okay, so for the validation step if no validator defined in ui it should fail the flow validation and for the decryption problem we can still be able to decrypt the list of secrets and mask i know how to do it, but there is another worth mentioning point is that with array of secrets there is no validator, like if we are defining a single secret input and that is another reason also to say that we should disable it, as they will be treated as plain strings or we should add validator to the array input level ?

loicmathieu:
We agree that SECRET input sould not be part of an ARRAY input.
But this should fail at flow validation time not linked to the use of a `validator`.

MTarek165:
@loicmathieu Ok, to prevent it from usage at flow level validation we should separate between data types and types that is allowed to be element type inside arrays or multiselect as I clarified earlier as  not all types should become element types such as:
1- Array of secret
2- Array of any multi value type (enum, select, array, multiselect) as they are not working properly inside array or multiselect

so instead of explicitly applying validations for them we can define ElementType enum with the supported types which are all types except secret, and any multi value type to be rejected directly at flow level validation (during parsing) . for example for now array of array is prevented using explicit @ArrayInputValidator, so instead of that remove Array from the supported list and so on for other types making the error consistent for all non supported types. 

so the summary of final changes I have done:
1- add validation for secret input earlier at parseType() before encryption
2- prevent using secret input and multi value types as element type by separation between all types and the allowed types to be used inside arrays or multiselect inputs

after_fix

core/src/main/java/io/kestra/core/models/flows/ElementType.java
```diff
@@ -0,0 +1,41 @@
...
+
+import io.kestra.core.models.flows.input.*;
+import io.micronaut.core.annotation.Introspected;
+
+/**
+ * The supported data types as element type for ItemTypeInterface types.
+ * SECRET, SELECT, MULTISELECT, ENUM, ARRAY are not supported element types.
+ */
+@Introspected
+public enum ElementType {
```
loicmathieu: Having two enumerations is a receipe for bug, a developer may add a new type in one enum and not on the other.

I think the best would be to use a single enum, and at validation time (inside the ArrayInputValidator) check that non-supported type are not used

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/models/flows/ElementType.java
MTarek165: Maybe but I need to refer to a couple of things:
1- keeping one enum didn't prevent bugs existence as adding new type may be missed at checks also at @ArrayInputValidator if we consider the same way of thinking
2- separating between both has semantic meaning not just for the sake of validation
3- whenever we add a new itemtype interface rather than array or multiselect we should add validator to it, BTW the @MultiselectInputValidator is missing, it should be added too.

Do you still see that keeping one enum is better? 
If so I would revert it back and go for it.

loicmathieu: Yes, one enum, as not supported in arrays and multiselect is an exception not the norm.
And yes, please add validation for multiselect

MTarek165:
@loicmathieu I have reverted back the idea and went to the discussed one. 
could you take a look again?

core/src/main/java/io/kestra/core/models/flows/input/SecretInput.java
```diff
@@ -24,14 +24,6 @@ public class SecretInput extends Input<EncryptedString> {
...
-        if (validator != null && !Pattern.matches(validator, input.getValue())) {
-            throw ManualConstraintViolation.toConstraintViolationException(
-                "it must match the pattern `" + validator + "`",
-                this,
-                SecretInput.class,
-                getId(),
-                input
-            );
-        }
+       // no implementation here
```
loicmathieu: Can you explain why?

core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
```diff
@@ -452,20 +455,30 @@ private Optional<AbstractMap.SimpleEntry<String, Object>> parseData(
...
+                        SecretInput secretInput = (SecretInput) data;
+                        if (secretInput.getValidator() != null && !Pattern.matches(secretInput.getValidator(), current.toString())) {
+                            throw ManualConstraintViolation.toConstraintViolationException(
+                                "it must match the pattern `" + secretInput.getValidator() + "`",
+                                secretInput,
+                                SecretInput.class,
+                                secretInput.getId(),
+                                current.toString()
+                            );
+                        }
```
loicmathieu: Can you move secret validation back to the SecretInput class but on another method than the existing validation method.
This would allow to better understand why it's done line that:
- The existing method would have a comment saying validation is done after decoding by the other method
- The other method will be colocated, so all secret input related code would still be inside the Secret Input

loicmathieu (changes requested):
Hi,
Sorry for the late review, I now agree with the PR and will merge it after you do the last small changes

core/src/main/java/io/kestra/core/models/flows/input/SecretInput.java
MTarek165: I clarified in description why I move it into parseType() as the main problem is the secret key needed for decryption, I can't pass it to validate() from FlowInputOutput without change in it and also I can't inject it into SecretInput as it is not a micronaut managed class, that is why i make this validation earlier on parseType since there we have the original value which also looks more natural than encryption -> decryption -> validation. Could you suggest something better ?

core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
MTarek165: If I understand it well, I should add another overloaded validate() method in SecretInput encapsulating this logic right?

loicmathieu (approved):
LGTM, thansk

loicmathieu:
Even if it was broken before, as validation will fail I consider it as a "small breaking change" so I didn't backport it.

MTarek165:
Maybe not for validation but it may need backporting for the sack of secret input item decryption and log masking issue as it is exposed in logs.

PR #14288, opened 2026-01-23
feat(ui): make PostHog tracking reliable and non-blocking

proddata:
Why:
- Early UI events were dropped before configs/PostHog were ready
- PostHog init could block critical flows and pageviews were inconsistent
- Backend analytics payloads should remain unchanged

What:
- Buffer API events until configs load and drain safely
- Queue PostHog captures until posthog-js initializes; drop stale items
- Improve pageview fields for PostHog only and avoid editor tab pageview pollution
- Reset/opt-out PostHog when analytics are disabled
- Add unit tests for buffering, PostHog queueing, and API flush

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -74,7 +74,7 @@
             uid: uid,
         });
 
-        await initPostHogForSetup(config);
+        void initPostHogForSetup(config);
 
         return config;
     }
--- a/ui/src/utils/tabTracking.ts
+++ b/ui/src/utils/tabTracking.ts
@@ -115,7 +115,7 @@ function sendTrackingEvent(eventData: any) {
         delete backendData.tab_type;
         delete backendData.metadata;
 
-        apiStore.events(backendData);
+        apiStore.events(backendData, {posthog: false});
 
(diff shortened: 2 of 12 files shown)


proddata:
This now seems to track all events reliable ...
image

ui/src/App.vue
```diff
@@ -74,7 +74,7 @@
             uid: uid,
         });
 
-        await initPostHogForSetup(config);
+        void initPostHogForSetup(config);
```
proddata, line 77: make PostHog init non-blocking

ui/src/utils/tabTracking.ts
```diff
@@ -115,7 +115,7 @@ function sendTrackingEvent(eventData: any) {
         delete backendData.tab_type;
         delete backendData.metadata;
 
-        apiStore.events(backendData);
+        apiStore.events(backendData, {posthog: false});
```
proddata, line 118: this send tab events twice to posthog before (as tab event and as page view)

MilosPaunovic (approved):

PR #14290, opened 2026-01-23
feat(ui): add Welcome/Product Tour menu & adjust welcome gating

proddata:
- **Product Tour menu entry**
      - Added to the left sidebar (top) and the Kestra dropdown.
      - The dropdown entry is always visible; the left menu entry only shows when there are no non‑tutorial flows.
  - **Welcome redirect behavior**
      - Redirect to /welcome now happens only after login, and only when there are no non‑tutorial flows.
      - Otherwise, login goes to from if provided, else the dashboard (home).
  - **Tutorial namespace centralized**

https://github.com/user-attachments/assets/4345635e-af28-4e94-9cf2-ed662e570d6b


akeller:
This is aligned to my expectations in the initial onboarding experience - I can do the product tutorial if I want and it's clearer that the tutorial flows exist in the tutorials namespace 👍

MilosPaunovic (approved):

PR #14298, opened 2026-01-23
feat(webhook): add responseContentType property for custom response content types

### LINKED ISSUE #14272: [Feature]: Support custom response content-type for webhook validation handshakes
## Use Case

Integration with **Microsoft Graph Change Notifications** (and similar services like Slack, Twilio, etc.) requires a validation handshake where the webhook endpoint must:

1. Return `HTTP 200 OK`
2. Return `Content-Type: text/plain`  
3. Echo back a validation token from the request

Reference: [Microsoft Graph Webhook Validation](https://learn.microsoft.com/en-us/graph/change-notifications-delivery-webhooks?tabs=http)

> *"The client must respond... A content type of text/plain. A body that includes the URL decoded plain text validation token."*

## Current Behavior

Kestra's `io.kestra.plugin.core.trigger.Webhook` always returns JSON (`WebhookResponse` or flow outputs). There's no way to return `text/plain` or customize the response body.

## Proposed Solution

Add one optional property to `Webhook.java`: `responseContentType`

```yaml
id: microsoft_graph_webhook
namespace: company.team

tasks:
  - id: handle_request
    format: "{{ trigger.parameters.validationToken[0] ?? 'processed: ' ~ trigger.body }}"

outputs:
  - id: response
    type: STRING
    value: "{{ outputs.handle_request.value }}"

triggers:
  - id: ms_graph
    key: my-secret-key
    wait: true
    returnOutputs: true
    responseContentType: "text/plain"  # NEW - changes response from JSON to text/plain
```

**Behavior:**
- Existing `wait: true` + `returnOutputs: true` returns flow outputs
- NEW `responseContentType: "text/plain"` changes the content type from JSON to text/plain

## Alternatives Considered

- Use Azure Event Hubs or Event Grid as middleware ([Event Hubs docs](https://learn.microsoft.com/en-us/graph/change-notifications-delivery-event-hubs), [Event Grid docs](https://learn.microsoft.com/en-us/azure/event-grid/subscribe-to-graph-api-events))
- Build custom plugin

Both add unnecessary complexity when the webhook approach is cleaner.

## Backwards Compatibility

✅ Fully backwards compatible - new property defaults to `null`, preserving existing JSON behavior.

---

I'm happy to submit a PR for this feature if the maintainers are interested! 🙂

andormarkus:
## Summary

Add support for custom response content-type in webhook triggers to enable integration with services like **Microsoft Graph** that require `text/plain` responses for validation handshakes.

Closes #14272

## Changes

- Add `responseContentType` property to `Webhook` trigger
  - Allowed values: `application/json` (default), `text/plain`
- Update `ExecutionController` to handle `text/plain` responses when `returnOutputs: true`
- Add validation for allowed content type values in `WebhookValidator`
- Add documentation example for Microsoft Graph webhook validation use case
- Add unit and integration tests

## Usage

```yaml
triggers:
  - id: webhook
    key: your-secret-key
    wait: true
    returnOutputs: true
    responseContentType: "text/plain"  # NEW
```

## Test Plan

- [x] Unit tests for validation (`WebhookTest.webhookResponseContentTypeValidation`)
- [x] Integration tests for plain text response (`ExecutionControllerRunnerTest.webhookWithPlainTextResponseContentType`)
- [x] Integration tests for validation token (`ExecutionControllerRunnerTest.webhookWithPlainTextValidationToken`)
- [x] Manual testing with curl commands simulating Microsoft Graph validation

## Backwards Compatibility

✅ Fully backwards compatible - `responseContentType` defaults to `null`, preserving existing JSON behavior.

## Files Changed

| File | Change |
|------|--------|
| `core/.../trigger/Webhook.java` | Add `responseContentType` property + example |
| `core/.../validator/WebhookValidator.java` | Add validation for allowed content types |
| `core/.../validations/WebhookTest.java` | Add validation tests |
| `core/.../flows/valids/webhook-plaintext.yaml` | New test flow |
| `webserver/.../ExecutionController.java` | Handle `text/plain` response |
| `webserver/.../ExecutionControllerRunnerTest.java` | Add integration tests |

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
@@ -679,13 +679,14 @@ public class ExecutionController {
     private HttpResponse<?> buildWebhookResponse(Object body, String responseContentType) {
         if (responseContentType != null && responseContentType.equals(MediaType.TEXT_PLAIN)) {
             String responseBody;
-            if (body instanceof String) {
-                responseBody = (String) body;
-            } else if (body instanceof Map<?, ?> map && map.size() == 1) {
-                // Extract single value from outputs map for plain text response
-                responseBody = String.valueOf(map.values().iterator().next());
+            if (body instanceof String s) {
+                responseBody = s;
             } else {
-                responseBody = String.valueOf(body);
+                try {
+                    responseBody = objectMapper.writeValueAsString(body);
+                } catch (Exception e) {
+                    responseBody = String.valueOf(body);
+                }
             }
             return HttpResponse.ok(responseBody).contentType(MediaType.TEXT_PLAIN_TYPE);
         }
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
@@ -644,8 +644,8 @@ public class ExecutionController {
                     .last()
                     .map(event -> {
                         if (webhook.getReturnOutputs()) {
-                            return HttpResponse.ok(event.getData().getOutputs());
-
+                            // Only apply custom responseContentType when returnOutputs is true
+                            return buildWebhookResponse(event.getData().getOutputs(), webhook.getResponseContentType());
                         } else {
                             return (HttpResponse<?>) HttpResponse.ok(WebhookResponse.fromExecution(
                                 event.getData(),
@@ -655,6 +655,7 @@ public class ExecutionController {
                     })
                     .doFinally(signalType -> streamingService.unregisterSubscriber(executionId, subscriberId));
             } else {
+                // Without wait, always return JSON (responseContentType only applies with returnOutputs)
                 return Mono.just(HttpResponse.ok(WebhookResponse.fromExecution(result, executionUrl(result))));
             }
         } catch (QueueException e) {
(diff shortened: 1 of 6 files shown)


webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -671,6 +672,27 @@ public static WebhookResponse fromExecution(Execution execution, URI url) {
...
+     * When responseContentType is set, the response will use that content type instead of the default application/json.
+     */
+    private HttpResponse<?> buildWebhookResponse(Object body, String responseContentType) {
+        if (responseContentType != null && responseContentType.equals(MediaType.TEXT_PLAIN)) {
+            String responseBody;
+            if (body instanceof String) {
+                responseBody = (String) body;
+            } else if (body instanceof Map<?, ?> map && map.size() == 1) {
+                // Extract single value from outputs map for plain text response
+                responseBody = String.valueOf(map.values().iterator().next());
```
loicmathieu: This is opinionated.
Maybe using Jackson to serialize it as JSON would be a better fit.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -671,6 +672,27 @@ public static WebhookResponse fromExecution(Execution execution, URI url) {
...
     }
 
+    /**
+     * Build webhook response with optional custom content type.
+     * When responseContentType is set, the response will use that content type instead of the default application/json.
+     */
+    private HttpResponse<?> buildWebhookResponse(Object body, String responseContentType) {
+        if (responseContentType != null && responseContentType.equals(MediaType.TEXT_PLAIN)) {
+            String responseBody;
+            if (body instanceof String) {
```
loicmathieu:
```suggestion
            if (body instanceof String s) {
```

Use pattern matching to avoid the cast

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
loicmathieu: There is already an `objectMapper` attribute in this class that you can use for that

loicmathieu (changes requested):

andormarkus:
Hi @loicmathieu 

I have pushed the required changes.

MilosPaunovic:
If you've addressed any of the requested changes from the PR review, @andormarkus, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

loicmathieu (approved):
LGMT, thanks

PR #14304, opened 2026-01-26
Fix escaped JSON insertion in `behavior` autocomplete for `PurgeFiles`

### LINKED ISSUE #14113: Autocomplete for behaviour in PurgeFiles inserts escaped JSON string
### Describe the issue

In `io.kestra.plugin.core.namespace.PurgeFiles`, when we autocomplete the behaviour field, it gets populated like this:
`behavior: {\"type\":\"version\",\"keepAmount\":1}`

Jam: https://jam.dev/c/df0ed730-2976-420d-a5ee-60d08d4b937d

### Expected Behaviour
Autocomplete should insert the behaviour value in a clean YAML format (without escaped quotes).

### Environment

- Image Hash and Time: 7da4d01 `Fri, Jan 9, 2026 10:32 PM`

Sumitsh28:
### ✨ Description

Fixes an issue where the `behavior` field autocomplete in `io.kestra.plugin.core.namespace.PurgeFiles` inserts an escaped JSON string instead of valid YAML.

The root cause was a default value defined at schema time using `@Builder.Default` with a complex object (`Property.ofValue(Version)`), which caused the UI to serialize the default as escaped JSON.  
This PR removes the schema-level default and applies the default purge behavior (keep last 1 version) at runtime instead, ensuring correct YAML insertion while preserving existing behavior.

---

### 🔗 Related Issue

Closes #14113 

---

### 🛠️ Backend Checklist

### Before
image

### After
image


loicmathieu:
We use default values in attributes everywhere so if it's causing an issue with autocompletion it's concerning as we will not fix possibly hundreds of plugins that uses the same mechanism.

Maybe @brian-mulier-p or @Skraye that have more knownledge on this area can have a look.

brian-mulier-p:
I agree with Loïc that it's not reliable to fix it in every plugin and it's not a true fix because you may prefer to be able to edit default value.

I would reprocess the autocompletion values with \" replaced with " directly in our default yaml completion override (frontend) and then later on check if we can figure out a way to replace json values with yaml values but that's for later and a tricky subject. A simple replace for now will achieve good enough result as yaml also accepts json (as long as we remove that \" escape).

I would honestly close this PR in favor of a frontend one.

Sumitsh28:
Hi @loicmathieu and @brian-mulier-p,

Thanks for the review and the detailed explanation. I completely agree, handling this centrally rather than patching individual plugins makes much more sense for scalability.

Since I've contributed to the frontend and have familiarity with the codebase, would you like me to take a stab at this frontend fix? 

If you're good with that, I can close this PR and open a new one to implement the `\"` replacement in the YAML completion override, just as Brian suggested. Let me know what you think!

MilosPaunovic:
Yup @Sumitsh28, absolutely, go for the FE solution PR here, thanks! 🍀

PR #14306, opened 2026-01-26
feat: namespace-prefix-filter

### LINKED ISSUE #9171: True implementation of STARTS_WITH_NAMESPACE_PREFIX filter for dashboards
### Feature description

For now a [workaround using regex](https://github.com/kestra-io/kestra/commit/c439d08a0ee124dc9d02649ec7eb3642db514b3a) directly in the code is used to handle it. It may be useful for users to have such filter available directly inside the dashboard DSL. Main challenge with that one is that in theory it should be scoped to and only to namespace field and the autocompletion should show it only if selected field is namespace 👍 

As a reminder, STARTS_WITH_NAMESPACE_PREFIX is a shortcut for `namespace.equals({namespace})` or `namespace.startsWith({namespace} + ".")`

hemantmm:
closes: #9171

### ✨ Description

What does this PR change?  

- Adds a first-class `Namespace Prefix` filter operation.
- Eliminates the requirement for regex.
- Provides consistent behavior for the JDBC repo.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/QueryFilter.java
+++ b/core/src/main/java/io/kestra/core/models/QueryFilter.java
@@ -68,7 +68,7 @@ public record QueryFilter(
             case ENDS_WITH -> EndsWith.<T>builder().field(field).value(value.toString()).build();
             case CONTAINS -> Contains.<T>builder().field(field).value(value.toString()).build();
             case REGEX -> Regex.<T>builder().field(field).value(value.toString()).build();
-            case PREFIX -> Regex.<T>builder().field(field).value("^" + value.toString().replace(".", "\\.") + "(?:\\..+)?$").build();
+            case PREFIX -> Prefix.<T>builder().field(field).value(value.toString()).build();
         };
     }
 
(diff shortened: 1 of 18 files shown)


loicmathieu (changes requested):
Can you add tests for it?

This looks good for me, but I still want @brian-mulier-p's advice on it as he's the requester.

Also, this would need adaptation on EE.

brian-mulier-p (approved):
LGTM  but needs EE counterpart I think so it will wait a bit for now

hemantmm:
Thanks for the review!

Since @brian-mulier-p already approved the changes and mentioned that it needs an EE counterpart, should I still add tests for the `PREFIX` filter on this PR, or will that be handled as part of the EE adaptation?

MilosPaunovic:
You should add tests @hemantmm.

hemantmm:
@MilosPaunovic, I have added the tests.

brian-mulier-p:
@loicmathieu I've done the counterpart PR for EE [here](https://github.com/kestra-io/kestra-ee/pull/7089/changes).
However I also spotted that the changes done on this PR are partial regarding the likeIgnoreCase to startsWith conversions. I've added [this commit](https://github.com/kestra-io/kestra/pull/14306/changes/03bbec7c7365c373b67b6fa36f6d696d9c200ab3) to fill the gap.
Do you see any potential issues in doing so? If so I'd prefer keeping the likeIgnoreCase everywhere for consistency. However I feel like we will never send a namespace with uppercase in it, do you see one?

PR #14320, opened 2026-01-27
feat(webserver): add endpoint to expose metrics for running workers.

ammeek:
### ✨ Description

What does this PR change?  
Add endpoint to expose metric for running worker via the api. The following metrics will be exposed for each running worker:
- worker.job.thread
- worker.job.pending
- worker.job.running

### 🔗 Related Issue
refs #[424](https://github.com/kestra-io/kestra-ee/issues/424)

### 🛠️ Backend Checklist

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/ClusterController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/ClusterController.java
@@ -3,6 +3,7 @@ package io.kestra.webserver.controllers.api;
 import io.kestra.core.repositories.ServiceInstanceRepositoryInterface;
 import io.kestra.core.server.*;
 import io.micronaut.context.annotation.Requires;
+import io.micronaut.data.model.Pageable;
 import io.micronaut.http.HttpResponse;
 import io.micronaut.http.annotation.*;
 import io.micronaut.http.exceptions.HttpStatusException;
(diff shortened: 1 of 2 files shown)


webserver/src/main/java/io/kestra/webserver/controllers/api/ClusterController.java
```diff
@@ -30,4 +35,19 @@ public HttpResponse<ServiceInstance> getService(@PathVariable("id") String id) t
             .map(HttpResponse::ok)
             .orElseGet(HttpResponse::notFound);
     }
-}
\ No newline at end of file
+
+    @ExecuteOn(TaskExecutors.IO)
+    @Get("/metrics")
+    @Operation(tags = {"Services"}, summary = "Get metrics for running workers")
+    public Map<String, Set<Metric>> metrics() {
```
loicmathieu, line 42: I think it should be a list (or set) of metrics and not a Map<String, Set> as you don't care of the server UID and those metrics should be usable in a standard way (list of metrics).

It needs to be tested what happened when there are multiple workers, I hope we add tags to differentiate the workers (pretty sure we have them).

Also, when there are no workers, you would need to emit metrics with value 0

loicmathieu (changes requested):

Issue #14323, opened 2026-01-27
handle flow validation issue with syncFromSource command


nkwiatkowski:
### Feature description

Log flows that can't be updated instead of failing on the first error

nkwiatkowski:
Fixed and backported up to 0.22

PR #14324, opened 2026-01-27
feat(CLI): handle flow update errors with syncFromSource command

nkwiatkowski:
- fix #https://github.com/kestra-io/kestra/issues/14323


cli/src/main/java/io/kestra/cli/commands/flows/FlowsSyncFromSourceCommand.java
```diff
@@ -32,18 +33,30 @@ public Integer call() throws Exception {
...
+            } catch (RuntimeException e){
+                stdErr("Unable to update flow %s".formatted(persistedFlow.getId()), e.getMessage());
+                flowsInError.add(persistedFlow.getId());
+            }
         }
+
         stdOut("%s flow(s) successfully updated!".formatted(count));
 
+        if (!flowsInError.isEmpty()) {
+            flowsInError.forEach(flowId -> stdOut("Flow %s hasn't been updated".formatted(flowId)));
```
AcevedoR: I dont know if we use sdtOut for everything, we could send it to stdErr

AcevedoR (approved):

PR #14338, opened 2026-01-28
chore(flows): deprecate io.kestra.plugin.core.log.Fetch task in favor…

loicmathieu:
… of io.kestra.plugin.kestra.logs.Fetch


brian-mulier-p (approved):
LGTM

PR #14348, opened 2026-01-28
feat(system): Kill Switch

loicmathieu:


core/src/main/java/io/kestra/core/services/IgnoreExecutionService.java
```diff
@@ -0,0 +1,117 @@
...
+
+            return null;
+        }
+    };
+
+    record NamespaceId(String tenant, String namespace) {
+        static @Nullable NamespaceId from(String namespaceId) {
+            String[] parts = IgnoreExecutionService.splitIdParts(namespaceId);
+
+            if (parts.length == 2) {
```
fhussonnois, line 105: Here, the tenant is required, and we always expect the `namespaceId` to be `<tenant>|<namespace>` but in method `FlowId#from` we manage the case where tenant is `null`. This is confusing. Also tenant should be nullable anymore.

core/src/main/java/io/kestra/core/services/IgnoreExecutionService.java
```diff
@@ -0,0 +1,117 @@
...
+            ignoredNamespaces.contains(new NamespaceId(tenant, namespace)) ||
+            ignoredFlows.contains(new FlowId(tenant, namespace, flow)) ||
+            (executionId != null && ignoredExecutions.contains(executionId));
+    }
+
+    private static String[] splitIdParts(String id) {
+        return id.split("\\|");
+    }
+
+    record FlowId(String tenant, String namespace, String flow) {
```
fhussonnois: Why you don't reuse the existing `io.kestra.core.models.flows.FlowId.of` ?

core/src/main/java/io/kestra/core/services/IgnoreExecutionService.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @VisibleForTesting
+    boolean ignoreExecution(String tenant, String namespace, String flow, String executionId) {
+        return (tenant != null && ignoredTenants.contains(tenant)) ||
+            ignoredNamespaces.contains(new NamespaceId(tenant, namespace)) ||
+            ignoredFlows.contains(new FlowId(tenant, namespace, flow)) ||
+            (executionId != null && ignoredExecutions.contains(executionId));
+    }
+
+    private static String[] splitIdParts(String id) {
```
fhussonnois, line 86: I think we already have that somewhere ?

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
```diff
@@ -546,9 +551,14 @@ private void executionQueue(Either<Execution, DeserializationException> either)
         }
 
         Execution message = either.getLeft();
-        if (skipExecutionService.skipExecution(message)) {
-            log.warn("Skipping execution {}", message.getId());
-            return;
+
+        EvaluationType evaluationType = killSwitchService.evaluate(message);
+        if (evaluationType != EvaluationType.PASS) {
```
fhussonnois: Maybe those conditions should be directly in the `EvalutionType` for code clarity. e.g., `EvaluationType.isKillSwitchable(Execution)`.

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
```diff
@@ -1528,6 +1545,57 @@ private Executor handleFailedExecutionFromExecutor(Executor executor, FailedExec
...
+    private void handleKillSwitchedExecution(EvaluationType evaluationType, Execution message) {
+        handleKillSwitchedExecution(evaluationType, message.getTenantId(), message.getId());
+    }
+
+    private void handleKillSwitchedWorkerTaskResult(EvaluationType evaluationType, WorkerTaskResult message) {
+        handleKillSwitchedExecution(evaluationType, message.getTaskRun().getTenantId(), message.getTaskRun().getExecutionId());
+    }
+
+    private void handleKillSwitchedExecution(EvaluationType evaluationType, String tenantId, String executionId) {
+        if (evaluationType == EvaluationType.IGNORE) {
```
fhussonnois: use a `switch` here

fhussonnois (changes requested):

core/src/main/java/io/kestra/core/services/IgnoreExecutionService.java
loicmathieu: This code was pre-existing, it's a rename, I don't know why Git didn't follow the file name change.
So yes, this was code created before mandatory tenant.

core/src/main/java/io/kestra/core/services/IgnoreExecutionService.java
loicmathieu: This code was pre-existing, it's a rename, I don't know why Git didn't follow the file name change.
This was created before FlowId even exist.

core/src/main/java/io/kestra/core/services/IgnoreExecutionService.java
loicmathieu: Not that I know

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
loicmathieu: done

AcevedoR (approved):

fhussonnois (approved):

Piyush-r-bhaskar (approved):
LGTM 🚀

PR #14351, opened 2026-01-29
fix(ui): log page table filter fixed

### LINKED ISSUE #14257: Logs page table doesn't show the correct number per page when filters load
### Describe the issue

Seems the number per page doesn't always get added to the URL when the other filters get loaded in.

Similar to https://github.com/kestra-io/kestra/issues/14212

Steps to reproduce:
- Click on logs
- Select page size above 25 (e.g. 100)
- Click on Flows
- Click back to logs: only 25 are displayed regardless of what was selected
- Refresh the page and the size parameter is added back

https://github.com/user-attachments/assets/d7134a38-33f7-4d1a-a881-e26d25b28a36

### Environment

- Kestra Version: `develop` OSS (b0ec148) and EE (5c5d2c7)

singhvibhanshu:
## What changes are being made and why?

This PR addresses issue #14257,
where the page size filter would desync from the URL after navigating away from the logs table page. The fix involves updating the `useDataTableActions.ts` composable file to save the user's preference to localstorage and automatically sync the URL when the page loads, ensuring the address bar correctly reflects the current active filter.

Closes #14257 

Proof of the fix:

https://github.com/user-attachments/assets/562019fa-ef92-4678-993e-c9764b61e3cc


Piyush-r-bhaskar (changes requested):
Hey there,

Thanks for your PR.

Unfortunately, Proper fix is not from this file. it might introduce regression in other places as well. can you try to handle at component level @singhvibhanshu

singhvibhanshu:
Hi @Piyush-r-bhaskar!

Based on your feedback, i've moved the fix from the global `useDataTableActions.ts` file to the specific component `LogsWrapper.vue` file to avoid regression in other tables.

I've added a wrapper around the `@page-demand` event to save the user's selected page size to `localStorage`

Also, i used a brief `setTimeout` during the restoration phase. This is necessary because the default table initialization logic tends to overwrite the URL query params immediately on load. The timeout ensures our saved preference is applied after the default logic finishes, guaranteeing the user's setting takes precedence

Piyush-r-bhaskar (approved):

PR #14352, opened 2026-01-29
fix: improve dependency badge styling and visibility

AshharAhmadKhan:
### ✨ Description

This PR improves the dependency count badge design to make it less visually alarming and cleaner when there are no dependencies.

**Changes:**
- Changed badge color from `danger` (red) to `info` (gray/blue) for a more subtle appearance
- Hide badge completely when dependency count is 0 (no more "0" badge displayed)
- Disable the Dependencies tab when no dependencies exist
- Applied across Flow, Execution, and Asset pages

### 🔗 Related Issue

Closes #14341

### 🎨 Frontend Checklist

- [ ] Code builds without errors (`npm run build`)
- [ ] All existing E2E tests pass (`npm run test:e2e`)
- [ ] Screenshots or video recordings attached showing the UI changes

### 🛠️ Backend Checklist

_No backend changes in this PR._

### 📝 Additional Notes

The changes are minimal and focused on UI presentation only. No API or backend logic was modified. The badge now uses the `info` type (gray/blue) which is consistent with informational badges elsewhere in the UI.

### 🤖 AI Authors

Why don't cats play poker in the jungle? Too many cheetahs! 🐱

PR #14354, opened 2026-01-29
fix(executions): loop until restart and replay with max duration reached limit

### LINKED ISSUE #14326: Replay on a LoopUntil doesn't reset the maxDuration
### Describe the issue

Image

I have this LoopWhile task
```
 - id: poll_release_status
    type: io.kestra.plugin.core.flow.LoopUntil
    condition: "{{ outputs.check_status_command.exitCode == 0 }}"
    failOnMaxReached: true
    checkFrequency:
      interval: PT5M
      maxDuration: PT6H
    tasks:
...
```

and it eventually reached the maxDuration and failed. My goal was to replay the LoopUntil as if it never ran yet (so go again for another 6 hours) but it fails instantly as it comes from an execution replay which brings the already elapsed time along with it.

### Environment

- Kestra Version: develop

MTarek165:
closes #14326
## Description 
1- The problem is the same for replay and restart and it happens at LoopUntil#reachedMaximums() when checking creationDate with maxDuration against current time, it always gets the first time the task is created adding maxDuration and compares against current time which is always passed since the last executed time that is why it is failing instantly after restart.
2- For now the state is changed to Running directly when restarting, keeping it difficult to track the first start time since Running states is cleaned from history between LoopUntil iterations.

## Solution
1- This can be done by adding Restart state for LoopUntil task before changing to Running to record time of restart and keeping it at the history as like Created to be easy to use it at comparison at LoopUntil#reachedMaximums() for subsequent iterations for the new restarted loop until execution.
2- Preserve Created and Restarted cases only for flowable task run history and get the recent to use it in checks (which is Restarted in our case)

## Testing
1- Added a LoopUntil restart and replay test which should assert that the final restarted or replayed execution is failed and we have restarted state for the LoopUntil task and ends with Failed state and the time elapsed between both is the same for max duration to make sure that the max duration period is respected after restart/replay of LoopUntil task

## Screen after fix
1- This screen shows the state history for replayed LoopUntil task with Max duration 3M and Interval 1M
2- A small note: the old history of running and failed doesn't appear between created and restarted after an iteration is done and starting a new one, since we are cleaning flowable task run history each time a new iteration start keeping only created and restarted for avoiding large history as stated but max duration is respected now as shown.

replay

### DIFF
--- a/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
+++ b/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
@@ -319,12 +319,12 @@ public class TaskRun implements TenantInterface {
     }
 
     public TaskRun resetAttempts() {
-        List<State.History> wantedHistories = this.state.getHistories()
+        List<State.History> creationHistories = this.state.getHistories()
             .stream()
             .filter(history -> history.getState().isCreated())
             .toList();
         return this.toBuilder()
-            .state(new State(wantedHistories.getLast().getState(), wantedHistories))
+            .state(new State(creationHistories.getLast().getState(), creationHistories))
             .attempts(null)
             .build();
     }
--- a/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
+++ b/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
@@ -319,8 +319,12 @@ public class TaskRun implements TenantInterface {
     }
 
     public TaskRun resetAttempts() {
+        List<State.History> wantedHistories = this.state.getHistories()
+            .stream()
+            .filter(history -> history.getState().isCreated())
+            .toList();
         return this.toBuilder()
-            .state(new State(State.Type.CREATED, List.of(this.state.getHistories().getFirst())))
+            .state(new State(wantedHistories.getLast().getState(), wantedHistories))
             .attempts(null)
             .build();
     }
--- a/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
+++ b/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
@@ -18,6 +18,7 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 @ToString
 @EqualsAndHashCode
@@ -319,12 +320,14 @@ public class TaskRun implements TenantInterface {
     }
 
     public TaskRun resetAttempts() {
-        List<State.History> creationHistories = this.state.getHistories()
+        State.Type lastCreationState = this.state.getHistories()
+            .reversed()
             .stream()
             .filter(history -> history.getState().isCreated())
-            .toList();
+            .findFirst().get()
+            .getState();
         return this.toBuilder()
-            .state(new State(creationHistories.getLast().getState(), creationHistories))
+            .state(new State(lastCreationState, this.state.getHistories()))
             .attempts(null)
             .build();
     }
--- a/core/src/main/java/io/kestra/core/services/ExecutionService.java
+++ b/core/src/main/java/io/kestra/core/services/ExecutionService.java
@@ -165,7 +165,7 @@ public class ExecutionService {
             .map(taskRun -> {
                 if (taskRun.getId().equals(flowableTaskRunId)) {
                     // Keep only CREATED/RESTARTED to avoid having large history
-                    // Keeping RESTARTED at history is useful for LoopUntil Task Restart case since we depend on it to check reached max duration
+                    // Keeping RESTARTED at history is useful for LoopUntil Task Restart/Replay case since we depend on it to check reached max duration
                     return taskRun.resetAttempts().incrementIteration();
                 }
 
--- a/core/src/main/java/io/kestra/core/services/ExecutionService.java
+++ b/core/src/main/java/io/kestra/core/services/ExecutionService.java
@@ -31,6 +31,7 @@ import io.kestra.core.storages.StorageInterface;
 import io.kestra.core.utils.GraphUtils;
 import io.kestra.core.utils.IdUtils;
 import io.kestra.core.utils.ListUtils;
+import io.kestra.plugin.core.flow.LoopUntil;
 import io.kestra.plugin.core.flow.Pause;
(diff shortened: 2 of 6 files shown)


MTarek165:
@MilosPaunovic It seems like the last PR action is buggy, since I’m not a new contributor 😂

MilosPaunovic:
@MTarek165 It's alright, as if you were new contributor, it would send you an automatic welcome message, for example, see https://github.com/kestra-io/kestra/pull/14352#issuecomment-3816714673.

MTarek165:
Nice but it seems like for this PR the tests are disabled, no triggered CI as it is the case for first time contributors.

MilosPaunovic:
It's unrelated, we seem to have some other CI problems, it will be sorted swiftly, thanks for pointing it out! 🍀

MTarek165:
Hi @brian-mulier-p, could you give some time and give a look on this PR?

MTarek165:
Hi @loicmathieu , sorry about that but could you please review this PR? 
since it was waiting for review for a month.

loicmathieu:
@MTarek165, @brian-mulier-p is already reviewing this PR.

brian-mulier-p:
@MTarek165 sorry it's been really busy weeks lately, I'm into it rn, I had to rebase everything but I'm not exactly sure I pulled all your latest changes, can you double check all your changes are there? Otherwise I recommand you to force push again from your local changes and I'll rebase it better.

core/src/main/java/io/kestra/core/models/executions/TaskRun.java
```diff
@@ -319,8 +319,12 @@ public TaskRun incrementIteration() {
     }
 
     public TaskRun resetAttempts() {
+        List<State.History> creationHistories = this.state.getHistories()
+            .stream()
+            .filter(history -> history.getState().isCreated())
+            .toList();
         return this.toBuilder()
-            .state(new State(State.Type.CREATED, List.of(this.state.getHistories().getFirst())))
+            .state(new State(creationHistories.getLast().getState(), creationHistories))
```
brian-mulier-p: That's not the right way to go imo as we will lose the history of the previous states. I think we better change the logic of the LoopUntil that should compare it's maxDuration based off the last isCreated instead of rewriting the whole state history.

core/src/main/java/io/kestra/plugin/core/flow/LoopUntil.java
```diff
@@ -181,8 +181,13 @@ private boolean reachedMaximums(RunContext runContext, Execution execution, Task
             if (printLog) {logger.warn("Max iterations reached");}
             return true;
         }
-
-        Instant creationDate = parentTaskRun.getState().getHistories().getFirst().getDate();
+        Instant creationDate = parentTaskRun.getState()
```
brian-mulier-p, line 184: This should be sufficient, the taskRun attempts rewrite above seems fishy

brian-mulier-p (changes requested):
LGTM except one of the change where I'm not confident about it.
Sorry for the review delay...
Also I tested locally and the issue example works perfectly, well done!

core/src/main/java/io/kestra/core/models/executions/TaskRun.java
MTarek165: Maybe, I was trying to align with the comment of avoiding too long history after my change, but agree it is not that problem we may remove overwriting step overall, this is better for user to see all transitions states.

core/src/main/java/io/kestra/core/services/ExecutionService.java
```diff
@@ -163,8 +164,8 @@ public Execution retryWaitFor(Execution execution, String flowableTaskRunId) {
             .stream()
             .map(taskRun -> {
                 if (taskRun.getId().equals(flowableTaskRunId)) {
-                    // Keep only CREATED/RUNNING
-                    // To avoid having large history
+                    // Keep only CREATED/RESTARTED to avoid having large history
+                    // Keeping RESTARTED at history is useful for LoopUntil Task Restart/Replay case since we depend on it to check reached max duration
```
MTarek165: @brian-mulier-p This is the referred comment.

core/src/main/java/io/kestra/core/services/ExecutionService.java
```diff
@@ -838,7 +839,7 @@ private TaskRun mapTaskRun(
             alterState = originalTaskRun.withState(newStateType).getState();
         } else {
             Task task = flow.findTaskByTaskId(originalTaskRun.getTaskId());
-            if (!task.isFlowable() || task instanceof WorkingDirectory) {
+            if (!task.isFlowable() || task instanceof WorkingDirectory || task instanceof LoopUntil) {
```
MTarek165, line 840: @brian-mulier-p Before merging this PR I have opinion on the change here beyond the context of the PR, related to Flowable transition to Running directly for restart/replay except for WorkingDirectory and LoopUntil tasks here as I see if there is no specific reason for that it is better to give all flowables a RESTARTED state as it is more natural and may have benefits like we have for LoopUntil.

MTarek165:
@brian-mulier-p I agree also with your point about clearing history, so after change now all history is preserved between iterations.
 see update for TaskRun#resetAttempts() 
preserve_history

core/src/main/java/io/kestra/core/services/ExecutionService.java
brian-mulier-p: idk, is there an impact on the resolution to remove the LoopUntil exception here? I can see why WorkingDirectory is there because if a subtask is restarted then the WorkingDirectory is also restarted because it's a special wrapper task that forces any subtask to be performed under the same worker so it means if any subtask or the WorkingDirectory task itself is restarted, the whole is restarted so it makes more sense to see the WorkingDirectory task also as restarted

MTarek165: Keeping LoopUntil as Restarted is essential step as it is a reference for the max duration when checking reaching maximum at LoopUntil between iterations.
I was wondering if we should make all flowables being Restarted, but since we don't know the consequences of this, let's keep aligned with the scope of the PR only.
You can merge it now 😀

brian-mulier-p (approved):
LGTM

PR #14361, opened 2026-01-30
chore(tests): enhance test coverage for WaitFor max duration functionality

MTarek165:
## Description
- the current waitForMaxDuration() test doesn't test the max duration well, since failed execution doesn't guarantee its working, since it can fail instantly without applying the duration properly like in restart/replay LoopUntil here #14326
- Although it is not happening for normal LoopUntil execution now but it is important to prevent regression and ensure the functionality is tested properly.


loicmathieu (approved):
LGTM, thansk

PR #14371, opened 2026-01-30
fix: base docker image was missing curl

AcevedoR:
eclipse-temurin:25-jre-jammy does not bundle curl anymore


loicmathieu (approved):
LGTM

PR #14383, opened 2026-01-31
perf(utils): replace TreeMap with HashMap in MapUtils

MTarek165:
## Description
- TreeMap provides no functional benefit here and only adds sorting overhead.
flattenToNestedMap() is on a critical path and is invoked at least twice per execution, even when no inputs, outputs, or labels are defined, and may also be triggered indirectly from other paths( like each time creating a new run context by adding execution labels), so for a non trivial executions it is called many times per execution.
- With thousands of executions with defined inputs/outputs/labels, this introduces unnecessary slight overhead. 
- Using HashMap aligns better with the existing model and avoids this cost.


loicmathieu (commented):
@fhussonnois did you remember if there was some rational to use a TreeMap here?

fhussonnois:
I don't remember the reason behind it, if there was one. Because, it's an utility class it's seems ok to use HashMap here.

loicmathieu (approved):
LGTM, thanks

PR #14385, opened 2026-02-02
perf(executions): optimize find last running task lookup at Execution

MTarek165:
## Description
- Streams.findLast() is equivalent to reduce((a, b) -> b) with O(n) time complexity, 
always scanning the entire list even when the target is near the end.
- While Streams.findLast() is appropriate for generic streams since we have no indexing, using it with 
Lists is inefficient since we have random access.
- The current implementation is inefficient, it filters all matching a condition then iterates over them again to get the last

Use reversed().findFirst() instead:
- reversed() creates a view in O(1) time (no copying)
- findFirst() short-circuits, stopping at the first match
- Overall complexity: O(k) where k = elements from end to first match

- This can make difference for iterative tasks (ForEach, LoopUntil) that can generate 
hundreds of task runs.

## JMH Benchmark Tests

Warmup: 5 iterations, 10 s each
Measurement: 5 iterations, 10 s each
Timeout: 10 min per iteration
Threads: 1 thread, will synchronize iteration
Units: Microseconds
## Results

- for 10 task run

Before: "io.kestra.core.executions.ExecutionsBenchmark.oldFindLastCreatedTaskRun":
  (min, avg, max) = (0.304, 0.364, 0.460)

After: "io.kestra.core.executions.ExecutionsBenchmark.newFindLastCreatedTaskRun":
  (min, avg, max) = (0.201, 0.252, 0.340)

Performance gain: 30.8%
1.44× faster than old one

- for 100 task run

Before: "io.kestra.core.executions.ExecutionsBenchmark.oldFindLastCreatedTaskRun":
  (min, avg, max) = (8.782, 11.471, 15.608)
  
After: "io.kestra.core.executions.ExecutionsBenchmark.newFindLastCreatedTaskRun":
  (min, avg, max) = (2.890, 3.885, 6.431)

Performance gain: 66.1%
2.95× faster than old one

- for 300 task run 

Before: "io.kestra.core.executions.ExecutionsBenchmark.oldFindLastCreatedTaskRun": 
(min, avg, max) = (24.755, 29.618, 46.953)

After: "io.kestra.core.executions.ExecutionsBenchmark.newFindLastCreatedTaskRun":
(min, avg, max) = (4.062, 4.551, 5.095)

Performance gain: 84.6%
6.51× faster than old one

## Observations
- The results show a significant performance gain between the two approaches.
- Even for only 10 task runs, the new approach is nearly 1.5 faster , and the improvement grows with more task runs.
- The new approach scales better because it leverages backward traversal to find the last matching element without iterating the entire list unnecessarily, which is crucial for thousands of executions since it is on a critical execution path.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/executions/Execution.java
+++ b/core/src/main/java/io/kestra/core/models/executions/Execution.java
@@ -576,6 +576,17 @@ public class Execution implements SoftDeletable<Execution>, TenantInterface, Has
             .findFirst();
     }
 
+    public Optional<TaskRun> findFirstRunning() {
+        if (this.taskRunList == null) {
+            return Optional.empty();
+        }
+
+        return this.taskRunList
+            .stream()
+            .filter(t -> t.getState().isRunning())
+            .findFirst();
+    }
+
     public Optional<TaskRun> findLastNotTerminated() {
         if (this.taskRunList == null) {
             return Optional.empty();
--- a/core/src/main/java/io/kestra/core/models/executions/Execution.java
+++ b/core/src/main/java/io/kestra/core/models/executions/Execution.java
@@ -8,7 +8,6 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Streams;
 import io.kestra.core.debug.Breakpoint;
 import io.kestra.core.exceptions.InternalException;
 import io.kestra.core.models.SoftDeletable;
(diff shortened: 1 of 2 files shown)


loicmathieu:
`reversed().findFirst()` didn't bruing the same sematic meaning, and is less readable.
Such optimization must bring justification either by an article or by bringing performance benchmark as a proof.

It may also be good to add a comment otherwize someone may reverse the optimization for readability concern.

MTarek165:
@loicmathieu I agree it’s slightly less readable at first glance. I can add a comment to clarify the intent.  
I’m confident this approach is more efficient. I can provide a benchmark , could you suggest a realistic workload size for TaskRun in production so I can use that and make realistic estimation?

loicmathieu:
For benchmarking, 10 taskrun seems to be a good fit.
If you know how JMH works, we have some JMH benchmarks in our code base, otherwise any other type of benchmark will work as soon as it proves to have a real benefit.

And yes, please add a comment

MTarek165:
@loicmathieu I have done benchmarks for set of (10, 100, 300) task runs and the performance gain is high as exepected, I have updated the proposal with the all details

core/src/main/java/io/kestra/core/models/executions/Execution.java
```diff
@@ -587,50 +586,65 @@ public Optional<TaskRun> findFirstRunning() {
             .findFirst();
     }
 
+    /**
```
loicmathieu:
```suggestion
    /*
```

This should be a developer comment not a JavaDoc.

loicmathieu (approved):
LGTM, thanks.
I'll push the comment change myself before merging.

PR #14388, opened 2026-02-02
feat(ui): introduce shared IconButton for action icons

proddata:
Add a reusable IconButton component that encapsulates tooltip behavior, consistent sizing, and optional router-link support. This reduces boilerplate across table action columns and ensures icon-only actions are accessible and visually consistent.

Migrate table action icons (flows, executions, templates, secrets, KV, triggers, logs, sidebar toggle) to use the new component and common styling.

Note: follow up in kestra-ee to align the remaining enterprise tables.

image

image

cc: @Nico-Kestra

### DIFF
--- a/ui/src/styles/layout/element-plus-overload.scss
+++ b/ui/src/styles/layout/element-plus-overload.scss
@@ -98,6 +98,34 @@
     }
 }
 
+.ks-icon-button {
+    color: var(--ks-content-primary);
+    width: 24px;
+    height: 24px;
+    min-width: 24px;
+    border-radius: var(--bs-border-radius);
+    text-align: center;
+    display: inline-flex;
+    justify-content: center;
+    align-items: center;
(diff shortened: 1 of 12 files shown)


ui/src/styles/layout/element-plus-overload.scss
```diff
@@ -98,6 +98,34 @@
     }
 }
 
+.ks-icon-button {
```
MilosPaunovic: Nitpick: I'd move this to the component itself, but on you whether you'd do it or keep like this.

MilosPaunovic (approved):

ui/src/styles/layout/element-plus-overload.scss
proddata: done ✅

MilosPaunovic (approved):

Issue #13700, opened 2025-12-16
Subflow execution with input parameter: fails after Kestra restart


conapi-stefan:
### Describe the issue

I configured a flow that triggers multiple subflows and passes input parameters.
After creation this works find. Then I restarted kestra and on the first execution of the main flow I get:

```
2025-12-16 15:20:23.887Created new execution [4UflJcry2c1IyTSpOwEuSK](http://localhost:8080/ui/main/executions/company.team/subflow1/4UflJcry2c1IyTSpOwEuSK) for flow [company.team.subflow1](http://localhost:8080/ui/main/flows/edit/company.team/subflow1)
2025-12-16 15:20:23.915Input productId was provided by parent execution 2xcs0KMVOfRH7EeghKuhkL for subflow company.team.subflow1 but isn't declared at the subflow inputs
2025-12-16 15:20:24.092Subflow execution [4UflJcry2c1IyTSpOwEuSK](http://localhost:8080/ui/main/executions/company.team/subflow1/4UflJcry2c1IyTSpOwEuSK) for flow [company.team.subflow1](http://localhost:8080/ui/main/flows/edit/company.team/subflow1) ends in FAILED state
```

To get it working again I have to:
-  Go to each subflow and add a defaults value. 
- Then run the main flow (invokesubflow ) again, it works.
- Now I can remove the defaults from subflow1 again.
- Run the invokesubflow again. Still works.

### Environment

- Kestra Version: 1.1.9 (also in 1.1.5)
- Docker with Postgres

[main_company.team_invokesubflow.yml](https://github.com/user-attachments/files/24193588/main_company.team_invokesubflow.yml)
[main_company.team_subflow1.yml](https://github.com/user-attachments/files/24193587/main_company.team_subflow1.yml)

wrussell1999:
@Pradumnasaraf can you try and reproduce this please. This is blocking a webinar in January.

Pradumnasaraf:
Hey,
I try to reproduce on version `1.1.9`, but I am unable to. Here is a Jam for reference: https://jam.dev/c/e30a214e-de43-4028-af21-9e628d88641c

conapi-stefan:
@Pradumnasaraf Here is a Jam where I reproduce it with attached docker compose: https://jam.dev/c/7758d52f-09c4-4fb0-b825-a14f045ab841

[docker.zip](https://github.com/user-attachments/files/24216886/docker.zip)

Pradumnasaraf:
[...]

PR #14390, opened 2026-02-02
fix(system): the WatchService should not be null

loicmathieu:
Using `@Nullable` here hides a configuration issue where watching is enabled but due to mis-configuration (ex: the directory to watch didn't exist) the WatchService cannot be created by Micronaut.

Part-of: https://github.com/kestra-io/kestra/issues/13700


brian-mulier-p (approved):
LGTM

loicmathieu:
Backported in 1.0 and 1.2

PR #14392, opened 2026-02-02
fix(kv): PurgeKV should also purge parent namespace

loicmathieu:
Even when the namespace has no flow.

Fixes https://github.com/kestra-io/kestra-ee/issues/5890


brian-mulier-p (approved):
LGTM

loicmathieu:
Backported to 1.2

PR #14404, opened 2026-02-03
perf(jdbc): Improve the defaultFilter so that it can correctly hit th…

bluesecret:
…e index.

Improve the default filter so that it always includes a `DELETE` condition in the SQL statement. This will ensure that indexes are correctly used and prevent full table scans.

- Case where there is no deleted field

887043da-9466-4b9a-a87f-1e93dc44e44a

---

- Cases of using the deleted field

image

---

### ✨ Description

What does this PR change?  
_Example: Replaces legacy scroll directive with the new API._

### 🔗 Related Issue

none

### 🛠️ Backend Checklist

- [√] Code compiles successfully and passes all checks
- [√] All unit and integration tests pass

### 📝 Additional Notes

none

### 🤖 AI Authors

none


loicmathieu (approved):
LGTM, thanks.
I need to adapt our EE code before merging

PR #14407, opened 2026-02-03
feat(core): refactor the webhook to have a AbstractWebhookTrigger as plugins

tchiotludo:


core/src/main/java/io/kestra/core/http/HttpResponse.java
```diff
@@ -78,14 +85,219 @@ public static <T> HttpResponse<T> from(ClassicHttpResponse httpResponse, T body,
...
+            .flatMap(headers -> headers.firstValue("Content-Type"))
+            .orElseThrow(() -> new IllegalStateException("Response has no Content-Type header [" + this.getHeaders() + "]"));
+    }
+
     /**
      * This data class is part of the plugin API. Changes are potentially breaking.
      */
     @Value
     @Builder
+    @AllArgsConstructor
```
fhussonnois, line 124: We should not maintain this kind of class.

core/src/main/java/io/kestra/plugin/core/trigger/WebhookContext.java
```diff
@@ -0,0 +1,27 @@
...
+import io.kestra.core.http.HttpRequest;
+import io.kestra.core.models.flows.Flow;
+import io.kestra.core.services.WebhookService;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+
+@Builder
+@AllArgsConstructor
+public class WebhookContext {
```
fhussonnois: Use java `record` for this kind of object.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -580,118 +593,42 @@ protected Publisher<HttpResponse<?>> webhook(
...
+            webhookService
+        );
 
+        // Call evaluate and create a failed execution if exception occurs
         try {
-            // inject the traceparent into the execution
-            Optional<TextMapPropagator> propagator = openTelemetry
-                .map(OpenTelemetry::getPropagators)
-                .map(ContextPropagators::getTextMapPropagator);
+            return webhook.evaluate(webhookContext).toMicronaut();
```
fhussonnois: I'm not sure, but before we had a reactive approach but now the `Webhook#evaluate` method is blocking when `wait=true` so we will keep a Micronaut's IO thread while waiting. Even if there is no risk of thread-pool being exhausted (under the hood it's a unbounded cachedThreadPool) with many concurrent `wait: true` webhooks on long-running flows we could have memory overhead.

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
```diff
@@ -171,42 +151,104 @@ This is useful for webhook validation handshakes that require specific content t
...
+                RunContext runContext = context.getWebhookService().runContext(context.getFlow(), event.getData());
+                int responseCode = runContext.render(this.responseCode).as(Integer.class).orElse(200);
+
+                if (this.getReturnOutputs()) {
+                    return buildOutputResponse(event.getData().getOutputs(), responseContentType, HttpResponse.Status.valueOf(responseCode));
+                } else {
+                    return HttpResponse.of(HttpResponse.Status.valueOf(responseCode), context.getWebhookService().executionResponse(event.getData()));
+                }
+            }))
+            .block();
```
fhussonnois: The evalute method should be fully reactive and return a `Mono<HttpResponse<?>` so the Controller can properly handle all the reactive stuff

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
```diff
@@ -171,42 +151,104 @@ This is useful for webhook validation handshakes that require specific content t
...
+        }
+
+        if (!this.wait) {
+            return HttpResponse.of(context.getWebhookService().executionResponse(execution));
+        }
+
+        return context
+            .getWebhookService()
+            .followExecution(execution, context.getFlow())
+            .last()
```
fhussonnois, line 208: I think `.last()` method will throw an Exception if there is no element. I will not be suprised this could happen ^^

core/src/main/java/io/kestra/core/http/HttpSseEvent.java
```diff
@@ -0,0 +1,66 @@
...
+
+import lombok.Builder;
+import lombok.Data;
+
+import java.time.Duration;
+
+
+@Data
+@Builder(toBuilder = true)
+public class HttpSseEvent<T> {
```
fhussonnois: Use a java `record` for that. `@Builder` is OK.

core/src/main/java/io/kestra/core/http/client/HttpClient.java
```diff
@@ -272,6 +278,135 @@ public <T> HttpResponse<T> request(HttpRequest request) throws HttpClientExcepti
...
+    /**
+     * Send an SSE (Server-Sent Events) request and consume events with typed data.
+     *
+     * @param request the HTTP request
+     * @param cls the class type for deserializing event data
+     * @param eventConsumer consumer that processes each SSE event with typed data
+     * @param <T> the type of data in the SSE events
+     * @return the HTTP response without the body, as events are consumed through the eventConsumer
+     */
+    public <T> HttpResponse<Void> sseRequest(
```
fhussonnois, line 290: For me it's over-engineering to polluate the HTTPClient, which is used by multiple plugins, with SSE which is currently only used by the core `SseRequest` task. 

The `SseRequest` should handle all the SSE parsing logic and because it's a core plugin we could have use the Micronaut `SseClient` directly. We are reinventing the wheel...

core/src/main/java/io/kestra/core/http/HttpRequest.java
```diff
@@ -64,6 +70,39 @@ public static HttpRequest from(org.apache.hc.core5.http.HttpRequest request) thr
             .build();
     }
 
+    public static HttpRequest from(io.micronaut.http.HttpRequest<?> request) {
```
fhussonnois: If the initial idea was to have an HTTPClient with no dependency on Micronaut we should not have this type of utility method here.

core/src/main/java/io/kestra/core/http/HttpResponse.java
```diff
@@ -78,14 +85,219 @@ public static <T> HttpResponse<T> from(ClassicHttpResponse httpResponse, T body,
             .build();
     }
 
+    public io.micronaut.http.HttpResponse<?> toMicronaut() {
```
fhussonnois: We should not introduce Micronaut HTTP dependency in our HTTP Client.

fhussonnois (changes requested):

core/src/main/java/io/kestra/core/utils/UriProvider.java
```diff
@@ -51,4 +52,14 @@ public URI flowUrl(FlowInterface flow) {
             flow.getNamespace() + "/" +
             flow.getId());
     }
+
+    public URI webhookUrl(FlowInterface flow, AbstractWebhookTrigger trigger) {
+        return this.build("/api/v1/" +
+            (flow.getTenantId() != null ? flow.getTenantId() + "/" : "") +
+            "executions/webhook" +
```
fhussonnois: Missing `/` after "webhook"

fhussonnois (changes requested):
All tests must be fixed!

core/src/main/java/io/kestra/core/http/HttpResponse.java
tchiotludo: part of our api for plugins, we don't want to have any dependencies expose to being abler to change the underlying dependencies.

core/src/main/java/io/kestra/core/http/client/HttpClient.java
tchiotludo: In AI era, SSE is everywhere, having a central way to do SSE is definitively an advantage on plugins side, sure it will be used somehow

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
tchiotludo: Moved back to reactive

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
tchiotludo: could never happen to have no elements, since we have at the initial state that we create on top

fhussonnois (approved):

PR #14442, opened 2026-02-04
fix(core): avoid infinite redirects on main dashboard

MilosPaunovic:
Closes https://github.com/kestra-io/kestra-ee/issues/6510.

Can you double check if this works now for you @proddata?


proddata:
It does seem to work.

I think the flashing redirect behaviour was there before.

https://github.com/user-attachments/assets/ac0a7fbe-4065-4cb8-a101-b693e2c84037

proddata (approved):

MilosPaunovic:
Backported to `releases/v1.2.x` branch.

PR #14446, opened 2026-02-04
fix(flows): import flows should do model validation

MTarek165:
As a follow up for #14055 importing flow doesn't apply model validation, discarding for example defined label validator NoSystemLabelValidation allowing to bypass system labels without rejecting it unlike defining it in flow at UI


loicmathieu:
Asking validation to our product team first as it would be a breaking change

loicmathieu:
Hi,
After discussion, we want to be able to import invalid flows even if they would then not be executable.
This is so a user can export a set of flows and found them inside the UI even if validation fail.

So thanks for your contribution but we will not proceed.

PR #14452, opened 2026-02-05
fix(labels): prevent system labels usage from schedule triggers

MTarek165:
- System labels defined on schedule triggers were not being excluded, which allowed them to appear in logs for scheduled executions if the flow bypassed earlier model validations and saved into the database which can happen in some cases (e.g. flow import, see #14446), system labels must be excluded from triggers and backfill just as flow labels when creating scheduled execution regardless of the earlier validation state to avoid inconsistencies and unintended exposure as there may be other ways to bypass validations rather than flow import.

- Refactor LabelService#labelsExcludingSystem() to handle labels uniformly across all contexts (flow, trigger, backfill, execution). The method is already null-safe, allowing removal of unnecessary null checks at call sites and resulting in cleaner, less verbose code.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/executions/Execution.java
+++ b/core/src/main/java/io/kestra/core/models/executions/Execution.java
@@ -194,7 +194,7 @@ public class Execution implements SoftDeletable<Execution>, TenantInterface, Has
             .kind(kind)
             .build();
 
-        List<Label> executionLabels = new ArrayList<>(LabelService.labelsExcludingSystem(flow));
+        List<Label> executionLabels = new ArrayList<>(LabelService.labelsExcludingSystem(flow.getLabels()));
         if (labels != null) {
             executionLabels.addAll(labels);
         }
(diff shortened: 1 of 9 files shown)


MTarek165:
Hi @loicmathieu , could you take a look on this PR and #14446 also? 
since there was a merge conflict here but I have resolved  and it is ready for review.

MTarek165:
Hi @loicmathieu , could you take a look on this PR and https://github.com/kestra-io/kestra/pull/14446 also?
since there was a merge conflict here but I have resolved and it is ready for review now, thanks for your efforts in reviews.

loicmathieu (approved):
LGTM, thanks

loicmathieu:
@MTarek165 can you rebase the branch and handle the conflicts? I'll merge after that.

MTarek165:
@loicmathieu I rebased it and compacted into important commits only.

PR #14468, opened 2026-02-05
scheduler queue migration

fhussonnois:


core/src/main/java/io/kestra/core/scheduler/events/SchedulerEvent.java
```diff
@@ -18,34 +20,44 @@
     @JsonSubTypes.Type(value = SchedulerEvent.VNodesAssignmentRelease.class, name = "VNODES_ASSIGNMENT_RELEASE"),
     @JsonSubTypes.Type(value = SchedulerEvent.VNodesAssignmentRejected.class, name = "VNODES_ASSIGNMENT_REJECTED"),
 })
-public interface SchedulerEvent {
-    
+public interface SchedulerEvent extends BroadcastEvent {
```
loicmathieu: Should extens `HasUID`

loicmathieu (approved):
LGMT however I would prefere if we can get rid of the SchedulerEventQueue and the TriggerEventQueue which sems unnecessary

PR #14469, opened 2026-02-05
feat(API): prevent changing state of a killed or non terminated taskRun

nkwiatkowski:
- fix #https://github.com/kestra-io/kestra-ee/issues/6529


loicmathieu (approved):
LGTM

PR #14473, opened 2026-02-06
Feat/batch subscription

loicmathieu:
Migrate to queue v2 the following queues:
- metric
- execution event
- execution
- log

Add support for batch consuming (supported only in JDBC).

Use byte[] serialization.

Add a FollowExecution queue for the follow execution endpoint.
Add a FollowLog queue for the follow log endpoint.


fhussonnois (approved):
LGTM

PR #14474, opened 2026-02-06
(fix): Remove minimum size requirement for values in SELECT input

### LINKED ISSUE #14472: Remove length requirement for SELECT inputs
### Describe the issue

With https://github.com/kestra-io/kestra/pull/12028 a minimum length requirement was introduced that requires  `values` to at least include 2 elements. However there might be use cases were one wants to use the `SELECT` as combobox with only one predefined value and the option `allowCustomValue: true`. 

Generally speaking I don't see a good reason to have any size requirement for the input.

### Environment

- Kestra Version: 1.2

proddata:
Reverts https://github.com/kestra-io/kestra/pull/12028

closes https://github.com/kestra-io/kestra/issues/14472


AcevedoR (approved):
ok for me unless Brian has an additional insight

PR #14484, opened 2026-02-09
build: add-missing-optional-deps-for-rolldown

elevatebart:
Add missing dependencies after upgrade of rolldown


MilosPaunovic (approved):

MilosPaunovic (commented):
Is it needed on `EE` also?

PR #14494, opened 2026-02-09
fix: no code pluginDefaults

elevatebart:
- [x] Saving pluginDefaults need to be well formatted
- [x] When selecting plugin DateBetween we woudl see NaN
- [x] Id would be visible and break interface


Piyush-r-bhaskar (approved):
Looks good 🚀

PR #14495, opened 2026-02-09
feat(core): Expose service instance metrics via API and add new pending worker messages metric.

ammeek:
### ✨ Description

What does this PR change?  
This PR adds the ability to query service instance metrics via the web server.
This PR also add a new metric for the amount of unconsumed messages on the worker queue.

### 🔗 Related Issue
[Part of #424](https://github.com/kestra-io/kestra-ee/issues/424)

### 🛠️ Backend Checklist


core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
```diff
@@ -90,6 +88,8 @@ public class MetricRegistry {
     public static final String METRIC_EXECUTOR_EXECUTION_QUEUED_COUNT_DESCRIPTION = "The total number of executions queued by the Executor";
     public static final String METRIC_EXECUTOR_EXECUTION_POPPED_COUNT = "executor.execution.popped.count";
     public static final String METRIC_EXECUTOR_EXECUTION_POPPED_COUNT_DESCRIPTION = "The total number of executions popped by the Executor";
+    public static final String WORKER_MESSAGES_UNCONSUMMATED_COUNT = "queue.worker.messages.unconsummated.count";
+    public static final String WORKER_MESSAGES_UNCONSUMMATED_COUNT_DESCRIPTION = "Total number of worker jobs in the queue that are not yet consumed by a worker";
```
loicmathieu: It's by all queues not only the worker queue right?
So better to name it `queue.messages.lag.count` I also think lag is more easy to understnad tha unconsummated.

core/src/main/java/io/kestra/core/queues/QueueInterface.java
```diff
@@ -42,6 +43,8 @@ default void delete(T message) throws QueueException {
         delete(null, message);
     }
 
+    Map<String, Integer> queueLagByWorkerGroup(Class<?> queueType);
```
loicmathieu: Worker groups is exposed by the consumer group not the queueType.
So here you compute queueLagByQueueType not by worker group.

Moreover, as worker group is not something exposed by the queue, you should not have a method talking about it but only about the abstract consumerGroup or here queueType

core/src/main/java/io/kestra/core/queues/WorkerQueueSizePoller.java
```diff
@@ -0,0 +1,53 @@
...
+import jakarta.inject.Singleton;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+@Slf4j
+@Singleton
+@Requires(property = "kestra.server-type", pattern = "(WEBSERVER|STANDALONE)", defaultValue = "STANDALONE")
```
loicmathieu: Don't set a default value they will always be one for this property

core/src/main/java/io/kestra/core/queues/WorkerQueueSizePoller.java
loicmathieu, line 1: I'm a little concern here for the cost of fetching the metric.
You do it each 30s even if nobody is querying it.

Moreover, as it may be costly, I would have prefer that we keep this under a flag, disabled by default.
I think it would be better to have a cache, with a configuration timeout, so instead of refreshing the value unconditionaly even if not used, we only fetch it if it's stale in the cache.

You can see of how we use Caffeine cache for different things in the code base

jdbc-mysql/src/main/java/io/kestra/runner/mysql/MysqlQueue.java
```diff
@@ -28,6 +28,16 @@ protected Condition buildTypeCondition(String type) {
         return AbstractJdbcRepository.field("type").eq(type);
     }
 
+    @Override
+    protected Condition buildConsumerCondition(Class<?> queueType) {
+        return DSL.or(List.of(
+            AbstractJdbcRepository.field("consumers").isNull(),
+            AbstractJdbcRepository.field("consumers").in(
+                QUEUE_CONSUMERS.allForConsumerNotIn(queueName(queueType))
```
loicmathieu, line 36: There will always be a queueType so this is not what you need to do

loicmathieu (changes requested):
This is not a full review, @fhussonnois will do the final one.
But I think you mix consumerGroup with queueType, the worker group will be passed using a consumerGroup.

core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
ammeek: @loicmathieu I think you are right about the using  the term lag instead of unconsummated. Currently I only query for the lag of the worker queue, since we only need to spin up a new worker if there are messages to be processed in the worker queue.

jdbc-mysql/src/main/java/io/kestra/runner/mysql/MysqlQueue.java
ammeek: Thanks for the feed back @loicmathieu, I tried removing the isNull check here and it unfortunately cause the unit tests to fail. Look at the MysqlQueue.receiveFetch method I can see that it uses the same condition to fetch messages from the queue. I believe that the notNull condition is needed.

core/src/main/java/io/kestra/core/queues/WorkerQueueSizePoller.java
ammeek: thank you for the feedback on this one @loicmathieu, I've moved to using the caffine cache and I am now a lot happier with the results.

core/src/main/java/io/kestra/core/queues/QueueInterface.java
ammeek: I've replace this method with a method with the following signature.  
``` java
Integer queueLagForConsumerGroup(String consumerGroup, Class<?> queueType);
```

ammeek:
@fhussonnois and @loicmathieu this pr is now ready for review.

core/src/main/java/io/kestra/core/queues/QueueLagPoller.java
```diff
@@ -0,0 +1,100 @@
...
+                );
+            });
+    }
+
+    @PostConstruct
+    void initQueueMetrics() {
+        QueueInterface<WorkerJob> workerJobQueue = workerJobQueueProvider.get();
+        this.register(
+            getQueueLagForConsumerGroup(WORKERJOB_NAMED, null, Worker.class, workerJobQueue),
+            MetricRegistry.TAG_WORKER_GROUP, "default",
```
loicmathieu: The issue is that if a user creates a worker group with the name default which is totally acceptable, it will clash with it

core/src/main/java/io/kestra/core/queues/QueueLagPoller.java
```diff
@@ -0,0 +1,100 @@
...
+public class QueueLagPoller {
+    private final MetricRegistry metricRegistry;
+    private final WorkerGroupExecutorInterface workerGroupExecutor;
+    private final BeanProvider<QueueInterface<WorkerJob>> workerJobQueueProvider;
+
+    private final Cache<CacheKey, Integer> queueLagCache = Caffeine.newBuilder()
+        .expireAfterWrite(Duration.ofSeconds(30))
+        .build();
+
+    private final Set<String> availableWorkerGroups = new HashSet<>();
```
loicmathieu: This should be a concurrent collection

loicmathieu: But anyway, I'm not fan of maintaining the list of worker groups here, we already have that inside the WorkerGroupExecutor

loicmathieu: I will remove this set and always fetch worker groups from the MetaStore, they are in cache their anyway

jdbc-mysql/src/main/java/io/kestra/runner/mysql/MysqlQueue.java
loicmathieu: Oh yes, we remove the consumers after consumption that why in the end it's null

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
```diff
@@ -55,6 +56,7 @@ public abstract class JdbcQueue<T> implements QueueInterface<T> {
 
     private final ExecutorService poolExecutor;
     private final ExecutorService asyncPoolExecutor;
+    private final WorkerGroupExecutorInterface workerGroupExecutor;
```
loicmathieu: Seems not used

scheduler/src/test/java/io/kestra/scheduler/SchedulerConditionTest.java
```diff
@@ -114,7 +114,7 @@ void schedule() throws Exception {
             }));
 
             scheduler.run();
-            assertTrue(queueCount.await(15, TimeUnit.SECONDS));
+            assertTrue(queueCount.await(30, TimeUnit.SECONDS));
```
loicmathieu: Not a good idea, it's already too long

webserver/src/main/java/io/kestra/webserver/controllers/api/ClusterController.java
```diff
@@ -30,4 +34,18 @@ public HttpResponse<ServiceInstance> getService(@PathVariable("id") String id) t
             .map(HttpResponse::ok)
             .orElseGet(HttpResponse::notFound);
     }
-}
\ No newline at end of file
+
+    @ExecuteOn(TaskExecutors.IO)
+    @Get("/metrics/{serviceType}")
+    @Operation(tags = {"Services"}, summary = "Get metrics for running workers")
```
loicmathieu:
```suggestion
    @Operation(tags = {"Services"}, summary = "Get metrics for running services")
```

core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
```diff
@@ -271,6 +292,14 @@ public DistributionSummary findDistributionSummary(String name) {
         return this.meterRegistry.find(metricName(name)).summary();
     }
 
+    /**
+     * Remove existing Meter in the meter registry
+     * @param meter The meter to remove
+     */
+    public void removeMeter(Meter meter) {
```
loicmathieu, line 298: I'm not sure it's a good practice to remove a metric.
Metrics should stay even after what produce them didn't exist anymore for post-mortem analysis

core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
```diff
@@ -145,6 +147,8 @@ public class MetricRegistry {
     public static final String TAG_STATE = "state";
     public static final String TAG_ATTEMPT_COUNT = "attempt_count";
     public static final String TAG_WORKER_GROUP = "worker_group";
+    public static final String TAG_QUEUE_NAME = "queue_name";
+    public static final String SERVICE_ID = "service_id";
```
loicmathieu: Service ID are ephemeral, this is not a good idea to add them as tag as it would lead to metrics explosion

worker/src/main/java/io/kestra/worker/DefaultWorker.java
```diff
@@ -202,7 +202,18 @@ public DefaultWorker(
...
         if (this.init.compareAndSet(false, true)) {
-            String[] tags = this.workerGroup == null ? new String[0] : new String[]{MetricRegistry.TAG_WORKER_GROUP, this.workerGroup};
+            List<String> tagList = new ArrayList<>();
+            tagList.add(MetricRegistry.TAG_WORKER_GROUP);
+            tagList.add(this.workerGroupKey != null ? this.workerGroupKey : "default");
+
+            if (this.id != null) {
+                tagList.add(MetricRegistry.SERVICE_ID);
+                tagList.add(this.id);
+            }
```
loicmathieu: This is not a good idea to add this tag

worker/src/main/java/io/kestra/worker/DefaultWorker.java
loicmathieu, line 1: Better to revert these changes as adding an ephemeral ID to metrics is not a good idea

core/src/main/java/io/kestra/core/queues/QueueLagPoller.java
```diff
@@ -0,0 +1,100 @@
...
+        WorkerGroupExecutorInterface workerGroupExecutor,
+        BeanProvider<QueueInterface<WorkerJob>> workerJobQueueProvider
+    ) {
+        this.metricRegistry = metricRegistry;
+        this.workerJobQueueProvider = workerJobQueueProvider;
+        this.workerGroupExecutor = workerGroupExecutor;
+    }
+
+
+    @Scheduled(fixedDelay = "30000s", initialDelay = "30s")
```
loicmathieu: You refresh each 8h?

webserver/src/main/java/io/kestra/webserver/services/SharedServiceInstanceMetricService.java
```diff
@@ -0,0 +1,133 @@
...
+            .forEach(metric -> {
+                    List<Metric.Tag> metricTags = new ArrayList<>(metric.tags());
+                    List<String> tags = new ArrayList<>(metric.tags().stream().map(
+                        (tag) -> List.of(tag.key(), tag.value())
+                    ).flatMap(List::stream).toList());
+
+                    if (metric.tags().stream().map(Metric.Tag::key).noneMatch(key -> key.equals("instance_id"))) {
+                        tags.add(MetricRegistry.SERVICE_ID);
+                        tags.add(serviceInstance.uid());
+                        metricTags.add(new Metric.Tag(MetricRegistry.SERVICE_ID, serviceInstance.uid()));
```
loicmathieu: Not a good idea, this would lead to metrics explosion

webserver/src/main/java/io/kestra/webserver/services/SharedServiceInstanceMetricService.java
```diff
@@ -0,0 +1,133 @@
...
+                        metric.name(),
+                        List.copyOf(metricTags.stream().map(Metric.Tag::key).toList()),
+                        serviceInstance.uid()
+                    );
+
+                    sharedMetricsValues.computeIfAbsent(
+                        metricKey, k -> new AtomicReference<>()
+                    ).set(metric.value());
+
+                    String filteredName = metric.name().replaceAll("^kestra\\.", "");
```
loicmathieu: Why are you doing that?

webserver/src/main/java/io/kestra/webserver/services/SharedServiceInstanceMetricService.java
```diff
@@ -0,0 +1,133 @@
...
+
+    private void cleanUp(List<ServiceInstance> serviceInstances) {
+        List<String> serviceInstanceIds = serviceInstances.stream().map(ServiceInstance::uid).toList();
+        List<MetricKey> toRemove = sharedMetricsValues.keySet().stream()
+            .filter(metricKey -> !serviceInstanceIds.contains(metricKey.serviceInstanceId()))
+                .toList();
+
+        toRemove.forEach(metricKey -> {
+            log.debug("Removing metric {} from shared metrics, as the associated service instance is no longer active", metricKey);
+            metricRegistry.removeMeter(sharedMetricsGauges.remove(metricKey));
```
loicmathieu: This is not a good idea, metrics should never be removed

webserver/src/main/java/io/kestra/webserver/services/SharedServiceInstanceMetricService.java
```diff
@@ -0,0 +1,133 @@
...
+public class SharedServiceInstanceMetricService {
+    private final ServiceType serverType;
+
+    private final MetricConfig metricConfig;
+
+    private final ServiceInstanceRepositoryInterface serviceInstanceRepository;
+
+    private final Map<MetricKey, AtomicReference<Number>> sharedMetricsValues = new HashMap<>();
+
+    private final Map<MetricKey, io.micrometer.core.instrument.Gauge> sharedMetricsGauges = new HashMap<>();
```
loicmathieu: They are set and used by different threads so you need to use concurrent collections

loicmathieu (changes requested):

webserver/src/main/java/io/kestra/webserver/services/SharedServiceInstanceMetricService.java
ammeek: @loicmathieu in the service instance table metrics are stored in the following format with the kestra prefix
```
kestra.metric.name
```
This means if we push the above value directly into the web servers metric registry then the resulting name will have two kestra prefixes e.g..

```
kestra.kestra.metric.name
```

webserver/src/main/java/io/kestra/webserver/services/SharedServiceInstanceMetricService.java
ammeek: Thanks @loicmathieu what do you think the idea behaviour here should be when there are no service instance pushing a metric for example when all worker are scaled to zero? Should the metric value reset to zero?

core/src/main/java/io/kestra/core/queues/QueueLagPoller.java
ammeek: @loicmathieu This seems like a valid issue. Unfortunately we are currently using the Micronaut management metric endpoint to fetch this value for the scaling work. The metrics endpoint only supports filtering for specific tag value not metrics without values. This means if we remove this value we won't be able to tell the difference between the cumulative metric value for all tags and the metric value for workers that are not inside of a specifically designed worker group. I can see three ways to resolve this issue, I've ordered them in preference from top to bottom:

- Add a new endpoint which will allow us to filter metric values with grater granularity. If we where to do this I would suggest that we make it match the k8 metric provider spec as we can then expose the values as external metrics which could be useful in the future.
- instead of defaulting to 'default' for as the tag value for workers that are not inside of a specifically designed worker group we could instead use the string value of 'null' this is significantly less likely to  picked by a end user.
- We could add a restriction to the names allow for worker groups to stop people from picking default. This solution seem the least appealing as it add a restriction to the user and decreases UX.

webserver/src/main/java/io/kestra/webserver/services/SharedServiceInstanceMetricService.java
loicmathieu: This is configurable, you must get the configured prefix from `kestra.metrics.prefix` and do a substring

webserver/src/main/java/io/kestra/webserver/services/SharedServiceInstanceMetricService.java
loicmathieu: Yes, setting the gauge to 0 will work

core/src/main/java/io/kestra/core/queues/QueueLagPoller.java
loicmathieu: We use `(default)` in a bunch of other places, it may be a good idea to use this.
Unfortunately, we didn't add validation to the Worker group key so we accept any character.

My point is `default` has a high risk of being used, `null` is a bit akward so something else you be found that use special char so it has less risk to be used by a user.

`(default)` as when we log a worker group with no key, `<default>` of `__default__` all work for me but not just `default`.

webserver/src/main/java/io/kestra/webserver/services/SharedServiceInstanceMetricService.java
```diff
@@ -0,0 +1,110 @@
...
+                Collectors.summingDouble((metric) -> metric.value().doubleValue())
+            )
+       );
+
+        sumedServiceInstanceMetrics.forEach((metricKey, value) -> {
+            sharedMetricsValues.computeIfAbsent(metricKey, k -> new AtomicReference<>()).set(value);
+
+            List<String> tags = new ArrayList<>(metricKey.tags().stream().map(
+                        (tag) -> List.of(tag.key(), tag.value())
+                    ).flatMap(List::stream).toList());
```
loicmathieu: Cloning the map seems un-necessary

loicmathieu (approved):
LGTM

PR #14505, opened 2026-02-09
Worker 2

fhussonnois:
Follow-up: https://github.com/kestra-io/kestra-ee/issues/6437


core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
```diff
@@ -214,11 +216,26 @@ public Counter counter(String name, String description, String... tags) {
...
+     *
+     * @param name   Name of the gauge being registered.
+     * @param description The metric description
+     * @param supplier A function that yields a double value for the gauge.
+     * @param tags   Sequence of dimensions for breaking down the name.
+     * @param <T>    The type of the number from which the gauge value is extracted.
+     * @return The number that was passed in so the registration can be done as part of an assignment
+     * statement.
+     */
+    public <T extends Number> Gauge registerGauge(String name, String description, Supplier<T> supplier, String... tags) {
```
loicmathieu:
```suggestion
    public <T extends Number> Gauge gauge(String name, String description, Supplier<T> supplier, String... tags) {
```

We never use the term `register` so I'll change the name to just gauge, but it's not a strong opinion, just a proposal.

core/src/main/java/io/kestra/core/models/ServerType.java
```diff
@@ -1,10 +1,11 @@
...
 
-public enum  ServerType {
+public enum ServerType {
     EXECUTOR,
     INDEXER,
     SCHEDULER,
     STANDALONE,
     WEBSERVER,
     WORKER,
+    WORKER_AGENT,
```
loicmathieu: Isn't it named a Controller? So why not `WORKER_CONTROLLER` ?

core/src/main/java/io/kestra/core/runners/WorkerJobEvent.java
```diff
@@ -0,0 +1,91 @@
...
+ * This allows WorkerJob instances to be dispatched via the KeyedDispatchQueueInterface,
+ * with the worker group key used for routing to specific worker groups.
+ * <p>
+ * The key is the worker group key (null or empty string for the default group).
+ *
+ * @param workerGroupKey The worker group key for routing. Null or empty string means the default worker group.
+ * @param job            The actual worker job payload.
+ *
+ */
+public record WorkerJobEvent(
```
loicmathieu, line 19: For all the queue v2 migration, I directly modified the existing message instead of creating a wrapper event like here.
I'm not against it, but I strongly prefer consistency so I wonder if it would not be better to modify the existing WorkerJob to contains the workerGroupKey and implement the necessary interfaces

core/src/main/java/io/kestra/core/runners/WorkerJobEvent.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    @Override
+    public String uid() {
+        return job.uid();
+    }
+
+    /**
+     * Normalizes worker group key: null and empty string both represent the default group.
+     */
+    private static String normalizeWorkerGroup(@Nullable String workerGroup) {
```
loicmathieu, line 85: This seems like something is not correct somewhere so we ends up with empty strings instead of null

executor/src/main/java/io/kestra/executor/DefaultExecutor.java
```diff
@@ -714,8 +696,8 @@ private void toExecution(ExecutorContext executor, boolean ignoreFailure) {
                     List<String> taskRunKeys = executor.getExecution().getTaskRunList().stream()
                         .map(taskRun -> taskRun.getId())
                         .toList();
-                    workerTaskResultQueue.deleteByKeys(taskRunKeys);
-                    workerJobQueue.deleteByKeys(taskRunKeys);
+                    // workerTaskResultQueue.deleteByKeys(taskRunKeys); TODO
+                    // workerJobQueue.deleteByKeys(taskRunKeys);  TODO
```
loicmathieu: You can remove those lines, they are no longuer necessary as messages are droped from the queue immediatly (except for dispatch which is an open point)

worker-controller/src/main/java/io/kestra/controller/config/ControllerConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+    Duration maxConnectionAge
+) {
+    public static final int DEFAULT_GRPC_PORT = 9096;
+    public static final String DEFAULT_GRPC_PORT_STRING = "9096";
+
+    /**
+     * Default max connection age: 30 minutes.
+     * This provides a reasonable balance between connection stability and load redistribution.
+     */
+    public static final String DEFAULT_MAX_CONNECTION_AGE_STRING = "PT30M";
```
loicmathieu: Maybe one hour?
I fear for benchmark use case which usually takes approx 30mn to be distrurbed by a rebalance...

worker-controller/src/main/java/io/kestra/controller/config/GrpcChannelConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @param maxRetryAttempts Specifies the maximum number of retry attempts for a gRPC call.
+ * @param keepAliveTime Defines the duration for gRPC connection keep-alive.
+ * @param shutdownTimeout Defines the maximum time to wait for graceful channel shutdown.
+ */
+@ConfigurationProperties("kestra.grpc.channel")
+public record GrpcChannelConfiguration(
+    @Bindable(defaultValue = "10")
+    int maxRetryAttempts,
+    @Bindable(defaultValue = "1h")
```
loicmathieu, line 23: Shouldn't it be aligned with the reconnection which is at 30mn? (and well, I propose to increase it to 1h)

worker-controller/src/main/java/io/kestra/controller/config/WorkerControllersConfiguration.java
```diff
@@ -0,0 +1,164 @@
...
+        String hostname,
+
+        @Bindable(defaultValue = DEFAULT_GRPC_PORT_STRING)
+        int defaultPort,
+
+        @Bindable(defaultValue = "SRV")
+        DnsRecordType recordType,
+
+        @Bindable(defaultValue = "PT30S")
+        Duration refreshInterval
```
loicmathieu, line 115: DNS refresh is usually seen as a security risk as there can be DNS attacks (not sure what's the official name). That's why for ex Java never refresh DNS during a JVM life.

worker-controller/src/main/java/io/kestra/controller/config/WorkerControllersConfiguration.java
```diff
@@ -0,0 +1,164 @@
...
+         */
+        public enum Policy {
+            /**
+             * Round-robin load balancing (default gRPC policy).
+             */
+            ROUND_ROBIN,
+            /**
+             * Pick first available endpoint.
+             */
+            PICK_FIRST
```
loicmathieu: I don't think it's really a load balancing policy that is useful.

As load balancing is a very complex thing, I wonder if we should not only have round robin at first and see later what we can do (using least used worker seems the best: check metrics and select the one that has the more thread avaible or something like that).

worker-controller/src/main/java/io/kestra/controller/grpc/services/GrpcConnectControllerService.java
```diff
@@ -0,0 +1,62 @@
...
+     * {@link WorkerGroupService}.
+     *
+     * @param request          the connect request containing the worker group key
+     * @param responseObserver the response observer to send the resolved worker group
+     */
+    @Override
+    public void connect(ConnectRequest request, StreamObserver<ConnectResponse> responseObserver) {
+        String workerGroupKey = request.getWorkerGroupKey();
+
+        log.debug("Worker connect request received with workerGroupKey: {}", workerGroupKey);
```
loicmathieu: Maybe log in INFO such important things

worker-controller/src/main/java/io/kestra/controller/grpc/services/GrpcConnectControllerService.java
```diff
@@ -0,0 +1,62 @@
...
+
+        String resolvedWorkerGroup = workerGroupService.resolveGroupFromKey(workerGroupKey);
+
+        if (resolvedWorkerGroup != null && !resolvedWorkerGroup.isEmpty()) {
+            log.info("Worker group resolved: '{}' for key '{}'", resolvedWorkerGroup, workerGroupKey);
```
loicmathieu: Here I will log in debug as it's a detail which worker group is resolved.

worker-controller/src/main/java/io/kestra/controller/grpc/services/GrpcWorkerControllerService.java
```diff
@@ -0,0 +1,241 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
+
+@Singleton
+@Slf4j
+public class GrpcWorkerControllerService extends WorkerControllerServiceGrpc.WorkerControllerServiceImplBase implements WorkerControllerService {
+
+    // QUEUES
```
loicmathieu: Please remove any unecessary comments

worker-controller/src/main/java/io/kestra/controller/grpc/services/GrpcWorkerControllerService.java
```diff
@@ -0,0 +1,241 @@
...
+                    // Create context for this worker stream
+                    WorkerStreamContext<WorkerJobResponse> context = new WorkerStreamContext<>(workerId, workerGroup, maxConcurrency, responseObserver);
+                    contextRef.set(context);
+
+                    // Register with dispatcher
+                    workerJobDispatcher.registerWorker(context);
+                    initialized = true;
+                }
+
+                WorkerStreamContext<WorkerJobResponse> context = contextRef.get();
```
loicmathieu, line 120: If I understand it, here context should never be null, but the code is not straighforward.

Maybe a if(context ==null) here would be great to log a warning for ex saying that the message will be ignored.
Or throw if this should never happen...

worker-controller/src/main/java/io/kestra/controller/grpc/services/GrpcWorkerControllerService.java
```diff
@@ -0,0 +1,241 @@
...
+
+    @Override
+    public void sendWorkerTaskResults(OpaqueData request, StreamObserver<OpaqueData> responseObserver) {
+        final MessageFormat messageFormat = MessageFormat.resolve(request.getHeader().getMessageFormat());
+        BatchMessage<WorkerTaskResult> message = messageFormat.fromByteString(request.getMessage(), TypeReferences.WORKER_TASK_RESULT);
+        message.records().forEach(workerTaskResult -> {
+            try {
+                workerTaskResultQueue.emit(workerTaskResult);
+            } catch (QueueException e) {
+                throw new RuntimeException(e);
```
loicmathieu, line 172: Just for my comprehension (and maybe add this as a comment), would this fail the gRPC publisher?
In this case would be better to log and call KestraContext.shutdown

Today, if we cannot emit a worker task result we log an error and that's all

worker-controller/src/main/java/io/kestra/controller/grpc/services/GrpcWorkerControllerService.java
```diff
@@ -0,0 +1,241 @@
...
+        message.records().forEach(workerTriggerResult -> {
+            // Get if an Execution is attached to the TriggerResult.
+            Execution execution = workerTriggerResult.execution();
+            if (execution != null) {
+                execution = execution.withTenantId(workerTriggerResult.id().getTenantId());
+            }
+
+            switch (workerTriggerResult.type()) {
+                case POLLING -> triggerEventQueue.send(new TriggerEvaluated(workerTriggerResult.id(), execution));
+                case REALTIME -> triggerExecutionPublisher.send(execution);
```
loicmathieu, line 192: Same here, what will happen if the queue throws an error?

worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerStreamContext.java
```diff
@@ -0,0 +1,193 @@
...
+     * <p>
+     * This is used for controller-side memory cleanup only.
+     * The actual job state is persisted in WorkerJobRunningStateStore before sending.
+     */
+    private final ConcurrentHashMap<String, PendingJob> inFlightJobs = new ConcurrentHashMap<>();
+
+    /**
+     * Timestamp of last activity (permit request or ACK) from this worker.
+     */
+    private volatile Instant lastActivity;
```
loicmathieu, line 54: I'll use an AtomicReference here as you can have 2 threads that update it concurrently.
But if a race is OK, you can keep it this was

worker-controller/src/main/java/io/kestra/controller/grpc/services/GrpcWorkerFlowMetaStoreService.java
```diff
@@ -0,0 +1,50 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * gRPC service implementation for worker meta store operations.
+ * Provides namespace, tenant, and flow metadata to workers via gRPC.
+ */
+@Singleton
+@Requires(property = "kestra.server-type", pattern = "(CONTROLLER|STANDALONE)")
+public class GrpcWorkerFlowMetaStoreService extends WorkerFlowMetaStoreServiceGrpc.WorkerFlowMetaStoreServiceImplBase implements WorkerControllerService {
```
loicmathieu: As it's for the controller, I would have use `Controller` somewhere in the name as I assume there is a corresponding meta store inside the Worker right?

worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
```diff
@@ -0,0 +1,531 @@
...
+     * If this was the last worker for the group, disposes the subscription immediately.
+     *
+     * @param workerId the worker's unique identifier
+     */
+    public void unregisterWorker(String workerId) {
+        Objects.requireNonNull(workerId, "workerId must not be null");
+
+        WorkerStreamContext<WorkerJobResponse> context = activeStreams.get(workerId);
+        if (context == null) {
+            return;
```
loicmathieu, line 165: Is it supposed to happen?.
Maybe log a warn if not, a debug if it is

worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
```diff
@@ -0,0 +1,531 @@
...
+            } else if (!hasAnyPermitsInGroup(workerGroup)) {
+                // Workers exist but no permits - pause
+                pauseSubscription(state, workerGroup);
+            }
+        } finally {
+            state.lock.unlock();
+        }
+
+        if (subscriberToClose != null) {
+            closeSubscriberQuietly(subscriberToClose, workerGroup);
```
loicmathieu, line 210: Why don't you close it on line 199 just after getting it to avoid storing it in a temporary variables and doing it lately?

worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
```diff
@@ -0,0 +1,531 @@
...
+        WorkerJob job = event.job();
+
+        GroupState state = groupStates.get(workerGroup);
+        if (state == null) {
+            log.error("No state for worker group '{}', re-queuing job {}", WorkerGroup.forLog(workerGroup), job.uid());
+            requeue(event);
+            return;
+        }
+
+        state.lock.lock();
```
loicmathieu, line 290: Why is it done inside the GroupState lock?
It didn't update the group set so it should be lock free right?

loicmathieu: In fact, the lock is only needed when you don't have anymore permits so its scope should be reduced so on the happy path, you don't lock

worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
```diff
@@ -0,0 +1,531 @@
...
+        }
+    }
+
+    /**
+     * Persists a job to the state store for recovery.
+     */
+    private void persistJobToStateStore(WorkerStreamContext<WorkerJobResponse> context, WorkerJob job) {
+        WorkerInstance workerInstance = new WorkerInstance(context.getWorkerId(), context.getWorkerGroup());
+        if (job instanceof WorkerTask workerTask) {
+            workerJobRunningStateStore.save(NoTransactionContext.INSTANCE, WorkerTaskRunning.of(workerTask, workerInstance));
```
loicmathieu, line 389: We previously store the running job inside the same transaction for queue that support it which is no longuer the case.
Can you open a point in a followup issue so we have a look at it later? If we don't want to support this anymore, we should just remove the TransactionContext for the method parameters.

worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
```diff
@@ -0,0 +1,531 @@
...
+
+        // Delete from state store
+        workerJobRunningStateStore.deleteByKey(NoTransactionContext.INSTANCE, job.uid());
+
+        // Re-queue the job
+        requeue(originalEvent);
+    }
+
+    /**
+     * Re-queues a job event back to the queue.
```
loicmathieu, line 415: This is a bit concerning as it breaks FIFO entirely, if there are a lot of pending jobs, this means the message will go back **at the end of the queue** potentially delaying a lot the execution

loicmathieu: Ideally, we should be able to nack a message so it's going back to the queue for this use case.
Maybe open a followup issue about it?

worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
```diff
@@ -0,0 +1,531 @@
...
+            return; // Already closed
+        }
+
+        log.info("Closing WorkerJobDispatcher with {} active workers", activeStreams.size());
+
+        // Close all queue subscriptions
+        groupStates.forEach((group, state) -> {
+            state.lock.lock();
+            try {
+                // Mark as closing - no new operations
```
loicmathieu, line 466: I don't understand why you lock/unlock and didn't close the suscriber inside the lock, this seems useless

loicmathieu: Or is it just a way to be sure that no other pending operations occurs?
In this case better to write a comment explaing that.
But still, why don't you close it inside the lock?

worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
loicmathieu, line 1: I have a few questions here:
- Can this component be started more than once? If not, it should be enforced in some way
- I have some concern of pausing/resuming very often a queue subscription, if I understand it correctly, this is to avoid an in-memory queue and never receive a message that we cannot send to a worker. This may impact performance as queue may take some time to transit from one state to the other. Moreover, as most queue implementation batch read messages even if you consumer them one by one it may not be working properly...

worker-controller/src/main/java/io/kestra/controller/GrpcChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+     */
+    private ManagedChannelBuilder<?> createStaticChannelBuilder() {
+        WorkerControllersConfiguration.StaticConfig staticConfig = controllersConfig.staticConfig();
+        if (staticConfig == null || staticConfig.endpoints() == null || staticConfig.endpoints().isEmpty()) {
+            throw new IllegalStateException("Static configuration requires at least one endpoint");
+        }
+
+        List<EquivalentAddressGroup> addresses = staticConfig.endpoints().stream()
+            .map(e -> {
+                log.debug("Adding static controller endpoint: {}:{}", e.host(), e.port());
```
loicmathieu, line 124: Isn't it worker endpoint and not controller?

worker-controller/src/main/java/io/kestra/controller/GrpcChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+            .userAgent(getUserAgent())
+            .keepAliveTime(grpcChannelConfiguration.keepAliveTime().toSeconds(), TimeUnit.SECONDS)
+            .keepAliveWithoutCalls(true)
+            .executor(sharedExecutorService);
+
+        // Configure load balancing policy
+        String loadBalancingPolicy = switch (controllersConfig.loadBalancing().policy()) {
+            case ROUND_ROBIN -> "round_robin";
+            case PICK_FIRST -> "pick_first";
+        };
```
loicmathieu: Can be a property of the enum

worker-controller/src/main/java/io/kestra/controller/GrpcChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+        Map<String, Object> config = new HashMap<>();
+        Map<String, Object> serviceMap = new HashMap<>();
+
+        config.put("healthCheckConfig", serviceMap);
+        serviceMap.put("serviceName", ""); // The empty string ("") service, signifying the health of the whole server
+        return config;
+    }
+
+    @PreDestroy
+    public void close() {
```
loicmathieu, line 197: If miss a closing log in INFO

worker/src/main/java/io/kestra/worker/fetchers/WorkerJobFetcher.java
```diff
@@ -0,0 +1,318 @@
...
+    public synchronized void init(final WorkerContext workerContext) {
+        this.workerJobQueue = workerQueueRegistry.getOrCreate(workerContext, WorkerJob.class);
+        this.workerContext = workerContext;
+    }
+
+    /**
+     * {@inheritDoc}
+     **/
+    @Override
+    protected void doOnLoop() throws Exception {
```
loicmathieu, line 111: I'm not sure I fully understand what it does.
It seems like a while(true) waiting indifinitly but awaken each 100ms for permit update

worker/src/main/java/io/kestra/worker/queues/InMemoryWorkerQueue.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void put(T event) {
+        try {
+            this.queue.put(event);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
```
loicmathieu: You should restore the interupt

worker/src/main/java/io/kestra/worker/queues/MonitoredWorkerQueue.java
```diff
@@ -0,0 +1,59 @@
...
+ * Decorate a queue with monitoring capabilities.
+ *
+ * @param <T>
+ */
+public class MonitoredWorkerQueue<T> extends AbstractDelegateWorkerQueue<T> {
+
+    public static final String QUEUE_SIZE = "queue.size";
+    public static final String QUEUE_REMAINING_CAPACITY = "queue.remaining.capacity";
+    public static final String QUEUE_ENQUEUED = "queue.enqueued";
+    public static final String QUEUE_DEQUEUED = "queue.dequeued";
```
loicmathieu: All these must be prefixed by `worker` as they are worker queue specific metrics

worker/src/main/java/io/kestra/worker/senders/WorkerIOSenderFactory.java
loicmathieu, line 1: It misses assets...

worker/src/main/java/io/kestra/worker/services/GrpcWorkerConnectionService.java
loicmathieu, line 1: Maybe log in INFO?

worker/src/main/java/io/kestra/worker/processors/WorkerTriggerProcessor.java
loicmathieu, line 1: I assume this is basically the old worker with only the code related to triggers so I didn't review it.

worker/src/main/java/io/kestra/worker/processors/WorkerTaskProcessor.java
loicmathieu, line 1: I assume this is the old Worker code for only tasks so I didn'ty review it

worker/src/main/java/io/kestra/worker/processors/WorkerJobProcessorFactory.java
```diff
@@ -0,0 +1,84 @@
...
+    private TracerFactory tracerFactory;
+    private Tracer tracer;
+
+    @PostConstruct
+    public void init() {
+        this.tracer = tracerFactory.getTracer(Worker.class, "WORKER");
+    }
+
+    @SuppressWarnings("unchecked")
+    public <T extends WorkerJob> WorkerJobProcessor<T> create(WorkerContext context, T job) {
```
loicmathieu, line 54: I'm concern creating such big object for each task processing is costly.
Moreover we already create a callable and a bunch of other objects, this seems a bit complex.

loicmathieu (commented):
First of all, this is huge! GG!

This is a first batch of review.

I have a generic concern about the Woker job execution.
Previously, we had a single DefaultWorker instance that directly process WorkerCallable.
Now we have a WorkerAgent, that uses a WorkerConsumer, that uses a loop that created a processor that call the callable (and I think I miss one or two layers).
This overall seems very complex and I struggle to find the added values of all theses layers.
I wonder if this can be simplified a little.

worker/build.gradle
```diff
@@ -1,10 +1,13 @@
 configurations {
+    tests
     implementation.extendsFrom(micronaut)
 }
 
 dependencies {
+    // Kestra
     annotationProcessor project(':processor')
     implementation project(":core")
+    implementation project(":worker-controller")
```
loicmathieu, line 10: This didn't seems right?
I think the proto/objects should be inside a separate project or in core to avoid having a dependency to the controller inside the worker.

core/src/main/java/io/kestra/core/runners/WorkerJobEvent.java
fhussonnois: The empty string is actually coming from protobuf message.

worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerStreamContext.java
fhussonnois: It's OK to have potentially a race condition - it's only to keep track of last active worker so on race the timestamp will be probably the same.

worker-controller/src/main/java/io/kestra/controller/grpc/services/GrpcWorkerControllerService.java
fhussonnois: This point is tracked in the follow-up issue and requires some time for investigation to have a clean strategy on queue error.

worker-controller/src/main/java/io/kestra/controller/grpc/services/GrpcWorkerControllerService.java
fhussonnois: This point is tracked in the follow-up issue and requires some time for investigation to have a clean strategy on queue error.

worker-controller/src/main/java/io/kestra/controller/config/WorkerControllersConfiguration.java
fhussonnois: This subject is tracked in the follow-up issue.

worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
fhussonnois: The `WorkerJobDispatcher` is a Micronaut singleton injected into the gRPC service  - there is no need to enforce a single instance. 

Queues are not paused or resumed frequently. `WorkerJobs` are delivered over a bidirectional gRPC stream, so as long as the connection between a worker and the controller remains open; and the worker has available threads the subscription stays running and continues polling for jobs. In overload scenarios, a subscription may temporarily transition back and forth between paused and resumed states as worker capacity fluctuates but again this would mean that all workers for a specific worker group are overloaded.

I added to follow-up issue that we should have metrics from `WorkerJobDispatcher`

loicmathieu: OK, it will only do that when no threads are available in a particular subscription.
It'll be a point of attention during my benchmarks anyway.

loicmathieu (approved):
LGTM

PR #14507, opened 2026-02-10
fix: runcontext init for mock runcontext

Malaydewangan09:
After bumping `plugin-dbt` to 1.2

Error 1:
```
io.kestra.core.models.tasks.RunnableTaskException: Cannot invoke "java.util.Optional.isPresent()" because "this.secretKey" is null
```

Error 2:
```
Cannot invoke "io.kestra.core.runners.VariableRenderer.render(String, java.util.Map)" because "this.variableRenderer" is null
java.lang.NullPointerException: Cannot invoke "io.kestra.core.runners.VariableRenderer.render(String, java.util.Map)" because "this.variableRenderer" is null
```

### DIFF
--- a/tests/src/main/java/io/kestra/core/utils/TestsUtils.java
+++ b/tests/src/main/java/io/kestra/core/utils/TestsUtils.java
@@ -234,7 +234,11 @@ abstract public class TestsUtils {
         Execution execution = TestsUtils.mockExecution(flow, inputs, null);
         TaskRun taskRun = TestsUtils.mockTaskRun(execution, task);
 
-        return runContextFactory.of(flow, task, execution, taskRun);
+        RunContext runContext = runContextFactory.of(flow, task, execution, taskRun);
+
+        runContextFactory.initializer().forExecutor((DefaultRunContext) runContext);
+
+        return runContext;
     }
 
     public static <T> Flux<T> receive(QueueInterface<T> queue) {


core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
```diff
@@ -57,7 +56,7 @@ public class DefaultRunContext extends RunContext {
     private VersionProvider version;
     private KVStoreService kvStoreService;
     private AssetManagerFactory assetManagerFactory;
-    private Optional<String> secretKey;
+    private Optional<String> secretKey =  Optional.empty();
```
loicmathieu: This should not be needed, the RunContextInitializer will init this property

loicmathieu (approved):
LGTM but I'll prefer it if you revert your changes to the DefaultRunContext as initializing the RunContext should be enought.

Worth notting, there are a lot of places in tests where we init it manually, would take advantage of using the TestUtils instead.

core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
Malaydewangan09: Reverted

Malaydewangan09:
I will merge after CI is green, and backport to 1.2 to be used by plugins.

PR #14513, opened 2026-02-10
fix(core): make fileURI resolve namespace file versions like read()

### LINKED ISSUE #14457: `fileURI()` expression not respecting namespace file versioning
### Describe the issue

`fileURI(path=...)` currently just concatenates `kestra:///<namespace>/_files/<path>` and does not account for namespace file revisions/versions. This is inconsistent with `read()`, which resolves the latest revision by default and supports an explicit version argument.

 As a result:
  - `fileURI("foo.txt")` always points to the unversioned path, even if the latest revision is stored as foo.txt.v<N>.
  - Users can easily end up with URIs that don’t match what `read("foo.txt")` would fetch, unless they manually append .v<N> to the path.

a use case might be something like:

```yaml
- id: get_desired_owners_data
   type: io.kestra.plugin.serdes.json.JsonToIon
   from: "{{ fileURI('my_file.txt')) }}
```

---

repro:

```yaml
id: test
namespace: company.team

tasks:
  - id: hello
    message: content {{ read(fileURI('my_file.json')) }}
  - id: hello2
    message: content {{ read('my_file.json') }}
```

**Slack Thread**: 💬 https://kestra-io.slack.com/archives/C03FQKXRK3K/p1770284285231839

> [!WARNING]
> One should also check other pebble functions related to files

### Environment

- Kestra Version: 1.2.3

related to https://github.com/kestra-io/kestra/issues/5617

MdTanwer:
### ✨ Description

Fixes the `fileURI()` Pebble function to resolve namespace file revisions/versions, making it consistent with the `read()` function. Previously, `fileURI("foo.txt")` would always return the unversioned path (`kestra:///namespace/_files/foo.txt`), even when the latest revision was stored as `foo.txt.v2`. Now `fileURI()` resolves to the latest version by default and supports an explicit `version` parameter, ensuring the returned URI matches what `read()` would use.

**Changes:**
- `fileURI()` now resolves the latest version when no version is specified
- Added optional `version` parameter support (e.g., `fileURI('file.txt', version=2)`)
- Validates that specified versions exist before returning URIs
- Maintains backward compatibility for non-existent files (returns unversioned URI)

### 🔗 Related Issue

This addresses the inconsistency reported in the Slack thread: https://kestra-io.slack.com/archives/C03FQKXRK3K/p1770284285231839

**Use case example:**
- id: get_desired_owners_data
   type: io.kestra.plugin.serdes.json.JsonToIon
   from: "{{ fileURI('my_file.txt') }}"

Previously, this would always point to the unversioned path, even if read('my_file.txt') would fetch the latest revision.

### 🔗 Related Issue

  Closes https://github.com/kestra-io/kestra/issues/14457_

🛠️ Backend Checklist
[x] Code compiles successfully and passes all checks
[x] All unit and integration tests pass

### 📝 Additional Notes

- The fix ensures fileURI() and read() are now consistent in how they handle namespace file versions
- Version 1 files don't have a .v1 suffix (only versions > 1 get the suffix), which is the expected behavior
- Other file-related Pebble functions (fileSize, fileExists, isFileEmpty) already handle versions correctly through their underlying methods
- The implementation validates version existence by attempting to access the file content, throwing a clear error if the version doesn't exist


loicmathieu (approved):
LGTM, thanks

loicmathieu:
Backported to 1.2

Issue #12320, opened 2025-10-23
Auto-delete expired keys from KV Store


anna-geller:
### Feature description

Acc. to @brian-mulier-p, his current 1.1 work redesigning the KV Store with metadata allows us now to add a scheduler to auto-delete expired KV pairs. 

Many users will appreciate it and it will avoid the need for an extra System flow https://kestra.io/blogs/release-1-0#task-to-purge-expired-kv-objects

iitzIrFan:
@anna-geller mam, I saw this issue and understood that, @brian-mulier-p is working on KV metadata caching and for auto-deletion one, my approach for this issue was to introduce a scheduled service that periodically scans for expired key-value pairs and deletes them in configurable batches, using error handling and logging to ensure safe, reliable and automated cleanup without manual intervention. 
If, I had got this issue correct I would love to make a PR for this to be reviewed and further improvements on this feature.
Thank you :)

anna-geller:
hey, sorry but this is something for Brian :) have a look at https://github.com/search?q=org%3Akestra-io+label%3A%22good+first+issue%22+is%3Aopen+no%3Aassignee&type=issues instead

Ben8t:
@nkwiatkowski I think something like this make senses and on par with some configuration we already have:

```
kestra:
  kv:
    purge-expired:
       enabled: true # default true
       interval: PT6H
```

nkwiatkowski:
Hi, I updated the yaml keys a bit to match what we already have in Kestra. Here is the configuration for the KV purge cleaner:

```
kestra:
  kv:
    purge-expired:
      enabled: true # default true
      initial-delay: PT5S # default PT6H 
      fixed-delay: PT5S # default PT6H
      batch-size: 10 # default 1000
```

PR #14517, opened 2026-02-10
feat(storage): auto delete expired kv entries

nkwiatkowski:
- fix #https://github.com/kestra-io/kestra/issues/12320


core/src/main/java/io/kestra/core/storages/kv/KVPurgeCleaner.java
```diff
@@ -0,0 +1,63 @@
...
+    private void purgeKVEntriesForTenant(String tenant) {
+        List<String> namespaces = findNamespaces(tenant);
+        for (String namespace : namespaces) {
+            try {
+                KVStore kvStore = kvStoreService.get(tenant, namespace, namespace);
+                List<KVEntry> expiredEntries = kvStore.listAll()
+                    .stream()
+                    .filter(kv -> kv.expirationDate() != null &&
+                        kv.expirationDate().isBefore(Instant.now()))
+                    .toList();
```
AcevedoR: what if either there is a lot of KV, or KVs with big V payload resulting in OOM ?

AcevedoR (changes requested):

core/src/main/java/io/kestra/core/storages/kv/KVPurgeCleaner.java
nkwiatkowski: I don't think it's a risk, kv usually are small, and this endpoint doesn't return the value, only the key and the metadata

core/src/main/java/io/kestra/core/storages/kv/KVPurgeScheduler.java
```diff
@@ -0,0 +1,63 @@
...
+import jakarta.inject.Singleton;
+import java.io.IOException;
+import java.time.Instant;
+import java.util.List;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@Requires(property = "kestra.kv.purge-expired.enabled", value = "true", defaultValue = "true")
+@Singleton
+public class KVPurgeScheduler {
```
loicmathieu:
```suggestion
public class KVPurgeCleaner {
```

Please don't call it a Scheduler as there is only one Scheduler, and you know it

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/storages/kv/KVPurgeCleaner.java
AcevedoR: Ok for the Value, but for a gigantic amount of keys there is a risk.
You just add a default configurable pagination of 1.000.000 that would do the job

loicmathieu: I'm pretty sure some users do add 1MN KV values, this is not because it's not intended to, that users will not do it.

So use a medium sized pagination, like 1000

nkwiatkowski: That's a good idea. I have implemented a better search and pagination to make sure we don't here this issue here.

core/src/main/java/io/kestra/core/storages/kv/KVPurgeCleaner.java
```diff
@@ -0,0 +1,91 @@
...
+    @Inject
+    private FlowRepositoryInterface flowRepository;
+
+    @Inject
+    private KvMetadataRepositoryInterface kvMetadataRepository;
+
+    @Value("${kestra.kv.purge-expired.batch-size:1000}")
+    private Integer batchSize;
+
+    @Scheduled(initialDelay = "${kestra.kv.purge-expired.initial-delay:PT6H}", fixedDelay = "${kestra.kv.purge-expired.fixed-delay:PT6H}")
```
loicmathieu:
```suggestion
    @Scheduled(initialDelay = "${kestra.kv.purge-expired.initial-delay:PT1H}", fixedDelay = "${kestra.kv.purge-expired.fixed-delay:PT1H}")
```

Nitpick but most of our internal purge mechanisms are using 1H delay so better to align here

loicmathieu (approved):
LGTM, but I'll let Roman how was the original reviewer to make a final approvement ;)

core/src/main/java/io/kestra/core/storages/kv/KVPurgeCleaner.java
nkwiatkowski: No problem, I will change it

AcevedoR (approved):

PR #14521, opened 2026-02-10
fix(core): inject local Worker envs into RunContext

fhussonnois:
- add dedicated RunContextSerializer to filter env variables
- re-inject local `envs` variables when building RunContext for workers

Related-to: kestra-io/kestra-ee#6622

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/RunContextSerializer.java
+++ b/core/src/main/java/io/kestra/core/runners/RunContextSerializer.java
@@ -1,0 +1,53 @@
+package io.kestra.core.runners;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+
+import java.io.IOException;
+import java.io.Serial;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Custom serializer for {@link RunContext}.
+ * <p>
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/kestra/core/runners/RunContextSerializer.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void serialize(RunContext value, JsonGenerator gen, SerializerProvider provider) throws IOException {
+
+        Map<String, Object> contextMap = new HashMap<>(3);
+        Optional.ofNullable(value.getSecretInputs()).ifPresent(items -> contextMap.put("secretInputs", items));
+        Optional.ofNullable(value.getTraceParent()).ifPresent(items -> contextMap.put("traceParent", items));
```
loicmathieu: This is a little concerning as if we add more base attributes to the runContext at some point, we would need to add them here also.
I wonder if we should not copy all variables and only remove the envs.

Otherwize, a comment inside the runContext that point here would be mandatory so we change the two places.

loicmathieu (approved):
LGMT but I'm not fan of it.

For 2.0, we should think of a better way to send variables to the Worker and to be sure a runcontext created inside the executor is preserved after re-hydratation in the worker

core/src/main/java/io/kestra/core/runners/RunContextSerializer.java
fhussonnois: Yes, we will probably have to do things differently in 2.0. For now, this was the minimal solution with the fewest possible changes.

Issue #14538, opened 2026-02-11
Purge KV from internal kv store should handle the version


nkwiatkowski:
### Describe the issue

Right now, the internal kv store ignore the version when it purge the kv. It should delete the entry for the proper version

### Environment

- Kestra Version: develop

nkwiatkowski:
merged and packported to 1.2

PR #14540, opened 2026-02-11
fix(storage): use version when purging KV entries

nkwiatkowski:
- fix #https://github.com/kestra-io/kestra/issues/14538


brian-mulier-p (approved):
Perfect, LGTM, to backport in 1.0 & 1.2

PR #14542, opened 2026-02-11
fix(docs): help homepage render semi properly

elevatebart:
- MDC does not allow self closing html items
- We remove multiline JSX components that would not be understood
- We make sure that there is always a return after an HTML/Vue object

Screenshot 2026-02-11 at 11 17 28


brian-mulier-p (approved):
LGTM

PR #14547, opened 2026-02-11
ci: add develocity buildScanUri to a tmp file

AcevedoR:
so we can use it later in CI to add it to the test report, and more


loicmathieu (approved):
LGTM

PR #14549, opened 2026-02-11
feat(system): expose some Kestra service to to run context

loicmathieu:
- Expose a curated list of services to the run context
- Expose additional services except for the Worker: this may be removed on 2.0 and is a transition to a better mechanism
- Make RunContext an interface
- Kill an execution from an ExecutionUpdatableTask

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/RunContext.java
+++ b/core/src/main/java/io/kestra/core/runners/RunContext.java
@@ -19,7 +19,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
-public abstract class RunContext implements PropertyContext {
+public interface RunContext extends PropertyContext {
 
     /**
      * Returns the trigger execution id attached to this context.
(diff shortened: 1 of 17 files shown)


core/src/main/java/io/kestra/core/runners/RunContext.java
```diff
@@ -28,118 +28,118 @@ public abstract class RunContext implements PropertyContext {
...
      *
      * Warning: this method can be called only once for an attempt.
      */
-    public abstract URI logFileURI();
+    URI logFileURI();
 
     // for serialization backward-compatibility
     @JsonIgnore
-    public abstract URI getStorageOutputPrefix();
+    URI getStorageOutputPrefix();
```
fhussonnois, line 109: I really don't know what is the purpose of this method

fhussonnois (approved):

core/src/main/java/io/kestra/core/runners/RunContext.java
loicmathieu: Being able to deserialize an old run context serialized with a storageOutputPrefix...
It's the kind of legacy dirt we have because we try to never break and have JsonDeserialization failing on unknown properties...

PR #14550, opened 2026-02-11
fix(core): runContext variable merge was not done for values, leading…

brian-mulier-p:
… to missing variables in some cases

part of https://github.com/kestra-io/kestra-ee/issues/6669


Skraye (approved):
LGTM

loicmathieu (commented):
So what's annoy me a little is that a comment in this method say that maps should be flat so you should at least remove that comment.
It's also surprising that it's used in the RunContext with maps that are by nature not flat...

brian-mulier-p:
I've removed the comment but yes it's weird as you said those maps are not flattened, idk...

PR #14559, opened 2026-02-12
fix: use URIFetcher when resolving inputfiles

Malaydewangan09:
- part-of: https://github.com/kestra-io/plugin-ansible/issues/54


core/src/main/java/io/kestra/core/models/tasks/runners/PluginUtilsService.java
```diff
@@ -203,9 +204,10 @@ private static void createInputFilesInternal(RunContext runContext, Path working
                     renderedFile = inputFiles.get(fileName);
                 }
 
-                if (renderedFile.startsWith("kestra://")) {
+                if (URIFetcher.supports(renderedFile)) {
```
fdelbrayelle:
```suggestion
                if (URIFetcher.supports(rFile)) {
```

fdelbrayelle (approved):

PR #14560, opened 2026-02-12
chore(system): set Java release flag to 21

loicmathieu:


AcevedoR (approved):
LGTM, will you need to do this as well for older releases ?

loicmathieu:
@AcevedoR no, it has only be done on 1.3.
I do switch to the recommanded way to define the Java version but we can keep old branches with the old way, not problem

PR #14563, opened 2026-02-12
fix(docs): flow validate merged methods for multipart and yaml to gen…

brian-mulier-p:
…erate proper swagger


loicmathieu (approved):
LGTM

PR #14581, opened 2026-02-12
feat(ui): replace legacy product tour with onboarding v2 

### LINKED ISSUE #9874: Make switch to use element-plus tour elements
Staring from version `2.5.0` of `element-plus` (which is already being used at Kestra), there is a native `tour` functionality.

https://element-plus.org/en-US/component/tour.html

Consider implications and cost of replacing the currently used one with this, might be more efficient.

**To be explored during cooldown before any actual work starting.**

### LINKED ISSUE #14558: Improve onboarding Product Tour to focus on first flow creation + execution
## Problem Statement  

The current onboarding Product Tour does not reliably help new users reach the core “Aha moment” in Kestra: creating, executing, and understanding a flow.

Current issues:

  - Examples were too complex too early.
  - The flow editor path was not hands-on enough (id, namespace, tasks, inputs, triggers).
  - The tour felt detached from real product usage.
  - Interaction behavior felt inconsistent and reduced trust.
  - Activation was not clearly optimized around successful execution(s).

Although the UI positions the tour as the way to “create your first flow” and explore Blueprints, the current experience does not effectively deliver on that promise.

---

## Proposed Solution  

Redesign the Product Tour as a short, hands-on walkthrough centered on building and running a minimal flow.

### Core Flow of the Tour

  1. Add flow id.
  2. Add namespace.
  3. Add first tasks entry with a Log task.
  4. Personalize the log message.
  5. Save flow.
  6. Execute flow.
  7. Check execution (Gantt + logs guidance).
  8. Click Edit flow to return to editor.
  9. Add inputs (runtime parameter).
  10. Use input in Log message.
  11. Save flow again.
  12. Execute flow again (with input value).
  13. Check execution again (logs guidance).
  14. Click Edit flow again.
  15. Add a cron trigger.

## Key Principles

  - Focus on real user actions in the real UI.
  - Keep the starter flow minimal and understandable.
  - Optimize for time-to-first-success and second-success.
  - Keep onboarding non-blocking with explicit cancel flow.
  - Ensure concept clarity in-step:
      - ID
      - Namespace
      - Tasks
      - Inputs
      - Triggers
  - Provide a clear transition path to Blueprints after completion.

proddata:
- remove legacy VueTour flow and obsolete onboarding button components

resolves https://github.com/kestra-io/kestra/issues/14558
Closes https://github.com/kestra-io/kestra/issues/9874.


MilosPaunovic (commented):
I've swiftly glanced through the code - that part looks alright so far in terms you're on point when it comes to things that need changing. Would have to give it a more thorough look once it's ready for review, especially because there's a lot of files affected by the PR.

Speaking about the behavior - I do like it very much, makes things easier for users to get the idea of what Kestra is and how it's supposed to work.

Below are some minor details I've notice during the test:

- This badge is not visible enough on dark mode 
image

- Starting the tutorial shows the dialog before the redirection to editor 
[Screencast from 2026-02-13 09-14-04.webm](https://github.com/user-attachments/assets/57c645ab-62a3-4cc4-b988-664844ed3edd)

- There is a `Complete` label at the bottom of dialog which I'm not sure what's the purpose of 
image
 Okay, I figured out after the step where I had to type something in, maybe a bit of design change where we don't show it when there's no user interaction needed.

- It was not clear to me that I should type the `id` field because the editor is not in the foreground here, but looks like it's either disabled or only in the background 
image

- There's console errors when I moved to this step 
image

- Here, a little highlighting of `Save` button might not hurt 
image

- Same goes for the next step about executing.

- Overlapping of dialogs - not good 
image

Overall, beside those little things, this was an excellent job so far @proddata! 🚀

MilosPaunovic:
And this PR will also close https://github.com/kestra-io/kestra/issues/9874, so I'll add it into description.

proddata:
> This badge is not visible enough on dark mode
✅ adjusted

> Starting the tutorial shows the dialog before the redirection to editor
[Screencast from 2026-02-13 09-14-04.webm](https://github.com/user-attachments/assets/57c645ab-62a3-4cc4-b988-664844ed3edd)
✅ should now only appear after editor loads

> There is a Complete label at the bottom of dialog which I'm not sure what's the purpose of
✅  It now only shows on steps that have a editor task

> It was not clear to me that I should type the id field because the editor is not in the foreground here, but looks like it's either disabled or only in the background

That is unfortunately the color scheme 😬 
Not sure it is easy to set the focus reliably on the editor.

> There's console errors when I moved to this step

Should be fixed

> Here, a little highlighting of Save button might not hurt

Highlights are added for all buttons

> Overlapping of dialogs - not good

Some steps have different default alignments now (i.e. in this case bottom right).

---

**Current State**


https://github.com/user-attachments/assets/0ccf3d90-35ef-4634-8991-11aed2d29295


In general I think this is good to review with maybe only small design tweaks left.

MilosPaunovic (approved):
There's still some console warnings when I first start the tour:
image


It might be a good idea that for example flow ID we use the same generator as we use for regular flow creation, so if someone already completed tour can do it once more

image


Other than these two things, I'd say we're good to merge.

PR #14583, opened 2026-02-12
fix(worker): implement execution killing via gRPC Controller-Worker stream

fhussonnois:
Add support for broadcasting ExecutionKilled events from the Controller to all connected Workers through the existing gRPC bidirectional stream.
- Extend WorkerJobResponse protobuf with killCommands field
- Add ExecutionKilledManager to centralize kill state and callbacks
- Subscribe to ExecutionKilled broadcast queue in WorkerJobDispatcher

### DIFF
--- a/worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
+++ b/worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
@@ -1,12 +1,20 @@
 package io.kestra.controller.grpc.services;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.protobuf.ByteString;
 import io.kestra.controller.grpc.WorkerJobPayload;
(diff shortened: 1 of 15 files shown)


worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
```diff
@@ -64,8 +74,13 @@
 @Slf4j
 public class WorkerJobDispatcher {
     
+    // TTL for killed execution IDs in the local cache to prevent unbounded growth
+    // This cache is used for pre-dispatch filtering of tasks belonging to killed executions avoiding unnecessary dispatch and allowing for faster kill response times.
+    private static final Duration KILLED_CACHE_TTL = Duration.ofMinutes(5);
```
loicmathieu: You use 24h in the Worker but 5mn here?
i think it would be better to align both to 24h as messages can arrives late

loicmathieu (approved):
LGMT, please align cache duration between the two this is easier to understand and have less risk of missing to kill early

PR #14620, opened 2026-02-13
Feat/queue v2 continuation

loicmathieu:


queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueCleaner.java
```diff
@@ -0,0 +1,97 @@
...
+                log.info("Cleaned {} records for the '{}' queue", deleted, queue.queueName());
+                totalDeleted.add(deleted);
+            });
+        });
+
+        return totalDeleted.longValue();
+    }
+
+    private int delete(org.jooq.Configuration configuration, Condition condition) {
+        if (configuration.dialect().family() == SQLDialect.MYSQL) {
```
fhussonnois, line 67: I really wonder if we shouldn't also perform a batch delete for Postgres, as a safety measure. batch.size could be configurable, if `-1 or 0` the no batch.

fhussonnois (approved):

queue-jdbc/src/main/java/io/kestra/queue/jdbc/client/JdbcQueueCleaner.java
loicmathieu: Postgres didn't suffer for the same issue as MySQL, don't remember why but some query are not an issue on Postgres.

PR #14625, opened 2026-02-13
fix(logs): redesign logs indexes and use single-query fetchPage on MySQL

fhussonnois:
Replace the inefficient ix_namespace_flow index where timestamp range made namespace unusable, with two proper indexes:
- ix_timestamp for general time-range queries
- ix_namespace_timestamp for namespace-scoped queries

Also align MySQL fetchPage (in MysqlRepository) with Postgres/H2 by using COUNT() OVER() window function in a single query instead of two separate queries (avoid multiple full-scans).


loicmathieu (approved):
MySQL change needs to be backed by performance results (before/after query duration on a big table) as it's like that for performance reason even if we didn't use DSL.count().over()

PR #14626, opened 2026-02-13
fix(cli): refactor Cli and version provider

fhussonnois:


cli/src/main/java/io/kestra/cli/services/DefaultStartupHook.java
loicmathieu, line 1: Why did you move it here?
It seems a bit counter-intuitive.

The main purpose of this code is, at some point, to be able to detect version change so I would prefer it to be in a more obvious place related to version as we would most probably use it in 2.0 to detect incompatibility (like not supported upgrade path)

loicmathieu (changes requested):

cli/src/main/java/io/kestra/cli/commands/plugins/PluginListCommand.java
```diff
@@ -16,16 +17,14 @@
     name = "list",
     description = "List all plugins already installed"
 )
+@Singleton
```
loicmathieu: This is stange, must really a command be a bean?

loicmathieu (changes requested):

core/src/test/java/io/kestra/core/services/VersionServiceTest.java
```diff
@@ -0,0 +1,184 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.*;
+
+@ExtendWith(MockitoExtension.class)
+class VersionServiceTest {
+
+    @Mock
+    private SettingRepositoryInterface settingRepository;
+
```
AcevedoR, line 26: We could just create a Test dedicated Fake implementation (with a simple hashmap) and use it instead of mocking. 
It could be reused a lot (in BasicAuthServiceTest for example to unflaky it)

core/src/main/java/io/kestra/core/services/VersionService.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Checks if the current instance version is stored in the settings repository and saves
+     * it if it's not present or if it differs from the software version.
+     */
+    public void maybeSaveOrUpdateInstanceVersion() {
+        Optional<String> settingVersion = getInstanceVersion();
+        final String softwareVersion = versionProvider.getVersion();
+        if (settingVersion.isEmpty() || !settingVersion.get().equals(softwareVersion)) {
+            settingRepository.save(Setting.builder()
```
AcevedoR, line 54: it would probably be a good spot to log (info or warn?) when a version existed and was changed

core/src/test/java/io/kestra/core/services/VersionServiceTest.java
fhussonnois: I'm not against it, but not in this PR because. I think providing a InMemory implementation of `SettingRepositoryInterface` is more a cooldown stuff, so we have a dedicated PR with all the changes.

AcevedoR (approved):
LGTM

AcevedoR:
followup up about the in memory settings repository:
- https://github.com/kestra-io/kestra/pull/14640

loicmathieu (approved):
LGTM

PR #14635, opened 2026-02-16
fix: make vue-tsc pass for EE as well

elevatebart:
- since `authStore.user` is nullable in EE it needs to be nullable in OSS too


AcevedoR (approved):

fdelbrayelle (approved):
LGTM

PR #14639, opened 2026-02-16
fix(core): make KVStore compatible with gRPC-based worker

fhussonnois:
- Add new gRPC service to manage KVMetadata
- Add new KVMetadataStateStore abstraction with repository and gRPC-based implementation
- Remove server-side methods from KVStore interface to exposes only worker-safe operations.
- Extract ExtendedKVStore interface to expose server-side methods
- Refactor KV core tasks

### DIFF
--- a/core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
+++ b/core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
@@ -1,32 +1,22 @@
 package io.kestra.core.storages.kv;
 
 import io.kestra.core.exceptions.ResourceExpiredException;
-import io.kestra.core.models.FetchVersion;
-import io.kestra.core.models.QueryFilter;
 import io.kestra.core.models.kv.PersistedKvMetadata;
-import io.kestra.core.repositories.ArrayListTotal;
-import io.kestra.core.repositories.KvMetadataRepositoryInterface;
+import io.kestra.core.runners.KVMetadataStateStore;
 import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.storages.StorageInterface;
 import io.kestra.core.storages.StorageObject;
-import io.kestra.core.utils.ListUtils;
-import io.micronaut.data.model.Pageable;
-import io.micronaut.data.model.Sort;
 import jakarta.annotation.Nullable;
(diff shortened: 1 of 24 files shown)


core/src/main/java/io/kestra/core/storages/kv/ExtendedKVStore.java
loicmathieu, line 1: I'm not fan of it, normally, we should exposes via the runcontext only things that can be used by runnable tasks.
Is the issue only for the PurgeKV task?
If so, I would prefer to avoid that ExtendedKVStore that uses both the metastore and the repository, and change the way the PurgeKV task is done by calling directly the repository in it as anyway it would be refactored later.

core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
```diff
@@ -35,29 +32,56 @@
  *
  */
 @Slf4j
-public class InternalKVStore implements KVStore {
+public class InternalKVStore implements ExtendedKVStore {
```
loicmathieu: It didn't feels right to have both the metastore and the repository here, there is a smell behind this.
It should only use the metastore and don't expose any methods not safe to be used on worker

loicmathieu (changes requested):
Looking at the PR, I found the existing KV stuff a bit akward.
We use the same KvStore object everywhere but it handle very different things as access to the KVStore from an execution (what it was designed for) and access KvStores from the API which should be using directly the repository IMHO.

I agree that this may need a bigger refactoring so I'll be OK to do that in a followup issue.

core/src/main/java/io/kestra/core/storages/kv/InternalKVStore.java
fhussonnois: `ExtendedKVStore` is removed

core/src/main/java/io/kestra/core/storages/kv/ExtendedKVStore.java
fhussonnois: Now, all methods not compatible with workers are moved from `KVStore` to  existing `KVStoreService`. The `PurgeTask` uses the new `DefaultRunContext#services` method to get access to it.

loicmathieu (approved):
LGTM

PR #14642, opened 2026-02-16
Feat/queue v2 misc

loicmathieu:
- Don't send trigger execution terminated event to the Scheduler for Webhook and Flow triggers.
- Introduce an AbstractPollingSubscriber, which uses a QueuePoller to poll a queue. It should be reusable in AMQP and Redis which have the same poll logic
- Remove the old JDBC queue
- Fix a bug for YAMLParser and some fialing unit tests


fhussonnois (approved):
It's much cleaner with the polling logic extracted to QueuePoller.

PR #14643, opened 2026-02-16
feat(system): store task run outputs out of the execution

loicmathieu:
See followup issue: https://github.com/kestra-io/kestra-ee/issues/6808

### DIFF
--- a/core/src/main/java/io/kestra/core/models/executions/Execution.java
+++ b/core/src/main/java/io/kestra/core/models/executions/Execution.java
@@ -7,7 +7,6 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import com.google.common.collect.ImmutableMap;
 import io.kestra.core.debug.Breakpoint;
 import io.kestra.core.exceptions.InternalException;
 import io.kestra.core.models.SoftDeletable;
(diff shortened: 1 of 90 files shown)


jdbc-mysql/src/main/resources/migrations/mysql/V3_12__outputs.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS task_outputs (
+    `key` VARCHAR(250) PRIMARY KEY,
+    `task_run_id` VARCHAR(150) NOT NULL,
+    `tenant_id` VARCHAR(150) NOT NULL,
+    `execution_id` VARCHAR(150) NOT NULL,
+    `value` TEXT,
```
fhussonnois: As you used BYTEA for Postgres, I think the best equivalent for MySQL is `LONGBLOB`

core/src/main/java/io/kestra/core/models/executions/Execution.java
```diff
@@ -8,6 +8,7 @@
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Streams;
```
fhussonnois: Unused import ?

fhussonnois (approved):

jdbc-mysql/src/main/resources/migrations/mysql/V3_12__outputs.sql
loicmathieu: `TEXT` is also stored out-of-column (columns are limited to 64k) but I wonder if switching to BLOB would not be great.
I'm not sure of the performance implication as it uses a separate API (stream based).
And in this case we would also be better to use also a BLOB in Postgres.

loicmathieu: I read the MySQL doc and you're right, the right type should be `LONGBLOB` for MySQL and most probably H2.

PR #14644, opened 2026-02-16
fix(core): add IN and NOT_IN for asset filters

brian-mulier-p:


AcevedoR (changes requested):
Please do add at least one test in the repository for these OPs
We had recently a team member explicitly asking to cover everything

fhussonnois (approved):

brian-mulier-p:
@AcevedoR @fhussonnois please have another look, I added generic tests for fields that don't have dedicated conditions on both ES & JDBC. I think it's less time-consuming and reliable and will help detect any exceptionally-conditionned fields + make sure we don't move some fields to some exceptional conditions without being aware of it.

See EE PR: https://github.com/kestra-io/kestra-ee/pull/6726

AcevedoR (approved):

AcevedoR (approved):

PR #14674, opened 2026-02-18
Revert "refactor: use the openapi output to determine the axios types (#14228)"

elevatebart:
because it breaks EE


loicmathieu (approved):
LGTM

mrlubos:
@elevatebart @loicmathieu do you need anything from my side to help out here?

PR #14676, opened 2026-02-18
refactor: use the openapi output to determine the axios types

elevatebart:
This one does not break EE auth


MilosPaunovic (approved):

MilosPaunovic (commented):
Shall we run `predev` here also, like in `EE` @elevatebart?

elevatebart:
If it feels necessary then we will.

PR #14682, opened 2026-02-18
fix(docker): register kill runnable early

Malaydewangan09:
Closes https://github.com/kestra-io/kestra-ee/issues/6382

### DIFF
--- a/core/src/main/java/io/kestra/core/exceptions/KilledException.java
+++ b/core/src/main/java/io/kestra/core/exceptions/KilledException.java
@@ -6,8 +6,7 @@ import java.io.Serial;
  * Exception thrown when a task runner is killed during execution.
  */
 public class KilledException extends KestraRuntimeException {
-    private static final String DEFAULT_MESSAGE = "Task runner was killed during execution.";
-    private static final String PHASE_MESSAGE = "Task runner was killed during %s.";
+    private static final String DEFAULT_MESSAGE = "Execution was killed.";
 
     /**
      * Creates a new {@link KilledException} with a default message.
@@ -20,10 +19,9 @@ public class KilledException extends KestraRuntimeException {
      * Creates a new {@link KilledException} with a message describing
      * the execution phase during which the kill occurred.
      *
-     * @param phase a short description of the execution phase
-     *              (e.g. "image pull", "container start").
+     * @param message the error message.
      */
-    public KilledException(String phase) {
-        super(PHASE_MESSAGE.formatted(phase));
+    public KilledException(String message) {
+        super(message);
     }
 }
--- a/core/src/main/java/io/kestra/core/exceptions/KilledException.java
+++ b/core/src/main/java/io/kestra/core/exceptions/KilledException.java
@@ -1,0 +1,29 @@
+package io.kestra.core.exceptions;
+
+import java.io.Serial;
+
+/**
+ * Exception thrown when a task runner is killed during execution.
+ */
+public class KilledException extends KestraRuntimeException {
+    private static final String DEFAULT_MESSAGE = "Task runner was killed during execution.";
+    private static final String PHASE_MESSAGE = "Task runner was killed during %s.";
+
+    /**
+     * Creates a new {@link KilledException} with a default message.
+     */
+    public KilledException() {
+        super(DEFAULT_MESSAGE);
+    }
+
+    /**
+     * Creates a new {@link KilledException} with a message describing
+     * the execution phase during which the kill occurred.
+     *
+     * @param phase a short description of the execution phase
+     *              (e.g. "image pull", "container start").
+     */
+    public KilledException(String phase) {
+        super(PHASE_MESSAGE.formatted(phase));
+    }
+}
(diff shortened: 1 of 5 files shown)


script/src/main/java/io/kestra/plugin/scripts/runner/docker/Docker.java
```diff
@@ -409,10 +419,16 @@ public TaskRunnerResult<DockerTaskRunnerDetailResult> run(RunContext runContext,
                     pullImage(dockerClient, image, renderedPolicy, logger);
                 }
 
+                if (getIsKilled().get()) {
+                    throw new IllegalStateException("Task was killed during image pull");
```
loicmathieu: Do you really need to throw an exception here?
Returning null didn't work?

I wonder if we should not move that to the parent TaskRunner as it would be usefull for other task runners?

If so:
- create a method `checkKilled()`  in TaskRunner
- throw a dedicated Exception: `KilledException` wich extends `KestraRuntimeException`

So it would be reusable.

Malaydewangan09: Okay, good call!
I'll try that.

core/src/main/java/io/kestra/core/exceptions/KilledException.java
```diff
@@ -0,0 +1,29 @@
+package io.kestra.core.exceptions;
+
+import java.io.Serial;
+
+/**
+ * Exception thrown when a task runner is killed during execution.
+ */
+public class KilledException extends KestraRuntimeException {
+    private static final String DEFAULT_MESSAGE = "Task runner was killed during execution.";
+    private static final String PHASE_MESSAGE = "Task runner was killed during %s.";
```
loicmathieu: This may be used later for other things than the task runner, can you find a more generic message?

core/src/main/java/io/kestra/core/exceptions/KilledException.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    /**
+     * Creates a new {@link KilledException} with a message describing
+     * the execution phase during which the kill occurred.
+     *
+     * @param phase a short description of the execution phase
+     *              (e.g. "image pull", "container start").
+     */
+    public KilledException(String phase) {
```
loicmathieu: Classic implementation is when you have a message, tu be able to se the full message not some part.
I would prefer that when you take a message in the constructor parameter, it would be the full message not only a part of it

script/src/main/java/io/kestra/plugin/scripts/runner/docker/Docker.java
loicmathieu, line 1: I wonder if this will not create a memory leak as the runnable capture the AtomicReference so it may not be properly garbage collected in all code path.

This goes to the right direction, but I would like to test some time to test this PR before merging to be sure.

script/src/test/java/io/kestra/plugin/scripts/runner/docker/DockerTest.java
```diff
@@ -197,8 +197,7 @@ void killAfterResume() throws Exception {
             // Assert that the container id is the same
             assertThat(resumeContainerId).isEqualTo(createContainerId);
 
-            // Kill the container and verify cleanup
-            resumeContainerThread.interrupt();
+            anotherTaskRunner.kill();
```
loicmathieu, line 200: Normally, this should also work for an interrupt.
I wonder if the best here would not be to have two test, one that interrupt and one that kill to be sure both work.

loicmathieu (changes requested):
Overall, this looks good to me.

However, I have a small concern about a possible memory leak so I would like to hold on of this one for a few days until I have the time to make more test on it.

script/src/test/java/io/kestra/plugin/scripts/runner/docker/DockerTest.java
Malaydewangan09: No, it didnt worked with interrupt.
I will try to test again.

core/src/main/java/io/kestra/core/exceptions/KilledException.java
Malaydewangan09: I thought about the same, but took some ref from other exceptions.
I'll update it.

script/src/test/java/io/kestra/plugin/scripts/runner/docker/DockerTest.java
Malaydewangan09: Yes, you were right, In the resume path, containerIdRef was never set, so interrupt didn't invoked kill.
Fixed it and now it works with interrupt and kill too.

loicmathieu: Great!

loicmathieu (changes requested):

loicmathieu (approved):
LGTM, nice PR!
Ready to be merged! I confirm it didn't cause a memory leak

Please open a followup issue to do the same for all taskrunner.

Malaydewangan09:
Merged, I'll open the followup issue once I am back from OOO!

PR #14696, opened 2026-02-18
fix(core): fix multiple race conditions in AbstractSubscriber state

fhussonnois:
Replace State enum with two independent AtomicBooleans (running, paused)
- Fix pause() threw IllegalStateException when subscriber was stopped
- Fix close() could hang forever (no timeout on stopped.await())
- Fix pause() before subscribe() was impossible (state started as STOPPED)
- Fix races between pause() and concurrent close()

Add unit-tests for AbstractSubscriber

Part-of: kestra-io/kestra-ee#6758

### DIFF
--- a/queue/src/main/java/io/kestra/queue/AbstractPollingSubscriber.java
+++ b/queue/src/main/java/io/kestra/queue/AbstractPollingSubscriber.java
@@ -17,7 +17,7 @@ import static org.awaitility.Awaitility.await;
 
 /**
  * A subscriber that uses a polling mechanism.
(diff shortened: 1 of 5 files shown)


loicmathieu (approved):
LGTM, more easy to gasp like that.

Small nitpick but you can totally dismiss them:
- We usually use the term `running` and not `active` but I agree that a subscription is active and it didn't have a run method so it may be better like that, no strong opinion
- We usually always allow to configure timeout. For ex in test we may want to lower it to avoid having hanging CI forever when things fail miserably

PR #14720, opened 2026-02-19
fix: add correct 'CANCELLED' type for Exit task while keeping compati…

### LINKED ISSUE #14719: CANCELLED vs CANCELED state mismatch in Exit plugin
I have a bug to report regarding ​`CANCELLED`​ state mistypes.

Kestra version: ​*1.2.6 *​(observed also in 1.2.5)

Description: Subflow returns state ​`CANCELLED`​, but ​`io.kestra.plugin.core.execution.Exit`​ would expect ​`CANCELED`​ state. We are trying dynamically set the state for exiting subflow, incorporating it with ​`transmitFailed: false`​ config (reasons are due to manual maintenance after any failures), and noticed this mistype in state.

Screenshots:
- Example from Subflow outputs view:

Image

- Example of ​`Exit`​ plugin description:

Image

Skraye:
…bility with previous and now deprecated 'CANCELED'

close #14719


loicmathieu (approved):
LGTM

PR #14726, opened 2026-02-19
Feat/asset triggers

brian-mulier-p:
part of https://github.com/kestra-io/kestra-ee/issues/6377

### DIFF
--- a/tests/src/main/java/io/kestra/core/runners/TestRunnerUtils.java
+++ b/tests/src/main/java/io/kestra/core/runners/TestRunnerUtils.java
@@ -4,6 +4,7 @@ import io.kestra.core.models.Label;
 import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.executions.ExecutionKilled;
 import io.kestra.core.models.executions.ExecutionKilledExecution;
+import io.kestra.core.models.executions.ExecutionKind;
 import io.kestra.core.models.flows.Flow;
 import io.kestra.core.models.flows.FlowInterface;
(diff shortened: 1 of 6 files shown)


tests/src/main/java/io/kestra/core/runners/TestRunnerUtils.java
```diff
@@ -67,9 +68,12 @@ public Execution runOne(String tenantId, String namespace, String flowId, Intege
...
 
-    public Execution runOne(String tenantId, String namespace, String flowId, Duration duration)
-        throws TimeoutException, QueueException {
-        return this.runOne(tenantId, namespace, flowId, null, null, duration, null);
+    public Execution runOne(String tenantId, String namespace, String flowId, Duration duration, ExecutionKind executionKind) throws TimeoutException, QueueException {
+        return this.runOne(tenantId, namespace, flowId, null, null, duration, null, executionKind);
+    }
+
+    public Execution runOne(String tenantId, String namespace, String flowId, Duration duration) throws TimeoutException, QueueException {
+        return this.runOne(tenantId, namespace, flowId, duration, ExecutionKind.NORMAL);
```
loicmathieu:
```suggestion
        return this.runOne(tenantId, namespace, flowId, duration, null);
```

By default it's null so better to set null here also

loicmathieu: Please align everywhere

loicmathieu (approved):
LGTM

PR #14751, opened 2026-02-23
fix(worker): add LogEntryEmitter to avoid depending on queues for logs

fhussonnois:


core/src/main/java/io/kestra/core/runners/DefaultLogEntryEmitter.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Inject
+    public DefaultLogEntryEmitter(DispatchQueueInterface<LogEntry> logQueue,
+                                  BroadcastQueueInterface<FollowLogEvent> followLogQueue) {
+        this.logQueue = logQueue;
+        this.followLogQueue = followLogQueue;
+    }
+
+    @Override
+    public void emits(LogEntry entry) {
```
loicmathieu: Maybe return a `CompletionStage` here, same for the other methdo

loicmathieu (approved):
LGTM

PR #14755, opened 2026-02-23
fix(worker): cleanup and fix WorkerTest

fhussonnois:


worker/build.gradle
```diff
@@ -20,6 +20,8 @@ dependencies {
     testImplementation project(':tests')
     testImplementation project(':jdbc')
     testImplementation project(':jdbc').sourceSets.test.output
+    testImplementation project(':queue').sourceSets.test.output
```
loicmathieu, line 23: Are you sure it's needed?
because it's a bad practice that may creates un-necessary module dependencies (which prevent usage of the Gradle cache)

loicmathieu (approved):
LGTM

worker/build.gradle
fhussonnois: Yes, otherwise I got an issue with Micronaut bean injection on queues

Issue #14187, opened 2026-01-15
Add syntax highlighting support for code properties editor


Ben8t:
Currently, code properties in the no-code editor lack syntax highlighting, which makes it difficult to read and edit code directly within the interface.

**Proposed Solution**
Add language-specific syntax highlighting for code properties, with support for common languages including:

- Python
- SQL
- JavaScript/TypeScript
- JSON
- YAML
- etc.

**Implementation Notes**
- The editor should apply appropriate syntax highlighting based on the selected language. This would improve the developer experience when working with code in the no-code interface.

- The same syntax highlighting functionality should be implemented for the global YAML editor to ensure consistency across the platform and improve readability when editing YAML configurations.

Eshaan-byte:
Hi @Ben8t can I work on this issue?

MilosPaunovic:
Related to https://github.com/kestra-io/kestra/issues/9647.

Skraye:
Update on the issue : 

Currently we already have syntax highlighting for code properties, but the way isn't very reliable

So after discussion we will add a new `language` property in the Plugins annotation to allows indicate a language for a plugin/property, and the UI will be able to pick it to define the syntax highlighting
PR for the backend part: https://github.com/kestra-io/kestra/pull/14722

Now what's left to do is modifying the plugin and the UI Part

Ben8t:
> Now what's left to do is modifying the plugin and the UI Part

@Skraye can you create a dedicated issue with @fdelbrayelle, and likely another issue for frontend (with @elevatebart ?) on this please? 
Let me know if I can help, but sounds like technical follow up issues right?

PR #14777, opened 2026-02-24
feat: use language prop from plugin in no code editor

Skraye:
relate to #14187

### DIFF
--- a/ui/src/components/no-code/components/tasks/TaskString.vue
+++ b/ui/src/components/no-code/components/tasks/TaskString.vue
@@ -83,59 +83,9 @@
 
     const pebble = ref(false);
 
-    // Function to detect programming language from task type
-    function detectLanguageFromTaskType(): string {
-        if (!props.task?.type) {
-            return "plaintext";
-        }
-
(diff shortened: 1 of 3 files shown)


ui/src/components/no-code/components/tasks/TaskString.vue
```diff
@@ -85,57 +85,12 @@
...
     // Computed property for editor language
     const editorLanguage = computed(() => {
-        // Only apply syntax highlighting for script and commands fields
-        if (props.root === "script" || props.root === "commands" ||
-            props.root?.endsWith(".script") || props.root?.endsWith(".commands")) {
-            return detectLanguageFromTaskType();
-        }
-
-        return "plaintext";
+        return detectLanguageFromTaskType();
```
MilosPaunovic: Can you simplify this by just returning the `return props.schema?.$language ?? "plaintext";` here and removing the method `detectLanguageFromTaskType`?

MilosPaunovic (changes requested):

MilosPaunovic (approved):

PR #14780, opened 2026-02-24
fix(core): fix task resubmit on worker failure

fhussonnois:
- Manage gRPC deadline per call for connection
- Refactor ServiceLivenessCoordinatorTest
 - Add defensive check for empty string workerGroupKey


worker/src/main/java/io/kestra/worker/senders/WorkerIOSenderFactory.java
```diff
@@ -20,6 +21,7 @@ public class WorkerIOSenderFactory {
      * Creates a sender for {@link WorkerTaskResult} events (sent per-item).
      */
     @Singleton
+    @Named
```
loicmathieu, line 24: This is a bit strange to name things without name... I assume it will use the name of the method

loicmathieu (approved):
LGTM

worker/src/main/java/io/kestra/worker/senders/WorkerIOSenderFactory.java
fhussonnois: That's what I thought too, but the qualifier named doesn't work without that.

PR #14832, opened 2026-02-26
fix(executions): replay cannot be done via a command

loicmathieu:
As commands are for updating the current execution, but replay creates a new one so it should be done  by sending the new execution directly to the execution queue.


fhussonnois (approved):
LGTM!

PR #14839, opened 2026-02-26
feat(core): implement SecurityIntegration ENABLED filter.

ammeek:
### ✨ Description

What does this PR change?  
Add SecurityIntegration ENABLED filter

### 🔗 Related Issue
relates to kestra-io/kestra-ee#6256

### 🛠️ Backend Checklist


AcevedoR (commented):
shouldnt SECURITY_INTEGRATION be present in the openapi spec ?

ammeek:
> shouldnt SECURITY_INTEGRATION be present in the openapi spec ?

Yea it should be, will regenerate the openapi spec.

ammeek:
@AcevedoR As mentioned [here](https://github.com/kestra-io/kestra-ee/pull/6875#issuecomment-4006286302) SECURITY_INTEGRATION should not be in the open api spec file.

AcevedoR (approved):

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
```diff
@@ -304,6 +304,10 @@ protected Condition getConditionOnField(
             return getDateCondition(value, operation, dateColumn);
         }
 
+        if (field == QueryFilter.Field.ENABLED) {
+            return getEnabledCondition(value, operation);
+        }
+
```
nkwiatkowski, line 310: I think this can be done like the other standards fields -> return defaultHandlers(field, value, operation);

ammeek: @nkwiatkowski unfortunately in the example of SecurityIntegration enabled is not a column available in the database and is stored in the value column. This means we need to unpack the json value from the column when trying to get SecurityIntegrations. The default imp of getEnabledCondition directly calls through to defaultHandlers so that by default enabled is treated like any other field.

nkwiatkowski (approved):
LGTM

PR #14849, opened 2026-02-27
chore(system): deprecate `kestra` commands in favor or `kestractl`

loicmathieu:
Closes https://github.com/kestra-io/kestra-ee/issues/6854

### DIFF
--- a/cli/src/main/java/io/kestra/cli/commands/flows/FlowValidateCommand.java
+++ b/cli/src/main/java/io/kestra/cli/commands/flows/FlowValidateCommand.java
@@ -15,6 +15,7 @@ import java.util.List;
     name = "validate",
     description = "Validate a flow"
 )
+@Deprecated(forRemoval = true, since = "1.3.0")
 public class FlowValidateCommand extends AbstractValidateCommand {
     @Inject
     private ModelValidator modelValidator;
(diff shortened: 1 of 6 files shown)


cli/src/main/java/io/kestra/cli/commands/flows/FlowValidateCommand.java
```diff
@@ -27,6 +28,8 @@ public class FlowValidateCommand extends AbstractValidateCommand {
 
     @Override
     public Integer call() throws Exception {
+        stdErr("WARNING: this command is deprecated, use `kestractl flows validate` instead");
```
AcevedoR, line 31:
```suggestion
        stdErr("WARNING: this command is deprecated, use `kestractl flows validate` instead. Unless you need to validate Flows in a offline mode (in CI for example) and don't want to rely on a running Kestra instance");
```

AcevedoR (changes requested):
doc request for flow validate

cli/src/main/java/io/kestra/cli/commands/flows/FlowValidateCommand.java
loicmathieu: I don't think we will support this anymore, it's a good question.

loicmathieu: Validated with Ben, we deprecate for removal in 2.0

AcevedoR (approved):

loicmathieu:
Backported to 1.3

PR #14854, opened 2026-02-27
chore: exclude spring-configuration-metadata.json from webserver to avoid Incremental build failures

AcevedoR:


loicmathieu (approved):
LGTM

PR #14856, opened 2026-02-27
fix: filter out deprecated definitions from plugin documentation

### LINKED ISSUE #14293: Hide Deprecated Definitions within a task docs

nancysangani:
Fixes #14293 

Deprecated definitions like `FlowCondition` and `FlowNamespaceCondition` were 
still showing in the Definitions section despite being marked as deprecated.

The API response already includes a `$deprecated: "true"` field on deprecated 
definitions. The `SchemaToHtml` component in ui-libs uses this field to hide 
deprecated properties, but the same check wasn't applied to definitions.

This fix adds a `filteredSchema` computed property in both `Plugin.vue` and 
`PluginDocumentation.vue` that filters out definitions with `$deprecated` before 
passing the schema to `SchemaToHtml` — a generic solution that works across all 
plugins, not just the ones in this issue.

**Tested:**
- Deprecated definitions no longer appear on the standalone Plugins page
- Deprecated definitions no longer appear in the in-app editor Documentation panel
- `npm run test:types` passes with no errors

Screenshot 2026-02-27 223250

## Screenshots 
Before:
image

After: only valid, non-deprecated definitions shown
Screenshot 2026-02-27 222328

### DIFF
--- a/ui/src/components/plugins/Plugin.vue
+++ b/ui/src/components/plugins/Plugin.vue
@@ -141,19 +141,9 @@
 
         const filteredDefinitions = Object.entries(schema.definitions).reduce(
             (acc, [key, value]: [string, any]) => {
-                const title: string = value?.title ?? "";
-
-                const shouldFilter =
-                    value?.deprecated === true ||
-                    value?.hidden === true ||
-                    key.includes("FlowCondition") ||
-                    key.includes("FlowNamespaceCondition") ||
-                    /condition for a (specific flow|flow namespace)/i.test(title);
-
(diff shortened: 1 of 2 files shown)


Piyush-r-bhaskar:
Closing in favour of https://github.com/kestra-io/ui-libs/pull/1157

PR #14911, opened 2026-03-03
chore: simplify openapi spec generation

AcevedoR:
forward gradle project version to micronaut, so it can generate a openapi spec with the proper version. Thanks to that updateOpenapiVersion is not needed anymore, it was causing a lot of gradle inc-build issues


loicmathieu (approved):
LGTM

PR #14922, opened 2026-03-04
feat(core): add separate purge options for execution and trigger logs in PurgeLogs

fdelbrayelle:
## Summary
- Add `purgeExecutionLogs` and `purgeNonExecutionLogs` boolean flags to the `PurgeLogs` task, allowing selective purging of execution logs (where `execution_id IS NOT NULL`) or non-execution logs like trigger logs (where `execution_id IS NULL`)
- Both flags default to `true` for full backward compatibility
- Output now includes `executionLogsCount` and `nonExecutionLogsCount` alongside the existing `count` (total)

## Changes
- **`PurgeLogs.java`** — Two new `Property<Boolean>` fields with `@Builder.Default` set to `true`; updated `Output` with separate counts
- **`ExecutionLogService.java`** — Updated `purge()` method returning `int[]` with per-category counts
- **`LogRepositoryInterface.java`** — Added `purgeExecutionLogs`/`purgeNonExecutionLogs` params to `deleteByQuery`; old 7-param signature kept as default method for backward compatibility
- **`AbstractJdbcLogRepository.java`** — Filters on `execution_id IS NOT NULL` (execution logs) / `execution_id IS NULL` (non-execution logs) based on flags
- **Tests** — 2 new test cases + 2 new test flow YAMLs

### Default behavior (purge all — backward compatible)

```yaml
id: qa_purge_all
namespace: company.team

tasks:
  - id: purge_logs
    type: io.kestra.plugin.core.log.PurgeLogs
    endDate: "{{ now() | dateAdd(1, 'HOURS') }}"
```

**Expected:** Both execution and non-execution logs are deleted. \`count\` = total, \`executionLogsCount\` + \`nonExecutionLogsCount\` = \`count\`.

### Purge only execution logs

```yaml
id: qa_purge_exec_only
namespace: company.team

tasks:
  - id: purge_logs
    type: io.kestra.plugin.core.log.PurgeLogs
    endDate: "{{ now() | dateAdd(1, 'HOURS') }}"
    purgeExecutionLogs: true
    purgeNonExecutionLogs: false
```

**Expected:** Only execution logs are deleted. \`nonExecutionLogsCount\` = 0. Non-execution logs (triggers, flow-level) remain in DB.

### Purge only non-execution logs (triggers, flow-level)

```yaml
id: qa_purge_non_exec_only
namespace: company.team

tasks:
  - id: purge_logs
    type: io.kestra.plugin.core.log.PurgeLogs
    endDate: "{{ now() | dateAdd(1, 'HOURS') }}"
    purgeExecutionLogs: false
    purgeNonExecutionLogs: true
```

**Expected:** Only non-execution logs are deleted. \`executionLogsCount\` = 0. Execution logs remain in DB.

### Purge neither (no-op)

```yaml
id: qa_purge_none
namespace: company.team

tasks:
  - id: purge_logs
    type: io.kestra.plugin.core.log.PurgeLogs
    endDate: "{{ now() | dateAdd(1, 'HOURS') }}"
    purgeExecutionLogs: false
    purgeNonExecutionLogs: false
```

**Expected:** No logs are deleted. All counts = 0.

## Test plan
- [x] Existing \`PurgeLogsTest\` tests pass (no regression)
- [x] New test: purge only execution logs — verifies \`executionLogsCount > 0\`, \`nonExecutionLogsCount == 0\`
- [x] New test: purge only non-execution logs — verifies \`executionLogsCount == 0\`, \`nonExecutionLogsCount > 0\`

part-of https://app.usepylon.com/support/issues/views/36469f3c-64a7-47dc-8e82-9d36026bfc31?issueNumber=1425&view=fs

fixes https://github.com/kestra-io/kestra-ee/issues/6938

### DIFF
--- a/core/src/main/java/io/kestra/core/services/ExecutionLogService.java
+++ b/core/src/main/java/io/kestra/core/services/ExecutionLogService.java
@@ -28,24 +28,32 @@ public class ExecutionLogService {
         this.logRepository = logRepository;
     }
 
+    public record PurgeResult(int executionLogsDeleted, int nonExecutionLogsDeleted) {}
+
     /**
      * Purges log entries matching the given criteria, with separate control over execution and non-execution logs.
      *
-     * @return an array of two ints: [executionLogsDeleted, nonExecutionLogsDeleted]
+     * @param tenantId    the tenant identifier
+     * @param namespace   the namespace of the flow
+     * @param flowId      the flow identifier
+     * @param executionId the execution identifier
+     * @param logLevels   the list of log levels to delete
+     * @param startDate   the start of the date range
+     * @param endDate     the end of the date range.
+     * @return the number of log entries deleted
      */
-    public int[] purge(String tenantId, String namespace, String flowId, String executionId, List<Level> logLevels, ZonedDateTime startDate, ZonedDateTime endDate, boolean purgeExecutionLogs, boolean purgeNonExecutionLogs) {
-        int executionLogsDeleted = 0;
-        int nonExecutionLogsDeleted = 0;
-
-        if (purgeExecutionLogs && purgeNonExecutionLogs) {
-            executionLogsDeleted = logRepository.deleteByQuery(tenantId, namespace, flowId, executionId, logLevels, startDate, endDate, true, true);
-        } else if (purgeExecutionLogs) {
-            executionLogsDeleted = logRepository.deleteByQuery(tenantId, namespace, flowId, executionId, logLevels, startDate, endDate, true, false);
-        } else if (purgeNonExecutionLogs) {
-            nonExecutionLogsDeleted = logRepository.deleteByQuery(tenantId, namespace, flowId, executionId, logLevels, startDate, endDate, false, true);
+    public PurgeResult purge(String tenantId, String namespace, String flowId, String executionId, List<Level> logLevels, ZonedDateTime startDate, ZonedDateTime endDate, boolean purgeExecutionLogs, boolean purgeNonExecutionLogs) {
+        if (!purgeExecutionLogs && !purgeNonExecutionLogs) {
+            return new PurgeResult(0, 0);
         }
 
-        return new int[]{executionLogsDeleted, nonExecutionLogsDeleted};
+        int deleted = logRepository.deleteByQuery(tenantId, namespace, flowId, executionId, logLevels, startDate, endDate, purgeExecutionLogs, purgeNonExecutionLogs);
+
+        // When both types are purged in a single call, we can't distinguish the individual counts
+        return new PurgeResult(
+            purgeExecutionLogs ? deleted : 0,
+            !purgeExecutionLogs && purgeNonExecutionLogs ? deleted : 0
+        );
     }
 
 
--- a/core/src/main/java/io/kestra/core/services/ExecutionLogService.java
+++ b/core/src/main/java/io/kestra/core/services/ExecutionLogService.java
@@ -29,38 +29,23 @@ public class ExecutionLogService {
     }
 
     /**
-     * Purges log entries matching the given criteria.
+     * Purges log entries matching the given criteria, with separate control over execution and non-execution logs.
      *
-     * @param tenantId    the tenant identifier
-     * @param namespace   the namespace of the flow
-     * @param flowId      the flow identifier
-     * @param executionId the execution identifier
-     * @param logLevels   the list of log levels to delete
-     * @param startDate   the start of the date range
-     * @param endDate     the end of the date range.
-     * @return the number of log entries deleted
+     * @return an array of two ints: [executionLogsDeleted, nonExecutionLogsDeleted]
      */
-    public int purge(String tenantId, String namespace, String flowId, String executionId, List<Level> logLevels, ZonedDateTime startDate, ZonedDateTime endDate) {
-        return logRepository.deleteByQuery(tenantId, namespace, flowId, executionId, logLevels, startDate, endDate);
-    }
-
-    /**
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/kestra/plugin/core/log/PurgeLogs.java
```diff
@@ -104,6 +105,20 @@ public class PurgeLogs extends Task implements RunnableTask<PurgeLogs.Output> {
     @NotNull
     private Property<String> endDate;
 
+    @Schema(
+        title = "Whether to purge execution logs",
+        description = "Execution logs are logs where the trigger ID is null. Default is true."
```
loicmathieu: This is a technical description which would not be understandable by users

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLogRepository.java
```diff
@@ -358,6 +363,12 @@ public int deleteByQuery(String tenantId, String namespace, String flowId, Strin
                     delete = delete.and(levelsCondition(logLevels));
                 }
 
+                if (Boolean.TRUE.equals(purgeExecutionLogs) && !Boolean.TRUE.equals(purgeTriggerLogs)) {
+                    delete = delete.and(field("trigger_id").isNull());
+                } else if (Boolean.TRUE.equals(purgeTriggerLogs) && !Boolean.TRUE.equals(purgeExecutionLogs)) {
+                    delete = delete.and(field("trigger_id").isNotNull());
+                }
```
loicmathieu: This would not work as there are no indices on `trigger_id` so you cannot do the purge by filtering it.

loicmathieu (changes requested):
You may use `execution_id` instead of `trigger_id` but in this case it would not be trigger logs but non-execution logs (as there might be flow level logs in some very specific areas).

Adding an index is for me a no-go as we already have too many of them and the cost of INSERT is already high.

core/src/main/java/io/kestra/plugin/core/log/PurgeLogs.java
fdelbrayelle: What do you propose instead? I took example on `PurgeExecutions` :man_shrugging:

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLogRepository.java
fdelbrayelle: How would you achieve the purge then?

loicmathieu: Don't know, needs investiguation

fdelbrayelle: I addressed your concern @loicmathieu

core/src/main/java/io/kestra/core/services/ExecutionLogService.java
```diff
@@ -29,19 +29,23 @@ public ExecutionLogService(LogRepositoryInterface logRepository) {
...
+        int executionLogsDeleted = 0;
+        int nonExecutionLogsDeleted = 0;
+
+        if (purgeExecutionLogs && purgeNonExecutionLogs) {
+            executionLogsDeleted = logRepository.deleteByQuery(tenantId, namespace, flowId, executionId, logLevels, startDate, endDate, true, true);
+        } else if (purgeExecutionLogs) {
+            executionLogsDeleted = logRepository.deleteByQuery(tenantId, namespace, flowId, executionId, logLevels, startDate, endDate, true, false);
+        } else if (purgeNonExecutionLogs) {
+            nonExecutionLogsDeleted = logRepository.deleteByQuery(tenantId, namespace, flowId, executionId, logLevels, startDate, endDate, false, true);
+        }
```
loicmathieu: It should be simplifieable to 

```java
logRepository.deleteByQuery(tenantId, namespace, flowId, executionId, logLevels, startDate, endDate, purgeExecutionLogs,purgeNonExecutionLogs );
```

core/src/main/java/io/kestra/core/services/ExecutionLogService.java
```diff
@@ -29,19 +29,23 @@ public ExecutionLogService(LogRepositoryInterface logRepository) {
...
-     * @param executionId the execution identifier
-     * @param logLevels   the list of log levels to delete
-     * @param startDate   the start of the date range
-     * @param endDate     the end of the date range.
-     * @return the number of log entries deleted
+     * @return an array of two ints: [executionLogsDeleted, nonExecutionLogsDeleted]
      */
-    public int purge(String tenantId, String namespace, String flowId, String executionId, List<Level> logLevels, ZonedDateTime startDate, ZonedDateTime endDate) {
-        return logRepository.deleteByQuery(tenantId, namespace, flowId, executionId, logLevels, startDate, endDate);
+    public int[] purge(String tenantId, String namespace, String flowId, String executionId, List<Level> logLevels, ZonedDateTime startDate, ZonedDateTime endDate, boolean purgeExecutionLogs, boolean purgeNonExecutionLogs) {
```
loicmathieu: I would prefer if it returns a `Pair` object or a record with 2 components

loicmathieu (approved):
Some small code improvement suggestion but it's OK for me so I already validate it

core/src/main/java/io/kestra/core/services/ExecutionLogService.java
fdelbrayelle: Yes you're right it's more meaningful to return a value object / record here :)

Issue #14623, opened 2026-02-13
Array out ot of bound exception during plugin validation


nkwiatkowski:
### Describe the issue

During the plugin validation of the Azure secret, we do have some arrays out of bound exception that start happening after some time, and required a restart to stop showing off.

```
2026-01-20T14:32:14.111379Z TRACE java.lang.IndexOutOfBoundsException: Index 1 out of bounds for length 1
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Unknown Source)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Unknown Source)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Unknown Source)
	at java.base/java.util.Objects.checkIndex(Unknown Source)
	at java.base/java.util.ArrayList.get(Unknown Source)
	at org.hibernate.validator.internal.properties.javabean.JavaBeanExecutable.getParameterName(JavaBeanExecutable.java:80)
	at org.hibernate.validator.internal.metadata.aggregated.ParameterMetaData$Builder.build(ParameterMetaData.java:165)
	at org.hibernate.validator.internal.metadata.aggregated.ExecutableMetaData$Builder.findParameterMetaData(ExecutableMetaData.java:436)
	at org.hibernate.validator.internal.metadata.aggregated.ExecutableMetaData$Builder.build(ExecutableMetaData.java:391)
	at org.hibernate.validator.internal.metadata.aggregated.BeanMetaDataBuilder$BuilderDelegate.build(BeanMetaDataBuilder.java:260)
	at org.hibernate.validator.internal.metadata.aggregated.BeanMetaDataBuilder.build(BeanMetaDataBuilder.java:133)
	at org.hibernate.validator.internal.metadata.BeanMetaDataManagerImpl.createBeanMetaData(BeanMetaDataManagerImpl.java:206)
	at org.hibernate.validator.internal.metadata.BeanMetaDataManagerImpl.getBeanMetaData(BeanMetaDataManagerImpl.java:165)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validate(ValidatorImpl.java:156)
	at io.micronaut.configuration.hibernate.validator.MicronautHibernateValidator.validate(MicronautHibernateValidator.java:69)
	at io.kestra.ee.secret.SecretInterfaceFactory.make(SecretInterfaceFactory.java:102)
[...]

PR #14936, opened 2026-03-05
fix(core): override micronaut DefaultParameterNameProvider to fix a b…

nkwiatkowski:
…ug where he uses the metadata to instantiate a bean

 - fix #https://github.com/kestra-io/kestra/issues/14623


loicmathieu (approved):
I hope micronaut didn't mess with the order :(

PR #14944, opened 2026-03-05
fix(jdbc): next task deduplication could have false-positive skips

brian-mulier-p:
I initially thought it was linked to such [failure](https://github.com/kestra-io/kestra/actions/runs/22729761619/job/65915400552?pr=14947#step:6:5106) but the more I look the more I think it's due to [error taskruns removal upon restart](https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/services/ExecutionService.java#L225), which removes the attempts and I'm not sure the ExecutorState has been cleared in-between so it leads to duplicate key and so the error handler is not executed.

Anyway I found it weird to do anyMatch and discard EVERY nexts instead of just filtering to keep only-seen-once at the end.

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
@@ -640,9 +640,10 @@ public class JdbcExecutor implements ExecutorInterface {
                         }
                         executor = executorService.process(executor);
 
-                        if (!executor.getNexts().isEmpty() && deduplicateNexts(execution, executorState, executor.getNexts())) {
+                        List<TaskRun> nexts = deduplicateNexts(execution, executorState, executor.getNexts());
+                        if (!nexts.isEmpty()) {
                             executor.withExecution(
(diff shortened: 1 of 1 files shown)


loicmathieu:
> Anyway I found it weird to do anyMatch and discard EVERY nexts instead of just filtering to keep only-seen-once at the end.

This is how it worked since day one, if you want to execute a list of nexts that was already executed, it means your message is a duplicate so we do nothing. This methods analyse taskruns but it's purpose is realy to deduplicate messages.

I'm a bit reluctant to validate this PR as I'm not sure we have tests for duplicate messages and the rational you give is more "I think it's better like that" than a full explaination. Don't get me wrong, I think this is a good direction to look at but as it's core to our deduplication mechanism I want to be sure we didn't endanger deduplication.

brian-mulier-p:
Kafka doesn't do that :thinking: it filters duplicated nexts but that's it, here we discard the whole payload (including potentially nexts that haven't been processed at all) whenever we see ONE duplicate...

brian-mulier-p:
Closing as this will be removed soon + Kafka does the same in fact I misread it.
Still some edge cases around parallel imo but whatever.

PR #14947, opened 2026-03-05
Fix/tenant id fallback

brian-mulier-p:


loicmathieu (commented):
Without explaination I have no idea what it is intended to fix

brian-mulier-p:
This one is just about adding a flag to disable the queue migration as some users that have large data amount may want to perform the queue migration after some in-between verification steps (support issue).

brian-mulier-p:
+ the EE PR counterpart was not properly handling the fallback tenant in the tenant selector.

loicmathieu:
Ah OK, good to me then

loicmathieu (approved):
LGTM

PR #14951, opened 2026-03-06
test(scheduler): Fix AbstractSchedulerWorkerGroupTest + Cli tenant create

Skraye:
Also modified the Kestra Test to allows running the controller without workers


loicmathieu (approved):
LGTM

PR #14967, opened 2026-03-06
fix(jdbc): avoid some duplicate consumption

brian-mulier-p:
Offset update and consumption is now done in the same transaction


loicmathieu (approved):
LGMT

brian-mulier-p:
Not working out of the box so for now closing, we may revisit that later.

PR #14969, opened 2026-03-07
fix(ui): prevent text clipping in execution overview state history 

### LINKED ISSUE #14878: Execution Overview page clips text in the left column when it's too narrow
### Describe the issue

The text clips for states when you make the column too narrow. It doesn't have to be very narrow (or page zoomed in) before state timestamps are clipped.

The column does have a horizontal scroll bar appear but it doesn't allow the text to be viewed or selected.

Image

https://github.com/user-attachments/assets/699ad3d1-e37d-473f-aa30-87179a96a3b8

### Environment

- Kestra Version: `1.3.0-rc1`

sanjay78632:
### ✨ Description
Fixes text clipping in the Execution Overview page's State History timeline when the left panel is too narrow.

The root cause was `el-timeline-item`'s `:timestamp` prop which renders timestamps using `position: absolute; left: -210px`. When the panel is narrower than ~210px, these timestamps are pushed outside the visible area and clipped by parent `overflow: hidden` — with no way to scroll to them.

**Fix:** Removed the `:timestamp` prop and moved both the timestamp and state name into the content slot using a flex layout (`.timeline-row`). Since they are now in normal document flow, they can never be clipped — and on very narrow panels they wrap gracefully onto separate lines.

### 🔗 Related Issue
Closes https://github.com/kestra-io/kestra/issues/14878.

### 🎨 Frontend Checklist

### 📝 Additional Notes
Only one file changed: `ui/src/components/executions/overview/components/sidebar/Timeline.vue`
- Removed `:timestamp` prop from `el-timeline-item`
- Added `.timeline-row` flex container inside the content slot
- Removed now-unused `.el-timeline-item__timestamp` absolute positioning styles

https://github.com/user-attachments/assets/e1f25bc5-1876-41d3-8582-1324f172da6a

-


MilosPaunovic (changes requested):
Hey @sanjay78632,

Good work so far! 🍀 

There is only a fact that we don't want to change the look (timestamp being on the left and state on the right) when there is enough space, and your PR does that.

We'd like to amend only the situation where the container width get too narrow.

Can you handle that?

PR #14970, opened 2026-03-07
Fix StorageService.partition() to always close RandomAccessFile writers

iamsanjaymalakar:
### Problem
`StorageService.partition()` creates multiple `RandomAccessFile` writers but closes them via a `forEach` wrapper that can throw early. If one `close()` fails (or if an exception occurs during reading/writing), remaining writers may not be closed.

### Fix
Move writer cleanup into a finally block so all `RandomAccessFile` instances are always closed. Close failures are aggregated (first exception thrown, subsequent ones suppressed).


core/src/main/java/io/kestra/core/services/StorageService.java
```diff
@@ -104,25 +104,42 @@ private static List<Path> split(RunContext runContext, String extension, String
...
 
-            index = index >= writers.size() - 1 ? 0 : index + 1;
+            return files.stream().filter(p -> p.toFile().length() > 0).toList();
+        } finally {
+            IOException first = null;
+            for (RandomAccessFile w : writers) {
+                try {
+                    w.close();
+                } catch (IOException e) {
+                    if (first == null) {
```
loicmathieu: Please log each exception via `runContext.logger().error()`.

core/src/main/java/io/kestra/core/services/StorageService.java
```diff
@@ -104,25 +104,42 @@ private static List<Path> split(RunContext runContext, String extension, String
...
+                        first = e;
+                    }
+                    else {
+                        first.addSuppressed(e);
+                    }
```
loicmathieu: I don't think this is correct suppressing and exception is for multiple exception on the same call stack (usually, you suppress the cause) as the two exceptions are not linked there is no point of suppressing it

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/services/StorageService.java
iamsanjaymalakar: Thank you for the suggestion.

loicmathieu (approved):
LGTM, thanks

PR #14975, opened 2026-03-09
fix(core): add JDBC dependency to queue beans to avoid shutting down …

brian-mulier-p:
…in the middle of a transaction

### DIFF
--- a/core/src/main/java/io/kestra/core/queues/QueueFactoryInterface.java
+++ b/core/src/main/java/io/kestra/core/queues/QueueFactoryInterface.java
@@ -9,7 +9,7 @@ import io.kestra.core.models.templates.Template;
 import io.kestra.core.models.triggers.Trigger;
 import io.kestra.core.runners.*;
 
-public interface QueueFactoryInterface {
+public interface QueueFactoryInterface<DEPENDENCY> {
     String EXECUTION_NAMED = "executionQueue";
     String EXECUTOR_NAMED = "executorQueue";
     String WORKERJOB_NAMED = "workerJobQueue";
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/kestra/core/queues/QueueFactoryInterface.java
```diff
@@ -9,7 +9,7 @@
 import io.kestra.core.models.triggers.Trigger;
 import io.kestra.core.runners.*;
 
-public interface QueueFactoryInterface {
+public interface QueueFactoryInterface<DEPENDENCY> {
```
loicmathieu:
```suggestion
public interface QueueFactoryInterface<D> {
```

Parameterized types are usualy denoted by a single letter, you can add a commend to describe what it is.

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueueDependencies.java
```diff
@@ -0,0 +1,10 @@
+package io.kestra.jdbc.runner;
+
+import io.kestra.jdbc.JooqDSLContextWrapper;
+import jakarta.annotation.Nullable;
+import jakarta.inject.Singleton;
+
+@Singleton
+@JdbcRunnerEnabled
+public record JdbcQueueDependencies(@Nullable JooqDSLContextWrapper jooqDSLContextWrapper) {
```
loicmathieu: Why is it nullable?

core/src/main/java/io/kestra/core/queues/QueueFactoryInterface.java
loicmathieu: Moreover, instead of having an ignored dependency which sounds suspicious you could have a real dependecy bean which would be a carrier for injected dependency and use it to build the queues instead of the application context.

As this code is soon depdecated, We can stay like that but better explain what's the purpoise of this dependency class.

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
```diff
@@ -336,18 +336,25 @@ public Runnable receive(String consumerGroup, Consumer<Either<T, Deserialization
 
         Timer timer = this.metricRegistry
             .timer(MetricRegistry.METRIC_QUEUE_RECEIVE_DURATION, MetricRegistry.METRIC_QUEUE_RECEIVE_DURATION_DESCRIPTION, tags);
+        StackTraceElement[] parentStackTrace = Thread.currentThread().getStackTrace();
```
loicmathieu: I'm not fan of retaining the stack trace forever here

loicmathieu: I think this part can be reverted

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
```diff
@@ -459,13 +466,14 @@ protected String queueName(Class<?> queueType) {
 
     @SuppressWarnings("BusyWait")
     protected Runnable poll(Supplier<Integer> runnable) {
+        AtomicBoolean queriedToStop = new AtomicBoolean(false);
```
loicmathieu: 3 booleans are not needed here.
2 should be enought: running and stopped (or active and stopped).
The thing is to switch running to false and await untill isClosed is true. This is what we do in 2.0

loicmathieu (changes requested):

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueueDependencies.java
brian-mulier-p: It's not I'll remove it, it was a placeholder for Kafka while I find the @JdbcRunnerEnabled

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
```diff
@@ -459,14 +460,15 @@ protected String queueName(Class<?> queueType) {
 
     @SuppressWarnings("BusyWait")
     protected Runnable poll(Supplier<Integer> runnable) {
-        AtomicBoolean running = new AtomicBoolean(true);
+        AtomicBoolean queriedToStop = new AtomicBoolean(false);
```
loicmathieu: It didn't seems to be used anymore right?

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
```diff
@@ -498,15 +500,29 @@ protected Runnable poll(Supplier<Integer> runnable) {
...
+            stopped.set(true);
         });
 
-        return () -> running.set(false);
+        return () -> {
+            running.set(false);
+            try {
+                Awaitility.await()
+                    .atMost(Duration.ofSeconds(30))
+                    .pollInterval(Duration.ofMillis(10))
```
loicmathieu, line 496: I usually didn't set the pollInterval as if I remember correctly awaitility by default use a backoff strategy which is better than a fixed poll

loicmathieu (approved):
LGTM, nice

PR #14976, opened 2026-03-09
chore(execution): move back the run context to an abstract class

loicmathieu:


fhussonnois (approved):
Approved, but I do not agree with this decision.

PR #14993, opened 2026-03-10
fix(log): ignore Prometheus queries

yuri1969:
### ✨ Description

There are clients querying the Prometheus endpoint using params. For example Blackbox Exporter with `/prometheus?module=http_2xx`.


loicmathieu (approved):
LGTM, thanks.
As I consider it as a bugfix, I'll backport it to 1.0 and 1.3

loicmathieu:
Backported to 1.0 and 1.3

PR #14994, opened 2026-03-10
fix(system): test can deadlock

loicmathieu:
Tests can deadlock as the application context was closed in the `ThreadUncaughtExceptionHandler` and by the Micronaut test support.

Also wait for termination on the queue service to force closing queue consumers.


fhussonnois (approved):
LGTM!

brian-mulier-p (approved):
LGTM

loicmathieu:
Tests last for 2h with this change so even if it seems good it causes other issues or it surfaces other issues so I'll switch it to draft for now

PR #14996, opened 2026-03-10
Support nullable worker-group key rendering

proddata:
This change fixes worker group key rendering so Pebble expressions can intentionally resolve to `null` and behave the same as if no `workerGroup.key` was set.

Expressions like `{{ null }}` or `{{ true ? null : 'work' }}` were effectively coerced to string output with `render(...)`.

fixes: https://github.com/kestra-io/kestra-ee/issues/6629

## What changed
- Added nullable string rendering path via `renderNullableString(...)` in run variable rendering.
  - preserve semantic `null`,
  - stringify typed non-string values consistently,
  - fallback to standard string rendering for mixed templates.
- Updated worker-group key resolution in executor and scheduler to use nullable rendering.

Users can now intentionally set `workerGroup.key` to `null` through e.g. a Pebble expression and get default routing behavior.

## Follow-up in EE required
A follow-up change is needed in `kestra-ee`:
- [`runner-kafka-ee/.../WorkerJobTopicNameExtractor`](https://github.com/kestra-io/kestra-ee/blob/da8b7fd05b1a088c2712b556eee802ae70808a51/runner-kafka-ee/src/main/java/io/kestra/ee/runner/kafka/streams/WorkerJobTopicNameExtractor.java#L47) should use `runContext.renderNullableString(...)` for worker-group key rendering to keep behavior aligned with OSS executor/scheduler routing.

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
+++ b/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
@@ -60,6 +60,34 @@ public class VariableRenderer {
         return this.render(inline, variables, this.variableConfiguration.getRecursiveRendering(), false);
     }
 
+    /**
+     * Render a string expression while preserving semantic {@code null} for single-expression templates.
+     * <p>
+     * For expressions that render to non-string objects, this method stringifies the typed result
+     * using the same writer used by regular string rendering.
+     */
+    public String renderNullableString(String inline, Map<String, Object> variables) throws IllegalVariableEvaluationException {
+        Object typedValue;
+        try {
+            typedValue = this.renderTyped(inline, variables);
+        } catch (IllegalArgumentException e) {
+            // Typed rendering cannot concatenate complex types with literals.
+            // Fall back to standard string rendering for those mixed templates.
+            return this.render(inline, variables);
+        }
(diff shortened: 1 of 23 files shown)


core/src/main/java/io/kestra/core/runners/VariableRenderer.java
```diff
@@ -60,6 +60,34 @@ public Object renderTyped(String inline, Map<String, Object> variables) throws I
...
+     */
+    public String renderNullableString(String inline, Map<String, Object> variables) throws IllegalVariableEvaluationException {
+        Object typedValue;
+        try {
+            typedValue = this.renderTyped(inline, variables);
+        } catch (IllegalArgumentException e) {
+            // Typed rendering cannot concatenate complex types with literals.
+            // Fall back to standard string rendering for those mixed templates.
+            return this.render(inline, variables);
+        }
```
loicmathieu: Trying one way then the other seems like a hack.
I would prefer a more robust solution, maybe even a broader solution to support null everywhere.

loicmathieu (changes requested):
Special casing only rendering when we render the worker group key may not be a good idea as we may want that to work globally.

PR #15002, opened 2026-03-10
chore(core): bring back deprecated methods of AbstractExecScript

loicmathieu:
Even if they should not be used anymore in 2.0 only for backward compatibility.


fhussonnois (approved):

loicmathieu:
@fdelbrayelle FYI this should make plugin script compiling again

PR #15030, opened 2026-03-11
chore(core): add utility method for closing ExecutorService

fhussonnois:


loicmathieu (approved):
LGTM.

As we usually use a 30s timeout, I'll add an overload without timeout for simplicity

Issue #14971, opened 2026-03-09
[Pebble] KV store Set: handle array output from task correctly


caxefaizan:
The KV store Set plugin errors when the value comes from a task output that is an array, failing with a type mismatch 
```
Cannot concat java.lang.String with java.util.ArrayList.
```
 
This blocks migrations where a complex object or list of dicts is intended to be stored. 
Investigate supporting array/object inputs for value and ensure proper type handling or conversion; add tests for array/list inputs and clarify value rendering when coming from task outputs.

Sample
Hardcoding works
```
tasks:
- id: add_to_kv_store
  type: io.kestra.plugin.core.kv.Set
  kvType:  JSON
  key: "temp"
  value: |
    {"a": [1,2,3 ] }
```
Pebble Expression doesnt
```
tasks:
- id: add_to_kv_store
  type: io.kestra.plugin.core.kv.Set
  kvType:  JSON
  key: "temp"
  value: |
    {"a": {{[1,2,3 ]}} }
```

proddata:
The problem seems to be the mixed render, which is not fully supported yet.

A workaround solution is putting the full payload in a pebble expression:

```yaml
  - id: t01
    type: io.kestra.plugin.core.output.OutputValues
    values:
      arr: [1,2,3]
  - id: add_to_kv_store
    type: io.kestra.plugin.core.kv.Set
    kvType:  JSON
    key: "temp"
    value: |
      {{ {"a": outputs.t01.values.arr } }}
```

nkwiatkowski:
Is this issue still active?

proddata:
> Is this issue still active?

I would say it is still valid to fix

PR #15034, opened 2026-03-11
fix(core): exception when multiple types are in a string pebble expre…

nkwiatkowski:
…ssion

 - fix #https://github.com/kestra-io/kestra/issues/14971


AcevedoR (approved):

loicmathieu (commented):
I understand the rational but I'm not comptfortable with this change.
Performance wize, it's not good when there are mixed types, we can consider that it's rare and we don't care.

But trying one approache, then another is a sign of bad design. I wonder if there isn't a way to detect mixed types. Moreover, here, the Pebble expression is only an array so it's not mixed type right?
Maybe I missunderstood something here

nkwiatkowski:
In this case , it is a mixed type because pebble start reading a string, then add the array, and fail during the concatenation of the string + the arrays

PR #15035, opened 2026-03-11
chore(system): avoid loading the credentials from DB for webhook

### LINKED ISSUE #14867: The AuthenticationFilter is called on webhook
### Describe the issue

The AuthenticationFilter is called even on anonymous endpoints like Webhook, as it load the authentication from the setting it incurs a small performance hit.

If it cannot be avoided, it may also be an idea to check whether or not we can cache the auth as in OSS there is only one so loading it from the settings on each request seems overkill

### Environment

- Kestra Version: develop

loicmathieu:
Closes #14867


Skraye (approved):
LGTM

PR #15037, opened 2026-03-11
Fix non-required inputs failing when prefill value is cleared at execution

### LINKED ISSUE #14946: Prefill doesn't work for some input types when prefill value removed on execute page
### Describe the issue

Prefill enables you to have inputs which aren't required have some sort of value that can be applied or removed at execution. There's a number of inputs where you are not able to remove the prefilled value at execution without an error preventing the execution from starting.

Maybe this is intended, but this should be documented. 

Reproducer flow:

Works:
```yaml
id: input_example
namespace: company.team

inputs:
  - id: not_required
    type: STRING
    prefill: "Hello"
    required: false

tasks:
  - id: hello
    message: "{{ inputs ?? '' }}"
```

Does not work:
```yaml
id: input_example
namespace: company.team

inputs:
  - id: not_required
    type: INT
    prefill: 1
    required: false

tasks:
  - id: hello
    message: "{{ inputs ?? '' }}"
```

If you try and execute that INT example, and remove the value in the input field, this error comes up when you press Execute:
Image

Input types that do not work ❌:
- [ ] INT
- [ ] FLOAT
- [ ] DATE
- [ ] DATETIME
- [ ] DURATION

Input types that don't really work with `required: false`
- [ ] ARRAY (ish) - sends an empty array if `prefill` present
- [ ] BOOL - either returns true or input is ignored, regardless of `prefill` value.
- [ ] TIME - doesn't let you set it to blank. Resets to 00:00:00 (but 12:00:00 in the execution)

What does work ✅: 
- [ ] STRING
- [ ] JSON
- [ ] URI
- [ ] SECRET
- [ ] YAML - but the formatting was off in the 'Execute the flow' menu (top is with prefill, bottom is with defaults - expected behaviour)

Image

### Environment

- Kestra Version: 1.3.0

nancysangani:
Fixes #14946

When a non-required input has a prefill value and the user clears it on the Execute dialog, the UI submits the string `"null"`. This was being passed directly to type parsers like `Integer.valueOf("null")`, causing an error for INT, FLOAT, DATE, DATETIME, and DURATION inputs.

Fixed by adding an early null/blank check in `FlowInputOutput.parseType()` before any type parsing occurs.


nancysangani:
@loicmathieu Please review it when you get a chance. Thanks!

MilosPaunovic:
Can you resolve a conflict here @nancysangani?

nancysangani:
@MilosPaunovic Resolved the merge conflict. Ready for review. Thanks!

loicmathieu (changes requested):
I wonder if this cannot be solved more gracefuly in the UI as it seems wrong to coalesce the null string to null here

nancysangani:
@loicmathieu fixed in the UI instead — null values from cleared non-required inputs (INT, FLOAT, DATE, DATETIME, DURATION) are now excluded from the form data before being sent to the backend, so the backend never receives "null" as a string. Reverted the backend change.
Please review it when you get a chance. Thanks!

loicmathieu:
@nancysangani thanks, as it's now UI only, I'll let @MilosPaunovic review it ;)

MilosPaunovic (approved):

PR #15043, opened 2026-03-12
Fix: Blueprints tags don't reset when navigating away #14965

### LINKED ISSUE #14965: Blueprints Tags don't reset when navigating away
### Describe the issue

Steps to reproduce:
1. Go to the flow blueprints page
2. Select a tag
3. Click onto another page (e.g. Flows or Dashboards)
4. Click back to Flow Blueprints and all blueprints are displayed
5. You now have to click the tag twice to get it to show just the tag.

https://github.com/user-attachments/assets/52c4941c-a9d5-4fb4-af05-6fdc9f65b0a9

### Environment

- Kestra Version: `1.3.0` EE and OsS

Ingole712521:
## Description

This PR fixes an issue where blueprint tags were not applied correctly when navigating away from the Blueprints Browser and returning to it.

Previously, when a user selected a tag and navigated to another page, returning to the Blueprints Browser could trigger an initial API request without the selected tag filter, even though the URL already contained the `selectedTag` query parameter.

This resulted in an incorrect initial request and could briefly display unfiltered results before the correct tag state was restored.

## Fix

The query construction now prioritizes tags derived from the route query (`selectedTag`) before falling back to the component's internal `selectedTags` state.

This ensures that the correct tag filter is applied immediately when the component loads or when the user navigates back to the page.

## Result

- The tag filter remains consistent with the URL state
- Prevents unfiltered blueprint results from appearing briefly
- Ensures the correct filtered results are loaded when returning to the page

Fixes #14965


Ingole712521:
Hi 👋

Sharing a short screen recording demonstrating the behavior after the fix.

The selected tag now remains correctly applied when navigating away from the Blueprints Browser and returning to the page.

Issue #14965

https://github.com/user-attachments/assets/bf5e4214-6710-40ac-ac52-f13853ce7719

MilosPaunovic (approved):
LGTM! 🚀

PR #15067, opened 2026-03-13
chore: replace instances of applicationContext DI with direct eager injection or Providers.

ammeek:
### ✨ Description

This PR standardizes dependency wiring by replacing programmatic bean lookups with framework-managed injection patterns across the codebase where applicable, improving lifecycle safety and consistency during startup/shutdown ordering.

  ### 🔗 Related Issue

  Relates To: kestra-io/kestra/issues/14982

  ### 🛠️ Backend Checklist

  ### 📝 Additional Notes
This PR does not include all of the instances of Di via application context but it;s alrady getting quite large so I will do the rest of them in a separate pr.

  This change follows the project guideline to avoid runtime bean retrieval via ApplicationContext.getBean() / findBean() when standard injection can be used. It aligns dependency declaration with Micronaut lifecycle management to reduce ordering-related instability.


cli/src/main/java/io/kestra/cli/commands/servers/WebServerCommand.java
```diff
@@ -26,15 +25,18 @@
...
 
     @Inject
     private IgnoreExecutionService ignoreExecutionService;
 
+    @Inject
+    private Indexer indexer;
+
+    @Inject
+    private Controller controller;
+
```
loicmathieu, line 40: Those two may not be used so it may be better to use a Provider here

cli/src/main/java/io/kestra/cli/StandAloneRunner.java
```diff
@@ -34,7 +33,19 @@ public class StandAloneRunner implements Runnable, AutoCloseable {
...
+    private Controller controller;
+
+    @Inject
+    private Worker worker;
+
+    @Inject
+    private Scheduler scheduler;
+
+    @Inject
+    private Indexer indexer;
```
loicmathieu: Those may not be used so it might be better to use a Provider

core/src/main/java/io/kestra/core/runners/VariableRenderer.java
```diff
@@ -23,23 +22,31 @@ public class VariableRenderer {
...
-        this.pebbleEngine = pebbleEngineFactory.create();
     }
-    
+
     public void setPebbleEngine(final PebbleEngine pebbleEngine) {
         this.pebbleEngine = pebbleEngine;
     }
-    
+
+    private PebbleEngine pebbleEngine() {
```
loicmathieu, line 38: Can't we avoid the double-checked locking idom here?
Why don't you construct the PebbelEngine from the factory using `create()` in the constructor as before?

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -99,10 +96,13 @@ public class ExecutorService {
     @Inject
     private TaskOutputService taskOutputService;
 
+    @Inject
+    private Provider<FlowMetaStoreInterface> flowMetaStoreInterfaceProvider;
+
     private FlowMetaStoreInterface flowExecutorInterface() {
```
loicmathieu: Can you remove this method and directly use the Provider instead?

queue/src/test/java/io/kestra/queue/AbstractDispatchQueueTest.java
loicmathieu, line 1: Interesting...
@fhussonnois WDYT as you worked on queue consumer lately?

loicmathieu (changes requested):

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -70,6 +67,7 @@ public class ExecutorService {
     @Inject
     private WorkerJobRunningStateStore workerJobRunningStateStore;
 
+    @Inject
```
loicmathieu, line 70: If you inject it here, why do you have a Provider on line 100?

loicmathieu (approved):
One small nit but overall LGTM

PR #15069, opened 2026-03-13
fix(core): JDBC queue closing was not working for commands

brian-mulier-p:


jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
```diff
@@ -529,18 +529,16 @@ public void resume() {
...
-        } catch (Exception e) {
-            log.warn("Error while closing queue", e);
-        }
+        this.closed.set(true);
         this.poolExecutor.shutdown();
         this.asyncPoolExecutor.shutdown();
+        try {
+            if(!this.poolExecutor.awaitTermination(30, TimeUnit.SECONDS) || !this.asyncPoolExecutor.awaitTermination(30, TimeUnit.SECONDS)) {
+                log.error("Couldn't wait for queue executors to close properly");
+            }
```
loicmathieu: If await return false, you must call shudownNow to force the shutdown of the executor and interrput the threads

loicmathieu (changes requested):

loicmathieu (approved):
LGTM

PR #15071, opened 2026-03-13
fix(keyboard-shortcuts): clarify font options for editor

wrussell1999:
The keyboard shortcuts mention about changing font size, but they mean specifically the Flow editor. I've changed the wording to match the settings menu as there was confusion about this being for all font, not just the editor.

Let me know if anything needs to happen to make this work for other languages.


MilosPaunovic:
What you would need in this case is to manually run this action:

https://github.com/kestra-io/kestra/actions/workflows/auto-translate-ui-keys.yml

On your branch `ui/keyboard-shortcuts`, with the `Whether to re-translate modified keys even if they already have translations.` option set to `true`. It will open a PR against this branch which should be merged, then this can be merged as well.

I've ran it this time and will handle the translation PR merging, but the above is the process for next time.

MilosPaunovic (approved):

wrussell1999:
This is very helpful, thank you!

PR #15076, opened 2026-03-15
fix(ui): remove escaped quotes from YAML JSON autocompletion

### LINKED ISSUE #14113: Autocomplete for behaviour in PurgeFiles inserts escaped JSON string

Sumitsh28:
### ✨ Description

Fixes an issue where autocompleting fields with complex `@Builder.Default` objects (like `behavior` in `PurgeFiles`) inserted escaped JSON strings instead of valid YAML/JSON.

As discussed in the previous backend PR #14304 , fixing this at the plugin schema level isn't scalable due to the widespread use of `@Builder.Default` across Kestra plugins.

This PR implements the fix on the frontend by intercepting the Monaco Editor YAML completion override in `yamlLanguageConfigurator.ts`. It sanitizes escaped quotes (`\"` and multiple escapes like `\\\"`) from the `label`, `insertText`, and `filterText`. Since YAML natively accepts JSON, this results in a clean insertion (e.g., `{"type":"version","keepAmount":1}`) without requiring complex YAML stringification logic.

### 🔗 Related Issue

Closes #14113 
Replaces PR #14304 

### 🎨 Frontend Checklist

* [x] Code builds without errors (`npm run build`)
* [x] All existing E2E tests pass (`npm run test:e2e`)
* [x] Screenshots or video recordings attached showing the `UI` changes

### Before
image

### After
image

### DIFF
--- a/ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
+++ b/ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
@@ -2,6 +2,7 @@ import {computed, watch} from "vue";
 import {useI18n} from "vue-i18n";
 import {configureMonacoYaml} from "monaco-yaml";
 import * as monaco from "monaco-editor/esm/vs/editor/editor.api";
+import {languages} from "monaco-editor/esm/vs/editor/editor.api";
 import {yamlSchemas} from "override/utils/yamlSchemas";
 import {StandaloneServices} from "monaco-editor/esm/vs/editor/standalone/browser/standaloneServices";
 import {ILanguageFeaturesService} from "monaco-editor/esm/vs/editor/common/services/languageFeatures";
@@ -9,11 +10,6 @@ import AbstractLanguageConfigurator from "./abstractLanguageConfigurator";
 import {YamlAutoCompletion} from "../../../services/autoCompletionProvider";
 import RegexProvider from "../../../utils/regex";
 import * as YamlUtils from "@kestra-io/ui-libs/flow-yaml-utils";
-import IPosition = monaco.IPosition;
-import IDisposable = monaco.IDisposable;
-import IModel = monaco.editor.IModel;
-import ProviderResult = monaco.languages.ProviderResult;
-import CompletionList = monaco.languages.CompletionList;
 import {
     endOfWordColumn,
     NO_SUGGESTIONS,
@@ -23,7 +19,11 @@ import {
 } from "./pebbleLanguageConfigurator";
 import {usePluginsStore} from "../../../stores/plugins";
 import {useBlueprintsStore} from "../../../stores/blueprints";
-import {languages} from "monaco-editor/esm/vs/editor/editor.api";
+import IPosition = monaco.IPosition;
+import IDisposable = monaco.IDisposable;
+import IModel = monaco.editor.IModel;
+import ProviderResult = monaco.languages.ProviderResult;
+import CompletionList = monaco.languages.CompletionList;
 import CompletionItem = languages.CompletionItem;
 
 type TaskLike = Record<string, unknown>;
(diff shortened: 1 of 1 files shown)


ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
```diff
@@ -194,7 +194,30 @@ export class YamlLanguageConfigurator extends AbstractLanguageConfigurator {
...
                     label: string;
                 })[]
-            )
+            )   
+                .map((s) => {
+                    const r = {...s};
+
+                    if (typeof r.label === "string") {
+                        r.label = r.label.replace(/\\+"/g, "\"");
+                    }
```
brian-mulier-p: are you sure the label should also have that change? To me only the insertText is escaped no? If the label contains the \ I'm ok with it but please do a raw replace (no regex) and don't capture every \ ahead of the ", just the one immediately before

ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
```diff
@@ -194,7 +194,30 @@ export class YamlLanguageConfigurator extends AbstractLanguageConfigurator {
...
+
+                    if (typeof r.insertText === "string") {
+                        r.insertText = r.insertText.replace(/\\+"/g, "\"");
```
brian-mulier-p: please do a raw replace (no regex) and don't capture every \ ahead of the ", just the one immediately before

ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
```diff
@@ -194,7 +194,30 @@ export class YamlLanguageConfigurator extends AbstractLanguageConfigurator {
...
+                    } else if (typeof r.insertText === "object" && r.insertText !== null) {
+                        const textObj = r.insertText as any;
+                        if (typeof textObj.value === "string") {
+                            textObj.value = textObj.value.replace(/\\+"/g, "\"");
```
brian-mulier-p: please do a raw replace (no regex) and don't capture every \ ahead of the ", just the one immediately before

ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
```diff
@@ -194,7 +194,30 @@ export class YamlLanguageConfigurator extends AbstractLanguageConfigurator {
...
+                        }
+                    }
+
+                    if (typeof r.filterText === "string") {
+                        r.filterText = r.filterText.replace(/\\+"/g, "\"");
```
brian-mulier-p: please do a raw replace (no regex) and don't capture every \ ahead of the ", just the one immediately before

brian-mulier-p (changes requested):
Almost LGTM 👍

ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
Sumitsh28: You are right! The raw label is just property name itself without any JSON. I removed the label replacement.

ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
Sumitsh28: Done! I've removed the regex and updated it to use a raw string replace instead.

MilosPaunovic:
If you've addressed any of the requested changes from the PR review, @Sumitsh28, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

brian-mulier-p (approved):
LGTM, I removed the second replace to avoid overriding unnecessary completions, if you have an example that doesn't work with it feel free to open another PR with an example but better to replace only the non-working example to avoid having too much side-effects

PR #15080, opened 2026-03-15
fix(pebble): prevent parent taskrun list mutation in CurrentEachOutputFunction

MTarek165:
# Problem
- currently, parent taskrun list maintain reference at memory hence mutating the list by reversing it is preserved in evaluation context and since pebble template uses the same context to evaluate the string template, so if we called the same function more than one it will fail as it will not start from root
- consider three nested ForEach so two parents exist say [v_inner, v_outer], the first time it will reverse parents in memory so it will be [v_outer, v_inner], and calling the function for the second time it will see the reversed list and reverse again returning to the original one hence starting to read outputs from inner to root causing failure!
- the same problem fixed here before merging this PR #12566

# Solution
- make parent taskrun list immutable so that it will be memory safe and each time we read parents and reverse to start from the root as exepected.

# Testing 
- calling the same function more than once at the same string template should works without failures

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/pebble/functions/CurrentEachOutputFunction.java
+++ b/core/src/main/java/io/kestra/core/runners/pebble/functions/CurrentEachOutputFunction.java
@@ -29,7 +29,9 @@ public class CurrentEachOutputFunction implements Function {
             for (Map<?, ?> parent : parents) {
                 Map<?, ?> taskrun = (Map<?, ?>) parent.get("taskrun");
                 if (taskrun != null) {
-                    if(outputs.get(taskrun.get("value")) == null) return null;
+                    if(outputs.get(taskrun.get("value")) == null) {
+                        return null;
+                    }
                     outputs = (Map<?, ?>) outputs.get(taskrun.get("value"));
                 }
             }
--- a/core/src/main/java/io/kestra/core/runners/pebble/functions/CurrentEachOutputFunction.java
+++ b/core/src/main/java/io/kestra/core/runners/pebble/functions/CurrentEachOutputFunction.java
@@ -29,12 +29,8 @@ public class CurrentEachOutputFunction implements Function {
             for (Map<?, ?> parent : parents) {
                 Map<?, ?> taskrun = (Map<?, ?>) parent.get("taskrun");
                 if (taskrun != null) {
-                    if(outputs.get(taskrun.get("value")) != null){
-                        outputs = (Map<?, ?>) outputs.get(taskrun.get("value"));
-                    }
-                    else {
-                        throw new PebbleException(null, "No execution outputs for the current task runs path", lineNumber, self.getName());
-                    }
+                    if(outputs.get(taskrun.get("value")) == null) return null;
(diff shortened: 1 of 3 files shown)


loicmathieu:
@MTarek165 can you provide a reproducer?

MTarek165:
1- This flow has 3 nested ForEach (more than one parent), calling the currentEachOutput() function twice on the same output taskId : 1-1-1_output, it should always compute the same result for the first call and the second but it will fail, now remove one function call to be only once it will work as known.
2- The failure occurs on the second call 
```
id:  test-current-output
namespace: io.kestra.tests
tasks:
  - id: 1_each
    values: '["v11", "v12"]'
    tasks:
      - id: 1-1_each
        values: '["v21", "v22"]'
        tasks:
          - id: 1-1-1_each
            values: '["v31", "v32"]'
            tasks:
            - id: 1-1-1_output
              format: "{{ parents[1].taskrun.value }}-{{ parents[0].taskrun.value }}-{{ taskrun.value }}"
            - id: 1-1-1_return
              format: "return-{{ currentEachOutput(outputs['1-1-1_output']).value }}-{{ currentEachOutput(outputs['1-1-1_output']).value }}"
              
```

core/src/main/java/io/kestra/core/runners/pebble/functions/CurrentEachOutputFunction.java
```diff
@@ -23,13 +24,17 @@ public Object execute(Map<String, Object> args, PebbleTemplate self, EvaluationC
...
             for (Map<?, ?> parent : parents) {
                 Map<?, ?> taskrun = (Map<?, ?>) parent.get("taskrun");
                 if (taskrun != null) {
-                    outputs = (Map<?, ?>) outputs.get(taskrun.get("value"));
+                    if(outputs.get(taskrun.get("value")) != null){
+                        outputs = (Map<?, ?>) outputs.get(taskrun.get("value"));
+                    }
+                    else {
+                        throw new PebbleException(null, "No execution outputs for the current task runs path", lineNumber, self.getName());
+                    }
```
loicmathieu: This would be a BC and I cannot see why we would fail if no outputs as the Pebble expression may use a conditional construct

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/runners/pebble/functions/CurrentEachOutputFunction.java
MTarek165: it is not a BC because before this change output is returned as null causing cast exception as you always assume that output map of tge target task is corresponding to the taskrun  but it happens when the provided output doesn't contain value of the parent task run as in our case due to the reverse problem as we are not starting from the root or it can happen if the provided task is not from the current nest(another case) so this is unhandled case.
But it should be updated to return null instead of throwing exception as you said in the other PR. It makes more sense  and i will make them similar for consistency.

loicmathieu (approved):
LGTM, thanks

loicmathieu:
Backported to 1.0 and 1.3

PR #15084, opened 2026-03-16
Fix/queue offset

### LINKED ISSUE #15073: [BUG] Integer overflow on queues offset column causes infinite task re-execution (JDBC/PostgreSQL backend)
When the `queues` table `offset` sequence exceeds `Integer.MAX_VALUE` (2,147,483,647), the worker enters an infinite loop re-executing the same tasks because queue messages can never be acknowledged.

### Context

Migration `V1_41__offset_bigint.sql` correctly changed the `offset` column from `SERIAL` to `BIGINT`:

```sql
ALTER SEQUENCE queues_offset_seq AS BIGINT;
ALTER TABLE queues ALTER COLUMN "offset" TYPE BIGINT;
```

But the Java code in `JdbcQueue.receiveImpl` still reads offsets as `Integer`:

```java
// JdbcQueue.java, line 397 (receiveImpl, inTransaction path)
result.map(record -> record.get("offset", Integer.class))

// JdbcQueue.java, line 411 (receiveImpl, non-inTransaction path)
fetch.map(record -> record.get("offset", Integer.class))
```

And `PostgresQueue.updateGroupOffsets` (line 86) receives `List<Integer>`.

### What happens

When `offset > 2,147,483,647`, `record.get("offset", Integer.class)` either:
1. Throws a `DataTypeException` (jOOQ cannot convert BIGINT to Integer), or
2. Silently overflows to a wrong positive value

In both cases, `updateGroupOffsets` fails to set `consumer_worker = true`:
- **Exception case**: thrown **after** `consumer.accept()` already dispatched tasks to the worker thread pool (irreversible side effect) → transaction rolls back → messages stay unacknowledged
- **Overflow case**: `UPDATE ... WHERE "offset" IN (<wrong values>)` → 0 rows matched → `consumer_worker` stays `false`

The poll loop (`immediateRepoll = true` by default) immediately re-fetches the same unacknowledged messages, dispatches them again, and so on — **every task is executed dozens of times in parallel**, creating exponential pod/process creation.

### How to reproduce

1. Use JDBC/PostgreSQL backend in distributed mode
2. Advance the `queues_offset_seq` sequence past `Integer.MAX_VALUE`:
   ```sql
   ALTER SEQUENCE queues_offset_seq RESTART WITH 2147483648;
   ```
3. Trigger any flow with a worker task
4. Observe: the task is executed repeatedly, `consumer_worker` stays `false` for all new messages

### Impact on our production

- 16 scheduled flows triggered → 128 pods created simultaneously (20 executions per task, saturating all 128 worker threads)
- 76,583 unacknowledged messages accumulated in the `queues` table
- Complete platform outage for ~24h

### Suggested fix

Replace `Integer.class` with `Long.class` in `JdbcQueue.receiveImpl`:

```java
// Line 397
result.map(record -> record.get("offset", Long.class))

// Line 411
fetch.map(record -> record.get("offset", Long.class))
```

And update `updateGroupOffsets` signature from `List<Integer>` to `List<Long>` in `JdbcQueue` and `PostgresQueue`.

### Contributing issue: unacknowledged messages never cleaned up

Migration `V1_31__queues_updated_date.sql` drops the DEFAULT on the `updated` column:

```sql
ALTER TABLE queues ALTER COLUMN updated DROP NOT NULL;
ALTER TABLE queues ALTER COLUMN updated DROP DEFAULT;
```

Since `produceFields()` doesn't set `updated`, new messages are inserted with `updated = NULL`. The `JdbcCleaner` deletes where `updated <= threshold`, which never matches NULL rows. Unacknowledged messages accumulate indefinitely, making the impact worse when the overflow bug triggers.

### Minor: inverted `compareAndSet` in `JdbcWorkerJobQueueService.close()`

Line 75:
```java
// Current (cleanup never executes):
if (!isStopped.compareAndSet(true, false)) { return; }

// Should be:
if (!isStopped.compareAndSet(false, true)) { return; }
```

### Workaround

Stop all components, then:
```sql
DELETE FROM queues;
ALTER SEQUENCE queues_offset_seq RESTART WITH 1;
```

### Environment

- Kestra Version: v1.1.15 (Helm chart 1.0.41)
- Backend: PostgreSQL 17.6 (AWS RDS)
- Deployment: Distributed on EKS (separate webserver, executor, scheduler, worker)

loicmathieu:
Fixes #15073

### DIFF
--- a/jdbc-h2/src/main/java/io/kestra/runner/h2/H2Queue.java
+++ b/jdbc-h2/src/main/java/io/kestra/runner/h2/H2Queue.java
@@ -55,7 +55,7 @@ public class H2Queue<T> extends JdbcQueue<T> {
     }
 
     @Override
-    protected void doUpdateGroupOffsets(DSLContext ctx, String consumerGroup, String queueType, List<Integer> offsets) {
+    protected void doUpdateGroupOffsets(DSLContext ctx, String consumerGroup, String queueType, List<Long> offsets) {
         var update = ctx.update(DSL.table(table.getName()))
             .set(
                 AbstractJdbcRepository.field("consumers"),
@@ -66,7 +66,7 @@ public class H2Queue<T> extends JdbcQueue<T> {
                 )
             )
             .set(AbstractJdbcRepository.field("updated"), LocalDateTime.now())
-            .where(AbstractJdbcRepository.field("offset").in(offsets.toArray(Integer[]::new)));
+            .where(AbstractJdbcRepository.field("offset").in(offsets.toArray(Long[]::new)));
 
         if (consumerGroup != null) {
             update = update.and(AbstractJdbcRepository.field("consumer_group").eq(consumerGroup));
--- a/jdbc-mysql/src/main/java/io/kestra/runner/mysql/MysqlQueue.java
+++ b/jdbc-mysql/src/main/java/io/kestra/runner/mysql/MysqlQueue.java
@@ -64,7 +64,7 @@ public class MysqlQueue<T> extends JdbcQueue<T> {
     }
 
     @Override
-    protected void doUpdateGroupOffsets(DSLContext ctx, String consumerGroup, String queueType, List<Integer> offsets) {
+    protected void doUpdateGroupOffsets(DSLContext ctx, String consumerGroup, String queueType, List<Long> offsets) {
         var update = ctx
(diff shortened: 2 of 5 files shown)


brian-mulier-p (approved):
LGTM

loicmathieu:
Backported to 1.0 and 1.3

PR #8107, opened 2025-03-27
Warning if a scheduled flow has inputs that have no defaults + no inputs set on the trigger, closes #2881

VTKop:
### What changes are being made and why?

This pull request closes #2881. It adds a warning in the flow editor when a flow uses an input that has no 'defaults' property defined and no value for the input is given in a flow trigger of type 'Schedule'.
This change was made to help users discover a possible issue with such a flow, since it will fail if the schedule trigger executes it.

### How the changes have been QAed?

The following flow can be used to QA the change. It should show a single warning in the validation notes of the flow editor, being that the every_minute trigger is missing an input for user2.
Adding a default input for user2, or adding an inputs field for the every_minute trigger with an input for user2 should clear the warning. Removing the default value from user should add the 'user' input to the warning.

```yaml
id: missing_default
namespace: qa

inputs:
  - id: user
    type: STRING
    defaults: John
  - id: user2
    type: STRING


tasks:
  - id: hello
    type: io.kestra.plugin.core.log.Log
    message: hey there {{ inputs.user }}! 👋

triggers:
  - id: every_minute
    type: io.kestra.plugin.core.trigger.Schedule
    cron: "*/1 * * * *"
```

PR #15090, opened 2026-03-16
feat(flows): validate Schedule triggers inputs

### LINKED ISSUE #2881: Warn in the UI (e.g. when clicking on "Save") if a scheduled flow has inputs that have no defaults + no inputs set on the trigger
## Use case
The following flow:

```yaml
id: missing_default
namespace: qa

inputs:
  - name: user
    type: STRING

tasks:
  - id: hello
    message: hey there {{ inputs.user }}! 👋

triggers:
  - id: every_minute
    type: io.kestra.core.models.triggers.types.Schedule
    cron: "*/1 * * * *"
```

Will fail every minute because the user forgot to enter a default value: `Missing variable: 'inputs' on 'hey there {{ inputs.user }}!`

## Question
Can we provide a warning in the UI when an input is used somewhere within a scheduled flow even though it has no defaults + no inputs set on the schedule?

if not possible/easy to do, please close the issue

this is only to help users discover such issues as early as possible

AcevedoR:
otherwise you can save an invalid Flow without knowing it, and it will fail systematically at each schedule

- fixes https://github.com/kestra-io/kestra/issues/2881
- this commit add tests from already existing PR https://github.com/kestra-io/kestra/pull/8107#issuecomment-4030735988 from https://github.com/VTKop

### DIFF
--- a/core/src/main/java/io/kestra/core/validations/validator/FlowValidator.java
+++ b/core/src/main/java/io/kestra/core/validations/validator/FlowValidator.java
@@ -10,7 +10,6 @@ import io.kestra.core.utils.ListUtils;
 import io.kestra.core.validations.FlowValidation;
 import io.kestra.plugin.core.trigger.Schedule;
 import io.micronaut.core.annotation.AnnotationValue;
-import io.micronaut.core.annotation.Introspected;
 import io.micronaut.core.annotation.NonNull;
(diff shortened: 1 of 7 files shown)


loicmathieu (approved):
LGTM

tchiotludo:

Issue #15039, opened 2026-03-12
Failed to use encrypted output from Subflow


danlsgiga:
### Describe the issue

After migrating from `v1.2.9` to `v1.3.2` our flow is failing with 

```
class io.kestra.core.models.flows.Output cannot be cast to class io.kestra.core.models.flows.input.SecretInput (io.kestra.core.models.flows.Output and io.kestra.core.models.flows.input.SecretInput are in unnamed module of loader 'app')
```

### Flows to reproduce

```
id: flow_secret_output_issue
namespace: test

tasks:
  - id: get_secret
    type: io.kestra.plugin.core.flow.Subflow
    namespace: test
    flowId: get_secret
    inputs:
      path: /secret/kestra
    wait: true
    transmitFailed: true
  
  - id: get_token
    type: io.kestra.plugin.scripts.shell.Commands
    containerImage: ubuntu
    taskRunner:
      type: io.kestra.plugin.scripts.runner.docker.Docker      
    retry:
      type: constant
      interval: PT30S
      maxAttempts: 10
      warningOnRetry: false        
    commands:
      - >
        echo -n '{"SECRET": "ASDF"}'
        > {{ execution.id }}
    outputFiles:
      - "{{ execution.id }}"

  - id: kv_set
    type: io.kestra.plugin.core.kv.Set
    key: "{{ execution.id }}"
    value: "{{ read(outputs.get_token.outputFiles[execution.id]) }}"
    kvType: STRING
    ttl: PT15S
    retry:
      type: constant
      interval: PT30S
      maxAttempts: 10
      warningOnRetry: false

outputs:
  - id: token
    type: SECRET
    value: "{{ kv(execution.id) }}"

afterExecution:
  - id: delete_storage_file
    type: io.kestra.plugin.core.storage.Delete
    uri: "{{ outputs.get_token.outputFiles[execution.id] }}"
    retry:
      type: constant
      interval: PT30S
      maxAttempts: 10
      warningOnRetry: false
```

```
id: get_secret
namespace: test

inputs:
  - id: token
    type: SECRET    
    displayName: Token
    required: false
  - id: path
    type: STRING
    displayName: Path

tasks:
  - id: get_secret
    type: io.kestra.plugin.scripts.shell.Commands
    containerImage: ubuntu
    taskRunner:
      type: io.kestra.plugin.scripts.runner.docker.Docker      
    retry:
      type: constant
      interval: PT30S
      maxAttempts: 10
[...]

PR #15091, opened 2026-03-16
fix(core): preserve secret flow outputs without SecretInput cast

proddata:
Fixes a regression due to https://github.com/kestra-io/kestra/pull/14282 in `SECRET` flow outputs introduced by secret input validation changes.

https://github.com/kestra-io/kestra/issues/15039

`FlowInputOutput` started casting every `SECRET`-typed value to `SecretInput`, which is valid for inputs but not for flow outputs. Since flow outputs are plain `Output` objects, that caused ClassCastException when using type: `SECRET` in outputs.

The fix keeps `SecretInput` validation for actual inputs, but skips that input-specific cast for outputs. Secret outputs are still encrypted as before.


loicmathieu (approved):
LGTM

PR #15092, opened 2026-03-16
fix(core): preserve secret flow outputs without SecretInput cast

proddata:
Backport of https://github.com/kestra-io/kestra/pull/15091


loicmathieu:
Backports can be directly cherry-picked to the target branch, we didn't usually open PR for them

PR #15103, opened 2026-03-16
chore: add flakyTests to check 

ammeek:
### ✨ Description

This PR updates the test task configuration so that `check` depends on `flakyTest`.

Previously, `test` and `flakyTest` were executed in separate Gradle invocations in CI, which resulted in multiple build scans and made it harder to view all test results together.

With this change, running `check` will execute both test suites within the same Gradle build.

This ensures that all test results are aggregated into a single build scan.

### 🔗 Related Issue

Relates to:  kestra-io/kestra-ee/issues/7065

### 🛠️ Backend Checklist

### 📝 Additional Notes

The behavior of `flakyTest` itself is unchanged. This PR only ensures that it is executed as part of the `check` lifecycle so both test suites run within the same Gradle build.

There is also a temporary change in this PR's change log to point the PR actions towards `chore/singular-test-build-scan` for testing. These changes will be reverted before the PR is merged.


AcevedoR (approved):
merge actions pr first and switch to kestra-io/actions/composite/setup-build@main and we are good!

PR #15105, opened 2026-03-17
chore: update openapi spec to include Filter description.

ammeek:
update openapi spec to include Filter description.


nkwiatkowski (approved):
LGTM

PR #15112, opened 2026-03-17
fix(cli): ignore namespace revision files during metadata migration

### LINKED ISSUE #13701: [NS-Files] Metadata migration shows `.vX` suffixed files
### Describe the issue

It occurs if some ns files are created and multiple revisions are created after v1.2 (or develop at time of writing) update but before running the migration command. In this case since we query the internal storage we end up seeing the `.vX` suffixed files and thinks they have to be migrated.

Image

Steps to reproduce:
- Kestra v1.2 (or develop)
- Create a ns file and update its content at least once to create a `.v1` file on the storage
- Run the migration command
- `nsFile.v1` will appear in the namespace files browser

### Environment

- Kestra Version: develop (or v1.2.0+ if not fixed before)

shivamwayal37:
### ✨ Description

This PR fixes an issue where namespace revision files (e.g. file.yaml.v1, file.yaml.v2) are incorrectly migrated as regular namespace files during the metadata migration.

During the migration, the storage layer returns both canonical namespace files and their revision versions. The previous implementation did not filter these revision files, causing them to appear in the namespace file browser after migration.

This change adds a check to ignore revision files matching *.v\\d+ before creating NamespaceFileMetadata.

Additionally, a unit test was added to verify that revision files are not migrated.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/13701

### 🛠️ Backend Checklist

### 📝 Additional Notes

A new test shouldNotMigrateRevisionFiles() was added to ensure that files with revision suffixes (e.g. test.yaml.v1) are ignored during metadata migration.

The test mocks the StorageInterface to return both canonical and revision files and verifies that revision files are not saved in NamespaceFileMetadataRepository.

Local testing was performed using:

`./gradlew test`
`./gradlew :cli:test --tests "*MetadataMigrationServiceTest"`


cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -95,22 +95,35 @@ public void kvMigration() throws IOException {
...
-            .flatMap(throwFunction(namespaceForTenant -> {
-                List<PathAndAttributes> list = listAllFromStorage(storageInterface, StorageContext::namespaceFilePrefix, namespaceForTenant.getKey(), namespaceForTenant.getValue());
-                return list.stream()
-                    .map(pathAndAttributes -> NamespaceFileMetadata.of(namespaceForTenant.getKey(), namespaceForTenant.getValue(), pathAndAttributes.path(), pathAndAttributes.attributes()));
-            }))
-            .forEach(throwConsumer(nsFileMetadata -> {
-                if (namespaceFileMetadataRepository.findByPath(nsFileMetadata.getTenantId(), nsFileMetadata.getNamespace(), nsFileMetadata.getPath()).isEmpty()) {
-                    namespaceFileMetadataRepository.save(nsFileMetadata);
-                    if (verbose) {
-                        System.out.println("Migrated namespace file metadata: " + nsFileMetadata.getNamespace() + " - " + nsFileMetadata.getPath());
```
brian-mulier-p, line 123: Keep that log if verbose is active

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -95,22 +95,35 @@ public void kvMigration() throws IOException {
...
+    public void nsFilesMigration(boolean dryRun) throws Exception {
```
brian-mulier-p: It's not a dryRun parameter from the command usage, it's actually a verbose parameter so keep as-is

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -95,22 +95,35 @@ public void kvMigration() throws IOException {
...
+                    FileAttributes attributes = storageInterface.getAttributes(tenant, namespace, uri);
+                    NamespaceFileMetadata metadata = NamespaceFileMetadata.builder()
+                            .tenantId(tenant)
+                            .namespace(namespace)
+                            .path(filename)
+                            .size(attributes.getSize())
+                            .build();
+
+                    if (!dryRun) {
+                        namespaceFileMetadataRepository.save(metadata);
```
brian-mulier-p: You need to keep the idempotency condition to avoid saving a new version over a file that has already been migrated (if file.exist don't save)

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -95,22 +95,35 @@ public void kvMigration() throws IOException {
...
+        Map<String, List<String>> namespacesPerTenant = namespacesPerTenant();
+        for (Map.Entry<String, List<String>> entry : namespacesPerTenant.entrySet()) {
+            String tenant = entry.getKey();
+
+            for (String namespace : entry.getValue()) {
+                List<URI> files = storageInterface.allByPrefix(tenant, namespace, URI.create("kestra://namespace/" + namespace + "/"), true);
+
+                for (URI uri : files) {
+                    String path = uri.getPath();
+                    String filename = path.substring(path.lastIndexOf("/") + 1);
```
brian-mulier-p: You can't do that, if you do it will not keep the directory structure

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -95,22 +95,35 @@ public void kvMigration() throws IOException {
...
+                    // Ignore revision files (.v1, .v2, etc.)
+                    if (filename.matches(".*\\.v\\d+$")) {
+                        continue;
+                    }
+                    FileAttributes attributes = storageInterface.getAttributes(tenant, namespace, uri);
+                    NamespaceFileMetadata metadata = NamespaceFileMetadata.builder()
+                            .tenantId(tenant)
+                            .namespace(namespace)
+                            .path(filename)
+                            .size(attributes.getSize())
```
brian-mulier-p: .version(1)

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -95,22 +95,35 @@ public void kvMigration() throws IOException {
...
+                            .build();
```
brian-mulier-p: You lack populating creation and update time. Please use NamespaceFileMetadata.of as before, it will handle all that for you

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -95,22 +95,35 @@ public void kvMigration() throws IOException {
...
-                    namespaceFileMetadataRepository.save(nsFileMetadata);
-                    if (verbose) {
-                        System.out.println("Migrated namespace file metadata: " + nsFileMetadata.getNamespace() + " - " + nsFileMetadata.getPath());
+    public void nsFilesMigration(boolean dryRun) throws Exception {
+        Map<String, List<String>> namespacesPerTenant = namespacesPerTenant();
+        for (Map.Entry<String, List<String>> entry : namespacesPerTenant.entrySet()) {
+            String tenant = entry.getKey();
+
+            for (String namespace : entry.getValue()) {
+                List<URI> files = storageInterface.allByPrefix(tenant, namespace, URI.create("kestra://namespace/" + namespace + "/"), true);
```
brian-mulier-p: Please don't reinvent the wheel, use the same method as before for retrieving files which was handling file not found

brian-mulier-p: Hardcoded namespace file prefix is wrong and lead to technical debt, please use the same prefix as before

brian-mulier-p (changes requested):
Too many side-effects and regressions. I suggest starting from scratch and adding only the logic of removing the ".vX" suffixed files and not change any behavior beside that

MilosPaunovic:
Hello @shivamwayal37,

shivamwayal37:
Thanks for the detailed feedback!

I understand that the PR introduced unintended changes. I’ll rework it to strictly keep the existing behavior and only add filtering for .vX revision files, without modifying any other logic.

I’ll also rebase on the latest develop and resolve conflicts accordingly.

shivamwayal37:
Hi @brian-mulier-p, 

Thanks for the review — I’ve addressed the feedback.
- Ignored namespace revision files (*.vX) during metadata migration
- Ensured only canonical files are processed
- Added test to verify revision files are not migrated
- Cleaned up stream handling to avoid null values

Could you please take another look when you have time?

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -36,44 +37,45 @@
...
+        KestraConfig kestraConfig,
+        KvMetadataRepositoryInterface kvMetadataRepository) {
         this.flowRepository = flowRepository;
         this.tenantService = tenantService;
-        this.kvMetadataRepository = kvMetadataRepository;
-        this.kvMetadataStateStore = kvMetadataStateStore;
         this.namespaceFileMetadataRepository = namespaceFileMetadataRepository;
         this.storageInterface = storageInterface;
         this.kestraConfig = kestraConfig;
+        this.kvMetadataRepository = kvMetadataRepository;
```
brian-mulier-p: No reordering + keep the constructor as-is to avoid breaking any inheritance

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -36,44 +37,45 @@
...
     }
 
     @VisibleForTesting
     public Map<String, List<String>> namespacesPerTenant() {
         String tenantId = tenantService.resolveTenant();
         return Map.of(
+
```
brian-mulier-p:
```suggestion
```

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -36,44 +37,45 @@
...
+                ).map(NamespaceInterface::asTree).flatMap(Collection::stream).distinct().toList()
+        
         );
     }
 
     public void kvMigration() throws IOException {
         this.namespacesPerTenant().entrySet().stream()
             .flatMap(namespacesForTenant -> namespacesForTenant.getValue().stream().map(namespace -> Map.entry(namespacesForTenant.getKey(), namespace)))
             .flatMap(throwFunction(namespaceForTenant ->
+
```
brian-mulier-p:
```suggestion
```

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -84,6 +86,7 @@ public void kvMigration() throws IOException {
                     .collect(Collectors.partitioningBy(kvEntry -> Optional.ofNullable(kvEntry.expirationDate()).map(expirationDate -> Instant.now().isAfter(expirationDate)).orElse(false)));
 
                 entriesByIsExpired.get(true).forEach(kvEntry ->
+
```
brian-mulier-p:
```suggestion
```

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -99,6 +102,7 @@ public void kvMigration() throws IOException {
                 return entriesByIsExpired.get(false).stream().map(kvEntry -> PersistedKvMetadata.from(namespaceForTenant.getKey(), kvEntry));
             }))
             .forEach(throwConsumer(kvMetadata ->
+
```
brian-mulier-p:
```suggestion
```

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -108,22 +112,32 @@ public void kvMigration() throws IOException {
...
+                            StorageContext::namespaceFilePrefix, namespaceForTenant.getKey(),
+                            namespaceForTenant.getValue());
+                    return list.stream()
+                            .map(pathAndAttributes -> {
+                                if (pathAndAttributes.path().matches(".*\\.v\\d+$")) {
+                                    return null;
+                                }
+                                return NamespaceFileMetadata.of(namespaceForTenant.getKey(), namespaceForTenant.getValue(), pathAndAttributes.path(), pathAndAttributes.attributes());
+                            })
+                            .filter(Objects::nonNull); //avoid nulls
```
brian-mulier-p: filter on !matches then map instead

cli/src/test/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationServiceTest.java
```diff
@@ -10,51 +10,121 @@
...
+
+        FileAttributes attributes = Mockito.mock(FileAttributes.class);
+        Mockito.when(attributes.getSize()).thenReturn(100L);
+
+        Mockito.when(storageInterface.getAttributes(Mockito.any(), Mockito.any(), Mockito.any()))
+                .thenReturn(attributes);
+
+        service.nsFilesMigration(false);
+
+        Mockito.verify(repo, never()).save(Mockito.argThat(meta -> meta.getPath().endsWith(".v1")));
```
brian-mulier-p: also assert the standard file was still saved

cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
```diff
@@ -108,22 +112,32 @@ public void kvMigration() throws IOException {
...
-                return list.stream()
-                    .map(pathAndAttributes -> NamespaceFileMetadata.of(namespaceForTenant.getKey(), namespaceForTenant.getValue(), pathAndAttributes.path(), pathAndAttributes.attributes()));
-            }))
-            .forEach(throwConsumer(nsFileMetadata ->
+                    List<PathAndAttributes> list = listAllFromStorage(storageInterface,
+                            StorageContext::namespaceFilePrefix, namespaceForTenant.getKey(),
+                            namespaceForTenant.getValue());
+                    return list.stream()
+                            .map(pathAndAttributes -> {
+                                if (pathAndAttributes.path().matches(".*\\.v\\d+$")) {
```
brian-mulier-p: Store a private static Pattern.compile then apply it to avoid overhead of Java rebuilding that regex over and over

brian-mulier-p (changes requested):
Some things are still off + please remove all the noise (additions of new lines, unnecessary whitespace addition)

shivamwayal37:
Thanks for the detailed feedback — I see the issues.

I’ll update the PR accordingly:

- Revert the constructor to its original form (no reordering)
- Remove all unrelated formatting/whitespace changes to reduce noise
- Refactor the stream logic to filter before mapping (no null handling)
- Extract the regex into a static Pattern to avoid recompilation
- Extend the test to also assert that canonical files are saved

I’ll push an updated version shortly.

shivamwayal37:
Hi @brian-mulier-p,

Thanks for the feedback — updated accordingly:

- Minimal change: only filters `.vX` files
- Static Pattern used for efficiency
- Stream refactored to filter before map
- Removed unrelated changes/noise
- Tests updated (exclude `.vX`, include canonical files)

Would appreciate another review when you have time 🙏

brian-mulier-p (approved):
LGTM

shivamwayal37:
Hi @loicmathieu,

The PR has been updated addressing all feedback and is now ready for review. Would appreciate your time to take a look 🙏

Many thanks @brian-mulier-p for the detailed guidance and approval  — really appreciate your help throughout!

shivamwayal37:
Thanks a lot @brian-mulier-p for the reviews and guidance throughout this PR — really appreciate your support!

This was a great learning experience, especially around keeping changes minimal and avoiding unintended side effects 🙌

PR #15141, opened 2026-03-18
fix: normalize Docker registry URLs for v2 Bearer token support

### LINKED ISSUE #66: Add Bearer-token support for Docker Registry v2
### Describe the issue

Right now the Docker plugin only does basic auth against `/v1/` registry https://index.docker.io/v1/, which limits us to the old registry endpoint. If you point it at a v2 server https://registry-1.docker.io/v2/, you get a 401 because there’s no token handshake.

It’d be great if the plugin could:
* Detect a `WWW-Authenticate: Bearer` challenge
* Fetch an access token from the registry’s auth service
* Retry the request with `Authorization: Bearer <token>`
* Cache the token until it expires

That way users can push and pull directly against `registry-1.docker.io/v2/` (and other v2-only registries) without falling back to the deprecated v1 API.

## Current workaround 

Atm, user needs to use basic auth against the deprecated v1 Registry API:

```yaml
id: build_dockerhub_image
namespace: company.team

tasks:
  - id: build
    type: io.kestra.plugin.docker.Build
    dockerfile: |
      FROM python:3.12-slim
      RUN pip install kestra
    tags:
      - annageller/kestra:latest
    push: true
    credentials:
      username: annageller
      password: dckr_pat_xxxxxxxx
      registry: https://index.docker.io/v1/ # works
      # registry: https://registry-1.docker.io/v2/ # doesn't work
``` 

### Environment

- Kestra Version: develop

fdelbrayelle:
## Summary
- Add `normalizeRegistryUrl()` to `DockerService` that maps Docker Hub v2 endpoints (`registry-1.docker.io/v2/`, `registry-1.docker.io`, etc.) to the canonical `https://index.docker.io/v1/` config key
- Strip `/v2/` path suffix from other registry URLs (e.g. `ghcr.io/v2/` → `ghcr.io`)
- The Docker daemon already handles v2 Bearer token exchange — credentials just need to be stored under the right key in `config.json`
- Workaround for upstream [docker-java/docker-java#2442](https://github.com/docker-java/docker-java/issues/2442) which lacks native v2 token auth

Fixes kestra-io/plugin-docker#66

## Test plan
- [x] Unit tests for `normalizeRegistryUrl()` covering Docker Hub v2 URLs, canonical v1 URL, other registries with/without `/v2/`, null input, ECR URLs
- [ ] Integration test: `docker pull` with `registry: https://registry-1.docker.io/v2/` credentials


Malaydewangan09 (commented):
LGTM, this solves the Docker Hub case. 
The original issue also mentioned other v2-only registries though, so Bearer token support for private registries is still not fully addressed, IMO.

So, maybe not close the issue?


[EDIT] your fix could work if daemon handles it all for all v2 registries?

script/src/main/java/io/kestra/plugin/scripts/runner/docker/DockerService.java
```diff
@@ -140,6 +140,39 @@ public static Path createConfig(RunContext runContext, @Nullable Object config,
...
+     * Normalizes a registry URL so that Docker Hub endpoints map to the canonical
+     * {@code https://index.docker.io/v1/} key expected by the Docker daemon, and
+     * other registries have the unnecessary {@code /v2/} path stripped.
+     */
+    static String normalizeRegistryUrl(String registry) {
+        if (registry == null) {
+            return null;
+        }
+
+        String canonical = "https://index.docker.io/v1/";
```
loicmathieu: Better to have it as a constant

loicmathieu (approved):
LGTM

fdelbrayelle:
## QA

```yaml
id: test_docker_registry_v2
namespace: company.team

tasks:
  - id: build_and_push
    type: io.kestra.plugin.docker.Build
    dockerfile: |
      FROM python:3.12-slim
      RUN pip install kestra
    tags:
      - fdelbrayelle/custom-test-image
    push: true
    credentials:
      registry: https://registry-1.docker.io/v2/
      username: "{{ secret('DOCKERHUB_USERNAME') }}"
      password: "{{ secret('DOCKERHUB_PASSWORD') }}"

  - id: pull_v2
    type: io.kestra.plugin.docker.Pull
    image: fdelbrayelle/custom-test-image
    credentials:
      registry: https://registry-1.docker.io/v2/
      username: "{{ secret('DOCKERHUB_USERNAME') }}"
      password: "{{ secret('DOCKERHUB_PASSWORD') }}"
```

Results:

image

script/src/main/java/io/kestra/plugin/scripts/runner/docker/DockerService.java
fdelbrayelle: Done: https://github.com/kestra-io/kestra/pull/15141/changes/1c225030a4826f6ae58e04ab882171764b649495

PR #15150, opened 2026-03-19
fix(core): condition must be required in the if task

anna-geller:
closes https://github.com/kestra-io/kestra-ee/issues/7055


core/src/test/java/io/kestra/core/docs/JsonSchemaGeneratorTest.java
```diff
@@ -200,6 +201,20 @@ void dag() throws URISyntaxException {
         });
     }
 
+    @SuppressWarnings("unchecked")
+    @Test
+    void ifTask() throws URISyntaxException {
```
loicmathieu: We usually don't add test for each property change here as otherwise we would have hundreds of them ;)

loicmathieu (approved):
LGTM

core/src/test/java/io/kestra/core/docs/JsonSchemaGeneratorTest.java
anna-geller: It's cut now. Thanks so much!

PR #15147, opened 2026-03-19
fix: replay to restart non-terminated parallel siblings

thyw1:
Closes  #11223

**Description**  
When replaying a single task inside a Parallel/DAG, non‑terminated sibling task runs from the previous execution were inherited as RUNNING in the new execution. These task runs are never rescheduled, so the execution can get stuck in RUNNING.  
This change restarts those non‑terminated siblings by remapping their original task runs to new taskRunIds and marking them RESTARTED.

**Behavior Change**  
Replay now restarts any non‑terminated sibling task runs (not part of the replay subtree), instead of inheriting them as RUNNING.

**Tests**  
`./gradlew :core:test --tests io.kestra.core.runners.ExecutionServiceTest.replayParallelRestartsRunningSibling`

PR #15153, opened 2026-03-19
fix: replay mark non-terminated sibling attempts as killed

thyw1:
This is a follow‑up to #15147 

When replaying a single task inside Parallel/DAG, sibling tasks that were still RUNNING in the original execution are now restarted (from previous fix). 
However the attempt history still shows the original RUNNING attempt, which is confusing.

This PR normalizes those attempts by marking the previous non‑terminated attempt as KILLED (via TaskRun.onRunningResend()) before remapping.
Behavior is unchanged (tasks are restarted), but attempt history becomes consistent.


loicmathieu (approved):
LGMT, can you update the test so that it checks that the attempts are also in the RESTARTED state

thyw1:
> LGMT, can you update the test so that it checks that the attempts are also in the RESTARTED state

From a replay on an unmodified flowable child node, I can see that `RESTARTED` is present on `taskRun.state.histories`, but not on `attempt.state` histories.
The attempts still represent actual worker executions (e.g. SUCCESS/SUCCESS), while the taskRun itself carries the RESTARTED transition.

Do you want the new test to assert the taskRun restart plus the new attempt record, or do you actually want to change the attempt semantics as well?

loicmathieu (approved):
LGTM, thanks

loicmathieu (changes requested):
> Do you want the new test to assert the taskRun restart plus the new attempt record,

Yes please, update an existing test to assert that the taskrun has been changed

loicmathieu:
@thyw1 are you willing to update the existing test as asked or do you want me to takeover this?

thyw1:
> @thyw1 are you willing to update the existing test as asked or do you want me to takeover this?

Apologies for the delay.

I updated the existing `replayParallelRestartsRunningSibling` test as requested.

It now asserts that the replayed sibling task run is changed to `RESTARTED`, that the restarted task run has a new id, and that `onRunningResend()` normalizes attempts by appending a `RESUBMITTED` attempt before remapping.

loicmathieu (approved):
LGMT, thanks

loicmathieu:
Test fail in an unrelated area, merging it.

loicmathieu:
Backported to 1.3 and 1.0

PR #15154, opened 2026-03-19
chore(system): don't create repository beans in workers

loicmathieu:


core/src/main/java/io/kestra/core/repositories/RepositoryBean.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Defines a queue as a Micronaut bean.
```
fhussonnois: Just the comment is wrong: Defines a repository.

fhussonnois (approved):
Nice! LGTM!

core/src/main/java/io/kestra/core/repositories/RepositoryBean.java
loicmathieu: :facepalm:

PR #15157, opened 2026-03-19
fix(build): replace Copy task with plain task for generateOpenapiSpec

brian-mulier-p:
Gradle 9.4 treats the `into` directory of a Copy task as a declared output. Since generateOpenapiSpec copied into the root project directory, Gradle inferred that this task owned the entire project tree, causing implicit dependency errors for every task in every module (jar, processResources, compileJava, sourcesJar, etc.).

Switching to a plain task with explicit inputs/outputs scoped to the single openapi.yml file eliminates the false ownership claim.

### DIFF
--- a/webserver/build.gradle
+++ b/webserver/build.gradle
@@ -53,13 +53,17 @@ tasks.withType(JavaCompile).configureEach {
         "-Amicronaut.openapi.expand.version=${project.version}"
     ]
 }
-tasks.register('generateOpenapiSpec', Copy) {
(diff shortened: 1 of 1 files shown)


ammeek (approved):
LGTM

AcevedoR:
good for me, ran validation script 3 and it was okay too https://develocity.kestra.io/c/v6hqnfukcxx5e/yyssoosnsbfzi/task-inputs?cacheability=cacheable,overlapping-outputs,validation-failure

PR #15160, opened 2026-03-19
feat(worker): forward cluster events to gRPC workers via unified events field

fhussonnois:
Replace the killCommands proto field with a polymorphic events field that carries both ExecutionKilled and ClusterEvent via WorkerBroadcastEvent. This allows cluster events like PLUGINS_SYNC_REQUESTED to reach remote gRPC workers which use an InMemoryBroadcastQueue for ClusterEvent.


loicmathieu (approved):
LGTM

PR #15164, opened 2026-03-20
fix(jdbc): stabilize `deduplicateNexts` for large payload values

### LINKED ISSUE #15148: Fix Executor failure when `taskRun.getValue()` is too large in `deduplicateNexts`
## Description
In high-scale data processing scenarios (e.g., a `ForEach` task processing a large list of items), the Executor may fail to persist its state due to database physical limits. The root cause lies in the `deduplicateNexts` method within the Executor implementations (`JdbcExecutor`). This method generates a deduplication key by concatenating various fields, including the full string representation of `taskRun.getValue()`.

When `taskRun.getValue()` contains a large amount of data (e.g., a multi-megabyte JSON or a long list of paths), the resulting key can exceed the **64KB limit** imposed by MySQL for JSON object keys (and similar limits in other databases).

## Error Message
Users will encounter an error similar to:
```text
org.jooq.exception.DataException: SQL [insert into `executorstate` (`key`, `value`) values (?, ?) on duplicate key update `value` = ?]; 
Data truncation: The JSON object contains a key name that is too long.
```

## Impact
**Executor Instability**: The entire database transaction rollbacks, preventing the execution from progressing. 

## Suggested Optimization
Implement **Partial Hashing** for the `value` component of the deduplication key:
1. **Threshold Check**: Check the length of `taskRun.getValue()`.
2. **Hashing**: If the length exceeds a safe threshold (e.g., 1KB), replace the value part in the key with a hashed version (e.g., `hashed_` + SHA256 hex).
3. **Backward Compatibility**: The deduplication logic should check both the new hashed key format and the legacy raw key format to ensure zero-downtime upgrades.

## Proposed Code Change (Conceptual)
```java
String rawValue = String.valueOf(taskRun.getValue());
// Only hash the value part if it exceeds 1KB to keep the key length safe
String safeValue = (rawValue.length() > 1 * 1024) ? 
    "hashed_" + Hashing.sha256().hashString(rawValue, StandardCharsets.UTF_8) : rawValue;

String safeKey = taskRun.getParentTaskRunId() + "-" + taskRun.getTaskId() + "-" + safeValue + ...;

// Check new key
if (executorState.getChildDeduplication().containsKey(safeKey)) {
    return false;
}
// Fallback check for legacy raw keys to ensure compatibility during upgrade
if (!safeKey.equals(rawKey) && executorState.getChildDeduplication().containsKey(rawKey)) {
    return false;
}
```

aulightzxy-web:
### ✨ Description

This PR improves backend deduplication stability in `JdbcExecutor` for large task-run payloads.

- Updates `deduplicateNexts(...)` to avoid oversized deduplication keys when `taskRun.value` is large.
- Uses a compatibility-safe dual-key strategy:
  - Keeps the legacy raw key format for backward-compatibility checks.
  - Uses a `safeKey` that hashes only the value segment (`sha256`) when value length is greater than 1024.
  - Persists deduplication entries with `safeKey` to reduce key size risk.
- Adds unit tests to validate:
  - hashed-key behavior for large payloads,
  - legacy raw-key compatibility for in-flight executions,
  - unchanged behavior for small payloads.

### 🔗 Related Issue

Closes #15148

### 🛠️ Backend Checklist

### 📝 Additional Notes

- Targeted validation executed:
  - `./gradlew :jdbc:test --tests io.kestra.jdbc.runner.JdbcQueueConfigurationTest --tests io.kestra.jdbc.runner.JdbcExecutorDeduplicationTest`
- The new deduplication tests are added in:
  - `jdbc/src/test/java/io/kestra/jdbc/runner/JdbcQueueConfigurationTest.java`

### 🤖 AI Authors

Why did the cat become a backend engineer? Because it always pawsed before shipping race conditions. 🐱

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
@@ -1,6 +1,7 @@
 package io.kestra.jdbc.runner;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.hash.Hashing;
 import io.kestra.core.contexts.KestraContext;
 import io.kestra.core.exceptions.DeserializationException;
 import io.kestra.core.exceptions.FlowNotFoundException;
@@ -61,6 +62,7 @@ import org.jooq.Configuration;
 import org.slf4j.event.Level;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.*;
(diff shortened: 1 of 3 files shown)


jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
```diff
@@ -1464,21 +1462,40 @@ private boolean deduplicateNexts(Execution execution, ExecutorState executorStat
...
+                // 3. Construct the structured safe key
+                String safeKey = taskRun.getParentTaskRunId() + "-" +
+                    taskRun.getTaskId() + "-" +
+                    safeValue + "-" +
+                    (taskRun.getAttempts() != null ? taskRun.getAttempts().size() : 0)
+                    + taskRun.getIteration();
+
+                // 4. Dual-check mechanism (compatible with existing in-flight jobs)
+                if (executorState.getChildDeduplication().containsKey(safeKey)) {
+                    log.trace("Duplicate Nexts on execution '{}' with key '{}'", execution.getId(), safeKey);
```
loicmathieu:
```suggestion
                    log.warn("Duplicate Nexts on execution '{}' with key '{}'", execution.getId(), safeKey);
```

jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
```diff
@@ -1464,21 +1462,40 @@ private boolean deduplicateNexts(Execution execution, ExecutorState executorStat
...
                     return false;
-                } else {
-                    executorState.getChildDeduplication().put(deduplicationKey, taskRun.getId());
-                    return true;
                 }
+
+                if (!safeKey.equals(rawKey) && executorState.getChildDeduplication().containsKey(rawKey)) {
+                    log.trace("Duplicate Nexts (Legacy) on execution '{}' with key '{}'", execution.getId(), rawKey);
```
loicmathieu:
```suggestion
                    log.warn("Duplicate Nexts (Legacy) on execution '{}' with key '{}'", execution.getId(), rawKey);
```

jdbc/src/test/java/io/kestra/jdbc/runner/JdbcQueueConfigurationTest.java
```diff
@@ -56,17 +66,95 @@ void shouldCompute6Steps() {
...
         );
     }
+}
 
-    @Test
-    void shouldComputeSingleStepWhenMinEqualsMax() {
-        var configuration = new JdbcQueue.Configuration();
-        configuration.maxPollInterval = Duration.ofSeconds(1);
-        configuration.minPollInterval = Duration.ofSeconds(1);
+class JdbcExecutorDeduplicationTest {
```
loicmathieu: Please extract to its own test class

loicmathieu (changes requested):

aulightzxy-web:
I have made the revisions as you requested.

loicmathieu (approved):
LGTM, thanks

loicmathieu:
Backported to 1.3

PR #15180, opened 2026-03-20
fix(core): fix GraphUtils.successors() to include error handler tasks

brian-mulier-p:
Closes kestra-io/kestra-ee#6959


Skraye (approved):
LGTM

PR #15186, opened 2026-03-23
feat(executions): make execution button actions context-dependent

### LINKED ISSUE #13702: [UI] Replace "Execute" Button on Execution page with context-dependend action button.
### Feature description

Currently, the primary action button on the Execution pages is “Execute”, which does not make much sense in this context, since the user is already viewing an execution. Instead, the primary action should depend on the current execution state—for example Pause for a running execution, Resume for a paused one, or Replay for a finished execution (successful or failed).

For users, this is helpful because the most relevant action is always accessible from anywhere on the execution page. A possible caveat is that for very fast executions the action may change quickly, but this seems like a reasonable trade-off and worth trying.

Image

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/13702.

- `Resume` is shown when the execution is `PAUSED`.
- `Pause` is shown when the execution is currently running.
- `Restart` is shown when the execution state is `FAILED`.
- `Replay` is shown for other terminated/finished states from `State.getTerminatedStates()`.
- `Execute` is shown only as a fallback when none of the state-specific actions above apply and the user has trigger permission.


MilosPaunovic:
@Pradumnasaraf Can you double check the behavior described in the description of this PR please? 🙏🏻

Piyush-r-bhaskar (approved):

Pradumnasaraf (approved):
Hey,

Tested and working as expected.
- Resume shows for PAUSED
- Pause shows for running executions
- Restart shows for FAILED
- Replay shows for other finished/terminated states
- Execute is not shown when any of the above applies

Love this chnage! A nice UX addition. Looks good to merge

PR #15192, opened 2026-03-23
refactor WorkerJob/WorkerTrigger models

fhussonnois:


loicmathieu (approved):
LGMT, I'm not a fan of this ExecutorWorkerTaks but I cannot see another way so let's ship it.

PR #15194, opened 2026-03-23
perf(core): improve resolving task runs in Execution and FlowableUtils

MTarek165:
- findTaskRunByTasks() is extremely important for resolving task runs at executor life cycle but it performs in a brute force way when finding task runs  from resolved tasks performing  in O(T*R) where T is the number of task runs and R is the number of resolved tasks which may perform worse under heavy load
- The proposed change is pre-storing either resolved tasks or task runs in a map by ids using HashMap lookup in O(1) avoiding brute force check turning performance to O(T + R) which is much better under heavy load
- This change enhances performance keeping the readability concern clear by encapsulating the logic into isResolvedTaskFor() and isTaskRunFor() methods

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowableUtils.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowableUtils.java
@@ -16,6 +16,7 @@ import io.kestra.plugin.core.flow.Dag;
 
 import java.util.*;
 import java.util.function.BiFunction;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -460,14 +461,13 @@ public class FlowableUtils {
         if (concurrency > 0 && runningCount > concurrency) {
             return Collections.emptyList();
         }
-
+        Map<String, List<TaskRun>> taskRunsByIds = taskRuns.stream().collect(Collectors.groupingBy(
+            TaskRun::getTaskId
+        ));
         // find all not created tasks
         List<ResolvedTask> notFinds = currentTasks
             .stream()
-            .filter(resolvedTask -> taskRuns
-                .stream()
-                .noneMatch(taskRun -> FlowableUtils.isTaskRunFor(resolvedTask, taskRun, parentTaskRun))
-            )
+            .filter(resolvedTask -> !FlowableUtils.isTaskRunFor(resolvedTask, parentTaskRun, taskRunsByIds))
             .toList();
 
         // first created, leave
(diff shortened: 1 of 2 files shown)


loicmathieu:
Thanks, I have a PR somewhere with some changes also in this area and have no time now to check if this one provides more improvements or not so I'll freeze it for now. Will need so time to sort this out later.

MilosPaunovic:
Hello @MTarek165,

core/src/main/java/io/kestra/core/runners/FlowableUtils.java
```diff
@@ -565,13 +565,39 @@ else if(obj instanceof Map mapObj) {
...
+        if (parentTaskRun != null && !parentTaskRun.getId().equals(taskRun.getParentTaskRunId())) {
+            return false;
+        }
+
+        List<ResolvedTask> matchingTasks = resolvedTasksByIds.get(taskRun.getTaskId());
+        if (matchingTasks == null) {
+            return false;
+        }
+
+        for (ResolvedTask rt : matchingTasks) {
```
loicmathieu, line 578: In the end, this will still be a nested loop

core/src/main/java/io/kestra/core/runners/FlowableUtils.java
```diff
@@ -565,13 +565,39 @@ else if(obj instanceof Map mapObj) {
...
+        return false;
+    }
+
+    public static boolean isTaskRunFor(ResolvedTask resolvedTask, TaskRun parentTaskRun, Map<String, List<TaskRun>> taskRunsByIds) {
+        List<TaskRun> matchingRuns = taskRunsByIds.get(resolvedTask.getTask().getId());
+        if (matchingRuns == null) {
+            return false;
+        }
+
+        for (TaskRun tr : matchingRuns) {
```
loicmathieu, line 592: In the end, this will still be a nested loop

loicmathieu (changes requested):
I'm not 100% sure that it provides a performance improvements, in some happy path we may avoid a nested loop but maybe not that much cases.

Anyway, if you rebase it, I may be able to test it in our benchmark and see if it provides any benefits

loicmathieu:
@MTarek165 thanks a lot for your contribution but I took another approach here that totally remove the nested loop: https://github.com/kestra-io/kestra/pull/15245

MTarek165:
@loicmathieu It is optimization, The point behind nested loop was the assumption that a task id can correspond to list of resolved tasks like in ForEach , I understand your PR using combination of id and value to make it unique, nice you did it better 😄
thanks too Loic.

PR #15199, opened 2026-03-24
fix(ai): no more NPE when sending no initial yaml

brian-mulier-p:


loicmathieu (approved):
LGTM

Pradumnasaraf (approved):
Looks good!

Now everything is working fine:

- Running simply the free trial or custom AI config (no switching)
- Switching from free trial to AI config and vice versa.
- The remaining message doesn't show up when we switch to the AI config from the free trial, and the message comes back when we switch back to the free one!

PR #15203, opened 2026-03-24
fix(storage-local): handle filenames with spaces and special characters in list/allByPrefix

### LINKED ISSUE #15200: bug: LocalStorage throws IllegalArgumentException for filenames containing spaces
## Bug Description

`LocalStorage` throws `IllegalArgumentException: Illegal character in path` when listing files or resolving namespace files whose filenames contain spaces (or other URI-illegal characters such as `#`, `?`, `%`).

## Affected Methods

Three methods in `storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java`:

1. `list()`
2. `listInstanceResource()`
3. `allByPrefix()`

## Steps to Reproduce

1. Upload a file with a space in its name to a namespace:

```
PUT /api/v1/namespaces/{namespace}/files?path=folder/file with space.txt
```

2. List the directory contents:

```
GET /api/v1/namespaces/{namespace}/files/directory?path=folder
```

3. Or execute a flow that uses `namespaceFiles` to copy namespace files — triggers `allByPrefix()`.

**Result:** `IllegalArgumentException: Illegal character in path at index N: ...file with space.txt`

**Expected:** Files with spaces in their names are returned normally.

### `list()` and `listInstanceResource()`

```java
// BUGGY — lines 144-149 and 163-169
URI relative = URI.create(
    getLocalPath(tenantId, null).relativize(
        Path.of(file.toUri())   // <- decodes %20 back to a space
    ).toString().replace("\\", "/")
    // the string now contains a raw space
);
// URI.create() treats space as illegal per RFC 2396 -> throws IllegalArgumentException
```

The sequence `Path -> toUri() -> Path.of(URI) -> relativize() -> toString()` introduces a decode-then-re-encode round-trip. `Path.of(file.toUri())` decodes `%20` back to a literal space, and then `URI.create()` (which does **not** auto-encode) rejects it.

### `allByPrefix()`

```java
// BUGGY — lines 99-113 and 122-125
String dirPath = dir.toString().replace("\\", "/");  // raw OS path, may contain spaces
uris.add(URI.create(dirPath + "/"));                  // crashes on spaces

uris.add(URI.create(file.toString().replace("\\", "/"))); // same issue

URI fsPathUri = URI.create(fsPath.toString().replace("\\", "/")); // same issue

.map(URI::getPath)   // should be getRawPath() to keep percent-encoding intact
```

`Path.toString()` returns the raw OS path (with spaces), and `URI.create()` does not percent-encode them.

### For `list()` and `listInstanceResource()`

Add a helper method that stays in the percent-encoded domain throughout:

```java
private static URI toRelativeUri(Path basePath, Path file) {
    String baseRawPath = basePath.toUri().getRawPath();
    if (!baseRawPath.endsWith("/")) {
        baseRawPath += "/";
    }
    String fileRawPath = file.toUri().getRawPath();
    String encodedRelative = fileRawPath.startsWith(baseRawPath)
        ? fileRawPath.substring(baseRawPath.length())
        : fileRawPath;
    return URI.create(encodedRelative);
}
```

Then replace the buggy block in both methods:

```java
// BEFORE
URI relative = URI.create(
    getLocalPath(tenantId, null).relativize(
        Path.of(file.toUri())
    ).toString().replace("\\", "/")
);

// AFTER
URI relative = toRelativeUri(getLocalPath(tenantId, null), file);
```

### For `allByPrefix()`

```java
// preVisitDirectory — BEFORE
String dirPath = dir.toString().replace("\\", "/");
uris.add(URI.create(dirPath + "/"));
return super.preVisitDirectory(Path.of(dirPath), attrs);

// preVisitDirectory — AFTER
if (includeDirectories) {
    uris.add(URI.create(dir.toUri().getRawPath() + "/"));
}
return super.preVisitDirectory(dir, attrs);

// visitFile — BEFORE
uris.add(URI.create(file.toString().replace("\\", "/")));

// visitFile — AFTER
uris.add(URI.create(file.toUri().getRawPath()));

// fsPathUri — BEFORE
URI fsPathUri = URI.create(fsPath.toString().replace("\\", "/"));

// fsPathUri — AFTER
URI fsPathUri = URI.create(fsPath.toUri().getRawPath());

// stream map — BEFORE
.map(URI::getPath)

// stream map — AFTER
.map(URI::getRawPath)
```

### Why `getRawPath()` instead of `getPath()`

`URI.getPath()` decodes percent-encoded characters (e.g. `%20` -> space), so passing its result to `URI.create()` crashes on spaces. `URI.getRawPath()` keeps the encoding intact (`%20` stays `%20`), which is a valid URI character.

### Why the new approach does not need `.replace("\\", "/")`

`Path.toUri().getRawPath()` always uses `/` as the separator — it follows the URI standard, which mandates `/` regardless of the host OS. `Path.toString()` is OS-dependent (backslash on Windows), hence the old `.replace()` workaround.

## Environment

- Kestra version: `1.4.0-SNAPSHOT` (also confirmed on earlier branches)
- Storage backend: `local`
- Java: 21

## Note on Non-ASCII Filenames

Filenames containing only non-ASCII characters (e.g. Chinese `中文.txt`) do **not** trigger the crash. Java's `URI.create()` permits characters above `0x7F` (classified as `other` per RFC 2396), but explicitly rejects space (`0x20`). Other characters that would also cause crashes include `#`, `?`, `[`, `]`.

FourFriends:
## Summary

Fixes #15200

`URI.create()` rejects raw spaces (RFC 2396), causing `IllegalArgumentException: Illegal character in path` when listing namespace files or running flows with `namespaceFiles` if any filename contains a space.

## Root Cause

Three methods were affected in `LocalStorage`:

**`list()` / `listInstanceResource()`**

The chain `Path.of(file.toUri()).relativize().toString()` decodes `%20` back to a literal space, which `URI.create()` then rejects:

```java
// BEFORE
URI relative = URI.create(
    getLocalPath(tenantId, null).relativize(
        Path.of(file.toUri())  // decodes %20 -> space
    ).toString().replace("\\", "/")
);
```

**`allByPrefix()`**

`Path.toString()` returns the raw OS path (spaces intact); `URI.create()` rejects it. Additionally, `URI::getPath` strips percent-encoding from the stream result:

```java
// BEFORE
uris.add(URI.create(file.toString().replace("\\", "/")));  // spaces crash URI.create
.map(URI::getPath)  // strips %20 encoding
```

## Fix

- Add a `toRelativeUri(Path basePath, Path file)` helper that stays in the percent-encoded domain throughout, using `toUri().getRawPath()` for both paths and a plain string prefix-strip instead of `Path.relativize()`.
- Replace all four `URI.create(path.toString()…)` calls in `allByPrefix()` with `URI.create(path.toUri().getRawPath())`, and change `URI::getPath` → `URI::getRawPath` in the stream pipeline.
- The explicit `.replace("\\", "/")` Windows workaround is no longer needed since `toUri().getRawPath()` always uses `/` per the URI standard.

```java
// AFTER — shared helper
private static URI toRelativeUri(Path basePath, Path file) {
    String baseRawPath = basePath.toUri().getRawPath();
    if (!baseRawPath.endsWith("/")) baseRawPath += "/";
    String fileRawPath = file.toUri().getRawPath();
    String encodedRelative = fileRawPath.startsWith(baseRawPath)
        ? fileRawPath.substring(baseRawPath.length())
        : fileRawPath;
    return URI.create(encodedRelative);
}
```

## Tests

Two test cases added to `StorageTestSuite`:

- `listWithSpaceInFilename()` — verifies `list()` returns files with spaces in their names without throwing.
- `filesByPrefixWithSpaceInFilename()` — verifies `allByPrefix()` returns correctly percent-encoded URIs (e.g. `file%20with%20space.txt`) for files with spaces.

## Note on Non-ASCII Filenames

Filenames with non-ASCII characters only (e.g. Chinese `中文.txt`) did not previously crash because Java's `URI.create()` permits characters above `0x7F`. The space character (`0x20`) is explicitly listed as illegal in RFC 2396. Other characters that would also trigger the crash include `#`, `?`, `[`, `]`.

### DIFF
--- a/storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java
+++ b/storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java
@@ -112,17 +112,16 @@ public class LocalStorage implements StorageInterface {
         Files.walkFileTree(fsPath, new SimpleFileVisitor<>() {
             @Override
             public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) throws IOException {
-                String dirPath = dir.toString().replace("\\", "/");
                 if (includeDirectories) {
-                    uris.add(URI.create(dirPath + "/"));
+                    uris.add(URI.create(dir.toUri().getRawPath() + "/"));
                 }
-                return super.preVisitDirectory(Path.of(dirPath), attrs);
+                return super.preVisitDirectory(dir, attrs);
             }
 
             @Override
             public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
                 if (!file.getFileName().toString().endsWith(".metadata")) {
-                    uris.add(URI.create(file.toString().replace("\\", "/")));
+                    uris.add(URI.create(file.toUri().getRawPath()));
                 }
                 return FileVisitResult.CONTINUE;
             }
@@ -135,10 +134,10 @@ public class LocalStorage implements StorageInterface {
             }
         });
 
-        URI fsPathUri = URI.create(fsPath.toString().replace("\\", "/"));
+        URI fsPathUri = URI.create(fsPath.toUri().getRawPath());
         return uris.stream().sorted(Comparator.reverseOrder())
             .map(fsPathUri::relativize)
-            .map(URI::getPath)
+            .map(URI::getRawPath)
             .filter(Predicate.not(String::isEmpty))
             .map(path ->
             {
@@ -160,11 +159,7 @@ public class LocalStorage implements StorageInterface {
                 .filter(path -> !path.getFileName().toString().endsWith(".metadata"))
                 .map(throwFunction(file ->
                 {
-                    URI relative = URI.create(
-                        getLocalPath(tenantId, null).relativize(
-                            Path.of(file.toUri())
-                        ).toString().replace("\\", "/")
-                    );
+                    URI relative = toRelativeUri(getLocalPath(tenantId, null), file);
                     return getAttributes(tenantId, namespace, relative);
                 }))
                 .toList();
@@ -180,17 +175,30 @@ public class LocalStorage implements StorageInterface {
                 .filter(path -> !path.getFileName().toString().endsWith(".metadata"))
                 .map(throwFunction(file ->
                 {
-                    URI relative = URI.create(
-                        getInstancePath(null).relativize(
-                            Path.of(file.toUri())
-                        ).toString().replace("\\", "/")
-                    );
+                    URI relative = toRelativeUri(getInstancePath(null), file);
                     return getInstanceAttributes(namespace, relative);
                 }))
                 .toList();
         } catch (NoSuchFileException e) {
             throw new FileNotFoundException(e.getMessage());
         }
+    }
+
+    /**
+     * Converts a file path to a relative URI against the given base path.
+     * Uses the encoded raw path from file.toUri() to safely handle filenames
+     * containing special characters such as spaces or non-ASCII characters.
+     */
+    private static URI toRelativeUri(Path basePath, Path file) {
+        String baseRawPath = basePath.toUri().getRawPath();
+        if (!baseRawPath.endsWith("/")) {
+            baseRawPath += "/";
+        }
+        String fileRawPath = file.toUri().getRawPath();
+        String encodedRelative = fileRawPath.startsWith(baseRawPath)
+            ? fileRawPath.substring(baseRawPath.length())
+            : fileRawPath;
+        return URI.create(encodedRelative);
     }
 
     @Override
--- a/storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java
+++ b/storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java
@@ -188,6 +188,9 @@ public class LocalStorage implements StorageInterface {
      * Converts a file path to a relative URI against the given base path.
      * Uses the encoded raw path from file.toUri() to safely handle filenames
      * containing special characters such as spaces or non-ASCII characters.
+     *
+     * @throws IllegalArgumentException if the file path is not under the base path,
+     *         preventing path traversal attacks
      */
     private static URI toRelativeUri(Path basePath, Path file) {
         String baseRawPath = basePath.toUri().getRawPath();
@@ -195,10 +198,12 @@ public class LocalStorage implements StorageInterface {
             baseRawPath += "/";
         }
         String fileRawPath = file.toUri().getRawPath();
-        String encodedRelative = fileRawPath.startsWith(baseRawPath)
-            ? fileRawPath.substring(baseRawPath.length())
-            : fileRawPath;
-        return URI.create(encodedRelative);
+        if (!fileRawPath.startsWith(baseRawPath)) {
+            throw new IllegalArgumentException(
+                "File path '" + fileRawPath + "' is not under base path '" + baseRawPath + "'"
+            );
+        }
+        return URI.create(fileRawPath.substring(baseRawPath.length()));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


MilosPaunovic:
Hello @FourFriends,

FourFriends:
@MilosPaunovic get, I will resolve it.

FourFriends:
The E2E test failure in this PR is a pre-existing CI infrastructure issue unrelated to this change. Checking the last several PR workflow runs, E2E tests have been consistently failing across all of
  them (e.g. runs for #15261, #15257, and others). 

The failure occurs within ~4 minutes — far too early for the Docker image build to complete — and no Playwright report is generated, suggesting the job
  crashes during setup rather than during actual test execution (likely due to missing registry credentials for fork PRs).

The Backend Tests pass successfully for this PR, confirming the fix itself is correct.

FourFriends:
@MilosPaunovic  please review this PR , thanks.

MilosPaunovic:
@loicmathieu Will be the one reviewing this one, and he'll do it as his earliest convenience, so no need to directly ping anyone.

FourFriends:
@loicmathieu can you help review this pr, thanks

storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java
```diff
@@ -178,6 +169,23 @@ public List<FileAttributes> listInstanceResource(@Nullable String namespace, URI
         }
     }
 
+    /**
+     * Converts a file path to a relative URI against the given base path.
+     * Uses the encoded raw path from file.toUri() to safely handle filenames
+     * containing special characters such as spaces or non-ASCII characters.
+     */
+    private static URI toRelativeUri(Path basePath, Path file) {
```
loicmathieu, line 177: I wonder if it is not subject to path traversal vulnerability.
I may need more time to analyse.

FourFriends: okay，please contact me if you have questions, thanks.

storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java
```diff
@@ -178,6 +169,23 @@ public List<FileAttributes> listInstanceResource(@Nullable String namespace, URI
...
+        String baseRawPath = basePath.toUri().getRawPath();
+        if (!baseRawPath.endsWith("/")) {
+            baseRawPath += "/";
+        }
+        String fileRawPath = file.toUri().getRawPath();
+        String encodedRelative = fileRawPath.startsWith(baseRawPath)
+            ? fileRawPath.substring(baseRawPath.length())
+            : fileRawPath;
```
loicmathieu, line 185: There is a potential path traversial issue here as we previously use Path.relalivize(), this can be mitiguated by a guard here:

```java
new IllegalArgumentException(
          "File path '" + fileRawPath + "' is not under base path '" + baseRawPath + "'");                                                                                                                                                                          
  }                                                                                                                                                                                                                                                                 
  String encodedRelative = fileRawPath.substring(baseRawPath.length());  
```

can you update your code?

storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java
loicmathieu: There is one, see the comment bellow

loicmathieu (changes requested):
LGTM.

@kestra-io/plugins after merging this fix inside the LocalStorage, it may cause issues in other storage if they suffer for the same special character support issue.

storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java
FourFriends: get。

FourFriends: @loicmathieu please help review this pr, thanks.

FourFriends:
@loicmathieu I would like to know if you have other questions?  thanks.

loicmathieu:
@FourFriends thanks for your contribution but I just merged https://github.com/kestra-io/kestra/pull/16178 wich fixes the same issue.

PR #15204, opened 2026-03-24
fix(core): sort and filter ForEachItem merge outputs for consistent ordering

fhussonnois:
The ForEachItemMergeOutputs task listed iteration directories without sorting or filtering, causing subflowOutputs to have non-deterministic ordering across runs depending on the storage backend.

Filter for directories only and sort numerically by iteration index.

Fixes kestra-io/kestra-ee#7122


loicmathieu (approved):
LGTM

PR #15209, opened 2026-03-24
Fix/ai copilot to nocode

brian-mulier-p:
closes https://github.com/kestra-io/kestra-ee/issues/5711

### DIFF
--- a/ui/src/components/inputs/EditorWrapper.vue
+++ b/ui/src/components/inputs/EditorWrapper.vue
@@ -1,73 +1,60 @@
 <template>
-    <div class="h-100 d-flex flex-column">
-        <img
-            v-if="['jpg', 'jpeg', 'png', 'gif', 'webp', 'webm', 'avif'].includes(extension)"
-            :src="`${apiUrl()}/namespaces/${namespace}/files?path=/${path}`"
-            class="image-preview"
-        >
-        <Editor
-            v-else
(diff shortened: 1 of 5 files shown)


ui/src/components/inputs/EditorWrapper.vue
```diff
@@ -430,76 +430,7 @@
...
-    }
-
-    @media (max-width: 480px) {
-        .prompt {
-            width: calc(100% - 1rem);
-            left: 0.5rem;
-            bottom: 2%;
-        }
-    }
+    @import "../ai/ai-copilot-popup.scss";
```
elevatebart: This feels very unsafe as the whole principle of a scoped css is that it "IS" the component. Having it in a separate file makes it so that it becomes harder to maintain and reason with. Plus tsh tooling odes not expect it.

ui/src/components/no-code/NoCode.vue
elevatebart, line 1: Most of the changes here seem disconnected from the NoCode component. 
Why not make it a separate component with a slot?

elevatebart (changes requested):
Thank you @brian-mulier-p.
Some light refactoring is necessary but this is going in the right direction.

brian-mulier-p:
Better?

elevatebart (approved):
Now that's cool

PR #15211, opened 2026-03-24
fix(core): clean up AiService conversation metadata on generation failure

fhussonnois:
The metadataByConversationId ConcurrentHashMap in AiService leaks entries when AI generation throws an exception. beforeGeneration() adds an entry but afterGeneration() is only called on success, so failed requests (timeouts, API errors, parsing failures) permanently leak metadata.

Wrap generation calls in try/catch to call afterGeneration() with an error span on failure, ensuring the map entry is always removed.


brian-mulier-p (approved):
LGTM, to do after develop merge + EE has similar service + api.kestra.io has similar service. All those should also receive a similar fix.

PR #15213, opened 2026-03-24
fix(ui): resolve monaco editor autocomplete instability

### LINKED ISSUE #14989: Autocomplete doesn't always display in editor
### Describe the issue

### Environment

Sumitsh28:
### ✨ Description

This PR resolves several instability issues with the Monaco Editor's YAML autocomplete behavior, specifically addressing edge cases where suggestions disappear, hang, or fail to show deprecated aliases.

**Key changes:**
- **Fix disappearing suggestions:** Added `incomplete: true` to the main completion provider's return object. This forces Monaco to bypass local caching and re-trigger our custom filtering logic on every keystroke (like backspace), keeping the widget open and accurate.
- **Fix missing aliases/deprecated tasks:** Swapped a `.filter()` for a `.map()` that applies `monaco.languages.CompletionItemTag.Deprecated` to deprecated plugins. This ensures aliases are still suggested but visually warned with a strikethrough, rather than completely hidden from the user.
- **Fix infinite "Loading...":** Wrapped the `valueAutoCompletion` async call in a `try...catch` block to gracefully return `NO_SUGGESTIONS` if the promise rejects, preventing the editor widget from hanging indefinitely.
- **Fix silent inline completion crashes:** Wrapped the `updateDocumentation` call in the inline completions provider in a `try...catch` block. This prevents the provider from crashing and failing to show multiline snippets if a plugin schema fetch fails (e.g., network error or invalid plugin class).

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/14989

https://github.com/user-attachments/assets/6f8d3ce6-d7c3-473d-96d7-d98908fb6500

### 🎨 Frontend Checklist


elevatebart (approved):
LGTM

PR #15246, opened 2026-03-25
fix(tests): fix flaky errorProcessing() in DispatchQueue tests

fhussonnois:
Use Awaitility to wait for shutdown() to be called, fixing a race condition where close() unblocks before markEnd(Throwable) calls KestraContext.shutdown().


loicmathieu (approved):
LGTM

PR #15248, opened 2026-03-25
feat(grpc): add TLS encryption support for worker-to-controller communication

fhussonnois:
Add optional TLS and mutual TLS (mTLS) support for gRPC channels between workers and controllers. When disabled (default), plaintext communication is preserved. Configuration via `kestra.grpc.tls` properties using P12/JKS keystores and truststores with password protection.


AcevedoR (approved):
clean, tested and straighforward. LGTM

PR #15261, opened 2026-03-26
feat(core): add @PluginProperty(secret = true) with validation for it

brian-mulier-p:
closes https://github.com/kestra-io/kestra-ee/issues/6418

### DIFF
--- a/core/src/main/java/io/kestra/core/utils/SecretUtils.java
+++ b/core/src/main/java/io/kestra/core/utils/SecretUtils.java
@@ -1,0 +1,78 @@
+package io.kestra.core.utils;
+
+import io.kestra.core.models.annotations.PluginProperty;
+import io.kestra.core.models.property.Property;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Utility for validating fields annotated with {@link PluginProperty#secret()}.
+ */
+public final class SecretUtils {
+
+    private SecretUtils() {}
+
+    /**
+     * Returns a list of violation messages for any {@link PluginProperty#secret()} field
+     * on the given object whose value is a plain-text string rather than a Pebble expression.
+     *
+     * @param obj the object to inspect (task, trigger, etc.)
+     * @return violation messages, empty if all secret fields are properly expressed
+     */
+    public static List<String> validateSecretFields(Object obj) {
+        List<String> violations = new ArrayList<>();
+        if (obj == null) {
+            return violations;
+        }
+
+        Class<?> clazz = obj.getClass();
+        while (clazz != null && clazz != Object.class) {
+            for (Field field : clazz.getDeclaredFields()) {
+                PluginProperty annotation = field.getAnnotation(PluginProperty.class);
+                if (annotation == null || !annotation.secret()) {
+                    continue;
+                }
+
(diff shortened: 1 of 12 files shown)


core/src/main/java/io/kestra/core/utils/SecretUtils.java
```diff
@@ -0,0 +1,78 @@
...
+                String strValue;
+                if (fieldValue instanceof String s) {
+                    strValue = s;
+                } else if (fieldValue instanceof Property<?> p) {
+                    strValue = getPropertyExpression(p);
+                } else {
+                    continue;
+                }
+
+                if (strValue != null && !strValue.contains("{{")) {
```
loicmathieu: Pebble can be used with `{%` for multi-line Pebble

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/utils/SecretUtils.java
AcevedoR: and comments `{#`

AcevedoR: see the default syntax at: pebble-4.1.1-sources.jar!/io/pebbletemplates/pebble/lexer/Syntax.java:209

```
  public static class Builder {

    private String delimiterCommentOpen = "{#";

    private String delimiterCommentClose = "#}";

    private String delimiterExecuteOpen = "{%";

    private String delimiterExecuteClose = "%}";

    private String delimiterPrintOpen = "{{";

    private String delimiterPrintClose = "}}";

    private String delimiterInterpolationOpen = "#{";

    private String delimiterInterpolationClose = "}";

    private String whitespaceTrim = "-";

    private boolean enableNewLineTrimming = true;
```

brian-mulier-p: I'll add it but I think there is a broader rework on that as we tend to use "{{" quite often

loicmathieu: Comment we don't care but in our doc we show the two {{ and {%, the other are not for delimitting a Pebble expression itself.
I think the best would be a little helper somewhere (we have a PebbleUtils) and start using it

brian-mulier-p: Yes that's what I was planning to do

brian-mulier-p:
@loicmathieu I've created the util and used it everywhere the app was checking for pebble blocks. you may review per commit as I split this PR into two commits: create PebbleUtils + move to its usage everywhere; then this PR original fix that now uses it.

EE PR: https://github.com/kestra-io/kestra-ee/pull/7216

brian-mulier-p:
Hey @MilosPaunovic I added the handling on the no code editor if you can give it a look:
nocode_secret_property_demo

The commit worth checking for you: https://github.com/kestra-io/kestra/pull/15261/changes/06a394bae022738e899822f0074035e88b05f602

MilosPaunovic (approved):
Changes related to `UI` are good to go! ✅

Malaydewangan09:
@brian-mulier-p, could you please backport your merged commits to `1.3` if possible, please? 🙏
Thanks!

We don't want to release 100 plugins between JAR and docker image

PR #15263, opened 2026-03-26
fix(executions): add execution-level Pebble expression eval endpoint

### LINKED ISSUE #10844: Render Expression on Execution overview page allows you to render invalid expressions
### Describe the issue

At the bottom of the Execution page, if you used a webhook trigger, it will show 3 values:
- `id`
- `type`
- `variables`

Image

These can be rendered with `Render Expression` but they use the following expression to do so:
- `{{ trigger.id }}`
- `{{ trigger.type }}`
- `{{ trigger.variables }}`

Image

but these don't exist in the workflow or in Debug Outputs. The key one that's a problem for users is `{{ trigger.variables }}` which contains useful stuff such as the webhook body goes from `{{ trigger.variables.body }}` to `{{ trigger.body }}`.  @aj-emerich was trying to copy his expression from this page into his workflow but was facing errors due to it being an invalid expression.

Either these expressions need to be implemented so they work inside of workflows, or removed to avoid confusion. `{{ trigger.variables }}` is the most problematic because people already use expressions that use child variables within it without referencing `variables`.

For example:
```yaml
id: webhook
namespace: company.team

triggers:
  - id: webhook
    key: abcdefg

tasks:
  - id: hello
    message: 
      - "{{ trigger.body }}"
      - "{{ trigger.id ?? 'invalid' }}"
      - "{{ trigger.type ?? 'invalid' }}"
      - "{{ trigger.variables ?? 'invalid' }}"
      - "{{ trigger }}"
```

returns the following:

Image

### Environment

- Kestra Version: develop

brian-mulier-p:
- Add new `POST /executions/{executionId}/eval` endpoint that evaluates Pebble expressions in the execution context (trigger, inputs, outputs, variables) without requiring a taskRunId
- Rewrite `DebugPanel.vue` to call the backend instead of doing fake client-side property traversal
- Add tests for the new endpoint (valid expression, invalid expression, secret masking)

Fixes https://github.com/kestra-io/kestra-ee/issues/7186
Closes https://github.com/kestra-io/kestra/issues/10844.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -3650,6 +3650,38 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ApiAsyncEvent"
+  /api/v1/{tenant}/executions/{executionId}/eval:
+    post:
+      tags:
+      - Executions
+      summary: Evaluate a variable expression for this execution
+      operationId: evalExpression
+      parameters:
+      - name: executionId
+        in: path
+        description: The execution id
+        required: true
+        schema:
+          type: string
+      - name: tenant
+        in: path
+        required: true
+        schema:
+          type: string
+      requestBody:
+        description: The Pebble expression that should be evaluated
+        content:
+          text/plain:
+            schema:
+              type: string
+        required: true
+      responses:
+        "200":
+          description: evalExpression 200 response
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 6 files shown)


Skraye (approved):
LGTM

PR #15276, opened 2026-03-26
fix: scheduler/trigger

fhussonnois:


webserver/src/main/java/io/kestra/webserver/models/api/ApiTriggerAndState.java
loicmathieu, line 1: Could be a record

loicmathieu (approved):
LGTM

PR #15286, opened 2026-03-27
fix(triggers): emit TriggerUpdated for unchanged triggers on flow revision change

### LINKED ISSUE #15281: Scheduler uses stale flow revision for existing triggers after flow update
## Bug

When a flow is updated with changes only to tasks (not triggers), the scheduler continues executing the old flow revision. For example, changing a log message in a task has no effect — executions still use the previous revision's message.

### Steps to Reproduce

1. Create a flow with a `Schedule` trigger (`cron: "* * * * *"`) and a Log task with message "ORIGINAL MESSAGE" — save it
2. Wait for at least one execution — confirm the log shows "ORIGINAL MESSAGE"
3. Edit the flow to change the message to "UPDATED MESSAGE" — save it (triggers unchanged)
4. Wait for the next execution — the log still shows "ORIGINAL MESSAGE"
5. Check the execution metadata — it references the old flow revision, not the latest

### Expected Behavior

After saving a new flow revision, subsequent trigger executions should use the latest revision.

### Actual Behavior

The scheduler's cached flow store never refreshes for unchanged triggers, so executions continue using the stale revision indefinitely.

brian-mulier-p:
## Summary
- When a flow was updated (e.g. task changes) but its triggers remained identical, no trigger event was emitted. This left the `CachedFlowMetaStore` holding a stale flow revision, so the trigger's evaluated tasks never reflected the new flow content.
- Adds `findUnchangedTrigger()` to identify triggers present in both revisions that are identical, and emits `TriggerUpdated` for them so the cache refreshes.

Closes #15281

### DIFF
--- a/core/src/main/java/io/kestra/core/services/FlowService.java
+++ b/core/src/main/java/io/kestra/core/services/FlowService.java
@@ -221,6 +221,10 @@ public class FlowService {
             triggersDeleted.forEach(
                 trigger -> sendTriggerEvent(new TriggerDeleted(TriggerId.of(flow, trigger)))
             );
+
+            if (flow.isDeleted()) {
+                return;
+            }
         }
 
         if (previous != null && !Objects.equals(previous.getRevision(), flow.getRevision())) {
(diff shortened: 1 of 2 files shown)


fhussonnois (approved):
LGTM!

Issue #1528, opened 2023-06-14
UI: make execution output for structured objects expandable/collapsible


vivane:
### Feature description

Execution outputs for structured objects are shown as one property per line. With large data, screen becomes hard to read. For example:
| Task   | Value | Name                        | Output   |
| ------ | ----- | --------------------------- | -------- |
| task-1 |       | objects.0.key               | file.txt |
| task-1 |       | objects.0.size              | 123      |
| task-1 |       | objects.0.owner.id          | 456      |
| task-1 |       | objects.0.owner.displayName | user     |

It might be more user-friendly, if structured data will be shown expandable/collapsible:
| Task   | Value | Name                        | Output   |
| ------ | ----- | --------------------------- | -------- |
| task-1 |       | objects                     | [...] +  |

...when user expands the array...
| Task   | Value | Name                        | Output   |
| ------ | ----- | --------------------------- | -------- |
| task-1 |       | objects                     | [...] -  |
|        |       | objects.0                   | {...} +  |

...when use expands the object in the array...
| Task   | Value | Name                        | Output   |
| ------ | ----- | --------------------------- | -------- |
| task-1 |       | objects                     | [...] -  |
|        |       | objects.0                   | {...} -  |
|        |       | objects.0.key               | file.txt |
|        |       | objects.0.size              | 123      |
|        |       | objects.0.owner             | {...} -  |

...when use expands the object in the object...
| Task   | Value | Name                        | Output   |
| ------ | ----- | --------------------------- | -------- |
| task-1 |       | objects                     | [...] -  |
|        |       | objects.0                   | {...} -  |
|        |       | objects.0.key               | file.txt |
|        |       | objects.0.size              | 123      |
|        |       | objects.0.owner             | {...} -  |
|        |       | objects.0.owner.id          | 456      |
|        |       | objects.0.owner.displayName | user     |

vivane:
Also, tasks themselves might also be expandable/collapsible...

Nico-Kestra:
Hello, maybe this will save us a meeting.
Here is a first draft for the Outputs page, thought out with Ludo last Friday.
We're using the same UX as on: https://jsonhero.io/j/q01i3ytnKWRu

![Executions_Outputs](https://github.com/kestra-io/kestra/assets/125994028/85405ef3-e8ec-45d8-a9c7-27e05e8850da)
[...]

PR #15270, opened 2026-03-26
fix(core): preserve page size when clearing all filters

MilosPaunovic:
Closes https://github.com/kestra-io/kestra-ee/issues/7131.

PR #15291, opened 2026-03-27
Properly reset filters and preserve other query params

karelkryda:
### ✨ Description

Fixes the "Reset" filter button not actually resetting filters to their default state.

Before:
- Clicking reset kept user-applied filters (e.g. selecting execution state `RUNNING` and clicking reset would keep it)
- Pagination params (`size`, `page`) were wiped from the URL, but the UI still showed the old values - causing a mismatch between displayed page size and actual results
- The reset logic re-parsed the current URL to build the "reset" state, effectively resetting to whatever was already applied

After:
- Reset properly clears all user-applied filters and returns to defaults (timeRange, default visible chips)
- Pagination size is preserved, page is dropped (back to page 1)
- Renamed `resetToPreApplied` → `resetToDefaults` to reflect actual behavior

### 🔗 Related Issue

Pylon: `#1528`

### 🎨 Frontend Checklist

### 📝 Additional Notes

All testing was done manually. Key scenarios verified:
- Reset clears user-added filters and restores defaults
- Pagination size param is preserved, page is dropped on reset
- Repeated clicks produce stable results (no reordering, no stale params)
- Tested on dashboard, flows, executions and logs pages

### DIFF
--- a/ui/src/components/filter/components/KSFilter.vue
+++ b/ui/src/components/filter/components/KSFilter.vue
@@ -74,7 +74,7 @@
         addFilter,
         removeFilter,
         updateFilter,
-        resetToPreApplied,
+        resetToDefaults,
         hasPreApplied,
         getPreApplied
     } = useFilters(
@@ -140,7 +140,7 @@
         toggleOptions,
         updateChart,
         refreshData,
-        resetToPreApplied,
+        resetToDefaults,
         hasPreApplied,
         getPreApplied,
         editSavedFilter: (filter: SavedFilter) => {
--- a/ui/src/components/filter/components/MainFilter.vue
+++ b/ui/src/components/filter/components/MainFilter.vue
@@ -91,7 +91,7 @@
 <script setup lang="ts">
     import {ref, inject, nextTick, computed} from "vue";
     import {useDebounceFn} from "@vueuse/core";
-    
+
     import {FilterOutline} from "../utils/icons";
 
     import FilterChip from "./layout/FilterChip.vue";
@@ -117,8 +117,8 @@
         return filter.configuration.value.keys?.find((key: any) => key.key === appliedFilter.key) ?? null;
     };
 
-    const setChipRef = (filterId: string, el: any) => el 
-        ? chipRefs.value[filterId] = el 
+    const setChipRef = (filterId: string, el: any) => el
+        ? chipRefs.value[filterId] = el
         : delete chipRefs.value[filterId];
 
     const handleAddFilter = (newFilter: AppliedFilter) => {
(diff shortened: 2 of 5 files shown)


Piyush-r-bhaskar:
Hey @karelkryda 

Which version you are targeting with this fix ?

- If it's for Clear All button -  It is broken indeed Post `1.3`
- If it's for pagination preservation with clear all - I feel it should have been fixed with https://github.com/kestra-io/kestra/pull/15270

karelkryda:
Hey @Piyush-r-bhaskar, 
It was intended as fix for pagination vs "Clear all" button (currently experiencing on version `1.2`) - I started it before @MilosPaunovic pushed that commit.

However I noticed that the button functionality was basically broken, as you said, so I fixed both issue here. 
Now I think the button works as expected and pagination (and other non-filter related query params) are properly preserved when using it.

Piyush-r-bhaskar:
Okie,

and if I see, `Clear All` button is not broken in `1.2` **for reseting to default filter** (PFA). correct ? So, I would be able to only backport your fix back till `1.3`. because other issue related to `pagination vs clear all` was already backported in 1.2 by Milos.


https://github.com/user-attachments/assets/5257500d-4ad5-44ab-aa98-14d1ee4edaae

karelkryda:
yep, "Clear all" button in term of clearing to default filters should work flawlessly in `1.2`, in that version only preservation of page size was discovered as issue.
So yeah, feel free to use this fix for `1.3`+ only.

Also fix by Milos was ported to `1.3` only iirc, would be nice if it could be ported to `1.2` as well. 

Thanks

Piyush-r-bhaskar (approved):
Thank you so much @karelkryda . Really appreciate your contribution 🎉

karelkryda:
@Piyush-r-bhaskar, I see you removed line 541 in your commit - `delete currentQuery.page`.

May l ask you why?

Piyush-r-bhaskar:
Yeah because if you are on say page other than 1 , you have this page added in url and you click clear all - page is removed only size is preserved. - should it be intended (i don't think so) -  now it preserves both page and size

karelkryda:
That is intended, if you reset filters, you should always end up on page one cuz clearing filters changes visible results.
Staying on page let's say two after changing filters is not what should happen.

Piyush-r-bhaskar:
But isn't when we add/change filter from page other than 1, we already removing page and coming back to 1

https://github.com/user-attachments/assets/2bb62858-6c70-4ad1-bdc6-f0d81bb22ac3


But yeah I will re add to remove the page, but It won't make much difference.

karelkryda:
Default page reset works **when changing filters**, but **clicking the button** without that line keeps it on page 2. That's different code path.

PR #15292, opened 2026-03-27
fix(tests): allow test loader extensions to work with MicronautTest

fhussonnois:
AbstractLoaderExtension and TriggerEvaluationExtension hardcoded the ApplicationContext lookup to the KestraTestExtension store namespace. Add a fallback to also check MicronautJunit5Extension namespace so loader annotations work with plain MicronautTest.


nkwiatkowski (approved):
LGTM

loicmathieu (approved):
LGTM

PR #15305, opened 2026-03-30
fix(tests): move SQL injection tests in the jdbc module

loicmathieu:
As obviously those didn't make sense in Elastic


brian-mulier-p (approved):
LGTM

PR #15306, opened 2026-03-30
refactor(core): add subscriber tracking lifebelt to AbstractQueue

fhussonnois:
Track all subscribers created via subscriber() so the queue can close any that were not closed by their callers. Uses the existing emit/doEmit template-method pattern: subscriber() is now final in the abstract queue classes and delegates to doSubscriber() in concrete implementations.


loicmathieu (approved):
LGTM, I have that in my mind also so I totally agree with it.

PR #15315, opened 2026-03-31
fix(worker): exclude repository-dependent beans from workers and add gRPC maintenance endpoint

Skraye:
Use @Requires(beans = XxxRepository.class) on all beans that inject repositories Add getMaintenanceMode gRPC endpoint
Make SettingRepository optional in InstanceService to prevent
  worker startup failure
Upgrade spotless dependencies

### DIFF
--- a/core/src/main/java/io/kestra/core/services/InstanceService.java
+++ b/core/src/main/java/io/kestra/core/services/InstanceService.java
@@ -8,6 +8,8 @@ import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 import lombok.extern.slf4j.Slf4j;
 
+import java.util.Optional;
+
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/kestra/core/services/InstanceService.java
```diff
@@ -33,6 +35,12 @@ public String fetch() {
     }
 
     private Setting fetchInstanceUuid() {
+        if (settingRepository == null) {
```
fhussonnois: I think there's no problem with returning `null` here for the worker. I'm not really fan of generating a random `instance_id` that won't have any meaning.

fhussonnois (changes requested):

PR #15247, opened 2026-03-25
Feat/queue perf

loicmathieu:
- Pause/Resume when there are no consumers on follow endpoints
- Use virtual threads for async queue message emission
- Directly process a message from the execition queue avoiding one message emission/reception from the executor
- Only use two indices on the `queues` table
- Use an UPDATE query inside the execution lock method: this shows 60% reduction in query time with respect to INSERT ON CONFLICT UPDATE (3.5ms -> 1.25ms)

PR #15317, opened 2026-03-31
fix(core): cache retry policy and deadlock predicate in JooqDSLContextWrapper

fhussonnois:
The retryer() method was rebuilding a new Random retry policy and RetryUtils.Instance on every call. Since JooqDSLContextWrapper is used by every JDBC queue poll cycle (~21 subscribers polling at 25ms intervals), this caused significant transient object allocation pressure contributing to GC overhead.

Cache the retry policy as a static constant and replace the anonymous predicate lambda with a static inner class to avoid repeated allocation.

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/JooqDSLContextWrapper.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/JooqDSLContextWrapper.java
@@ -18,6 +18,15 @@ import jakarta.inject.Singleton;
 
 @Singleton
 public class JooqDSLContextWrapper {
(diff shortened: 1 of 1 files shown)


loicmathieu:
I already have this one here: https://github.com/kestra-io/kestra/pull/15247/changes/80b5bd6d9763c0b249278eff593486769b2a241a ;) 

This PR had multiple perf improvements that I planed to merge at some point but didn't have time yet: https://github.com/kestra-io/kestra/pull/15247

loicmathieu (approved):
LGTM

PR #15318, opened 2026-03-31
fix(core): select only needed columns in JDBC queue subscriber queries

fhussonnois:
Replace SELECT * with SELECT offset, value in the 4 subscriber methods of JdbcQueueClient. All methods only use these 2 fields — the other columns (type, routing_key, key, created) were fetched and deserialized by jOOQ on every poll cycle then immediately discarded.


loicmathieu (approved):
LGTM

Issue #15040, opened 2026-03-12
MCP Tool Trigger


anna-geller:
### Feature description

Goal: each flow becomes an individual MCP tool

We want to add a global MCP server per tenant/instance:
- All flows with the `McpToolTrigger` are automatically exposed as tools
- JSON schema is auto-generated from flow inputs/outputs
- 1 - 1.5 weeks development estimate

Example from @ammeek:

```yaml
id: custom_schema_mcp_tool_flow
namespace: company.team

triggers:
  - id: mcp_tool_triggger
    type: io.kestra.core.models.triggers.McpToolTrigger
    description: "Call this tool to generate a greeting for the user."
    title: "User Greeting Tool With Time Of Day"
    additionalToolArguments:
      time_of_day: |
        {
          "$schema": "http://json-schema.org/draft-07/schema#",
          "title": "TimeOfDay",
          "type": "string",
          "enum": ["morning", "afternoon"]
        }

inputs:
  - id: first_name
    type: STRING
    description: "Your first name"
  - id: last_name
    type: STRING
    description: "Your last name"


tasks:
  - id: hello
    type: io.kestra.plugin.core.log.Log
    message: "Good {{trigger.time_of_day}} {{inputs.first_name}} {{inputs.last_name}}. It's nice to meet you."


outputs:
  - id: greeting
    type: STRING
    value: "Good {{trigger.time_of_day}} {{inputs.first_name}} {{inputs.last_name}}. It's nice to meet you."
    displayName: A friendly greeting
```

bishalbera:
Can I try this ?

loicmathieu:
@bishalbera no, it's part of a bigger project around MCP integration in Kestra and someone from our internal team will have a look, sorry.

anna-geller:
@ammeek can you write a brief update on what's the status of this & ETA before Dev Weekly on Monday?

ammeek:
@anna-geller 
Current Status & Demo
The OSS UI and back-end development are going well and are mostly complete. I'm just polishing up a few minor details. We decided to hold off on starting the Enterprise Edition (EE) features just yet so we can focus on getting the OSS components completely wrapped up. I'll be setting up a full demo of this for the Product Team and Florian next week.

What's Shippable for 2.0
Since we're aiming for a release candidate in the next 2–3 weeks, Florian and I agreed to prioritize core functionality and push a few of the heavier items to 2.1.

For 2.0, we'll be shipping:

[...]

PR #15325, opened 2026-03-31
feat: validate JSON flow inputs against jsonSchema

ammeek:
### ✨ Description

This PR adds runtime validation for JSON flow inputs when jsonSchema is provided on JsonInput.

Changes included:

- Implements JsonInput.validate(...) in:
    - kestra/core/src/main/java/io/kestra/core/models/flows/input/JsonInput.java
- Uses the same JSON schema validation stack used in the MCP Java SDK (com.networknt:json-schema-
validator) to avoid custom/partial schema logic.
- Supports validation for both:
    - object/map input values
    - string input values containing JSON
- Dependency updates:
    - kestra/core/build.gradle adds com.networknt:json-schema-validator
    - kestra/platform/build.gradle pins com.networknt:json-schema-validator:2.0.0

This is implemented to support MCP worker work where structured JSON input validation is required.

### 🔗 Related Issue

Relates to https://github.com/kestra-io/kestra/issues/15040.

### 🛠️ Backend Checklist

### 📝 Additional Notes

- The validator implementation intentionally mirrors MCP Java SDK’s NetworkNT approach (SchemaRegistry and dialect-based schema validation) for consistency.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/input/JsonInput.java
+++ b/core/src/main/java/io/kestra/core/models/flows/input/JsonInput.java
@@ -1,7 +1,19 @@
 package io.kestra.core.models.flows.input;
 
-import io.kestra.core.models.flows.Input;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.networknt.schema.Error;
+import com.networknt.schema.Schema;
+import com.networknt.schema.SchemaRegistry;
+import com.networknt.schema.dialect.Dialects;
+
+import io.kestra.core.models.flows.Input;
+import io.kestra.core.models.validations.ManualConstraintViolation;
+import io.kestra.core.serializers.JacksonMapper;
 import jakarta.validation.ConstraintViolationException;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/kestra/core/models/flows/input/JsonInput.java
```diff
@@ -11,8 +22,50 @@
 @Getter
 @NoArgsConstructor
 public class JsonInput extends Input<Object> {
+    private static final ObjectMapper MAPPER = JacksonMapper.ofJson();
+    private static final SchemaRegistry SCHEMA_REGISTRY = SchemaRegistry.withDialect(Dialects.getDraft202012());
```
loicmathieu, line 29: We already have a JSON Schema library `com.github.victools.jsonschema` can't it be used instead of a new one?

platform/build.gradle
```diff
@@ -113,6 +113,7 @@ dependencies {
         api group: 'com.github.victools', name: 'jsonschema-module-jakarta-validation', version: jsonschemaVersion
         api group: 'com.github.victools', name: 'jsonschema-module-jackson', version: jsonschemaVersion
         api group: 'com.github.victools', name: 'jsonschema-module-swagger-2', version: jsonschemaVersion
+        api group: 'com.networknt', name: 'json-schema-validator', version: '2.0.0'
```
loicmathieu: Version 3.0.1 exist why do you use an old one?

core/src/main/java/io/kestra/core/models/flows/input/JsonInput.java
ammeek: Unfortunately the com.github.victools.jsonschema lib only support generation of json schemas it does not support validating them.

platform/build.gradle
ammeek: I thought that 2.0.0 was the version that the MCP client was using which I thought would be the latest. I've just doubled check and the mcp client repo uses 3.0.0. I'll update this to use 3.0.1. Thanks for catching that.

core/src/test/java/io/kestra/core/models/flows/input/JsonInputSchemaFlowIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+import io.kestra.core.runners.FlowInputOutput;
+import io.kestra.core.runners.TestRunnerUtils;
+
+import jakarta.inject.Inject;
+
+import static io.kestra.core.tenant.TenantService.MAIN_TENANT;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@KestraTest(startRunner = true)
```
fhussonnois, line 23: I wonder if these tests are really relevant. In Kestra we used to always create integration test for validating flows. But, here the `JsonInput` validation rules are already tested with pure unit-tests `JsonInputTest`. 

runner-based tests are costly and I've the sentiment that we just test that ValidationException are properly handled but we already have tests for that. 

WDYT @loicmathieu ?

fhussonnois (approved):

PR #15328, opened 2026-03-31
feat(flow, execution): `Loop` task with transparent sub-executions

loicmathieu:
Closes https://github.com/kestra-io/kestra-ee/issues/7112*  

## Description

Loop iterations are now run in separate sub-executions, called loop executions.
These loup execution has an execution kind of type LOOP and a `LoopRun` attribute which contains the context of the loop including its value and iteration, and all of it's parents values and iterations.

Each loop iteration then executes on an isolated context.

A concurrencyLimit allow to start loop executions concurrently.

The loop context is available to expressions in the `item` variable.

## Examples:

A basic loop with a single task will creates 3 loop executions:

```yaml
id: loop-serial
namespace: io.kestra.tests

tasks:
  - id: loop
    values: ["value 1", "value 2", "value 3"]
    tasks:
      - id: item
        type: io.kestra.core.tasks.debugs.Return
        format: "{{item.index}} - {{item.value}}"
```

Multiple loops in the same execution:

```yaml
id: loop-multiple
namespace: io.kestra.tests

tasks:
  - id: loop1
    values: ["value 1", "value 2", "value 3"]
    tasks:
      - id: return1
        format: "{{item.index}} - {{ item.value }}"
  - id: loop2
    values: ["value 1", "value 2", "value 3"]
    tasks:
      - id: return2
        format: "{{item.index}} - {{ item.value }}"
```

Flowables can be used inside a loop, they could directly access item information (no need for parent.item.value) as the item is boutond to the loop execution not the loop direct subtasks

```yaml
id: loop-flowable
namespace: io.kestra.tests

tasks:
  - id: loop
    values: ["value 1", "value 2", "value 3"]
    tasks:
      - id: before_if
        format: "Before if {{ item.value }}"
      - id: if
        condition: '{{ item.value == "value 2" }}'
        then:
          - id: after_if
            format: "After if {{ item.value }}"
```

Unlimited nested loops are possible, you can use item.parent to access the direct parent value and index, and `item.parent[idx]`  to access any parents in the hierarchy.

```yaml
id: loop-nested
namespace: io.kestra.tests

tasks:
  - id: loop1
    values: [1, 2, 3]
    tasks:
      - id: loop2
        values: [1, 2, 3]
        tasks:
          - id: loop3
            values: [ 1, 2, 3 ]
            tasks:
            - id: item
              type: io.kestra.core.tasks.debugs.Return
              format: "{{item.index}} - {{item.value}}"
            - id: parent
              type: io.kestra.core.tasks.debugs.Return
              format: "{{item.parent.index}} - {{item.parent.value}}"
            - id: parents
              type: io.kestra.core.tasks.debugs.Return
              format: "{{item.parents[0].index}} - {{item.parents[0].value}} - {{item.parents[1].index}} - {{item.parents[1].value}}"
```

`transmitFailed` can be used to control wether or not the loop should fail if one of its sub-tasks fail:

```yaml
id: loop-failed-no-transmit
namespace: io.kestra.tests

tasks:
  - id: loop
    values: ["value 1", "value 2", "value 3"]
    transmitFailed: false
    tasks:
      - id: fail
```

## Remaining work:

- [x] Nested loops
- [ ] A loop has multiple iterations so maybe rename item -> iteration
- [ ] Kill
- [ ] Support Map as values
- [ ] Support internal storage as values
- [ ] Support loop iteration input
- [ ] Support loop iteration output (as loop execution output?) and provide a way to merge them post loop
- [ ] Should we support a wait to trigger and forget loop iterations?
- [ ] UI
- [ ] ...

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/flow/Loop.java
+++ b/core/src/main/java/io/kestra/plugin/core/flow/Loop.java
@@ -1,0 +1,347 @@
+package io.kestra.plugin.core.flow;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.kestra.core.exceptions.IllegalVariableEvaluationException;
+import io.kestra.core.models.annotations.Example;
+import io.kestra.core.models.annotations.Plugin;
+import io.kestra.core.models.annotations.PluginProperty;
+import io.kestra.core.models.executions.Execution;
+import io.kestra.core.models.executions.ExecutionKind;
+import io.kestra.core.models.executions.NextTaskRun;
+import io.kestra.core.models.executions.TaskRun;
+import io.kestra.core.models.flows.State;
+import io.kestra.core.models.hierarchies.GraphCluster;
+import io.kestra.core.models.hierarchies.RelationType;
+import io.kestra.core.models.tasks.FlowableTask;
+import io.kestra.core.models.tasks.ResolvedTask;
+import io.kestra.core.models.tasks.Task;
+import io.kestra.core.runners.FlowableUtils;
+import io.kestra.core.runners.RunContext;
+import io.kestra.core.utils.GraphUtils;
+import io.kestra.core.utils.MapUtils;
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotEmpty;
+import jakarta.validation.constraints.NotNull;
+import jakarta.validation.constraints.PositiveOrZero;
+import lombok.*;
+import lombok.experimental.SuperBuilder;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+@SuperBuilder
+@ToString
+@EqualsAndHashCode
+@Getter
+@NoArgsConstructor
+@Schema(
+    title = "Execute child tasks for each value in a list.",
+    description = """
+        Renders `values` (JSON array, YAML list, or expression) and runs the child task group once per item. The current item is available as `item.value`; `item.index` exposes the index.
+
+        Control parallelism with `concurrencyLimit` (0 = unlimited, 1 = fully serialized, N = up to N concurrent task groups). To run tasks inside each group in parallel, wrap them in a `Parallel` task."""
+)
+@Plugin(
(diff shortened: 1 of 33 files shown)


core/src/main/java/io/kestra/plugin/core/flow/Loop.java
```diff
@@ -0,0 +1,347 @@
...
+    private final Integer concurrencyLimit = 1;
+
+    @Builder.Default
+    @Schema(
+        title = "Flag specifying whether to fail the current task if any loop iteration fails or is killed."
+    )
+    @PluginProperty
+    private final Boolean transmitFailed = true;
+
+    // FIXME there are a lot of duplication with Sequential but as it needs to return a different output it cannot extend it
```
AcevedoR, line 246: One FIXME and other TODO were added, should we create issues for those ? none of them need immediate fixing ?

core/src/test/resources/flows/valids/loop-nested.yaml
```diff
@@ -0,0 +1,25 @@
...
+    type: io.kestra.plugin.core.flow.Loop
+    values: [1, 2, 3]
+    tasks:
+      - id: loop2
+        type: io.kestra.plugin.core.flow.Loop
+        values: [1, 2, 3]
+        tasks:
+          - id: loop3
+            type: io.kestra.plugin.core.flow.Loop
+            values: [ 1, 2, 3 ]
```
AcevedoR, line 15: it would be easier to debug/understand with different values for each loop (ex: 1,2,3. a,b,c.  y,x,z)

AcevedoR (approved):
LGTM, but can you double check all TODOs?

core/src/main/java/io/kestra/plugin/core/flow/Loop.java
loicmathieu: No, as we're evolving the feature that will be delivered in multiple PR, I'll handle those in subsequent PRs.

core/src/test/resources/flows/valids/loop-nested.yaml
loicmathieu: Maybe.

PR #15331, opened 2026-03-31
feat(variables): return pebble function args with defaults for autocompletion

### LINKED ISSUE #8350: Dynamic Pebble function fetching
### Feature description

Follow up to https://github.com/kestra-io/kestra/pull/8245, we should provide a backend endpoint that will expose all our Pebble functions (maybe filters, ...) with metadata about their args, type and default value so that the autocompletion is not hardcoding them.

### Context

Currently, autocompletion for Pebble functions is **manually** handled [in flowAutoCompletionProvider.ts](https://github.com/kestra-io/kestra/blob/develop/ui/src/override/services/flowAutoCompletionProvider.ts#L21) + [flowAutoCompletionProvider.ts](https://github.com/kestra-io/kestra/blob/develop/ui/src/override/services/flowAutoCompletionProvider.ts#L218).

We would like to change that manual (semi-hardcoded) process to be handled by the backend to avoid situations like we currently have, where newly added functions like `"{{ tasksWithState('FAILED') }}"` are not autocompleted:

Image

Image

brian-mulier-p:
closes https://github.com/kestra-io/kestra/issues/8350

### DIFF
--- a/ui/src/services/autoCompletionProvider.ts
+++ b/ui/src/services/autoCompletionProvider.ts
@@ -4,15 +4,25 @@ import {useAxios} from "../utils/axios";
 
 export const QUOTE = "'";
 
+export interface FunctionArgument {
+    name: string;
+    defaultValue: string | null;
+}
+
+export interface PebbleFunctionDef {
+    name: string;
+    arguments: FunctionArgument[];
+}
+
 let cachedFilters: string[] | null = null;
-let cachedFunctions: string[] | null = null;
+let cachedFunctions: PebbleFunctionDef[] | null = null;
 
 export function resetExpressionCache() {
     cachedFilters = null;
     cachedFunctions = null;
 }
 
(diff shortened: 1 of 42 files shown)


ui/src/services/autoCompletionProvider.ts
```diff
@@ -29,18 +39,34 @@ async function fetchExpressionFilters(): Promise<string[]> {
...
         }
     }
     return cachedFunctions;
 }
 
+/**
+ * Converts a PebbleFunctionDef into a Monaco snippet string using named-argument syntax.
+ * Arguments with null defaults are omitted.
+ */
+export function functionToSnippet(fn: PebbleFunctionDef): string {
```
AcevedoR, line 58: maybe you can quickly add some unit tests on this with real world data. Seems a quickwin :)

AcevedoR (approved):
LGTM besised that a quick unit test could be added easily

PR #15351, opened 2026-04-01
feat(core): regexp filter

### LINKED ISSUE #15332: [Expressions] Add regex functionality
### Feature description

Add a suite of new pebble expressions which allow regex operations. As an example we could consider the following:
- `regexMatch()` - returns a simple true/false if the provided string matches the given regex
- `regexReplace()` - replace in a string where it matches the given regex
- `regexExtract()` - extract the portion of a given string if it matches the regex

These could open up simpler ways to build pipelines - e.g. extract an ID directly from an input with an expression rather than running an additional task to do so first

tchiotludo:
Closes https://github.com/kestra-io/kestra/issues/15332

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/pebble/filters/RegexExtractFilter.java
+++ b/core/src/main/java/io/kestra/core/runners/pebble/filters/RegexExtractFilter.java
@@ -1,0 +1,67 @@
+package io.kestra.core.runners.pebble.filters;
+
+import io.pebbletemplates.pebble.error.PebbleException;
+import io.pebbletemplates.pebble.extension.Filter;
+import io.pebbletemplates.pebble.template.EvaluationContext;
+import io.pebbletemplates.pebble.template.PebbleTemplate;
+
+import java.text.MessageFormat;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Pebble filter that extracts the first portion of a string matching a regular expression.
+ *
+ * <p>Usage: {@code {{ "order-12345-done" | regexExtract(regex="\d+") }}}</p>
+ *
+ * <p>An optional {@code group} argument (default {@code 0}) selects which capture group to return.
+ * Group {@code 0} returns the entire match; groups {@code 1..n} return specific capture groups.</p>
+ *
+ * <p>Returns {@code null} if no match is found.</p>
+ *
+ * @see RegexMatchFilter
+ * @see RegexReplaceFilter
+ */
+public class RegexExtractFilter implements Filter {
+
+    public static final String FILTER_NAME = "regexExtract";
+
+    private static final String ARGUMENT_REGEX = "regex";
+    private static final String ARGUMENT_GROUP = "group";
+
+    private static final List<String> ARGS = List.of(ARGUMENT_REGEX, ARGUMENT_GROUP);
+
+    @Override
+    public List<String> getArgumentNames() {
+        return ARGS;
+    }
+
+    @Override
+    public Object apply(Object input, Map<String, Object> args, PebbleTemplate self, EvaluationContext context, int lineNumber) throws PebbleException {
+        if (input == null) {
+            return null;
+        }
+
+        if (args.get(ARGUMENT_REGEX) == null) {
+            throw new PebbleException(
+                null,
+                MessageFormat.format("The argument ''{0}'' is required.", ARGUMENT_REGEX),
+                lineNumber,
+                self.getName()
+            );
+        }
+
+        String regex = args.get(ARGUMENT_REGEX).toString();
+        int group = args.containsKey(ARGUMENT_GROUP) && args.get(ARGUMENT_GROUP) != null
+            ? ((Number) args.get(ARGUMENT_GROUP)).intValue()
+            : 0;
+
+        Matcher matcher = Pattern.compile(regex).matcher(input.toString());
+        if (matcher.find()) {
+            return matcher.group(group);
+        }
+        return null;
+    }
+}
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/kestra/core/runners/pebble/filters/RegexMatchFilter.java
```diff
@@ -0,0 +1,52 @@
...
+            throw new PebbleException(
+                null,
+                MessageFormat.format("The argument ''{0}'' is required.", ARGUMENT_REGEX),
+                lineNumber,
+                self.getName()
+            );
+        }
+
+        String regex = args.get(ARGUMENT_REGEX).toString();
+        return Pattern.compile(regex).matcher(input.toString()).find();
```
brian-mulier-p: Idk if it's wanted but the regex doesn't have to fully match the string with the current implementation (the regex must have a "^" at beginning and "$" at the end if we want to do so, or alternatively use .matches() instead)

core/src/test/java/io/kestra/core/runners/pebble/filters/RegexFilterTest.java
```diff
@@ -0,0 +1,110 @@
...
+    void regexExtractFullMatch() throws IllegalVariableEvaluationException {
+        String result = variableRenderer.render("{{ 'order-12345-done' | regexExtract(regex='\\d+') }}", Map.of());
+        assertThat(result).isEqualTo("12345");
+    }
+
+    @Test
+    void regexExtractCaptureGroup() throws IllegalVariableEvaluationException {
+        String result = variableRenderer.render("{{ 'order-12345-done' | regexExtract(regex='order-(\\d+)', group=1) }}", Map.of());
+        assertThat(result).isEqualTo("12345");
+    }
```
brian-mulier-p, line 92: Better to test with 3 numbers and use group 2 to assert it takes the right group number

brian-mulier-p (approved):
Small changes but overall LGTM

core/src/main/java/io/kestra/core/runners/pebble/filters/RegexMatchFilter.java
brian-mulier-p: Please move Pattern.compile(regex) to a static constant

core/src/main/java/io/kestra/core/runners/pebble/filters/RegexReplaceFilter.java
```diff
@@ -0,0 +1,65 @@
...
+                null,
+                MessageFormat.format("The argument ''{0}'' is required.", ARGUMENT_REPLACEMENT),
+                lineNumber,
+                self.getName()
+            );
+        }
+
+        String regex = args.get(ARGUMENT_REGEX).toString();
+        String replacement = args.get(ARGUMENT_REPLACEMENT).toString();
+        return Pattern.compile(regex).matcher(input.toString()).replaceAll(replacement);
```
brian-mulier-p: Please move Pattern.compile(regex) to a static constant

core/src/main/java/io/kestra/core/runners/pebble/filters/RegexExtractFilter.java
```diff
@@ -0,0 +1,67 @@
...
+                self.getName()
+            );
+        }
+
+        String regex = args.get(ARGUMENT_REGEX).toString();
+        int group = args.containsKey(ARGUMENT_GROUP) && args.get(ARGUMENT_GROUP) != null
+            ? ((Number) args.get(ARGUMENT_GROUP)).intValue()
+            : 0;
+
+        Matcher matcher = Pattern.compile(regex).matcher(input.toString());
```
brian-mulier-p: Please move Pattern.compile(regex) to a static constant

core/src/main/java/io/kestra/core/runners/pebble/filters/RegexExtractFilter.java
```diff
@@ -0,0 +1,67 @@
...
+        if (matcher.find()) {
```
brian-mulier-p, line 77: Either try catch throwing a PebbleException upon catching an IndexOutOfBoundsException

OR add `&& group <= matcher.groupCount` to return null if there is not enough caught groups (maybe worth adding a test)

brian-mulier-p (changes requested):

core/src/main/java/io/kestra/core/runners/pebble/filters/RegexReplaceFilter.java
tchiotludo: didn't get it, since the regexp is not static, seems not possible

core/src/main/java/io/kestra/core/runners/pebble/filters/RegexMatchFilter.java
tchiotludo: same here, regexp is not static ^^

tchiotludo: It the easiest implementation and expected for end users, people that want a strict version is used to manage the `^` `$`, but lazy one will have a win quickly

core/src/main/java/io/kestra/core/runners/pebble/filters/RegexExtractFilter.java
tchiotludo: again ^^

core/src/main/java/io/kestra/core/runners/pebble/filters/RegexExtractFilter.java
tchiotludo: good catch, added a proper message

core/src/main/java/io/kestra/core/runners/pebble/filters/RegexReplaceFilter.java
brian-mulier-p: I'm dumb forget about that :clown_face:

brian-mulier-p (approved):
LGTM

PR #15357, opened 2026-04-01
fix(core): unwrap Optional types in JSON schema to avoid null type

### LINKED ISSUE #2929: Optional in json schema are shown with both "null" and the actual type under the Optional
### Describe the issue

tbh idk if it's worth anytime but at least it's tracked feel free to close if you're ok to live with that, I am
Image

If any change has to be made, basically the `type` property of the json schema is returned as an array as ["null", "string"] for eg. then it's parsed and displayed as-is in ui-libs.

Basically it's just underlining the fact that it's not a required property but as it's in definition it doesn't make much sense so... idk

brian-mulier-p:
Closes https://github.com/kestra-io/docs/issues/2929


nkwiatkowski (approved):
LGTM

PR #15359, opened 2026-04-01
fix(replay): show inputs if applicable

### LINKED ISSUE #15293: Replay modal shows options for inputs even when flow doesn't have any
### Describe the issue

When you go to replay an execution (both from the execution level or task run level), you are presented with the option to reuse original inputs or modify inputs. However there's a number of scenarios where these radio buttons are confusing as they are unused.

Image

Scenario | Expected | Actual
-- | -- | --
Flow replay on a revision with inputs (execution had inputs) | Show "Reuse original inputs" / "Modify inputs" radio buttons | Inputs section always shows regardless of whether the flow has inputs
Flow replay on a revision without inputs | Hide the Inputs section entirely | Inputs section still shows with reuse/modify options (misleading)
Flow replay — user switches revision from one without inputs to one with inputs | Prompt user to provide inputs since there are none to reuse | No handling — shows reuse option even though there are no original inputs to reuse
Task run replay on a revision without inputs | Hide the Inputs section | Inputs section still shows with reuse/modify options
Task run replay on a revision with inputs | Show reuse/modify options (task inherits inputs from original execution) | Same as flow replay — no distinction between the two

The tricky bit is dealing with a flow revision which now has inputs but the previous revision didn't. I guess this would fail regardless as there's been too many changes and you just need to create a new execution at this point. cc @kestra-io/products for some suggestions how to handle this scenario.

Here's a mock up I made playing with Claude Code where it can hide the inputs section when it's not applicable.

Image

This is also a good thing for the docs/education around when Replay isn't suitable because too many changes (to expressions or tasks cause replaying from task runs halfway likely to fail anyway). cc @aj-emerich / @akeller 

### Environment

- Kestra Version: `1.3+`

wrussell1999:
Fixes https://github.com/kestra-io/kestra/issues/15293

It does also rename the Restart modal which is touched on in more detail in this PR: https://github.com/kestra-io/kestra/pull/15196

https://github.com/user-attachments/assets/dd53b210-30c8-488a-a199-c8a51427a3e9


MilosPaunovic:
Hey @wrussell1999, so this PR will suppress https://github.com/kestra-io/kestra/pull/15196 and also close https://github.com/kestra-io/kestra-ee/issues/7130?

MilosPaunovic (approved):
Does as advertised! ✅

PR #15362, opened 2026-04-01
fix(core): cache compiled JsonQuery expressions in H2Functions

fhussonnois:
JsonQuery.compile() was called on every invocation of the H2 JQ SQL functions (JQ_STRING, JQ_BOOLEAN, etc.), recompiling the same jq expression for each row processed. JFR profiling showed this caused 3.3 GB of allocations (58% of queue-h2 thread allocation) due to JavaCC's SimpleCharStream allocating large int[]/char[] buffers on each parse.

Cache compiled JsonQuery instances in a ConcurrentHashMap keyed by expression string. JsonQuery is immutable (holds only a final Expression field) and thread-safe to share.

### DIFF
--- a/jdbc-h2/src/main/java/io/kestra/runner/h2/H2Functions.java
+++ b/jdbc-h2/src/main/java/io/kestra/runner/h2/H2Functions.java
@@ -2,6 +2,7 @@ package io.kestra.runner.h2;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


loicmathieu (approved):
Ah ah, saw this one in an IntelliJ profile but didn't fix it yet as H2 is not perfomant oriented.
By the way 3GB allocation is not that an issue today for a run as long as our CI.

Anyway, still something worth it.
I had some concurrency concern but I think that each H2 query uses a single lock so it should be OK.

PR #15363, opened 2026-04-01
chore(build): remove Allure and AspectJ weaver from test configuration

AcevedoR:
## Summary

- Removes `aspectjweaver` javaagent from test JVM args — it starts a JFR Event Stream background thread that causes `OutOfMemoryError: Java heap space` in CI
- Removes `allure-junit5` test dependency and `allure-bom` platform dependency

## Test plan

- [ ] CI passes without OOM errors in JFR Event Stream thread


loicmathieu (approved):
LGTM

PR #15366, opened 2026-04-02
feat(core): add failFast property to Parallel task

### LINKED ISSUE #15365: Not Urgent | Suggestion - a property for core.Parallel Task that Kills all other sub tasks when one task fails
### Feature description

Hello,

Right now it is not intended behavior for a `io.kestra.plugin.core.flow.Parallel` to KILL other sub tasks that are running within it when one fails.  

I suggest adding property to have this kill sub tasks when one fails.

This is not urgent, just suggestion, it is solveable with the following YAML:

```yaml
      - id: parallel
        type: io.kestra.plugin.core.flow.Parallel
        concurrent: 5
        errors:
          - id: kill_fast
            type: io.kestra.plugin.kestra.executions.Kill
            propagateKill: true
            executionId: "{{ execution.id }}"
            auth:
              apiToken: "{{ secret('KESTRA_API_TOKEN') }}"
          - id: fail_main
        tasks:
...
```

The errors sub tasks I added essentially does what this property would do, and solves it, just for future users the property might be helpful.

Curious what you think?

anna-geller:
## Changes

Adds a `childFailurePolicy` enum property to all parallel flowable tasks (`Parallel`, `Dag`, `ForEach`) that controls behavior when a child task fails. Defaults to `FAIL_FAST`.

Closes https://github.com/kestra-io/kestra/issues/15365

### Enum values

- **`FAIL_FAST`** (default): kill running siblings immediately, don't start new ones
- **`STOP`**: let running siblings finish, don't start new ones (previous default behavior)
- **`CONTINUE`**: keep starting and running all siblings, fail only after all complete

### What changed

- **BREAKING CHANGE:** Default behavior is now `FAIL_FAST`. Set `childFailurePolicy: STOP` to restore the previous behavior.
- Supported on `Parallel`, `Dag`, and `ForEach` (when `concurrencyLimit != 1`)
- Works with `allowFailure`, `allowWarning`, and retry policies
- Error and finally handlers still execute after the policy is applied

### Implementation Details

- New `ChildFailurePolicy` enum (`FAIL_FAST`, `STOP`, `CONTINUE`) with `@JsonCreator` for case-insensitive parsing
- `getChildFailurePolicy()` default method on `FlowableTask` interface — generic for any flowable task
- `FlowableUtils.resolveParallelNexts()`, `resolveDagNexts()`, and `resolveConcurrentNexts()` accept the enum directly
- `CONTINUE` mode bypasses `findTaskDependingFlowState()` to keep scheduling normal tasks despite failures
- `handleFailFastKilling()` in `ExecutorService` kills siblings only for `FAIL_FAST` policy
- KILLED→FAILED state conversion in the executor distinguishes failFast kills from manual execution kills

### Tests

- `parallel-failfast.yaml` — FAIL_FAST behavior (siblings killed)
- `parallel-failfast-with-errors.yaml` — FAIL_FAST with error/finally handlers
- `parallel-failfast-allow-failure.yaml` — FAIL_FAST with allowFailure (should NOT trigger)
- `parallel-failfast-disabled.yaml` — STOP mode (siblings complete normally)
- `parallel-continue.yaml` — CONTINUE mode (all siblings run despite failure)
- `dag-failfast.yaml` — FAIL_FAST for DAG tasks
- `foreach-failfast.yaml` — FAIL_FAST for ForEach with concurrent execution


core/src/main/java/io/kestra/core/runners/FlowableUtils.java
loicmathieu: I wonder if this one is still used

loicmathieu: Same here, it should not be used anymore

core/src/main/java/io/kestra/plugin/core/flow/Dag.java
```diff
@@ -191,6 +201,21 @@ public Optional<State.Type> resolveState(RunContext runContext, Execution execut
             this.isAllowFailure(),
             this.isAllowWarning()
         );
+
+        // When failFast killed siblings due to a child failure, guessFinalState returns KILLED
+        // because it prioritizes KILLED over FAILED. Override to FAILED since the root cause
+        // is a task failure, not a manual kill.
+        if (state.isPresent() && state.get() == State.Type.KILLED) {
```
loicmathieu: I wonder if it would not make a real KILL ends in failed when you enable failFast.

core/src/main/java/io/kestra/plugin/core/flow/ForEach.java
```diff
@@ -245,6 +255,21 @@ public Optional<State.Type> resolveState(RunContext runContext, Execution execut
             this.isAllowFailure(),
             this.isAllowWarning()
         );
+
+        // When failFast killed siblings due to a child failure, guessFinalState returns KILLED
+        // because it prioritizes KILLED over FAILED. Override to FAILED since the root cause
+        // is a task failure, not a manual kill.
+        if (state.isPresent() && state.get() == State.Type.KILLED) {
```
loicmathieu: Same here

core/src/main/java/io/kestra/plugin/core/flow/Parallel.java
```diff
@@ -198,5 +208,20 @@ public Optional<State.Type> resolveState(RunContext runContext, Execution execut
             this.isAllowFailure(),
             this.isAllowWarning()
         );
+
+        // When failFast killed siblings due to a child failure, guessFinalState returns KILLED
+        // because it prioritizes KILLED over FAILED. Override to FAILED since the root cause
+        // is a task failure, not a manual kill.
+        if (state.isPresent() && state.get() == State.Type.KILLED) {
```
loicmathieu: same here

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -761,6 +766,71 @@ private ExecutorContext handleCreatedKilling(ExecutorContext executor) throws In
         return executor;
     }
 
+    /**
+     * When a parallel flowable task (Parallel, Dag, ForEach) has failFast enabled and a child has
+     * failed (with no pending retries), mark all still-running sibling task runs as KILLED so the
+     * block can transition to error/finally handling.
+     */
+    private ExecutorContext handleFailFastKilling(ExecutorContext executor) {
```
loicmathieu: I'm not convinced this is the best place to do that, and the code seems overly complicated.
I wonder if this is not something that should be bring more broadly to all flowable, or at least all Flowable that executes subtasks.

I think there might be a better design here.

loicmathieu (changes requested):
Overall, even if it seems to work, I wonder if this is the right way to do it.
Moreover, I think there is a bug when you want to kill a ForEach that has failFast enabled.

There are also too much comment without real added value which is not something we want in our code base.

executor/src/main/java/io/kestra/executor/ExecutorService.java
anna-geller: Thanks a lot for the review. I'll address your feedback when I'm back after Easter

anna-geller:
Review feedback should now be addressed. Can you check again? Thanks a lot!

anna-geller:
@loicmathieu Could you review again, please? I included Ludo's request to make this an Enum https://kestra-io.slack.com/archives/C052XEFFC90/p1775217276344109?thread_ts=1775217034.551139&cid=C052XEFFC90

core/src/main/java/io/kestra/core/models/tasks/FlowableTask.java
```diff
@@ -70,6 +71,14 @@ public interface FlowableTask<T extends Output> {
      */
     boolean isAllowWarning();
 
+    /**
+     * Policy for handling child task failures.
+     * Defaults to {@code FAIL_FAST}; sequential flowable tasks may override to {@code STOP}.
+     */
+    default Property<ChildFailurePolicy> getChildFailurePolicy() {
+        return Property.ofValue(ChildFailurePolicy.FAIL_FAST);
```
loicmathieu, line 79: This is a BC as the current mechanism is STOP.

core/src/main/java/io/kestra/core/models/tasks/ChildFailurePolicy.java
```diff
@@ -0,0 +1,27 @@
...
+@Schema(
+    enumAsRef = true,
+    title = "Policy for handling child task failures in parallel flowable tasks."
+)
+public enum ChildFailurePolicy {
+    @Schema(title = "Kill all running siblings immediately on first failure.")
+    FAIL_FAST,
+
+    @Schema(title = "Stop starting new siblings but let running ones finish.")
+    STOP,
```
loicmathieu: We usually use FAIL / KILL and never STOP, this would bring a new term / concept

core/src/main/java/io/kestra/plugin/core/flow/ForEach.java
```diff
@@ -208,6 +210,18 @@ public class ForEach extends Sequential implements FlowableTask<VoidOutput> {
...
 
+    @Builder.Default
+    @Schema(
+        title = "Policy for handling child task failures when concurrencyLimit allows parallel execution.",
+        description = """
+            - `FAIL_FAST`: kill running task groups immediately.
+            - `STOP`: let running groups finish but don't start new ones.
+            - `CONTINUE`: keep starting and running all groups, fail only after all complete.
+
+            Has no effect when `concurrencyLimit` is 1 (serial execution)."""
```
loicmathieu, line 221: We usually fail validation for incorrectly configured tasks, this may be a good idea to do that here also

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -1256,6 +1327,15 @@ public void addWorkerTaskResult(ExecutorContext executor, Supplier<FlowWithSourc
...
 
         TaskRun taskRun = workerTaskResult.getTaskRun();
+        try {
+            TaskRun existing = executor.getExecution().findTaskRunByTaskRunId(taskRun.getId());
+            if (existing.getState().getCurrent() == State.Type.KILLED && taskRun.getState().getCurrent() != State.Type.KILLED) {
+                return;
+            }
+        } catch (InternalException ignored) {
+            // ignored
+        }
```
loicmathieu: I don't understand why you have to do that here.

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -806,6 +829,54 @@ private ExecutorContext handleCreatedKilling(ExecutorContext executor) throws In
         return executor;
     }
 
+    private ExecutorContext handleFailFastKilling(ExecutorContext executor) {
```
loicmathieu: This should be done while processing flowable tasks and not as a new step in the execution processing to avoid creating two times a run context which is a costly operation

loicmathieu (changes requested):
FAIL_FAST is supposed to be killing the execution but it is not what it did, it just set tasks to KILLED but they may still be running.
I think a real KILL execution should be performed.

I also wonder if any change should be done to the executor at all, the flowable task may just return a state of KILLING and if I remember correctly the executor would detect it and kills the execution

core/src/main/java/io/kestra/core/models/tasks/FlowableTask.java
anna-geller: Correct. That's why the PR has a breaking changes label.

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -262,6 +267,25 @@ private Optional<WorkerTaskResult> childWorkerTaskResult(FlowWithSource flow, Ex
                 runContext.logger().error("Unable to resolve state from the Flowable task: {}", e.getMessage(), e);
                 state = Optional.of(State.Type.FAILED);
             }
+            // When FAIL_FAST killed siblings, guessFinalState returns KILLED.
+            // Convert to FAILED when the root cause is a child failure, not a manual execution kill.
```
loicmathieu, line 272: Are you sure it's a good idea?
Because FAIL_FAST would effictively kill child tasks so for me it's correct for the task to ends in KILLED.

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -521,6 +542,37 @@ private ExecutorContext handleFlowableTasks(ExecutorContext executor) throws Int
...
+                        continue;
+                    }
+                    List<ResolvedTask> childTasks = flowableParent.childTasks(runContext, parentTaskRun);
+                    if (!execution.hasFailedNoRetry(childTasks, parentTaskRun)) {
+                        continue;
+                    }
+                    List<TaskRun> children = execution.findTaskRunByTasks(childTasks, parentTaskRun);
+                    for (TaskRun child : children) {
+                        if (child.getState().isRunning() && !child.getState().isFailed()) {
+                            execution = execution.withTaskRun(child.withState(State.Type.KILLED));
```
loicmathieu, line 567: This is not how to KILL a task!
In fact the only way to KILL a task, so to interrupt running task which seems to be what you want to do is to KILL an execution. And as far as I remember, if you return KILLED (or KILLING) from a flowable `resolveState()` method this is what it will do.

Overall, I think this goes in the wrong direction, it messes with the taskrun state out of our state machine.

loicmathieu (changes requested):
I also notice that a lot of helpful comments has been removed :( can you add them back?

anna-geller:
@loicmathieu, I see that this has been abandoned even though it seemed like a fairly simple enhancement that will unblock Jack's use case. Can you pick it up and backport it to 1.3?

anna-geller:
this PR is abandoned so I'm closing it - @loicmathieu perhaps you can plan it for 2.1?

PR #15380, opened 2026-04-03
feat(worker): add gRPC server events reporting relay and worker configs propagation

fhussonnois:
Workers now receive configuration from the controller during the connect handshake via a generic worker_configs map in the proto. This includes server events reporting enablement, gated by WorkerReportableScheduler.

- Add WorkerConfigsProvider interface and DefaultWorkerConfigsProvider
- Add worker_configs map to ConnectResponse proto
- Add WorkerReportableScheduler for controller-gated reporting on workers
- Add sendRaw/requestRaw to ServerEventSender for pre-serialized relay
- Exclude repository-dependent Reportables from WORKER server type


loicmathieu (approved):
LGTM

PR #15381, opened 2026-04-03
feat(system): expose the trace context in the run variables

loicmathieu:
Closes https://github.com/kestra-io/kestra-ee/issues/6826

I also take the opportunity to remove an unnecessary clone of the variable map inside the `DefaultRunContext` as the `RunVariable` already return an unmodifiable map.


ammeek (approved):
LGTM

PR #15389, opened 2026-04-06
fix(docker): add fail-fast docker socket validation, improve error detection, add tests, and skip integration tests when daemon unavailable

### LINKED ISSUE #15161: Unclear error message on missing docker socket
### Context

I was following the quickstart as mentioned https://kestra.io/docs/quickstart#start-kestra.
I however changed the command to exclude the docker socket mount and the /tmp as I don't like to share that with a product I'm exploring. 

`docker run --pull=always -it -p 8080:8080 \
  --name kestra --restart=always \
  -v kestra_data:/app/storage \
  kestra/kestra:latest server local`

Trying to run a task, this failed with a standard Java error which isn't very descriptive

Image 

Only when I opened the DEBUG logging, I got a better indication. 

Image

### Expectation

I expect a better error message than just the default Java "missing file" error. This is something people who are a bit scared to share their docker.sock while exploring a framework might run into, as running with docker is the default. 

I would include in the error message something similar to: 

> Running with docker is not supported because the docker.sock file could not be accessed. Either make it available or run your task locally with the taskRunner.type: io.kestra.plugin.core.runner.Process.

### Environment

- Kestra Version: 1.3.3

shivamwayal37:
### ✨ Description

This PR improves Docker error handling by introducing a fail-fast validation for Unix socket-based Docker hosts.

- Adds pre-validation to detect missing or inaccessible Docker Unix sockets before execution
- Improves detection of Docker socket-related errors by inspecting nested exception causes
- Provides clearer and more actionable error messages when Docker is unavailable
- Adds dedicated unit tests (DockerErrorHandlingTest) to cover these scenarios
- Updates Docker integration tests to skip execution when the Docker daemon/socket is not available

This ensures a clear separation between environment/setup failures and container execution failures, improving reliability and debuggability.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/15161

### 🛠️ Backend Checklist

### 📝 Additional Notes

- IllegalStateException is intentionally used for pre-execution environment validation failures instead of TaskException, to avoid mixing distinct failure modes.
- Docker integration tests are conditionally skipped when the Docker daemon is not accessible, preventing false negatives in environments without Docker.


shivamwayal37:
It looks like the CI failure is unrelated to this PR:

- Failure occurs in `BlueprintControllerTest`
- Error: `Address already in use (port 28181)`
- This PR only modifies Docker runner logic

This seems like a flaky test or port conflict in CI.

Would you prefer:
1. I re-run the workflow?
2. Or open a separate PR to make the test use a random port?

Happy to address it separately if needed.

script/src/main/java/io/kestra/plugin/scripts/runner/docker/Docker.java
```diff
@@ -652,29 +658,80 @@ public void onComplete() {
...
-                            "Please make sure you properly mounted the Docker socket into your Kestra container (`-v /var/run/docker.sock:/var/run/docker.sock`) and that your user or group has at least the read and write privilege. "
-                            +
-                            "Tried socket: " + resolvedHost,
-                        e
-                    );
-                }
-            } catch (ClassNotFoundException ignored) {
-                // If we can't check if the stacktrace class is a Socket, we just ignore the exception
-                throw e;
+            if (isDockerSocketAccessError(e)) {
```
loicmathieu, line 658: As you manually validae the Docker socket is accessible before, I don't understand why you do that again here.

loicmathieu (changes requested):
This seems overly complicated.
What I don't unrestand is that we had a code block to handle specifically this kind of exception but it seems it didn't work, maybe fix that code block instead of removing it and replacing it?

shivamwayal37:
Thanks for the feedback!

You're right — the current approach introduces duplication between the pre-validation and the exception handling block.

My intention was to:
- Fail fast for missing Unix sockets before execution
- Improve detection of Docker socket-related errors from nested exceptions

But I see that this overlaps with the existing logic.

I'll refactor this to:
- Reuse and improve the existing error handling block instead of duplicating it
- Remove redundant checks between pre-validation and catch block

Will push an updated version shortly.

shivamwayal37:
Thanks for the review @loicmathieu!

I’ve refactored the implementation to remove duplication and simplify the flow:

- Reused and improved the existing error handling instead of replacing it
- Removed redundant checks between pre-validation and the catch block
- Kept fail-fast validation only for early Unix socket detection

This makes the behavior more consistent and easier to follow.

Regarding the CI failure: it occurs in jdbc-postgres (PostgresServiceLivenessCoordinatorTest) and appears unrelated to this PR. The script module (where these changes were made) is passing all tests successfully. This looks like a flaky or environment-related issue, but I’m happy to re-run the workflow or investigate further if needed.

loicmathieu (approved):
LGTM

loicmathieu:
Backported to 1.3 and 1.0

shivamwayal37:
Thanks @loicmathieu and @MilosPaunovic for the review and quick merge!

Glad the refactor aligns well. Appreciate the backport as well 🙌

PR #15392, opened 2026-04-06
feat(logs): add batchSize to PurgeLogs

Malaydewangan09:
- closes https://github.com/kestra-io/kestra-ee/issues/6139
- Companion PR: https://github.com/kestra-io/kestra-ee/pull/7331


core/src/main/java/io/kestra/core/repositories/LogRepositoryInterface.java
```diff
@@ -105,4 +105,9 @@ default int deleteByQuery(String tenantId, String namespace, String flowId, Stri
 
     int deleteByQuery(String tenantId, String namespace, String flowId, String executionId, List<Level> logLevels, ZonedDateTime startDate, ZonedDateTime endDate, boolean purgeExecutionLogs,
         boolean purgeNonExecutionLogs);
+
+    default int deleteByQuery(String tenantId, String namespace, String flowId, String executionId, List<Level> logLevels, ZonedDateTime startDate, ZonedDateTime endDate,
```
loicmathieu: I'm not fan of default method.
Is the previous method really used somewhere? I'll prefer to always add the limit.

loicmathieu: But maybe set it as Optional as some database may better work without the limit.

core/src/main/java/io/kestra/plugin/core/log/PurgeLogs.java
```diff
@@ -121,6 +121,14 @@ public class PurgeLogs extends Task implements RunnableTask<PurgeLogs.Output> {
     @Builder.Default
     private Property<Boolean> purgeNonExecutionLogs = Property.ofValue(true);
 
+    @Schema(
+        title = "The number of log rows deleted per batch",
+        description = "Deletion is performed in batches to limit transaction size. Tune this down if you hit database transaction size limits (e.g. MySQL group_replication_transaction_size_limit)."
+    )
+    @Builder.Default
+    @NotNull
+    private Property<Integer> batchSize = Property.ofValue(1000);
```
loicmathieu: I wonder if we should really set a limit by default.

On internal purge, we use 10000 by default, but we only use it for MySQL which is not possible to do here we don't want plugins to behaves by default differently from DB to DB so I'm not sure what's the best option here.

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLogRepository.java
loicmathieu, line 1: I don't think this is the right approach here, you select then delete which would not be very performant.
The best is to delete with a LIMIT and until the delete query returns 0.

You can check what we do for internal Queue purging which implements such algorithm for MySQL

loicmathieu (changes requested):

core/src/main/java/io/kestra/plugin/core/log/PurgeLogs.java
Malaydewangan09: Agreed, no default. 
Users on MySQL hitting `group_replication_transaction_size_limit` can set it explicitly. 

No point adding overhead for databases that don't need it.

core/src/main/java/io/kestra/core/repositories/LogRepositoryInterface.java
Malaydewangan09: hmm, it was not used, good point!

loicmathieu (approved):
LGTM

PR #15395, opened 2026-04-06
fix(executor): defer WorkerJobRunningStateStore and CREATED state event to fix flaky test initialization

ammeek:
### ✨ Description
defer WorkerJobRunningStateStore and CREATED state event to fix flaky test initialization                                                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                        
### 🛠️  Backend Checklist                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                               
### DIFF
--- a/executor/src/main/java/io/kestra/executor/DefaultExecutor.java
+++ b/executor/src/main/java/io/kestra/executor/DefaultExecutor.java
@@ -153,11 +153,12 @@ public class DefaultExecutor extends AbstractService implements Executor {
         this.numberOfThreads = threadCount != 0 ? threadCount : Math.max(4, KestraContext.getContext().getAllocatedCpuCores());
         this.workerTaskResultExecutorService = executorsUtils.maxCachedThreadPool(numberOfThreads, "executor-worker-task-result-executor");
         this.executionExecutorService = executorsUtils.maxCachedThreadPool(numberOfThreads, "executor-execution-event-executor");
-        setState(ServiceState.CREATED);
     }
 
     @PostConstruct
-    void initMetrics() {
+    void init() {
+        setState(ServiceState.CREATED);
+
         // create metrics to store thread count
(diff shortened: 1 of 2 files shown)


Malaydewangan09 (approved):

fdelbrayelle (approved):

loicmathieu (commented):
@fhussonnois WDYT?
Should we always swtich to CREATED in a postconstruct?

@ammeek It seems a bit odd to me, I agree to use a `Provider` to break dependency cycle but I'm not sure I understand why you have to set the CREATED state later

ammeek:
@loicmathieu setState publishes a ServiceStateChangeEvent which is consumed by the ServiceLivenessManager which I thought caused this bean to be initialised before the beans it needs are available. After taking a second look I can see that ServiceLivenessManager is a Context bean so the event publishing does not cause this issue like I thought.

I think that the event publishing should still be done in the post construct as it's not required to initialise the bean.

I'll take another look into what is causing the issue and update the description of this PR to include a more detailed description of the fix I find for it.

ammeek:
The failing JavaSecurityTest has stumped me. It looks like the bean definition for ServerConfig$Liveness does not exists. This is strange because the correct application.yaml values have been set. I thought that this might have been caused by the class being compiled before the correct configs are loaded but this does not seem to be the case.

I'll continue to investigate this. @loicmathieu if you have any idea of how i can go about this it would be appreciated.

https://github.com/kestra-io/kestra-ee/actions/runs/24029838381/job/70090388092

ammeek:
After some investigation last week we found out that the JavaSecurityTest was a lot less flaky than we originally thought .The test report was caching the results from earlier runs on the CI pipeline. Since JavaSecurityTest is no longer a blocking issue I am going to drop the investigation.

@fhussonnois would you be able we weigh in on Loïc comment? 

> @fhussonnois WDYT? Should we always switch to CREATED in a postconstruct?

PR #15268, opened 2026-03-26
fix(core): remove redundant details action button from flow and execution tables

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/13635.

@Pradumnasaraf Can you give this a check once Docker image is ready please?

PR #15402, opened 2026-04-07
fix(core): label check to add as filter and routerlink

### LINKED ISSUE #15401: Clicking on labels tag not adding filter.
### Describe the issue

Clicking on `labels` tag not adding filter and it just open the particular id in one click.

- single-click anywhere on the row navigates away
- No "Open in new tab" or copy the URL or `cmd/ctrl + click`

### Environment

- Kestra Version: develop

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra/issues/15401

### DIFF
--- a/ui/src/components/executions/Executions.vue
+++ b/ui/src/components/executions/Executions.vue
@@ -1100,6 +1100,10 @@
     color: var(--ks-content-primary);
 }
 
+:deep(.executions-table) .el-table__row {
+    cursor: pointer;
+}
+
 :deep(a.execution-id) code {
     color: var(--ks-content-link);
 }
--- a/ui/src/components/executions/Executions.vue
+++ b/ui/src/components/executions/Executions.vue
@@ -71,7 +71,7 @@
                     :defaultSort="{prop: 'state.startDate', order: 'descending'}"
                     tableLayout="auto"
                     fixed
-                    @row-click="(row: any) => onRowDoubleClick(executionParams(row))"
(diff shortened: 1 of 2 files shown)


MilosPaunovic (changes requested):
Hey, double click was removed in https://github.com/kestra-io/kestra/pull/15268.

Can you add `.prevent.stop` to the label click to sort the problem here?

Piyush-r-bhaskar:
Ahh sure, but still we have an issue with not able to open the id in new tab ?

MilosPaunovic:
Ooh, yes, that's true, so you should add back the router link on `id` field, absolutely right.

Piyush-r-bhaskar:
yeah, and I am going to remove `dblclick` and use the `.prevent.stop`

ui/src/components/executions/Executions.vue
```diff
@@ -1089,7 +1100,7 @@
     color: var(--ks-content-primary);
 }
 
-:deep(.executions-table) .el-table__row {
```
MilosPaunovic, line 1092: We'd also want to keep cursor pointer on the whole row, as now it's clickable,

MilosPaunovic (changes requested):

ui/src/components/executions/Executions.vue
Piyush-r-bhaskar: Absolutely right 🤩

MilosPaunovic (approved):
Perfect now! ✅

PR #15403, opened 2026-04-07
Feat invalid Unicode surrogate crash in JSON serialization

### LINKED ISSUE #14806: invalid unicode sequence crashes kestra executor pod
### Describe the issue

A recent incident on our kestra deployment was caused by a broken `insert` query shown below: 
```
2026-02-24 17:50:01 UTC:kestra_user@kestrav2:[3240]:ERROR:  invalid input syntax for type json
2026-02-24 17:50:01 UTC:kestra_user@kestrav2:[3240]:DETAIL:  Unicode low surrogate must follow a high surrogate.
2026-02-24 17:50:01 UTC:kestra_user@kestrav2:[3240]:CONTEXT:  JSON data, line 1: ...5\uDC59\uD835\uDC5E2 = lower limit of the \uD835....
2026-02-24 17:50:01 UTC:kestra_user@kestrav2:[3240]:STATEMENT:  insert into queues ("value", "type", "key") values (cast($1 as jsonb), CAST($2 AS queue_type), $3)
2026-02-24 17:50:01 UTC:kestra_user@kestrav2:[3240]:ERROR:  current transaction is aborted, commands ignored until end of transaction block
2026-02-24 17:50:01 UTC:kestra_user@kestrav2:[3240]:STATEMENT:  insert into executions ("key", "value") values ($1, cast($2 as jsonb)) on conflict ("key") do update set "value" = cast($3 as jsonb)

```

which stopped again after someone ran a `delete` query on the `queues` table manually. Further research has shown that the string `\uDC59\uD835\uDC5E2` is an invalid utf-16 string that is being cast to `jsonb`, to be inserted into multiple tables, and that is failing and causing the kestra executor to error out. More info on the utf-16 stuff here: https://paragraph.com/@sds.eth/here-be-dragons-using-javascript-to-store-json-in-a-postgres-database 

So there are two problems I can see: we're generating or getting UGC or AIGC that is invalid utf-16 encoded, and Kestra is trying to insert that into a DB with encoding constraints. For the first part: we can try to not have the summary directly in the text output of certain workflows as in the screenshot below, **for the second part, here I am to tell you to sanitize workflows in and output before inserting it into the typed jsonb columns.**

A minimal reproducer workflow is given below, it can be saved (with an error) but executing it will immediately kill kestra (tested on a local single-container setup with the latest kestra). Don't run this on a production system: 

```
id: whale_848283
namespace: company.team

tasks:
  - id: hello
    message: "\uDC59\uD835\uDC5E2 = lower limit of t"
```

### Environment

- Kestra Version: v1.1.5
Running as a separate executor pod in an EKS kubernetes cluster.

harsh21234i:
Added support for handling invalid UTF-16 surrogate sequences during JSON serialization.

- Introduced SurrogateStrippingStringSerializer to replace invalid surrogate characters with U+FFFD
- Integrated serializer into JacksonMapper
- Added comprehensive tests covering edge cases and reproducing issue #14806

This prevents crashes when serializing malformed Unicode strings (e.g., PostgreSQL jsonb rejection). 
Fixes #14806


core/src/main/java/io/kestra/core/serializers/SurrogateStrippingStringSerializer.java
```diff
@@ -0,0 +1,72 @@
...
+            if (Character.isSurrogate(input.charAt(i))) {
+                hasSurrogate = true;
+                break;
+            }
+        }
+        if (!hasSurrogate) {
+            return input;
+        }
+
+        // Slow path: rebuild string replacing invalid surrogate code units.
```
loicmathieu: Replacing may be seen very opinionated, by what?

core/src/main/java/io/kestra/core/serializers/SurrogateStrippingStringSerializer.java
```diff
@@ -0,0 +1,72 @@
...
+            throws IOException {
+        gen.writeString(sanitize(value));
+    }
+
+    public static String sanitize(String input) {
+        if (input == null) {
+            return null;
+        }
+
+        // Fast path: return original object if no surrogates present.
```
loicmathieu: This would be a high price for 99.999999% of strings that are valid.

loicmathieu (changes requested):
I understand the rational but I'm not fan.
Doing this is highly opinionated and would incure a performance cost of reading the whome JSON byte array.

I would prefer that we fail gracely if there is an unsupported character sequence, we already handled that in the queue and he executor I think (we catch the Postgres specific error message and fail the exec).

A better approach may also to have validation whenever possible to avoid such un-processable strings to be set  (like in inputs).

harsh21234i:
Hello @loicmathieu ,
I updated the PR to remove the global surrogate sanitization approach and keep the fix on the existing Postgres queue failure path.

I also tightened the queue-side detection so unsupported Unicode payload failures are still converted to UnsupportedMessageException, including lone surrogate cases, and kept regression coverage on the Postgres queue tests without relying on one exact Postgres error string.

I couldn’t fully verify the Postgres test locally because my local test environment does not currently have Postgres available on localhost:5432, but the code compiles and the change is now targeted to the existing recoverable failure path rather than global string replacement.

harsh21234i:
Backend tests themselves passed; the failing job is in the reporting step after Gradle and looks related to fork PR permissions ("Resource not accessible
  by integration", "Missing organization token"), so it seems unrelated to the code change in this PR.

loicmathieu (approved):
LGTM

PR #15414, opened 2026-04-07
fix(jdbc): avoid race condition between database migrations and editi…

### LINKED ISSUE #15413: fix(core): Kestra fails to start against a fresh PostgreSQL database
Starting Kestra against a fresh (empty) PostgreSQL database fails immediately with:

```
Exception in thread "main" io.micronaut.context.exceptions.BeanInstantiationException: 
Bean definition [io.kestra.core.server.ServiceLivenessManager] could not be loaded: 
Error instantiating bean of type [io.kestra.core.utils.VersionProvider]
Message: SQL [select "value" from settings where (true and "key" = ?) fetch next ? rows only]; 
ERROR: relation "settings" does not exist
```

Kestra starts fine if the database was previously initialized (tables already exist from a prior run).

brian-mulier-p:
…on persistance

closes https://github.com/kestra-io/kestra/issues/15413

### DIFF
--- a/core/src/main/java/io/kestra/core/utils/EditionProvider.java
+++ b/core/src/main/java/io/kestra/core/utils/EditionProvider.java
@@ -5,8 +5,6 @@ import java.util.Optional;
 import io.kestra.core.models.Setting;
 import io.kestra.core.repositories.SettingRepositoryInterface;
 
-import jakarta.annotation.PostConstruct;
-import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 
 @Singleton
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/kestra/core/utils/EditionProvider.java
```diff
@@ -5,8 +5,6 @@
 import io.kestra.core.models.Setting;
 import io.kestra.core.repositories.SettingRepositoryInterface;
 
-import jakarta.annotation.PostConstruct;
-import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 
 @Singleton
```
fhussonnois, line 10: I really think `utils` package should not contain Micronaut beans.

loicmathieu: yeah, utils should basically only be static utility classes

loicmathieu (approved):
I approve it as it solves the issue, the remaining suggestion from Florian was pre-dating the fix so it can be done later

PR #15445, opened 2026-04-08
fix: replace Sleep with CountDownLatchTask in liveness coordinator tests

ammeek:
✨ Description                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                 
Replaces Sleep-based WorkerTask fixtures in AbstractServiceLivenessCoordinatorTest with CountDownLatchTask, eliminating the race condition where the task could complete before worker.close() was called and the liveness coordinator never saw 
a mid-flight job.                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                 
Also adds annotationProcessor project(':processor') to tests/build.gradle so CountDownLatchTask is registered as a plugin and can be deserialized from the JDBC queue. Updates getTaskForCountDownLatch to set id, type, and support a           
WorkerGroup parameter.


Skraye (approved):
LGTM

PR #15451, opened 2026-04-08
fix(logs): better handling on repository filters inheritance

brian-mulier-p:
closes https://github.com/kestra-io/kestra-ee/issues/7327

The only part worth checking is the defaultFilters refactoring at the end, other changes are pure spotless (idk why it applies now, I thought there was a big apply at some point).

The goal is to make sure there are no gaps in our default filter handling and prevent breakage due to inclusion of "deleted" condition in logs which are no longer soft-deletable


nkwiatkowski (approved):
LGTM

PR #15455, opened 2026-04-08
refactor(triggers): replace Execution with lightweight TriggerEvaluationResult in WorkerTriggerResult

fhussonnois:
The full Execution object (23 fields) was being transported over gRPC from worker to controller, but trigger-created executions only populate 9 fields.

Introduce TriggerEvaluationResult, a self-contained record carrying only the necessary data (executionId, stateType, trigger, labels, flowRevision, flowVariables). The full Execution is reconstructed on the controller/scheduler side via toExecution(TriggerId) without any Flow lookup.

Also removes the deprecated io.kestra.core.runners.WorkerTriggerResult class which had no remaining usages.


core/src/main/java/io/kestra/core/worker/models/TriggerEvaluationResult.java
```diff
@@ -0,0 +1,83 @@
...
+ * @param flowRevision  the flow revision at evaluation time.
+ * @param flowVariables the flow-level variables (typically small or null).
+ */
+public record TriggerEvaluationResult(
+    @JsonProperty String executionId,
+    @JsonProperty State.Type stateType,
+    @JsonProperty ExecutionTrigger trigger,
+    @JsonProperty @Nullable List<Label> labels,
+    @JsonProperty @Nullable Integer flowRevision,
+    @JsonProperty @Nullable Map<String, Object> flowVariables
```
loicmathieu: Is it really needed? Normally triggers should pass variables via the `ExecutionTrigger` not from the flowVariables that are normally defined inside the flow.

loicmathieu (commented):
Overall I'm not against it but it seems to be a hack inside the Worker/Scheduler where a proper fix would have been to change the contract of triggers.

Here, triggers create an execution, then you decide that on this execution, only a subset of fields are worth it (it may lead to subtle bugs, especially as some triggers may do unexpected things) and only transfer those to recreate the execution later.

For me, the trigger evaluation should have been changed to return the new `TriggerEvaluationResult` it would have been cleaner but I can understand that it's not really feasible.

This looks like a _shadow contract_ that the trigger is not aware of so I'm a bit concern about it

core/src/main/java/io/kestra/core/models/triggers/PollingTriggerInterface.java
```diff
@@ -24,9 +23,22 @@ public interface PollingTriggerInterface extends WorkerTriggerInterface {
...
+    default Optional<TriggerEvaluationResult> eval(ConditionContext conditionContext, TriggerContext context) throws Exception {
+        return evaluate(conditionContext, context).map(TriggerEvaluationResult::from);
+    }
+
     /**
      * Evaluate the trigger and create an execution if needed.
+     *
+     * @deprecated Override {@link #eval} instead.
      */
+    @Deprecated
```
loicmathieu:
```suggestion
    @Deprecated(forRemoval=true, since="2.0.0")
```

core/src/main/java/io/kestra/core/models/triggers/PollingTriggerInterface.java
loicmathieu, line 1: As there is a default implementation for the new method but no default for the old method, dev would still need to implement the old one!

I think the best would be to add a default implementation to the old method that would throw an UnimplementedException with a message saying that they must implement the new method.
That would catch up mis-implementation.

core/src/main/java/io/kestra/core/models/triggers/RealtimeTriggerInterface.java
loicmathieu, line 1: Same here, deprecation config and default method

core/src/main/java/io/kestra/core/models/triggers/TriggerService.java
```diff
@@ -9,10 +9,78 @@
 import io.kestra.core.models.flows.State;
 import io.kestra.core.models.tasks.Output;
 import io.kestra.core.runners.RunContext;
+import io.kestra.core.services.LabelService;
 import io.kestra.core.utils.IdUtils;
 import io.kestra.core.utils.ListUtils;
 
 public abstract class TriggerService {
+
+    // -- New API returning TriggerEvaluationResult --
```
loicmathieu: Claude keep adding these kind of non-useful comment :( 
I'll remove them as they didn't bring any meaning

core/src/main/java/io/kestra/core/models/triggers/TriggerService.java
```diff
@@ -9,10 +9,78 @@
...
+            conditionContext.getFlow().getRevision()
+        );
+    }
+
+    // -- Deprecated API returning Execution --
+
+    /**
+     * @deprecated Use {@link #generateEvaluationResult(AbstractTrigger, ConditionContext, TriggerContext, Map)} instead.
+     */
+    @Deprecated
```
loicmathieu: add forRemoval and since to all your deprecated annotation

loicmathieu (changes requested):

loicmathieu (approved):
LGMT, nice refacto

PR #15469, opened 2026-04-09
fix(plugins): reset version when navigating between plugins

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra-ee/issues/6971

there was two issue due to this:

- version related - `Illegal argument: Unsupported version for type `
- and interrelated to version , `404` when version not found for few plugins.


MilosPaunovic (approved):

MilosPaunovic (approved):
Nice catch! 🚀

PR #15477, opened 2026-04-09
fix(core): pull $group and $index up from anyOf in JSON schema generator

fdelbrayelle:
## Summary

- For `Property<Boolean>` (and any `Property<T>` with multiple target types), victools generates an `anyOf` schema and places `$group`/`$index` inside each `anyOf` item instead of at the parent level
- `pullDocumentationAndDefaultFromAnyOf` already hoisted `default`, `title`, `description`, and `$deprecated` — this extends it to also hoist `$group` and `$index`
- Without this fix, properties like `failedOnMissing` (a `Property<Boolean>` annotated with `@PluginProperty(group = "reliability")`) appeared in the "Optional" section of the no-code editor instead of their own group section

## Test plan

- [ ] Added `TaskWithGroupedBooleanProperty` test task and `propertyBooleanWithGroupShouldHaveGroupAtTopLevel` test in `JsonSchemaGeneratorTest`
- [ ] Test asserts `$group` is at the top level for both the `properties()` and `schemas()` code paths
- [ ] Confirmed with `./gradlew :core:test --tests "io.kestra.core.docs.JsonSchemaGeneratorTest.propertyBooleanWithGroupShouldHaveGroupAtTopLevel"`

part-of: https://github.com/kestra-io/kestra-ee/issues/6712

### DIFF
--- a/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
+++ b/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
@@ -186,7 +186,9 @@ public class JsonSchemaGenerator {
                             Map.entry("default", Optional.empty()),
                             Map.entry("title", Optional.empty()),
                             Map.entry("description", Optional.empty()),
-                            Map.entry("$deprecated", Optional.empty())
+                            Map.entry("$deprecated", Optional.empty()),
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
LGTM

PR #15479, opened 2026-04-09
Conditions as Pebble functions

loicmathieu:


core/src/main/java/io/kestra/core/runners/pebble/functions/IsPublicHolidayFunction.java
```diff
@@ -0,0 +1,83 @@
...
+            : null;
+
+        LocalDate localDate;
+        try {
+            localDate = DateUtils.parseLocalDate(dateArg.toString());
+        } catch (InternalException e) {
+            throw new PebbleException(e, "The 'isPublicHoliday()' function could not parse 'date': " + e.getMessage(), lineNumber, self.getName());
+        }
+
+        HolidayManager holidayManager = HolidayManager.getInstance(ManagerParameters.create(countryCode));
```
brian-mulier-p, line 61: Try catch in case of wrong country code, this is the reason for failing test

brian-mulier-p (changes requested):
Add a try catch and we're good

core/src/main/java/io/kestra/core/runners/pebble/functions/IsPublicHolidayFunction.java
loicmathieu: I removed it because, well, we allow Pebble function to fail right?
So I may just change the test as I updated manually the function.

brian-mulier-p (approved):
LGTM

PR #15480, opened 2026-04-09
fix(system): add split attribute to the excludes command line option

loicmathieu:


fhussonnois (approved):

loicmathieu:
Backported to 1.0, 0.24 and 0.23

PR #15482, opened 2026-04-09
fix(flows): fix OpenAPI spec generation for FlowController endpoints

### LINKED ISSUE #15483: openapi spec generation fails on Flows
since we bumped micronaut platform https://github.com/kestra-io/kestra/pull/15426 something changed in our openapi spec generation -> most Flows endpoints are not generated anymore

AcevedoR:
- fixes https://github.com/kestra-io/kestra/issues/15483

## Summary

- The Micronaut OpenAPI annotation processor was skipping all `FlowController` endpoints mapped to `{namespace}/{id}` (GET, PUT, DELETE), generating an empty path `{}` in `openapi.yml`
- Root cause: processing `FlowWithSource` → `Flow`/`Task` → `AbstractRetry` → concrete subtypes (`Constant`, `Exponential`, `Random`) triggered a `PostponeToNextRoundException` because Micronaut constraint validators on those subtypes are processed in the same annotation processing round, making their metadata unavailable to the OpenAPI processor
- Fix: annotate the `retry` field in both `Flow` and `Task` with `@Schema(implementation = Object.class)` to prevent the processor from following the `@JsonSubTypes` chain on `AbstractRetry`; add class-level `@Schema` documentation to the concrete retry subtypes

## Test plan

- [ ] Run `./gradlew generateOpenapiSpec --rerun-tasks` and verify no `PostponeToNextRoundException` warning in the output
- [ ] Verify `/api/v1/{tenant}/flows/{namespace}/{id}` in `openapi.yml` contains `get`, `put`, and `delete` operations (was previously `{}`)
- [ ] Verify no existing tests are broken: `./gradlew :webserver:test`


core/src/main/java/io/kestra/core/models/flows/Flow.java
```diff
@@ -103,6 +103,14 @@ public List<Task> getFinally() {
     @Valid
     List<Output> outputs;
 
+    // implementation = Object.class prevents the Micronaut OpenAPI annotation processor from following
+    // the @JsonSubTypes on AbstractRetry, which causes a PostponeToNextRoundException at compile time
+    // due to the Micronaut constraint validators on the concrete retry subtypes (Constant, Exponential, Random).
```
loicmathieu, line 108: Did you openned and issue about that?
May be worth openning one and adding the link here so we can go back to normal when fixed inside Micronaut

loicmathieu (approved):
The fix is a bit disapointing but if there are no other way around, let's do that.

Did you check autocompletion? It may no longuer work for retry which would not be very convenient.

PR #15484, opened 2026-04-09
fix: eliminate all Java compiler warnings (204 → 0)

AcevedoR:
## Summary

- Reduces Java compiler warnings from **204 to 0** across all 13 OSS modules
- Changes are purely mechanical (no logic changes): `@SuppressWarnings`, `serialVersionUID`, minor API replacements
- Organized into 6 batches, each independently safe and testable

## Changes by category

| Category | Fix applied |
|----------|-------------|
| `[serial]` missing serialVersionUID (7 classes) | Added `private static final long serialVersionUID = 1L` |
| `[rawtypes]`/`[unchecked]` (68 warnings) | Added `@SuppressWarnings` at narrowest scope; used `TypeReference<Map<String,Object>>` in CSV export endpoints to avoid raw-type cascade |
| `@SafeVarargs` varargs heap pollution (14) | Added `@SafeVarargs` + `@SuppressWarnings("varargs")` on final/static methods |
| Jackson2/Jackson3 Lombok ambiguity (24) | Set `lombok.addJacksonAnnotations=jackson2` in `lombok.config` |
| `@SuperBuilder` ignoring initializer (5) | Added `@Builder.Default` to affected fields |
| `[deprecation]` — Await class (21+) | `@SuppressWarnings("deprecation")` on containing methods |
| `[deprecation]` — other APIs (6) | Replaced `ObjectNode.fields()` → `properties()`, `Artifact.getFile()` → `getPath()`, StringUtils wrappers → plain String methods; suppressed RouteMatchUtils and gRPC StaticNameResolver |
| `[try]` — AutoCloseable + InterruptedException (12) | Added `default void close() throws IOException` to `GenericQueueInterface` (narrows contract safely); `@SuppressWarnings("try")` on `PluginManager`, `LocalPluginManager`, and use sites |
| `[this-escape]` (5) | `@SuppressWarnings("this-escape")` on constructors calling overridable methods |
| Static method qualification / nullable on primitive | Fixed directly at call site |

## Non-trivial semantic fixes (not just suppression)

- **`GenericQueueInterface`**: added `default void close() throws IOException {}` — narrows the inherited `throws Exception` contract. Verified no OSS or EE implementations throw `InterruptedException` from `close()`.
- **`QueryFilter.asValues()`**: widened `String[]` to `Object[]` before `Arrays.asList()` to eliminate varargs ambiguity without changing behavior.
- **CSV export endpoints** (`ExecutionController`, `FlowController`, `TriggerController`): used `TypeReference<Map<String,Object>>` instead of raw `Map.class` in `objectMapper.convertValue()` — this also fixes a latent raw-type propagation that was causing a compile error.
- **`JsonSchemaGenerator`**: replaced deprecated `ObjectNode.fields()` with `ObjectNode.properties()` (Jackson 2.21+).
- **`MavenPluginDownloader`**: replaced deprecated `Artifact.getFile()` with `Artifact.getPath()` (Maven Resolver 2.x).

## Test plan

- [ ] All modules compile cleanly: `./gradlew compileJava --rerun 2>&1 | grep "warning:" | grep -v micronaut` → no output
- [ ] Build succeeds: `./gradlew build -x test -x integrationTest -x testCodeCoverageReport --parallel`
- [ ] Core tests pass: `./gradlew :core:test`
- [ ] Webserver tests pass: `./gradlew :webserver:test`
- [ ] JDBC tests pass: `./gradlew :jdbc:test`
- [ ] Worker/controller tests pass: `./gradlew :worker:test :worker-controller:test`

### DIFF
--- a/cli/src/main/java/io/kestra/cli/StandAloneRunner.java
+++ b/cli/src/main/java/io/kestra/cli/StandAloneRunner.java
@@ -5,13 +5,14 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.kestra.core.runners.*;
 import io.kestra.core.server.Service;
-import io.kestra.core.utils.Await;
 import io.kestra.core.utils.ExecutorsUtils;
+
+import org.awaitility.Awaitility;
+import org.awaitility.core.ConditionTimeoutException;
 import io.kestra.core.worker.Controller;
 import io.kestra.executor.DefaultExecutor;
 
@@ -52,7 +53,6 @@ public class StandAloneRunner implements Runnable, AutoCloseable {
     private ExecutorService poolExecutor;
 
     @Override
-    @SuppressWarnings("deprecation")
     public void run() {
         running.set(true);
 
@@ -84,8 +84,10 @@ public class StandAloneRunner implements Runnable, AutoCloseable {
         }
 
         try {
-            Await.until(() -> servers.stream().allMatch(s -> Optional.ofNullable(s.getState()).orElse(Service.ServiceState.RUNNING).isRunning()), null, runningTimeout);
-        } catch (TimeoutException e) {
+            Awaitility.await().atMost(runningTimeout).until(
+                () -> servers.stream().allMatch(s -> Optional.ofNullable(s.getState()).orElse(Service.ServiceState.RUNNING).isRunning())
+            );
+        } catch (ConditionTimeoutException e) {
             throw new RuntimeException(
                 servers.stream().filter(s -> !Optional.ofNullable(s.getState()).orElse(Service.ServiceState.RUNNING).isRunning())
                     .map(Service::getClass)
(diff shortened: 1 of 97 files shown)


tchiotludo:
OMG :100: 
any way to prevent new compilation warning to happen, it's awful on the DX ! 
Thanks for jumping into the subject ^^

AcevedoR:
yeah if we have 0 warnings then we can pass something like -Werror to fail compile when there is warnings

loicmathieu (changes requested):
Hi,
This is cool!
Some realy feedback, didn't review the code yet.

> Added @SafeVarargs 

How did you check that the varagrs are safe from heap pollution?

> [deprecation] — Await class (21+)

I'm not fan, if we decided to deprecate it, but hide the deprecation, nobody will ever do anything about it. In this case I'll suggest either replacing its usage by Awaitility or acknowledge we will not remove it and promote it to non-deprecated.

> [try] — AutoCloseable + InterruptedException (12)

I'm not sure it's a good idea to hide this warning anywhere as it's usually a resource leak (and it will definitly in tests as the JVM is reused)

> [this-escape] (5)

I have mix feelings about it, it's a new warning sometimes hard to understand but it indicate potential subtle bugs at instantiasion (and, well, we have from time to time test that fails miserably we don't know why so it's a lead).
Did you check that those are un-harmed?

AcevedoR:
I'll do it iteratively instead https://github.com/kestra-io/kestra/pull/15503

PR #15486, opened 2026-04-10
chore(docs): optimize AGENTS.md for agent accuracy

fhussonnois:
Restructure guidelines with critical code patterns (DI, error handling) prioritized at the top in DO/DON'T format. Remove generic "Common Development Tasks" section, consolidate duplicate rules, and reorder sections by violation frequency to improve AI agent code accuracy.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,15 +16,199 @@ This document provides essential information for AI coding agents working on the
 - **Write clear, maintainable, and well-documented code**
 - **Build & test are mandatory**
 
-## Project Overview
+## Project
 
-It's a monorepo built with Java (backend) and Vue (frontend), using Gradle as the build system.
+Monorepo built with Java (backend) and Vue (frontend), using Gradle as the build system.
 
-**Core Technologies:**
-- **Backend:** Java 25, Micronaut Framework, Lombok, JUnit 5
+## Tech Stack
+- **Backend:** Java 25, Micronaut Framework, Lombok
 - **Frontend:** Vue 3, TypeScript, Vite, Element Plus, Pinia
 - **Build:** Gradle 8.x with multi-project structure (77 submodules)
 - **Testing:** JUnit 5, Mockito, AssertJ, Vitest, Playwright
+
+## Critical Code Patterns
+
+### Dependency Injection
+
+**DO**: Use constructor injection with final fields.
+
+```java
+@Singleton
+public class MyService {
+    private final SomeDependency dependency;
+
+    @Inject
+    public MyService(SomeDependency dependency) {
+        this.dependency = Objects.requireNonNull(dependency);
+    }
+}
+```
+
+**DON'T**: Use field injection (`@Inject` on fields directly). Always prefer constructor injection.
+
+### Class Structure
+
+```java
+// 1. Package declaration and imports
+// 2. Class-level annotations (@Slf4j, @Singleton, etc.)
+// 3. Class declaration with Javadoc
+// 4. Static constants (UPPER_SNAKE_CASE)
+// 5. Injected fields (@Inject)
+// 6. Constructors
+// 7. Public methods
+// 8. Protected methods  
+// 9. Private methods
+// 10. Inner classes/records
+```
+
+### Annotations
+- **Micronaut:** `@Singleton`, `@Inject`, `@Controller`, `@Replaces`, `@Requires`
+- **Validation:** `@Valid`, `@NotNull`, `@Nullable`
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -16,120 +16,38 @@ This document provides essential information for AI coding agents working on the
...
-npm run build-storybook # Build
+    @Inject
+    public MyService(SomeDependency dependency) {
+        this.dependency = Objects.requireNonNull(dependency);
+    }
+}
 ```
 
-## Code Style Guidelines
+**DON'T**: Use field injection (`@Inject` on fields directly). Always prefer constructor injection.
```
loicmathieu, line 47: The only issue with constructor injection is when there are like 20 dependencies, at this stage I prefere field injection (highly debatable, should also be a sign of needing a refactoring to split it)

loicmathieu: Need to check but usually Sonar rules flag constructors with more than 7 or 9 parameters

AGENTS.md
```diff
@@ -227,13 +144,16 @@ public enum MyEnum {
...
 - Use `@KestraTest` for Kestra-specific integration test configuration
+- Use `@MicronautTest` only when `@KestraTest` is not appropriate
 - Place tests in same package structure as source code
 - Simple unit test with mocks over complex integration tests when possible
 - Add // Given-When-Then comments for clarity
 
-**Test Structure:**
+**DO**:
+- Always use naming conventions for test methods (e.g., `shouldPerformActionWhenCondition`)
+- Use `@KestraTest` for tests that require running Kestra services (e.g., Executor, Scheduler)
```
loicmathieu, line 150: Maybe specify using `@MicronautTest` otherwise at it tends to always use `@KestraTest`.

AGENTS.md
```diff
@@ -263,43 +185,111 @@ class ServiceTest {
...
+
+# Create executable JAR
+./gradlew executableJar --parallel
+
+# Run application locally
+./gradlew runLocal
+
+# Run standalone server
+./gradlew runStandalone
+```
```
loicmathieu: I cannot see why it should do that, I'll remove that

AGENTS.md
```diff
@@ -263,43 +185,111 @@ class ServiceTest {
...
+
+```bash
+# Clean build
+./gradlew clean
+
+# Full build (includes tests)
+./gradlew build
+
+# Build without tests (faster)
+./gradlew build -x test -x integrationTest -x testCodeCoverageReport --refresh-dependencies --no-daemon --parallel
```
loicmathieu, line 220: Would it be more simpler to only compile? The only caveat would be for UI.
But I would also add a suggestion to at least compile the affected modules as once it changes code without recompilation and ... it fails.

```suggestion
./gradlew build compileTestJava
```

AGENTS.md
```diff
@@ -263,43 +185,111 @@ class ServiceTest {
...
+
+```bash
+# Run all tests (excludes flaky tests)
+./gradlew test
+
+# Run only unit tests (fastest)
+./gradlew unitTest
+
+# Run integration tests
+./gradlew integrationTest
```
loicmathieu, line 233: We don't have any

loicmathieu (changes requested):
Maybe a single line explaining what Kestra is?

AGENTS.md
```diff
@@ -360,9 +330,40 @@ docker compose -f docker-compose-ci.yml down
...
+- `cli` - Command Line Interface
+- `core` - Core functionality
+- `webserver` - Web server
+- `ui` - Vue 3 frontend application
+- `executor` - The component responsible for managing execution state
+- `scheduler` - The component responsible for scheduling polling and schedule triggers
+- `worker` - The component that executes tasks and manages worker instances
+- `worker-controller` - The component that manages worker instances and job distribution
+- `indexer` - The component responsible for indexing executions
+- `plateform` - provides the Platform Bill of Materials (BOM) for dependency management
```
loicmathieu, line 331: I think those are really not enough explained, especially the executor/schedule/worker.
core also contains so many things, including flowable tasks that may be explained separatly.

loicmathieu (changes requested):

fdelbrayelle (approved):

AGENTS.md
fhussonnois: Agree, but details of each module should not be in AGENTS.md but in dedicated `arch` md files that the agant can load on demand.

AGENTS.md
fhussonnois: All KestraTest are integrationTest - we have a dedicated task for that

PR #15490, opened 2026-04-10
fix(core): add ReDoS protection for user-supplied regex operations

### LINKED ISSUE #15488: Regex Pebble filters are vulnerable to ReDoS (catastrophic backtracking)
## Problem

All regex-capable Pebble expression filters (`regexMatch`, `regexReplace`, `regexExtract`, and `replace` with `regexp=true`) call `Pattern.compile(regex)` with user-supplied patterns and no safeguards against catastrophic backtracking.

A pathological regex like `(a+)+$` against input `"aaaaaaaaaaaaaaaaaaaaaaaaaaaaab"` (29 `a`s + `b`) causes the engine to explore ~2^29 backtracking paths, hanging the worker thread indefinitely.

### Steps to Reproduce

1. Create a flow using a regex filter with a vulnerable pattern:
   ```yaml
   - id: check
     format: "{{ inputs.data | regexMatch(regex='(a+)+$') }}"
   ```
2. Execute with `inputs.data` set to a long string of `a`s followed by a non-matching character (e.g., 30+ `a`s + `b`)
3. The worker thread hangs for minutes or longer

### Proposed Mitigation

1. **Timeout guard** — Wrap regex operations in a `Future` with a configurable timeout (e.g., 5s). A shared utility like `RegexUtils.withTimeout()` would cover all regex filters. I think this would be the fastest and easiest to implement, and would neutralize the threat on its own.
2. **Input/pattern length limits** — Reject patterns above a configurable max length and inputs above a max size to bound the worst-case backtracking.
3. **Static pattern analysis (optional)** — Detect known-dangerous patterns (nested quantifiers, overlapping alternations) at compile time and reject them.

Related: #15351

brian-mulier-p:
Closes #15488

### DIFF
--- a/core/src/main/java/io/kestra/core/utils/RegexConfiguration.java
+++ b/core/src/main/java/io/kestra/core/utils/RegexConfiguration.java
@@ -20,7 +20,7 @@ public class RegexConfiguration {
      * Maximum duration for a single regex operation before it is aborted.
      * Defaults to 30 seconds.
      */
-    private Duration timeout = Duration.ofSeconds(30);
+    private Duration timeout = Duration.ofSeconds(10);
 
     @PostConstruct
     void init() {
--- a/core/src/main/java/io/kestra/core/utils/RegexConfiguration.java
+++ b/core/src/main/java/io/kestra/core/utils/RegexConfiguration.java
@@ -1,0 +1,29 @@
+package io.kestra.core.utils;
+
+import java.time.Duration;
+
+import io.micronaut.context.annotation.ConfigurationProperties;
+import jakarta.annotation.PostConstruct;
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * Configuration for regex timeout. Applies to all user-supplied regex operations
+ * (Pebble filters, input validators, storage split, etc.).
+ */
+@Getter
+@Setter
+@ConfigurationProperties("kestra.regex")
+public class RegexConfiguration {
+
+    /**
+     * Maximum duration for a single regex operation before it is aborted.
+     * Defaults to 30 seconds.
+     */
+    private Duration timeout = Duration.ofSeconds(30);
+
+    @PostConstruct
+    void init() {
+        RegexUtils.setTimeout(timeout);
+    }
+}
--- a/core/src/main/java/io/kestra/core/utils/RegexUtils.java
+++ b/core/src/main/java/io/kestra/core/utils/RegexUtils.java
@@ -1,0 +1,202 @@
+package io.kestra.core.utils;
+
+import java.time.Duration;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Utilities for running regex operations with a timeout guard to prevent ReDoS
+ * (catastrophic backtracking) attacks.
+ *
+ * <p>All methods wrap the input {@link CharSequence} so that every {@code charAt()} call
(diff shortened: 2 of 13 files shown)


core/src/main/java/io/kestra/core/utils/RegexUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private RegexUtils() {
+    }
+
+    /**
+     * Sets the global regex timeout.
+     *
+     * @param duration the maximum duration for any regex operation.
+     */
+    public static void setTimeout(Duration duration) {
```
loicmathieu, line 34: I'm not comfortable with that, it may allow bad usage.
Maybe add a defensive check via an AtomicBoolean so it can be called only once?

core/src/main/java/io/kestra/core/utils/RegexConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+@ConfigurationProperties("kestra.regex")
+public class RegexConfiguration {
+
+    /**
+     * Maximum duration for a single regex operation before it is aborted.
+     * Defaults to 30 seconds.
+     */
+    private Duration timeout = Duration.ofSeconds(30);
+
+    @PostConstruct
```
loicmathieu, line 25: I was not aware you can do that in configs, seems like either a very bad idea of a clever idea, not sure.

core/src/main/java/io/kestra/core/utils/RegexUtils.java
```diff
@@ -0,0 +1,150 @@
...
+ * Utilities for running regex operations with a timeout guard to prevent ReDoS
+ * (catastrophic backtracking) attacks.
+ *
+ * <p>All methods wrap the input {@link CharSequence} so that every {@code charAt()} call
+ * checks a deadline. If the deadline is exceeded, a {@link RegexTimeoutException} is thrown,
+ * terminating the backtracking without needing a separate thread.</p>
+ */
+public final class RegexUtils {
+
+    private static volatile Duration timeout = Duration.ofSeconds(30);
```
loicmathieu: This is already very large!
But some JVM stop may lead to false positive (like full GC or other kind of thread starvation induced issues)

loicmathieu (commented):
I'll let our security officer to due the review as it's security related @tchiotludo

core/src/main/java/io/kestra/core/utils/RegexConfiguration.java
brian-mulier-p: Why not? it's a regular bean 🤔

core/src/main/java/io/kestra/core/utils/RegexUtils.java
brian-mulier-p: I know but I'm afraid to be disruptive by setting a low limit, I was about to put 5 seconds or so but... idk if there is a good limit tbh, that's why it's configurable and a high limit by default to simulate previous behavior (aka no timeout) with a little guardrail for very abusive behaviors, I'm not against changing it but I need a new default limit :P

loicmathieu: This is I think the biggest issue, it's not easy to define what is the sane limit...
Maybe 15s? or 10s?

tchiotludo (approved):

PR #15504, opened 2026-04-10
feat(flow): Flow trigger dependsOn

loicmathieu:
Implements the new `dependsOn` mechanism for flow triggers.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/triggers/Window.java
+++ b/core/src/main/java/io/kestra/core/models/triggers/Window.java
@@ -1,0 +1,119 @@
+package io.kestra.core.models.triggers;
+
+import io.kestra.core.models.annotations.PluginProperty;
+import io.kestra.core.validations.WindowValidation;
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotNull;
+import lombok.Builder;
+import lombok.Getter;
+
+import java.time.Duration;
+import java.time.LocalTime;
+
+@Builder
+@Getter
+@WindowValidation
+@Schema(
+    title = "Window configuration",
+    description = """
+        Defines the time window within which all `dependsOn` conditions must be met for the trigger to fire.
+        The window type is inferred from the fields that are set:
+        - `deadline` set: daily time deadline window (conditions must be met before the given time each day).
+        - `from` and `to` both set: daily time window (conditions must be met within the given time range each day).
+        - `lookback` set: sliding window (conditions must be met within the past duration).
+        - otherwise: duration window (default, conditions must be met within a fixed duration, configurable via `every` and `offset`)."""
+)
+public class Window {
+    @Schema(
+        title = "Daily deadline",
+        description = "Use this to define a `DAILY_TIME_DEADLINE` window: the `dependsOn` conditions must be met before this time each day. Mutually exclusive with `from`, `to`, `lookback`, `every`, and `offset`."
+    )
+    @PluginProperty
+    private LocalTime deadline;
+
+    @Schema(
+        title = "Daily window start time",
+        description = "Use this together with `to` to define a `DAILY_TIME_WINDOW`: the `dependsOn` conditions must be met within the time range `[from, to]` each day. Mutually exclusive with `deadline`, `lookback`, `every`, and `offset`."
+    )
+    @PluginProperty
+    private LocalTime from;
+
+    @Schema(
+        title = "Daily window end time",
+        description = "Use this together with `from` to define a `DAILY_TIME_WINDOW`: the `dependsOn` conditions must be met within the time range `[from, to]` each day. Mutually exclusive with `deadline`, `lookback`, `every`, and `offset`."
+    )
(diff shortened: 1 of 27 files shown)


loicmathieu:
Missing features for now:

- [ ] Scoped outputs
- [ ] onMiss SLA
- [ ] mode and minSatisfied
- [ ] stable condition ID (they are generated the same as for precondition)
- [ ] silentFailures should creates FAILED execution

Missing a decision what to do for the existing precondition.

core/src/main/java/io/kestra/core/models/triggers/Window.java
```diff
@@ -0,0 +1,119 @@
...
+        description = "Use this to define a `SLIDING_WINDOW`: the `dependsOn` conditions must be met within the past duration relative to the current time. Mutually exclusive with `deadline`, `from`, `to`, `every`, and `offset`."
+    )
+    @PluginProperty
+    private Duration lookback;
+
+    @Schema(
+        title = "Whether the trigger can fire only once per window",
+        description = """
+            When `false` (the default), the window state is NOT reset after a successful evaluation, meaning the trigger can fire again within the same window each time conditions are satisfied.
+            When `true`, after a successful evaluation the window state is reset, so the same set of conditions must be met again within the window to trigger a new execution."""
```
brian-mulier-p, line 73: specify the evaluation loop interval: if false, for a duration of 2 hours with condition being satisfied during the whole 2 hours, how many time will it trigger?

core/src/main/java/io/kestra/core/validations/validator/WindowValidator.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+            return false;
+        }
+        return true;
+    }
+
+    private boolean validateSlidingWindow(Window value, ConstraintValidatorContext context) {
+        if (value.getDeadline() != null || value.getFrom() != null || value.getTo() != null || value.getOffset() != null) {
+            context.disableDefaultConstraintViolation();
+            if (value.getDeadline() != null) {
```
brian-mulier-p, line 92: you miss getEvery != null check

core/src/main/java/io/kestra/plugin/core/trigger/Flow.java
```diff
@@ -301,6 +296,23 @@ public class Flow extends AbstractTrigger implements TriggerOutput<Flow.Output>
...
+    @PluginProperty
+    private List<Dependency> dependsOn;
+
+    @Valid
+    @Schema(
+        title = "Window configuration for the dependsOn trigger",
+        description = "Configure the time window within which all dependsOn conditions must be met."
+    )
+    @PluginProperty
+    private Window window;
```
brian-mulier-p, line 312: Please add a validation in FlowValidator to make sure that if any of those is specified, the other one is also (dependsOn mandates window and window mandates dependsOn). Add a validation test accordingly

brian-mulier-p (changes requested):
Tiny changes

core/src/main/java/io/kestra/core/models/triggers/Window.java
loicmathieu: The flow trigger behavior is a bit complex so I think the whole documentation would deserve a rework as a lot of parameters depends on each other so it's not that trivial to document each independantly.
That being said, I'll leave this as is for now.

core/src/main/java/io/kestra/plugin/core/trigger/Flow.java
loicmathieu: There is a default window so you can specify dependsOn without a window.
And you can specify a window without any dependsOn, it will fire on every flow on the instance, this capability already exist.
And you can use `when`.
There is a warning in this case at flow validation.

core/src/main/java/io/kestra/core/validations/validator/WindowValidator.java
loicmathieu: Added

core/src/main/java/io/kestra/core/validations/validator/WindowValidator.java
```diff
@@ -0,0 +1,111 @@
...
+            if (value.getLookback() != null) {
+                context.buildConstraintViolationWithTemplate("Window of type `DAILY_TIME_WINDOW` cannot have a lookback duration.").addConstraintViolation();
+            }
+            return false;
+        }
+        return true;
+    }
+
+    private boolean validateSlidingWindow(Window value, ConstraintValidatorContext context) {
+        if (value.getDeadline() != null || value.getFrom() != null || value.getTo() != null || value.getOffset() != null) {
```
brian-mulier-p:
```suggestion
        if (value.getDeadline() != null || value.getFrom() != null || value.getTo() != null || value.getOffset() != null || value.getEvery() != null) {
```

brian-mulier-p (changes requested):
last change

brian-mulier-p (approved):
LGTM

PR #15505, opened 2026-04-10
feat(execution): make the VariableRender tolerant about any runtime e…

loicmathieu:
…xception

Some PebbleFunction may throws RuntimeException, even if it would be cleaner to handle all possible exceptions in all functions, this is practically impossible. To harden the executor and the scheduler, we encapsulate any RuntimeException inside the checked IllegalVariableEvaluationException.


loicmathieu:
## Breaking Change

I also removed the alternative renderer hook which should only impact the only customer we know that uses it to have a backward compat with Handlebar.

brian-mulier-p (approved):
LGTM but I don't see the runtime exception wrapping you're mentionning 🤔

PR #15509, opened 2026-04-12
fix(tests): add indexer to shared test deps

Malaydewangan09:
closes https://github.com/kestra-io/kestra-ee/issues/7394


Malaydewangan09:
We can also later add something like `@KestraTest(startIndexer = true)` to start Indexer only for tests that really need it. 

A bit of drawback coukd be that this becomes a core change and would eventually require identifying needed tests in all the plugins too, which is the per-plugin decision we want to avoid for now.

loicmathieu (approved):
LGTM

Issue #15299, opened 2026-03-30
[Filters] Update endpoint where the UI still uses legacyQuery.


ammeek:
### Description

There are several places in the UI that still use the legacy query format that have not yet been updated to use the new query filter pattern. This is causing issues with the 1.5 migration.

Entities that need to support query filter patten:
- Tenant
- Apps
- Plugins
- Audit Log

The current plan is that I will implement the backend changes needed to support thease entities then @Piyush-r-bhaskar will then implement the front end changes.

Piyush-r-bhaskar:
For which all it's already completed from BE @ammeek  ?

ammeek:
@Piyush-r-bhaskar All of the PR have now been merged for the BE change.

Piyush-r-bhaskar:
Perfect, I will take care of the UI for those all. Thanks @ammeek

PR #15513, opened 2026-04-13
fix: audit query filters

ammeek:
### ✨ Description

This PR cleans up the audit log filters and introduces a new utility for controller tests:

Removes the CONTAINS operator for resources: resources is a singular item, not a collection, making CONTAINS conceptually incorrect. In v1.3, the UI and API used CONTAINS as a workaround for an EQUALS operation. I originally retained it to preserve backward compatibility, but since these audit log filters are slated for the 2.0 release, we can safely drop this legacy behavior and enforce the correct operator.

Adds QueryFilterTestUtils: This new utility helps serialize QueryFilter objects into the exact format supported by our endpoint. This abstracts away the boilerplate and streamlines the creation of controller tests.

### 🔗 Related Issue
Relates to kestra-io/kestra#15299 

### 🛠️ Backend Checklist

### DIFF
--- a/core/src/main/java/io/kestra/core/models/QueryFilter.java
+++ b/core/src/main/java/io/kestra/core/models/QueryFilter.java
@@ -278,7 +278,7 @@ public record QueryFilter(
         RESOURCES("resources") {
             @Override
             public List<Op> supportedOp() {
-                return List.of(Op.CONTAINS, Op.IN);
+                return List.of(Op.IN);
             }
         },
         DETAILS("details") {
--- a/core/src/test/java/io/kestra/core/models/QueryFilterTest.java
+++ b/core/src/test/java/io/kestra/core/models/QueryFilterTest.java
@@ -463,6 +463,138 @@ public class QueryFilterTest {
                     Op.EQUALS,
                     Op.NOT_EQUALS
                 )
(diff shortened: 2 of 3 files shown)


brian-mulier-p:
Can you clarify the intent of the PR? I think it's the wrong PR description as this one only removes CONTAINS (why? explain concretely why you removed it)

ammeek:
> Can you clarify the intent of the PR? I think it's the wrong PR description as this one only removes CONTAINS (why? explain concretely why you removed it)

Thanks for pointing this out, @brian-mulier-p. Removing CONTAINS for resources is intentional, and I'll update the PR description to reflect that!

The reasoning is that resources is a singular item, not a collection. In v1.3, we used CONTAINS as EQUALS. I originally kept CONTAINS for backward compatibility, but since these audit log filters are targeting the 2.0 release, we can safely drop that legacy behavior now.

brian-mulier-p (changes requested):
QueryFilter tests should be added / modified accordingly (regular allowed / disallowed Op checks)

brian-mulier-p (approved):
LGTM

brian-mulier-p (approved):
LGTM

PR #15517, opened 2026-04-13
fix(ui): improve plugin autocomplete ranking

### LINKED ISSUE #15176: Auto-complete search does not surface relevant results
### Describe the issue

For example, open up the auto-complete and search "purge" 

Image

In general it would be good to improve the ranking of results. For instance, users searching for things like "if" or "kestra" (for Kestra plugin) need to scroll through many results to find the plugin they require

### Environment

- Kestra Version: 1.3

harsh21234i:
## What changed

  Improved YAML plugin autocomplete ranking for type values so more relevant matches surface earlier.

  This replaces the previous ad hoc sort bumping with deterministic segment-aware ranking for plugin class names. It now prefers:
  - last-segment prefix matches
  - exact matches in earlier segments
  - prefix matches in earlier segments
  - substring matches after that
Screenshot 2026-04-13 161847

  This makes searches like purge,if, and kestra return more relevant results near the top of the suggestion list.
  Issue #15176 reports that autocomplete search does not surface relevant results.
  The previous ordering could bury the expected plugin under many weaker matches.

  Manual:
  - started backend with ./gradlew runLocal
  - started frontend with cd ui && npm.cmd run dev
  - verified autocomplete behavior in the flow editor.

  Fixes #15176


harsh21234i:
hey @Piyush-r-bhaskar  can you look into this ?

ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
```diff
@@ -314,8 +360,7 @@ export class YamlLanguageConfigurator extends AbstractLanguageConfigurator {
                             suggestion.sortText === undefined &&
                             suggestion.label.includes(wordAtPosition)
                         ) {
-                            suggestion.sortText =
-                                sortBumperText + suggestion.label;
+                            suggestion.sortText = `08-09-999-${suggestion.label.length.toString().padStart(4, "0")}-${suggestion.label.toLowerCase()}`;
```
Piyush-r-bhaskar: Hardcoded sort key for non-dotted labels is both a DRY violation and a regression. It forces rank 08 on every non-dotted matching label, placing them below ALL dotted labels including those with zero match (rank 8 = no match).

Additionally, the comparison is case-sensitive: `suggestion.label.includes(wordAtPosition) ` compares mixed-case labels (e.g., "MyTask") against lowercase input (e.g., "my"), which will almost never match.

ui/tests/unit/services/yamlLanguageConfiguratorRanking.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+        expect(pluginSuggestionSortText(
+            "io.kestra.plugin.core.execution.PurgeExecutions",
+            "purge",
+        ) < pluginSuggestionSortText(
+            "io.kestra.plugin.core.execution.ExecutionLabels",
+            "purge",
+        )).toBe(true);
+    });
+
+    it("prioritizes exact final-segment matches for short searches like if", () => {
```
Piyush-r-bhaskar: trivial test, not meaningful
Comparing condition. `If` vs `flow.Switch` for input "if". `Switch` contains no "if" anywhere so it gets rank 8 (no match) — it's trivially rank 0 < rank 8. Should instead compare two labels that both match but at different quality ?

Piyush-r-bhaskar (changes requested):
Hey @harsh21234i 

Thank You for the PR 🎉

Few changes if you could look into before another review.

ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
```diff
@@ -28,6 +28,95 @@ import CompletionItem = languages.CompletionItem;
...
+
+    const normalizedLabel = label.toLowerCase();
+    const segments = getPluginSuggestionSegments(label);
+    const lastSegmentIndex = segments.length - 1;
+
+    let bestRank = Number.POSITIVE_INFINITY;
+    let bestDistanceFromEnd = Number.POSITIVE_INFINITY;
+    let bestPosition = Number.POSITIVE_INFINITY;
+
+    segments.forEach((segment, index) => {
```
Piyush-r-bhaskar: The loop through `segments.forEach()` continues even after finding the best match. For large `cls` with many segment, this could be slightly infficient.

harsh21234i:
Hey @Piyush-r-bhaskar   
 replaced the hardcoded non-dotted fallback sort key with a shared plain- label sort helper ,made the non-dotted matching path case-insensitive ,updated the `if` test to compare two matching labels with different
  match quality, changed the segment scan to an early-exit loop instead of `forEach`.

ui/tests/unit/services/yamlLanguageConfiguratorRanking.spec.ts
```diff
@@ -32,12 +32,12 @@ describe("YamlLanguageConfigurator ranking", () => {
         )).toBe(true);
     });
 
-    it("prioritizes exact final-segment matches for short searches like if", () => {
+    it("prioritizes exact final-segment matches over weaker final-segment matches", () => {
         expect(pluginSuggestionSortText(
             "io.kestra.plugin.core.condition.If",
```
Piyush-r-bhaskar: is this dummy ? 

io.kestra.plugin.core.condition.If
io.kestra.plugin.core.condition.IfCondition

harsh21234i: Yes, it’s a weak/dummy comparison.

  I’ll replace it with a stronger test case where both labels match the same
  input but with clearly different match quality, so the ranking assertion
  is actually meaningful.

Piyush-r-bhaskar: What you used not even existing in Kestra

io.kestra.plugin.core.condition.If
io.kestra.plugin.core.condition.IfCondition

or currently pushed  "io.kestra.plugin.if.core.Condition",


I would prefer to use `cls` what we have. 
and give me result of the test and reply from what you understand please not what AI understood.

harsh21234i: I’ve updated it to use real Kestra classes now:
  io.kestra.plugin.core.trigger.Flow and io.kestra.plugin.core.flow.If.

  For the input flow, both match, but io.kestra.plugin.core.trigger.Flow is
  an exact final-segment match, while io.kestra.plugin.core.flow.If only
  matches in an earlier segment,

harsh21234i:
Hey @Piyush-r-bhaskar  I’ve updated the test to use real Kestra classes and re-ran the focused unit test successfully.

Piyush-r-bhaskar:
Seeing all the changes together looks like just a refactoring and when I looked if issue is reproducible without your branch. It was not. @harsh21234i 

So, few question:

- What was the issue you saw exactly ? and How you tried fixing it ?
- Could you share the reproducible steps for me as well ?


See attached video from one of many environment I checked: and seems ot reproducible. am i looking correctly ?

https://github.com/user-attachments/assets/10abfd45-bcee-478f-8a8c-92ceb18f8b70

Piyush-r-bhaskar:
Closing as our QA confirmed issue non reproducibility.

PR #15523, opened 2026-04-13
fix(worker): retry once on UNAUTHENTICATED when sending to controller

fhussonnois:
When a worker's gRPC access token expires during task execution, the result-send call fails with UNAUTHENTICATED and the message is silently dropped — task results, logs, and metrics are lost.

Wrap the send with RetryOnUnauthenticatedObserver so that a single retry goes through the auth interceptor again and picks up a refreshed token. The retry delegates to the base observer, so further failures (including RESOURCE_EXHAUSTED) are still handled and no infinite retry loop can occur.


Skraye (approved):
LGTM

PR #15525, opened 2026-04-14
fix(core): streamline execution status change

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra-ee/issues/7342

**Original issue:** Changing status was crashing the instance - same issue **_also reproducible in 1.3.9. same cause but different error._**
**Cause:** Old code passed state: `this.selectedStatus` from inside the `setup()` function, where this is `nnull` — meaning the **state being sent was always null  in payload**
**Fix Summary:**  The commit moves `selectedStatus` into the `setup() `ref and uses selectedStatus.value, now we have payload populated with state --> **This needs backport to 1.3**

---

**After fixing above I found another Issue when you change Status: Gantt/Overview will not be updated until manual refresh. (Only in develop because of related changes in kestra V2)**

**Cause**: In Kestra V2 the backend made `POST /executions/{id}/change-status` and `POST /executions/{id}/state` async — they now return `ApiAsyncEvent `(`{eventId}`) instead of the updated `Execution`. The old frontend branched on `response.data.id === props.execution.id`, which was always `false` (`id is undefined`), so it skipped the wait and **rendered against stale state.**
**Fix Summary:** Added `waitForStateChange` that polls `GET /executions/{id}` until `state.histories.length` grows past the pre-command value — the deterministic signal the async event was processed. Both components now: fire the command → await waitForStateChange → emit("follow"), so the new SSE starts from the already-updated state.

### DIFF
--- a/ui/src/components/executions/ChangeExecutionStatus.vue
+++ b/ui/src/components/executions/ChangeExecutionStatus.vue
@@ -55,17 +55,14 @@
 
 <script setup lang="ts">
     import {ref, computed} from "vue";
-    import {useRouter, useRoute} from "vue-router";
     import {useI18n} from "vue-i18n";
 
     import SwapHorizontal from "vue-material-design-icons/SwapHorizontal.vue";
 
     import {State, Status} from "@kestra-io/ui-libs";
-    import * as ExecutionUtils from "../../utils/executionUtils";
     import permission from "../../models/permission";
     import action from "../../models/action";
     import {useToast} from "../../utils/toast";
-    import {useAxios} from "../../utils/axios";
 
     import {Execution, useExecutionsStore} from "../../stores/executions";
     import {useAuthStore} from "override/stores/auth";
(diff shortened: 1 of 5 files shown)


MilosPaunovic (approved):

Piyush-r-bhaskar:
@Pradumnasaraf , could you QA this please ? 

also we have same issue (null state in payload) in 1.3.9 which caused  `"Internal server error: Unable to serialize the message"` I will open separate PR for that for you to QA.

Pradumnasaraf (commented):
Hey @Piyush-r-bhaskar,

It's working now and not crashing the instance. But whenever I change the task state, it treats it as an execution restart for the whole flow (a pop-up in the execution overview, of course, executions are not restarted). Is that expected?

https://jam.dev/c/5808391e-f7f3-411f-a2f6-713e60a27680

Piyush-r-bhaskar:
probably , I will see how it behaved in 1.2 once.

Piyush-r-bhaskar:
it is expected both codewise and and as i see in earlier version.

Pradumnasaraf (approved):
Perfect then.

Looks good to merge!

PR #15529, opened 2026-04-14
fix(core): warn instead of reject on plain-text secret properties

Malaydewangan09:
closes https://github.com/kestra-io/kestra-ee/issues/6418


fdelbrayelle (approved):

anna-geller:
@Malaydewangan09 can we merge and close the issue? PR seems to be approved.

PR #15535, opened 2026-04-14
chore(flow): rename Check.condition to when

loicmathieu:
Closes https://github.com/kestra-io/kestra-ee/issues/7424#event-24476541864


fhussonnois (approved):
LGTM!

PR #15540, opened 2026-04-14
fix(core): allow disabling extended TCP keep-alive socket options in HttpClient

Skraye:


AcevedoR (approved):
LGTM, conf with a default

fhussonnois (approved):

PR #15545, opened 2026-04-14
fix(test): reduce flaky test failures in queue and runner tests

brian-mulier-p:
Companion PR to kestra-io/kestra-ee#7432

- **Dispatch queue `closingConsumer` flake**: Guard consumer callback by expected message keys so stale Kafka topic messages from other tests don't pollute the assertion (was: `[13, 14]` instead of `[1, 2]`)
- **Broadcast queue `pause` timeout**: Increase `DEFAULT_TIMEOUT_SECONDS` from 15s to 30s — Kafka consumer group rebalancing after pause/resume cycles can exceed 15s under CI load
- **Runner test timeouts**: Increase negative assertion timeout from 1s to 3s (ES deletion settling) and default await timeout from 20s to 45s (ES indexing latency under CI load)

### DIFF
--- a/core/src/test/java/io/kestra/core/runners/MultipleConditionTriggerCaseTest.java
+++ b/core/src/test/java/io/kestra/core/runners/MultipleConditionTriggerCaseTest.java
@@ -168,7 +168,7 @@ public class MultipleConditionTriggerCaseTest {
         assertThrows(
             RuntimeException.class, () -> runnerUtils.awaitFlowExecution(
                 e -> e.getState().getCurrent().equals(Type.SUCCESS),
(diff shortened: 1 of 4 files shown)


Skraye (approved):
LGTM

loicmathieu (commented):
Increasing timeouts everywhere should not be needed, I doubt if an execution is not terminated in 20s it wouild be in 45s as most of the time our tests executions last a few hundreds of milliseconds.
Doing that would increase feedback time when a test fails, and would most probably lead to CI action timeout.

Instead of changing the default, if they are test that really need a longuer timeout to await execution, it should be done at the test side not inside the TestUtils

PR #15576, opened 2026-04-15
fix: guard invalid unicode queue payloads

### LINKED ISSUE #14806: invalid unicode sequence crashes kestra executor pod

harsh21234i:
Hey @loicmathieu 
  ## Summary

  Updates the fix for invalid Unicode payload handling so malformed Unicode
  does not still destabilize the instance.

  This follows up on the reopened `#14806` report that the issue was still
  reproducible and could still kill the instance.

  ## What changed

  - guard `DefaultLogEntryEmitter` against unsupported queue payloads so
  invalid log entries are dropped instead of letting the async queue failure
  bubble out
  - broaden PostgreSQL JSONB error detection in `JdbcQueueClient` so lone
  surrogate / invalid Unicode payload failures are converted into
  `UnsupportedMessageException`
  - keep the worker-controller recovery path that strips outputs and retries
  when queue emission fails with `UnsupportedMessageException`
  - add regression coverage for:
    - unsupported log-entry emission
    - lone high surrogate payloads
    - lone low surrogate payloads

  ## Why

  The previous handling covered some invalid payload cases, but malformed
  Unicode could still escape through log emission. That left a crash path
  open even after the earlier fix.

  Notes:

  - The Postgres queue regression test updates are included, but I could not
    fully run the Postgres-backed test class in this environment because no
    local PostgreSQL instance was available.

  ## Related

  - Closes #14806


harsh21234i:
any update @loicmathieu ?

harsh21234i:
any update?

core/src/main/java/io/kestra/core/runners/DefaultLogEntryEmitter.java
loicmathieu, line 1: I'm not sure this has any benefits as it's async, it would not fail anything if the log event cannot be delivered.

I would validate a smaller change with only a log for any failures as using async emission means an error would be lost.

loicmathieu (changes requested):
You PR describe way more changes that what was actually done.

MilosPaunovic:
Hey @harsh21234i, are there any updates on this?

MilosPaunovic:
Closing due to inactivity.

@harsh21234i  if you're willing to take the comments into account and tweak the PR, we can re-open.

PR #15577, opened 2026-04-15
feat(flow): Flow trigger mode

loicmathieu:
Support Mode.ALL, Mode.ANY and Mode.AT_LEAST to trigger a flow based on all, only one or a fixed number of conditions.

### DIFF
--- a/core/src/test/java/io/kestra/core/runners/MultipleConditionTriggerCaseTest.java
+++ b/core/src/test/java/io/kestra/core/runners/MultipleConditionTriggerCaseTest.java
@@ -332,6 +332,56 @@ public class MultipleConditionTriggerCaseTest {
         ));
     }
 
+    public void flowTriggerAnyMode() throws TimeoutException, QueueException {
+        // Run only flow-a — flow-b is not run
+        Execution execution = runnerUtils.runOne(
+            MAIN_TENANT, "io.kestra.tests.trigger.any.mode",
+            "flow-trigger-any-mode-flow-a"
+        );
+        assertThat(execution.getTaskRunList().size()).isEqualTo(1);
+        assertThat(execution.getState().getCurrent()).isEqualTo(State.Type.SUCCESS);
+
+        // Trigger fires because mode is ANY and one condition (flow-a) is satisfied
+        Execution triggerExecution = runnerUtils.awaitFlowExecution(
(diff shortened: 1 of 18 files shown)


loicmathieu:
Missing features for now:

- [ ] Scoped outputs
- [ ]  onMiss SLA
- [ ]  stable condition ID (they are generated the same as for precondition)
- [ ]  silentFailures should create FAILED execution

Missing a decision on what to do for the existing precondition.

core/src/test/java/io/kestra/core/runners/MultipleConditionTriggerCaseTest.java
```diff
@@ -328,10 +328,53 @@ public void flowTriggerDependsOnFireOnceTrue() throws TimeoutException, QueueExc
...
+        assertThat(execution.getState().getCurrent()).isEqualTo(State.Type.SUCCESS);
+
+        execution = runnerUtils.runOne(
+            MAIN_TENANT, "io.kestra.tests.trigger.at.least.mode",
+            "flow-trigger-at-least-mode-flow-b"
+        );
+        assertThat(execution.getTaskRunList().size()).isEqualTo(1);
+        assertThat(execution.getState().getCurrent()).isEqualTo(State.Type.SUCCESS);
+
+        // Trigger fires because mode is AT_LEAST with minSatisfied=1 and 2 conditions are satisfied
```
brian-mulier-p:
```suggestion
        // Trigger fires because mode is AT_LEAST with minSatisfied=2 and 2 conditions are satisfied
```

Ideally we want to check that after a single flow run (flow-a) the listener is NOT YET launched but I know timings are painful as it's not deterministic to wait for something NOT to happen so... whatever

brian-mulier-p (changes requested):
You miss the condition change in FlowTriggerService (L195) to use the new mode property rather than an ALL condition for purge

core/src/test/java/io/kestra/core/runners/MultipleConditionTriggerCaseTest.java
loicmathieu: It can be done, it would not catch false positive but it may miss true failures.
I'll add that.

brian-mulier-p (approved):
LGTM

PR #15579, opened 2026-04-15
feat(execution): task output endpoint

loicmathieu:
- Change the return of the by-taskrun endpoint to directely return the map of outputs
- Add an endpoint to list the task output information without including any output content

Companion PR: https://github.com/kestra-io/kestra-ee/pull/7446


nkwiatkowski (approved):
LGTM

PR #15581, opened 2026-04-15
fix(executions): followExecution was broken because awaitility used a separate thread

AcevedoR:
we now wrap Awatility inside our Await to have a centralized configuration. Thanks to that by default Awatility use the same thread for polling.

### DIFF
--- a/core/src/main/java/io/kestra/core/utils/Await.java
+++ b/core/src/main/java/io/kestra/core/utils/Await.java
@@ -1,22 +1,36 @@
 package io.kestra.core.utils;
 
+import org.awaitility.Awaitility;
+import org.awaitility.core.ConditionFactory;
+
 import java.time.Duration;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BooleanSupplier;
 import java.util.function.Supplier;
 
-/**
- * @deprecated use {@link org.awaitility.Awaitility} instead
- */
-@Deprecated
+
 public class Await {
     private static final Duration defaultSleep = Duration.ofMillis(100);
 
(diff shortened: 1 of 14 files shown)


core/src/main/java/io/kestra/core/utils/Await.java
```diff
@@ -1,22 +1,36 @@
...
- * @deprecated use {@link org.awaitility.Awaitility} instead
- */
-@Deprecated
+
 public class Await {
     private static final Duration defaultSleep = Duration.ofMillis(100);
 
+    public static ConditionFactory await(){
+        Awaitility.pollInSameThread();
+        Awaitility.setDefaultPollDelay(defaultSleep);
```
loicmathieu: I'm not sure I understand it, would it turn this for the whole JVM?
It seems very dangerous, can't this be set on the ConditioFactrory instead?

I wonder also if we should not be smarter, I think it supports backoff poll delay so maybe configuring 10ms to 100ms poll delay (or even more) would bring some improvements.

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/utils/Await.java
```diff
@@ -1,22 +1,37 @@
...
+        return Awaitility
+            .await()
+            .pollInSameThread()
+            .pollDelay(defaultSleep);
+    }
+
+    /**
+     * @deprecated use {@link #await()} instead
+     */
+    @Deprecated
```
loicmathieu: I'll didn't deprecate it but use `await()` with the default timeout instead

Same for all methods that didn't take a timeout and duration as they are not that problematic

loicmathieu (approved):

PR #15594, opened 2026-04-16
fix(scheduler): compute nextEvaluationDate on trigger create and reset

fhussonnois:
onTriggerCreated and onResetTrigger produced trigger states with a null nextEvaluationDate. The fetcher then computed a future cron tick and immediately evaluated it, but Schedule.evaluate() rejected it via its future-guard while processSchedulableTrigger already advanced nextEvaluationDate one tick further — silently skipping the first scheduled execution.

Compute nextEvaluationDate at the source in both event handlers, matching the existing pattern in TriggerScheduler.onStart().

Closes: kestra-io/kestra-ee#7410


Pradumnasaraf (commented):
Hey, the issue is still there. 

Picking the schedule from another flow if the trigger name is the same (I think this is the root cause of all issues). Plus, the schedule is not getting updated correctly. Tested with a fresh instance!

Jam: https://jam.dev/c/fda7e165-fd76-4295-8d2c-e5f22a5e70be

Pradumnasaraf (approved):
Looks good to merge!

It was a confusion on my side. No changes needed.

PR #15595, opened 2026-04-16
chore(system): replace DSL.trueCondition by DSL.noCondition

loicmathieu:
jOOQ will then generates a query with no WHERE clause or no AND clause instead of WHERE true or WHERE 1=1


loicmathieu:
All tests failures are on unrelated areas and pass locally

AcevedoR (approved):

PR #15598, opened 2026-04-16
fix: add PARENT_ID query filter field to openapi spec

ammeek:
### ✨ Description

The openapi spec on development is currently out of date and does not include PARENT_ID query filter.


loicmathieu:
Sorry about that, thanks for fixing it.

PR #15600, opened 2026-04-16
feat(flow): Flow trigger outputs should be scoped to the flow

loicmathieu:


brian-mulier-p (approved):
LGTM

brian-mulier-p (approved):
LGTM, don't forget to flag it as a BC

loicmathieu:
Oh yes, thanks, added the label.
Any the issue is also flagged as a BC and we might deprecate/remove some part of the existing Flow trigger so it will definitly be listed as a BC.

PR #15601, opened 2026-04-16
feat(core): automate all migration scripts

nkwiatkowski:
- fix #https://github.com/kestra-io/kestra-ee/issues/1672


cli/src/test/java/io/kestra/cli/commands/configs/sys/NoConfigCommandTest.java
```diff
@@ -29,51 +24,18 @@ void shouldSucceedWithNamespaceKVCommandWithoutParamsAndConfig() {
...
             String[] args = {};
             Integer call = PicocliRunner.call(KvCommand.class, ctx, args);
 
             assertThat(call).isZero();
             assertThat(out.toString()).contains("Usage: kestra namespace kv");
         }
     }
-
-    @Test
-    void shouldFailWithDeleteFlowCommandWithoutConfig() throws URISyntaxException {
```
loicmathieu, line 42: Why are you deleting this test? It shoudl still pass

jdbc-h2/src/main/java/io/kestra/repository/h2/migration/H2MigrationLock.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * No-op {@link MigrationLock} for H2.
+ *
+ * <p>H2 is an embedded, single-process database, so distributed locking is not needed.
+ * <p>Active only when H2 is the <em>repository</em> backend, not just the queue, to avoid
+ * conflicting with the Elasticsearch repository backend when H2 is used only as the queue.
+ */
+@Singleton
+@Requires(property = "kestra.repository.type", pattern = "h2|memory")
+public class H2MigrationLock implements MigrationLock {
```
loicmathieu, line 19: It would be safier to use an in-memory Lock like the one provided by the JDK.
I woud suggest using `java.util.concurrent.locks.ReentrantLock`

jdbc-h2/src/main/resources/migrations/baseline-queue-h2.sql
```diff
@@ -0,0 +1,14 @@
+-- H2 queue baseline schema (Queue 2.0)
+-- Creates the JDBC queue table used by the memory/H2 queue backend.
+
+CREATE TABLE IF NOT EXISTS queues (
```
loicmathieu, line 5: I think it miss the drop of the table before re-creation

jdbc-h2/src/main/resources/migrations/baseline-h2.sql
```diff
@@ -0,0 +1,381 @@
...
+
+CREATE INDEX IF NOT EXISTS ix_last_deleted_tenant_namespace_path_version ON namespace_file_metadata ("last", "deleted", "tenant_id", "namespace", "path", "version");
+CREATE INDEX IF NOT EXISTS ix_last_deleted_tenant_namespace_path ON namespace_file_metadata ("last", "deleted", "tenant_id", "namespace", "path");
+CREATE INDEX IF NOT EXISTS ix_last_deleted_tenant_namespace_parent_path ON namespace_file_metadata ("last", "deleted", "tenant_id", "namespace", "parent_path");
+CREATE INDEX IF NOT EXISTS ix_last_deleted_tenant_namespace_version ON namespace_file_metadata ("last", "deleted", "tenant_id", "namespace", "version");
+CREATE INDEX IF NOT EXISTS ix_last_deleted_tenant_path_version ON namespace_file_metadata ("last", "deleted", "tenant_id", "path", "version");
+
+
+/* ----------------------- locks ----------------------- */
+CREATE TABLE IF NOT EXISTS locks (
```
loicmathieu: I don't understand why you have locks and task_outputs which are new in 2.0 inside both the baseline and the V2 scripts

jdbc-h2/src/main/java/io/kestra/repository/h2/migration/V2_0UpgradeMigration.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public String description() {
+        return "OSS H2 upgrade: apply Kestra 2.0 schema changes on Flyway-managed databases";
+    }
+
+    @Override
+    public String checksum() {
+        return CHECKSUM;
```
loicmathieu, line 53: Shoulnd't it be a real checksum computed by the system?

jdbc-mysql/src/main/java/io/kestra/repository/mysql/migration/MysqlMigrationLock.java
```diff
@@ -0,0 +1,88 @@
...
+ * <p>The lock is session-scoped and persists across transactions, making it safe for
+ * multi-node deployments. The lock times out after 5 minutes if not acquired.
+ */
+@Slf4j
+@Singleton
+@MysqlRepositoryEnabled
+public class MysqlMigrationLock implements MigrationLock {
+
+    private static final String LOCK_NAME = "kestra_migration";
+    private static final int LOCK_TIMEOUT_SECONDS = 300;
```
loicmathieu, line 29: Some migration can take hours or days, using such low timeout would be problematic

jdbc-postgres/src/main/java/io/kestra/repository/postgres/migration/PostgresMigrationHistoryStore.java
```diff
@@ -0,0 +1,82 @@
...
+        // Quote the identifier to handle case-sensitive or special-character schema names
+        String quoted = schemaToCreate.replace("\"", "\"\"");
+        stmt.execute("CREATE SCHEMA IF NOT EXISTS \"" + quoted + "\"");
+    }
+
+    /**
+     * Parses the PostgreSQL {@code search_path} and returns the first non-system schema name,
+     * or {@code null} if none is found.
+     */
+    static String resolveSchemaFromSearchPath(final Statement stmt) throws SQLException {
```
loicmathieu, line 62: I'm surprise we have to do that, we don't care of the schema anywhere else

loicmathieu (changes requested):
You delete all existing Java migration script but you didn't integrate them inside the automatic migration script.

You also remove the V2TriggerMigrationCommand without replacement!

The whole purpose of this mechanism was to avoid having migrations as commands but this didn't seems to be integrated yet.

cli/src/test/java/io/kestra/cli/commands/configs/sys/NoConfigCommandTest.java
nkwiatkowski: This test doesn't really make sens, it is passing, but not for the same error now. We are testing that without configuration you can't access your backend.

jdbc-h2/src/main/resources/migrations/baseline-h2.sql
nkwiatkowski: I can make it it only the second script, I wanted to have the init script to give a full 2.0 ready database to have it in one place, but I can leave it only in the second script.

jdbc-postgres/src/main/java/io/kestra/repository/postgres/migration/PostgresMigrationHistoryStore.java
nkwiatkowski: We need it here because the schema used to be created by flyway. Now that it's not the case anymore, we need to handle the schema creation if we come from a fresh database.

cli/src/main/java/io/kestra/cli/commands/migrations/V2ExecutionOutputMigrationCommand.java
loicmathieu, line 1: Sorry to didn't tell you before but this is no longuer needed as the migration will be done "on the fly" with a compatibility layer.
Please remove this migration command.

cli/src/main/java/io/kestra/cli/commands/migrations/V2ExecutionResubmitMigrationCommand.java
```diff
@@ -0,0 +1,77 @@
...
+                {
+                    // save outputs in the new table
+                    ListUtils.emptyOnNull(execution.getTaskRunList()).forEach(taskRun ->
+                    {
+                        try {
+                            taskOutputService.saveOutputs(taskRun, taskRun.getOutputs());
+                        } catch (InternalException e) {
+                            Log.error(e, "Failed to save outputs for execution: " + execution.getId() + " task run: " + taskRun.getId());
+                        }
+                    });
```
loicmathieu: Remove this part, we now have a compatibility layer so it's unnecessary

cli/src/test/java/io/kestra/cli/commands/configs/sys/NoConfigCommandTest.java
```diff
@@ -29,51 +24,18 @@ void shouldSucceedWithNamespaceKVCommandWithoutParamsAndConfig() {
         ByteArrayOutputStream out = new ByteArrayOutputStream();
         System.setOut(new PrintStream(out));
 
-        try (ApplicationContext ctx = ApplicationContext.builder().deduceEnvironment(false).start()) {
+        try (ApplicationContext ctx = ApplicationContext.builder()
+            .deduceEnvironment(false)
+            .properties(Map.of(
+                "kestra.repository.type", "memory",
+                "kestra.queue.type", "memory"
+            ))
```
loicmathieu: Here you configure a backend, defeating the test

cli/src/test/java/io/kestra/cli/commands/configs/sys/NoConfigCommandTest.java
loicmathieu, line 42: This is problematic to remove this test as we want to be sure it works when it needs, and fail when it needs

cli/src/main/java/io/kestra/cli/commands/migrations/UnlockMigrationCommand.java
loicmathieu, line 1: Can you add a test for it?

cli/src/main/java/io/kestra/cli/commands/migrations/V2ExecutionResubmitMigrationCommand.java
loicmathieu, line 1: Can you add a test for it?

jdbc-h2/src/main/java/io/kestra/repository/h2/migration/H2MigrationLock.java
```diff
@@ -0,0 +1,65 @@
...
+    public H2MigrationLock(
+        @Property(name = "kestra.migration.lock-acquire-timeout", defaultValue = "PT1H") final Duration lockTimeout) {
+        this.lockTimeout = lockTimeout;
+    }
+
+    @Override
+    public void acquire() throws Exception {
+        if (!lock.tryLock(lockTimeout.toMillis(), TimeUnit.MILLISECONDS)) {
+            throw new IllegalStateException(
+                "Could not acquire H2 migration lock within " + lockTimeout
```
loicmathieu:
```suggestion
                "Could not acquire H2 migration lock within " + lockTimeout + " you can increase this value by setting `kestra.migration.lock-acquire-timeout`"
```

jdbc-h2/src/main/java/io/kestra/repository/h2/migration/V2_0Migration.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public String description() {
+        return "OSS H2 init: create full schema from scratch";
+    }
+
+    @Override
+    public String checksum() {
+        return CHECKSUM;
```
loicmathieu: I would not only compute the checksum() when called to avoid it to be created eaglery.
Normally, the checksum method is called only once so it's safe to compute it on demand.

jdbc-mysql/src/main/java/io/kestra/repository/mysql/migration/MysqlMigrationLock.java
```diff
@@ -0,0 +1,131 @@
...
+        log.debug("Acquiring MySQL migration lock '{}'", LOCK_NAME);
+        lockConnection = dataSource.getConnection();
+        try (PreparedStatement ps = lockConnection.prepareStatement("SELECT GET_LOCK(?, ?)")) {
+            ps.setString(1, LOCK_NAME);
+            ps.setInt(2, (int) lockTimeout.toSeconds());
+            try (ResultSet rs = ps.executeQuery()) {
+                if (!rs.next() || rs.getInt(1) != 1) {
+                    lockConnection.close();
+                    lockConnection = null;
+                    throw new SQLException(
```
loicmathieu: It's not an SQLException!
I think you created a dedicated exception for that right?

jdbc-mysql/src/main/java/io/kestra/repository/mysql/migration/MysqlMigrationLock.java
loicmathieu, line 1: I'm a little concern about this connection sharing in a volatile attribute.
Is it really needed?
If so, I think it should be safer to guard each methods that access the shared connections via synchronized methods or blocks.

jdbc-mysql/src/main/java/io/kestra/repository/mysql/migration/V2_0Migration.java
```diff
@@ -0,0 +1,53 @@
...
+ * Creates the full Kestra OSS schema from scratch on fresh MySQL installations.
+ * For databases already migrated by Flyway (Kestra &le; 1.3), this script is skipped
+ * automatically by {@link io.kestra.core.migration.MigrationRunner} (schema already exists).
+ */
+@Singleton
+@MysqlRepositoryEnabled
+public class V2_0Migration extends AbstractSQLMigrationScript {
+
+    private static final String SCRIPT_ID = "0-init";
+    private static final String CHECKSUM = MigrationScript.checksumOfResources("/migrations/baseline-mysql.sql");
```
loicmathieu: Same remark here for checksum (apply to all migration scripts)

jdbc-postgres/src/main/java/io/kestra/repository/postgres/migration/PostgresMigrationLock.java
loicmathieu, line 1: Same remark as for MySQL, if sharing the connection is really needed better to protect it via synchronized methods or blocks

jdbc-postgres/src/main/java/io/kestra/repository/postgres/migration/V2_0Migration.java
```diff
@@ -0,0 +1,53 @@
...
+ * Creates the full Kestra OSS schema from scratch on fresh PostgreSQL installations.
+ * For databases already migrated by Flyway (Kestra &le; 1.3), this script is skipped
+ * automatically by {@link io.kestra.core.migration.MigrationRunner} (schema already exists).
+ */
+@Singleton
+@PostgresRepositoryEnabled
+public class V2_0Migration extends AbstractSQLMigrationScript {
+
+    private static final String SCRIPT_ID = "0-init";
+    private static final String CHECKSUM = MigrationScript.checksumOfResources("/migrations/baseline-postgres.sql");
```
loicmathieu: Same remark for checksum computation

jdbc/src/main/java/io/kestra/jdbc/migration/AbstractSQLMigrationScript.java
```diff
@@ -0,0 +1,222 @@
...
+     *
+     * @param dataSource   the data source to obtain a connection from
+     * @param resourcePath classpath resource path to the SQL file (e.g.
+     *                     {@code "/migrations/baseline-h2.sql"})
+     * @throws IOException  if the resource cannot be read
+     * @throws SQLException if a statement fails to execute
+     */
+    public static void executeSqlScript(final DataSource dataSource, final String resourcePath)
+        throws IOException, SQLException {
+        try (InputStream is = AbstractSQLMigrationScript.class.getResourceAsStream(resourcePath)) {
```
loicmathieu: I wonder if using the Thread current classloader may not be better: Thread.currentThread().getContextClassLoader().

This  is usually safer, especially when using frameworks like Micronaut which may have special classloaders

jdbc/src/main/java/io/kestra/jdbc/migration/V2_0TriggerMigration.java
loicmathieu, line 1: This is exactly what I envision when I discussed this with you: mid-level using jOOQ, use internal records for temporary structure, straighforward 👍

loicmathieu (changes requested):
Overall, it's very good, so what's requested are details.

tchiotludo:

No EE spec drift (EE branch: `feat/automate_migration_scripts`).

loicmathieu (approved):
LGTM

PR #15606, opened 2026-04-16
feat(flow): stable condition ID inside the Flow trigger dependsOn

loicmathieu:

### DIFF
--- a/core/src/main/java/io/kestra/core/utils/Hashing.java
+++ b/core/src/main/java/io/kestra/core/utils/Hashing.java
@@ -5,6 +5,7 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
 import com.google.common.hash.HashCode;
+import org.apache.commons.lang3.StringUtils;
 
 /**
  * Utilities for hashing.
@@ -20,6 +21,18 @@ public final class Hashing {
      */
     public static String hashToString(final String value) {
         return getHashString(value).toString();
+    }
+
+    /**
+     * Returns a consistent hash value for the given input strings, using
+     * a non-cryptographic hash function.
+     * Each string will be concatenated before hashing.
+     *
+     * @param values the values to be hashed.
+     * @return the string hash value.
+     */
+    public static String hashToString(final String... values) {
+        return getHashString(StringUtils.join(values)).toString();
     }
 
     /**
--- a/core/src/main/java/io/kestra/plugin/core/trigger/Flow.java
+++ b/core/src/main/java/io/kestra/plugin/core/trigger/Flow.java
@@ -7,6 +7,7 @@ import java.util.stream.Stream;
 
 import io.kestra.core.models.property.Property;
 import io.kestra.core.models.triggers.Window;
+import io.kestra.core.utils.*;
 import org.apache.commons.lang3.stream.Streams;
 import org.slf4j.Logger;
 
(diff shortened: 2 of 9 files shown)


core/src/main/java/io/kestra/core/utils/Hashing.java
```diff
@@ -22,6 +23,18 @@ public static String hashToString(final String value) {
...
+    /**
+     * Returns a consistent hash value for the given input strings, using
+     * a non-cryptographic hash function.
+     * Each string will be concatenated before hashing.
+     *
+     * @param values the values to be hashed.
+     * @return the string hash value.
+     */
+    public static String hashToString(final String... values) {
+        return getHashString(StringUtils.join(values)).toString();
```
brian-mulier-p, line 35: You should add a separator between the values to avoid my.namespace|flow to have the same hash than my.namespac|eflow (ofc terrible example but you get the thing)

core/src/main/java/io/kestra/plugin/core/trigger/Flow.java
```diff
@@ -540,6 +537,16 @@ public Mode getMode() {
...
             return minSatisfied;
         }
+
+        private String hash(Dependency dependency) {
+            return Hashing.hashToString(
+                dependency.namespace,
+                dependency.flowId,
+                dependency.when != null ? dependency.when.toString() : null,
+                ListUtils.emptyOnNull(dependency.states).stream().map(Enum::name).collect(Collectors.joining(",")),
+                MapUtils.emptyOnNull(dependency.labels).entrySet().stream().map(entry -> entry.getKey() + ":" + entry.getValue()).collect(Collectors.joining(","))
```
brian-mulier-p: I would sort the lists to avoid shifts due to serialization or similar tbh but as you judge on that one

core/src/test/java/io/kestra/core/runners/FlowTriggerCaseTest.java
```diff
@@ -94,6 +108,68 @@ public void triggerWithPause() throws TimeoutException, QueueException {
...
+            tasks:
+              - id: only
+                type: io.kestra.plugin.core.debug.Return
+                format: "It works"
+            """;
+        GenericFlow invertedListenerFlow = GenericFlow.fromYaml(MAIN_TENANT, invertedSource);
+        flowService.update(invertedListenerFlow, currentListenerFlow);
+
+        // the flow metastore is updated async so we wait a little
+        Thread.sleep(250);
```
brian-mulier-p: can't you assert the metastore in awaitility rather than arbitrary wait (that involves potentially extra waiting time that accumulates + is not deterministic in case of CI load)

brian-mulier-p (changes requested):

core/src/main/java/io/kestra/plugin/core/trigger/Flow.java
loicmathieu: Yes, true, if I didn't sort by label key, this would not be stable

core/src/test/java/io/kestra/core/runners/FlowTriggerCaseTest.java
loicmathieu: Done

core/src/main/java/io/kestra/core/utils/Hashing.java
loicmathieu: Would be rather rare but OK

brian-mulier-p (approved):
LGTM

PR #15607, opened 2026-04-16
fix(worker): stop heartbeat after terminal state to fix flaky liveness coordinator test

loicmathieu:
## Summary

- **Root cause**: `ServiceLivenessManager` (heartbeat) continuously refreshed `updatedAt` in the database even after a worker transitioned to a terminal state (`TERMINATED_FORCED`, `TERMINATED_GRACEFULLY`). Because `isTerminationGracePeriodElapsed` uses a strict `isBefore` comparison, this prevented the `ServiceLivenessCoordinator` from ever detecting the elapsed grace period — causing `AbstractServiceLivenessCoordinatorTest.shouldResubmitTaskWhenWorkerIsStopped` to be flaky.
- Set `isStateUpdatable = false` **before** persisting terminal state transitions in `ServiceLivenessManager`, so the heartbeat scheduler immediately stops refreshing `updatedAt` once a worker has terminated.
- Extract `CREATED` event handling to bypass the `isStateUpdatable` guard — a new worker of the same `ServiceType` can start after a previous one terminated, and the stale `false` flag must not block its registration. (`ServiceRegistry` is keyed by `ServiceType`, not service ID.)
- Document the shared-`AtomicBoolean` invariant on `LocalServiceState.with()`, which the fix relies on: callers holding a pre-update `holder` still affect the newly registered instance because the same `AtomicBoolean` instance is threaded through `with()`.

### DIFF
--- a/core/src/main/java/io/kestra/core/server/LocalServiceState.java
+++ b/core/src/main/java/io/kestra/core/server/LocalServiceState.java
@@ -19,10 +19,14 @@ public record LocalServiceState(Service service,
     }
 
     /**
-     * Convenient method for constructing a new {@link LocalServiceState} from a given instance.
+     * Constructs a new {@link LocalServiceState} from a given instance, preserving the same
+     * {@link AtomicBoolean} reference for {@code isStateUpdatable}. This means that any caller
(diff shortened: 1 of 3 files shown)


fhussonnois (approved):
LGTM! This seems logical

PR #15608, opened 2026-04-16
Fix working directory cache deletion

### LINKED ISSUE #15408: PurgeCurrentExecutionFiles breaks Task Caching
### Describe the issue

If you use PurgeCurrentExecutionFiles with Task Caching it fails as the file doesn't exist but assumes it will.

https://github.com/user-attachments/assets/d1e48b63-40d1-463b-bdcd-e86e8c76cb79

To reproduce, use this flow:

```yaml
id: caching
namespace: company.team

tasks:
  - id: transactions
    uri: https://huggingface.co/datasets/kestra/datasets/resolve/main/csv/cache_demo/transactions.csv

  - id: products
    uri: https://huggingface.co/datasets/kestra/datasets/resolve/main/csv/cache_demo/products.csv
    taskCache:
      enabled: true
      ttl: PT24H

  - id: duckdb
    type: io.kestra.plugin.jdbc.duckdb.Query
    store: true
    inputFiles:
      products.csv: "{{ outputs.products.uri }}"
      transactions.csv: "{{ outputs.transactions.uri }}"
    sql: |-
      SELECT
        t.transaction_id,
        t.timestamp,
        t.quantity,
        t.sale_price,
        p.product_name,
        p.category,
        p.cost_price,
        p.supplier_id,
        (t.sale_price - p.cost_price) * t.quantity AS profit
      FROM
        read_csv_auto('transactions.csv') AS t
      JOIN
        read_csv_auto('products.csv') AS p
      USING (product_id);

  - id: purge
    type: io.kestra.plugin.core.storage.PurgeCurrentExecutionFiles
```

### Environment

- Kestra Version: develop

shivamwayal37:
### ✨ Description

Fixes an issue where cached files (via taskCache) are not recreated during execution, but cleanup logic still attempts to delete them, causing failures.

This PR ensures missing files are handled safely during purge and adds tests to cover this scenario.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/15408

### 🛠️ Backend Checklist

### 📝 Additional Notes

The fix takes a safe approach by ensuring deletion logic does not fail when expected files are absent (e.g., due to caching). A unit test has been added to prevent regressions.


shivamwayal37:
Hi,
Just a gentle follow-up on this PR whenever you have time.

This fixes an issue where cached files may not exist during purge, causing deletion failures. The fix ensures safe handling of missing files and includes tests to prevent regressions.

Would really appreciate your feedback. Thanks!

loicmathieu (changes requested):
I think the correct fix should be to purge any cache files when purging execution files.

I don't know if it's easily doable, but it should be better and more robust than analysing any existing cache file for an `uri` key and backoff if missing, this is very fragile as the internal storage URI may be in a key with a different name (and deep nested inside a Map which may be a bit counterperformant to check)

shivamwayal37:
Thanks for the suggestion, that makes sense.

You're right that checking for missing files is fragile and depends on how URIs are stored. I'll update the implementation to instead purge related cache files when execution files are deleted, which should be more robust.

I'll push an update shortly.

shivamwayal37:
Thanks for the feedback — updated the implementation.

- Removed worker-side workaround
- Moved cleanup to InternalStorage.deleteExecutionFiles()
- Cache files now store executionId as metadata
- Purge deletes cache by metadata (no URI parsing or archive inspection)

Updated tests to verify execution and cache files are both removed.

Let me know if you’d prefer a different approach for linking cache to executions.

worker/src/main/java/io/kestra/worker/processors/WorkerTaskProcessor.java
loicmathieu, line 1: Can you rollback the changes on this file?

core/src/main/java/io/kestra/core/storages/InternalStorage.java
```diff
@@ -227,13 +232,47 @@ private Optional<Long> getCacheFileLastModifiedTime(String cacheId, @Nullable St
         return isFileExist(uri) ? Optional.of(this.storage.getAttributes(context.getTenantId(), context.getNamespace(), uri).getLastModifiedTime()) : Optional.empty();
     }
 
+    private List<URI> deleteExecutionCacheFiles() throws IOException {
+        if (context.getExecutionId() == null) {
+            return List.of();
+        }
+
+        List<URI> deletedCacheFiles = new ArrayList<>();
+        List<URI> flowFiles = this.storage.allByPrefix(context.getTenantId(), context.getNamespace(), context.getFlowStorageURI(), false);
```
loicmathieu, line 238: This would effectively browse all the files related to this namespace which may be a lot of file!
I think the best would be to change the cache mechanism to include the executionId so we know which file to delete.

loicmathieu (changes requested):

shivamwayal37:
Thanks, that makes sense.

One architectural question before I implement it:

`getCacheFile(hash, taskRunValue, ttl)` currently resolves caches independently of execution, which allows cross-execution cache reuse.

If we move cache storage to an execution-scoped path (for example including `executionId` in the cache URI), then cache lookup would no longer find entries created by previous executions unless we either:

* scan execution-scoped prefixes, or
* maintain a separate shared lookup/index path.

Should the new implementation preserve cross-execution cache reuse semantics, or is execution-scoped caching acceptable here?

loicmathieu:
@shivamwayal37 you're right, I think here we follow the wrong path, sorry, we should not remove a cache related to an execution, it would defeat the purpose of the cache.

So, back to the original issue I'm not sure anymore why it fails as PurgeExecutionFiles should not have removed the cache file.

So now, I would say, it require more investiguation, sorry about that.

shivamwayal37:
No worries, thanks for taking another look.

I’ll investigate the original failure path more deeply and try to identify exactly why the cached file reference becomes invalid during purge. I’ll update the PR once I have a clearer root cause.

MilosPaunovic:
Hey @shivamwayal37, are there any updates on this?

shivamwayal37:
Thanks for the follow-up.

I revisited the earlier approaches and agree that deleting cache entries during execution purge is not the correct fix since it would break cross-execution cache reuse.

At the moment, I’m still investigating the original failure path to better understand why the purge flow attempts to delete files that are no longer present when task caching is involved.

I don’t have a reliable root cause yet, so I’d rather avoid pushing another speculative fix. My next step is to isolate the exact sequence that produces the stale file reference and reduce it to a minimal reproducible case.

I’ll continue investigating and update the PR once I have clearer findings. Thanks again for the patience and guidance.

shivamwayal37:
@loicmathieu, @MilosPaunovic

I dug deeper into the failure path and found what appears to be the root cause.

The issue is not that `PurgeCurrentExecutionFiles` deletes cache files. Rather, task output caching can persist execution-scoped `kestra://.../executions/{executionId}/...` URIs.

A task may produce an output URI pointing to a file stored under a specific execution. That output is serialized into the task cache and can later be restored for a different execution. If the original execution files have been purged, the restored output contains references to files that no longer exist.

In other words, the cache lifetime (flow/task scoped) can outlive the execution-scoped files referenced by the cached outputs.

Because of that, deleting cache files during execution purge does not seem like the correct fix, as it would break cross-execution cache reuse semantics.

I'm currently evaluating possible fixes, including:

* invalidating cache entries when restored outputs contain missing execution-scoped URIs; or
* rematerializing/excluding execution-scoped file references during cache restore.

Before proceeding, does this align with your understanding of the issue?

loicmathieu:
> invalidating cache entries when restored outputs contain missing execution-scoped URIs; or

IMHO, this is the only sane thing to do as if a file is missing the cache could not be reused.

Maybe this is the kind of things that should not be fixed, if you want to cache outputs, but delete the file underneath this output via a purge task, you made a mistake...

shivamwayal37:
Thanks, that makes sense.

Just to confirm the expected behavior: if a cached output contains a `kestra://` URI and the referenced file no longer exists, should the cache be considered invalid and the task re-executed?

Or should this be treated as an unsupported use case and documented instead?

I want to make sure I'm implementing the intended behavior before proceeding.

loicmathieu:
> if a cached output contains a kestra:// URI and the referenced file no longer exists, should the cache be considered invalid and the task re-executed?

I think we should ignore that, we don't want to check each possible cache entry for existing URL and check each of them: it would be too costly.

If someone purge more aggressively than a configured cache TTL, this is not something we should handle, it's a design issue not a bug inside Kestra.

shivamwayal37:
Thanks for confirming. That makes sense.

Based on your feedback, it sounds like this behavior is expected rather than a bug in Kestra. In that case I'm happy to close this PR since none of the proposed fixes preserve the intended cache semantics.

Thanks for the discussion and the guidance throughout the investigation.

MilosPaunovic:
@loicmathieu Should this PR be closed or is there something else to explore?

MilosPaunovic:
@loicmathieu, another ping here, wondering if this PR should be closed or is there something else to explore?

Issue #13648, opened 2025-12-12
[Plugin Scripts] Mark the `beforeCommands` logs as `DEBUG`


caxefaizan:
### Feature description

Currently, beforeCommands are logged at **INFO** level. In practice, these commands usually succeed but tend to produce a large volume of log output due to installations, upgrades, and other setup steps.

This negatively impacts the user experience as a flow having several script tasks with some before commands generate huge logs and to find the meaningful information they have to scroll a lot.

It would be nice to mark them as `DEBUG` so the log filter could be kept on `INFO` by default to avoid having to look at them.
If an error occurs, it should still be logged at ERROR level, ensuring failures remain clearly visible.

saifxyzyz:
does this need triage? I'd like to work on this

loicmathieu:
It's not easy as beforeCommands and commands are contatenated into a single command in the end, so we have no way to know when receiving a log that it's emitted during the beforeCommands or commands set of commands...

saifxyzyz:
oh I might have misunderstood, I thought I'd just change log.info to log.debug haha

fdelbrayelle:
@caxefaizan It's merged! Does it need a backport?

caxefaizan:
not really. thank you

fdelbrayelle:
@caxefaizan Allright, so this will be a part of the 2.0.0 by beginning of June then! Thanks!

PR #15613, opened 2026-04-17
fix(scripts): log beforeCommands output at DEBUG level

fdelbrayelle:
## Summary

- Injects `echo '##kestra:log:debug##'` / `echo '##kestra:log:info##'` sentinel lines around `beforeCommands` in the merged shell script (`CommandsWrapper`)
- `DefaultLogConsumer` detects these markers, flips a `debugMode` flag, and silently swallows the marker lines (not counted, not logged)
- `PluginUtilsService.parseOut` gains a `debug` overload: when `debug=true` and stdout, logs at `DEBUG` instead of `INFO`; stderr always stays at `ERROR` for visibility

This preserves shell state sharing between `beforeCommands` and the main `commands` (single script execution) while reducing log noise for setup steps like `pip install -r requirements.txt`.

## Test plan

- [ ] `DefaultLogConsumerTest` — 3 new unit tests: markers are swallowed (not counted), regular lines are counted, stderr always counted even in debug mode
- [ ] Manual: run a Shell or Python task with `beforeCommands: ["pip install requests"]` — setup output should disappear at INFO level and reappear at DEBUG
- [ ] Confirm stderr from a failing `beforeCommands` still surfaces at ERROR level

**What to verify:**
1. `pip install requests` and `apt-get install` output → **DEBUG** level (hidden at default log level, visible when log level set to DEBUG)
2. Main `commands` output (`Status: 200`, `main command`) → **INFO** level (visible by default)
3. `beforeCommands` stderr → **ERROR** level (always visible regardless of log level)

Part-of: https://github.com/kestra-io/kestra/issues/13648

### DIFF
--- a/core/src/main/java/io/kestra/core/models/tasks/runners/DefaultLogConsumer.java
+++ b/core/src/main/java/io/kestra/core/models/tasks/runners/DefaultLogConsumer.java
@@ -8,7 +8,11 @@ import io.kestra.core.runners.RunContext;
  * Default implementation of an @{link {@link AbstractLogConsumer}}
  */
 public class DefaultLogConsumer extends AbstractLogConsumer {
+    private static final String LOG_DEBUG_MARKER = "##kestra:log:debug##";
+    private static final String LOG_INFO_MARKER = "##kestra:log:info##";
+
     private final RunContext runContext;
+    private volatile boolean debugMode = false;
 
     public DefaultLogConsumer(RunContext runContext) {
         this.runContext = runContext;
@@ -20,7 +24,15 @@ public class DefaultLogConsumer extends AbstractLogConsumer {
     }
 
     public void accept(String line, Boolean isStdErr, Instant instant) {
-        outputs.putAll(PluginUtilsService.parseOut(line, runContext.logger(), runContext, isStdErr, instant));
+        if (LOG_DEBUG_MARKER.equals(line)) {
+            debugMode = true;
+            return;
+        }
+        if (LOG_INFO_MARKER.equals(line)) {
+            debugMode = false;
+            return;
+        }
+        outputs.putAll(PluginUtilsService.parseOut(line, runContext.logger(), runContext, isStdErr, instant, debugMode));
 
         if (isStdErr) {
             this.stdErrCount.incrementAndGet();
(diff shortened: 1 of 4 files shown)


core/src/test/java/io/kestra/core/models/tasks/runners/DefaultLogConsumerTest.java
```diff
@@ -0,0 +1,57 @@
...
+    @Test
+    void stderrLinesAreCountedEvenInDebugMode() {
+        var runContext = runContextFactory.of("id", "namespace", IdUtils.create());
+        var consumer = new DefaultLogConsumer(runContext);
+
+        consumer.accept("##kestra:log:debug##", false);
+        consumer.accept("stderr during beforeCommands", true);
+
+        assertThat(consumer.getStdErrCount()).isEqualTo(1);
+        assertThat(consumer.getStdOutCount()).isZero();
```
Malaydewangan09, line 60: Cn we add a test, which could verify lines are actually logged at DEBUG level?

fdelbrayelle: Done

core/src/main/java/io/kestra/core/models/tasks/runners/DefaultLogConsumer.java
```diff
@@ -8,7 +8,11 @@
  * Default implementation of an @{link {@link AbstractLogConsumer}}
  */
 public class DefaultLogConsumer extends AbstractLogConsumer {
+    private static final String LOG_DEBUG_MARKER = "##kestra:log:debug##";
```
Malaydewangan09, line 11: If a user prints `kestra:log:debug` from their task, all subsequent stdout logs silently drop to DEBUG?

fdelbrayelle: Er yes it should... But well who will do this if not on purpose?

fdelbrayelle:
## QA

Flow:

```yaml
id: test_before_commands_log_level
namespace: company.team

tasks:
  - id: python_before_commands
    beforeCommands:
      - pip install requests
    script: |
      import requests
      r = requests.get("https://httpbin.org/get")
      print(f"Status: {r.status_code}")

  - id: shell_before_commands
    beforeCommands:
      - echo "setup step — should be DEBUG"
      - apt-get install -y curl 2>/dev/null || true
    commands:
      - echo "main command — should be INFO"
      - curl -s https://httpbin.org/get | head -5

  - id: stderr_always_error
    beforeCommands:
      - echo "this goes to stdout — should be DEBUG"
      - echo "this goes to stderr — should be ERROR" >&2
    commands:
      - echo "done"
```

Results with INFO level:

image

Results with DEBUG level:

image

image

Malaydewangan09 (approved):

PR #15623, opened 2026-04-19
feat(triggers): add trigger catalog with Add/Manage tabs

anna-geller:
Splits the tenant-level Triggers page into two tabs:
- **Add** (browse trigger plugins and add them to a flow) and
- **Manage** (the existing management table).

Closes https://github.com/kestra-io/kestra-ee/issues/7181

It will need to add https://github.com/kestra-io/kestra/issues/15040 later to the Core triggers

Demo https://share.descript.com/view/jBEGSZzAH8S

## What changes

- New `GET /api/v1/plugins/triggers` endpoint feeds the Add tab. It walks every `RegisteredPlugin`, filters internal and deprecated classes, and buckets each trigger into **Core**, **Realtime**, or **App**:
  - **Core**: bundled with Kestra (`getExternalPlugin() == null`) plus classes living in `io.kestra.plugin.core.*` and `io.kestra.core.*` even when they ship in a separate EE plugin jar (so `AssetEvent` lands under Core with the EE badge).
  - **Realtime**: external plugins implementing `RealtimeTriggerInterface`.
  - **App**: everything else (typically `PollingTriggerInterface`).
- EE detection combines the plugin manifest's `X-Kestra-License` attribute with a package-name fallback (`io.kestra.ee.*`, `io.kestra.plugin.ee.*`, or any package containing `.ee.`) since uber-jars strip per-module manifests.
- **Add tab**: catalog cards per trigger, grouped by category, with hover tooltips rendering the full `@Schema` description as markdown.
- **Configure modal** (opens on card click): collects namespace, flow, trigger ID, and shows an editable YAML **Source** tab plus a plugin **Documentation** tab. On submit it stashes a draft in a Pinia store and redirects to the flow editor, which splices the trigger into the target flow via `@kestra-io/ui-libs/flow-yaml-utils`.
- **Manage tab**: unchanged behavior.

## What this does not cover

- No in-modal property editing. Filling in plugin-specific properties still happens in the flow editor after the redirect. A schema-driven form inside the modal would need the no-code renderer (`Task.vue` and its ~15 injection keys) wired to a synthetic flow context, which is a larger project + can create confusion about when trigger editing should take place (Flow editor is for it).
- No drag-and-drop from the catalog into an open flow editor.
- Trigger deprecation and alias handling follow the existing plugin rules; no changes there.


Piyush-r-bhaskar:
Though the BE test passed, some Claude suggestions on Backend changes: Could you validate these @ammeek and also see if we need any EE equivalent changes for new endpoint ?

- **[CRITICAL]** - `webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java:152-169` -  listTriggerPlugins returns a raw `List<TriggerPluginDto>`.  - APIs should not return a response being a JSON array which cannot be evolved in a backwards- compatible way." Every future enhancement (counts, filter metadata, EE hints) will need a new endpoint. Wrap in a response object: `record TriggerPluginListResponse(List<TriggerPluginDto> triggers) {}`.

- **[CRITICAL]** No tests added for new endpoint.

- **[CRITICAL]** `webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java:152` - No authorization annotation on the new endpoint.

- **[MAJOR]** `webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java:204-207`
EE detection via `packageName.contains(".ee.")` matches any segment literally named "ee",
which is a substring match across package boundaries. It works today but is a time bomb
for any plugin that happens to have a nested namespace like `...fee.internal` (would match)
or `...teepot.` (would not, but shows the fragility). Use a segment-level check:
`Arrays.stream(packageName.split("\\.")).anyMatch("ee"::equals)` — and add a test matrix.

Piyush-r-bhaskar (approved):
Looks good now FE code wise 🎉

Nice initiative and work Anna 🚀

webserver/src/main/java/io/kestra/webserver/controllers/api/TriggerPluginDto.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @param type fully qualified class name (for example {@code io.kestra.plugin.core.trigger.Schedule})
+ * @param name human-readable name (Schema#title if set, otherwise simple class name)
+ * @param description one-line description from the plugin @Schema
+ * @param group category bucket ({@code core}, {@code realtime}, or {@code app})
+ * @param ee true when the trigger is only available in Enterprise Edition (bundled with EE core, or shipped by a plugin distributed under an Enterprise license)
+ * @param icon icon key resolvable via {@code GET /api/v1/plugins/icons}
+ * @param deprecated whether the trigger is deprecated
+ */
+public record TriggerPluginDto(
```
ammeek, line 14: The only place we use the dto is in the PluginController, we should probably move it to be a inner record.
It would also be good if we update it's name to follow the convention already set for our api dto objects.
For example: ApiAuditLogIte, ApiPluginArtifact

webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java
```diff
@@ -138,6 +139,74 @@ public List<Plugin> listPlugins() {
...
+        description = "Feeds the 'Add Trigger' catalog UI. Returns one entry per non-internal, non-deprecated " +
+            "trigger class, classified as core (bundled with Kestra Core), realtime (implements " +
+            "RealtimeTriggerInterface) or app (implements PollingTriggerInterface). Optionally filter by " +
+            "group. The 'ee' flag is true when the trigger ships with an Enterprise Edition module."
+    )
+    public List<TriggerPluginDto> listTriggerPlugins(
+        @Parameter(description = "Optional category filter: core, realtime, or app")
+        @Nullable @QueryValue(value = "group") TriggerPluginCategory group,
+        @Parameter(description = "Include deprecated triggers")
+        @Nullable @QueryValue(value = "includeDeprecated", defaultValue = "false") Boolean includeDeprecated
```
ammeek, line 156: Ideally we would not return a json array in the root of our response. My suggestion here is to add a page and size property here so that the responses are paged and match the pattern we already have defined in our api then used the PagedResults.

webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java
```diff
@@ -138,6 +139,74 @@ public List<Plugin> listPlugins() {
...
+    ) {
+        return pluginRegistry.plugins().stream()
+            .flatMap(registeredPlugin -> registeredPlugin.getTriggers().stream()
+                .filter(c -> !io.kestra.core.models.Plugin.isInternal(c))
+                .filter(c -> !c.getName().startsWith("org.kestra."))
+                .filter(c -> Boolean.TRUE.equals(includeDeprecated) || !io.kestra.core.models.Plugin.isDeprecated(c))
+                .map(c -> toTriggerPluginDto(registeredPlugin, c))
```
ammeek, line 163: We are using the io.kestra.core.models.Plugin class multiple times in this controller so we should probably import it. Same for the other instances of classes that use the full path ref in this controller multiple times.

ammeek (changes requested):
Few small changes

ammeek:
> * **[CRITICAL]** - `webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java:152-169` -  listTriggerPlugins returns a raw `List<TriggerPluginDto>`.  - APIs should not return a response being a JSON array which cannot be evolved in a backwards- compatible way." Every future enhancement (counts, filter metadata, EE hints) will need a new endpoint. Wrap in a response object: `record TriggerPluginListResponse(List<TriggerPluginDto> triggers) {}`.

Ideally we would not return a json array in the root of our response. My suggestion here is to add a page and size property here so that the responses are paged and match the pattern we already have defined in our api then used the PagedResults.

> * **[CRITICAL]** No tests added for new endpoint.

New tests would be Ideally here as we have introduced new functionality to our api which our application now relies on.

> * **[CRITICAL]** `webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java:152` - No authorization annotation on the new endpoint.

The PluginController falls under the "/api/v1/**: path so the AuthenticationFilter will handle the auth without needing a annotation.

> * **[MAJOR]** `webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java:204-207`
>   EE detection via `packageName.contains(".ee.")` matches any segment literally named "ee",
>   which is a substring match across package boundaries. It works today but is a time bomb
>   for any plugin that happens to have a nested namespace like `...fee.internal` (would match)
>   or `...teepot.` (would not, but shows the fragility). Use a segment-level check:
>   `Arrays.stream(packageName.split("\\.")).anyMatch("ee"::equals)` — and add a test matrix.

I don't think this is valid, java contains does not use regex for it's contains it just check if a string contains another sub string exactly.

Piyush-r-bhaskar (approved):

Piyush-r-bhaskar:
@ammeek I have commited requested changes in BE using `claude` , Can you review and make any changes still needed on backend and approve this PR , so that @anna-geller  can merge it.

tchiotludo:

```diff
1731a1732,1750
>   /api/v1/plugins/triggers:
>     get:
>       tags:
>       - Plugins
>       summary: Get list of trigger plugins grouped by category
>       description: "Feeds the 'Add Trigger' catalog UI. Returns one entry per non-internal,\
>         \ non-deprecated trigger class, classified as core (bundled with Kestra Core),\
>         \ realtime (implements RealtimeTriggerInterface) or app (implements PollingTriggerInterface)."
>       operationId: listTriggerPlugins
>       responses:
>         "200":
>           description: listTriggerPlugins 200 response
>           content:
>             application/json:
>               schema:
>                 $ref: "#/components/schemas/PagedResults_PluginController.ApiTriggerPlugin_"
>       security:
>       - bearerAuth: []
>       - basicAuth: []
21346a21366,21378
>     PagedResults_PluginController.ApiTriggerPlugin_:
>       required:
>       - results
>       - total
>       type: object
>       properties:
>         results:
>           type: array
>           items:
>             $ref: "#/components/schemas/PluginController.ApiTriggerPlugin"
>         total:
>           type: integer
>           format: int64
21671a21704,21720
>     PluginController.ApiTriggerPlugin:
>       type: object
>       properties:
>         type:
>           type: string
>         name:
>           type: string
>         description:
>           type: string
>         group:
>           $ref: "#/components/schemas/TriggerPluginCategory"
>         ee:
>           type: boolean
>         icon:
>           type: string
>         deprecated:
>           type: boolean
23256a23306,23313
>     TriggerPluginCategory:
>       type: string
>       enum:
>       - CORE
>       - REALTIME
>       - APP
>       - UNKNOWN
>       x-type: String
```

ammeek:
Almost everything on the backend looks good to me. I've removed the filtering logic from the backend endpoint and the front and client code as it was never being used, also if we want to add filtering logic to the backend we should probably follow the new query filter pattern.

I've also added a sanity check test and stopped the backend from mapping the schema title to be the plugin name as the schema tilte is a description rather than a name.

ammeek (approved):

Piyush-r-bhaskar:
Good to merge @anna-geller 🚀

anna-geller:
@Piyush-r-bhaskar sorry I haven't seen your comment, could you resolve conflicts and merge? 🙏

PR #15632, opened 2026-04-20
feat(jdbc): support delete/purge/purgebycondition inside AbstractJdbcCrudRepository for central use

### LINKED ISSUE #13001: Support delete/purge inside the AbstractJdbcCrudRepository
Today, this repository didn't support delete/deleteByQuery operations (soft delete) and purge (hard delete).

What's challenging is that not all entities support soft delete so some implements delete/deleteByQuery via hard delete.

shivanshsharma13:
### ✨ Description

This PR adds 3 new methods to `AbstractJdbcCrudRepository` to provide a unified deletion mechanism across all JDBC repositories.

- **`delete(T item)`** - Soft-deletes the entity if it implements `SoftDeletable` (persists with `deleted = true`), otherwise hard-deletes it. Returns the deleted item in both cases. Child repositories that need existence validation or `CrudEvent` publishing can override this method.

- **`purge(T item)`** - Always hard-deletes the entity regardless of `SoftDeletable`. Returns the number of deleted rows. Child repositories that need `CrudEvent` publishing can override this method.

- **`purgeByCondition(DSLContext context, Condition condition)`** - Bulk hard-delete by condition. Centralizes the raw `context.delete().where().execute()` pattern currently repeated across multiple repositories.

**Existing overrides (unaffected):**
- `AbstractJdbcExecutionRepository.delete()` overrides base `delete()` - adds existence validation via `findById` and fires `CrudEvent.delete()`
- `AbstractJdbcExecutionRepository.purge()` overrides base `purge()` - additionally fires `CrudEvent.delete()`
- `AbstractJdbcFlowRepository.deleteWithoutAcl()` remains unchanged - has its own flow-specific deletion logic

**Repositories not yet using the new methods:**
- `AbstractJdbcSettingRepository`, `AbstractJdbcServiceInstanceRepository`, and `AbstractJdbcLockRepository` still call `this.jdbcRepository.delete()` directly instead of the new base `delete()` method
- `AbstractJdbcLogRepository`, `AbstractJdbcMetricRepository`, and `AbstractJdbcFlowTopologyRepository` still use raw `context.delete().where().execute()` directly instead of the new `purgeByCondition()` (new method)

Closes https://github.com/kestra-io/kestra/issues/13001

### 🛠️ Backend Checklist

- [X] Code compiles successfully and passes all checks
- [X] All unit and integration tests pass

### 📝 Additional Notes

I left the `purgeByCondition` implementation unchanged in this PR to keep the scope focused. Migrating them to use the new base methods can be done incrementally in follow-up PRs without any risk of breaking existing behavior (Or will do in here if needed now!)

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
@@ -145,28 +145,44 @@ public abstract class AbstractJdbcCrudRepository<T> extends AbstractJdbcReposito
     }
 
     /**
-     * Hard-delete an item from the database
-     * Use this for physical purge operations.
+     * Hard-delete an item from the database.
+     * Use this for physical purge operations on a single item.
      *
      * @param item the item to purge.
+     * @return {@code true} if the item was deleted, {@code false} if it was not found.
      */
-    public Integer purge(T item) {
-        return this.jdbcRepository.delete(item);
+    public boolean purge(T item) {
+        return this.jdbcRepository.delete(item) > 0;
     }
 
     /**
-     * Hard-delete all items matching the given condition within a transaction context.
-     * Use this for bulk purge operations.
+     * Hard-delete all items matching the given condition for the given tenant.
      *
-     * @param context   the DSL context (transaction-aware).
+     * @param tenantId  the tenant ID used to build the default filter.
      * @param condition the condition to match items for deletion.
      * @return the number of deleted rows.
      */
-    protected Integer purgeByCondition(DSLContext context, Condition condition) {
-        return context
-            .delete(this.jdbcRepository.getTable())
-            .where(condition)
-            .execute();
+    protected Integer purge(String tenantId, Condition condition) {
+        return purge(defaultFilter(tenantId), condition);
+    }
+
+    /**
+     * Hard-delete all items matching the given conditions.
+     *
+     * @param defaultFilter the base filter condition (e.g. tenant isolation).
+     * @param condition     the condition to match items for deletion.
+     * @return the number of deleted rows.
+     */
+    protected Integer purge(Condition defaultFilter, Condition condition) {
+        return this.jdbcRepository
+            .getDslContextWrapper()
+            .transactionResult(configuration ->
+                DSL.using(configuration)
+                    .delete(this.jdbcRepository.getTable())
+                    .where(defaultFilter)
+                    .and(condition)
+                    .execute()
+            );
     }
 
     /**
--- a/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
@@ -24,7 +24,7 @@ import reactor.core.publisher.FluxSink;
  * If the child repository uses a default filter, it should override it.
  * <p>
  * For example, to avoid supporting allowDeleted:
- * 
+ *
  * <pre>
  * {@code
  * &#64;Override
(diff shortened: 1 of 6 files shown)


shivanshsharma13:
Hi @loicmathieu, just a gentle ping on PR whenever you get a chance. No rush, appreciate your time!

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
```diff
@@ -124,6 +124,51 @@ public void deleteWithoutAcl(T item) {
...
+        return item;
+    }
+
+    /**
+     * Hard-delete an item from the database
+     * Use this for physical purge operations.
+     *
+     * @param item the item to purge.
+     */
+    public Integer purge(T item) {
```
loicmathieu:
```suggestion
    public boolean purge(T item) {
```

As you purge a single item it's beter to return a boolean that indicate that it do purge the item

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
```diff
@@ -124,6 +124,51 @@ public void deleteWithoutAcl(T item) {
...
+
+    /**
+     * Hard-delete all items matching the given condition within a transaction context.
+     * Use this for bulk purge operations.
+     *
+     * @param context   the DSL context (transaction-aware).
+     * @param condition the condition to match items for deletion.
+     * @return the number of deleted rows.
+     */
+    protected Integer purgeByCondition(DSLContext context, Condition condition) {
```
loicmathieu: This is not how I would do it:
- You should not pass the DSLContext but use the `jdbcRepository` to open a transaction.
- You should have two variantes: one that takes a tenantId and a condition and build the defaultFilter()
- One that accepts two conditions: the first being the defaultFilter

And you can just name them purge().

loicmathieu (changes requested):
It would also be great to find possible places where those new methods could be used (execution and logs for ex)

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
shivanshsharma13: Yes, Agreed!

`purge(T item)` — return type changed from `Integer` to `boolean`. Since this purges a single known item, a boolean is the correct semantic. Implementation: return `this.jdbcRepository.delete(item) > 0`

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcCrudRepository.java
shivanshsharma13: - `purge(Execution execution)` - uses the tenantId variant so the defaultFilter(tenantId) is applied.
- `purge(List<Execution> executions)` - no tenant guard since the list can span multiple tenants, the `execution_id` IN condition is the only predicate needed

shivanshsharma13:
Acknowledged the comments please review!

loicmathieu (approved):
LGMT, thanks.

I need to do some adaptation in our EE code so I'll merge it after that

loicmathieu:
@shivanshsharma13 I have no idea why but some tests fails after pour changes. Can you have a look please?

shivanshsharma13:
Ok, sure let me check

shivanshsharma13:
**Done fixes for build failures!**

- **Fixed `purge(Execution)` in `AbstractJdbcMetricRepository` to use `DSL.noCondition()`** - same as the log repository fix. Using `execution.getTenantId()` as the filter causes a silent no-op when the tenant is not set (resolves to `WHERE tenant_id IS NULL`, matching nothing). Since execution IDs are globally unique, no tenant scoping is needed here.

- **Added explicit `delete` override in `AbstractJdbcKvMetadataRepository`** - adding `delete(T)` to `AbstractJdbcCrudRepository` shadows the `KvMetadataRepositoryInterface` default method without this override, deletion skips `saveKvMetadata()` and calls `persist` directly with a null version, causing a `NOT NULL` constraint violation. The override routes back through `save(item.toDeleted())` so versioning is handled correctly.

PR #15633, opened 2026-04-20
fix(core): prevent ClassCastException in decodeInput for already-decrypted secret inputs

Skraye:
## Summary
- `RunVariables.decodeInput()` blindly casts input values to `Map`, causing a `ClassCastException` when a `SecretInput` value is already a plain `String` (e.g. during subflow execution end)
- Use `instanceof` checks instead of unchecked casts for both flat and nested input paths
- Add null guard on the `inputs` parameter for the recursive nested path

Seems to happen for execution that were paused and the Kestra instance get an upgrade


loicmathieu (approved):
LGTM

PR #15646, opened 2026-04-21
fix(tests): replace sampletestfile.com with embedded server

Skraye:
- `sampletestfile.com` is down, breaking `DownloadTest.run()` and `RequestTest.head()`
- Replaced external HTTP calls with the Micronaut embedded server already used by other tests in the same classes
- Tests are now self-contained and won't break when external sites go down


brian-mulier-p (approved):
LGTM

PR #15648, opened 2026-04-21
fix(tests): reuse injected ApplicationContext in RequestTest

nkwiatkowski:
## Summary
- Replace 23 `ApplicationContext.run()` calls in `RequestTest` with the injected `ApplicationContext` from `@KestraTest`
- Each `ApplicationContext.run()` bootstraps a full Micronaut context (scanning all beans, initializing `@Context` beans like `MigrationRunner`), which can deadlock with the outer test context competing for migration locks and DB connections
- SSL tests (`selfSigned`, `selfSignedFailed`) keep a dedicated context since they need the `testssl` environment config
- Follows the same pattern already used by `DownloadTest`

## Test plan
- [x] `RequestTest` — all 28 tests pass (25s)
- [x] `DownloadTest` — no regressions
- [x] Compilation clean

### DIFF
--- a/core/src/test/java/io/kestra/plugin/core/http/DownloadTest.java
+++ b/core/src/test/java/io/kestra/plugin/core/http/DownloadTest.java
@@ -205,25 +205,21 @@ class DownloadTest {
 
     @Test
     void failed() throws Exception {
-        try (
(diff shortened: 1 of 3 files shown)


Skraye (approved):
LGTM

PR #15649, opened 2026-04-21
Fix/realtime trigger resubmit

loicmathieu:
Fixes https://github.com/kestra-io/kestra-ee/issues/7326

### DIFF
--- a/executor/src/main/java/io/kestra/executor/DefaultServiceLivenessCoordinator.java
+++ b/executor/src/main/java/io/kestra/executor/DefaultServiceLivenessCoordinator.java
@@ -6,6 +6,7 @@ import java.util.Optional;
 import java.util.Random;
 import java.util.Set;
 
+import io.kestra.core.models.triggers.RealtimeTriggerInterface;
 import org.slf4j.event.Level;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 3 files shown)


executor/src/main/java/io/kestra/executor/DefaultServiceLivenessCoordinator.java
```diff
@@ -155,18 +155,15 @@ protected void handleAllWorkersForUncleanShutdown(Instant now) {
                 return;
             }
 
-            // List of workers for which we don't know the actual state of tasks executions.
-            // Re-emit all WorkerJobs for unclean workers
             boolean isUncleanShutdownService = isUncleanShutdownService(serviceInstance, now);
-            if (isUncleanShutdownService) {
+            // Transit GRACEFUL or UNCLEAN SHUTDOWN worker to NOT_RUNNING.
+            if (isUncleanShutdownService || serviceInstance.is(Service.ServiceState.TERMINATED_GRACEFULLY)) {
```
fhussonnois, line 160: I don't think it's safe to call the `reEmitWorkerJobsForWorker` when `TERMINATED_GRACEFULLY` is detected, because in a rare cases we could have a risk to resubmit a few tasks if some entries are not removed yet from the WorkerJJobRunning table. 

I would prefer to have a distinct method: `reEmitRealTimeTriggerForWorker` called on: `TERMINATED_GRACEFULLY`

fhussonnois (changes requested):

executor/src/main/java/io/kestra/executor/DefaultServiceLivenessCoordinator.java
loicmathieu: > resubmit a few tasks if some entries are not removed yet from the WorkerJJobRunning table.

Your concern is about tasks that were terminated but not yet removed from the table because everything is async right?
We have no idea whether a trigger is a realtime trigger or not so re-emitting only for them is not that simple.

fhussonnois (approved):

PR #15650, opened 2026-04-21
fix(system): possible NoSuchElement exception after migratoin to 1.2

### LINKED ISSUE #15620: Stuck executions after upgrade from v1.0.1 to v1.2.3 and manual attempts patch causing queue growth
## Content:
We are currently using Kestra v1.0.1 and planning to upgrade to v1.2.3.

Due to business constraints, there are some executions that cannot be completed before the upgrade.
Therefore, in our test environment, we stopped Kestra, performed the upgrade, and then restarted the service.

After the upgrade, we encountered the following issue:

## Problem
For executions with the following structure:
Flow-A (RUNNING) → Flow-B (RUNNING) → Flow-C (PAUSED)

After restart, when attempting to Resume Flow-C:

Some executions do not contain the attempts field in the taskRunList
This results in errors or even infinite retry/loop behavior

## Workaround (Data Patch)
To address this, we implemented a manual data patch while Kestra was stopped:

Target executions:
- State = RUNNING or PAUSED
- taskRunList entries missing attempts

Patch logic:
- For RUNNING: copy state.current into a newly created attempts structure
- For PAUSED: construct attempts similarly, but normalize state as RUNNING

Not handled:
- flowable/runnable lifecycle consistency
- queues table updates

## Result
After applying the patch:

The infinite loop / resume error no longer occurs
However, we observed a significant increase in messages in the queues table

We are unsure whether this queue growth is related to our patch or expected behavior after recovery.

## Questions
- Is this kind of manual patching approach valid or risky?
- Is constructing attempts from state.current a correct assumption?
- Could this patch cause abnormal queue growth?
- Is there any official or recommended way to fix such stuck executions after upgrade?

Any guidance or best practices would be greatly appreciated.

loicmathieu:
As there was previously no attemps on flowable tasks

Fixes https://github.com/kestra-io/kestra/issues/15620

### DIFF
--- a/executor/src/main/java/io/kestra/executor/ExecutorService.java
+++ b/executor/src/main/java/io/kestra/executor/ExecutorService.java
@@ -279,9 +279,11 @@ public class ExecutorService {
                     List<TaskRunAttempt> attempts = Optional.ofNullable(parentTaskRun.getAttempts())
                         .map(ArrayList::new)
                         .orElseGet(ArrayList::new);
-                    State.Type endedState = endedTask.get().getTaskRun().getState().getCurrent();
-                    TaskRunAttempt updated = attempts.getLast().withState(endedState);
-                    attempts.set(attempts.size() - 1, updated);
+                    if (!attempts.isEmpty()) { // can occur on migration from pre-1.2
+                        State.Type endedState = endedTask.get().getTaskRun().getState().getCurrent();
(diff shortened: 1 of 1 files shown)


nkwiatkowski (approved):
LGTM

loicmathieu:
Backported to 1.3

PR #15654, opened 2026-04-21
chore(system): only remove worker job running for worker tasks

loicmathieu:
The method where it was, were also used for flowable tasks which didn't have any worker_job_running.

Moving the removal inside the WorkerTaskResultMessageHandler avoid sending DELETE request that would never delete anything to the database.

### DIFF
--- a/executor/src/main/java/io/kestra/executor/ExecutorService.java
+++ b/executor/src/main/java/io/kestra/executor/ExecutorService.java
@@ -62,9 +62,6 @@ public class ExecutorService {
     private WorkerGroupMetaStore workerGroupMetaStore;
 
     @Inject
-    private WorkerJobRunningStateStore workerJobRunningStateStore;
-
-    @Inject
     protected FlowMetaStoreInterface flowExecutorInterface;
 
     @Inject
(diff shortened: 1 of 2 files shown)


executor/src/main/java/io/kestra/executor/handler/WorkerTaskResultMessageHandler.java
```diff
@@ -38,6 +42,11 @@ public Optional<ExecutorContext> handle(WorkerTaskResult message) {
             ExecutorContext current = new ExecutorContext(execution);
 
             if (execution.hasTaskRunJoinable(message.getTaskRun())) {
+                // remove the worker job running
+                if (message.getTaskRun().getState().isTerminated()) {
+                    workerJobRunningStateStore.deleteByKey(message.getTaskRun().getId());
```
fhussonnois: I'm wondering if for task this should not be done in the `GrpcWorkerControllerService.java` (i.e., the Controller) like for triggers. So that, the `WorkerJobRunningStateStore` will be the sole responsability of the Controller - already responsible for doing the `save` - and the Liveness Coordinator (responsible for re-submit).

fhussonnois: Could be done here: https://github.com/kestra-io/kestra/blob/develop/worker-controller/src/main/java/io/kestra/controller/grpc/services/GrpcWorkerControllerService.java#L176

loicmathieu: Agree, I'll move it

loicmathieu: Done

fhussonnois (approved):

PR #15659, opened 2026-04-21
fix(worker): route null or blank workerGroup.key to the default worker group

fhussonnois:
Empty or null keys now short-circuit to the default worker-group.

Related-to: kestra-io/kestra-ee#6629


nkwiatkowski (approved):
LGTM

PR #15682, opened 2026-04-22
feat: add MCP server management, McpTool trigger, and AI agent integration UI

ammeek:
### ✨ Description

Implements the full MCP (Model Context Protocol) feature for Kestra OSS.
A new core trigger that exposes a flow as a named MCP tool:

* `toolName` — tool identifier (required)
* `title` — human-readable name shown to AI agents (required)
* `toolDescription` — what the tool does (optional)
* `mcpServer` — which server to register on, defaults to `"default"`
* `annotations` — tool behaviour hints with spec-defined defaults:
    * `readOnly: false`
    * `openWorld: false`
    * `destructive: true`
    * `idempotent: false`

Flow inputs/outputs are automatically converted to a JSON schema tool spec. Executions created via MCP carry a `system.from:mcp` label for observability.

#### MCP Server entity

Tenant-scoped configuration object (`McpServer`) with fields: `id`, `name`, `description`, `systemPrompt`, `serverType` (PRIVATE/PUBLIC), `authType` (BASIC/API_TOKEN), `enabled`, `isDefault`. Stored with soft-delete support across H2, PostgreSQL, and MySQL via new Flyway migrations.

A default server is auto-provisioned per tenant on startup with a curated system prompt that instructs AI agents to invoke flows only when appropriate.

### Walkthrough

  1. Create an MCP server                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                   
  Navigate to Tenant → MCP Servers in the sidebar. A default server is already provisioned for every tenant — you can use it as-is or create a dedicated one. Click + Create Server, give it a name (e.g. data-team), an optional description and  
  system prompt, choose Private (the default, requires authentication) or Public, and pick Basic Auth as the auth type. Hit Create Server and the card appears in the grid.
                                                                                                                                                                                                                                                   
  2. Expose a flow as a tool                                                                                                                                                                                                                       
   
  Add an McpToolTrigger to any flow you want to expose:                                                                                                                                                                                            
                                                            
  triggers:
    - id: mcp
      type: io.kestra.plugin.core.trigger.McpToolTrigger
      toolName: hello_world                                                                                                                                                                                                                        
      title: Hello World greeting tool
      toolDescription: Returns a personalised greeting. Call this when the user asks for a greeting.                                                                                                                                               
      mcpServer: data-team                                  

  The trigger's toolName is what the AI agent sees. The toolDescription is what it uses to decide whether to call the tool — write it as if you're telling the AI when and why to use it. Flow inputs and outputs are automatically converted into 
  a JSON schema, so the AI knows exactly what parameters to pass.
                                                                                                                                                                                                                                                   
  3. Connect a client                                       

  Back on the MCP Servers page, click Connect on the data-team card. The modal shows the SSE endpoint URL and ready-to-paste configuration for each client:                                                                                        
   
  - Claude Desktop — copy the JSON block into claude_desktop_config.json (path shown in the modal for macOS and Windows)                                                                                                                           
  - Claude Code — copy and run the claude mcp add command in your terminal
  - Cursor — paste the server URL into Cursor Settings → MCP → Add new MCP server                                                                                                                                                                  
   
  4. Call the tool                                                                                                                                                                                                                                 
                                                            
  Once connected, the AI agent discovers the tool list from the SSE endpoint. Ask Claude (or whichever client you connected) to do something that maps to your flow — e.g. "Say hello to Alice" — and it will invoke hello_world with user=Alice as
   the input. A Kestra execution is created, tagged with system.from:mcp and system.mcp-server-id so you can filter it in the Executions view. The flow runs, and the output is streamed back to the AI agent and returned to the user.


cli/src/main/java/io/kestra/cli/services/DefaultStartupHook.java
```diff
@@ -25,10 +27,14 @@ public class DefaultStartupHook implements StartupHookInterface {
...
     private Optional<SettingRepositoryInterface> settingRepositoryProvider;
 
+    @Inject
+    Optional<McpServerService> mcpServerService;
+
     @Override
     public void start(AbstractCommand cmd) {
         if (cmd instanceof ServerCommandInterface && !(cmd instanceof WorkerCommand)) {
             saveKestraVersion();
+            ensureDefaultMcpServer();
```
fhussonnois: EE startup hook calls this default start-up hook and the MAIN tenant is only relevant for OSS.

core/src/main/java/io/kestra/core/mcp/models/McpServer.java
```diff
@@ -0,0 +1,124 @@
...
+ * Represents an MCP (Model Context Protocol) server configuration.
+ */
+public record McpServer(
+    @Hidden
+    @Pattern(regexp = "^[a-z0-9][a-z0-9_-]*")
+    String tenantId,
+
+    @NotNull
+    @NotBlank
+    String name,
```
fhussonnois: In Kestra, we use the `id` property as a unique identifier, just as we do in Flow, Dashboard, Apps, etc. The `name` property serves more as a "display name."

core/src/main/java/io/kestra/core/mcp/services/McpServerService.java
```diff
@@ -0,0 +1,43 @@
...
+public class McpServerService {
+
+    private final McpServerRepositoryInterface mcpServerRepository;
+
+    @Inject
+    public McpServerService(McpServerRepositoryInterface mcpServerRepository) {
+        this.mcpServerRepository = mcpServerRepository;
+    }
+
+    public void ensureDefaultMcpServer(final String tenantId) {
```
fhussonnois: rename this method `createDefaultMcpServerIfNotExist`

core/src/main/java/io/kestra/core/mcp/services/McpServerService.java
```diff
@@ -0,0 +1,43 @@
...
+        if (mcpServerRepository.get(tenantId, McpServer.DEFAULT_NAME).isPresent()) {
```
fhussonnois: Please, add an `exists` method to the repository - no need to select and deserialize the full row from the database for checking existance

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcMcpSessionRepository.java
```diff
@@ -0,0 +1,137 @@
...
+    public Optional<McpSession> find(String tenantId, String serverId, String sessionId) {
+        return jdbcRepository
+            .getDslContextWrapper()
+            .transactionResult(configuration -> {
+                DSLContext context = DSL.using(configuration);
+                Record record = context
+                    .select(VALUE_FIELD)
+                    .from(jdbcRepository.getTable())
+                    .where(this.defaultFilter(tenantId))
+                    .and(field("server_id", String.class).eq(serverId))
```
fhussonnois: create static fields for column (e.g., `field("server_id", String.class)`)

ui/src/components/EeBadge.vue
```diff
@@ -0,0 +1,23 @@
+<template>
```
fhussonnois: use the existing lock icon for EE feature do not introduce a new one.

webserver/src/main/java/io/kestra/mcp/McpConfig.java
```diff
@@ -0,0 +1,22 @@
...
+import io.micronaut.context.annotation.ConfigurationProperties;
+import lombok.Data;
+import lombok.Getter;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+
+@Data
+@ConfigurationProperties("kestra.mcp")
+public class McpConfig {
```
fhussonnois: Use a `record` for Config objet

webserver/src/main/java/io/kestra/mcp/McpSessionService.java
```diff
@@ -0,0 +1,162 @@
...
+    static final Duration SESSION_PURGE_AGE = Duration.ofHours(48);
+
+    @Inject
+    public McpSessionService(McpSessionRepositoryInterface sessionRepository, BroadcastQueueInterface<McpSession> mcpSessionQueue) {
+        this.sessionRepository = sessionRepository;
+        this.mcpSessionQueue = mcpSessionQueue;
+    }
+
+    @PostConstruct
+    public void startNotificationConsumer() {
```
fhussonnois: Just use the `start()` / `stop()` naming convention for your methods—there’s no need for an extra `NotificationConsumer()`—so this will align with the naming conventions we already use in other classes.

webserver/src/main/java/io/kestra/mcp/McpServerHandlerTransport.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private McpAsyncServer buildServer(
+        HandlerKey handlerKey,
+        KestraFluxStreamableServerTransportProvider serverTransport
+    ) {
+        var mcpServerSpec = io.modelcontextprotocol.server.McpServer.async(serverTransport)
+            .capabilities(
+                McpSchema.ServerCapabilities.builder()
+                    // List changes is set as false as we manually handle the update
+                    .tools(false)
```
fhussonnois: The value should be `true`.  Any server that calls  notifyToolsListChanged() must advertise listChanged: true.

core/src/main/java/io/kestra/core/mcp/models/McpSession.java
```diff
@@ -0,0 +1,36 @@
...
+import io.kestra.core.models.HasUID;
+import io.kestra.core.queues.event.BroadcastEvent;
+import io.kestra.core.utils.IdUtils;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record McpSession(
+    String tenantId,
+    String serverId,
+    String sessionId,
+    String sseNode,
```
fhussonnois, line 14: If I understand correctly, do `serverId` and `sseNode` refer to the same thing? I see you compare sseNode to `ServerInstance.INSTANCE_ID`

queue-jdbc/src/main/java/io/kestra/queue/jdbc/JdbcQueueFactory.java
```diff
@@ -162,6 +168,24 @@ public DispatchQueueInterface<WorkerTaskResult> workerTaskResultQueue(JdbcDepend
...
+    public BroadcastQueueInterface<McpSession> mcpSessionQueue(JdbcDependencies dependencies) {
+        return new JdbcBroadcastQueue<>(
+            McpSession.class, dependencies.queueService(), dependencies.jdbcQueueClient(), dependencies.executorsUtils(), dependencies.metricRegistry(),
+            dependencies.ignoreExecutionService()
+        );
+    }
+
+    @QueueBean
+    @Override
+    public BroadcastQueueInterface<McpServer> mcpQueue(JdbcDependencies dependencies) {
```
fhussonnois: I think you should create a dedicated `McpSessionEvent` interface  to carry, if needed, the McpSession. But, we should have clear domain event. From my understanding you have 3 kind of event: `CREATED`, `OWNERSHIP_CHANGED`, `DELETED`. So maybe : 

```
record McpSessionEvent(
   McpSession session,
   McpSessionEventType
) {

    enum McpSessionEventType {
        CREATED, OWNERSHIP_CHANGED, DELETED;
   }
}
```
 so consumers aren't inferring intent from the McpSession props

fhussonnois (changes requested):
Could you rebase your branch so that it contains only two commits—one for the backend and one for the UI?

webserver/src/main/java/io/kestra/mcp/FlowToolSchemaMapper.java
```diff
@@ -0,0 +1,331 @@
...
+            baseSchema.put("pattern", input.getValidator());
+        }
+
+        return baseSchema;
+    }
+
+    private static Map<String, String> buildItemsForMultiselectInput(MultiselectInput arrayInput, List<String> enumValues) {
+        return Map.of(
+            "type", getJsonSchemaType(arrayInput.getItemType()),
+            "enum", String.join(", ", enumValues)
```
fhussonnois: enums should be passed as array not comma-joined string

fhussonnois (changes requested):

webserver/src/main/java/io/kestra/mcp/FlowToolSchemaMapper.java
```diff
@@ -0,0 +1,331 @@
...
+        );
+    }
+
+    private static Map<String, Object> toSecretType(SecretInput input, Map<String, Object> baseSchema) {
+        return baseSchema;
+    }
+
+    private static Map<String, Object> toSelectType(SelectInput input, Map<String, Object> baseSchema) {
+        baseSchema.put(
+            "enum", input.getValues()
```
fhussonnois, line 291: Schema is wrong here: missing`type`

fhussonnois (changes requested):

core/src/main/java/io/kestra/core/mcp/models/McpSession.java
ammeek: @fdelbrayelle serverId refers to the MCP server id and sseNode refer to the web server node that currently handles the persistent see connection for sending async notifications.

fdelbrayelle: wrong ping I guess @ammeek ?

jdbc-postgres/src/main/resources/migrations/upgrade-v2.0-mcp-postgres.sql
loicmathieu, line 1: Is it really needed to store all those fields?
Only fields that you filter on must be stored.

jdbc-postgres/src/main/resources/migrations/postgres/V3_4__mcp.sql
```diff
@@ -0,0 +1,43 @@
+CREATE TABLE IF NOT EXISTS mcp (
+    "key" VARCHAR(250) NOT NULL PRIMARY KEY,
+    "value" JSONB NOT NULL,
+    "tenant_id" VARCHAR(150) GENERATED ALWAYS AS (value ->> 'tenantId') STORED,
+    "id" VARCHAR(250) NOT NULL GENERATED ALWAYS AS (value ->> 'id') STORED,
+    "description" TEXT GENERATED ALWAYS AS (value ->> 'description') STORED,
+    "instructions" TEXT GENERATED ALWAYS AS (value ->> 'instructions') STORED,
+    "server_type" VARCHAR(50) GENERATED ALWAYS AS (value ->> 'serverType') STORED,
+    "auth_type" VARCHAR(50) GENERATED ALWAYS AS (value ->> 'authType') STORED,
+    "enabled" BOOLEAN GENERATED ALWAYS AS (CAST(value ->> 'enabled' AS BOOLEAN)) STORED,
```
loicmathieu: We usuallu use disabled and not enabled, not important but may be more coherent

jdbc-postgres/src/main/resources/migrations/postgres/V3_4__mcp.sql
```diff
@@ -0,0 +1,43 @@
...
+    "fulltext" TSVECTOR GENERATED ALWAYS AS (
+        FULLTEXT_INDEX(CAST(value ->> 'id' AS VARCHAR))
+    ) STORED,
+    "created" TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updated" TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+CREATE INDEX IF NOT EXISTS mcp__deleted_tenant ON mcp ("deleted", "tenant_id");
+CREATE INDEX IF NOT EXISTS mcp__id_deleted_tenant ON mcp ("id", "deleted", "tenant_id");
+CREATE INDEX IF NOT EXISTS mcp__enabled_deleted_tenant ON mcp ("enabled", "deleted", "tenant_id");
```
loicmathieu: Each index has a cost, you can use an index for multiple query filter as long as all the leftmost filters are used.
That's why we usually start indexes with `tenant_id` as it's always included, then `deleted` again, always included, ...

jdbc-postgres/src/main/resources/migrations/postgres/V3_4__mcp.sql
```diff
@@ -0,0 +1,43 @@
...
+    "server_id"  VARCHAR(150) GENERATED ALWAYS AS (value ->> 'serverId') STORED,
+    "session_id" VARCHAR(150) NOT NULL GENERATED ALWAYS AS (value ->> 'sessionId') STORED,
+    "sse_node"   VARCHAR(250) GENERATED ALWAYS AS (value ->> 'sseNode') STORED,
+    "user_id"    VARCHAR(150) GENERATED ALWAYS AS (value ->> 'userId') STORED,
+    "created_at" TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+CREATE INDEX IF NOT EXISTS mcp_session__tenant_server ON mcp_session ("tenant_id", "server_id");
+CREATE INDEX IF NOT EXISTS mcp_session__session ON mcp_session ("tenant_id", "session_id");
+CREATE INDEX IF NOT EXISTS mcp_session__sse_node ON mcp_session ("sse_node");
```
loicmathieu: Same here, if you can use less indices this would be great

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcMcpServerRepository.java
```diff
@@ -0,0 +1,113 @@
...
+                    .from(jdbcRepository.getTable())
+                    .where(this.defaultFilter(tenantId))
+                    .and(field("id", String.class).eq(id))
+                    .fetchAny();
+                return record == null ? Optional.empty() : Optional.of(jdbcRepository.map(record));
+            });
+    }
+
+    @Override
+    public ArrayListTotal<McpServer> list(Pageable pageable, String tenantId) {
```
loicmathieu: This is usually called listAll as it list all records

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcMcpServerRepository.java
```diff
@@ -0,0 +1,113 @@
...
+            toSave = new McpServer(mcpServer.tenantId(), mcpServer.id(),
+                mcpServer.description(), mcpServer.instructions(), mcpServer.serverType(), mcpServer.authType(),
+                mcpServer.enabled(), mcpServer.isDefault(), mcpServer.deleted(),
+                Instant.now(), Instant.now());
+        } else {
+            toSave = new McpServer(mcpServer.tenantId(), mcpServer.id(),
+                mcpServer.description(), mcpServer.instructions(), mcpServer.serverType(), mcpServer.authType(),
+                mcpServer.enabled(), mcpServer.isDefault(), mcpServer.deleted(),
+                previousMcpServer.created(), Instant.now());
+        }
```
loicmathieu: It's always better to add a `with` method inside the record for such thing, more readable

loicmathieu (changes requested):
I didn't check the webserver / UI part so it's only a partial review on the core/jdbc part

tchiotludo:

Spec generated with EE branch `feat/flow-trigger-based-mcp`. Diff vs [client-sdk](https://github.com/kestra-io/client-sdk/blob/main/kestra-ee.yml):
```diff
21968a21969
>       - MCP_SERVER
```

fhussonnois (approved):

core/src/main/java/io/kestra/plugin/core/trigger/McpToolTrigger.java
```diff
@@ -0,0 +1,246 @@
...
+    ) {
+        Execution execution = Execution.builder()
+            .inputs(input)
+            .flowId(flow.getId())
+            .state(new State())
+            .id(IdUtils.create())
+            .flowRevision(flow.getRevision())
+            .namespace(flow.getNamespace())
+            .tenantId(flow.getTenantId())
+            .kind(ExecutionKind.NORMAL)
```
loicmathieu: Don't set kind, for nomal kind we use null

webserver/src/main/java/io/kestra/mcp/McpServerChangeNotifier.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static io.kestra.plugin.core.trigger.McpToolTrigger.DEFAULT_SERVER_ID;
+
+@Slf4j
+@Context
```
loicmathieu: If you can avoid using `@Context` it would be better as it move this bean and all its dependencies to an early initialization stage.

For beans never injected, you can use a `StartupEvent`  method.

loicmathieu (commented):
I didn't review in detail the PR and I saw that Florian already accepted it so my remarks didn't need to be taken into detail if you didn't found them legitimate.

I'm a bit concern by the news queues added, each queue adds a polling query in JDBC so we need to add new queue with care and here I'm not sure new queues are mandatory, I wonder if we cannot do the same without them. Just challenging the implementation, no strong pushback.

ammeek:
Thank you for the feedback @loicmathieu, I've made the changes based on your feedback. 
@fhussonnois also raised concerns about adding new queue when they are not needed, originally we had a MCP server and MCP session queue which would give us notifications when either entity was updated. We decided that the MCP server queue was not needed and we opted to use the ClusterEvent to send notification about mcpServer updates.

This leaves us with one new queue, the mcp session queue. This queue is needed to send notification across webserver node. It may be possible to move to a push based model via grpc which would allow us to get ride of the session queue.

I am happy to raise a follow up issues to look into this once the already scoped features for this epic have been implemented.

loicmathieu:
@ammeek yes, a followup issue is OK.

For the clustereventqueue, if you add a new subsriber, this is not really different to a new queue as all subscriber will poll the table. This can also be a followup issue as to reduce the number of susbscriber of the clusterevent queue and use a fanout mechanism. This issue is not very specific to your PR but we may want to try to subscribe to this queue once and then send to multiple consumers using async in-JVM events or anything else.

PR #15693, opened 2026-04-22
test(queue): settle after resume in broadcast pause test

fhussonnois:
Avoids a race where messages emitted immediately after resume() reach the broker before the subscriber has re-attached — dropped on non-persistent brokers (Redis pub/sub). No effect on persistent brokers.


Skraye (approved):
LGTM

PR #15701, opened 2026-04-23
fix: test task caching and flaky test no-source issues

ammeek:
✨ Description

Fixes two build issues introduced by enabling Develocity build caching:

1. Test tasks were cached after failure. ignoreFailures = true caused all test tasks to always exit with code 0, so Develocity cached "successful" runs that contained failing JUnit XML results. Subsequent runs restored those stale failures from cache. Removed ignoreFailures from commonTestConfig so failed tasks exit non-zero and are not cached. Added org.gradle.continue=true to gradle.properties in both repos so all module tests still run even when one fails.
2. Flaky tests never ran. The flakyTest task was missing testClassesDirs and classpath, which are required for custom Test tasks but not automatically inherited. Without them Gradle marked the task NO-SOURCE and skipped it. Added the missing properties, filter { failOnNoMatchingTests = false } (some modules have no flaky tests), and outputs.doNotCacheIf to prevent non-deterministic results being cached.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -252,7 +252,6 @@ subprojects {subProj ->
             t.timeout = java.time.Duration.ofMinutes(
                 (System.getenv('GRADLE_TEST_TIMEOUT_MIN') ?: '30').toInteger()
             )
-            t.ignoreFailures = true
             t.finalizedBy jacocoTestReport
             t.doFirst {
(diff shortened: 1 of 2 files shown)


brian-mulier-p (approved):
lgtm

PR #15710, opened 2026-04-24
fix(flow): duplicate multiple condition

### LINKED ISSUE #14196: MultipleCondition trigger causes duplicated executions with 2+ executor replicas
### Describe the issue

Hello Kestra team 👋 ,
We are seeing a specific behavior in a Kubernetes deployment of Kestra using the official Helm chart and Postgres 17. When running with two or more executor replicas, the `MultipleCondition` trigger produces duplicate flow executions for a single triggering event

This seems to happen randomly, but is usually reproducible within 5 minutes of testing. Notably, once the behavior starts, it tends to persist for a period of time where every subsequent run is duplicated. This is followed by periods when the issue does not occur. It occurs more frequently on GKE/CloudSQL, likely due to the slight increase in latency between pods and the database compared to a local `kind` cluster.

It's also worth mentioning that this wasn't a problem on v0.19.x.

## Steps to Reproduce
1. Deploy Kestra using the attached Helm chart.
2. Ensure the backend is PostgreSQL (as per the attached manifests).
3. Import the attached flows: `root-scheduled`, `dependency-1`, and `dependency-2`.
4. Run the `root-scheduled` flow manually via the UI multiple times.
5. Check the `dependency-1` and `dependency-2` execution lists, where you can see duplicated executions.

## Expected Behavior
Each trigger should produce exactly one execution of the dependent flow, regardless of the number of executor replicas.

## Evidence
https://github.com/user-attachments/assets/8da03b36-4216-4792-91dd-9f3b8dd7c989

**Video Walkthrough:**
- **00:06**: Execution of `root-scheduled`
- **00:09**: `dependency-1` shows duplicated executions
- **00:16**: Execution history cleared
- **00:20**: New execution of `root-scheduled`
- **00:22**: Success (no duplicates this time)
- **00:24**: New execution of `root-scheduled`
- **00:26**: **2 duplicates** appear on `dependency-1`
- **00:32**: New execution of `root-scheduled`
- **00:36**: Another **2 duplicates** on `dependency-1`

We’d really appreciate your help in resolving this. Currently, relying on a single executor prevents us from running all our existing flows, so we’ve had to roll back some of them to the old Kestra instance. Thank you! :)

### Environment

* **Kestra Version: v1.1.5 and v1.2.0** 
* **Infrastructure:** GKE + CloudSQL (Reproduced also on local `kind` cluster)
* **Database:** PostgreSQL 17
* **Executor Replicas:** 2+

This is the complete environment we deployed in our local `kind` cluster to replicate the behavior
[affected-flows.zip](https://github.com/user-attachments/files/24674641/affected-flows.zip)
[kestra-helm-chart.zip](https://github.com/user-attachments/files/24674629/kestra-helm-chart.zip)
[postgres-deployment.zip](https://github.com/user-attachments/files/24674637/postgres-deployment.zip)

loicmathieu:
Fixes https://github.com/kestra-io/kestra/issues/14196

MultipleCondition evaluation could lead to duplicate flow trigger because there was done outside of any lock. Wrap them inside a transaction, guarded by a SELECT FOR UPDATE.

We also need to implement a safe way to insert the window if not existing, reusing the pattern previously used by concurrency limit.


Skraye (approved):
LGTM

PR #15711, opened 2026-04-24
feat(async): blocking single-action + async batch for executions and triggers

fhussonnois:
Single-action mutating endpoints now block and return the updated resource. By-ids / by-query endpoints return 202 with {operationId, totalItems}.

Introduces AsyncOperation marker, AsyncOperationProcessedEvent broadcast, AsyncEventStreamingService, AsyncActionWaiter, ApiAsyncOperationResponse. Executor and scheduler handlers emit processed events for tagged commands.


core/src/main/java/io/kestra/core/async/AsyncOperationService.java
```diff
@@ -0,0 +1,60 @@
...
+        String operationId = message.operationId();
+        if (operationId == null) {
+            return;
+        }
+        try {
+            asyncOperationProcessedEventQueue.emit(new AsyncOperationProcessedEvent(
+                operationId, tenantId, itemId, outcome, error, Instant.now()
+            ));
+        } catch (QueueException e) {
+            log.error("Failed to emit AsyncOperationProcessedEvent for op {} item {}", operationId, itemId, e);
```
loicmathieu:
```suggestion
            log.error("Failed to emit AsyncOperationProcessedEvent for op {} item {}, ignoring it", operationId, itemId, e);
```

As the failure is not propagated, better to specify it inside the log

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -1087,11 +1093,52 @@ public Mono<ApiAsyncEvent> replayExecutionWithinputs(
...
+        var replayedExecution = executionService.replay(execution, flow, taskRunId, revision, breakpoints, true);
+
+        AsyncOperationProcessedEvent processed;
+        try {
+            processed = asyncOperationWaiter.submitAndWait(
+                execution.getId(),
+                operationId -> {
+                    try {
+                        // emit the replayed execution (new run, no operationId tagging)
+                        executionQueue.emit(replayedExecution);
```
loicmathieu, line 1116: I think you can emit the new execution outside the asynchronous processing.
I'll move it just after the call to the executionService.
The async operation is to add labels to the existing exec only

loicmathieu (commented):
I wonder if it is the right approach from the ExecutionController to return direclty the execution.
It may be better to return a `Publisher<Execution>` and let Micronaut call and resolves it in the correct thread.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -2419,4 +2424,68 @@ protected boolean validateExecutionACL(Execution execution) {
...
+                },
+                asyncWaitTimeout
+            );
+        } catch (TimeoutException e) {
+            throw new HttpStatusException(HttpStatus.GATEWAY_TIMEOUT,
+                "Operation timed out waiting for state transition");
+        }
+        if (processed.outcome() == AsyncOperationProcessedEvent.Outcome.FAILED) {
+            throw new HttpStatusException(HttpStatus.CONFLICT,
+                actionName + " failed: " + processed.error());
```
brian-mulier-p: add executionId & operationId to the error message

brian-mulier-p (approved):
LGTM

Issue #12747, opened 2025-11-06
Clean up the navigation in the upper-right corner to keep only one main action button and hide all other actions behind the 3-dots menu


tchiotludo:
On flows pages: 
<img width="786" height="224" alt="Image" src="https://github.com/user-attachments/assets/f8660fc5-33c7-4dea-bb1b-556e10c1f52d" />

Vs on executions pages: 
<img width="835" height="261" alt="Image" src="https://github.com/user-attachments/assets/d02657d4-08b8-487e-8938-3f65e1fae71a" />

To prevent this behavior, on all pages (everywhere on the application) 
We should: 
- Have only 1 main button
- All others button should move to an hamburger menu

anna-geller:
I agree that what's shown in the screenshot could be seen as a bug + quick-win, but the proposed solution is to clean up all action menus across the application, and that's not a quick win anymore and it's not a bug so adjusted labels, FYI :) 

great issue for cooldown to clean this up @kestra-io/frontend 🙏

sagnik26:
Hi @tchiotludo , is this issue open for contributors who have already completed a first PR from a good-first-issue? I’d be happy to take it up if allowed. Thanks!

MilosPaunovic:
Hi @sagnik26,

This is something we need to handle internally. 

If you're still willing to contribute, may I suggest that you check our [curated list of good first issues](https://go.kestra.io/contributing) to see if there is something you find interesting enough and think you can tackle independently. It can be further filtered by desired labels to narrow down the results (suggesting to filter by either `area/backend`, `area/plugin` or `area/frontend` ones, based on your preferences).

sagnik26:
@MilosPaunovic , Thanks for the clarification. Just to understand, is there a particular label, tag, or indicator that shows when an issue is being handled internally? It would help with filtering.

MilosPaunovic:
There is a label for other way around, and it's `good first issue`. You should look for those.

sagnik26:
@MilosPaunovic Got it. I was under the impression that I couldn't pick up a good-first-issue if I have already completed a similar contribution. Thanks for clarifying! Sorry for the off-topic discussion on this issue.

MilosPaunovic:
If not stated differently in the issue description, you're free to pick them up.

PR #15713, opened 2026-04-24
feat(flows): collapse secondary actions into overflow dots menu

RemiBarthe:
Part of https://github.com/kestra-io/kestra/issues/12747

**Navigation Bar Actions Refactor:**

* Introduced new `NavBarActions`, `NavBarAction`, and `NavBarActionsDropdown` components to standardize how navigation bar actions are rendered. These components automatically group secondary actions into a dropdown if there are more than one, improving usability on smaller screens. 
* Updated `Header.vue` and `Flows.vue` to use the new `NavBarActions` system, replacing previous `<ul>`/`<li>` and `<el-button>` lists with a cleaner, more maintainable structure using the new components. 
* Removed redundant or now-unnecessary CSS and markup related to previous action button layouts.

**Dashboard Selector Refactor:**

* Extracted dashboard selector content into a new `Content.vue` component, improving code clarity and separation of concerns. The selector now supports both dropdown and dialog modes, depending on context. 

**Top Navigation Bar Cleanup:**

* Simplified the `TopNavBar.vue` component by moving the logs deletion button and its associated logic.


MilosPaunovic (changes requested):
Hey @RemiBarthe, at first glance, there is no more primary button on the first page after login:

image

And there seem to be some warnings in the console, probably what's making the button not appear:
image

Could you tackle that first and I'll continue the review afterwards?

MilosPaunovic (approved):
Overall, excellent work! 🚀 

On small nitpick:

When reloading the page on main `Dashboard`, for a quick second there is a three-dots button which is then replaced by the dashboard selector. Can we handle that in a different way, maybe show nothing until it's computed which button needs to be shown? That same behavior is observed on all the pages, not just limited to `Dashboard` one.

After handling that one, I'm okay with the merge.

MilosPaunovic (changes requested):
Ahh, also, it seems like the `Storybook` tests are failing inside the CI, can you check them as well?

RemiBarthe:
> Overall, excellent work! 🚀
> 
> On small nitpick:
> 
> When reloading the page on main `Dashboard`, for a quick second there is a three-dots button which is then replaced by the dashboard selector. Can we handle that in a different way, maybe show nothing until it's computed which button needs to be shown? That same behavior is observed on all the pages, not just limited to `Dashboard` one.
> 
> After handling that one, I'm okay with the merge.

Sure, on it !



> Ahh, also, it seems like the `Storybook` tests are failing inside the CI, can you check them as well?

I'll have a look too

MilosPaunovic (approved):

PR #15717, opened 2026-04-25
feat(flows): add draft property to flow revisions

tchiotludo:
## Summary

Closes kestra-io/kestra-ee#6697.

Introduces a `draft` boolean property on flow revisions. A draft revision is **never picked up implicitly** when an execution starts without an explicit revision — webhooks, schedules, subflows and manual triggers all resolve to the latest non-draft revision instead. Executions can still target a draft by passing the revision explicitly.

### Confirmed semantics

- **Execute on a draft-latest flow without revision** → silently fall back to the latest non-draft revision (404 if none).
- **Scheduler on a draft-latest flow** → falls back to the latest non-draft revision; saving a draft does not stop existing schedules.
- **Revision bump** → `draft` is part of the revision diff, so toggling `draft: true` → `false` (Save after Save-as-draft) creates a new revision and behaves as a normal "publish".

### Backend changes

- New `draft` field on `AbstractFlow` (default `false`), propagated through `Flow`, `FlowWithSource` and `FlowForExecution`.
- New repository methods on `FlowRepositoryInterface`: `findByIdForExecution`, `findByIdForExecutionWithoutAcl`, `findByIdWithSourceForExecution`, `findByIdWithSourceForExecutionWithoutAcl`, `findAllWithSourceForExecutionForAllTenants`. Each returns the latest non-draft revision.
- JDBC: new `JdbcFlowRepositoryService.lastNonDraftRevision` filters drafts before the row-number partition so flows whose head is a draft fall back to the previous non-draft revision; flows with only draft revisions are not returned.
- Generated `draft` column on the `flows` table for H2, MySQL and PostgreSQL via new flyway migrations + a covering index. `COALESCE` keeps legacy rows whose JSON does not yet carry the field as published.
- Wired the new lookups into every implicit execution start path: `ExecutionController` webhook + bulk replay, `FlowService.getFlowIfExecutableOrThrow`, the core `DefaultFlowMetaStore.findById` (subflow path), the scheduler's `DefaultFlowMetaStore` (per-vNode trigger registration) and `DefaultSchedulableTriggerFetcher` (defense-in-depth `flow.isDraft()` skip).

### Frontend changes

- `draft?: boolean` added to the `Flow` type and to the JSON schema for editor autocompletion.
- `EditorButtons.vue` now renders a split button: **Save** (publish, sets `draft: false`) and **Save as draft** (`draft: true`). Wording stays "save" — no "Deploy" / "Publish" verbs per request.
- New `saveAsDraft` action on the flow store; both `save` / `saveAll` now write the draft flag explicitly so the persisted revision matches the action the user took. The flag is set with a small inline YAML helper because `YAML_UTILS.updateMetadata` drops keys outside its allow-list.
- A "Draft" badge is shown in `FlowRootTopBar`, the flow list rows pick up a `draft` CSS class, and `FlowRun.vue` shows a warning when the latest revision is a draft.

### Tests

- `AbstractFlowRepositoryTest` covers: round-trip of `draft`; default of `false`; `findByIdForExecution` returning the previous non-draft when the head is a draft; empty result when only drafts exist; `findAllWithSourceForExecutionForAllTenants` filtering correctly.
- `ExecutionControllerTest` covers the webhook → 404 case for a draft-only flow, backed by a new fixture `webhook-draft.yaml`.
- New Vitest spec verifies the YAML mutation behavior used by `saveAsDraft` and that the action is exposed on the store.

## Test plan

- [ ] `./gradlew :core:test --tests "*FlowRepositoryTest"` passes.
- [ ] `./gradlew :jdbc-h2:test :jdbc-postgres:test :jdbc-mysql:test` apply the new migrations and run the inherited repository tests on each backend.
- [ ] `./gradlew :webserver:test --tests "ExecutionControllerTest"` covers the new webhook draft case.
- [ ] `./gradlew :scheduler:test` exercises the scheduler with the new `findAllWithSourceForExecutionForAllTenants` source.
- [ ] `cd ui && npm run check:types && npx vitest run tests/unit` — performed locally; 53 unit tests pass.
- [ ] Manual smoke in the UI: create a flow, **Save**, then **Save as draft** → list shows "Draft" class, topbar shows the badge, `Execute` warns, schedule does not fire on the draft revision; clicking **Save** again republishes and re-arms the schedule.

https://claude.ai/code/session_01Ema5CYQ3pEzFQd6HPXeERn

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ema5CYQ3pEzFQd6HPXeERn)_


anna-geller (approved):
I think only this use case needs to be addressed: when you iterate on a workflow and you have some invalid/deprecated syntax (as in this example), it should be possible to save it as a draft:

```yaml
id: bad_retry
namespace: company.team

tasks:
  - id: hello
    message: Hello World! 🚀
    retry:
      interval: PT1M
      type: constant
      maxAttempt: 3
```

Image

MilosPaunovic:
Hey @tchiotludo! 👋

tchiotludo:
rebase is done !

scheduler/src/main/java/io/kestra/scheduler/internals/DefaultSchedulableTriggerFetcher.java
```diff
@@ -95,7 +95,9 @@ public List<TriggerEvaluationContext> getSchedulableTriggers(final Clock clock,
                 // has not yet been processed. In these cases, 
                 final String triggerId = triggerState.getTriggerId();
                 Optional<AbstractTrigger> maybeTrigger = flow.getTriggers().stream().filter(it -> it.getId().equals(triggerId)).findFirst();
-                if (flow.isDisabled() || maybeTrigger.isEmpty() || maybeTrigger.get().isDisabled()) {
+                // Defense in depth: the meta-store already filters drafts, but a stale trigger
+                // event may briefly point to one if the meta-store hasn't refreshed yet.
+                if (flow.isDisabled() || flow.isDraft() || maybeTrigger.isEmpty() || maybeTrigger.get().isDisabled()) {
```
loicmathieu: I'm not sure this is correct, it should be filtered inside the metastore, if a draft arrives here it's a bug and it means that if a previous version exist it will not be taken into account.

scheduler/src/main/java/io/kestra/scheduler/TriggerScheduler.java
```diff
@@ -348,11 +352,17 @@ private void processSchedulableTrigger(Clock clock, ZonedDateTime scheduleTime,
 
         // May send a new execution - if Schedulable trigger or on error
         final String tenantId = triggerState.getTenantId();
+        final FlowInterface flowToExecute = triggerEvaluationContext.flow();
         evaluationResult.ifPresent(res ->
         {
             var execution = res.toExecution(triggerContext)
                 .withScheduleDate(scheduleTime.toInstant())
                 .withTenantId(tenantId);
+            // Defense in depth: if the flow has constraint violations (it should not after save
```
loicmathieu: I'm not convinced by this defense in depth, we filter draft flows form the metastore so they should never arrives here!
And validating a flow on each evaluation would be very costly.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -740,6 +779,34 @@ public Publisher<ExecutionResponse> createExecution(
...
+        String errorMessage
+    ) {
+        Execution failedExecution = Execution.newExecution(flow, null, parsedLabels, scheduleDate)
+            .toBuilder()
+            .kind(kind.orElse(null))
+            .build()
+            .withState(State.Type.FAILED);
+        runContextFactory.of(flow, failedExecution).logger().error(errorMessage);
+        try {
+            executionQueue.emit(failedExecution);
```
loicmathieu: You need to send a `Create` command and wait for it to be processed

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -659,8 +660,46 @@ public Publisher<ExecutionResponse> createExecution(
...
+                }
+            }
+            throw e;
+        }
+
+        // Drafts can be saved with constraint violations. When the user explicitly executes one
+        // (by passing the revision) the request is accepted but the execution is created already
+        // FAILED, with the validation error logged through the run context so it is persisted
+        // and visible in the UI's execution log.
+        Optional<ConstraintViolationException> violations = flowService.validateForExecution(flow);
```
loicmathieu: If only draft flows must be re-validated, do this part conditionnaly

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -659,8 +660,46 @@ public Publisher<ExecutionResponse> createExecution(
...
         List<Label> parsedLabels = parseLabels(labels);
+        final Flow flow;
+        try {
+            flow = flowService.getFlowIfExecutableOrThrow(tenantId, namespace, id, revision);
+        } catch (NoSuchElementException e) {
+            // No non-draft revision was found. If the flow exists but only has draft revisions,
+            // we accept the request and emit a FAILED execution explaining why instead of a bare
+            // 404, so the user sees the failure in the UI executions list.
+            if (revision.isEmpty()) {
+                Optional<Flow> latestAny = flowRepository.findByIdWithoutAcl(tenantId, namespace, id, Optional.empty());
```
loicmathieu, line 700: This will do 2 DB queries each time the flow is in draft, I wonder if we could find a way to avoid that to lower miss cost.
This can be improved in a followup issue

loicmathieu (changes requested):
@elevatebart can you review the frontend part?
@fhussonnois can you review the scheduler part?

loicmathieu:
@flcarre please followup on the UI

flcarre:
Here is what it's look like in the ui
https://github.com/user-attachments/assets/141ce0cb-07fe-49af-b73e-ef087de80ea6

loicmathieu:
@flcarre thanks for the UI part, seems good and neat.

scheduler/src/main/java/io/kestra/scheduler/internals/DefaultSchedulableTriggerFetcher.java
AcevedoR: should be ok now

scheduler/src/main/java/io/kestra/scheduler/TriggerScheduler.java
AcevedoR: should be ok now

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
AcevedoR: should be ok now

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -697,8 +698,48 @@ public Publisher<ExecutionResponse> createExecution(
...
+                if (latestAny.isPresent() && latestAny.get().isDraft()) {
+                    Flow draftFlow = latestAny.get();
+                    return emitFailedExecution(
```
loicmathieu: From the API, we usually throw and not create a FAILED execution, this is lighter and as we it's done sync it's better to warn the client that may think the execution has been created correclty.

The same for the second bloc under

loicmathieu (changes requested):
You need to rename your revisions, that version is already taken!

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
AcevedoR: done

loicmathieu (approved):
LGTM

PR #15532, opened 2026-04-14
fix(ui): update logs route-level filtering to prevent flickering

RedHolger:
**Description:**

Changing from DynamicScrolling behaviour to RecycleScroller has fixed the issue. Dynamic scroller keeps measuring/remeasuring rows as content is changed, which causes positional flickers while RecycleScroller has fixed item size. For a heavy logging scenario in temporal logs DynamicScrolling will cause recalculations and constant changes, therefore using the latter fixes the problem.

**Closes https://github.com/kestra-io/kestra/issues/15382.**

**Proof:**
https://jam.dev/c/a4ee435a-fdcf-4803-8ef7-2655b7945baf

PR #15724, opened 2026-04-27
fix(executions): resolve log viewer scroll oscillation and flickering

karelkryda:
### ✨ Description

Replaces nested `DynamicScroller` (`vue-virtual-scroller`) with plain DOM scrolling in `TaskRunDetails.vue` to fix log viewer scroll oscillation and flickering.

The nested `DynamicScroller` setup (outer scroller for task runs, inner scroller for log lines) caused an infinite `ResizeObserver` feedback loop where `scrollHeight` oscillated between two values, making the log view flicker constantly during and after execution. Debug logging confirmed the oscillation persisted even with no new log lines arriving.

This also resolves the infinite scroll-to-bottom loop observed on `v1.2`, where the `@resize` handler on the inner `DynamicScroller` called `scrollToBottomFailedTask()` on every item resize, which triggered `scrollToBottom()` in a recursive `requestAnimationFrame` loop, which caused further resizes - creating a self-sustaining cycle that made it impossible to scroll up through log history. On the `develop` branch this manifested as visible flickering and scroll position oscillation.

Changes:

- Replace outer `DynamicScroller`/`DynamicScrollerItem` with `div` + `v-for`
- Replace inner `DynamicScroller`/`DynamicScrollerItem` with `div` + `v-for` + `overflow-y: auto`
- Implement auto-scroll via `MutationObserver` with pinned-to-bottom detection (20px threshold)
- Use `scrollTo` with `behavior: "smooth"` for live log tailing
- Preserve scroll position when user scrolls up to read history
- Remove `@resize` handler and `sizeDependencies` that contributed to the feedback loop
- Remove unused `vue-virtual-scroller` CSS import

#### Before: Blocked scroll on v1.2
https://github.com/user-attachments/assets/7e9af058-d4ed-40cc-9f33-6c1162357101

#### Before: Flickering on develop
https://github.com/user-attachments/assets/2dee7335-0779-4347-9949-5d7360dbf82a

#### After
https://github.com/user-attachments/assets/cecb9783-d155-4b19-90df-84d3a6bf5077

### 🔗 Related Issue

N/A

### 🎨 Frontend Checklist

### 📝 Additional Notes

N/A


karelkryda:
oh gosh how i hate merge conflicts ... hopefully its done now 😄

Piyush-r-bhaskar:
This PR is trying to fix the same issue for which https://github.com/kestra-io/kestra/pull/15532 is raised. So, I will request Brian's review on this too.

karelkryda:
Hey @brian-mulier-p, I've read the feedback on #15532 regarding performance.
I'm currently working on a solution that removes the virtual scroller (which causes the flickering feedback loop) and replaces it with bounded DOM rendering - only rendering a portion of the logs at a time with lazy loading on scroll.

Still refining it, will push an update soon.

karelkryda:
I've spent some time investigating the performance side as well. I experimented with windowed DOM rendering (only rendering a portion of log lines with lazy loading on scroll-up), but managing scroll position with variable-height items turned out to be quite complex and felt like it would expand the scope of this PR significantly.

From my benchmarking, the current approach handles executions (tasks) up to ~5k log lines smoothly, which should cover the vast majority of real-world use cases. For larger volumes (10k+), I think the right approach would be backend pagination combined with frontend lazy loading.

brian-mulier-p (changes requested):
Hello! Sorry for the late review. Unfortunately the current implementation is not one that will fit as we want to allow very large scaling which is why we introduced the DynamicScroller at some point (which recycles DOM elements). I know this is a complex thing to handle but we need to stick with some solution that recycles / reuse elements for performance reasons. As there is another PR adressing the same issue but keeping the dynamic scroller, I will close this one. Still thanks for all your effort on this one. Feel free to have a look at some other issues 👍

PR #15730, opened 2026-04-27
fix: only remove worker job running on terminal task state

ammeek:
✨ Description       

GrpcWorkerControllerService was deleting the WorkerTaskRunning record from the database on every result received from the worker, including intermediate  
RUNNING state results.
                                                                                                                                                          
Because a worker sends a RUNNING result as soon as it begins executing a task, the record was being removed before the task had actually finished. If the 
worker was then forcefully terminated before sending a final result, the liveness coordinator's recovery scan found no WorkerTaskRunning entries for that
worker and had nothing to resubmit — causing the task to be silently lost rather than re-dispatched to a healthy worker.                                  
                                                          
The fix restores the isTerminated() guard so that the WorkerTaskRunning record is only deleted once the task has reached a terminal state (SUCCESS,       
FAILED, KILLED, etc.), keeping the record alive for the full duration of execution and ensuring the liveness coordinator can correctly resubmit tasks when
 a worker dies mid-execution.


loicmathieu (approved):
LGTM

PR #15732, opened 2026-04-27
feat(ui): move save filter action into saved filters popover

### LINKED ISSUE #12745: Filter are not usable on mobile
Image

- Saved filter should be only a button 
- Consider move the save button for the current search inside the saved filter dropdown (even on desktop), leading this more usable one: 
  Image

RemiBarthe:
closes https://github.com/kestra-io/kestra/issues/12745

Related to https://github.com/kestra-io/kestra-ee/issues/7772.

**Saved Filter Functionality Refactor:**

- Moved the `SaveFilters` button and dialog into the `SavedFilters` popover, so users now save or edit filters directly from the saved filters list rather than as a separate button. The related props and events are updated for seamless integration. 
- Updated the `SavedFilters` component to accept new props (`disabled`, `appliedFilters`, `editingFilter`) and emit new events (`save`, `update`, `close-edit`) to support the consolidated workflow.
- Removed the old `SaveFilters` button from the right filter bar and its associated styles, as its functionality is now handled within the saved filters popover.

**UI/UX and Style Improvements:**

- Adjusted the layout and styling of filter components for better responsiveness and consistency, including changes to flex wrapping, minimum widths, and button appearance.
- Updated the `SaveFilters` button to use a smaller, link-style appearance and improved its disabled state styling for clarity.


MilosPaunovic (changes requested):
Can we make the `Saved Filters` only a button on small screens, but keep the label included on larger ones?

After a bit of thinking, it might be better if @Piyush-r-bhaskar gives this a proper review, as he's much more knowledgable about this topic than I am.

Piyush-r-bhaskar (changes requested):
I would have waited little bit more and ask for a tweak from Nico with 2.0 design. Can we do that ?

if not , I agree with Milos, Let's keep the current UI for large screen and manage the Saved Filters part as a button with no label on mobile for now.

RemiBarthe:
> I would have waited little bit more and ask for a tweak from Nico with 2.0 design. Can we do that ?
> 
> if not , I agree with Milos, Let's keep the current UI for large screen and manage the Saved Filters part as a button with no label on mobile for now.

Nicolas is going to take a look, I'll wait for his feedback

PR #15749, opened 2026-04-28
chore(core)!: remove trigger conditions

loicmathieu:


core/src/test/java/io/kestra/core/runners/pebble/functions/IsDateBeforeFunctionTest.java
```diff
@@ -0,0 +1,112 @@
...
+import jakarta.inject.Inject;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collections;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+@MicronautTest
```
AcevedoR, line 15: do we really need @MicronautTest for this test ?

AcevedoR (approved):
LGTM

core/src/test/java/io/kestra/core/runners/pebble/functions/IsDateBeforeFunctionTest.java
loicmathieu: No, should not, but anyway, it has a high chance of being removed as I would most probably remove the functions.

PR #15753, opened 2026-04-28
feat(cli): add configuration-schema command for JSON Schema generation

Skraye:
Add a `configuration-schema` CLI command that generates a JSON Schema describing all Kestra configuration properties (from `@ConfigurationProperties` annotated classes) and installed plugin storage configurations. The schema is generated via classpath scanning and reflection, then uploaded to GCS during CI for both develop and release builds.

This also replaces scattered `@Value` annotations with proper `@ConfigurationProperties` records so the schema generator can discover them.

Companion PRs:
- https://github.com/kestra-io/kestra-ee/pull/7562
- https://github.com/kestra-io/api.kestra.io/pull/432
- https://github.com/kestra-io/actions/pull/87


tchiotludo:

Spec generated with EE branch `feat/configuration-schema-generator`. Diff vs [client-sdk](https://github.com/kestra-io/client-sdk/blob/main/kestra-ee.yml):
```diff
22381a22382,22393
>         preview:
>           nullable: true
>           allOf:
>           - $ref: "#/components/schemas/ServerConfig.Preview"
>         standalone:
>           nullable: true
>           allOf:
>           - $ref: "#/components/schemas/ServerConfig.Standalone"
>         service:
>           nullable: true
>           allOf:
>           - $ref: "#/components/schemas/ServerConfig.Service"
22405a22418,22453
>     ServerConfig.Preview:
>       type: object
>       properties:
>         initialRows:
>           type: integer
>           format: int32
>           default: 100
>         maxRows:
>           type: integer
>           format: int32
>           default: 5000
>     ServerConfig.Service:
>       type: object
>       properties:
>         purge:
>           nullable: true
>           allOf:
>           - $ref: "#/components/schemas/ServerConfig.Service.Purge"
>     ServerConfig.Service.Purge:
>       type: object
>       properties:
>         retention:
>           type: string
>     ServerConfig.Standalone:
>       type: object
>       properties:
>         running:
>           nullable: true
>           allOf:
>           - $ref: "#/components/schemas/ServerConfig.Standalone.Running"
>     ServerConfig.Standalone.Running:
>       type: object
>       properties:
>         timeout:
>           type: string
>           default: PT1M
```

tchiotludo:

Spec generated with EE branch `feat/configuration-schema-generator`. Diff vs [client-sdk](https://github.com/kestra-io/client-sdk/blob/main/kestra-ee.yml):
```diff
22378,22415d22377
<     ServerConfig:
<       required:
<       - terminationGracePeriod
<       type: object
<       properties:
<         workerTaskRestartStrategy:
<           nullable: true
<           allOf:
<           - $ref: "#/components/schemas/WorkerTaskRestartStrategy"
<         terminationGracePeriod:
<           type: string
<           default: 5m
<         liveness:
<           $ref: "#/components/schemas/ServerConfig.Liveness"
<     ServerConfig.Liveness:
<       required:
<       - enabled
<       - heartbeatInterval
<       - initialDelay
<       - interval
<       - timeout
<       type: object
<       properties:
<         enabled:
<           type: boolean
<           default: true
<         interval:
<           type: string
<           default: 5s
<         timeout:
<           type: string
<           default: 45s
<         initialDelay:
<           type: string
<           default: 45s
<         heartbeatInterval:
<           type: string
<           default: 3s
22485,22486d22446
<         config:
<           $ref: "#/components/schemas/ServerConfig"
23603,23608d23562
<     WorkerTaskRestartStrategy:
<       type: string
<       enum:
<       - NEVER
<       - IMMEDIATELY
<       - AFTER_TERMINATION_GRACE_PERIOD
```

cli/src/main/java/io/kestra/cli/StandAloneRunner.java
```diff
@@ -109,6 +107,13 @@ public void run() {
         }
     }
 
+    private Duration getRunningTimeout() {
+        if (serverConfig.standalone() != null && serverConfig.standalone().running() != null) {
```
loicmathieu: (optional) You may use Optional and chain them to avoid the chain of null-check

core/src/main/java/io/kestra/core/contexts/configuration/StorageTypeConfiguration.java
```diff
@@ -0,0 +1,9 @@
+package io.kestra.core.contexts.configuration;
+
+import io.micronaut.context.annotation.ConfigurationProperties;
+import io.micronaut.core.annotation.Nullable;
+
+@ConfigurationProperties("kestra.storage")
+public record StorageTypeConfiguration(
```
loicmathieu:
```suggestion
public record StorageConfiguration(
```

core/src/main/java/io/kestra/core/contexts/KestraBeansFactory.java
```diff
@@ -65,7 +65,7 @@ public StorageInterface storageInterface(final StorageInterfaceFactory storageIn
     }
 
     public String getStoragePluginId(StorageInterfaceFactory storageInterfaceFactory) {
-        return storageType.orElseThrow(
+        return Optional.ofNullable(storageTypeConfiguration.type()).orElseThrow(
```
loicmathieu: (optional) You can use Optional as a type inside the config record.

core/src/main/java/io/kestra/core/runners/pebble/functions/AbstractFileFunction.java
```diff
@@ -46,8 +46,8 @@ abstract class AbstractFileFunction implements KestraFunction {
     @Inject
     protected Provider<NamespaceFactory> namespaceFactory;
 
-    @Value("${" + LocalPath.ENABLE_FILE_FUNCTIONS_CONFIG + ":true}")
-    protected boolean enableFileProtocol;
+    @Inject
+    protected Provider<LocalFilesConfiguration> localFilesConfiguration;
```
loicmathieu: Why using a Provider here, I cannot imagine a dependency cycle for configuration bean

core/src/main/java/io/kestra/core/runners/VariableRenderer.java
```diff
@@ -268,10 +267,14 @@ public VariableConfiguration() {
             this.cacheEnabled = true;
             this.cacheSize = 1000;
             this.recursiveRendering = false;
+            this.redactedEnvVars = java.util.List.of(
+                "KESTRA_PLUGINS_PATH", "KESTRA_CONFIGURATION_PATH", "KESTRA_CONFIGURATION", "KESTRA_JAVA_OPTS"
+            );
```
loicmathieu: This should stay configurable so a user may add more redacted env vars

core/src/main/java/io/kestra/core/runners/WorkingDirFactory.java
```diff
@@ -12,8 +13,8 @@
 @Singleton
 public class WorkingDirFactory {
 
-    @Value("${kestra.tasks.tmp-dir.path}")
-    protected Optional<String> tmpdirPath;
+    @Inject
+    protected WorkingDirConfiguration workingDirConfiguration;
```
loicmathieu: This prop is not related to working dir (even if only used here), please create a `TasksConfiguration` object in case we need more config in it

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -211,20 +201,36 @@ public class ExecutionController {
...
+
+    private Integer initialPreviewRows;
+    private Integer maxPreviewRows;
+    private String kestraUrl;
     private boolean enableLocalFilePreview;
+    private Duration asyncWaitTimeout;
 
     @Inject
-    private ObjectMapper objectMapper;
+    void initConfig(
```
loicmathieu: (optional) we usually do field or constructor injection but never method injection like that so I'm not comptfortable using it

webserver/src/main/java/io/kestra/webserver/controllers/api/StaticFilter.java
```diff
@@ -31,17 +33,19 @@ public class StaticFilter implements HttpServerFilter {
...
     @Nullable
-    @Value("${kestra.webserver.html-title}")
     protected String htmlTitle;
-
     @Nullable
-    @Value("${kestra.webserver.html-head}")
     protected String htmlHead;
 
+    @jakarta.inject.Inject
+    void initConfig(WebserverConfiguration webserverConfiguration) {
```
loicmathieu: same here use constructor or field injection

core/src/main/java/io/kestra/core/contexts/KestraConfig.java
```diff
@@ -10,10 +12,10 @@
 public class KestraConfig {
```
fhussonnois: We could probably drop this class in favor of `SystemFlowsConfiguration` ?

webserver/src/main/java/io/kestra/webserver/controllers/api/StaticFilter.java
```diff
@@ -31,17 +33,19 @@ public class StaticFilter implements HttpServerFilter {
...
-
     @Nullable
-    @Value("${kestra.webserver.html-title}")
     protected String htmlTitle;
-
     @Nullable
-    @Value("${kestra.webserver.html-head}")
     protected String htmlHead;
 
+    @jakarta.inject.Inject
```
fhussonnois: Add the import for Inject

webserver/src/main/java/io/kestra/webserver/services/TriggerStateService.java
```diff
@@ -58,17 +56,17 @@ public class TriggerStateService {
...
-                               final FlowRepositoryInterface flowRepository,
-                               final TriggerEventQueue triggerEventQueue,
-                               final BroadcastQueueInterface<ExecutionKilled> executionKilledQueue,
-                               final AsyncOperationWaiter asyncOperationWaiter,
-                               @Value("${kestra.async-operations.wait-timeout:PT30S}") final Duration asyncWaitTimeout) {
+        final FlowRepositoryInterface flowRepository,
+        final TriggerEventQueue triggerEventQueue,
+        final BroadcastQueueInterface<ExecutionKilled> executionKilledQueue,
+        final AsyncOperationWaiter asyncOperationWaiter,
+        final io.kestra.webserver.configuration.AsyncOperationsConfiguration asyncOperationsConfiguration) {
```
fhussonnois: Add `import`

loicmathieu (changes requested):
I'm not sure I understand the purpose of this change (no description, no linked issues).

I was thinking it would creates a JSON schema for our configuration, but it creates some mega schema with all plugin properties and the config properties which are two very different things.
Moreover, it do that without our existing JSON schema support, manually, which may be complex to maintain in the long term.

Some additional notes:
- There are a lot of package prefix classes in code: use import
- I think a lot of null check can be avoided by using Optional fields and/or default fields (even full object)

fhussonnois (changes requested):
Why not adding all single config props to the `KestraConfiguration` ?

core/src/main/java/io/kestra/core/runners/VariableRenderer.java
Skraye: Those are defaults value and can be customized

Skraye:
@loicmathieu I have updated the description and added the companion PR (nice reminder)

> I was thinking it would creates a JSON schema for our configuration, but it creates some mega schema with all plugin properties 

Not really, its a JSON schema for our configuration, **we do not load all the plugin in the Schema**, the plugin we load are only the storage plugin because they are part of the Kestra Configuration

> it do that without our existing JSON schema support

Our current JSON schema isn't handling the same domain, sure they both generate a JSON schema and maybe some code could be centralized, but they do not handle the same data.

@fhussonnois 
> Why not adding all single config props to the KestraConfiguration ?

Because I'm not confident in having a catch-all configuration that could grow too much, but this is something that can be think again later

core/src/main/java/io/kestra/core/runners/VariableRenderer.java
loicmathieu: (optional) we never use inner classes for configuration, it may be better to extract it so it's more discoverable.

cli/src/main/java/io/kestra/cli/schema/ConfigurationSchemaGenerator.java
loicmathieu, line 1: Overall, I found this class too low level, I'm pretty sure frameworks exposes an easier way to generate a JSON schema (and maybe a framework that we already have in our dependencies) but I'm validating it anyway.

cli/src/main/java/io/kestra/cli/schema/ConfigurationSchemaGenerator.java
```diff
@@ -0,0 +1,579 @@
...
+                    target.put("type", "object");
+                    target.set("additionalProperties", itemSchema);
+                }
+            } else {
+                addPropertiesViaReflection(target, clazz);
+            }
+        }
+
+        if (pluginRegistry != null) {
+            addPluginStorageSchemas(root, pluginRegistry);
```
loicmathieu, line 89: Shouldn't we do the same for secrets?

cli/src/main/java/io/kestra/cli/schema/ConfigurationSchemaGenerator.java
```diff
@@ -0,0 +1,579 @@
...
+        }
+        return false;
+    }
+
+    private void discoverConfigClasses() {
+        SoftServiceLoader<BeanDefinitionReference> loader =
+            SoftServiceLoader.load(BeanDefinitionReference.class);
+
+        var iterator = loader.iterator();
+        while (true) {
```
loicmathieu:
```suggestion
        while (iterator.hasNext()) {
```

This is more idiomatic

cli/src/main/java/io/kestra/cli/schema/ConfigurationSchemaGenerator.java
```diff
@@ -0,0 +1,579 @@
...
+
+        if (!required.isEmpty()) {
+            ArrayNode requiredArray = target.has("required")
+                ? (ArrayNode) target.get("required")
+                : target.putArray("required");
+            required.forEach(requiredArray::add);
+        }
+    }
+
+    private void applySchemaAnnotation(ObjectNode node, Annotation[] annotations) {
```
loicmathieu, line 394: schema also allow to define a more specific types than the one discover by reflection, shouldn't it be supported also?

loicmathieu (approved):
LGTM, I added some comments and small changes but you can merge without them.

PR #14716, opened 2026-02-19
feat(kv): introduce Put task

fdelbrayelle:
closes https://github.com/kestra-io/kestra-ee/issues/3631

### QA :heavy_check_mark: 

Put a first value:

```yaml
id: kv_put_json
namespace: company.team

tasks:
  - id: update_my_json
    type: io.kestra.plugin.core.kv.Put
    key: my-json-key
    value:
      - key: "json-key-1"
        value: "my new value"
      - key: "nested.field"
        value: 123
```

**Result:**

image

Get it:

```yaml
id: kv_store_get
namespace: company.team

tasks:
  - id: kv_get
    type: io.kestra.plugin.core.kv.Get
    key: my-json-key
```

**Result:**

image

image

Update the values:

```yaml
id: kv_put_json2
namespace: company.team

tasks:
  - id: update_my_json
    type: io.kestra.plugin.core.kv.Put
    key: my-json-key
    value:
      - key: "json-key-1"
        value: "my new value (UPDATED)"
      - key: "nested.field"
        value: 456
```

**Result:**

image

Get it again:

```yaml
id: kv_store_get
namespace: company.team

tasks:
  - id: kv_get
    type: io.kestra.plugin.core.kv.Get
    key: my-json-key
```

**Result:**

image

Flow to distinct the difference between `Set` and `Put`:

```yaml
id: kv_set_put_get_demo
namespace: company.team

variables:
  key: kv_demo_key
  put_patch:
    - key: "user.city"
      value: "Lyon"
    - key: "status"
      value: "active"

tasks:
  - id: cleanup
    type: io.kestra.plugin.core.kv.Delete
    key: "{{ vars.key }}"
    errorOnMissing: false

  - id: set_initial
    type: io.kestra.plugin.core.kv.Set
    key: "{{ vars.key }}"
    value: '{"user":{"name":"alice","city":"Paris"},"status":"new"}'
    kvType: JSON
    overwrite: true

  - id: get_after_set
    type: io.kestra.plugin.core.kv.Get
    key: "{{ vars.key }}"
    errorOnMissing: true

  - id: put_precise_update
    type: io.kestra.plugin.core.kv.Put
    key: "{{ vars.key }}"
    value: "{% raw %}{{ vars.put_patch }}{% endraw %}"

  - id: get_after_put
    type: io.kestra.plugin.core.kv.Get
    key: "{{ vars.key }}"
    errorOnMissing: true

  - id: debug_after_put
    type: io.kestra.plugin.core.log.Log
[...]

PR #15754, opened 2026-04-28
test(tasks): fix qaKv sanity check size assertion (6 -> 7)

AcevedoR:
## Summary

`SanityCheckTest.qaKv` asserts `hasSize(6)` against the `sanity-checks/kv.yaml` flow, but the flow has had **7 tasks** since #14716 added the `put` task. The test fails deterministically with `Expected size: 6 but was: 7`.

The failure has been hidden because the test is tagged `@FlakyTest` (`@Tag("flaky")`), so regular `./gradlew test` skips it and the failure only surfaces in the non-blocking `flakyTest` run.

## Root cause

| When | Commit | What |
|---|---|---|
| 2025-10-10 | `ff18fc40e` (#11914) | Added `@FlakyTest` to `qaKv` for an earlier timing-related flake |
| 2026-02-19 | `746d7c9ae` (#14716) | Added 7th task (`put`) to `kv.yaml`, did not update the test |

The 7-task flow is the source of truth; the `hasSize(6)` is stale.

## Fix

Update assertion to `hasSize(7)`. `@FlakyTest` is intentionally left in place — it predates the put-task change and likely covers a separate timing concern that is out of scope here.

## Test plan

- [x] Reproduced failure locally: `./gradlew :core:flakyTest --tests "io.kestra.core.tasks.test.SanityCheckTest.qaKv" --rerun-tasks` -> `Expected size: 6 but was: 7`
- [x] After fix, same command passes (`✔ qaKv(Execution)`)
- [ ] CI green

### DIFF
--- a/core/src/test/java/io/kestra/core/tasks/test/SanityCheckTest.java
+++ b/core/src/test/java/io/kestra/core/tasks/test/SanityCheckTest.java
@@ -37,7 +37,7 @@ class SanityCheckTest {
     @Test
     @ExecuteFlow("sanity-checks/kv.yaml")
     void qaKv(Execution execution) {
-        assertThat(execution.getTaskRunList()).hasSize(6);
+        assertThat(execution.getTaskRunList()).hasSize(7);
         assertThat(execution.getState().getCurrent()).isEqualTo(State.Type.SUCCESS);
     }
 


nkwiatkowski (approved):
LGTM

PR #15757, opened 2026-04-28
fix(tasks): inject outputFiles into Pebble context in CommandsWrapper.render()

### LINKED ISSUE #13765: Consistent outputFiles Path Resolution for All Task Types (JDBC, Script, etc.)
### Problem Statement

There currently appear to be three different behaviors for how `outputFiles` are handled, at least across JDBC and Script tasks.

#### Script Example

For Script tasks, `outputFiles` must be referenced as a plain string literal inside the script body:

```yaml
tasks:
  - id: node
    type: io.kestra.plugin.scripts.node.Script
    outputFiles:
      - "out.txt"
    script: |
      const fs = require('fs');
      fs.writeFileSync('out.txt', 'hello from node');
```
Using a Pebble expressions like `{{ outputFiles["out.txt"]}` causes to an error.

When using a remote Task Runner, Pebble is required, but only for the working directory (not the file reference itself):

```yaml
tasks:
  - id: node
    type: io.kestra.plugin.scripts.node.Script
    taskRunner:
      type: io.kestra.plugin.ee.azure.runner.Batch
    outputFiles:
      - "out.txt"
    script: |
      const fs = require('fs');
      fs.writeFileSync('{{ workingDir }}/out.txt', 'hello from node');
```

#### JDBC Example

also see https://github.com/kestra-io/plugin-scripts/issues/307

For JDBC tasks, using a plain string literal like `'out.csv'` inside the sql statement does not work as expected. Instead, a Pebble expression must be used:

```yaml
tasks:
  - id: duckdb
    type: io.kestra.plugin.jdbc.duckdb.Queries
    outputFiles:
      - "out.csv"
    sql: |
      COPY (SELECT 'hello from sql' as msg) TO '{{ outputFiles["out.csv"] }}' WITH (FORMAT csv, HEADER true);
```

Using `'out.csv'` directly in the SQL leads to an empty file.

> [!NOTE]
> from @Malaydewangan09 
> _In JDBC tasks, **outputFiles** are explicitly converted into real filesystem paths and injected into the Pebble render context before rendering the SQL. That's why `{{ outputFiles.* }}` can be resolved there._
>
> _In script tasks, **outputFiles** are only rendered as a list of filenames and passed to the runner to be collected after execution. They are never added to the Pebble render context when the script is rendered, so Pebble cannot resolve outputFiles and fails._

### Proposed solution

Make **outputFiles** usage consistent across task types by allowing Pebble-based file references such as `{{ outputFiles["out.txt"] }}` everywhere, in particular:
	- Script tasks (local and remote / task runner based)
	- JDBC tasks

This would provide a single, predictable way to reference **outputFiles** paths across Kestra.
If we decide to align this across tasks, we could keep the legacy methods for file access available for backwards compatibility, but should update docs accordingly. 

- [ ] check how this affects glob-patterns for outputfiles

fdelbrayelle:
## Summary

- `CommandsWrapper.render(RunContext, Property<String>)` was missing the `outputFiles` map injection that already existed in `run()`
- `Script.run()` calls `render()` before `run()`, so `{{ outputFiles.name }}` resolved against a context without `outputFiles` and threw `IllegalVariableEvaluationException`
- Added the same non-glob `name → workingDir/name` map building to `render()` so the variable is available at script render time

## Related

Fixes https://github.com/kestra-io/kestra/issues/13765


fdelbrayelle:
Closing — branch was cut from releases/v1.3.x instead of develop. New PR incoming.

PR #15759, opened 2026-04-28
feat(helm): migrate kestra and kestra-starter charts to GCS

MarcoSaba:
## Summary

Moves `charts/kestra` and `charts/kestra-starter` from `kestra-io/helm-charts` into this repo, making it the single source of truth for chart releases. A new `helm-release.yml` workflow packages and publishes both charts to GCS on every version tag using Workload Identity Federation.

## Changes

- `feat(helm)/migrate-artifacts-with-cicd`: one-shot workflow that copied 157 historical `.tgz` artifacts and `index.yaml` from `kestra-io/helm-charts` gh-pages to GCS bucket `helm-charts-kestra-host`
- `feat(helm): move kestra and kestra-starter charts into repo`: chart directories copied from `helm-charts`, `sources` URL updated to `kestra-io/kestra`
- `feat(helm): add helm-release workflow for GCS publishing`: on tag push, bumps `version`/`appVersion`, packages both charts, merges into existing `index.yaml`, uploads to GCS via WIF auth

close kestra-io/helm-charts#284


tchiotludo:

gperreymond (approved):

.github/workflows/release-docker.yml
```diff
@@ -38,4 +38,15 @@ jobs:
...
\ No newline at end of file
+      GH_PERSONAL_TOKEN: ${{ secrets.GH_PERSONAL_TOKEN }}
+
+  helm-release:
+    name: Helm release
+    needs: publish-docker
+    if: startsWith(github.ref, 'refs/tags/v')
+    permissions:
+      contents: read
+      id-token: write
```
MarcoSaba, line 50: helm-release.yml declares permissions: id-token: write at workflow level. That applies when the workflow runs standalone (workflow_dispatch). When called via workflow_call, the token is already minted by the helm-release job in release-docker.yml before helm-release.yml even loads. The callee can only restrict the caller's token, never expand it.

PR #15761, opened 2026-04-28
ci: pass GITHUB_TO_SLACK_MAP to slack-status on develop failure

brian-mulier-p:
- forward the new \`users-map\` input on the \`slack-status\` composite action so the responsible committer is @-mentioned in Slack when develop fails (notifications today are silently ignored because nobody is pinged)
- value comes from the org-level \`GITHUB_TO_SLACK_MAP\` variable kept out of this public repo and managed by terraform in kestra-io/infra
- companion PRs: kestra-io/actions#88 (action change), kestra-io/infra#199 (variable), companion EE PR in kestra-io/kestra-ee


nkwiatkowski (approved):
LGTM

PR #15785, opened 2026-04-29
fix(scheduler): inject trigger variables into when-expression render context

AcevedoR:
## Summary

- `TriggerScheduler.evaluate()` called `render(trigger.getWhen())` without passing the `ConditionContext` variables that carry `trigger.date`
- Pebble expressions like `{{isDayWeekInMonth(trigger.date, 'MONDAY', 'FIRST')}}` received a null date argument, threw a `PebbleException`, and the catch block emitted a spurious `FAILED` execution instead of skipping the non-matching slot
- Fix: pass `conditionContext.getVariables()` as the additional variables map to `render()` so `trigger.date` is available during `when`-expression evaluation

## Test plan

- [ ] `TriggerSchedulerTest.shouldSucceedScheduleConditionalScheduleTriggerGivenValidTimeZone` passes (was failing with `expected: 0 but was: 1`)
- [ ] `TriggerSchedulerTest.shouldStoreNextConditionMatchingDateOnFirstEvaluationWhenConditionFails` passes (same root cause)
- [ ] Full scheduler test suite (`./gradlew :scheduler:test :scheduler:flakyTest`) — 74 regular + 2 flaky — all green


tchiotludo:

PR #15801, opened 2026-04-29
fix(ui): revert vue-virtual-scroller to 2.0.1 — v3.x breaks DynamicScroller in headless CI

### LINKED ISSUE #14989: Autocomplete doesn't always display in editor

brian-mulier-p:
Closes #14989

## Root cause

Scheduled E2E runs failed on `flow.spec.ts` — both `should create and execute the example Flow` and `should create and execute a Flow with input` timed out waiting for task names in the Gantt panel.

Root cause: `vue-virtual-scroller` v3.x (including both 3.0.1 and 3.0.2) renders 0 items in headless Chrome because `clientHeight = 0` at `onMounted` time for conditionally-rendered scrollers. Both the outer Gantt scroller (`v-else-if="execution && executionsStore.flow"`) and the inner log-lines scroller inside `TaskRunDetails` exhibit this. Multiple fix attempts (inline height, `watch(ganttScroller)` + nextTick, requestAnimationFrame) gave at best 1/6 CI pass rate — the `clientHeight = 0` window is a race that cannot be reliably closed within v3's rendering model.

v2.0.1 was the last confirmed-green version (CI run 25101607686).

## Fix

Revert `vue-virtual-scroller` from v3.x back to `2.0.1`. Add a dependabot ignore for `>=3.0.0` to prevent automatic re-bump until upstream ships a stable v3 that works in headless environments.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -143,6 +143,6 @@ updates:
       - dependency-name: "axios"
         versions: ["*"]
 
-      # Pin vue-virtual-scroller patch updates — 3.0.2 broke Gantt DynamicScroller rendering (CI run 25118331295)
+      # Reverted to 2.0.1 — v3.x DynamicScroller renders 0 items in headless CI (clientHeight=0 at mount time)
       - dependency-name: "vue-virtual-scroller"
-        update-types: ["version-update:semver-patch"]
+        versions: [">=3.0.0"]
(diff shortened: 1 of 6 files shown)


tchiotludo:

MilosPaunovic:
Hey @brian-mulier-p! 👋

brian-mulier-p:
I'll handle it but anyway it's not the right move to fix the E2E I'll create another PR

PR #15806, opened 2026-04-30
fix(helm): remove erroneously committed chart dependency tgz files

MarcoSaba:
## Summary
The `charts/kestra/charts/` directory contained bundled `.tgz` dependency files that should not be committed — they are generated at build time and gitignored in the canonical chart source. Their presence caused `helm lint` to fail because Helm enforces that sub-charts found in `charts/` must be declared in `Chart.yaml`'s `dependencies` block.

## Changes
- Remove `charts/kestra/charts/*.tgz` from git tracking
- Add `charts/kestra/charts/` and `charts/kestra-starter/charts/` to `.gitignore`

ref kestra-io/helm-charts#284


tchiotludo:

PR #15808, opened 2026-04-30
fix(helm): fetch kestra-starter postgres dep before packaging

MarcoSaba:
## Summary
kestra-starter/Chart.yaml declares a postgres dependency (groundhog2k chart, version read dynamically from Chart.yaml) but it was never fetched during the release workflow, causing helm package charts/kestra-starter to fail.

## Changes
- Add groundhog2k helm repo and pull postgres into charts/kestra-starter/charts/ in the bundle step, before lint and package run

ref kestra-io/helm-charts#284


tchiotludo:

PR #15811, opened 2026-04-30
fix(scheduler): serialize when-condition candidate dates in trigger timezone

AcevedoR:
## Summary

- `Schedule.whenConditionMatch()` was calling `candidateOutput.toMap()` which serializes `ZonedDateTime` using `JacksonMapper.MAPPER` with `setTimeZone(TimeZone.getDefault())`
- On CI machines running UTC, midnight Paris dates like `2025-11-03T00:00+01:00` were serialized as `2025-11-02T23:00:00Z`
- `isDayWeekInMonth` then extracted the wrong local date (Nov 2 instead of Nov 3), causing `findNextDateMatchingConditions` to skip the expected Monday and advance to next month's first Monday
- The scheduler missed the scheduled execution, making the test fail only on UTC-timezone hosts

## Fix

Use `candidateOutput.toMap(ZoneId.of(this.timezone))` so candidate dates are always serialized in the trigger's configured timezone — matching the pattern already used in `Schedule.evaluate()`.

Removes `@FlakyTest` from `shouldSucceedScheduleConditionalScheduleTriggerGivenValidTimeZone` since the test is now deterministic on all timezones.

## Test plan

- [x] `shouldSucceedScheduleConditionalScheduleTriggerGivenValidTimeZone` passes under `./gradlew :scheduler:test` (no longer flaky)
- [x] Full scheduler test suite: 75 passing, 0 failing
- [x] `ScheduleTest` and `IsDayWeekInMonthFunctionTest` in core: 37 passing, 0 failing

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/trigger/Schedule.java
+++ b/core/src/main/java/io/kestra/plugin/core/trigger/Schedule.java
@@ -524,9 +524,12 @@ public class Schedule extends AbstractTrigger implements Schedulable, TriggerOut
      * schedule dates (so when can render {@code {{ trigger.date }}}).
      */
     private boolean whenConditionMatch(RunContext runContext, Output candidateOutput) throws InternalException {
+        Map<String, Object> outputMap = this.timezone != null
+            ? candidateOutput.toMap(ZoneId.of(this.timezone))
+            : candidateOutput.toMap();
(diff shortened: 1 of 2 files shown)


tchiotludo:

loicmathieu (approved):
LGTM

PR #15814, opened 2026-04-30
fix(worker): prevent kill events being lost when worker job queue is full

AcevedoR:
⚠️  full Claude PR, to review

## Summary

- **Root cause**: `WorkerJobFetcher.handleJobResponse()` calls `workerJobQueue.put(job)` which blocks when the local queue is at capacity. Because gRPC serialises stream message delivery per stream (via its internal `SerializingExecutor`), a kill event arriving as the next message was never dispatched until the blocked `put()` returned — which only happens when the queue drains, which requires the task to be killed, which requires the kill event. Deadlock.
- **Fix**: Offload the blocking `put()` to a virtual thread so the gRPC `onNext()` callback returns immediately. Kill-event responses are still processed synchronously (before the virtual-thread branch).
- Three additional hardening fixes for adjacent race windows:
  - `AbstractWorkerJobProcessor`: `killRequested` flag so a kill arriving before `callJob()` sets `currentWorkerCallable` is not silently dropped
  - `AbstractWorkerCallable.call()`: deferred self-interrupt when `kill()` ran before `call()` could set `currentThread`
  - `ExecutionKilledManager.register()`: check `killedExecutions` cache at registration time so queued tasks are killed immediately instead of running to completion
- Test: replaced `Thread.sleep(500)` with a deterministic Awaitility await on `State.Type.RUNNING`, and removed `@FlakyTest`

## Test plan

- [x] `WorkerTest#shouldKillTasksWhenExecutionKillEventReceived` passes in 4/4 consecutive local runs
- [x] Full `:worker:test` suite (53 tests) passes

### DIFF
--- a/worker/src/main/java/io/kestra/worker/fetchers/WorkerJobFetcher.java
+++ b/worker/src/main/java/io/kestra/worker/fetchers/WorkerJobFetcher.java
@@ -308,29 +308,33 @@ public class WorkerJobFetcher extends WorkerLoop {
             }
         }
 
-        List<String> acks = new ArrayList<>();
-        for (WorkerJobPayload payload : response.getJobsList()) {
-            try {
-                String jobId = payload.getJobId();
-                WorkerJob job = MessageFormats.JSON.fromByteString(payload.getJobData(), WorkerJob.class);
-
-                log.debug("Received job: {}", jobId);
-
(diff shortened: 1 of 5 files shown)


tchiotludo:

PR #15816, opened 2026-05-01
feat: add lastTriggeredDate, source, and locked query filters

### LINKED ISSUE #13884: [Filters] Triggers: add filters & operators
### Describe the issue

Implement/enable Triggers page filtering.

Filters & supported operators

- SOURCE: EQUALS

Image

- LOCKED: EQUALS

Image

- LAST TRIGGERED DATE: EQUALS (same as timeRange)

Image

- LABELS:

Image

has any of (OR)
has none of (NOT OR)
has all of (AND)
contains (LIKE)
does not contain (NOT LIKE)
is set (IS NOT NULL)
is not set (IS NULL)

- TIME_RANGE: EQUALS

### Environment

- Kestra Version: develop

ammeek:
### ✨ Description                                                                                                                                      
                                                                                                                                                          
This PR adds query filter support for the following Field, Resource and Operator Combinations.                                                            
                                                                                                                                                          
| Resource | Field               | Operator                                                                                     |
|----------|---------------------|----------------------------------------------------------------------------------------------|
| TRIGGER  | LAST_TRIGGERED_DATE | EQUALS, NOT_EQUALS, GREATER_THAN, GREATER_THAN_OR_EQUAL_TO, LESS_THAN, LESS_THAN_OR_EQUAL_TO |
| TRIGGER  | SOURCE              | EQUALS                                                                                       |
| TRIGGER  | LOCKED              | EQUALS                                                                                       |
| TRIGGER  | NEXT_TRIGGERED_DATE | EQUALS, NOT_EQUALS, GREATER_THAN, GREATER_THAN_OR_EQUAL_TO, LESS_THAN, LESS_THAN_OR_EQUAL_TO |                   


Breaking changes:
The following fields have been deprecated in favour of the new TIME_RANGE pattern introduced in 2.0.
Field.START_DATE/Field.END_DATE
                                                                                                                                    
### 🔗 Related Issue                                                                                                                                      
                                                                                                                                                          
closes: https://github.com/kestra-io/kestra/issues/13884)                                                                                   
                                                                                                                                                        
### 🎨 Frontend Checklist                                                                                                                                 
                                                                                                                                                        
### 🛠️  Backend Checklist                                                                                                                                
                                                                                                                                                          

tchiotludo:

Spec generated with EE branch `fix/trigger-query-filters`. Diff vs [client-sdk](https://github.com/kestra-io/client-sdk/blob/main/kestra-ee.yml):
```diff
21894a21895,21897
>       - SOURCE
>       - LOCKED
>       - LAST_TRIGGERED_DATE
```

MilosPaunovic:
Hey @ammeek! 👋

webserver/src/main/java/io/kestra/webserver/utils/QueryFilterUtils.java
```diff
@@ -130,4 +130,49 @@ public static List<QueryFilter> replaceTimeRangeWithComputedDateFilter(List<Quer
         validateTimeline(updatedFilters);
         return updatedFilters;
     }
+
+    public enum TriggerDateFilter {
+        NEXT_EXECUTION_DATE(QueryFilter.Field.NEXT_EXECUTION_DATE),
```
AcevedoR: wouldnt it have been possible to not create another data structure, and reuse QueryFilter with 2 dedicated fields that can be filtered for triggers specificaly ?

AcevedoR (changes requested):

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcTriggerRepository.java
```diff
@@ -329,6 +334,52 @@ public List<TriggerState> findTriggersEligibleForScheduling(ZonedDateTime now, S
...
+    protected Condition nextExecutionDateCondition(Object value, QueryFilter.Op operation) {
+        return triggerDateFieldCondition(value, operation, NEXT_EVALUATION_DATE_COLUMN, QueryFilter.Field.NEXT_EXECUTION_DATE);
+    }
+
+    private Condition triggerDateFieldCondition(Object value, QueryFilter.Op operation, String column, QueryFilter.Field field) {
+        // Accept ISO-8601 durations (e.g. PT24H) as "last N hours" — same semantics as TIME_RANGE
+        try {
+            Duration duration = value instanceof Duration d ? d : Duration.parse(value.toString());
+            ZonedDateTime threshold = ZonedDateTime.now().minus(duration);
+            return applyDateCondition(threshold.toOffsetDateTime(), QueryFilter.Op.GREATER_THAN_OR_EQUAL_TO, column);
```
AcevedoR, line 369: is it on purpose that GREATER_THAN_OR_EQUAL_TO is fixed here ?

AcevedoR (changes requested):

webserver/src/main/java/io/kestra/webserver/utils/QueryFilterUtils.java
ammeek: The challenge he is that we need to expose the restriction of only allowing NEXT_EXECUTION_DATE and LAST_TRIGGERED_DATE to the api spec we generate. Look at the available methods it looks like we are able to use the Parameter.schema annotation to do this.

Example:
``` java
@Parameter(
    description = "Which trigger date field the time interval is applied to",
    schema = @Schema(
        allowableValues = {"NEXT_EXECUTION_DATE", "LAST_TRIGGERED_DATE"}
    )
) @Nullable @QueryValue QueryFilter.Field dateFilter
```

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcTriggerRepository.java
ammeek: Yes

webserver/src/main/java/io/kestra/webserver/utils/QueryFilterUtils.java
AcevedoR: I still dont understand why you cannot just reuse the QueryFilter API like everywhere else and you had to add the dateFilter field, I must be missing something 🤔

Can you not from the UI just call either: 
- `filters[NEXT_EXECUTION_DATE][GREATER_THAN]=2026-10-10` if the user wants to filter on the NEXT_EXECUTION_DATE
- or `filters[LAST_TRIGGERED_DATE][GREATER_THAN]=2026-10-10` if the user wants to filter on the LAST_TRIGGERED_DATE 
?

ammeek: The dateFilter parameter isn't a new pattern; it's the same one already used by ExecutionController.searchExecutions, which takes `@Nullable` @QueryValue DateFilter dateFilter and resolves it via QueryFilterUtils.replaceTimeRangeWithComputedDateFilter(filters, dateFilter) against START_DATE / END_DATE / START_OR_END_DATE. Triggers does the equivalent against NEXT_EXECUTION_DATE / LAST_TRIGGERED_DATE. It was introduced by @brian-mulier-p in #15905, and the UI's shared filter widget already expects this shape when a resource exposes more than one date column (see the "Apply to" segmented control in FilterSelect.vue and the dateFilterOptions / keyLabelProvider / meta plumbing on AppliedFilter).

The reason you can't just write filters[NEXT_EXECUTION_DATE][GREATER_THAN]=… directly is that the UI's duration/time-range picker emits a relative filter — TIME_RANGE (e.g. filters[timeRange][EQUALS]=PT5M) — not an absolute timestamp. That filter is resolved server-side at request time by QueryFilterUtils.rewriteTriggerDateFilters, which uses TimeLineSearch.extractFrom(filters) + DateUtils.validateTimeline(...) to turn the duration into a concrete ZonedDateTime against the column chosen by dateFilter. Same machinery ExecutionController relies on, and the same default-START_DATE shortcut that LogController.search uses.

Resolving relative time ranges server-side is what makes URLs like "last 5 minutes" bookmarkable — if the UI baked an absolute timestamp into the query string before sending, a bookmarked "within the last 5 minutes" page would freeze on the time it was created instead of sliding forward. That's also why the rewrite logic accepts TIME_RANGE, START_DATE, and END_DATE and projects them onto the resource-specific column — the picker stays generic, the backend knows the column. The alternative (resolving on the frontend) works but loses that bookmark behavior, which is why none of the other search endpoints do it that way.

One alternative worth calling out: the UI could keep the relative TIME_RANGE in the page's URL (so the bookmark still slides forward) and only at request time resolve it to an absolute timestamp against the explicit field — i.e. rewrite filters[timeRange][EQUALS]=PT5M + dateFilter=NEXT_EXECUTION_DATE into filters[nextExecutionDate][GREATER_THAN_OR_EQUAL_TO]=2026-05-21T… before calling the API.

AcevedoR: thanks for the explanation, I didnt know this changes was added recently, I thought you were adding it ^^'

AcevedoR: we are discussing about the topic currently, I'll let you know tomorrow, by the way, did you plan to report the change to the SDK too ? (just a question)

AcevedoR: I wonder if we could not try to simplify by making the QueryFilter API accept either a duration or a timestamp for dates fields. Then the backend does the computation, and you still have a bookmarkable link

ammeek: @AcevedoR I think you are correct, allows the same filter to accept a relative and absolute time range value should be a good solution to this problem. I think this would best be handled via a follow up issue.

AcevedoR (approved):

webserver/src/main/java/io/kestra/webserver/utils/QueryFilterUtils.java
AcevedoR: yes !

PR #15838, opened 2026-05-05
fix(ui): guard DynamicScroller mounts until items are non-empty

brian-mulier-p:
Global goal of this PR is to repair E2E that were no longer passing at all for a while.

It fixes a headless Chrome layout race with vue-virtual-scroller v3.

v3 computes viewport as `el.scrollTop + el.clientHeight` with no ResizeObserver fallback. When a `DynamicScroller` mounts inside conditional rendering and `clientHeight=0`, items never render (deadlock — nothing to scroll, no scroll events, no recovery).

Fix: defer each scroller until its items array is non-empty, so the container has settled layout by mount time.

### DIFF
--- a/ui/src/components/executions/Gantt.vue
+++ b/ui/src/components/executions/Gantt.vue
@@ -35,6 +35,7 @@
                 </template>
                 <template #default>
                     <DynamicScroller
+                        v-if="filteredSeries.length > 0"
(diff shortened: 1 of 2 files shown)


tchiotludo:

Piyush-r-bhaskar (approved):
LGTM 🚀

PR #15858, opened 2026-05-05
fix(workflow): add missing closing quote in version echo statement

elevatebart:


tchiotludo:

PR #15860, opened 2026-05-05
build(deps): replace nested @popperjs/core CJS build with ESM fork globally

MilosPaunovic:
## Summary

- Promotes the `@popperjs/core → npm:@sxzz/popperjs-es@^2.11.7` npm override from bootstrap-scoped to global
- Removes the now-redundant bootstrap-specific override
- Fixes all 113 previously failing design-system unit test files (573 tests now passing)

## Root cause

The design-system package installed its own nested `node_modules/@popperjs/core` using the original CJS-only distribution (no `exports` field). When element-plus's ESM files tried to `import { placements } from "@popperjs/core"`, Node.js loaded the CJS build and reported a named-export error, crashing every design-system test suite.

The root `node_modules` was already correctly using `@sxzz/popperjs-es` (the ESM fork) via an existing bootstrap-scoped override, but that override didn't apply to element-plus inside the design-system. A global override fixes all nested installs.

## Test plan

- [ ] `npm run test:unit -- --coverage` in `ui/` passes (was 113 failed, now 0)


tchiotludo:

PR #15865, opened 2026-05-05
test(tests): fix flaky focusNode unit test by using fake timers for rAF

MilosPaunovic:
## Summary

- `mountWithChart` awaited only `setTimeout(0)` callbacks, but jsdom polyfills `requestAnimationFrame` as `setTimeout(cb, 16)` — so `poll()` (which populates `storedPositions` via `capturePositions()`) hadn't fired before `selectNode` was called
- This left `storedPositions` empty → `focusNode` skipped → `setOption` with `zoom: 1.8` never called
- The race only manifests in CI where timer scheduling is stricter

## Fix

Replace manual `setTimeout(0)` waits with `vi.useFakeTimers()` + `vi.runAllTimersAsync()` so the `requestAnimationFrame` callback fires deterministically before assertions run.

## Test plan

- [ ] `npm run test:unit` passes all 17 tests in `useDependencies.spec.ts`

### DIFF
--- a/ui/tests/unit/dependencies/composables/useDependencies.spec.ts
+++ b/ui/tests/unit/dependencies/composables/useDependencies.spec.ts
@@ -314,6 +314,7 @@ describe("useDependencies composable", () => {
             ],
             count: 3,
         });
+        vi.useFakeTimers();
         const wrapper = mount({
             template: "<div></div>",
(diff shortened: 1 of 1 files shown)


tchiotludo:

PR #15867, opened 2026-05-05
refactor: move condition variables to RunContext and update related methods

### LINKED ISSUE #15787: Remove ConditionContext variables in favor of RunContext variables
There are variables in both the ConditionContext and the RunContext making things more complex than needed.
All variables should be stored inside the RunContext instead to avoid potential rendering of expressions and forgot to pass the ConditionContext's variables to the run context at rendering time.

iitzIrFan:
### ✨ Description

Refactored variable handling by removing `ConditionContext.variables` and introducing per-evaluation conditionVariables passed explicitly to `RunContext.render(...)`. Updated scheduler and worker paths to propagate these variables consistently, ensuring correct rendering without mutating shared context.

### 🔗 Related Issue

closes: https://github.com/kestra-io/kestra/issues/15787

### 🛠️ Backend Checklist


iitzIrFan:
Note: We have not completely Remove ConditionContext variables in favor of RunContext variables, but we are keeping them outside and passes them around.

loicmathieu:
Now I'm not sure it's better without the condition context, @fhussonnois WDYT?

PR #15873, opened 2026-05-05
fix(backup): add findDistinctNamespace to KV and namespace-file metadata repos

brian-mulier-p:
- Add default `findDistinctNamespace(tenantId)` to `KvMetadataRepositoryInterface` and `NamespaceFileMetadataRepositoryInterface` (returns `List.of()` — non-breaking for external implementers)
- Override in `AbstractJdbcKvMetadataRepository` and `AbstractJdbcNamespaceFileMetadataRepository` using `select + groupBy` on `last=true` rows
- Add `findDistinctNamespace` test in abstract KV and NSFile metadata repo test classes

Companion EE PR: https://github.com/kestra-io/kestra-ee/pull/7655

### DIFF
--- a/core/src/main/java/io/kestra/core/repositories/KvMetadataRepositoryInterface.java
+++ b/core/src/main/java/io/kestra/core/repositories/KvMetadataRepositoryInterface.java
@@ -3,6 +3,7 @@ package io.kestra.core.repositories;
 import java.io.IOException;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
 import io.kestra.core.models.FetchVersion;
 import io.kestra.core.models.QueryFilter;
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/kestra/core/repositories/NamespaceFileMetadataRepositoryInterface.java
```diff
@@ -11,6 +11,10 @@
 import io.micronaut.data.model.Pageable;
 
 public interface NamespaceFileMetadataRepositoryInterface {
+    default List<String> findDistinctNamespace(String tenantId) {
```
loicmathieu: I don't see the point of having a default no-op implementation always overriden

core/src/main/java/io/kestra/core/repositories/KvMetadataRepositoryInterface.java
```diff
@@ -11,6 +11,10 @@
 import io.micronaut.data.model.Pageable;
 
 public interface KvMetadataRepositoryInterface {
+    default List<String> findDistinctNamespace(String tenantId) {
+        return List.of();
```
loicmathieu: I don't see the point of having a default no-op implementation always overriden

loicmathieu (changes requested):

loicmathieu (changes requested):
Moreover, a `Set` would be better as return type as it enforces no duplicates so it's more obvious.
I also wonder if you should not use the distrinct clause in your SQL queries

loicmathieu (approved):
LGTM

PR #15884, opened 2026-05-06
fix(core): run migrations in @PostConstruct before other beans initialize

nkwiatkowski:
## Summary
- Move all migration logic into `initOnStartup()` (`@PostConstruct`) which fires before other `@Context` beans initialize, fixing a crash on fresh EE instances where `auto=false` (default)
- Make `initOnStartup()` `final` and delegate to overridable `autoRun()` hook (Micronaut's compile-time AOP does not inherit `@PostConstruct` to overriding methods)
- Remove the `maybeRunMigrations()` / `shouldAutoMigrate()` / `run()` indirection chain from `AbstractCommand` — no longer needed

Companion PR: kestra-io/kestra-ee (same branch `fix/migration-fresh-instance`)

## Test plan
- [x] `MigrationRunnerTest` passes
- [ ] Fresh H2 instance with `auto=false`: auto-migrates and starts (was crashing with "Table not found")
- [ ] Existing instance with `auto=false`, no pending: starts normally
- [ ] Existing instance with `auto=false`, pending migrations: fails with `MigrationPendingException`
- [ ] `kestra migrate run` still works
- [ ] `kestra migrate unlock` still works

### DIFF
--- a/cli/src/main/java/io/kestra/cli/AbstractCommand.java
+++ b/cli/src/main/java/io/kestra/cli/AbstractCommand.java
@@ -20,7 +20,6 @@ import io.kestra.core.plugins.PluginManager;
 import io.kestra.core.plugins.PluginRegistry;
 import io.kestra.core.utils.Rethrow;
 import io.kestra.core.migration.MigrationRunner;
-import io.kestra.core.migration.MigrationRunnerInterface;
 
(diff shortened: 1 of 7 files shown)


loicmathieu (approved):
LGTM

PR #15890, opened 2026-05-06
feat(helm): gate GCS upload behind helm-docs, kubeconform, and image-tag checks

MarcoSaba:
## Summary

Extends `helm-release.yml` with three validation gates that must all pass before any artifact is uploaded to `gs://helm-charts-kestra-host/`:

- **helm-docs sync** (before Chart.yaml version bump) — fails if `README.md` is out of sync with `*.gotmpl` templates
- **kubeconform `-strict` validation** on packaged `.tgz` artifacts — replaces the silent `helm template > /dev/null` check with actual schema validation
- **Docker Hub image-tag check** — verifies `kestra/kestra:<appVersion>` exists before publishing a chart that references a non-existent image

Also pins Helm to `v3.12.3` (matching the validated combo from the source `helm-charts` repo) and installs `kubeconform v0.7.0`.


Skraye (approved):
LGTM

PR #15907, opened 2026-05-06
refactor(executions): throw ConflictException for state conflicts in ExecutionController

fhussonnois:
State-mutating endpoints (restart, pause, force-run, kill, resume, unqueue, change-status, ...) inconsistently threw IllegalStateException (mapped to 409) or IllegalArgumentException (mapped to 422) when the execution was in the wrong state. Replace those with ConflictException so all state conflicts return 409, and normalize the error messages to a consistent "Cannot <action>: <reason>." format.


loicmathieu (approved):
LGTM

tchiotludo:

```diff
21968a21969
>       - MCP_SERVER
```

loicmathieu:
By the way, we have the same error message inside the ExecutionService I think, you may want to align them even if they should normally not be used as we should always checked the correct state before calling it.

PR #15918, opened 2026-05-07
fix(queue): swallow RejectedExecutionException during shutdown

fhussonnois:
When the queue executor is shut down (via @PreDestroy) while a bean is still iterating over its subscribe() chain, the rejection escaped to ThreadUncaughtExceptionHandler and triggered Runtime.exit(1), pre-empting graceful shutdown. Treat it as a benign shutdown-time race: log debug, mark the subscriber ended, and return.


loicmathieu (commented):
I'm surprised the exception is on the `queueService.execute()` method as it's only called one at subscription time.

I would think the exception arrives inside the Runnable pass to the execute method so I would add a catch at line 93 before the finally that mark the end of the subscription instead.

PR #15923, opened 2026-05-07
Feat/loop features

loicmathieu:
- Add support for kill, retry, change taskrun state, restart and replay in loops (fixes https://github.com/kestra-io/kestra-ee/issues/7608)
- Change the loop output format and add a LoopOutputFunction to more easily retrieve loop outputs (part-of https://github.com/kestra-io/kestra-ee/issues/7565)
- Fix sublow in loop (fixes https://github.com/kestra-io/kestra-ee/issues/7644)


core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
```diff
@@ -106,19 +107,37 @@ public static <T extends Task & ExecutableTask<?>> Optional<SubflowExecution<?>>
...
+                        List<Execution> childExecutions;
+                        if (currentExecution.getLoopRun() != null) {
+                            // In a loop context, loopExecution IDs change on every restart so trigger.variables.executionId
+                            // is not stable. Instead, use stable loop iteration identifiers (loopTaskRunId + loopIndex)
+                            // stored in the child execution's trigger variables to find the correct one.
+                            String loopTaskRunId = currentExecution.getLoopRun().taskRunId();
+                            int loopIndex = currentExecution.getLoopRun().index();
+                            String childNamespace = runContext.render(currentTask.subflowId().namespace());
+                            String childFlowId = runContext.render(currentTask.subflowId().flowId());
+                            childExecutions = executionRepository.findByFlowId(currentExecution.getTenantId(), childNamespace, childFlowId, Pageable.UNPAGED)
```
brian-mulier-p: This is a risky performance overhead. I'd advise to AT LEAST filter by parent execution id (I think it's populated?). Ideally direct targetting via variables traversal would be better but... JSONB :)

core/src/test/java/io/kestra/plugin/core/flow/LoopCaseTest.java
```diff
@@ -509,4 +512,34 @@ public void loopSwitch(Execution execution) throws InternalException {
...
+        assertThat(execution.getState().getCurrent()).isEqualTo(State.Type.SUCCESS);
+
+        var subExecutions = executionRepository.findLoopSubExecutions(execution.getTenantId(), execution.getId());
+        assertThat(subExecutions).hasSize(2);
+
+        // for each sub-execution, check their subflow
+        var subflowExecution1 = findSubflowExecution(subExecutions.getFirst());
+        assertThat(subflowExecution1.getState().getCurrent()).isEqualTo(State.Type.SUCCESS);
+        assertThat(subflowExecution1.getTaskRunList()).hasSize(1);
+        var subflowExecution2 = findSubflowExecution(subExecutions.get(0));
```
brian-mulier-p: findSubflowExecution(subExecutions.get(1))

core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
```diff
@@ -135,9 +154,29 @@ public static <T extends Task & ExecutableTask<?>> Optional<SubflowExecution<?>>
...
+                                Map<String, Object> updatedVars = existingVars != null ? new HashMap<>(existingVars) : new HashMap<>();
+                                updatedVars.put("executionId", currentExecution.getId());
+                                updatedVars.put("taskRunId", currentTaskRun.getId());
+                                restartedChild = restartedChild.withTrigger(
+                                    ExecutionTrigger.builder()
+                                        .id(restartedChild.getTrigger().getId())
+                                        .type(restartedChild.getTrigger().getType())
+                                        .variables(updatedVars)
+                                        .logFile(restartedChild.getTrigger().getLogFile())
+                                        .build()
```
brian-mulier-p: might be better to add a toBuilder to ExecutionTrigger so that you don't need to copy over every fields and avoid it being error prone / broken upon any new members added in ExecutionTrigger?

core/src/main/java/io/kestra/plugin/core/flow/Loop.java
brian-mulier-p: I think it's false? Any sub-execution will inject that {{item}} variable so even with two-story nest it will see this item.value? Please change that description to reflect that.

However it might be a good move to specify (either through enhancing this example with a nested Loop -> Return("this will only see {{item.value}} of nested_loop, not parent_loop") so that user are aware that whenever a Loop task is invoked it discards the parent one.

brian-mulier-p (changes requested):

core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
loicmathieu: It's worth than that :( 
When restarting an execution that has a failing loop, it restarts the whole loop instead of restarting the subflow execution that fail.

I propose for now to creates a new subflow execution and wait for a proper fix about restarting only the failing loop sub-execution

core/src/test/java/io/kestra/plugin/core/flow/LoopCaseTest.java
loicmathieu: Fixed

core/src/main/java/io/kestra/core/runners/ExecutableUtils.java
loicmathieu: Done

core/src/main/java/io/kestra/plugin/core/flow/Loop.java
loicmathieu: Yes, it's the main changes, it's the same for all tasks.
I updated the doc but anyway, AJ will do a pass and enhance all this.

tchiotludo:

```diff
21968a21969
>       - MCP_SERVER
```

brian-mulier-p (approved):
LGTM

loicmathieu:
Followup issue: restarting a Loop creates new loop sub-execution instead of restarting the failing one.
https://github.com/kestra-io/kestra/issues/15945

PR #15968, opened 2026-05-11
fix: correctly copy the Kestra executable

Skraye:


AcevedoR:
I guess the CI was changed recently? explaining why its needed now ?

fhussonnois (approved):

Skraye:
> I guess the CI was changed recently? explaining why its needed now ?

We use the generated exe to generate the JSON schema of the config now

PR #15986, opened 2026-05-11
fix(test): stabilize PluginSearchCommandTest and flowConcurrencyKilled flakiness

brian-mulier-p:
- `PluginSearchCommandTest`: use random WireMock port; fixes `Failed to bind to /0.0.0.0:28181` collision with webserver tests that also hardcode port 28181.
- `flow-concurrency-queue-killed.yml`: bump sleep `PT10S → PT1M` so exec1 stays RUNNING long enough for the kill signal to arrive on slow Kafka backends.
- `TriggerSchedulingLoopTest.shouldPauseAndResumeLoopGivenPausedState`: add 500ms grace period + `clearInvocations` after `pause()` so the in-flight iteration tail is discarded before `verifyNoMoreInteractions`.

Note: `flowTriggerWithPause` flakiness is addressed in a separate PR (#15989) which fixes the root cause in `DefaultExecutor` (intermediate state transitions dropped when two state changes collapse in one executor cycle, see #15988).


ammeek (approved):
LGTM

PR #15989, opened 2026-05-11
fix(executor): emit Flow trigger for every state transition in a cycle

### LINKED ISSUE #15988: Flow trigger misses state transitions when multiple state changes collapse in one executor cycle
## Bug

A Flow trigger configured to listen on multiple states (e.g. `RUNNING`, `PAUSED`, `SUCCESS`) can miss intermediate state transitions when two state changes occur close together in the same executor processing cycle.

### Steps to Reproduce

1. Create a flow with a `Pause` task with a very short pause duration (e.g. `PT0.5S`).
2. Create a listener flow with a Flow trigger listening on `[RUNNING, PAUSED, SUCCESS]` states of the first flow.
3. Run the first flow and wait for it to complete.
4. Count the listener executions — there should be 4 (`RUNNING`, `PAUSED`, `RUNNING` on resume, `SUCCESS`).

### Expected Behavior

4 listener executions are created, one per state transition of the parent flow.

### Actual Behavior

On slow backends (e.g. PostgreSQL under CI load), only 3 listener executions are created. The `RUNNING` state after pause resume is silently skipped.

### Notes

This is intermittent — reproduces more reliably on slow environments (e.g. PostgreSQL CI) because tight timing makes two transitions land in the same executor cycle.

Test: `AbstractRunnerTest.flowTriggerWithPause` / `FlowTriggerCaseTest.triggerWithPause`

brian-mulier-p:
Fixes #15988

When two state changes collapse into a single executor cycle (e.g. a pause-resume delay fires and the executor pipeline immediately advances the execution to `SUCCESS`), `DefaultExecutor` previously fired `processFlowTriggers` only once — for the net `originalState → finalState` delta. Intermediate states were silently skipped, causing Flow trigger listeners to miss expected executions.

- `ExecutorContext` now records every distinct state in `stateTransitions` (appended on each `withExecution` call when the new state differs from the last)
- `DefaultExecutor.toExecution` iterates `stateTransitions[1..n]` and calls `processFlowTriggers` with an `execution.withState(state)` snapshot for each — making trigger evaluation independent of executor batching
- New unit tests in `ExecutorContextTest` verify transition tracking and deduplication

`AbstractRunnerTest.flowTriggerWithPause` is the regression test (was `// flaky on CI but never fail locally` — this fix makes it deterministic).

### DIFF
--- a/executor/src/main/java/io/kestra/executor/DefaultExecutor.java
+++ b/executor/src/main/java/io/kestra/executor/DefaultExecutor.java
@@ -759,9 +759,14 @@ public class DefaultExecutor extends AbstractService implements Executor {
             boolean isTerminated = executor.getFlow() != null && executionService.isTerminated(executor.getFlow(), executor.getExecution());
 
             Execution execution = executor.getExecution();
-            // handle flow triggers on state change
-            if (!execution.getState().getCurrent().equals(executor.getOriginalState())) {
-                processFlowTriggers(execution);
+            // Fire flow triggers for every distinct state transition that occurred in this cycle.
+            // A single cycle can advance through multiple states (e.g. PAUSED → RUNNING → SUCCESS
+            // when a pause-resume delay fires and the executor immediately completes the next task).
+            // Iterating stateTransitions[1..n] ensures each intermediate state reaches the trigger
+            // pipeline, regardless of how many transitions collapsed into one executor cycle.
+            List<State.Type> transitions = executor.getStateTransitions();
+            for (int i = 1; i < transitions.size(); i++) {
+                processFlowTriggers(execution.withState(transitions.get(i)));
             }
 
             // IMPORTANT: this must be done before emitting the last execution message so that all consumers are notified that the execution ends.
--- a/executor/src/main/java/io/kestra/executor/ExecutorContext.java
+++ b/executor/src/main/java/io/kestra/executor/ExecutorContext.java
@@ -41,16 +41,21 @@ public class ExecutorContext {
     private final List<SubflowExecutionResult> subflowExecutionResults = new ArrayList<>(0);
     private final List<Execution> loopExecutions = new ArrayList<>(0);
     private State.Type originalState;
+    // Tracks every distinct state this execution passes through within a single cycle.
+    // Index 0 = state at cycle entry (== originalState); each subsequent entry is a new state.
+    private final List<State.Type> stateTransitions = new ArrayList<>(1);
 
     public ExecutorContext(Execution execution) {
         this.execution = execution;
         this.originalState = execution.getState().getCurrent();
+        this.stateTransitions.add(this.originalState);
     }
 
     public ExecutorContext(Execution execution, FlowWithSource flow) {
         this.execution = execution;
         this.flow = flow;
         this.originalState = execution.getState().getCurrent();
+        this.stateTransitions.add(this.originalState);
     }
 
     public Boolean canBeProcessed() {
@@ -68,6 +73,10 @@ public class ExecutorContext {
         this.execution = execution;
         this.from.add(from);
         this.executionUpdated = true;
+        State.Type newState = execution.getState().getCurrent();
+        if (!newState.equals(stateTransitions.getLast())) {
+            stateTransitions.add(newState);
+        }
 
         return this;
     }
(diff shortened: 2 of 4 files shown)


AcevedoR (approved):

executor/src/main/java/io/kestra/executor/DefaultExecutor.java
```diff
@@ -759,9 +759,14 @@ private void toExecution(ExecutorContext executor, boolean ignoreFailure) {
...
-            if (!execution.getState().getCurrent().equals(executor.getOriginalState())) {
-                processFlowTriggers(execution);
+            // Fire flow triggers for every distinct state transition that occurred in this cycle.
+            // A single cycle can advance through multiple states (e.g. PAUSED → RUNNING → SUCCESS
+            // when a pause-resume delay fires and the executor immediately completes the next task).
+            // Iterating stateTransitions[1..n] ensures each intermediate state reaches the trigger
+            // pipeline, regardless of how many transitions collapsed into one executor cycle.
+            List<State.Type> transitions = executor.getStateTransitions();
+            for (int i = 1; i < transitions.size(); i++) {
+                processFlowTriggers(execution.withState(transitions.get(i)));
```
loicmathieu, line 769: This would generate a WARN most of the time as they would generally be a single state transition which would correspond to the current execution state (so withState will WARN).

loicmathieu: Moreover, as you add the original state to the list of transitions it may do  double processing for the same transition?

executor/src/main/java/io/kestra/executor/ExecutorContext.java
```diff
@@ -68,6 +73,10 @@ public ExecutorContext withExecution(Execution execution, String from) {
         this.execution = execution;
         this.from.add(from);
         this.executionUpdated = true;
+        State.Type newState = execution.getState().getCurrent();
```
loicmathieu, line 76: This seems totally unrelated to the purpose of this method.

loicmathieu (commented):
Overall, I would revert this change and fix the issue in the place it should be: when resuming an execution, we should process trigger (this is what we do in other places: process trigger when we change the state for ex from queued to running)

executor/src/main/java/io/kestra/executor/DefaultExecutor.java
brian-mulier-p: it starts at i=1 so no if there is only the current state it will not process it and otherwise it will process only the new states

executor/src/main/java/io/kestra/executor/ExecutorContext.java
brian-mulier-p: why ? Did you expand github when you reviewed? It's for the withExecution method so it's essentially like if you used the ExecutorContext(Execution) constructor

executor/src/main/java/io/kestra/executor/DefaultExecutor.java
loicmathieu: Didn't saw it, but as I said, it generates a WARN log which is not a good idea.

brian-mulier-p: I'm not sure to get why it would generate a WARN log?

loicmathieu: `execution.withState(transitions.get(i))` will generates a log if the state is already the same. Run the code you will see it.

Here you are updating the execution which normally should be an immutable object, that's why I'm not comptfortable with this whole change.

Instead I would have favor calling the processTrigger where it should be called (on resume I assume?)

brian-mulier-p: Well there's many cases where the scenario of multiple state in a single pass would occur tbh, markAs, pause, resume, ... I think it will be hard to maintain to go out of the regular processing for all these cases, not to mention it means we have to detect that triggers were already processed for a particular state change so it doesn't duplicate...

The gap I see in my implementation to make sure no warning log is sent would be to processFlowTrigger(execution) rather than withState if the execution is already with the same state (btw withState is creating a new execution with duplicate attribute so no mutation there and it's discarded as soon as the loop turn is done)

I have a hard time not seeing a drift over time due to potentially duplicate flow triggers being popped off if we don't stay in a single place to handle those flow triggers

PR #15995, opened 2026-05-12
fix(copilot): clarify that AI generation limit resets daily at 12:00:01 UTC

brian-mulier-p:
Closes https://github.com/kestra-io/kestra-ee/issues/7738


anna-geller (approved):
LGTM, thanks for fixing this so quickly!

PR #16009, opened 2026-05-12
fix(ui): render a visible Gantt bar for very short tasks

### LINKED ISSUE #16002: Show a minimum-width bar for very short tasks in the Gantt chart
### Describe the issue

Always show some duration bar in the Gantt chart even for super-quick tasks like log:
Image

Quick win issue

### Environment

- Kestra Version: develop

ChrisJr404:
## What

Very short tasks (`log`, expression-only tasks, etc.) currently render a Gantt row whose colored progress fill is sub-pixel and effectively invisible.

The bar wrapper already had `min-width: 5px`, but the inner Element Plus progress fill is drawn as a percentage of that wrapper, and the running portion is clamped to 10% of the bar. 10% of 5px is half a pixel, so nothing is painted.

This change:
- bumps `.task-progress` `min-width` from 5px to 24px so the bar is wide enough to host a visible colored portion
- adds `min-width: 3px` on the inner `kel-progress-bar__inner` so the colored running portion is always at least a few pixels wide regardless of how the percentage shakes out
- applies the existing `Math.max(item.width, 3)%` clamp to child tasks (`parentEndPercent` branch) too, so nested short tasks behave the same as root-level short tasks

## Before / after

Before: a `log` task in a 30s execution rendered as a 1px hairline, invisible against the row background.

After: the same task renders as a clearly visible bar in the task's state color at its correct horizontal offset.

## Testing

- `npm run build` in `ui/` passes
- `npm run check:types` passes
- `npm run test:lint` reports no new errors or warnings for the changed file

Closes #16002


Piyush-r-bhaskar:
Hey @ChrisJr404 

This has been handled in internal design revamp. Thanks for your efforts 🚀

PR #16014, opened 2026-05-12
fix(storage): parenTransversalGuard was not handling encoded characters

AcevedoR:
- fixes https://github.com/kestra-io/kestra-ee/issues/7686


brian-mulier-p (approved):
LGTM

PR #16056, opened 2026-05-15
fix(tests): increase liveness coordinator test timeouts to prevent flakiness

AcevedoR:
## Summary

- `shouldReEmitTasksWhenWorkerIsDetectedAsNonResponding` had `runningLatch.await(5s)` and `resubmitLatch.await(10s)`
- The resubmit path requires at least 8s under ideal conditions: 3s liveness timeout + 5s task re-execution — leaving only 2s of headroom under CI load
- The sibling test `shouldReEmitTasksToTheSameWorkerGroup` already uses 30s for the same reason
- `origin/develop` has already applied the same fix — this is a backport to `releases/v1.0.x`

## Test plan

- [x] `MysqlServiceLivenessCoordinatorTest.shouldReEmitTasksWhenWorkerIsDetectedAsNonResponding` passes locally with the new timeouts
- [ ] Verify CI passes on this branch


nkwiatkowski (approved):
LGTM

PR #16083, opened 2026-05-18
feat(executions): make the create execution endpoint await for execution creation

AcevedoR:
- using executionCommandQueue and a new Create event
- advance on https://github.com/kestra-io/kestra-ee/issues/7713

### DIFF
--- a/core/src/main/java/io/kestra/core/exceptions/FlowNotFoundException.java
+++ b/core/src/main/java/io/kestra/core/exceptions/FlowNotFoundException.java
@@ -3,6 +3,7 @@ package io.kestra.core.exceptions;
 import java.io.Serial;
 
 import io.kestra.core.models.executions.Execution;
+import io.kestra.core.models.executions.ExecutionId;
 
 /**
  * Exception that can be thrown when a Flow is not found.
@@ -31,5 +32,8 @@ public class FlowNotFoundException extends NotFoundException {
 
     public FlowNotFoundException(final Execution execution) {
         super(FLOW_NOT_FOUND_MESSAGE.formatted(execution.getTenantId(), execution.getNamespace(), execution.getFlowId(), execution.getFlowRevision(), execution.getId()));
+    }
+    public FlowNotFoundException(final ExecutionId executionId, Integer flowRevision) {
+        super(FLOW_NOT_FOUND_MESSAGE.formatted(executionId.tenantId(), executionId.namespace(), executionId.flowId(), flowRevision, executionId.executionId()));
     }
 }
--- a/core/src/main/java/io/kestra/core/executor/command/Replay.java
+++ b/core/src/main/java/io/kestra/core/executor/command/Replay.java
@@ -1,0 +1,62 @@
+package io.kestra.core.executor.command;
+
+import java.time.Instant;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.kestra.core.events.EventId;
+import io.kestra.core.models.executions.Execution;
+
+import jakarta.annotation.Nullable;
+import lombok.With;
+
+/**
+ * Command to replay an existing execution, optionally from a specific task run with new inputs.
+ * The new execution is created with a pre-generated {@code executionId}, and the source execution
+ * is identified by {@code sourceExecutionId}.
+ */
+public record Replay(
+    String tenantId,
+    String namespace,
+    String flowId,
+    String executionId,
+    String sourceExecutionId,
+    Instant timestamp,
+    EventId eventId,
+
+    @With @Nullable String operationId,
+    @With @JsonProperty @Nullable String taskRunId,
+    @With @JsonProperty @Nullable Integer revision,
+    @With @JsonProperty @Nullable String breakpoints,
+    @With @JsonInclude(JsonInclude.Include.NON_EMPTY) @Nullable Map<String, Object> inputs
+) implements ExecutionCommand {
+
+    /**
+     * Creates a {@code Replay} command from the given source execution and pre-generated new execution ID.
+     *
+     * @param sourceExecution the execution to replay
+     * @param newExecutionId  pre-generated ID for the new execution
+     * @param taskRunId       optional task run ID to restart from
+     * @param revision        optional flow revision override
+     * @param breakpoints     optional comma-separated breakpoint IDs
+     */
+    public static Replay from(Execution sourceExecution, String newExecutionId,
+                              @Nullable String taskRunId, @Nullable Integer revision,
+                              @Nullable String breakpoints) {
+        return new Replay(
+            sourceExecution.getTenantId(),
+            sourceExecution.getNamespace(),
+            sourceExecution.getFlowId(),
+            newExecutionId,
+            sourceExecution.getId(),
+            Instant.now(),
+            EventId.create(),
+            null,
+            taskRunId,
+            revision,
+            breakpoints,
+            sourceExecution.getInputs()
+        );
+    }
+}
--- a/core/src/main/java/io/kestra/core/scheduler/service/DefaultTriggerExecutionPublisher.java
+++ b/core/src/main/java/io/kestra/core/scheduler/service/DefaultTriggerExecutionPublisher.java
@@ -4,7 +4,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.kestra.core.events.CrudEvent;
+import io.kestra.core.executor.command.Create;
+import io.kestra.core.executor.command.ExecutionCommand;
 import io.kestra.core.models.executions.Execution;
+import io.kestra.core.models.executions.ExecutionId;
 import io.kestra.core.models.flows.State;
 import io.kestra.core.queues.DispatchQueueInterface;
 import io.kestra.core.queues.QueueException;
@@ -20,31 +23,44 @@ public class DefaultTriggerExecutionPublisher implements TriggerExecutionPublish
     private static final Logger LOG = LoggerFactory.getLogger(DefaultTriggerExecutionPublisher.class);
 
     private final ApplicationEventPublisher<CrudEvent<Execution>> executionEventPublisher;
-    private final DispatchQueueInterface<Execution> executionQueue;
+    private final DispatchQueueInterface<ExecutionCommand> executionCommandQueue;
     private final RunContextLoggerFactory runContextLoggerFactory;
 
     @Inject
     public DefaultTriggerExecutionPublisher(ApplicationEventPublisher<CrudEvent<Execution>> executionEventPublisher,
-        DispatchQueueInterface<Execution> executionQueue,
+        DispatchQueueInterface<ExecutionCommand> executionCommandQueue,
         RunContextLoggerFactory runContextLoggerFactory) {
         this.executionEventPublisher = executionEventPublisher;
-        this.executionQueue = executionQueue;
+        this.executionCommandQueue = executionCommandQueue;
         this.runContextLoggerFactory = runContextLoggerFactory;
     }
 
     public void send(final Execution execution) {
         try {
-            this.executionQueue.emit(execution);
-            this.executionEventPublisher.publishEvent(CrudEvent.create(execution));// FIXME probably a bug, should not exist here
+            Create cmd = toCreate(execution);
+            // If the trigger already produced a terminal execution (e.g. input rendering failed),
+            // preserve that state so the executor doesn't restart it from CREATED.
+            if (execution.getState().isTerminated()) {
+                cmd = cmd.withStateType(execution.getState().getCurrent());
+            }
+            this.executionCommandQueue.emit(cmd);
         } catch (QueueException e) {
             try {
                 Execution failedExecution = fail(execution, e);
-                this.executionQueue.emit(failedExecution);
-                this.executionEventPublisher.publishEvent(CrudEvent.create(execution));// FIXME same
+                this.executionCommandQueue.emit(toCreate(failedExecution).withStateType(State.Type.FAILED));
             } catch (QueueException ex) {
                 LOG.error("Unable to emit the execution", ex);
             }
         }
+    }
+
+    private Create toCreate(Execution execution) {
+        return Create.of(new ExecutionId(execution.getTenantId(), execution.getNamespace(), execution.getFlowId(), execution.getId(), execution.getFlowRevision()))
+            .withKind(execution.getKind())
+            .withTrigger(execution.getTrigger())
+            .withLabels(execution.getLabels())
+            .withInputs(execution.getInputs())
+            .withScheduleDate(execution.getScheduleDate());
     }
 
     private Execution fail(Execution message, Exception e) {
--- a/executor/src/main/java/io/kestra/executor/DefaultExecutor.java
+++ b/executor/src/main/java/io/kestra/executor/DefaultExecutor.java
@@ -1,15 +1,5 @@
 package io.kestra.executor;
 
-import java.time.Duration;
-import java.time.Instant;
-import java.util.*;
-import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
-import org.slf4j.event.Level;
-
 import io.kestra.core.contexts.KestraContext;
 import io.kestra.core.exceptions.DeserializationException;
 import io.kestra.core.exceptions.FlowNotFoundException;
@@ -18,9 +8,11 @@ import io.kestra.core.executor.command.ExecutionCommand;
 import io.kestra.core.killswitch.EvaluationType;
 import io.kestra.core.killswitch.KillSwitchService;
 import io.kestra.core.metrics.MetricRegistry;
-import io.kestra.core.models.Label;
 import io.kestra.core.models.executions.*;
-import io.kestra.core.models.flows.*;
+import io.kestra.core.models.flows.Concurrency;
+import io.kestra.core.models.flows.FlowInterface;
+import io.kestra.core.models.flows.FlowWithSource;
+import io.kestra.core.models.flows.State;
 import io.kestra.core.models.flows.sla.ExecutionMonitoringSLA;
 import io.kestra.core.models.flows.sla.SLA;
 import io.kestra.core.models.flows.sla.Violation;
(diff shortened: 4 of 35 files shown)


core/src/main/java/io/kestra/core/exceptions/FlowNotFoundException.java
```diff
@@ -32,4 +33,7 @@ public FlowNotFoundException(final String message) {
     public FlowNotFoundException(final Execution execution) {
         super(FLOW_NOT_FOUND_MESSAGE.formatted(execution.getTenantId(), execution.getNamespace(), execution.getFlowId(), execution.getFlowRevision(), execution.getId()));
     }
+    public FlowNotFoundException(final ExecutionId executionId, Integer flowRevision) {
```
loicmathieu, line 36: Why is flow revision not inside the new `ExecutionId` ?

executor/src/main/java/io/kestra/executor/DefaultExecutor.java
```diff
@@ -363,7 +368,40 @@ private void executionCommandQueue(Either<ExecutionCommand, DeserializationExcep
             }
         }
 
-        Optional<ExecutorContext> maybeExecutor = executionCommandMessageHandler.handle(message);
+        Optional<ExecutorContext> maybeExecutor = Optional.empty();
```
loicmathieu: I would do it differently:
- Creating the execution inside the state store should move to the ExecutionCommandMessageHandler or the ExecutionService
- After calling the handler, you can send back the async operation processed event as the event is processed
- The optimization to call the executionEventMessageHandler can be generalized to all commands

I wonder also if we should remove/deprecate or refactor the current execution queue handling. Maybe it should send a Create event to avoid duplication?

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -661,12 +662,13 @@ public Publisher<ExecutionResponse> createExecution(
         @Parameter(description = "Specific execution kind") @QueryValue Optional<ExecutionKind> kind) {
         Flow flow = flowService.getFlowIfExecutableOrThrow(tenantService.resolveTenant(), namespace, id, revision);
         List<Label> parsedLabels = parseLabels(labels);
-        final Execution current = Execution.newExecution(flow, null, parsedLabels, scheduleDate).toBuilder()
+
+
+        final Execution dummyExecutionOnlyForInputs = Execution.newExecution(flow, null, parsedLabels, scheduleDate).toBuilder()
```
loicmathieu: This dummy execution is not ideal, why not storing the full execution object inside the Create command?

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/exceptions/FlowNotFoundException.java
AcevedoR: done

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
AcevedoR: I created a new function to get inputs from an executionid, we dont want anymore to pass a full execution when it's not needed, thats why we try to keep Create as small and precise as possible

loicmathieu: Yeah but copying all usefull fields from exec to Create then to exec seems a loss of time with a high risk of being buggy when you add a field in the exec and forgot to copy it in Create.

So even if I understood the rational, here, it didn't makes sense  because what you want to create is an Execution, and you have it just here!

executor/src/main/java/io/kestra/executor/DefaultExecutor.java
AcevedoR: I removed usaged of execution queue in the ExecutionController + I extracted the state store creation and other logic into CreateCommandHandler, WDYT ?

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
AcevedoR: it's not a loss of time, it's insuring clear boundaries between components requesting the creation of an Execution via the Create comment event VS the executor processing this event and creating an Execution. I already started reworking other usages of the executionsQueue, soon I hope we will only have one clear entrypoint in the Executor to create Executions :).
Also thanks to having a dedicated event object, we now for sure what is transiting inside the queue, and was is the contract to create a new execution (from outside the executor)

core/src/main/java/io/kestra/core/executor/command/Replay.java
```diff
@@ -0,0 +1,62 @@
...
+
+import jakarta.annotation.Nullable;
+import lombok.With;
+
+/**
+ * Command to replay an existing execution, optionally from a specific task run with new inputs.
+ * The new execution is created with a pre-generated {@code executionId}, and the source execution
+ * is identified by {@code sourceExecutionId}.
+ */
+public record Replay(
```
loicmathieu, line 19: Replay creates a new executionId so it's not necessary to have a separate record for it.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -682,72 +670,86 @@ public Publisher<ExecutionResponse> createExecution(
...
 
-                try {
-                    // inject the traceparent into the execution
+                    if (Check.Behavior.FAIL_EXECUTION.equals(behavior)) {
+                        Logs.logExecutionId(createCommand.executionFullId(), log, Level.WARN, "Flow execution failed because one or more condition checks evaluated to false.");
+                        createCommand = createCommand.withStateType(State.Type.FAILED);
+                    }
+
+                    // inject the traceparent from the current OTel context into the command so it's propagated to the execution
+                    Map<String, String> traceCarrier = new HashMap<>();
```
loicmathieu, line 712: Ideally you should create a CreateTestMapSetter to mimic what was done inside the ExecutionTestMapSetter, this is OTEL idiomatic.
it can be done later.
By the way, is the ExecutionTestMapSetter still in use?

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -682,72 +670,86 @@ public Publisher<ExecutionResponse> createExecution(
...
-                    eventPublisher.publishEvent(CrudEvent.create(executionWithInputs));
+                    Create finalCreateCommand = createCommand;
+                    return awaitBlockingAction(
+                        executionId, "Create",
+                        operationId -> executionCommandQueue.emit(finalCreateCommand.withOperationId(operationId))
+                    ).flatMap(res -> {
+                        var executionUrl = executionUrl(finalCreateCommand.executionFullId());
+                        if(res.status().getCode() != 200){
+                            // early return, could not even acknowledge creation command
+                            // FIXME return an error here
```
loicmathieu: So Mono.error()?

executor/src/main/java/io/kestra/executor/DefaultExecutor.java
```diff
@@ -740,6 +740,7 @@ private void toExecution(ExecutorContext executor, boolean ignoreFailure) {
                 // We need to detect that and reset them as they will never reach the reset code later on this method.
                 if (execution.getTrigger() != null && execution.getState().isFailed() && ListUtils.isEmpty(execution.getTaskRunList())) {
                     sendTriggerExecutionTerminated(execution);
+                    this.followExecutionEventQueue.emitAsync(new FollowExecutionEvent(execution, ExecutionEventType.TERMINATED));
```
loicmathieu, line 673: Seems unrelated, can you explain what it is for?

executor/src/main/java/io/kestra/executor/handler/ReplayCommandHandler.java
```diff
@@ -0,0 +1,93 @@
...
+import io.kestra.executor.ExecutionStateStore;
+import io.kestra.executor.ExecutorContext;
+
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+import lombok.extern.slf4j.Slf4j;
+
+@Singleton
+@Slf4j
+public class ReplayCommandHandler {
```
loicmathieu: I think this is not necessary.
Replay creates a new execution so you can call replay from the ExecutionController then send a Create command. The executor didn't need to know that the new execution has been created from a replay as it would anyway treat it as any new execution.

executor/src/main/java/io/kestra/executor/handler/ExecutionCommandMessageHandler.java
```diff
@@ -30,22 +30,34 @@ public class ExecutionCommandMessageHandler implements ExecutorMessageHandler<Ex
     private final FlowMetaStoreInterface flowMetaStore;
     private final TaskOutputService taskOutputService;
     private final AsyncOperationService asyncOperationService;
+    private final CreateCommandHandler createCommandHandler;
```
loicmathieu: This is not idiomatic and didn't match how we treat other commands, this indicate that something is not clearly bounded here.

executor/src/main/java/io/kestra/executor/handler/CreateCommandHandler.java
```diff
@@ -0,0 +1,71 @@
...
+            var flow = flowMetaStore
+                .findById(createCommand.tenantId(), createCommand.namespace(), createCommand.flowId(), Optional.ofNullable(createCommand.flowRevision()))
+                .orElseThrow(() -> new FlowNotFoundException(createCommand.executionFullId(), createCommand.flowRevision()));
+
+            var newExecution = executionService.create(createCommand, flow);
+
+            // Persist the execution before processing the state-machine event.
+            // Any persistence failure is a hard error — the execution must exist in the DB
+            // before we signal success back to the caller.
+            executionStateStore.create(newExecution);
```
loicmathieu: I will move this line inside the execution service so we can avoid this command handler alltogether and do everything inside the main `ExecutionCommandHandler`.

As for the optimization to directly call the `ExecutionEventMessageHandler`, I think this can be moved to the `DefaultExecutor.executionCommandQueue()` itself as every time we receive a command for an execution, we should try to process it except if already terminated (which can be tested easily).

There is also an issue by using the `ExecutionCommand` queue, if the execution is kill switched, it would not be stored inside the DB which was the case before...

loicmathieu (changes requested):

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
AcevedoR: added a todo

executor/src/main/java/io/kestra/executor/DefaultExecutor.java
AcevedoR: I dont remember what test failure I had, but I was in a case where I never received the followExecutionEventQueue, because we have a early retrun

executor/src/main/java/io/kestra/executor/handler/ExecutionCommandMessageHandler.java
AcevedoR: I changed it liked you asked to only have all in ExecutionCommandMessageHandler

core/src/main/java/io/kestra/core/scheduler/service/DefaultTriggerExecutionPublisher.java
```diff
@@ -20,33 +23,42 @@ public class DefaultTriggerExecutionPublisher implements TriggerExecutionPublish
...
-        this.executionQueue = executionQueue;
+        this.executionCommandQueue = executionCommandQueue;
         this.runContextLoggerFactory = runContextLoggerFactory;
     }
 
     public void send(final Execution execution) {
         try {
-            this.executionQueue.emit(execution);
+            this.executionCommandQueue.emit(toCreate(execution));
             this.executionEventPublisher.publishEvent(CrudEvent.create(execution));// FIXME probably a bug, should not exist here
```
loicmathieu: Can you remove the call to the executionEventPublisher (the second also), it is indead a bug as it cannot work.

executor/src/main/java/io/kestra/executor/DefaultExecutor.java
loicmathieu: OK

webserver/src/main/java/io/kestra/mcp/McpToolService.java
```diff
@@ -132,7 +135,11 @@ private Mono<Execution> runFlowForMcpTask(
         )));
 
         try {
-            executionQueue.emit(execution);
+            executionCommandQueue.emit(Create.of(new ExecutionId(execution.getTenantId(), execution.getNamespace(), execution.getFlowId(), execution.getId(), execution.getFlowRevision()))
```
loicmathieu, line 141: Creating a Create command from an execution is duplicated in a lot of places, can you add a dedicated factory method inside the Create command to avoid duplication?

loicmathieu (approved):
There are still a bunch of points that I'm not fan of:
- I still think a Replay command is not necessary as you can reuse the Create command (a replay creates a new execution).
- I'm not fan of how you implement the Create inside the handler: an handler that calls another handler is a bad smell, I would have call the other handler inside the DefaultExecutor (as was done inside the executionQueue method)
- I still see a lot of double-mapping Execution -> Create -> Execution: again a smell, still think it would have been better to add the Execution object directly inside the command.
- The executionQueue still exist....

core/src/main/java/io/kestra/core/scheduler/service/DefaultTriggerExecutionPublisher.java
AcevedoR: done

core/src/main/java/io/kestra/core/scheduler/service/DefaultTriggerExecutionPublisher.java
```diff
@@ -20,33 +23,46 @@ public class DefaultTriggerExecutionPublisher implements TriggerExecutionPublish
...
 
     public void send(final Execution execution) {
         try {
-            this.executionQueue.emit(execution);
-            this.executionEventPublisher.publishEvent(CrudEvent.create(execution));// FIXME probably a bug, should not exist here
+            Create cmd = toCreate(execution);
+            // If the trigger already produced a terminal execution (e.g. input rendering failed),
+            // preserve that state so the executor doesn't restart it from CREATED.
+            if (execution.getState().isTerminated()) {
+                cmd = cmd.withStateType(execution.getState().getCurrent());
```
loicmathieu, line 44: So, well, this is not a Create execution right?

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/scheduler/service/DefaultTriggerExecutionPublisher.java
AcevedoR: It's a create directly in terminated state because of for example failed input, seems legit to call this a create anyway no ?

loicmathieu: Well, you can ask a friend but for me, it's not logical...

nkwiatkowski: Performance wise, I don't think adding this if here for something that will almost never happen is a good idea unless it's fixing a bug.

executor/src/main/java/io/kestra/executor/handler/ExecutionCommandMessageHandler.java
```diff
@@ -90,6 +111,112 @@ public Optional<ExecutorContext> handle(ExecutionCommand message) {
...
+            );
+            outcome = AsyncOperationProcessedEvent.Outcome.FAILED;
+            error = e.getMessage();
+            return Optional.empty();
+        } finally {
+            asyncOperationService.emitProcessedIfAsync(command, command.tenantId(), command.executionId(), outcome, error);
+        }
+    }
+
+    private Optional<ExecutorContext> handleReplay(Replay command) {
```
loicmathieu, line 166: Miss kill switch, again, as said, I didn't see the added value of a Replay command, better to replay the exec in the endpoint and send a Create command

executor/src/main/java/io/kestra/executor/handler/ExecutionCommandMessageHandler.java
```diff
@@ -90,6 +111,112 @@ public Optional<ExecutorContext> handle(ExecutionCommand message) {
...
+            var flow = flowMetaStore
+                .findById(command.tenantId(), command.namespace(), command.flowId(), Optional.ofNullable(command.flowRevision()))
+                .orElseThrow(() -> new FlowNotFoundException(command.executionFullId(), command.flowRevision()));
+
+            var newExecution = executionService.create(command, flow);
+
+            // Evaluate the kill switch before persisting: a brand-new execution has no running
+            // tasks, so we can write it directly in its terminal state without going through
+            // the KILLING intermediate state or the kill queue.
+            EvaluationType evaluationType = killSwitchService.evaluate(newExecution);
```
loicmathieu: kill switch code is duplicated with the DefaultExecutor and didn't reuse the same log messages

loicmathieu (changes requested):

executor/src/main/java/io/kestra/executor/KillSwitchActionService.java
loicmathieu, line 1: Miss a test

executor/src/main/java/io/kestra/executor/handler/ExecutionCommandMessageHandler.java
```diff
@@ -30,22 +37,51 @@ public class ExecutionCommandMessageHandler implements ExecutorMessageHandler<Ex
...
+
+        // For existing executions, check kill switch before taking the lock.
+        EvaluationType evaluationType = killSwitchService.evaluate(message);
+        if (evaluationType != EvaluationType.PASS) {
+            var execution = executionStateStore.findById(message.executionId());
+            if (execution != null && evaluationType.isKillSwitched(execution)) {
+                killSwitchActionService.handle(evaluationType, execution.getTenantId(), execution.getId());
+                return Optional.empty();
+            }
+        }
```
loicmathieu, line 83: Kill switch would be better evaluated inside the Defaultexecutor

loicmathieu (approved):

executor/src/main/java/io/kestra/executor/KillSwitchActionService.java
AcevedoR: done !

executor/src/main/java/io/kestra/executor/handler/ExecutionCommandMessageHandler.java
AcevedoR: killswitch done in this case

PR #16092, opened 2026-05-18
fix(tests): fix macOS-specific test failures in local-files and echo commands

AcevedoR:
## Summary

- **`LocalPathFactory`**: resolve allowed paths via `toRealPath()` at construction time so that symlinks (e.g. `/tmp` → `/private/tmp` on macOS) are accounted for when checking file authorization. Same resolution applied to plugin-level allowed paths. Without this, `checkPath()` resolved the _file_ URI through `toRealPath()` but compared it against raw unresolved strings, causing false rejections on macOS.
- **6 test `createFile()` helpers**: explicitly pass `new File("/tmp")` to `File.createTempFile()` so temp files land under `/tmp` (which resolves to `/private/tmp`) rather than the macOS default `java.io.tmpdir` (`/private/var/folders/...`), which has no relation to the `/tmp` allowed path used by the tests.
- **`echo -n` → `printf '%s'`**: macOS `/bin/echo` does not support the `-n` flag and prints it literally, producing `-n text` instead of `text`. Replaced with `printf '%s'` which is POSIX-portable and behaves identically on both macOS and Linux.

## Test plan

- [ ] `URIFetcherTest.shouldFetchFromLocalFileWhenAllowedGlobally`
- [ ] `URIFetcherTest.shouldFetchFromLocalFileWhenAllowedForPlugin`
- [ ] `FilesServiceTest.localFileAsInputFile`
- [ ] `FileExistsFunctionTest.shouldSucceedProcessingAllowedFile`
- [ ] `FileSizeFunctionTest.shouldSucceedProcessingAllowedFile`
- [ ] `IsFileEmptyFunctionTest.shouldSucceedProcessingAllowedFile`
- [ ] `ReadFileFunctionTest.shouldSucceedProcessingAllowedFile`
- [ ] `ProcessTest.inputAndOutputFiles`
- [ ] `CommandsWrapperOutputFilesTest.shouldResolveOutputFilesPebbleExpressionInScript`

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/LocalPathFactory.java
+++ b/core/src/main/java/io/kestra/core/runners/LocalPathFactory.java
@@ -17,13 +17,23 @@ import jakarta.inject.Singleton;
 
 @Singleton
 public class LocalPathFactory {
-    private final List<String> globalAllowedPaths;
+    private final List<Path> globalAllowedPaths;
 
     @Inject
     public LocalPathFactory(LocalFilesConfiguration localFilesConfiguration) {
         this.globalAllowedPaths = localFilesConfiguration.allowedPaths() != null
-            ? localFilesConfiguration.allowedPaths()
+            ? localFilesConfiguration.allowedPaths().stream().map(LocalPathFactory::resolveAllowedPath).toList()
             : Collections.emptyList();
+    }
+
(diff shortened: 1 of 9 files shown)


nkwiatkowski (approved):
LGTM

PR #16098, opened 2026-05-18
Feat: Worker Group 2.0

fhussonnois:
**Core**
  - Replace WorkerGroup task field with WorkerSelector (selector by tags + match strategy: ALL/ANY, + WorkerQueueFallback)
  - Add Rfc1123Label validation for queue tags
  - Add WorkerQueueRouting model (tag → queue resolution with disposition)
  - Rename WorkerGroupMetaStore → WorkerQueueMetaStore; rename WorkerGroupService → WorkerQueueService
  - Add NoMatchingWorkerQueueException
  - Expose WorkerQueues / WorkerGroups helpers and QueueSubscription model

**Worker**
  - Tag-based worker queue subscription on connect (QueueSubscription)
  - Per-worker metrics isolated with worker_group tag; exclude per-task gauges from heartbeat
  - WorkerQueueRegistry / MonitoredWorkerQueue updates to support multi-queue fetch

**Controller (worker-controller)**
  - New WorkerJobDispatcher rewrite for tag-routed dispatch
  - Add WorkerCapacityPolicy + SinglePoolCapacityPolicy + WorkerCapacityPolicyFactory
  - Add WorkerCapacityMetricsPublisher (live capacity metrics)
  - Add WorkerQueueResolver, WorkerLifecycleListener, extended WorkerStreamContext
  - Proto updates: worker_controller.proto, connect_controller.proto

**Executor / Scheduler**
  - Minor adjustments to event handlers and liveness coordinator to align with new queue routing


core/src/main/java/io/kestra/core/models/tasks/WorkerSelector.java
```diff
@@ -0,0 +1,53 @@
...
+    @Schema(description = "Strategy when no worker is available. Defaults to FAIL when null.")
+    WorkerQueueFallback fallback
+) {
+    /**
+     * Backwards-compatible 2-argument constructor. Equivalent to
+     * {@code new WorkerSelector(tags, null, fallback)}.
+     */
+    public WorkerSelector(List<String> tags, WorkerQueueFallback fallback) {
+        this(tags, null, fallback);
+    }
```
loicmathieu, line 44: It's a new feature, I don't understand it's backward compat to what...

core/src/main/java/io/kestra/core/worker/WorkerGroups.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Shared constants for the Worker Group concept — the well-known service-property
+ * key under which a worker publishes its resolved group id, and the default sentinel
+ * id used when no specific group is resolved.
+ */
+public final class WorkerGroups {
+
+    /** Default Worker Group id sentinel, used when no specific group is resolved. */
+    public static final String DEFAULT_ID = "default";
```
loicmathieu, line 13: To avoid potential clashes with what a user may want to use, to may want to use a reserved prefix for such constant, for example `_default` and use a regex to avoid starting a worker group with an underscore at creation time.

core/src/main/java/io/kestra/core/worker/WorkerQueues.java
```diff
@@ -0,0 +1,103 @@
...
+ * readable log messages, without depending on any concrete Worker Queue entity.
+ */
+public final class WorkerQueues {
+
+    /**
+     * Reserved Worker Queue id used as a sentinel for "the global default queue". No
+     * Worker Queue row is ever persisted with this id; subscriptions and resolved
+     * routings carry it directly to mean "dispatch to the default queue".
+     */
+    public static final String DEFAULT_ID = "default";
```
loicmathieu, line 22: Same here, if possible use _default or <default> or something else that cannot be used by a user

worker-controller/src/main/java/io/kestra/controller/grpc/services/SinglePoolCapacityPolicy.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Trivial capacity policy: a single shared pool of {@code maxConcurrency}
+ * interchangeable slots, no per-queue accounting and no borrowing.
+ *
+ * <p>This is the default policy. Deployments that need per-queue reservations
+ * or capacity borrowing replace the {@link WorkerCapacityPolicyFactory} bean
+ * with one that returns a richer implementation.
+ */
+@ThreadSafe
+final class SinglePoolCapacityPolicy implements WorkerCapacityPolicy {
```
loicmathieu, line 20: I'm not sure what this class is about?

loicmathieu (commented):
Executor-related code LGTM.

I wonder if we could use a different term than WorkerQueue as it resonates with our queue, which may lead to misunderstanding.

You can no longuer start a worker with a groupId via command line argument? Your call but I think it was convenient.

As for the Worker code, I have the impression that now there is another call between the Worker and the Controller for each task execution. This might have a performance impact so I wonder if it can be combined inside an existing call (when you send an event on job termination to re-compute permits/reservations).

ammeek (approved):
LGTM

PR #16099, opened 2026-05-18
feat: add DeleteImage task to remove Docker images

Skraye:
EDIT: Adds a standalone DeleteImage task instead of a deleteImage property on the Docker runner, as suggested in review

### DIFF
--- a/script/src/main/java/io/kestra/plugin/scripts/runner/docker/DeleteImage.java
+++ b/script/src/main/java/io/kestra/plugin/scripts/runner/docker/DeleteImage.java
@@ -1,0 +1,119 @@
+package io.kestra.plugin.scripts.runner.docker;
+
+import com.github.dockerjava.api.DockerClient;
+import com.github.dockerjava.api.exception.NotFoundException;
+
+import io.kestra.core.models.annotations.Example;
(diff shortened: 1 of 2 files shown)


loicmathieu (commented):
I understand the rational but it's not very logical to create an image with the `docker.Build` task, use it with the `Docker` task runner and delete it after usage from the task runner.
It think it would be better to use a `docker.Delete` task, this boolean seems like a good way to shoot you in the foot.

Anyway, @kestra-io/plugins should review it

fdelbrayelle:
Quite agree with you @loicmathieu. Could you share the initial issue to have the initial context @Skraye :pray: ?

Skraye:
Updated with a new task as you suggested to delete an image 

@fdelbrayelle There is no issue opened but it relates to https://github.com/kestra-io/devops-cloud-instances/issues/62

loicmathieu:
This should be inside the Docker plugin instead of the core plugin.

fdelbrayelle:
Yes accordingly to your context and your doc ("This is useful when images are built dynamically (e.g. via `docker.Build`) and should not"), you should open a PR on plugin-docker not on the core :+1:

Skraye:
My bad, I will come back in the good repo!

PR #16102, opened 2026-05-19
fix(topology): stop click propagation on all Controls buttons

fdelbrayelle:
## Summary
- `ControlButton` clicks were bubbling up to parent `router-link` elements, causing the router to navigate to the nearest tab route (e.g. `/edit`) and resulting in a 404 page.
- Added `.stop` modifier to all `ControlButton @click` handlers in `Topology.vue`.

## Test plan
- [ ] Click the info icon (bottom-left Controls) in the topology view — should toggle extra details, not redirect to 404
- [ ] Click fit-view, toggle-orientation, and export buttons — no unwanted navigation

### DIFF
--- a/ui/packages/topology/src/Topology.vue
+++ b/ui/packages/topology/src/Topology.vue
@@ -92,16 +92,16 @@
         </template>
 
         <Controls v-if="controlsShown" :showInteractive="false" :showFitView="false">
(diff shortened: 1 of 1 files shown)


fdelbrayelle:
Closing — fix should go into the existing PR #15400 instead.

PR #16110, opened 2026-05-19
feat(core): add InputStream/OutputStream-based methods to FileSerde

nkwiatkowski:
## Summary
- Adds `InputStream`-based `readAll()` and `reader()` methods to `FileSerde` that delegate to the existing `Reader`-based implementations
- Adds `OutputStream`-based `writeAll()` that delegates to the existing `Writer`-based implementation
- **No behavior change** — purely additive forward-compatibility layer so plugins compiled against these method signatures work on both 1.x and 2.0

## Context
In 2.0, `FileSerde` will switch to binary ION for storage writes (~20-40% size reduction). Binary ION requires `InputStream`/`OutputStream` (not `Reader`/`Writer`) to avoid byte corruption from character decoding. Adding these signatures on 1.x means plugins can migrate to the new API now and remain compatible with both versions.

Companion PRs for v1.2.x, v1.1.x, and v1.0.x will follow (same commit, cherry-picked).

## Test plan
- [x] `./gradlew :core:compileJava` passes
- [ ] CI passes on all release branches

### DIFF
--- a/core/src/main/java/io/kestra/core/serializers/FileSerde.java
+++ b/core/src/main/java/io/kestra/core/serializers/FileSerde.java
@@ -198,6 +198,56 @@ public final class FileSerde {
         }
     }
 
+    // region InputStream-based read methods (compatible with both text and binary ION in 2.0+)
+
+    /**
+     * For performance, it is advised to wrap the input stream inside a BufferedInputStream, see {@link #BUFFER_SIZE}.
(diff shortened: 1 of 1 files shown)


loicmathieu (approved):
LGTM, warning, if you rename the methods in the main PR don't forgot to rename them here also.

PR #16118, opened 2026-05-19
fix(triggers): expose Schedule timezone autocomplete in JSON schema

fhussonnois:
Schedule trigger's `timezone` field had no enum in the generated JSON schema, so Monaco YAML autocomplete only suggested the JVM default timezone (e.g. when running with -Duser.timezone=Asia/Tokyo, only Asia/Tokyo was offered).

Teach JsonSchemaGenerator to recognize the existing @TimezoneId validation annotation and emit
`anyOf: [{enum: getAvailableZoneIds() U SHORT_IDS}, {pattern: offsetRegex}]` so the editor offers full timezone autocomplete while still accepting offset-style values (`+02:00`, `GMT-05:00`, `Z`) that ZoneId.of() parses.

Apply @TimezoneId to Schedule.timezone, which both drives the new schema enum and adds runtime validation for invalid timezone strings.

Refs kestra-ee#7576

### DIFF
--- a/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
+++ b/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
@@ -3,9 +3,11 @@ package io.kestra.core.docs;
 import java.lang.reflect.*;
 import java.time.Duration;
 import java.time.LocalTime;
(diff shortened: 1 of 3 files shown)


ammeek (approved):
LGTM

PR #16143, opened 2026-05-20
fix(core): prevent path traversal in inputFiles filename resolution

AcevedoR:
## Summary

- `FilesService.inputFiles()` was constructing `File` objects directly from rendered filenames with `new File(workingDir, renderedName)`, allowing keys like `../escape.txt` to escape the task working directory sandbox
- Routes filename resolution through the existing `workingDir().resolve()` guard, which already rejects paths that escape the base directory (checks both `../` segments and absolute path escape via `resolved.startsWith(baseDir)`)
- Adds two regression tests covering the traversal and absolute-path cases

Fixes [GHSA-q3fw-mvgv-pjr2](https://github.com/kestra-io/kestra/security/advisories/GHSA-q3fw-mvgv-pjr2) (medium severity, CVSS 6.5 — arbitrary file write on worker host).

> **Note:** This advisory is currently in draft. Consider publishing it alongside this PR merge, or coordinating a patch release.

## Test plan

- [x] `shouldRejectPathTraversalInInputFileNames` — `../escape.txt` key throws `IllegalArgumentException`
- [x] `shouldRejectAbsolutePathInInputFileNames` — `/etc/passwd` key throws `IllegalArgumentException`
- [x] All 10 existing `FilesServiceTest` tests still pass

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FilesService.java
+++ b/core/src/main/java/io/kestra/core/runners/FilesService.java
@@ -38,7 +38,7 @@ public abstract class FilesService {
         inputFiles
             .forEach(throwBiConsumer((fileName, input) ->
             {
-                var file = new File(runContext.workingDir().path().toString(), runContext.render(fileName, additionalVars));
+                var file = runContext.workingDir().resolve(Path.of(runContext.render(fileName, additionalVars))).toFile();
 
                 if (!file.getParentFile().exists()) {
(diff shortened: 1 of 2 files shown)


AcevedoR:
I guess we want to prevent too absolute path @loicmathieu right ? /etc/passwd

loicmathieu (approved):
LGTM

PR #16145, opened 2026-05-20
feat(design-system): export color palette CSS vars, add missing EP wrappers and fix build exports

Skraye:
- Expose full color palette as --ks-{color}-{shade} custom properties
- Add KsSubMenu, KsPageHeader, KsDescriptions, KsCarousel, KsResult, KsBacktop wrappers
- Add bootstrap containers and grid to styleBase
- Fix tsdown export keys to always use ./ prefix


ui/packages/design-system/tsdown.config.ts
```diff
@@ -17,7 +17,7 @@ function findVueFiles(dir: string): string[] {
 const componentsDir = path.resolve(import.meta.dirname, "src/components")
 const componentEntries = Object.fromEntries(
     findVueFiles(componentsDir).map(file => {
```
elevatebart, line 29: What is that for ?

ui/packages/design-system/src/assets/styles/ks-tokens.scss
```diff
@@ -1,10 +1,84 @@
 @use "variables" as v;
+@use "color-palette" as palette;
 
 :root {
     // base colors
     #{--ks-black}: #{v.$black};
     #{--ks-white}: #{v.$white};
 
+    // color palette
```
elevatebart: Why are we exposing this as pure CSS?
A concern I would have is that... people use it to override their component and make inconsistent color UIs and colors that do not follow dark/light themes. I truly prefer semantic colors the way Nicolas set them up in the templates.

ui/packages/design-system/src/assets/styles/bootstrap.scss
```diff
@@ -8,6 +8,8 @@ $prefix: 'kbs-';
 @import "bootstrap/scss/maps";
 @import "bootstrap/scss/mixins";
 @import "bootstrap/scss/reboot";
+@import "bootstrap/scss/containers";
+@import "bootstrap/scss/grid";
```
elevatebart: This adds significant bloat to the package.

elevatebart (commented):
Some comments

ui/packages/design-system/src/assets/styles/ks-tokens.scss
Skraye: You do use those color inside Kestra, for example for the state label, so I need the same colors exposed

ui/packages/design-system/tsdown.config.ts
Skraye: Two fixes:
* normalizes path separators in export keys so they're always forward slashes — without it, builds can produce broken keys like components\Data\KsAvatar. 
* customExports ensures all export keys are prefixed with ./ as required by the Node.js package exports spec — tsdown sometimes generates keys without it

ui/packages/design-system/src/assets/styles/bootstrap.scss
Skraye: I remove it but this is still heavily used in the core anyway and I was trying to align on it

ui/packages/design-system/tsdown.config.ts
elevatebart: I published a fix for the / starting issue a week ago. There was customEntries and they were absolutely unnecessary. Sorry for the confusion on this.

Can you, if you intend to merge this change, share a bad version of package.json exports before your change and a good one after so I can see what changed?

Skraye: agree its work without now, removed!

elevatebart (approved):

PR #16146, opened 2026-05-20
chore: fix queue migration script

ammeek:
### ✨ Description

Fixes a bean conflict that caused EE tests using in-memory queues to fail with `Index "IX_TYPE__OFFSET" not found` against an H2 datasource. All three `V2_0QueueIndexAndType` migration classes (`h2`, `mysql`, `postgres`) were gated on `@Requires(property = "kestra.queue.type", pattern = "<dialect>|memory")`, so when tests ran with `kestra.queue.type=memory` all three matched — and the MySQL/Postgres dialect SQL ran against the H2 datasource and blew up.

**This PR:**
- Introduces three new annotations (`@H2QueueEnabled`, `@MysqlQueueEnabled`, `@PostgresQueueEnabled`) alongside the existing `@*RepositoryEnabled` annotations, encapsulating the queue-type discriminator each module already uses on sibling queue migrations.
- Applies them to `V2_0QueueIndexAndType` in `jdbc-h2`, `jdbc-mysql`, and `jdbc-postgres`, removing the `|memory` overlap from the MySQL and Postgres variants and bringing them in line with `V2_0QueueMigration` / `V2_0QueueUpgradeMigration`.

### 🛠️ Backend Checklist

### 📝 Additional Notes

- **Follow-up suggestion:** Add package-level `@Requires` conditions (via `package-info.java`) to the `io.kestra.repository.{h2,mysql,postgres}.migration` packages so every bean in a dialect-specific migration package is automatically gated to that dialect. That would make this class of bug structurally impossible — a future migration added to the wrong package, or with a too-permissive `@Requires`, still wouldn't activate against the wrong `DataSource`.

### DIFF
--- a/jdbc-h2/src/main/java/io/kestra/repository/h2/H2QueueEnabled.java
+++ b/jdbc-h2/src/main/java/io/kestra/repository/h2/H2QueueEnabled.java
@@ -1,0 +1,12 @@
+package io.kestra.repository.h2;
+
+import java.lang.annotation.*;
+
+import io.micronaut.context.annotation.Requires;
+
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.PACKAGE, ElementType.TYPE })
+@Requires(property = "kestra.queue.type", pattern = "h2|memory")
+public @interface H2QueueEnabled {
+}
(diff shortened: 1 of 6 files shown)


loicmathieu (approved):
LGTM

PR #16147, opened 2026-05-20
fix(worker): close race between kill event and callable registration

fhussonnois:
A kill event arriving after `register()` but before `callJob()` sets `currentWorkerCallable` was silently dropped: the killable callback ran on a null callable, and the early `isExecutionKilled` cache check in `runTask` had already returned false. The task then ran to completion.

Track the kill request on the processor so it survives the gap, and honor `killed` at the start of `AbstractWorkerCallable.call()` for the sub-window where `interrupt()` was a no-op (currentThread not yet set).


ammeek (approved):
LGTM

PR #16163, opened 2026-05-21
fix(plugins): prevent ClassLoader mismatch when multiple plugin versions coexist

fdelbrayelle:
## Summary

- When two versions of the same plugin JAR are simultaneously present (e.g. during a rolling upgrade or when both v1.12.3 and v1.12.4 of a plugin are in the plugins folder), both are loaded with separate `ClassLoader` instances.
- `PluginDeserializer` could resolve the outer `Task` class from one `ClassLoader` and a nested `AdditionalPlugin` field (e.g. a `ModelProvider` subtype used as a task field) from a different version's `ClassLoader`.
- Java reflection then rejects `field.set()` with `Can not set [FieldType] field [Task].[field] to [Value]`, because `FieldType_CLv3.isAssignableFrom(Value_CLv4)` returns `false` — even though they are logically the same type.

**Concrete reproduction**: `plugin-ai` v1.12.3 and v1.12.4 both present → flow using `IngestDocument` (with a `ModelProvider provider` field) fails validation with `Can not set io.kestra.plugin.ai.domain.ModelProvider field io.kestra.plugin.ai.rag.IngestDocument.provider to io.kestra.plugin.ai.provider.GoogleGemini`.

## Changes

**`PluginRegistry`** — adds a new `default` method `findClassByIdentifier(String, ClassLoader preferredClassLoader)`. The default implementation delegates to `findClassByIdentifier(String)`, keeping all existing implementations (including EE) backward-compatible.

**`DefaultPluginRegistry`** — overrides the new method with a two-phase lookup:
1. Fast path: direct key lookup; returns immediately if the class belongs to `preferredClassLoader`.
2. Slow path: scans all registered values for a class whose name matches the bare type and whose `ClassLoader` is `preferredClassLoader`. Falls back to the standard lookup if nothing matches.

**`PluginDeserializer`** — adds a `ThreadLocal<ClassLoader> CURRENT_PLUGIN_CLASS_LOADER` that is:
- Set to `pluginType.getClassLoader()` when the outermost plugin in a deserialization chain is resolved (detected by `get() == null`).
- Passed to every `findClassByIdentifier` call so nested plugin lookups prefer the same `ClassLoader`.
- Cleared in a `finally` block once the outermost `treeToValue` call completes.

## Test plan

- [ ] `DefaultPluginRegistryClassLoaderTest` — unit tests for the new `findClassByIdentifier(String, ClassLoader)` overload: prefers the right CL, falls back when no match, behaves identically when `null` is passed.
- [ ] `PluginDeserializerTest` — existing tests still pass; the `ThreadLocal` is correctly set and cleared.
- [ ] Manual: install two versions of `plugin-ai` simultaneously and validate a flow using `IngestDocument` with a `GoogleGemini` provider — should no longer throw `Can not set ModelProvider field`.

### DIFF
--- a/core/src/main/java/io/kestra/core/plugins/DefaultPluginRegistry.java
+++ b/core/src/main/java/io/kestra/core/plugins/DefaultPluginRegistry.java
@@ -316,6 +316,47 @@ public class DefaultPluginRegistry implements PluginRegistry {
 
     /**
      * {@inheritDoc}
+     *
+     * <p>Searches {@code pluginClassByIdentifier} for all entries whose class was loaded by
+     * {@code preferredClassLoader} and whose class name matches the unversioned {@code identifier}.
+     * This handles the case where a versioned registry (e.g. EE) stores entries under
+     * {@code "type:version"} keys: by scanning all values we can find the entry for the right CL
+     * even when the lookup key carries no version.</p>
+     **/
+    @Override
+    public Class<? extends Plugin> findClassByIdentifier(final String identifier, final ClassLoader preferredClassLoader) {
+        if (preferredClassLoader == null) {
+            return findClassByIdentifier(identifier);
+        }
+        requireNonNull(identifier, "Cannot found plugin for null identifier");
+
+        // Extract the bare type name (strip any ":version" suffix that may be present in EE registries).
+        var typeName = PluginIdentifier.parseIdentifier(identifier).getLeft();
+
+        lock.lock();
+        try {
+            // Fast path: direct key lookup first (covers the non-versioned OSS case).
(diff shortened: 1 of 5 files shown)


fdelbrayelle:
Closing in favor of [kestra-ee#7927](https://github.com/kestra-io/kestra-ee/pull/7927) — the root cause is in EE-side registration (`EEPluginRegistry.getPluginIdentifier` was storing `AdditionalPlugin` subtypes under `ClassTypeIdentifier` instead of `VersionedPluginIdentifier` like it does for tasks), so the right fix lives in kestra-ee with zero hot-path impact. The original ThreadLocal/slow-scan approach in this PR added cost to the deserialization hot path and was not the right shape.

PR #16164, opened 2026-05-21
fix(core): fix flaky BadFlowableTest.flowableWithParentFail race condition

AcevedoR:
## Summary

- `flowableWithParentFail` was flaky in CI because it asserted sub-execution states immediately after the parent execution completed
- The flow uses `concurrencyLimit: 0` (parallel loops), so the parent can reach `FAILED` while one sub-execution is still `RUNNING`
- Added `Await.until()` to wait for all sub-executions to reach a terminal state before asserting — the same pattern already used in `FinallyTest`

## Root cause

The assertion on line 37:
```java
assertThat(subExecutions).extracting("state.current").containsOnly(State.Type.FAILED);
```
was seeing `[FAILED, RUNNING]` because `@ExecuteFlow` only waits for the **parent** execution to terminate, not the concurrent loop sub-executions.

## Test plan

- [x] `./gradlew :core:test --tests "io.kestra.plugin.core.flow.BadFlowableTest" --rerun-tasks` passes (2/2 green)
- [x] Matches existing pattern in `FinallyTest.loopWithErrors` / `loopParallelWithErrors`

### DIFF
--- a/core/src/test/java/io/kestra/plugin/core/flow/BadFlowableTest.java
+++ b/core/src/test/java/io/kestra/plugin/core/flow/BadFlowableTest.java
@@ -1,6 +1,7 @@
 package io.kestra.plugin.core.flow;
 
 import io.kestra.core.repositories.ExecutionRepositoryInterface;
+import io.kestra.core.utils.Await;
 import jakarta.inject.Inject;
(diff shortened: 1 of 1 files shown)


AcevedoR:
we might need to backport too

PR #16165, opened 2026-05-21
fix(tests): fix two flaky CI tests

AcevedoR:
## Summary

Fixes two tests that were consistently flaking in CI, identified via Develocity data from the past week (37 and 23 failures respectively).

### `WorkerTest.shouldKillTasksWhenExecutionKillEventReceived`

**Root cause:** `worker.getRunningJobs().size() >= 2` is satisfied as soon as the Worker registers the jobs internally, before the task-run state machine transitions to `RUNNING`. Sending the kill immediately after caused a race: the kill arrived before `RUNNING` was written, producing `[CREATED, KILLED]` instead of `[CREATED, RUNNING, KILLED]`, which broke `assertThat(histories).hasSize(3)`.

**Fix:** Extend the await condition to also require that the kill target has emitted a `RUNNING` state via `workerTaskResultQueue` — the same pattern used in `AbstractServiceLivenessCoordinatorTest`.

### `AbstractServiceLivenessCoordinatorTest.shouldResubmitTaskWhenWorkerIsStopped`

**Root cause:** `runningLatch.await(10, TimeUnit.SECONDS)` timed out on slow CI runners before the worker picked up the task and emitted `RUNNING` state, causing the parameterized test variants `[1]` and `[2]` to fail.

**Fix:** Increase both `runningLatch` timeouts from 10 s to 30 s (consistent with the `resubmitLatch` timeout already set to 30 s in the same test).

## Test plan

- [x] `./gradlew :worker:test --tests "io.kestra.worker.WorkerTest.shouldKillTasksWhenExecutionKillEventReceived" --rerun-tasks` — passes
- [x] `./gradlew :jdbc-h2:test --tests "io.kestra.runner.h2.H2ServiceLivenessCoordinatorTest.shouldResubmitTaskWhenWorkerIsStopped" --rerun-tasks` — both parameterized variants pass

### DIFF
--- a/executor/src/test/java/io/kestra/executor/AbstractServiceLivenessCoordinatorTest.java
+++ b/executor/src/test/java/io/kestra/executor/AbstractServiceLivenessCoordinatorTest.java
@@ -129,7 +129,7 @@ public abstract class AbstractServiceLivenessCoordinatorTest {
             }
         });
         workerJobEventQueue.emit(workerGroupKey, WorkerJobEvent.of(workerTask, workerGroupKey));
-        assertThat(runningLatch.await(10, TimeUnit.SECONDS)).isTrue();
+        assertThat(runningLatch.await(30, TimeUnit.SECONDS)).isTrue();
 
         // WHEN - stop first worker. The task is guaranteed to still be running because it is
         // blocked on holdLatch.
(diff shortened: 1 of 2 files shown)


nkwiatkowski (approved):
LGTM

PR #16166, opened 2026-05-21
fix(tests): backport flaky WorkerTest and liveness coordinator fixes to releases/v1.3.x

AcevedoR:
## Summary

Backport of flaky test fixes from \`develop\` to this release branch.

- **WorkerTest.killed**: replaced \`Thread.sleep(500)\` with a deterministic \`Await.until()\` that waits for the kill target to emit \`RUNNING\` state via the result queue before sending the kill event. The sleep was insufficient on slow CI runners, causing the kill to race the RUNNING transition and produce \`[CREATED, KILLED]\` instead of \`[CREATED, RUNNING, KILLED]\`.

- **JdbcServiceLivenessCoordinatorTest**: increased \`runningLatch\` timeouts from 5s/10s to 30s to accommodate slow CI runners where the worker takes longer to pick up a task and emit RUNNING state.


nkwiatkowski (approved):
LGTM

PR #16170, opened 2026-05-21
feat: support AND/OR boolean logic in query filters

### LINKED ISSUE #11587: [Filters] New filter operators for Labels
### Feature description

We want to support multiple filter operators on the Labels filter to a.o. clarify AND/OR/NOT conditions and to overall improve the filtering experience:

Image

Note that the MySQL implementation may not work fully here; the same applies to older Postgres versions - BE needs to validate

ammeek:
### ✨ Description

Adds support for AND/OR boolean logic in query filters across the API and JDBC repository layer.

- `QueryFilter` is now polymorphic — either a **leaf** (`field`/`operation`/`value`) or a **node** (`logical` operator + non-empty `children`). The compact constructor enforces leaf-XOR-node so invalid combinations are rejected on construction.
- The LHS-brackets URL format gains optional `[and][N]` and `[or][N]` prefixes for grouping. The existing flat format continues to work unchanged.
  - Example: `?filters[namespace][EQUALS]=io.kestra&filters[or][0][state][EQUALS]=RUNNING&filters[or][1][state][EQUALS]=FAILED`
  - Translates to: `WHERE namespace = 'io.kestra' AND (state = 'RUNNING' OR state = 'FAILED')`
- The JDBC repository dispatcher is now a recursive walker that emits `Condition.and` / `Condition.or` via jOOQ. Per-leaf handlers are unchanged.
- New `kestra.query-filter` configuration:
  - `max-depth` (default **3**, hard floor 3) caps nesting depth per URL param.
  - `max-width` (default **20**, hard floor 20) caps children per tree node.
  - `resources.<NAME>.max-{depth,width}` overrides caps per `QueryFilter.Resource`.
  - Below-floor values are clamped at startup with a WARN log; the bean never exposes a value below the floor.
- `@QueryFilterFormat` now requires a `QueryFilter.Resource` value so the binder can resolve per-Resource caps at parse time. All controller callsites updated.
- Defensive fixes in `DateUtils`, `SecretService`, and `AbstractJdbcExecutionRepository` so leaf-only call-sites no longer NPE when handed a node filter.
- Integration tests on **H2, PostgreSQL, and MySQL** exercising `(A AND B) OR (C AND D)`-style queries end-to-end against a live database.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/11587

### 📝 Additional Notes

- **Wire-format compatibility:** existing URLs (`filters[field][op]=value`) parse identically to before. The 5 pre-existing binder tests pass unchanged.
- **No UI changes in this PR** — the Vue 3 filter builder still emits flat AND lists. The new URL syntax is API-only until a follow-up adds a UI affordance for OR groups.
  - Full arbitrary tree nesting from the spec (`((1 OR 2) OR 3) AND 4`). The implementation supports DNF (a list of AND-groups OR'd together), which covers ~all realistic filter UIs and can be relaxed later if needed.

### DIFF
--- a/core/src/test/java/io/kestra/core/models/QueryFilterTest.java
+++ b/core/src/test/java/io/kestra/core/models/QueryFilterTest.java
@@ -11,6 +11,7 @@ import org.junit.jupiter.params.provider.MethodSource;
 
 import io.kestra.core.exceptions.InvalidQueryFiltersException;
 import io.kestra.core.models.QueryFilter.Field;
+import io.kestra.core.models.QueryFilter.Logical;
 import io.kestra.core.models.QueryFilter.Op;
 import io.kestra.core.models.QueryFilter.Resource;
 import io.kestra.core.models.dashboards.filters.AbstractFilter;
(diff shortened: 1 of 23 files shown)


core/src/test/java/io/kestra/core/models/QueryFilterTest.java
```diff
@@ -584,6 +585,63 @@ static Stream<Arguments> validOperationFilters() {
         ).flatMap(s -> s);
     }
 
+    @Test
+    void should_allow_leaf_construction() {
```
fhussonnois: No `_` in test name :)

webserver/src/main/java/io/kestra/webserver/configuration/QueryFilterConfiguration.java
```diff
@@ -0,0 +1,90 @@
...
+ * below the floor are clamped at startup with a WARN-level log; the bean never exposes a value below the floor.
+ * <p>
+ * Per-Resource overrides are independent of the global value — they may be tighter or looser than the global cap,
+ * but still subject to the same floor.
+ */
+@Slf4j
+@Getter
+@Setter
+@ConfigurationProperties("kestra.query-filter")
+public class QueryFilterConfiguration {
```
fhussonnois: We use `record` for configuration-props class no lombok. This props should be probably under `kestra.webserver.queryFilter` because it's only used by the webserver.

fhussonnois (changes requested):

webserver/src/main/java/io/kestra/webserver/configuration/QueryFilterConfiguration.java
ammeek: Thanks florian, I've moved this to be a record. I think it's best to leave `QueryFilterConfiguration ` in `io.kestra.webserver.configuration` as that is were the rest of the configuration for the web server lives.

fhussonnois (approved):

PR #16181, opened 2026-05-21
feat(core): add maintenance status endpoint

### LINKED ISSUE #14455: [API] maintenance mode status
### Feature description

I recommend implementing a public (or low-privilege) GET endpoint. Removing the superuser requirement would allow third-party monitoring tools or status pages to check system availability without compromising security through elevated permissions.

Sample:
GET /api/v1/instance/maintenance/status

``` json
{
  "status": "maintenance",
  "maintenance_details": {
    "is_active": true,
    "reason": "Scheduled database migration and security patches",
    "start_time": "2026-02-05T08:00:00Z",
    "estimated_end": "2026-02-05T11:00:00Z",
  }
}
```

Skraye:
Add GET /cluster/maintenance/status endpoint returning maintenance mode state and per-service readiness breakdown.

Closes #14455

Companion PR: https://github.com/kestra-io/kestra-ee/pull/7425

### DIFF
--- a/core/src/main/java/io/kestra/core/models/MaintenanceStatusResponse.java
+++ b/core/src/main/java/io/kestra/core/models/MaintenanceStatusResponse.java
@@ -1,0 +1,30 @@
+package io.kestra.core.models;
+
+import java.util.Map;
+
+import io.kestra.core.server.ServiceType;
+
+/**
+ * Response payload for the maintenance status endpoint.
+ *
+ * @param maintenance whether maintenance mode is currently enabled.
+ * @param ready       whether all services have transitioned to MAINTENANCE state
+ *                    (i.e. the instance is safe to update).
+ * @param services    per-type breakdown: how many instances of each service type
+ *                    are in MAINTENANCE vs total running.
+ */
+public record MaintenanceStatusResponse(
+    boolean maintenance,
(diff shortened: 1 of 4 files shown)


webserver/src/main/java/io/kestra/webserver/controllers/api/ClusterController.java
```diff
@@ -49,4 +57,11 @@ public Set<Metric> metrics(@QueryValue ServiceType serviceType) {
             .flatMap(Set::stream)
             .collect(Collectors.toSet());
     }
+
+    @ExecuteOn(TaskExecutors.IO)
+    @Get("/maintenance/status")
+    @Operation(tags = { "Cluster", "Maintenance" }, summary = "Get maintenance mode status", description = "Returns maintenance mode status and per-service readiness.")
+    public MaintenanceStatusResponse maintenanceStatus() {
```
loicmathieu: Why adding it here instead of in EE?
The rest can stay here but as maintenance mode is EE only better to move the endpoint in EE right?

loicmathieu (approved):
LGMT, maybe move the endpoint and its tests in EE.
Maybe also add a test inside the MaintenanceServiceTest  also?

PR #16193, opened 2026-05-22
fix: avoid render function stackOverflow by adding a configurable limit default to 16

Skraye:
Render() inside render() can create infiniting looping
Added a customizable max limit for recursive rendering

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
+++ b/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
@@ -22,6 +22,7 @@ import jakarta.inject.Singleton;
 public class VariableRenderer {
     private static final Pattern RAW_PATTERN = Pattern.compile("(\\{%-*\\s*raw\\s*-*%}(.*?)\\{%-*\\s*endraw\\s*-*%})");
     public static final int MAX_RENDERING_AMOUNT = 100;
+    public static final String RENDER_DEPTH_VAR = "__kestra_render_depth";
 
     private volatile PebbleEngine pebbleEngine;
     private final VariableConfiguration variableConfiguration;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/kestra/core/runners/pebble/functions/RenderFunction.java
loicmathieu, line 1: A ThreadLocal is costly, I wonder if we would be able to avoid using one by adding a counter inside the Pebble context itself.
Or maybe by overloading the render function.

If not possible use a `StableValue` instead.

By the way, the current code has a memory leak did you spot it ;)

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/runners/pebble/functions/RenderFunction.java
Skraye: Remove the ThreadLocal (and its memory leak :p) and went for an overload of the render function

core/src/main/java/io/kestra/core/runners/VariableRenderer.java
```diff
@@ -22,6 +22,7 @@
 public class VariableRenderer {
     private static final Pattern RAW_PATTERN = Pattern.compile("(\\{%-*\\s*raw\\s*-*%}(.*?)\\{%-*\\s*endraw\\s*-*%})");
     public static final int MAX_RENDERING_AMOUNT = 100;
+    public static final String RENDER_DEPTH_VAR = "__kestra_render_depth";
```
loicmathieu, line 25: I wonder if we should not name such variables `__kestra.render_depth` sounds more like a real prefix thing.
No strong opinion

loicmathieu (approved):
LGTM, very nice and neat! Well done

PR #16197, opened 2026-05-22
feat: add AND/OR condition groups to KsFilter

### LINKED ISSUE #11587: [Filters] New filter operators for Labels

ammeek:
### ✨ Description

Adds AND/OR condition-group support to the `KsFilter` data-table filter.

Capabilities:
- **OR by default at the top level.** Multiple top-level groups combine with OR; the operator chip between groups can be flipped to AND.
- **Drag chips between groups.** Chips can be dragged from one group to another to recompose the query.
- **Drag-to-wrap (one level of nesting).** Dropping one group onto another wraps them into an AND/OR sub-group. Wrappers are limited to a single nesting level — anything deeper is handled by the raw editor.
- **Drag fields from "Add filter" into a group.** Dragging a row from the customize dropdown drops a new chip into the target group with a styled chip-shaped drag preview.
- **Raw-query editor fallback.** When the URL has more nesting than the chip UI can render, a raw newline-delimited editor is shown instead.
- **Same field, different comparators in one group.** Two chips on the same field with distinct comparators (e.g. `startDate > X AND startDate < Y`) are now allowed.
- **Empty chips self-clean.** Closing the edit popover without picking a value removes the chip so misleading placeholder chips never persist.

URLs round-trip via the existing `filters[and|or][N][and|or][M][field][op]=value` backend parser, so legacy flat URLs continue to load and existing saved filters keep working.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/11587


ui/packages/design-system/src/components/Data/KsDataTable/filter/RawFilter.vue
```diff
@@ -0,0 +1,118 @@
...
+</template>
+
+<script setup lang="ts">
+    import {computed, inject, ref, watch} from "vue"
+    import {FILTER_CONTEXT_INJECTION_KEY} from "./utils/filterInjectionKeys"
+
+    const filter = inject(FILTER_CONTEXT_INJECTION_KEY)!
+    const draft = ref(filter.rawQuery?.value ?? "")
+
+    watch(() => filter.rawQuery?.value, (latest) => {
```
Piyush-r-bhaskar: **Dead watcher** — `isDirty` is computed against `filter.rawQuery?.value`, the same value that just changed. Inside the callback `isDirty` is always `true` (or values equal → early return), so `draft.value = latest` is unreachable.

ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue
```diff
@@ -0,0 +1,242 @@
+<template>
+    <div
+        v-if="isWrapperGroup(unit)"
+        class="filter-wrapper-container"
+        :class="{
+            'filters-hidden': filter.searchInputFullWidth?.value,
+            'drag-over': dragOverGroupId === unit.id,
+        }"
+        @dragover="(e) => $emit('drag-over', e, unit.id)"
```
Piyush-r-bhaskar, line 9: **Nested drop events bubble.** Both the wrapper container and the inner child dropzone bind `@dragover`/`@drop`. Drops on a child fire twice (the second is a no-op only by accident — `findLeafById` returns `undefined` for the wrapper id). The `drag-over` highlight also lands on the wrapper instead of the child. Add `.stop` to the inner dropzone's emits ?

ui/packages/design-system/src/components/Data/KsDataTable/filter/MainFilter.vue
```diff
@@ -103,6 +128,117 @@
...
+    const addFieldChipToGroup = (keyName: string, targetGroupId: string) => {
+        const key = filter.configuration.value.keys?.find((k) => k.key === keyName)
+        if (!key) return
+        const comparator = key.comparators?.[0]
+        const newFilter: AppliedFilter = {
+            id: `${key.key}-${Date.now()}-${Math.random().toString(36).slice(2, 6)}`,
+            key: key.key,
+            keyLabel: key.label,
+            comparator: comparator!,
+            comparatorLabel: comparator as unknown as string,
```
Piyush-r-bhaskar: I think `comparatorLabel: comparator as unknown as string` — should be `COMPARATOR_LABELS[comparator]`. Chip briefly shows raw symbol (`IN`, `*=`) instead of the human label until edit applies.

ui/packages/design-system/src/components/Data/KsDataTable/filter/segments/CustomizeFilters.vue
```diff
@@ -60,41 +63,78 @@
...
-            }
-            emits("add-filter", newFilter)
+    const addFilterForKey = (key: FilterKeyConfig) => {
+        const comparator = key.comparators?.[0]
+        const newFilter: AppliedFilter = {
+            id: `${key.key}-${Date.now()}-${Math.random().toString(36).slice(2, 6)}`,
+            key: key.key,
+            keyLabel: key.label,
+            comparator,
+            comparatorLabel: comparator,
```
Piyush-r-bhaskar: here too `comparatorLabel: comparator`. Should be `COMPARATOR_LABELS[comparator]`

ui/packages/design-system/src/components/Data/KsDataTable/filter/MainFilter.vue
```diff
@@ -103,6 +128,117 @@
...
+
+    const addFieldChipToGroup = (keyName: string, targetGroupId: string) => {
+        const key = filter.configuration.value.keys?.find((k) => k.key === keyName)
+        if (!key) return
+        const comparator = key.comparators?.[0]
+        const newFilter: AppliedFilter = {
+            id: `${key.key}-${Date.now()}-${Math.random().toString(36).slice(2, 6)}`,
+            key: key.key,
+            keyLabel: key.label,
+            comparator: comparator!,
```
Piyush-r-bhaskar: `comparator: comparator!` — non-null assertion crashes if a `FilterKeyConfig` has an empty `comparators` array. Guard or default to `Comparators.EQUALS` ?

ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue
```diff
@@ -0,0 +1,242 @@
...
+                class="filter-group-dropzone"
+                :class="{
+                    'drag-over': dragOverGroupId === childLeaf.id,
+                    'empty': childLeaf.filters.length === 0,
+                }"
+                @dragover="(e) => $emit('drag-over', e, childLeaf.id)"
+                @dragleave="(e) => $emit('drag-leave', e, childLeaf.id)"
+                @drop="(e) => $emit('drop', e, childLeaf.id)"
+            >
+                <span
```
Piyush-r-bhaskar, line 31: Drag handle `<span>` is not focusable and has no keyboard alternative. Screen-reader users can't trigger reorder. Wrap in a `<button>` or document the gap.

ui/packages/design-system/src/components/Data/KsDataTable/filter/composables/useFilters.ts
```diff
@@ -183,113 +204,181 @@ export function useFilters(
...
+                    TIME_RANGE_KEY,
                     createTimeRangeFilter(
                         timeRangeConfig,
-                        new Date(dateFilters.startDate?.value),
-                        new Date(dateFilters.endDate?.value),
+                        new Date(slot.dateFilters[START_DATE_FIELD].value),
+                        new Date(slot.dateFilters[END_DATE_FIELD].value),
                         comparator,
                         meta,
                     ),
```
Piyush-r-bhaskar: This file started getting dirty and not maintainable..take care of it please.

Piyush-r-bhaskar (changes requested):
**Non-existent design-system tokens (fallbacks mask the issue)**
The token names don't exist in `ks-tokens.scss`; the inline fallback is what actually renders. no magic numbers, no orphan tokens please

- `var(--ks-border-radius-sm, 0.15rem)` → real token is `--ks-radius-sm`
  - `ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue:164`
  - `ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue:189`
  - `ui/packages/design-system/src/components/Data/KsDataTable/filter/segments/LogicalSeparator.vue:55`
  - `ui/packages/design-system/src/components/Data/KsDataTable/filter/segments/LogicalChooser.vue:45`
  - `ui/packages/design-system/src/components/Data/KsDataTable/filter/RawFilter.vue:99`
- `var(--ks-font-family-monospace, monospace)` → real token is `--ks-font-family-mono`
  - `ui/packages/design-system/src/components/Data/KsDataTable/filter/RawFilter.vue:92`
  - `ui/packages/design-system/src/components/Data/KsDataTable/filter/RightFilter.vue:199`
- `var(--ks-background-hover, …)` → no such token, use a component-scoped one
  - `ui/packages/design-system/src/components/Data/KsDataTable/filter/segments/LogicalChooser.vue:49`

---

**Text characters used as icons**
*"Never inline raw SVG, font-icon classes, or emoji as UI state."* Replace with `<KsIcon name="…" />` (vue-material-design-icons).

- `ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue:39` — `⋮⋮` drag handle
- `ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue:90` — `⋮⋮` drag handle
- `ui/packages/design-system/src/components/Data/KsDataTable/filter/segments/GroupActions.vue:13` — `⤴` unwrap
- `ui/packages/design-system/src/components/Data/KsDataTable/filter/segments/GroupActions.vue:28` — `✕` remove
- `ui/packages/design-system/src/components/Data/KsDataTable/filter/RightFilter.vue:77` — `{ }` / `≡` view-mode toggle

ui/packages/design-system/src/components/Data/KsDataTable/filter/segments/LogicalChooser.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+    <div class="logical-chooser">
+        <button
```
Piyush-r-bhaskar: use `KsButton` please

ui/packages/design-system/src/components/Data/KsDataTable/KsFilter.locale.ts
```diff
@@ -36,6 +36,21 @@ export default {
             "reset": "Clear all",
             "reset_all": "Reset all filters",
             "reset_tooltip": "Reset to default",
+            "or": "OR",
```
Piyush-r-bhaskar, line 39: use `claude` for other lang too...workflow won't generate for keys under this file.

ui/packages/design-system/src/components/Data/KsDataTable/filter/MainFilter.vue
```diff
@@ -92,16 +116,128 @@
...
+        // Mirror onDragEnd so the highlight clears if the user drops outside any dropzone.
+        draggingEntity.value = null
+        dragOverGroupId.value = null
+    }
+
+    const onDragEnd = () => {
+        draggingEntity.value = null
+        dragOverGroupId.value = null
+    }
+
```
Piyush-r-bhaskar, line 240: Could you find any `@vueuse/core` which could shorten the logic. I feel we do have it.

Piyush-r-bhaskar (changes requested):

Piyush-r-bhaskar:
Also you need to rebase, I am seeing old layout of kestra from your branch.

ui/packages/design-system/src/components/Data/KsDataTable/filter/MainFilter.vue
ammeek: useDropZone is the only fit, but it only covers the hover/leave half and binds to a single element — with our nested v-for dropzones it'd mean a per-dropzone component plus re-adding what it doesn't handle (self-drop, move/copy, nested .stop) so I'd leave it as-is.

Piyush-r-bhaskar (approved):
Looks good, once you fix those two visual changes.

PR #15817, opened 2026-05-03
feat: add query filter support to namespace search.

ammeek:
### ✨ Description
This PR adds query filter support for the following Field, Resource and Operator combinations.
                                            
| Resource  | Field     | Operator    | 
|-----------|-----------|-------------|
| Namespace | Namespace | EQUALS      |                                                                                                                   
| Namespace | Namespace | NOT_EQUALS  |     
| Namespace | Namespace | CONTAINS    |                                                                                                                   
| Namespace | Namespace | STARTS_WITH |
| Namespace | Namespace | ENDS_WITH   |                                                                                                                   
| Namespace | Namespace | REGEX       |     
| Namespace | Namespace | IN          |                                                                                                                   
| Namespace | Namespace | NOT_IN      |
| Namespace | Namespace | PREFIX      |                                                                                                                   
                                            
**Backend changes:**                                                                                                                                      
- Introduced `SearcheableFactory` in OSS webserver providing an in-memory `Searcheable<Namespace>` bean (qualified `NAMESPACE`) with all operators above
plus free-text search                                                                                                                                     
- Refactored `NamespaceController.searchNamespaces` in OSS to use `Searcheable` for filtering, sorting, and pagination instead of manual stream logic —
fixes the previously ignored `filters` and `sort` parameters                                                                                              

                                            
**Frontend changes:**                                                                                                                                     
- `namespacesFilter.ts` — added the `namespace` filter key with all supported comparators and comparator selection enabled
- `Namespaces.vue` — data loading now passes all `filters[*]` query params from the URL to the search API and watches the full `route.query` for changes; 
layout restructured to place `KSFilter` outside the `el-row` grid to fix filter popover positioning
                                                                                                                                                          
### 🔗 Related Issue                                                                                                                                      
relates to: https://github.com/kestra-io/kestra-ee/issues/6255                                                                                       
                                                                                                                                                          
### 🎨 Frontend Checklist                                                                                                                                 
- [x] Code builds without errors (`npm run build`)
- [x] All existing E2E tests pass (`npm run test:e2e`)                                                                                                    
- [x] Screenshots or video recordings attached showing the `UI` changes
                                                                                                                                                          
### 🛠️  Backend Checklist
- [x] Code compiles successfully and passes all checks                                                                                                    
- [x] All unit and integration tests pass

PR #16202, opened 2026-05-25
feat: expand log filter fields

### LINKED ISSUE #15954: [Filters] [Logs] Support "at or below" filter operators
### Feature description

Add at or below - Include this level and all higher severities.

Image

ammeek:
### ✨ Description

Reworks log filtering to fix a misleading `EQUALS` semantic and consolidate the per-execution log endpoints around the `QueryFilter` machinery.

**Backend (model + repository + controller):**

- `QueryFilter.Field`: renamed `MIN_LEVEL` → `LEVEL` (JSON wire value `"level"` unchanged). The old `EQUALS` op silently meant "at or above" — replaced with explicit `GREATER_THAN_OR_EQUAL_TO` and added `LESS_THAN_OR_EQUAL_TO`. `EQUALS`/`NOT_EQUALS` on `LEVEL` are now rejected.
- Added `TASK_ID`, `TASK_RUN_ID`, `ATTEMPT_NUMBER` to `Resource.LOG.supportedField()`. JDBC column-name override on `AbstractJdbcLogRepository` maps `TASK_RUN_ID → taskrun_id` (single mismatch with the default snake-case derivation). New `handleAttemptNumberField` parses values to `Integer` before binding so Postgres' strict type contract is honoured (H2 silently coerced before).
- `LogController.listLogsFromExecution` / `downloadLogsFromExecution` / `followLogsFromExecution`: dropped the typed `?minLevel=`, `?taskId=`, `?taskRunId=`, `?attempt=` query params. Endpoints now take a single `@QueryFilterFormat List<QueryFilter> filters` plus the existing `executionId` path variable. `buildExecutionFilters` appends the path-var `EXECUTION_ID = EQUALS` filter to the user-supplied list.
- `LogStreamingService` now stores `List<QueryFilter>` per subscriber and applies them via a new `FollowLogEventMatcher` interface (backed by a `Searchable<FollowLogEvent>` registered in `SearchableFactory`). Live SSE streaming now respects the full filter set, not just LEVEL — closes a gap where any non-`level` filter was silently ignored on `/follow`.
- `ILogs.whereWithGlobalFilters`: added the LEVEL branch so dashboard log charts can scope by level.

**Searchable utility:**

- Added `Searchable.matches(item, filters)` extracted from `filter()` for single-item evaluation.
- Added a `Function<? super T, ?>`-based overload of `searchableQueryFilterExtractor` that synthesises the predicate from the standard string-semantic for each operator (EQUALS, NOT_EQUALS, CONTAINS, STARTS_WITH, ENDS_WITH, REGEX, IN, NOT_IN, PREFIX). Lets factories register a field with a one-liner method reference instead of writing the same lambda shape repeatedly. Unsupported ops (`GREATER_THAN` etc.) throw at registration time.

**Frontend:**

- `logLevelQuery.ts`: route helpers now return `{value, direction}` instead of a bare string. Handles `GREATER_THAN_OR_EQUAL_TO`, `LESS_THAN_OR_EQUAL_TO`, and the legacy `EQUALS` shape (read-only — writes use the new form). New `levelToRequestParams` helper.
- `FilterKeyConfig.comparatorLabels` (optional per-field label override) — the level filter dropdown now shows **"At or Above" / "At or Below"** instead of "Greater Than or Equal To" / "Less Than or Equal To". Other filters using the same comparators are untouched.
- `logFilter.ts`, `logExecutionsFilter.ts`, `ganttExecutionFilter.ts`: comparators + labels for level; added `taskId`, `taskRunId`, `attemptNumber` filter entries to the main logs page and the per-execution Logs tab.
- `LogsWrapper.vue`, `Logs.vue`, `Gantt.vue`: typed `useRouteFilterPolicy<LevelFilterValue>` and propagated the richer value to children.
- `TaskRunDetails.vue`: `level` prop renamed to `levelFilter` (object). Internal fetch builds request params via `levelToRequestParams(levelFilter)` and explicit `filters[taskId][EQUALS]` / `filters[attemptNumber][EQUALS]` keys.
- `ErrorAlert.vue`: migrated from `?minLevel=ERROR` to `?filters[level][GREATER_THAN_OR_EQUAL_TO]=ERROR`.

**Tests added:**

- `QueryFilterTest`: LEVEL valid/invalid op sets.
- `DashboardDataGlobalFiltersTest`: 6 LEVEL cases on `ILogs.whereWithGlobalFilters`.
- `AbstractLogRepositoryTest`: refactored to `FiltersTestCase` pattern. 13 cases covering LEVEL ≥/≤, TASK_ID (EQUALS/NOT_EQUALS/IN), TASK_RUN_ID (EQUALS/IN — exercises the `taskrun_id` column-name override), ATTEMPT_NUMBER (EQUALS/NOT_EQUALS/IN — exercises the integer-parsing path).
- `LogStreamingServiceTest`: new file. 8 parameterized cases verifying that emitted `FollowLogEvent`s only reach subscribers whose `QueryFilter` list they match.
- `LogControllerTest`: 21 parameterized cases (7 × 3 endpoints — list / download / follow) end-to-end through the HTTP client (follow uses direct controller invocation for the historical replay assertion).
- `SearchableTest`: 5 `matches()` cases + 21 default-operator parameterized cases + builder-time rejection of unsupported ops.

**Breaking changes (major release):**

- `Field.MIN_LEVEL` → `Field.LEVEL` in `QueryFilter` (JSON `"level"` unchanged).
- `LEVEL` no longer accepts `EQUALS` / `NOT_EQUALS` — only `GREATER_THAN_OR_EQUAL_TO` and `LESS_THAN_OR_EQUAL_TO`.
- `/logs/{executionId}`, `/logs/{executionId}/download`, `/logs/{executionId}/follow`: typed `?minLevel=`, `?taskId=`, `?taskRunId=`, `?attempt=` query params removed. Use `?filters[level][GREATER_THAN_OR_EQUAL_TO]=...&filters[taskId][EQUALS]=...` etc.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/15954

### 🎨 Frontend Checklist

- [ ] Screenshots attached showing the level filter dropdown rendering "At or Above" / "At or Below"

### 🛠️ Backend Checklist

- [ ] All unit and integration tests pass (verified locally on H2; Postgres-specific `attempt_number` int-coercion is covered by the new `handleAttemptNumberField` helper)

### DIFF
--- a/core/src/test/java/io/kestra/core/repositories/AbstractLogRepositoryTest.java
+++ b/core/src/test/java/io/kestra/core/repositories/AbstractLogRepositoryTest.java
@@ -10,8 +10,11 @@ import java.util.stream.Stream;
 
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.FieldSource;
 import org.junit.jupiter.params.provider.MethodSource;
 import org.slf4j.event.Level;
+
+import lombok.Builder;
 
 import io.kestra.core.exceptions.InvalidQueryFiltersException;
 import io.kestra.core.models.QueryFilter;
@@ -145,8 +148,8 @@ public abstract class AbstractLogRepositoryTest {
             QueryFilter.builder().field(Field.EXECUTION_ID).value("Id").operation(Op.ENDS_WITH).build(),
             QueryFilter.builder().field(Field.EXECUTION_ID).value(List.of("executionId")).operation(Op.IN).build(),
             QueryFilter.builder().field(Field.EXECUTION_ID).value(List.of("anotherId")).operation(Op.NOT_IN).build(),
-            QueryFilter.builder().field(Field.MIN_LEVEL).value(Level.DEBUG).operation(Op.EQUALS).build(),
-            QueryFilter.builder().field(Field.MIN_LEVEL).value(Level.ERROR).operation(Op.NOT_EQUALS).build()
+            QueryFilter.builder().field(Field.LEVEL).value(Level.DEBUG).operation(Op.GREATER_THAN_OR_EQUAL_TO).build(),
+            QueryFilter.builder().field(Field.LEVEL).value(Level.INFO).operation(Op.LESS_THAN_OR_EQUAL_TO).build()
         );
     }
 
@@ -172,7 +175,9 @@ public abstract class AbstractLogRepositoryTest {
             QueryFilter.builder().field(Field.TRIGGER_EXECUTION_ID).value("test").operation(Op.EQUALS).build(),
             QueryFilter.builder().field(Field.CHILD_FILTER).value(ChildFilter.CHILD).operation(Op.EQUALS).build(),
             QueryFilter.builder().field(Field.WORKER_ID).value("test").operation(Op.EQUALS).build(),
-            QueryFilter.builder().field(Field.EXISTING_ONLY).value("test").operation(Op.EQUALS).build()
+            QueryFilter.builder().field(Field.EXISTING_ONLY).value("test").operation(Op.EQUALS).build(),
+            QueryFilter.builder().field(Field.LEVEL).value(Level.INFO).operation(Op.EQUALS).build(),
+            QueryFilter.builder().field(Field.LEVEL).value(Level.INFO).operation(Op.NOT_EQUALS).build()
         );
     }
 
@@ -192,8 +197,8 @@ public abstract class AbstractLogRepositoryTest {
         assertThat(find.getFirst().getExecutionId()).isEqualTo(save.getExecutionId());
         var filters = List.of(
             QueryFilter.builder()
-                .field(QueryFilter.Field.MIN_LEVEL)
-                .operation(QueryFilter.Op.EQUALS)
+                .field(QueryFilter.Field.LEVEL)
+                .operation(QueryFilter.Op.GREATER_THAN_OR_EQUAL_TO)
                 .value(Level.WARN)
                 .build(),
             QueryFilter.builder()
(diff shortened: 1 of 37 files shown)


core/src/test/java/io/kestra/core/repositories/AbstractLogRepositoryTest.java
```diff
@@ -424,4 +429,380 @@ void purge() {
...
+
+        assertThat(results)
+            .extracting(LogEntry::getExecutionId)
+            .containsExactlyInAnyOrderElementsOf(
+                testCase.expectedLogs().stream().map(LogEntry::getExecutionId).toList()
+            );
+    }
+
+    @Builder
+    public record FiltersTestCase(
```
fhussonnois, line 803: Please create a follow-up issue for refactoring (during cooldown). We have a `FiltersTestCase` for almost all repositories. The model is always the same only tye change; so we could have:

```
public record FiltersTestCase<T>(List<T> actual, List<T> expected, QueryFilter filter)
```

Then, one class for each type, e.g., `LogFilterTestCases` to avoid creating to much noises in repository test classes

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
```diff
@@ -521,26 +526,57 @@ private Condition handleChildFilter(Object value, Op operation) {
...
+            case NOT_EQUALS -> DSL.field(columnName).ne(toInteger(value));
+            case IN -> DSL.field(columnName).in(toIntegerList(value));
+            case NOT_IN -> DSL.field(columnName).notIn(toIntegerList(value));
+            default -> throw new InvalidQueryFiltersException(
+                "Unsupported operation for ATTEMPT_NUMBER: " + operation
+            );
+        };
+    }
+
+    private static Integer toInteger(Object value) {
```
fhussonnois, line 586: Add a //TODO here - we should create reusable classes for type conversion - that's a global technical debt in Kestra

webserver/src/main/java/io/kestra/webserver/controllers/api/NamespaceController.java
fhussonnois, line 137: Is it not possible to drop the `getNamespace` method and use Searchable as you did in the `searchNamespace` ?

fhussonnois (approved):
LGTM, just a few comments/questions

webserver/src/main/java/io/kestra/webserver/controllers/api/NamespaceController.java
ammeek: I believe that this comment was linked to a changes which was addressed by the https://github.com/kestra-io/kestra/pull/15817 this pr used to be based off that prs branch because it contained changes this pr needed. I believe that the change was addressed in https://github.com/kestra-io/kestra/pull/15817

PR #16213, opened 2026-05-26
fix(triggers): redesign admin Triggers Manage tab

fhussonnois:
- Force-expand backfilled rows with BackfillBanner (per-row pause/resume/stop)
- Consolidate row actions into a hamburger menu (Details / Restart / Unlock / Delete)
- Add Details drawer with trigger vars
- Make all date columns sortable; add evaluatedAt column with header tooltip + info icon
- Replace inline backfill status icons with a KsTag (running / paused)
- Gate per-row backfill button on Schedule triggers
- Slim selection payload to {namespace, flowId, triggerId}
- Pass tenant param on flows/update router-link
- Extend KsDataTable with a forceExpandedRowKeys prop (pre-expands rows + hides their chevron) and use it from TriggersManage instead of a feature-level :deep selector


MilosPaunovic (approved):
LGTM!

PR #16214, opened 2026-05-26
chore: deprecate TOOL plugin sub group

jymaire:
Deprecate plugin sub group. See https://kestra.io/plugins for current plugin sub groups.


model/src/main/java/io/kestra/core/models/annotations/PluginSubGroup.java
```diff
@@ -26,6 +26,7 @@ enum PluginCategory {
         CLOUD,
         STORAGE,
```
fdelbrayelle, line 34: All of what is NOT on https://kestra.io/plugins should be deprecated (like `STORAGE`)

fdelbrayelle (approved):

PR #16243, opened 2026-05-26
fix(ui): URL-controlled KsDataTable pagination (fixes logs page drift)

flcarre:
## Summary
Fixes a logs pagination bug (URL says `page=2` but the UI renders page 1 and the backend gets `page=1`) and hardens the shared data table behind it.

- **`KsDataTable` is now fully controlled.** Removed the internal `internalPage`/`internalSize` mirror; `props.currentPage`/`props.pageSize` are the single source of truth (URL-bound in callers). The component emits `update:currentPage`/`update:pageSize` (v-model) and reloads via one watcher. This kills the URL↔component drift that caused the bug.
- **Stringified filter watcher.** Callers watched a `computed` that spreads `route.query` into a fresh object with `{deep:true}` — which fires on *every* route change (Vue compares by reference; `deep` doesn't add structural equality). It reset the page to 1 on every pagination click. Replaced with a `JSON.stringify`'d `filterQueryKey` so it only fires on real filter changes.
- **Hostile-input clamping.** `?size=10000000` could become a 10M-row request and `?page=1e308` a garbage offset. `KsDataTable` clamps page/size at the single chokepoint (page ≥ 1 integer ≤ 1e6, size ∈ [1, 1000]) so every caller is protected.
- **Embedded-table URL namespacing.** An embedded `LogsWrapper` (e.g. a `TriggersManage` expand row) shared the outer table's `page` key — paginating the logs moved the triggers table. Embedded instances now use `logsPage`/`logsSize`; both tables keep their page in the URL, no collision.
- Applied the controlled pattern across all 9 data-table callers (Logs, Flows, Executions, KV, Secrets, Triggers, FlowsSearch, Blueprints; MetricsTable via v-model).

## Test Coverage
- New `KsDataTable` contract tests (no internal page state, v-model emits) + 9 clamp tests (negative / fractional / NaN / Infinity / 0 page, size DoS cap, page cap, legitimate pair untouched).
- New `logsWrapperPaginationWiring.spec.ts` harness covering the URL→API→reset wiring.
- **659/659 unit tests pass.** lint clean.

## Pre-Landing Review
Ran `/review` + an independent adversarial subagent. Findings:
- **[FIXED] Uncapped page** → `?page=1e308` reached the backend. Capped at `MAX_PAGE=1e6`.
- **[FIXED] Embedded/outer table page-key collision** in TriggersManage → namespaced to `logsPage`/`logsSize`.
- **[CLEARED] Double-fetch on filter change** — hypothesized, verified it does NOT happen (single fetch).
- **[DEFERRED]** Pre-existing stale-response race in `callLoad` (rapid clicks) — not a regression of this branch, spun out to a follow-up.

## Verification
Verified live in the browser (authed): URL `page`/`size` propagate to the API exactly, clamping neutralizes hostile values (200 instead of 422/10M-row), and the embedded `logsPage` does not leak into the triggers search. Not verifiable in the dev instance: the embedded log table reading `logsPage` (no log data on dev triggers).

## Notes
- Doc added to `ui/AGENTS.md`: the controlled-pagination contract and the deep-watch/computed-spread trap.
- No version bump / CHANGELOG (Kestra uses centralized Gradle versioning, no per-PR VERSION file).

## Test plan
- [x] `vitest --project=unit` — 659 passed
- [x] lint clean on changed files
- [x] live browser verification of URL↔API↔UI sync + clamping

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,4 @@ core/src/main/resources/gradle.properties
 charts/kestra/charts/
 charts/kestra-starter/charts/
 .claude/settings.local.json
+.gstack/
--- a/ui/AGENTS.md
+++ b/ui/AGENTS.md
@@ -70,6 +70,8 @@ Reject (or ask to fix) anything that:
 - Adds a CSS class that overrides `.el-...` selectors.
 - Duplicates a component that already exists in the design system.
 - Adds a `Ks*` component without a Storybook story or test.
+- Mounts `KsDataTable` without binding `:currentPage` / `:pageSize` (or `v-model:currentPage` / `v-model:pageSize`) — pagination is controlled; see "Data tables & pagination state".
+- Watches a `computed` that returns a fresh object (spread / `{...}`) with `{deep: true}` — that fires on every dependency change regardless of content. See "The deep-watch / computed-spread trap".
 
 ### Accessibility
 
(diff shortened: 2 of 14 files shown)


flcarre:
Superseded by #16244 (same commits, branch renamed from the auto-generated name to `fix/datatable-pagination-url-drift`). Closing this one.

PR #16276, opened 2026-05-27
fix(flow): Use the tenant from the URL when updating a concurrency limit

loicmathieu:
Fixes https://github.com/kestra-io/kestra-ee/issues/7982


AcevedoR (approved):

loicmathieu:
Backported to 1.3

PR #16293, opened 2026-05-27
fix(storage): reject backslash path traversal in LocalStorage (GHSA-qw4v-6w32-xx9h)

AcevedoR:
modified PR of @StarPlatinu

## Summary

Fixes the path traversal reported in advisory **GHSA-qw4v-6w32-xx9h** (authenticated arbitrary file read via the execution file-download API).

### Root cause

`LocalStorage.getPath()` ran the parent-traversal guard **before** `windowsToUnixPath()` converted Windows-style backslashes to `/`:

```java
parentTraversalGuard(uri);                                              // (1) validate
return Paths.get(basePath.toString(), windowsToUnixPath(uri.getPath())); // (2) canonicalize, then use
```

`FileUtils.isParentTraversal()` used `File.separator` (always `/` on Linux/containers), so it only recognised forward-slash traversal (`../`, `/..`). A payload using backslashes (`..\..\..`) passed the guard unchanged, was then rewritten to `../../../` by `windowsToUnixPath()`, and escaped the storage base directory — exposing any file readable by the Kestra process (`/etc/passwd`, `/proc/self/environ`, the H2 database, cross-namespace internal storage, …). CWE-22 / CWE-180.

### Fix (defense in depth)

1. **`FileUtils.isParentTraversal`** — add a `String` overload that normalises `\` → `/` before segment matching; the `URI` overload delegates to it. Drops the `File.separator` dependency entirely. This also hardens the shared `StorageInterface.parentTraversalGuard` used by every storage backend.
2. **`LocalStorage.getPath`** — canonicalize separators **before** validating, then add a containment check (`resolved.startsWith(basePath)`) so a resolved path can never escape the base directory even if a future guard regresses.

### Tests

- `FileUtilsTest` — backslash, mixed-separator and percent-encoded payloads are now rejected; filenames that merely contain `..` (e.g. `my..file.txt`) are still allowed; null inputs handled safely.
- `LocalStorageTest` — `shouldRejectBackslashParentTraversal` / `shouldRejectForwardSlashParentTraversal` assert `storageInterface.get()` throws on traversal URIs.

Verified the patched logic against the original PoC payloads: all traversal variants are blocked, all legitimate paths still resolve.

### Notes for maintainers

- Patch targets `develop`. For older release lines where the guard is inlined in `StorageInterface.parentTraversalGuard`, the same `isParentTraversal` hardening should be backported.
- The `ExecutionController.validateFile()` prefix check (`path.getPath().startsWith(prefix)`) is weak but no longer reachable for traversal once the storage layer rejects it; tightening it to a normalised containment check would be a good follow-up hardening.

### DIFF
--- a/core/src/main/java/io/kestra/core/utils/FileUtils.java
+++ b/core/src/main/java/io/kestra/core/utils/FileUtils.java
@@ -3,7 +3,6 @@ package io.kestra.core.utils;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 
-import java.io.File;
 import java.net.URI;
 import java.util.Optional;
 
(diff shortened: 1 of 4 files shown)


loicmathieu (approved):
LGTM.

I wonder if some tests should be added to the `StorageTestSuite` to assert that no other storage implementation shares the same issue.

PR #16295, opened 2026-05-27
fix(helm): replace losisin action with direct helm-docs binary call

### LINKED ISSUE #16294: fix(helm): chart README badges show stale version in published .tgz
## Problem

Since the Helm chart migration to the GCS-backed pipeline, every published chart package contains stale version badges in its `README.md`. For example, `kestra-1.3.20.tgz` contains:

```
![Version: 1.0.53](...)  ![AppVersion: v1.3.14](...)
```

instead of:

```
![Version: 1.3.20](...)  ![AppVersion: v1.3.20](...)
```

Artifact Hub reads the `README.md` from inside the packaged `.tgz` and displays it on the package page — so the displayed badges are wrong for every release from v1.3.16 onwards.

## Root cause

`helm-release.yml` has a **Regenerate helm-docs after version bump** step that runs `losisin/helm-docs-github-action@v1` after bumping `Chart.yaml` to the release version. This action does **not** write the updated `README.md` back to the working directory (acts as a no-op for file writes), so `helm package` bundles the pre-bump README.

The `helm-docs` binary is already installed earlier in the same job (step **Install kubeconform and helm-docs**). The action wrapper is unnecessary and broken.

## Fix

Replace both `uses: losisin/helm-docs-github-action@v1` steps with `run: helm-docs` to use the already-installed binary directly. Verified locally: running `helm-docs` after a `Chart.yaml` version bump correctly regenerates `README.md` with the new version badges.

MarcoSaba:
## Summary

Every published Helm chart `.tgz` from v1.3.16–v1.3.20 contains stale version badges in its `README.md` (showing `Version: 1.0.53` / `AppVersion: v1.3.14`). Artifact Hub displays this README on the package page. The root cause is `losisin/helm-docs-github-action@v1` silently not writing files after the Chart.yaml version bump. Since `helm-docs` is already installed in the same job, this PR replaces both action usages with a direct binary call.

## Changes

- Replace `uses: losisin/helm-docs-github-action@v1` with `run: helm-docs` in **Verify helm-docs are up to date** step
- Replace `uses: losisin/helm-docs-github-action@v1` with `run: helm-docs` in **Regenerate helm-docs after version bump** step

close kestra-io/kestra#16294

### DIFF
--- a/.github/workflows/helm-release.yml
+++ b/.github/workflows/helm-release.yml
@@ -51,7 +51,7 @@ jobs:
           tar -xzf /tmp/hd.tar.gz -C /usr/local/bin helm-docs
 
       - name: Verify helm-docs are up to date
-        uses: losisin/helm-docs-github-action@v2
+        run: helm-docs
 
       - name: Fail on uncommitted README diffs
         run: git diff --exit-code -- 'charts/**/README.md'
@@ -91,7 +91,7 @@ jobs:
           mv /tmp/chart-starter.yaml charts/kestra-starter/Chart.yaml
 
       - name: Regenerate helm-docs after version bump
-        uses: losisin/helm-docs-github-action@v2
+        run: helm-docs
 
       - name: Lint kestra chart
         run: |


gperreymond (approved):

Skraye (approved):
LGTM

PR #16305, opened 2026-05-27
fix(core): register streaming subscriber before resuming queue

AcevedoR:
## Summary

- In `ExecutionStreamingService.registerSubscriber()`, the polling queue was **resumed before the subscriber was added to the map**
- On multi-core machines, the polling thread can wake up between `resume()` and `subscribers.put()` and deliver a terminal `FollowExecutionEvent` to an empty subscriber map, dropping the event
- `Flux.last()` in `TestSuiteService` then hangs indefinitely, making `run_shouldBePersisted` (and similar streaming-backed tests) flaky in CI

**Root cause**: `registerSubscriber` snapshot:
```java
// BEFORE (racy):
if (MapUtils.isEmpty(subscribers) && this.queueSubscriber.isPaused()) {
    this.queueSubscriber.resume();   // ← polling thread can run here
}
subscribers.computeIfAbsent(...).put(...);  // ← subscriber added too late
```

**Fix**: capture `wasEmpty`, register the subscriber first, then resume:
```java
boolean wasEmpty = MapUtils.isEmpty(subscribers);
subscribers.computeIfAbsent(...).put(...);  // ← subscriber visible before resume
if (wasEmpty && this.queueSubscriber.isPaused()) {
    this.queueSubscriber.resume();
}
```

## Test plan

- [x] `MysqlTestSuiteControllerTest` (all 27 tests) pass locally with the fix
- [x] `run_shouldBePersisted` targeted test passes 10/10 times

### DIFF
--- a/core/src/main/java/io/kestra/core/services/ExecutionStreamingService.java
+++ b/core/src/main/java/io/kestra/core/services/ExecutionStreamingService.java
@@ -108,16 +108,13 @@ public class ExecutionStreamingService {
     public void registerSubscriber(String executionId, String subscriberId, FluxSink<Event<Execution>> sink, Flow flow) {
         // it needs to be synchronized as we get and remove if empty, so we must be sure that nobody else is adding a new one in-between
         synchronized (subscriberLock) {
-            boolean wasEmpty = MapUtils.isEmpty(subscribers);
-
             // Register the subscriber BEFORE resuming the queue. If the queue is resumed first,
             // the polling thread can deliver a terminal FollowExecutionEvent between resume() and put(),
             // missing the event and leaving Flux.last() hanging forever.
             subscribers.computeIfAbsent(executionId, k -> new ConcurrentHashMap<>())
                 .put(subscriberId, Pair.of(sink, flow));
 
-            // resume the subscription if it was paused with no subscribers
-            if (wasEmpty && this.queueSubscriber.isPaused()) {
+            if (this.queueSubscriber.isPaused()) {
                 this.queueSubscriber.resume();
             }
         }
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/kestra/core/services/ExecutionStreamingService.java
```diff
@@ -108,13 +108,18 @@ void startQueueConsumer() {
...
+            boolean wasEmpty = MapUtils.isEmpty(subscribers);
 
+            // Register the subscriber BEFORE resuming the queue. If the queue is resumed first,
+            // the polling thread can deliver a terminal FollowExecutionEvent between resume() and put(),
+            // missing the event and leaving Flux.last() hanging forever.
             subscribers.computeIfAbsent(executionId, k -> new ConcurrentHashMap<>())
                 .put(subscriberId, Pair.of(sink, flow));
+
+            // resume the subscription if it was paused with no subscribers
+            if (wasEmpty && this.queueSubscriber.isPaused()) {
```
loicmathieu: I'm not 100% sure we need to check if the subscribers were empty, after all, we should just trust isPause...

loicmathieu (commented):
The same issue should exists in the `LogStreamingService` and the `ExecutionDependenciesStreamingService` .
Can you check and do the necessary changes?

core/src/main/java/io/kestra/core/services/ExecutionStreamingService.java
AcevedoR: done for the other 2, keeping the check it doesnt hurt

loicmathieu (approved):
LGTM

PR #16320, opened 2026-05-28
fix(execution): potential path traversal issues in namespace files

loicmathieu:
Fixes https://github.com/kestra-io/kestra-ee/issues/7988

### DIFF
--- a/core/src/main/java/io/kestra/core/storages/NamespaceFile.java
+++ b/core/src/main/java/io/kestra/core/storages/NamespaceFile.java
@@ -7,6 +7,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import io.kestra.core.models.namespaces.files.NamespaceFileMetadata;
+import io.kestra.core.utils.FileUtils;
 import io.kestra.core.utils.WindowsUtils;
 
 import jakarta.annotation.Nullable;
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/kestra/core/storages/NamespaceFile.java
```diff
@@ -161,12 +162,15 @@ public static Path normalize(Path path) {
         if (!compatiblePath.startsWith("/")) {
             compatiblePath = "/" + compatiblePath;
         }
-        return Path.of(compatiblePath);
+        if (FileUtils.isParentTraversal(compatiblePath)) {
+            throw new IllegalArgumentException("File should be accessed with their full path and not using relative '..' path.");
```
fhussonnois, line 166: For later: could be nice to create a dedicated Exception for that error as we have it in multiple places.

webserver/src/test/java/io/kestra/webserver/controllers/api/NamespaceFileControllerTest.java
```diff
@@ -367,6 +367,25 @@ void forbiddenPaths() {
         assertForbiddenErrorThrown(() -> client.toBlocking().exchange(HttpRequest.DELETE("/api/v1/main/namespaces/" + namespace + "/files?path=/_flows/test.txt", null)));
     }
 
+    @Test
+    void pathTraversalShouldBeRejected() throws IOException, URISyntaxException {
```
fhussonnois, line 371: Let's try to have the same `shouldXxx` everywhere - this avoid LLM picking random naming convention ;)

fhussonnois (approved):
LGTM! minor comments but no blocker for merge

core/src/main/java/io/kestra/core/storages/NamespaceFile.java
loicmathieu: When we'll stabilize all this, yes

webserver/src/test/java/io/kestra/webserver/controllers/api/NamespaceFileControllerTest.java
loicmathieu: too late (and I'm too lazy to change that and wait again for CI)

loicmathieu:
Backported to 1.3

PR #16329, opened 2026-05-28
fix(triggers): populate namespace dropdown in trigger configure modal

MilosPaunovic:
## Summary

- Replace the custom remote `KsSelect` in `TriggerConfigureModal` with the shared `NamespaceSelect` component, which correctly loads namespaces on mount via the autocomplete endpoint
- Add an `autoDefault` prop to `NamespaceSelect` (default `true`) so callers can opt out of the default namespace pre-selection
- Fix `NamespaceSelect` passing a raw string as `ids` when the model value is an empty string

## Test plan

- [ ] Open the Triggers page → click configure on any trigger → namespace dropdown should be populated
- [ ] Selecting a namespace should load the flow dropdown
- [ ] Other uses of `NamespaceSelect` (KV, Secrets, Flows search) should be unaffected

Closes https://github.com/kestra-io/kestra-ee/issues/8068.


Pradumnasaraf (approved):
Looks good to merge!

Image

PR #16339, opened 2026-05-28
add missing padding and margin on FlowRoot and LogsWrapper

flcarre:
### ✨ Description

What does this PR change?  
add missing padding and margin on FlowRoot and LogsWrapper

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra-ee/issues/8011

image
image


ui/src/components/logs/LogsWrapper.vue
```diff
@@ -302,6 +302,7 @@
             border-radius: var(--kel-border-radius-round);
             overflow: hidden;
             padding: 1rem;
+            margin: 0 24px;
```
RemiBarthe: We usually try to avoid hardcoded px values when possible. Prefer using the existing CSS variables first, and if there isn’t one that fits the need, use rem instead

`var(--ks-spacing-5)`

RemiBarthe (approved):

PR #16341, opened 2026-05-28
feat(flows): support label/value pairs in SELECT and MULTISELECT inputs

brian-mulier-p:
## Summary
SELECT/MULTISELECT inputs accept either a plain string or `{label, value}`. UI shows the label, expressions resolve to the value.

The `@Regex` annotation on `values` has been dropped — it was copy-pasted from `Input.validator` (where it correctly checks a regex pattern) onto option values, which are literal strings, not regexes. So it only ever rejected strings that happened to be malformed regex syntax, which nobody writes.

Closes kestra-io/kestra-ee#7488


brian-mulier-p:
The single failing test `ClusterControllerTest.shouldGetWorkerMetrics()` is flaky and also fails on `develop` ([run 26637520434](https://github.com/kestra-io/kestra/actions/runs/26637520434)) — unrelated to this PR.

elevatebart (approved):
Front-end works

PR #16349, opened 2026-05-29
fix(core): delegate TAGS filtering to repository default handlers

### LINKED ISSUE #16323: fix(core): TAGS filter always throws despite advertised operations in QueryFilter
## Problem

`QueryFilter.Field.TAGS.supportedOp()` advertises six operations as valid:
`IN`, `NOT_IN`, `PREFIX`, `CONTAINS`, `STARTS_WITH`, `ENDS_WITH`

`validateQueryFilters()` therefore passes any of these operations without error.

However, `AbstractJdbcRepository.tagsCondition()` unconditionally throws `InvalidQueryFiltersException` for every operation:

```java
protected Condition tagsCondition(Object value, QueryFilter.Op operation) {
    throw new InvalidQueryFiltersException("Unsupported operation for TAGS field: " + operation);
}
```

No OSS subclass overrides `tagsCondition`. This creates a gap between the validation layer (which says the operations are valid) and the execution layer (which always throws).

## Impact

Any query using a TAGS filter — e.g. filtering Apps by tag — passes pre-flight validation and then fails at execution time with a misleading `InvalidQueryFiltersException("Unsupported operation for TAGS field: IN")`, which looks like a client error but is actually a missing implementation.

## Expected behavior

Either:
1. `tagsCondition()` should have a working default implementation (e.g. delegating to `defaultHandlers`), or
2. `TAGS.supportedOp()` should not advertise operations that have no backing implementation, or
3. Subclasses that support TAGS filtering must override `tagsCondition()` and this requirement should be enforced (e.g. via `abstract` or clear documentation).

## Steps to reproduce

Send a TAGS filter request to any endpoint backed by a repository that does not override `tagsCondition` (e.g. App search with `?filters=[{field:"tags",operation:"IN",value:["my-tag"]}]`).

sagarkhandagre998:
Fixes #16323 

### Summary

This PR fixes an inconsistency between query filter validation and JDBC execution for `QueryFilter.Field.TAGS`.

`TAGS.supportedOp()` currently advertises support for:

* `IN`
* `NOT_IN`
* `PREFIX`
* `CONTAINS`
* `STARTS_WITH`
* `ENDS_WITH`

These filters pass `validateQueryFilters()`, but `AbstractJdbcRepository.tagsCondition()` always throws `InvalidQueryFiltersException`, causing runtime failure.

### Changes

Updated `AbstractJdbcRepository.tagsCondition()` to delegate to the existing repository filtering logic instead of unconditionally throwing.

```java
return defaultHandlers(QueryFilter.Field.TAGS, value, operation);
```

### Result

This aligns execution behavior with the operations already exposed by `QueryFilter.Field.TAGS.supportedOp()` and prevents TAGS filters from failing after validation.

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
@@ -475,6 +475,10 @@ public abstract class AbstractJdbcRepository {
         throw new InvalidQueryFiltersException("getSuperAdminCondition must be overridden for JSONB-backed superAdmin field");
     }
 
+    protected Condition tagsCondition(Object value, QueryFilter.Op operation) {
+        return defaultHandlers(QueryFilter.Field.TAGS, value, operation);
+    }
+
     // Generate the condition for Field.STATE
     @SuppressWarnings("unchecked")
     protected Condition generateStateCondition(Object value, QueryFilter.Op operation) {
--- a/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
@@ -510,10 +510,6 @@ public abstract class AbstractJdbcRepository {
         return defaultHandlers(QueryFilter.Field.NAME, value, operation);
     }
 
-    protected Condition tagsCondition(Object value, QueryFilter.Op operation) {
-        throw new InvalidQueryFiltersException("Unsupported operation for TAGS field: " + operation);
-    }
-
     protected Condition typeCondition(Object value, QueryFilter.Op operation) {
         return defaultHandlers(QueryFilter.Field.TYPE, value, operation);
     }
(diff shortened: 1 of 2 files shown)


sagarkhandagre998:
@loicmathieu Please take a look

loicmathieu (changes requested):

Can you add a test so we know that it works now?

MilosPaunovic:
Also, could you resolve merge conflicts in this PR please @sagarkhandagre998?

sagarkhandagre998:
@MilosPaunovic @loicmathieu Test passed locally.

loicmathieu (approved):
LGTM

loicmathieu:
@sagarkhandagre998 it didn't compile, can you fix it and makes sure everything compile and your test work before updating the PR

sagarkhandagre998:
> @sagarkhandagre998 it didn't compile, can you fix it and makes sure everything compile and your test work before updating the PR

sorry for that , I forgot to remove the duplicate method , Now its fine.

Test Passed Locally

sagarkhandagre998:
@loicmathieu Can you please run the workflow

loicmathieu (approved):
LGTM

PR #16372, opened 2026-05-29
fix(system): init metrics whenever possible

loicmathieu:
This would set them to 0 and avoid NoData in Prometheus.

Fixes https://github.com/kestra-io/kestra-ee/issues/4206


nkwiatkowski (approved):
LGTM

PR #16374, opened 2026-05-29
fix(execution): markAs should also mark the taskrun attempt

### LINKED ISSUE #15622: SLAs successfully FAIL Workflow, but keep task as RUNNING
### Describe the issue

SLAs successfully FAIL Workflow, but keeps task as RUNNING. It appears it running infinite, I don't think it's actually running though, but UI keeps counting it up.

CANCEL appears to stop and KILL the task, just happening on FAIL.

Image 1 showing SLA hit.
Image

Image 2 showing FAILed task but running in the log section.
Image

workflow
```yaml
id: helloWorld
namespace: demo
description: |
  This flow logs a message to the console, and used for demoing and testing purposes.

sla:
  - id: maxDuration
    type: MAX_DURATION
    duration: PT1S
    behavior: FAIL
    labels:
      sla: miss
      reason: durationExceeded

tasks:
  - id: hello
    script: |
      sleep 10
```

Thank you, let me know if you need any other details.

### Environment

- Kestra Version: 1.1.10
*we are upgrading to LTS in a week or so, sorry if this is already addressed in LTS.

loicmathieu:
Fixes https://github.com/kestra-io/kestra/issues/15622

Note that SLA FAIL/CANCEL is inherently racy as late worker tasks may update the execution so it may not fix all edge cases.

### DIFF
--- a/core/src/test/java/io/kestra/core/runners/SLATestCase.java
+++ b/core/src/test/java/io/kestra/core/runners/SLATestCase.java
@@ -1,10 +1,13 @@
 package io.kestra.core.runners;
 
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeoutException;
 
 import io.kestra.core.models.Label;
 import io.kestra.core.models.executions.Execution;
+import io.kestra.core.models.executions.TaskRun;
+import io.kestra.core.models.executions.TaskRunAttempt;
 import io.kestra.core.models.flows.State;
 import io.kestra.core.queues.QueueException;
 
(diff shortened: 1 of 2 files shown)


AcevedoR (commented):
> Note that SLA FAIL/CANCEL is inherently racy as late worker tasks may update the execution so it may not fix all edge cases.

could we not implement a rule that prevent late updates from worker tasks in this case ?

loicmathieu:
> could we not implement a rule that prevent late updates from worker tasks in this case ?

No it's more important that a user know that a task ends successfully as otherwize he may restart a long potentially with side effect task by mistake.
We must always process late worker task result.
If there are still races, we should try to mitiguate them... or leave with them...

AcevedoR (approved):

loicmathieu:
Backported to 1.0 and 1.3

PR #16380, opened 2026-05-29
fix(scheduler): restrict JdbcCleaner to executor to prevent deadlocks

nkwiatkowski:
## Summary
- `JdbcCleaner` runs on all Kestra components (webserver, executor, worker, scheduler). When two components fire their cleanup simultaneously, they deadlock on the `queues` table, causing executor stalls requiring manual restarts.
- Added `@Requires(property = "kestra.server-type", pattern = "(EXECUTOR|STANDALONE)")` to scope the cleaner to the executor only, following the same pattern used by `JdbcServiceLivenessCoordinator`.

Closes: kestra-io/kestra-ee#7919

## Test plan
- [x] Existing `*CleanerTest` tests pass (H2, Postgres, MySQL) — test configs use `STANDALONE` which matches the new pattern
- [ ] Verify in a multi-component deployment that only the executor runs the cleaner


AcevedoR (approved):
LGTM

PR #16386, opened 2026-05-29
fix(execution): change state warning

loicmathieu:
Changing a state display a warning if it was to the same type.

Here, it's most of the time the case so we need to do it conditionnaly.


loicmathieu:
@brian-mulier-p this is what we discussed recently after your PR on flow trigger processing on intermediatary states

nkwiatkowski (approved):
LGTM

PR #16389, opened 2026-05-29
fix(security): derive CSRF cookie secure flag from request scheme

nkwiatkowski:
## Summary
- The CSRF cookie was hardcoded with `secure(true)`, preventing browsers from sending it over HTTP
- All POST/PUT/DELETE requests failed with 403 Forbidden in non-HTTPS environments (local dev, HTTP proxies)
- Changed to `request.isSecure()` so the cookie's secure flag matches the transport — HTTPS users still get `Secure` cookies, HTTP users get functional ones

## Test plan
- [x] `CsrfTokenFilterTest` passes
- [x] `StaticFilter` tests pass
- [ ] CI green


loicmathieu (approved):
LGTM

PR #16390, opened 2026-05-29
fix(design-system): square the icon-only copy button

flcarre:
## What

The copy-to-clipboard button rendered as a **52×32px rectangle** around a 14×14 icon because the icon-only case kept the default text-button horizontal padding (`8px 15px`). This made the background look oversized with a wrong aspect ratio, most visibly on the Blueprint **Source** block.

This constrains the button to a **square (2rem)** when no `label` is provided, mirroring the existing `KsIconButton` sizing pattern (fixed `width`/`height`/`min-width`, no `:deep`, no `.el-button` override). An accessible label is also added for the icon-only variant.

The labelled variant (when `label` is set) is unchanged.

## Scope

`CopyToClipboard` is shared across 6 icon-only usages (Blueprint Source, execution logs toolbar, file preview, execution outputs, webhook curl, curl). All now render a consistent square button — e.g. on the execution logs toolbar the copy icon now lines up with the adjacent download/refresh icons.

## Verification

- Blueprint Source copy button: `52×32` → `28×28` (square), icon centered.
- Checked in **light and dark** mode.
- Spot-checked execution logs page: 3 copy buttons all square, no layout breakage.

## Screenshot
image

Closes https://github.com/kestra-io/kestra-ee/issues/8115


elevatebart (approved):
LGTM

PR #16393, opened 2026-05-29
Feat/small perf improvement

loicmathieu:
- Avoid creating an ArrayList to compute metrics in the jOOQ execute listener
- Avoid computing twice a RunContext for some message and merge flowable processing in a single pass
- Small improvement to dynamic task run processing (don't rely on exception for control flow!)


nkwiatkowski (approved):
LGTM

PR #16399, opened 2026-05-30
fix(ui): render trigger-specific properties in TriggerConfigureModal

### LINKED ISSUE #16077: [Triggers] Configure modal "Form" tab is missing the trigger-specific properties (e.g. cron for Schedule)
### Describe the issue

On the Triggers → Add page, clicking Add on the Schedule card opens the configure modal. The Form tab only shows three fields: Namespace, Flow, and Trigger ID. The cron expression is missing.

The same problem applies to other triggers. Webhook key, HTTP url, MCP Tool toolName, etc., are also not available in the modal.

Steps to reproduce:
1. Go to Triggers → Add.
2. Click Add on the Schedule card.
3. Only Namespace, Flow, and Trigger ID fields are shown.

Image

### Expected Behaviour
The modal should also show the trigger-specific fields (cron for Schedule, key for Webhook, url for HTTP, etc.) so the trigger can be fully configured from the modal.

### Environment

- Kestra Version: 2.0 develop
- Setup: Standalone with Postgres

blaspat:
## Description
Add trigger-specific properties (e.g. cron for Schedule, key for Webhook) to the Form tab of the TriggerConfigureModal.

Previously, the form only showed Namespace, Flow, and Trigger ID fields. The cron expression and other trigger-specific fields were missing, forcing users to configure them manually in the YAML editor after adding the trigger.

Fixes #16077


RemiBarthe:
Thanks for tackling this! Unfortunately the current approach doesn't work yet.

1. It doesn't render the trigger fields. Testing Schedule, the form shows $schema, properties, required, title… — the top-level keys of the JSON Schema wrapper, not the class fields (cron, timezone…). 
Capture d’écran 2026-06-01 à 11 20 27

2. Don't hand-roll a schema => form renderer. The no-code editor already has a generic one (`ui/src/components/no-code/components/tasks/TaskObject.vue`)

blaspat:
Hi @RemiBarthe ,
I've already updated the PR using `TaskObject.vue`. Could you take another look?
Thanks

MilosPaunovic:
Hey @blaspat, 

Sorry for us taking so long to review this.
Can you resolve the merge conflicts that exist now and we'll get right on it?

flcarre (changes requested):
Surfacing the trigger-specific fields is a real improvement. Two things I'd love you to check:

1. **Schema injection context.** `TaskObject` (and its children) resolve `$ref`/object-typed fields through `inject(FULL_SCHEMA_INJECTION_KEY)` / `SCHEMA_DEFINITIONS_INJECTION_KEY` / `BLOCK_SCHEMA_PATH_INJECTION_KEY`. The other callers (`NoCode.vue`, `TaskEditor.vue`, `DashboardNoCodeEditor.vue`) `provide()` those before mounting it, but I don't see them provided here, so object-typed inherited fields like `labels` or `workerSelector` may render blank instead of a proper editor. Could you provide the same triad from the fetched trigger schema? This tends to fail silently, which is why static checks pass.

2. **Required-field gating.** `canSubmit` currently checks only `namespace`/`flowId`/`triggerId`, so a Schedule with an empty `cron` can still be added. Would it make sense to also require the schema's required keys, e.g. `triggerSchema.required?.every(k => !!model[k])`?

The schema traversal and reactivity handling otherwise look solid.

blaspat:
Addressed ficarre's two review points:

1. **Schema injection context** — Now providing , , and  from the fetched trigger plugin schema, so TaskObject can resolve  fields (inherited labels, workerSelector etc. will render correctly).

2. **Required-field validation** —  now checks that all schema-required fields (e.g. cron for Schedule) have values before allowing submission, in addition to the existing namespace/flowId/triggerId checks.

Also merged latest upstream develop to resolve the merge conflicts.

blaspat:
Addressed ficarre's two review points:

1. **Schema injection context** — Now providing `FULL_SCHEMA_INJECTION_KEY`, `SCHEMA_DEFINITIONS_INJECTION_KEY`, and `BLOCK_SCHEMA_PATH_INJECTION_KEY` from the fetched trigger plugin schema, so TaskObject can resolve `$ref` fields (inherited labels, workerSelector etc. will render correctly).

2. **Required-field validation** — `canSubmit` now checks that all schema-required fields (e.g. cron for Schedule) have values before allowing submission, in addition to the existing namespace/flowId/triggerId checks.

Also merged latest upstream develop to resolve the merge conflicts.

flcarre (changes requested):
Thanks for wiring the injection triad, I traced it through the form components and `$ref` fields resolve correctly now. Two things:

- The schema fetch doesn't pass `all: true`, so `labels`/`workerSelector` are still excluded server-side (hidden base-class fields). If the intent is to defer those to the flow editor, fine by me, let's just say it explicitly.
- `schema` is typed `JSONSchema`, which doesn't match `FULL_SCHEMA_INJECTION_KEY`'s required `{definitions, $ref}` shape, does `npm run check:types` pass locally?

blaspat:
Thanks for the review,

1. `all: true` added to the load call. Now `labels`, `workerSelector`, and other inherited fields come through from the server.
2. Type mismatch, yes `vue-tsc` would flag this. Added an explicit type assertion matching what `DashboardNoCodeEditor` does, so the provided ref satisfies {`definitions`, `$ref`}.

flcarre (approved):
Thanks for the work !

PR #16417, opened 2026-06-01
chore(deps): Upgrade Pebble to 4.1.2

### LINKED ISSUE #15048: Passing Pebble-combined lists to a subflow rewrites the first output list with all combined values
### Describe the issue

We have a flow with the following steps and inputs.

## Input
```yaml
inputs:
  - id: stage_tables
    type: ARRAY
    itemType: STRING
    required: true
    defaults: |
      ["a1", "a2", "a3"]
```

The KV store contains the key **`dq_excluded_tables`** with the JSON value:

```json
["test"]
```

## Tasks

1. **Python task**

   A Python script sets the Kestra output **`failed_tables`** to:

   ```
   failed_table1
   ```

2. **Log task**

   Logging the output from step 1 correctly prints:

   ```
   failed_table1
   ```

3. **Subflow call**

   We call a subflow (which only contains a log task for simulation purposes).

   The subflow is called with the following input:

   ```yaml
   excluded_tables: "{{ (outputs.test_produce_output.vars.failed_tables | default([])) + (kv('dq_excluded_tables', errorOnMissing=false) | default([])) + (inputs.stage_tables | default([])) }}"
   ```

   This combines:
   - KV store value
   - `stage_tables` input
   - `failed_tables` output from step 1

4. **Log task**

   The same log task as in step 2 is executed again.

   **Unexpected behavior:**  
   Instead of printing only:

   ```
   failed_table1
   ```

   it prints:
   - `failed_table1`
   - all values from the KV store
   - all values from `stage_tables`

## Additional Observations

- If we **remove the subflow call (step 3)**, step 4 behaves correctly and prints only:
  ```
  failed_table1
  ```

- This suggests that when sending the **`excluded_tables`** parameter, the KV and input values somehow get appended to the first item (`outputs.failed_tables`), although this should not happen.

- **Important:** If we **reverse the order** in the `excluded_tables` combination and move `outputs.failed_tables` to the **last position**, the issue does **not** occur.

Testing YAML + Python files attached.
[test.michaela_kestra_bug.zip](https://github.com/user-attachments/files/25936497/test.michaela_kestra_bug.zip)

**Screenshots**:
(1) Including **transfer (step3)** task:
Image

(1) Output:
Image

(2) When **transfer (step3) task is commented out**:
Image

(2) Output:

Image

---
### Additional details/investigation

The root cause seems to be as follows:

1. Pebble's implementation of List1 + List2 is: List1.addAll(List2) ([code](https://www.javatips.net/api/Pebble-master/src/main/java/com/mitchellbosecke/pebble/utils/OperatorUtils.java)) so basically it mutates List1 in place
2. If Pebble needs to just print the expression (i.e. in a Log task) the intermediate result of the expression (the mutated list) is not preserved, BUT when Pebble needs to use (serialize?) the expression as input for a subflow task, it gets preserved and so subsequent uses of the variable List1 are mutated.

### Environment

- Kestra Version: 1.2.3 (non-enterprise)

loicmathieu:
Closes #16407
Fixes #15048


nkwiatkowski (approved):

loicmathieu:
Backported to 1.3

PR #16419, opened 2026-06-01
fix(tests): make shouldGetWorkerMetrics robust against async metrics persistence

AcevedoR:
## Summary

- `shouldGetWorkerMetrics` was flaky because it asserted metrics non-empty immediately after `@BeforeEach` waited for `worker.getState() != null`
- Worker metrics are persisted to `ServiceInstanceRepository` asynchronously via liveness heartbeats (`ServiceLivenessManager`), so the repository could be empty at assertion time on a loaded CI runner
- Wrapped the API call in `Awaitility.untilAsserted()` (10 s / 200 ms poll) so the test retries until running-worker metrics actually appear in the repository

## Test plan

- [ ] `./gradlew :webserver:test --tests "io.kestra.webserver.controllers.api.ClusterControllerTest"` passes


loicmathieu (approved):
LGTM

PR #16421, opened 2026-06-01
fix(design-system): capture KsDrawer resize drag with an overlay

flcarre:
Follow-up to kestra-io/kestra#16388 (resizable KsDrawer). The resize listeners were on `document`, so an editor under the cursor (Monaco, in the EE audit-log diff) could swallow `mousemove` and stall the drag. On `mousedown`, attach a full-viewport overlay that captures move/up during the drag, then remove it on release — robust over editors and iframes.

Verified on a local EE audit-diff drawer: drag resizes continuously over the Monaco diff; the header expand/collapse icon stays coherent (collapse >=95%, expand below).

Generated with Claude Code

### DIFF
--- a/ui/packages/design-system/src/components/Feedback/KsDrawer.vue
+++ b/ui/packages/design-system/src/components/Feedback/KsDrawer.vue
@@ -92,19 +92,22 @@
         const panel = (event.target as HTMLElement).closest(".kel-drawer") as HTMLElement | null
         if (!panel) return
 
+        const overlay = document.createElement("div")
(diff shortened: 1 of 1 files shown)


flcarre:
Superseded by the native-resize fix: the overlay approach is unnecessary because Element Plus's native ElDrawer resize already listens on `window` (works over Monaco). The real bug was our custom handle conflicting with EP's native dragger — fixed by dropping the custom handle entirely. See the replacement PR.

PR #16425, opened 2026-06-01
refactor(executor): move kill-switch logic into message handlers

AcevedoR:
- advance on https://github.com/kestra-io/kestra-ee/issues/7292?issue=kestra-io%7Ckestra-ee%7C8164

## Summary

- Every `DefaultExecutor` queue handler now follows the same 3-line pattern as `executionCommandQueue`: deserialize check → delegate to handler → done
- Kill-switch logic is moved into each dedicated message handler; `DefaultExecutor` no longer holds `KillSwitchService` or `KillSwitchActionService` injections
- New `ExecutionMessageHandler` encapsulates the legacy `executionQueue` path (persist-always + kill-switch guard + delegate to `ExecutionEventMessageHandler`)
- Fixes two pre-existing bugs in the process:
  - `executionEventQueue`: null-dereference when `findById` returned null before the `isKillSwitched` check
  - `workerTaskResultQueue`: called `killSwitchActionService.handle()` unconditionally without `findById + isKillSwitched`, causing redundant kills for already-killing executions

## Test plan

- [ ] 58 unit tests passing (`./gradlew :executor:test --tests "io.kestra.executor.handler.*"`)
- [ ] New Mockito-based kill-switch test classes for every handler following the `ExecutionCommandMessageHandlerTest` pattern
- [ ] All existing integration tests (`@KestraTest`) unchanged and passing


executor/src/test/java/io/kestra/executor/handler/ExecutionEventMessageHandlerKillSwitchTest.java
loicmathieu, line 1: Please move these tests inside the existing `ExecutionEventMessageHandlerTest` one test by component is enought (do this also for the other tests).

executor/src/main/java/io/kestra/executor/handler/ExecutionMessageHandler.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        EvaluationType evaluationType = killSwitchService.evaluate(message);
+        if (evaluationType.isKillSwitched(message)) {
+            killSwitchActionService.handle(evaluationType, message.getTenantId(), message.getId());
+            return Optional.empty();
+        }
+
+        var eventType = message.getState().isCreated() ? ExecutionEventType.CREATED : ExecutionEventType.UPDATED;
+        return executionEventMessageHandler.handle(new ExecutionEvent(message, eventType));
```
loicmathieu: I'm not fan of an handler calling another one.
As it's an optimization, I think it would be better inside the DefaultExecutor.

loicmathieu (changes requested):

executor/src/test/java/io/kestra/executor/handler/ExecutionEventMessageHandlerKillSwitchTest.java
AcevedoR: done

executor/src/main/java/io/kestra/executor/handler/ExecutionMessageHandler.java
AcevedoR: I put all this logic back into the DefaultExecutor, but I'm not sure if its better or not, WDYT ?

loicmathieu (approved):
LGTM

PR #16427, opened 2026-06-01
 feat(core): fallback SDK auth to OSS basic auth

### LINKED ISSUE #16161: Configure SDK default auth with OSS auth
### Feature description

In OSS, tasks using the SDK must configure it globally inside the application configuration.
As in OSS, there is a single user, we might be able to pre-configure the SDK global auth based on the OSS single authentication.

This would allow OSS users to use any tasks talking to the SDK without any configuration at all.

sagarkhandagre998:
## Summary

Configure SDK default authentication from OSS basic-auth configuration by default.

## Changes

- map `kestra.tasks.sdk.authentication.username` to `kestra.server.basic-auth.username`
- map `kestra.tasks.sdk.authentication.password` to `kestra.server.basic-auth.password`
- ignore blank SDK auth values in `RunContextSDKFactory`

## Notes

The SDK auth properties are resolved using Micronaut placeholders with an empty default:

`"${kestra.server.basic-auth.username:}"`

When `kestra.server.basic-auth` is not configured, these placeholders may resolve to empty strings instead of missing values.

Without filtering, empty strings could be treated as configured credentials by `RunContextSDKFactory`, resulting in SDK authentication being considered present with blank username/password.

Blank values are therefore ignored so they behave the same as an unset configuration.

## Fixes #16161


loicmathieu (changes requested):
I think an easier approach would be to do it directly inside the `application.yaml` file so that no code is needed and it has no risk to do anything on EE.

sagarkhandagre998:
> > I think an easier approach would be to do it directly inside the `application.yaml` file so that no code is needed and it has no risk to do anything on EE.
> 
@loicmathieu  Can you please review this now .

sagarkhandagre998:
@loicmathieu Do we also need a test for this ?

loicmathieu (approved):
LGTM, thanks

PR #16438, opened 2026-06-01
fix(design-system): drive KsDrawer full-screen icon from native resize

flcarre:
## Problem

On the Audit Logs page (EE), opening the "Changes applied to the object" diff drawer shows a narrow side-by-side view: the JSON gets jammed and wrapped, making it hard to read and compare, with no way to enlarge the drawer.

## What this does

Lets the user **resize the drawer** to read the diff comfortably:

- Drag the drawer's edge to widen it, all the way to full screen.
- Or use the header button to toggle full screen.
- The header icon reflects the current state — *expand* when there's room to grow, *collapse* when already at full width.

It relies on Element Plus's native drawer resize, so dragging keeps working even over the embedded Monaco diff editor; the width on release drives the full-screen state and the icon.

## Result

The audit-log diff drawer can be dragged wider or sent full screen, giving a readable side-by-side comparison. Covered by a unit test asserting the icon reflects the full-screen state.

Closes kestra-io/kestra-ee#8104

EE wiring of the audit-log diff drawer: kestra-io/kestra-ee#8148


RemiBarthe (approved):
Much cleaner 👍

PR #16441, opened 2026-06-01
refactor(core): move AsyncOperationWaiter from webserver to core

AcevedoR:
## Summary

- Moves `AsyncOperationWaiter` from `io.kestra.webserver.services` to `io.kestra.core.services` so it can be reused from `WebhookService` and other core services without creating a dependency on the webserver module.
- Updates all callers in `ExecutionController`, `TriggerStateService`, and `AsyncOperationWaiterTest` to import from the new location.
- No behavior changes.

## Test plan

- [ ] `./gradlew :core:compileJava :webserver:compileJava` passes (verified locally)
- [ ] Existing `AsyncOperationWaiterTest` tests still pass (no logic changed)

### DIFF
--- a/core/src/main/java/io/kestra/core/services/AsyncOperationWaiter.java
+++ b/core/src/main/java/io/kestra/core/services/AsyncOperationWaiter.java
@@ -1,4 +1,4 @@
-package io.kestra.webserver.services;
+package io.kestra.core.services;
 
 import io.kestra.core.async.AsyncOperationProcessedEvent;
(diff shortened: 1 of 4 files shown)


loicmathieu (approved):
LGTM but I wonder if the opposite would not be better.
Webhook are part of the webserver, the service might be moved their instead

AcevedoR:
this WebhookService is passed in the WebhookContext and used at least in io.kestra.plugin.core.trigger.Webhook
So I'm not sure if it can be done

PR #16442, opened 2026-06-01
Fix: trigger execution

fhussonnois:
When a polling or schedule trigger has allowConcurrent=false, record the id
of the execution holding the lock on the trigger state and clear it when the
execution terminates or the trigger is reset. Expose it through ApiTriggerState.

No DB migration: the execution_id generated column on the triggers table
already reads the 'executionId' JSON key in all three dialects.


ammeek (approved):
LGTM

PR #16444, opened 2026-06-01
fix(core): make Webhook wait for execution to be created

AcevedoR:
- fixes https://github.com/kestra-io/kestra-ee/issues/8170

## Summary

Since v2, the webhook API was fire-and-forgetting the `Create` execution command: the HTTP response could return before the executor had persisted the execution, causing callers to receive an execution ID that did not yet exist in the repository.

This PR fixes the race by:

- **Rewrites `WebhookService.startExecution`** to use `AsyncOperationWaiter.submit()` instead of a bare `executionCommandQueue.emit()`. The `Create` command is now tagged with an `operationId`, and the method returns `Mono<AsyncOperationProcessedEvent>` — the response is only sent once the executor confirms the execution is created.
- **Updates callers** (`Webhook`, `WebhookTestPlugin`) to chain on the returned `Mono`, mapping it to the appropriate `HttpResponse` and propagating errors reactively.
- **Moves `AsyncOperationsConfiguration`** from `io.kestra.webserver.configuration` to `io.kestra.core.async` so it is accessible to core services; updates `ExecutionController` and `TriggerStateService` imports accordingly.

## Test

`ExecutionControllerRunnerTest.shouldPersistExecutionBeforeWebhookResponds` — calls the webhook endpoint, then immediately asserts `executionRepository.findById(id)` is present with no polling. This would be a race before the fix; it is guaranteed after.

### DIFF
--- a/core/src/main/java/io/kestra/core/services/WebhookService.java
+++ b/core/src/main/java/io/kestra/core/services/WebhookService.java
@@ -7,6 +7,8 @@ import java.util.stream.Collectors;
 import org.apache.hc.core5.http.NameValuePair;
 import org.apache.hc.core5.net.URIBuilder;
 
+import io.kestra.core.async.AsyncOperationProcessedEvent;
+import io.kestra.core.async.AsyncOperationsConfiguration;
 import io.kestra.core.events.CrudEvent;
 import io.kestra.core.models.Label;
 import io.kestra.core.models.executions.Execution;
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/kestra/core/services/WebhookService.java
```diff
@@ -156,13 +167,25 @@ public void startExecution(Execution execution) throws QueueException {
...
-        eventPublisher.publishEvent(CrudEvent.create(execution));
+            .withTraceParent(traceParent.orElse(null));
+
+        return asyncOperationWaiter.submit(
+            execution.getId(),
+            operationId -> {
+                try {
+                    executionCommandQueue.emit(command.withOperationId(operationId));
+                    eventPublisher.publishEvent(CrudEvent.create(execution));
+                } catch (Exception e) {
```
loicmathieu, line 183: This is a smell!

By the way, you can use `throwConsumer()` instead it's more elegant

loicmathieu (approved):
LGTM

Issue #8740, opened 2025-05-13
Allow referencing multiple `pluginDefaults` blocks in a single flow by the ID of that pluginDefaults block


anna-geller:
### Feature description

In a workflow with many tasks of the same type (e.g. 20+), user wants to use different `pluginDefaults` configs depending on the task type—some are I/O-intensive, others are CPU-intensive.

Currently, there's no way to define multiple `pluginDefaults` blocks and reference it in a task by ID. User would like to do something like this:

```yaml
pluginDefaults:
  - ref: io-config
    type: io.kestra.plugin.scripts.python.Script
    values:
      taskRunner:
        type: io.kestra.plugin.scripts.runner.docker.Docker
        cpu:
          cpus: 1
        memory:
          memory: "512Mi"
      containerImage: io-image:latest

  - ref: cpu-config
    type: io.kestra.plugin.scripts.python.Script
    values:
      taskRunner:
        type: io.kestra.plugin.scripts.runner.docker.Docker
        cpu:
          cpus: 10
        memory:
          memory: "1024Mi"
      containerImage: cpu-image:latest
```

And then in tasks:

```yaml
pluginDefaultsRef: io-config
```

This would help avoid redefining settings per task. For those stored on a namespace level in EE this would become even more useful -- user could define many plugin defaults and reference the one they need to keep their workflows DRY.

## Next steps

- add core task property called `pluginDefaultsRef`
- add autocomplete (dropdown) on a namespace level if possible, if not then on a tenant level in EE
- it should work with both OSS (when `pluginDefaults` set in a flow or configuration has `ref`) and EE namespace-level `pluginDefaults` — `ref` being an optional property

anna-geller:
Use cases:
- switching between environments
- applying slightly different config for one set of tasks etc

ETA of complexity: 1 week of work

anna-geller:
@AcevedoR A quick note that we will need to make this work not only for tasks but also for triggers and task runners -- wherever we rely on plugins.

Ben8t:
Moving this back to triage and Core squad - it isn't pluginDefaults as defined and more something to define in the core of flow/tasks.

pluginDefaults supply **fallback values for a plugin *type*** when none is provided (closest scope wins). This request is the opposite shape: an explicit, per-task selection of a named config bundle. That's a reusable **`taskBlock`** defined at Namespace/Tenant level and referenced from a flow — a different primitive worth its own design.

[...]

PR #16446, opened 2026-06-01
feat(plugindefaults): allow referencing named pluginDefaults by ref (#8740)

brian-mulier-p:
Implements kestra-io/kestra#8740.

Adds a named-reference mechanism on top of `pluginDefaults`:

- `pluginDefaults` entries accept an optional `ref` id; tasks, triggers and task runners accept `pluginDefaultsRef` to opt into a named bundle.
- A `ref` bundle applies only to plugins that reference it, never by type matching. A plugin that sets `pluginDefaultsRef` ignores non-forced type-matched defaults.
- Refs resolve across levels (flow + namespace in EE) with whole-bundle shadowing — one winner per ref. Non-forced: highest level wins (flow > namespace > global). Forced: lowest/admin level wins (global > namespace > flow) so enforcement can't be bypassed.
- Forced (admin-enforced) defaults always apply even to a plugin declaring `pluginDefaultsRef`; a WARN is logged that the referenced bundle is ineffective for the enforced properties.
- An unknown `pluginDefaultsRef` is an error: flow validation reports it in the editor, and tasks/triggers fail at runtime with the new `PluginDefaultsRefNotFoundException`.
- Editor autocomplete suggests the flow's own `ref` ids (namespace-level refs added in the EE companion PR).

Companion EE PR added below.

Tested: `PluginDefaultServiceTest` (unit + override), H2 runner e2e for the runtime failure, autocomplete provider spec.

### DIFF
--- a/core/src/main/java/io/kestra/core/exceptions/PluginDefaultsRefNotFoundException.java
+++ b/core/src/main/java/io/kestra/core/exceptions/PluginDefaultsRefNotFoundException.java
@@ -1,0 +1,17 @@
+package io.kestra.core.exceptions;
+
+import java.io.Serial;
+
+/**
+ * Thrown when a plugin (task, trigger or task runner) references named plugin-defaults via
+ * {@code pluginDefaultsRef} that cannot be resolved — either none exists with that {@code ref} at flow,
+ * namespace or global level, or the one that exists is scoped to a different plugin type.
+ */
+public class PluginDefaultsRefNotFoundException extends KestraRuntimeException {
+    @Serial
+    private static final long serialVersionUID = 1L;
+
+    public PluginDefaultsRefNotFoundException(String ref) {
+        super("No plugin-defaults exists for pluginDefaultsRef: '" + ref + "'");
+    }
+}
--- a/core/src/main/java/io/kestra/core/services/FlowService.java
+++ b/core/src/main/java/io/kestra/core/services/FlowService.java
@@ -30,6 +30,7 @@ import io.kestra.core.models.topologies.FlowTopology;
 import io.kestra.core.models.triggers.AbstractTrigger;
 import io.kestra.core.models.triggers.TriggerId;
 import io.kestra.core.models.triggers.WorkerTriggerInterface;
+import io.kestra.core.models.validations.ManualConstraintViolation;
 import io.kestra.core.models.validations.ModelValidator;
 import io.kestra.core.models.validations.ValidateConstraintViolation;
 import io.kestra.core.plugins.PluginRegistry;
@@ -130,7 +131,9 @@ public class FlowService {
         // Validate Flow with defaults values
         // Do not perform a strict parsing validation to ignore unknown
         // properties that might be injecting through default values.
-        modelValidator.validate(pluginDefaultService.injectAllDefaults(parsed, false));
+        FlowWithSource withDefaults = pluginDefaultService.injectAllDefaults(parsed, false);
+        modelValidator.validate(withDefaults);
+        validatePluginDefaultsRefs(withDefaults);
 
         FlowWithSource created = flowRepository.create(flow);
 
@@ -162,7 +165,9 @@ public class FlowService {
         // Validate Flow with defaults values
         // Do not perform a strict parsing validation to ignore unknown
         // properties that might be injecting through default values.
-        modelValidator.validate(pluginDefaultService.injectAllDefaults(parsed, false));
+        FlowWithSource withDefaults = pluginDefaultService.injectAllDefaults(parsed, false);
+        modelValidator.validate(withDefaults);
+        validatePluginDefaultsRefs(withDefaults);
 
         FlowWithSource updated = flowRepository.update(flow, previous);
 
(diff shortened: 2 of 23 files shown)


core/src/main/java/io/kestra/core/services/PluginDefaultService.java
```diff
@@ -571,4 +643,90 @@ Object recursiveDefaults(Object object, Map<String, List<PluginDefault>> default
...
+            // unknown ref: keep 'pluginDefaultsRef' so a surviving key signals an unresolved reference
+            // (flagged as a validation error in the editor and failed at runtime).
+            log.warn("No pluginDefaults bundle found for ref '{}' referenced by plugin '{}'", refId, plugin.get("type"));
+            return plugin;
+        }
+
+        Map<String, Object> result;
+        if (pluginDefault.isForced()) {
+            // forced bundle overrides the plugin's own values
+            result = MapUtils.deepMerge((Map<String, Object>) plugin, pluginDefault.getValues());
```
fhussonnois: refs are applied after type-based defaults, so a `forced: true` ref will silently bypasses an enforced forced type (at namespace or global config) default here.

core/src/main/java/io/kestra/core/services/PluginDefaultService.java
```diff
@@ -571,4 +643,90 @@ Object recursiveDefaults(Object object, Map<String, List<PluginDefault>> default
...
+
+        PluginDefault pluginDefault = refDefaults.get(refId);
+        if (pluginDefault == null) {
+            // unknown ref: keep 'pluginDefaultsRef' so a surviving key signals an unresolved reference
+            // (flagged as a validation error in the editor and failed at runtime).
+            log.warn("No pluginDefaults bundle found for ref '{}' referenced by plugin '{}'", refId, plugin.get("type"));
+            return plugin;
+        }
+
+        Map<String, Object> result;
```
fhussonnois: You are applying the default ref without checking that the plugin `type` matches.

core/src/main/java/io/kestra/core/services/FlowService.java
```diff
@@ -370,6 +370,14 @@ public List<ValidateConstraintViolation> validate(final String tenantId, final L
                 constraintsBuilder.namespace(flow.getNamespace());
 
                 modelValidator.validate(flowWithDefaults);
+
+                java.util.Set<String> unresolvedRefs = pluginDefaultService.unresolvedPluginDefaultsRefs(flowWithDefaults);
```
fhussonnois: The validate is only done from the UI perspective (i.e., the `validate` API), but is not called when creating/update the flow. So the best is probably to have a dedicate method to validate that all refs exist in to call it also in the `create`/`update` method of the `FlowService`

fhussonnois (changes requested):

core/src/main/java/io/kestra/core/exceptions/PluginDefaultsRefNotFoundException.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Thrown when a plugin (task, trigger or task runner) references a named plugin-defaults bundle via
+ * {@code pluginDefaultsRef} that does not exist at flow, namespace or global level.
+ */
+public class PluginDefaultsRefNotFoundException extends KestraRuntimeException {
+    @Serial
+    private static final long serialVersionUID = 1L;
+
+    public PluginDefaultsRefNotFoundException(String ref) {
+        super("No pluginDefaults bundle found for pluginDefaultsRef '" + ref + "'.");
```
fhussonnois: if the exception message is user-facing, the best is just to use `No plugin-defaults exists for pluginDefaultsRef: '%s'` to avoid introducing a concept of `bundle`

fhussonnois (changes requested):

core/src/main/java/io/kestra/core/services/PluginDefaultService.java
brian-mulier-p: You're right, my bad I lost myself upon the many scenario that can occur with that forced flag and refs

core/src/main/java/io/kestra/core/services/PluginDefaultService.java
brian-mulier-p: Nice catch :+1: Adding it

core/src/main/java/io/kestra/core/services/FlowService.java
brian-mulier-p: I'll add that, I thought of runtime + actual validate, but maybe at some points we'd better call the full validate method also in create & update as it doesn't make much sense to duplicate per-validation those :thinking:

worker/src/main/java/io/kestra/worker/processors/internals/WorkerTaskCallable.java
```diff
@@ -58,6 +59,12 @@ protected void kill(final boolean markAsKilled) {
 
     @Override
     public State.Type doCall() throws Exception {
+        // a surviving 'pluginDefaultsRef' means the referenced plugin-defaults could not be resolved
+        String unresolvedRef = workerTask.getTask().getPluginDefaultsRef();
+        if (unresolvedRef != null) {
```
fhussonnois: PluginDefaults are resolved before the job being pushed to controller. Sending a bad worker-job to fail it right after  is not good idea. The job/task should fail during the plugin-default injection.

fhussonnois (changes requested):

worker/src/main/java/io/kestra/worker/processors/internals/WorkerTriggerCallable.java
```diff
@@ -31,6 +32,12 @@ public WorkerTriggerCallable(RunContext runContext, ConditionContext conditionCo
 
     @Override
     public State.Type doCall() throws Exception {
+        // a surviving 'pluginDefaultsRef' means the referenced plugin-defaults could not be resolved
+        String unresolvedRef = workerTrigger.getTrigger().getPluginDefaultsRef();
```
fhussonnois: same comment than WorkerTaskCallable

worker/src/main/java/io/kestra/worker/processors/internals/WorkerTriggerRealtimeCallable.java
```diff
@@ -42,6 +43,12 @@ public WorkerTriggerRealtimeCallable(
 
     @Override
     public State.Type doCall() throws Exception {
+        // a surviving 'pluginDefaultsRef' means the referenced plugin-defaults could not be resolved
+        String unresolvedRef = workerTrigger.getTrigger().getPluginDefaultsRef();
```
fhussonnois: same comment than WorkerTaskCallable

fhussonnois (changes requested):

anna-geller:
This goes into backburner as no time to finish and backport to 1.3

once we manage, this feature will be usable only on 1.3 so due to this limited lifetime of this feature, it's deprioritized for now

brian-mulier-p:
Superseded — this feature now targets `releases/v1.3.x` only (product decision), delivered via the backport PR #16995. Closing this develop-targeted PR.

Issue #16376, opened 2026-05-29
Update all gRCP related libs


loicmathieu:
### Describe the issue

Our code didn't seems compatible with updating gRCP libs as there respective dependabot PRs all fail.

Migration should be done manually
https://github.com/kestra-io/kestra-ee/pull/7900
https://github.com/kestra-io/kestra-ee/pull/7658
https://github.com/kestra-io/kestra-ee/pull/7184
https://github.com/kestra-io/kestra-ee/pull/7183
https://github.com/kestra-io/kestra/pull/15846

### Environment

- Kestra Version: develop

PR #16448, opened 2026-06-01
build(deps): upgrade gRPC to 1.81.0 and protobuf to 4.33.2

fhussonnois:
protoc 4.x emits @com.google.protobuf.Generated on generated stubs, a class absent from the BOM-pinned protobuf-java 3.25.9, so the protoc bump failed to compile. Move the platform BOM to protobuf-java 4.33.2 (the version Google Cloud libraries-bom 26.83.0 already ships, paired with grpc 1.81.0) and bump the worker-controller protoc/grpc codegen to match. The 3.25.9 pin was a stale lowest-common-denominator that force-downgraded mysql-connector (4.31.1) and the Google Cloud libs (4.33.2); ORC is irrelevant as it shades protobuf. opentelemetry-proto moves to 1.5.0-alpha (the protobuf-4 line).

Companion to the kestra-ee worker-controller-ee bump.

Refs #16376

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,9 +26,6 @@ updates:
     open-pull-requests-limit: 50
     labels: ["dependency-upgrade", "area/backend"]
     ignore:
-      # Ignore versions of Protobuf >= 4.0.0 because Orc still uses version 3
-      - dependency-name: "com.google.protobuf:*"
-        versions: ["[4,)"]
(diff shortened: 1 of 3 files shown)


loicmathieu (approved):
LGTM, we have an ignore rule in `dependabot.yml` for protobuf can you remove it before merging?

PR #16449, opened 2026-06-01
fix(design-system): fix filter chip OR-group sync, drag grouping, and group operator

ammeek:
# Description 

Fixes three issues in the shared filter chip UI:

- Adding a second OR-group condition that duplicates an existing value (e.g. (taskId=fail AND level>=ERROR) OR (taskId=fail)) was silently dropped from the URL until the chip was edited.
- Dragging a chip onto a group that already had a chip with the same key discarded one of them instead of grouping them with AND.
- A group's logical operator looked editable per-junction, so changing one "AND" appeared to unexpectedly change the others.

### DIFF
--- a/ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue
+++ b/ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue
@@ -18,7 +18,8 @@
                 :linked="hoveredWrapperId === unit.id"
                 inner
                 @change="(op) => filter.setWrapperLogical(unit.id, op)"
-                @hover="(v) => hoveredWrapperId = v ? unit.id : null"
+                @mouseenter="hoveredWrapperId = unit.id"
+                @mouseleave="hoveredWrapperId = null"
             />
             <div
                 class="filter-group-dropzone"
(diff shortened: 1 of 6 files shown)


ammeek:
Jam recording of this feature: https://jam.dev/c/7eba2b16-4dcb-44b4-a782-c79d73eab97c

ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue
```diff
@@ -15,8 +15,10 @@
                 v-if="childIndex > 0"
                 :logical="unit.logical"
                 :disabled="filter.readOnly?.value"
+                :linked="hoveredWrapperId === unit.id"
                 inner
                 @change="(op) => filter.setWrapperLogical(unit.id, op)"
+                @hover="(v) => hoveredWrapperId = v ? unit.id : null"
```
elevatebart: I would have avoided inventing a new event name here and used mouseenter/mouseleave to avoid confusion. But it can stay this way if need be.

elevatebart (approved):
One smol comment. 

I relly like that you added tests. Thank you 🙏

ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue
ammeek: Good call, switched to forwarding native mouseenter/mouseleave instead of the custom hover event.

PR #16454, opened 2026-06-01
fix(worker): render default Worker Queue as 'default' in logs

fhussonnois:
The default Worker Queue is stored internally as an empty string, which surfaced as '' in dispatcher logs. Format it as 'default' via WorkerQueues.forLog, and route the two direct call sites that logged the raw id through forLog as well.


ammeek (approved):
LGTM

PR #16476, opened 2026-06-02
feat(executions): keep the time filter as a global scope, outside conditional groups

### LINKED ISSUE #16433: [Filters] Quick interval selector doesn't fit conditional groups
### Describe the issue

The Executions' view quick time interval selector (15m/1h/...) does not fit to the conditional groups functionality.

## Verification scenario

1. Add a new conditional group to the default filters
2. Keep the `OR` operator
3. Add a new filter - **two distinct filter groups joined by `OR` get created**
4. Use the Interval quick filter

Which filter group should receive the changed time interval filter? The first one, the second one, or both?

This combination of filters seems to be confusing.

### Environment

- Kestra Version: develop

flcarre:
## Summary

The execution-list time range was a regular chip inside the `AND`/`OR` group builder, so it could end up in an `OR` branch and the quick-interval selector was ambiguous (*"which group should receive the interval?"* — issue #16433). A time range is always an **AND** scope, never an `OR` member, so it does not belong inside the group builder.

This makes the time filter a **global scope** rendered outside the conditional groups, and introduces a reusable `groupable` capability in the design-system filter to express that.

**Design-system** — new optional `groupable` flag on `FilterKeyConfig`. When `false`, a filter:
- is excluded from the *Add filters* menu and cannot be dragged into a group
- renders in a dedicated slot in `MainFilter`, separated from the groups by a divider, with a persistent `+ <label>` placeholder when unset that reopens the full editor (predefined + custom range + applyTo)
- always encodes at the top level (`filters[<key>][...]`), never under an `or][n]` / `and][n]` prefix, so the backend `AND`s it with the group expression — `timeRange AND (g1 OR g2)` — matching the AND/OR query support added in #16170

**Executions** — set `groupable: false` on the `timeRange` key of the execution and flow-execution filters, and move the quick-interval presets above the filter builder so they stay visible.

Also extracts a shared `buildNewFilter(key)` chip factory, removing duplicated chip construction across the add-menu, drag-drop and global-slot flows.

## Test Coverage
- `helpers.spec.ts`: timeRange stays global (`filters[timeRange][EQUALS]`) across multiple `OR` groups; custom range (`startDate`/`endDate`) stays global in multi-group; global timeRange decodes alongside grouped filters without a `groupIndex`.
- UI filter unit tests: 55 passing. Design-system filter logic tests passing.

## Verification
Verified live on the executions list: quick interval keeps the time global with conditional groups active; the time chip renders outside the `OR`/`AND` groups and is no longer addable/draggable into one; deleting it shows the `+ Interval` placeholder that reopens the editor (predefined + custom + applyTo).

## Screenshots

**Time filter as a global slot, presets on top**
1-layout

**Always outside the AND/OR groups**
2-time-outside-or-groups

**Full editor preserved (predefined + custom range + apply to)**
3-time-editor

**Empty state — `+ Interval` placeholder reopens the editor**
4-empty-state

## Notes
- Pure frontend change; no backend changes (the AND/OR support it relies on landed in #16170).
- The DS component test suites fail locally on an unrelated monaco-editor + jsdom interaction (`document.queryCommandSupported`) present on `develop`, not touched by this diff.

Closes #16433

### DIFF
--- a/ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue
+++ b/ui/packages/design-system/src/components/Data/KsDataTable/filter/FilterGroupRenderer.vue
@@ -25,7 +25,7 @@
                 class="filter-group-dropzone"
                 :class="{
                     'drag-over': dragOverGroupId === childLeaf.id,
-                    'empty': childLeaf.filters.length === 0,
+                    'empty': renderableFilters(childLeaf.filters).length === 0,
                 }"
                 @dragover.stop="(e) => $emit('drag-over', e, childLeaf.id)"
                 @dragleave.stop="(e) => $emit('drag-leave', e, childLeaf.id)"
@@ -41,7 +41,7 @@
                     @dragend="$emit('drag-end')"
                 ><Drag /></span>
                 <div
-                    v-for="appliedFilter in childLeaf.filters"
+                    v-for="appliedFilter in renderableFilters(childLeaf.filters)"
                     :key="appliedFilter.id"
                     class="filter-chip-wrap"
                     :class="{'dragging': isDraggingFilter(appliedFilter.id)}"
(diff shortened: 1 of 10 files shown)


RemiBarthe (approved):

flcarre:
Folded into the consolidated executions filter redesign on `claude/loving-shtern-acfe14`.

The global time-scope work from this PR was cherry-picked and reconciled with the new rules-pill + advanced-panel UI there:
- `timeRange` is `groupable: false`, rendered as a standalone global chip, always encoded at the top level, and excluded from groups, the advanced panel, and the "Add filters" picker.
- The `buildNewFilter` / `groupable` plumbing and the `helpers.spec.ts` tests came over as-is (all green).

Verified end-to-end on the running app: the global chip works, the "N rules" pill counts only conditional filters, and the quick state filter hides once filters become complex.

Closing in favor of that branch — everything from here ships there.

PR #16483, opened 2026-06-02
fix(ui): keep AI Copilot open when selecting a provider from the dropdown

brian-mulier-p:
- The `onClickOutside` handler in `AiCopilotWrapper.vue` still ignored the stale `.ai-provider-pill-popper` class from the old pill `el-dropdown`; the design-system refactor swapped it for `KsSelect`, whose dropdown teleports to `body` with a different popper class — so clicking a provider option counted as an outside click and closed the copilot panel, making it impossible to pick a second model.
- Introduce a shared `AI_PROVIDER_POPPER_CLASS` constant, set as `popperClass` on both provider `KsSelect`s and used by the click-outside ignore list, so producer and consumer can no longer drift apart.
- Add a regression test: clicking a provider option keeps the copilot open (verified red against the stale selector), while a true outside click still closes it.

Fixes kestra-io/kestra-ee#8172


MilosPaunovic (approved):
The code looks pretty straight-forward, so I'll approve, just make sure to handle the merge conflict before the merge itself.

PR #16517, opened 2026-06-03
feat: surface logs for all execution kinds and add a kind filter

ammeek:
### ✨ Description

Logs produced by non-`NORMAL` execution kinds (`PLAYGROUND`, `TEST`, and `LOOP`) were effectively hidden from much of the UI.

The most visible symptom was that Playground execution logs appeared in the standard execution log view but not in the temporal/raw log view, and never appeared on the global Logs page.

The root cause was that log queries were hard-filtered to:

```text
execution_kind = NORMAL
```

causing logs from all other execution kinds to be silently excluded.

### Fix

Make logs from all execution kinds visible by default and provide filtering controls for users who want to narrow results by execution kind.

#### Log Query Behavior

* Log search endpoints now return all execution kinds by default.
* Per-execution log endpoints (`list`, `download`, and `follow`) now return all execution kinds by default.
* Dashboard aggregations continue to filter to `NORMAL` executions to avoid introducing noise into operational charts.

#### Execution Kind Filter

Added a new **Kind** filter for the `LOG` resource with support for:

* `EQUALS`
* `NOT_EQUALS`
* `IN`
* `NOT_IN`

The filter is implemented across all log backends:

* JDBC
* Elasticsearch (EE)
* Live SSE follow matching

#### Logs Page UI

Added a **Kind** filter to the Logs page.

To preserve the existing production-focused experience, the filter defaults to:

```text
NORMAL
```

Users can explicitly include:

* `PLAYGROUND`
* `TEST`
* `LOOP`

when investigating those execution types.

### Additional Log Viewer Fixes

While validating the execution-kind changes, several related issues in the log viewer were addressed.

#### Temporal/Raw View Live Updates

The temporal/raw log view now streams live updates.

Previously it performed a single fetch, making active executions appear stuck until the page was refreshed.

#### Filter Propagation to SSE Follow

Active filters are now forwarded to the log follow (SSE) endpoint, including:

* Level filters
* Task filters

This keeps the live stream consistent with the static log view.

#### Unsupported Parameters Removed

Stopped sending unsupported `startDate` and `endDate` query parameters to:

```text
/logs/search
```

The effective time window is already derived server-side from the `timeRange` filter.

#### Playground Log Filters

Removed the irrelevant **Task Run** and **Attempt** filters from the Playground logs tab.

#### Default Filter Handling

Default log filters are now applied only when arriving on a clean route.

If filters are already present in the URL, they are respected as-is rather than having defaults layered on top of them.

🔗 Related Issue

Closes https://github.com/kestra-io/kestra-ee/issues/7348

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLogRepository.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLogRepository.java
@@ -74,8 +74,11 @@ public abstract class AbstractJdbcLogRepository extends AbstractJdbcCrudReposito
 
     @Override
     public ArrayListTotal<LogEntry> find(Pageable pageable, @Nullable String tenantId, @Nullable List<QueryFilter> filters) {
-        // All execution kinds are exposed by default; callers narrow with an explicit KIND filter.
+        // Default to NORMAL kind only; an explicit KIND filter overrides that and selects the requested kind(s).
         var condition = this.filter(filters, DATE_COLUMN, Resource.LOG);
+        if (!QueryFilter.hasField(filters, QueryFilter.Field.KIND)) {
+            condition = NORMAL_KIND_CONDITION.and(condition);
+        }
         return findPage(pageable, tenantId, condition);
     }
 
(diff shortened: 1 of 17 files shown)


jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLogRepository.java
loicmathieu, line 1: We want by default to only expose NORMAL kind logs, this was done on purpose.

What should be done instead, is to only filter on NORMAL kind if not filter exist for the kind.
No filter => only normal
Kind filter => only the specified kind

loicmathieu (changes requested):

loicmathieu (approved):
Backend part LGTM, there are too much frontend changes for me, @elevatebart can you review the frontend?

elevatebart (approved):
Looks good!

- why not use the new sdk for SSE. I think it could have avoided headaches.
- if you are to ship your own, I might have liked a composable to ship it all as one function.

PR #16525, opened 2026-06-03
fix(build): restore linux/arm64 for base images, eliminate double plugin download

AcevedoR:
## Problem

The previous quick-fix (`66fdff877`) dropped `linux/arm64` from both base images to stop the bleeding. That unblocked the nightly CI but left arm64 users without a compatible base image — and would break the arm64 leg of every release Docker build.

Root cause: QEMU arm64 emulation was silently timing out for the `eclipse-temurin` JRE layer. Buildx pushed only the amd64 manifest entry while reporting the job as `[ok]`.

## Solution

**1. Native runners per platform**
Replace the single QEMU-based multi-arch build with a `matrix` of native runners:
- `ubuntu-latest` → `linux/amd64`
- `ubuntu-24.04-arm` → `linux/arm64`

Each job builds and pushes a temporary arch-suffixed image (`latest-no-plugins-amd64`, `latest-no-plugins-arm64`). A `merge-*` job then calls `docker buildx imagetools create` to assemble the final multi-arch manifest — no QEMU, no silent failures.

**2. Plugin download happens once, outside Docker**
`Dockerfile.base.with-plugins` is simplified to a single `COPY plugins/ /app/plugins/`. The CI pipeline:
1. Downloads all OSS plugins via `kestractl` on one amd64 runner
2. Uploads them as a GHA artifact
3. Each platform build job downloads the same artifact and COPYs it in

Java JARs are platform-independent, so the same files work for both amd64 and arm64. Maven Central is only hit once per nightly run — no duplication, no 429 rate-limiting.

## Pipeline shape

```
build-no-plugins (amd64)  ──┐
build-no-plugins (arm64)  ──┤─ merge-no-plugins ──┐
download-plugins ───────────┘  (parallel)          ├─ build-with-plugins (amd64) ──┐
                                                    └─ build-with-plugins (arm64) ──┴─ merge-with-plugins
```

## Dry-run behaviour

`dry-run: true` builds both platform images without pushing (validates `Dockerfile.base` compiles for both platforms). All downstream jobs are skipped since nothing was pushed.

## Test plan

- [ ] Manually trigger the workflow with `dry-run: false` — verify all 5 jobs pass
- [ ] `docker manifest inspect ghcr.io/kestra-io/kestra-base:latest-no-plugins` shows both `linux/amd64` and `linux/arm64`
- [ ] `docker manifest inspect ghcr.io/kestra-io/kestra-base:latest` shows both platforms
- [ ] Main build on develop goes green after a fresh base image push

### DIFF
--- a/.github/workflows/global-push-base-image.yml
+++ b/.github/workflows/global-push-base-image.yml
@@ -2,14 +2,16 @@ name: Build and Push Base Image
 
 # This workflow maintains two base images:
 #
-# 1. kestra-base:latest-no-plugins — a lightweight image with just the JRE and uv pre-installed.
-#    Used as the foundation for the image below.
+# 1. kestra-base:latest-no-plugins — lightweight image with just the JRE and uv pre-installed.
+#    Built for linux/amd64 and linux/arm64 from Dockerfile.base.
 #
-# 2. kestra-base:latest — built on top of latest-no-plugins, with all OSS plugins pre-downloaded.
-#    This is a plugin cache: downstream Kestra image builds pull from it instead of fetching
-#    190+ plugins from Maven Central on every build. A nightly refresh keeps the cache current.
-#    Partial downloads (e.g. due to Maven Central rate-limiting) are acceptable — even a partial
-#    cache is better than none, and missing plugins are picked up on the next nightly run.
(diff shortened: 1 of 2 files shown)


AcevedoR:
Superseded — redoing with sequential per-arch approach

PR #16530, opened 2026-06-03
fix(system): only send an auth event when usage tracking is enabled

loicmathieu:
Fixes https://github.com/kestra-io/kestra-ee/issues/6610


AcevedoR (approved):

loicmathieu:
Backported to 1.0 and 1.3

PR #16542, opened 2026-06-04
Add env Pebble function

### LINKED ISSUE #16396: envs() function notation
### Feature description

Hi guys ! As with the secret retrieval function, it would be convenient to have the same call format for environment variables (' envs("xxxx") '). This would allow calling an environment variable based on a dynamic name (for example, according to a database return value or status) which is not currently possible.

jianongHe:
### ✨ Description

Adds an `env(name, default)` Pebble function.

It reads from the execution context `envs` map provided by RunContext, not from `System.getenv()`. If the value is missing or empty, it returns the optional default value.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/16396

### 🛠️ Backend Checklist

- [x] Code compiles successfully and passes focused checks
- [x] Focused unit/doc tests pass

### 📝 Additional Notes

Before: `env(...)` was not registered as a Pebble function, so users had to access `envs` directly.

After: `{{ env('DATABASE_HOST', 'localhost') }}` returns the RunContext env value, or `localhost` when that value is missing or empty.

Tested with:

```bash
./gradlew :core:test --tests io.kestra.core.runners.pebble.functions.EnvFunctionTest
./gradlew :core:test --tests io.kestra.core.runners.pebble.PebbleExpressionServiceTest --tests io.kestra.core.runners.pebble.PebbleDocBasicMdDriftTest
```

Results: 6 EnvFunction tests passing, plus 9 Pebble docs/autocomplete tests passing.

### 🤖 AI Authors

AI assistance was used for the implementation and test run notes. Template note: this change is purr-posefully small.


core/src/main/java/io/kestra/core/runners/pebble/functions/EnvFunction.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Map;
+
+import io.kestra.core.runners.RunVariables;
+
+import io.pebbletemplates.pebble.error.PebbleException;
+import io.pebbletemplates.pebble.template.EvaluationContext;
+import io.pebbletemplates.pebble.template.PebbleTemplate;
+import jakarta.inject.Singleton;
+
+@Singleton
```
loicmathieu: As there is nothing to inject inside this function, it can be a regular class and not a Micronaut bean.
So when adding it inside the extension, you can directly instantiate it.

loicmathieu (changes requested):
Hi, thanks a lot for your contribution.

Only one small changes and I'll merge it.

core/src/main/java/io/kestra/core/runners/pebble/functions/EnvFunction.java
jianongHe: Done, removed `@Singleton` and now instantiate it directly in `Extension.java`. Tests pass.

loicmathieu (approved):
LGTM, thanks

PR #16544, opened 2026-06-04
fix(core): apply forced plugin defaults in admin-first order

brian-mulier-p:
- Fixes forced plugin-default precedence ([kestra-ee#8262](https://github.com/kestra-io/kestra-ee/issues/8262)): forced defaults were applied closest-first (`Lists.reverse`), so a namespace-forced default beat a global-forced one and a child namespace beat its parent. They now apply **admin-first** — global beats namespace, parent namespace beats child — the intended enforcement direction (reverse of non-forced, where flow > namespace > global).
- Root cause: the forced map was built from `Lists.reverse(...)` while the defaults list flipped to flow-first in v0.19.0; combined, the closest forced level won. Removing the reversal lets list-order-last (admin-most) win.
- Non-forced precedence is unchanged; added comments explaining why the two merge directions are intentionally asymmetric.
- Adds a `globalForcedDefaultBeatsNamespaceForcedDefault` test.
- Adds a JMH benchmark characterizing type-matched plugin-default injection (`noDefaults` + `typeDefaults`) and pins the `jmh-benchmarks` toolchain to JDK 25.
- Independent of the `pluginDefaultsRef` feature ([#8740](https://github.com/kestra-io/kestra/issues/8740) / [#16446](https://github.com/kestra-io/kestra/pull/16446)); backportable to 1.0.x/1.1.x (impacted since v0.19.0).

### DIFF
--- a/core/src/main/java/io/kestra/core/services/PluginDefaultService.java
+++ b/core/src/main/java/io/kestra/core/services/PluginDefaultService.java
@@ -20,7 +20,6 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.Lists;
 
 import io.kestra.core.exceptions.FlowProcessingException;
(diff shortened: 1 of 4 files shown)


brian-mulier-p:
To backport to 1.3

fhussonnois (approved):
LGTM!

PR #16558, opened 2026-06-04
fix(executions): confirm before discarding a filled execution input form

flcarre:
## What

Closing the flow-execution input dialog (outside-click, X button, or `Escape`) used to silently discard everything the user had typed. For forms with many inputs, one accidental click meant re-filling the whole form from scratch.

Now:
- **Empty / pristine form** → closes immediately (nothing to lose).
- **Form with data** → a confirmation first: *"You have unsaved inputs. Are you sure you want to discard this execution?"*

This is the behavior agreed on the issue (keep all four close paths, only guard when there's something to lose, no "don't show again" toggle).

## How

- `FlowRun.vue` exposes an `isDirty` computed: true when the user has non-default inputs, a filled execution label, or a schedule date.
- `TriggerFlow.vue` gates **both** execute dialogs (the per-flow dialog and the select-flow dialog) through a shared `confirmDiscardIfDirty(isDirty, done)` helper.
- A reentrancy guard (`isConfirmingClose`, cleared in `.finally` over confirm *and* cancel) prevents rapid repeated close attempts from stacking confirmation prompts.
- New i18n key `discard execution confirmation` (en only; the auto-translate action backfills other locales).

## Testing

- Unit: `ui/tests/unit/components/flows/triggerFlowDiscardConfirmation.spec.ts` (8 tests — `isDirty` predicate + gating contract).
- `vue-tsc` type-check clean, oxlint clean.
- Manual (local dev): verified pristine close (no prompt), dirty close → confirm, Cancel keeps the form + data, OK discards and closes, the select-flow dialog now prompts too, rapid double-Escape yields a single confirm, and a successful submit closes without a spurious prompt.

Closes https://github.com/kestra-io/kestra-ee/issues/7818

> Per the issue, this is treated as a UX bug to **backport to LTS 1.0 and 1.3**.

### DIFF
--- a/ui/src/components/flows/FlowRun.vue
+++ b/ui/src/components/flows/FlowRun.vue
@@ -198,6 +198,14 @@
         flow.value && !flow.value.disabled && !haveBadLabels.value,
     )
 
+    const isDirty = computed(() =>
+        Object.keys(inputsNoDefaults.value).length > 0 ||
+        executionLabels.value.some(label => label.key || label.value) ||
+        scheduleDate.value !== undefined,
+    )
+
+    defineExpose({isDirty})
+
     const hasWebhookTriggers = computed(() => {
         if (!flow.value?.triggers) {
             return false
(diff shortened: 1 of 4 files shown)


flcarre:
Superseded by #16573, which contains this exact fix as its first commit plus the generalization into a global discard-guard rule. Keeping a single develop PR for the discard-confirmation work. The minimal LTS backports (#16561, #16563) remain as-is.

PR #16565, opened 2026-06-04
fix(ui): restore outdated revision warning and save confirmation

### LINKED ISSUE #16560: concurrent flow editing seems broken
### Describe the issue

When N > 1 users are concurrently editing the same flow based on the most recent revision there is no mechanism in place that would inform users if the flow has changed since they started editing. It seems what silently happens in the background is last write win semantics which may not be what users expect. Also, it's unclear what the intended behaviour of Kestra is in such cases.

https://github.com/user-attachments/assets/362db9fa-b303-417b-bf52-13960371bbf1

### Environment

- Kestra Version: 1.3.20 OSS

brian-mulier-p:
closes #16560


elevatebart (approved):
I like the testing part specialy

PR #16585, opened 2026-06-05
fix: show meaningful empty state in Gantt view

ammeek:
✨ Description

Replaces the misleading/blank empty states in the execution Gantt view with a single, consistent design, retires the old ExecutionPending placeholder, and fixes a related 400 it surfaced in the concurrency view.

The problem. When the Gantt had no task runs to plot, the card rendered completely blank. For executions cancelled or failed by a concurrency limit (behavior: CANCEL / FAIL) — terminal state, zero task runs — it then showed "No tasks are running yet", which is wrong since nothing will ever run. Separately, CREATED/QUEUED executions showed a different, older placeholder (queued_visual.svg + its own layout).

The new empty state. When there are no task runs, the Gantt now hides the filter bar (mirroring the versioned-plugins empty screen) and renders a KsEmptyState — a boxed illustration plus the execution's status as a KsExecutionStatus badge ("Execution is: <state>") — with a state-aware hint:

- Created / not yet started → "No tasks are running yet." + "Once the execution starts, you'll see progress updates here."
- Terminal, no task runs (the concurrency CANCEL/FAIL case) → "This execution ended without running any task. Check its overview and logs for the reason."
- Task runs exist but filters/search hid them all → "No tasks match the selected filters…" (filter bar stays; message shown in the card).

Routing / old UI retired. Previously both CREATED and QUEUED rendered ExecutionPending (the old EmptyTemplate + queued_visual.svg + copy). Now only QUEUED routes to ExecutionPending; CREATED and terminal states use the unified KsEmptyState. ExecutionPending itself was reduced to either FlowConcurrency (once the flow is loaded) or a loading spinner — so the old queued_visual.svg placeholder no longer renders in any state (including the brief flow-loading transient). queued_visual.svg is now unreferenced.

QUEUE behavior unchanged. A queued execution still shows the FlowConcurrency overview (running-slots card + the flow's executions), since it's waiting on a slot and that context is more useful than a generic placeholder.

The 400. Verifying the QUEUED case exposed a bug: the concurrency view embeds the executions list, which built its /executions/search request from the shared route query and forwarded the Gantt's filters[level] (a log-level filter) → Field LEVEL is not supported for resource EXECUTION. The list now narrows query filters to the fields its filter configuration supports (config-derived, so it stays in sync; timeRange's startDate/endDate encoding and nested and/or groups handled), applied at every endpoint boundary — search, charts, and the CSV export, which previously bypassed sanitisation entirely.

Closes https://github.com/kestra-io/kestra-ee/issues/7346


Piyush-r-bhaskar:
We have a design for this . can you look and help implement for this usecase https://www.figma.com/design/AOqwOR5l0YFIPAd59NbWhb/_2.0?node-id=27302-7641&t=VqqkLuZpCvSmV1IL-0

Piyush-r-bhaskar (approved):

PR #16589, opened 2026-06-05
fix(core): exporting in flows, executions and triggers

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra-ee/issues/8339


MilosPaunovic (approved):
Good catch! 🚀

MilosPaunovic:
Can you check whether this is related to it also https://github.com/kestra-io/kestra-ee/issues/8073 @Piyush-r-bhaskar?

Piyush-r-bhaskar:
Nope it's different with 403, but seems can be fixed similar

PR #16605, opened 2026-06-08
feat(design-system): redesign login page [2.0]

flcarre:
## Summary

- Replace full Kestra wordmark with the black-tile Kestra monogram (`monogram.svg`, exported from the 2.0 Figma)
- Remove "Enter your credentials…" subtitle
- Rewrite with bespoke `login-field` / `login-btn` components matching Figma spec: 30px inputs with inline outline icons, eye toggle (filled `Eye`/`EyeOff`) on password field, 32px primary button, DS font-size tokens
- Layout per Figma: 32px gaps between monogram, title and form; "Trouble signing in?" sits 16px below the Login button inside the form column; placeholders use `--ks-text-primary`
- Inline validation errors below each field (red border + error text)
- Update `setup.login_title` to "Sign in to Kestra"; drop `setup.login_subtitle` key from all 12 locale files

Implements Figma node [28889-37123 (Sign in OSS)](https://www.figma.com/design/AOqwOR5l0YFIPAd59NbWhb/_2.0?node-id=28889-37123&m=dev).
Companion OSS change for kestra-io/kestra-ee#8348.

https://github.com/user-attachments/assets/7c135d77-9e78-4724-b532-9d63a18eaff9

## Test plan

- [x] Visit `/login` on OSS basic-auth setup — verify monogram, "Sign in to Kestra" title, no subtitle, compact inputs with inline icons, eye toggle, purple Login button (verified in light and dark themes against Figma)
- [x] Submit with invalid credentials → inline error appears below each field (red border + message)
- [ ] Submit with valid credentials → redirects correctly
- [ ] "Trouble signing in?" opens troubleshooting docs in new tab
- [x] `cd ui/src/translations && node check.js` → clean for all locales

- Add the onboarding grid background behind the login so login and onboarding stay visually consistent


ui/src/components/basicauth/BasicAuthLogin.vue
```diff
@@ -1,68 +1,62 @@
...
-            <KsText tag="h1" class="basic-auth-login__title">{{ $t("setup.login_title") }}</KsText>
-            <KsText tag="p" class="basic-auth-login__subtitle">{{ $t("setup.login_subtitle") }}</KsText>
-        </div>
-
-        <KsForm @submit.prevent :model="credentials" ref="form" :rules="rules">
+    <div class="oss-login">
+        <img src="../../assets/monogram.svg" alt="Kestra" class="oss-login__icon" />
+        <h1 class="oss-login__title">{{ $t("setup.login_title") }}</h1>
+
+        <form class="oss-login__form" @submit.prevent="handleSubmit">
```
flcarre: Native `required` + `type="email"` make the browser pop its own validation bubble on submit, which pre-empts your localized inline errors. Adding `novalidate` keeps the JS validation as the single source of truth.

```suggestion
        <form class="oss-login__form" @submit.prevent="handleSubmit" novalidate>
```

ui/src/components/basicauth/BasicAuthLogin.vue
```diff
@@ -1,68 +1,62 @@
...
+                        v-model="credentials.username"
+                        :placeholder="$t('email')"
+                        type="email"
+                        required
+                        autocomplete="username"
+                        @blur="validateEmailField"
+                        @input="errors.username = undefined"
+                    />
+                </div>
+                <p v-if="errors.username" class="login-field-error">{{ errors.username }}</p>
```
flcarre: **Design-system bypass (`ui/AGENTS.md` #1/#8).** This hand-rolled field replaces `KsInput`, and the file adds ~150 lines of bespoke SCSS for what `KsInput`/`KsButton`/`KsForm` already provide. The compact inline-icon input is a real DS gap, so the compliant fix is to extend `KsInput` (add an icon slot, reuse its existing `showPassword` toggle) or introduce a `KsLoginField` component with a Storybook story and a unit test, then build the form from DS components. A local fork drifts when `KsInput` evolves.

**Accessibility:** the `<input>` has only an `id` and a placeholder, no `<label>` and no `aria-label` (placeholder is not a label, WCAG 1.3.1), and the error `<p>` below is not linked to the input via `aria-describedby` nor is `aria-invalid` set on the error state. Screen-reader users will not get the field name or the error.

flcarre (commented):
## Independent code review

**Verdict: REQUEST CHANGES** (1 blocker, 6 major, 4 minor).
Posted as a `COMMENT` review because GitHub does not allow request-changes on your own PR; treat the verdict above as the intent.

Reviewed against `ui/AGENTS.md` (design-system golden rules). Every finding below was confirmed against the PR head, not inferred from the diff.

### Blocker

**i18n: `setup.validation.password_required` exists only in `en.json`.** It is missing from all 12 other locales (`de, es, fr, hi, it, ja, ko, pl, pt, pt_BR, ru, zh_CN`), so `ui/src/translations/check.js` reports 12 missing keys and the translation CI check fails. The "check.js clean" item in the description does not hold. Add the key to every locale in the same PR. Suggested values:

| locale | value |
|---|---|
| de | `Passwort ist erforderlich` |
| es | `La contraseña es obligatoria` |
| fr | `Le mot de passe est requis` |
| hi | `पासवर्ड आवश्यक है` |
| it | `La password è obbligatoria` |
| ja | `パスワードは必須です` |
| ko | `비밀번호는 필수입니다` |
| pl | `Hasło jest wymagane` |
| pt | `A palavra-passe é obrigatória` |
| pt_BR | `A senha é obrigatória` |
| ru | `Пароль обязателен` |
| zh_CN | `密码为必填项` |

### Major (see inline comments)

1. **Design-system bypass.** The form hand-rolls raw `<form>/<input>/<button>` plus ~150 lines of bespoke SCSS in place of `KsForm`/`KsInput`/`KsButton`. `ui/AGENTS.md` rules #1 and #8 say compose or extend `Ks*`, do not fork. The compact inline-icon input is a genuine DS gap, so the compliant path is to extend `KsInput` (icon slot, reuse its `showPassword`) or add a `KsLoginField` with a story and a test. If 2.0 ships this page outside the DS on purpose, it should be a tracked tech-debt follow-up rather than silent.
2. **`:deep(svg)`** appears three times (icon and eye blocks). Rule #4 forbids it; using `KsIcon` with its `:size` prop removes the need entirely.
3. **Hardcoded colors** in `.login-btn` (`color: white`, spinner `rgba(255,255,255,0.3)`, `border-top-color: white`). Rule #2: `--ks-btn-primary-text` already exists.
4. **Hover via `opacity`** instead of `--ks-btn-primary-bg-hover` (not theme-aware).
5. **`border-radius: 0.5rem`** hardcoded twice; `--ks-radius-base` exists and the old code used a radius token.
6. **Accessibility:** the eye toggle is `tabindex="-1"` (keyboard users cannot reveal the password) and has no accessible name; inputs have no `<label>` or `aria-label`, and errors are not wired via `aria-invalid` / `aria-describedby`.

### Minor

- **`login_title` content drift:** `en` became "Sign in to Kestra" but the 12 other locales still render their old "Sign in" string (no product name). The key exists everywhere so `check.js` passes; just a localization-staleness nit.
- **Native validation competes with the bespoke errors:** `required` + `type="email"` makes the browser show its own validation bubble on submit, pre-empting the localized inline messages. Add `novalidate` to the `<form>` so the JS validation is the single source (inline suggestion).
- **`width: 320px`** fixed (was `max-width: 400px; width: 100%`): minor responsiveness regression on very narrow viewports.
- **`3.9286rem`** icon size is an odd magic value.

### Verified, not an issue

All other `--ks-*` tokens exist (no phantom references). `email_temporary_not_allowed`, `Logo.vue`, and `mailchecker` are still used by `BasicAuthSetup.vue`, so nothing is orphaned. Dropping password-complexity and disposable-email checks on a login form is correct (you authenticate a stored credential; re-applying creation-time policy would block valid users). `KsMessage.error("Login failed")` is a pre-existing hardcoded string. The monogram `#000000` tile is an intentional brand asset.

ui/src/components/basicauth/BasicAuthLogin.vue
```diff
@@ -1,68 +1,62 @@
...
+                        id="input-password"
+                        v-model="credentials.password"
+                        :type="showPasswordText ? 'text' : 'password'"
+                        :placeholder="$t('password')"
+                        required
+                        autocomplete="current-password"
+                        @blur="validatePasswordField"
+                        @input="errors.password = undefined"
+                    />
+                    <button class="login-field__eye" type="button" @click="showPasswordText = !showPasswordText" tabindex="-1">
```
flcarre: Icon-only toggle with `tabindex="-1"`: a keyboard-only user cannot reveal the password they are typing, and the button has no accessible name (`ui/AGENTS.md` a11y). Drop `tabindex` and add an `aria-label`. The two keys below need adding to the locale files (or reuse existing ones if present).

```suggestion
                    <button class="login-field__eye" type="button" @click="showPasswordText = !showPasswordText" :aria-label="showPasswordText ? $t('setup.hide_password') : $t('setup.show_password')">
```

ui/src/components/basicauth/BasicAuthLogin.vue
```diff
@@ -255,78 +211,167 @@
...
-
-        &__logo {
-            width: 140px;
-            margin-bottom: 1.25rem;
+    .oss-login {
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+        gap: var(--ks-spacing-6);
+        width: 320px;
```
flcarre, line 218: Two magic numbers here: `width: 320px` is a fixed layout width (the old code used `max-width: 400px; width: 100%`, which is responsive), and below it `3.9286rem` for the icon is an odd derived value. Prefer `var(--ks-spacing-*)` / a token where one fits, and a clean rem otherwise.

ui/src/components/basicauth/BasicAuthLogin.vue
```diff
@@ -255,78 +211,167 @@
...
-            margin-bottom: 1.75rem;
+    .login-field {
+        display: flex;
+        align-items: center;
+        gap: var(--ks-spacing-2);
+        height: 30px;
+        padding: var(--ks-spacing-2);
+        background: var(--ks-bg-input);
+        border: 1px solid var(--ks-border-strong);
+        border-radius: 0.5rem;
```
flcarre: Rule #6: use the radius token (the old code used one). `--ks-radius-base` resolves to `0.5rem`.

```suggestion
        border-radius: var(--ks-radius-base);
```

ui/src/components/basicauth/BasicAuthLogin.vue
```diff
@@ -255,78 +211,167 @@
...
-        &__submit {
-            margin-bottom: 0;
+        &__icon {
+            flex-shrink: 0;
+            color: var(--ks-text-secondary);
+
+            :deep(svg) {
+                width: 14px;
+                height: 14px;
+            }
```
flcarre: `:deep()` is forbidden in feature code (rule #4); the same block repeats for `.login-field__eye` below. It is only needed because the icons are raw `vue-material-design-icons` components. Switch to `<KsIcon name="account-outline" :size="14" />` (and `lock-outline` / `eye` / `eye-off`) and size via the prop, which removes all three `:deep` blocks.

ui/src/components/basicauth/BasicAuthLogin.vue
```diff
@@ -255,78 +211,167 @@
...
+            background: transparent;
+            border: none;
+            outline: none;
+            font-size: var(--ks-font-size-xs);
+            color: var(--ks-text-primary);
+            font-family: inherit;
+
+            &::placeholder {
+                color: var(--ks-text-primary);
+            }
```
flcarre: Placeholder color set to `--ks-text-primary` (same as typed text) makes an empty field look pre-filled and removes the visual distinction between placeholder and value. Even if this matches Figma, it is an a11y/UX concern worth confirming; `--ks-text-secondary` or a dimmer token reads better.

ui/src/components/basicauth/BasicAuthLogin.vue
```diff
@@ -255,78 +211,167 @@
...
+        font-size: var(--ks-font-size-xs);
+        color: var(--ks-text-error);
+    }
+
+    .login-btn {
+        width: 100%;
+        height: 32px;
+        background: var(--ks-btn-primary-bg-default);
+        border: none;
+        border-radius: 0.5rem;
```
flcarre: Same as the field above: use the radius token.

```suggestion
        border-radius: var(--ks-radius-base);
```

ui/src/components/basicauth/BasicAuthLogin.vue
```diff
@@ -255,78 +211,167 @@
...
+        color: white;
```
flcarre: Hardcoded color (rule #2). `--ks-btn-primary-text` exists and resolves to white.

```suggestion
        color: var(--ks-btn-primary-text);
```

ui/src/components/basicauth/BasicAuthLogin.vue
```diff
@@ -255,78 +211,167 @@
...
+        cursor: pointer;
+        box-shadow: 0 1px 2px var(--ks-shadow-element);
+        font-family: inherit;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+
+        &:hover:not(:disabled) {
+            opacity: 0.9;
+        }
```
flcarre: Use the hover token instead of `opacity` so the hover state stays theme-aware. `--ks-btn-primary-bg-hover` exists.

```suggestion
        &:hover:not(:disabled) {
            background: var(--ks-btn-primary-bg-hover);
        }
```

ui/src/components/basicauth/BasicAuthLogin.vue
```diff
@@ -255,78 +211,167 @@
...
+        &:disabled {
+            opacity: 0.6;
+            cursor: not-allowed;
+        }
+
+        &__spinner {
+            width: 14px;
+            height: 14px;
+            border: 2px solid rgba(255, 255, 255, 0.3);
+            border-top-color: white;
```
flcarre: Hardcoded colors on the spinner (rule #2). Derive from the button text token.

```suggestion
            border: 2px solid color-mix(in srgb, var(--ks-btn-primary-text) 30%, transparent);
            border-top-color: var(--ks-btn-primary-text);
```

MilosPaunovic (approved):

PR #16620, opened 2026-06-08
fix(core): migrate binary ION readers to streaming and fix memory regressions

### LINKED ISSUE #16554: Split output chunks cannot be read by Aggregate (or IonToJson)
## Reproducer

```yaml
id: bug_split_aggregate_incompatibility
namespace: company.qa

tasks:
  - id: download
    uri: https://huggingface.co/datasets/kestra/datasets/raw/main/csv/orders.csv

  - id: to_ion
    type: io.kestra.plugin.serdes.csv.CsvToIon
    from: "{{ outputs.download.uri }}"
    header: true

  - id: split
    type: io.kestra.plugin.core.storage.Split
    from: "{{ outputs.to_ion.uri }}"
    rows: 25

  - id: aggregate_direct
    type: io.kestra.plugin.transform.Aggregate
    from: "{{ outputs.to_ion.uri }}"
    outputType: STORE
    groupBy: [customer_email]
    aggregates:
      orders:
        expr: count()
        type: INT

  - id: aggregate_on_chunk
    type: io.kestra.plugin.transform.Aggregate
    from: "{{ outputs.split.uris[0] }}"
    outputType: STORE
    groupBy: [customer_email]
    aggregates:
      orders:
        expr: count()
        type: INT
```

## Expected

Both `aggregate_direct` and `aggregate_on_chunk` succeed.

### Actual

`aggregate_direct` succeeds. `aggregate_on_chunk` fails with:

```
bad character [239, "\xef"] encountered where a token was supposed to start at line 1 offset 2
```

`IonToJson` fails on the same `Split` output chunk with an identical error.

## Notes

- The only difference between the two tasks is whether the input comes from `CsvToIon` output or a `Split` chunk. The `Aggregate` config is identical.
- The `\xef` byte at offset 2 is consistent with a UTF-8 BOM (`EF BB BF`), suggesting `Split` may be writing chunk files with a BOM prefix that the ION text reader cannot handle.
- `IonToJson` fails on the same chunk with the same error, so the issue is in what `Split` writes, not in `Aggregate` specifically.

### Environment

- Kestra Version: develop

nkwiatkowski:
## Summary

After the binary ION switch (`66f239b3e`), several core components loaded entire ION files into `ArrayList` buffers before processing. This caused:
- **O(N²) memory usage** for Loop iterations (`FlowableUtils.readLoopValuesFromUri` loaded the full file on every call with `count=1`)
- **Unbounded memory** for Split operations on large files (`StorageService.splitIon` buffered all records)
- **Binary ION corruption** when `StorageService.split()` read binary ION through UTF-8 text streams (`bad character [239, "\xef"]`)

### Changes
- **StorageService**: stream ION records via `MappingIterator` directly from `InputStream` — no intermediate `ArrayList`
- **FlowableUtils**: stream-and-skip pattern — count/skip records in the consumer callback without storing them (O(limit) memory instead of O(N))
- **FilterItems / DeduplicateItems**: migrated from `BufferedReader`/`readLine()` to `FileSerde.read()`/`FileSerde.write()` for binary ION support
- **FileSerde**: removed premature `try-with-resources` parser close in `createMappingIterator`, added `createBinarySequenceWriter` for EE exporters
- Added binary ION input test coverage for FilterItems, DeduplicateItems, and FlowableUtils

## Test plan
- [x] `StorageServiceTest` — 12/12 pass (ION + text split paths)
- [x] `FilterItemsTest` — 7 tests including 2 new binary ION input tests
- [x] `DeduplicateItemsTest` — 4 tests including 1 new binary ION input test
- [x] `FlowableUtilsTest` — 21 tests including 2 new binary ION input tests

Closes #16554

### DIFF
--- a/core/src/main/java/io/kestra/core/serializers/FileSerde.java
+++ b/core/src/main/java/io/kestra/core/serializers/FileSerde.java
@@ -253,31 +253,31 @@ public final class FileSerde {
     }
 
     private static <T> MappingIterator<T> createMappingIterator(ObjectMapper objectMapper, Reader reader, TypeReference<T> type) throws IOException {
-        try (var parser = objectMapper.createParser(reader)) {
-            return objectMapper.readerFor(type).readValues(parser);
-        }
+        var parser = objectMapper.createParser(reader);
+        return objectMapper.readerFor(type).readValues(parser);
     }
 
     private static <T> MappingIterator<T> createMappingIterator(ObjectMapper objectMapper, Reader reader, Class<T> type) throws IOException {
-        try (var parser = objectMapper.createParser(reader)) {
-            return objectMapper.readerFor(type).readValues(parser);
-        }
+        var parser = objectMapper.createParser(reader);
+        return objectMapper.readerFor(type).readValues(parser);
     }
 
     private static <T> MappingIterator<T> createMappingIterator(ObjectMapper objectMapper, InputStream inputStream, TypeReference<T> type) throws IOException {
-        try (var parser = objectMapper.createParser(inputStream)) {
-            return objectMapper.readerFor(type).readValues(parser);
-        }
+        var parser = objectMapper.createParser(inputStream);
+        return objectMapper.readerFor(type).readValues(parser);
     }
 
     private static <T> MappingIterator<T> createMappingIterator(ObjectMapper objectMapper, InputStream inputStream, Class<T> type) throws IOException {
-        try (var parser = objectMapper.createParser(inputStream)) {
-            return objectMapper.readerFor(type).readValues(parser);
-        }
+        var parser = objectMapper.createParser(inputStream);
+        return objectMapper.readerFor(type).readValues(parser);
     }
 
     public static <T> SequenceWriter createSequenceWriter(ObjectMapper objectMapper, Writer writer, TypeReference<T> type) throws IOException {
         return objectMapper.writerFor(type).writeValues(writer);
+    }
+
+    public static <T> SequenceWriter createBinarySequenceWriter(OutputStream outputStream, TypeReference<T> type) throws IOException {
+        return BINARY_OBJECT_MAPPER.writerFor(type).writeValues(outputStream);
     }
 
     public static <T> SequenceWriter createJsonSequenceWriter(Writer writer, TypeReference<T> type) throws IOException {
--- a/core/src/main/java/io/kestra/core/services/StorageService.java
+++ b/core/src/main/java/io/kestra/core/services/StorageService.java
@@ -28,6 +28,7 @@ import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.utils.RegexUtils;
 import io.kestra.core.storages.StorageSplitInterface;
 
+import com.amazon.ion.util.IonStreamUtils;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.micronaut.core.convert.format.ReadableBytesTypeConverter;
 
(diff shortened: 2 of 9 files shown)


core/src/main/java/io/kestra/core/serializers/FileSerde.java
```diff
@@ -253,33 +253,33 @@ public static <T> Mono<Long> writeAll(Flux<T> values, SequenceWriter seqWriter)
     }
 
     private static <T> MappingIterator<T> createMappingIterator(ObjectMapper objectMapper, Reader reader, TypeReference<T> type) throws IOException {
-        try (var parser = objectMapper.createParser(reader)) {
```
loicmathieu, line 256: This seems wrong, why did you remove the try-with-resource letting the parser unclosed?

core/src/main/java/io/kestra/core/services/StorageService.java
```diff
@@ -18,47 +23,196 @@
...
+    private static final ObjectMapper ION_MAPPER = JacksonMapper.ofIon();
+
     public static List<URI> split(RunContext runContext, StorageSplitInterface storageSplitInterface, URI from) throws IOException, IllegalVariableEvaluationException {
         String fromPath = from.getPath();
         String extension = ".tmp";
         if (fromPath.indexOf('.') >= 0) {
             extension = fromPath.substring(fromPath.lastIndexOf('.'));
         }
 
+        boolean isIon = extension.equals(".ion");
```
loicmathieu: This is a bit annoying as we may have ION files in files not with the `ion` extension

core/src/main/java/io/kestra/core/serializers/FileSerde.java
nkwiatkowski: I removed it because it's closed twice. When it's used, it's always with a finally with close:
    public static <T> Flux<T> readAll(MappingIterator<T> mappingIterator) throws IOException {
        return Flux.<T> create(sink ->
        {
            mappingIterator.forEachRemaining(sink::next);
            sink.complete();
        }, FluxSink.OverflowStrategy.BUFFER)
            .doFinally(throwConsumer(ignored -> mappingIterator.close()));
    }

core/src/main/java/io/kestra/core/services/StorageService.java
```diff
@@ -18,62 +23,213 @@
...
-                    bufferedReader, runContext.render(storageSplitInterface.getPartitions()).as(Integer.class).orElseThrow()
-                );
-            } else if (storageSplitInterface.getRows() != null) {
-                Integer renderedRows = runContext.render(storageSplitInterface.getRows()).as(Integer.class).orElseThrow();
-                splited = StorageService.split(
-                    runContext, extension, runContext.render(storageSplitInterface.getSeparator()).as(String.class).orElseThrow(),
-                    bufferedReader, (bytes, size) -> size >= renderedRows
-                );
+        try (BufferedInputStream inputStream = new BufferedInputStream(runContext.storage().getFile(from), FileSerde.BUFFER_SIZE)) {
+            inputStream.mark(FileSerde.BUFFER_SIZE);
```
loicmathieu:
```suggestion
            inputStream.mark(4);
```

core/src/main/java/io/kestra/core/services/StorageService.java
```diff
@@ -18,62 +23,213 @@
...
+            if (IonStreamUtils.isIonBinary(header)) {
+                return splitIon(runContext, storageSplitInterface, inputStream, extension);
             } else {
-                throw new IllegalArgumentException("Invalid configuration with no size, count, rows, nor regexPattern");
+                return splitText(runContext, storageSplitInterface, inputStream, extension);
+            }
+        }
+    }
+
+    private static List<URI> splitIon(RunContext runContext, StorageSplitInterface storageSplitInterface, InputStream inputStream, String extension) throws IOException, IllegalVariableEvaluationException {
```
loicmathieu: Those new methods seems a lot duplicated, I wonder if we should not istead have single methods that read differently but share the same if/else branches on the storageSplitInterface

loicmathieu (changes requested):

loicmathieu (approved):
LGTM

PR #16621, opened 2026-06-08
fix(worker-controller): deliver storage discovery results on the gRPC SynchronizationContext

fhussonnois:
StorageNameResolver invoked Listener2.onResult2 from its background refresh thread, but gRPC requires that callback to run on the channel's SynchronizationContext. The scheduled refresh therefore threw IllegalStateException, the discovered controller endpoints were dropped, and on cold start the worker never received an address — connect failed with DEADLINE_EXCEEDED.

Pass the channel's SynchronizationContext into the resolver (via NameResolver.Args) and dispatch onResult2 through it. Add a regression test asserting scheduled resolutions are delivered on the context.


ammeek (approved):
LGTM

PR #16622, opened 2026-06-08
fix(worker-controller): drop unsupported DNS SRV record type

fhussonnois:
The DNS discovery 'recordType: SRV' option never performed SRV resolution. Both record types built a 'dns:///<host>' target handed to gRPC's built-in DNS name resolver, which only does A/AAAA lookups of the hostname and never queries _grpc._tcp.<host> SRV records (grpc-java has no SRV endpoint discovery; see grpc/grpc-java#2109). SRV mode therefore A-looked-up the bare hostname and failed with UnknownHostException.

Remove the recordType knob and DnsRecordType enum; DNS discovery is now A-records only (dns:///<host>:<defaultPort>). Also extract the shared PT30S default into a DEFAULT_REFRESH_INTERVAL constant.


ammeek (approved):
LGTM

PR #16627, opened 2026-06-09
fix(flows): downgrade read-only id/namespace toast from error to warning, show once per session

flcarre:
## Summary

- The \"Read-only property\" toast was firing on every editor validation cycle (every keystroke on `id:` / `namespace:`) with `variant: "error"`, making the flow editor appear broken for users editing those fields
- Downgraded to `variant: "warning"` — the constraint is expected behavior, not an error
- Added a per-session deduplication flag (`readonlyToastShown`) so the toast fires at most once per editing session; the flag resets when a new flow is loaded
- Applied the same fix to the dashboard editor (`dashboard.ts`)

Fixes https://github.com/kestra-io/kestra-ee/issues/8366 — 49 users / 59 flows affected in a single week

## Test plan

- [ ] Open an existing flow in the YAML editor, try typing in the `id:` field → toast appears once as a **warning** (yellow), not repeatedly as an error
- [ ] Reload another flow → confirm a new attempt triggers the toast once again (flag reset on load)
- [ ] Open an existing dashboard, try editing the `id:` field → same behavior
- [ ] Create a new flow (isCreating = true) → no toast at all

### DIFF
--- a/ui/src/stores/dashboard.ts
+++ b/ui/src/stores/dashboard.ts
@@ -39,6 +39,7 @@ export const useDashboardStore = defineStore("dashboard", () => {
     const defaultDashboards = ref<DashboardSettings>()
     const chartErrors = ref<string[]>([])
     const isCreating = ref<boolean>(false)
+    const readonlyToastShown = ref(false)
 
     const sourceCode = ref("")
     const sourceCodeOrigin = ref("")
@@ -186,6 +187,7 @@ export const useDashboardStore = defineStore("dashboard", () => {
         activeDashboard.value = data
         sourceCode.value = data.sourceCode ?? ""
         sourceCodeOrigin.value = sourceCode.value
+        readonlyToastShown.value = false
 
         return activeDashboard.value
     }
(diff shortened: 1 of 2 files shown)


flcarre (commented):
## Review — request changes (scope only; the toast fix itself is good)

This branch contains **two unrelated commits**, but the PR only describes one:

- `fix(flows): downgrade read-only id/namespace toast...` — the change this PR is about (`flow.ts` + `dashboard.ts`).
- `feat(design-system): redesign login page to match 2.0 Figma` — a full rewrite of `BasicAuthLogin.vue` plus removal of `setup.login_subtitle` across all 13 locales. This one is undocumented here and overlaps the dedicated open PR #16605, which rewrites the same file and is already ahead of this version (it drops the login-time password-complexity check that this commit still has). Whichever lands first, the other will conflict on `BasicAuthLogin.vue`.

**Recommendation:** rebase this branch to drop the login commit so the PR is purely the toast fix. The login work should live only in #16605.

### Toast fix (in scope) — looks good ✅
- `readonlyToastShown` is reset in `loadFlow()` / `load()`, so the warning fires once per loaded flow as described.
- The `error` to `warning` downgrade is type-safe (`"warning"` is a valid `Message` variant).
- `replaceIdAndNamespace` still runs on every cycle, so the read-only constraint stays enforced; only the toast is deduplicated. Correct.

Once the login commit is dropped, this is an approve from my side.

<details>
<summary>If the login rewrite is intended for #16605, it currently diverges from the design system</summary>

- raw `<form>` / `<input>` / `<button>` instead of `KsForm` / `KsInput` / `KsButton` (and re-implements `KsPassword`'s show-toggle and `KsButton`'s loading by hand)
- two `:deep(svg)` selectors
- hardcoded colors including `color: white` and `rgba(255, 255, 255, 0.3)`
- several magic `px` / `rem` values where tokens exist (e.g. `border-radius: 0.5rem` vs `var(--ks-radius-lg)`)
- password-visibility toggle has `tabindex="-1"` and no `aria-label`; the field error is not linked via `aria-describedby`
</details>

elevatebart (approved):
We need a better UX for flow edition (figure out a way to make the top copyable but not modifyable)

flcarre:
Thanks for the review! Tracked the read-only-but-copyable follow-up in #17615.

PR #16632, opened 2026-06-09
fix(worker): reset reconnect backoff only on confirmed connection

fhussonnois:
The reconnect backoff was reset to the 500ms minimum in sendInitialRequest(), right after the initial request was enqueued. But onNext() only buffers the message and does not prove the transport is live, so on a controller outage the reset fired on every failed attempt and the exponential doubling never survived — the worker hot-looped reconnects at a flat 500ms (~2/s).

Defer the reset to confirmConnection(), invoked once the channel reaches READY (ConnectivityState polled in the worker loop). The backoff now grows 500ms -> 1s -> ... -> 30s and resets only on a genuine (re)connection. This also stops the misleading 'Connected to controller' log and the flow-metastore cache invalidation from firing on every failed attempt during an outage.


ammeek (approved):
LGTM

PR #16642, opened 2026-06-09
fix(worker): handle oversized gRPC payloads cleanly instead of hanging

fhussonnois:
An oversized worker<->controller message left the execution stuck in RUNNING, in both directions. Now the task fails cleanly:
- worker->controller: TaskRun.fail() no longer throws on an immutable attempts list, so the RESOURCE_EXHAUSTED fallback can resend the failed result.
- controller->worker: the worker advertises its maxInboundMessageSize and the controller fails an oversized job before dispatch instead of sending a frame the worker can never receive.


ammeek (approved):
LGTM

PR #16648, opened 2026-06-09
feat: update outputs design for UI 2.0

elevatebart:
closes https://github.com/kestra-io/kestra-ee/issues/8379


MilosPaunovic (approved):
Love the new look! ❤️

yuri1969:
Got a few details regarding the `FILE` input/output from brief testing.

# Inputs

Inputs with `type: FILE` get rendered as plain string URI.

# Outputs

* Output file name seems to be omitted
* Output size (in bytes) seems to be omitted
* Reaching the Download button requires scrolling down
  * To improve the UX the button should be placed somewhere near the top.
* Nested outputs gets completely replaced by a `FILE` output as with:
```yaml
tasks:
  - id: download
    uri: https://google.com
```

PR #16649, opened 2026-06-09
test(worker): mark AbstractGrpcMetaStoreTest as @FlakyTest

AcevedoR:
## Summary

- `GrpcWorkerNamespaceMetaStoreTest` and siblings fail intermittently in CI with:
  ```
  java.io.IOException: Failed to bind to address 0.0.0.0/0.0.0.0:49646
  ```
- Root cause: multiple `AbstractGrpcMetaStoreTest` subclasses in the same module can share a Micronaut test context (same configuration = same cached context). The shared context evaluates `${random.port}` once. When one class stops the controller in `@AfterAll` and another restarts it in `@BeforeAll`, the OS port may still be in `TIME_WAIT` causing a bind failure on the same port.
- Adding `@FlakyTest` to the abstract base covers all subclasses in both OSS (`worker`) and EE (`worker-ee`) via JUnit 5 annotation inheritance.

## Test plan
- [ ] No logic changes — annotation only
- [ ] CI will retry these classes on bind failure

### DIFF
--- a/worker/src/test/java/io/kestra/worker/stores/AbstractGrpcMetaStoreTest.java
+++ b/worker/src/test/java/io/kestra/worker/stores/AbstractGrpcMetaStoreTest.java
@@ -4,6 +4,7 @@ import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.TestInstance;
 
+import io.kestra.core.junit.annotations.FlakyTest;
 import io.kestra.core.worker.Controller;
 import io.kestra.core.worker.models.WorkerInfo;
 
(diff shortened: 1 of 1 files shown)


AcevedoR:
Superseded by #16772, which fixes the root cause instead of marking tests @FlakyTest.

Root cause: all concrete subclasses shared one Micronaut context because their annotations were identical. `${random.port}` was evaluated once; when a subclass stopped the gRPC Controller and the next tried to restart it on the same port, the OS port was still in TIME_WAIT. The fix adds a unique `@Property(name = "test.context.id", ...)` sentinel per subclass to force isolated contexts and fresh port allocation. Companion EE change in kestra-io/kestra-ee#8603.

PR #16653, opened 2026-06-09
feat(core): add subflow() Pebble function

brian-mulier-p:
Closes kestra-io/kestra-ee#7850


core/src/main/java/io/kestra/core/models/executions/ExecutionKind.java
```diff
@@ -4,11 +4,16 @@
...
+ *   non-NORMAL kinds
  * - NORMAL: anything else, for backward compatibility NORMAL is not persisted but null is used instead
  */
 public enum ExecutionKind {
     NORMAL,
     TEST,
     PLAYGROUND,
-    LOOP
+    LOOP,
+    SUBFLOW_FUNCTION
```
loicmathieu: I don't understand, a SUBFLOW execution is a standard execution as the one created by the  `Subflow` tasl.

Even if the issue originally asked it for populating inputs, I think this may be used in a lot of other places so I would not set it as a specific KIND as those executions are "not normal": not shown inside the UI, not taken into account by flow triggers, not shown as dependencies, ...
I don't think this applies here

core/src/main/java/io/kestra/core/services/WebhookService.java
```diff
@@ -211,6 +217,36 @@ public Flux<Event<Execution>> followExecution(Execution execution, Flow flow) {
...
+     * the streaming service: a fast subflow can terminate (and have its streaming events dropped, since
+     * no subscriber is registered for its id yet) before we could subscribe. Polling the repository is
+     * race-free for a fire-and-wait caller.
+     *
+     * @param execution The execution to start
+     * @param flow The flow associated with the execution
+     * @param completionTimeout The maximum time to wait for the execution to terminate
+     * @return the terminal execution
+     */
+    public Execution runAndWait(Execution execution, Flow flow, Duration completionTimeout) {
```
loicmathieu: This has nothing to do here, it's not related to webhook

loicmathieu: Better to move it to the `ExecutionService` for example

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunctionConfiguration.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @param completionTimeout maximum time to wait for the synchronously-run subflow to terminate.
+ *        Kept short by default because the call blocks input rendering (e.g. the execute form).
+ * @param maxDepth maximum nesting depth of {@code subflow()} calls on the same rendering thread,
+ *        guarding against runaway recursion (a subflow whose own inputs call {@code subflow()}).
+ */
+@ConfigurationProperties("kestra.pebble.subflow-function")
+public record SubflowFunctionConfiguration(
+    @Bindable(defaultValue = "PT1M") Duration completionTimeout,
+    @Bindable(defaultValue = "3") int maxDepth
```
loicmathieu, line 23: A subflow function is recursive if it calls the flow it's defined on right? So how can it be only 2-level recursive and not 3?

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunctionConfiguration.java
```diff
@@ -0,0 +1,21 @@
...
+ * Configuration for the {@code subflow()} Pebble function.
+ *
+ * @param completionTimeout maximum time to wait for the synchronously-run subflow to terminate.
+ *        Kept short by default because the call blocks input rendering (e.g. the execute form).
+ * @param maxDepth maximum nesting depth of {@code subflow()} calls on the same rendering thread,
+ *        guarding against runaway recursion (a subflow whose own inputs call {@code subflow()}).
+ */
+@ConfigurationProperties("kestra.pebble.subflow-function")
+public record SubflowFunctionConfiguration(
+    @Bindable(defaultValue = "PT1M") Duration completionTimeout,
```
loicmathieu: It's very opinionated and low.
I understand for inputs you don't want it to be more than 1mn, but I'm sure this function will be used elsewhere where this timeout would be a little too aggressive

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
```diff
@@ -0,0 +1,166 @@
...
+    private static final String ID_ARG = "id";
+    private static final String INPUTS_ARG = "inputs";
+    private static final String REVISION_ARG = "revision";
+
+    /**
+     * Guards against runaway recursion: a subflow's own inputs may call {@code subflow()}, which is
+     * rendered synchronously on this same thread before any blocking wait. Incremented on entry and
+     * decremented in a finally block.
+     */
+    private static final ThreadLocal<Integer> DEPTH = ThreadLocal.withInitial(() -> 0);
```
loicmathieu, line 70: This would only work for inputs.
I think another check should be done to prohibits a subflow function to call the current flow as if the recursivity is inside another thread (for ex if it's used in a task) it may not hit that guardrails.

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
```diff
@@ -0,0 +1,166 @@
...
+ */
+@Slf4j
+@Singleton
+public class SubflowFunction implements KestraFunction {
+    public static final String NAME = "subflow";
+
+    private static final String NAMESPACE_ARG = "namespace";
+    private static final String ID_ARG = "id";
+    private static final String INPUTS_ARG = "inputs";
+    private static final String REVISION_ARG = "revision";
```
loicmathieu, line 61: Maybe allow setting labels also

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
```diff
@@ -0,0 +1,166 @@
...
+            }
+
+            Execution execution;
+            try {
+                execution = Execution.newExecution(
+                    targetFlow,
+                    (f, e) -> flowInputOutput.get().readExecutionInputs(f, e, inputs),
+                    List.of(new Label(Label.FROM, NAME)),
+                    Optional.empty(),
+                    ExecutionKind.SUBFLOW_FUNCTION
```
loicmathieu: I don't think it's a good idea, I already explained why.
But you can use the `system.from` label to denote executions created from the subflow function

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
```diff
@@ -0,0 +1,166 @@
...
+                throw new PebbleException(e, "Failed to run subflow '" + namespace + "'.'" + id + "': " + e.getMessage(), lineNumber, self.getName());
+            }
+
+            State.Type state = terminated.getState().getCurrent();
+            if (state != State.Type.SUCCESS && state != State.Type.WARNING) {
+                throw new PebbleException(null, "Subflow '" + namespace + "'.'" + id + "' ended in state " + state
+                    + " (execution " + terminated.getId() + ").", lineNumber, self.getName());
+            }
+
+            return terminated;
```
loicmathieu: I would not return the full execution object.
From the `Subflow` task, we only return the execution outputs, I think it would be best to do the same.

If you need more information, you can return a specific object, but not the full execution object as it may change in incompatible ways and exposes too much information which would clutter the execution context.

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunctionConfiguration.java
loicmathieu: Maybe a better option would be:
- allow setting the timeout inside the function
- defaultTimeout: the timeout if not set inside the function
- maxTimeout: a maximum that cannot be exceeded, if a function set an higher value this is a runtime exception

loicmathieu (changes requested):
There is an architecture issue here as this function uses the execution repository which is not something we should do inside a function.

So basically, today, this function seems to be designed to be used only for inputs from the webserver even if it's not limited to. I wonder if we should add a check for that and restrict to inputs and the webserver (inputs can be resolved by the executor for ex for a Subflow task).

I'm not comptfortable with using the WebhookService and the streamingService, I didn't check in details but we might have possible resource leak here if the subflow function timeout or is interrupted...

core/src/main/java/io/kestra/core/models/executions/ExecutionKind.java
brian-mulier-p: This was in the original issue requirement. I didn't put a dedicated kind at first but then I saw back the requirement and thought about how many executions such function could generate and that it could effectively clutter the execution graphs quite a lot, idk see with @kestra-io/products but I think both sides are valid

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunctionConfiguration.java
brian-mulier-p: Not sure to understand what you meant, Flow A could call subflow B that could call subflow A (or C then A, whatever)...

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunctionConfiguration.java
brian-mulier-p: Why not

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
brian-mulier-p: Alternatively we could leverage system labels to detect such scenario regardless of thread

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
brian-mulier-p: Idk it's restricting quite a lot, you can't detect an subflow failure then, maybe I can use a wrapper object but imo it's very valuable to return more than the output itself. In any case I think the {outputs} wrapping object is at least needed so that we can extend what's return later, we don't want to do subflow(...).outputId directly...

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunctionConfiguration.java
loicmathieu: Nevermind, I wrote that before seeing how it was actually implemented.

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
loicmathieu: I don't think it would be a good idea as labels are a user-facing features and here it's a technical detail.
I would do a simple check to see if the subflow is the current flow (gathered from the execution context)

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
loicmathieu: OK so a wrapper object with a curated list of attributes (execution outputs, state, labels to start with) seems to be a good starting point.

core/src/main/java/io/kestra/core/models/executions/ExecutionKind.java
loicmathieu: KIND is to differentiate between regular executions (production) and test/debug/playground.
Here it's a regular execution so I think it should use null/NORMAL as all other executions

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
brian-mulier-p: It's not enough, if you have a Flow A -> Flow B -> Flow A recursiveness scenario it would not catch it so I really think there's no way around than having hard limit counter. I don't think it's an issue to have a system.subflowLevel label :thinking: (Naming TBD)

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
brian-mulier-p: good catch

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
brian-mulier-p: But if we limit it to webserver we can probably rollback to the threadlocal?

loicmathieu: yeah, or a scoped variable

brian-mulier-p:
@loicmathieu I've handled all the changes

core/src/main/java/io/kestra/core/runners/pebble/functions/SubflowFunction.java
```diff
@@ -0,0 +1,248 @@
...
+
+    @SuppressWarnings("unchecked")
+    private List<Label> buildLabels(Object rawLabels, PebbleTemplate self, int lineNumber) {
+        List<Label> labels = new ArrayList<>();
+        if (rawLabels != null) {
+            if (!(rawLabels instanceof Map)) {
+                throw new PebbleException(null, "The 'subflow' function 'labels' must be a map of string keys to values.", lineNumber, self.getName());
+            }
+            ((Map<String, Object>) rawLabels).forEach((key, value) -> {
+                if (value != null) {
```
loicmathieu, line 200: You need to check that no system labels (except system.correlationId) are set.

loicmathieu (approved):
A small change request for validating against passing system labels and also as we discussed it may be great to only emit the function bean inside STANDALON/WEBSERVER.

No need for another pass of review, I'm validating it anyway

aj-emerich:
@brian-mulier-p I am going to work on documenting this this week. I may need your review to ensure it's captured correctly. Great feature 👏

PR #16655, opened 2026-06-09
feat: add FORM input type for grouped flow execution forms

brian-mulier-p:
Implements the FORM input type from kestra-io/kestra-ee#8330 — a new flow input that groups related inputs under a `displayName` + `description`.

- Backend: new `FORM` input expanded to dotted-id leaves so existing resolution + `flattenToNestedMap` produce a nested payload; FORM never reaches resolution. Validation rejects duplicate / prefix-conflict expanded paths, requires dotted `dependsOn` refs for form children (bad refs rejected loudly), and rejects nested FORM / FORM-with-defaults.
- Flow Execute modal: Next/Back wizard — one step per top-level FORM, a contiguous run of ungrouped inputs as its own step, recap with per-section Edit, localStorage persistence (clear-on-success).
- Progressive autocomplete: `inputs.<form>.` suggests the FORM's children.
- Translations for the new wizard keys (`back` / `next` / `done` / `review your inputs`) across all supported locales.

⚠️ **Design deviation from the issue's note:** the FORM is a real structural wrapper — submission is **NESTED** (`{"environment":{"region":"EU"}}`) and tasks reference `{{ inputs.environment.region }}`. This intentionally diverges from the issue comment describing FORM as UI-only grouping with a flat payload.

Companion EE PR: kestra-io/kestra-ee#8447

<details>
<summary>Flow To Test</summary>

```yaml
id: flow_with_form_input
namespace: company.team

inputs:
  - id: requester
    type: STRING
    required: true
    description: Who is requesting this environment

  - id: environment
    type: FORM            # needed for the backend to parse the group; carries displayName + description
    displayName: Environment Setup
    description: Where the environment runs and what it is for
    inputs:
      - id: region
        type: SELECT
        required: true
        defaults: eu-central-1
        values:
          - eu-central-1
          - eu-west-1
          - us-east-1
      - id: data_center
        type: SELECT
        required: true
        # Issue #7850: values are fetched from a subflow at form-render time.
        # dependsOn re-runs the fetch whenever region changes, so the dropdown
        # only ever shows data centers that exist in the selected region.
        # No Pause hack, no extra post-task. The form fetches its own options.
        dependsOn:
          inputs:
            - environment.region
        expression: "{{ subflow(namespace='company.system', id='fetch_datacenters', inputs={'region': inputs.environment.region}).outputs.datacenters }}"

tasks:
  - id: hello
    message: Hello World! 🚀
```
</details>


core/src/main/java/io/kestra/core/runners/RunVariables.java
loicmathieu, line 1: General feedback but not sure there is a better design.
I feel now the code base is clutted with this `Input.expandToLeaves()` method, I wonder if we should not do it once somewhere...

core/src/main/java/io/kestra/plugin/core/trigger/McpToolTrigger.java
```diff
@@ -181,7 +182,7 @@ public Execution evaluate(
         List<Label> additionalLabels
     ) {
         Execution execution = Execution.builder()
-            .inputs(input)
+            .inputs(MapUtils.flattenToNestedMap(input))
```
loicmathieu, line 185: If it needs to be done everywhere, we might miss some places here :(

loicmathieu (approved):
Overall it's good to me but I'm not fan of all those `MapUtils.flattenToNestMap()` and `Input.explandToLeaves()` this clutter the codebase and I fear we already miss one (subflow for ex?).

I really think it would lead to bugs as the code evolved and we forgot to call them :(

loicmathieu:
@elevatebart can you review the frontend part?

elevatebart (approved):
Testing it out, I found that one cannot use the linked flow even by fixing it.

The feature missing is: 
- One cannot use the output of an input to create the options for the next input.

If it is clear that it is out of scope, then merge away.

ui/src/components/inputs/InputsForm.vue
```diff
@@ -334,6 +390,125 @@
...
+
+    const steps = computed<WizardStep[]>(() => isWizard.value ? buildWizardSteps(props.initialInputs as any) : [])
+
+    const currentStep = ref(0)
+    const current = computed<WizardStep | undefined>(() => steps.value[currentStep.value])
+    const recapIndex = computed(() => steps.value.length - 1)
+    const returnToRecap = ref(false)
+    const navLoading = ref(false)
+    // Only swap the Next label to the loading message once the validate round-trip is slow
+    // enough to notice (>500ms). A fast validate shows just the brief spinner — no jarring label flash.
```
elevatebart: It could have been sane to move this into a composable outside of the main component. I find it hard to parse.

elevatebart (approved):

ui/tests/unit/utils/inputs.spec.ts
elevatebart, line 1: This should very much be a storybook test as visual feedback would give more value

core/src/main/java/io/kestra/core/runners/RunVariables.java
brian-mulier-p: I added a method directly in flow to get resolved inputs, this should help, I can't think of better way than that :'(

ui/tests/unit/utils/inputs.spec.ts
brian-mulier-p: There are some storybook test already, just not there

brian-mulier-p:
> Testing it out, I found that one cannot use the linked flow even by fixing it.
> 
> The feature missing is:
> 
> * One cannot use the output of an input to create the options for the next input.
> 
> If it is clear that it is out of scope, then merge away.

I'm not sure to follow @elevatebart I couldn't reproduce such behavior, if you have a dependsOn there is no such issue, the validate retriggers whenever an input value changes and updates options accordingly

elevatebart:
In [this comment](https://github.com/kestra-io/kestra-ee/issues/8330#issuecomment-4629970933) there is an example flow. It does not work and the code does not seem to yield a list of data-center on a region. 

Do you have another example flow for me to test?

PR #16678, opened 2026-06-10
fix: use --ks-text instead of --ks-content that does not exist anymore

elevatebart:
Cleanup since the update of the themes


flcarre (commented):
I dont see `--ks-text-tertiary` and `--ks-text-inverse` variables, are you sure we have them ?

elevatebart:
> I dont see `--ks-text-tertiary` and `--ks-text-inverse` variables, are you sure we have them ?

You are right. I will make a list in storybook of all the colors we do have. In the meantime we do not have `--ks-content-tertiary` or `--ks-content-inverse` either. So it does not improve nor deter from the current status quo.

flcarre (approved):

PR #16679, opened 2026-06-10
fix: return empty manifest instead of 404 when plugin UI is missing

ammeek:
### ✨ Description

Fixes an issue where opening community blueprints that reference uninstalled plugins could incorrectly redirect users to the global **Page Not Found** view.

`POST /api/v1/plugins/pluginUiManifest` previously behaved as an all-or-nothing endpoint. If none of the requested task classes could be resolved from the plugin registry, it threw a `NotFoundException` (`HTTP 404`).

This commonly occurred when viewing community blueprints that reference plugins not installed on the current instance, such as:

* `io.kestra.plugin.ai.*`
* `io.kestra.plugin.googleworkspace.*`
* `io.kestra.plugin.slack.*`

For example:

```text
/ui/main/blueprints/flow/community/ai-gdpr-dpa-reviewer
```

Although the blueprint itself was valid, the manifest request returned a 404, which the frontend interpreted as a fatal routing error and redirected to the global **Page Not Found** page.

#### `PluginController.getPluginUiManifest`

Removed the two `NotFoundException` throws.

The endpoint already supports partial manifest generation, so the all-miss case now behaves consistently with partial matches:

* **Before:** returns `404 Not Found`
* **After:** returns `200 OK` with an empty or partial manifest

Example response:

```json
{
  "manifest": {}
}
```

This allows consumers to gracefully handle missing plugins without treating them as a request failure.

### Result

Community blueprints that reference unavailable plugins now open normally:

* Installed plugins continue to provide their custom UI.
* Missing plugins simply contribute no manifest entries.
* Users no longer see a false "Page Not Found" error when viewing otherwise valid blueprints.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra-ee/issues/8402
Closes https://github.com/kestra-io/kestra-ee/issues/8446

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java
@@ -19,7 +19,6 @@ import io.kestra.core.models.ui.TaskWithVersion;
 import io.kestra.core.plugins.PluginRegistry;
 import io.kestra.core.plugins.RegisteredPlugin;
 import io.kestra.core.repositories.ArrayListTotal;
-import io.kestra.core.utils.ListUtils;
 import io.kestra.core.utils.MapUtils;
 import io.kestra.webserver.converters.QueryFilterFormat;
 import io.kestra.webserver.responses.PagedResults;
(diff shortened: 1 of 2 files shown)


anna-geller:
closes https://github.com/kestra-io/kestra-ee/issues/8446

fhussonnois (approved):

PR #16681, opened 2026-06-10
refactor(tests): compose @Test into @FlakyTest and remove redundant annotations

AcevedoR:
## Summary

- Add `@Test` as a meta-annotation to `@FlakyTest` so methods annotated with only `@FlakyTest` are recognized as JUnit 5 tests without a separate `@Test`.
- Remove the now-redundant `@Test` from all existing method-level `@FlakyTest` sites across the codebase (9 files).

**Before:**
```java
@Override
@FlakyTest(description = "...")
@Test
@LoadFlows(...)
public void someTest() { ... }
```

**After:**
```java
@Override
@FlakyTest(description = "...")
@LoadFlows(...)
public void someTest() { ... }
```

## Note

Companion EE PR: https://github.com/kestra-io/kestra-ee/pull/8443

Class-level `@FlakyTest` usage is unaffected — `@Test` on an annotation type is only effective when `@FlakyTest` is applied at the method level.

### DIFF
--- a/cli/src/test/java/io/kestra/cli/services/FileChangedEventListenerTest.java
+++ b/cli/src/test/java/io/kestra/cli/services/FileChangedEventListenerTest.java
@@ -60,6 +60,7 @@ class FileChangedEventListenerTest {
     }
 
     @FlakyTest(description = "OS file watcher events are non-deterministic; CI filesystems may delay or batch inotify events")
+    @Test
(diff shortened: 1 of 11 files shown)


AcevedoR:
Reverting — the @FlakyTest compose approach was rejected; keeping @Test explicit on overrides.

PR #16688, opened 2026-06-10
fix(core): don't disable new label row when its key includes similar prefix

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra-ee/issues/8479


Pradumnasaraf (approved):
Looks good to merge! That was a fast PR 🚀

Image

PR #16696, opened 2026-06-10
fix(Duration): task node duration blinking

RemiBarthe:
closes https://github.com/kestra-io/kestra-ee/issues/8418

This pull request makes improvements to the display and formatting of durations in the `Duration.vue` component. The main focus is on ensuring consistent numeric alignment and formatting for durations, as well as minor style adjustments elsewhere.

**Duration formatting and display improvements:**

* Ensured durations that are bare seconds (e.g., `5s`) are always displayed with two decimal places (e.g., `5.00s`) for consistency.
* Introduced a scoped style for `.ks-duration-value` to use tabular numbers, improving alignment of numeric values in the UI.

**Other style adjustments:**

* Removed the monospace font family from the `.status-tag__text` class in `TaskNode.vue` to standardize text appearance.

### DIFF
--- a/ui/packages/topology/src/misc/Duration.vue
+++ b/ui/packages/topology/src/misc/Duration.vue
@@ -11,7 +11,7 @@
             </span>
         </template>
         <template #default>
-            <span v-html="duration" />
+            <span class="ks-duration-value" v-html="duration" />
         </template>
     </KsTooltip>
 </template>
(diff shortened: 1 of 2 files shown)


ui/packages/topology/src/misc/Duration.vue
```diff
@@ -101,13 +101,18 @@
...
 
     function computeDuration() {
-        duration.value =
-            filteredHistories.value.length === 0
-                ? "&nbsp;"
-                : Utils.humanDuration(delta() / 1000, {
-                    maxDecimalPoints: 2,
-                    units: ["h", "m", "s"],
-                })
+        if (filteredHistories.value.length === 0) {
```
Piyush-r-bhaskar, line 104: Do we really need this all changes ? i think `font-variant-numeric: tabular-nums;` should be enough. try once if it works ?

Piyush-r-bhaskar (changes requested):

Piyush-r-bhaskar (approved):

ui/packages/topology/src/misc/Duration.vue
RemiBarthe: humanDuration doesn't display decimals for whole seconds (2s instead of 2.00s), which causes the text width to change every time a full second is reached, making the live timer blink.

I tried fixing it by forcing a fixed width, but the result didn't look great. With these changes, the display remains consistent.

Piyush-r-bhaskar: okie...go ahead

Piyush-r-bhaskar (approved):

PR #16387, opened 2026-05-29
fix(flows): Flow Editor 2.0 UI fixes and UX polish

flcarre:
## Summary

Frontend fixes and UX polish for the 2.0 Flow Editor (EE-reported, fixed in the shared OSS components):

**Bug fixes**
- **Blueprint plugin icons too large** (`BlueprintDetail.vue`): the detail "Plugins" section rendered icons in fixed 100×100px tiles. Now token-based ~3rem chips; removed a dead `:deep(.wrapper)` override. Fixes https://github.com/kestra-io/kestra-ee/issues/8063.
- **No-Code taskRunner field missing its icon** (`Element.vue`): an empty `taskRunner` showed `KsTaskIcon`'s blank-page fallback. It now renders a clear placeholder affordance. Fixes https://github.com/kestra-io/kestra-ee/issues/8060.

**UX quick wins**
- No-Code empty complex-field placeholder reads as an action: centered `+` icon + an "Add {Field}" label (i18n-interpolated via `no_code.add_field`), with the delete button hidden until a value is set.
- Plugin docs panel shows a `KsSkeleton` while plugins load (was a bare "Loading plugins…" text).
- Flow editor stat counters (Revisions / Dependencies) now have a visible hover state — the previous hover used an undefined token (`--ks-background-card` → `--ks-bg-hover`).

**Cleanup**
- Removed dead `:deep(.kel-breadcrumb__*)` `!important` overrides in the Docs-tab breadcrumb (`PluginList.vue`) — the DS emits `.ks-breadcrumb__*`, so they never matched. Tokenized a magic-number gap.

### Related EE issues — verified already fixed on OSS develop (closed)
- https://github.com/kestra-io/kestra-ee/issues/8062 (new flow showed revisions before save) — `FlowEditorStats` already guards with `hasFlow` + `isCreating`.
- https://github.com/kestra-io/kestra-ee/issues/8064 (Docs breadcrumb looked off) — the breadcrumb was migrated to `KsBreadcrumb` and renders correctly; this PR only removes the leftover dead CSS.

## Screenshots
01-blueprint-icons_AFTER
02-taskrunner_AFTER
03-docs-loading_AFTER
04-stat-hover_AFTER

## Test plan
- [ ] CI: UI lint, type-check, and unit tests
- [x] ESLint clean on all changed files (local)
- [x] Manual QA on dev server (light + dark)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #16697, opened 2026-06-10
feat(executions): redesign Set labels popup for clearer add/edit

flcarre:
## Summary

Reworks the execution **Set labels** popup so adding and editing labels reads clearly, per [kestra-ee#8480](https://github.com/kestra-io/kestra-ee/issues/8480). Today every row (including rows for labels that already exist) carries its own `+` and `−`, which feels off next to an already-created label.

This applies the same principle as [#16387](https://github.com/kestra-io/kestra/pull/16387) (turning a redundant inline `+` into a single clear "Add" affordance):

- **One "Add label" button at the top** of the list opens a fresh empty row.
- **Existing label rows keep only a remove (`−`) action** — no more per-row `+`. Their key stays locked (unchanged); the value remains editable.
- Removing all rows is now allowed (clears the labels); the top button re-adds.
- Trigger renamed **"Set extra labels" → "Manage labels"** (it manages add/edit/remove, not just add).
- Primary action renamed **"OK" → "Save"**.

The change lives in the shared `LabelInput`, so the **flow-run** ("Other properties → Execution labels") and **bulk set-labels** dialogs get the same layout for free.

## Demo

> _Drag-and-drop the `manage-labels-demo.mp4` here (GitHub only accepts video uploads via the browser)._ Add → type → Save → reopen, showing the new label persist as a locked existing row.

## Screenshots (before / after)

| Before | After |
|--------|-------|
| Per-row `+`/`−` on every row, incl. existing labels | Single "Add label" on top; existing rows keep only `−` |

https://github.com/user-attachments/assets/c831b15d-e491-4e51-bb84-5f3c1bb61d06

## Test plan

- [x] New unit test `ui/tests/unit/components/labels/LabelInput.spec.ts` (7 tests) — add row, remove-to-zero, existing-key lock, value editable, new-row editable key, `update:labels` emit
- [x] ESLint clean on changed files
- [x] `check.js`: no extra keys across all 12 locales
- [x] Manual QA on dev server (light): execution popup, flow-run labels, no console errors
- [ ] CI: UI lint, type-check, unit tests

Closes https://github.com/kestra-io/kestra-ee/issues/8480

### DIFF
--- a/ui/src/components/labels/LabelInput.vue
+++ b/ui/src/components/labels/LabelInput.vue
@@ -1,29 +1,32 @@
 <template>
-    <div
-        class="d-flex w-100 mb-2"
-        v-for="(label, index) in locals"
-        :key="index"
-    >
-        <div class="flex-grow-1 d-flex align-items-center">
+    <div class="label-input">
+        <KsButton class="label-input__add" :icon="Plus" @click="addItem">
+            {{ $t("add label") }}
+        </KsButton>
+
+        <div
+            class="label-input__row"
+            v-for="(label, index) in locals"
+            :key="index"
+        >
             <KsInput
-                class="form-control me-2"
+                class="label-input__field"
                 :placeholder="$t('key')"
                 :modelValue="(label.key as string | undefined)"
                 :disabled="existingRows.has(label)"
                 @update:model-value="update(index, $event, 'key')"
             />
             <KsInput
-                class="form-control me-2"
+                class="label-input__field"
                 :placeholder="$t('value')"
                 :modelValue="(label.value as string | undefined)"
                 @update:model-value="update(index, $event, 'value')"
             />
-        </div>
-        <div class="flex-shrink-1">
-            <KsButtonGroup class="d-flex">
-                <KsButton :icon="Plus" @click="addItem" :tooltip="$t('add label')" />
-                <KsButton :icon="Minus" @click="removeItem(index)" :tooltip="$t('remove label')" />
-            </KsButtonGroup>
+            <KsButton
+                :icon="Minus"
+                :tooltip="$t('remove label')"
+                @click="removeItem(index)"
+            />
         </div>
     </div>
 </template>
@@ -57,9 +60,6 @@
 
     const removeItem = (index: number) => {
         locals.value.splice(index, 1)
-        if (locals.value.length === 0) {
-            addItem()
-        }
         emit("update:labels", locals.value)
     }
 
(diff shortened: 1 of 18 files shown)


ui/src/components/labels/LabelInput.vue
```diff
@@ -86,3 +83,26 @@
...
+        &__row {
+            display: flex;
+            align-items: center;
+            gap: var(--ks-spacing-2);
+        }
+
+        &__field {
+            flex: 1;
+            min-width: 0;
+        }
```
elevatebart: Prefer to avoid using concatenated css classes.
- They often convey BEM which is used to segregate CSS rules but is made moot by using the `scoped` attribute.
- Vue dev tools allow you to navigate from a class to its definition, but not if the class is split.
- Search tool will never find a contructed class.
- Please KISS (**Keep** **It** as **Simple** as if I were **Stupid**)

NOTE: I know it was written by AI. We should tell claude to stop doing that. Lets add a rule in the `AGENTS.md`.

ui/tests/unit/components/labels/LabelInput.spec.ts
elevatebart, line 1: This is the moment where I NEED storybook tests because the visual part of this issue is paramount to its success. 
Only terminal output can be seen in this test. Tests are meant to check that value is not lost.
We are improving UX not terminal output. So we should be able to see UX in the tests.

elevatebart (approved):
Minor improvements requested, nothing blocking

ui/src/components/labels/LabelInput.vue
flcarre: Always good to keep in mind ! Thanks !

flcarre: Done in 29efb2ab2 — flattened to literal class selectors (`.label-input-row`, `.label-input-add`, `.label-input-field`), no more `&__` construction, so they're greppable and devtools can jump to them. Also added a rule to `ui/AGENTS.md` so AI stops generating constructed classes.

ui/tests/unit/components/labels/LabelInput.spec.ts
flcarre: Agreed — a `LabelInput.stories.tsx` already existed (Default / WithValue / WithExistingLabels) but was display-only. In 29efb2ab2 I added `play` assertions to it so the new UX is verified in the browser: Add-label button visible, existing-label keys locked, one remove action per row, and "Add label" appending an editable row. Kept this unit test for the pure logic (remove-to-zero, emit payload).

PR #16700, opened 2026-06-10
fix: widen misc store theme type to SelectedTheme

### LINKED ISSUE #16520: misc store `theme` ref typed as "light"|"dark" but holds raw dark-2/syncWithSystem
### Problem

`ui/src/override/stores/misc.ts` declares the theme ref as:

```ts
const theme = ref<"light" | "dark">("light")
```

but `switchTheme` (`ui/src/utils/utils.ts`) assigns the raw selected theme, which can be `"dark-2"` or `"syncWithSystem"`:

```ts
miscStore.theme = theme // theme ∈ "light" | "dark" | "dark-2" | "syncWithSystem"
```

Because `switchTheme(miscStore: any, …)` is untyped, TypeScript can't catch the mismatch. The ref silently holds values outside its declared union.

### Why it matters

This is the type-laundering that enabled the `dark-2` bug class (editor white background, charts color seed, plugin code highlight). Those consumers were all migrated to `getTheme()` in PR #16310, which resolves `dark-2 → dark` and `syncWithSystem`. But the underlying type still lies, so any new consumer doing `miscStore.theme === "dark"` will treat `dark-2` and `syncWithSystem` as not-dark and silently reintroduce the bug.

### Suggested fix

Widen the ref to the real union and keep the convention of resolving through `getTheme()` for any dark/light decision:

```ts
import type {SelectedTheme} from "../../utils/utils"
const theme = ref<SelectedTheme>("syncWithSystem")
```

Note: widening alone does not block `=== "dark"` mistakes (comparing a 4-member union to a literal is valid TS), so pair it with a lint note or a doc rule that theme decisions go through `getTheme()`. Verify no new TS errors surface in EE (`misc.ts` is a shared override store).

### Context

Deferred from PR #16310 review (out of that PR's diff, shared store with EE blast radius).

Siddhu2k04:
### ✨ Description

This PR fixes a theme type mismatch in the misc store.

Previously, `ui/src/override/stores/misc.ts` declared the theme state as:

```ts
ref<"light" | "dark">
```

However, `switchTheme()` could assign additional valid theme values such as `"dark-2"` and `"syncWithSystem"`, causing the store to hold values outside its declared TypeScript union.

This change updates the store to use the correct `SelectedTheme` type, ensuring type safety and consistency with the application's theme selection logic.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/16520

### 🎨 Frontend Checklist

* [x] Code builds without errors (`npm run build`)
* [x] Updated theme typing to match actual runtime values
* [ ] Screenshots not required (no UI changes)

### 📝 Additional Notes

* Replaced the narrow `"light" | "dark"` union with `SelectedTheme`.
* No functional UI changes.
* This is a TypeScript type-safety improvement that aligns the store state with the values assigned by `switchTheme()`.
* Theme resolution behavior remains unchanged and continues to rely on `getTheme()` where applicable.

### 🤖 AI Authors

Why did the cat become a programmer? Because it loved catching bugs! 🐱


Siddhu2k04:
I've submitted a fix for #16520. The PR updates the misc store theme typing to match the actual runtime theme values.

Could you please review it when convenient? Thank you!

flcarre (approved):
Thanks for taking this on 🙌 This is exactly what #16520 asked for: the misc store `theme` ref now carries the real `SelectedTheme` union, so the `any`-laundering in `switchTheme()` that enabled the `dark-2` bug class can no longer slip values past the type checker.

I verified the blast radius on `develop`:

- Every consumer of `miscStore.theme` uses it either as a reactive invalidation signal (`void miscStore.theme`, resolved through `getTheme()`) or as a `:key`. There are zero `=== "dark"` / `=== "light"` literal comparisons against it, so widening the union does not surface any latent comparison bugs.
- The new `"syncWithSystem"` default is safe: `App.vue` calls `switchTheme(miscStore)` on boot before render, so the ref holds the resolved theme before any consumer reads it.

Approving. A few non-blocking notes for whenever you (or a follow-up) want them:

1. **Optional dedup (`utils.ts`)** — `theme = localStorage.getItem("theme") as SelectedTheme` casts an arbitrary stored string straight to the union. It is no worse than before (the previous line did the same against a `string`), but the file already exports `getSelectedTheme()` a few lines below, which performs the same null-coalesce to `"syncWithSystem"`. Reusing it would drop both the cast and the duplication.
2. **Follow-up (`BasicSettings.vue`)** — the two `value as SelectedTheme` casts exist only because `ThemePicker` emits `string`. Tightening `ThemePicker`'s emit type to `SelectedTheme` would let both casts go away. Out of scope for this PR.
3. **EE verification** — `ui/src/override/stores/misc.ts` is a shared override store. Worth a quick `check:types` against kestra-ee before merge, in case EE ships its own `misc.ts` typing or passes its store into the now-tightened `switchTheme(miscStore: {theme: SelectedTheme})`.

Nice, focused change. 👍

PR #16703, opened 2026-06-10
fix(scheduler): prevent flow-listener from resurrecting a terminated trigger's execution lock

fhussonnois:
The flow-change listener re-runs over all flows on any flow change and, for Schedulables with recoverMissedSchedules=NONE, rewrites the trigger from a start-of-method snapshot. If the executor cleared the execution lock in between, the stale executionId is written back, re-locking the trigger forever and stopping the schedule until manually unlocked.

Never rewrite a locked trigger from the listener, and decide from a fresh read of the trigger store instead of the snapshot.

Related-to: kestra-io/kestra-ee#8507

### DIFF
--- a/scheduler/src/test/java/io/kestra/scheduler/SchedulerInitializeTriggersTest.java
+++ b/scheduler/src/test/java/io/kestra/scheduler/SchedulerInitializeTriggersTest.java
@@ -1,0 +1,71 @@
+package io.kestra.scheduler;
+
+import java.time.ZonedDateTime;
+import java.util.Optional;
+
+import org.junit.jupiter.api.Test;
+
+import io.kestra.core.models.triggers.RecoverMissedSchedules;
+import io.kestra.core.models.triggers.Trigger;
+import io.kestra.plugin.core.trigger.Schedule;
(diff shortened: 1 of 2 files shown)


scheduler/src/test/java/io/kestra/scheduler/SchedulerInitializeTriggersTest.java
```diff
@@ -0,0 +1,78 @@
...
+        Optional<Trigger> result = AbstractScheduler.computeScheduleInitialization(
+            schedule, locked, RecoverMissedSchedules.NONE, null
+        );
+
+        assertThat(result).isEmpty();
+    }
+
+    // Given an unlocked Schedule trigger whose stored next date is stale,
+    // When re-initialized with recoverMissedSchedules=NONE,
+    // Then the next date is recomputed and the trigger stays unlocked (executionId remains null).
```
ammeek: Should these comments not be in the methods body?

ammeek (approved):
Looks good to me, minor comment about the comment placement in the tests

PR #16716, opened 2026-06-11
refactor(executions): render tab content via router-view (pilot)

elevatebart:
## What & why

This is the **pilot** for moving the tabs system off the custom `TabBody` render function and onto vue-router's `<router-view>`, starting with the **Executions detail** page. Three goals:

### 1. Remove `KsRouterTab`
The design-system `KsRouterTab` component redefined its own tab type, which we want to eliminate. It's deleted (component + story + unit test) and purged from the design-system barrel (`index.ts`): the import, the `export type {RouterTab}`, both component registrations, and the global-components type entry. `KsTabs` / `KsTabPane` remain as the primitives.

`Tabs.vue` now consolidates on the app's existing **`RouteTab`** type (no tab type lives in the design system anymore) and rebuilds the horizontal bar from `KsTabs` / `KsTabPane`, porting the small logic that lived in `KsRouterTab` (active-name sync, the `#tab-label` slot with lock badge / count, routed vs. embedded link modes, disabled styling).

### 2. `<router-view>` for Executions tabs
`routes.ts` splits the single `executions/update` route (`.../:id/:tab?`) into a **parent** (`.../:id`, `ExecutionRoot.vue`) plus one **lazy child record per tab**. Child `path` equals the tab token so **URLs are byte-identical** (`.../:id/logs`, `.../:id/gantt`, …); per-tab flags (`maximized`, `noOverflow`, `locked`, `embed`) and static props (`isReadOnly` for dependencies, `topbar:false` for assets) live on the child record's `meta`/`props`.

A function `redirect` on the parent maps `to.params.tab || localStorage 'executeDefaultTab' || 'overview'` to the matching child, so legacy deep-links (`{name:'executions/update', params:{tab:'logs'}}`, used by e.g. `MetricsTable`, `links.ts`) and bare `/:id` URLs still resolve. A new `useActiveTab()` composable returns the active tab from the matched child route (`route.meta.tab`), falling back to `route.params.tab` — keeping the ~30 existing `route.params.tab` readers working during the roll-out. Full-reload navigations that build hrefs via `resolve().href` (`Restart.vue`, `submitTask.ts`, `LowCodeEditor.vue`) now target child route names so the tab is preserved.

`Tabs.vue` keeps its dynamic `<component :is>` path for the embedded (`embedActiveTab`) and blueprint-modal cases (no URL segment), and for pages not yet migrated. The `<router-view>` wrapper injects **no handlers** — only the section class from route meta and the `embed` flag.

### 3. Tab-body events via Pinia stores
The `@follow` emit chain is collapsed: the real leaf emitters — `ChangeExecutionStatus.vue` and `ChangeStatus.vue` — now call `executionsStore.followExecution()` directly instead of emitting `follow`. The `@follow` relays in the Executions-exclusive tab components (`Overview`, `Banner`, `Gantt`, `Logs`) and the `@follow` prop on the tabs body in `ExecutionRoot.vue` are removed. `useExecutionRoot` keeps `follow()` for its own lifecycle/popstate needs, and the `onMounted` default-tab `router.replace` is dropped (the parent redirect subsumes it).

> Note: the shared `TaskRunActions` / `TaskRunLine` / `TaskRunDetails` / `Topology` components keep their `follow` relays untouched because they're also used by `LowCodeEditor` (flow-execution path), which is out of pilot scope.

## Pilot scope

Only the **Executions** detail page is migrated to `<router-view>`.

## Deferred roll-out (follow-up)

- Flows detail tabs
- Namespaces detail tabs
- `RouteTabsSidebar` (vertical mode)
- admin pages

## Verification

- `npm run check:types` — green
- `npm run test:lint` — green (0 errors; remaining warnings are pre-existing)
- `npm run test:unit` — **884 passed (112 files)**; no dangling `KsRouterTab` import

Reasoned through: each Executions tab loads at its identical URL; legacy `{name:'executions/update', params:{tab}}` pushes land on the right tab via the parent redirect; a state change re-subscribes via the store; dependencies count/disabled still drive the bar.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KB8EH45DDZ7aBBLreySehJ)_


elevatebart:
closed in favor of https://github.com/kestra-io/kestra/pull/17494

PR #16724, opened 2026-06-11
fix(core): use SDK client in MCP store for token

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra-ee/issues/8215


ammeek (approved):
LGTM

Pradumnasaraf (approved):
Looks good, 

Only one small nit-pick is that whenever we create an MCP server, we don't get that toast message, as we get for an existing one.

https://github.com/user-attachments/assets/07e33a93-f043-4833-af93-5c2481a646af

PR #16725, opened 2026-06-11
fix(design-system): respect browser font size user preferences

flcarre:
## Summary

Remove the hardcoded `html { font-size: 14px }` that was preventing the browser's font size preference (set in browser Settings → Appearance) from taking effect. Rewrite the design system font-size SCSS variables as explicit `rem` values relative to the browser's default root (16px), so all sizes scale correctly when a user raises their preference.

**What changes for users:** The UI now respects the browser's font size setting. A user who configured "Large" in their browser will see larger text across the whole app. This fixes WCAG 1.4.4 compliance.

**What doesn't change:** At default browser settings (16px), all rendered sizes are bit-for-bit identical — `0.875rem` = 14px, `0.75rem` = 12px, etc.

**Files changed:**
- `ui/packages/design-system/src/assets/styles/index.scss` — remove `html { font-size: 14px }`
- `ui/packages/design-system/src/assets/styles/variables.scss` — rewrite `$font-size-*` variables from `calc(N/14) * 1rem` (relative to a forced 14px root) to explicit rem fractions relative to the 16px browser default

## Test Coverage

CSS-only change — no logic added. Pixel equivalents at default settings are unchanged.

## Pre-Landing Review

No issues found. Change is a straightforward SCSS accessibility fix.

## Design Review

No component structure or visual hierarchy changed — only the root font-size mechanism. Pixel values at default browser settings are identical.

---

Closes https://github.com/kestra-io/kestra-ee/issues/8536

Ref: https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/


ui/packages/design-system/src/assets/styles/index.scss
```diff
@@ -66,14 +66,11 @@
...
 // global
-html {
-    font-size: 14px;
-}
-
 html, body {
     height: 100%;
     background: var(--ks-bg-base);
     color: var(--ks-text-primary);
+    font-size: var(--ks-font-size-base);
```
elevatebart: This line seems to still be a setting of the font-size on html (root). Wasn't the point of this change to remove this rule ?

flcarre: Im scared we'll have regressions if we dont keep a global definition, ill check this more deeply

Thought a global definition, but in rem would be ok

elevatebart (approved):

PR #16731, opened 2026-06-11
feat(core): add full arg to the secret() pebble function

Malaydewangan09:
Adds an optional `full` argument to `secret()` to get the whole secret in one call, shaped as `{value, metadata}`: `value` is the secret, `metadata` holds any extra fields. Default behaviour is unchanged. Each field is masked in logs.

```
{% set cred = secret('prod-mssql', full=true) %}
{{ cred.value }} / {{ cred.metadata.username }}
```

Part of kestra-io/secret-delinea#15

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/pebble/functions/SecretFunction.java
+++ b/core/src/main/java/io/kestra/core/runners/pebble/functions/SecretFunction.java
@@ -2,6 +2,7 @@ package io.kestra.core.runners.pebble.functions;
 
 import java.io.IOException;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
@@ -13,6 +14,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import io.kestra.core.runners.RunVariables;
 import io.kestra.core.secret.SecretException;
 import io.kestra.core.secret.SecretNotFoundException;
+import io.kestra.core.secret.SecretObject;
 import io.kestra.core.secret.SecretService;
 import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.services.NamespaceService;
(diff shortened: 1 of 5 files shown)


brian-mulier-p (changes requested):
I'm not a fan of the naming there:
"full" doesn't make much sense, I'd prefer "json" (true means we will parse the string as json to extract the object, or "fromJson".

In any case imo we should not introduce such parameter, it's up-to-caller to parse its secret but here we assume the secret is a json object. It's too delicate imo, better to add a "fromJson(secret(...))" on the consumer side.

Malaydewangan09:
We're not changing what secret managers return. `secret('k')` returns the same string as today, nothing is converted to JSON. The parsing only happens when the caller explicitly passes full=true.

`fromJson(secret(...))` only works if the fields are in the string. On Delinea, secret('k') returns only the password, so the username and domain are never there to parse. 

And we can't make Delinea return the full object as JSON either, that would break existing flows expecting the bare password

So `full=true` lets the caller ask for all fields, and the secret manager answers through a new method (kestra-io/kestra-ee#8542). Same mechanism subkey already uses.

Malaydewangan09:
About json: the flag doesn't really parse JSON in the main case. On Delinea the fields come natively, no JSON involved, parsing is only the fallback for managers storing JSON strings (Vault, AWS). 

So json would be misleading there. full just says what you get.

loicmathieu:
Agree with @brian-mulier-p here, we should not put too much functionality inside the secret function but the user should rely on function composition so calling `fromJson(secret(...)).mykey`.

Malaydewangan09:
On Delinea `secret('k')` returns only the password, so for `fromJson(secret(...)).username` to work, the plugin has to return the whole secret as JSON. 

We can't make that the default since it would break existing flows expecting the bare password, so it'd anyway need an opt-in flag on the Delinea config (e.g. fullObject: true) that switches the response to a JSON object.

Do you want to propose this opt-in flag here?

brian-mulier-p (approved):
LGTM

PR #16744, opened 2026-06-12
fix(variables): let KVSTORE-scoped users find and read KVs without NAMESPACE

nkwiatkowski:
## What

OSS side of kestra-io/kestra-ee#7023 for `releases/v1.0.x`. Companion to **kestra-io/kestra-ee#8561**.

On v1.0.x a user with only KV Store permission on a namespace could neither **find** their KVs nor **read** a value:
1. The KV page enumerates namespaces through the NAMESPACE-gated `search`/`autocomplete` endpoints → a KV-only user got an empty list.
2. The value is only reachable through the edit drawer, gated by `KVSTORE.UPDATE` → a READ-only user had no way to read it.

This PR:
- Adds an optional **`resource`** hint to namespace discovery: a field on `ApiAutocomplete` and a `resource` query param on `/namespaces/search`. OSS ignores it; **Enterprise** (companion PR) honors it to authorize discovery through that resource permission.
- The KV page passes `resource=KVSTORE` when listing namespaces (`useNamespaces` iterator) and in the add-KV namespace picker (`NamespaceSelect` gains a `resource` prop, forwarded to autocomplete).
- Adds a **read-only value viewer** (eye icon) for users who can read but not update a KV, fetching the value via the existing READ-authorized `GET /namespaces/{ns}/kv/{key}`. Value formatting extracted into a pure, unit-tested helper.

## Tests

- `kvValueDisplay.spec.ts` (STRING / NUMBER / BOOLEAN / JSON / DATETIME-tz) — 5/5 pass.
- `check:types` clean, eslint clean.

Refs kestra-io/kestra-ee#7023

### DIFF
--- a/ui/src/components/kv/KVTable.vue
+++ b/ui/src/components/kv/KVTable.vue
@@ -70,6 +70,18 @@
             </template>
         </el-table-column>
 
+        <el-table-column v-if="!paneView" column-key="view" class-name="row-action">
+            <template #default="scope">
+                <el-tooltip v-if="!canUpdate(scope.row) && canRead(scope.row)" :content="$t('show')">
+                    <el-button
+                        :icon="Eye"
+                        link
+                        @click="viewKvModal(scope.row.namespace, scope.row.key, scope.row.description)"
+                    />
+                </el-tooltip>
+            </template>
+        </el-table-column>
+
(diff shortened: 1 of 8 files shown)


MilosPaunovic (approved):

loicmathieu (approved):
LGTM

PR #16748, opened 2026-06-12
feat(core): add config flag to disable the UI and run the webserver API-only

### LINKED ISSUE #16747: Add a headless mode to the webserver to disable the UI
### Feature description

Allow running the Kestra webserver in headless mode, serving only the API without the bundled UI.

```yaml
kestra:
  webserver:
    headless: true   # default: false
```

When enabled:
- `/ui/**` endpoints return `404` (UI static assets are not served)
- `GET /` no longer redirects to `/ui/`
- All `/api/v1/**` endpoints (and `/ping`) keep working as usual

This is useful for API-only deployments, e.g. instances running behind a custom frontend or exposed only to programmatic clients, where the bundled UI should not be reachable.

Default behavior (flag absent or `false`) is unchanged.

Ben8t:
## Summary
- Add a `kestra.webserver.ui.enabled` configuration flag (default `true`). Setting it to `false` runs the webserver API-only, with the bundled UI disabled.
- The `/ui/**` static-resources route is turned off purely through configuration via Micronaut's `StaticResourceConfiguration.enabled`: `enabled: ${kestra.webserver.ui.enabled:true}` in `cli/src/main/resources/application.yml`.
- `RedirectController` (the `/` → `/ui/` redirect) and `StaticFilter` (index.html rewriting) are not registered when the flag is `false`, via `@Requires(property = "kestra.webserver.ui.enabled", notEquals = "false", defaultValue = "true")`.
- No new runtime classes — the net main-source diff is +5 lines. Default behavior is unchanged when the flag is absent.

Fixes https://github.com/kestra-io/kestra/issues/16747

> Reworked from the initial approach: the first version added a `kestra.webserver.headless` flag plus a dedicated `HeadlessUiFilter` that 404'd `/ui/**`, because a YAML placeholder can't negate a boolean (`enabled = !headless`). Inverting the polarity to a positively-named `ui.enabled` flag lets the static route be disabled with pure config, so the filter was removed.

## Usage

```yaml
kestra:
  webserver:
    ui:
      enabled: false
```

## Note for reviewers
The `root` static-resources mapping (`/**` → `classpath:root`) stays active when the UI is disabled, so favicon-type assets under `root/` are still served (`GET /favicon.ico` → 200). This is intended — API-only means no UI app, not no static bytes — but flagging it in case you'd prefer it gated too.

## Test plan
- [x] `UiDisabledWebserverTest` (`ui.enabled=false`): `/ui/` and `/` return 404, `/ping` returns 200, `RedirectController` and `StaticFilter` beans absent
- [x] `UiEnabledWebserverTest` (default): `/` redirects 307 → `/ui/`, both beans present
- [x] `./gradlew :webserver:test --tests "*Ui*WebserverTest"` — both classes green
- [x] Live (executable jar, `server local`): with `ui.enabled: false` → `/ui/` 404, `/` 404, `/ping` 200, `/api/v1/main/configs` 401 (API alive), `/favicon.ico` 200; default → `/` 307 → `/ui/`, `/ui/` 200
- [x] Verified on Kestra EE (companion change in `kestra-ee`): EE `RedirectController` (`@Replaces`) and `EEStaticFilter` (`extends StaticFilter`) don't inherit `@Requires`, so they get the same condition; validated live in both modes
image


fhussonnois (approved):
LGTM!

PR #16754, opened 2026-06-13
fix: log Kestra UI URL as last line when server is ready

### LINKED ISSUE #15987: CLI: log Kestra UI URL when the server is ready
### Feature description

When starting Kestra via the CLI, there's no log line telling the user where to open the UI. New users have to guess the URL or check the docs.

We already have what we need to print it: `micronaut.server.host`, the actual bound port from `EmbeddedServer`, and `micronaut.server.context-path` for reverse-proxied setups.

After the server is up, print one line:

```
Kestra is ready at http://localhost:8080
```

Apply to `server standalone`, `server local`, and `webserver`. Skip headless components.

RizwanAhamed13:
Fixes #15987

## What

After the server starts, print one clean, distinctive line at the very end of startup output:

```
✅ Kestra is ready! Open the UI at: http://localhost:8080
```

## Why

The URL was already logged on line 191 via `log.info(...)` but was buried in the middle of other startup logs with a logger prefix, making it easy to miss — especially for new users.

## Changes

- Added a `System.out.println` **after** all startup work completes in `maybeStartWebserver()`
- Used `System.out.println` instead of `log.info` so there's no logger prefix cluttering the line — the terminal shows it clean and auto-linkifies the URL
- Added `\n` before the message so there's a blank line above it, making it visually stand out
- Applies to `server standalone`, `server local`, and `webserver` — headless components are skipped via the existing `ServerCommandInterface` check (unchanged)

## Before

```
15:32:01.123 [main] INFO  io.kestra.cli.AbstractCommand -- Server is running at http://localhost:8080
15:32:01.456 [main] INFO  io.kestra.core... -- (more logs)
```

## After

```
15:32:01.123 [main] INFO  io.kestra.cli.AbstractCommand -- Server is running at http://localhost:8080
15:32:01.456 [main] INFO  io.kestra.core... -- (more logs)

✅ Kestra is ready! Open the UI at: http://localhost:8080
```

### DIFF
--- a/cli/src/main/java/io/kestra/cli/AbstractCommand.java
+++ b/cli/src/main/java/io/kestra/cli/AbstractCommand.java
@@ -62,7 +62,7 @@ public abstract class AbstractCommand extends BaseCommand implements Callable<In
     private io.kestra.core.utils.VersionProvider versionProvider;
 
     @Inject
-    private Optional<EmbeddedServer> embeddedServer;
+    protected Optional<EmbeddedServer> embeddedServer;
 
     @Inject
     private BeanProvider<FlowAutoLoader> flowAutoLoaderService;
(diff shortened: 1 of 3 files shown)


RizwanAhamed13:
Closing in favor of #16755, which guarantees the banner is the actual last line (this PR couldn't beat the post-readiness async log burst). Thanks!

PR #16755, opened 2026-06-13
feat(cli): print UI URL as the last line once the server is fully ready

### LINKED ISSUE #15987: CLI: log Kestra UI URL when the server is ready

RizwanAhamed13:
Closes #15987

### What & why
On startup it's hard to spot the UI URL — it's logged in the middle of dozens of async service logs with a logger prefix. This adds a distinctive, prefix-free banner printed once the server is fully ready:

```

✅ Kestra is ready! Open the UI at: http://localhost:8080

```

- Blank line above/below so it stands out
- Plain `stdout` (no `io.kestra.cli...` logger prefix) so the URL is auto-linkified by the terminal
- Applied to `server standalone`, `server local`, and `server webserver`
- Removed the duplicate URL log from `AbstractCommand.maybeStartWebserver()`


cli/src/main/java/io/kestra/cli/ReadinessBanner.java
loicmathieu, line 1: This is too complicated and would add races.
Anyway, you cannot guarantee that no other logs pops up (for ex if there are scheduled triggers).

I would remove this class.

cli/src/main/java/io/kestra/cli/commands/servers/StandAloneCommand.java
```diff
@@ -137,12 +138,17 @@ public Integer call() throws Exception {
                 standAloneRunner.setIndexerEnabled(false);
             }
 
+            // run() blocks until every service (executor, worker, scheduler, indexer, controller)
+            // confirms RUNNING state, so the server is fully ready once it returns.
```
loicmathieu: Remove this comment, it's useless

loicmathieu (changes requested):
We already have a log line with the URL, it should be removed to avoid duplicated logs

RizwanAhamed13:
Done! Addressed all feedback:

Removed ReadinessBanner entirely — replaced with a simple System.out.println
Removed the duplicate URL log from AbstractCommand.maybeStartWebserver()
Removed the useless comment in StandAloneCommand
Thanks for the clear direction.

loicmathieu (approved):
LGTM

PR #16760, opened 2026-06-15
refactor: translate more in CI than the last commit

elevatebart:
Also, I migrated the translation script to TS

Closes https://github.com/kestra-io/kestra-ee/issues/8631.


MilosPaunovic (approved):
This all seems quite alright! 🚀 

Maybe we should cover the EE counterpart at the same time?

I know I should've created the issues for this, it's still on the TODO list for either later today or sometime tomorrow.

PR #16761, opened 2026-06-15
feat(helm): delegate helm release to kestra-io/actions@main

MarcoSaba:
## Summary

- Removes `helm-release.yml` — it is no longer needed
- Updates `release-docker.yml` to call `kestra-io/actions/.github/workflows/kestra-oss-helm-release.yml@main` directly

## Why

`release-docker.yml` previously called `helm-release.yml` via `uses: ./.github/workflows/helm-release.yml`. GitHub resolves that path at the **caller's ref** (e.g. `refs/tags/v1.3.22`). Release branches carried an older version of the workflow that skipped the `helm-docs` regeneration step after bumping `Chart.yaml`. Every published chart tarball therefore shipped a `README.md` with stale version badges (e.g. `kestra-1.3.22.tgz` showed `Version: 1.0.53`), which ArtifactHub rendered verbatim.

Routing through `@main` in `kestra-io/actions` ensures the workflow code is always current — no more stale READMEs in published charts.

## Manual helm-only releases

These are now triggered via `workflow_dispatch` directly on `kestra-io/actions/.github/workflows/kestra-oss-helm-release.yml` from the actions repo's Actions tab.

Related: kestra-io/actions#128 / kestra-io/actions#129


Skraye (approved):
LGTM

PR #16777, opened 2026-06-15
fix: propagate forced execution to flowable children

zgxme:
### ✨ Description

This PR propagates `forceExecution` from a flowable parent task run to the child task runs it resolves.

`afterExecution` task runs are marked with `forceExecution=true` so they can run after a killed execution. When an `afterExecution` task is a flowable wrapper such as `If`, only the wrapper previously received that flag. Its resolved child tasks could still be blocked by the worker killed-execution guard, leaving the flowable wrapper unable to settle and preventing following `afterExecution` tasks from running.

This change keeps the existing behavior for normal flowable tasks and only propagates the flag when the parent task run was already forced.

### 🔗 Related Issue

No issue.

### 🛠️ Backend Checklist

### 📝 Additional Notes

Added coverage for:

- flowable tasks in `afterExecution`
- killed executions where an `afterExecution` `If` task resolves child tasks and a following `afterExecution` task must still run

Local verification:

- `git diff --check`
- Attempted `./gradlew :jdbc-h2:test --tests io.kestra.runner.h2.H2RunnerTest.shouldCallFlowableTasksAfterExecution`, but local Gradle configuration failed before test execution because this machine only has JDK 24 and the project requires a Java 25 toolchain.

### 🤖 AI Authors

Why did the cat review the PR? To make sure every purr-mission was inherited correctly.

### DIFF
--- a/core/src/test/java/io/kestra/core/runners/AbstractRunnerTest.java
+++ b/core/src/test/java/io/kestra/core/runners/AbstractRunnerTest.java
@@ -713,6 +713,12 @@ public abstract class AbstractRunnerTest {
     }
 
     @Test
+    @ExecuteFlow("flows/valids/after-execution-flowable.yaml")
+    public void shouldCallFlowableTasksAfterExecution(Execution execution) {
+        afterExecutionTestCase.shouldCallFlowableTasksAfterExecution(execution);
+    }
+
+    @Test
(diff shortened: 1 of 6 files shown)


loicmathieu (approved):
LGTM, thanks

PR #16786, opened 2026-06-15
fix(worker): commit worker job running entries before dispatching to workers

fhussonnois:
The worker-controller dispatches jobs from inside the JDBC dispatch-queue poll transaction, and the dispatch-time save of the worker_job_running entry silently joined that thread-bound transaction, which only commits after the whole polled batch is processed. A fast task could complete within that window: its terminal result, processed on another connection, deleted nothing (the entry was not visible yet) and the entry leaked forever once the batch committed — to be wrongly resubmitted as a zombie job the next time its worker disconnected.

Saving through a dedicated connection acquired directly from the pool restores the intended ordering: the entry is durable and visible before the job is sent, so the terminal result's delete always finds it.

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/JooqDSLContextWrapper.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/JooqDSLContextWrapper.java
@@ -1,18 +1,25 @@
 package io.kestra.jdbc;
 
+import java.sql.Connection;
 import java.sql.SQLException;
 import java.time.Duration;
 import java.util.function.Predicate;
 
 import javax.sql.DataSource;
 
+import org.jooq.ConnectionProvider;
 import org.jooq.DSLContext;
 import org.jooq.TransactionalCallable;
 import org.jooq.TransactionalRunnable;
+import org.jooq.exception.DataAccessException;
+import org.jooq.impl.DSL;
+import org.jooq.impl.DefaultConnectionProvider;
+import org.jooq.impl.DefaultTransactionProvider;
 
 import io.kestra.core.models.tasks.retrys.Random;
 import io.kestra.core.utils.RetryUtils;
 
+import io.micronaut.data.connection.jdbc.advice.DelegatingDataSource;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 
@@ -28,15 +35,20 @@ public class JooqDSLContextWrapper {
     private static final DeadlockPredicate DEADLOCK_PREDICATE = new DeadlockPredicate();
 
     private final DSLContext dslContext;
+    private final DataSource rawDataSource;
 
     /**
-     * @param dataSource explicit dependency to ensure Micronaut destroys this bean before the DataSource.
+     * @param dataSource used by {@link #requireNewTransaction(TransactionalRunnable)}, and an
+     *        explicit dependency to ensure Micronaut destroys this bean before the DataSource.
      *        Without it, the @EachBean-derived DSLContext/Configuration may be destroyed
      *        together with the DataSource, leaving this wrapper with a stale DSLContext.
      */
     @Inject
     public JooqDSLContextWrapper(DSLContext dslContext, DataSource dataSource) {
         this.dslContext = dslContext;
+        // Unwrap any Micronaut Data AOP proxy: the wrapped DataSource hands back the current
+        // thread's transaction-bound connection instead of a new one.
+        this.rawDataSource = DelegatingDataSource.unwrapDataSource(dataSource);
     }
 
     private static <T> RetryUtils.Instance<T, RuntimeException> retryer() {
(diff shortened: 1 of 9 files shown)


jdbc/src/main/java/io/kestra/jdbc/runner/AbstractJdbcWorkerJobRunningStateStore.java
```diff
@@ -30,7 +30,13 @@ public WorkerJobRunning save(TransactionContext txContext, WorkerJobRunning work
             var dslContext = txContext.unwrap(JdbcTransactionContext.class).getDslContext();
             this.jdbcRepository.persist(workerJobRunning, dslContext, this.jdbcRepository.persistFields(workerJobRunning));
         } else {
-            this.jdbcRepository.persist(workerJobRunning);
```
loicmathieu, line 33: I'm a bit surprised as for me, if you didn't use the same dslContext, you are in a separate transaction on the same connection which should not have the issue your raised.

Using two connections and not one is problematic as it uses more resources

jdbc/src/main/java/io/kestra/jdbc/JooqDSLContextWrapper.java
```diff
@@ -61,6 +73,39 @@ public <T> T transactionResult(TransactionalCallable<T> transactional) {
...
+     * other connections — even when a transaction is already open on the current thread.
+     * <p>
+     * Regular {@link #transaction(TransactionalRunnable)} calls are bound to the calling thread
+     * and silently join a caller-owned transaction (e.g. the dispatch-queue poll transaction),
+     * deferring their writes until that transaction commits.
+     * <p>
+     * When the calling thread already has a connection checked out, this briefly holds a second
+     * one — keep the work short.
+     */
+    public void transactionOnNewConnection(TransactionalRunnable transactional) {
```
loicmathieu: Instead of using a new connection, it would be better to start a new transaction in the same connection.

jdbc/src/main/java/io/kestra/jdbc/runner/AbstractJdbcWorkerJobRunningStateStore.java
fhussonnois: JDBC connection has exactly one transaction at a time. `DSLContext`  is just a facade over a `Configuration` which is actually tied to a `ConnectionProvider` + `TransactionProvider` (i.e., `MicronautTransactionProvider`. 

Creating a new `DSLContext` inherits the `MicronautTransactionProvider` which reuse the same connection when a transaction is already active on the thread. So we just rejoins the poll transaction, and any nested transaction made inside the poll is only visible once the outer-transaction is commited. 

Micronaut code.:
* https://github.com/micronaut-projects/micronaut-sql/blob/991abff48d922cf81ebf403c8b3bdc5e997a5424/jooq/src/main/java/io/micronaut/configuration/jooq/MicronautTransactionProvider.java#L60
* https://github.com/micronaut-projects/micronaut-data/blob/v4.14.4/data-tx/src/main/java/io/micronaut/transaction/support/AbstractTransactionOperations.java#L491

Most of the time, creating a new transaction means getting a new connection from the underlying HikariDataSource (it would be the same if where using MicronautData with `Transactional(propagation=REQUIRE_NEW)`). 

So regarding the cost itself, we don't create a new physical connection. `unwrapDataSource(...) ` returns the `HikariDataSource`, so it's a borrow from the pool for a single-row upsert whcih returned immediately.

loicmathieu: OK, using REQUIRE_NEW would have been a better option as it would have suspended the current running transaction (not sure what it does actually), it would also reuse the same context propagation lowering the cost.

But we can live with it as is.

jdbc/src/main/java/io/kestra/jdbc/runner/AbstractJdbcWorkerJobRunningStateStore.java
```diff
@@ -30,7 +30,13 @@ public WorkerJobRunning save(TransactionContext txContext, WorkerJobRunning work
...
+            // Commit on a dedicated connection: the entry must be visible to concurrent result
+            // processing as soon as the job is sent to the worker, even when the caller holds an
+            // open thread-bound transaction (the dispatch-queue poll transaction) — otherwise the
+            // terminal result of a fast task deletes nothing and the entry leaks forever.
+            this.jdbcRepository.getDslContextWrapper().transactionOnNewConnection(configuration ->
```
loicmathieu:
```suggestion
            this.jdbcRepository.getDslContextWrapper().requireNewTransaction(configuration ->
```

So we match the standard naming and if later we can hook inside the Micronaut transaction manager to reuse it we could have a cleaner approach

loicmathieu (approved):
LGTM

PR #16790, opened 2026-06-15
fix(dashboards): "All" quick-filter tab shows every execution state

### LINKED ISSUE #16812: Dashboard "Executions" quick-filter: the "All" tab is empty while state tabs show executions
## Description

On the dashboard executions table widget (the quick-filter tab bar **All / Running / Paused / Success / Warning / Failed**), the **All** tab can show *"There are no results to be shown"* while a state tab such as **Success** shows plenty of executions. Seen on deployed 2.0 / develop environments.

## Steps to reproduce

1. Use a dashboard whose executions `Table` chart carries a `where: STATE IN [...]` constraint — e.g. the "Executions In Progress" widget, or any custom / DB-stored dashboard scoped by state.
2. Have executions whose state is **not** in that `where` set (e.g. only `SUCCESS` executions, none in progress).
3. Open the dashboard. The **All** tab is active by default.

## Actual

- **All** → "There are no results to be shown."
- **Success** → the executions appear.

(Reproduced live: 3 `SUCCESS` executions, 0 in progress → "All" empty, "Success" lists all 3.)

## Expected

The **All** tab shows at least everything the individual state tabs can show (it should not be silently narrower than them).

## Root cause

The quick filter auto-enables on any executions `Table` with a `STATE` column. The backend merges the runtime global filter into the chart's static `where` by **per-field replacement** (`IExecutions.whereWithGlobalFilters`): a global `STATE` filter *removes* the chart's `STATE` `where`. The per-state tabs send a `STATE` filter (so they escape the chart-level `where`), but the **All** tab sent no filter, so it stayed bound by the chart's `where` — empty whenever no execution matched that constraint.

## Fix

PR #16790 — the **All** tab now emits the union of all tab states when the chart constrains `STATE` (overriding the `where` through the existing replacement semantics), and seeds it on initial load (not only on click).

flcarre:
Closes #16812

## Problem

On the dashboard Executions table, clicking the **All** quick-filter tab shows nothing, while clicking a state tab (e.g. **Success**) shows plenty of executions. Reproduced on deployed environments.

## Cause

The quick-filter tabs auto-enable on any executions table with a `STATE` column. The backend merges the runtime global filter into the chart's static `where` by **per-field replacement** (`IExecutions.whereWithGlobalFilters`): a global `STATE` filter *removes* the chart's `STATE` `where`.

So the per-state tabs (Success / Failed / …) escape any chart-level `STATE` constraint, while the **All** tab — which sent no state filter — stayed bound by it. On any dashboard whose chart constrains `STATE` via `where` (custom / DB-stored dashboards, or the pre-#16616 "Executions In Progress" widget), **All** only showed those states and was empty whenever none were present, even though the per-state tabs were full.

#16616 only removed the `where` from the bundled default dashboard; this fixes the underlying feature for every dashboard.

## Fix

`TableQuickFilter.vue`: the **All** tab now emits the union of every tab's states **when the chart constrains `STATE`**, overriding that `where` through the existing replacement semantics — so **All** always shows at least what any individual tab can. Charts without a `STATE` `where` still send no filter (truly all states, `RESUBMITTED` included), so nothing changes for the default dashboards.

## Tests

Added `tableQuickFilter.spec.ts`:
- no `STATE` where → All emits no filter
- `STATE` where → All emits the full state set
- a state tab emits its own states

`check:types` and ESLint clean.

### DIFF
--- a/ui/src/components/dashboard/sections/Table.vue
+++ b/ui/src/components/dashboard/sections/Table.vue
@@ -60,6 +60,7 @@
     import {isPaginationEnabled, useChartGenerator} from "../composables/useDashboards"
     import {FilterObject} from "../../../utils/filters"
     import TableQuickFilter from "./TableQuickFilter.vue"
+    import {stateFilterForTab} from "./quickFilters"
     import Date from "./table/columns/Date.vue"
     import Duration from "./table/columns/Duration.vue"
     import Link from "./table/columns/Link.vue"
@@ -124,7 +125,7 @@
 
     const data = ref()
     const activeTab = ref("all")
-    const stateFilter = ref<FilterObject | null>(null)
+    const stateFilter = ref<FilterObject | null>(stateFilterForTab(props.chart, "all"))
     const pageNumber = ref(1)
     const pageSize = ref(25)
 
--- a/ui/src/components/dashboard/sections/TableQuickFilter.vue
+++ b/ui/src/components/dashboard/sections/TableQuickFilter.vue
@@ -64,6 +64,8 @@
         },
     ] as const
 
+    const ALL_STATES = QUICK_FILTER_TABS.flatMap((tab) => tab.states)
+
     type TabKey = (typeof QUICK_FILTER_TABS)[number]["key"]
 
     const props = defineProps<{chart: Chart}>()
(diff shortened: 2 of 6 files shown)


flcarre:
Self-review (`/review`) + independent adversarial pass caught a gap in the first commit and it's fixed in 23662e2b8:

- The override only applied on tab **click**; the **initial render** of a STATE-constrained chart still fetched with no override, so "All" was empty on first paint (the exact reported symptom). Now the tab→filter logic lives in `quickFilters.ts` and `Table.vue` seeds its initial filter from it — single fetch, correct on load.

Tests: `quickFilters.spec.ts` (logic, incl. an "All ⊇ every tab" invariant check), `tableInitialFilter.spec.ts` (guards the initial-load wiring), `tableQuickFilter.spec.ts` (click emits). `check:types` + ESLint clean.

Known low-impact limitations (not addressed here): a STATE condition nested inside an `Or` composite `where` isn't detected (the backend's replacement can't strip it either — pre-existing); `RESUBMITTED` is a TaskRun-attempt state that doesn't surface as an execution state, so it's intentionally absent from the tab set.

MilosPaunovic (approved):

PR #16802, opened 2026-06-15
feat(executions): add UI support for execution breakpoints

### LINKED ISSUE #15583: Execution debugging
### Describe the issue

Since the playground functionality, we have the possibility in the backend to suspend an execution at a breakpoint, and resume it later.

We made this so it is not tied to the playground functionality at all so we can later bring this functionality on itself inside the UI.

The following changes describes UI only changes to support that.

## Start and execution with a breakpoint

When starting an execution, inside the `Additional properties` options, we can add a breakpoint (or a list of breakpoints) so that, when the execution arrives at this task, it will suspend.

Then, when an execution is suspended at a breakpoint, allow resuming it, optionally setting new breakpoints.

This would be a kind of light debugger for executions.

## Add breakpoints to a running execution

Optionaly, we can also allow adding breakpoints to a running execution, this can be useful for operational teams that would stop an execution but not kill it: add a breakpoint to the next task.

This would be racy as the execution could start the task and the breakpoint ignored so we should add a warning about it.

### Environment

- Kestra Version: develop

chaitanyakatore:
### ✨ Description

This PR adds UI support for setting and resuming from execution breakpoints.

Key additions:
* **Trigger/Replay Dropdown:** A "Breakpoints" multi-select dropdown in the execution panel (under "Other properties") that displays all task IDs.
* **Resume Button & Modal:** A "Resume" button shown on executions in the `BREAKPOINT` state, opening a modal to configure breakpoints before resuming.
* **State Update:** Updated design system state rules to treat `BREAKPOINT` as an active/killable state.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/15583

### 🎨 Frontend Checklist

https://github.com/user-attachments/assets/b0c7a5bb-03d6-4dc7-8132-c80c978dbc06


ui/src/components/executions/overview/components/actions/ResumeFromBreakpoint.vue
```diff
@@ -0,0 +1,111 @@
...
+            })
+            if (props.execution.breakpoints) {
+                selectedBreakpoints.value = props.execution.breakpoints.map((b: any) => b.value ? `${b.id}.${b.value}` : b.id)
+            }
+        }
+    })
+</script>
+
+<style scoped>
+.w-100 {
```
elevatebart: Meh! It looks too much like a utility class!

elevatebart (approved):
I would have loved a storybook test

ui/src/components/executions/overview/components/actions/ResumeFromBreakpoint.vue
chaitanyakatore: @elevatebart I have updated `w-100` to `.breakpoints-select` could you please review

PR #16804, opened 2026-06-16
feat(logs): filter to an exact level from the logs quick-filter

### LINKED ISSUE #16813: Logs: filter by an exact level, not just "this level and above"
## Context

[#16306](https://github.com/kestra-io/kestra/issues/16306) surfaced the **log level** quick-filter on the Logs views. As specified there, it reused the **existing single-select min-level filter** (a severity *threshold*) and was explicitly "no new filtering capability, no backend change".

So today, clicking the `DEBUG` level button applies `level >= DEBUG` and the view shows **DEBUG + INFO + WARN + ERROR**.

## Problem

For a one-click level selector that's not what users expect. Clicking `DEBUG` should show **only DEBUG** logs. The "and above" threshold is the right default for the *dropdown*, but the quick-filter buttons read as "show me this level".

This requirement was never written down as a clear ticket — #16306 scoped the opposite (threshold only, no backend), and a backend test (`DashboardDataGlobalFiltersTest.shouldRejectUnsupportedLevelOperationForLogs`) actively asserted that an exact (`EQUALS`) level filter is **rejected**. This issue specifies the intended behavior.

## Requirement

- **Quick-filter level buttons → exact level.** Clicking a level filters to **exactly** that level. Clicking another switches to it (single-select).
- **"Add filters" dropdown keeps the thresholds** (`At or Above` / `At or Below`) **and gains an `Equals`** operator, so a level can also be picked exactly from the dropdown.
- The exact chip renders as a bare value (e.g. `Log Level DEBUG`), no operator word; threshold chips keep `At or Above` / `At or Below`.

## Backend change (new capability — out of scope of #16306)

The `LEVEL` query-filter field previously supported only `GREATER_THAN_OR_EQUAL_TO` / `LESS_THAN_OR_EQUAL_TO`. Supporting exact selection requires adding `EQUALS` (exact match) to that field across every consumer (JDBC repo, dashboard `ILogs`, the follow-log realtime matcher, and the EE Elasticsearch repo). This **reverses** the prior constraint that `EQUALS` on level is rejected.

## Scope

- **In scope / delivered:** global **Logs** page (and embedded flow/namespace tabs). Implemented in #16804.
- **Follow-ups:**
  - **EE Elasticsearch** needs the parallel `EQUALS` case in its log repository.
  - **Per-execution Logs tab** uses a threshold-only (`minLevel`) endpoint, so exact-level there is a separate change (best done client-side).

Implemented by #16804.

flcarre:
Closes #16813

## What

On the **Logs** page, clicking a level button under the chart (e.g. `DEBUG`) now shows **only that level**, instead of "that level and above". The chip reads `Log Level DEBUG`. The **Add filters** dropdown keeps the `At or Above` / `At or Below` thresholds and gains an `Equals` option.

Before, clicking `DEBUG` applied a `>= DEBUG` threshold, so the result still included INFO/WARN/ERROR — confusing for a one-click level selector.

## How

Two layers:

**Backend** — the `LEVEL` query-filter field only supported `GREATER_THAN_OR_EQUAL_TO` / `LESS_THAN_OR_EQUAL_TO`. Added `EQUALS` (exact match):
- `QueryFilter.Field.LEVEL.supportedOp()` includes `EQUALS`
- `AbstractJdbcRepository.handleLevelField` and `ILogs` (dashboard) translate `EQUALS` to a single-level `IN` condition

**Frontend** — the level quick-filter buttons and the level chip now use an exact `filters[level][EQUALS]=X`:
- `logLevelQuery`: new `"exact"` direction (read/normalize/request all map to the single `EQUALS` key)
- `FilterKeyConfig.exactEquals` renders an `EQUALS` chip as a bare value (no comparator word), so the chip reads `Log Level DEBUG` while the dropdown keeps the thresholds + `Equals`
- `LogsWrapper.selectLevel` writes `EQUALS` and rebuilds the query preserving key order, so the level chip no longer jumps position

## Testing

- **Backend:** `QueryFilterTest` (EQUALS now valid on LEVEL) + `H2LogRepositoryTest` (new case: `level = INFO` returns only INFO) — green.
- **Frontend:** `logLevelQuery`, `routeDecoderExactEquals`, `FilterChip` unit tests — green. Lint + typecheck clean.
- **Manual QA** (local H2): clicking `DEBUG` → only DEBUG rows + `200` (no 422); dropdown `At or Above WARN` → WARN+ERROR; operator dropdown shows `Equals / At or Above / At or Below`.

## Follow-ups (not in this PR)

- **Enterprise / Elasticsearch:** the same `EQUALS` case must be added to `AbstractElasticSearchRepository.handleLevelFilter` in `kestra-ee` (2 lines). EE + Postgres is already covered by the JDBC change here.
- **Execution logs tab:** its logs endpoint is `minLevel`-only (threshold), so exact-level there is a separate change (best done client-side); deferred.

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/services/SearchableFactory.java
+++ b/webserver/src/main/java/io/kestra/webserver/services/SearchableFactory.java
@@ -133,6 +133,8 @@ public class SearchableFactory {
                 (event, v) -> event.level() != null && event.level().toInt() >= toLevel(v).toInt())
             .searchableQueryFilterExtractor(QueryFilter.Field.LEVEL, QueryFilter.Op.LESS_THAN_OR_EQUAL_TO,
                 (event, v) -> event.level() != null && event.level().toInt() <= toLevel(v).toInt())
+            .searchableQueryFilterExtractor(QueryFilter.Field.LEVEL, QueryFilter.Op.EQUALS,
+                (event, v) -> event.level() != null && event.level().toInt() == toLevel(v).toInt())
 
             .searchableQueryFilterExtractor(QueryFilter.Field.EXECUTION_ID, FollowLogEvent::executionId,
                 QueryFilter.Op.EQUALS, QueryFilter.Op.NOT_EQUALS, QueryFilter.Op.CONTAINS,
--- a/core/src/main/java/io/kestra/core/models/QueryFilter.java
+++ b/core/src/main/java/io/kestra/core/models/QueryFilter.java
@@ -357,7 +357,7 @@ public record QueryFilter(
         LEVEL("level") {
             @Override
             public List<Op> supportedOp() {
-                return List.of(Op.GREATER_THAN_OR_EQUAL_TO, Op.LESS_THAN_OR_EQUAL_TO);
+                return List.of(Op.EQUALS, Op.GREATER_THAN_OR_EQUAL_TO, Op.LESS_THAN_OR_EQUAL_TO);
             }
         },
         PATH("path") {
--- a/core/src/main/java/io/kestra/plugin/core/dashboard/data/ILogs.java
+++ b/core/src/main/java/io/kestra/plugin/core/dashboard/data/ILogs.java
@@ -45,6 +45,7 @@ public interface ILogs extends IData<ILogs.Fields> {
                 {
                     Level level = f.value() instanceof Level l ? l : Level.valueOf((String) f.value());
                     List<Level> levels = switch (f.operation()) {
+                        case EQUALS -> List.of(level);
                         case GREATER_THAN_OR_EQUAL_TO -> LogEntry.findLevelsByMin(level);
                         case LESS_THAN_OR_EQUAL_TO -> LogEntry.findLevelsByMax(level);
                         default -> throw new IllegalArgumentException("Unsupported operation for LEVEL: " + f.operation());
--- a/core/src/test/java/io/kestra/core/models/QueryFilterTest.java
+++ b/core/src/test/java/io/kestra/core/models/QueryFilterTest.java
@@ -224,6 +224,7 @@ public class QueryFilterTest {
             buildQueryFiltersForOperations(
                 Field.LEVEL, Resource.LOG,
                 Set.of(
+                    Op.EQUALS,
                     Op.GREATER_THAN_OR_EQUAL_TO,
                     Op.LESS_THAN_OR_EQUAL_TO
                 )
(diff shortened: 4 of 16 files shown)


loicmathieu:
Who decides to change that behavior?

This is pretty standard and what a user might expect: when you filter logs on a level, it's a minimum level, as levels above are usually even more important than the one at the level you select.

flcarre:
> Who decides to change that behavior?
> 
> This is pretty standard and what a user might expect: when you filter logs on a level, it's a minimum level, as levels above are usually even more important than the one at the level you select.

Its only for those buttons 
image

In the new log page 
image

We don't change the behavior of filters elsewhere

It's checked with @Ben8t and @Nico-Kestra

webserver/src/main/java/io/kestra/webserver/services/SearchableFactory.java
```diff
@@ -132,6 +132,8 @@ public Searchable<FollowLogEvent> followLogEventSearchable(SystemFlowsConfigurat
                 (event, v) -> event.level() != null && event.level().toInt() >= toLevel(v).toInt())
             .searchableQueryFilterExtractor(QueryFilter.Field.LEVEL, QueryFilter.Op.LESS_THAN_OR_EQUAL_TO,
                 (event, v) -> event.level() != null && event.level().toInt() <= toLevel(v).toInt())
+            .searchableQueryFilterExtractor(QueryFilter.Field.LEVEL, QueryFilter.Op.EQUALS,
+                (event, v) -> event.level() != null && event.level().toInt() == toLevel(v).toInt())
```
flcarre, line 137: 🟠 **[Tests] Missing `LEVEL + EQUALS` coverage for the follow-log matcher**

This new exact-level predicate is the realtime (follow-log) counterpart of the JDBC/dashboard `EQUALS` change, but it has no unit coverage. `FollowLogEventSearchableTest` is the dedicated test for this in-memory matcher (the repository path is covered separately by `AbstractLogRepositoryTest`), and it only exercises `GREATER_THAN_OR_EQUAL_TO` / `LESS_THAN_OR_EQUAL_TO` for `LEVEL` (lines 176-179). If this predicate were dropped or its comparison inverted, no test would fail.

`baseEvent.level` is `INFO`, so add a match + non-match case after line 179 of `FollowLogEventSearchableTest.java`:

```java
new PredicateCase(baseEvent, filter(Field.LEVEL, Op.EQUALS, Level.INFO), true),
new PredicateCase(baseEvent, filter(Field.LEVEL, Op.EQUALS, Level.ERROR), false),
```

flcarre:
Close by https://github.com/kestra-io/kestra/pull/17070

PR #16816, opened 2026-06-16
fix(core): keep JDBC queue consumer alive on transient DB errors

### LINKED ISSUE #16794: JDBC queue consumer thread dies on transient DB errors, causing unbounded queues table growth
## Summary

A transient database error during a JDBC queue poll permanently kills the consumer thread, so messages stop being consumed and accumulate unbounded in the `queues` table.

## Root cause

In `JdbcQueue.poll(Supplier<Integer>)` (`jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java`), the polling loop only catches `CannotCreateTransactionException`:

```java
try {
    Integer count = runnable.get();
    ...
} catch (CannotCreateTransactionException e) {
    if (log.isDebugEnabled()) {
        log.debug("Can't poll on receive", e);
    }
}
```

Any **other** throwable from `runnable.get()` — e.g. a MySQL lock-wait-timeout `DataAccessException`, or a `RetryUtils$RetryFailed` from the transaction retryer — escapes the `while` loop. The `poolExecutor` task then returns and **the consumer thread is gone for good** (no supervision/restart; `running` stays `true`). With no consumer, messages are never marked consumed (`consumers` stays `NULL`) and the queue grows without bound — "no consumer, no clean".

A related issue: `JooqDSLContextWrapper.predicate()` only retries deadlocks (SQLState `40001` / `40P01`). A MySQL lock-wait-timeout (`ER_LOCK_WAIT_TIMEOUT`, vendor code `1205`) is not retried, so a producing thread (e.g. a worker emitting a result) crashes on the first transient lock contention instead of retrying.

## Impact (observed on a high-throughput MySQL deployment)

- `queues` table grew to tens of millions of rows / tens of GB with `consumers = NULL`.
- The scheduler's `SELECT ... FOR UPDATE SKIP LOCKED` locked millions of rows per poll, starving worker `INSERT`s, which hit `innodb_lock_wait_timeout` and crashed workers in a loop.

## Fix

1. `JdbcQueue.poll()` — add a broad `catch (Exception e)` after the existing catch so a transient error logs and backs off but never kills the polling thread.
2. `JooqDSLContextWrapper.predicate()` — also retry MySQL `ER_LOCK_WAIT_TIMEOUT` (vendor code `1205`).

## Notes

- First targeting `releases/v1.2.x` (where this was reported); backports to other maintained branches to follow.
- The Queue V2 architecture on `develop` removes the `consumers` column / delete-on-consume, but `AbstractPollingSubscriber` there also terminates the poller on non-transaction exceptions — a similar resilience hardening is worth a follow-up.

nkwiatkowski:
Backport of #16814 to `releases/v1.3.x`.

## What

A transient database error during a JDBC queue poll permanently kills the consumer thread, so messages stop being consumed and accumulate unbounded in the `queues` table. The two defects are present verbatim on this branch.

Closes #16794.

## Changes

- **`JdbcQueue.poll()`** — add a broad `catch (Exception e)` after the existing `CannotCreateTransactionException` catch: log at `error`, back off to `maxPollInterval`, and keep polling. A transient DB error (lock-wait-timeout `DataAccessException`, `RetryUtils$RetryFailed`) can no longer terminate the consumer.
- **`JooqDSLContextWrapper.predicate()`** — also retry MySQL `ER_LOCK_WAIT_TIMEOUT` (vendor code `1205`), so a producing thread retries transient lock contention instead of crashing.
- **`JdbcQueueTest.shouldKeepConsumingAfterConsumerThrows`** — regression test (H2/MySQL/Postgres): a consumer that throws must not stop later messages from being delivered.

## Testing

`./gradlew :jdbc-h2:test --tests "io.kestra.runner.h2.H2QueueTest"` — 5/5 pass, including the new test.

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/JooqDSLContextWrapper.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/JooqDSLContextWrapper.java
@@ -47,7 +47,10 @@ public class JooqDSLContextWrapper {
             // standard deadlock
             cause.getSQLState().equals("40001") ||
             // postgres deadlock
-                cause.getSQLState().equals("40P01");
+                cause.getSQLState().equals("40P01") ||
+            // MySQL lock wait timeout (ER_LOCK_WAIT_TIMEOUT), so a transient lock contention is retried
+            // instead of bubbling up and crashing the producing thread (e.g. a worker emitting a result).
+                cause.getErrorCode() == 1205;
         };
     }
 
--- a/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueue.java
@@ -491,6 +491,12 @@ public abstract class JdbcQueue<T> implements QueueInterface<T> {
                         if (log.isDebugEnabled()) {
                             log.debug("Can't poll on receive", e);
                         }
+                    } catch (Exception e) {
+                        // Any other error (e.g. a lock-wait-timeout escaping the transaction retryer) must NOT
+                        // kill the polling thread: if it dies, the consumer is silently lost and messages pile up
+                        // forever with consumers = NULL (no consumer, no clean). Log it, back off, and keep polling.
+                        log.error("Unexpected error while polling the '{}' queue, retrying after backoff", queueType(), e);
+                        sleep = configuration.maxPollInterval;
                     }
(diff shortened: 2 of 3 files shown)


loicmathieu (approved):
LGTM, note that now the PR description is no longuer accurate, may be good to update it before merging

PR #16831, opened 2026-06-16
Implemented different Parameter testing on different search endpoints through repositories

### LINKED ISSUE #8533: Add code coverage in repositories and search
Most of the search method use a list of parameters, but only one parameter is used during unit testing. We then don't know which parameter are really used, and if they work.

It will allow us to find a lot of bugs before it's too late.

rishabhrawat05:
### ✨ Description

Adds test methods that verify different parameter combinations across repository search methods to improve test coverage and ensure consistent search behavior.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/8533

### 🛠️ Backend Checklist

### 📝 Additional Notes

In AbstractMetricRepositoryTest, the `all()` test previously expected metrics to be returned for `endDate + 1`. However, the implementation of `aggregateByFlowId()` iterates only up to `endDate` (exclusive of dates beyond it), so the expected count has been adjusted to match the actual behavior of the method.

### DIFF
--- a/core/src/test/java/io/kestra/core/repositories/AbstractFlowTopologyRepositoryTest.java
+++ b/core/src/test/java/io/kestra/core/repositories/AbstractFlowTopologyRepositoryTest.java
@@ -111,4 +111,37 @@ public abstract class AbstractFlowTopologyRepositoryTest {
 
         assertThat(list.size()).isEqualTo(3);
     }
+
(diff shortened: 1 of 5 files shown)


loicmathieu (approved):
LGTM, thanks

PR #16832, opened 2026-06-16
feat(settings): add default triggers tab configuration option

RemiBarthe:
This pull request adds support for configuring the default tab shown on the Triggers page. Users can now select their preferred default tab ("add" or "manage") via the settings, and this preference is persisted and respected across sessions. The implementation includes updates to the settings UI, local storage handling, and translations.

Settings UI and logic:

* Added a new setting in `BasicSettings.vue` that allows users to choose the default Triggers tab (`add` or `manage`). The selection is saved to local storage and used as the default when visiting the Triggers page. [[1]](diffhunk://#diff-15fd5025633138038c25fef93288dae67ee6c262c8a6fb5c2c9873a8f9e113c9R66-R74) [[2]](diffhunk://#diff-15fd5025633138038c25fef93288dae67ee6c262c8a6fb5c2c9873a8f9e113c9R305) [[3]](diffhunk://#diff-15fd5025633138038c25fef93288dae67ee6c262c8a6fb5c2c9873a8f9e113c9R517-R521)
* Introduced logic in `Triggers.vue` to read the user's preferred default tab from local storage, defaulting to `add` if not set or invalid.

Localization:

* Added new translation keys and descriptions for the Triggers default tab setting in `en.json`. [[1]](diffhunk://#diff-fbae0f44db543a83ef06ee63dc76c769cd27bea439d08dccc20c42b7ed49bd82R992) [[2]](diffhunk://#diff-fbae0f44db543a83ef06ee63dc76c769cd27bea439d08dccc20c42b7ed49bd82R1004)

Settings metadata:

* Updated settings metadata to include the new Triggers default tab option, ensuring correct display and description in the UI.

Tab options:

* Defined available options for the Triggers default tab in the settings component, supporting both "add" and "manage" tabs.

closes https://github.com/kestra-io/kestra-ee/issues/8575


flcarre (changes requested):
## Review — `feat(settings): add default triggers tab configuration option`

Nice, well-scoped addition and the translations cover all 13 locales. One blocker though: as implemented the setting has no effect on normal navigation, so the feature doesn't actually work yet.

### 🔴 Blocker — the stored preference is never applied on normal navigation

In `Triggers.vue`, `DEFAULT_TAB` (read from `localStorage`) is only referenced inside the watch guard:

```ts
watch(() => route.params.tab, (tab) => {
    if (tab !== undefined && !VALID_TABS.includes(tab as ValidTab)) {
        router.replace({name: "admin/triggers", params: {...route.params, tab: DEFAULT_TAB}})
    }
}, {immediate: true})
```

The route is `/:tenant?/admin/triggers/:tab?` and the sidebar link navigates to `{name: "admin/triggers"}` with **no** `tab` param. So on a normal click `route.params.tab` is `undefined`, the guard's `tab !== undefined` condition is false, and it does nothing. `Tabs.vue` then resolves the active tab via:

```ts
const key = props.embedActiveTab ?? (route?.params?.tab as string | undefined)
return props.tabs.find(t => t.name === key) ?? props.tabs[0]   // tabs[0] === "add"
```

`key` is `undefined`, so it falls back to `tabs[0]` (`"add"`). Result: setting the preference to `"manage"` changes nothing — the stored value is only consulted when a user manually types an *invalid* tab into the URL, which never happens in real use.

**Fix:** make the watcher handle the `undefined` case and redirect to populate the param:

```ts
watch(() => route.params.tab, (tab) => {
    const resolved = VALID_TABS.includes(tab as ValidTab) ? tab as ValidTab : DEFAULT_TAB
    if (tab !== resolved) {
        router.replace({name: "admin/triggers", params: {...route.params, tab: resolved}})
    }
}, {immediate: true})
```

### 🟡 Minor — `"triggersDefaultTab"` is a raw string literal in 3 places

The key appears as an unguarded magic string in `Triggers.vue` and twice in `BasicSettings.vue`. A typo in any one silently breaks the cross-component contract. Sibling settings such as `AUTO_REFRESH_INTERVAL` live in `storageKeys` — consider adding `TRIGGERS_DEFAULT_TAB` there and referencing it everywhere. (Consistency, not correctness — the analogous `executeDefaultTab`/`flowDefaultTab` keys are raw strings today too.)

### 🟡 Minor — cast over a possibly-`null` value

`Triggers.vue`: `localStorage.getItem("triggersDefaultTab") as ValidTab` casts `string | null` to `ValidTab` before the `VALID_TABS.includes(...)` guard. It's runtime-safe, but `?? ""` before the `includes` check keeps the cast honest:

```ts
const stored = localStorage.getItem(storageKeys.TRIGGERS_DEFAULT_TAB) ?? ""
const DEFAULT_TAB: ValidTab = VALID_TABS.includes(stored as ValidTab) ? stored as ValidTab : "add"
```

### ✅ i18n / security / performance

Translations are complete across all 13 locales and `check.js` is clean. No security or performance concerns. The transliteration of "trigger" in some non-Latin locale descriptions matches the existing convention in those files — no action needed.

---

The blocker is the only thing standing between this and shipping — once the watcher applies `DEFAULT_TAB` on entry, the rest is polish.

flcarre (approved):
## Re-review: blocker resolved

The default-tab preference now applies on normal navigation. Traced end-to-end:

- Route `/:tenant?/admin/triggers/:tab?` has an optional `tab` param, and the sidebar link navigates to `{name: "admin/triggers"}` with no param, so `route.params.tab` is `undefined` on a normal click.
- The updated watcher now resolves `undefined` to `DEFAULT_TAB` and `router.replace`s to populate the param. It re-fires once, finds `tab === resolved`, and exits cleanly: loop-safe, `replace` (no history entry), tenant param preserved. `Tabs.vue` then resolves the active tab from the populated param, so a `manage` preference correctly lands on Manage.

Verified the rest:
- **i18n**: all 12 non-en locales got both `fields.triggers_default_tab` and `descriptions.triggers_default_tab`; `triggers_tabs_add` / `triggers_tabs_manage` already exist; reserved term `trigger` kept in English; `check.js` clean.
- **Design system**: `KsSelect` / `KsOption`, no `element-plus` / `:deep` / hex / `--el-*`; option values (`add` / `manage`) match `VALID_TABS`.
- **Security / performance**: nothing to flag (localStorage-only, single read at setup).

Two non-blocking notes, no change required:
- A sub-microtask flash of the `add` tab can occur for users whose stored default is `manage`, since `router.replace` settles asynchronously after the synchronous `{immediate: true}` watch. This is inherent to the same redirect-based default-tab pattern used by `flowDefaultTab`, not a regression.
- The `"triggersDefaultTab"` localStorage key is a raw string literal in a few sites rather than living in `storageKeys`, which matches the existing `executeDefaultTab` / `flowDefaultTab` convention. A good follow-up cleanup for the whole family, out of scope here.

LGTM.

PR #16841, opened 2026-06-16
fix(flows): scope yaml autocomplete and validation to matched task type

### LINKED ISSUE #14646: Autocomplete suggests invalid log level "FATAL" in io.kestra.plugin.core.log.Log
### Describe the issue

While using the io.kestra.plugin.core.log.Log task, the IDE/YAML autocomplete suggests a log level called FATAL. However, this level does not seem to be supported by Kestra.

If I use level: `FATAL`, the flow fails to execute correctly. As per the documentation, the valid levels appear to be: `DEBUG`, `INFO`, `WARN`, `ERROR` and `TRACE`.

Image

So the autocomplete is suggesting an option that is not actually valid and leads to a runtime error.

This can be confusing for users, especially because the suggestion comes directly from the autocomplete and looks official.

**Expected behavior**

The autocomplete should only suggest valid log levels supported by Kestra.

**Actual behavior**

Autocomplete includes FATAL, which causes the flow to fail if used.

**Environment**

Kestra version: v1.1

Happy to provide more details or logs if needed.

### Environment

- Kestra Version: develop

ferdinandlaine:
### ✨ Description

The code editor now uses the `type:` property as a discriminator to prioritize the appropriate `anyOf` schema branch for suggestions and validation.

This fixes:

1. **Wrong schema branch selection** — properties from other task types bleed through: e.g. `FATAL` or `WARNING` are suggested for a Log task's `level` property, and validation errors are reported on the wrong property.
2. **Validation errors without a `type`** — when a task has properties but no `type` yet, the validator picks a branch based on property overlap and reports misleading errors.
3. **Incomplete `type` value suggestions** — once a task has existing properties, most `type` suggestions disappear because non-matching branches are dropped.

This is done by patching `monaco-yaml`, which exposes no public API for `anyOf` branch prioritization or validation error filtering — the underlying logic is bundled from `yaml-language-server` at publish time.

Note that this PR is not a bug fix — `monaco-yaml` behaves as expected for generic schemas — it's an adaptation to Kestra's plugin model, where the `type` property acts as a discriminator for the entire task schema. A more structural approach at the schema level could make this patch unnecessary and is worth considering as a follow-up.

### 👀 Preview

https://github.com/user-attachments/assets/a4294f18-a4ea-4c79-9f6b-195046935339

### 🔗 Related Issue

Fixes #14646

### 🎨 Frontend Checklist

- [x] All existing E2E tests pass
- [x] Video recording attached showing the changes


elevatebart:
Hello @ferdinandlaine,

This PR is a feat of engineering. Thank you for your research. We finally managed to upgrade monaco-yaml and by connection, the yaml language server. 

Can you first check that the bug you fixed still exists in the updated server?
If it still exists, do you want me to recreate your patch in 5.5.1?
We need to resolve merge conflicts before releasing.

Thank you so much for your help and...

Great job! 🎉

ferdinandlaine:
Hi @elevatebart,

The issue still exists with monaco-yaml@5.5.1.
My branch has been reset and minimal changes re-applied.
The partial class name matching fix will be submitted as a separate PR.  
The PR description has also been updated with more details on the choice to patch monaco-yaml.

Ready for review!

elevatebart (approved):

PR #16842, opened 2026-06-16
fix(core): propagate the real secret/Pebble error instead of the Handlebars fallback

nkwiatkowski:
## What

Backport of two `develop` commits to `releases/v0.22.x` so that when a Pebble expression fails for a **runtime / infrastructure** reason (e.g. `{{ secret('...') }}` while Vault returns 502/503, or the older `addSecretConsumer is null` NPE), Kestra **fails fast with the real error** instead of masking it.

Cherry-picked from `develop`:
- `77f0ee6667` chore(execution): remove the alternate variable renderer
- `a80052ed6a` feat(execution): make the VariableRender tolerant about any runtime exception

## Why

Customer report (Leroy Merlin, ticket #1916, on v0.22.43): flows using `{{ secret('...') }}` failed during a Vault incident with a cryptic error:

```
com.github.jknack.handlebars.HandlebarsException: found: ''hr-squad'', expected: 'id'
{{ secret('hr-squad') }}
    at io.kestra.plugin.templates.runners.VariableRenderer.alternativeRender(VariableRenderer.java:304)
    at io.kestra.core.runners.VariableRenderer.renderOnce(VariableRenderer.java:118)
```

**Root cause:** on a Vault 502/503 the secret manager throws `IOException`, which `SecretFunction` wraps in a `PebbleException`. `VariableRenderer.renderOnce` then routed that through the legacy **Handlebars** `alternativeRender` fallback (provided by `plugin-templates`), which cannot parse native Pebble syntax and throws a `HandlebarsException` — discarding the real Vault error. Same mechanism masked the previous `addSecretConsumer is null` NPE variant.

After this change the original error propagates as `IllegalVariableEvaluationException` (wrapping the actual IO/secret failure), and any raw `RuntimeException` thrown by a Pebble function is wrapped cleanly rather than crashing the worker/scheduler.

## Notes
- Already fixed on `develop` (2.0); not present in any released line — this backports it to 0.22.x. Companion PRs target `releases/v1.0.x` (LTS) and `releases/v1.3.x`.
- Removing the core hook is runtime-safe with the existing `plugin-templates` JAR (the override is simply never called). Retiring the now-dead override in `plugin-templates` is a separate follow-up.

## Tests
`./gradlew :core:test` for `VariableRendererTest` and the affected `*FunctionTest` classes pass locally. 0.22.x-specific tests that asserted a raw runtime exception from `render(...)` were updated to assert the wrapped `IllegalVariableEvaluationException` + cause.

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
+++ b/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
@@ -114,7 +114,7 @@ public class VariableRenderer {
             OutputWriter writer = stringify ? new JsonWriter() : new TypedObjectWriter();
             compiledTemplate.evaluate(writer, variables);
             result = writer.output();
-        } catch (IOException | PebbleException e) {
+        } catch (IOException | RuntimeException e) {
             if (e instanceof PebbleException pebbleException) {
                 throw properPebbleException(pebbleException);
             }
--- a/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
+++ b/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
@@ -115,15 +115,10 @@ public class VariableRenderer {
             compiledTemplate.evaluate(writer, variables);
             result = writer.output();
         } catch (IOException | PebbleException e) {
-            String alternativeRender = this.alternativeRender(e, (String) inline, variables);
-            if (alternativeRender == null) {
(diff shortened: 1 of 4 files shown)


nkwiatkowski:
Closing: the customer still relies on the alternate (Handlebars) variable renderer, so it will be kept. Another team will handle this issue.

PR #16847, opened 2026-06-16
feat(core): widen locks table, fail executions on locked-asset writes

brian-mulier-p:
- Widen locks.key 250->700 and locks.id 150->500 (2.0.13-widen-locks, all three dialects) so asset lease ids (tenant_assetId) fit. H2 drops+re-adds the generated id column (idempotent); Postgres/MySQL alter in place.
- `AssetService.asyncUpsert` now declares a checked `InternalException` alongside `QueueException`, so the EE write path can fail an execution writing to an asset locked by another owner.
- `ExecutorService`'s asset-output upsert switches from a `forEach` to a plain for-loop so that `InternalException` propagates out (instead of being swallowed) and fails the execution; ordering-dependent by design — if output #k is rejected, outputs #k+1.. are not emitted.
- Document `SystemTask` as EXCEPTIONAL usage (runs on the executor, not horizontally scalable; repository-backed work belongs on scalable workers reaching data via the SDK) — a guardrail against reaching for it by default.
- Note in `AbstractJdbcLockRepository` that the `locks` table is now shared with EE's durable Lease store; the server-death sweep is keyed by server-instance id and never collides with lease owner ids (execution/user ids).
- The actual Lease/LeaseStore/LeaseHeldException lock abstraction lives entirely in EE (core-ee/locks) — this PR carries only the OSS-side plumbing (schema widening + locked-write failure propagation) that the EE feature depends on.
- Companion EE PR: kestra-io/kestra-ee#8667


core/src/main/java/io/kestra/core/models/tasks/SystemTask.java
```diff
@@ -12,6 +12,26 @@
...
+ * A high-volume task (one invoked by many flows / many executions) implemented as a SystemTask will
+ * funnel that load onto the executor and become a scalability bottleneck. If such a task needs a
+ * repository-backed operation, it must run on a <b>regular worker</b> and reach the operation
+ * indirectly — via a gRPC controller service (the MetaStore pattern) or the Kestra client SDK — so
+ * the work stays on the scalable worker tier and only the data access crosses to the trusted
+ * process.</p>
+ *
+ * <p>This note is also a guardrail for AI agents and contributors: prefer the gRPC/SDK indirection
+ * for anything that isn't clearly rare and low-volume. Reach for {@code SystemTask} only when no
+ * scalable alternative is reasonable, and say why.</p>
```
loicmathieu: Please remove this noisy comment

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -1383,16 +1383,20 @@ public void addWorkerTaskResult(ExecutorContext executor, Supplier<FlowWithSourc
...
+                // don't update output assets if task fail
                 if (!taskRun.getState().isFailed()) {
-                    taskRun.getAssets().getOutputs().forEach(asset ->
-                    {
+                    // Plain for-loop (not forEach) so a checked InternalException from asyncUpsert (EE: a write
+                    // to an asset locked by a different owner) propagates out of addWorkerTaskResult; on the
+                    // worker-task-result path it reaches WorkerTaskResultMessageHandler, which fails the
+                    // execution. Fail-fast and ordering-dependent by design: if output #k is rejected, outputs
+                    // #k+1.. are not emitted, while earlier upserts and the lineage event above already were.
+                    for (var asset : taskRun.getAssets().getOutputs()) {
```
loicmathieu: Can you summarize this comment to avoid too much documentation growth

jdbc-h2/src/main/java/io/kestra/repository/h2/migration/V2_0_19WidenLocksColumnsMigration.java
loicmathieu, line 1: 1.0.13 is already taken (twice)

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLockRepository.java
```diff
@@ -75,6 +75,9 @@ public void deleteById(String category, String id) {
 
     @Override
     public List<Lock> deleteByOwner(String owner) {
+        // The `locks` table is shared with the EE durable Lease store. This server-death sweep is keyed by
+        // owner = ServerInstance.INSTANCE_ID; leases are owned by execution or user ids, which never
+        // collide with a server instance id, so a lease is never swept here.
```
loicmathieu: Don't think it adds value here, please remove this comment

loicmathieu (approved):
Please rename the migrations and remove/summarize comments before merging

fhussonnois (approved):
I know, I'm a little annoying about this, but could you make sure that all the test methods follow our naming convention? :)

jdbc-h2/src/main/java/io/kestra/repository/h2/migration/V2_0_19WidenLocksColumnsMigration.java
brian-mulier-p: Was in my plan to update the version right before the merge for sure :+1:

PR #16848, opened 2026-06-16
fix(core): prevent worker crash on interrupted log upload

### LINKED ISSUE #16623: Bug: Worker crashes entirely (full JVM shutdown) when a thread is interrupted during the terminal-state log upload to S3
## Summary
When a worker thread is interrupted (e.g. the execution is killed, the task is
cancelled, etc **while it is uploading a
task's file-backed logs to S3**, the AWS SDK throws
`software.amazon.awssdk.core.exception.AbortedException: Thread was interrupted`
(caused by `SdkInterruptedException`). This exception is **not caught** inside
`DefaultWorker.runAttempt` / `DefaultRunContext.logFileURI`, so it propagates to
`io.kestra.core.utils.ThreadUncaughtExceptionHandler`, whose `uncaughtException`
handler calls `KestraContext$Initializer.shutdown()` and **shuts down the entire
worker JVM**.

The result is that a single interrupted task takes down the whole worker, killing
every other in-flight execution running on that worker and forcing a cold restart
(`process_start_time_seconds` resets). What should be an isolated, recoverable
per-task failure becomes a full server outage. This also causes worker flow execution
resubmission to skyrocket during these restarts during peak hours.

This started for us right after we enabled `logToFile: true` across all flows.
With `logToFile: true`, task logs bypass the database and are written to a local
file; when the execution reaches a terminal state Kestra uploads that file to the
internal storage backend (S3). The interruptible S3 upload on that terminal-state
path is exactly where the edge case lives.

## Consistent findings on our investigations
1. A worker thread (`worker_NNNN`) finishes/►is running a task attempt and, on the
   terminal-state path, uploads the file-backed logs to S3 via
   `DefaultRunContext.logFileURI` → `InternalStorage.putFile` → `S3Storage.put` →
   `S3Storage.exists` → `S3Client.headObject`.
2. The thread is interrupted (execution killed / task cancelled / worker shutdown
   signal). The AWS SDK detects the interrupt mid-call and throws
   `AbortedException` (caused by `SdkInterruptedException`).
3. `AbortedException` is a `RuntimeException` and is **not handled** by the worker
   task-execution code, so it reaches the thread's uncaught-exception handler.
4. `ThreadUncaughtExceptionHandler.uncaughtException()` calls
   `KestraContext$Initializer.shutdown()`, which stops the Micronaut application
   context (closing the Hikari datasource, etc.) → the **entire worker JVM exits**.

This means **any** uncaught exception on **any** worker thread is treated as fatal
and terminates the whole server; the interrupted S3 log upload is a very easy way
to trigger it once `logToFile: true` is enabled.

## Steps to reproduce
1. Configure S3 as the internal storage backend.
2. Set `logToFile: true` (globally or on a flow) so task logs are written to a file
   and uploaded to S3 when the execution reaches a terminal state.
3. Start a flow/task that produces logs.
4. While the terminal-state log upload to S3 is in flight, interrupt the worker
   thread — e.g. **kill the execution** or cancel the task
5. Observe that the worker logs `Caught an exception ... Shutting down.` with an
   `AbortedException`, and the **entire worker process shuts down** (not just the
   one task), taking all other concurrent executions down with it.

## Expected behavior
An interrupted/aborted S3 log upload should fail (or be skipped/retried) for **that
single task run only**. It should never escalate to a full worker JVM shutdown.

## Actual behavior
The `AbortedException` propagates to `ThreadUncaughtExceptionHandler`, which shuts
down the whole worker via `KestraContext.shutdown()`. All in-flight executions on
the worker are aborted and the process restarts cold.

## Full stacktrace (exactly as it appears in our worker logs)

Primary exception — interrupted S3 upload escaping to the uncaught handler:

```
Caught an exception in Thread[#23563,worker_3783,5,main]. Shutting down. software.amazon.awssdk.core.exception.AbortedException: Thread was interrupted at software.amazon.awssdk.core.exception.AbortedException$BuilderImpl.build(AbortedException.java:102) at software.amazon.awssdk.core.exception.AbortedException.create(AbortedException.java:38) at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.handleInterruptedException(ApiCallTimeoutTrackingStage.java:148) at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.translatePipelineException(ApiCallTimeoutTrackingStage.java:109) at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:64) at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:43) at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallMetricCollectionStage.execute(ApiCallMetricCollectionStage.java:50) at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallMetricCollectionStage.execute(ApiCallMetricCollectionStage.java:32) at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206) at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206) at software.amazon.awssdk.core.internal.http.pipeline.stages.ExecutionFailureExceptionReportingStage.execute(ExecutionFailureExceptionReportingStage.java:37) at software.amazon.awssdk.core.internal.http.pipeline.stages.ExecutionFailureExceptionReportingStage.execute(ExecutionFailureExceptionReportingStage.java:26) at software.amazon.awssdk.core.internal.http.AmazonSyncHttpClient$RequestExecutionBuilderImpl.execute(AmazonSyncHttpClient.java:210) at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.invoke(BaseSyncClientHandler.java:103) at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.doExecute(BaseSyncClientHandler.java:173) at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.lambda$execute$1(BaseSyncClientHandler.java:80) at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.measureApiCallSuccess(BaseSyncClientHandler.java:182) at software.amazon.awssdk.core.internal.handler.BaseSyncClientHandler.execute(BaseSyncClientHandler.java:74) at software.amazon.awssdk.core.client.handler.SdkSyncClientHandler.execute(SdkSyncClientHandler.java:45) at software.amazon.awssdk.awscore.client.handler.AwsSyncClientHandler.execute(AwsSyncClientHandler.java:53) at software.amazon.awssdk.services.s3.DefaultS3Client.headObject(DefaultS3Client.java:7623) at io.kestra.storage.s3.S3Storage.exists(S3Storage.java:134) at io.kestra.storage.s3.S3Storage.mkdirs(S3Storage.java:536) at io.kestra.storage.s3.S3Storage.put(S3Storage.java:358) at io.kestra.storage.s3.S3Storage.put(S3Storage.java:342) at io.kestra.core.storages.StorageInterface.put(StorageInterface.java:186) at io.kestra.core.storages.InternalStorage.putFile(InternalStorage.java:170) at io.kestra.core.storages.InternalStorage.putFile(InternalStorage.java:189) at io.kestra.core.storages.InternalStorage.putFile(InternalStorage.java:178) at io.kestra.core.runners.DefaultRunContext.logFileURI(DefaultRunContext.java:406) at io.kestra.worker.DefaultWorker.runAttempt(DefaultWorker.java:965) at io.kestra.worker.DefaultWorker.run(DefaultWorker.java:743) at io.kestra.worker.DefaultWorker.handleTask(DefaultWorker.java:378) at io.kestra.worker.DefaultWorker.lambda$run$8(DefaultWorker.java:296) at io.micrometer.core.instrument.internal.TimedRunnable.run(TimedRunnable.java:49) at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) at java.base/java.lang.Thread.run(Unknown Source) Caused by: software.amazon.awssdk.core.exception.SdkInterruptedException at software.amazon.awssdk.core.internal.http.InterruptMonitor.checkInterrupted(InterruptMonitor.java:40) at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:52) at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:35) at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.executeWithTimer(ApiCallTimeoutTrackingStage.java:82) at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:62) ... 33 common frames omitted
```

Escalation — the uncaught handler then shuts the whole server down (observed
immediately afterward on the same thread while closing the Micronaut context):

```
HikariPool-1 - Interrupted during closing java.lang.InterruptedException at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.awaitNanos(Unknown Source) at java.base/java.util.concurrent.ThreadPoolExecutor.awaitTermination(Unknown Source) at com.zaxxer.hikari.pool.HikariPool.shutdown(HikariPool.java:246) at com.zaxxer.hikari.HikariDataSource.close(HikariDataSource.java:350) at io.micronaut.configuration.jdbc.hikari.DatasourceFactory.close(DatasourceFactory.java:116) at io.micronaut.configuration.jdbc.hikari.$DatasourceFactory$Definition.dispose(Unknown Source) at io.micronaut.inject.DisposableBeanDefinition.dispose(DisposableBeanDefinition.java:44) at io.micronaut.context.DefaultBeanContext.destroyBean(DefaultBeanContext.java:1214) at io.micronaut.context.DefaultBeanContext.destroyBean(DefaultBeanContext.java:1185) at io.micronaut.context.DefaultBeanContext.stop(DefaultBeanContext.java:447) at io.micronaut.context.DefaultApplicationContext.stop(DefaultApplicationContext.java:236) at io.micronaut.context.DefaultApplicationContext.stop(DefaultApplicationContext.java:76) at io.micronaut.context.LifeCycle.close(LifeCycle.java:61) at io.kestra.core.contexts.KestraContext$Initializer.shutdown(KestraContext.java:175) at io.kestra.core.utils.ThreadUncaughtExceptionHandler.uncaughtException(ThreadUncaughtExceptionHandler.java:26) at java.base/java.lang.Thread.dispatchUncaughtException(Unknown Source)
```

### Environment

- Kestra version: **1.3.21**
- Queue/Repository backend: **JDBC (PostgreSQL 16)**
- Internal storage: **S3** (`io.kestra.storage.s3.S3Storage`)
- Relevant config: **`logToFile: true`** applied to all flows

chaitanyakatore:
### ✨ Description

When a task execution is cancelled/killed during log file upload (e.g. upload to S3), the storage layer throws a runtime exception (such as `AbortedException: Thread was interrupted`). Since this exception was not caught inside `DefaultRunContext.logFileURI()`, it propagated to Kestra's uncaught exception handler, resulting in a fatal JVM shutdown of the entire worker process.

This PR wraps the log file upload process in `DefaultRunContext.logFileURI()` in a generic catch block that:
- Catches any `Exception` to prevent worker crash.
- Logs a warning rather than escalating to a full worker JVM shutdown.
- Safely checks for thread interruption status and restores the thread's interrupted flag (via `Thread.currentThread().interrupt()`).
- Safely returns `null`.
- Includes unit tests covering both generic `IOException` and interrupted `RuntimeException` failures during log upload to prevent future regressions.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/16623

### 🛠️ Backend Checklist


core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
```diff
@@ -406,8 +406,15 @@ public URI logFileURI() {
...
-                throw new RuntimeException(e);
+            } catch (Exception e) {
+                logger().warn("Failed to upload log file to storage", e);
+                if (Thread.currentThread().isInterrupted() || 
+                    e instanceof InterruptedException || 
+                    e.getCause() instanceof InterruptedException || 
+                    e.getClass().getName().contains("Interrupt") || 
+                    (e.getCause() != null && e.getCause().getClass().getName().contains("Interrupt"))) {
+                    Thread.currentThread().interrupt();
+                }
```
loicmathieu: Can you explain this? I feel this is not necessary

chaitanyakatore: @loicmathieu  When S3 (or any other external storage client) uploads the log file and the thread is cancelled/killed, the AWS SDK catches the JDK InterruptedException internally, clears the thread's interrupted flag, and wraps it in a proprietary runtime exception: AbortedException (caused by SdkInterruptedException).

There are two reasons why this block is necessary here:

- Restoring the Interrupted Status: If we just swallow the exception, the thread's interrupted status remains cleared, and the parent worker loop won't know the thread was cancelled/interrupted, which can lead to threads hanging or not terminating during shutdown. Calling Thread.currentThread().interrupt() is the standard Java best practice to restore that cancelled state.
- Classpath / Plugin Isolation: Because the core module does not have a dependency on the S3/AWS SDK storage plugin (which is isolated in storage-s3), we cannot use type checks like e instanceof SdkInterruptedException directly as it would cause compilation failures. Checking the class name for "Interrupt" is a safe, classpath-independent way to detect that the storage plugin's upload was aborted due to thread interruption.

loicmathieu: This seems to broad, I think only checking directly the InterruptedException and the cause is an InterruptedException i enought.
Otherwise, I'd prefer an exact match on the class name instead of contains

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
chaitanyakatore: @loicmathieu I have made the changes, please review

core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
```diff
@@ -406,8 +406,19 @@ public URI logFileURI() {
...
+                    e instanceof InterruptedException || 
+                    e.getCause() instanceof InterruptedException || 
+                    "software.amazon.awssdk.core.exception.AbortedException".equals(e.getClass().getName()) || 
+                    "com.amazonaws.AbortedException".equals(e.getClass().getName()) || 
+                    (e.getCause() != null && (
+                        "software.amazon.awssdk.core.exception.AbortedException".equals(e.getCause().getClass().getName()) || 
+                        "com.amazonaws.AbortedException".equals(e.getCause().getClass().getName())
+                    ))) {
+                    Thread.currentThread().interrupt();
+                }
```
loicmathieu: It's not correct, you need to re-interupt the thread in all cases so you need to use 2 if and not only one.

core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
```diff
@@ -406,8 +406,19 @@ public URI logFileURI() {
...
                 return logFileURI;
-            } catch (IOException e) {
-                throw new RuntimeException(e);
+            } catch (Exception e) {
+                logger().warn("Failed to upload log file to storage", e);
+                if (Thread.currentThread().isInterrupted() || 
+                    e instanceof InterruptedException || 
+                    e.getCause() instanceof InterruptedException || 
+                    "software.amazon.awssdk.core.exception.AbortedException".equals(e.getClass().getName()) || 
+                    "com.amazonaws.AbortedException".equals(e.getClass().getName()) ||
```
loicmathieu: Can you externalize these two checks in a dedicated method to avoir duplication and allow to add more clause later if we discover the same issue in other implementation?

core/src/test/java/com/amazonaws/AbortedException.java
loicmathieu, line 1: Why?

core/src/test/java/software/amazon/awssdk/core/exception/AbortedException.java
loicmathieu, line 1: Why?

loicmathieu (changes requested):

danlsgiga:
@chaitanyakatore any traction on this?

loicmathieu:
@chaitanyakatore are you willing to continue on this PR and take into account my remaining feedback?

chaitanyakatore:
> @chaitanyakatore are you willing to continue on this PR and take into account my remaining feedback?

I will work on your feedback and try to complete it

MilosPaunovic:
Hey @chaitanyakatore, are there any updates on this?

chaitanyakatore:
Hi @loicmathieu , I've updated the PR based on your feedback:

- Split thread interruption re-assert logic into two separate if blocks.
- Externalized AWS AbortedException checks into dedicated helper methods (isAbortedException and isInterrupted) to 
- eliminate duplication and allow future extension.
- Removed standalone dummy exception files from src/test/java and cleaned up the unit tests.
Please take another look when you get a chance!

danlsgiga:
Would love to see this in the next release

danlsgiga:
Hey folks, any updates on this?

core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
```diff
@@ -432,13 +432,35 @@ public URI logFileURI() {
                     logger().warn("Unable to delete the log file {}", logger.getLogFile().toPath());
                 }
                 return logFileURI;
-            } catch (IOException e) {
-                throw new RuntimeException(e);
+            } catch (Exception e) {
+                logger().warn("Failed to upload log file to storage", e);
+                if (Thread.currentThread().isInterrupted()) {
+                    Thread.currentThread().interrupt();
+                }
```
loicmathieu: This seems useless, you check the interrupted status and if true you re-interrupt only. This is a no-op.

Can you remove it, then I'll merge this.

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
chaitanyakatore: Hi @loicmathieu, I've removed the redundant Thread.currentThread().isInterrupted() check as requested. Thanks!

loicmathieu (approved):
LGTM

PR #16849, opened 2026-06-16
fix(storage): retry temp file deletion to handle transient Windows file locks

ammeek:
Stops the spurious `Failed to delete temporary file '...'` warning that can occur on Windows after files are uploaded to internal storage.

The root cause is that `InternalStorage.putFile(File)` deletes temporary files immediately after upload using a single `Files.delete()` call. On Windows, recently closed files can remain temporarily locked, causing the delete to fail and log a warning even though the upload succeeded. This change adds retry logic for temporary file deletion and makes the retry behaviour configurable, so warnings are only logged when deletion still fails after all retry attempts.


fhussonnois (approved):
LGTM!

PR #16865, opened 2026-06-17
feat(worker): expose task completion rate in worker heartbeat

fhussonnois:
Add a per-Worker RateMeter that derives the task completion rate (tasks/sec) from the existing worker.ended.count counter — EWMA-smoothed (ALPHA=0.3), sampled on each heartbeat and emitted as the worker.tasks.rate metric in the heartbeat set. No new Micrometer meter is registered; the rate is an operational figure for the Worker Group UI only.


ammeek (approved):
LGTM

PR #16869, opened 2026-06-17
fix(core): hide progress bar in KPI widget when value is 0

MilosPaunovic:
Looks weird with progress bar if it's on 0, so I removed it on that case. If it goes against the designs, let's close this.

image


Piyush-r-bhaskar (approved):
That's a good idea 💡

PR #16871, opened 2026-06-17
fix(core): return 204 instead of 409 when webhook conditions are not met

### LINKED ISSUE #16601: Webhook throw 409 HTTP error code when trigger condition does not match
### Describe the issue

While implementing a web-hook trigger based on an Expression I noticed that a passing condition would result in 200 code but an unmet one resulted in a 409 CONFLICT HTTP error code while it should have been 204 as per documentation.

Image

The flow passing with a matching condition, returning 200
```
id: webhook-with-auth
namespace: nve.team

tasks:
  - id: logBody
    message: "{{ trigger.body }}"

triggers:
  - id: authenticated_webhook
    key: "api" # Replace with a strong, secret key for your webhook
    wait: true
    returnOutputs: true
    conditions:
     - type: io.kestra.plugin.core.condition.Expression
       expression: "{{ 10 <= 10 }}"
```

Same flow with a non-matching expression, returning 409 Conflict

```
id: webhook-with-auth
namespace: nve.team

tasks:
  - id: logBody
    message: "{{ trigger.body }}"

triggers:
  - id: authenticated_webhook
    key: "api" # Replace with a strong, secret key for your webhook
    wait: true
    returnOutputs: true
    conditions:
     - type: io.kestra.plugin.core.condition.Expression
       expression: "{{ 10 < 10 }}"
```

It is to be noted that in this case the Execution is not even logged, and this is the log I get from the Docker container:
```
kestra-1  | 11:34:27.558 INFO  default-eventLoopGroup-2-1 io.kestra.webserver.access 2026-06-06T11:34:27.167Z | POST /api/v1/main/executions/webhook/nve.team/webhook-with-auth/api HTTP/1.1 | status: 409 | ip: 90.xx.xx.xx | length: - | duration: 10
```

### Environment

- Kestra Version: 1.3.21
- Docker
- 6.17.0-1016-oracle #16~24.04.1-Ubuntu SMP aarch64 GNU/Linux

Vidhaankhare16:
### ✨ Description

The `Webhook` trigger returned `409 CONFLICT` when a trigger condition did not match, but its documented contract is `204 No Content` for "conditions not met".

The cause is that `WebhookService.newExecution(...)` returned `Optional.empty()` for two different reasons — conditions not met **and** input rendering failed — which `Webhook.evaluate` collapsed into a single `409`. This PR keeps `Optional.empty()` for "conditions not met" → `204`, and throws a new `WebhookInputRenderException` on render failure → `422` (previously an undocumented `409`), so the two cases are no longer conflated.

### 🔗 Related Issue

Closes #16601

### 🛠️ Backend Checklist

### 📝 Additional Notes

- The render-failure path now returns `422` (was an also-incorrect `409`). Happy to use a different code if preferred — it's a one-line change.
- All webhook tests pass locally across `:core` and `:webserver`, including `webhookWithCondition` → `204`.

### 🤖 AI Authors

I used Claude code to understand and fix the issue better but verified everything and ran all tests myself.

🐱 And because the template asks: why did the cat return a `204`? The condition didn't match, so it had *no content* to give.

### DIFF
--- a/core/src/main/java/io/kestra/core/services/WebhookService.java
+++ b/core/src/main/java/io/kestra/core/services/WebhookService.java
@@ -27,6 +27,7 @@ import io.kestra.core.utils.IdUtils;
 import io.kestra.core.utils.UriProvider;
 import io.kestra.plugin.core.trigger.AbstractWebhookTrigger;
 import io.kestra.plugin.core.trigger.WebhookContext;
+import io.kestra.plugin.core.trigger.WebhookInputRenderException;
 import io.kestra.plugin.core.trigger.WebhookResponse;
 
 import io.micronaut.context.event.ApplicationEventPublisher;
@@ -101,7 +102,8 @@ public class WebhookService {
      * @param context The webhook context containing request, path, flow, and services
      * @param trigger The webhook trigger
      * @param output The trigger output to attach to the execution
-     * @return The prepared execution, or empty if conditions are not met
+     * @return The prepared execution, or empty if the trigger conditions are not met
+     * @throws WebhookInputRenderException if the trigger inputs cannot be rendered or processed
      */
     public Optional<Execution> newExecution(WebhookContext context, Flow flow, AbstractWebhookTrigger trigger, io.kestra.core.models.tasks.Output output) {
         Execution execution = Execution.builder()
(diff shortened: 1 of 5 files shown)


loicmathieu (approved):
LGTM, thanks.

Vidhaankhare16:
Hi @loicmathieu, All 6,949 tests have passed but the workflow is failing to report results due to missing GitHub token permissions (checks & pull-requests write). Can you please look into this? Thank you!

loicmathieu:
@Vidhaankhare16 this is a known issue, I'll merge it anyway

PR #16881, opened 2026-06-17
fix(core): fail fast on pebble function errors

ammeek:
## ✨ Description

When a Pebble function such as `secret()` failed because the underlying secret backend (for example Vault) was unavailable or returned a server error, the original exception was masked instead of being propagated.

The failure path looked like this:

1. The secret backend throws a `SecretException`.
2. `SecretFunction` wraps it in a `PebbleException`, preserving the original cause.
3. `VariableRenderer.renderOnce(...)` catches every `PebbleException` and unconditionally falls back to the legacy Handlebars renderer (`alternativeRender`).

That fallback exists only to support legacy Handlebars syntax. Since the template is valid Pebble, Handlebars attempts to parse expressions such as `secret(...)` and fails with a misleading parse error like:

```text
found: '...', expected: 'id'
```

As a result, the real infrastructure error (for example a Vault connectivity failure or HTTP 5xx response) was lost.

This PR changes `renderOnce(...)` to inspect the exception cause before falling back to Handlebars. If the `PebbleException` wraps a non-Pebble exception, it indicates that template parsing succeeded and a function or filter failed during execution. In that case, the Handlebars fallback is skipped and the exception is propagated through `properPebbleException(...)`, preserving both the original cause and the template location.

Pure Pebble parsing and attribute-resolution errors (which have no wrapped runtime cause) continue to fall back to Handlebars as before, preserving compatibility with legacy templates.

This also aligns the behavior with `develop`, where the legacy Handlebars fallback has been removed entirely, so runtime exceptions are already propagated directly.

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
+++ b/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
@@ -115,7 +115,15 @@ public class VariableRenderer {
             compiledTemplate.evaluate(writer, variables);
             result = writer.output();
         } catch (IOException | PebbleException e) {
-            String alternativeRender = this.alternativeRender(e, (String) inline, variables);
+            // The Handlebars fallback only exists to bridge legacy Handlebars-vs-Pebble syntax
+            // differences. If the failure originates from a function that hit a real error - e.g.
+            // the secret() function failing because the secret backend (Vault) is unreachable -
+            // re-rendering with Handlebars cannot help and only masks the real cause, surfacing a
+            // cryptic "found: '...', expected: 'id'" syntax error instead. Skip the fallback in that
+            // case and let the error propagate (properPebbleException keeps the real cause in the chain).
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/kestra/core/runners/VariableRenderer.java
```diff
@@ -115,7 +115,15 @@ public Object renderOnce(Object inline, Map<String, Object> variables, boolean s
             compiledTemplate.evaluate(writer, variables);
             result = writer.output();
         } catch (IOException | PebbleException e) {
-            String alternativeRender = this.alternativeRender(e, (String) inline, variables);
+            // The Handlebars fallback only exists to bridge legacy Handlebars-vs-Pebble syntax
```
fhussonnois: IMHO, this comment is to verbose

fhussonnois (approved):

Issue #16874, opened 2026-06-17
core + client-sdk: resolve template expressions in plugin artifact/topology display


fdelbrayelle:
## Parent epic
Sub-issue of #12696

## Problem

When a plugin emits topology details (pre-execution artifacts), raw Pebble expressions are displayed instead of resolved values. Example from `io.kestra.plugin.ee.aws.runner.Batch`:

```yaml
taskRunner:
  type: io.kestra.plugin.ee.aws.runner.Batch
  region: "{{ vars.region }}"
  computeEnvironmentArn: "{{ vars.compute_environment_arn }}"
  jobQueueArn: "{{ vars.job_queue_arn }}"
  accessKeyId: "{{ secret('AWS_ACCESS_KEY_ID') }}"
```

The topology node currently shows `{{ vars.region }}` instead of `us-east-1`. This affects all plugins that emit topology/artifact data with templated property values.

## Expression taxonomy & resolution rules

| Category | Syntax | Pre-exec | Post-exec | Display rule |
|---|---|:---:|:---:|---|
| Flow variables | `{{ vars.xxx }}` | ✅ | ✅ | Resolve from flow `variables` map |
| Flow metadata | `{{ flow.id }}`, `{{ flow.namespace }}`, etc. | ✅ | ✅ | Resolve from flow definition |
| Inputs | `{{ inputs.xxx }}` | ❌ | ✅ | Pre-exec: show raw; post-exec: resolve |
| Outputs | `{{ outputs.task.xxx }}` | ❌ | ✅ | Pre-exec: show raw; post-exec: resolve |
| Execution context | `{{ execution.id }}`, `{{ execution.startDate }}`, etc. | ❌ | ✅ | Pre-exec: show raw; post-exec: resolve |
| Trigger context | `{{ trigger.xxx }}` | ❌ | ✅ | Pre-exec: show raw; post-exec: resolve |
| Task run | `{{ taskrun.xxx }}`, `{{ parents.xxx }}` | ❌ | ✅ | Pre-exec: show raw; post-exec: resolve |
| Server globals | `{{ globals.xxx }}` | ✅ | ✅ | Resolve via API |
| KV store | `{{ kv('key') }}` | ❌ | ❌ | Show raw |
| Secrets | `{{ secret('KEY') }}` | 🔒 | 🔒 | **Always** mask as `[secret: KEY_NAME]` — never resolve |
[...]

PR #16884, opened 2026-06-18
feat(topology): resolve display expressions in topology details

jymaire:
## What

Implements #16874. The backend computes a parallel **`renderedProperties`** map per task node during flow-graph build and ships it alongside the raw task to the `topology-details` slot, so plugin artifacts can display resolved values instead of raw Pebble expressions.

Single authoritative server-side Pebble implementation — no TypeScript reimplementation, no bulk variable context leaked to the client.

## Security model — allowlist, not denylist

Display resolution runs Pebble against task properties, so it is security-sensitive. The display engine uses an **allowlist**: only pure, side-effect-free functions are ever invoked.

| Category | Behavior |
|----------|----------|
| `fromJson`, `fromIon`, `yaml`, calendar helpers | resolved |
| `secret(...)` | masked `[secret: KEY]` — real service **never invoked** |
| `env(...)` | masked `[env: NAME]` — env vars **never read** |
| everything else (`kv`, `render`, `decrypt`, `fetchContext`, file IO, `now`/`uuid`, **any future/plugin function**) | kept raw, **never invoked** |

Additional hardening:
- **`envs.*` direct map access** is stripped from the display context, so it cannot bypass the `env()` mask.
- **`renderedProperties` is cleared in `forExecution()`**, so it does not bypass the EXECUTION-READ field reduction (`TaskForExecution.of()`).
- Segment-level resolution keeps unresolvable segments raw (e.g. `{{ vars.region }}-{{ now() }}` → `us-east-1-{{ now() }}`).
- `{% raw %}` blocks preserved via collision-proof (UUID) placeholders.

## Frontend
- `topology-details` slot contract gains an optional `resolvedProperties` field (backward-compatible; raw `task` retained).
- `LowCodeEditor` passes `node?.task?.renderedProperties` through.

## Tests
26 tests passing — covers vars/flow resolution, inputs raw pre-exec / resolved post-exec, allowlist (`fromJson` resolves, `kv`/`fetchContext` raw), secret/env masking with a **"never invoked" `SecretService` spy**, the `envs.*` bypass, and the `forExecution()` strip.

## Not included
- `package-lock.json` peer-dep churn and untracked `cli/data/` were intentionally left out to keep the diff focused.
- Redacting literal `@PluginProperty(secret=true)` values is pre-existing behavior (the raw task already ships those) — out of scope; candidate for a separate hardening ticket.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/hierarchies/AbstractGraphTask.java
+++ b/core/src/main/java/io/kestra/core/models/hierarchies/AbstractGraphTask.java
@@ -2,6 +2,7 @@ package io.kestra.core.models.hierarchies;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 
 import io.kestra.core.models.executions.TaskRun;
 import io.kestra.core.models.tasks.TaskForExecution;
@@ -21,6 +22,8 @@ public abstract class AbstractGraphTask extends AbstractGraph {
     private final TaskRun taskRun;
     private final List<String> values;
     private final RelationType relationType;
+    /** Task properties with Pebble expressions resolved for display. Null when resolution was not attempted. */
+    private Map<String, Object> renderedProperties;
 
     public AbstractGraphTask(String uid, TaskInterface task, TaskRun taskRun, List<String> values, RelationType relationType) {
         super(uid);
@@ -54,9 +57,19 @@ public abstract class AbstractGraphTask extends AbstractGraph {
         return String.join("_", list);
     }
 
+    /** Sets display-resolved properties on this node and returns {@code this} for chaining. */
+    public AbstractGraphTask withRenderedProperties(Map<String, Object> resolved) {
+        this.renderedProperties = resolved;
+        return this;
+    }
+
     @Override
     public AbstractGraph forExecution() {
         this.setTask(TaskForExecution.of(this.getTask()));
+        // forExecution() reduces the visible task to a safe subset for principals that only have
+        // EXECUTION-READ. The resolved properties hold the full task configuration, so they must be
+        // stripped here too, otherwise they would bypass that reduction.
+        this.renderedProperties = null;
 
         return this;
     }
--- a/core/src/main/java/io/kestra/core/runners/pebble/PebbleEngineFactory.java
+++ b/core/src/main/java/io/kestra/core/runners/pebble/PebbleEngineFactory.java
@@ -4,11 +4,25 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Proxy;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import io.kestra.core.runners.VariableRenderer;
 import io.kestra.core.runners.configuration.VariableConfiguration;
+import io.kestra.core.runners.pebble.functions.DayOfMonthFunction;
+import io.kestra.core.runners.pebble.functions.DayOfWeekFunction;
+import io.kestra.core.runners.pebble.functions.EnvFunction;
+import io.kestra.core.runners.pebble.functions.FromIonFunction;
+import io.kestra.core.runners.pebble.functions.FromJsonFunction;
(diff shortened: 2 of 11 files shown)


jymaire:
**Re: the `env` resolution row in the spec**

Flagging a deviation between the spec's env row and the current implementation, plus a proposed resolution.

**Spec row (verbatim):**
> Env Vars (`{{ envs.xxx }}` / `{{ env('VAR') }}`): ⚠️ Pre-exec, ⚠️ Post-exec — "Resolve backend-side only (never expose raw server env to frontend)"

Note this is marked **⚠️** (resolve-with-caution), distinct from secrets which are **🔒** (always mask, never resolve).

**What the PR currently does:**
1. `env()` is wrapped in `envMaskProxy` → always returns `[env: NAME]`, never resolves.
2. `GraphService.buildDisplayVariables()` does `variables.remove("envs")` on both pre- and post-exec paths, so `{{ envs.xxx }}` has no map to read either.

So both forms named in the spec row are effectively masked. The behavior we shipped is the **🔒 secret** behavior, not the **⚠️ env** behavior — env never resolves in either phase.

**Why this happened (and why it isn't simply a bug):** the spec row is internally in tension. "Resolve backend-side" and "never expose raw server env to frontend" pull in opposite directions, because resolving e.g. `{{ env('KESTRA_DATASOURCES_POSTGRES_PASSWORD') }}` would put a credential-bearing value into `renderedProperties` and onto the frontend — which is exactly what the second half forbids. The topology view is readable by anyone with flow read access, so naive full resolution turns it into a server-env exfiltration channel. Full masking is the safe over-correction.

**Proposed resolution — an env allowlist:** the only behavior that satisfies *both* halves of the row is resolving a curated subset:
- Resolve `env('VAR')` / `envs.VAR` **only** for env names on an explicit display-safe allowlist (config-driven, e.g. `kestra.display.resolvable-env-vars`, default empty).
- Mask everything else as `[env: NAME]`, as today.
- Keep a filtered `envs` submap in the display context (instead of removing it wholesale) so `{{ envs.xxx }}` resolves for allowlisted names only.

This honors "resolve backend-side" for safe values and "never expose raw server env" for everything else, and the server-side-only `renderedProperties` architecture already guarantees names/maps never reach the client.

**Question for the spec owner:** is the allowlist the intended reading of ⚠️, or do we prefer to keep full masking and amend the row to "masked by default, allowlist-resolvable"? Either way the spec and code should be brought into agreement — right now the code is stricter than the row as written.

core/src/main/java/io/kestra/core/models/hierarchies/AbstractGraphTask.java
```diff
@@ -54,9 +57,19 @@ public String getUid() {
...
+        return this;
+    }
+
     @Override
     public AbstractGraph forExecution() {
         this.setTask(TaskForExecution.of(this.getTask()));
+        // forExecution() reduces the visible task to a safe subset for principals that only have
+        // EXECUTION-READ. The resolved properties hold the full task configuration, so they must be
+        // stripped here too, otherwise they would bypass that reduction.
+        this.renderedProperties = null;
```
jymaire, line 72: try to remove it

core/src/main/java/io/kestra/core/runners/pebble/PebbleEngineFactory.java
```diff
@@ -47,6 +61,153 @@ public PebbleEngine createWithCustomSyntax(Syntax syntax, Class<? extends Extens
...
+     *
+     * <ul>
+     *   <li>{@code secret(...)} → {@code [secret: KEY]} without invoking the real service.</li>
+     *   <li>{@code env(...)} → {@code [env: NAME]} without reading env variables.</li>
+     *   <li>Functions in {@link #SAFE_DISPLAY_FUNCTIONS} (pure parsing / calendar helpers) work normally.</li>
+     *   <li>Every other function throws {@link DisplayUnrenderableException} before invocation,
+     *       so the caller keeps the raw segment.</li>
+     * </ul>
+     */
+    public PebbleEngine createForDisplay() {
```
jymaire, line 99: createRestricted

PR #16886, opened 2026-06-18
fix(cli): reconstruct namespace file version history during metadata migration

jymaire:
## What

The `kestra migrate metadata nsfiles` command discarded namespace file version history. It filtered out every `.vN` revision object and indexed only the bare file as a single version (`version=1, last=true`).

`PurgeFiles` builds its purge list from the metadata table, so it only ever saw **one version per path** → `keepAmount: 1` left nothing to skip → **nothing was purged**, and old `.vN` revisions stayed on storage forever. This is the root cause of kestra-ee#8665.

## Fix

Group every stored object (bare file + `.vN` revisions) by logical path and replay the revisions oldest-first through `save()`, which increments the version and flags the latest as `last` — exactly the state normal sequential uploads produce.

The idempotency guard is a **version threshold** rather than a binary skip, so the same command repairs both populations:

| Install state | versions already indexed | result |
|---|---|---|
| Never migrated (future runs) | 0 | replay all → v1, v2, v3 |
| Migrated by the broken command | 1 | backfill v2, v3 |
| Fully migrated | N (max) | replay nothing (idempotent) |

Re-running `kestra migrate metadata nsfiles` is the single remediation path for already-affected installs — no new command, no config surface.

## Scope

`releases/v1.3.x` only — the migration command was removed on `develop` (2.0), so no backport is needed.

## Tests

- **Unit** `MetadataMigrationServiceTest`:
  - `shouldMigrateRevisionsAsVersionsOfTheSamePath` — all three revisions migrate under the logical path, never under a `.vN` path.
  - `shouldBackfillMissingRevisionsForAlreadyMigratedPath` — with a pre-existing v1 row, only v2/v3 are saved.
- **Integration** `NsFilesMetadataMigrationCommandTest`:
  - `shouldReconstructVersionHistoryFromStorageRevisions` — fresh migration ends at v1/v2/v3, v3 `last`, idempotent on re-run.
  - `shouldBackfillVersionsForAlreadyMigratedFile` — simulates the broken state (v1-only row + full storage history), runs the migration, asserts backfill to v1/v2/v3 and idempotency.

closes https://github.com/kestra-io/kestra-ee/issues/8665

### DIFF
--- a/cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
+++ b/cli/src/main/java/io/kestra/cli/commands/migrations/metadata/MetadataMigrationService.java
@@ -7,6 +7,7 @@ import java.nio.file.NoSuchFileException;
 import java.time.Instant;
 import java.util.*;
 import java.util.function.Function;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -34,7 +35,9 @@ import static io.kestra.core.utils.Rethrow.throwFunction;
 
 @Singleton
 public class MetadataMigrationService {
-    private static final Pattern REVISION_FILE_PATTERN = Pattern.compile(".*\\.v\\d+$");
+    // Captures the logical path (group 1) and the revision number (group 2) of a stored namespace
+    // file. Version 1 is stored under the bare path, later versions under a ".vN" suffix.
+    private static final Pattern REVISION_FILE_PATTERN = Pattern.compile("(.*)\\.v(\\d+)$");
 
     protected FlowRepositoryInterface flowRepository;
     protected TenantService tenantService;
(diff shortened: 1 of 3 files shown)


jymaire:
## 🧪 QA Report — Namespace File Version History Reconstruction

**Type:** CLI migration QA (not a plugin). Exercised the actual `migrate metadata nsfiles` command end-to-end against a live instance — not unit tests.

**Environment**
- **Edition:** Kestra OSS, built from this branch — `v1.3.23-7-g73a2dd218` (commit `73a2dd218`, the PR head)
- **Backend:** Postgres (JDBC repository + queue), local storage at `/tmp/kestra/storage/`
- **Command under test:** `kestra migrate metadata nsfiles --log-migrations`
- **Fixture:** namespace `qa.migration`, file `/script.py` uploaded 3× via the Files API → storage holds `script.py` (v1), `script.py.v2`, `script.py.v3`; a flow was added to the namespace so the migration's namespace discovery (`flowRepository.findDistinctNamespace` + system namespace) picks it up.

### Results summary

| # | Scenario | Pre-migration state | Expected | Result |
|---|----------|---------------------|----------|--------|
| A | **Full reconstruction** | metadata rows for `/script.py` deleted (un-migrated) | v1, v2, v3 rebuilt; v3 `last` | ✅ PASS |
| B | **Partial backfill** (the bug this PR fixes) | only v1 present, `last=true` (simulates the old broken run that indexed only the bare file) | v2, v3 backfilled above threshold; `last` moves v1→v3 | ✅ PASS |
| C | **Idempotency / non-regression** | fully migrated (v1–v3) | no-op, no errors, rows unchanged | ✅ PASS |
| D | **`.v1` fail-loud guard** (hardening) | invalid `script.py.v1` object planted in storage | loud abort with actionable message, no silent corruption | ✅ PASS |
| D′ | **Recovery after remediation** | offending `.v1` object removed | migration succeeds, full chain restored | ✅ PASS |

---

### Scenario A — Full reconstruction
Deleted all `/script.py` metadata rows (3), then ran the migration.

```
Migrated namespace file metadata: qa.migration - /script.py (v1)
Migrated namespace file metadata: qa.migration - /script.py (v2)
Migrated namespace file metadata: qa.migration - /script.py (v3)
✅ Namespace Files Metadata migration complete.
```
DB after:
```
/script.py v1 last=false
/script.py v2 last=false
/script.py v3 last=true
```
The **full version history** is rebuilt (not just the bare v1), and the newest revision is correctly flagged `last`.

### Scenario B — Partial backfill (core fix)
Simulated the previously-broken state: kept only **v1 with `last=true`**, deleted v2/v3.

Before:
```
/script.py v1 last=true
```
Migration output — only the missing revisions are replayed (threshold = max already-indexed version = 1):
```
Migrated namespace file metadata: qa.migration - /script.py (v2)
Migrated namespace file metadata: qa.migration - /script.py (v3)
✅ Namespace Files Metadata migration complete.
```
After:
```
/script.py v1 last=false
/script.py v2 last=false
/script.py v3 last=true
```
Exactly the reported defect: history beyond the bare file is now reconstructed, the backfill is **idempotent** (only versions above the indexed threshold are replayed), and `last` is correctly moved from v1 to v3.

### Scenario C — Idempotency / non-regression
Re-ran on the fully-migrated state:
```
✅ Namespace Files Metadata migration complete.
```
No `Migrated …` lines emitted, row count stays at 3 — safe to re-run.

### Scenario D — `.v1` fail-loud guard
Planted an invalid `script.py.v1` object in storage (collides with the bare-path v1). The migration aborts loudly instead of silently corrupting the version chain:
```
❌ Namespace Files Metadata migration failed: Storage object with '.v1' suffix found — this collides
with the bare path and would corrupt the version chain. Tenant='main' namespace='qa.migration'
path='/script.py.v1'. Remove or rename the offending object and re-run the migration.
java.lang.IllegalStateException: Storage object with '.v1' suffix found …
```
Exit code 1, no rows written. After removing the offending object (**D′**), the migration succeeds again and restores v1/v2/v3 — confirming the message's remediation instructions work.

---

### Notes
- A benign `PluginScanner` warning appears at startup (an EE `secret-vault` JAR present in the local OSS plugins dir) — unrelated to the migration, which completes normally.
- Topology / plugin-artifact checks are **not applicable** (CLI change, and Kestra < 2.0 on the `v1.3.x` line).
- Discovery caveat worth flagging for reviewers: the migration only processes namespaces returned by `flowRepository.findDistinctNamespace()` (plus the system namespace). A namespace that contains files but **no flow** is not visited — fine for real upgrades (such namespaces are rare), but worth being aware of.

**Verdict: ✅ All scenarios pass.** The fix correctly reconstructs full namespace-file version history, backfills partially-migrated paths idempotently, and fails loud on the `.v1` collision edge case.

jymaire:
see comments on issue

PR #16898, opened 2026-06-18
feat: per-taskrun logs for dynamically-generated taskruns (core API + nested UI)

brian-mulier-p:
Plugins that synthesize dynamic taskruns via `dynamicWorkerResult` (e.g. the Ansible and dbt plugins, building one taskrun per unit of work) had no way to attach logs to them, so all the per-unit output landed on the parent task's root taskrun — and the execution views rendered the dynamic taskruns flat.

Core:
- Adds `RunContext.dynamicWorkerResult(WorkerTaskResult, List<DynamicTaskRunLog>)` — registers a dynamic taskrun together with its log lines in a single call.
- The logs ride with the taskrun being registered, so they can only target it; `executionId`/`tenantId`/`namespace`/`flow` are taken from the run context (a plugin can't target another execution or tenant), the attempt is fixed to 0, and secrets are masked — plugins never build (nor can forge) a `LogEntry`.
- The default registers the taskrun and drops the logs, so runtimes that don't support per-taskrun logs (and older ones) still show the taskrun.
- Adds the `DynamicTaskRunLog(Level, String)` record and `RunContextLogger.emitDynamicTaskRunLogs`; the appender's secret masking is extracted into a shared helper. Covered by `RunContextLoggerTest`.

UI — nest dynamic taskruns under their parent (they carry `parentTaskRunId`):
- Gantt: indent child rows by depth.
- Logs: order taskruns parent → child and indent children.
- Topology "Show task logs": include the node's dynamic child taskruns so their logs show alongside the task's own.

Companion to the plugin-ansible PR that uses this API for Ansible.

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/RunContextLogger.java
+++ b/core/src/main/java/io/kestra/core/runners/RunContextLogger.java
@@ -15,6 +15,7 @@ import com.google.common.base.Throwables;
 
 import io.kestra.core.exceptions.IllegalVariableEvaluationException;
 import io.kestra.core.models.executions.LogEntry;
+import io.kestra.core.models.executions.TaskRun;
 
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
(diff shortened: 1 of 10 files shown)


anna-geller:
We discussed that there's no risk to backport it also to 1.3, as it will be visible only to EE customers who update the Ansible plugin. 

This will help our customer(s) get this much faster.

core/src/main/java/io/kestra/core/runners/RunContextLogger.java
```diff
@@ -189,6 +190,57 @@ public void emitLogs(List<LogEntry> logEntries) {
...
+     * Otherwise the lines go through a child logger bound to a {@link LogEntry} whose execution,
+     * tenant, namespace and flow are taken from this context's bound entry — a caller cannot target
+     * another execution or tenant — with the dynamic taskrun's id and a fixed attempt 0 (these
+     * taskruns have a single attempt and the log view groups by the 0-based attempt). The level
+     * filter and the secrets known to this context are inherited so nothing else is lost.
+     */
+    public void emitDynamicTaskRunLogs(TaskRun dynamicTaskRun, List<DynamicTaskRunLog> logs) {
+        if (this.logEntry == null || this.logEmitter == null || logs == null || logs.isEmpty()) {
+            return;
+        }
```
loicmathieu: This didn't seems to be needed

core/src/main/java/io/kestra/core/runners/RunContextLogger.java
```diff
@@ -189,6 +190,57 @@ public void emitLogs(List<LogEntry> logEntries) {
...
+            .build();
+
+        RunContextLogger taskRunLogger = new RunContextLogger(this.logEmitter, boundLogEntry, null, false);
+        taskRunLogger.loglevel = this.loglevel; // inherit this context's level filter as-is (logback Level)
+        taskRunLogger.useSecrets.addAll(this.useSecrets);
+
+        org.slf4j.Logger logger = taskRunLogger.logger();
+        for (DynamicTaskRunLog log : logs) {
+            logger.atLevel(log.level()).log(log.message());
+        }
```
loicmathieu: This seems overkill, you're redoing what's done in other methods.
Basically you should create a list of LogEntry then call `emitLogs` with it.

loicmathieu: Or emit them one by one as if it was done directly inside the task

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/runners/RunContextLogger.java
brian-mulier-p: I'm not really a fan of building the LogEntries within the plugin task... It opens too much rewrite doors and the plugin should not be aware of the LogEntry model imo...
The above scopes where the logs can be emitted

loicmathieu: yes, I was asking to compute them in this new methods and delegate to exisiting methods to avoid duplication

brian-mulier-p: Everything go through Appenders so the emitLogs already does that, and emitLogs is too raw, no secret masking or anything. Here I'm relying on the initializeLogger and derive logger for subtaskruns so that logs appear at the right place. I did my best to go back on the initial path so I don't think there is much duplication here. Only duplicate / weird stuff is the logToFile handling because I didn't want to create 1 log file per subtask but rather keep them in the parent task file to avoid cluttering everything.

loicmathieu: makes sense

loicmathieu (approved):
LGTM, @elevatebart can you have a quick check on the frontend part?

ui/src/components/executions/Gantt.vue
```diff
@@ -99,7 +99,10 @@
                                                 <ChevronRight v-if="!selectedTaskRuns.includes(item.id)" />
                                                 <ChevronDown v-else />
                                             </div>
-                                            <div class="task-label">
+                                            <div
+                                                class="task-label"
+                                                :style="item.depth ? {paddingLeft: `${item.depth * 1.5}rem`} : undefined"
```
elevatebart: Is there a better way to do that within the design system? @flcarre ?

ui/src/components/executions/TaskRunLine.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-    <div v-if="!hideHeader" class="taskrun-header">
+    <div
+        v-if="!hideHeader"
+        class="taskrun-header"
+        :style="depth ? {paddingLeft: `calc(1rem + ${depth * 1.5}rem)`} : undefined"
```
elevatebart: Is there a better way to do that within the design system? @flcarre ?

ui/src/components/logs/TaskRunDetails.vue
```diff
@@ -376,11 +377,41 @@
...
+                ordered.push({...node, depth})
+                const children = childrenByParent[node.id]
+                if (children) {
+                    walk(children, depth + 1)
+                }
+            }
+        }
+        walk(roots, 0)
+        return ordered
+    })
```
elevatebart: Could we give a type to TaskRunList once and for all so we avoid all this `any`?

elevatebart (approved):
Overall approved. A few small remarks that can be attacked later.

PR #16909, opened 2026-06-19
fix(executions): correct Postgres duration sort for executions over one minute

### LINKED ISSUE #16908: Postgres: sorting executions by duration is wrong for executions ≥ 1 minute (state_duration drops minutes/hours)
### What happens
On a **PostgreSQL** backend, sorting the Executions list by **Duration** does not order executions by their real duration once any execution lasts a minute or more. A 7-minute execution can appear *below* a 3-minute one. Sorting effectively happens on the *sub-minute fraction* of the duration only.

Example (sort by duration, as seen by a customer):
```
2m 59.77s
3m 59.74s
7m 59.45s   <-- 7 minutes, but sorted between two 3-minute runs
3m 59.42s
```
The order tracks the `.77 / .74 / .45 / .42` seconds-fraction, not the total duration.

### Root cause
The `executions.state_duration` generated column is defined as:
```sql
state_duration BIGINT GENERATED ALWAYS AS
  (EXTRACT(MILLISECONDS FROM PARSE_ISO8601_DURATION(value #>> '{state, duration}'))) STORED
```
`EXTRACT(MILLISECONDS FROM interval)` returns **only the seconds field of the interval × 1000** (range 0–59999); it discards minutes and hours. So a `7m59.45s` duration is stored as `59450` instead of `479450`. The UI sorts on `state_duration`, so the order is wrong for anything ≥ 1 minute.

```sql
SELECT EXTRACT(MILLISECONDS FROM '7 min 59.45 sec'::interval);              -- 59450  (bug)
SELECT (ROUND(EXTRACT(EPOCH FROM '7 min 59.45 sec'::interval) * 1000))::bigint; -- 479450 (correct)
```

### Scope
- **PostgreSQL only.** MySQL (`value ->> '$.state.duration' * 1000`) and H2 (`JQ_DOUBLE(value, '.state.duration')`) read the value as total seconds and sort correctly.
- Present on `1.0.x`, `1.3.x`, and `develop` (2.0).
- Distinct from #11593 / #13132 (which only fixed NULL durations for *running* executions). That fix is already shipped and does not address this.

### Fix
Use `EXTRACT(EPOCH FROM interval)` (total seconds across all fields) × 1000:
```sql
(ROUND(EXTRACT(EPOCH FROM PARSE_ISO8601_DURATION(value #>> '{state, duration}')) * 1000))::bigint
```
Since `state_duration` is a `STORED` generated column, the migration drops and re-creates it (backfilling existing rows) and recreates the `executions_state_duration` index.

### Affected versions
1.0.x, 1.3.x, 2.0 (develop) — PostgreSQL.

### Steps to reproduce
1. PostgreSQL backend.
2. Run flows producing executions with durations both under and over 1 minute.
3. Sort the Executions list by Duration → multi-minute executions are mis-ordered relative to sub-minute ones.

nkwiatkowski:
## What
Fix sorting executions by **Duration** on PostgreSQL. The `executions.state_duration` generated column used `EXTRACT(MILLISECONDS FROM interval)`, which returns only the seconds field of the interval (0–59999) and drops minutes/hours. A `7m59.45s` execution was stored as `59450` instead of `479450`, so duration sort was wrong for any execution ≥ 1 minute. Replaced with `EXTRACT(EPOCH FROM interval) * 1000` (total milliseconds).

PostgreSQL only — MySQL and H2 already read the value as total seconds. Distinct from #11593/#13132 (NULL duration for running execs), which is unrelated to this.

Closes #16908

## How
Edit the fresh-install baseline and add incremental migration **2.0.11** (`V2_0_11StateDurationFixMigration` + SQL) for existing installs. Column kept nullable.

`state_duration` is a `STORED` generated column, so the migration drops and re-creates it (backfilling all rows; takes a brief `ACCESS EXCLUSIVE` lock on large tables) and recreates the `executions_state_duration` index.

## Tests
Added `findShouldSortByTotalDurationAcrossMinuteBoundary` asserting a 5-minute execution sorts above a 20-second one (and vice versa). Verified passing on Postgres and H2; confirmed it fails without the fix.

---
Companion PRs: develop + releases/v1.3.x + releases/v1.0.x.


jdbc-postgres/src/main/java/io/kestra/repository/postgres/migration/V2_0_12StateDurationFixMigration.java
loicmathieu, line 1: File name should match class name

jdbc-postgres/src/main/resources/migrations/2.0.12-fix-state-duration-postgres.sql
loicmathieu, line 1: Migration name should be 11

loicmathieu (changes requested):

jdbc-postgres/src/main/java/io/kestra/repository/postgres/migration/V2_0_11StateDurationFixMigration.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * <p>
+ * The original expression used {@code EXTRACT(MILLISECONDS FROM interval)}, which only returns the
+ * seconds field of the interval and dropped minutes and hours, breaking duration sorting for
+ * executions lasting one minute or more. It is replaced with {@code EXTRACT(EPOCH FROM interval) * 1000}
+ * so the column stores the total duration in milliseconds.
+ */
+@Singleton
+@PostgresRepositoryEnabled
+public class V2_0_11StateDurationFixMigration extends AbstractSQLMigrationScript {
```
loicmathieu: Can you rename to 12 as 11 will be soon taken by plugin auto-download and I already asked him to change that 3 times ;)

loicmathieu (approved):

PR #16550, opened 2026-06-04
feat(executions): redesign the filter bar and make execution states actionable

flcarre:
## What & why

The condition-group filters (AND/OR conditions, from #16197) were too complex to actually use. Working with the designer, we reworked the **whole filter UX and UI** — that rework is what this PR is. It started from a focused issue and grew into a full redesign of the filter bar, in the shared `KsFilter` design-system component, so it lands across the list pages (Executions, Logs, Flows, Dashboards). It's a big diff (58 files).

It builds directly on the recent filter foundation already on `develop` — AND/OR condition groups (#16197), quick filters (#16311), and `KsSearch` (#16203) — and redesigns the experience on top of them.

> A small quick-win rode along because it touched the same code: execution **status badges are a dead-click target** in PostHog (users click `FAILED`/`SUCCESS` expecting to filter by that state), so we made them clickable. That's a few lines — a footnote, not the point of this PR.

A full file-by-file walkthrough is in a **collapsible comment below**.

## What changed

**Filter bar redesign (the bulk of the PR)**
- Notion-style bar: applied conditions collapse into an `N rules` pill that opens an **advanced builder** modal — conditions, AND/OR groups, per-field operators, value editors, and join-operator (`AND`/`OR`) badges so it's clear how a new entry joins. A single condition stays inline as a chip.
- **Quick state pills** (KILLED / QUEUED / FAILED / PAUSED / RUNNING) and quick **log-level** pills; the row auto-hides once filters become complex (grouped or >1 condition), where a single top-level toggle would mislead.
- **Global time scope** chip on the right — always encoded top-level, kept out of conditional groups (integrates #16476).
- `</>` **code-editor mode** — inline `filters[field][op]=value` editor with apply / clear; same height as the rest of the bar.
- **Right cluster**: saved filters, refresh, a Customize-table-columns dropdown, and a Settings dropdown (show chart, periodic refresh).
- Uniform `--ks-spacing-*` / Figma 2.0 status color + icon tokens; `Clear all` clears everything.

**Actionable execution states (the quick-win)**
- `KsExecutionStatus` gains a `clickable` prop (+ inset hover ring). Clicking a status badge filters the list by that state (`filters[state][IN]`, page reset, time range preserved); re-clicking the sole active state toggles it off.
- Wired across the executions list, Gantt, overview, task-run rows, and dashboard tables (navigates to the filtered executions list when embedded elsewhere).

**Plumbing & correctness**
[...]

PR #16919, opened 2026-06-19
feat(executions): drag-to-select time-range brush on the logs chart

### LINKED ISSUE #16603: Logs chart: drag-to-select time-range brush (Grafana/GCP/PostHog style)
## Summary

On the logs timeseries chart (log counts per date, grouped by level) shown above the log list on the global `/logs` page, let users **drag horizontally on the chart to select a time window** that filters the logs to that range — and **drag the left/right edges of an existing selection to extend or shrink it**, drag the middle to pan it, and double-click to clear. This is the time-range brush interaction familiar from **Grafana**, **Google Cloud Logging (Logs Explorer)**, and **PostHog**.

Today, narrowing the time range requires the `QuickFilters` preset segmented control (Last 24h, etc.). Direct manipulation on the chart is faster, more precise, and is the expected pattern for log/observability tooling.

Follow-up to the enriched execution logs work in https://github.com/kestra-io/kestra/pull/16580 (initiative: kestra-io/kestra-ee#8305). Frontend-only is strongly preferred, consistent with that PR (no backend changes).

## The one constraint that drives the whole design

The chart x-axis is `type: "category"` with **pre-formatted date *string* buckets** (`TimeSeries.vue` → `parseValue` → `KestraUtils.getDateFormat`), **not** a numeric `time` axis. Consequences every implementer must internalise:

- A pixel maps to a **bucket index**, not a continuous timestamp. Selection granularity = the bar bucket granularity. A drag snaps to bucket boundaries (left edge of bucket *i* → right edge of bucket *j*). This actually matches Grafana's behaviour on a bar histogram, so it's the right behaviour, not a compromise.
- Pixel→time conversion needs the **real per-bucket timestamps**, which `parseValue` currently discards (it formats to display strings). The implementation must keep a parallel `{startMs, endMs}` array per category index; derive the last bucket's width from the previous interval.
- `getDateFormat` returns lossy formats (`yyyy-MM`, `yyyy-'W'ww`) for ranges > ~180d/365d — these cannot be cleanly inverted to an instant. **v1 should disable the brush at those zoom levels** (day/hour/minute formats only) or snap to the bucket's first instant and document the imprecision.
- **Time zone:** the inverse mapper must parse a bucket label in the same TZ that produced it (`TIMEZONE_STORAGE_KEY` / `dateUtils`), or the window is off by the TZ offset. This is the most likely subtle bug — the pure mapper must take `tz` explicitly.

## Data flow (verified against the code)

The plumbing already exists for `QuickFilters` and is reused inverted:

1. zrender `mouseup` → start/end pixel X → `instance.convertFromPixel({xAxisIndex: 0}, x)` → two bucket indices → `{startMs, endMs}` → ISO `startDate`/`endDate` (via a pure mapper, in the user TZ).
2. `TimeSeries.vue` emits `@select`; `Sections.vue` forwards it (needs a small passthrough listener on its `<component :is>`); `LogsWrapper.vue` owns the URL time state.
3. `LogsWrapper` writes the route in **one** `router.push`: `normalizeRouteTimeRangeFilter(route.query, undefined)` (this already deletes every `filters[timeRange][*]`, legacy `timeRange`, `startDate`, `endDate`) then sets absolute `startDate`/`endDate` + `pageKey = 1`.
4. The `filterQueryKey` watch in `LogsWrapper` fires `dataTable.resetAndReload()` (resets to page 1, honours the controlled `:currentPage`/`:pageSize` contract). `loadQuery`'s `if (!startDate || !endDate)` guard is skipped because both are provided, so the brushed dates win.

Single push → single `filterQueryKey` change → single reload. No double-fire, no fight with the controlled pagination. Use the dynamic `pageKey`/`sizeKey` (not hardcoded `page`) so `embed` mode keeps working.

## Interaction model (gesture vocabulary)

**Idle (no selection)** — cursor `crosshair` over plot.
- **Click-drag horizontally** → creates a selection; edges snap to bucket boundaries; cursor `col-resize` while dragging; the floating range label tracks the cursor live.
- **Single click on a bar** (movement < 4px threshold) → selects exactly that one bucket (`start=bucket.startMs`, `end=bucket.endMs`). Discoverable, reversible. A 4px threshold disambiguates click vs drag.
- **Click on empty plot background** → clears any selection (same as Esc).

**Selected (active window)**
- Cursor over band body → `grab`/`grabbing`; over an edge handle (±6px hit zone) → `col-resize`; over the dimmed outside region → `crosshair` (starts a new selection, discards old — Grafana behaviour).
- **Drag a handle** → resize that edge, snap to buckets, opposite edge pinned; if dragged past the other edge, swap (don't lock).
- **Drag the band body** → pan the window preserving its duration, clamped to the data extent.
- **Double-click** → clear, revert to the active preset/default (primary fast reset).
- **Esc** (chart focused) → clear.
- A persistent clear affordance also lives in the preset row (see below), so reset is never gesture-only.

**Commit on `mouseup` only**, never on every `mousemove` — during drag only the overlay + label update (no network). Keeps the 60fps drag feel and avoids hammering `findLogs`.

## Visual design (all `--ks-*` tokens, light + dark)

- **Selection band fill:** brand primary at low alpha — `cssVar("--ks-button-primary-background", 0.12)` (or `--ks-content-link` at 0.12). Primary `#8405FF` is the system's active/selection colour and reads in both themes at low alpha.
- **Edges:** 2px solid vertical bars at `--ks-button-primary-background`, full plot height; a 12×20px rounded grip "pill" (radius `--ks-radius-base`) centred on each edge with ticks in `--ks-button-primary-content`. Hit area ±6px (wider than the visual).
- **Dimming outside:** a scrim of `cssVar("--ks-bg-base", 0.55)` over out-of-window bars — washes toward white in light mode, darkens in dark mode because the token flips (no hardcoded black/white scrim). Full-vs-dimmed contrast is the primary "selected" signal — colour-independent (satisfies "never state by colour alone").
- **Floating range label:** a `KsTag`/tooltip-styled pill anchored to the top-centre of the band, following during drag. Line 1 = `start → end` via `dateUtils` (user TZ + format prefs); line 2 = `durationUtils.humanDuration(end − start)` (e.g. "4h 30m"). Tokens: `--ks-tooltip-background`/`--ks-tooltip-content`, `--ks-radius-base`, `--ks-spacing-2`, `--ks-font-size-xs`. `pointer-events: none` during drag.

## Coexistence with QuickFilters presets (the crux)

Time state must become **one mutually-exclusive system in the URL**: either a `timeRange` preset **or** an explicit `startDate`+`endDate` window, never both (the backend already prefers explicit dates).

1. Brush commit → write `startDate`/`endDate`, **delete `timeRange`**. Preset click → write `timeRange`, **delete `startDate`/`endDate`**.
2. The `KsSegmented` preset control gains a synthetic **"Custom"** state (the `custom` i18n key already exists). When the URL holds explicit dates, `selectedTimeRange` resolves to `"CUSTOM"` so the segmented control highlights Custom instead of leaving "Last 24h" falsely lit (PostHog pattern). Custom is display-only / reflects state; clicking a real preset exits it.
3. Picking a preset clears the brush — `TimeSeries.vue` watches the resolved window and removes the overlay when the URL no longer has explicit dates (no orphan overlay).
4. The brush operates **within the currently-rendered buckets** (fetched for the active window). To widen beyond the canvas, pick a bigger preset first (Cloud Logging behaviour).

Net feel: preset = coarse/named, brush = fine/custom; the Custom chip is the visible bridge. Keep all time UI in the `QuickFilters` row + the chart — do **not** add time as a `KsFilter` facet chip (that reintroduces two competing pickers).

## Feedback & states

- **Loading after commit:** the existing `isLoading` + `filterQueryKey` watch already trigger the table refetch + `v-ks-loading` spinner — zero new plumbing. Keep the chart interactive during refetch; the band must persist from the URL (not local state), no flicker.
- **Too-small selection:** if the drag spans < 1 bucket but > the click threshold, snap up to 1 bucket; never commit `start === end`.
- **Zero results:** show the existing `KsEmpty`; keep the band so the user can widen it (don't auto-clear).
- **Live / periodic refresh:** a brushed window is **absolute and pinned** — `refresh()` re-queries the same bounds (the `startDate` computed returns `route.query.startDate` verbatim), unlike a now-relative preset. Band stays put across refreshes.

## Accessibility

- Drag is pointer-only — must not be a mouse-only island. Primary keyboard/non-pointer path: behind the "Custom" chip, open a `KsPopover` with start/end `KsDatePicker`s writing the same `startDate`/`endDate` (reuses DS components, better than fake arrow-key brushing).
- Optional stretch: chart container `tabindex=0` `role="application"` + `aria-label`; arrow keys move the focused edge by a bucket, Shift = ×10, Tab toggles edge, Esc clears; announce via `aria-live="polite"` holding the `start → end · duration` string.
- The canvas band is invisible to SR — meaning must live in DOM (the Custom chip's accessible name = the live range; `aria-live` announces commits).
- **Touch:** horizontal drag conflicts with scroll — `touch-action: pan-y` on the plot, larger ±12px edge hit targets. On small screens (`smallerOrEqual("sm")`, already used in `TimeSeries.vue`) consider disabling chart brushing and relying on presets + the date-picker popover (Cloud Logging degrades this way).

## Open implementation decision (brush component vs custom overlay)

The two planning passes diverged here — to settle during implementation:

- **Custom `getZr().on(mousedown/move/up)` + `graphic`/`markArea` overlay** (eng lens). Pros: reuses the instance-access pattern already in `TimeSeries.vue` (`getEchartsInstance()` + `instance.on("click", ...)`); `GraphicComponent` is already registered; **zero changes to the shared `KsEchart` wrapper**; full control of snap-to-bucket and custom handles. Cons: more hand-written hit-testing for edge handles.
- **ECharts native `brush` component (`brushType: 'lineX'`)** (design lens). Pros: persistent region, edge handles, `outOfBrush` dimming, `brushEnd` events for free; closest to Grafana/PostHog feel. Cons: `BrushComponent` (+ likely `ToolboxComponent`) is **not registered** in `KsEchart` → touches the design-system package and affects every chart; on a category axis `brush` returns category indices (still needs index→date); always-on/no-toolbox is fiddly.

Recommendation leans **custom overlay for v1** (no DS-wide blast radius), revisit native `brush` if/when we promote this into `KsEchart` as a reusable `brushable` feature with a real numeric time axis.

## Scope

**v1 (tracer bullet):** `selectable` prop on `TimeSeries` (default false, opt-in from the global logs chart only — **disabled for the compact `short`/`execution` chart variants**); click-drag to create; snap-to-bucket; emit `@select`; `LogsWrapper` writes `startDate`/`endDate` + clears preset; double-click to clear; brush disabled for lossy `yyyy-MM`/`yyyy-'W'ww` formats.

**v2+:** edge-drag resize handles + middle-drag pan; echo the current `startDate`/`endDate` back as a visible band on load (shared URLs show the brush); the keyboard date-picker popover; promote to a reusable `KsEchart` `brushable` feature on a numeric time axis.

## Testing

- **Unit (extract + TDD):** `bucketLabelToDateRange(label, format, tz)` (the inverse of `getDateFormat` — carries all the TZ + lossy-format risk); `pixelSelectionToBucketIndices(xStart, xEnd, axisData)` (+ clamp / right-to-left normalisation); `buildBrushQuery(routeQuery, isoStart, isoEnd, pageKey)` (asserts all `timeRange` keys cleared + page=1).
- **e2e (Playwright):** the zrender drag, overlay render, single `resetAndReload` to page 1, double-click clear, TZ correctness end-to-end (drag is not unit-testable — needs a real canvas).

## Effort

~0.5 day CC / ~3–5 dev-days human for v1 (plumbing + the pure mapper with TZ + tests + e2e); +2–3 days for v2 edge-drag handles. **Biggest unknown:** the category-axis bucket-label → instant inversion across time zones and lossy formats. If arbitrary precision becomes a hard requirement, the real fix is converting the chart to a numeric `type: "time"` x-axis — a larger change to `TimeSeries.vue` (`parsedData`/`echartsOption`/tooltip), out of scope for a frontend-only tracer bullet.

## Key files

- `ui/src/components/dashboard/sections/TimeSeries.vue` — chart; pixel↔time mapping, per-bucket `{startMs,endMs}`, overlay/handles, emit committed range.
- `ui/src/components/logs/LogsWrapper.vue` — owns URL time state; add `onBrushTimeRange`, make `onQuickFilterTimeRange` clear explicit dates, resolve `selectedTimeRange` to `"CUSTOM"`.
- `ui/src/components/filter/QuickFilters.vue` — synthetic "Custom" segmented option + keyboard date-picker popover.
- `ui/src/components/dashboard/sections/Sections.vue` — forward the chart `@select` event.
- `ui/packages/design-system/src/components/Charts/KsEchart.vue` — only if going the native-`brush` route (register `BrushComponent`).
- `ui/src/utils/utils.ts` (`getDateFormat`) — the format the inverse mapper must reverse.

---

_Issue drafted from two grounded planning passes (engineering + design lenses) over the real chart stack._

flcarre:
## Summary

- Adds a Grafana-style drag-to-select brush overlay on the global `/logs` TimeSeries chart
- Users drag horizontally to select a time window; the log list filters to that window
- Brush persists as absolute `startDate`/`endDate` in the URL (survives refresh + shared links)
- Edge-resize, body-pan, double-click/Esc/click-empty-area clear gestures all implemented
- Single click (< 4px movement) selects exactly one bucket
- `KsSegmented` preset interval control added to the `/logs` QuickFilters (was hidden before); preset and brush are mutually exclusive in the URL
- "CUSTOM" chip in the segmented control opens a date-picker popover for keyboard/non-pointer path
- `selectable` prop on `TimeSeries` is opt-in (default `false`), disabled for compact variants (`short`/`execution`) and non-invertible date formats (`yyyy-MM`, ISO week)
- Three pure utility functions (`bucketLabelToDateRange`, `pixelSelectionToBucketIndices`, `buildBrushTimeRangeQuery`) written TDD-first with 16 Vitest unit tests
- All colors via `--ks-*` tokens; no `:deep()`, no hardcoded px for theme values

closes: https://github.com/kestra-io/kestra/issues/16603

## Test plan

- [ ] `npm run check:types` passes (pre-existing errors in unrelated files only)
- [ ] `npm run test:unit` — 28 new tests pass, only 2 pre-existing `SourceSearchResults.spec.ts` failures remain
- [ ] `npm run test:lint` passes (warnings only, all pre-existing)
- [ ] `node src/translations/check.js` — no missing/extra keys (no new i18n keys added, using existing `datepicker.custom`, `from`, `to`, `save`, `cancel`)

### DIFF
--- a/ui/src/components/dashboard/sections/Sections.vue
+++ b/ui/src/components/dashboard/sections/Sections.vue
@@ -63,6 +63,8 @@
                             :dashboardId="dashboard.id"
                             :filters
                             :showDefault="props.showDefault"
+                            v-bind="chartProps(chart)"
+                            @select="(payload: any) => emit('chart-select', {chartId: chart.id, payload})"
                         />
                     </div>
                 </div>
@@ -97,12 +99,28 @@
         refreshCharts,
     })
 
+    const emit = defineEmits<{
+        "chart-select": [{chartId: string; payload: {startDate: string; endDate: string} | null}]
+    }>()
+
     const props = defineProps<{
         dashboard: Dashboard;
         charts?: Chart[];
         showDefault?: boolean;
         padding?: boolean;
+        selectableChartId?: string;
+        brushStart?: string;
+        brushEnd?: string;
     }>()
+
+    function chartProps(chart: Chart): Record<string, unknown> {
+        if (chart.id !== props.selectableChartId) return {}
+        return {
+            selectable: true,
+            brushStart: props.brushStart,
+            brushEnd: props.brushEnd,
+        }
+    }
 
     const labels = (chart: Chart) => ({
         title: getChartTitle(chart),
--- a/ui/src/components/dashboard/sections/TimeSeries.vue
+++ b/ui/src/components/dashboard/sections/TimeSeries.vue
@@ -12,14 +12,68 @@
             @toggle="onLegendToggle"
         />
 
-        <KsEchart
-            ref="ksEchartRef"
-            class="canvas"
-            :options="echartsOption"
-            :loading="false"
-            :tooltipType="TooltipType.EXTERNAL"
-            @echarts-click="onChartClick"
-        />
+        <div
+            ref="chartWrapper"
+            class="canvas-wrapper"
+            :class="{'canvas-wrapper--selectable': isSelectable}"
+        >
+            <KsEchart
+                ref="ksEchartRef"
+                class="canvas"
+                :options="echartsOption"
+                :loading="false"
+                :tooltipType="TooltipType.EXTERNAL"
+                @echarts-click="onChartClick"
+            />
+
+            <template v-if="isSelectable">
+                <div
+                    v-if="brushState.active"
+                    class="brush-scrim brush-scrim--left"
+                    :style="scrimLeftStyle"
+                />
+                <div
+                    v-if="brushState.active"
+                    class="brush-scrim brush-scrim--right"
+                    :style="scrimRightStyle"
+                />
+
+                <Motion
+                    v-if="brushState.active"
+                    as="div"
+                    class="brush-band"
+                    :layout="true"
+                    :style="bandStyle"
+                    :initial="{opacity: 0, scaleX: 0.8}"
+                    :animate="{opacity: 1, scaleX: 1}"
+                    :exit="{opacity: 0, scaleX: 0.8}"
+                    :transition="{duration: 0.15, ease: 'easeOut'}"
+                >
+                    <div
+                        class="brush-edge brush-edge--left"
+                        @mousedown.stop="startEdgeDrag('left', $event)"
+                    />
+                    <div
+                        class="brush-label"
+                        @mousedown.stop="startBandDrag($event)"
+                    >
+                        <span v-if="brushLabel">{{ brushLabel }}</span>
+                    </div>
+                    <div
+                        class="brush-edge brush-edge--right"
+                        @mousedown.stop="startEdgeDrag('right', $event)"
+                    />
+                </Motion>
+
+                <div
+                    v-if="isDragging"
+                    class="brush-drag-overlay"
+                    @mousemove="onDragMove"
+                    @mouseup="onDragEnd"
+                    @mouseleave="onDragEnd"
+                />
+            </template>
+        </div>
     </div>
     <KsNoData
         v-else-if="!props.short || (props.execution && generated?.total === 0)"
@@ -28,7 +82,7 @@
 </template>
 
 <script setup lang="ts">
-    import {computed, ref, watch} from "vue"
+    import {computed, ref, watch, onMounted, onUnmounted, nextTick} from "vue"
     import {useRoute} from "vue-router"
 
     import moment from "moment"
@@ -36,6 +90,8 @@
     import {BarChart, LineChart} from "echarts/charts"
     import {useBreakpoints, breakpointsElement} from "@vueuse/core"
     import {KsEchart, TooltipType, cssVar, durationUtils} from "@kestra-io/design-system"
+    import {Motion} from "motion-v"
+    import momentTz from "moment-timezone"
 
     import {Chart, useChartGenerator} from "../composables/useDashboards"
     import {getConsistentHEXColor, useLegendToggle} from "../composables/charts"
@@ -43,6 +99,11 @@
     import ChartLegend from "./ChartLegend.vue"
     import {getDateFormat, useTheme} from "../../../utils/utils"
     import {FilterObject} from "../../../utils/filters"
+    import {storageKeys} from "../../../utils/constants"
+    import {
+        bucketLabelToDateRange,
+        pixelSelectionToBucketIndices,
+    } from "../../../utils/logsBrushMappers"
 
     use([BarChart, LineChart])
 
@@ -57,6 +118,9 @@
         execution?: boolean;
         flow?: string;
         namespace?: string;
+        selectable?: boolean;
+        brushStart?: string;
+        brushEnd?: string;
     }>(), {
         dashboardId: undefined,
         filters: () => [],
@@ -65,7 +129,14 @@
         execution: false,
         flow: undefined,
         namespace: undefined,
+        selectable: false,
+        brushStart: undefined,
+        brushEnd: undefined,
     })
+
+    const emit = defineEmits<{
+        select: [{startDate: string; endDate: string} | null]
+    }>()
 
     const route = useRoute()
     const verticalLayout = useBreakpoints(breakpointsElement).smallerOrEqual("sm")
@@ -137,7 +208,6 @@
             const column = chartOptions?.column ?? ""
             const colorByColumn = (chartOptions as Record<string, any>)?.colorByColumn as string | undefined
 
-            // Get the fields for stacks (columns without `agg` and not the xAxis column)
             const fields = Object.keys(columns)
                 .filter(key => !aggregatorKeys.includes(key))
                 .filter(key => key !== column)
@@ -163,7 +233,6 @@
                 const current = acc[stack]
                 const parsedDate = parseValue(params[column])
 
-                // Check if the date is already processed
                 if (!current.unique.has(parsedDate)) {
                     current.unique.add(parsedDate)
                     current.data.push({
@@ -171,7 +240,6 @@
                         y: params[field],
                     })
                 } else {
-                    // Update existing stack value for the same date
                     const existing = current.data.find((v: {x: unknown; y: number}) => v.x === parsedDate)
                     if (existing) existing.y += params[field]
                 }
@@ -193,7 +261,6 @@
 
         const yDataset = reducer(rawData, aggregator.value[0][0], "y")
 
-        // Sorts the dataset array alphabetically by label for a consistent order across time ranges.
         const yDatasetData = Object.values(getData(aggregator.value[0][0], yDataset)).sort((a: any, b: any) =>
             (a.label ?? "").localeCompare(b.label ?? ""),
         )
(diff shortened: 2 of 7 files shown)


flcarre:
Merged `develop` to clear the conflicts. One part of the branch could not be carried over, so the scope is now narrower than the description above.

**Dropped: the `KsSegmented` interval control and the `CUSTOM` date-picker chip.** The filter-bar redesign (#16550, #17070) removed the `<QuickFilters>` row from the logs page — `QuickFilters.vue` now only backs the download dialog, and `hasComplexFilters` / `selectTimeRange` are gone from `LogsWrapper`. Re-adding that row would have reverted part of the redesign, and leaving the `CUSTOM` chip on the download-dialog instance would have shipped a picker whose `update:customDates` nothing handles. So `QuickFilters.vue` and its spec are back to their `develop` state, and `LogsWrapper` keeps only the brush wiring.

**Kept:** the brush itself — `TimeSeries.vue` overlay, the `selectable` / `brushStart` / `brushEnd` plumbing through `Sections.vue`, `onChartSelect` in `LogsWrapper`, and `logsBrushMappers` with its 16 unit tests. The brush still writes absolute `startDate` / `endDate` into the URL, which `develop`'s `loadData` reads unchanged, so the feature works as described.

The keyboard/non-pointer path the `CUSTOM` chip provided is now the filter bar's own time-range filter rather than a chip next to the chart. If you would rather have the chip back, that is a call about the filter bar's design, not something the merge could decide.

Other conflicts resolved: `TimeSeries.vue` keeps the brush overlay around `develop`'s `KsEchart` props (`maxPixelRatio`, `stickyTooltip`) and drops the `FilterObject` import that `develop` replaced with `QueryFilter`; `Sections.vue` keeps both the selectable-chart props and `develop`'s lazy chart blocks.

Verified on the merge result: `check:types` clean, `npm run build` clean, `oxlint` and `eslint` clean on the touched files, and `logsBrushMappers.test.ts` + `QuickFilters.spec.ts` pass (23 tests).

---

PR #16923, opened 2026-06-22
feat(executions): propagate configured execution labels to the logging MDC

### LINKED ISSUE #16338: Propagate flow execution labels to MDC context
### Feature description

With https://github.com/kestra-io/kestra/pull/16067 now merged 🫶 , it would be really nice to have flow execution labels included as part of log metadata. This is specially interesting for using the `system.correlationId` label so we can easily trace logs based on correlationId. This will allow people to query for logs related to an entire parent execution (since the correlationId is always the execution Id of the parent / triggering execution) and alternatively also based on any other labels.

Right now, its hard to filter logs based on the entire execution pipeline because log metadata only contains information regarding the executionId of a specific flow or task. When teams are using multiple levels of subflows, its hard to correlate all the executionIds across the full scope of the execution.

manerow:
Closes #16338

### What

Adds an opt-in `kestra.logging.labels` allow-list. When configured, the matching execution labels are propagated into the SLF4J **MDC** of execution and task logs, so forwarded logs can be filtered by business labels (e.g. `team`, `env`) in the logging backend.

```yaml
kestra:
  logging:
    labels:
      - team
      - env
```

### Why

Per #16338, only `executionId`/`flowId`/etc. reach log metadata today, so logs can't be correlated by label. As discussed in the issue, this is implemented as a **config-driven allow-list with no defaults**, mirroring `kestra.metrics.labels`, to keep the MDC memory footprint and backend cardinality under control.

### How

- New `LoggingConfiguration` (`kestra.logging.labels`), empty by default.
- `RunContextLoggerFactory` filters execution labels to the allow-listed keys and passes them to `RunContextLogger`.
- `RunContextLogger` merges the filtered labels with the existing `LogEntry` MDC context (LogEntry keys win on collision).
- Labels are carried to the worker via a new `labels` field on `WorkerTask`.

### Scope / notes

- **MDC only** - logs are not persisted with labels (no `logs` table column). Out of scope here.
- Covers the primary execution and task-dispatch paths. Working-directory nested tasks and liveness resubmissions don't yet carry labels (the `Execution` isn't in scope at those build sites); can follow up if wanted.
- Trace-context-to-logs (`trace_id`/`span_id`) is intentionally a **separate PR**, per maintainer feedback on the issue.

### Tests

- `RunContextLoggerLabelsTest` - asserts only allow-listed labels reach the MDC, and that nothing is propagated when unconfigured.
- `H2RunnerTest` green (executor touched).

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/RunContextLogger.java
+++ b/core/src/main/java/io/kestra/core/runners/RunContextLogger.java
@@ -42,6 +42,7 @@ public class RunContextLogger implements Supplier<org.slf4j.Logger> {
     private Level loglevel;
     private final List<String> useSecrets = new ArrayList<>();
     private final boolean logToFile;
+    private final Map<String, String> mdcLabels;
 
     @Getter
     private File logFile;
@@ -51,9 +52,14 @@ public class RunContextLogger implements Supplier<org.slf4j.Logger> {
     public RunContextLogger() {
         this.loggerName = "unit-test";
         this.logToFile = false;
+        this.mdcLabels = Map.of();
     }
 
     public RunContextLogger(LogEntryEmitter logEmitter, LogEntry logEntry, org.slf4j.event.Level loglevel, boolean logToFile) {
+        this(logEmitter, logEntry, loglevel, logToFile, Map.of());
+    }
+
+    public RunContextLogger(LogEntryEmitter logEmitter, LogEntry logEntry, org.slf4j.event.Level loglevel, boolean logToFile, Map<String, String> mdcLabels) {
         if (logEntry.getTaskId() != null) {
             this.loggerName = baseLoggerName(logEntry) + "." + logEntry.getTaskId();
         } else if (logEntry.getTriggerId() != null) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/kestra/core/runners/WorkerTask.java
```diff
@@ -35,6 +38,9 @@ public class WorkerTask extends WorkerJob {
     @Nullable
     private ExecutionKind executionKind;
 
+    @Nullable
+    private List<Label> labels;
```
loicmathieu: I'm not sure this is a good idea as we want to keep our messages as light as possible.

@fhussonnois WDYT?

manerow: Good point, agreed. The labels are already on the wire: `RunVariables` puts them into `WorkerTaskData.variables` (the `labels` variable used by `{{ labels.* }}`), and that key isn't stripped for workers, so the dedicated field was redundant.

I dropped it and now read the labels from `workerTask.getData().variables()` in `RunContextLoggerFactory`, so the worker message size is unchanged. It also covers the working-directory and resubmission paths the explicit field missed.

loicmathieu (approved):
LGTM, thanks

PR #16924, opened 2026-06-22
feat(executions): add OpenTelemetry trace context to log MDC

manerow:
Follow-up to #16338 (kept separate from the labels PR #16923, per maintainer feedback).

### What

Surfaces the active OpenTelemetry `trace_id` and `span_id` into the SLF4J **MDC** of execution and task logs. This lets logs be correlated with the traces Kestra already produces (trace-to-logs in Grafana, Datadog, Jaeger, ...).

### Why

Even with OTel tracing enabled, logs currently carry no trace identifier, so they can't be tied back to spans. Kestra already starts a span around execution handling (executor) and task execution (worker), so the active span is available when logs are emitted.

### How

`LogEntry.toMap()` (the source of the log MDC) now reads `Span.current().getSpanContext()` and adds `trace_id`/`span_id` when the context is valid.

- **On by default when OTel tracing is enabled** - no extra config, as requested.
- **No-op when tracing is disabled** - `SpanContext.isValid()` is `false`, so nothing is added and behaviour is unchanged.

### Note on the "standard Micronaut way"

A standard `opentelemetry-logback-mdc` instrumentation exists, but it only covers the server's global logback context. Execution/task logs use a dedicated per-execution `LoggerContext` and set an explicit MDC map, so the standard instrumentation can't reach them - hence injecting the context at the `LogEntry.toMap()` chokepoint. Adding the instrumentation + a `%X{trace_id}` logback pattern for the server's own daemon logs could be a separate follow-up.

### Tests

- `LogEntryTraceContextTest` - using the OpenTelemetry SDK testing extension, asserts `trace_id`/`span_id` are present in the MDC within an active span and absent without one.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/executions/LogEntry.java
+++ b/core/src/main/java/io/kestra/core/models/executions/LogEntry.java
@@ -17,6 +17,8 @@ import io.kestra.core.models.triggers.TriggerId;
 import io.kestra.core.queues.event.DispatchEvent;
 import io.kestra.core.utils.IdUtils;
 
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanContext;
 import io.swagger.v3.oas.annotations.Hidden;
 import jakarta.annotation.Nullable;
 import jakarta.validation.constraints.NotNull;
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
LGTM

PR #16928, opened 2026-06-22
fix(dashboards): cap bar chart height and add vertical scroll for many namespaces

MilosPaunovic:
## Summary
- Caps the executions-per-namespace bar chart at 500px height
- Wraps the chart in a scrollable container (`overflow-y: auto`) so all namespaces remain accessible when the list is long
- Natural height calculation is unchanged — charts with few namespaces render at their normal size

Closes https://github.com/kestra-io/kestra-ee/issues/8626.


Piyush-r-bhaskar (approved):

MilosPaunovic:
Closes https://github.com/kestra-io/kestra-ee/issues/8626.

PR #16954, opened 2026-06-22
docs(flows): document Check 'when' boolean and eval-error semantics

nkwiatkowski:
## What

Adds javadoc to `Check.when` (`core/.../models/flows/check/Check.java`) documenting two intentional-but-undocumented behaviors of flow Checks, surfaced during 2.0 QA:

- The expression must render to a **real boolean** `true`/`false`. A value that renders to anything else (the string `"true"`/`"yes"`, a number, …) is treated as a **failed** check — evaluation uses a strict `Boolean.TRUE.equals(...)` in `FlowService.getFailedChecks`.
- If the expression **cannot be evaluated** (undefined variable, syntax error), the check fails safe to `BLOCK_EXECUTION` + `ERROR`, overriding the declared `behavior`/`style` — a broken gate must never let a run through.

No behavior change — javadoc only. Companion user-facing docs PR: kestra-io/docs (Checks page).

Refs kestra-io/kestra-ee#8711, part of kestra-io/kestra-ee#8184.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/check/Check.java
+++ b/core/src/main/java/io/kestra/core/models/flows/check/Check.java
@@ -29,6 +29,15 @@ public class Check {
 
     /**
      * The condition to evaluate.
+     * <p>
+     * The expression must render to a real boolean {@code true} or {@code false}. A value that renders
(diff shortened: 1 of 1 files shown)


loicmathieu (approved):
LGTM

PR #16955, opened 2026-06-22
docs(flows): document Check 'condition' boolean and eval-error semantics (1.3.x)

nkwiatkowski:
## What

Backport of #16954 to `releases/v1.3.x`. Adds javadoc to `Check.condition` (named `condition` on this branch; `when` on develop) documenting two intentional behaviors of flow Checks, surfaced during 2.0 QA:

- The expression must render to a **real boolean**; a non-boolean (string `"true"`/`"yes"`, a number, …) is treated as a **failed** check (strict `Boolean.TRUE.equals(...)`).
- An **un-evaluatable** expression fails safe to `BLOCK_EXECUTION` + `ERROR`, overriding the declared `behavior`/`style`.

No behavior change — javadoc only.

Refs kestra-io/kestra-ee#8711, part of kestra-io/kestra-ee#8184.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/check/Check.java
+++ b/core/src/main/java/io/kestra/core/models/flows/check/Check.java
@@ -28,6 +28,15 @@ public class Check {
 
     /**
      * The condition to evaluate.
+     * <p>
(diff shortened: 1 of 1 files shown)


loicmathieu (approved):
LGTM

PR #16956, opened 2026-06-22
feat(execution): Quotas

loicmathieu:


core/src/main/java/io/kestra/core/models/flows/quota/Quota.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Represents a quota for a flow.
+ * The duration is the identifier.
+ */
+@SuperBuilder
+@Getter
+@NoArgsConstructor
+public class Quota {
+    @NotNull
+    private Duration duration;
```
AcevedoR, line 22: maybe missing a validation that this should not be 0

core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
```diff
@@ -144,6 +144,8 @@ public class MetricRegistry {
     public static final String METRIC_EXECUTOR_SLA_MONITOR_LOOP_DURATION_DESCRIPTION = "SLA monitor loop duration inside the Executor";
     public static final String METRIC_EXECUTOR_EXECUTION_DELAY_LOOP_DURATION = "executor.loop.delay.duration";
     public static final String METRIC_EXECUTOR_EXECUTION_DELAY_LOOP_DURATION_DESCRIPTION = "Execution delay loop duration inside the Executor";
+    public static final String METRIC_EXECUTOR_QUOTA_EXCEEDED_COUNT = "executor.quota.exceeded.count";
```
AcevedoR: this metrics should be suffixed with TOTAL and not COUNT

see the doc at https://github.com/kestra-io/kestra/blob/develop/docs/architecture/METRICS_GUIDELINES.md

executor/src/test/java/io/kestra/executor/handler/ExecutionEventMessageHandlerTest.java
```diff
@@ -135,4 +146,48 @@ void shouldNotApplyKillActionWhenExecutionIsAlreadyKilling() {
 
         verify(killSwitchActionService, never()).handle(any(), any(), any());
     }
+
+    @Test
+    void shouldFailExecutionWhenQuotaExceededWithFailBehavior() {
```
AcevedoR, line 151: It seems to lack multiple tests about quotas vs concurrency limits.
I know it's hard in this code area to have such tests, but we need to find a way, I will try something

AcevedoR (changes requested):

core/src/main/java/io/kestra/core/models/flows/quota/Quota.java
loicmathieu: `@Positive` didn't work for duration, I need to have a look.
If I'll add validation, I would restrict to at least 1mn I think.

core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
loicmathieu: done

core/src/main/java/io/kestra/core/models/flows/quota/Quota.java
loicmathieu: done

executor/src/test/java/io/kestra/executor/handler/ExecutionEventMessageHandlerTest.java
loicmathieu: I'll add that in a followup PR

AcevedoR (approved):

PR #16957, opened 2026-06-22
fix(storage): return 4xx for storage/authorization failures instead of 500

### LINKED ISSUE #16897: Storage endpoints: storage/authorization failures return HTTP 500 (one leaks the absolute FS path) instead of 4xx
Two related error-handling problems on the storage / file endpoints: storage failures surface as HTTP **500** (instead of a clean 4xx) and one of them **leaks the absolute internal-storage filesystem path** in the response body.

## 1. Over-long filename → 500 leaking the absolute storage base path
Uploading a namespace file whose name exceeds the filesystem component limit (255 bytes on ext4) raises the OS `ENAMETOOLONG` which propagates raw to the client.

```bash
B=http://localhost:8080/api/v1/main; A=(-u 'demo:DemoDemo1!'); NS=io.kestra.qa
LONG=$(printf 'x%.0s' {1..300})
printf data > /tmp/d
curl -s "${A[@]}" -X POST "$B/namespaces/$NS/files?path=/data/$LONG.txt" -F fileContent=@/tmp/d
```
Response (HTTP 500):
```
Internal server error: /home/<user>/workspace/.../main/io/kestra/qa/_files/data/xxxxxx…xxx.txt
```
Two problems:
- **Information disclosure**: the body reveals the server's absolute internal-storage filesystem path.
- **Wrong status / no graceful handling**: an over-long name should be a clean `4xx`. `StorageInterface#limit(URI, int)` (`core/.../storages/StorageInterface.java:453`, max 255) is designed to truncate names but is not applied on the namespace-file `put` path.

## 2. `file://` scheme: authorization denial returns 500 instead of 4xx
On the execution file endpoint, requesting a `file://` path that is not in the allow-list is correctly **blocked** (content is *not* served), but the denial is returned as a 500.

```bash
curl -s "${A[@]}" "$B/executions/<execId>/file?path=file%3A%2F%2F%2Fetc%2Fpasswd"
```
Response (HTTP 500):
```
Internal server error: The path /etc/passwd is not authorized. Path must be allowed
either globally inside the Kestra configuration using the `kestra.local-files.allowed-paths` property.
```
No data is exposed (the allow-list gate works), but an authorization failure should be `403`/`422`, not `500`. Source: `ExecutionController#validateFile` / `downloadFileFromExecution` (`webserver/.../ExecutionController.java:846`, `:905`).

## Suggested fix
- Map storage/authorization failures to appropriate 4xx status codes and strip absolute filesystem paths from client-facing error bodies.
- Apply `StorageInterface#limit(...)` (or validate name length) before writing a namespace file.

## Impact
Low/medium — primarily error-handling and information disclosure; no unauthorized data access observed. The FS-path leak in (1) is the same class as a similar leak seen in backup/restore error messages.

## Environment
Kestra 2.0 EE, `develop` @ `6075c7fb2`, ELS+Kafka backend, `local` storage (`LocalStorage`). Found during 2.0 QA of the Internal Storage interface.

nkwiatkowski:
## What

Fixes #16897 — two storage-endpoint error-handling problems found during 2.0 QA (`kind/security`):

1. **Over-long namespace filename → 500 leaking the absolute FS path.** Uploading a file whose name exceeds the filesystem limit raised `ENAMETOOLONG`, which surfaced as a 500 whose body contained the server's absolute internal-storage path (`/home/.../_files/...`). Now rejected **before** writing with a clean **422** ("File name exceeds the maximum length of 255 characters.") — no path leaked.
2. **`file://` allow-list denial → 500.** A denied `file://` path threw a `SecurityException` with no handler, surfacing as 500. Added a `SecurityExceptionHandler` mapping `SecurityException` → **403** (also covers "Local file preview is disabled").

## Changes
- `NamespaceFileController.putNamespaceFile`: guard the file-name length (≤ 255) → `IllegalArgumentException` (mapped to 422 by the existing handler).
- New `webserver/.../exceptions/SecurityExceptionHandler` → 403.

## Test
`NamespaceFileControllerTest.createFileWithTooLongNameReturnsCleanError`: over-long name → 422, body contains "maximum length" and **not** the internal `_files` path nor "Internal server error". Passes on H2 locally.

## Backports
Present in `releases/v1.0.x` and `releases/v1.3.x` — companion backport PRs follow.

Fixes #16897 · found via kestra-io/kestra-ee#8184

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/NamespaceFileController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/NamespaceFileController.java
@@ -48,6 +48,10 @@ import static io.kestra.core.utils.Rethrow.throwConsumer;
 @Controller("/api/v1/{tenant}/namespaces")
 public class NamespaceFileController {
     public static final String FLOWS_FOLDER = "_flows";
+
+    // Maximum length of a single file-name component on common filesystems (e.g. 255 bytes on ext4).
+    private static final int MAX_FILE_NAME_LENGTH = 255;
+
     @Inject
     private StorageInterface storageInterface;
     @Inject
@@ -240,6 +244,16 @@ public class NamespaceFileController {
             return Collections.emptyList();
         }
         forbiddenPathsGuard(path);
+
+        // Reject over-long names before writing: otherwise the filesystem raises ENAMETOOLONG, which
+        // surfaces as a 500 leaking the absolute internal-storage path. The limit is per path component
+        // (255 bytes on common filesystems), so we check each segment rather than the whole path — a
+        // valid multi-component path must not be rejected. Return a clean 4xx instead.
+        for (Path component : Path.of(filePath)) {
+            if (component.toString().length() > MAX_FILE_NAME_LENGTH) {
+                throw new IllegalArgumentException("A file or folder name exceeds the maximum length of " + MAX_FILE_NAME_LENGTH + " characters.");
+            }
+        }
 
         Namespace namespaceStorage = namespaceFactory.of(tenantId, namespace, storageInterface);
         return namespaceStorage.putFile(Path.of(path.getPath()), inputStream);
--- a/webserver/src/main/java/io/kestra/webserver/exceptions/SecurityExceptionHandler.java
+++ b/webserver/src/main/java/io/kestra/webserver/exceptions/SecurityExceptionHandler.java
@@ -1,0 +1,26 @@
+package io.kestra.webserver.exceptions;
+
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.http.HttpResponse;
+import io.micronaut.http.HttpStatus;
+import io.micronaut.http.MediaType;
+import io.micronaut.http.annotation.Produces;
+import io.micronaut.http.server.exceptions.ExceptionHandler;
+import jakarta.inject.Singleton;
+
+/**
+ * Maps a {@link SecurityException} to a clean {@code 403 FORBIDDEN} instead of letting it surface as a
+ * {@code 500}. This covers authorization denials such as a disabled local-file preview or a path that is
+ * not in the {@code kestra.local-files.allowed-paths} allow-list.
+ */
+@Produces(value = MediaType.TEXT_PLAIN)
+@Singleton
+@Requires(classes = { SecurityException.class, ExceptionHandler.class })
+@SuppressWarnings("rawtypes")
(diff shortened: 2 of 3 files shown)


webserver/src/main/java/io/kestra/webserver/controllers/api/NamespaceFileController.java
```diff
@@ -241,6 +245,13 @@ private List<NamespaceFile> putNamespaceFile(String tenantId, String namespace,
         }
         forbiddenPathsGuard(path);
 
+        // Reject over-long file names before writing: otherwise the filesystem raises ENAMETOOLONG,
+        // which surfaces as a 500 leaking the absolute internal-storage path. Return a clean 4xx instead.
+        String fileName = Path.of(filePath).getFileName().toString();
```
loicmathieu: He you're checking the full filename not each components right?
So a fileName with two components of 200 chars will be prohibited even if it would success creation

webserver/src/main/java/io/kestra/webserver/exceptions/SecurityExceptionHandler.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Maps a {@link SecurityException} to a clean {@code 403 FORBIDDEN} instead of letting it surface as a
+ * {@code 500}. This covers authorization denials such as a disabled local-file preview or a path that is
+ * not in the {@code kestra.local-files.allowed-paths} allow-list.
+ */
+@Produces(value = MediaType.TEXT_PLAIN)
+@Singleton
+@Requires(classes = { SecurityException.class, ExceptionHandler.class })
+@SuppressWarnings("rawtypes")
+public class SecurityExceptionHandler implements ExceptionHandler<SecurityException, HttpResponse> {
```
loicmathieu, line 21: Is it really needed? This is a custom exception and I don't think it's necessaryly leaking sensitive information.
Can you give me more details about it?

loicmathieu (changes requested):

webserver/src/main/java/io/kestra/webserver/exceptions/SecurityExceptionHandler.java
nkwiatkowski: It's not about information leaking, it's about changing the response status: 403 instead of the 500 without the handler

loicmathieu: OK then

loicmathieu (approved):
LGTM

PR #16974, opened 2026-06-23
fix: workerTaskRestartStrategy IMMEDIATELY not respected for OOMKilled workers

### LINKED ISSUE #16970: workerTaskRestartStrategy is not being respected
### Describe the issue

When defining `workerTaskRestartStrategy: IMMEDIATELY` in the config, Kestra is still waiting for the time period configured in `terminationGracePeriod` (in my case 15 minutes), to reschedule executions to the new workers.

It seems the default `workerTaskRestartStrategy: AFTER_TERMINATION_GRACE_PERIOD` is being used regardless.

Image

This graph shows the moment the single worker we have in this cluster died due to OOMKill and exactly 15 minutes later the executions were retriggered on the new worker allocation. 

### Environment

- Kestra Version: 1.3.21

Palak24Ol:
Fixes #16970

## Root Cause
When a worker is OOMKilled it transitions directly to `TERMINATED_FORCED` state.

- `handleAllNonRespondingServices()` — correctly handles `IMMEDIATELY` but only processes workers in RUNNING state, so it never sees a `TERMINATED_FORCED` worker.
- `handleAllWorkersForUncleanShutdown()` — is the only path triggered, but its `isUncleanShutdownService()` always waited for `terminationGracePeriod` regardless of the configured strategy.

This caused `IMMEDIATELY` to be completely ignored, and the default `AFTER_TERMINATION_GRACE_PERIOD` behavior was always used.

## Fix
Added an `isImmediately` check in `isUncleanShutdownService()` to skip the grace period wait when `workerTaskRestartStrategy` is `IMMEDIATELY`. Minimal surgical change — only the logic in one private method was modified.


fhussonnois:
Hi @Palak24Ol, thank you for your PR. Unfortunately, I don't think the fix is correct. 

The `TERMINATED_FORCED` state is set by the worker itself during its graceful-shutdown hook when the job executor fails to drain before the deadline. For for an OOMKill (i.e., a SIGKILL) the worker will not transition to `TERMINATED_FORCED` (no more JVM) .  It just stops heartbeating, and the executor marks it `RUNNING` -->
  `DISCONNECTED` via the non-responding path... which already handles `IMMEDIATELY` strategy.
  
The current PR will introduce a duplicate resubmission risk on the `DISCONNECTED` branch (i.e., `handleAllNonRespondingServices`) because the non-responding path already re-emitted jobs for an `IMMEDIATELY` worker, 

In addition, you'vre removes intentional safety on `TERMINATED_FORCED`. The grace-period wait there is deliberate to ensure that all in-flight task runs had enough time to be fully handled by the executors. Short-circuiting it with `IMMEDIATELY` reintroduces exactly the duplicate-execution risk that wait was protecting against.

loicmathieu:
@Palak24Ol we didn't hear for you since 3 weeks? Are you willing to take into account the feedback from @fhussonnois ? Otherwise I'll close the PR.

Thanks anyway for your work.

loicmathieu:
Hearing no feedback since more than one month from the author, I'm closing it.

@Palak24Ol if you still want to contribute this, please take into account the feedback of Florian and re-open the PR.

PR #16977, opened 2026-06-23
fix(core): smooth right panel open/close transition matching the left sidebar

MilosPaunovic:
## Summary

- Replaces the Vue `<Transition>` (v-if mount/unmount) with a CSS class-toggle approach, identical to how the left sidebar animates
- The previous approach caused a visible snap: `KsSplitter` uses `useElementSize` (ResizeObserver) which reports size `0` on first mount and fires asynchronously — creating a layout jump during the enter transition
- Keeping the splitter always mounted lets it measure once and stabilize; the `is-closed` class toggles `width: 0` and animates via CSS, matching the left sidebar's easing curve (`cubic-bezier(0.22, 1, 0.36, 1)`, 0.32s)

## Test plan

- [ ] Open/close the right panel (News, Docs, Help buttons) — should animate smoothly like the left sidebar
- [ ] Open/close the left sidebar — confirm both transitions feel identical
- [ ] Resize the right panel with the drag handle — should still work
- [ ] Verify `prefers-reduced-motion` disables the transition

Closes https://github.com/kestra-io/kestra-ee/issues/8763.

### DIFF
--- a/ui/src/components/ContextDrawer.vue
+++ b/ui/src/components/ContextDrawer.vue
@@ -1,53 +1,51 @@
 <template>
-    <Transition name="drawer">
-        <div v-if="hasButtons && activeTab" class="contextDrawer" :style="{'--drawer-width': `${drawerWidth}px`}">
-            <KsSplitter
-                class="drawerSplitter"
-                :style="{width: `${maxDrawerWidth}px`}"
-            >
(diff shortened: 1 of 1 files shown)


Pradumnasaraf (approved):
Looks good!

Issue #16982, opened 2026-06-23
Native/off-heap memory grows over time from SSE follow-stream connection churn (Netty direct buffers)


nkwiatkowski:
## Describe the bug

On long-running standalone deployments, the pod's resident memory (RSS / `container_memory_working_set_bytes`) grows steadily over days and can trip the container memory limit, even though the **JVM heap is healthy** — it sawtooths and garbage-collects cleanly, and heap dumps show nothing abnormal. The growth correlates with UI usage (the executions detail and logs views that live-follow over Server-Sent Events).

The growth is **off-heap**: `jvm_memory_used_bytes{area="heap"}` stays flat while `container_memory_working_set_bytes` climbs. `MALLOC_ARENA_MAX=2` reduces but does not stop it (it constrains glibc arenas, not Netty's own pooled direct arenas).

## Root cause

Three compounding factors in the SSE live-follow path:

**1. The UI `EventSource` is never closed on error.** The browser's `EventSource` auto-reconnects (~every 3s) on any error unless `.close()` is called. The UI `onerror` handlers either surface a toast without closing, or have no `onerror` at all:

- `ui/src/stores/executions.ts` — `followExecution` `onerror` (line 493) sets a "connection lost" message but never calls `closeSSE()`.
- `ui/src/components/executions/Logs.vue` — `streamLogs` (lines 274–291) sets `onmessage` but **no `onerror`**, so a log-stream error falls through to the browser's default silent infinite reconnect.
- `ui/src/components/dependencies/composables/useDependencies.ts` — `openSSE` `onerror` (line 552) sets a message but never calls `closeSSE()`.

Any stream error (the 1h server timeout firing, a proxy idle-cut, a transient network blip, an abandoned tab) therefore makes the browser silently reconnect forever, each reconnect opening a **fresh server-side SSE connection** (new subscriber, new `FluxSink`, new Netty channel).

**2. No direct-memory ceiling, and Netty's pool never trims.** No `-XX:MaxDirectMemorySize` / `io.netty.maxDirectMemory` is configured (Docker image, Helm chart, `application.yml`), so the off-heap ceiling defaults to ≈ max heap. Each SSE response is written through Netty's `PooledByteBufAllocator` (direct buffers); under sustained connection churn the pool grows to its high-water mark, fragments, and does not return memory to the OS → monotonic RSS growth invisible to heap tooling.

[...]

PR #16984, opened 2026-06-23
fix(executions): close SSE streams on error to stop off-heap leak

nkwiatkowski:
## What

The UI live-follow streams (execution, logs, dependencies) use the native `EventSource`, which **auto-reconnects (~every 3s) on any error unless explicitly closed**. The `onerror` handlers showed a "connection lost" toast but never called `close()`, and the logs stream had no `onerror` at all. So any stream error (the 1h server-side timeout firing, a proxy idle-cut, a transient network blip, an abandoned tab) made the browser silently reconnect forever. Each reconnect opens a **fresh server-side SSE connection** whose Netty pooled direct buffers are not promptly reclaimed → off-heap memory (pod RSS) grows over time while the JVM heap stays healthy.

This closes the `EventSource` in all three `onerror` handlers so an error is terminal instead of triggering unbounded reconnect churn:

- `ui/src/stores/executions.ts` — `followExecution` `onerror` now calls `closeSSE()`.
- `ui/src/components/executions/Logs.vue` — `streamLogs` now sets an `onerror` that calls `closeLogsSSE()`.
- `ui/src/components/dependencies/composables/useDependencies.ts` — `openSSE` `onerror` now calls `closeSSE()`.

## Why

This is the highest-leverage, lowest-risk part of #16982 — it stops the client from generating unbounded server-side connection churn, which is what turns UI usage into monotonic RSS growth. The existing code already treats `onerror` as terminal (it shows a "connection lost" message); this completes that intent.

## Notes / follow-ups (tracked in #16982)

- Server-side: proactively complete idle terminal-execution follow streams and bound the `OverflowStrategy.BUFFER` sinks.
- Defaults: ship a sane `-XX:MaxDirectMemorySize` / `io.netty.maxDirectMemory` in the Docker image / Helm chart.
- Operator workaround in the meantime: set `-XX:MaxDirectMemorySize` to cap off-heap and confirm via `jvm_buffer_memory_used_bytes{id="direct"}`.

## Test

- `npm run check:types` ✅
- `eslint` ✅ (pre-commit lint-staged passed)

Refs #16982

### DIFF
--- a/ui/src/components/dependencies/composables/useDependencies.ts
+++ b/ui/src/components/dependencies/composables/useDependencies.ts
@@ -555,6 +555,11 @@ export function useDependencies(
                 title:   t("error"),
                 message: t("something_went_wrong.loading_execution"),
             }
+
+            // Close on error: EventSource auto-reconnects unless explicitly closed,
+            // and each reconnect leaks a server-side SSE connection (Netty direct
+            // buffers) over time. See kestra-io/kestra#16982.
+            closeSSE()
         }
     }
 
--- a/ui/src/components/executions/Logs.vue
+++ b/ui/src/components/executions/Logs.vue
@@ -287,6 +287,13 @@
                     sseFlushTimer = setTimeout(flushSseBuffer, 200)
                 }
             }
+            // Close on error: without this, EventSource auto-reconnects (~every 3s)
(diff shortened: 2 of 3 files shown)


loicmathieu (approved):
LGTM

PR #16987, opened 2026-06-23
fix(executions): close SSE streams on error to stop off-heap leak (1.0.x)

nkwiatkowski:
## What

Backport of #16984 to `releases/v1.0.x`.

The UI live-follow streams use the native `EventSource`, which **auto-reconnects (~every 3s) on any error unless explicitly closed**. The `onerror` handlers showed a "connection lost" toast but never called `close()`, and some streams had no `onerror` at all. So any stream error (the 1h server-side timeout firing, a proxy idle-cut, a transient network blip, an abandoned tab) made the browser silently reconnect forever. Each reconnect opens a fresh server-side SSE connection whose Netty pooled direct buffers are not promptly reclaimed → off-heap memory (pod RSS) grows over time while the JVM heap stays healthy.

This closes the `EventSource` on every error path:

- `ui/src/stores/executions.ts` — `followExecution` `onerror` now calls `closeSSE()`.
- `ui/src/components/logs/TaskRunDetails.vue` — `followLogs` `onerror` now calls `closeLogsSSE()`; the target-execution (`rawSSE`) follow gains an `onerror` that calls `closeTargetExecutionSSE()`.
- `ui/src/components/dependencies/composables/useDependencies.ts` — `openSSE` `onerror` now calls `closeSSE()`.

Consumers that follow through the store's managed stream (`Topology.vue`, `FlowPlayground.vue`, `ExecutionRoot.vue`) do not override `onerror`, so they inherit the store-level fix automatically.

## Why

This is the off-heap memory growth some deployments see: pod RSS climbs over days, correlated with UI usage, while the JVM heap stays healthy. Stopping the client's unbounded reconnect churn is the highest-leverage, lowest-risk mitigation.

## Operator workaround (in the meantime)

Set `-XX:MaxDirectMemorySize=<cap, e.g. 512m>` to bound off-heap memory; confirm via `jvm_buffer_memory_used_bytes{id="direct"}` vs RSS. Keep `MALLOC_ARENA_MAX=2` and `-XX:ActiveProcessorCount=<pod CPU limit>`.

## Test

- `eslint` ✅ (pre-commit lint-staged passed; 0 errors)
- (`check:types` is a no-op on this release line)

Refs #16982

### DIFF
--- a/ui/src/components/dependencies/composables/useDependencies.ts
+++ b/ui/src/components/dependencies/composables/useDependencies.ts
@@ -404,6 +404,11 @@ export function useDependencies(container: Ref<HTMLElement | null>, subtype: typ
 
         sse.value.onerror = () => {
             coreStore.message = {variant: "error", title: t("error"), message: t("something_went_wrong.loading_execution")};
+
+            // Close on error: EventSource auto-reconnects unless explicitly closed,
+            // and each reconnect leaks a server-side SSE connection (Netty direct
+            // buffers) over time. See kestra-io/kestra#16982.
+            closeSSE();
         };
     };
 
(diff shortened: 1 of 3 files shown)


elevatebart (approved):
good job!

PR #16997, opened 2026-06-23
feat(system): expose SSE follow-connection metrics in Prometheus

nkwiatkowski:
## What

Each SSE streaming service registers, tagged by `sse_type` (`logs` / `execution` / `dependencies`):

- **Gauge** `kestra_sse_connections_active{sse_type=…}` — currently-open follow connections, read live from the service's `subscribers` map (`subscribers.values().stream().mapToInt(Map::size).sum()`).
- **Counter** `kestra_sse_connections_opened_total{sse_type=…}` — cumulative opens; its `rate()` exposes the reconnect churn that drives the leak.

Instrumented services:
- `core/.../services/LogStreamingService.java`
- `core/.../services/ExecutionStreamingService.java`
- `webserver/.../services/ExecutionDependenciesStreamingService.java`

Metric name + tag constants added to `core/.../metrics/MetricRegistry.java`, next to the existing `queue.subscribers.active` block. Registration follows the existing `QueueLagPoller` gauge pattern; the gauge `Supplier` runs per scrape (O(active executions), negligible; `ConcurrentHashMap` iteration is thread-safe).

## Why

Lets operators correlate SSE connection count/churn with memory growth, confirm the fix in #16984 works, and alert on reconnect storms.

## Test

- Manual: open an execution's logs + a dependency view, `curl localhost:8081/prometheus | grep sse_connections` → `kestra_sse_connections_active{sse_type="logs"}` rises while following and falls after closing; `..._opened_total` increments per open.

Refs #16982

### DIFF
--- a/core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
+++ b/core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
@@ -200,6 +200,10 @@ public class MetricRegistry {
     public static final String METRIC_QUEUE_SUBSCRIBERS_PAUSED_TOTAL_DESCRIPTION = "The total number of queue subscribers pause operations";
     public static final String METRIC_QUEUE_SUBSCRIBERS_RESUMED_TOTAL = "queue.subscribers.resumed.total";
     public static final String METRIC_QUEUE_SUBSCRIBERS_RESUMED_TOTAL_DESCRIPTION = "The total number of queue subscribers resume operations";
+    public static final String METRIC_SSE_CONNECTIONS_ACTIVE = "sse.connections.active";
+    public static final String METRIC_SSE_CONNECTIONS_ACTIVE_DESCRIPTION = "Number of currently-open Server-Sent Events (SSE) follow connections";
+    public static final String METRIC_SSE_CONNECTIONS_OPENED_TOTAL = "sse.connections.opened.total";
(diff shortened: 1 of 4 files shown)


nkwiatkowski:
Reworked per review feedback (thanks @loicmathieu): moved the instrumentation from the three streaming services to the **controller Flux pipeline** — closest to the actual SSE connection lifecycle. A reusable `SseConnectionMetrics.track(flux, type)` decorator increments on `doOnSubscribe` and decrements in `doFinally` (complete/error/cancel/timeout), so the active gauge tracks real open/close (and correctly counts short-circuited connections too). Metric names/tags unchanged.

Note: the MCP SSE endpoints build their event-stream Flux inside the manually-constructed `KestraFluxStreamableServerTransportProvider` (not a DI bean) and expose `Flux<Event<?>>`, so I left them out for now to keep this surgical — the decorator makes adding `sse_type="mcp"` a trivial follow-up.

nkwiatkowski:
Addressed @loicmathieu's `doFinally` ordering point: there's now **exactly one** `doFinally`. `track(flux, type, onClose)` folds the caller's cleanup (`unregisterSubscriber`) into its single `doFinally` as `try { onClose.run(); } finally { active.decrementAndGet(); }`, so the active-connection decrement runs **last** — after the subscriber is unregistered — deterministically on every terminal signal (complete/error/cancel/timeout). (Reactor has no `doLast`, and `doAfterTerminate` doesn't fire on `cancel`, so a single ordered `doFinally` is the only way to guarantee the order.)

loicmathieu (approved):
LGTM

PR #16999, opened 2026-06-23
feat(system): expose SSE follow-connection metrics in Prometheus (1.3.x)

nkwiatkowski:
## What

Backport of #16997 to `releases/v1.3.x`.

Each SSE streaming service registers, tagged by `sse_type` (`logs` / `execution` / `dependencies`):

- **Gauge** `kestra_sse_connections_active{sse_type=…}` — currently-open follow connections, read live from the service's `subscribers` map.
- **Counter** `kestra_sse_connections_opened_total{sse_type=…}` — cumulative opens; its `rate()` exposes the reconnect churn that drives the leak.

Instrumented: `LogStreamingService`, `ExecutionStreamingService` (core), `ExecutionDependenciesStreamingService` (webserver). Metric name + tag constants added to `MetricRegistry`, plus a `Supplier`-based `gauge(...)` overload (already present on develop) so the value is read lazily per scrape.

## Why

Lets operators on 1.3.x correlate SSE connection count/churn with memory growth, confirm the fix in #16985 works, and alert on reconnect storms.

## Test

- Manual: open an execution's logs + a dependency view, `curl localhost:8081/prometheus | grep sse_connections` → `kestra_sse_connections_active{sse_type="logs"}` rises while following and falls after closing; `..._opened_total` increments per open.

Refs #16982

### DIFF
--- a/core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
+++ b/core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
@@ -138,6 +138,10 @@ public class MetricRegistry {
     public static final String METRIC_QUEUE_RECEIVE_DURATION_DESCRIPTION = "Queue duration to receive and consume a batch of messages";
     public static final String METRIC_QUEUE_POLL_SIZE = "queue.poll.size";
     public static final String METRIC_QUEUE_POLL_SIZE_DESCRIPTION = "Size of a poll to the queue (message batch size)";
+    public static final String METRIC_SSE_CONNECTIONS_ACTIVE = "sse.connections.active";
+    public static final String METRIC_SSE_CONNECTIONS_ACTIVE_DESCRIPTION = "Number of currently-open Server-Sent Events (SSE) follow connections";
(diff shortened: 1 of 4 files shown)


nkwiatkowski:
Reworked to match #16997: instrumentation moved from the streaming services to the **controller Flux pipeline** via a reusable `SseConnectionMetrics.track(flux, type)` decorator (inc on `doOnSubscribe`, dec in `doFinally`). Backend-only; `:core` + `:webserver` compile clean.

nkwiatkowski:
Addressed @loicmathieu's `doFinally` ordering point: there's now **exactly one** `doFinally`. `track(flux, type, onClose)` folds the caller's cleanup (`unregisterSubscriber`) into its single `doFinally` as `try { onClose.run(); } finally { active.decrementAndGet(); }`, so the active-connection decrement runs **last** — after the subscriber is unregistered — deterministically on every terminal signal (complete/error/cancel/timeout). (Reactor has no `doLast`, and `doAfterTerminate` doesn't fire on `cancel`, so a single ordered `doFinally` is the only way to guarantee the order.)

loicmathieu (approved):
LGTM

PR #17001, opened 2026-06-23
fix(core): allow expression-only MULTISELECT inputs

### LINKED ISSUE #16657: MULTISELECT input requires `values:` even when options come from `expression:` (asymmetric with SELECT)
## Bug

A `MULTISELECT` flow input whose options are generated dynamically (via `expression:` / the `subflow()` function) fails flow validation unless a dummy `values: []` is added. The equivalent `SELECT` input validates fine with no `values:` at all.

The two input types are inconsistent: `MultiselectInput.values` is annotated `@NotNull`, whereas `SelectInput.values` is not — even though both rebuild their option list from `expression:` at render time.

### Steps to Reproduce

Define a flow with a `MULTISELECT` input that has only an `expression:` (no static `values:`):

```yaml
inputs:
  - id: zones
    type: MULTISELECT
    dependsOn:
      inputs:
        - region
    expression: "{{ subflow(namespace='company.infra', id='catalog_zones', inputs={'region': inputs.region}).outputs.zones }}"
```

Save / validate the flow.

### Expected Behavior

The flow validates, the same way an equivalent `SELECT` input (with only `expression:`, no `values:`) does.

### Actual Behavior

Validation fails because `values` must not be null. You have to add a meaningless `values: []` to make it pass.

### Notes

- `MultiselectInput.values` -> `@NotNull` (`core/src/main/java/io/kestra/core/models/flows/input/MultiselectInput.java`, ~line 32)
- `SelectInput.values` -> no annotation (`core/src/main/java/io/kestra/core/models/flows/input/SelectInput.java`, ~line 32)

Suggested fix: drop `@NotNull` from `MultiselectInput.values` so it matches `SelectInput`.

kumarroshan12:
### ✨ Description

This PR fixes an inconsistency between `SELECT` and `MULTISELECT` inputs when options are provided dynamically through `expression`.

Previously, `MULTISELECT` validation failed with:

`zones.values: must not be null`

when only `expression` was provided, even though the equivalent `SELECT` input was valid.

Changes:

* Removed the `@NotNull` constraint from `MultiselectInput.values`
* Added a validation test covering expression-only `MULTISELECT` inputs

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/16657

### 🛠️ Backend Checklist

* [x] Code compiles successfully and passes all checks
* [x] Added a regression test covering the reported issue
* [x] All relevant unit tests pass

### 📝 Additional Notes

The issue was unassigned at the time this work was started.

This change aligns `MULTISELECT` behavior with `SELECT`, which already supports dynamically generated options via `expression` without requiring a static `values` list.


kumarroshan12:
Hi @loicmathieu, whenever you have some time, could you please take a look at this PR?

This fixes the validation inconsistency where `MULTISELECT` required `values` even when options were provided dynamically through `expression`, while `SELECT` already supported this behavior. I've also added a regression test to cover the reported issue.

Thanks for your time.

loicmathieu (approved):
LGTM

loicmathieu:
@kumarroshan12 `FlowTest.inputValidation()` now fail, can you have a look please?

kumarroshan12:
Hi @loicmathieu, thanks for the review!

I've updated `FlowTest.inputValidation()` to reflect the new `MULTISELECT` validation behavior and pushed the changes. All tests pass locally now.

Please let me know if anything else needs to be addressed. Thanks!

loicmathieu (approved):
LGTM, thanks for your contribution

PR #17003, opened 2026-06-23
fix: namespace files revision history modal

elevatebart:
closes https://github.com/kestra-io/kestra-ee/issues/8818


ui/src/components/layout/Revisions.vue
```diff
@@ -368,6 +368,16 @@
         padding-bottom: 1rem;
     }
 
+    .revision-grid {
+        display: grid;
+        grid-template-columns: repeat(2, minmax(0, 1fr));
+        margin-right: 2rem;
```
flcarre: **F2 — spacing token** (DS rule #6): `2rem` maps to `--ks-spacing-6`.

```suggestion
        margin-right: var(--ks-spacing-6);
```

ui/src/components/layout/Revisions.vue
```diff
@@ -389,11 +399,18 @@
         }
     }
 
-    .revision-crud-info {
-        flex-shrink: 0;
-        white-space: nowrap;
+    .revision-crud-info{
+        width:calc(100% - 1rem);
+        margin-right: 1rem;
```
flcarre: **F2 — spacing tokens** (DS rule #6): `1rem` maps to `--ks-spacing-4`.

```suggestion
    .revision-crud-info {
        width: calc(100% - var(--ks-spacing-4));
        margin-right: var(--ks-spacing-4);
```

ui/src/components/layout/Revisions.vue
```diff
@@ -389,11 +399,18 @@
...
+        :deep(> div) {
```
flcarre: **F1 — DS golden rule #4 (no `:deep()`).** This `:deep(> div)` reaches across the slot boundary into the root element of the `<Crud>` component that fills `<slot name="crud">`. In OSS, `Crud` renders an empty `<span/>`, so this rule only takes effect against the **EE** `Crud` override (the created/updated-by block) — i.e. OSS styling is coupled to EE slot-content internals, which is exactly what rule #4 exists to prevent (it'll silently break if `Crud`'s markup changes).

Heads up: it's not a one-line swap. Moving `overflow`/`text-overflow: ellipsis` onto `.revision-crud-info` itself won't truncate, because the text lives inside `Crud`'s own element. The clean options are to let the slot-content component own its truncation (apply the `overflow`/`ellipsis` inside `Crud`), or expose it via a prop — rather than reaching in from here. If a cross-boundary reach is genuinely unavoidable for slot content, let's call it out explicitly and agree on it rather than leave a bare `:deep()`.

flcarre (changes requested):
Thanks @elevatebart — the fix is correct and addresses all three symptoms in kestra-io/kestra-ee#8818. I checked out the branch and verified the parts that looked risky are actually safe:

- ✅ Teleporting the dialog body (`appendToBody`) keeps the scoped `.revision-history-dialog-body` height rule applying — Vue stamps `data-v-*` on the vnode, so teleport doesn't strip it.
- ✅ `KsRow`/`KsCol` → `grid` with `minmax(0, 1fr)` is the right overflow-safe primitive (the flex gutter is what let the `KsSelect` dropdown overflow its column); no DS component wraps this, so the raw `<div>` grid is justified here.
- ✅ `appendToBody` correctly escapes the `.sidebar` `overflow-x: hidden` stacking context that was clipping the Restore button.

Two design-system golden-rule items to address before merge (inline):

1. **`:deep(> div)` in `.revision-crud-info`** — rule #4 (no `:deep()`).
2. **Hardcoded spacing** (`2rem` / `1rem`) — rule #6, should use `--ks-spacing-*`.

Everything else looks good.

flcarre (approved):

PR #17004, opened 2026-06-23
feat(system): expose SSE follow-connection metrics in Prometheus (1.0.x)

nkwiatkowski:
## What

Backport of #16997 to `releases/v1.0.x`.

Instrumented at the controller `Flux` pipeline via a reusable `SseConnectionMetrics.track(flux, type, onClose)` decorator, applied to the logs, execution and dependencies follow endpoints. Tagged by `sse_type`:

- **Gauge** `kestra_sse_connections_active{sse_type=…}` — currently-open follow connections (incremented on `doOnSubscribe`, decremented in `doFinally`).
- **Counter** `kestra_sse_connections_opened_total{sse_type=…}` — cumulative opens; `rate()` exposes reconnect churn.

The caller's stream cleanup (`unregisterSubscriber`) is folded into the **single** `doFinally` the decorator installs, so the active-connection decrement runs deterministically **last** — after the subscriber is unregistered — on every terminal signal (complete / error / cancel / timeout). The non-SSE "create-and-wait" execution endpoints (trigger/webhook `wait=true`) are intentionally **not** instrumented — they return JSON, not `text/event-stream`.

## Why

Lets operators on 1.0.x correlate SSE connection count/churn with memory growth and confirm the fix in #16987 works.

## Test

Refs #16982


loicmathieu (approved):
LGTM

PR #17007, opened 2026-06-23
fix(flows): match task type by partial class name in suggestions

ferdinandlaine:
### ✨ Description

This fixes partial class name matching where typing `execution` cannot match `SentryExecution` since the class name is matched as a whole. `getLabelSegments` now splits class names into words so partial input matches correctly.

### 👀 Preview

https://github.com/user-attachments/assets/d80f7619-d507-41c7-b83d-284af8807051

### 🎨 Frontend Checklist

- [x] All existing E2E tests pass
- [x] Video recording attached showing the changes


elevatebart (approved):
Nice !!!

PR #17031, opened 2026-06-24
fix(core): fix exponential retry delay — first retry was 0s, growth w…

loicmathieu:
…as linear not exponential

nextRetryDate() was computing interval * factor * (attemptCount-1), which produced a 0s delay on the first retry and grew linearly instead of exponentially. Switch to interval * factor^(attemptCount-1) to match the Failsafe withBackoff() semantics already used by toPolicy(). Also preserve the full double precision of delayFactor (was truncated via intValue()) and use Math.round instead of a cast to avoid IEEE 754 truncation undercount.

Add nextRetryDate unit tests for all three retry types: Exponential, Constant, and Random.

Fixes kestra-io/kestra-ee#8821


AcevedoR (approved):

loicmathieu:
Backported to 1.0 and 1.3

PR #17032, opened 2026-06-24
fix(ui): reliably re-trigger editor autocomplete in YAML and Pebble contexts

### LINKED ISSUE #14989: Autocomplete doesn't always display in editor

brian-mulier-p:
Fixes #14989.

The flow editor's meaningful completions — plugin `type:` values and Pebble `{{ }}` expressions — live inside YAML *string* tokens. Monaco's default `quickSuggestions` (`{strings: false}`) never auto-triggers there, so once the suggest widget was dismissed (backspace, space, or moving the cursor) it would not reappear while typing, and cursor moves into/between `{{ }}` blocks failed to reopen it.

- Enable `quickSuggestions.strings` for YAML/Pebble editors so the editor re-shows completions as you type inside `type:` values and `{{ }}` expressions.
- Re-trigger suggestions when the cursor enters a Pebble block, tracking the block by **identity** (its opening `{{` offset) on every cursor change rather than inside the debounced callback. This survives the cursor-move debounce on a fast out-and-back round-trip **and** catches moving straight from one `{{ }}` block to another (no non-block position in between). Staying in the same block is not a fresh entry, so dismissing the widget with Escape and nudging the cursor won't reopen it. Extracted into `createPebbleEntryTracker` / `pebbleBlockKeyAtOffset` helpers with unit tests.

Verified live (shared design-system, exercised through the EE flow editor):
- Typing a partial `type:` value shows task suggestions; after backspace, resuming typing re-shows them.
- Space+backspace inside `{{ }}` re-shows suggestions.
- Moving the cursor out of and back into a `{{ }}` block — and directly between two blocks — re-shows suggestions (real mouse + keyboard).
- Escape-dismiss then nudging within the same block stays closed; plain string values (e.g. `message: hello world`) stay quiet — no spurious popups.

### DIFF
--- a/ui/packages/design-system/src/components/Form/KsEditor.vue
+++ b/ui/packages/design-system/src/components/Form/KsEditor.vue
@@ -90,6 +90,13 @@
         return isOffsetInPebbleBlock(editor.getValue(), absoluteOffset)
     }
 
+    function cursorPebbleBlockKey(editor: monaco.editor.ICodeEditor): number | null {
+        const cursorPos = editor.getPosition()
+        if (!cursorPos) return null
+        const absoluteOffset = editor.getModel()?.getOffsetAt(cursorPos) ?? 0
+        return pebbleBlockKeyAtOffset(editor.getValue(), absoluteOffset)
+    }
+
     function uid(): string {
         return Math.random().toString(36).slice(2, 11)
     }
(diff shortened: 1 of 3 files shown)


elevatebart (approved):
Deep PR, took me a minute to wrap my head around it.

I wonder if I should ask claude to explain it to me like I am 5.

Good work.

PR #17037, opened 2026-06-24
fix(executor): emit FollowExecutionEvent in the same transaction

loicmathieu:
…to fix webhook wait:true hang

In the catch block of toExecution(), the lock lambda previously returned null (execution never persisted) and called followExecutionEventQueue.emitAsync() inside the transaction, allowing the FollowExecutionEvent INSERT to commit before the execution row was visible. ExecutionStreamingService would then fail to find the execution, drop the terminal event, and leave webhook wait:true callers blocked forever on Flux.last().

Fix: return a non-null ExecutorContext from the lock lambda so the FAILED execution is persisted, move both queue emits outside the lambda (post-commit), and correct the FollowExecutionEvent type from UPDATED to TERMINATED.

Closes kestra-io/kestra-ee#8824

### DIFF
--- a/core/src/test/java/io/kestra/core/services/ExecutionStreamingServiceTest.java
+++ b/core/src/test/java/io/kestra/core/services/ExecutionStreamingServiceTest.java
@@ -1,0 +1,99 @@
+package io.kestra.core.services;
+
+import java.util.Collections;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.junit.jupiter.api.Test;
+
+import io.kestra.core.junit.annotations.KestraTest;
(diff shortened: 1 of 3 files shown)


AcevedoR (approved):

AcevedoR (approved):

loicmathieu:
## Performance impact

More tests would be needed to assess the performance impact in details but prelimirary tests shows that:
- It shows no regression up to 3000 exec/mn - 6000 tasks/mn (the new supported perf level of 2.0)
- It shows performance regression at 3500 exec/mn (1.5s -> 7s) but here we are above what Kestra can support in terms of performance with our benchmark standard deployment so it could be a benchmark artifact

Overall: the impact is OK

PR #17040, opened 2026-06-24
fix(namespaces): make namespace file reads resilient to index/storage drift

fhussonnois:
When the metadata index pointed at a revision whose storage object no longer existed, reads failed with a hard 404. Now getFileContent falls back to the most recent lower revision present in storage, and purge() deletes metadata before the object so an interrupted purge can't strand the index ahead of storage.

Related-to: kestra-io/kestra-ee#8067


brian-mulier-p (approved):
LGTM

PR #17042, opened 2026-06-24
feat(core): REUSABLE_INPUTS input type, validation, and editor seam

brian-mulier-p:
Open-source primitives for reusable inputs (the Enterprise feature that stores and resolves the definitions lives in the companion kestra-ee PR). Kept in OSS because the input `type` is a core enum used as Jackson's discriminator, so the type must be registered here for any flow to parse `type: REUSABLE_INPUTS`.

- New `REUSABLE_INPUTS` input type referencing a definition by `ref` (+ optional `namespace`/`revision`). Marked `@EeOnly` so it is excluded from the OSS-served flow schema — the open-source editor neither proposes nor flags it, while the EE-served schema includes it.
- `ReusableInputsExpander` interface with an `expand()` inlining primitive (the reusable-inputs counterpart of `Input.expandToLeaves`). The OSS default errors ("requires Enterprise Edition"); the EE build replaces it with a store-backed implementation.
- `FlowInterface.resolvableInputs(expander)` / `inlinedInputs(expander)` centralise reusable-inputs inlining alongside FORM flattening, so resolution paths inline references before flattening to dotted leaves; `FlowForExecution.of(flow, expander)` makes the execution form render and submit the resolved inputs under the reference id.
- `FlowValidator` rejects `@EeOnly` inputs at validation time on open-source (the EE build overrides to allow them), so a `REUSABLE_INPUTS` flow can't be saved on OSS.
- Editor: a generic `registerEditionProviders()` seam on the YAML language configurator (swapped per-edition via the `override` alias) lets the EE build attach the reusable-inputs hover / go-to-definition without shipping any reusable-inputs UI code in open-source; plus reusable-inputs sibling-input autocompletion in the definition editor. The execution form also skips an un-inlined reference node defensively.
- Docs: AGENTS.md now requires i18n for all user-facing UI strings.

Companion EE PR: https://github.com/kestra-io/kestra-ee/pull/8926

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/FlowInterface.java
+++ b/core/src/main/java/io/kestra/core/models/flows/FlowInterface.java
@@ -24,6 +24,9 @@ import io.kestra.core.models.flows.quota.Quota;
 import io.kestra.core.models.flows.sla.SLA;
 import io.kestra.core.models.tasks.WorkerSelector;
 import io.kestra.core.queues.event.BroadcastEvent;
+import io.kestra.core.runners.ReusableInputsExpander;
+
+import io.micronaut.core.annotation.Nullable;
 import io.kestra.core.serializers.JacksonMapper;
 
 /**
@@ -66,8 +69,33 @@ public interface FlowInterface extends FlowId, SoftDeletable<FlowInterface>, Ten
      */
     @JsonIgnore
     default List<Input<?>> resolvableInputs() {
+        return resolvableInputs(null);
+    }
+
+    /**
+     * Same as {@link #resolvableInputs()} but also inlines every {@code REUSABLE_INPUTS} reference (via the given
+     * {@link ReusableInputsExpander}) before flattening — use this overload on resolution paths that may encounter a
+     * reference. A {@code null} expander means "no reusable inputs to resolve" (FORM flattening only), so open-source
+     * call sites and flows without references can pass {@code null}.
+     */
+    default List<Input<?>> resolvableInputs(@Nullable ReusableInputsExpander reusableInputsExpander) {
+        return Input.expandToLeaves(inlinedInputs(reusableInputsExpander));
+    }
+
+    /**
+     * The authored inputs with every {@code REUSABLE_INPUTS} reference inlined (the block's inputs spliced in, ids
+     * prefixed by the reference id) but {@code FORM} grouping left intact — the structure the execute form renders.
+     * {@link #resolvableInputs(ReusableInputsExpander)} layers FORM flattening on top of this. A {@code null} expander
+     * returns the authored inputs unchanged.
+     */
+    default List<Input<?>> inlinedInputs(@Nullable ReusableInputsExpander reusableInputsExpander) {
         List<Input<?>> inputs = getInputs();
-        return inputs == null ? List.of() : Input.expandToLeaves(inputs);
+        if (inputs == null) {
+            return List.of();
+        }
+        return reusableInputsExpander == null
+            ? inputs
+            : reusableInputsExpander.expand(getTenantId(), getNamespace(), inputs);
     }
 
     List<Output> getOutputs();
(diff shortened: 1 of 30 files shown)


nkwiatkowski (approved):

core/src/main/java/io/kestra/core/models/flows/FlowInterface.java
```diff
@@ -64,8 +67,35 @@ public interface FlowInterface extends FlowId, SoftDeletable<FlowInterface>, Ten
...
+        return resolvableInputs(null);
+    }
+
+    /**
+     * Same as {@link #resolvableInputs()} but also inlines every {@code REUSABLE_INPUTS} reference (via the given
+     * {@link ReusableInputsExpander}) before flattening — use this overload on resolution paths that may encounter a
+     * reference. A {@code null} expander means "no reusable inputs to resolve" (FORM flattening only), so open-source
+     * call sites and flows without references can pass {@code null}.
+     */
+    @JsonIgnore
```
loicmathieu:
```suggestion
```

core/src/main/java/io/kestra/core/models/flows/FlowInterface.java
```diff
@@ -64,8 +67,35 @@ public interface FlowInterface extends FlowId, SoftDeletable<FlowInterface>, Ten
...
+        return Input.expandToLeaves(inlinedInputs(reusableInputsExpander));
+    }
+
+    /**
+     * The authored inputs with every {@code REUSABLE_INPUTS} reference inlined (the block's inputs spliced in, ids
+     * prefixed by the reference id) but {@code FORM} grouping left intact — the structure the execute form renders.
+     * {@link #resolvableInputs(ReusableInputsExpander)} layers FORM flattening on top of this. A {@code null} expander
+     * returns the authored inputs unchanged.
+     */
+    @JsonIgnore
```
loicmathieu:
```suggestion
```

core/src/main/java/io/kestra/core/runners/NoOpReusableInputsExpander.java
```diff
@@ -0,0 +1,23 @@
...
+import io.kestra.core.models.flows.input.ReusableInputsInput;
+
+import jakarta.inject.Singleton;
+
+/**
+ * Open-source default: reusable inputs are an Enterprise Edition feature, so any attempt to resolve a
+ * {@code REUSABLE_INPUTS} reference fails. The EE build replaces this with a store-backed implementation.
+ */
+@Singleton
+public class NoOpReusableInputsExpander implements ReusableInputsExpander {
```
loicmathieu:
```suggestion
public class ReusableInputsExpander implements ReusableInputsExpander {
```

It's not a Noop as it throws....

loicmathieu (approved):
Some minor nits but overall LGTM

core/src/main/java/io/kestra/core/models/flows/FlowInterface.java
brian-mulier-p: Removed — this overload takes a parameter, so Jackson never treats it as a bean getter; the `@JsonIgnore` was redundant.

core/src/main/java/io/kestra/core/models/flows/FlowInterface.java
brian-mulier-p: Removed — same reasoning, this overload also takes a parameter.

core/src/main/java/io/kestra/core/runners/NoOpReusableInputsExpander.java
brian-mulier-p: Renamed, but to `DisabledReusableInputsExpander` rather than `ReusableInputsExpander` — that name collides with the interface it implements. Agreed it's not a no-op, so the class name shouldn't say so.

PR #17045, opened 2026-06-24
fix(namespaces): disable save button in file editor when no edits

MilosPaunovic:
## Summary

- Disable the Save icon button in the namespace files editor when there are no unsaved edits
- Guard the Ctrl+S keyboard shortcut for namespace files the same way
- Add visual feedback (dimmed icon, `cursor: not-allowed`) via `.save-disabled` CSS class

## Test plan

- [ ] Open Namespaces > system > Namespace Files, open a file without editing — Save button should be visually disabled and non-clickable
- [ ] Edit a file — Save button should become active
- [ ] Save the file — button goes back to disabled state
- [ ] Ctrl+S on an unmodified file should be a no-op (no unnecessary API call)

Closes https://github.com/kestra-io/kestra-ee/issues/8820.


Pradumnasaraf (approved):
Looks good to merge!

Image

PR #17046, opened 2026-06-24
feat: concurrency-queue-size

### LINKED ISSUE #15779: Add queueSize as parameter to concurrency.QUEUE
### Feature description

We have multiple flows (Flow A1, A2, A3, etc.) and one common flow (Flow X) that should be started after these flows finish.

Flows An can be started at any time (in parallel, individually, or simultaneously), but the end result should always trigger Flow X.

Currently, we have a concurrency limit of 1 and the Queue option enabled on Flow X. However, this results in more executions than necessary. For example:

Flow A1 finishes, and Flow X starts.

While Flow X is running, Flow A2 and Flow A3 complete.

Two new copies of Flow X are added to the queue.

I only need one additional execution to catch the updates from both A2 and A3; any subsequent queued runs should be skipped.

Could we add a Queue Size parameter (or something like that to skip if already queued) to limit the number of executions in the queue and cancel any others?

hemantmm:
closes: #15779

---

### ✨ Description

What does this PR change?  

Added `queueSize` to flow concurrency when using `behavior: QUEUE`.

This lets a flow cap how many executions may wait in the concurrency queue. When the queue is already full, additional executions are cancelled instead of being queued indefinitely.

https://github.com/user-attachments/assets/e8550ebe-250d-4ca8-ba31-d97c6f8b6f3b

Example:

```yaml
concurrency:
  behavior: QUEUE
  limit: 1
  queueSize: 1


hemantmm:
It shows that all tests executed successfully, and only the automated PR-comment step failed with a GitHub token permission error.

hemantmm:

hemantmm:

hemantmm:
Hey @loicmathieu, all tests are passing successfully. The remaining failure is from the workflow attempting to post a PR comment (Resource not accessible by integration, HTTP 403), which is a GitHub Actions permission issue rather than a code or test failure.

hemantmm:

PR #17048, opened 2026-06-24
feat(plugins): redesign flow editor documentation panel

flcarre:
Closes https://github.com/kestra-io/kestra-ee/issues/8879

## What & why

Redesigns the flow-editor **documentation panel** (the plugin/task docs shown while editing a flow) for a shadcn-grade reading experience. This panel reaches ~98% of flow-editor users and is the highest-volume editor surface, so its clarity matters. The redesign covers both states of the panel: the default intro page (no task at cursor) and the task-bound view. It reworks the Flow properties presentation flagged in the linked issue (the old 2-column table is replaced by clean stacked rows with intentional spacing).

## Changes

**Intro / default page** (`PluginDocumentation.vue`)
- Hero, a contextual callout, a reference search, and tabbed sections (Overview / Flow / Tasks / Inputs / Pebble / Examples) split from the bundled `basic.md`.
- A "Start here" guide whose 3 steps are now real buttons that jump to the matching reference tab (id/namespace, tasks, inputs).
- A "Learn more" link grid.
- The intro now renders immediately instead of being gated behind the plugins fetch, so it survives a slow or flaky `/plugins` load.

**Task-bound view**
- Redesigned header: package kicker, task icon, name, kind chip, copyable full `type`, version, release link, and a summary line.
- A controlled section nav (Overview / Properties N / Outputs N / Examples).
- A compact, filterable properties reference (All / Required) via a new `compact` mode in `SchemaToHtml` / `SchemaPropertiesSection`.

**Design system** (`KsTable.vue`)
- Exposes a `--ks-table-row-hover-bg` token (defaults to `--ks-bg-hover`), so the panel can disable row-hover on read-only markdown tables through a CSS variable instead of a `:deep` override.

## How it was verified
- `npm run check:types` (design-system + app + test), `eslint`, and `node check.js` all green.
- Live QA on the running app driving every flow (intro, actionable steps, search, bound task view, properties): 0 console / page errors, all states render in light and dark mode.
- A short demo video of the flows was recorded; I will drag-drop the mp4 into this PR.

https://github.com/user-attachments/assets/a2688f41-2b7c-4421-9825-42cbd426c243

## Notes
- No backend changes. `PluginList.vue` shows no net diff (an exploratory change was reverted).
- This is a UI redesign and would benefit from a design sign-off before merge.


flcarre:
@anna-geller @tchiotludo @Nico-Kestra @loicmathieu proposol of an improvement of the doc tab as saw yesterday. What do you think of this ?

tchiotludo:
definitely better, love it :heart:

Piyush-r-bhaskar (changes requested):
Hey,

Issue is about revamp for Flow Properties Docs only but you have changed the Plugin documentation as well in panel. is this related ? 
If Yes, could we have a FIGMA to track please ?
If no, revert the changes for Schema and Plugin documentation and we are good.

flcarre:
Yes it's intentional: both docs share the same panel and the same components (`SchemaToHtml`), restyling only the flow properties would have left it half old / half new.

No Figma here, it was designed code-first with the DS tokens, the demo video above is the reference.

@Nico-Kestra can you take a look design-wise?

Piyush-r-bhaskar:
can you fix the conflicts please @flcarre

Piyush-r-bhaskar (approved):

PR #17054, opened 2026-06-25
feat(topology): render Pebble expressions for display in the topology graph

jymaire:
## Summary

- Adds `POST /api/v1/{tenant}/expressions/render` endpoint that resolves a list of Pebble expressions for safe display (secrets and env vars masked, unresolvable expressions kept raw)
- `GraphService` now builds a richer pre-execution variable context that merges namespace-level variables (via `NamespaceVariablesProvider`, no-op in OSS, overridden in EE) with flow-level `variables:` block
- Topology slots receive the flow source string so plugin UIs can resolve draft (unsaved) expressions before an execution exists

## Depends on
- EE: https://github.com/kestra-io/kestra-ee/pull/8836

## Test plan
- [ ] `POST /api/v1/expressions/render` resolves flow `variables:` block expressions pre-exec
- [ ] `GraphService` tests pass (`./gradlew :core:test --tests "GraphServiceTest"`)
- [ ] Topology "Show Details" modal resolves `{{ vars.* }}` expressions without an execution


jymaire:
## Why `java.lang.reflect.Proxy` in `PebbleEngineFactory`

`PebbleEngine` is built via a `PebbleEngine.Builder` that accepts `Extension` objects. An `Extension` exposes a `Map<String, Function>` of named functions, but there is no public API to swap individual functions out after the engine is built — the extension map is consumed at construction time and baked in.

The two new engine variants (`createForDisplay`, `createWithMaskedFunctions`) need to intercept specific function invocations without replacing whole extensions or forking Pebble's internals. Dynamic proxies are the surgical solution:

- **`wrapExtension`** proxies an `Extension` to intercept the `getFunctions()` call, returning a remapped version of the function map. Every other `Extension` method (filters, tests, global variables, …) is transparently delegated to the original.
- **`interceptExecute`** proxies a single `Function` to intercept its `execute()` call with a custom lambda. Every other `Function` method (`getArgumentNames`, etc.) is delegated to the original, so the proxy is invisible to the engine except at invocation time.

The result is that **the engine itself sees fully conformant `Extension` and `Function` objects** — no Pebble internals are touched. The alternative (subclassing each extension or function) would require knowing the concrete types at compile time, which is impossible since extensions come from plugins loaded at runtime.

core/src/main/java/io/kestra/core/runners/pebble/PebbleEngineFactory.java
```diff
@@ -79,27 +170,61 @@ private PebbleEngine.Builder newPebbleEngineBuilder() {
...
+    @FunctionalInterface
+    private interface ExecuteInterceptor {
+        Object intercept(Map<String, Object> args) throws Throwable;
+    }
+
+    /**
+     * Returns a proxy for {@code initial} that replaces its {@code execute} method with {@code handler}.
+     * All other {@link Function} methods delegate to the original.
+     */
+    private static Function interceptExecute(Function initial, ExecuteInterceptor handler) {
```
loicmathieu: Using a proxy seems over complicated.
As you are iterating over the list of functions that the extension provide, a simpler approach would have to replace non allowed function but a NoOpFunction which returns the expression.

core/src/main/java/io/kestra/core/runners/pebble/DisplayUnrenderableException.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Thrown by display-engine function proxies to signal that a given expression segment
+ * is non-deterministic or has side effects and must therefore be kept raw rather than resolved.
+ * The caller ({@code DisplayExpressionRenderer}) catches this and falls back to the original template text.
+ */
+public class DisplayUnrenderableException extends RuntimeException {
+
+    public DisplayUnrenderableException() {
+        super(null, null, true, false); // suppress stack-trace capture — this is a control-flow signal
```
loicmathieu: Control flow with an expression is a smell

core/src/main/java/io/kestra/core/runners/DisplayExpressionRenderer.java
```diff
@@ -0,0 +1,158 @@
...
+
+        return resolveSegments(template, variables);
+    }
+
+    /**
+     * Splits the template into alternating literal and expression segments,
+     * renders each expression independently, and stitches the result.
+     * Expressions inside {@code {% raw %}...{% endraw %}} blocks are never rendered.
+     */
+    private String resolveSegments(String template, Map<String, Object> variables) {
```
loicmathieu: I don't understand why you have to do that, Pebble do that automatically no?

core/src/main/java/io/kestra/core/services/NamespaceVariablesProvider.java
```diff
@@ -0,0 +1,32 @@
...
+ * <p>This OSS implementation is a no-op that returns an empty map — there are no
+ * namespace-level variables in the open-source edition. Enterprise Edition replaces this
+ * with a real implementation backed by the namespace meta-store, exposing variables
+ * inherited from parent namespaces and making them available in the topology "Show details"
+ * view before any execution exists.
+ *
+ * <p>Flow-level variables take precedence over namespace variables.
+ */
+@Singleton
+public class NamespaceVariablesProvider {
```
loicmathieu: Why do you have to do that?
You would render the expression using RunVariables created from an execution right? So namespace variables would be there right?

webserver/src/main/java/io/kestra/webserver/controllers/api/ExpressionController.java
```diff
@@ -0,0 +1,136 @@
...
+                .orElseThrow(() -> new HttpStatusException(HttpStatus.NOT_FOUND, "Unable to find flow '" + request.namespace() + "." + request.flowId() + "'"));
+
+            return flowVariables(flow);
+        }
+
+        return Map.of();
+    }
+
+    private Map<String, Object> flowVariables(Flow flow) {
+        // Without an execution, RunVariables exposes flow.* but not vars.*, so we build vars explicitly.
```
loicmathieu: This is IMOH a bug that should be fixed

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/services/NamespaceVariablesProvider.java
jymaire: Because we also want to display expression before the execution during flow edit

core/src/main/java/io/kestra/core/runners/pebble/PebbleEngineFactory.java
jymaire: Proxy removed.
About the NoOpFunction, with the changes done might not be still needed ?

core/src/main/java/io/kestra/core/runners/pebble/DisplayUnrenderableException.java
jymaire: removed

webserver/src/main/java/io/kestra/webserver/controllers/api/ExpressionController.java
jymaire: removed

core/src/main/java/io/kestra/core/runners/pebble/PebbleEngineFactory.java
```diff
@@ -47,21 +62,99 @@ public PebbleEngine createWithCustomSyntax(Syntax syntax, Class<? extends Extens
...
+            if (SAFE_DISPLAY_FUNCTIONS.contains(name)) {
+                return entry;
+            } else if (name.equals(SecretFunction.NAME)) {
+                // Returns [secret: KEY] without touching the real secret service.
+                return Map.entry(name, new InterceptingFunction(entry.getValue(),
+                    args -> "[secret: " + args.getOrDefault("key", "?") + "]"));
+            }
+            // Everything else (including env(), kv(), now(), uuid(), read(), …) is removed: Pebble cannot
+            // invoke it, the expression fails to evaluate, and the caller keeps it raw.
+            return null;
```
loicmathieu: Instead of failing shouldn't we use the same InterceptingFunction which would return the original expression?

core/src/main/java/io/kestra/core/runners/DisplayExpressionRenderer.java
```diff
@@ -0,0 +1,91 @@
...
+     * Resolves a single template string for display.
+     *
+     * <p>Rendering is all-or-nothing: if any expression in the template cannot be resolved, Pebble
+     * aborts the render and the original template is returned unchanged.
+     *
+     * @param template  the raw Pebble template (may be null or contain no expressions)
+     * @param variables the variable context (flow-level, or execution-level)
+     * @return the resolved display string, or the original template if it cannot be fully resolved
+     */
+    public String resolveForDisplay(String template, Map<String, Object> variables) {
```
loicmathieu:
```suggestion
    public String render(String template, Map<String, Object> variables) {
```

It's the same primitive as the existing render but for a single string

core/src/main/java/io/kestra/core/runners/DisplayExpressionRenderer.java
```diff
@@ -0,0 +1,91 @@
...
+        if (template == null || !template.contains("{{")) {
```
loicmathieu:
```suggestion
        if (template == null || !template.contains("{")) {
```

Pebble allow `{{` or `{%`.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExpressionController.java
```diff
@@ -0,0 +1,124 @@
...
+    public record RenderExpressionRequest(
+        @NotEmpty @Size(max = 500) @Schema(description = "The raw Pebble expressions to render") List<String> expressions,
+        @Nullable @Schema(description = "Resolve against this execution's context") String executionId,
+        @Nullable @Schema(description = "Resolve against this flow's context (with flowId)") String namespace,
+        @Nullable @Schema(description = "Resolve against this flow's context (with namespace)") String flowId,
+        @Nullable @Schema(description = "Resolve against this flow source's context (YAML)") String flow
+    ) {}
+
+    @Introspected
+    public record RenderedExpressions(
```
loicmathieu, line 119: This record didn't brings any benefits, why don't you directly return the Map?

loicmathieu (changes requested):

webserver/src/main/java/io/kestra/webserver/controllers/api/ExpressionController.java
jymaire: I tried to remove it but it broke the feature. It is used as a contract by frontend

webserver/src/main/java/io/kestra/webserver/controllers/api/ExpressionController.java
```diff
@@ -0,0 +1,124 @@
...
+
+        return Map.of();
+    }
+
+    private Map<String, Object> flowVariables(Flow flow) {
+        // RunVariables exposes flow.* and flow-level vars.* even without an execution.
+        return runContextFactory.of(flow, Map.of()).getVariables();
+    }
+
+    @Introspected
```
loicmathieu: Can you remove the `@Introspected` annotation, it's useless,

webserver/src/main/java/io/kestra/webserver/controllers/api/ExpressionController.java
```diff
@@ -0,0 +1,124 @@
...
+    public record RenderExpressionRequest(
+        @NotEmpty @Size(max = 500) @Schema(description = "The raw Pebble expressions to render") List<String> expressions,
+        @Nullable @Schema(description = "Resolve against this execution's context") String executionId,
+        @Nullable @Schema(description = "Resolve against this flow's context (with flowId)") String namespace,
+        @Nullable @Schema(description = "Resolve against this flow's context (with namespace)") String flowId,
+        @Nullable @Schema(description = "Resolve against this flow source's context (YAML)") String flow
+    ) {}
+
+    @Introspected
```
loicmathieu: remove it please

loicmathieu (approved):
LGTM, please remove the two annotations before merging

PR #17070, opened 2026-06-26
feat(executions): redesign execution & log filter UX (grouped states, compact operator, live apply)

flcarre:
## What

Reworks execution & log filtering on the Executions and Logs surfaces, plus the underlying design-system filter popover.

### Changes
1. **Removed the redundant quick-filter rows** — the execution-state pill row on the Executions list (added in #16550) and the log-level pill rows on Logs / the execution Logs tab / Gantt. They duplicated the canonical `KsFilter` controls (same query params) and weren't part of the design-system spec. The Logs download dialog's level/interval picker is kept.
2. **Grouped, collapsible state filter** — states are grouped into **Running / Paused / Completed / Failed**; a one-click bucket toggle selects the whole group; single-state buckets render as plain rows; unknown states fall into an "Other" bucket. Non-state filters keep the flat list.
3. **Compact operator** — the comparator is folded into the popover header: a segmented control for ≤3 operators (glyphs ≥/≤ or short labels, with tooltips) and a compact dropdown for more; replaces the old full-width labeled select.
4. **Live apply** — the Apply button is removed (Reset kept); changes apply immediately (text debounced, time-range waits for both ends); the popover stays open and empty non-default filters are cleaned up on close.

Group labels are translated across all 13 locales (`KsFilter.locale.ts`, registered via `registerDesignSystemI18n`).

### Product notes
- `WARNING` is grouped under **Completed** (with SUCCESS/SKIPPED/RETRIED) and `KILLING` under **Failed**, per product decision (note: the engine marks `WARNING` as `isFailed:true`, so it is deliberately neither plain success nor plain failure).
- The state-group taxonomy intentionally differs (for now) from the dashboard table tabs (`sections/quickFilters.ts`); aligning the dashboard is a follow-up.

### Test plan
- `npm run check:types`, `npm run test:lint` — clean
- Design-system unit tests (696) incl. new `FilterComparatorSelect` (5) & `FilterMultiSelectGroups` (12) — pass
- App filter tests (81) — pass
- `node ui/src/translations/check.js` — clean (all locales)
- Manual QA (light + dark): grouped state filter, one-click bucket, expand, In/Not In & ≥/≤ operators, namespace compact dropdown, live apply, no pill rows

https://github.com/user-attachments/assets/2041f05b-717f-430e-b2ac-065e6dfb4913


ui/packages/design-system/tests/units/Data/KsFilter/FilterComparatorSelect.test.ts
elevatebart, line 1: We will need to need to discuss this choice:

This unit test file of a component is not helping me understand how the feature is supposed to work.
The tests themselves rely on internals like css classes or named event. 
They are inherently brittle and will cost us in compute and in brain time.

- If the test do not help us refactor in the future
- If they do not document a functional behavior and help a reviewer understand
- If running them does not check that the visual value is kept

What is their purpose?

ui/packages/design-system/tests/units/Data/KsFilter/FilterMultiSelectGroups.test.ts
```diff
@@ -0,0 +1,209 @@
...
+    {value: "KILLING", label: "KILLING"},
+    {value: "PAUSED", label: "PAUSED"},
+    {value: "BREAKPOINT", label: "BREAKPOINT"},
+    {value: "SUCCESS", label: "SUCCESS"},
+    {value: "WARNING", label: "WARNING"},
+    {value: "FAILED", label: "FAILED"},
+    {value: "KILLED", label: "KILLED"},
+    {value: "CANCELLED", label: "CANCELLED"},
+    {value: "SKIPPED", label: "SKIPPED"},
+    {value: "RETRIED", label: "RETRIED"},
```
elevatebart: I believe this list exists in the design system

elevatebart (approved):
My only nitpick is with unit tests in general. I find them hard to read and not valuable enough to be run and fixed after every change.

You tell me

PR #17071, opened 2026-06-26
fix(repository): complete SQL-injection backport on 1.0.x (test compile + flow-label NOT_EQUALS)

nkwiatkowski:
## What

Finishes the incomplete SQL-injection label-filter backport on `releases/v1.0.x`, which left the branch failing to compile on both `kestra` (OSS) and `kestra-ee`.

Two gaps remained after the earlier backport commits (`4d567de62`, `b6bc2da93`):

1. **Test compile failure** — `AbstractSQLInjectionTest` lost the `Flow` / `FlowWithSource` / `GenericFlow` / `FlowRepositoryInterface` imports and the injected `flowRepository` field during the merge, while the test body still uses them. `:jdbc:compileTestJava` failed → broke the whole build (and EE, which compiles the OSS `jdbc` module directly).

2. **Flow-label `NOT_EQUALS` behavior** — the `isNull().or(...)` fix was only backported for Postgres. On H2 and MySQL an absent label key produced `NULL <> 'x'` (= NULL, not true), so `NOT_EQUALS` with an absent key returned no row. This is the `flowLabelFilterShouldResistSqlInjection` assertion.

## Changes

- `AbstractSQLInjectionTest`: restore 4 imports + the `flowRepository` field.
- `H2FlowRepositoryService`: `NOT_EQUALS` → `valueField.isNull().or(valueField.ne(value))`.
- `MysqlFlowRepositoryService`: add the `NOT_EQUALS` branch (+ import), mirroring v1.3.x.

All three mirror the already-green `releases/v1.3.x` backport. Postgres already had the fix.

## Verification

- `./gradlew :jdbc:compileTestJava` ✅
- `./gradlew :jdbc-h2:test --tests "*SQLInjection*"` → **3 tests, 0 failures, 0 skipped** ✅
- `:jdbc-mysql` / `:jdbc-postgres` compile ✅ (their DB-specific injection tests run via CI testcontainers)

### DIFF
--- a/jdbc-h2/src/main/java/io/kestra/repository/h2/H2FlowRepositoryService.java
+++ b/jdbc-h2/src/main/java/io/kestra/repository/h2/H2FlowRepositoryService.java
@@ -49,7 +49,7 @@ public abstract class H2FlowRepositoryService {
                 Field<String> valueField = DSL.field("JQ_STRING(\"value\", CONCAT('.labels[]? | select(.key == \"', {0}, '\") | .value'))", String.class, DSL.val(H2Functions.escapeJqString((String) key), String.class));
                 Condition condition = switch (operation) {
                     case EQUALS -> value == null ? valueField.isNull() : valueField.eq((String) value);
-                    case NOT_EQUALS -> value == null ? valueField.isNotNull() : valueField.ne((String) value);
+                    case NOT_EQUALS -> value == null ? valueField.isNotNull() : valueField.isNull().or(valueField.ne((String) value));
                     default -> throw new UnsupportedOperationException("Unsupported operation: " + operation);
                 };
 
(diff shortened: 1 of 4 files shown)


nkwiatkowski:
Pushed a follow-up commit for the MySQL CI failure.

**Root cause:** `MysqlRepository.fullTextCondition` on 1.0.x used `MATCH ... AGAINST` only. In the test transaction InnoDB's FULLTEXT index doesn't see the just-inserted row, so a literal query returned 0 rows — and bare `%` / `_` weren't escaped. Backported the v1.3.x implementation (`FULLTEXT MATCH` **OR** `LIKE '%escaped%'` with `%`/`_`/`\` escaped), keeping the 1.0.x constructor signature. H2 and Postgres `fullTextCondition` already matched v1.3.x.

**Verified locally against real databases (all green):**

| DB | `*SQLInjectionTest` |
|----|---------------------|
| H2 | 3/3 ✅ |
| MySQL | 3/3 ✅ |
| Postgres | 3/3 ✅ |

loicmathieu (approved):
LGTM

PR #17073, opened 2026-06-26
fix(storage): optimise logs table indexes

loicmathieu:
## Summary

Fixes a MySQL timeout on large log datasets (~15 GB) when browsing flow-scoped logs
(filters: `tenant_id`, `namespace`, `flow_id`, `timestamp >=`). 

Fixes kestra-io/kestra-ee#6714.

**Root cause:** no index covered `flow_id`. The best existing index,
`ix_tenant_namespace_timestamp (tenant_id, namespace, timestamp, level)`, forced MySQL to
range-scan the full time window and filter `flow_id` as a residual — effectively a full scan.

**Changes (drop 2, add 1 — net 7 → 6 indexes per dialect):**

- **DROP** `ix_execution_id / logs_execution_id` — redundant strict prefix of
  `ix_execution_id__task_id`; all `execution_id`-only queries use the composite prefix.
- **DROP** `ix_timestamp / logs_timestamp` — redundant given `ix_tenant_timestamp
  (tenant_id, timestamp, level)`; every timestamp-range query on the logs table always carries
  a `tenant_id` predicate via `defaultFilter(tenantId)` → `buildTenantCondition`.
- **ADD** `ix_tenant_namespace_flow_id_timestamp (tenant_id, namespace, flow_id, timestamp, level)` —
  lets the engine seek directly to the `(tenant, namespace, flow)` slice and range-scan only
  that flow's timestamps.

Applied as migration `2.0.11-logs-flow-index` to all three dialects (MySQL, Postgres, H2).
Baselines are intentionally left unchanged (they are checksummed by `V2_0Migration` and
modifying them would fail startup on existing installations).

## Test plan

- [x] H2 tests: 930/930 passing (`./gradlew :jdbc-h2:test`)
- [ ] MySQL tests: `./gradlew :jdbc-mysql:test` (requires Docker)
- [ ] Postgres tests: `./gradlew :jdbc-postgres:test` (requires Docker)
- [ ] Idempotency: run migration twice on fresh and existing DB — should be a no-op on second run
- [ ] On MySQL: `EXPLAIN SELECT value FROM logs WHERE tenant_id=? AND namespace=? AND flow_id=? AND timestamp>=?` confirms `key = ix_tenant_namespace_flow_id_timestamp`
- [ ] On MySQL: `EXPLAIN SELECT value FROM logs WHERE execution_id=?` still resolves via `ix_execution_id__task_id` prefix after `ix_execution_id` is dropped

### DIFF
--- a/jdbc-mysql/src/main/resources/migrations/2.0.11-logs-flow-index-mysql.sql
+++ b/jdbc-mysql/src/main/resources/migrations/2.0.11-logs-flow-index-mysql.sql
@@ -1,0 +1,23 @@
+-- Drop the redundant ix_execution_id index: it is a strict leftmost prefix of
+-- ix_execution_id__task_id (execution_id, task_id) and provides no extra coverage.
+SET @idx_exists = (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema = DATABASE() AND table_name = 'logs' AND index_name = 'ix_execution_id');
+SET @sql = IF(@idx_exists > 0, 'ALTER TABLE logs DROP INDEX ix_execution_id', 'SELECT 1');
+PREPARE stmt FROM @sql;
+EXECUTE stmt;
+DEALLOCATE PREPARE stmt;
+
+-- Drop the redundant ix_timestamp index: every timestamp-range query on logs always carries a
+-- tenant_id predicate, so ix_tenant_timestamp (tenant_id, timestamp, level) covers it fully.
+SET @idx_exists = (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema = DATABASE() AND table_name = 'logs' AND index_name = 'ix_timestamp');
+SET @sql = IF(@idx_exists > 0, 'ALTER TABLE logs DROP INDEX ix_timestamp', 'SELECT 1');
+PREPARE stmt FROM @sql;
+EXECUTE stmt;
+DEALLOCATE PREPARE stmt;
+
+-- Add a composite index for the flow-scoped logs query (tenant + namespace + flow_id + timestamp
+-- range). Fixes full-scan on large log datasets when filtering by namespace, flow_id and timestamp.
+SET @idx_exists = (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema = DATABASE() AND table_name = 'logs' AND index_name = 'ix_tenant_namespace_flow_id_timestamp');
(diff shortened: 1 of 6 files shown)


jdbc-mysql/src/main/resources/migrations/2.0.11-logs-flow-index-mysql.sql
```diff
@@ -0,0 +1,23 @@
...
+SET @idx_exists = (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema = DATABASE() AND table_name = 'logs' AND index_name = 'ix_timestamp');
+SET @sql = IF(@idx_exists > 0, 'ALTER TABLE logs DROP INDEX ix_timestamp', 'SELECT 1');
+PREPARE stmt FROM @sql;
+EXECUTE stmt;
+DEALLOCATE PREPARE stmt;
+
+-- Add a composite index for the flow-scoped logs query (tenant + namespace + flow_id + timestamp
+-- range). Fixes full-scan on large log datasets when filtering by namespace, flow_id and timestamp.
+SET @idx_exists = (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema = DATABASE() AND table_name = 'logs' AND index_name = 'ix_tenant_namespace_flow_id_timestamp');
+SET @sql = IF(@idx_exists = 0, 'ALTER TABLE logs ADD INDEX ix_tenant_namespace_flow_id_timestamp (tenant_id, namespace, flow_id, timestamp, level), ALGORITHM=INPLACE, LOCK=NONE', 'SELECT 1');
```
AcevedoR, line 20: is there no way to find an async/lighter way to do this migration for big Log tables ?

loicmathieu: > ALTER TABLE logs ADD INDEX ix_tenant_namespace_flow_id_timestamp (tenant_id, namespace, flow_id, timestamp, level), ALGORITHM=INPLACE, LOCK=NONE

This is async in MySQL as LOCK=NONE

AcevedoR: cool then lgtm

AcevedoR: maybe worth adding a comment if you can ^^

AcevedoR (approved):

PR #17077, opened 2026-06-26
fix(core): match enum task states in tasksWithState()

AcevedoR:
## What

`tasksWithState()` silently returns an empty list whenever a task's `state` is a `State.Type` enum rather than a `String`. This makes `{{ tasksWithState('X') is (not) empty }}` always evaluate as *empty* inside `runIf`, even when matching tasks exist.

## Why

The `tasks` Pebble variable (`RunVariables.computeTasksMap`) stores each task's state as `taskRun.getState().getCurrent()` — a **`State.Type` enum**. `TasksWithStateFunction` only matched `valueObj instanceof String`, so the enum was never matched.

The two render contexts differ in how `state` is typed:

| Render context | Where | `tasks.<id>.state` type | Result |
|---|---|---|---|
| `runIf` | **executor** (`ExecutionEventMessageHandler`, `ExecutorService`) — in-memory vars | `State.Type` **enum** | ❌ always empty |
| `message` / outputs | **worker** — vars round-tripped through JSON | `String` | ✅ works |

So `tasksWithState()` is effectively broken in any `runIf`, while it works in task messages — exactly the discrepancy reported.

## Fix

Normalize the state value (String **or** enum) to its uppercase name before comparison, in both the direct-`state` and nested-value (loop/iteration) branches.

## Test (TDD)

Added `TasksWithStateFunctionTest`:
- `shouldMatchWhenStateIsString` — worker-side representation (was already green)
- `shouldMatchWhenStateIsEnum` — executor/`runIf` representation (**red → green**)
- `shouldMatchNestedValueWhenStateIsEnum` — loop/iteration nested branch (**red → green**)

## Reproducer

```yaml
id: taskswithstate_runif
namespace: company.team
tasks:
  - id: hello
    message: Hello World!
  - id: run_if_no_success
    runIf: "{{ tasksWithState(state='SUCCESS') is empty }}"
    message: "{{ tasksWithState(state='SUCCESS') }}"
```

Before: `run_if_no_success` **runs** (runIf sees the SUCCESS list as empty). After: it is correctly **skipped**.

Fixes Pylon #1984. OSS-only — EE consumes this function from `core` and inherits the fix.

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/pebble/functions/TasksWithStateFunction.java
+++ b/core/src/main/java/io/kestra/core/runners/pebble/functions/TasksWithStateFunction.java
@@ -50,35 +50,28 @@ public class TasksWithStateFunction implements KestraFunction {
                         transformedTask.put("taskId", taskId);
 
                         if ("state".equals(key)) {
-                            if (valueObj instanceof String) {
-                                String state = ((String) valueObj).toUpperCase();
+                            if (valueObj instanceof Map) {
+                                Map<String, Object> nestedMap = (Map<String, Object>) valueObj;
+                                String state = toState(nestedMap.get("state"));
+                                if (stateToFilter.equals(state)) {
+                                    transformedTask.put("state", state);
+                                    transformedTask.put("value", "state");
+                                    filteredTasks.add(transformedTask);
+                                }
+                            } else {
+                                String state = toState(valueObj);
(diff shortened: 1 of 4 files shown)


nkwiatkowski (approved):

AcevedoR:
need backport

PR #17086, opened 2026-06-26
fix(namespaces): preserve spaces and distinguish '+' in namespace file paths

### LINKED ISSUE #16896: Namespace files: filename with space collides with literal '+' → silent overwrite / data loss
## Summary
Namespace files whose name contains a **space** collide with files whose name contains a literal **`+`**: they map to the same internal-storage object, so uploading one silently overwrites the other. Spaces are also persisted/displayed as `+` everywhere (listings, search, UI).

## Root cause
`NamespaceFileController` round-trips the `path` query parameter through `URLEncoder.encode(path, StandardCharsets.UTF_8)` and then parses it back with `new URI(...).getPath()`:

- upload: `webserver/src/main/java/io/kestra/webserver/controllers/api/NamespaceFileController.java:223`
- read / stats / revisions / list / create-dir / delete: lines `84`, `102`, `127`, `155`, `181`, `329`

`URLEncoder.encode` uses `application/x-www-form-urlencoded` rules, so a space becomes `+`. `URI.getPath()` does **not** decode `+` back into a space (that is form-encoding semantics, not URI semantics), so the literal `+` becomes part of the stored object name. Both `" "` and `"+"` therefore resolve to the same path. Reads apply the identical mangling, which makes the round-trip self-consistent via the API and masks the bug — but the object on disk and in every listing genuinely carries a `+`, and the two logical names share one object.

## Reproduction
```bash
B=http://localhost:8080/api/v1/main; A=(-u 'demo:DemoDemo1!'); NS=io.kestra.qa
printf 'SPACE-version' > /tmp/s; printf 'PLUS-version' > /tmp/p

# upload "a b.txt" (space, %20)
curl -s "${A[@]}" -X POST "$B/namespaces/$NS/files?path=/c/a%20b.txt" -F fileContent=@/tmp/s
# upload "a+b.txt" (literal plus, %2B)
curl -s "${A[@]}" -X POST "$B/namespaces/$NS/files?path=/c/a%2Bb.txt" -F fileContent=@/tmp/p

curl -s "${A[@]}" "$B/namespaces/$NS/files/directory?path=/c"   # -> single entry "a+b.txt"
curl -s "${A[@]}" "$B/namespaces/$NS/files?path=/c/a%20b.txt"   # -> "PLUS-version" (the space file is gone)
```

**Expected:** two distinct files `a b.txt` and `a+b.txt`, each with its own content.
**Actual:** one file `a+b.txt`; reading back `a b.txt` returns the `+`-file's content. The first upload is silently lost.

## Impact
Medium — filenames with spaces are common. An upload can silently clobber an unrelated existing file, and the stored/displayed name is wrong (`my file.txt` → `my+file.txt`).

## Suggested fix
Stop using `URLEncoder.encode` over the whole path. Treat `path` as an already-decoded string and build the URI with proper per-segment URI encoding (or `URI`'s multi-arg constructor), so `+` and space remain distinct and spaces are preserved.

## Environment
Kestra 2.0 EE, `develop` @ `6075c7fb2`, ELS+Kafka backend, `local` storage (`LocalStorage`). Found during 2.0 QA of the Internal Storage interface.

loicmathieu:
## Summary

- Fixes a silent data-loss bug where uploading a file named `a b.txt` (space) would silently overwrite an existing file named `a+b.txt` (literal `+`), and vice versa, because both names collapsed to the same internal storage object.
- Root cause: `URLEncoder.encode()` (form-encoding rules: space→`+`) was applied to the already-decoded `path` query parameter in `NamespaceFileController`, and `URI.getPath()` does not reverse form-encoding — so `" "` and `"+"` both became `"+"` in the stored path.
- Same bug existed one level deeper in `NamespaceFile.of(String, String, int)`, where `URI.create()` was used to build the internal `kestra://` storage URI and would throw `IllegalArgumentException` for any path containing a space.

**Changes:**
- `NamespaceFileController`: replace all 7 `new URI(URLEncoder.encode(path, UTF_8))` calls with a shared `toFileUri(path)` helper using `new URI(null, null, path, null)` (multi-arg constructor encodes URI-illegal chars only, space→`%20`, `+`→`+`). Remove now-unused `URLEncoder`/`StandardCharsets` imports. Fix ZIP import path (same `URI.create()` issue for space-named entries).
- `NamespaceFile.of(String, String, int)`: replace `URI.create(KESTRA_PROTOCOL + rawPath)` with `new URI(KESTRA_SCHEME, "", rawPath, null)` so the internal storage URI is properly encoded.

## Test plan

- [ ] `NamespaceFileControllerTest.shouldNotCollideSpaceAndPlusInFileName` — new integration test: upload `a b.txt` and `a+b.txt`, read each back via HTTP, assert distinct content; check directory listing shows two entries with correct names.
- [ ] `NamespaceFileTest.shouldDistinguishSpaceAndPlusInStorageUri` — new unit test: asserts `NamespaceFile.of()` produces `%20`-encoded URI for space filenames, preserves `+` for plus filenames, and that `storagePath()` round-trips correctly.
- [ ] All 22 `NamespaceFileControllerTest` tests pass (was 21 before).
- [ ] All 18 `NamespaceFileTest` tests pass (was 17 before).

Closes #16896

### DIFF
--- a/core/src/main/java/io/kestra/core/storages/NamespaceFile.java
+++ b/core/src/main/java/io/kestra/core/storages/NamespaceFile.java
@@ -1,6 +1,7 @@
 package io.kestra.core.storages;
 
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.file.Path;
 import java.util.Objects;
 import java.util.regex.Matcher;
@@ -146,12 +147,21 @@ public record NamespaceFile(
             storagePath += ".v" + revision;
         }
 
-        return new NamespaceFile(
-            pathWithoutLeadingSlash,
-            URI.create(StorageContext.KESTRA_PROTOCOL + namespacePrefixPath.resolve(storagePath).toString().replace("\\", "/") + (isDirectory(path) ? "/" : "")),
-            namespace,
-            revision
-        );
+        // Use the multi-argument URI constructor so that URI-illegal characters (e.g. spaces)
+        // are percent-encoded (space → %20) while legal URI path characters (e.g. '+') are
+        // preserved unchanged — keeping "a b.txt" and "a+b.txt" as two distinct stored objects.
+        String uriPath = namespacePrefixPath.resolve(storagePath).toString().replace("\\", "/")
+            + (isDirectory(path) ? "/" : "");
+        try {
+            return new NamespaceFile(
+                pathWithoutLeadingSlash,
+                new URI(StorageContext.KESTRA_SCHEME, "", uriPath, null),
+                namespace,
+                revision
+            );
+        } catch (URISyntaxException e) {
+            throw new IllegalArgumentException("Invalid namespace file path: " + path, e);
+        }
     }
 
     public static Path normalize(Path path) {
--- a/core/src/test/java/io/kestra/core/storages/NamespaceFileTest.java
+++ b/core/src/test/java/io/kestra/core/storages/NamespaceFileTest.java
@@ -173,4 +173,25 @@ class NamespaceFileTest {
         assertThat(toLogicalPath(NamespaceFile.normalize(Path.of("/file..with..dots")))).isEqualTo("/file..with..dots");
     }
 
+    @Test
+    void shouldDistinguishSpaceAndPlusInStorageUri() {
+        // Regression: "a b.txt" (space) and "a+b.txt" (literal '+') must map to distinct storage URIs.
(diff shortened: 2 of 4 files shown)


nkwiatkowski (approved):

loicmathieu:
Backported to 1.3

PR #17099, opened 2026-06-29
fix(executions): nullable inputs cannot be passed from parent flow to subflow

### LINKED ISSUE #6624: Not possible to dynamically pass nullable inputs from parent flow to subflow
### Describe the issue

Hello team,

I have a flow that triggers subflow via Subflow task
```
id: TEST_NULL_ZPRACOVANI
namespace: csutrf
inputs:
- id: integerNull
  type: INT
  required: false
labels: []
tasks:
- id: TEST_NULL_1_subflow
  namespace: csutrf
  flowId: TEST_NULL_1
  inputs:
    integerNull: "{{ inputs.integerNull }}"
concurrency:
  limit: 1
  behavior: CANCEL

id: TEST_NULL_1
namespace: csutrf
inputs:
- id: integerNull
  type: INT
  required: false
labels: []
tasks:
- id: log1
  message: "{{ inputs.integerNull }}"
```

I need to pass input variable from parent flow to the subflow. It works fine until the value is null, when I trigger parent flow with input that has flag `required = false`. The Subflow task throws the following error
```
integerNull: Invalid input for integerNull, For input string: "", but received `` 
```
I tried to modify the expression
```
"{{ inputs.integerNull }}" -> "{{ inputs.integerNull ?? 'null' }}" 
```
But it also throws an error
```
integerNull: Invalid input for integerNull, For input string: "null", but received null
```

Is it possible to somehow pass an input with a null value to the subflow?

P.S.: It's also not possible to achieve with TemplatedTask as it supports only Runnable tasks

### Environment

- Kestra Version: 0.18.4

loicmathieu:
## Summary

Fixes #6624.

When a parent flow passes an optional non-text input (e.g. `INT`) to a subflow via a Pebble expression such as `{{ inputs.integerNull }}` and the parent value is null, Pebble renders the null reference as an empty string `""`. `FlowInputOutput.resolveInputValue` only treated literal `null` as "absent", so `""` fell through to `Integer.valueOf("")` which threw:

```
integerNull: Invalid input for integerNull, For input string: "", but received ``
```

**Root cause:** `Subflow.createSubflowExecutions` calls `runContext.render(this.inputs)` — Pebble renders a null value as `""`. The input resolution code only normalized literal `null`, not empty strings.

**Fix:** In `FlowInputOutput.resolveInputValue`, coerce `""` to `null` for non-text input types before the default/required check. Text-like types (`STRING`, `SELECT`, `EMAIL`, `SECRET`) keep `""` as a legitimate value. This fix applies to all entry paths (subflow, webhook, API) since they all share `resolveInputValue`.

## Changes

- `core/.../runners/FlowInputOutput.java` — 2-line normalization + `isTextType` helper
- `core/.../runners/FlowInputOutputTest.java` — 6 new unit tests (INT/FLOAT/STRING/EMAIL edge cases)
- `core/.../flow/SubflowRunnerTest.java` — end-to-end test reproducing the reported scenario
- 2 new YAML flow fixtures for the end-to-end test

## Test plan

- [x] `FlowInputOutputTest` — 45/45 passing
- [x] `SubflowRunnerTest` — 5/5 passing (includes new `shouldPassNullableInputFromParentToSubflow`)
- [x] `H2RunnerTest` — 100/100 passing

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
@@ -358,6 +358,11 @@ public class FlowInputOutput {
 
             Object value = resolvable.get().value();
 
+            // Pebble renders a null reference as ""; treat "" as absent for non-text types.
+            if (value instanceof String s && s.isEmpty() && !isTextType(input.getType())) {
+                value = null;
+            }
+
             // resolve default if needed; a `defaults` that is a Pebble expression (e.g. subflow()/secret())
             // can itself fail to render — that is also a broken field, so flag it as a render error.
             if (value == null && input.getDefaults() != null) {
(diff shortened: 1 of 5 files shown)


nkwiatkowski (approved):

loicmathieu:
backported to 1.3

PR #17101, opened 2026-06-29
fix(core): executor service metric names collide for worker agent and…

loicmathieu:
… system worker

Fixes https://github.com/kestra-io/kestra-ee/issues/8389


fhussonnois (approved):
LGTM!

PR #17103, opened 2026-06-29
fix(core): throw an illegal exception when a sort is null or empty

loicmathieu:
Defense in depth


nkwiatkowski (approved):

loicmathieu:
Backported to 1.0 and 1.3

PR #17115, opened 2026-06-29
ci: run EE compile check for OSS PRs via Kestra webhook

AcevedoR:
Run the kestra-ee `compileJava` check for an OSS PR's ref through a Kestra webhook and surface the result inline as a required status — replacing the async `oss-updated` repository_dispatch round-trip.

here is the Flow that does everything: https://prd.kestra-tech.kestra.cloud/ui/main/flows/edit/kestra.engineering.ci/run_backend_ee_ci_webhook/edit

The flow resolves the matching kestra-ee ref itself (same-name branch, else `develop`), so no branch pre-check is needed. Consumes the `KESTRA_CI_EEBUILD_WEBHOOK_URL` repo secret (published by terraform in flows-engineering).

Adds a shared composite action `.github/actions/ee-compile-check`, used by both the PR job (`pull-request.yml`) and an on-demand manual workflow (`ee-compile-check-manual.yml`).

The legacy `oss-updated` EE trigger is kept but disabled (`if: false && …`) for quick rollback.

### DIFF
--- a/.github/actions/ee-compile-check/action.yml
+++ b/.github/actions/ee-compile-check/action.yml
@@ -1,0 +1,135 @@
+name: "EE compile check via Kestra webhook"
+description: >-
+  Run the kestra-ee compileJava check for an OSS ref on a Kestra instance via a
+  synchronous webhook, and surface the result inline (job summary + log group).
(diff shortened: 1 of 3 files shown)


Skraye (approved):
IMO Kestra should handle the feedback to the PR

But this still LGTM

Issue #132, opened 2020-06-29
Log isolation between execution


tchiotludo:
Log on the worker jvm can be shared and displayed on other execution. 
Maybe because SL4J is shared.. 
Need to have a proper instance completely isolated

PR #17117, opened 2026-06-29
feat(ui): stream task logs to the topology-details plugin slot

fdelbrayelle:
## What

Adds an optional `logs` prop to the `topology-details` plugin UI slot, fed by a live `followLogs` SSE subscription, so plugin slots can follow per-step progress **while a task is running**.

## Why

Plugin `topology-details` slots currently receive `metrics` and `execution`, but neither carries live per-step progress during a run:

- **Metrics** accumulate in-memory and flush only *after* the task run completes (`WorkerTaskProcessor`), then go through the indexer. The UI also only loads them on demand (Metrics tab).
- **Outputs** likewise only exist post-completion.
- The **execution SSE follow** stream emits only on execution *state* changes, so it's static through a single long-running task.

The **log stream is the only channel that delivers data mid-task** (`/logs/{id}/follow`). This PR routes it to the slot so a plugin (e.g. the Kubernetes runner) can advance a live lifecycle stepper instead of jumping straight from "started" to "all done".

## How

- **`topology-details` contract**: add optional `logs: LogEntry[]`.
- **`LowCodeEditor`**: pass per-task logs to the slot, filtered by `taskId` (mirrors the existing `taskMetrics` pattern).
- **`executions/Topology`**: own the `followLogs` SSE lifecycle — subscribe only while the execution is running, buffer + throttle-flush into a reactive array, and **close on error** to avoid the EventSource reconnect leak documented in #16982.

The prop is **optional and additive** — existing slots are unaffected.

## Companion PR

The consumer side lives in the Kubernetes EE plugin (`plugin-ee-kubernetes` PR #132), which emits `[lifecycle] <step>` markers and parses them from these streamed logs to drive the live stepper.

## Test

- `oxlint` + `eslint` clean on the changed files.
- Manual QA of the live stepper is done via the companion plugin PR.

### DIFF
--- a/ui/src/components/executions/Topology.vue
+++ b/ui/src/components/executions/Topology.vue
@@ -58,52 +58,47 @@
     // FIXME: any - SSE objects don't have a consistent type in this codebase
     const sseBySubflow = ref<Record<string, any>>({}) // FIXME: any
 
-    // Live task logs, streamed only while the execution is running, so plugin
+    // Live task logs, polled while the execution is running, so plugin
     // topology-details slots can follow per-step progress in real time (metrics
     // and outputs only materialise once the task run completes).
+    //
+    // Polled, not streamed: the follow SSE only reliably delivers the snapshot
+    // replayed at connect time, and its realtime push isn't a dependable catch-up
+    // path — a stable connection never reconnects, so any lifecycle marker emitted
+    // after connect froze the stepper until the execution ended. A 1s GET snapshot
+    // is plenty (steps last seconds) and avoids the reconnect loop that caused the
+    // kestra-io/kestra#16982 off-heap leak.
     const executionLogs = ref<any[]>([]) // FIXME: any
-    let logsSSE: EventSource | undefined
-    let logsBuffer: any[] = []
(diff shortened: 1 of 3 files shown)


ui/src/components/executions/Topology.vue
```diff
@@ -57,6 +58,68 @@
     // FIXME: any - SSE objects don't have a consistent type in this codebase
     const sseBySubflow = ref<Record<string, any>>({}) // FIXME: any
 
+    // Live task logs, streamed only while the execution is running, so plugin
+    // topology-details slots can follow per-step progress in real time (metrics
+    // and outputs only materialise once the task run completes).
+    const executionLogs = ref<any[]>([]) // FIXME: any
```
jymaire, line 71: fixme: to do now ?

fdelbrayelle:
Superseded by #17166 — replaces the log-poll/text-marker workaround here with a typed `progress` field on `LogEntry`/`FollowLogEvent` and a `RunContext#emitProgress` API, riding the existing follow-logs SSE instead of a 1s poll.

PR #17118, opened 2026-06-29
feat(ui): FORM wizard bottom progress bar + single primary button

brian-mulier-p:
Replaces the top numbered stepper in the FORM execution wizard with a label-less segmented progress bar at the bottom of the Inputs pane, and ensures only one primary button is shown at a time.

- `KsSteps` gains a `variant="bar"` segmented progress meter; the default `variant="steps"` rendering is unchanged, so existing usages (basic-auth setup, EE instance setup / worker-group wizard) are unaffected. Per-step state is driven by each `KsStep`'s `status` via provide/inject.
- `InputsForm` drops the top stepper, renders the sticky bottom bar (hidden on the recap, restored when editing a section from it), and shows the active FORM's `displayName` as a header above its fields — only when it has one (ungrouped inputs and FORMs without a displayName show none). The old `.wizard-steps` CSS is removed.
- A segment fills only once its step has been passed via **Next** (a sticky `visited` set in `useInputsWizard` + a new `stepStatus`), so progress reflects what's actually completed.
- The dialog-footer Execute (`FlowRunActions`) is gated on a new `FlowRun.showExecuteButton`, so the wizard shows **Next** while stepping and **Execute** only on the recap — fixing the case where both showed at once.

Unit tests cover the bar variant, the wizard `visited`/`stepStatus` state, and the Execute gating. Verified in-browser on an EE instance: the bar fills after Next, the header shows only for named FORMs, a single primary button at a time, the recap hides the bar, Edit returns and keeps prior segments filled, and FORM children show bare labels.

### DIFF
--- a/ui/packages/design-system/src/components/Navigation/KsSteps/KsSteps.vue
+++ b/ui/packages/design-system/src/components/Navigation/KsSteps/KsSteps.vue
@@ -1,5 +1,9 @@
 <template>
+    <div v-if="variant === 'bar'" class="ks-stepbar" role="list" v-bind="$attrs">
+        <slot />
+    </div>
     <ElSteps
+        v-else
         :class="{'kel-steps--small': size === 'small'}"
         v-bind="({...filteredProps(), ...$attrs} as any)"
     >
@@ -11,6 +15,7 @@
 
 <script setup lang="ts">
     import {ElSteps} from "element-plus"
+    import {provide, toRef} from "vue"
     import {useFilteredProps} from "../../../utils/filteredProps"
 
     defineOptions({inheritAttrs: false})
@@ -24,9 +29,14 @@
         simple?: boolean
         alignCenter?: boolean
         size?: "default" | "small"
+        variant?: "steps" | "bar"
     }>()
 
-    const filteredProps = useFilteredProps(props, ["size"])
+    const filteredProps = useFilteredProps(props, ["size", "variant"])
+
+    // In the "bar" variant each child KsStep renders itself as a single progress segment (instead of
+    // an ElStep), reading this injected variant — the same provide/inject pattern ElSteps/ElStep use.
+    provide("ksStepsVariant", toRef(props, "variant"))
 
     defineSlots<{
         default?(): unknown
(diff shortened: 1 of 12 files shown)


ui/src/components/inputs/InputsForm.vue
```diff
@@ -1000,31 +1011,13 @@
...
-    // Completed steps: drop the green glow too — a plain green check reads cleaner.
-    :deep(.kel-step__head.is-success .kel-step__icon) {
-        box-shadow: none;
-    }
+.wizard-progress {
+    position: sticky;
+    bottom: 0;
+    margin-top: 1rem;
+    padding: 0.75rem 0 0.25rem;
+    background: var(--ks-bg-body, var(--ks-bg-surface));
```
flcarre: `--ks-bg-body` isn't declared in any of the three theme files (`ks-theme-light/dark/dark-2.scss`) — this only renders correctly because the fallback lands on `--ks-bg-surface`. Since the fallback *is* the intended value, reference it directly so we're not relying on a phantom token (the same phantom already exists in `ksTheme.ts`/`PluginLayout.vue` — no need to propagate it).

```suggestion
    background: var(--ks-bg-surface);
```

ui/packages/design-system/src/components/Navigation/KsSteps/KsSteps.vue
```diff
@@ -36,4 +46,25 @@
...
+    .ks-stepbar {
+        display: flex;
+        align-items: center;
+        gap: var(--ks-spacing-2, 0.5rem);
+        width: 100%;
+
+        &__seg {
+            flex: 1 1 0;
+            height: 0.375rem;
+            border-radius: 999px;
```
flcarre: Optional nit: `--ks-radius-xl` exists and gives the same pill on a 6px-tall bar, so we can avoid the hardcoded px.

```suggestion
            border-radius: var(--ks-radius-xl);
```

flcarre (approved):
Clean change — the single-primary-button gating, the bottom bar, and the FORM `displayName` header all hang together well, and the test coverage (bar variant, `visited`/`stepStatus`, Execute gating) is solid. Verified the provide/inject contract (a standalone `KsStep` falls through to `ElStep`), the `visited` Set reactivity, and that `showExecuteButton` is correct for every OSS consumer.

Approving. One small token cleanup worth doing inline, plus an optional nit.

Heads-up (out of OSS scope): `FlowRunInstance.showExecuteButton` is now a **required** field — any EE code that builds a synthetic `FlowRunInstance` (rather than using the `FlowRun` component ref) will need the field added to keep TS compiling.

PR #17121, opened 2026-06-30
feat(blueprints): disable blueprints that need uninstalled plugins

MilosPaunovic:
## What

A blueprint that references a task from a plugin not installed on the instance previously failed only **after** the user opened it in the editor, with no upfront warning. This detects the situation and disables such blueprints, clearly surfacing which plugin(s) are missing.

- New `useBlueprintPlugins` composable cross-references a blueprint's `includedTasks` against the instance's installed task types (`GET /api/v1/plugins`, exposed by the plugins store as `allTypes`) — the same source the editor uses to validate, so it flags exactly the cases that would have errored.
- `BlueprintCard`: greys out the unavailable task icons, shows a warning indicator, and disables **Use** with a tooltip naming the missing plugin(s).
- `BlueprintOverview`: dims missing tasks and renders a *Missing plugins* alert; exposes a `missing-plugins-action` slot so EE can add an install shortcut.
- `BlueprintDetailView` / detail override: disable **Open in Edit Flow** when a plugin is missing.
- Translations added for all 13 languages (`check.js` clean).

## Companion PR

EE (install link + EE blueprint views): https://github.com/kestra-io/kestra-ee/pull/9073

## How tested

- `npm run check:types` (ui): clean
- ESLint clean on changed files
- New unit test `useBlueprintPlugins.spec.ts` (11 cases)
- Manually verified on a local instance (screenshots in the EE PR)

Relates to https://github.com/kestra-io/kestra-ee/issues/8390.


Piyush-r-bhaskar (approved):

Pradumnasaraf (changes requested):
On the Blueprints page, the "open in Edit flow" is disabled when the plugin is not installed, which is expected. But in the embedded Blueprint inside the flow editor, it gives me an option to use "open in Edit flow". 

https://jam.dev/c/32f3e500-5c48-4e8a-9bfb-925666b63f82

MilosPaunovic:
Thanks @Pradumnasaraf — fixed. The embedded blueprint view (blueprint browser inside the flow editor) now handles missing plugins like the standalone detail: `BlueprintEmbedView` exposes the missing-plugin state to its actions slot, so **Open in Edit Flow** is disabled with a tooltip and the *Missing plugins* alert renders in the embedded overview.

While verifying on a local instance I also caught (and fixed) a false positive: the old check compared `includedTasks` against `pluginsStore.allTypes`, which in the editor is a partial, subgroup-shaped payload that never contains conditions — so core-only blueprints (e.g. ones using a core condition) were wrongly flagged. Detection now loads the installed task/trigger types from `/api/v1/plugins` into a dedicated cache, matches on the **exact class** (so sibling artifacts like `scripts.python` vs `scripts.shell` are distinguished), and skips condition classes (the endpoint doesn't return them). Nothing is disabled until the installed set is known.

Verified live in the editor: a core-only blueprint stays enabled with no alert; a blueprint needing an uninstalled plugin is disabled with the alert. Unit tests updated to cover these cases.

MilosPaunovic:
@Pradumnasaraf sorry about the AI generated comment, I was not aware it will be there, but, nonetheless, can you check the PR again please?

Pradumnasaraf (approved):
Thanks. It's working now. Looks good

Pradumnasaraf:
> @Pradumnasaraf sorry about the AI generated comment, I was not aware it will be there, but, nonetheless, can you check the PR again please?

No worries! Done. We commented at the same time ^^

PR #17125, opened 2026-06-30
[backport releases/v1.3.x] feat(ui): FORM wizard bottom progress bar + single primary button

brian-mulier-p:
Backport of #17118 to `releases/v1.3.x`.

1.3 has a different UI structure than develop (no in-repo design-system, no `FlowRunActions`), so this is a structural port rather than a clean cherry-pick:

- The label-less segmented progress bar is **inlined** in `InputsForm.vue` (plain markup + CSS) rather than a `KsSteps variant="bar"` — 1.3's stepper was raw Element Plus `<el-steps>`, removed here. Uses 1.3 tokens (`--ks-border-active`, `--ks-border-primary`, `--ks-background-card`).
- The active FORM's `displayName` shows as a header above its fields, only when set (`WizardStep.displayName`).
- `useInputsWizard` gains a sticky `visited` set + `stepStatus` so a segment fills only once its step is passed via **Next** (verbatim from develop — the composable is byte-identical in 1.3).
- Single primary button needs no change: 1.3's `FlowRun` already gates the Execute button on `!hasFormInputs || inputsOnRecap` and has no `FlowRunActions` split (that refactor post-dates 1.3).

Updated the `InputsForm` Wizard storybook story for the inlined bar.

Original PR: #17118 — feat(ui): FORM wizard bottom progress bar + single primary button.

### DIFF
--- a/ui/src/components/inputs/InputsForm.vue
+++ b/ui/src/components/inputs/InputsForm.vue
@@ -1,14 +1,13 @@
 <template>
     <template v-if="initialInputs">
-        <!-- Wizard progress: one step per fillable step (FORM titles, "Inputs" for ungrouped runs).
-             active = currentStep, so earlier steps render as finished and the recap marks them all done. -->
-        <el-steps v-if="isWizard" :active="currentStep" finishStatus="success" class="wizard-steps" data-testid="wizard-steps">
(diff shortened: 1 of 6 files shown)


brian-mulier-p:
No need to review as it's pure backport

PR #17131, opened 2026-06-30
fix: executions-nested-output-file-preview

### LINKED ISSUE #17085: Unable to preview an output if it's nested
### Describe the issue

The new output page has a _JSON view_ to view structured output; as a result, files nested inside the output didn't have anymore the download/preview button.

Image

### Environment

- Kestra Version: develop

hemantmm:
closes: #17085

---

### ✨ Description

What does this PR change?  

- Fixed nested file output previews in the execution Input/Output tree.

- The new structured JSON/tree output view allowed users to inspect nested output values, but nested file URIs no longer triggered the preview/download UI.

https://github.com/user-attachments/assets/ed8bc90e-2b05-4dd2-97a2-7e4b9300311b


MilosPaunovic:
Hey @hemantmm, 

Can you resolve the merge conflicts that exist now and we'll get right on it?

hemantmm:
Sure.

hemantmm:

flcarre:
Thanks for picking this up @hemantmm! 🙏 The nested file preview works now, and pulling `normalizeFilePreview` out into a tested pure function (with the added `Array.isArray` guard) is a really nice cleanup.

While testing I ran into one behavior I think is worth a small tweak before merging, so I wanted to share it with a repro.

### The behavior
`onSelectPath` now sets `selectedValue` on *every* tree-row click, not just file leaves. When you click an intermediate object/array row, the tree re-roots to that child, but `selectedBase` (the tree's `basePath`) isn't updated to match. So the paths the tree rebuilds afterwards drop the segment(s) you drilled through, and the generated Pebble expression ends up wrong.

### Repro
Task output: `{values: {a: {b: "one", c: "two"}, nestedfile: "kestra:///…"}}`

1. Select the task output → the tree appears.
2. Click the `a` row (the row body, not the caret) → the tree re-roots to `{b, c}`.
3. Click `b` → the Expression panel shows `{{ outputs.bundle.b }}`, but the correct path is `{{ outputs.bundle.values.a.b }}` (the UI even surfaces an "Invalid prefix path" error).

qa-blocker-evidence

Clicking `b` directly, without first drilling into `a`, gives the correct path, so it's specific to drilling into an object. One related side effect: clicking any non-file leaf replaces the tree with a bare scalar and there's no obvious way back except reselecting from the sidebar.

### A possible direction
Since the file preview only needs the *value* (not a re-root), one option is to leave the tree's root `selectedValue` / `basePath` untouched for in-tree clicks and drive the preview from a separate ref for the "previewed leaf." Alternatively, only swap `selectedValue` for leaf rows and keep `selectedBase` in sync when you do. Whatever fits best for you.

Thanks again for the contribution, and happy to help if anything here is unclear!

hemantmm:
Thanks for the feedback @flcarre.

You were right — I was reusing `selectedValue` for nested file preview, which unintentionally re-rooted the tree while `selectedBase` still pointed to the original task output.

I will be updating the implementation so in-tree clicks no longer replace `selectedValue`. They now only update the selected expression path, while previewing is driven by a separate `previewedValue` ref. This keeps the original tree root/base path intact and still allows nested file values to open in `FilePreview`.

flcarre (commented):
Nice fix, and the demo video is helpful. A few small things:

- The new test covers `normalizeFilePreview`, but not the `previewedValue`/`onSelectPath` path that is the actual subject of the fix. A small `ExecutionVariableExplorer` test (or extracting the detection into a pure helper) would guard against a regression here.
- For a single-key wrapper like `{ myFile: { uri: ... } }`, the auto-select sets `previewedValue` to the outer object, so the preview only appears after an extra click. Mirroring `onSelectPath` (`previewedValue = selectedValue[onlyKey]`) in that branch would make the auto-select actually preview the file.
- The `vitest.config.js` coverage block looks unrelated to this fix; maybe split it out to keep the diff focused?

The nested traversal itself is a genuine general fix, which is great.

hemantmm:
Thanks for the detailed review!

You’re right, in-tree clicks should not re-root `selectedValue`, because `selectedBase` still points to the original task output and that can produce truncated Pebble paths.

I will update the logic so tree-row clicks only update the selected expression path and store the clicked value separately in `previewedValue`. File preview is now derived from that separate value, while the tree root remains unchanged.

I will also handle the single-key wrapper case by setting `previewedValue` to the only child during auto-select, so `{ myFile: { uri: ... } }` previews immediately without an extra click.

And will add a focused `ExecutionVariableExplorer` test for the sidebar selection path, and will remove the unrelated `vitest.config.js` coverage change to keep the PR focused.

flcarre (approved):
Thanks for the fix, @hemantmm — this is a clean, well-targeted solution to #17085. 🙌

The core model is exactly right: keeping `previewedValue` (the deepest node the user has drilled into, which drives the file preview) distinct from `selectedValue` (the tree root) is the correct separation, and it's applied consistently across all three selection paths — the initial select, the single-key auto-select, and the tree-click `onSelectPath`. Emitting the row `value` alongside `path` from `VariableTreeView` is a tidy way to carry the nested value through. I also really appreciate the test coverage: the three new component specs exercise precisely the regression from the issue (nested file preview, single-key auto-select, and intermediate-node navigation not re-rooting the tree), and they assert user-facing behavior rather than internals.

A couple of small, **non-blocking** observations:

- **`ui/src/stores/flow.ts` cast** — the `as Parameters<typeof FlowsAPI.updateFlow>[0]` change is unrelated to the nested-file-preview fix. It's genuinely necessary (it resolves a `vue-tsc` error inherited from the recent SDK refactor and mirrors the identical cast already used on the sibling `createFlow` call), so no objection to it landing — but if it's easy, splitting it into its own small `fix(flows):` commit would keep the history cleaner since it touches an unrelated store.
- **`normalizeFilePreview` extraction** — nice bonus: adding the `Array.isArray(data.content)` guard makes the Ion workaround strictly safer than the previous unguarded `.some(...)`, and it's now unit-tested. 👍
- **(optional)** A tiny standalone `VariableTreeView.spec.ts` would make that component's new `select` emit contract self-verifying; right now it's covered transitively through `ExecutionVariableExplorer.spec.ts`, which is fine.

Approving — great work. 🚀

hemantmm:
Thanks a lot for the thoughtful review and approval, @flcarre!

Agreed that the `flow.ts` cast is unrelated to the nested preview fix. I kept it because it was needed to get `vue-tsc` green after the SDK refactor, but I’ll move it into a separate `fix(flows): ...` commit to keep the history cleaner.

Good point about `VariableTreeView.spec.ts` as well. Since the behavior is already covered through `ExecutionVariableExplorer.spec.ts`, I’ll leave it as-is for this PR unless you think a direct component spec would be valuable.

Thanks again for the detailed review and the kind words! 🙌

flcarre (approved):

PR #17132, opened 2026-06-30
feat(helm): add kestra 2.0 grpc worker-controller support

### LINKED ISSUE #16791: Update the Helm Chart to be with compatible 2.0
### Feature description

**Mandatory:**
- Current helm chart in Develop must be updated to be compatible with the changes made in 2.0

We keep a single helm repo, so all change can be made in develop and Kestra 2 release will publish the new helm

MarcoSaba:
## Summary

Adds Kestra 2.0 gRPC worker-controller compatibility to the Helm chart. The chart now exposes port 50051 on all pods and the standalone Service by default. A new optional `controller` deployment can be enabled for fully distributed mode, backed by a dedicated ClusterIP Service at `<release>-controller:50051`.

## Changes

- Add `deployments.controller` (disabled by default), `service.ports.grpc` (50051), and `controller-service.yaml` for distributed gRPC control plane
- Document standalone, standalone+worker, and fully distributed deployment patterns with upgrade prerequisites

close kestra-io/kestra#16791


MarcoSaba:
**Follow-up (not blocking this PR):** Enterprise Edition adds a worker registration/auth step on top of the OSS gRPC controller (`kestra.ee.worker.auth.enabled`, opt-in, off by default). Workers need a kestra.worker.auth.registrationToken` to bootstrap credentials — either pre-set in config or generated post-deploy via the EE worker-group UI.

This PR only covers the OSS chart and plaintext gRPC (no auth). But once an EE chart/overlay exists, it will need a way to inject that registration token securely (e.g. a Secret + `extraEnvFrom`, following the chart's existing `configurations.secrets` pattern) rather than a plain values.yaml field.

AcevedoR (commented):
If you tried it already and made it work it should be fine.
I'm adding Florian as a reviewer he his the author of this feature.

Additionally did you try to use a Core plugin like https://kestra.io/plugins/core/execution/io.kestra.plugin.core.execution.labels ? it would be interesting just to be sure it's working well: some of our Core plugins access the database, which is an edge case in v2 architecture, so we implemented an authentication for these edge cases, worth to check it works well with what you configured

AcevedoR (approved):

MarcoSaba:
Tested io.kestra.plugin.core.execution.Labels in distributed mode on rc2 (gRPC worker-controller, Postgres backend). Created and ran a flow via the API, confirmed the execution completed SUCCESS and the label was actually persisted (testKey: testValue showed up in the execution's final state), no auth errors in worker, controller, or executor logs. The DB-access edge case works fine with the current chart setup.

PR #17134, opened 2026-06-30
fix(worker): emit failed result when task fails on its own during graceful shutdown drain

### LINKED ISSUE #17124: Task that fails during a worker's graceful shutdown is orphaned in RUNNING forever
## Bug

If a task reaches a terminal **FAILED** state during the window in which its worker is gracefully shutting down (SIGTERM / rolling deploy / scale-down), the worker discards the task result and the execution stays stuck in **RUNNING** indefinitely. It is never marked FAILED and never resubmitted.

This is not limited to tasks that were interrupted by the shutdown — a task that fails **on its own** (e.g. a script container exiting non-zero) during the shutdown drain window is also affected.

### Steps to Reproduce

1. Start a flow with a task that runs for a little while and then fails on its own (e.g. a `Script`/`Commands` task on a remote task runner whose process exits with code 1).
2. While that task is running, gracefully terminate the worker executing it (SIGTERM — the normal Kubernetes rolling-restart / scale-down path), so the worker drains and reports `TERMINATED_GRACEFULLY`.
3. Let the task's failure land during the worker's termination-grace drain window.

### Expected Behavior

The task is marked FAILED (or is reliably resubmitted to another worker and reaches a terminal state). The execution leaves RUNNING.

### Actual Behavior

The execution is stuck in RUNNING forever. The task's last logs show the real failure (e.g. `Container exited with code 1`) but no terminal state is ever recorded. The only remedy is to manually Kill the execution.

Observed on a Kestra **1.3.23** instance: an execution stuck RUNNING for 11 days whose `index` task logged a container exit-1 failure, then went silent. The stuck executions cluster on a few discrete dates — the signature of worker restarts/deploys, not random failures.

### Likely cause (from source reading, v1.3.23)

- `DefaultWorker.run()` (~lines 755-761): when `shutdown.get()` is true and the task ended in a failed state (and restart strategy != NEVER), the worker **returns without emitting the `WorkerTaskResult`**, assuming it will be resubmitted. `shutdown` is set at the very start of `close()`, so the whole grace period is affected. This cannot distinguish a shutdown-interrupted task from a genuinely-failed one, so it drops real terminal failures.
- The resubmission contract does not cover gracefully-terminated workers: `AbstractServiceLivenessCoordinator.filterAllUncleanShutdownServices` excludes `TERMINATED_GRACEFULLY` instances, and `JdbcServiceLivenessCoordinator.handleAllWorkersForUncleanShutdown` resubmits only realtime triggers (not regular `WorkerTask`s) for graceful workers. So a graceful worker's dropped task results are never re-run. (An *unclean* `TERMINATED_FORCED` death self-heals, because those tasks *are* resubmitted.)

### Suggested direction

Fix at the worker: only drop/skip-emit for tasks that were actually interrupted by the shutdown (e.g. gate on the killed/interrupt flag); always emit the terminal state for tasks that completed on their own during the drain window. Resubmitting graceful workers' tasks instead would risk duplicating tasks that genuinely succeeded/failed during drain.

brian-mulier-p:
Backport of #17129 to `releases/v1.3.x`.

Closes #17124

During a worker's graceful shutdown drain window, `shutdown.get()` returns `true` for the entire window — not only when the worker forcibly interrupts a still-running task. This caused tasks that reached a FAILED state on their own during the drain to have their `WorkerTaskResult` silently dropped, leaving the execution stuck in RUNNING forever.

Fix: introduce `shutdownInterrupted` (`AtomicBoolean` set only when `executorService.shutdownNow()` is actually called) and gate the early-return on both `shutdown.get() && shutdownInterrupted.get()`. Tasks that fail organically during the drain window fall through and emit their terminal result normally.

`shutdownInterrupted` is set to `true` before `shutdownNow()` in both paths:
- the `skipGracefulTermination` (force-shutdown) path
- the `waitForTasksCompletion` path when the grace period elapses

Regression tests in `WorkerTest` cover:
- task fails on its own during drain (result must be emitted — was the bug)
- task interrupted by forced shutdown (result must be dropped for resubmission)


brian-mulier-p:
Wrong head branch (picked up develop worktree instead of 1.3 branch). Recreating.

Issue #9134, opened 2025-05-28
Filter on flow page and sending error and the charts are not displayed


tchiotludo:
![Image](https://github.com/user-attachments/assets/5e957a95-cf34-43d3-aa67-595022457b1e)

the search is `text!=sanitychecks​ text=core` 

https://postgres-ee.preview.dev.kestra.io/ui/main/flows?filters[q][NOT_EQUALS]=sanitychecks%E2%80%8B&filters[q][EQUALS]=core

```
io.kestra.webserver.controllers.api.DashboardController$PreviewRequest["globalFilter"] > io.kestra.webserver.models.GlobalFilter["filters"] > java.util.ArrayList[0] > io.kestra.core.models.QueryFilter["field"]: Failed to convert argument [previewRequest] for value [null] due to: Cannot construct instance of `io.kestra.core.models.QueryFilter$Field`, problem: Unsupported field 'text'. Expected one of: [childFilter, workerId, endDate, level, triggerId, triggerExecutionId, labels, q, existingOnly, scope, namespace, state, flowId, startDate, timeRange] at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 684] (through reference chain: io.kestra.webserver.controllers.api.DashboardController$PreviewRequest["globalFilter"]->io.kestra.webserver.models.GlobalFilter["filters"]->java.util.ArrayList[0]->io.kestra.core.models.QueryFilter["field"])
```

MilosPaunovic:
This seems to be solved now, as on the link provided in the description there is no more error.

Might be related to preview Docker image not being up to date (see https://github.com/kestra-io/kestra-ee/issues/3925#issuecomment-2929385897 for more details).

PR #17164, opened 2026-07-01
feat(system): add 'migrate plan' to preview pending database migrations

AcevedoR:
### What

Adds a new read-only CLI command:

```
kestra migrate plan          # list the pending database migration scripts
kestra migrate plan --sql    # also print the SQL of each pending migration
```

It lists the migration scripts that `kestra migrate run` **would** apply, then exits **without applying them or modifying the database**. `--sql` additionally prints the SQL each migration would run.

### Why

Backport of the intent behind #17135 (`sys database plan`, on `releases/v1.3.x`) to `develop`. Customers preparing an upgrade want to preview which migrations are planned — and, especially, the SQL — before running them. On `develop` this is particularly useful for the 1.x → 2.0 jump.

### How (differs from the 1.3 version — Flyway is gone on develop)

`develop` replaced Flyway with the custom `MigrationScript` / `MigrationRunner` framework:

- `PlanMigrationCommand` mirrors `RunMigrationCommand`: its `propertiesOverrides()` sets `MigrationRunner.setSkipAutoRun(true)` so the eager `@Context` `MigrationRunner` does **not** auto-apply migrations during context startup.
- It then calls the framework's read-only `MigrationRunner.pendingScripts()` (no lock, applies nothing) and prints each pending script's `scriptId` + `description`.
- Registered as a subcommand of the existing `migrate` group, alongside `run` and `unlock`.

#### `--sql` — single source of truth for the resource path

`sqlResources()` on the `MigrationScript` interface is the **one place** a SQL migration declares its resource path (default empty for pure-Java migrations). The base classes derive everything else from it, so the path is never duplicated:

- `AbstractSQLMigrationScript` derives `checksum()` and `migrate()` from `sqlResources()`.
- `AbstractV2_0_01UpgradeMigration` derives `checksum()` from `sqlResources()`.

Concrete migrations therefore only declare `scriptId` / `description` / `sqlResources` (+ a `dataSource()` accessor) — `checksum()`/`migrate()` are inherited (net removal of code). `migrate plan --sql` reads and prints the SQL for each pending migration; Java-based migrations report they have none.

This is behavior-preserving: a single-element resource list yields the same checksum (same varargs call) and the same single execution as before.

- OSS-only change: EE inherits the `migrate` group and subclasses the runner, so `migrate plan` works in EE too with no EE code. Base methods are non-abstract defaults, so EE subclasses that keep their own `checksum()`/`migrate()` are unaffected.

### Testing

`PlanMigrationCommandTest` (3 tests, passing) via `PicocliRunner` against a fresh in-memory H2:
- auto-run skipped → lists all pending migrations (`0-init` + the `N pending migration(s):` header)
- `--sql` → prints the actual SQL (asserts a `CREATE TABLE`)
- auto-run enabled → `No pending migrations.`

No migration regressions: `MigrationRunnerTest` (15), `AbstractSQLMigrationScriptTest` (11), `JdbcMigrationHistoryStoreTest` (7) all pass.

Also QA'd end-to-end on Postgres: seeded a DB to 1.3 (Flyway-managed) then ran `migrate plan --sql` — it detected the legacy schema, **skipped the init/baseline scripts**, and listed exactly the 9 pending 2.0 migrations with their real Postgres DDL (and `(no SQL — Java-based migration)` for the Java ones), without applying anything.

### DIFF
--- a/cli/src/main/java/io/kestra/cli/commands/migrations/MigrationCommand.java
+++ b/cli/src/main/java/io/kestra/cli/commands/migrations/MigrationCommand.java
@@ -13,6 +13,7 @@ import picocli.CommandLine;
     mixinStandardHelpOptions = true,
     subcommands = {
         RunMigrationCommand.class,
+        PlanMigrationCommand.class,
         UnlockMigrationCommand.class
     }
 )
--- a/cli/src/main/java/io/kestra/cli/commands/migrations/PlanMigrationCommand.java
+++ b/cli/src/main/java/io/kestra/cli/commands/migrations/PlanMigrationCommand.java
@@ -1,0 +1,57 @@
+package io.kestra.cli.commands.migrations;
+
+import io.kestra.cli.AbstractCommand;
+import io.kestra.core.migration.MigrationRunner;
+import io.kestra.core.migration.MigrationScript;
+import jakarta.inject.Inject;
+import lombok.extern.slf4j.Slf4j;
+import picocli.CommandLine;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * CLI command that lists the pending database migration scripts without applying them.
+ *
+ * <p>This is the read-only counterpart of {@code kestra migrate run}: it reports the migrations
+ * that would be applied (e.g. what an upgrade would run) and then exits without touching the data.
+ * It is useful to prepare an upgrade by knowing exactly which migrations are planned.
+ *
+ * <p>Usage: {@code kestra migrate plan}
+ */
+@Slf4j
+@CommandLine.Command(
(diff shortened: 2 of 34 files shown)


nkwiatkowski (approved):

AcevedoR:
Companion EE PR (makes `migrate plan --sql` show EE migration SQL too): kestra-io/kestra-ee#9134 — must merge together.

AcevedoR:
Companion EE PR updated to kestra-io/kestra-ee#9152 (was #9134 — branch renamed to `feat/migrate-plan-command`). Must merge together.

PR #17171, opened 2026-07-01
fix(core): return a clear error when importing a file without an extension

anxkhn:
### Description

`HasSource.readSourceFile` handles uploaded flow files (used by the flow-import endpoint). When an upload is neither YAML nor a `.zip`, it rejects it with:

```java
throw new IllegalArgumentException("Cannot import file of type " + fileName.substring(fileName.lastIndexOf('.')));
```

If the uploaded filename has no extension, `fileName.lastIndexOf('.')` returns `-1`, so `substring(-1)` throws `StringIndexOutOfBoundsException`. That surfaces as a 500 on the import endpoint instead of the intended, descriptive `IllegalArgumentException`.

This guards the extension lookup: if there is no `.`, the whole filename is reported instead of calling `substring(-1)`. The descriptive error is now returned for extensionless filenames too.

### Related Issue

No existing issue. Self-identified while reading the import path. Happy to open a tracking issue if you'd prefer one.

### Backend Checklist

Added `HasSourceTest.shouldThrowIllegalArgumentWhenImportingFileWithoutExtension`. On `develop` it fails with `StringIndexOutOfBoundsException: Range [-1, 12)`; it passes with the fix. Verified with `./gradlew :core:test --tests "io.kestra.core.models.HasSourceTest"`.

### Additional Notes

Small guard plus a regression test; no change for files that already have an extension.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/HasSource.java
+++ b/core/src/main/java/io/kestra/core/models/HasSource.java
@@ -83,7 +83,9 @@ public interface HasSource {
                     }
                 }
             } else {
-                throw new IllegalArgumentException("Cannot import file of type " + fileName.substring(fileName.lastIndexOf('.')));
+                int extensionIndex = fileName.lastIndexOf('.');
+                String type = extensionIndex >= 0 ? fileName.substring(extensionIndex) : fileName;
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
LGTM, thanks

PR #17172, opened 2026-07-01
fix(core): report the full conflicting key path in flattenToNestedMap warning

anxkhn:
### Description

`MapUtils.flattenToNestedMap` reassembles dotted keys (for example form inputs) into a nested map and warns when a key conflicts with an existing value. For a conflict at an intermediate key it built the reported key with:

```java
var invalidKey = String.join(",", Arrays.copyOfRange(keys, 0, i));
```

The keys were split on `"."` (`entry.getKey().split("\\.")`), so this had two problems:

- It rejoined with `","` instead of `"."`, so the reported key did not match the user's original dotted key.
- `Arrays.copyOfRange(keys, 0, i)` is exclusive of `i`, so it dropped the segment that actually conflicts.

A conflict at `k1.k2` was therefore logged as `k1`. This changes it to `String.join(".", Arrays.copyOfRange(keys, 0, i + 1))` so the warning points at the real key. The other conflict branch in the same method already reports the correct key.

### Related Issue

No existing issue. Self-identified while reading the input-expansion path. Happy to open a tracking issue if you'd prefer one.

### Backend Checklist

Added `MapUtilsTest.shouldReportFullDottedKeyPathOnConflict`, which captures the log output and asserts the full path. It fails on `develop` (reports `k1`) and passes with the fix. Verified with `./gradlew :core:test --tests "io.kestra.core.utils.MapUtilsTest"`.

### Additional Notes

Diagnostic-only change: the map result is unchanged, only the warning message is corrected.

### DIFF
--- a/core/src/main/java/io/kestra/core/utils/MapUtils.java
+++ b/core/src/main/java/io/kestra/core/utils/MapUtils.java
@@ -202,7 +202,7 @@ public class MapUtils {
                 if (!currentMap.containsKey(key)) {
                     currentMap.put(key, new HashMap<>());
                 } else if (!(currentMap.get(key) instanceof Map)) {
-                    var invalidKey = String.join(",", Arrays.copyOfRange(keys, 0, i));
+                    var invalidKey = String.join(".", Arrays.copyOfRange(keys, 0, i + 1));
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
LGTM, thanks

PR #17183, opened 2026-07-01
Optimize JdbcQueueConfiguration by lowering poll intervals

### LINKED ISSUE #17168: perf(queue): lower default maxPollInterval and pollSwitchInterval in JdbcQueueConfiguration
## Problem

The JDBC queue's adaptive back-off escalates to long polling intervals too aggressively:

```
minPollInterval:    25 ms
maxPollInterval:   500 ms   ← worst-case latency for an idle queue
pollSwitchInterval: 60 s    ← time before escalation to max begins
```

A queue that has been idle for more than 60 seconds polls at 500 ms intervals. For execution processing this means a flow that hasn't run recently sees up to **500 ms of queuing latency** before its first message is picked up — the dominant contributor to perceived "cold start" latency on JDBC backends.

In practice, 500 ms is also the polling interval during off-hours when the scheduler fires an isolated trigger: the first execution always feels slow.

## Current defaults (`JdbcQueueConfiguration.java`)

| Property | Current | Rationale for change |
|---|---|---|
| `maxPollInterval` | 500 ms | 500 ms worst-case latency is too high for interactive workloads |
| `pollSwitchInterval` | 60 s | Queue stays at slow intervals for a full minute after going idle |

## Proposed defaults

| Property | Proposed | Notes |
|---|---|---|
| `maxPollInterval` | 100 ms | Limits worst-case latency to 100 ms, still avoids hammering the DB |
| `pollSwitchInterval` | 10 s | Recovers to the ceiling faster after a quiet period |

These remain conservative enough to avoid excessive DB load during genuine idle periods (one SELECT every 100 ms is negligible), while cutting worst-case latency by 5×.

## Impact

- Worst-case queuing latency: 500 ms → 100 ms
- Recovery after idle burst: starts within 10 s instead of 60 s
- Slight increase in SELECT rate during sustained idle periods (negligible for any production DB)
- Values remain fully configurable via `kestra.queue.min-poll-interval` / `max-poll-interval` / `poll-switch-interval`

## Related

Part of a broader JDBC queue latency analysis. See also: skip-sleep-on-partial-batch improvement and the PostgreSQL LISTEN/NOTIFY investigation.

raza-khan0108:
### ✨ Description

This PR lowers the default adaptive back-off polling intervals in `JdbcQueueConfiguration` and the CLI application configuration to reduce worst-case queuing latency on JDBC backends:

- **`maxPollInterval`**: Lowered from `PT0.5S` (500 ms) to `PT0.1S` (100 ms).
- **`pollSwitchInterval`**: Lowered from `PT60S` (60 s) to `PT10S` (10 s).

Previously, a flow that hadn't run recently saw up to 500 ms of queuing latency before its first message was picked up. By capping the maximum polling interval at 100 ms and reducing the time before back-off recovery from 60 s to 10 s, we cut perceived "cold start" latency by 5× while keeping database query load conservative during idle periods (1 SELECT every 100 ms).

Additionally, the step computation unit tests in `QueuePollerConfigurationTest` have been updated to reflect and verify these new default intervals.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/17168

### 🛠️ Backend Checklist

### 📝 Additional Notes

- Cuts worst-case idle queuing latency by 5× (from 500 ms down to 100 ms).
- Recovery after an idle period starts within 10 s instead of 60 s.
- Values remain fully configurable via `kestra.jdbc.queues.min-poll-interval`, `max-poll-interval`, and `poll-switch-interval`.

### DIFF
--- a/cli/src/main/resources/application.yml
+++ b/cli/src/main/resources/application.yml
@@ -204,8 +204,8 @@ kestra:
   jdbc:
     queues:
       min-poll-interval: 25ms
-      max-poll-interval: 500ms
-      poll-switch-interval: 60s
+      max-poll-interval: 100ms
+      poll-switch-interval: 10s
 
     cleaner:
       initial-delay: 1h
--- a/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueueConfiguration.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcQueueConfiguration.java
@@ -8,8 +8,8 @@ import io.micronaut.core.bind.annotation.Bindable;
 @ConfigurationProperties("kestra.jdbc.queues")
 public record JdbcQueueConfiguration(
     @Bindable(defaultValue = "PT0.025S") Duration minPollInterval,
-    @Bindable(defaultValue = "PT0.5S") Duration maxPollInterval,
-    @Bindable(defaultValue = "PT60S") Duration pollSwitchInterval,
+    @Bindable(defaultValue = "PT0.1S") Duration maxPollInterval,
+    @Bindable(defaultValue = "PT10S") Duration pollSwitchInterval,
     @Bindable(defaultValue = "100") Integer pollSize,
     @Bindable(defaultValue = "5") Integer switchSteps,
     @Bindable(defaultValue = "true") Boolean immediateRepoll) {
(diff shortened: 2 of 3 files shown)


loicmathieu:
@raza-khan0108 thanks for your contribution but #17168 was assigned to me for a reason: we need to benchmark and analyse the impact on DB resources before changing those default.

PR #17185, opened 2026-07-02
fix(executions): keep the redirect to overview

Piyush-r-bhaskar:
closes https://github.com/kestra-io/kestra-ee/issues/9113


MilosPaunovic (commented):
Hey @Piyush-r-bhaskar,

I think that the current behavior is the one we actually want, that `Play` button for execution was introduced so the users would be able to quickly execute many flows without being redirected every time.

Piyush-r-bhaskar:
Okay okay...then this is not relevant. Thanks Milos.

PR #17196, opened 2026-07-02
fix(webserver): use constant-time comparison for secret webhook key and BasicAuth token cache

loicmathieu:
## Summary
- The webhook trigger endpoint compared the caller-supplied `key` against the flow's configured secret webhook key using `String.equals()`, which short-circuits on the first differing character. Since this endpoint is public and unauthenticated by design, the timing side-channel is directly exploitable by a remote attacker to recover a valid webhook key.
- The Basic-Auth fast-path cache check compared the SHA-256 digest of the incoming token against the last successfully verified token's digest using `String.equals()`, exposing the same class of timing side-channel on an authentication-bypass path.
- Both are replaced with `MessageDigest.isEqual()`, which the JDK specifies as constant-time. Same fix pattern already applied to `SecureApiToken.matches()` in the EE repo.

Fixes https://github.com/kestra-io/kestra-ee/issues/9142
Fixes https://github.com/kestra-io/kestra-ee/issues/9141

## Test plan
- [x] `./gradlew :webserver:test --tests "io.kestra.webserver.services.BasicAuthServiceTest"` — 28/28 passing
- [x] `./gradlew :webserver:test --tests "io.kestra.webserver.controllers.api.ExecutionControllerRunnerTest.webhook"` — passing

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
@@ -15,6 +15,7 @@ import java.nio.charset.IllegalCharsetNameException;
 import java.nio.charset.StandardCharsets;
 import java.nio.charset.UnsupportedCharsetException;
 import java.nio.file.Path;
+import java.security.MessageDigest;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.ZonedDateTime;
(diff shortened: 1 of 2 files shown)


nkwiatkowski (approved):

loicmathieu:
Backported to 1.0 and 1.3

PR #17198, opened 2026-07-02
feat(system): increase the default database pool size from 10 to 25

loicmathieu:
We have 20 queues in OSS (more in EE) and queues are polling frequently the queues table and process messages inside a transaction so the connection is retain for the full select/process/delete sequence.

On a busy instance, the default 10 connections can all be used by queues without talking about other queries form the UI. Using a default of 25 (if not configured otherwise by the user) offers better performance and is a sane default as production databases can usually suppport way more open connections.

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/DatasourcePoolSizeDefaulter.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/DatasourcePoolSizeDefaulter.java
@@ -1,0 +1,41 @@
+package io.kestra.jdbc;
+
+import java.util.Objects;
+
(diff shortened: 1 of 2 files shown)


loicmathieu:
It increase supported throughput in our reference scenario from 3500 exec/mn to 4000 exec/mn

loicmathieu:
## benchmark results

The pool-sizing change (max=25 vs max=10) is a clear, substantial win: it raises the throughput ceiling from ~2500–3000 req/s to ~3500 req/s before latency degrades, and critically prevents the catastrophic, non-recovering backlog collapse develop suffers at 4000+ req/s (99% completion vs ~1% completion at step 8).

AcevedoR (approved):

PR #17206, opened 2026-07-02
fix(chart): allow explicit replicas 0 on a deployment

Skraye:
Helm's `default` function treats `0` as an empty value, so setting `replicas: 0` on a specific deployment (e.g. `deployments.executor.replicas: 0`) was silently falling back to `common.replicas` and rendering `1`. This prevents declaring a dormant deployment scaled to zero, which we need for KEDA-driven autoscaling of cloud instances (a deployment must exist at 0 replicas so KEDA can scale it 0→1).

Replaced `default` with an explicit `hasKey` check so a deployment-level `replicas: 0` is respected, while the fallback to `common.replicas` (including offboarding's `common.replicas: 0`) is unchanged. Added a helm-unittest suite covering the four replica resolution cases.

### DIFF
--- a/charts/kestra/templates/deployment.yaml
+++ b/charts/kestra/templates/deployment.yaml
@@ -47,7 +47,8 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
-  replicas: {{ hasKey $content "replicas" | ternary $content.replicas $common.replicas }}
+  {{- /* kindIs "invalid" is true only when replicas is unset or explicit null, so an explicit `replicas: 0` is preserved (unlike Sprig's `default`, which treats 0 as empty) */}}
+  replicas: {{ kindIs "invalid" $content.replicas | ternary $common.replicas $content.replicas }}
   revisionHistoryLimit: {{ default $common.revisionHistoryLimit $content.revisionHistoryLimit }}
   selector:
     matchLabels:
--- a/charts/kestra/templates/deployment.yaml
+++ b/charts/kestra/templates/deployment.yaml
@@ -47,7 +47,6 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
-  {{- /* kindIs "invalid" is true only when replicas is unset or explicit null, so an explicit `replicas: 0` is preserved (unlike Sprig's `default`, which treats 0 as empty) */}}
   replicas: {{ kindIs "invalid" $content.replicas | ternary $common.replicas $content.replicas }}
   revisionHistoryLimit: {{ default $common.revisionHistoryLimit $content.revisionHistoryLimit }}
   selector:
--- a/charts/kestra/templates/deployment.yaml
+++ b/charts/kestra/templates/deployment.yaml
@@ -47,7 +47,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
-  replicas: {{ default $common.replicas $content.replicas }}
+  replicas: {{ hasKey $content "replicas" | ternary $content.replicas $common.replicas }}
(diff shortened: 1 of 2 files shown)


charts/kestra/templates/deployment.yaml
```diff
@@ -47,7 +47,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
-  replicas: {{ default $common.replicas $content.replicas }}
+  replicas: {{ hasKey $content "replicas" | ternary $content.replicas $common.replicas }}
```
MarcoSaba: `hasKey` only checks the key exists — an explicit `replicas: null` counts as "present" and renders blank instead of falling back to `common.replicas`. Tested on a real cluster: `kubectl apply` accepts it silently and Kubernetes defaults it to `1`, ignoring `common.replicas` entirely.

Suggested fix:
```diff
- replicas: {{ hasKey $content "replicas" | ternary $content.replicas $common.replicas }}
+ replicas: {{ kindIs "invalid" $content.replicas | ternary $common.replicas $content.replicas }}
```

`kindIs "invalid"` catches actual `nil` (absent or null), while `0` still passes through. Verified locally against both `--set` and `-f values.yaml` (unset, `0`, `null`, `2`) — all pass.

MarcoSaba:
Here is a reproducer on how to test the fix added by myself:

# Testing `replicas: 0` on the Kestra Helm Chart

This validates the fix for `deployments.<name>.replicas: 0` (and `null`) on the Kestra Helm chart — used by KEDA-driven autoscaling, which requires a real Deployment object at 0 replicas so it can scale it 0→1 on demand.

Assumes a Kubernetes cluster is already available and `kubectl`/`helm` are configured against it.

## 1. Create a namespace

```bash
kubectl create namespace kestra-test
```

## 2. Create a test values file

Save as `~/test-replicas-zero.yaml`:

```yaml
image:
  tag: "v1.3.26"

dind:
  enabled: false

deployments:
  standalone:
    enabled: false
  webserver:
    enabled: true
  executor:
    enabled: true
  indexer:
    enabled: true
  scheduler:
    enabled: true
  worker:
    enabled: true
    replicas: 0

configurations:
  application:
    kestra:
      repository:
        type: memory
      storage:
        type: local
        local:
          base-path: /tmp/kestra-storage
      queue:
        type: memory
      server:
        basic-auth:
          enabled: false
```

## 3. Install the chart

```bash
helm install kestra-test . \
  --namespace kestra-test \
  -f ~/test-replicas-zero.yaml
```

## 4. Verify the worker deployment is at 0 replicas, not 1

```bash
kubectl get deployment -n kestra-test
```

Expected: `kestra-test-worker` shows `0/0` ready — the Deployment exists, it's just scaled to zero. Every other component (webserver, executor, indexer, scheduler) should show `0/1` (still starting) or `1/1` once ready.

Confirm the exact value:

```bash
kubectl get deployment kestra-test-worker -n kestra-test -o jsonpath='{.spec.replicas}'
```

Expected output: `0`

## 5. Simulate KEDA scaling it back up

KEDA scales by patching `spec.replicas` directly on the live object — this proves that mechanism works on top of the chart's rendering:

```bash
kubectl scale deployment kestra-test-worker -n kestra-test --replicas=1
```

## 6. Confirm the fix applies to every deployment type, not just worker

The `replicas` field is resolved by one shared template line that every entry under `deployments.*` (and `workerGroups.*` for EE) goes through — so the same `replicas: 0` behavior should hold for any component. Verify with a second one:

```bash
helm upgrade kestra-test ~/repositories/kestra/charts/kestra \
  --namespace kestra-test \
  -f ~/test-replicas-zero.yaml \
  --set deployments.executor.replicas=0

kubectl get deployment kestra-test-executor -n kestra-test -o jsonpath='{.spec.replicas}'
```

Expected output: `0`

gperreymond (approved):

PR #17222, opened 2026-07-03
chore(core): store translation files in `en.json` key order

MilosPaunovic:
Generates and stores UI translation files in en.json key order (instead of alphabetically) so regeneration no longer opens reorder-only PRs.


elevatebart (approved):
simple enough

PR #17223, opened 2026-07-03
fix: make basic auth cleaner (no clear cookie on client)

elevatebart:
closes https://github.com/kestra-io/kestra-ee/issues/9053


ui/src/override/components/auth/Auth.vue
```diff
@@ -74,7 +74,7 @@
     }))
 
     const logout = () => {
-        BasicAuth.logout()
+        void BasicAuth.logout()
```
loicmathieu: You don't need to await?

ui/src/utils/basicAuth.ts
```diff
@@ -1,19 +1,29 @@
...
+    }
     return true
 }
 
-export function signIn(username: string, password: string) {
-    const trimmedUsername = username.trim()
-    const encoded = btoa(`${trimmedUsername}:${password}`)
-    document.cookie = `BASIC_AUTH=${encoded};path=/;samesite=strict`
+export function signIn() {
+    sessionStorage.setItem(AUTH_FLAG_KEY, "true")
```
loicmathieu, line 27: Is it safe? Anybody can set this in the browser right?

ui/src/main.ts
```diff
@@ -35,7 +35,7 @@ const app = createApp(App)
 
 const handleAuthError = (error: Error, to: {fullPath: string}) => {
     if (error.message?.includes("401")) {
-        BasicAuth.logout()
+        void BasicAuth.logout()
```
loicmathieu: Same here and under, I'm not sure I undersand void here and in another place you await

webserver/src/main/java/io/kestra/webserver/filter/AuthenticationFilter.java
```diff
@@ -48,6 +48,8 @@ public Publisher<MutableHttpResponse<?>> doFilter(HttpRequest<?> request, Server
             {
                 String normalizedPath = normalizePath(request.getPath());
                 boolean isConfigEndpoint = "/api/v1/configs".equals(normalizedPath)
+                    || "/api/v1/login".equals(normalizedPath)
+                    || "/api/v1/logout".equals(normalizedPath)
```
loicmathieu: Logout should not bypass auth so I don't think it shoul dbe here

loicmathieu (changes requested):

ui/src/override/components/auth/Auth.vue
elevatebart: Intentional — this `logout()` is a synchronous click handler that needs to call `router.push` right away rather than block on the network round-trip to `/api/v1/logout`. `BasicAuth.logout()` clears the `sessionStorage` login flag synchronously before it even starts the `fetch`, so `isLoggedIn()` already reflects "logged out" the instant this runs, and the fetch itself is wrapped in a try/catch as best-effort cookie cleanup. We only `await` it in `BasicAuthLogin.vue`'s `validateCredentials`, where the call happens inside an already-async flow and nothing needs to navigate immediately.

---

ui/src/main.ts
elevatebart: Same rationale as the `Auth.vue` thread: both `handleAuthError` and `beforeLogout` need to return synchronously (a route redirect object, or void from an interceptor callback) without waiting on the `/api/v1/logout` network call. `BasicAuth.logout()` clears the `sessionStorage` auth flag synchronously up front, so `isLoggedIn()` is already correct regardless of whether the fetch has resolved — the network call is just best-effort cookie cleanup (wrapped in try/catch). The one place we `await` it (`BasicAuthLogin.vue`) is already inside an async function where nothing depends on immediate navigation.

---

ui/src/utils/basicAuth.ts
elevatebart: Yes, anyone can set it — but it's not a security boundary, just a client-side UI hint for routing decisions (`App.vue`'s login/setup redirect, the axios interceptor's `isLoggedIn` check in `main.ts`, `misc.ts`). Actual authorization is enforced entirely server-side by `AuthenticationFilter`/`BasicAuthService#isAuthenticated`, which validates the HttpOnly `BASIC_AUTH` cookie issued by the server on `/login` — that cookie can't be read or forged from JS. If someone manually sets `kestraBasicAuthenticated=true` in sessionStorage without a valid cookie, the UI optimistically skips the login screen, but every subsequent API call still 401s, and the `!isLoggedIn()` check in the response interceptor is exactly what triggers the redirect back to `/login` when that happens. So worst case is a brief flash before being bounced back — no real bypass of auth.

---

webserver/src/main/java/io/kestra/webserver/filter/AuthenticationFilter.java
elevatebart: This is intentional rather than an oversight: `/logout` needs to bypass `AuthenticationFilter` so it can clear the `BASIC_AUTH` cookie even when that cookie is no longer valid (e.g. credentials changed since it was issued, or the value is stale/corrupted). If it weren't bypassed, `isAuthenticated()` would already be failing in exactly the case logout most needs to handle — the request would get a bare 401 without ever reaching the cookie-clearing logic, leaving an invalid cookie stuck in the browser forever.

It's still safe against CSRF: `CsrfTokenFilter` runs independently of this filter and enforces `X-CSRF-TOKEN` validation on any state-changing request that carries the auth cookie (`hasCookieAuth`), regardless of whether `AuthenticationFilter` requires auth for that path — so a forged cross-site POST to `/logout` is already rejected there. And logging a session out has no meaningful blast radius beyond forcing a re-login.

Happy to add a short code comment on the bypass list calling this out explicitly if that'd help future readers.

---

elevatebart: Fair point — pushed in f88a6fd9b: `/api/v1/logout` no longer bypasses `AuthenticationFilter`. Added `logout_shouldRequireAuthentication` (unauthenticated → 401) and updated `logout_shouldClearCookie` → `logout_shouldClearCookie_whenAuthenticated` to authenticate via Basic auth first.

---

ui/src/main.ts
elevatebart: When logging out, we do not need to know if the logout was succesful to pursue removing cookies and session tokens.

ui/src/override/components/auth/Auth.vue
loicmathieu: Understood

loicmathieu (approved):
LGTM

ui/src/override/components/auth/Auth.vue
elevatebart: removed void since it brings nothing but confusion

PR #17224, opened 2026-07-03
fix(tasks): don't validate Content-Length for content-encoded HTTP downloads

AcevedoR:
## What

The `Download` HTTP task fails with `IllegalStateException: Invalid size` whenever the server returns a **content-encoded** (e.g. `Content-Encoding: gzip`) response.

## Root cause

`Download.run()` compares the number of bytes written to storage against the response's `Content-Length` header. For a gzip response, `Content-Length` is the **compressed** transfer size, while the Apache HTTP client transparently **decompresses** the body — so the bytes written are the **decoded** size. The two never match, so the task throws.

Per [RFC 9110 §8.6](https://www.rfc-editor.org/rfc/rfc9110#section-8.6), `Content-Length` describes the encoded message body, so the check is only meaningful for identity encoding.

## How this was found

It surfaced as a **consistent** (non-flaky) failure of `core > io.kestra.core.tasks.test.SanityCheckTest > qaNamespaceFiles`. Its first task downloads a file from `raw.githubusercontent.com`, which is now served gzip-encoded:

```
Invalid size, got 383, expected 245   # decoded 383 bytes vs compressed Content-Length 245
```

→ `download` task FAILED → execution never reached SUCCESS → `hasSize(8)` assertion failed. Reproduced locally on the first run (deterministic, not a timeout).

## Fix

Skip the `Content-Length` size check when a non-identity `Content-Encoding` header is present.

## Tests

Adds `DownloadTest.gzip`, an embedded-server regression test that serves a gzip-encoded body (compressed Content-Length ≠ decoded size). Verified it **fails without** this change (`Invalid size, got 2816, expected 83`) and **passes with** it.

- `SanityCheckTest.qaNamespaceFiles` → green
- `DownloadTest` → 13/13 passing

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/http/Download.java
+++ b/core/src/main/java/io/kestra/plugin/core/http/Download.java
@@ -98,7 +98,14 @@ public class Download extends AbstractHttp implements RunnableTask<Download.Outp
                         size.set(0L);
                     }
 
-                    if (r.getBody() != null) {
+                    // Content-Length describes the size of the transferred (encoded) body, while `size` counts the
+                    // bytes actually written after any content-coding has been transparently decoded (e.g. gzip).
+                    // The two only match for identity encoding, so skip the check when the response is content-encoded.
+                    boolean contentEncoded = r.getHeaders().firstValue("Content-Encoding")
+                        .map(encoding -> !encoding.isBlank() && !"identity".equalsIgnoreCase(encoding))
+                        .orElse(false);
+
+                    if (r.getBody() != null && !contentEncoded) {
                         r.getHeaders().firstValue("Content-Length").ifPresent(header ->
                         {
                             long length = Long.parseLong(header);
--- a/core/src/test/java/io/kestra/plugin/core/http/DownloadTest.java
+++ b/core/src/test/java/io/kestra/plugin/core/http/DownloadTest.java
@@ -1,8 +1,10 @@
 package io.kestra.plugin.core.http;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
+import java.util.zip.GZIPOutputStream;
 
 import org.apache.commons.io.IOUtils;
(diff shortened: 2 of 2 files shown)


AcevedoR:
## Root-cause timeline — what introduced this and why it surfaced now

### Two layers

**Latent bug — `6dfe29adf6` "feat(core): use http abstraction on http tasks" (2024-12-31).**
That migration rewrote `Download` to validate bytes-written against the raw `Content-Length` **header** (`r.getHeaders().firstValue("Content-Length")`) instead of the old `entity.getContentLength()`. This silently relied on the HTTP client stripping `Content-Length`/`Content-Encoding` from the response *after* transparently decompressing gzip.

**Actual trigger — `093e999182` "build(deps): bump httpclient5 5.5.1 → 5.6.2" (2026-07-01, dependabot).**

### Why it didn't fail for ~18 months — confirmed by bisection

| Code | httpclient5 | Result |
|---|---|---|
| unpatched | **5.5.1** | ✅ pass |
| unpatched | **5.6.2** | ❌ fail (`Invalid size, got 383, expected 245`) |
| this PR | either | ✅ pass |

httpclient5 **≤ 5.5.1** removed `Content-Length`/`Content-Encoding` from the message after transparently gunzipping, so `firstValue("Content-Length")` was empty and the size check was skipped entirely. **5.6.2 no longer strips them**, so the *compressed* `Content-Length: 245` now survives onto the *decompressed* response (383 bytes) and the check finally runs → throws.

So there was no regression in Kestra's own logic on 2026-07-01 — a minor HTTP-client bump removed the undocumented header-stripping behavior the size check had been quietly depending on since Dec 2024. Any user downloading a gzip-served endpoint hits this the moment they run on httpclient5 5.6.x; `SanityCheckTest.qaNamespaceFiles` (which downloads a live gzip URL from `raw.githubusercontent.com`) was just the canary.

This PR makes the check correct regardless of the client's header-stripping behavior by skipping it when a non-identity `Content-Encoding` is present.

brian-mulier-p (approved):
LGTM

PR #17227, opened 2026-07-03
refactor: migrate UI stores to the kestra-sdk client

elevatebart:
### ✨ Description

Migrates every Pinia store under `ui/src/stores/` (plus `ui/src/composables/useBaseNamespaces.ts`) from hand-built `axios` calls against string-concatenated URLs to the typed `@kestra-io/kestra-sdk` client, following the pattern already used by a handful of call sites (`useBaseNamespaces.ts`'s `NamespaceAPI`, `FilePreview.vue`, `ExpressionDebugger.vue`, `AiCopilot.vue`).

**Stores migrated:** `mcp.ts`, `trigger.ts`, `kvs.ts`, `secrets.ts`, `core.ts`, `service.ts`, `plugins.ts`, `dashboard.ts`, `logs.ts`, `blueprints.ts` (local/custom half), `flow.ts`, `executions.ts`, and the SDK-coverable parts of `useBaseNamespaces.ts`.

**Left on raw axios / `EventSource`, on purpose:**
- SSE streaming (`followExecution`, `followExecutionDependencies`, `followLogs`) — no SDK streaming equivalent.
- Blob/CSV/text exports (`exportFlowByIds`, `exportFlowByQuery`, `exportFlowAsCSV`, `exportExecutionsAsCSV`, `exportDashboard`, `exportTriggersAsCSV`, `exportFileDirectory`).
- Multipart/form-data calls (`importFlows`, `resume`, `validateResume`, `replayExecutionWithInputs`, `triggerExecution`, `validateExecution`, namespace file create/import).
- Calls to the external `api.kestra.io` host (`api.ts`, `pluginsEnrichment.ts`, `doc.ts`, the "community" half of `blueprints.ts`).
- `plugins.ts`'s `loadIcon()` — the SDK only exposes the full-catalog `pluginIcons()`, no per-class icon lookup endpoint.

Added `routeQueryToQueryFilters()` in the design system as a prerequisite: generated SDK search functions only forward parameters declared in their OpenAPI signature, so this converts the raw pre-encoded `filters[...]` `route.query` keys that list pages build today into the typed `QueryFilter[]` array the SDK's search functions expect, porting `QueryFilterFormatBinder`'s regex and AND/OR tree-building from the backend so nested groups and `labels` sub-keys round-trip correctly.

**Reusing SDK model types instead of duplicating them.** `flow.ts`'s `Flow`/`Trigger` and `executions.ts`'s `Execution`/`Label`/`Histories` are now built on the SDK's generated types (`FlowWithSource`, `AbstractTrigger`, `ApiExecution`, `Label`, `StateHistory`) instead of hand-rolled interfaces, overridden only where the generated OpenAPI schema is genuinely missing something the backend actually returns (recursive subtasks, Schedule's `backfill`, Webhook's `key`, execution `inputs`/`variables`/`outputs`). This:
- Surfaced and fixed a real bug: `Flow.labels` was typed as a map but is actually `Label[]` at runtime, so `readOnlySystemLabel` silently never matched.
- Removed several now-unnecessary `as unknown as Execution` casts and duplicate local `Flow`/`Execution`/`Trigger` interfaces in `WebhookCurl.vue`, `SubFlowLink.vue`, `ForceRun.vue`, `Unqueue.vue`, `ExecutionPending.vue`, `ExecutionRootTopBar.vue`, and `executionUtils.ts`.
- Bumped `@kestra-io/kestra-sdk` mid-migration to a version that closed two gaps this PR had previously worked around with raw axios: `flow.ts`'s `saveFlow`/`createFlow` now use `FlowsAPI.updateFlow`/`createFlow` (the SDK finally exposes the `draft` query param), and `dashboard.ts`'s `loadDefaults()` now uses the new `dashboards-admin` submodule's `defaultDashboards()` (the correct tenant-scoped `DashboardController` route) instead of raw axios.

Also fixed test coverage that this migration exposed as stale: several `tests/storybook/**/*.stories.{jsx,tsx}` files mocked the backend via `setMockClient(axios)`, which only swaps the axios instance returned by `useClient()` — it does not reconfigure the separate internal `client` singleton that every generated SDK submodule function actually calls through. Converted the affected stories (`Triggers`, `Executions`, `KSFilter`, `FileExplorer`, `LowCodeEditor`) to `vi.mock("@kestra-io/kestra-sdk/...", ...)`, and `Table.stories.tsx` to a direct Pinia store method stub (`useDashboardStore().generate = ...`) after the module-mock approach proved unreliable for that specific submodule. The two unit specs asserting on `saveFlow`'s draft param (`flowDraftSave.spec.ts`, `flowOutdatedSave.spec.ts`) were updated the same way once `saveFlow` moved off raw axios.

### 🔗 Related Issue

None.

### 🎨 Frontend Checklist

- [ ] Screenshots or video recordings attached showing the `UI` changes — no user-visible UI changes; this is a transport-layer refactor plus test fixes.

### 📝 Additional Notes

- No behavior change is intended anywhere except `dashboard.ts`'s `loadDefaults()`, where the migration briefly introduced a real regression (every page load 404'd inside `loadDefaults()` before falling back to the bundled default dashboard, breaking the home page's "Default Dashboard" heading) — caught via CI's E2E suite and fixed in this PR (and now fixed properly via the SDK rather than a raw-axios workaround).
- Storybook/Playwright browser tests could not be run locally in this environment (Playwright browser revision mismatch), so all test fixes in this PR were verified via pushing and re-checking GitHub Actions CI runs rather than locally.
- Full working session: https://claude.ai/code/session_0113pS2MhU3711wyXwUFdy2S

### 🤖 AI Authors

Why was the cat sitting on the computer? It wanted to keep an eye on the `mouse` — much like this PR keeps an eye on every `axios.get` that used to sneak past code review. 🐱


Piyush-r-bhaskar:
Very interesting ❤️

ui/src/stores/executions.ts
```diff
@@ -73,57 +76,16 @@ interface LogsState {
...
-    },
-    metadata: {
-        originalCreatedDate: string;
-        attemptNumber: number;
-    },
+// inputs/variables/outputs and taskRunList[].outputs aren't modeled on the SDK's ApiExecution/
+// ApiTaskRun (a backend OpenAPI-spec gap), even though the backend's Execution/TaskRun models
+// have them - overridden here so this app's execution objects keep exposing them.
+export type Execution = ApiExecution & {
+    taskRunList?: (ApiTaskRun & {outputs?: Record<string, any>})[]
```
elevatebart: `execution.outputs` do not exist in version 2.0. This quirk is an artefact of 1.3 implementation that was never removed. 
The use of this field and its definition should be removed.

ui/src/stores/flow.ts
```diff
@@ -25,14 +27,13 @@ const textYamlHeader = {
...
+// backfill (Schedule) and key (Webhook) are trigger-type-specific runtime config, not part of
+// the SDK's generic AbstractTrigger schema (which models fields common to every trigger type).
+export type Trigger = AbstractTrigger & {
     backfill?: {
         start?: string;
     };
+    key?: string;
 }
 
 export interface Task {
```
elevatebart: Why is the Task interface not using the `Task` from `kestra-sdk` ?

ui/src/stores/executions.ts
elevatebart: Removed in a7b7539 — dropped `outputs` from the `Execution` type, and removed its dead usages (the "Outputs" column in `Executions.vue` and the "Flow Outputs" section in `ExecutionVariableExplorer.vue`, plus their now-orphaned translations). `taskRunList[].outputs` (task-level outputs, actively used by `Outputs.vue`) is unaffected.

---

ui/src/stores/flow.ts
elevatebart: Fair point — fixed in a7b7539: `Task` now extends the SDK's `Task` type, with just the recursive `tasks` field added back on top (same pattern as `Trigger`/`Flow`), since the SDK's flat generic `Task` schema doesn't model subtask nesting.

---

elevatebart:
CI's E2E run flagged a real bug from `8198ce5`'s `triggerExecution()` migration to `ExecutionsAPI.createExecution()` — reverted just that function to raw axios in `f54bc11`.

The SDK's `createExecution()` relies on `formDataBodySerializer`, which builds the multipart body via `Object.entries(body)`. That only works when `body` is a plain object. `triggerExecution`'s callers pass either a plain object (`playground.ts`'s `defaultInputValues`) *or* an already-built `FormData` instance (`submitTask.ts`'s `inputsToFormData()`) — and `Object.entries()` on an existing `FormData` returns nothing, so every input silently vanished from the request. `flow.spec.ts`'s "create and execute a Flow with input" E2E test caught it (the execution never got its input, so the graph never rendered the executed task). `Utils.toFormData()` already normalizes both shapes correctly, so reverting to it is the simplest fix — happy to help find a cleaner SDK-based way to express "body is either a FormData or a plain object" if you'd rather keep it on the SDK path.

---

ui/src/stores/dashboard.ts
```diff
@@ -79,22 +75,20 @@ export const useDashboardStore = defineStore("dashboard", () => {
         return dashboardList.value
     }
 
+    // Tenant-scoped settings, hardcoded to the "main" tenant (this OSS build is single-tenant) —
+    // bypasses normal tenant routing on purpose, same as before this endpoint moved to the SDK.
+    const DEFAULT_DASHBOARDS_TENANT_ID = "main"
```
elevatebart: Why did we have to rederfine this here? Isn't there a default tenant in the sdk?

elevatebart: And since it is ued only once, let's hardcode it.

ui/src/stores/dashboard.ts
```diff
@@ -175,60 +169,57 @@ export const useDashboardStore = defineStore("dashboard", () => {
...
             return undefined
         }
 
-        if (res.status === 404){
-            return undefined
-        }
-
-        activeDashboard.value = res.data
-        sourceCode.value = res.data.sourceCode ?? ""
+        activeDashboard.value = data as Dashboard
```
elevatebart: Why is this necessary? 
`activeDashboard.value` should be of the return type of `DashboardsAPI.dashboard(` whatever it is.

ui/src/stores/dashboard.ts
```diff
@@ -175,60 +169,57 @@ export const useDashboardStore = defineStore("dashboard", () => {
...
-        const res = await axios.post(`${apiUrl()}/dashboards/validate`, source, header)
-        return res.data
+        return DashboardsAPI.validateDashboard({body: source ?? ""})
     }
 
     async function generate(id: Dashboard["id"], chartId: Chart["id"], parameters: Parameters) {
-        const res = await axios.post(`${apiUrl()}/dashboards/${id}/charts/${chartId}`, parameters, {validateStatus})
-        return res.data
+        try {
+            return await DashboardsAPI.dashboardChartData({id, chartId, ...parameters} as any)
```
elevatebart: Why is there `as any` here? This is usually a bad practice and removes features from strong typing. Fix it

ui/src/stores/dashboard.ts
```diff
@@ -175,60 +169,57 @@ export const useDashboardStore = defineStore("dashboard", () => {
...
+            throw e
+        }
     }
 
     async function validateChart(source: string) {
-        const res = await axios.post(`${apiUrl()}/dashboards/validate/chart`, source, header)
-        chartErrors.value = res.data
-        return res.data
+        const data = await DashboardsAPI.validateChart({body: source})
+        chartErrors.value = data as unknown as string[]
```
elevatebart: This is an odd cast. What is `data` not an array of string in the first place?

ui/src/stores/dashboard.ts
```diff
@@ -175,60 +169,57 @@ export const useDashboardStore = defineStore("dashboard", () => {
...
+        return data
     }
 
     async function chartPreview(request: Request) {
-        const res = await axios.post(`${apiUrl()}/dashboards/charts/preview`, request)
-        return res.data
+        return DashboardsAPI.previewChart(request as any)
     }
```
elevatebart: The `Request` type should be equivalent to the first param of `previewChart` making the cast unnecessary.

ui/src/stores/executions.ts
```diff
@@ -73,57 +76,15 @@ interface LogsState {
...
-    state: string;
-    date: string;
-}
+export type {Label, StateHistory as Histories} from "@kestra-io/kestra-sdk"
 
-export interface Execution{
-    id: string;
-    namespace: string;
-    flowId: string;
+export type Execution = Omit<SDKExecution, "deleted" | "tenantId" | "taskRunList"> & {
```
elevatebart: Is there a way to avoid repeating the list of fields in this type modification? DRY...

ui/src/stores/dashboard.ts
elevatebart: Done in 4e24af5 — inlined it. (There is a default tenant in the SDK, but it auto-fills params literally named `tenant`; this endpoint's path param is `id`, which isn't covered by that mechanism, so it still needs to be passed explicitly — `"main"` is the only value that actually routes correctly here since `TenantController` hardcodes that segment.)

---

ui/src/stores/dashboard.ts
elevatebart: Fixed in 4e24af5. `activeDashboard` is a local `Dashboard` type (in `components/dashboard/types.ts`), not the raw SDK response type — the app's `Dashboard` differs from `DashboardControllerDashboardResponse` in one field: `charts` is polymorphic on the backend (`DataChart`, `DataChartKPI`, ...) but the SDK's generated response type only models the lean common subset (id/type/chartOptions). So `activeDashboard.value = {...data, charts: data.charts as unknown as Chart[]}` — everything else comes straight from `data` untouched, only `charts` is bridged. `title`/`deleted` are also widened back to optional on the local type since a dashboard can be a locally-built preview placeholder that never persisted either field.

---

ui/src/stores/dashboard.ts
elevatebart: Fixed in 4e24af5 — replaced with `as globalThis.Parameters<typeof DashboardsAPI.dashboardChartData>[0]`. The local `Parameters` type (imported from `dashboard/types.ts`) was shadowing the global TS utility type of the same name, so I couldn't reference it directly; `globalThis.Parameters<T>` disambiguates and gets us back real type-checking against the SDK function's actual parameter shape instead of `any`.

---

ui/src/stores/dashboard.ts
elevatebart: Fixed in 4e24af5. `data` (from `DashboardsAPI.validateChart`) is the full chart validation response, not a string array — the string array we want is `data.constraints`. Changed to `chartErrors.value = data.constraints ? [data.constraints] : []`, no cast needed.

---

ui/src/stores/dashboard.ts
elevatebart: Fixed in 4e24af5. `Request` is now `export type Request = DashboardControllerPreviewRequest` (the SDK's actual param type for `previewChart`), so `chartPreview()` just does `return DashboardsAPI.previewChart(request)` with no cast.

---

ui/src/stores/executions.ts
elevatebart: Fixed in 4e24af5 with a small local generic helper:
```ts
type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
```
so the widened fields are listed once (`PartialBy<SDKExecution, "deleted">`, `PartialBy<TaskRun, "namespace" | "executionId" | "flowId">`) instead of being spelled out via repeated `Omit<...> & {field?: ...}` blocks.

---

Piyush-r-bhaskar (approved):
Thank you for this. It's really a nice refactoring 🎉

flcarre (changes requested):
.

PR #17228, opened 2026-07-03
feat(core): OSS enablers for EE Policies

fhussonnois:
- add the policyRefs flow field and the POLICY JSON-schema type
- remove flow plugin defaults, rename PluginDefaultService to FlowParsingService
- expose map-level parse seam (readFlowAsMap/parseFlowFromMap)
- add the Patterns compiled-regex cache utility


MilosPaunovic:
## Frontend review

Clean, well-scoped change overall — rips out the old Plugin Defaults no-code plumbing and repoints it at Policies, adds a nice `rowSelectable` to `KsDataTable` (with a test), and tidies up design-system tokens. A few things worth a look before merge.

### 🔴 `TaskNamespaces.vue` — `allowCreate` and `data-type` are silently dropped
```vue
<NamespaceSelect data-type="flow" v-model="modelValue" multiple allowCreate />
```
`NamespaceSelect` declares only `multiple, readOnly, clearable, taggable, placeholder, fit, autoDefault` and sets `inheritAttrs: false` with no `v-bind="$attrs"`, so neither `allowCreate` nor `data-type="flow"` reaches anything — both are no-ops. The prop that enables typing new values is `taggable` (mapped internally to `:allowCreate="taggable"`). As written, users **can't create new namespace entries** here. Suggest: use `taggable` instead of `allowCreate`, and drop `data-type="flow"`.

### 🟡 `getTaskComponent.ts` — name-based routing is broad
```ts
if (key === "namespaces" && property.type === "array") return "namespaces"
if (key === "tenants"    && property.type === "array") return "tenants"
```
This keys purely on field name + array type, so *any* schema with a `namespaces`/`tenants` array now renders the namespace/tenant picker, not just policy targets. Intentional? If not, consider scoping it (e.g. gate on the enclosing policy/rule `$ref`) so an unrelated plugin field named `namespaces`/`tenants` isn't hijacked.

### 🟡 `previewSource` has no reset path in OSS
`previewSource` (store) forces the editor `readOnly` whenever it's `!== undefined`. I don't see where it's cleared in OSS (presumably EE-driven). Please confirm there's a reset on flow/tab switch so a stale preview can't leave the editor stuck read-only.

### 🟢 `KsTag` — `danger` vs `error` are now near-identical
Both resolve to `bg: var(--ks-bg-error)` / `border: transparent`; only the text token differs (`--ks-status-error` vs `--ks-text-error`). Two visually-equivalent variants (one real ElTag type, one class-based) invites confusion — consider consolidating or documenting when to use which. Also confirm `ElTag` accepts `primary` in the exported `KsTagType`.

### 🟢 Minor
- Missing trailing newline (EOF): `TaskTenants.vue`, `taskList/Element.vue`.
- `visibleTags`/`hiddenTags` (`slice(0,3)`/`slice(3)`) is duplicated between `TaskTenants.vue` and `NamespaceSelect.vue` — candidate for a small shared composable.
- `Empty.vue` fallback reorder now lets `typeLinks.learnMore` win over the `demoCta` `"#"`. Correct for Policies, but it's a behavior change for every `demoCta` page whose type has a registry `learnMore` (they now deep-link to docs instead of `"#"`). Confirm that's intended across all demo pages.
- Apps/Assets now fall back to the registry's `watch?v=` video URLs instead of the old `embed/` URLs they passed. Every other registry entry already uses `watch?v=`, so likely fine — worth one visual check that the video still plays.

### ✅ Verified clean (no action)
- `--ks-bg-active` (newly referenced in `TaskObjectField.vue`/`Element.vue`) is defined in all theme files.
- i18n keys `demos.policies.{label,header,title,message}` and `tenant.names` all exist.
- `apps`/`assets` registry entries exist with values matching the removed props, so the demo-file cleanup is safe.

MilosPaunovic (approved):
FE side is now good to go, amazing work, as always @Piyush-r-bhaskar! 🚀

ammeek (approved):
LGTM

PR #17235, opened 2026-07-06
feat(tenants): tenant types infrastructure vertical

RemiBarthe:
## What is in it

**Asset lifecycle on the base class**

- `status`, `ttl` and `owner` move from an EE subtype onto `Asset` itself, stored under the `system.` prefix already reserved for system labels. Every asset type carries them and no subtype redeclares them.
- The `ttl` format is what the expiry buckets compare, as text. It was an undocumented invariant every writer happened to respect; it is now a constraint that rejects an offset other than `Z`, which would otherwise sort by wall clock instead of by instant.

**Asset types from the plugin registry**

- The registry already indexed asset types, the plugin DTO never exposed them. It now lists them like every other plugin element, so a UI can stop hardcoding fully-qualified class names.
- `Custom` is marked internal: it is the fallback an unknown type deserializes into, not a type to offer in a picker.

**Extension points**

- The product tour becomes variant-aware, so an edition can ship its own scenes without forking the tour.
- The data table gains the column and filter machinery the new screens need.

## Companion

EE: kestra-io/kestra-ee#8595.


core/src/main/java/io/kestra/core/models/assets/AssetAction.java
```diff
@@ -0,0 +1,15 @@
+package io.kestra.core.models.assets;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotBlank;
+
+public record AssetAction(
```
brian-mulier-p: Let's share a record between case actions and asset actions:
rename CaseAction to FlowAction (and keep the optional label for each so it's friendlier), move all references from case actions to this renamed one and do the same for your assetActions (you can keep the same member name though to avoid changing migrations, I'm fine with keeping assetActions and caseActions naming, I'm just speaking about the wrapper object)

core/src/main/java/io/kestra/core/models/QueryFilter.java
```diff
@@ -216,6 +216,20 @@ public List<Op> supportedOp() {
...
+                return List.of(Op.EQUALS, Op.NOT_EQUALS, Op.IN, Op.NOT_IN);
+            }
+        },
+        @JsonProperty("leaseStatus")
+        LEASE_STATUS("leaseStatus") {
+            @Override
+            public List<Op> supportedOp() {
+                return List.of(Op.EQUALS, Op.NOT_EQUALS, Op.IN, Op.NOT_IN);
+            }
+        },
```
brian-mulier-p: What is it used for? LOCKED should be enough?

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
```diff
@@ -410,10 +410,18 @@ protected final Condition getConditionOnField(
...
             return findMetadataCondition((Map<?, ?>) value, operation);
         }
 
+        if (QueryFilter.Field.ASSET_STATUS == field) {
+            return assetStatusCondition(value, operation);
+        }
+
+        if (QueryFilter.Field.LEASE_STATUS == field) {
+            return leaseStatusCondition(value, operation);
+        }
```
brian-mulier-p: As said above I think LOCKED is enough, if so remove that otherwise keep it

ui/packages/design-system/src/components/Data/KsDataTable/filter/ColumnToggleRow.vue
```diff
@@ -0,0 +1,91 @@
+<template>
```
brian-mulier-p, line 1: Please create a storybook test to showcase this component, what it can be used for and if needed a small prop-based testing

core/src/main/java/io/kestra/core/models/QueryFilter.java
```diff
@@ -216,6 +216,20 @@ public List<Op> supportedOp() {
                 return List.of(Op.EQUALS, Op.NOT_EQUALS, Op.IN, Op.NOT_IN, Op.CONTAINS);
             }
         },
+        @JsonProperty("assetStatus")
+        ASSET_STATUS("assetStatus") {
+            @Override
+            public List<Op> supportedOp() {
+                return List.of(Op.EQUALS, Op.NOT_EQUALS, Op.IN, Op.NOT_IN);
+            }
+        },
```
brian-mulier-p: Can't you use "status" and override the filter behavior in the Asset repository directly rather than adding yet another field?

ui/packages/design-system/src/components/Data/KsDataTable/filter/composables/useTableColumns.ts
```diff
@@ -7,6 +7,7 @@ export interface ColumnConfig {
     default: boolean;
     description?: string;
     condition?: () => boolean;
+    group?: string;
```
brian-mulier-p, line 12: Add a small doc explaining what it can be used for please

ui/src/components/layout/Labels.vue
```diff
@@ -49,7 +49,10 @@
     interface Label {
         key?: string;
         value: string;
+        /** Replaces the displayed value, keeping the `key:` prefix. */
         display?: string;
+        /** Replaces the whole displayed text, `key:` prefix included. */
+        text?: string;
```
brian-mulier-p: It brings confusion. Please instead provide a boolean `keyPrefix?: boolean` (default true) which you turn off in your case and put the full text you want in the display. Override of an override is never a good idea

brian-mulier-p: Alternative would be having a 
`format?: (label: Omit<Label, 'format'>) => string;`
with default (label) => `${label.key}: ${label.value}`
or something like that, up-to-you to decide, both are ok for me.

ui/src/components/onboarding/tour/TourOverlay.vue
```diff
@@ -172,41 +163,58 @@
...
+    const actions = variant.useActions()
+    const scenes = variant.scenes
+    const sceneIds = sceneIdsOf(scenes)
+    const stepGroups = stepGroupsOf(scenes)
+    const totalSteps = scenes.length
+
+    const {trackOnboarding} = useOnboardingAnalytics({sceneIds, guideId: variant.id})
     const toast = useToast()
 
+    const tk = (suffix: string) => `${variant.i18nPrefix}.${suffix}`
```
brian-mulier-p: what do tk mean? I guess translationKey but it's not clear, please if that's the case name the function `translationKey` so it's easier to grasp.

brian-mulier-p: Or `tourTranslation(originalKey)` that does the full `t(variant.i18nPrefix + "." + suffix)` for you

ui/src/composables/useTableColumns.ts
```diff
@@ -8,6 +8,7 @@ export interface ColumnConfig {
     minWidth?: number;
     description?: string;
     condition?: () => boolean;
+    group?: string;
```
brian-mulier-p: js doc

brian-mulier-p (changes requested):
Good PR, nice work, still some questioning points but it'll make it 👍

core/src/main/java/io/kestra/core/models/QueryFilter.java
RemiBarthe: Two different concepts:

LOCKED is a boolean resolved from the active AssetLocks lease (is someone holding a lock right now?), while LEASE_EXPIRY filters on the asset's TTL (EXPIRING_7D, EXPIRING_30D, EXPIRED, NO_EXPIRY), which powers the expiry filters and counters.

I agree the name made this look like a duplicate though, since the lock mechanism is also backed by a Lease. I've renamed it back to LEASE_EXPIRY, which matches LeaseExpiryBucket and the existing UI naming.

brian-mulier-p (approved):
LGTM

PR #17238, opened 2026-07-06
build(system): allow multi-digit release candidates in Start release guard

anxkhn:
> Note: the repo PR template has Frontend and Backend checklist sections that say
> to delete them when not applicable. This change is CI-only (a GitHub Actions
> workflow + its inline bash guard), so both checklists are removed and the
> validation done is described under Additional Notes.

---

### ✨ Description

The manually dispatched **Start release** workflow
(`.github/workflows/global-start-release.yml`) validates the `releaseVersion`
input with a bash regex before it updates `gradle.properties` and pushes the
`v<version>` tag. Two problems in that guard:

1. **The enforced regex only allows a single RC digit.** The RC fragment is
   `-rc[0-9]`, which matches exactly one digit, so once a release line reaches its
   tenth release candidate the workflow rejects it. `0.21.0-rc9` is accepted but
   `0.21.0-rc10` (and `rc11`, `rc42`, ...) fail with `Invalid release version`,
   blocking a dispatch for those RCs.

2. **The error message advertises a different regex than the one enforced.** On
   failure it printed `^[0-9]+(\.[0-9]+)(\.[0-9]+)-(rc[0-9])?(-SNAPSHOT)?$`, with
   the `-` moved outside the `rc` group. That makes the dash mandatory, so the
   printed pattern would (as written) reject the workflow's own documented example
   `0.21.1`, giving self-contradictory guidance to whoever hits the error.

This PR changes `-rc[0-9]` to `-rc[0-9]+` in the enforced guard so any number of
RC digits is accepted, and rewrites the error message to print the exact pattern
that is enforced so the two stay in sync. Clearly malformed inputs are still
rejected (`0.21`, `1.2.3.4`, a bare `-rc`, `-rcx`).

```diff
-          if ! [[ "$RELEASE_VERSION" =~ ^[0-9]+(\.[0-9]+)(\.[0-9]+)(-rc[0-9])?(-SNAPSHOT)?$ ]]; then
-            echo "Invalid release version. Must match regex: ^[0-9]+(\.[0-9]+)(\.[0-9]+)-(rc[0-9])?(-SNAPSHOT)?$"
+          if ! [[ "$RELEASE_VERSION" =~ ^[0-9]+(\.[0-9]+)(\.[0-9]+)(-rc[0-9]+)?(-SNAPSHOT)?$ ]]; then
+            echo "Invalid release version. Must match regex: ^[0-9]+(\.[0-9]+)(\.[0-9]+)(-rc[0-9]+)?(-SNAPSHOT)?$"
```

### 🔗 Related Issue

No existing issue. This is a small release-infrastructure fix found while reading
the release workflows. Happy to open a tracking issue first if the maintainers
prefer that flow.

### 📝 Additional Notes

The workflow's own `workflow_dispatch` guard is the test surface (there is no unit
harness for a shell block). Behavior was validated by running the enforced regex
under bash `[[ =~ ]]` (the same engine the workflow uses at runtime), before and
after the change:

- Now accepted (were rejected before): `0.21.0-rc10`, `0.21.0-rc11`,
  `0.21.0-rc42`, `0.21.0-rc123`, `0.21.0-rc10-SNAPSHOT`.
- Still accepted: `0.21.1`, `1.0.0`, `10.20.30`, `0.21.0-rc1`, `0.21.0-rc9`,
  `0.21.0-SNAPSHOT`.
- Still rejected: `0.21`, `1.2.3.4`, `0.21.0-rc`, `0.21.0-rcx`, `0.21.0-rc10-FOO`,
  `abc`.

The enforced regex and the printed error message are now byte-for-byte identical.
No production code is touched.


loicmathieu:
@AcevedoR can you take over the review?

.github/workflows/global-start-release.yml
```diff
@@ -27,8 +27,8 @@ jobs:
             exit 1
           fi
 
-          if ! [[ "$RELEASE_VERSION" =~ ^[0-9]+(\.[0-9]+)(\.[0-9]+)(-rc[0-9])?(-SNAPSHOT)?$ ]]; then
-            echo "Invalid release version. Must match regex: ^[0-9]+(\.[0-9]+)(\.[0-9]+)-(rc[0-9])?(-SNAPSHOT)?$"
+          if ! [[ "$RELEASE_VERSION" =~ ^[0-9]+(\.[0-9]+)(\.[0-9]+)(-rc[0-9]+)?(-SNAPSHOT)?$ ]]; then
```
AcevedoR: the '-' change you did below is missing here

AcevedoR (changes requested):

MilosPaunovic:
If you've addressed any of the requested changes from the PR review, @anxkhn, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

.github/workflows/global-start-release.yml
anxkhn: fixed in 23fff9b6 by grouping the separator with rc in both the validation regex and the error message.

anxkhn:
@AcevedoR could you please re-review?

AcevedoR:
thanks a lot! 
It forced me to recheck other internal workflow to realize this change was needed elsewhere as well

AcevedoR (approved):

PR #17240, opened 2026-07-06
refactor: serve plugin icons as real SVG resources instead of data URIs

elevatebart:
### ✨ Description

Replaces `KsTaskIcon`'s client-side base64 decode/recolor/encode of every plugin icon with real, browser-cacheable `image/svg+xml` resources — and along the way, removes every eager bulk icon fetch (local and external) in favor of per-icon, on-demand resolution.

**Backend**
- New `GET /api/v1/plugins/icons/{cls}/icon.svg` endpoint serves the decoded SVG directly, for use in `` or CSS `mask-image`. Falls back to the group/subgroup icon index when the class-level lookup misses, so plugin-group icons (used by the plugin catalog) resolve too, not just task/trigger classes.
- `PluginIcon` gains two fields:
  - `monochrome` — true when the SVG uses `currentColor` (computed once, server-side).
  - `hash` — a content hash of the icon bytes, so the frontend can cache `icon.svg` **indefinitely** (`max-age=31536000, immutable`) by busting the URL with `?v={hash}` only when the actual bytes change.
  - Kept a backward-compatible 3-arg constructor for callers built against the DTO's pre-redesign shape (e.g. EE).
- `RegisteredPlugin` gained `iconAndMonochrome()` (Class- and String-keyed) replacing direct base64 encoding.
- `PluginCatalogService` (backs the EE plugin-manager's "available plugins" catalog) no longer bundles every plugin group's base64 icon into the manifest on every load. `PluginManifest` now carries the plugin `group` key instead, and a new `icon(groupId, artifactId)` lazily resolves + caches a single artifact's icon bytes on demand — the companion this PR's EE counterpart needs to serve icons as standalone image resources instead of inlining them.

**Frontend**
- `KsTaskIcon` now renders non-monochrome icons as a plain `` pointed at the real endpoint, and monochrome icons as a CSS `mask-image` (recolorable via the `variable` prop or `--ks-text-primary`) — no more `atob`/`btoa`/regex substitution on every reactive render. A new `customIcon` prop lets callers supply an already-resolved icon (URL or data URI) for cases where the bytes come from elsewhere entirely (e.g. a plugin catalog row).
- Every registered task/trigger class still gets an `icons` map entry regardless of whether it ships an icon — `KsTaskIconData` carries an explicit `hasIcon` flag so `KsTaskIcon` falls back to a generic icon instead of pointing at a URL that would 404.
- Icons for ecosystem plugins not installed on the local instance (e.g. shown in Blueprints, or plugin-catalog groups) come from the external `api.kestra.io` plugin catalog. This is no longer fetched eagerly in bulk — `pluginsStore.loadIcon()` resolves them lazily, one class at a time, only for classes actually rendered, falling back to the ecosystem catalog when the local instance doesn't know the class. Blueprint- and plugin-catalog-related components now pass `loadIcon` alongside `icons` for this fallback.
- Ecosystem icon existence is probed via a plain `new Image()` load rather than `fetch`/`axios`: reading a cross-origin response body needs CORS headers that `api.kestra.io`'s per-class endpoint doesn't send, so an XHR-based sniff (originally used to also detect `monochrome`) was hitting real CORS errors in the browser. A bare image load needs no such headers, at the cost of no longer being able to detect `monochrome` for ecosystem icons (they render at their native colors).
- Handles `KESTRA_BASE_PATH` correctly for reverse-proxy path-prefixed deployments (trims a trailing slash to avoid an accidental protocol-relative `//` URL).

### 🎨 Frontend Checklist

- [x] All unit tests updated and passing (KsTaskIcon, plugins store, pluginUtils)
- [x] Storybook stories updated: default/monochrome/fallback/registered-without-icon/ecosystem-catalog/custom-icon variants
- [x] Verified interactively via Storybook + Playwright screenshots (light/dark mode, mask recoloring, fallback paths)

### 🛠️ Backend Checklist

- [x] New/updated unit tests: `PluginControllerTest` (`icon.svg` endpoint, group-icon fallback, 404 for unknown classes, hash stability/uniqueness, immutable cache header), `RegisteredPluginTest` (monochrome detection), `PluginCatalogServiceTest` (lazy `icon()` resolution)

### 📝 Additional Notes

Iterated through several rounds against a real local instance to catch issues that only show up with a live plugin registry rather than mocked data: a `KESTRA_BASE_PATH` double-slash bug, a "registered but iconless" class incorrectly getting a `/icon.svg` URL, the external ecosystem-catalog icon source needing its own resolution path entirely, that same external source never having sent a `monochrome` field to begin with, dropping the ~8MB eager bulk fetch (both local and external) in favor of lazy per-class resolution, a real CORS failure once the ecosystem fallback started reading response bodies cross-origin, and plugin-group/subgroup icons 404ing because the class-level lookup didn't fall back to the group icon index.

This PR is paired with [kestra-io/kestra-ee#9249](https://github.com/kestra-io/kestra-ee/pull/9249) (same branch name, so cross-repo CI picks up both together) — that PR extends this redesign to EE's plugin-manager UI, which had the same "inline everything as base64" problem for its own plugin catalog.

OSS backend/frontend suites are fully green on the latest commit.

### 🤖 AI Authors

Why did the SVG refuse to compile? It had too many *currentColor* issues and needed to mask its `fill`ings. 🐱


core/src/main/java/io/kestra/core/docs/PluginIcon.java
```diff
@@ -11,4 +11,19 @@ public class PluginIcon {
     String name;
     String icon;
     Boolean flowable;
+    Boolean monochrome;
+    /**
+     * Content hash of {@link #icon}, null when there is no icon. Lets callers cache-bust a
+     * long-lived, immutable URL for the icon bytes (see PluginController's {@code icon.svg}
+     * endpoint) — the URL only needs to change when this hash does.
+     */
```
elevatebart:
```suggestion
```

core/src/main/java/io/kestra/core/docs/PluginIcon.java
```diff
@@ -11,4 +11,19 @@ public class PluginIcon {
...
+    String hash;
+
+    /**
+     * Backward-compatible constructor for callers built against the DTO's shape before
+     * {@link #monochrome} and {@link #hash} were added.
+     */
```
elevatebart:
```suggestion


```

core/src/main/java/io/kestra/core/plugins/PluginCatalogService.java
```diff
@@ -231,6 +190,60 @@ private List<PluginManifest> load() {
...
+     * <p>
+     * The icon is proxied from the Kestra API on demand and cached per plugin group, so the catalog
+     * of icons is never fetched or transferred as a whole — only icons that are actually requested
+     * (i.e. rendered) are loaded. Returns empty when icon resolution is disabled for this instance,
+     * the artifact/group is unknown, or the Kestra API has no icon for the group.
+     *
+     * @param groupId    the Maven group identifier of the artifact.
+     * @param artifactId the Maven artifact identifier.
+     * @return the raw SVG bytes, or empty if none could be resolved.
+     */
```
elevatebart:
```suggestion
```

core/src/main/java/io/kestra/core/plugins/RegisteredPlugin.java
```diff
@@ -311,8 +311,25 @@ public String version() {
...
+
+    public String icon(String iconName) {
+        return iconAndMonochrome(iconName).map(IconAndMonochrome::icon).orElse(null);
+    }
+
+    /**
+     * Resolves, sanitizes and base64-encodes the SVG icon for a class, together with whether the
+     * SVG is a single-color icon (uses {@code currentColor} rather than fixed brand colors) and
+     * can therefore be recolored via CSS instead of shipping a fixed color.
+     */
```
elevatebart:
```suggestion
```

ui/packages/design-system/src/components/Kestra/KsTaskIcon.vue
```diff
@@ -4,35 +4,53 @@
...
+         * serve locally (e.g. an ecosystem plugin catalog entry) — a data URI or an absolute URL.
+         * Takes priority over the local `/icon.svg` endpoint when present.
+         */
+        iconUrl?: string;
+        /**
+         * Content hash of the icon bytes, appended as a cache-busting query param to the local
+         * `/icon.svg` URL so the browser can cache it indefinitely — the URL only needs to change
+         * when this hash does.
+         */
+        hash?: string;
```
elevatebart:
```suggestion
        hasIcon: boolean;
        iconUrl?: string;
        hash?: string;
```

RemiBarthe (approved):

PR #17243, opened 2026-07-06
feat: Policies OSS UI

Piyush-r-bhaskar:


ui/packages/design-system/src/components/Data/KsTag/KsTag.vue
```diff
@@ -69,19 +73,19 @@
             text: var(--ks-status-success),
         ),
         warning: (
-            bg: var(--ks-log-background-warn),
-            border: var(--ks-log-border-warn),
-            text: var(--ks-status-warn),
+            bg: #FF8B611A,
```
fhussonnois: I guess you should use --ks var here ? same for other `pg` props

Piyush-r-bhaskar: yes indeed, Thanks.

PR #17248, opened 2026-07-06
build: remove husky, use gradle-managed git hook

Skraye:
Removes Husky from the UI package; the Gradle `spotless-conventions` plugin now owns the git hook.

Husky and the plugin both set `core.hooksPath`, and Husky's `prepare` script re-pointed it to `ui/.husky/_` on every `npm i`, overriding the Gradle hook — and Husky's pre-commit only ran frontend `lint-staged`, so `spotlessApply` on staged Java files was skipped. This is the cause of the recurring Java formatting drift. The Gradle hook already runs `lint-staged` for the UI, so no coverage is lost.

After merging, existing clones must re-point their hook once: `./gradlew clean build` (or `git config core.hooksPath .github/.hooks`).

EE companion: kestra-io/kestra-ee#9208

### DIFF
--- a/ui/.husky/pre-commit
+++ b/ui/.husky/pre-commit
@@ -1,1 +1,0 @@
-cd ui && npx lint-staged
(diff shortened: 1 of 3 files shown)


brian-mulier-p (approved):
LGTM

PR #17252, opened 2026-07-06
refactor/type converter

fhussonnois:
Consolidates duplicated ad-hoc parsing (boolean/number/date/duration/enum/
list) into io.kestra.core.utils.TypeConverter: null-safe, passthrough when
already typed, strict toString() parse otherwise. Failures throw
TypeConversionException, which extends IllegalArgumentException on purpose
so the existing HTTP 400 mapping and repository filter wrapping apply as-is.

Migrated call sites: flow input parsing, KV, JDBC/webserver filter coercion,
purge/version tasks, scheduler, CLI. Conversion errors now surface as
consistent 400s instead of raw parse exceptions (500) on some paths.

Part-of: https://github.com/kestra-io/kestra-ee/pull/9216

### DIFF
--- a/core/src/test/java/io/kestra/core/utils/TypeConverterTest.java
+++ b/core/src/test/java/io/kestra/core/utils/TypeConverterTest.java
@@ -1,0 +1,137 @@
+package io.kestra.core.utils;
+
+import io.kestra.core.exceptions.TypeConversionException;
+import org.junit.jupiter.api.Named;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeParseException;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class TypeConverterTest {
+
+    private enum TestEnum {
+        FIRST, SECOND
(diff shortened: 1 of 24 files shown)


core/src/test/java/io/kestra/core/utils/TypeConverterTest.java
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    @Test
+    void shouldReturnSameInstanceWhenBoolean() {
+        assertThat(TypeConverter.toBoolean(Boolean.TRUE)).isSameAs(Boolean.TRUE);
+    }
+
+    @Test
+    void shouldParseTrueWhenStringTrueIgnoringCase() {
+        assertThat(TypeConverter.toBoolean("true")).isTrue();
```
AcevedoR: dont hesitate to do TableTesting via @ParameterizedTest

AcevedoR (approved):

core/src/main/java/io/kestra/plugin/core/execution/PurgeExecutions.java
```diff
@@ -141,7 +142,7 @@ public PurgeExecutions.Output run(RunContext runContext) throws Exception {
             renderedNamespace,
             runContext.render(flowId).as(String.class).orElse(null),
             runContext.render(startDate).as(String.class).map(ZonedDateTime::parse).orElse(null),
-            ZonedDateTime.parse(runContext.render(endDate).as(String.class).orElseThrow()),
```
loicmathieu, line 144: In such case we know it's not null and a string so using `ZoneDateTime.parse()` will be quicker.

loicmathieu (approved):
LGTM

A small concern as for the performance overhead if we know it's not null and a string, direct parsing via Inetger/Longf/ZoneDateTime.parse() will always be quicker as it avoids null check and toString.
I reported it once but it apply to multiple call site.
I acknowledge that consistency may be better here so I let it up to you

PR #17255, opened 2026-07-06
build: run backend check serially with --rerun-tasks (CI experiment — do not merge)

AcevedoR:
## ⚠️ CI experiment — do not merge

Overrides the `backend` job (normally the reusable `kestra-oss-backend-tests.yml`) with an inline copy whose Gradle step runs:

```bash
./gradlew check --rerun-tasks --no-parallel
```

instead of `./gradlew check javadoc --parallel`, to observe how the test suite behaves with **zero Gradle parallelism** and **every test forced to run**.

### Notes
- `--no-parallel` is needed on top of dropping `--parallel`: `gradle.properties` sets `org.gradle.parallel=true`, so the plain command would still parallelize.
- `--rerun-tasks` defeats up-to-date checks and the build cache, so no test task can be skipped.
- The `file-changes` gate is dropped for this job because the backend paths-filter excludes `.github/**` — this PR would otherwise skip the backend job entirely.
- `timeout-minutes` raised 90 → 300 since the serial run will be much slower.

Companion EE PR: kestra-io/kestra-ee#9215 (the EE branch was pushed first, so the `trigger-ee` dispatch here is skipped in favor of the EE PR's own CI).

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -115,19 +115,67 @@ jobs:
     secrets:
       GITHUB_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  # EXPERIMENT (do not merge): inlined from
+  # kestra-io/actions/.github/workflows/kestra-oss-backend-tests.yml@main to run the
+  # Gradle check serially with every task forced to re-run.
+  # - `needs`/`if` on file-changes are intentionally dropped: the backend filter
+  #   excludes `.github/**`, so this workflow-only PR would otherwise skip the job.
+  # - `--no-parallel` is required on top of dropping `--parallel`: gradle.properties
+  #   sets `org.gradle.parallel=true`, which would silently re-enable parallelism.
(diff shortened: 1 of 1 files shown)


AcevedoR:
Serial run verdict: **7,330/7,330 tests executed (0 from cache), 1 failure** — `H2RunnerTest.flowTriggerWithPause()` ("3 Execution found for flow trigger-flow-listener-with-pause, but 4 where awaited"), a timing race independent of parallelism. Companion EE run (kestra-io/kestra-ee#9215) was fully green serially: 6,316/6,316, 0 failures. See the EE PR for the full experiment summary.

AcevedoR:
Serial vs parallel (identical `--rerun-tasks` workload): serial ~42 min, 7,330/7,330 executed, 1 failure (`flowTriggerWithPause` timing race); parallel 27 min, 7,333/7,333 executed, 0 failures. Parallelism costs nothing in correctness here and saves ~35% wall clock. Full comparison incl. EE: kestra-io/kestra-ee#9215.

PR #17259, opened 2026-07-06
fix(executions): restrict draft revision execution to the playground

AcevedoR:

Draft revisions must only ever run as **PLAYGROUND-kind executions**. This PR enforces that rule in the backend and fixes the two UI run entry points, which had the draft behaviour inverted.

#### Backend enforcement (`ExecutionController`)

- `createExecution` rejects an explicit draft revision unless `kind=PLAYGROUND` (422, checked before draft validation).
- The two replay endpoints reject an explicit draft revision when the source execution is not a playground one, so a draft cannot be smuggled into a NORMAL execution through a replay. Playground replays keep working since the replayed execution inherits the parent's kind.
- EE is covered too: its `ExecutionController` delegates to `super.createExecution`, and EE test suites emit the `Create` command directly (never through this endpoint), so TEST-kind runs are unaffected.

#### UI

- **Flow page (Execute dialog)** passed the loaded flow's revision to the create-execution API, so when the latest revision was a draft it executed the draft. It now omits the `revision` parameter for drafts, letting the backend resolve the **latest published revision** (and return the existing "save a published revision first" error for draft-only flows). Non-draft flows behave exactly as before.
- **Playground** passed no revision, so after a draft save the backend resolved the latest published revision and silently ran **stale code** instead of what is in the editor (or failed outright for draft-only flows). It now passes the current revision explicitly (with `kind=PLAYGROUND`), matching the playground's replay path which already did.
- **"Save & Execute"** in the editor now also omits the revision for drafts (running the latest published one, like the flow page) instead of sending a draft revision with `kind=NORMAL`, which the backend now refuses.
- The `draft_flow_warning` alert in the Execute dialog claimed the execution would run the draft, which is no longer true — reworded it in all 13 languages to say the latest published revision runs instead, and to point users at publishing or the playground to run their draft.

### Checks

- New tests in `ExecutionControllerRunnerTest`: draft + no kind → 422 (playground-only message); draft + `kind=PLAYGROUND` → execution created on the draft revision with PLAYGROUND kind. Existing invalid-draft validation test updated to pass the new gate. All 4 draft tests pass.
- `node check.js` (translations): no missing/extra keys
- `npm run check:types`: passes
- `FlowRunActions.spec.ts`: 2/2 pass

Fixes kestra-io/kestra-ee#8680


loicmathieu (approved):
LGMT.

i suggest reviewing the comments inside the UI as they seems a bit too verbose

AcevedoR:
> i suggest reviewing the comments inside the UI as they seems a bit too verbose
done

PR #17260, opened 2026-07-06
feat(dashboards): broaden chart export to all data charts, add CSV/ION format, default dashboard sentinel

brian-mulier-p:
- Renames the chart export routes from CSV-only `.../export/to-csv` to `.../export?format=CSV|ION`
- Broadens the export guard from Table-only to every data chart (Table, TimeSeries, Pie, Bar, KPI); Markdown stays excluded
- Adds a reserved `_default` dashboard id, resolved server-side to the built-in home dashboard, so it can be addressed without a saved dashboard (guarded against user creation) — known low-probability edge case: an installation with a pre-existing saved dashboard literally named `_default` from before this change would have it shadowed by the sentinel; not migrated automatically since leading-underscore ids were never a realistic pre-existing pattern
- Moves the three bundled default dashboard definitions from frontend assets to backend resources, served via a new endpoint — removes prior frontend/backend drift on those files
- Adds a chart Export dropdown (CSV/ION) to every data chart on the dashboard view
- Adds `dashboardId`/`chartId` autocompletion in both the no-code task form and the YAML/Monaco editor, falling back to the default dashboard's charts when `dashboardId` is empty
- Fixes `getTaskComponent.ts`'s sibling-property dispatch, which read sibling keys from the flow-schema `definitions` map instead of the task's own resolved schema properties and always produced an empty list — this silently broke the existing `Subflow` task's `flowId` autocompletion too, not just the new fields added here
- Companion to kestra-io/client-sdk#323 and kestra-io/plugin-kestra#157 (new `Export` task, kestra-ee#9112); not to be merged before Kestra 2.0 - draft until then


webserver/src/main/java/io/kestra/webserver/controllers/api/DashboardController.java
```diff
@@ -193,6 +247,22 @@ public HttpResponse<DashboardResponse> updateDashboard(
         return HttpResponse.ok(new DashboardResponse(this.save(existingDashboard.get(), dashboardToSave, dashboard)));
     }
 
+    private void assertNotReservedId(String id, Dashboard dashboardParsed) {
+        if (DEFAULT_DASHBOARD_ID.equals(id)) {
+            throw new ConstraintViolationException(
```
fhussonnois: Just throw our `InvalidException` instead of the `ConstraintViolationException`.  We tend to use to much the Bean Validation API as a general exception.  e.g.: `new InvalidException("Dashboard id '" + DEFAULT_DASHBOARD_ID + " is reserved.'")`

webserver/src/main/java/io/kestra/webserver/controllers/api/DashboardController.java
```diff
@@ -109,6 +127,39 @@ public DashboardResponse getDashboard(
...
+            }
+            return new String(is.readAllBytes(), StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
     @ExecuteOn(TaskExecutors.IO)
     @Post(consumes = MediaType.APPLICATION_YAML)
     @Operation(tags = { "Dashboards" }, summary = "Create a dashboard from yaml source")
```
fhussonnois: Add to the `@Operation`: 
```
responses = {
@ApiResponse(
    responseCode = "422",
    description = "If the dashboard id is reserved ('" + DashboardController.DEFAULT_DASHBOARD_ID +"')"
  )
}
```

webserver/src/test/java/io/kestra/webserver/controllers/api/DashboardControllerTest.java
```diff
@@ -686,11 +694,258 @@ void exportADefaultDashboardChartToCsv() {
...
 
         // export CSV
-        byte[] csvBytes = client.toBlocking().retrieve(POST(DASHBOARD_PATH + "/charts/export/to-csv", previewRequest), Argument.of(byte[].class));
+        byte[] csvBytes = client.toBlocking().retrieve(POST(DASHBOARD_PATH + "/charts/export", previewRequest), Argument.of(byte[].class));
         var csv = new String(csvBytes, StandardCharsets.UTF_8);
         assertThat(csv).isEqualTo("chart_namespace,chart_execution_id\r\n%s,%s\r\n".formatted(fakeNamespace, fakeExecutionId));
     }
 
+    @Test
+    void exportNonTableDataChartToCsv() {
```
fhussonnois: Please follow test method naming convention ;)

webserver/src/main/java/io/kestra/webserver/controllers/api/DashboardController.java
```diff
@@ -109,6 +127,39 @@ public DashboardResponse getDashboard(
...
+    @Operation(tags = { "Dashboards" }, summary = "Get the built-in default dashboard definitions")
+    public Map<String, String> getDefaultDashboardDefinitions() {
+        return Map.of(
+            "main", readClasspathResource(DEFAULT_MAIN_DEFINITION_RESOURCE),
+            "flow", readClasspathResource(DEFAULT_FLOW_DEFINITION_RESOURCE),
+            "namespace", readClasspathResource(DEFAULT_NAMESPACE_DEFINITION_RESOURCE)
+        );
+    }
+
+    private Dashboard defaultDashboard() {
```
fhussonnois: Move this method to the `Dashboard` class as a public static method.

webserver/src/main/java/io/kestra/webserver/controllers/api/DashboardController.java
```diff
@@ -69,6 +73,16 @@
 public class DashboardController {
     public static final Pattern DASHBOARD_ID_PATTERN = Pattern.compile("^id:.*$", Pattern.MULTILINE);
 
+    /**
+     * Reserved id resolving to the built-in default dashboard definition. Not a real stored dashboard, never
+     * shown in dashboard id autocompletion — it's a backend/API sentinel only (e.g. used by export tooling
+     * when no explicit dashboardId is provided).
+     */
+    public static final String DEFAULT_DASHBOARD_ID = "_default";
```
fhussonnois: This const should probably be moved to the `Dashboard` class

fhussonnois (approved):
LGTM minor the few comments to be fixed before merge.

webserver/src/test/java/io/kestra/webserver/controllers/api/DashboardControllerTest.java
brian-mulier-p: Done before the reminder on the other PR, will make Claude change these for sure :+1:

PR #17262, opened 2026-07-06
feat(core): add 'sys purge-queue' CLI command and fix queue purge lock timeout

nkwiatkowski:
## Context

A customer's `queues` table (MySQL) grew to ~8.6M rows and the executor's scheduled cleaner could no longer purge it — it hit `Lock wait timeout exceeded` and the uncaught-exception handler shut the server down. Because **the executor is the component that runs the purge**, once it can't boot it can never clean up. Operators were reduced to running manual `DELETE ... LIMIT` scripts against the DB.

## Root cause

`JdbcCleaner` batches MySQL deletes (`.limit(10_000)`) but ran the **entire batch loop inside a single `dslContextWrapper.transaction(...)`**. All batch-DELETEs accumulate row locks in one uncommitted transaction until the loop finishes, so on millions of rows it always trips `innodb_lock_wait_timeout`. Statement-level batching doesn't help when locks are held until commit — which is exactly why a manual script that **commits between batches** works.

## Changes

- **`JdbcQueueCleaner`** (new) — server-type-agnostic bean holding the purge logic. The MySQL branch now **commits per batch** (each `.limit(10_000)` delete in its own transaction), so the purge can grind through millions of rows without holding locks. Exposes `purge()` (configured retention) and `purge(Duration)` (global override).
- **`JdbcCleaner`** — reduced to a thin `@Scheduled` delegate; the executor's scheduled purge inherits the per-batch-commit fix.
- **`sys purge-queue`** CLI command — purges the `queues` table **without starting any server component**, so an instance whose executor can't boot can still be recovered. Uses the configured `kestra.jdbc.cleaner` retention, with an optional `--retention` override. Guards on `kestra.queue.type` (h2/mysql/postgres).

Usage: stop the components, then `kestra sys purge-queue -c <config>`.

## Tests

- `AbstractJdbcCleanerTest` passes on **H2, MySQL, and Postgres** (covers the refactored `purge()` incl. the MySQL per-batch path).
- New `PurgeQueueCommandTest`: the command resolves `JdbcQueueCleaner` and runs to success in a **no-server-type** context (the crux of the feature), plus graceful handling of non-JDBC queue types.
- Full `cli sys` command package still green.

### DIFF
--- a/cli/src/main/java/io/kestra/cli/commands/sys/PurgeQueueCommand.java
+++ b/cli/src/main/java/io/kestra/cli/commands/sys/PurgeQueueCommand.java
@@ -1,0 +1,59 @@
+package io.kestra.cli.commands.sys;
+
+import java.time.Duration;
+import java.util.Optional;
+
+import io.kestra.cli.AbstractCommand;
+import io.kestra.jdbc.runner.JdbcQueueCleaner;
+
+import io.micronaut.context.ApplicationContext;
+import jakarta.inject.Inject;
+import lombok.extern.slf4j.Slf4j;
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = "purge-queue",
+    description = { "Purge expired messages from the JDBC 'queues' table",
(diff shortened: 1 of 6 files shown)


loicmathieu (approved):
LGTM

PR #17264, opened 2026-07-06
test(executions): extract state-store contracts and prove the in-memory fakes against them

AcevedoR:
Stacked on #17257 (base branch: `test/executor-testkit-phase0`) — rebase onto `develop` once that PR merges.

> **Rebased on fresh `develop` (2026-07-07):** develop now contains the merged #17246 inline fix; the 2nd commit's `ExecutionDelayProcessor` extraction supersedes it structurally (same outbox ordering, enforced by construction), so the rebase resolved in favor of the extraction.

## What

Turns the two abstract state-store test suites into reusable **behavioral contracts** with a `store()` seam, and holds the executor testkit's in-memory fakes to them. This closes the main credibility gap of the unit-test kit: until now nothing proved that `InMemoryExecutionStateStore` / `InMemoryMultipleConditionStateStore` actually behave like the JDBC/Elasticsearch stores they stand in for.

## How — zero scenarios deleted or modified

The scenarios of `AbstractExecutionStateStoreTest` (5) and `AbstractMultipleConditionStateStoreTest` (22) moved **verbatim** one level up into annotation-free superclasses:

```
ExecutionStateStoreContract                    MultipleConditionStateStoreContract
  (plain JUnit, abstract store())                (plain JUnit, abstract store())
        ▲                                              ▲
        ├─ AbstractExecutionStateStoreTest             ├─ AbstractMultipleConditionStateStoreTest
        │    (@MicronautTest + @Inject, unchanged      │    (@MicronautTest + @Inject)
        │     name/field/flow() for subclasses)        │
        │      ▲ H2 / Postgres / Mysql / EE ES         │      ▲ H2 / Postgres / Mysql / EE ES
        │                                              │
        └─ InMemoryExecutionStateStoreTest (kit)       └─ InMemoryMultipleConditionStateStoreTest (kit)
```

- The legacy `Abstract*` classes keep their names, annotations, injected fields (including the `protected executionStateStore` that `ElasticsearchExecutionStateStoreTest` uses directly in its fencing test) and the `flow()` helper — **every existing JDBC and EE subclass compiles and runs unchanged**.
- The new kit subclasses run the same scenarios against the fakes in plain JUnit — no Micronaut, no database.

## Verified

- `:executor:test`: **216 tests, 0 failures** (189 from the base branch + 27 contract runs against the fakes)
- `:jdbc-h2:test --tests "H2ExecutionStateStoreTest" --tests "H2MultipleConditionStateStoreTest"`: 27 tests, 0 failures — the container-injected path still runs every inherited scenario
- kestra-ee `repository-elasticsearch-ee:compileTestJava`: green, no EE change needed

---

## 2nd commit — the #17246 bug class, made testable and impossible

**Relationship to #17246: this absorbs that fix structurally.** If #17246 merges first, this rebases to keep the extraction + tests; if this merges first, #17246 can be closed.

`DefaultExecutor.executionDelayLoop()` published each delayed execution's event from **inside** the delay store's JDBC transaction — on Kafka the consumer could observe the event before the execution row committed, find nothing, and silently drop `CREATE_NEW_EXECUTION` retries (see #17246 for the full trace).

Instead of just moving the emission (a fix any future edit can regress), the per-delay logic moves **verbatim** into a new `ExecutionDelayProcessor` that **has no queue dependency at all** — it collects and returns the `ExecutorContext`s; `DefaultExecutor`'s loop emits them only after `processExpired()` returns, i.e. after the transaction commits. The transactional-outbox rule is now enforced by construction, not by code-review vigilance.

Two regression nets on top:

- **`ExecutionDelayProcessorTest`** (8 tests, plain JUnit over the harness, ~ms): the outbox invariant (processing persists the replayed execution under its own id and emits **nothing** on any of the 7 queue channels — this is the test that fails if anyone re-inlines an emission), plus the per-delay-type decision matrix (replay minting with `attemptNumber+1` + REPLAY label, scheduled-start resume, KILLING/terminated guards, missing-flow failure capture, not-yet-due, missing-execution skip). The end-to-end twin is `KafkaRunnerRetryTest.retryNewExecution*` — 30s+ over a real broker vs. milliseconds here.
- **Contract gap closed**: the state-store contract had no different-id lock scenario, and the in-memory fake was silently diverging from JDBC (it persisted the replayed execution under the *locked* id, overwriting the original). New contract test `shouldInsertReturnedExecutionAndKeepOriginalWhenLockReturnsDifferentId` now runs against H2/Postgres/MySQL/ES **and** the fixed fake.

Verified: `:executor:test` **225 tests, 0 failures**; `H2ExecutionStateStoreTest` green including the new scenario; `H2RunnerTest` green (executor production change gate); kestra-ee `jdbc-ee:compileJava` + `repository-elasticsearch-ee:compileTestJava` green.

---

## 3rd commit — concurrency-limit lifecycle sagas + a third store contract

`ConcurrencyDecisionTest` only covered the pure behavior switch (`processExecutionRunning`). This commit covers the machinery around it:

- **`ConcurrencyLifecycleTest`** (8 sagas, plain JUnit, ~ms): through the real `ExecutionEventMessageHandler` — a RUNNING execution claims a counter slot; QUEUE parks the execution in the queued store *without* claiming a slot and short-circuits processing (`handleConcurrencyLimit`); CANCEL/FAIL terminate without touching the counter; counters are keyed per tenant/namespace/flow and **shared across revisions**; killing a QUEUED execution removes it from the queued store without releasing a slot; a restarted FAILED execution re-enters the concurrency gate. Full-runner twins: the `FlowConcurrencyCaseTest` scenarios (seconds each over a real backend).
- **`ConcurrencyLimitStateStoreContract`** (5 scenarios): counting via `countThenProcess`, decrement floor at zero, FIFO pop with the freed slot handed to the popped execution, and the queued-protection guard (inflated counter → nothing dequeued until slots actually free). Runs against **H2** via the new `AbstractConcurrencyLimitStateStoreTest` shell and against the testkit fake — same pattern as the other two contracts.
- **Fake-fidelity bug caught by the contract**: `InMemoryConcurrencyLimitStateStore.decrementAndPop` did not re-increment the counter when popping a queued execution, while `AbstractJdbcConcurrencyLimitStateStore` hands the freed slot to the popped execution inside the pop consumer. Fixed.

Verified: `:executor:test` **235 tests, 0 failures**; `H2ConcurrencyLimitStateStoreTest` 5/5 green; kestra-ee compile gates green.

> ~~Noted for a follow-up: the queued-pop path in `DefaultExecutor.toExecution` emits the popped execution **inside** the `decrementAndPop` transaction (same shape as the #17246 bug class fixed by the 2nd commit) — same `ExecutionDelayProcessor`-style extraction applies.~~ **Done in the 5th commit below.**

---

## 4th commit — more concurrency sagas + the quota gate

Four more concurrency-limit lifecycle sagas in `ConcurrencyLifecycleTest` (now 12):

- **Boundary**: with a limit of 2, both slots fill and run — the limit does not trip one short; the third arrival queues.
- **Ordering**: executions queued behind a full slot park in arrival order — the order `decrementAndPop` will replay them in.
- **Tenant keying**: same flow id AND namespace in two tenants → independent counters (completes the namespace/revision keying matrix).
- **Limit lowered between revisions**: two executions admitted under revision 1's limit of 3 keep their slots; a revision-2 arrival with limit 1 queues — the *latest* flow definition's limit governs the shared, revision-less counter.

New **`QuotaGateTest`** (6 sagas): quota *evaluation* is EE (the OSS `QuotaService` throws unconditionally), but the gate in `ExecutionEventMessageHandler` is OSS executor logic — driven here through the harness's stubbed `QuotaService`:

- FAIL/CANCEL stop the execution terminally before any task run, `updatedFrom("processQuotas")`.
- An exceeded quota short-circuits **before** the concurrency gate: a free slot is never claimed, nothing reaches the queued store.
- A passing quota is a pure pass-through (execution runs, claims its slot).
- A flow **without** quotas never consults the service — load-bearing, since the OSS implementation throws on every call.
- A `failedThenRestarted` execution re-enters the quota gate like a fresh one.

Verified: `:executor:test` **245 tests, 0 failures**; kestra-ee `jdbc-ee:compileJava` + `repository-elasticsearch-ee:compileTestJava` green.

---

## 5th + 6th commits — the two remaining outbox violations in `DefaultExecutor`, extracted

The follow-up flagged in the 3rd commit, done. Both remaining emit-inside-transaction sites in `DefaultExecutor` now use the same seam as `ExecutionDelayProcessor` — the transactional-outbox rule is enforced by construction across all three loops.

**`ConcurrencySlotReleaseProcessor`** (5th commit): the queued-execution pop emitted the popped execution and fired flow triggers **inside** the `decrementAndPop` transaction. The release guards and the decrement/pop moved verbatim into the new processor (no queue dependency); it returns the popped execution already marked RUNNING and `DefaultExecutor` emits it after the transaction commits. Side effect: the three release guards — queued-then-killed, concurrency short-circuit termination, duplicate KILLED events — become unit-testable for the first time. `ConcurrencySlotReleaseProcessorTest` (7 tests, ~ms) covers the outbox invariant, the slot handover, and every guard, including the subtle one: a queued-then-killed execution *passes* the duplicate-kill guard (KILLING → KILLED) yet must not release a slot it never claimed.

**`SLAMonitorProcessor`** (6th commit): the SLA-monitor loop called `toExecution()` — emitting the execution's TERMINATED/UPDATED events — **inside** `SLAMonitorStateStore#processExpired`'s transaction. The per-monitor logic moved verbatim into the new processor; the loop emits the returned contexts after the transaction commits. `SLAMonitorProcessorTest` (7 tests): outbox invariant, FAIL/CANCEL/labels-only violation behaviors, SLA-removed and execution-terminated races, deadline boundary.

Writing that test found a **new production bug, fixed in the same commit**: removing the **last** SLA from a flow while a monitor is pending leaves `flow.getSla()` null → the loop NPE'd inside `processExpired()`'s transaction → rollback → the monitor is never consumed → the SLA loop re-throws every tick and wedges for **all** monitors. Fixed with `ListUtils.emptyOnNull` (the null-safety `SLAService` already uses).

**Known residual, documented + pinned by a test, not fixed here:** for FAIL/CANCEL SLA behaviors, `ExecutorService#processViolation` → `markAs` emits the kill request from inside the execution lock. That emission is shared with the execution-changed SLA path (`handleExecutionChangedSLA`), so it needs its own seam — an emission accumulator on `ExecutorContext` — rather than a local fix.

`bug-report.md` updated accordingly (§1.2–1.4).

Verified: `:executor:test` **259 tests, 0 failures**; `H2RunnerTest` 100/100 (covers the SLA + concurrency scenarios end-to-end via `AbstractRunnerTest`); kestra-ee `jdbc-ee:compileJava` + `repository-elasticsearch-ee:compileTestJava` green.

---

## 7th commit — concurrency drain sagas, groundwork for hierarchical (namespace/tenant) limits

Preparation for the upcoming namespace/tenant-level concurrency limit feature (kestra-ee#8660): pin the flow-level semantics that feature will have to preserve — or deliberately change — so the implementation starts against a regression net instead of a blank page.

`ConcurrencyLifecycleTest` grows to 16 sagas (+4):

- **The full drain saga** through the handler + `ConcurrencySlotReleaseProcessor` pair (the exact orchestration of `DefaultExecutor#toExecution`): slot handover on termination, FIFO pop order, decrement back to zero when the queue empties.
- **A popped execution never claims a second slot**: `DefaultExecutor`'s executionQueue consumer persists the message first, then the handler sees it as an UPDATED event — the concurrency gate only guards CREATED/restarted-failed executions.
- **Pop scope is keyed per flow**: a freed slot never bleeds into another flow's queue, even in the same namespace — the exact semantic namespace-level limits will deliberately relax.
- **Unlimited flows are invisible to the machinery**: no counter touch, nothing queued, no release — the invariant hierarchical limits will flip (a flow without limits must count toward its namespace's).

`ConcurrencySlotReleaseProcessorTest` grows to 9 (+2): a formerly queued execution that terminates in **error during its actual run** still releases its slot — its history (CREATED → QUEUED → RUNNING → FAILED/CANCELLED) is one step away from the concurrency short-circuit shape (CREATED → FAILED/CANCELLED) that must NOT release; the guard distinguishes them only via `histories[size-2].isCreated()`.

Verified: `:executor:test` **268 tests, 0 failures**; kestra-ee `jdbc-ee:compileJava` + `repository-elasticsearch-ee:compileTestJava` green.


AcevedoR:
Consolidated into #17257 — the three stacked harness PRs are regrouped into one (same commits, identical SHAs). The state-store contracts, outbox seams, and their sagas are all part of #17257 now.

PR #17277, opened 2026-07-07
fix(kv): translate duplicate-key and duration validator errors

brian-mulier-p:
- The KV Store create/edit form's duplicate-key and invalid-duration validators returned raw i18n keys (`kv.duplicate`, `datepicker.error`) instead of translated messages — both now go through `t()`, matching the other validators in the same file.


brian-mulier-p:
Simple call of i18n, no need for review

PR #17298, opened 2026-07-08
feat(plugins): support kv() and secret() in task version field

### LINKED ISSUE #17280: feat(plugins): support kv() (and secret()) in task version field
## Feature request

Allow the plugin `version` field on a task to be set via a KV lookup, e.g.:

```yaml
id: gcs_upload
type: io.kestra.plugin.gcp.gcs.Upload
version: "{{ kv('GCP_PLUGIN_VERSION') }}"
```

Use case: pin different plugin versions per environment (dev/staging/prod) from the same flow YAML, without maintaining separate files per environment.

## Motivation

- Org split: the team operating the Kestra instance is not the team authoring/deploying flows. The `application.yml` workaround below requires instance-team/infra access to change — it doesn't give the flow team self-service control.
- Current pain: plugin `version` is hardcoded per task in each flow today. Bumping a version means editing and redeploying every flow that pins it.
- Per-namespace control: since the KV store is namespace-scoped, `kv()` resolution gives per-namespace version control for free — bumping the KV value for one namespace changes only that namespace's flows, without touching or redeploying flows in other namespaces.

## Why this doesn't work today

`version` (like `type`) is read as a raw string directly off the JSON/YAML node in `PluginDeserializer.extractPluginRawIdentifier` ([core/src/main/java/io/kestra/core/plugins/serdes/PluginDeserializer.java](https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/plugins/serdes/PluginDeserializer.java#L156)) — it's used to pick which plugin class/classloader to deserialize the task into, *before* any Pebble rendering happens. `Task#version` ([core/src/main/java/io/kestra/core/models/tasks/Task.java](https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/models/tasks/Task.java#L44)) is a plain `String`, not `Property<String>`, for the same reason.

No `RunContext`/execution exists at that point — but tenant + namespace *are* known from the flow's own metadata at parse time, and the KV store is namespace-scoped (not execution-scoped), so a `kv()` lookup is technically reachable before class resolution — it's just not implemented.

## Proposed scope

Add a pre-deserialization render pass on the `version` node only (not `type` — out of scope, higher risk), restricted to a minimal Pebble context:
- Allowed: `kv()`, `secret()`, `envs()` (all resolvable from flow tenant/namespace metadata alone).
- Not allowed: `outputs`, `inputs`, task/flow run vars — no execution exists yet.

Runs once at flow parse/save time (flows are parsed once and the `Flow` object is cached/reused across executions, not re-parsed per run). Consequences to design for:
- If the underlying KV value changes later, the resolved `version` stays stale until the flow is re-saved/re-synced. Needs to be documented clearly (not a bug).
- KV store unreachable / tenant not ready at flow-validation time → flow validation must fail with a clear error, not silently fall back to no version.

## Existing workaround (already shipped, no code change needed)

`kestra.plugins.defaults` (`PluginGlobalDefaultConfiguration`) already lets you set a plugin version per type in `application.yml`, with Micronaut env-var placeholders resolved per environment:

```yaml
kestra:
  plugins:
    defaults:
      - type: io.kestra.plugin.gcp.gcs.Upload
        values:
          version: "${GCP_PLUGIN_VERSION:1.2.0}"
```

This solves "different plugin version per environment" today. The marginal value of `kv()` over this is that KV is self-service per-namespace (no admin access / server restart needed to change it), whereas `application.yml` requires infra access.

Related to [Pylon #2025](https://app.usepylon.com/support/issues/views/all-issues?issueNumber=2025&view=fs)

fdelbrayelle:
## Summary

Allows a task's `version` field to be resolved via `kv()` / `secret()` / `env()` at flow parse/save time, e.g.:

```yaml
id: gcs_upload
type: io.kestra.plugin.gcp.gcs.Upload
version: "{{ kv('GCP_PLUGIN_VERSION') }}"
```

This lets teams pin plugin versions per namespace via the (self-service, namespace-scoped) KV store instead of the `kestra.plugins.defaults` `application.yml` workaround, which requires infra/admin access to change.

- `PluginDefaultService` now renders any `version` field containing a Pebble expression on the raw flow map, right after plugin-default injection and before `YamlParser.parse` triggers `PluginDeserializer` classloader resolution — so by the time the plugin identifier (`type:version`) is extracted, `version` is already a concrete literal.
- Only `flow` (tenant/namespace/id) and `envs` are exposed to the renderer, so `kv()`, `secret()`, and `env()` work, but `outputs`/`inputs`/task-run vars are not resolvable (no execution exists yet) — referencing them fails with a clear error instead of silently resolving to nothing, thanks to Pebble's `strictVariables` mode.
- Rendering only runs when a `version` string actually contains a Pebble delimiter (`PebbleUtil.containsOpeningBlockDelimiter`), so plain literal versions (the common case) pay no extra cost.
- A render failure (missing KV key, disallowed variable, etc.) throws `KestraRuntimeException`, so flow save/parse fails loudly rather than silently dropping the version.
- `type` is intentionally untouched, matching the scope called out in the issue.

Since this hooks into `PluginDefaultService.parseFlowWithAllDefaults`, it runs once at flow create/update and once on first execution warm-up (`DefaultFlowMetaStore.findByExecutionThenInjectDefaults`), then the parsed `FlowWithSource` is cached — it is **not** re-evaluated on every execution. If the underlying KV value changes, the flow must be re-saved for the new version to take effect.

Closes #17280

## Test plan

- [x] `PluginDefaultServiceTest#shouldRenderKvExpressionInTaskVersion` — `kv()` in `version` resolves to the stored value on a real task.
- [x] `PluginDefaultServiceTest#shouldFailClearlyWhenVersionExpressionReferencesMissingKv` — missing KV key fails flow parsing with a clear error.
- [x] `PluginDefaultServiceTest#shouldFailClearlyWhenVersionExpressionReferencesExecutionScopedVariable` — referencing `outputs.*` in `version` fails clearly instead of being silently ignored.
- [x] `./gradlew :core:test --tests "io.kestra.core.services.*"` and related suites (`YamlParserTest`, `PluginDeserializerTest`, pebble function tests, `DefaultFlowMetaStoreTest`) pass.
- [x] `./gradlew :jdbc-h2:test --tests "io.kestra.runner.h2.H2RunnerTest"` passes (executor constraint per contributing guidelines).

### DIFF
--- a/core/src/main/java/io/kestra/core/services/PluginDefaultService.java
+++ b/core/src/main/java/io/kestra/core/services/PluginDefaultService.java
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 
 import io.kestra.core.exceptions.FlowProcessingException;
+import io.kestra.core.exceptions.IllegalVariableEvaluationException;
 import io.kestra.core.exceptions.KestraRuntimeException;
 import io.kestra.core.models.Plugin;
 import io.kestra.core.models.executions.Execution;
@@ -33,12 +34,16 @@ import io.kestra.core.models.flows.FlowWithSource;
 import io.kestra.core.models.flows.PluginDefault;
 import io.kestra.core.models.flows.PluginDefaultSpec;
 import io.kestra.core.plugins.PluginRegistry;
+import io.kestra.core.runners.RunContextCache;
 import io.kestra.core.runners.RunContextLogger;
 import io.kestra.core.runners.RunContextLoggerFactory;
+import io.kestra.core.runners.RunVariables;
+import io.kestra.core.runners.VariableRenderer;
 import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.serializers.YamlParser;
 import io.kestra.core.utils.Logs;
 import io.kestra.core.utils.MapUtils;
+import io.kestra.core.utils.PebbleUtil;
 
 import io.micronaut.core.annotation.Nullable;
 import jakarta.annotation.PostConstruct;
@@ -71,14 +76,22 @@ public class PluginDefaultService {
 
     protected final PluginRegistry pluginRegistry;
 
+    private final VariableRenderer variableRenderer;
+
+    private final RunContextCache runContextCache;
+
     @Inject
     public PluginDefaultService(
         @Nullable PluginGlobalDefaultConfiguration pluginGlobalDefault,
         RunContextLoggerFactory runContextLoggerFactory,
-        PluginRegistry pluginRegistry) {
+        PluginRegistry pluginRegistry,
+        VariableRenderer variableRenderer,
+        RunContextCache runContextCache) {
         this.pluginGlobalDefault = pluginGlobalDefault;
         this.runContextLoggerFactory = runContextLoggerFactory;
         this.pluginRegistry = pluginRegistry;
+        this.variableRenderer = variableRenderer;
+        this.runContextCache = runContextCache;
     }
 
     @PostConstruct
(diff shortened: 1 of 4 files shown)


fdelbrayelle:
@paulgrainger85 Since @fhussonnois and Platform are removing a lot of code around this for 2.0 I'll close the issue and my PR and Florian will iterate on the [Pylon](https://app.usepylon.com/support/issues/views/all-issues?issueNumber=2025&view=fs) to add extra details.

PR #17305, opened 2026-07-08
fix(namespaces): make namespace_file_metadata case-sensitive on MySQL

AcevedoR:
### What this does

On **MySQL**, namespace files whose paths differ only by **case** (e.g. `MyFile.sql` vs `myfile.sql`) are treated as the **same file**. This corrupts `namespace_file_metadata` and breaks `SyncNamespaceFiles` / `PurgeFiles` / any `putFile` for the affected namespace. Postgres and H2 are unaffected.

### Root cause

`namespace_file_metadata` is created with the case-insensitive collation `utf8mb4_unicode_ci` (`jdbc-mysql/.../baseline-mysql.sql`). Namespace file paths are semantically case-sensitive — they map 1:1 to storage URIs, which are case-sensitive on virtually every backend. On MySQL this causes:

1. **Primary-key collision** — the PK `key` is the file UID and embeds the path (`tenantId_namespace_path_revision`, see `NamespaceFileMetadata.uid()` / `IdUtils.fromParts`). `..._/MyFile.sql_1` and `..._/myfile.sql_1` collide under a case-insensitive PK → inserting the second fails with `Duplicate entry ... for key PRIMARY`.
2. **Wrong lookups** — `AbstractJdbcNamespaceFileMetadataRepository.findByPath` filters with `field("path").in(...)`; on MySQL it matches a differently-cased row, so `save()`/`putFile()` (`InternalNamespace`) believe the file already exists and take the overwrite/version-bump branch against the wrong record; `fetchOne`(→`fetchAny`) returns the wrong file's metadata.

Postgres/H2 compare case-sensitively by default → **MySQL-only**.

### Fix

Switch the identity/lookup columns of `namespace_file_metadata` — the primary key `key` and the `path` / `parent_path` generated columns — to the case-sensitive `utf8mb4_bin` collation. Purely schema-level; no Java production code change.

- New migration `2.0.15-namespace-file-metadata-case-sensitive-mysql.sql` + bean `V2_0_15NamespaceFileMetadataCaseSensitiveMigration` (idempotent — guarded per-column, safe to re-run; converts existing installs).
- Regression test `AbstractNamespaceFileMetadataRepositoryTest.findByPathIsCaseSensitive` — runs on H2 / Postgres / MySQL.

> Note: the `FULLTEXT(path)` index used for the namespace-file *search* filter becomes case-sensitive as a side effect. Acceptable given file identity correctness is the priority.

### Testing

- Reproduced the collision at the DB level on MySQL 9.7 (duplicate-key on insert + lowercase lookup returning the mixed-case row).
- Verified the migration converts `ci`→`bin`, preserves existing rows, keeps the `FULLTEXT` index, and is idempotent.
- `AbstractNamespaceFileMetadataRepositoryTest` — **20 passing** on each of H2, Postgres, MySQL (including the new test).

### Related

Fixes kestra-io/kestra-ee#9254 (from an internal support ticket).

### DIFF
--- a/core/src/test/java/io/kestra/core/repositories/AbstractNamespaceFileMetadataRepositoryTest.java
+++ b/core/src/test/java/io/kestra/core/repositories/AbstractNamespaceFileMetadataRepositoryTest.java
@@ -60,6 +60,45 @@ public abstract class AbstractNamespaceFileMetadataRepositoryTest {
     }
 
     @Test
+    void findByPathIsCaseSensitive() throws IOException {
+        // Given: two namespace files whose paths differ only by case.
+        // Paths map 1:1 to case-sensitive storage URIs, so they are two distinct files.
+        String tenantId = TestsUtils.randomTenant();
+        String namespace = TestsUtils.randomNamespace();
+
+        NamespaceFileMetadata upperCase = NamespaceFileMetadata.builder()
+            .tenantId(tenantId)
+            .namespace(namespace)
+            .path("/MyFile.sql")
+            .size(1L)
+            .build();
+        NamespaceFileMetadata lowerCase = NamespaceFileMetadata.builder()
+            .tenantId(tenantId)
+            .namespace(namespace)
+            .path("/myfile.sql")
+            .size(1L)
+            .build();
+
+        // When: both are saved
+        namespaceFileMetadataRepositoryInterface.save(upperCase);
+        namespaceFileMetadataRepositoryInterface.save(lowerCase);
+
+        // Then: each is stored as its own revision-1 file and lookups are case-sensitive.
+        // On MySQL with a case-insensitive collation this fails: saving the second one is treated
+        // as a new revision of the first, and findByPath returns the wrong row (see Pylon #2018).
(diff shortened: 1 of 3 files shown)


AcevedoR:
v1.3.x backport: #17310

AcevedoR:
**QA note**

The identical scenario was reproduced end-to-end on real EE (MySQL) for the v1.3.x backport #17310: on the case-insensitive collation, creating `/MyFile.sql` then `/myfile.sql` silently collapses into a single file with wrong content; after switching `key`/`path`/`parent_path` to `utf8mb4_bin` the two files coexist with correct content. See #17310 for the full transcript.

This PR ships the same collation change (as MigrationRunner migration `2.0.15-namespace-file-metadata-case-sensitive`) and is covered by `AbstractNamespaceFileMetadataRepositoryTest.findByPathIsCaseSensitive` — 20 passing on H2, Postgres and MySQL, with the migration auto-applied on startup.

(End-to-end EE run used a 1.3.0-scoped license so it was performed on the 1.3 line; the develop DDL and product code path are identical.)

nkwiatkowski (approved):

PR #17307, opened 2026-07-08
feat: purge-flow-revisions

### LINKED ISSUE #17175: Purge Flows versions
### Feature description

Like the "PurgeFiles" task that purges old versions of namespace files, could we have a "PurgeFlows" task that purges old versions of flows?

hemantmm:
closes: #17175 

---

### ✨ Description

What does this PR change?  

Adds `io.kestra.plugin.core.flow.PurgeFlows`, a core task that purges old flow revisions by namespace, optional flow ID pattern, and version retention behavior.

The task supports:
- `namespaces`
- `namespacePattern`
- `includeChildNamespaces`
- `flowPattern`
- `behavior.type: version`
- `behavior.keepAmount`
- `behavior.before`

By default, the task keeps the latest revision of each flow and purges older revisions.

## Why

Kestra already has `PurgeFiles` for old namespace file versions. This adds equivalent cleanup behavior for flow revisions.

## Testing

```bash
./gradlew :core:test --tests "io.kestra.plugin.core.flow.PurgeFlowsTest"
```

https://github.com/user-attachments/assets/18ffb99c-d65e-4ff5-b454-6fcb8e7aa51c

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/flow/PurgeFlows.java
+++ b/core/src/main/java/io/kestra/plugin/core/flow/PurgeFlows.java
@@ -3,6 +3,7 @@ package io.kestra.plugin.core.flow;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicLong;
 
+import io.kestra.core.exceptions.ValidationErrorException;
 import io.kestra.core.models.annotations.Example;
 import io.kestra.core.models.annotations.Plugin;
 import io.kestra.core.models.flows.FlowWithSource;
@@ -12,7 +13,6 @@ import io.kestra.core.models.tasks.Task;
 import io.kestra.core.repositories.FlowRepositoryInterface;
 import io.kestra.core.runners.DefaultRunContext;
 import io.kestra.core.runners.RunContext;
-import io.kestra.plugin.core.purge.PurgeTask;
 
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.Valid;
@@ -28,7 +28,7 @@ import lombok.experimental.SuperBuilder;
 @Schema(
     title = "Purge old flow revisions.",
     description = """
-        Deletes old flow revisions using a purge `behavior` (default keeps 1 latest revision), optional flow ID glob, and Namespace filters (`namespaces` list or `namespacePattern`). Child Namespaces are included by default.
+        Deletes old flow revisions using a purge `behavior` (default keeps 1 latest revision), with optional namespace prefix and flow ID filters.
 
         The latest revision of each flow is always kept."""
 )
@@ -44,35 +44,25 @@ import lombok.experimental.SuperBuilder;
                 tasks:
                   - id: purge_flows
                     type: io.kestra.plugin.core.flow.PurgeFlows
-                    namespaces:
-                      - company
-                    includeChildNamespaces: true
-                    flowPattern: "*_deprecated"
+                    namespace: company
                     behavior:
-                      type: version
                       keepAmount: 2
                 """
         )
     }
 )
-public class PurgeFlows extends Task implements PurgeTask<FlowWithSource>, RunnableTask<PurgeFlows.Output> {
+public class PurgeFlows extends Task implements RunnableTask<PurgeFlows.Output> {
     @Schema(
-        title = "Flow ID pattern, e.g. '*'",
-        description = "Delete only old revisions for flows whose ID matches the glob pattern."
+        title = "Namespace whose flows need to be purged, or namespace of the flow that needs to be purged",
+        description = "If `flowId` isn't provided, this is a namespace prefix, else the namespace of the flow."
     )
-    private Property<String> flowPattern;
+    private Property<String> namespace;
 
     @Schema(
-        title = "List of namespaces to delete old flow revisions from",
-        description = "If not set, all namespaces will be considered. Can't be used with `namespacePattern` - use one or the other."
+        title = "The flow ID whose old revisions should be purged",
+        description = "You need to provide the `namespace` property if you want to purge a flow."
     )
-    private Property<List<String>> namespaces;
-
-    @Schema(
-        title = "Glob pattern for the namespaces to delete old flow revisions from",
-        description = "If not set, all namespaces will be considered. Example: `company.*`. Can't be used with `namespaces` - use one or the other."
-    )
-    private Property<String> namespacePattern;
+    private Property<String> flowId;
 
     @Schema(
         title = "Purge behavior",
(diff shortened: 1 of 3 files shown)


hemantmm:
It shows that all tests executed successfully, and only the automated PR-comment step failed with a GitHub token permission error.

success ✅ > tests: 7359, success: 7338, skipped: 21, failed: 0 (🔄 7359 executed).

hemantmm:

core/src/main/java/io/kestra/plugin/core/flow/PurgeFlows.java
```diff
@@ -0,0 +1,141 @@
...
+                """
+        )
+    }
+)
+public class PurgeFlows extends Task implements PurgeTask<FlowWithSource>, RunnableTask<PurgeFlows.Output> {
+    @Schema(
+        title = "Flow ID pattern, e.g. '*'",
+        description = "Delete only old revisions for flows whose ID matches the glob pattern."
+    )
+    private Property<String> flowPattern;
```
loicmathieu: I think it should miror `PurgeExecutions` not `PurgeFiles` so I'll only keep a single namespace and flowId

core/src/main/java/io/kestra/plugin/core/flow/PurgeFlows.java
```diff
@@ -0,0 +1,141 @@
...
+    @Valid
+    @NotNull
+    private Property<Version> behavior = Property.ofValue(Version.builder().keepAmount(1).build());
+
+    @Schema(
+        title = "Delete old flow revisions from child namespaces",
+        description = "Defaults to true. This means that if you set `namespaces` to `company`, it will also delete old flow revisions from `company.team`, `company.data`, etc."
+    )
+    @Builder.Default
+    private Property<Boolean> includeChildNamespaces = Property.ofValue(true);
```
loicmathieu: This is not needed, the namespace should be a namespace prefix

core/src/main/java/io/kestra/plugin/core/flow/Version.java
```diff
@@ -0,0 +1,80 @@
...
+import lombok.experimental.SuperBuilder;
+
+@SuperBuilder
+@Getter
+@NoArgsConstructor
+public class Version {
+    @NotNull
+    @JsonInclude
+    @Builder.Default
+    protected String type = "version";
```
loicmathieu: It didn't seems to be used

core/src/main/java/io/kestra/plugin/core/flow/Version.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Schema(
+        title = "How many revisions should be kept for each matching flow.",
+        description = "The latest revision is always kept."
+    )
+    @Min(1)
+    private Integer keepAmount;
+
+    List<FlowWithSource> revisionsToPurge(String tenantId, String namespace, String flowId, FlowRepositoryInterface flowRepository) {
+        List<FlowWithSource> revisions = flowRepository.findRevisions(tenantId, namespace, flowId, false);
```
loicmathieu, line 35: We now have the principle of draft revision, you should keep the last "non-draft" revision and not just the last revision

loicmathieu (changes requested):

core/src/main/java/io/kestra/plugin/core/flow/Version.java
hemantmm: Thanks for pointing this out. I will update the purge logic to keep the latest non-draft revision, instead of only relying on the latest revision.

core/src/main/java/io/kestra/plugin/core/flow/Version.java
hemantmm: Thanks for the review, I will remove the unused `type` field since `PurgeFlows` only supports this version-based behavior for now.

core/src/main/java/io/kestra/plugin/core/flow/PurgeFlows.java
hemantmm: Agreed. I will update the task to mirror `PurgeExecutions` instead of `PurgeFiles`: it now uses a single `namespace` and optional `flowId`, rather than `namespaces`, `namespacePattern`, and `flowPattern`.

core/src/main/java/io/kestra/plugin/core/flow/PurgeFlows.java
hemantmm: Makes sense. I will remove `includeChildNamespaces` and change to the namespace behavior to work as a namespace prefix when `flowId` is not provided.

hemantmm:
This failure looks unrelated to the PurgeFlows changes. The only failed test is `TriggerControllerTest.shouldUnlockTriggerWhenLocked()` in `webserver`, while this PR only changes the core `PurgeFlows` task/tests. The final failure is also from the test-report comment step failing with GitHub 403 permissions.

thomas-depierre:
@loicmathieu, can you check his PR please?

core/src/main/java/io/kestra/plugin/core/flow/PurgeFlows.java
```diff
@@ -0,0 +1,138 @@
...
+                  - id: purge_flows
+                    type: io.kestra.plugin.core.flow.PurgeFlows
+                    namespace: company
+                    behavior:
+                      keepAmount: 2
+                """
+        )
+    }
+)
+public class PurgeFlows extends Task implements RunnableTask<PurgeFlows.Output> {
```
loicmathieu: You should also implement `SystemTask` or it will not work on remote workers via gRPC (new 2.0 feature).

loicmathieu (changes requested):

core/src/main/java/io/kestra/plugin/core/flow/PurgeFlows.java
hemantmm: Thanks! Good catch. I've updated PurgeFlows to implement SystemTask so it works correctly with remote workers over gRPC.

core/src/main/java/io/kestra/plugin/core/flow/PurgeFlows.java
```diff
@@ -0,0 +1,139 @@
...
+    @Override
+    public Output run(RunContext runContext) throws Exception {
+        AtomicLong count = new AtomicLong();
+        Version renderedBehavior = runContext.render(behavior).as(Version.class).orElseThrow();
+        String tenantId = runContext.flowInfo().tenantId();
+        FlowRepositoryInterface flowRepository = ((DefaultRunContext) runContext).services().additionalService(FlowRepositoryInterface.class);
+        String renderedNamespace = runContext.render(this.namespace).as(String.class).orElse(null);
+        String renderedFlowId = runContext.render(this.flowId).as(String.class).orElse(null);
+        if (renderedNamespace == null && renderedFlowId != null) {
+            throw new ValidationErrorException(List.of("Property `namespace` is required when `flowId` is set."));
```
loicmathieu:
```suggestion
            throw new IllegalArgumentException("Property `namespace` is required when `flowId` is set.");
```

We don't use validation exception in runtime code, validation is a build feature so it happens on flow save

loicmathieu (approved):
One last small change and I'll merge it.

core/src/main/java/io/kestra/plugin/core/flow/PurgeFlows.java
hemantmm: Done, updated to use `IllegalArgumentException` instead. Thanks!

hemantmm:
@loicmathieu, I have updated to use `IllegalArgumentException`.

thomas-depierre:
Thank you for your work!

thomas-depierre:
Will it be possible to backport this to version 1.3?

loicmathieu (approved):
LGTM, thanks a lot

loicmathieu:
@thomas-depierre we didn't backport features.
However, if you are a customer, you can ask for a specific backport via the support portal.

PR #17308, opened 2026-07-08
feat(tasks): report per-state terminated iteration counts for Loop task

### LINKED ISSUE #16107: Reports all Loop iteration states and not only terminated
### Feature description

The `Loop` task currently reports only running and terminated sub-executions via its outputs:
```json
{
  "iterationCount":3,
  "runningIterations":0,
  "terminatedIterations":3
}
```

Those are used inside the Loop UI progress bar.

As a sub-executions can ends in different state, and `transmitFailed: false` allows to complete all iterations even is one fail, we should not return terminatedIterations but a per-state terminated count and update the progress bar UI accordingly.

loicmathieu:
## Summary
- The `Loop` task's `terminatedIterations` output was a single aggregate count, but `transmitFailed: false` lets all iterations complete even if some fail — losing which sub-executions succeeded vs. failed. It now reports a per-state breakdown (e.g. `{"SUCCESS": 2, "FAILED": 1}`), accumulated incrementally by the executor as each sub-execution terminates.
- The Loop progress bar in the UI (`TaskRunDetails.vue`) now renders a segmented, per-state-colored bar (via `State.color()`) instead of a single fill, with a tooltip showing the count per state.

Closes #16107

## Test plan
- [x] `LoopExecutionEventMessageHandlerTest` — added a test covering the `transmitFailed: false` mixed-state accumulation path (SUCCESS + FAILED in the same map); all 7 tests pass
- [x] `H2RunnerTest` (loop* tests, 25 scenarios) — migrated assertions to the new map shape; all pass
- [x] `TaskOutputServiceTest` — migrated assertions to the new map shape; all pass
- [x] UI: `npm run check:types` and lint (oxlint + eslint) pass on the changed component
- [x] Translations: `iterations` key added to `en.json` and propagated to all 12 locales; `npm run translations:check` reports no missing/extra keys


loicmathieu:
@elevatebart for the UI part, I created a new component for a segmented progress bar inside the design system. Please have a careful look, you're free to re-implement it differently if you think it's not the good approach.

elevatebart (approved):

ui/src/components/logs/TaskRunLoopProgress.vue
```diff
@@ -0,0 +1,95 @@
...
+    })
+
+    const consolidatedTerminalStates = computed(() => {
+        return loopTerminatedSegments.value.reduce((acc, segment) => {
+            return acc + segment.count
+        }, 0)
+    })
+</script>
+
+<style>
```
elevatebart:
```suggestion
<style lang="scss" scoped>
```

ui/src/components/logs/TaskRunLoopProgress.vue
```diff
@@ -0,0 +1,95 @@
...
+            </KsButton>
+        </div>
+    </div>
+</template>
+
+<script setup lang="ts">
+    import {computed} from 'vue';
+    import {State} from '@kestra-io/design-system';
+    import {RouterLink} from 'vue-router';
+
```
elevatebart:
```suggestion
    import {computed} from "vue";
    import {State} from "@kestra-io/design-system";
    import {RouterLink} from "vue-router";

```

PR #17310, opened 2026-07-08
fix(namespaces): make namespace_file_metadata case-sensitive on MySQL (v1.3.x backport)

AcevedoR:
Backport of #17305 to `releases/v1.3.x`.

### What this does

On **MySQL**, namespace files whose paths differ only by **case** (e.g. `MyFile.sql` vs `myfile.sql`) are treated as the **same file**, corrupting `namespace_file_metadata` and breaking `SyncNamespaceFiles` / `PurgeFiles` / any `putFile` for the affected namespace. Postgres and H2 are unaffected.

### Root cause

`namespace_file_metadata` is created with the case-insensitive collation `utf8mb4_unicode_ci` (`V1_51__ns_files_metadata.sql`). Namespace file paths are case-sensitive — they map 1:1 to storage URIs. On MySQL this causes:

1. **PK collision** — the PK `key` is the file UID and embeds the path (`NamespaceFileMetadata.uid()`); `..._/MyFile.sql_1` and `..._/myfile.sql_1` collide under a case-insensitive PK → `Duplicate entry ... for key PRIMARY`.
2. **Wrong lookups** — `findByPath` (`field("path").in(...)`) matches a differently-cased row, so `save()`/`putFile()` overwrite the wrong record.

### Fix

Switch `key`, `path` and `parent_path` to the case-sensitive `utf8mb4_bin` collation via migration `V1_57__namespace_file_metadata_case_sensitive.sql`. Purely schema-level; no Java production change.

> Note: the `FULLTEXT(path)` index used for the namespace-file *search* filter becomes case-sensitive as a side effect — acceptable given identity correctness is the priority.

### Testing

- Reproduced the collision at the DB level on MySQL 9.7 (duplicate-key on insert + lowercase lookup returning the mixed-case row).
- Verified the migration converts `ci`→`bin`, preserves rows, keeps the `FULLTEXT` index.
- `AbstractNamespaceFileMetadataRepositoryTest` — **20 passing** on H2 and MySQL (including the new `findByPathIsCaseSensitive`).

### Related

Backport of #17305 · fixes kestra-io/kestra-ee#9254.

### DIFF
--- a/core/src/test/java/io/kestra/core/repositories/AbstractNamespaceFileMetadataRepositoryTest.java
+++ b/core/src/test/java/io/kestra/core/repositories/AbstractNamespaceFileMetadataRepositoryTest.java
@@ -60,6 +60,45 @@ public abstract class AbstractNamespaceFileMetadataRepositoryTest {
     }
 
     @Test
+    void findByPathIsCaseSensitive() throws IOException {
+        // Given: two namespace files whose paths differ only by case.
+        // Paths map 1:1 to case-sensitive storage URIs, so they are two distinct files.
+        String tenantId = TestsUtils.randomTenant();
+        String namespace = TestsUtils.randomNamespace();
+
+        NamespaceFileMetadata upperCase = NamespaceFileMetadata.builder()
+            .tenantId(tenantId)
+            .namespace(namespace)
+            .path("/MyFile.sql")
+            .size(1L)
+            .build();
+        NamespaceFileMetadata lowerCase = NamespaceFileMetadata.builder()
+            .tenantId(tenantId)
+            .namespace(namespace)
+            .path("/myfile.sql")
+            .size(1L)
+            .build();
+
+        // When: both are saved
+        namespaceFileMetadataRepositoryInterface.save(upperCase);
+        namespaceFileMetadataRepositoryInterface.save(lowerCase);
+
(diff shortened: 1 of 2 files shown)


AcevedoR:
**QA — reproduced on real EE (v1.3.25, MySQL 9.7) via ee-sandbox**

Ran a published `kestra-ee:v1.3.25` standalone on MySQL (case-insensitive `utf8mb4_unicode_ci` table, i.e. pre-fix) and drove the real namespace-files API.

_Before fix_ — `POST .../files?path=/MyFile.sql` then `POST .../files?path=/myfile.sql` both returned 200, but the second silently corrupted state: DB held only `/MyFile.sql` (v1 + v2), the directory listed **one** file, and `GET /myfile.sql` served `/MyFile.sql`'s content. Matches the customer symptom.

_After applying this migration's DDL_ (collation → `utf8mb4_bin`), repeating the same two calls on a fresh namespace produced **two distinct v1 files**, both listed, each serving its own content (`/MyFile.sql`→`SELECT 1;`, `/myfile.sql`→`SELECT 2;`). No collision.

Also: `AbstractNamespaceFileMetadataRepositoryTest` 20 passing on H2 + MySQL; migration auto-applied by Flyway on startup against a clean DB.

nkwiatkowski (approved):

PR #16366, opened 2026-05-29
fix(core): inject plugin defaults before computing validation warnings

nkwiatkowski:
## Summary
- Inject all plugin defaults (flow-level, namespace-level, global) **before** computing warnings and deprecation paths during flow validation
- Previously, `warnings()` and `deprecationPaths()` ran on the flow before defaults were applied, causing false warnings in the UI for properties provided by plugin defaults
- Added test: `shouldReturnNoWarningsWhenPropertiesProvidedByPluginDefaults`

Fixes #14471

## Test plan
- [x] Existing `FlowServiceTest` tests pass
- [x] New test validates that a flow with required properties provided only via `pluginDefaults` produces no warnings or constraint violations

PR #17315, opened 2026-07-08
fix(flows): stop flagging pluginDefaults-supplied required properties in the editor

### LINKED ISSUE #14471: Flow code editor shows warning for missing properties in `pluginDefaults`
### Describe the issue

https://github.com/user-attachments/assets/d84cf85b-ca7b-47ce-a14e-f2f03ffdc8a0

Can we make the validation smarter so it can recognize when properties are in the `pluginDefaults` and not actually missing? With my limited screen size I thought these properties were entirely missing at first.

### Environment

- Kestra Version: develop

nkwiatkowski:
## What

The flow **code editor** shows yellow "missing property" warnings for required task properties that are actually provided by `pluginDefaults`.

### Why the previous fix didn't work

PR #16366 injected plugin defaults before computing warnings in `FlowService.validate`. But that path can't produce this symptom:
- `FlowService.warnings(...)` only emits subflow / trigger-`dependsOn` / timeout / plaintext-secret warnings — **never** "missing property".
- Those warnings feed the `FlowEditorStats` banner, not the editor gutter.
- The yellow underline is produced **client-side** by `monaco-yaml`, validating the raw buffer against the static, type-only JSON schema from `GET /plugins/schemas/flow`. It has no knowledge of `pluginDefaults`.

### Approach

Post-filter monaco-yaml's markers: drop a `Missing property "x"` warning when an effective plugin default supplies `x` for the enclosing task's `type` (exact/prefix match, mirroring `PluginDefaultService`).

- **Flow-level** defaults are read from the editor buffer (live, no network).
- **Namespace-level + global** defaults come from a new `GET namespaces/{id}/effective-plugindefaults` endpoint, fetched and cached per namespace. A failed/forbidden fetch degrades gracefully — the warning stays.

`monaco-yaml`'s `validate` is all-or-nothing, so disabling client validation isn't an option (it would drop legitimate type/enum/unknown-property diagnostics) — hence the marker post-filter.

## Changes

- `pluginDefaultsMarkers.ts` — pure, unit-tested helper (`filterDefaultsCoveredMarkers`, `isPropertyCoveredByDefaults`).
- `yamlLanguageConfigurator.ts` — global `onDidChangeMarkers` filter gated to `flow-*` models; idempotent (no re-publish loop).
- `useBaseNamespaces.ts` — `loadEffectivePluginDefaults` store action.
- `PluginDefaultService.resolveEffectiveDefaults(...)` public wrapper + `GET namespaces/{id}/effective-plugindefaults` on `NamespaceController` (wrapped DTO).

## Security

The new endpoint is public in OSS (OSS has no per-namespace RBAC). The EE companion PR **overrides it with a `NAMESPACE:VIEW` check** — see kestra-io/kestra-ee#9261.

## Tests

- Vitest: 11 cases for the marker filter.
- `NamespaceControllerTest`: endpoint returns the wrapped shape.
- `check:types` + eslint clean.

Fixes #14471

### DIFF
--- a/ui/src/composables/useBaseNamespaces.ts
+++ b/ui/src/composables/useBaseNamespaces.ts
@@ -187,10 +187,6 @@ export const useBaseNamespacesStore = () => {
         // NOOP IN OSS
     }
 
-    // Effective plugin defaults (namespace-level + global) applicable to a namespace.
-    // Used by the flow editor to avoid flagging required properties supplied by a plugin default.
-    // Returns [] on any non-200 (e.g. 403 when the user lacks namespace read access), so the
-    // editor degrades gracefully to showing the warning.
     async function loadEffectivePluginDefaults({id}: {id: string}): Promise<EffectiveDefault[]> {
         try {
             const response = await axios.get(`${apiUrl()}/namespaces/${id}/effective-plugindefaults`, VALIDATE)
--- a/ui/src/composables/useBaseNamespaces.ts
+++ b/ui/src/composables/useBaseNamespaces.ts
@@ -1,5 +1,6 @@
 import {ref} from "vue"
 import {apiUrl} from "override/utils/route"
+import type {EffectiveDefault} from "./monaco/languages/pluginDefaultsMarkers"
 import * as Utils from "../utils/utils"
 import {useClient, type PagedResultsNamespace} from "@kestra-io/kestra-sdk"
 import * as NamespaceAPI from "@kestra-io/kestra-sdk/namespaces"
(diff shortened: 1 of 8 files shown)


ui/src/composables/useBaseNamespaces.ts
```diff
@@ -186,6 +187,22 @@ export const useBaseNamespacesStore = () => {
         // NOOP IN OSS
     }
 
+    // Effective plugin defaults (namespace-level + global) applicable to a namespace.
+    // Used by the flow editor to avoid flagging required properties supplied by a plugin default.
+    // Returns [] on any non-200 (e.g. 403 when the user lacks namespace read access), so the
+    // editor degrades gracefully to showing the warning.
```
elevatebart: Meh

```suggestion
```

elevatebart (approved):

PR #17318, opened 2026-07-08
feat(system): accept relative durations on every date filter field

ammeek:
## ✨ Description

Removes the special-case `TIME_RANGE` filter and folds its functionality into the existing date fields.

Every date field (`startDate`, `endDate`, `created`, `updated`, `expirationDate`, `nextExecutionDate`, `lastTriggeredDate`) now accept:

* an absolute timestamp
* a relative ISO-8601 duration (for example `P7D` or `PT24H`).

### Backend

Relative durations are resolved to absolute instants in a single place: the `@QueryFilterFormat` binder. From that point on, controllers and repositories only deal with absolute dates.

Whether a duration is interpreted as "now minus" or "now plus" is determined by the field via `QueryFilter.Field#dateOrientation()`:

* **Past-oriented fields** (for example `startDate`, `created`, `updated`) → `now - duration`
* **Future-oriented fields** (for example `nextExecutionDate`, `expirationDate`) → `now + duration`

As part of this change, the dedicated `TIME_RANGE` filter field has been removed.

### Frontend

The single `timeRange` filter chip has been replaced with per-field date filters:

* Executions, Logs and Dashboards: `startDate` / `endDate`
* Triggers: `nextExecutionDate` / `lastTriggeredDate`

Each date filter supports:

* Relative presets (for example "Last 7 days")
* Absolute date selection
* Comparator selection

### Why?

This removes one of the last special cases in the query filter API.

Instead of having a bespoke `timeRange` concept alongside normal date filters, every date field is now filtered in exactly the same way using the standard query filter operators. The API becomes more consistent, easier to understand, and simpler for SDKs, MCP servers, and AI tooling to consume, since they only need to reason about real fields and standard operators.

### ⚠️ Breaking change

The legacy `TIME_RANGE` filter has been removed.

Before:

```text
filters[timeRange][EQUALS]=P7D
```

After:

```text
filters[startDate][GREATER_THAN_OR_EQUAL_TO]=P7D
```

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra-ee/issues/9263


AcevedoR (commented):
the idea seems cool, I would love to see your point of view on this @brian-mulier-p the timelord

brian-mulier-p:
As I said to @AcevedoR, I'll have a proper look next week especially around the start & endDate filters handling as I know there was a lot of confusion regarding the fact that it was applying a range to startDate but the naming vs what it acts on was confusing. I've fixed that by adding a three-option tab in the date selector so you can define on which the range applies, idk if your current code went for startDate applies to actual startDate field (so you provide a range to startDate, it takes the field startDate as-is and checks if within range) or there is still the same issue).
If you took care of it you may remove the parameter I added a while ago (dateFilter I think?)

brian-mulier-p:
After first-step review, I'm not sure to understand multiple things:
- as I said above, dateFilter should be removed if we are on a true per-field filtering, because it's confusing to be able to give a timerange to both startDate AND endDate while the current implementation of your PR didn't change the target of those fields, they are still dependant on dateFilter (START_DATE, END_DATE OR START_OR_END_DATE). However multiple ranges targetting in the end the same field doesn't make sense so we should truly redirect those filtering to their own fields each time (so if I input a range in startDate, it means startDate is IN that range, same goes on with endDate, no magical endDate range apply to startDate anymore).
- If we implement the above, I fear our current filter system doesn't allow having a START_OR_END_DATE behavior which was mandatory for one of our user: he wants to see executions which has been active (so either started or terminated) in a given range. However we don't have a way to specify an OR filter (because if we go the full path, it would mean we put two filters with an OR in-between: startDate in RANGE_A OR endDate in RANGE_A)
- I'm not sure to understand why Logs would have both startDate / endDate anymore, I would expect a BC with only `date` filter because logs don't have a concept of start / end but only a timestamp at which it got emitted so a `date` filter with a range should cover every usecases?

brian-mulier-p:
After more thoughts I confused myself between timeRange (duration-based range from now) and a date-to-date filtering so two fields per-entity seems needed you're right. However what's confusing is to have timerange for both startDate and endDate at once for entities that don't map these to their actual fields. Unless I didn't see it, it seems there's no restriction for that and I find that case confusing:
What is `filters[startDate][GREATER_THAN_OR_EQUAL_TO]=P7D&filters[endDate][GREATER_THAN_OR_EQUAL_TO]=P6D` doing? When it applies to their respective fields I would understand but on executions both map depending on dateFilter currently so it seems countering each other?

ammeek:
Thank you for the feedback, @brian-mulier-p. I've taken a second look and applied the relevant changes based on your suggestions.

The two main changes I've made are:

* Removing the `dateFilter` query param. As you noted, "dateFilter should be removed if we are on a true per-field filtering."
* Removing `startDate` and `endDate` from several Resources where they no longer make sense.

The `TIME_RANGE` field can now be completely removed, as it is no longer used in any meaningful way. I've decided to tackle this in a later PR, as the AI Copilot will need several updates.

I've also made a few other small updates, mostly removing redundant logic.

brian-mulier-p:
It looks better but still no UI way to filter executions "active in range" (started OR ended within window), old dateFilter=START_OR_END_DATE gone and it was originally done per a user ask so we need to have a way to do that. Maybe add a UI way to have an OR condition since it's already implemented on the backend side?

PR #17329, opened 2026-07-09
fix(logs): widen logs/metrics task_id column to 256 to match Task.id max size

### LINKED ISSUE #17328: Over-long taskId (>150 chars) crash-loops JDBC instances (value too long for VARCHAR(150) logs/metrics task_id)
## What happens

On a JDBC backend, an over-long `taskId` on a `LogEntry` (or `MetricEntry`) crash-loops the whole instance.

The JDBC indexer persists the entry as JSON into `logs.value`; every dialect defines a generated projection column:

```sql
task_id VARCHAR(150) GENERATED ALWAYS AS (value ->> 'taskId') STORED
```

When `taskId` exceeds 150 characters, evaluating that generated column fails:

```
org.jooq.exception.DataException: ... ERROR: value too long for type character varying(150)
```

Because the persist runs **before** the queue ack (`JdbcQueue.receiveImpl`, `inTransaction = false`), the SQL error propagates to the fail-fast handler in the poll loop (`JdbcQueue.java`, the generic `catch (Exception)` → `KestraContext.getContext().shutdown()`). The message is never acked, so on restart it is redelivered and crashes the instance again → **crash loop**.

```
ERROR jdbc-queue-LogEntry_0 io.kestra.jdbc.runner.JdbcQueue Fatal error while polling the
'io.kestra.core.models.executions.LogEntry' queue. Initiating shutdown.
```

## Root cause

Two independent facts:

1. **Model-vs-schema mismatch.** `Task.id` is validated `@Size(max = 256)` (`Task.java`), but the generated `task_id` columns are only `VARCHAR(150)`. Even a legitimate, pattern-valid task id of 151–256 chars overflows.
2. **Unvalidated dynamic task ids.** A `taskId` can also be set to an arbitrary value via the dynamic sub-taskrun path (`RunContext.dynamicWorkerResult(...)` → `RunContextLogger.deriveLoggerFor` → `LogEntry.taskId(dynamicTaskRun.getTaskId())`). `TaskRun.taskId` / `LogEntry.taskId` carry no pattern/size validation. Plugins that surface per-item logs under synthetic sub-taskruns (e.g. the Ansible plugin emitting `<host> | <play> : <task name>`) can therefore produce a `taskId` that never went through flow-definition validation.

Only `logs` and `metrics` promote `taskId` to a length-limited column (the `flows` table promotes only flow-level `id`/`namespace`; `executions` never extracts `taskId`), which is why only those two tables overflow.

## Affected tables

| Table | Column | Width (before) |
|-------|--------|----------------|
| `logs` | `task_id` (generated) | `VARCHAR(150)` |
| `metrics` | `task_id` (generated, PG `NOT NULL`) | `VARCHAR(150)` |

(EE `asset_usages` / `asset_lineage_events` already use `VARCHAR(256)` on Postgres/MySQL; only their H2 variants lagged at 150.)

## Fix

Widen the generated `task_id` columns from `VARCHAR(150)` → `VARCHAR(256)` (matching `Task.id`'s `@Size(max = 256)`) via new Flyway migrations, in all three dialects, using the in-place `ALTER`/`MODIFY` pattern (no index rebuild, preserves the generated projection design). No Java change; no data loss (the full taskId lives in the JSON `value`).

- **Postgres** — catalog-only change (no table rewrite), effectively instant.
- **H2** — instant.
- **MySQL** — modifying a `STORED GENERATED` column rebuilds the table (`ALGORITHM=COPY`); there is no online (`LOCK=NONE`) path for a stored generated column. Fine on small tables; large MySQL `logs` tables may see a blocking rebuild during the startup migration (or run it via gh-ost / pt-online-schema-change beforehand).

## Known residual (follow-up)

Because `taskId` arrives unvalidated on the dynamic-taskrun path, a value **>256** would still overflow. Widening to 256 fixes every legitimate task id and the reported case. Fully closing the class of issue would require bounding/validating `TaskRun.taskId` / `LogEntry.taskId` on the dynamic-taskrun path — tracked as a separate follow-up.

## Reproduction

Persist a `LogEntry` (or `MetricEntry`) whose `taskId` is longer than 150 characters on a JDBC backend; the indexer poll thread logs the fatal error and the server shuts down.

## Fix PRs

- [ ] OSS 1.3.x: kestra-io/kestra#17329
- [ ] EE 1.3.x: kestra-io/kestra-ee#9274
- [ ] OSS v2 (develop): kestra-io/kestra#17331
- [ ] EE v2 (develop): kestra-io/kestra-ee#9277

nkwiatkowski:
## What

Widen the generated `task_id` column from `VARCHAR(150)` → `VARCHAR(256)` on the `logs` and `metrics` tables (Postgres, MySQL, H2), to match `Task.id`'s `@Size(max = 256)`.

## Why

Closes #17328. A `taskId` longer than 150 chars (e.g. a plugin-synthesized Ansible label `<host> | <play> : <task name>`) overflows the `VARCHAR(150)` generated column. The JDBC indexer treats the SQL error as fatal and calls `KestraContext.shutdown()`; the un-acked message is redelivered on restart, crash-looping the instance.

## How

- 3 Flyway migrations (`V1_57` pg/mysql, `V1_56` h2) using the in-place `ALTER`/`MODIFY` pattern — preserves indexes and the generated-projection design. No Java change, no data loss (the full taskId lives in the JSON `value`).
- Regression tests in the shared abstract bases (`AbstractLogRepositoryTest`, `AbstractMetricRepositoryTest`) persisting a 200-char taskId; confirmed red before the migration, green after.

### Migration cost
- **Postgres / H2**: metadata-only / instant.
- **MySQL**: modifying a `STORED GENERATED` column rebuilds the table (`ALGORITHM=COPY`, no online path). Large MySQL `logs` tables may see a blocking rebuild at startup — can be run via gh-ost / pt-osc beforehand.

## Companion PR

EE H2 asset-table parity (`asset_usages`, `asset_lineage_events` → 256): kestra-io/kestra-ee (linked below once opened).

## Residual (follow-up)

`taskId` is unvalidated on the dynamic-taskrun path, so a value >256 would still overflow — a separate follow-up to bound/validate `TaskRun.taskId` / `LogEntry.taskId`.

### DIFF
--- a/core/src/test/java/io/kestra/core/repositories/AbstractLogRepositoryTest.java
+++ b/core/src/test/java/io/kestra/core/repositories/AbstractLogRepositoryTest.java
@@ -417,6 +417,21 @@ public abstract class AbstractLogRepositoryTest {
     }
 
     @Test
+    void shouldPersistTaskIdLongerThan150Chars() {
+        // A plugin-generated taskId (e.g. Ansible "<host> | <play> : <task>") can exceed the legacy
+        // VARCHAR(150) task_id column and crash-loop the indexer; the column now allows up to 256.
+        String tenant = TestsUtils.randomTenant(this.getClass().getSimpleName());
+        String longTaskId = "a".repeat(200);
+        LogEntry log = logEntry(tenant, Level.INFO, "execLongTaskId").taskId(longTaskId).build();
+
+        LogEntry saved = logRepository.save(log);
+
+        List<LogEntry> found = logRepository.findByExecutionIdAndTaskId(tenant, saved.getExecutionId(), longTaskId, null);
+        assertThat(found).hasSize(1);
+        assertThat(found.getFirst().getTaskId()).isEqualTo(longTaskId);
+    }
+
+    @Test
     void purge() {
         String tenant = TestsUtils.randomTenant(this.getClass().getSimpleName());
         logRepository.save(logEntry(tenant, Level.INFO, "execution1").build());
--- a/core/src/test/java/io/kestra/core/repositories/AbstractMetricRepositoryTest.java
+++ b/core/src/test/java/io/kestra/core/repositories/AbstractMetricRepositoryTest.java
@@ -190,6 +190,22 @@ public abstract class AbstractMetricRepositoryTest {
         assertThat(results).isEqualTo(1.0);
     }
 
+    @Test
+    void shouldPersistTaskIdLongerThan150Chars() {
+        // A plugin-generated taskId (e.g. Ansible "<host> | <play> : <task>") can exceed the legacy
+        // VARCHAR(150) task_id column and crash-loop the indexer; the column now allows up to 256.
+        String tenant = TestsUtils.randomTenant(this.getClass().getSimpleName());
+        String executionId = FriendlyId.createFriendlyId();
+        String longTaskId = "a".repeat(200);
+        MetricEntry counter = MetricEntry.of(taskRun(tenant, executionId, longTaskId), counter("counter"), null);
+
+        metricRepository.save(counter);
+
+        List<MetricEntry> results = metricRepository.findByExecutionIdAndTaskId(tenant, executionId, longTaskId, Pageable.from(1, 10));
+        assertThat(results.size()).isEqualTo(1);
+        assertThat(results.getFirst().getTaskId()).isEqualTo(longTaskId);
+    }
+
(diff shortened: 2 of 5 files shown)


nkwiatkowski:
Companion EE PR (H2 asset-table parity): kestra-io/kestra-ee#9274

loicmathieu (approved):
LGTM

PR #17330, opened 2026-07-09
refactor(executions): use constructor injection in ExecutorService and the executor message handlers

AcevedoR:


loicmathieu (approved):
LGTM

PR #17331, opened 2026-07-09
fix(logs): widen logs/metrics task_id to 256 (v2 / MigrationRunner)

nkwiatkowski:
## What

Widen the generated `task_id` column from `VARCHAR(150)` → `VARCHAR(256)` on the `logs` and `metrics` tables for Kestra v2, via `MigrationRunner` increments.

## Why

The v2 counterpart of #17329. Root issue #17328: a `taskId` longer than 150 chars (e.g. a plugin-synthesized Ansible label `<host> | <play> : <task>`) overflows the generated `VARCHAR(150)` `task_id` column; the JDBC indexer treats the SQL error as fatal and shuts the server down, crash-looping the instance.

## How

Two increments (both `task_id`→256, guarded on MySQL):
- **`2.0.16-widen-task-id`** (repository-type gated, primary datasource): widens `metrics` and the main-DB `logs` table — covers the default/co-located log store (Kestra enforces a co-located store matches the repository dialect).
- **`2.0.16-widen-logs-task-id-<dialect>`** (`kestra.logs.type` gated, `LogJdbcDataSourceProvider`): widens a **dedicated** external log DB. Mirrors `V2_0_14LogsMigration`; its **scriptId encodes the datasource** (`-<dialect>[-<table>]`) so switching the log-repo backend re-applies it.

**MySQL guard:** modifying a `STORED GENERATED` column rebuilds the table (`ALGORITHM=COPY`). Each ALTER is guarded by an `information_schema.CHARACTER_MAXIMUM_LENGTH < 256` check so it skips the rebuild for installs already at 256 (e.g. upgraded from Kestra 1.3.x, which widened it via Flyway). Postgres/H2 are metadata-only.

## Tests

TDD regression tests added to `AbstractMetricRepositoryTest` and `AbstractLogDataStoreTest` (persist a 200-char taskId); red before the migrations, green after. Full H2 log + metric suites pass.

## Companion / residual

- EE companion (H2 asset-table parity): kestra-io/kestra-ee (linked in a comment).
- Residual: a pure back-compat install with **no `kestra.logs.type`** keeps logs in the baseline main-DB `logs` table — covered by the primary-DS migration above. A `taskId` >256 is still out of range (unvalidated dynamic-taskrun path) — separate follow-up.

### DIFF
--- a/core/src/test/java/io/kestra/core/repositories/AbstractMetricRepositoryTest.java
+++ b/core/src/test/java/io/kestra/core/repositories/AbstractMetricRepositoryTest.java
@@ -198,6 +198,22 @@ public abstract class AbstractMetricRepositoryTest {
         assertThat(results).isEqualTo(1.0);
     }
 
+    @Test
+    void shouldPersistTaskIdLongerThan150Chars() {
+        // A plugin-generated taskId (e.g. Ansible "<host> | <play> : <task>") can exceed the legacy
+        // VARCHAR(150) task_id column and crash-loop the indexer; the column now allows up to 256.
+        String tenant = TestsUtils.randomTenant(this.getClass().getSimpleName());
+        String executionId = FriendlyId.createFriendlyId();
+        String longTaskId = "a".repeat(200);
+        MetricEntry counter = MetricEntry.of(taskRun(tenant, executionId, longTaskId), counter("counter"), null);
+
+        metricRepository.save(counter);
+
+        List<MetricEntry> results = metricRepository.findByExecutionIdAndTaskId(tenant, executionId, longTaskId, Pageable.from(1, 10));
+        assertThat(results.size()).isEqualTo(1);
+        assertThat(results.getFirst().getTaskId()).isEqualTo(longTaskId);
+    }
+
     private Counter counter(String metricName) {
         return Counter.of(metricName, 1);
     }
--- a/jdbc-h2/src/main/java/io/kestra/repository/h2/migration/V2_0_16WidenLogsTaskIdMigration.java
+++ b/jdbc-h2/src/main/java/io/kestra/repository/h2/migration/V2_0_16WidenLogsTaskIdMigration.java
@@ -1,0 +1,76 @@
+package io.kestra.repository.h2.migration;
+
+import java.util.List;
+import java.util.Map;
+
+import javax.sql.DataSource;
+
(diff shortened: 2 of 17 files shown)


nkwiatkowski:
Companion EE v2 PR (H2 asset-table parity): kestra-io/kestra-ee#9277

nkwiatkowski:
Reworked per review (force-pushed):

- **Split the migrations cleanly.** The metrics widen (`2.0.16-widen-metrics-task-id`) now runs on the standard backend / default datasource and **only touches `metrics`** — the `logs` ALTER is removed from it. The log table is widened solely by `2.0.16-widen-logs-task-id-<dialect>` via `LogJdbcDataSourceProvider`. No table is altered twice (this removes the redundant double-run, which was a costly rebuild on MySQL).
- **The logs widen now always applies.** Instead of gating on `kestra.logs.type` being set (which missed the default case), it's gated on the **effective** log dialect `kestra.logs.type ?? kestra.repository.type` (new `LogStoreTypeResolver` + a per-dialect `Condition`, modelled on `DatasourceProvider.H2RepositoryOrQueue`). So it runs whether a log repo is configured or logs use the current backend.
- **scriptId encodes the datasource.** `2.0.16-widen-logs-task-id-<dialect>` now also appends `-dedicated` (and a non-default table name), so switching the log-repo backend re-applies it — and it fixes a real cross-datasource scriptId collision (co-located primary vs dedicated log DB sharing one migration history).
- **MySQL size guard retained** on both scripts (`information_schema.CHARACTER_MAXIMUM_LENGTH < 256`) — skips the rebuild when already 256 (customers from 1.3.x).

Full `jdbc-h2` suite green (the one failure was `H2RunnerTest.flowTriggerAnyMode`, a known-flaky flow-trigger test, passes in isolation).

nkwiatkowski:
Follow-up (force-pushed): removed the `-dedicated` token from the logs `scriptId` — the scriptId is now just `2.0.16-widen-logs-task-id-<dialect>` (+ table name when non-default), matching `V2_0_14LogsMigration`.

The scriptId collision that `-dedicated` had masked is instead fixed at the test level: `H2LogDataStoreDedicatedTest` now uses an isolated primary datasource so its log-migration history no longer leaks into the shared in-memory `public` DB used by the other repository tests. Full `jdbc-h2` suite green (965 tests).

loicmathieu (approved):
LGTM

PR #17333, opened 2026-07-09
refactor(design-system): move TaskIcon out of the design system into the app

elevatebart:
### ✨ Description

`TaskIcon` and `PluginCard` depend on the app's plugin-icon API, so they don't belong in the design system. Moved both to `ui/src/components/plugins/` and dropped the `Ks` prefix. `KsEditor` and the topology package now resolve the icon via a `TASK_ICON_INJECTION_KEY` provide/inject (app provides `TaskIcon` at bootstrap) instead of importing it.

### 🎨 Frontend Checklist

- [ ] `npm run build` / `npm run test:e2e` — not run here
- No visual change (relocation/rename only). `vue-tsc`, `vitest` (TaskIcon/PluginCard/KsEditor/topology), and `oxlint` all pass locally.

### 📝 Additional Notes

Companion PR: kestra-io/kestra-ee#9278.

### 🤖 AI Authors

Why did the cat leave the design system? It wanted to paw-sition itself in the app. 🐱


ui/packages/design-system/src/composables/taskIcon.ts
```diff
@@ -0,0 +1,45 @@
...
+const FallbackTaskIcon = defineComponent({
+    name: "FallbackTaskIcon",
+    inheritAttrs: false,
+    props: {
+        cls: {type: String, default: undefined},
+        customIcon: {type: Object as PropType<TaskIconProps["customIcon"]>, default: undefined},
+        icons: {type: Object as PropType<TaskIconProps["icons"]>, default: undefined},
+        onlyIcon: {type: Boolean, default: false},
+        variable: {type: String, default: undefined},
+        loadIcon: {type: Function as PropType<TaskIconProps["loadIcon"]>, default: undefined},
```
elevatebart: This props definition is there twice. Could we merge the two or induce the interface from the object?

flcarre (approved):

PR #17335, opened 2026-07-09
[Filters] Fix save filter dialog to display condition groups

### LINKED ISSUE #16434: [Filters] Save filter dialog does not display condition groups
### Describe the issue

The filter save dialog currently does not support condition groups. The displayed filter preview might confuse users.

Image

### Environment

- Kestra Version: develop

snehal242005:
Closes #16434

## What does this PR do?
The "Save filter" dialog's condition preview previously flattened all filter 
conditions into a single list, losing any AND/OR group structure. This meant 
users couldn't verify how their grouped conditions were actually organized 
before saving.

This PR:
- Fixes `SaveFilters.vue` to properly render nested condition groups 
  (bordered group boxes with AND/OR operators between them), instead of a 
  flat list. Falls back to the original flat rendering for simple, 
  ungrouped filters.
- Passes the actual `groups`/`topLogical` data down from `RightFilter.vue` 
  instead of the pre-flattened list.
- Persists `groups`/`topLogical` when saving a filter (`useSavedFilters.ts`), 
  so reloading a saved filter with grouped conditions correctly restores 
  the nesting instead of losing it. Legacy saved filters (flat list only) 
  still load correctly via a fallback.

## How to test?
1. Open the Executions page and build a filter with multiple condition 
   groups (e.g., group A: Status=FAILED AND Namespace=prod; group B: 
   Status=WARNING), joined by OR
2. Click "Save filter" 
3. The preview should show each group in a bordered box with "AND" between 
   conditions inside a group, and "OR" between the groups — not a flat list
4. Save the filter, reload the page, and re-open it — the grouping should 
   be preserved

## Mandatory checks
- [x] Tests created for changes (added/updated 4 test files)
- [x] Manually tested changed features in running Kestra
- [x] Checked documentation: no user-facing docs need updates for this fix


flcarre (approved):
Solid work, and I like the round-trip tests. One narrow thing worth a look: `loadSavedFilter` now goes through `replaceTree`, which updates the tree but not the dismissed-keys set, so after loading a saved filter that omits a `visibleByDefault` key, the "Reset" affordance (`canReset`) can drift out of sync. Reconciling dismissed keys inside `loadSavedFilter` (or folding it into `replaceTree` so every caller benefits) would close that. The dead `logical: "AND"` on single-leaf boxes and a Storybook story for the new grouped preview are minor follow-ups. Nothing blocking from me.

PR #17339, opened 2026-07-09
feat(executions): namespace and tenant concurrency limits — executor engine (kestra-ee#8660)

AcevedoR:
## What

The executor-side engine for kestra-io/kestra-ee#8660 — **concurrency limits at the namespace or tenant level** — extracted to the minimal production changes. The test evidence lives on the test-harness PR #17257 (12 lifecycle sagas incl. the issue's worked example, plus state-store contract scenarios run on H2).

Spec implemented:

- Evaluation order **flow → namespace → parent namespaces (innermost first) → tenant**; the **first limit reached defines the behavior** (QUEUE / CANCEL / FAIL).
- An execution that runs claims **one slot in every applicable scope**; executions of flows without their own `concurrency` property count toward namespace/tenant limits.
- On termination every scope is released and the oldest queued execution that fits **all of its own scopes** is popped — including executions of other flows sharing a freed scope; candidates blocked by one of their own scopes are skipped (no head-of-line starvation) and reconsidered when their scope frees.

## How

Following the review, **all multi-scope store implementations live in EE** (kestra-ee#9286: the JDBC engine in `jdbc-ee` and the Elasticsearch one) — OSS only carries the seam, and the `jdbc` module is untouched.

1. **`refactor`: `ConcurrencySlotReleaseProcessor`.** The slot-release/pop block moves verbatim out of `DefaultExecutor#toExecution` into a processor with no queue dependency: it returns the popped execution and `DefaultExecutor` emits it **after** `decrementAndPop()` has committed — previously the emission happened inside the state-store transaction (same transactional-outbox rule as the delay loop).
2. **`feat`: the scoped-limit seam.**
   - `ScopedConcurrencyLimit` (core): a `Concurrency` plus the scope it applies to (FLOW / NAMESPACE / TENANT); counters are keyed by its uid (`tenant|namespace|flowId`, blank parts for broader scopes — the FLOW uid matches today's `ConcurrencyLimit` key).
   - `ConcurrencyLimitResolver` (core): resolves the ordered limit list for a flow. **OSS returns the flow's own `concurrency` only** — namespace/tenant limit definitions are the EE part (`@Replaces` resolver, kestra-ee#9286).
   - `ConcurrencyLimitStateStore` gains two scoped operations as **default methods** — multi-counter `countThenProcess` (read all, decide, claim-all atomically) and `releaseThenPop` (decrement all, widest-scope FIFO pop with per-candidate fit check). The defaults delegate to the existing flow-keyed methods for the single-flow-scope case, so every backend keeps working unchanged; multi-scope throws `UnsupportedOperationException` — unreachable in OSS, and EE overrides it.
   - The `ExecutionEventMessageHandler` gate and the release processor go through the resolver + scoped store API (single code path).
3. **`fix`: release the scopes the execution was *admitted* under, not the currently defined ones.** Without this, removing (or redefining) a namespace/tenant limit while an admitted execution runs leaks its counter forever (e.g. a tenant counter stuck at 1, blocking the tenant until a manual reset): the release re-resolved the limits at termination and found nothing to decrement. The gate now stamps the claimed scope uids on `ExecutionMetadata#concurrencyScopes` when the execution claims its slots (popped queued executions are stamped the same way), and the release decrements exactly the stamped scopes — falling back to the current definitions for unstamped executions. Reproduced first in `shouldReleaseTheScopesTheExecutionWasAdmittedUnderWhenTheLimitWasRemovedMidRun` (red: counter stuck at 1 → green).
4. **Review follow-ups**: the multi-scope JDBC implementation moved out of the `jdbc` module into kestra-ee (r3585688379), and the now-unused flow-scoped `processExecutionRunning` overload was removed (r3585696676).

Net effect in OSS today: behavior is exactly the current flow-level semantics — the resolver only ever returns the flow scope, and the scoped default methods delegate to the same store operations as before. The multi-scope path only activates once a resolver provides namespace/tenant limits.

## Testing

- `:executor:test` and `H2RunnerTest` green (AGENTS.md executor rule)
- `AbstractScopedConcurrencyLimitStateStoreTest` (new, in `executor` test output): 7 contract scenarios — claim-all accounting, shared namespace counters across flows, cross-flow pop, blocked-candidate skip + reconsideration, scope isolation, and the admitted-scope release after a mid-run limit removal. Runs in EE against the JDBC store (`jdbc-ee-h2`, 7/7) and Elasticsearch (kestra-ee#9286).
- `ExecutionEventMessageHandlerTest`: +2 — the gate stamps the claimed scopes on admission, and stamps nothing when no limit applies
- kestra-ee `jdbc-ee:compileJava` + `repository-elasticsearch-ee:compileTestJava` compile clean
- Unit coverage (multi-scope sagas + store contracts on the testkit fake) is on #17257, the test-harness PR

## Next (EE, separate PRs)

The kestra-ee companion PR (same branch name, kestra-ee#9286): `concurrency` on the Namespace/Tenant models, the `@Replaces` resolver walking the namespace tree, and the JDBC + Elasticsearch multi-scope stores. Then the Limits-tab UI and runner-level tests.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/executions/ExecutionMetadata.java
+++ b/core/src/main/java/io/kestra/core/models/executions/ExecutionMetadata.java
@@ -1,6 +1,7 @@
 package io.kestra.core.models.executions;
 
 import java.time.Instant;
+import java.util.List;
 
 import jakarta.validation.constraints.NotNull;
 import lombok.Builder;
@@ -18,6 +19,15 @@ public class ExecutionMetadata {
 
     @NotNull
     Instant originalCreatedDate;
+
+    /**
+     * The uids of the concurrency scopes this execution claimed a slot in when it was admitted.
+     * The release decrements exactly these scopes, so removing or changing a namespace/tenant
+     * limit while the execution runs cannot leak the counter of a scope it was admitted under.
+     * Null when the execution never claimed a slot (or predates the scoped limits).
+     */
+    @With
+    List<String> concurrencyScopes;
 
     public ExecutionMetadata nextAttempt() {
         return this.toBuilder()
--- a/core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
+++ b/core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
@@ -69,22 +69,6 @@ public record ScopedConcurrencyLimit(
         return new ScopedConcurrencyLimit(scope, tenantId, namespace, flowId, concurrency);
     }
 
-    /**
-     * Whether the given execution coordinates fall inside this scope: the flow itself for a
-     * flow scope, the namespace or any descendant for a namespace scope, the whole tenant for
-     * a tenant scope.
-     */
-    public boolean covers(String tenantId, String namespace, String flowId) {
-        if (!Objects.equals(this.tenantId, tenantId)) {
-            return false;
-        }
-        return switch (this.scope) {
-            case FLOW -> Objects.equals(this.namespace, namespace) && Objects.equals(this.flowId, flowId);
-            case NAMESPACE -> Objects.equals(this.namespace, namespace) || (namespace != null && namespace.startsWith(this.namespace + "."));
-            case TENANT -> true;
-        };
-    }
-
     /** {@inheritDoc} */
     @Override
     public String uid() {
--- a/core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
+++ b/core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
@@ -21,14 +21,15 @@ import jakarta.validation.constraints.NotNull;
  * @param tenantId the tenant, never null
  * @param namespace the namespace the limit applies to, null for a tenant-scoped limit
  * @param flowId the flow the limit applies to, null unless the scope is {@link Scope#FLOW}
- * @param concurrency the limit and its behavior
+ * @param concurrency the limit and its behavior — null only for release-time scopes whose
+ *        definition was removed while the admitted execution ran (see {@link #fromUid})
  */
 public record ScopedConcurrencyLimit(
     @NotNull Scope scope,
     @NotNull String tenantId,
     @Nullable String namespace,
     @Nullable String flowId,
-    @NotNull Concurrency concurrency) implements HasUID {
+    @Nullable Concurrency concurrency) implements HasUID {
 
     public enum Scope {
         FLOW,
@@ -46,6 +47,26 @@ public record ScopedConcurrencyLimit(
 
     public static ScopedConcurrencyLimit ofTenant(String tenantId, Concurrency concurrency) {
         return new ScopedConcurrencyLimit(Scope.TENANT, tenantId, null, null, concurrency);
+    }
+
+    /**
+     * Rebuild a scope from its uid — the form persisted on an execution's metadata when it
+     * claims its slots. Tenant, namespace and flow ids cannot contain {@code |}, so the parts
+     * split back unambiguously; blank parts mark the broader scopes.
+     *
+     * @param concurrency the currently defined limit of that scope, or null when its
+     *        definition was removed since the execution was admitted
+     */
+    public static ScopedConcurrencyLimit fromUid(String uid, @Nullable Concurrency concurrency) {
+        String[] parts = uid.split("\\|", -1);
+        if (parts.length != 3) {
+            throw new IllegalArgumentException("Invalid concurrency scope uid: " + uid);
+        }
+        String tenantId = parts[0];
+        String namespace = parts[1].isEmpty() ? null : parts[1];
+        String flowId = parts[2].isEmpty() ? null : parts[2];
+        Scope scope = flowId != null ? Scope.FLOW : namespace != null ? Scope.NAMESPACE : Scope.TENANT;
+        return new ScopedConcurrencyLimit(scope, tenantId, namespace, flowId, concurrency);
     }
 
     /**
--- a/core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
+++ b/core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
@@ -1,0 +1,72 @@
+package io.kestra.core.runners;
+
+import java.util.Objects;
+
+import io.kestra.core.models.HasUID;
+import io.kestra.core.models.flows.Concurrency;
+import io.kestra.core.models.flows.FlowInterface;
+
+import jakarta.annotation.Nullable;
+import jakarta.validation.constraints.NotNull;
+
+/**
+ * A concurrency limit definition together with the scope it applies to.
+ * <p>
+ * A flow-scoped limit comes from the flow's own {@code concurrency} property; namespace and
+ * tenant scoped limits are an EE feature. The running counter of a scope is keyed by
+ * {@link #uid()}: for the {@link Scope#FLOW} scope it matches {@link ConcurrencyLimit#uid()},
+ * for broader scopes the absent parts are kept empty.
+ *
+ * @param scope the level this limit applies to
+ * @param tenantId the tenant, never null
+ * @param namespace the namespace the limit applies to, null for a tenant-scoped limit
+ * @param flowId the flow the limit applies to, null unless the scope is {@link Scope#FLOW}
+ * @param concurrency the limit and its behavior
+ */
+public record ScopedConcurrencyLimit(
+    @NotNull Scope scope,
+    @NotNull String tenantId,
+    @Nullable String namespace,
+    @Nullable String flowId,
+    @NotNull Concurrency concurrency) implements HasUID {
+
+    public enum Scope {
+        FLOW,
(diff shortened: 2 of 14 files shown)


AcevedoR:
EE companion PR (same branch name): kestra-io/kestra-ee#9286 — namespace/tenant limit definitions on the Namespace/Tenant models, the `@Replaces` resolver, and the Elasticsearch scoped store implementation. Merge this OSS PR first; the EE PR compiles against this branch.

core/src/main/java/io/kestra/core/models/executions/ExecutionMetadata.java
```diff
@@ -19,6 +20,15 @@ public class ExecutionMetadata {
...
     Instant originalCreatedDate;
 
+    /**
+     * The uids of the concurrency scopes this execution claimed a slot in when it was admitted.
+     * The release decrements exactly these scopes, so removing or changing a namespace/tenant
+     * limit while the execution runs cannot leak the counter of a scope it was admitted under.
+     * Null when the execution never claimed a slot (or predates the scoped limits).
+     */
+    @With
+    List<String> concurrencyScopes;
```
loicmathieu, line 30: I'm not fan, if an execution was started it means all limits was OK so why keeping it here? You should be able at the end of the execution to free any existing limit for that flow with a guard that a limit never goes under 0 in case the limit was added mid-execution.

core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
```diff
@@ -0,0 +1,93 @@
...
+ * for broader scopes the absent parts are kept empty.
+ *
+ * @param scope the level this limit applies to
+ * @param tenantId the tenant, never null
+ * @param namespace the namespace the limit applies to, null for a tenant-scoped limit
+ * @param flowId the flow the limit applies to, null unless the scope is {@link Scope#FLOW}
+ * @param concurrency the limit and its behavior — null only for release-time scopes whose
+ *        definition was removed while the admitted execution ran (see {@link #fromUid})
+ */
+public record ScopedConcurrencyLimit(
```
loicmathieu, line 27: I don't see the value of this new ScopedConcurrencyLimit with respect to the existing ConcurrencyLimit.

core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
```diff
@@ -0,0 +1,93 @@
...
+    @NotNull Scope scope,
+    @NotNull String tenantId,
+    @Nullable String namespace,
+    @Nullable String flowId,
+    @Nullable Concurrency concurrency) implements HasUID {
+
+    public enum Scope {
```
loicmathieu, line 34: Should not be needed, or otherwise we should do the same for quotas to be consistent

executor/src/main/java/io/kestra/executor/ConcurrencyLimitStateStore.java
```diff
@@ -48,4 +58,85 @@ public interface ConcurrencyLimitStateStore {
...
+     * The default implementation only supports the single flow-scoped limit case (the OSS
+     * semantics) and delegates to {@link #countThenProcess(FlowInterface, BiFunction)};
+     * implementors must override it to support namespace or tenant scoped limits.
+     *
+     * @param flow the flow whose execution is evaluated
+     * @param limits the applicable limits in evaluation order, never empty
+     * @param consumer decides from the running counts; returns the processed execution and
+     *        whether to claim one slot in every scope
+     */
+    default ExecutionRunning countThenProcess(FlowInterface flow, List<ScopedConcurrencyLimit> limits,
```
loicmathieu, line 77: I don't think maintaining two stores, one in OSS and one in EE is worth it.
The resolver would not resolves anything from a namespace of a tenant so it's sufficient.

I would merge this one with the one in EE.

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -99,13 +99,32 @@ public ExecutorService(
 
     public ExecutionRunning processExecutionRunning(FlowInterface flow, int runningCount, ExecutionRunning executionRunning) {
```
loicmathieu: Is it still needed?

executor/src/main/java/io/kestra/executor/ConcurrencyLimitStateStore.java
loicmathieu, line 1: I'm surprise you added new methods and not update existing ones...

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/models/executions/ExecutionMetadata.java
AcevedoR: Freeing "any existing limit" at the end doesn't work for the opposite case: a limit removed mid-execution no longer exists, so it's never freed and the counter stays stuck (the never-goes-under-0 guard never triggers). There's a test reproducing this in AbstractScopedConcurrencyLimitStateStoreTest.

 And for a limit added mid-execution, the guard isn't enough either: the execution frees a slot it never took, so other running executions' slots are freed too early and the limit gets exceeded.

That's why we keep the scopes on the execution: we free exactly what was taken at start, whatever the limits are now. It stays null when no limit applies.

executor/src/main/java/io/kestra/executor/ConcurrencyLimitStateStore.java
AcevedoR: agreed I'll merge them

core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
AcevedoR: removing

executor/src/main/java/io/kestra/executor/ExecutorService.java
AcevedoR: no I'll remove it

core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
AcevedoR: ConcurrencyLimit is the stored counter (it has running but no limit or behavior). ScopedConcurrencyLimit is the resolved limit: the Concurrency itself plus the scope it applies to. Merging them would put a Concurrency on a stored entity and a meaningless running on what the resolver returns

executor/src/main/java/io/kestra/executor/ConcurrencyLimitStateStore.java
AcevedoR: I kept the existing methods untouched so no implementation had to change: the new methods delegate to them for the flow-scope case, and only EE overrides them for multi-scope. The executor itself now only calls the new ones. If you prefer a single set of methods, I can change the existing signatures instead — each store (JDBC, Elasticsearch) would then have to handle the flow-scope case itself instead of relying on the default. Your call.

core/src/main/java/io/kestra/core/models/executions/ExecutionMetadata.java
loicmathieu: > a limit removed mid-execution no longer exists, so it's never freed

What should be done is to remove the counter when we remove the limit. I have an openned issue about that, then when you decrement, if it didn't exist, you skip it.

If we need it for your PR, we should then do https://github.com/kestra-io/kestra-ee/issues/8930 before it so we are at the target.

loicmathieu: > And for a limit added mid-execution, the guard isn't enough either: the execution frees a slot it never took, so other running executions' slots are freed too early and the limit gets exceeded.

this is acceptable if it brings simplicity in the implementation: as we have check to never decrement under 0 it would only have a temporal small risk of surge. Acceptable.

core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
loicmathieu: No need to merge them, you have the behavior in the flow/namespace/tenant no need to duplicate it.
As I said, for quota I didn't had to create a separate structure so you should not need one here.

We might need to discuss this together...

core/src/main/java/io/kestra/core/runners/ScopedConcurrencyLimit.java
```diff
@@ -0,0 +1,77 @@
...
```
loicmathieu, line 34: Should not be needed, or otherwise we should do the same for quotas to be consistent

executor/src/main/java/io/kestra/executor/ConcurrencyLimitStateStore.java
loicmathieu: Exposing methods in an interface, not used outside of the interface implementation is a bad architecture.
I'll refactor that before merge.

loicmathieu: Moreover, I'm not fan of default methods with plain implementation.

loicmathieu (approved):

PR #17340, opened 2026-07-09
fix(core): fix service start/stop lifecycle races behind the 'cannot start' CI failures

### LINKED ISSUE #16969: Flaky: {Mysql,Postgres}SchedulerScheduleTest.schedule() — HikariCP pool closed mid-test kills the Trigger queue poller
## Summary

`io.kestra.schedulers.mysql.MysqlSchedulerScheduleTest > schedule()` and `io.kestra.schedulers.postgres.PostgresSchedulerScheduleTest > schedule()` (both inheriting `io.kestra.scheduler.SchedulerScheduleTest.schedule()`) fail intermittently on CI. Observed on `releases/v1.0.x` (Main Workflow runs `27765498169` 2026-06-18 and `27674607099` 2026-06-17).

## Symptom

```
org.opentest4j.AssertionFailedError:
expected: 0L
 but was: 4L
	at io.kestra.scheduler.SchedulerScheduleTest.schedule(SchedulerScheduleTest.java:177)
```

The test fires a 5-hour backfill (6 executions) and waits on a `CountDownLatch(6)` for up to 1 minute (`SchedulerScheduleTest.java:171`), then asserts the latch reached 0. The latch freezes (e.g. at 4) and the test fails after ~61s.

## Root cause (from CI logs)

Mid-test, on the scheduler's queue-poller thread:

```
ERROR jdbc-queue-Trigger_0  io.kestra.jdbc.runner.JdbcQueue  Fatal error while polling the 'io.kestra.core.models.triggers.Trigger' queue. Initiating shutdown.
io.micronaut.data.connection.jdbc.exceptions.CannotGetJdbcConnectionException: Failed to obtain JDBC Connection
  Caused by: java.sql.SQLException: HikariDataSource (HikariPool-23) has been closed.
```

The HikariCP connection pool is **closed while the scheduler is still polling the Trigger queue**. `JdbcQueue` treats the resulting `CannotGetJdbcConnectionException` as fatal and **permanently shuts the queue down**, so no further backfill executions are processed and the latch can never reach 0.

This is a context / connection-pool **teardown race**: the JDBC variants run under `@KestraTest(rebuildContext = true)` (the in-memory `scheduler`-module `SchedulerScheduleTest.schedule()` passes in the same runs), which closes and rebuilds the datasource around each test method. A queue-poller thread appears to outlive the pool it depends on.

## Why a timeout bump does NOT fix it

Once the pool is closed and the queue poller has shut down, the latch can **never** reach 0 — waiting longer cannot help. The fix must address the lifecycle race, not the wait window.

## Possible directions

- Ensure JDBC queue pollers / scheduler background threads are fully stopped **before** the Micronaut context (and its Hikari pool) is torn down between rebuilt contexts.
- Reconsider `JdbcQueue` treating a closed-pool-during-shutdown as a *fatal* error that shuts the whole queue down (vs. a benign teardown condition).

## Reproduction

Not reproducible locally to date (passes 13/13, including under full CPU saturation) — appears to need a loaded CI runner. Needs CI-level reproduction to validate any fix.

## Related

The wall-clock flake in the same triage (`DateFilterTest.now()`) was a separate, unrelated issue, fixed in #16968.

loicmathieu:
## Context

CI sometimes fails with Kestra unable to start: `[DefaultExecutor] not started in time`, `[WorkerAgent] not started in time`, or the cryptic `Condition with Lambda expression in io.kestra.queue.AbstractPollingSubscriber was not fulfilled within 10 seconds`. Digging into the CI logs of a failed run showed this was not just an overloaded CI server: a chain of real races caused **one test class to shut down the application context of the next test class while it was still starting**, cascading into `initializationError` for every subsequent class in the fork.

The chain, in short: a test finished so fast that its context began closing while the executor was still starting → the executor's `doStop()` crashed halfway (concurrent list modification) and never cancelled its scheduled background loops → those orphaned loops kept running after the context (and its datasource) was closed, failed, and called `KestraContext.getContext().shutdown()` — but that static pointer already referenced the **next** test's freshly started context, which got killed mid-startup. Repeat.

This PR fixes each link of that chain, plus the same class of bugs found in the other services by auditing everything that extends `AbstractService`.

## The races and their fixes

**1. The runners reported "ready" before services actually started** — `TestRunner` and `StandAloneRunner` counted a service as started as soon as it was *created* (or even when it had no state at all). Tests could therefore start — and finish, and tear down — while a service was still in the middle of its startup. Both runners now wait until every service has actually reached `RUNNING` (or `MAINTENANCE`), and the timeout error names the service that lagged along with its actual state. This also unmasks real failures: a worker whose connection to the controller failed used to be silently reported as started.

**2. Stopping a service could race its own in-flight startup** — every service (executor, worker, scheduler, controller, indexer) starts asynchronously on a runner thread pool, while `stop()` runs on the context-close thread. When the two overlapped, `stop()` tore down a half-built service: it missed resources created moments later (queue subscribers, scheduled loops, a gRPC server), which then leaked past the context's death. `AbstractService` now owns a small **lifecycle guard** used by all services through a single `guardedStart()` template:
- a stop that arrives *before* startup simply cancels it — nothing is ever created;
- a stop that arrives *during* startup waits for it to finish (bounded by a timeout, so a startup hung on a dead database can never block shutdown forever), then tears everything down;
- a service whose startup finished while a stop was pending no longer advertises itself as `RUNNING`.

**3. Emergency shutdowns could target the wrong application** — several failure paths called `KestraContext.getContext().shutdown()`. That static always points to the *newest* context in the JVM, so a leftover thread from a closed test context would shut down the next test's healthy context. All these paths (executor watchers, queue subscribers, thread-pool uncaught-exception handlers, liveness callback — and the license checker in EE) now hold a reference to **their own** context, captured when they are created: at worst a stale thread now shuts down a context that is already closed, a no-op.

**4. A hidden 10-second wait with an unreadable failure** — queue subscribers ended their subscription with a bare Awaitility wait whose only possible outcome, when it fired, was the cryptic 10-second `ConditionTimeoutException` seen on CI. The wait was a no-op on the happy path and is removed; a subscriber deactivated by a concurrent shutdown now simply doesn't start its polling loop.

**5. `DefaultIndexer` never reported its state** — it shadowed its parent's `state` field, so `getState()` returned `null` forever. Harmless under the old lax readiness check, fatal under the fixed one; the shadowed fields are removed.

**6. A stopping worker could still take work** — the controller dispatches jobs to any connected stream without checking the worker's state, so a worker whose stop was already pending could fetch a job seconds before being killed, causing a duplicate partial run. The worker no longer starts its job fetcher when a stop is pending.

**7. Smaller windows of the same family** — the scheduler now unsubscribes its maintenance/rebalance listeners *before* tearing down (a listener firing at the wrong moment could re-create consumers mid-teardown); the executor's background-loop watchers now learn about a stop at its very beginning instead of halfway through (no more spurious "Executor fatal exception" logs during clean shutdowns); and two JVM-wide statics leaked by tests (`SchedulerClock` frozen at a 2025 date, and a stub `KestraContext`) are now restored after the tests that set them — the frozen clock was itself a source of order-dependent failures.

Fixes #16969

## Validation

- Unit tests for the new lifecycle guard, the subscriber behavior, and the context-bound exception handler.
- `H2RunnerTest` (100 tests) plus the `core`/`queue`/`executor`/`worker`/`scheduler`/`indexer`/`worker-controller` suites all green.
- CI signatures to watch after merge — these should disappear: `not started in time`, `AbstractPollingSubscriber was not fulfilled within 10 seconds`, and `Kestra server - Shutdown initiated` from `*-exception-watcher` threads during teardown.

## Notes for reviewers

- One deliberate behavior decision: a Kestra that cannot reach its database during **normal operation** still shuts down, as before — the fixes only stop *teardown noise* (errors caused by the shutdown itself) from being treated as fatal.
- EE compiles against this branch unchanged (audited: no EE code constructs or overrides the touched classes); a small companion EE branch applies the same owning-context fix to the license checker.
- `kestra server standalone` now genuinely waits for all services before reporting ready — startup may appear marginally slower, but "ready" now means ready.

### DIFF
--- a/cli/src/main/java/io/kestra/cli/StandAloneRunner.java
+++ b/cli/src/main/java/io/kestra/cli/StandAloneRunner.java
@@ -74,6 +74,7 @@ public class StandAloneRunner implements Runnable, AutoCloseable {
 
         poolExecutor = executorsUtils.cachedThreadPool("standalone-runner");
         poolExecutor.execute(defaultExecutor);
+        servers.add(defaultExecutor);
 
         if (controllerEnabled) {
             Controller controller = controllerProvider.get();
@@ -105,16 +106,22 @@ public class StandAloneRunner implements Runnable, AutoCloseable {
         servers.add(systemWorker);
 
         try {
-            Await.await().atMost(getRunningTimeout()).until(
-                () -> servers.stream().allMatch(s -> Optional.ofNullable(s.getState()).orElse(Service.ServiceState.RUNNING).isRunning())
-            );
+            Await.await().atMost(getRunningTimeout()).until(() -> servers.stream().allMatch(StandAloneRunner::isStarted));
         } catch (ConditionTimeoutException e) {
             throw new RuntimeException(
-                servers.stream().filter(s -> !Optional.ofNullable(s.getState()).orElse(Service.ServiceState.RUNNING).isRunning())
-                    .map(Service::getClass)
+                servers.stream().filter(s -> !isStarted(s))
+                    .map(s -> s.getClass().getSimpleName() + " (state: " + s.getState() + ")")
                     .toList() + " not started in time"
             );
         }
+    }
+
+    /**
+     * A service is only considered started once it reached RUNNING (or MAINTENANCE).
+     */
+    private static boolean isStarted(Service service) {
+        Service.ServiceState state = service.getState();
+        return Service.ServiceState.RUNNING == state || Service.ServiceState.MAINTENANCE == state;
     }
 
     private Duration getRunningTimeout() {
--- a/core/src/main/java/io/kestra/core/contexts/KestraContext.java
+++ b/core/src/main/java/io/kestra/core/contexts/KestraContext.java
@@ -99,6 +99,9 @@ public abstract class KestraContext {
 
     /**
      * Shutdowns the Kestra application.
+     * WARNING: this method is not safe to be called from the global KestraContext static instance as in unit test,
+     * this instance may have already been replaced by a fresh one, so you might shut down the next test run!
+     * The safe pattern is to inject a KestraContext instance and call this method on it.
      */
     public void shutdown() {
         // noop
--- a/core/src/main/java/io/kestra/core/exceptions/TypeConversionException.java
+++ b/core/src/main/java/io/kestra/core/exceptions/TypeConversionException.java
@@ -5,7 +5,8 @@ import io.kestra.core.utils.TypeConverter;
 /**
  * Exception thrown by {@link TypeConverter} when a value cannot be converted to the target type.
  *
- * <p>Extends {@link IllegalArgumentException} so that existing {@code IllegalArgumentException}
+ * <p>
+ * Extends {@link IllegalArgumentException} so that existing {@code IllegalArgumentException}
  * handling (e.g. the webserver's HTTP 400 mapping and repository filter wrapping) applies to
  * conversion failures without additional wiring. The original parse failure is preserved as the cause.
  */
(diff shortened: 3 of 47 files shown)


fhussonnois (approved):
LGTM!

nkwiatkowski (approved):

AcevedoR (approved):

Issue #17332, opened 2026-07-09
Frontend: recover from stale lazy-chunk 404s (add a vite:preloadError handler)


flcarre:
## Problem

The UI code-splits many views and components with dynamic `import()` (`defineAsyncComponent`, route-level splitting). Each lazy chunk is served under a content-hashed URL. When that URL goes stale, fetching the chunk fails with a 404 and the user is left on a broken view until they manually refresh.

This happens in two situations:

- **Production, after a deploy** — a user who kept a tab open from before the deploy still references the old chunk hashes. When a lazy component next mounts (navigating, or a re-render remounting an async component), the old chunk is gone → 404.
- **Dev, after an HMR rebuild** — the same effect, transiently.

There is currently no recovery: the app registers no `vite:preloadError` handler (the only `location.reload` in the codebase is an unrelated one in `Restart.vue`).

## Reproduce (production-like)

1. Open the flow editor (or any view that lazy-loads components).
2. Ship a new frontend build so the chunk hashes change.
3. In the still-open tab, trigger a lazy component mount (navigate, or cause a re-render that remounts an async component).
4. Observe `Failed to fetch dynamically imported module` / a 404 on the chunk, and a broken view that only recovers on a manual refresh.

## Expected

The app recovers automatically — reload once to fetch the new chunks — instead of leaving the user on a broken state.

## Proposed fix

Add a `vite:preloadError` listener in `ui/src/main.ts` that reloads the page once, guarded against a reload loop (skip if a reload was already attempted within the last few seconds, e.g. via a `sessionStorage` timestamp) so a genuinely-missing chunk does not loop:

```ts
window.addEventListener("vite:preloadError", () => {
    // reload once, loop-guarded, to refetch chunks whose hashed URL went stale
})
```

Notes:

- `vite:preloadError` covers the production preload path. The dev-server "Failed to fetch dynamically imported module" rejection is a separate, dev-only surface; the scope here is the production case.
- Noticed while working on the no-code block editor, whose canvas mounts several lazy components — a debounced re-render remounting one of them makes the stale-chunk 404 easy to hit during active development, but the underlying gap is general to the app.

MilosPaunovic:
Hi there, @reachsrinivaschennupati-tech! 👋

Thanks for your interest in contributing and for opening this PR!

As a general open-source convention, it’s best to **first comment on the issue to express your interest** and wait until you're **assigned** before starting any work. This helps avoid overlapping efforts and keeps contributions coordinated.  

[...]

PR #17342, opened 2026-07-09
fix(ui): recover from stale lazy chunks

reachsrinivaschennupati-tech:
## Summary

Adds a Vite `vite:preloadError` handler during UI startup so the application can recover when a deployment invalidates previously referenced hashed lazy chunks.

## Problem

After a new frontend deployment, users with an older app shell loaded in the browser can request stale lazy-loaded chunk URLs. When those chunk files no longer exist, the preload error can surface to the user instead of recovering gracefully.

## Solution

- Prevent the Vite preload error from bubbling to the UI.
- Store a preload recovery marker in sessionStorage.
- Reload the page once per session so the browser fetches the latest app shell.
- Avoid reload loops if the preload error persists.

## Testing

- Verified the handler prevents the preload error from surfacing.
- Verified the reload guard prevents repeated reload loops.
- Tested with mocked storage to lock in the expected reload behavior.


reachsrinivaschennupati-tech:
Hi maintainers, this PR is ready for review when you have a chance.

The change is intentionally scoped to Vite stale lazy-chunk recovery: it handles `vite:preloadError`, prevents the error from surfacing to the UI, and reloads only once within a guarded window to avoid reload loops.

Validation noted in the PR:
- Verified the handler prevents the preload error from surfacing.
- Verified the reload guard prevents repeated reload loops.

Happy to refine the recovery window or storage key naming if you prefer a different convention.

flcarre (changes requested):
One thing on the guard logic I wanted to flag: `getLastPreloadErrorReload()` returns `Number(... ?? 0)`, so on a fresh tab `lastReload` is `0` (not `undefined`), and `now - 0` is never `< 10_000`. The first `preloadError` reloads right away (good), but if the chunk is genuinely gone (deleted after a deploy, or offline), the next error after 10s reloads again, and it keeps re-firing every ~10s. The linked issue asks for a single loop-guarded reload, so a one-shot guard might match the intent better:

```ts
function hasReloaded() {
    try {
        return sessionStorage.getItem(PRELOAD_ERROR_RELOAD_KEY) !== null;
    } catch {
        return false;
    }
}

window.addEventListener("vite:preloadError", (event) => {
    event.preventDefault();
    if (hasReloaded()) return;
    try {
        sessionStorage.setItem(PRELOAD_ERROR_RELOAD_KEY, String(Date.now()));
    } catch {
        return;
    }
    window.location.reload();
});
```

A small unit test on the predicate with a mocked `sessionStorage` would also lock it in. Does that match what you had in mind?

reachsrinivaschennupati-tech:
@flcarre  Thanks for catching that, you’re right. The timestamp window still allows another reload after 10s, which doesn’t fully match the one-shot loop guard intent.

I have updated the guard to be one-shot per session instead of timestamp-window based, also added a focused unit test with mocked storage to lock in the expected reload behavior.

Please review.

flcarre (commented):
Thanks for tackling this @reachsrinivaschennupati-tech 🙏 Nice, focused fix. The `vite:preloadError` hook is exactly the right seam, the once-reload loop guard is the correct shape for the infinite-reload risk, the util is cleanly testable via the injected `Storage`, and it's correctly typed (`vite/client` augments `WindowEventMap`, so no cast is needed). 👍

A few non-blocking suggestions inline, mostly around observability and matching the issue's guard semantics:

1. **`preventDefault()` placement** (inline on `main.ts`): avoid silently swallowing the error on the paths where no reload happens.
2. **Dead timestamp / guard semantics** (inline on `preloadErrorReload.ts`): the stored `Date.now()` is never read, so it's worth either aligning with the issue's time-window intent or renaming to a boolean.
3. **Test the listener wiring**: the unit tests nicely cover the util, but the actual loop-guard behaviour in `main.ts` (dispatch `vite:preloadError` twice, expect `reload()` called exactly once) is the part issue #17332 really cares about, and it's easy to lock in with a small `main.ts`-level test that mocks `window.location.reload`.

None of these block the idea, the core approach is solid. Nice work! 🚀

ui/src/main.ts
```diff
@@ -2,6 +2,19 @@ import {createApp} from "vue"
...
+    event.preventDefault()
+
+    if (hasReloadedAfterPreloadError()) {
+        return
+    }
+
+    if (markPreloadErrorReloaded()) {
+        window.location.reload()
+    }
+})
```
flcarre: Small robustness idea: `event.preventDefault()` runs before the guard check, so on the paths where no reload happens (guard already tripped, or `sessionStorage` throws, e.g. Safari private mode), the chunk-load failure is suppressed with no signal at all, leaving the user on a broken view with nothing in the console or error tracking. What do you think about only preventing the default when a reload is actually about to happen, and logging otherwise?

```suggestion
window.addEventListener("vite:preloadError", (event) => {
    if (hasReloadedAfterPreloadError()) {
        console.error("Stale lazy chunk detected, but a reload was already attempted this session", event.payload)
        return
    }

    if (markPreloadErrorReloaded()) {
        event.preventDefault()
        window.location.reload()
    } else {
        console.error("Stale lazy chunk detected, but the reload guard could not be persisted", event.payload)
    }
})
```

ui/src/utils/preloadErrorReload.ts
```diff
@@ -0,0 +1,18 @@
+export const PRELOAD_ERROR_RELOAD_KEY = "kestra:vite-preload-error-reload-at"
```
flcarre: One thing worth aligning: the key is named `...-reload-at` and stores `Date.now()`, but `hasReloadedAfterPreloadError` only checks `!== null`, so the timestamp is written and never read. The linked issue (#17332) suggested a time-window guard (skip only if a reload happened "within the last few seconds"), whereas this is effectively once-per-tab-session, indefinitely. Both are reasonable, but they diverge in one case: a long-lived tab (common in Kestra for editing/monitoring) that survives a second deploy hits a fresh stale chunk and never auto-recovers, because the key is still set from the first time.

Could go either way, pick whichever you prefer:
- **Time-window** (matches the issue): read the stored timestamp back and allow a new reload once `Date.now() - stored > WINDOW_MS`.
- **Once-per-session** (simpler, matches the PR description): keep the behaviour, but store a boolean and rename the key to drop the misleading `-at`, so the value isn't dead data, e.g. `kestra:vite-preload-error-reloaded`.

reachsrinivaschennupati-tech: @flcarre Thank you for the suggestion, makes sense. I have included all your suggestions the coverage for the listener wiring as well. I extracted the preload-error listener setup into a small testable helper, so the unit test can dispatch two `vite:preloadError` events and verified that reload is called only once without importing the full app bootstrap. Please review once at your end.

reachsrinivaschennupati-tech:
@flcarre I have addressed the review suggestions, could you please review.

ui/src/utils/preloadErrorReload.ts
```diff
@@ -0,0 +1,48 @@
...
+    logger?: Pick<Console, "error">
+}
+
+export function hasReloadedAfterPreloadError(storage: Storage = window.sessionStorage) {
+    try {
+        return storage.getItem(PRELOAD_ERROR_RELOAD_KEY) !== null
+    } catch {
+        return false
+    }
+}
```
flcarre: Optional: switch to a time-bounded guard so a later, unrelated stale-chunk incident in a long-lived tab can still recover (paired with the `setItem` change below):

```suggestion
export function hasReloadedAfterPreloadError(storage: Storage = window.sessionStorage, windowMs = 10_000) {
    try {
        const last = Number(storage.getItem(PRELOAD_ERROR_RELOAD_KEY) ?? 0)
        return last > 0 && Date.now() - last < windowMs
    } catch {
        return false
    }
}
```

ui/src/utils/preloadErrorReload.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    window.addEventListener("vite:preloadError", handler)
+
+    return () => window.removeEventListener("vite:preloadError", handler)
+}
+
+export function markPreloadErrorReloaded(storage: Storage = window.sessionStorage) {
+    try {
+        storage.setItem(PRELOAD_ERROR_RELOAD_KEY, "true")
```
flcarre: Store the timestamp the guard above reads, instead of a permanent `"true"`:

```suggestion
        storage.setItem(PRELOAD_ERROR_RELOAD_KEY, String(Date.now()))
```

flcarre (commented):
Thanks for picking this up, @reachsrinivaschennupati-tech — this is a clean, well-scoped fix for #17332. The dependency-injectable `storage`/`reload`/`logger` design makes the handler genuinely testable, the sessionStorage key is nicely namespaced, and it fails safe (falling back to the original error rather than swallowing it). Nice work. 👍

I verified the mechanics against Vite's own runtime and they're correct: `WindowEventMap["vite:preloadError"]` type-resolves via `vite/client`, and Vite's `handlePreloadError` re-throws only when `preventDefault()` wasn't called — so calling it on the recovery branch and skipping it on the guard branch is exactly right.

One design point I'd love your take on, plus a couple of optional nits below. None of this is blocking — happy to merge once we've talked through the guard semantics.

### Loop-guard: permanent marker vs. time window

The issue's proposed fix asked specifically for a *time-window* guard ("skip if a reload was already attempted **within the last few seconds**"), and interestingly your first commit implemented exactly that before a later commit switched it to a permanent per-session boolean. I think the time-window version is worth keeping, because of long-lived tabs (Kestra dashboards often stay open for hours):

- Deploy #1 → stale chunk → auto-reloads once ✅ (marker set)
- Deploy #2, hours later, same tab → stale chunk → marker is still set → handler takes the log-only branch, doesn't call `preventDefault()`, so Vite re-throws and the user is stuck on a broken view until a manual refresh.

A cooldown window keeps the loop protection (a reload still can't retrigger itself within a few seconds) while letting a genuinely-later stale-chunk incident recover too. Suggestions inline. Was there a reason you moved away from the timestamp approach? If there's a subtlety I'm missing, I'm glad to be corrected.

### Optional, non-blocking

- **Test coverage for the defensive branches.** The `catch { return false }` paths and the "could not be persisted" log-only branch (Safari private mode / quota-exceeded) are the whole reason for the try/catch, but they're currently untested. A storage mock whose `getItem`/`setItem` throw, asserting `logger.error` fires and `reload` does *not*, would lock them in.
- **Storage mock** uses `as unknown as Storage`, which bypasses the type checker — fine today, just means a future `removeItem`/`clear` call would compile against a mock that silently returns `undefined`.

Thanks again for the contribution — this is a real papercut and a good fix.

flcarre (approved):
Thanks again @reachsrinivaschennupati-tech! I went ahead and pushed a small follow-up commit (803a4f5) that swaps the permanent per-session marker for a timestamped cooldown window (`PRELOAD_ERROR_RELOAD_WINDOW_MS = 10_000`), so a later stale-chunk incident in a long-lived tab can still auto-recover while keeping the loop protection. I also extended the tests to cover recovery after the window elapses and the storage-cannot-persist fallback.

Verified locally: the util file's unit tests pass (6/6), ESLint is clean, and `vue-tsc` reports no errors in the changed files. Approving. 🚀

PR #17360, opened 2026-07-10
fix(ui): correctly position a task's Gantt bar when a later sibling runs longer

### LINKED ISSUE #17274: Gantt chart misplaces a quick task's execution segment when followed by a long-running sibling inside `Sequential`, making correct sequential execution look parallel
## Summary
The Execution Gantt chart renders a quick-completing task's actual execution segment near the END of a following long-running sibling's duration instead of at its true (near-instant, early) position — making correctly sequential tasks inside a `Sequential` block visually appear to overlap/run in parallel.
Verified via task logs that the real execution order and timing were correct end-to-end; only the Gantt rendering is wrong.

## Expected Behavior
Inside a `Sequential` block, each child's rendered bar should reflect its own real start/end timestamps. A near-instant task followed by a 10s `Sleep` should render as: the quick task's bar completing almost immediately, then the Sleep task's bar starting right after and spanning ~10s.

## Actual Behavior
A quick `kv.Set` task's executed/colored segment is rendered near the END of its row's width — overlapping with where the following 10s `Sleep` task iss hown running — instead of at the very beginning where it actually executed.

Image

Confirmed via the execution's task logs that real execution had zero overlap:
```
12:04:33.881  <preceding task> done
12:04:34.523  <Sleep task> starts: "Waiting for PT10S"
12:04:44.671  <next task> fails — 10.148s later, exactly matching the sleep
```
The quick `kv.Set` task ran instantly between the first two log lines (not logged itself, so it completed in well under 0.6s) — yet its Gantt segment appeared aligned with the END of the Sleep task's bar, not the start.

## Steps to Reproduce
1. Create and run the flow below.
2. Open the Execution's Gantt chart.
3. Observe the quick task's colored segment rendered near the end of the
   block's timeline instead of at the very start.

```yaml
id: gantt_repro
namespace: sandbox

tasks:
  - id: block
    type: io.kestra.plugin.core.flow.Sequential
    tasks:
      - id: quick_task
        key: "repro_key"
        value: "v2"

      - id: slow_task
        duration: PT10S

      - id: fail_task
        errorMessage: "intentional failure for repro"
```

## Environment
- Kestra Version: 1.3.26 
- Deployment: Docker Compose, single node, `server standalone`
- Repository: postgres

## Impact
Purely visual — no functional execution bug (confirmed via task logs). But it actively misleads anyone diagnosing timing/ordering from the Gantt view; we initially suspected a real Sequential-ordering bug until we checked raw timestamps.

riddhij-7:
Fixes #17274
### ✨ Description

What does this PR change?  
Fixes the Execution Gantt chart rendering a task's colored/executed segment at the wrong position when a quick finishing task is followed by a longer running sibling inside a Sequential (or Parallel) block, making correctly sequential execution visually look like it overlapped.
Root cause: In Gantt.vue, every child task's bar width was unconditionally stretched from its own start to its parent's end (parentEndPercent = parent.start + parent.width), regardless of how much earlier the child actually finished. For a quick task followed by a long-running sibling, this stretched the quick task's bar across time it never ran in, so its colored/executed segment (positioned as a percentage of that stretched container) rendered near the end instead of at its true, near instant position.
This stretching was originally introduced to fix a cosmetic nested alignment gap (a single child's bar not quite reaching its parent's edge), but it was applied unconditionally to every child, not just the one case it was meant for.

Fix: Only stretch a task's bar to align with its parent's end when the task's own real finish time is already within a small epsilon (200ms) of the parent's, i.e. a genuine rounding/precision gap, not a sibling that still has time left to run. Extracted this into a small, pure, independently tested function: ui/src/utils/ganttSeries.ts (computeTaskBarPercents), following the same pattern already used by taskRunHierarchy.ts.
An earlier version of this fix picked the "last child by start timestamp" to stretch, which broke under Parallel blocks (siblings start at roughly the same time, so scheduling jitter could flag the wrong child). The final approach depends only on how close a child's end is to its parent's end, never on start order,  see the added test for that regression.

### 🔗 Related Issue
Closes [https://github.com/kestra-io/kestra/issues/17274](https://github.com/kestra-io/kestra/issues/17274)

### 🎨 Frontend Checklist
**Before:**
image
**After:**
image

### 📝 Additional Notes
Testing:
Added ui/tests/unit/utils/ganttSeries.spec.ts covering: the original Sequential repro from #17274, the Parallel start-order regression, the single-child nested-alignment case, an orphan-parent edge case, and a custom-epsilon case.
Manually reproduced the issue using the exact repro flow from #17274, plus additional manual verification against a Parallel block and a single-child wrapper, confirming correct rendering before/after the fix in each case.
npx vitest run tests/unit/utils/ all passing.
vue-tsc --noEmit no type errors.


ui/src/components/executions/Gantt.vue
```diff
@@ -497,17 +512,14 @@
                 const lastIndex = task.state.histories.length - 1
                 stopTs = ts(task.state.histories[lastIndex].date)
             }
-
+
```
RemiBarthe: trailing whitespace

ui/src/utils/ganttSeries.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+const DEFAULT_ALIGNMENT_EPSILON_MS = 200
+
+/**
+ * @param tasks               
+ * @param executionStart      
+ * @param executionDelta      
+ * @param alignmentEpsilonMs 
+ * @return
```
RemiBarthe: empty @return in the JSDoc, worth filling in or dropping

RemiBarthe (changes requested):
Really nice work on this 🙌

Just a few small cleanup nits before merge (no behavior impact)
Once those are tidied up, good to go. thanks again ! 👍

ui/src/components/executions/Gantt.vue
riddhij-7: Fixed!

ui/src/utils/ganttSeries.ts
riddhij-7: Fixed!

riddhij-7:
@RemiBarthe, review comments addressed.

RemiBarthe (approved):
LGTM, thanks again for the contribution ! 👍

PR #17362, opened 2026-07-10
feat(flow): add input with subflow function inside the topology

loicmathieu:
Closes https://github.com/kestra-io/kestra-ee/issues/8589

### DIFF
--- a/core/src/main/java/io/kestra/core/topologies/FlowTopologyService.java
+++ b/core/src/main/java/io/kestra/core/topologies/FlowTopologyService.java
@@ -12,6 +12,8 @@ import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.flows.Flow;
 import io.kestra.core.models.flows.FlowInterface;
 import io.kestra.core.models.flows.FlowWithSource;
+import io.kestra.core.models.flows.input.MultiselectInput;
+import io.kestra.core.models.flows.input.SelectInput;
 import io.kestra.core.models.hierarchies.Graph;
 import io.kestra.core.models.tasks.ExecutableTask;
 import io.kestra.core.models.topologies.FlowNode;
@@ -25,6 +27,7 @@ import io.kestra.core.runners.RunContext;
 import io.kestra.core.runners.RunContextFactory;
 import io.kestra.core.services.ConditionService;
 import io.kestra.core.utils.ListUtils;
+import io.kestra.core.utils.PebbleUtil;
 
 import io.micronaut.core.annotation.Nullable;
 import jakarta.inject.Inject;
@@ -161,6 +164,10 @@ public class FlowTopologyService {
             return FlowRelation.FLOW_TRIGGER;
         }
 
+        if (this.isInputChild(parent, child)) {
+            return FlowRelation.SUBFLOW_FUNCTION;
+        }
+
         return null;
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/kestra/core/topologies/FlowTopologyService.java
```diff
@@ -208,6 +215,31 @@ protected boolean isTriggerChild(Flow parent, Flow child) {
...
+            .map(MultiselectInput.class::cast)
+            .anyMatch(selectInput -> isSubflowFunctionFor(selectInput.getExpression(), child));
+
+        return selectMatch || multiSelectMatch;
+    }
+
+    // best effort: we use regexes as a substitute for proper parsing of Pebble expressions
+    private boolean isSubflowFunctionFor(String expression, Flow child) {
+        return expression != null && PebbleUtil.containsOpeningBlockDelimiter(expression) &&
+            expression.matches(".*subflow\\s*\\(.*\\).*") &&
```
brian-mulier-p, line 238: `private static final Pattern SUBFLOW_CALL_PATTERN = Pattern.compile(".*subflow\\s*\\(.*\\).*");`

core/src/main/java/io/kestra/core/topologies/FlowTopologyService.java
```diff
@@ -208,6 +215,31 @@ protected boolean isTriggerChild(Flow parent, Flow child) {
...
+            expression.matches(".*id\\s*=\\s*['\"]?" + child.getId() + "['\"]?\\s*[\\s,)].*") &&
+            expression.matches(".*namespace\\s*=\\s*['\"]?" + child.getNamespace() + "['\"]?\\s*[\\s,)].*");
+    }
```
brian-mulier-p, line 241:
```
// best effort: we use regexes as a substitute for proper parsing of Pebble expressions
private boolean isSubflowFunctionFor(String expression, Flow child) {
    if (expression == null || !PebbleUtil.containsOpeningBlockDelimiter(expression) || !SUBFLOW_CALL_PATTERN.matcher(expression).matches()) {
        return false;
    }

    String id = Pattern.quote(child.getId());
    String namespace = Pattern.quote(child.getNamespace());
    String idArg = "\\bid\\s*=\\s*['\"]?" + id + "['\"]?\\s*[\\s,)]";
    String namespaceArg = "\\bnamespace\\s*=\\s*['\"]?" + namespace + "['\"]?\\s*[\\s,)]";

    // named parameters, any order, e.g. subflow(namespace = 'io.kestra.ee', id = 'child') or subflow(id = 'child', namespace = 'io.kestra.ee')
    boolean namedMatch = expression.matches(".*" + idArg + ".*" + namespaceArg + ".*|.*" + namespaceArg + ".*" + idArg + ".*");

    // unnamed parameters, positional order (namespace, id), e.g. subflow('io.kestra.ee', 'child')
    boolean unnamedMatch = expression.matches(".*subflow\\s*\\(\\s*['\"]?" + namespace + "['\"]?\\s*,\\s*['\"]?" + id + "['\"]?\\s*[\\s,)].*");

    return namedMatch || unnamedMatch;
}
```

This should be more resilient and will manage named arg regardless of their position and unnamed args (position-based)

brian-mulier-p: Add a test for unnamed args

brian-mulier-p (changes requested):
Can you please add support for unnamed args? Otherwise LGTM

core/src/main/java/io/kestra/core/topologies/FlowTopologyService.java
loicmathieu: For positioned args, this may lead to false positives, and we try to avoid false positives on topology and prefer to be stricter.
So I'm not confident of doing it as there are optional parameters that might interfere especialy inputs or labels that are free form

brian-mulier-p: I'm not sure to understand why it could lead to false positives? The first two positional args are namespace & id and I doubt it will ever change so I'm not sure to get it :thinking:

loicmathieu: Your pattern for named arg is not OK: it didn't check that named args are inside the subflow function, and didn't take into account that they are not ordered.

For positional args, a regex can match inputs instead of namespace/flow so I think we might not support it, at least for now. I'll let that for a followup if we really want to support it.

```
subflow("myNamespace", "myFlow", {"input1": "other", "input2, "well'})
```

This would match positional arg with "other" as namespace and "input2" as flow.

brian-mulier-p: I don't think because the positional check is checking for subflow(<namespace>,<id> (with the regex verbosity ofc) so it always matches the first 2 arguments, nothing else so I don't think it's an issue. Anyway I'll not block the PR for this.

For the named arg regex you're right but I think it's just a matter of adding `.*subflow\\s*\\(` ahead :thinking:

brian-mulier-p (approved):

PR #17366, opened 2026-07-10
feat(logs): capability-aware LogDataStoreInterface (aggregate / purge / cursor pagination)

nkwiatkowski:
## What

Makes the pluggable `LogDataStoreInterface` **capability-aware** so a future external log store (GCP Cloud Logging, S3, Datadog, …) that cannot aggregate, cannot purge on demand, and paginates by a forward cursor is a first-class backend and the whole system degrades gracefully. **Backend only — no GCP implementation in this PR.**

Context / analysis: kestra-io/kestra-ee#7054 (capability matrix across GCP/S3/Azure/Datadog/Splunk is in the issue).

## Changes

- **Capabilities, honoured by the store itself:**
  - `QueryBuilderInterface.canAggregate()` and `LogDataStoreInterface.canPurge()` (default `true` → JDBC/Elasticsearch unchanged).
  - `AbstractJdbcLogDataStore` returns an empty aggregation / zero KPI when it can't aggregate, and no-ops purge/delete (returns `0`, without firing the delete audit) when it can't purge. Consumers (dashboard engine, purge callers) stay backend-agnostic — no `instanceof`/capability checks pushed into them.
- **Pagination:** paginated `find*` now return Micronaut `Page<LogEntry>` (offset + total) or `CursoredPage<LogEntry>` (forward cursor, no total). New `PaginationType { OFFSET, CURSOR }` + `paginationType()` advertises the mode. `PagedResults` gained `type` + an opaque `nextCursor`; `GET /logs/search` accepts a `?cursor=` token. **Offset responses keep `{results, total}`, so the current UI is unchanged.**
- **Shared contract test:** `AbstractLogDataStoreTest` now branches on the store's declared capabilities (aggregate vs empty, purge vs no-op, offset vs cursor), making it the single suite that exercises every combination. A test-only `ReducedLogDataStore` (extends H2, all capabilities off) proves the degraded paths.

## Not in this PR / follow-ups

- No GCP/external implementation yet.
- Frontend: offset UI is untouched; a cursor UI (load-more, no total) is deferred to the external-store work. The mode can be advertised to the UI via `/api/v1/configs` (`logs.paginationType`) when that lands.
- **Open question (needs a call):** the capability flags currently drive only the store's own graceful degradation + the test contract — they are not consulted elsewhere in production code. Whether to surface/consult them more widely is left for a follow-up.

## Tests

`AbstractLogDataStoreTest` (H2 = all caps, `ReducedLogDataStore` = none) green with every branch exercised; `LogControllerTest` (offset), `LogControllerCursorTest` (cursor response shape + `?cursor=` round-trip), `PagedResultsTest` (offset vs cursor shapes) all pass.


jdbc-h2/src/test/java/io/kestra/repository/h2/ReducedLogDataStore.java
```diff
@@ -0,0 +1,75 @@
...
+        Pageable offsetPageable = pageable.isUnpaged()
+            ? Pageable.UNPAGED
+            : Pageable.from((int) (offset / size) + 1, size, pageable.getSort());
+
+        List<LogEntry> content = super.find(offsetPageable, tenantId, filters).getContent();
+
+        // One cursor per row (Micronaut expects the list to match the content); every entry carries the same
+        // next-page token, so the last one — which the wrapper reads — resumes at the next page.
+        String nextToken = String.valueOf(offset + size);
+        List<Pageable.Cursor> cursors = new ArrayList<>();
```
loicmathieu:
```suggestion
        List<Pageable.Cursor> cursors = new ArrayList<>(content.size());
```

Better to pre-size the array list

webserver/src/main/java/io/kestra/webserver/responses/PagedResults.java
```diff
@@ -16,9 +19,29 @@ public class PagedResults<T> {
...
     private List<T> results;
 
-    @JsonInclude
-    @NotNull
-    private long total;
+    /**
+     * Exact total row count — present in offset mode (and for the {@link ArrayListTotal} path used by other
+     * endpoints), omitted in cursor mode (a cursor store has no total).
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
```
loicmathieu: I think this annotation is useless as it's the default

loicmathieu: Same for the other

loicmathieu (approved):
Overall LGTM, some small nits but you can merge as is if you want

PR #17372, opened 2026-07-13
fix(ui): unify fitHeight layout across listing pages for pinned pagination

brian-mulier-p:
- KsDataTable gains a `fitHeight` prop + `.fit` CSS modifier so a table can bound its height to its container instead of growing the page
- Fixed an element-plus bug where empty-state tables still reserved scroll height, causing a phantom page scrollbar
- Tabs.vue gains a `fullContainer` tab option, giving the active tab's content flex-bounded height for tabs that host a full listing page
- Added missing `fullContainer` field to the namespaces-local `Tab` type (useHelpers.ts) so the KV Store tab wiring type-checks
- Wired fitHeight/full-container across OSS listing pages: Flows, Executions, Secrets, SecretsTable, KVs, KVTable, ConcurrencyLimits, LogsWrapper, BlueprintsBrowser, Blueprints override
- Fixes kestra-io/kestra-ee#9299 — pagination bar scrolled off-screen on long listing pages instead of staying pinned at the bottom
- Layout/CSS change — verified live across ~15 pages (empty, single-page, multi-page states); no meaningful unit-test coverage for this kind of change

### DIFF
--- a/ui/src/components/admin/triggers/TriggersManage.vue
+++ b/ui/src/components/admin/triggers/TriggersManage.vue
@@ -8,6 +8,7 @@
             :currentPage="urlPage"
             :pageSize="urlSize"
             :defaultSort="{prop: 'flowId', order: 'ascending'}"
+            fitHeight
             :selectable="canCheck"
             :selectionMapper="selectionMapper"
             :rowClassName="getClasses"
@@ -859,6 +860,10 @@
 
 <style scoped lang="scss">
     .triggers-manage {
+        display: flex;
+        flex-direction: column;
+        min-height: 0;
+
         :deep(tr.no-expand .kel-table__expand-icon) {
             pointer-events: none;
 
--- a/ui/src/components/kv/KVs.vue
+++ b/ui/src/components/kv/KVs.vue
@@ -10,7 +10,7 @@
             </ul>
         </template>
     </TopNavBar>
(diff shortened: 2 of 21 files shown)


ui/src/components/admin/triggers/TriggersManage.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div class="triggers-manage">
+    <div class="triggers-manage d-flex flex-column fill-height">
```
elevatebart: Can we avoid using utility classes if we a class is already defined?

ui/src/components/kv/KVs.vue
```diff
@@ -10,7 +10,7 @@
             </ul>
         </template>
     </TopNavBar>
-    <section class="d-flex flex-column fill-height container padding-bottom">
```
elevatebart, line 13: I love this !! ❤️ Let's remove as many utility classes as we can.

ui/src/components/secrets/SecretsTable.vue
```diff
@@ -610,6 +610,10 @@
     })
 </script>
 <style scoped lang="scss">
+    .d-flex.flex-column.fill-height {
+        min-height: 0;
+    }
```
elevatebart: No, please use a real class instead.

elevatebart (approved):
A few NIT, really easy to fix

PR #17374, opened 2026-07-13
fix: design-system-duration-picker-disabled

### LINKED ISSUE #17303: [Design System] KsDurationPicker is missing a `disabled` prop
### Describe the issue

`KsDurationPicker` (`ui/packages/design-system/src/components/Form/KsDurationPicker.vue`) does not expose a `disabled` prop, and none of its internal fields (the 7 `KsInputNumber` unit inputs, plus the custom-duration `KsInput`) receive one either.

Every other form control in the design system (`KsInput`, `KsInputNumber`, `KsSelect`, `KsSwitch`, ...) supports `:disabled`, so `KsDurationPicker` is the odd one out. This becomes a problem when composing a form where some rows/sections must be fully read-only: there is no clean way to disable the picker, and CSS-based workarounds (e.g. `pointer-events: none`) only block mouse/touch interaction — the inner number inputs remain keyboard-focusable and editable via Tab, which is an accessibility inconsistency versus sibling fields that are properly `:disabled`.

**Suggested fix:** add a `disabled?: boolean` prop to `KsDurationPicker` and forward it to each nested `KsInputNumber` and the custom-duration `KsInput`, mirroring the pattern used by other design-system form components.

### Environment

- Kestra Version: develop

hemantmm:
closes: #17303

---

### ✨ Description

What does this PR change?  

Added a `disabled` prop to `KsDurationPicker` and forwards it to all nested duration inputs.

## Changes

- Added `disabled?: boolean` to `KsDurationPicker`.
- Passed `disabled` to the 7 `KsInputNumber` unit fields.
- Passed `disabled` to the custom ISO duration `KsInput`.
- Added unit coverage for disabled nested inputs.
- Added a Storybook disabled state.

https://github.com/user-attachments/assets/f664a639-e56d-49d2-8ad4-24bc0b46cb6d


hemantmm:

flcarre:
Thanks for this @hemantmm — really clean, well-scoped fix that does exactly what #17303 asks for. 👏

A few things I checked and liked:

- **Propagation is correct end-to-end.** `:disabled` binds as the child's own declared prop on `KsInputNumber`/`KsInput`, so it flows through to a real native `disabled` attribute on each inner `<input>` — no ambiguity with the component's `inheritAttrs:false` + `$attrs` root binding.
- **It fixes the actual a11y motivation.** Native `disabled` removes the inputs from tab order, which is precisely the gap the old `pointer-events:none` workaround left open (mouse-blocked but still Tab-focusable).
- **Tests have real teeth.** Full mount rather than shallow, so the unit test genuinely exercises DOM propagation, and the Storybook `play` also confirms the ARIA roles while disabled.
- **No blast radius** — the optional, default-`false` prop can't break existing usage, and the new Storybook story nicely matches the file's existing style pattern.

LGTM ✅

flcarre (approved):
LGTM ✅ Clean, well-scoped fix that does exactly what #17303 asks — `disabled` propagates end-to-end to the native inputs, fixes the tab-order a11y gap the old `pointer-events:none` workaround left open, and the tests have real teeth (full mount + Storybook `play`). Thanks @hemantmm!

hemantmm:
Thanks a lot for the detailed review and approval @flcarre!

Glad the disabled propagation, accessibility behavior, and test coverage all look good. Thank you for taking the time to verify the native input behavior and tab-order fix. 🙌

PR #17379, opened 2026-07-13
chore: upgrade spotless conventions plugins

Skraye:
EE equivalent: https://github.com/kestra-io/kestra-ee/pull/9311


Skraye:
close as update was done in another PR

PR #17096, opened 2026-06-29
feat(flows): add the no-code Blocks editor

flcarre:
## Summary

Adds a no-code **Blocks editor**: a canvas-based visual editor for flows, wired as the default engine of the no-code tab. The legacy schema-driven form is kept behind a `NOCODE_ENGINE=legacy` localStorage flag.

- **Canvas** (`ui/src/components/no-code/blocks/`): renders each flow section (`tasks`, `errors`, `finally`, `afterExecution`, `triggers`) as a list of cards. Leaf tasks render as `LeafBlockCard`; flowables (If / Switch / Parallel / Sequential / ForEach / Dag / …) render as expandable `FlowableClusterCard` with per-branch lanes (`then` / `else` / `cases` / `tasks`), including DAG `dependsOn` editing.
- **Mutations** are path-based YAML operations (`ui/src/utils/flowableBlockOps.ts`): insert, delete, duplicate (with nested-id renaming), reorder and move, plus a bounded undo history.
- **Editing**: click a card to edit it in the shared MultiPanel dock, or open it in a split; keyboard model (`keymap.ts`, `useBlockEditorKeyboard.ts`) with j/k navigation, activation, duplicate/delete, and a `/` command palette (`BlockCommandMenu.vue`).
- **Flow-level properties** panel (`FlowPropertiesEdit.vue`): namespace, description, inputs, outputs, variables, concurrency, retry, sla, checks, pluginDefaults, workerSelector, disabled.
- **Validation**: each block card surfaces its own missing/invalid fields, grouped from the flow's validation constraints by task id.
- **Structured revision diff** (`RevisionStructuredDiff.vue` / `revisionDiff.ts`): an opt-in, per-block diff view alongside the existing text diff on the Revisions tab.
- The **properties-panel sub-task lists** (If `then`/`else`, `tasks`/`errors`/`finally`, single-task fields) now render the same Blocks card as the canvas.
- **Backend**: `@Valid` on `Dag.DagTask.task` so an invalid nested DAG sub-task is caught at flow validation; `@Schema` documentation on `AbstractFlow.disabled`.
- **Tests**: unit specs (`flowableBlockOps`, `revisionDiff`, `BranchLane`, `TaskList`, keymap), Storybook stories, and a Playwright e2e suite under `ui/tests/e2e/blocks/`.
- **i18n**: all 13 locales updated (`block_editor.*`, `revision_diff.*`, …).

## Test plan

- [x] `npm run check:types` passes
[...]

Issue #9616, opened 2025-06-18
[No-code] Autocomplete for expressions not working in no code editor


wrussell1999:
### Describe the issue

My input expression didn't autocomplete for my input. It did work in the flow editor. Would be good to have it work in no code too.

![Image](https://github.com/user-attachments/assets/00c9a8ff-b940-4413-8d08-05d4bc3132f5)

### Environment

- Kestra Version: `0.23.0` EE

wrussell1999:
Blocker for video: https://www.notion.so/kestra-io/Can-I-build-a-workflow-with-no-code-20136907f7b580eea527cf4541abc485?source=copy_link

wrussell1999:
Another example - `inputFiles` for Script tasks:

![Image](https://github.com/user-attachments/assets/0239582d-2c25-4bfd-ae3f-92032d135915)

varsh-21:
**Autocomplete Engine-**

Is it being initialized correctly in no-code mode?

Are the necessary schemas/context variables being passed to the editor?

Check if the issue is limited to no-code editor or global.

Check if autocomplete works in the code editor but not in the no-code editor. This helps isolate UI issues.

**If it's a frontend-only bug, you’ll likely be working in:**

ui/src/pages/flow/designer/...
or
ui/src/components/editor/...

anna-geller:
duplicate of existing https://github.com/kestra-io/kestra/issues/9321 -- to be fair closing this one as Izaac was first :)

PR #17386, opened 2026-07-13
feat(flows): add a guided flow-creation funnel

flcarre:
## What

Extracted from #17096 so the new no-code Blocks editor and the flow-creation funnel can be reviewed and shipped independently.

`/flows/create` now lands on a guided funnel instead of dropping straight into the editor:

- **Blank flow** — id + namespace form, then opens the editor with the starter template.
- **Browse blueprints** — links to the community blueprints catalog.
- **System flow (recipe builder)** — guided notification-flow builder: trigger panels (Schedule / Execution / Webhook / any other trigger), notification channel grid, and a live summary that generates the flow YAML.
- **Import YAML** — paste or upload an existing flow definition.

The namespace view's blueprints tab becomes `SystemBlueprintsTab`, hosting the same recipe builder scoped to the configured system namespace (this also replaces the hardcoded `system` namespace with the `systemNamespace` config). In-place system-blueprint browsing is intentionally dropped in favour of a link out to the community catalog.

**Requires the EE companion https://github.com/kestra-io/kestra-ee/pull/9616 to merge in lockstep**: `useHelpers().tabs` becomes a `ComputedRef` so the system-flow tab can react to `configs.systemNamespace` arriving after the first render, and the EE `useTabs` override spreads it.

Deep links that carry a **flow source** keep their behavior and go straight to the editor: `blueprintId`, `blueprintSource`, `blueprintSourceYaml`, `copy`, `onboardingPreset`, `recipePreset`, `ai`, `createTrigger` (`ui/src/utils/flowCreationLanding.ts`).

`?namespace=` is deliberately **not** in that list: it is context, not a flow source. The three links that carry it (`namespaces/Actions.vue`, `flows/Flows.vue`, `onboarding/tour/TourFinale.vue`) now land on the funnel with the namespace **preselected**, rather than skipping it.

## Changes vs the version embedded in #17096

- `KsTaskIcon` (removed from the design system on develop) replaced with the app-level `plugins/TaskIcon.vue`.
- Adapted to the new `PluginIconMap` icon typing and the nullable blueprint `source`.
- Fixed the `NewFlowLanding` Storybook story: it relied on a `parameters.stubs` option Storybook ignores; it now registers the named routes its router-links resolve.
- Added the missing `datepicker.invalid duration` translation to the 12 non-English locales (pre-existing red on the shared translation check).

## Tests

- Unit: recipe state composable, YAML generation, import parsing, landing form, import form (51 tests).
- Storybook: landing, recipe (incl. dark mode), import (5 stories).
- `check:types`, ESLint and the shared translation check are green.

## Demo

Guided flow-creation funnel — landing, then the stepped system-flow builder (Trigger → Notify → Review on a KsSteps rail, radio trigger tiles, Custom channel, live YAML at review) through to the editor hand-off:

https://github.com/user-attachments/assets/9f3f1164-31f3-4d5e-bacd-ce0dfc561877


flcarre:
Live QA pass on a fresh local instance (backend + vite on this branch), walking the three funnel paths as a real user:

- **Blank flow** → editor opens with the starter YAML for the typed id/namespace ✅
- **System flow** → SystemBlueprintsTab renders the recipe builder; trigger panels work including *Other trigger* (6 trigger rows with icons via the app-level TaskIcon), states selection, live summary and **Preview YAML** generate the expected flow ✅ (notification channels correctly show as unavailable on a distribution without the Slack/Teams/Email plugins)
- **Import YAML** → pasting a definition opens the editor with that YAML ✅

One real bug found and fixed during the pass (ef8013de04): on a fresh instance the namespace select was filterable-only, so with zero existing namespaces the blank-flow path was a dead end — `allow-create` now commits a typed namespace, with a unit test guarding it.

MilosPaunovic (commented):
## Overview

`/flows/create` gains a landing funnel (blank flow form / blueprints link / system-flow recipe builder / YAML import) in front of `MultiPanelFlowEditorView`, gated by `ui/src/utils/flowCreationLanding.ts` — any query key in a hardcoded allowlist bypasses it. The namespace "blueprints" tab is replaced by `SystemBlueprintsTab` hosting the same recipe builder, and the hardcoded `system` namespace becomes `miscStore.configs.systemNamespace`. Recipe → YAML goes through `sessionStorage` (`RECIPE_PRESET_KEY`) + `?recipePreset=true`.

Structurally this is good work: pure logic extracted into testable utils (`recipeToYaml.ts`, `importYamlUtils.ts`, `flowIdentifiers.ts`), state in a composable, presentation split into small components. Design-system compliance is genuinely clean — I verified every `--ks-*` token used exists in `ks-theme-light.scss`/`ks-tokens.scss`, no `:deep()`, no hex, no `element-plus` imports, `data-test` everywhere, no hardcoded user-facing strings. All 13 locales carry real generated translations, not English placeholders. I ran the new suites locally: **79 tests, 9 files, all green**.

Two blockers, though.

---

## 🔴 Blockers

### 1. Breaks kestra-ee at runtime — no companion PR exists

`ui/src/components/namespaces/utils/useHelpers.ts` changes the exported `tabs` from `Tab[]` to `ComputedRef<Tab[]>`. The OSS override is updated to `baseTabs.value`, but EE's override is not:

```ts
// kestra-ee/ui-ee/src/override/components/namespaces/useTabs.ts:41
const merged: Tab[] = [
    ...baseTabs,          // ← was an array, now a ComputedRef
```

Verified against the installed Vue: `[...computed(() => [1,2])]` throws `TypeError: c is not iterable`. That takes out every namespace detail page in EE, plus `useNamespaceTabsSidebar`, and fails EE `check:types`. I searched kestra-ee for a companion PR on `feat/flow-creation-funnel` or referencing this one — there is none. Needs a paired EE PR merged in lockstep (`...baseTabs.value`).

### 2. The generated recipe YAML is invalid for 3 of the 4 notify channels

I checked each emitted FQCN against its published plugin schema. `TeamsExecution` under `io.kestra.plugin.microsoft365.teams` does exist (good — that one's fine), but the property sets don't line up:

| Channel / path | Emitted | Problem |
|---|---|---|
| Slack — **both** paths | `channel: "#alerts"` | Neither `SlackExecution` nor `SlackIncomingWebhook` has a `channel` property (`url, messageText, payload, executionId, customMessage, customFields, templateRenderMap, options`). The channel is determined by the webhook URL. |
| Teams — non-execution triggers | `message: "Flow … triggered"` | `TeamsIncomingWebhook` takes `url` + `payload`, not `message`. |
| Email — both paths | `to: [state.emailTo]`, no `host` | `MailSend.to` is a *string* ("semicolon-delimited list of RFC2822 addresses"), not a list; and with no `host` the task can't send. |

Kestra fails on unrecognised task properties, so a user who picks Slack (the headline channel) lands in the editor with a red validation banner on a flow the product just generated for them. It also means the "Slack channel" input in `NotifyGrid.vue` is collecting a value that can't be expressed — either drop the field or move it into the `payload` JSON.

Worth generating one flow per channel × trigger type and pasting each through `/api/v1/flows/validate` before merge; the unit tests assert the shape the code produces, so they can't catch this class of bug.

---

## 🟡 Should fix

**Silent channel drop on trigger-type change.** `useFlowRecipe.ts:62-85` computes `channelAvailability` from the trigger type, but `hasNotifyChannel` (and therefore `isValid`) only reads `recipe.notify.*`. Select `execution` + Slack, then switch to `schedule`: if the webhook variant isn't installed, availability flips to `false`, the tile greys out, `recipe.notify.slack` stays `true`, `buildNotifyTasks` silently omits the task, and `isValid` is still `true`. The user creates a flow containing only the `placeholder` Log task with no warning. `hasNotifyChannel` should intersect with `channelAvailability`, or switching trigger type should clear now-unavailable channels.

**Imported YAML without `id`/`namespace` produces an unsaveable flow.** In `FlowCreate.vue:117-118` the import branch never sets `shouldApplyGeneratedMetadata = true`, so `withGeneratedFlowMetadata` isn't applied. Paste a fragment that's just `tasks:` and the editor opens with a source that has no `id`/`namespace` while `flowStore.flow.id` holds a random one — save fails. `withGeneratedFlowMetadata` already guards with `!("id" in parsedFlow)`, so it's safe to run unconditionally on this path. Relatedly, `parseImportYaml` computes and returns `id`/`namespace` that nothing consumes — that looks like the wiring that got lost.

**Blueprint detail in the system-namespace tab is now unreachable.** `useHelpers.ts:95` keeps `blueprintDetail: true`, which makes `Tabs.vue:301` pass an `onGoToDetail` listener — but `SystemBlueprintsTab` never emits `goToDetail`, so `selectedBlueprintId` can never be set and the inline detail view is dead. Either drop the flag or forward the event. The tab also loses in-place system-blueprint browsing entirely (it's now an outbound link to the community catalog) — intentional? Worth calling out in the description if so.

**`namespace` isn't in the bypass allowlist**, so three existing entry points now land on the funnel instead of the editor: `onboarding/tour/TourFinale.vue:107` ("Start building" at the end of the onboarding tour), `override/components/namespaces/Actions.vue:10`, and `flows/Flows.vue:15`. Defensible (the namespace is prefilled), but it contradicts "any meaningful query param bypasses the landing" in the description, and the tour ending on a form rather than an editor is a UX change someone should sign off on.

**Two dead translation keys shipped to all 13 locales:** `recipe.blueprints_section_title` and `recipe.summary.yaml_preview` have no reference anywhere in `src/` or `tests/`.

---

## 🟢 Nits

- **Duplication.** The 500-namespace fetch (`useNamespaces(500).all()` + loading/error refs) is copy-pasted between `NewFlowLanding.vue:180-190` and `FlowRecipe.vue:208-219` — extract a composable. `"0 9 * * *"` is hardcoded in three places and `"FAILED, WARNING"` in two; `DEFAULT_WEBHOOK_KEY` shows the right pattern.
- **Props mutated as state.** Every panel does `v-model="recipe.someField"` on a prop (and `SchedulePanel` assigns `props.recipe.cron = …` directly). It works because `recipe` is a shared reactive, but it's one-way-data-flow inverted; `v-model` on the child or emits would be conventional.
- **`SelectableTile` looks like a design-system component living in feature code** — generic name, generic API (`role`/`indicator`/`layout`/`selected`/`disabled`), ~90 lines of CSS, no story or unit test of its own. Per `ui/AGENTS.md` (rules 8–9, and the ">50 lines of CSS" smell) this belongs in `packages/design-system` with a story, or should be expressed via existing `KsRadio`/`KsCheckTag`.
- **Two trigger cards share `type: "other"`** (`FlowRecipe.vue:252` and `:268`). Only masked by `card.disabled` on the first; enabling "case" later would highlight both.
- **Stale-flow flicker.** `handleLandingProceed`/`handleImportSubmit` set `showLanding = false` *before* awaiting `setupFlow`, so `MultiPanelFlowEditorView` can mount for a tick against a `flowStore.flow` left over from a previous route. Flip the flag after the await.
- `Math.random().toString(36).slice(2, 7)` for the recipe flow id — 5 base-36 chars, no collision check, and not user-editable before creation. `getRandomID()` is the house helper.
- Hardcoded 9-entry timezone list in `SchedulePanel`; `Intl.supportedValuesOf("timeZone")` is a one-liner and there's no existing util to reuse.
- `ImportYaml` reads the uploaded file with no size guard — `accept=".yml,.yaml"` is only a hint.
- `border-left: var(--ks-border-width-thick, 4px)` — the token exists, so drop the px fallback (`FlowRecipe.vue:347`).

---

## Test coverage

Strong on the extracted logic (`recipeToYaml` 24 cases incl. round-trip through the real `flowYamlUtils`, `flowIdentifiers`, `shouldShowLanding`, `useFlowRecipe` incl. a state-leak test). Gaps:

- **`importYamlUtils.spec.ts` mocks `flowYamlUtils` entirely** (lines 3-17) with a hand-rolled `/^(\w+):\s*(.+)$/` parser. It exercises the wrapper's branching but proves nothing about real YAML — the `parse_error` and `invalid_mapping` paths in particular. Use the real parser here; it's synchronous and dependency-free.
- **`FlowCreate.vue` is untested** — the `showLanding` gate, the `overrideId`/`overrideNamespace`/`importYaml` plumbing, and the `RECIPE_PRESET_KEY` sessionStorage handoff are the integration seams this PR adds, and all three are uncovered. `SystemBlueprintsTab` has no test at all.
- The Storybook stories have no `play` functions, so they're render snapshots rather than the behavioural tests the description implies.
- No Playwright coverage for the funnel, despite it now sitting in front of the most-used creation path.

## Conventions

Commit/PR title conventions look right. One thing to settle before merge: this touches shared OSS code that EE overrides, so the EE companion PR needs to land simultaneously — and it's worth deciding whether any of this is backport-bound.

flcarre:
Thanks, this was a genuinely useful review. Both blockers were real and both are addressed. I checked each finding against the source rather than taking it on trust. Two of your diagnoses turned out to be slightly off in my favour, and I've noted those below so nobody re-litigates them later.

Pushed in `acdd95165f`, `3f70788186`, `77b591c50a`, `48ba5a7f63`.

## Blockers

**1. EE tabs breakage.** Confirmed exactly as you describe: `ui-ee/src/override/components/namespaces/useTabs.ts:41` spreads `...baseTabs`, which is now a `ComputedRef`. It's the only consumer of OSS `useHelpers()` in `ui-ee` (grepped), so the companion is a one-liner. Paired PR: https://github.com/kestra-io/kestra-ee/pull/9616. It needs to merge in lockstep with this one.

**2. Generated recipe YAML.** You were right that the property sets don't line up, and I fixed all three cases. Two corrections to the detail:

- **The FQCNs are all correct.** I checked each one against its plugin repo: `io.kestra.plugin.slack.notifications.{SlackExecution,SlackIncomingWebhook}` in `plugin-slack`, `io.kestra.plugin.microsoft365.teams.{TeamsExecution,TeamsIncomingWebhook}` in `plugin-microsoft365`, `io.kestra.plugin.email.MailSend` in `plugin-email`. All six exist.
- **Slack `channel` is valid on the execution path.** `SlackExecution extends SlackTemplate`, and `SlackTemplate` declares `channel` (`plugin-slack/.../SlackTemplate.java:38`), and the plugin's own second `@Example` uses `channel: "#errors"` with `SlackExecution`. It is *not* valid on `SlackIncomingWebhook`, which `SlackTemplate` extends rather than the reverse. So the bug was narrower than "both paths": only the non-execution path.

What changed:

| Channel / path | Before | Now |
|---|---|---|
| Slack, execution | `channel` | kept, valid via `SlackTemplate` |
| Slack, non-execution | `channel` + JSON `payload` | `messageText`, no `channel`; the channel input is only offered for execution triggers |
| Teams, non-execution | `message` | `payload` with a MessageCard, matching the plugin's own example |
| Teams, execution | `url` + `executionId` | unchanged (correct) |
| Email, both | `to: [list]`, no `host` | `to` as a single string, plus `host`/`port`/`username`/`password` from secrets |
| Email, execution | `MailSend` | `MailExecution`, so the channel matches the Slack and Teams pattern |

On "the unit tests assert the shape the code produces, so they can't catch this class of bug": fair, and that's the part worth fixing structurally. There's now a `notify task properties match the plugin schemas` block that asserts the *constraints* rather than the output: that `channel` is absent from the webhook variant, that Teams sends no `message`, that `to` is a string and not an array, that a host is present. Those fail if someone reintroduces any of the four.

## Should fix

- **Silent channel drop.** Fixed. `hasNotifyChannel` now intersects with `channelAvailability`, so `isValid` can't be true on a recipe whose only channel will be omitted, and the Notify step names the channels it's going to leave out instead of dropping them silently. The selection itself survives, so switching trigger type back restores it. Covered by a test that installs only the execution-trigger Slack task and flips the trigger type.
- **Imported fragment unsaveable.** Fixed. The import path sets `shouldApplyGeneratedMetadata`, and since `withGeneratedFlowMetadata` already guards on the keys being absent it's a no-op for a definition that carries its own. Two tests: metadata preserved when present, generated when absent. You were right that `parseImportYaml`'s `id`/`namespace` return is the wiring that got lost; it's still unused by the component, but it's now genuinely exercised by the spec rather than by a regex stub (see below).
- **Dead `blueprintDetail`.** Dropped. The flag only ever attached `onGoToDetail`, and `SystemBlueprintsTab` never emits it. In-place system-blueprint browsing being replaced by a link out is intentional; I've said so in the description.
- **`namespace` bypass.** Kept out of the allowlist on purpose, and I've reworded the description, which was the part that was wrong. The reasoning: the namespace is context, not an instruction to skip the funnel, and the landing now preselects it (`NewFlowLanding` reads `route.query.namespace` and offers it as an option even when it isn't in the fetched list). Thanks for catching `TourFinale.vue:107`, I'd only found the two Flows entry points. "Start building" landing on a form rather than the editor is the one call here I'd genuinely like a second opinion on; happy to add `namespace` to the allowlist if you'd rather the tour ends in the editor.
- **Dead translation keys.** `recipe.blueprints_section_title` and `recipe.summary.yaml_preview` removed from all 13 locales.

## Nits

Done: the duplicated 500-namespace fetch is now `useNamespaceOptions`; `"0 9 * * *"`, the default states, the Slack channel, the webhook key and the mail recipient are single exported constants; `getRandomID()` replaces the `Math.random` suffix; the EE-locked tile has a `null` type so it no longer collides with the real `"other"`; `showLanding` flips after the `await`; the timezone list comes from `Intl.supportedValuesOf`; the upload is rejected over 1 MB; the `4px` fallback is gone.

Not done, with reasons:

- **`SelectableTile` in the design system.** I agree, and it's the right end state. I've left it in feature code for this PR because moving it means a DS component, a story, a unit test and a token pass, and I'd rather not grow a 6k-line PR further. Happy to file it as a follow-up unless you consider it merge-blocking.
- **Props mutated as state.** Also agree in principle. `recipe` is one shared reactive object threaded through five panels; converting to `v-model` per field is a real refactor of the whole wizard, not a local fix. Same offer: follow-up issue.

## Test coverage

- `importYamlUtils.spec.ts` now runs against the real `flowYamlUtils` parser. You were right that the regex stub proved nothing. Switching to the real parser is what makes `parse_error` and `invalid_mapping` mean something, and it turned up that duplicate keys and an unclosed mapping both land in `parse_error` while a bare scalar lands in `invalid_mapping`.
- `FlowCreate.spec.ts` is new: the landing gate per query param, the `overrideId`/`overrideNamespace`/`importYaml` plumbing, metadata generation on import, and the `RECIPE_PRESET_KEY` hand-off. `SystemBlueprintsTab.spec.ts` is new too.
- One thing worth flagging from your develop merge: `#17743`'s leak guard caught my first version of `FlowCreate.spec.ts`. It mocked `stores/flow`, and with `isolate: false` that mock was served to `flowOutdatedSave.spec.ts` and `Actions.spec.ts` from the shared registry and broke them. It now drives the real store from a fresh pinia and spies on `initYamlSource`. Good guard.
- Playwright coverage for the funnel: agreed, but out of scope here.

Green locally: 1077 unit, 187 Storybook, `check:types`, ESLint. The translation check has 117 pre-existing misses on `develop` (`onboarding.tour.*`, `test_event.*` from #17720). None from this PR, and all 116 keys it adds are present in every locale.

MilosPaunovic (commented):
Second pass on `5846f526`. Reviewed the four amendment commits (`acdd95165f`, `3f70788186`, `77b591c50a`, `48ba5a7f63`) plus the develop merge, and checked the claims against plugin source rather than taking them on trust.

## Both blockers cleared

**EE tabs.** kestra-io/kestra-ee#9616 is exactly the one-liner needed (`...baseTabs.value`, inside the existing `computed`). Confirmed it's the only `useHelpers()` consumer in `ui-ee`.

⚠️ **One thing to flag on that PR**: it can't merge first. Against current OSS `develop`, `baseTabs` is still a plain array, so `...baseTabs.value` becomes `[...undefined]` — which throws too. EE `check:types` and `namespaceUseTabs.spec.ts` will be red on #9616 until this PR lands. Worth a note over there so nobody "fixes" the red by reverting it.

**Recipe YAML.** Your two corrections are right, and mine was the sloppy half:

- `SlackExecution extends SlackTemplate` — confirmed at `SlackExecution.java:93`, and the plugin's own `@Example` at `:68` uses `channel: "#errors"`. `channel` is valid on the execution path; I'd generalised from the rendered docs page, which doesn't list inherited properties. `SlackIncomingWebhook extends AbstractSlackWebhookConnection` and declares `messageText` at `:228`, no `channel` — so the narrower diagnosis was correct.
- All six FQCNs exist; `io.kestra.plugin.email.MailExecution` is real (`MailExecution extends MailTemplate extends MailSend implements ExecutionInterface`).

The email fix holds up on details I hadn't checked and would otherwise have flagged:

- `MailSend.to` is `Property<String>` (`:261`) — the string is right, the array was wrong.
- `transportStrategy` defaults to `SMTPS` (`:223`), so `port: 465` matches the default. If it had defaulted to plain `SMTP`, 465 would have been a silent runtime hang.
- `MailExecution` overrides `htmlTextContent` (`:90`), so omitting it on the execution path is correct rather than incomplete.

The `notify task properties match the plugin schemas` block is the right structural answer — asserting absence (`channel` undefined on the webhook variant, no `message` on Teams, `to` not an array) is what makes this class of bug catchable.

## Everything else verified fixed

`hasNotifyChannel` now intersects with `channelAvailability` and the Notify step names what it's dropping; the import path sets `shouldApplyGeneratedMetadata`; `blueprintDetail` is gone; both dead keys are gone; the `too_large` key is present with a real translation in all 13 locales (checked each). Every nit landed — `useNamespaceOptions`, the exported constants, `getRandomID()`, `type: null` for the EE-locked tile with a guard in `selectTrigger`, the flag flipped after the `await`, `Intl.supportedValuesOf`, the 1 MB cap, the `4px` fallback.

Independently on the branch: **1120 unit tests / 133 files pass**, `check:types` green across design-system + app + test, `test:lint` **0 errors** (3 warnings, all pre-existing in `Vars.vue`/`Restart.vue`), `translations:check` reports no extra keys and nothing missing under `recipe.*` or `new_flow_landing.*`. `FlowCreate.spec.ts` drives the real store off a fresh pinia — no leak into the shared registry.

## Your open question — TourFinale

I'd keep `namespace` out of the allowlist as you have it, and have `TourFinale` opt out explicitly instead. The tour is a guided hand-off that ends "now build one yourself"; dropping the user on a chooser after twenty minutes of guidance is a step backwards, but that's specific to the tour, not to `namespace` — the two `Flows.vue`/`Actions.vue` entry points genuinely do want the funnel. A dedicated bypass key on that one push gets both.

## Small remainders (none blocking)

- **`recipeToYaml.spec.ts`**: `expect(Object.keys(webhookTask)).toEqual(["id", "type", "url", "messageText"])` is order-sensitive — reordering `buildSlackTask` breaks it for no behavioural reason. `expect.arrayContaining` plus an explicit `not.toHaveProperty("channel")` says the same thing without the coupling.
- **`recipe.then.unavailable_channels_warning`** reads "*Slack, Microsoft Teams* will be left out: the plugin is not installed for this trigger type" — singular noun over an interpolated list. Either `<i18n-t>` with a count, or reword to "the required plugins are not installed".
- **Stacked warnings**: when *every* selected channel is unavailable, the dropped-channels alert and the no-channel alert both render. Cosmetic.
- `SelectableTile` → design system, and the `v-model`-per-field refactor, as follow-ups is the right call for a PR this size.

LGTM once https://github.com/kestra-io/kestra-ee/pull/9616 is queued behind this one.

MilosPaunovic (approved):
Third pass, covering the seven branch commits since `5846f526` (`0766536e61` through `48af7a1302`) plus the two develop merges. Checked the effective merge-base diff rather than the per-commit noise: 53 frontend-only files, exactly what the description claims.

## All remainders from the second pass are fixed

- **Order-sensitive webhook assertion**: now `expect.arrayContaining` plus an explicit `not.toHaveProperty("channel")`, which keeps the property the test exists to guard without coupling to key order.
- **Singular-over-list warning**: reworded so the channel list leads and nothing agrees with its cardinality. Nicer than the `<i18n-t>` count option, since it avoids hand-writing the plural forms Russian and Polish would need.
- **Stacked warnings**: the generic no-channel alert is suppressed while the dropped-channel alert shows; Next stays disabled either way.
- **TourFinale**: resolved with a dedicated `blank` bypass instead of allowlisting `namespace`, keeping the tour namespace preselected. Exactly the split the second pass argued for - and reusing the same key the system-tab escape hatch needed anyway is tidier than a tour-specific one.

## New work verified

- **`blank` bypass**: sound. The key is consumed nowhere in `setupFlow` - it only steps aside from the landing, and setup falls through to the starter template with `query.namespace` honored. The system-tab link has a `?? "system"` fallback if the config has not arrived yet. Covered by the allowlist spec and a new SystemBlueprintsTab spec.
- **E2E funnel walk**: walking the blank-flow form as a user would, rather than bypassing with a query flag, keeps the real entry point covered. One non-issue checked: `` `flow_${testUUID}`.slice(0, 19) `` always ends on an underscore (uuid index 13 is a separator), but a trailing `_` is valid per both the funnel's `FLOW_ID_PATTERN` and the backend regex.
- **Escaped `@`**: end state verified - all 13 locales carry `{'@'}` in `email_to_placeholder`, the fingerprint entry (`e9f27946c09e`) matches the sha256 of the current escaped English value so the key will not churn on the next regen, and the generator prompt now treats `{'...'}` blocks as literals. Since the tooling is shared from OSS, the prompt fix propagates to EE for free.
- **The regen commit** (`53f36e803f`) nets out to +151 additive lines per locale against the merge base - develop got its own regen via the fingerprinting PR, so the churn cancels. `en.json` is untouched by it and the parity/placeholder/drift gate is green.

## Merge order

Unchanged from the second pass: this PR first, then https://github.com/kestra-io/kestra-ee/pull/9616 immediately after. Its red "Frontend - Tests / Check" is the expected pre-merge coupling (EE spreads `baseTabs.value`, which does not exist on OSS develop yet) and the failing run predates yesterday's push here - it needs a re-trigger once this lands.

No new bot findings, no unresolved threads. LGTM.

PR #17388, opened 2026-07-13
[WIP] feat: AI Copilot v2 agentic loop

ammeek:
Closes https://github.com/kestra-io/kestra-ee/issues/7910.
Closes https://github.com/kestra-io/kestra-ee/issues/7911.
Closes https://github.com/kestra-io/kestra-ee/issues/7912.
Closes https://github.com/kestra-io/kestra-ee/issues/7913.

Introduces AI Copilot v2 — a from-scratch, in-process agentic loop for the Kestra chat surface, replacing one-shot generation with a tool-driven orchestrator. This PR delivers Weeks 1 & 2 of the KIP: the core architecture (validated end-to-end), the read/authoring tool families, permission scoping, and the first chat UI.

Core loop & data model (Week 1)
- Agent Orchestrator — a hand-written low-level loop over ChatModel/StreamingChatModel on a bounded executor (not AiServices), bound to a provider via providerId (Provider Registry unchanged).
- Thread + Message model — AgentThread / AgentMessage with ToolCall shape; single-node in-memory stores (InMemoryThreadStore, InMemoryMessageStore) behind ThreadStore/MessageStore interfaces. Status transitions IDLE → RUNNING → AWAITING_CONFIRMATION → IDLE persist to the durable message log, making a turn resumable.
- Mode Profiles — one code path for all three modes; each mode is a system-prompt persona (ask.md / edit.md / plan.md) + a tool allow-list by family + a write policy.
- Tool Catalog + AiPlatformTool — @Tool/@P reflection produces the ToolSpecification (no hand-written schema), with a single dispatch entry point.
- Confirmation Gate — suspends on MUTATE/ACT; APPROVE resumes and dispatches, REJECT records a rejected TOOL_RESULT (Edit resumes, Plan aborts) with an optional steering reason.
- Plan-mode plan card — PROPOSED_ACTION with a null toolCall, approve-then-per-step-confirm, reusing the confirm machinery (no separate endpoint).
- Copilot Gate — existing RBAC gate guarding the chat endpoint.

Tools (Weeks 1 & 2)
- Read family (grounds Ask/Edit + authoring): read-execution, list-executions, read-flow, list-flows, search-plugins, get-plugin-schema, validate-flow, read-execution-logs.
- Act/mutate: restart-execution.
- Ask docs Q&A via the Kestra.io MCP (agent-as-MCP-client) for docs-grounded answers, plus execution-failure analysis over read-execution + read-execution-logs.
- Authoring sub-agents — author-flow, author-dashboard reuse the existing PluginFinder → restricted schema → YAML Builder → validate pipeline verbatim and return a non-mutating Draft (emitted as artefact_draft).
- Permission scoping — specsFor(mode, principal) coarse pre-filter of the advertised tool set (UX/token only) + per-call RBAC at dispatch; denial surfaces as a rejected tool result.


webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/AiPlatformTool.java
```diff
@@ -0,0 +1,15 @@
+package io.kestra.webserver.services.ai.agent.tool;
+
+import io.kestra.webserver.services.ai.agent.domain.AgentToolFamily;
+import io.kestra.webserver.services.ai.agent.domain.AgentWritePolicy;
+
+/**
+ * A native platform tool — a deterministic platform call. OSS declares no permissions — EE extends
```
fhussonnois: Let's avoid EE implementation details in comment - this comment applied to multiple classes in the PR

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/ReadFlowTool.java
```diff
@@ -0,0 +1,56 @@
...
+        @P(name = "namespace", value = "The namespace of the flow") String namespace,
+        @P(name = "flowId", value = "The id of the flow") String flowId,
+        @P(name = "revision", value = "Optional flow revision to read; omit to read the latest revision", required = false) Integer revision,
+        @TenantId @P(name = "tenantId", value = "The tenant to run against; omit to use your current tenant", required = false) String tenantId) {
+        String tenant = AgentCallContext.resolveTenant(tenantId);
+
+        FlowWithSource flow = flowRepository.findByIdWithSource(tenant, namespace, flowId, Optional.ofNullable(revision), false)
+            .orElseThrow(
+                () -> new IllegalArgumentException(
+                    "Flow not found: '" + namespace + "." + flowId + "'"
```
fhussonnois: Use `"".formatted(String...)` for readability

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/RestartExecutionTool.java
```diff
@@ -0,0 +1,68 @@
...
+        @P(name = "executionId", value = "The id of the execution to restart") String executionId,
+        @P(name = "revision", value = "Optional flow revision to restart with; omit to use the execution's own revision", required = false) Integer revision,
+        @TenantId @P(name = "tenantId", value = "The tenant to run against; omit to use your current tenant", required = false) String tenantId) {
+        String tenant = AgentCallContext.resolveTenant(tenantId);
+
+        Execution execution = executionRepository.findById(tenant, executionId)
+            .orElseThrow(() -> new IllegalArgumentException("Execution not found: '" + executionId + "'"));
+
+        if (!execution.getState().canBeRestarted()) {
+            throw new IllegalStateException(
```
fhussonnois, line 67: Same, use "".formatted(String...) for readability

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/ReadExecutionTool.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    public AgentWritePolicy writePolicy() {
+        return AgentWritePolicy.AUTO;
+    }
+
+    @Tool(
+        name = "read-execution",
+        value = "Read a single Kestra execution: its state, duration and per-task-run breakdown (failed runs include their state history). Read-only; use this to diagnose or inspect a specific run when you already know its id."
+    )
+    public String readExecution(
```
fhussonnois: Why are you returning a `String` ? Please use strong typed type for Tool. Langchain4J is supposed to handle serialization properly.

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/RestartExecutionTool.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        String operationId = IdUtils.create();
+        try {
+            executionCommandQueue.emit(Restart.from(execution, revision).withOperationId(operationId));
+        } catch (QueueException e) {
+            throw new IllegalStateException("Failed to enqueue restart for execution '" + executionId + "': " + e.getMessage(), e);
+        }
+
+        return "Restart requested for execution '" + executionId + "' (operationId=" + operationId + ").";
```
fhussonnois: Same here: I would prefer a strongly-typed object - no text-log

fhussonnois (changes requested):
Tools should return strongly-typed object, i.e., List, Object, etc. langchain4j handles the serialization for us that's why the ToolExecutor always return String

webserver/src/main/java/io/kestra/webserver/services/ai/agent/AiThreadManager.java
```diff
@@ -0,0 +1,143 @@
...
+                .createdAt(Instant.now())
+                .build()
+        );
+    }
+
+    public void appendCancelled(final String threadId, final String traceId) {
+        append(threadId, traceId, AgentMessageRole.SYSTEM, AgentMessageType.CANCELLED, null, null, null);
+    }
+
+    private String deriveTitle(final String threadId) {
```
fhussonnois: We should probably use the LLM to rename the session directly with a custom internal tool.

webserver/src/main/java/io/kestra/webserver/services/ai/agent/AgentOrchestrator.java
```diff
@@ -0,0 +1,382 @@
...
+        String traceId = thread.uid() + "-turn-" + (threadManager.load(thread.uid()).size() + 1);
+        try {
+            ResolvedProfile profile = modeProfiles.resolve(context.mode(), context.tenant(), context.principal());
+            StreamingChatModel model = aiServiceManager.getAiService(context.providerId()).streamingChatModel(List.of());
+
+            threadManager.appendUser(thread.uid(), traceId, context.prompt());
+
+            List<ChatMessage> messages = new ArrayList<>();
+            messages.add(SystemMessage.from(profile.systemPrompt()));
+            messages.addAll(ChatMessageAdaptor.project(threadManager.load(thread.uid())));
```
fhussonnois: Here you can do the projection first, then initialize the messages with projection+size

fhussonnois (changes requested):

webserver/src/main/java/io/kestra/webserver/services/ai/agent/ConfirmationRegistry.java
```diff
@@ -0,0 +1,20 @@
+package io.kestra.webserver.services.ai.agent;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+import jakarta.inject.Singleton;
+
+@Singleton
+public class ConfirmationRegistry {
```
fhussonnois: This class should probably not exist - we can't maintain a in-èmemory state for suspened turns. This will never work in production. I think I already did a review on that class ? Only MessageStore/ThreadStore should be used for replaying a conversation.

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/RestartExecutionTool.java
```diff
@@ -0,0 +1,68 @@
...
+    public AgentWritePolicy writePolicy() {
+        return AgentWritePolicy.CONFIRM;
+    }
+
+    @Tool(name = "restart-execution", value = "Restart a failed or paused Kestra execution from its failed tasks, creating a new run. The execution must be in a terminated or paused state.")
+    public String restartExecution(
+        @P(name = "executionId", value = "The id of the execution to restart") String executionId,
+        @P(name = "revision", value = "Optional flow revision to restart with; omit to use the execution's own revision", required = false) Integer revision,
+        @TenantId @P(name = "tenantId", value = "The tenant to run against; omit to use your current tenant", required = false) String tenantId) {
+        String tenant = AgentCallContext.resolveTenant(tenantId);
```
fhussonnois: I don't understand why you still have this ? if the tenant is passed throgh tool args why you still have to resolve stomething ?

fhussonnois: Ok, so the tenant is a property of the authenticated conversation, never of a tool call. It enters exactly once at the controller, from `tenantService.resolveTenant() ` and goes to Context.tenant(), and every tool reads it from there. Let's have a conversation is single-tenant by construction otherwise it's too error prone with potential security risk. So drop: `@]TenantId` and only use `AgentCallContext`, you probably need to create a dedicated `@CallContext `annotation to hide the param from the LLM.

fhussonnois (changes requested):

webserver/src/main/java/io/kestra/webserver/services/ai/agent/AgentCallContext.java
```diff
@@ -0,0 +1,69 @@
...
+
+import io.micronaut.core.annotation.Nullable;
+
+/**
+ * Per-dispatch context bound to the executor thread while a tool call runs, so {@code @Tool} methods
+ * can read the caller's tenant and principal, the turn's provider and conversation, and publish
+ * artefact drafts back to the turn without depending on the orchestrator.
+ */
+public final class AgentCallContext {
+    private static final ThreadLocal<Context> CURRENT = new ThreadLocal<>();
```
fhussonnois: No thread-local, this is a code-smell!

Could you please have a look at this langchain method : `https://github.com/langchain4j/langchain4j/blob/main/langchain4j/src/main/java/dev/langchain4j/service/tool/ToolExecutor.java#L36` we should be able to pass the CallContext using the `InvocationContext`. This should also allow us to simplify a lot the  `QueryFilterToolExecutor` and have a cleaner implementation that relies more on the langchain4J `DefaultToolExecutor`.

Then, we should be able to do something like in each tool: 

```
     public String restartExecution(
         @P(name = "executionId", value = "The id of the execution to restart") String executionId,
         @P(name = "revision", value = "Optional flow revision to restart with; omit to use the execution's own revision", required = false) Integer revision,
        InvocationParameters params) {
        String tenant = AgentCallContext.from(params).resolveTenant(tenantId);
}
```

fhussonnois (changes requested):

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/RestartExecutionTool.java
ammeek: Sounds good, just checking that we are happy with the user only being able to access resources from the current tenant they are using. In a conversation we had last week you suggested that a super user may want to access resources from multiple tenants at the same time.

core/src/main/java/io/kestra/core/ai/agent/models/AgentMode.java
```diff
@@ -0,0 +1,41 @@
...
+ * <ul>
+ * <li>{@link #ASK} — read + (non-mutating) authoring tools only; never mutates.</li>
+ * <li>{@link #EDIT} — read + mutation of the in-focus artefact; each write is confirmed.</li>
+ * <li>{@link #PLAN} — cumulative over Edit and adds act tools; multi-step, confirmed per step.</li>
+ * </ul>
+ */
+public enum AgentMode {
+    ASK(EnumSet.of(AgentToolFamily.READ)),
+    EDIT(EnumSet.of(AgentToolFamily.READ, AgentToolFamily.MUTATE)),
+    PLAN(EnumSet.of(AgentToolFamily.READ, AgentToolFamily.MUTATE, AgentToolFamily.ACT));
```
fhussonnois: There is an error: `AgentToolFamily.ACT` is for `EDIT` not `PLAN`

core/src/main/java/io/kestra/core/ai/agent/models/AgentPrincipal.java
```diff
@@ -0,0 +1,4 @@
+package io.kestra.core.ai.agent.models;
+
+public interface AgentPrincipal {
```
fhussonnois, line 8: Add a javdoc here - what is the purpose of this interface

core/src/main/java/io/kestra/core/ai/agent/models/AgentThreadStatus.java
```diff
@@ -0,0 +1,7 @@
+package io.kestra.core.ai.agent.models;
+
+public enum AgentThreadStatus {
+    IDLE,
```
fhussonnois, line 5: Add one line javadoc for each enum no prose

core/src/main/java/io/kestra/core/ai/agent/AgentThread.java
```diff
@@ -0,0 +1,74 @@
...
+
+import io.micronaut.core.annotation.Nullable;
+import lombok.Builder;
+import lombok.With;
+
+/**
+ * The durable unit of Copilot conversation memory: one row in the {@code agent_thread} table. Holds the
+ * metadata/governance record for a conversation — never the messages themselves (those are appended to
+ * {@code agent_message}). Keyed by its globally-unique {@link #uid()} and scoped by {@link #tenant()};
+ * {@link #userId()} is the owning principal (populated in EE, implicit in OSS). Soft-deleted rather
```
fhussonnois: Remove EE/OSS implementation details

jdbc-h2/src/main/java/io/kestra/repository/h2/migration/V2_0_17AiCopilotMigration.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * OSS H2 AI Copilot migration script.
+ *
+ * <p>
+ * Creates the {@code agent_thread} and {@code agent_message} tables backing the Copilot conversation
+ * store. Activates only when H2 (or the in-memory H2) is the repository backend.
+ */
+@Singleton
+@Requires(property = "kestra.repository.type", pattern = "h2|memory")
+public class V2_0_17AiCopilotMigration extends AbstractSQLMigrationScript {
```
fhussonnois: Double check if the script number is still OK with  `develop` branch

core/src/main/java/io/kestra/core/ai/agent/repositories/AiThreadRepositoryInterface.java
```diff
@@ -0,0 +1,93 @@
...
+     * Threads are private to the user that created them (the KIP {@code (tenant, userId, uid)} key),
+     * so this returns empty when the thread exists but is owned by a different user — callers use it
+     * to enforce per-user access without leaking a thread's existence.
+     *
+     * @param tenant the tenant the thread belongs to.
+     * @param userId the user that must own the thread.
+     * @param uid the unique identifier of the thread.
+     * @return the thread, or an empty {@link Optional} if it does not exist, is deleted, or is owned by another user.
+     */
+    Optional<AgentThread> find(String tenant, String userId, String uid);
```
fhussonnois: Do we really need that method for doing the user-check ?

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcAgentThreadRepository.java
```diff
@@ -0,0 +1,98 @@
...
+     * Lists the tenant's non-deleted threads (the default filter excludes deleted rows) and keeps only
+     * those owned by {@code userId}. The {@code agent_thread} table has no {@code user_id} column, so
+     * ownership is matched against the deserialized record rather than in SQL — mirroring
+     * {@link #find(String, String, String)}.
+     * {@inheritDoc}
+     */
+    @Override
+    public List<AgentThread> findAllForUser(String tenant, String userId) {
+        return find(tenant, DSL.trueCondition()).stream()
+            .filter(thread -> Objects.equals(thread.userId(), userId))
```
fhussonnois: Add the `user_id`column to the table avoding in-memory filtering and add an index: e.g.:
```
CREATE INDEX IF NOT EXISTS ix_agent_thread__tenant_deleted_user
    ON agent_thread ("tenant_id", "deleted", "user_id");
```

jdbc-postgres/src/main/resources/migrations/2.0.17-ai-copilot-postgres.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE IF NOT EXISTS agent_thread (
+    "key"       VARCHAR(250) NOT NULL PRIMARY KEY,
+    "value"     JSONB NOT NULL,
+    "tenant_id" VARCHAR(150) GENERATED ALWAYS AS (value ->> 'tenant') STORED,
+    "deleted"   BOOLEAN NOT NULL GENERATED ALWAYS AS (CAST(value ->> 'deleted' AS BOOLEAN)) STORED,
+    "created"   TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updated"   TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+CREATE INDEX IF NOT EXISTS agent_thread__tenant_deleted ON agent_thread ("tenant_id", "deleted");
```
fhussonnois: This index can be removed once you add the new one with `user_id` column, same for other DB

core/src/main/java/io/kestra/core/ai/agent/models/AgentMessage.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * One append-only step of a Copilot turn: one row in the {@code agent_message} table, linked to its
+ * parent thread by {@link #threadId()}. Rows are only ever inserted; a thread's history is loaded
+ * ordered by its monotonic {@link #uid()}, so messages always load in the order they were appended
+ * (e.g. a tool call always loads before its result).
+ */
+@Builder
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record AgentMessage(
```
fhussonnois, line 22: This class should be in `model` package

core/src/main/java/io/kestra/core/ai/agent/models/AgentThread.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * The durable unit of Copilot conversation memory: one row in the {@code agent_thread} table. Holds the
+ * metadata/governance record for a conversation — never the messages themselves (those are appended to
+ * {@code agent_message}). Keyed by its globally-unique {@link #uid()} and scoped by {@link #tenant()};
+ * {@link #userId()} is the owning principal (populated in EE, implicit in OSS). Soft-deleted rather
+ * than hard-deleted so history is retained until purge.
+ */
+@Builder(toBuilder = true)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record AgentThread(
```
fhussonnois, line 24: This class should be in `model` package

core/src/main/java/io/kestra/core/ai/agent/models/AgentMessage.java
fhussonnois, line 22: Add a tenant field into that class, and add the coresponding `tenant_id` column to align with Kestra convention and also to allow us purging all messages based only on the tenant if necesarry without performing a join on the thread table

core/src/main/java/io/kestra/core/ai/agent/repositories/AiMessageRepositoryInterface.java
```diff
@@ -0,0 +1,23 @@
...
+     */
+    AgentMessage append(AgentMessage message);
+
+    /**
+     * Loads the full message history of a thread, ordered chronologically by uid.
+     *
+     * @param threadId the unique identifier of the thread.
+     * @return the thread's messages, or an empty list if the thread has none.
+     */
+    List<AgentMessage> load(String threadId);
```
fhussonnois: Add `tenantId` to the method to enforce tenant isolation

jdbc-h2/src/main/resources/migrations/2.0.17-ai-copilot-h2.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE IF NOT EXISTS agent_thread (
```
fhussonnois: table should be prefixed with `ai_` same for all databases and the message table

fhussonnois (changes requested):

webserver/src/main/java/io/kestra/webserver/services/ai/agent/internals/TurnWindow.java
```diff
@@ -0,0 +1,52 @@
...
+        LinkedHashSet<String> turns = new LinkedHashSet<>();
+        for (AgentMessage message : history) {
+            turns.add(message.traceId());
+        }
+        if (turns.size() <= maxTurns) {
+            return history;
+        }
+
+        // Keep only the trailing maxTurns traceIds.
+        Set<String> kept = turns.stream().skip(turns.size() - (long) maxTurns).collect(java.util.stream.Collectors.toSet());
```
fhussonnois: Import `java.util.stream.Collectors`

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/AuthorDashboardTool.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Authoring sub-agent for dashboards: runs the existing one-shot generation pipeline
+ * (plugin finder → restricted schema → YAML builder) through {@link AiServiceInterface#generateDashboard},
+ * validates the result, and publishes it as a non-mutating {@link ArtefactDraft}. Nothing is saved —
+ * the draft is surfaced to the user for review and apply.
+ */
+@Singleton
+@Requires(bean = AiServiceManager.class)
+public class AuthorDashboardTool implements AiAuthoringTool {
```
fhussonnois, line 33: Move this tool to EE, custom dashboards are moving to EE

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/AiTool.java
```diff
@@ -0,0 +1,16 @@
...
+ * Common contract of everything the tool catalog can register: the object carrying the single
+ * {@code @Tool}-annotated method the model-facing specification is derived from.
+ *
+ * <p>
+ * Tools here carry no authorization logic. A replacement bean can override the same {@code @Tool}
+ * method to check the caller's access, then delegate to {@code super}.
+ * </p>
+ */
+public interface AiTool {
+    default Object toolInstance() {
```
fhussonnois: For me this method is not justified. If all tool return `this` the nyou don't need it.

fhussonnois (changes requested):

webserver/src/main/java/io/kestra/webserver/services/ai/agent/AgentOrchestrator.java
```diff
@@ -0,0 +1,514 @@
...
+        this.modeProfiles = modeProfiles;
+        this.threadManager = threadManager;
+        this.modelCallTimeout = configuration.modelCallTimeout();
+        this.maxContextTurns = configuration.maxContextTurns();
+        this.maxSequentialToolsInvocations = configuration.maxSequentialToolsInvocations();
+    }
+
+    public void runTurn(final AgentTurnContext context, final TurnEventSink sink) {
+        AgentThread thread = context.thread();
+        String traceId = thread.uid() + "-turn-" + (threadManager.load(thread.uid()).size() + 1);
```
fhussonnois: Just use: `String traceId = IdUtils.create();`, no need to load the the entire message history for generating an id

webserver/src/main/java/io/kestra/webserver/services/ai/agent/AgentOrchestrator.java
```diff
@@ -0,0 +1,514 @@
...
+            List<ChatMessage> projected = ChatMessageAdaptor.project(TurnWindow.lastNTurns(threadManager.load(thread.uid()), maxContextTurns));
+            List<ChatMessage> messages = new ArrayList<>(projected.size() + 2);
+            messages.add(SystemMessage.from(profile.systemPrompt()));
+            messages.addAll(projected);
+            // Caller-supplied context for this turn only: appended at the end so the model sees it as the
+            // latest input, but never persisted to the thread history (it is not appended via threadManager).
+            additionalContextMessage(context.additionalContext()).ifPresent(messages::add);
+
+            runLoop(
+                new AgentLoopContext(
```
fhussonnois, line 107: Move the ` new AgentLoopContext(...)` on a dedicated line for readibility

webserver/src/main/java/io/kestra/webserver/services/ai/agent/AgentOrchestrator.java
```diff
@@ -0,0 +1,514 @@
...
+        log.warn("Copilot turn for thread {} hit the sequential tool-invocation cap ({})", ctx.thread().uid(), maxSequentialToolsInvocations);
+        String message = "I reached the maximum number of tool steps (%d) for this turn and stopped before finishing. Ask me to continue if you'd like me to keep going."
+            .formatted(maxSequentialToolsInvocations);
+        threadManager.appendAssistantText(ctx.thread().uid(), ctx.traceId(), message);
+        sink.emit(AgentEvents.TOKEN, new AgentEvents.TokenEvent(message));
+        finishTurn(ctx);
+        done(sink, AgentThreadStatus.IDLE);
+    }
+
+    private static int toolInvocationsInTurn(final List<AgentMessage> log, final String traceId) {
```
fhussonnois, line 483: The count is wrong. In `runLoop` method you increment +1 per turn containing toolExecutionRequest and that's OK. But here, each individual tool_call is recorded. 

You can: 
1. Change the guard from "increment per round" to "check budget before the round" (line 279–284):
```
  // The model requested tools. Cap the number of tool invocations in a turn to guard against a
  // runaway reasoning loop . On exhaustion the turn ends gracefully, not failed. The check is per
  // round (before any of the round's tools run) so a tool-call/result pair is never split; the
  // count is bumped per persisted tool call, so a resume rebuilds the exact same number.
  if (ctx.toolInvocations().get() >= maxSequentialToolsInvocations) {
      stopForToolBudget(ctx, sink);
      return;
  }
  ctx.messages().add(ai);
```
2. Increment right where the row is persisted (right after line 300):
```
 threadManager.appendToolCall(ctx.thread().uid(), ctx.traceId(), ai.text(), ChatMessageAdaptor.toToolCall(req, entry.kind(), entry.family()));
ctx.toolInvocations().incrementAndGet();
```

webserver/src/main/java/io/kestra/webserver/services/ai/agent/AiThreadManager.java
```diff
@@ -0,0 +1,195 @@
...
+                .build()
+        );
+    }
+
+    /**
+     * Mints a monotonically-increasing message uid: a zero-padded per-node sequence prefix followed by
+     * a random id. Message history is loaded ordered by this uid, so the prefix guarantees messages
+     * sort in append order (a tool call always before its result).
+     */
+    private String newMessageUid() {
```
fhussonnois, line 195: This can't work with multiple webserver instance. Instead use a UUIDv7 we should provide enough in our context for ordering and uniqueness: see that Kestra class: `io.kestra.core.events.Event`

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/ToolCatalog.java
```diff
@@ -0,0 +1,212 @@
...
+            return tool != null;
+        }
+
+        /** Authoring tools are non-mutating drafts, advertised in every mode. */
+        public boolean isAuthoring() {
+            return kind == AgentToolCall.Kind.AUTHORING;
+        }
+    }
+
+    private Map<String, ToolEntry> registry() {
```
fhussonnois, line 96: You can perform the tool registration part of the instantiation of the object to remove the double-checked locking idiom.

fhussonnois (changes requested):

ui/src/components/ai/copilot/ProposedActionCard.vue
```diff
@@ -0,0 +1,235 @@
...
+    .proposed-action-arg-key {
+        flex: 0 0 auto;
+        min-width: 5rem;
+        color: var(--ks-text-secondary);
+    }
+
+    .proposed-action-arg-value {
+        margin: 0;
+        min-width: 0;
+        font-family: var(--font-family-monospace, monospace);
```
Piyush-r-bhaskar:
```suggestion
        font-family: var(--ks-font-family-mono);
```

Piyush-r-bhaskar (commented):
Would it be possible to cut the comments where it's not necessary, it's adding too much noise, 
and also classes seems too long - do you want to handle this in this iteration or next @MilosPaunovic ?

ui/src/components/ai/copilot/CopilotArtefactDraft.vue
```diff
@@ -0,0 +1,117 @@
...
+<script setup lang="ts">
+    import {computed} from "vue"
+    import {useI18n} from "vue-i18n"
+    import FileDocumentOutline from "vue-material-design-icons/FileDocumentOutline.vue"
+    import {useApplyDraft} from "./useApplyDraft"
+    import type {ArtefactDraftEvent} from "./types"
+
+    const props = defineProps<{draft: ArtefactDraftEvent}>()
+
+    const {t} = useI18n()
```
Piyush-r-bhaskar: instead of defining this let's use `$t`

Piyush-r-bhaskar (approved):
LGTM on FE side . Fantastic work @MilosPaunovic ❤️

fhussonnois (approved):
LGTM! @ammeek  before mering please fix the issue with the `ExecutorService` in the `AiAgentCopilot` class we discussed.

@Ben8t must be able to use the feature on the docker-image before merging ;)

PR #1116, opened 2023-03-29
fix(core,ui) Pause task

loicmathieu:
Fixes #1103
Fixes #1114

@yuri1969 I tested 10k flow executions and all ends up green with this patch so I think it will fix #1114.

My test flow is:

```yaml
id: wait-and-delay
namespace: io.kestra.tests


tasks:
  - id: wait-and-resume
    type: io.kestra.core.tasks.flows.Pause
    description: "Wait 10 sec and then resume"
    delay: PT10S
    tasks:
      - id: hello-world
        type: io.kestra.core.tasks.log.Log
        message: Hello World
```

As soon as this PR is merged this would be great if you can test it and assert your issues are fixed.

PR #17393, opened 2026-07-14
fix(core): correct findLastNotTerminated predicate to skip terminated and paused runs

anxkhn:
### ✨ Description

`Execution.findLastNotTerminated()` is meant to return the last task run that is still active (neither terminated nor paused), and several callers use it to pick the run they then mutate.

The filter predicate was a tautology:

```java
.filter(t -> !t.getState().isTerminated() || !t.getState().isPaused())
```

A state is never both terminated and paused: `State.Type.isTerminated()` is true for `FAILED, WARNING, SUCCESS, KILLED, CANCELLED, RETRIED, SKIPPED, RESUBMITTED`, while `isPaused()` is true only for `PAUSED`. Because the two are mutually exclusive, at least one operand of the `||` is always true, so the filter matched every element and `findFirst()` returned the last task run in the list regardless of its state, instead of the last still-active one.

This is a regression from #1116, which changed the original `!t.getState().isTerminated()` to `!t.getState().isTerminated() || !t.getState().isPaused()` to also skip paused runs. By De Morgan's law, "neither terminated nor paused" is `!terminated && !paused`, so the operator should be `&&`.

The fix swaps `||` for `&&` and keeps the existing `reversed().stream().findFirst()` shape that the method comment mandates for performance (see #14385):

```java
.filter(t -> !t.getState().isTerminated() && !t.getState().isPaused())
```

Why it matters: callers that select the last active run and then mutate its state can otherwise re-stamp an already terminated or paused run. The three non-test consumers are:
- `ExecutorService.markAs(...)` (used when killing or failing an execution) does `withStateAndAttempt(state)` on the returned run;
- `plugin.core.execution.Exit` transitions the returned run (and its parents) to the exit state;
- `Execution.failedExecutionFromExecutor(...)` attaches a failed attempt to the returned run.

Added regression tests cover a trailing terminated run, a trailing paused run, and an all-terminated list.

### 🔗 Related Issue

No existing issue. This is a self-found correctness bug in `Execution.findLastNotTerminated()`. Happy to open a tracking issue first if the maintainers prefer; the change is a one-operator fix plus tests.

### 🛠️ Backend Checklist

### 📝 Additional Notes

- The change is a single operator (`||` -> `&&`) on `Execution.java`, plus three colocated unit tests in `ExecutionTest`.
- Verified red before the fix: reverting the source to `||` makes exactly the three new tests fail with the expected assertions; with the fix the full `ExecutionTest` class is green.
- Ran the runner tests that consume this method to confirm no behavioral regression on the kill / pause / exit paths: `ExitTest` and `PauseTest` both pass.
- The method comment explicitly asks not to replace the `reversed().findFirst()` shape (performance, #14385); this PR keeps it.

---

## Verification commands (for reviewer / Anas)

```bash
export JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home
# Fix + new tests green:
./gradlew :core:test --tests "io.kestra.core.models.executions.ExecutionTest"
# Consumers unaffected:
./gradlew :core:test --tests "io.kestra.plugin.core.execution.ExitTest"
./gradlew :core:test --tests "*PauseTest"
```

### DIFF
--- a/core/src/main/java/io/kestra/core/models/executions/Execution.java
+++ b/core/src/main/java/io/kestra/core/models/executions/Execution.java
@@ -682,7 +682,7 @@ public class Execution implements SoftDeletable<Execution>, TenantInterface, Has
         return this.taskRunList
             .reversed()
             .stream()
-            .filter(t -> !t.getState().isTerminated() || !t.getState().isPaused())
+            .filter(t -> !t.getState().isTerminated() && !t.getState().isPaused())
             .findFirst();
     }
 
--- a/core/src/test/java/io/kestra/core/models/executions/ExecutionTest.java
+++ b/core/src/test/java/io/kestra/core/models/executions/ExecutionTest.java
@@ -203,6 +203,62 @@ class ExecutionTest {
     }
 
     @Test
+    void shouldFindLastNotTerminatedWhenLastTaskRunIsTerminated() {
+        // Given a task run list whose last run is terminated (SUCCESS) but is preceded by a running one
+        Execution execution = Execution.builder()
+            .taskRunList(List.of(
+                TaskRun.builder().id("running").state(new State(State.Type.RUNNING, new State())).build(),
+                TaskRun.builder().id("success").state(new State(State.Type.SUCCESS, new State())).build()
+            ))
+            .build();
+
+        // When looking for the last not terminated task run
+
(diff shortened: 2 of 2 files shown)


loicmathieu (approved):
LGTM, thanks

Issue #17093, opened 2026-06-26
[regression] LocalStorage.putFile() throws FileNotFoundException on STORE fetch type (JDBC plugin) — introduced in v1.3.24 OSS


CedricOB52857:
## Description

When using the JDBC SQL Server plugin (or any JDBC plugin) with `fetchType: STORE`, task execution fails with a `FileNotFoundException` inside `LocalStorage.putFile()`. The error occurs because the parent directories for the output `.ion` file are not created before the `FileOutputStream` is opened.

This is a **regression introduced in v1.3.24**. The same configuration works correctly on v1.3.23.

---

## Steps to reproduce

1. Deploy Kestra with `storage.type: local` and a local path (e.g. `/app/storage`)
2. Create a flow with a JDBC task (e.g. `io.kestra.plugin.jdbc.sqlserver.Query`) using `fetchType: STORE`
3. Trigger an execution
4. Observe task failure

---

## Expected behavior

The worker creates all necessary parent directories before writing the output file, as it did in v1.3.23.

---

## Actual behavior

Task fails immediately with:

```
java.io.FileNotFoundException: /app/storage/main///executions//tasks///.ion (No such file or directory)
	at java.base/java.io.FileOutputStream.open0(Native Method)
	at java.base/java.io.FileOutputStream.open(Unknown Source)
	at java.base/java.io.FileOutputStream.(Unknown Source)
	at io.kestra.storage.local.LocalStorage.putFile(LocalStorage.java:216)
	at io.kestra.storage.local.LocalStorage.put(LocalStorage.java:200)
	at io.kestra.core.storages.StorageInterface.put(StorageInterface.java:188)
	at io.kestra.core.storages.InternalStorage.putFile(InternalStorage.java:170)
	at io.kestra.plugin.jdbc.AbstractJdbcQuery.run(AbstractJdbcQuery.java:109)
```

---

## Environment

| Key | Value |
|-----|-------|
| Kestra version (broken) | v1.3.24, v1.3.25 |
| Kestra version (working) | v1.3.23 |
| Deployment | Docker Compose — standalone mode |
| Storage backend | `local` (mapped volume: `/mnt/kestra-storage → /app/storage`) |
| OS | Ubuntu 24.04 LTS |
| Plugin | `io.kestra.plugin.jdbc.sqlserver.Query` (reproducible with any JDBC plugin) |
| `fetchType` | `STORE` |

---

## Additional context

[...]

PR #17396, opened 2026-07-14
fix(storage): surface a descriptive error when local storage cannot create parent directories

nikhiln64:
### ✨ Description

`LocalStorage.putFile()` ignored the boolean result of `File#mkdirs`, so any failure to create the parent directory hierarchy surfaced later as a misleading `FileNotFoundException (No such file or directory)` from `FileOutputStream` pointing users at a path problem when the real cause was something else entirely.

This PR replaces the ignored `mkdirs()` call with `Files.createDirectories()`, which throws a descriptive exception naming the offending path:
- `AccessDeniedException` when the process lacks permission on an existing parent
- `FileAlreadyExistsException` when a regular file occupies a path segment

### 🔗 Related Issue

Relates to https://github.com/kestra-io/kestra/issues/17093 (diagnosability half; the UID regression itself was resolved by the base-image pin in v1.3.27).

### 🛠️ Backend Checklist

- [x] All unit and integration tests pass `./gradlew :storage-local:test` (67 tests, includes the inherited `StorageTestSuite` contract)

### 📝 Additional Notes

- The new test (`shouldFailWithDescriptiveErrorWhenParentDirectoryCannotBeCreated`) was verified red against the previous implementation (fails with `FileNotFoundException`) and green with the fix it deterministically forces the failure by placing a regular file where a parent directory must be created, so it doesn't depend on permission bits (which would be bypassed when CI runs as root).
- `createDirectoryFromPath()` already checks its `mkdirs()` result and throws; left untouched.
- Developed with AI assistance (Claude Code), reviewed and driven by a human. As requested by the template, a cat joke: why did the cat get assigned the storage bug? Because it's an expert at putting things in boxes. 🐱

### DIFF
--- a/storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java
+++ b/storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java
@@ -209,10 +209,11 @@ public class LocalStorage implements StorageInterface {
     }
 
     private static URI putFile(URI uri, StorageObject storageObject, File file) throws IOException {
-        File parent = file.getParentFile();
-        if (!parent.exists()) {
-            parent.mkdirs();
-        }
+        // Files.createDirectories throws a descriptive exception (e.g. AccessDeniedException,
+        // FileAlreadyExistsException) when the parent hierarchy cannot be created, unlike
+        // File#mkdirs whose boolean result was previously ignored and let the subsequent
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
LGTM, thanks

PR #17397, opened 2026-07-14
fix(core): bound the plugin documentation cache to stop unbounded heap growth

### LINKED ISSUE #16983: Unbounded static cache ClassPluginDocumentation.CACHE leaks heap as plugin docs are browsed
## Describe the bug

`core/src/main/java/io/kestra/core/docs/ClassPluginDocumentation.java` declares a process-lifetime static cache:

```java
private static final Map<PluginDocIdentifier, ClassPluginDocumentation<?>> CACHE = new ConcurrentHashMap<>();
```

The cache key includes the plugin class name **+ version** (`PluginDocIdentifier` builds `pluginClass.getName() + ":" + version`), and the values are large generated documentation objects (JSON schemas, `$defs`, outputs, metrics). It has **no `maximumSize` and no expiry**, and is never cleared — it lives for the entire JVM lifetime.

It is populated by the plugin-docs endpoints (`GET .../plugins/{cls}` and `GET .../plugins/{cls}/versions/{version}` in `webserver/.../controllers/api/PluginController.java`), which the editor hits every time a user opens a task. In EE, multiple versions of the same plugin can be installed, so the `version` dimension makes the keyspace grow as users browse plugins/versions over time → unbounded heap growth.

## Root cause

A `static`, process-lifetime `ConcurrentHashMap` with no eviction policy, keyed on an effectively unbounded keyspace (class + version), holding large values.

## Expected behavior

The documentation cache should be bounded (e.g. Caffeine with `maximumSize` and/or `expireAfterAccess`) so it cannot grow without limit.

## Proposed fix

Replace the raw `ConcurrentHashMap` with a bounded cache (Caffeine `maximumSize` + `expireAfterAccess`), consistent with other bounded caches in the codebase.

## Notes

- This is **on-heap** — it is visible in heap dumps / `jmap -histo`.
- The `version` cache-key dimension and the `{cls}/versions/{version}` endpoint are recent additions; this is primarily a concern on `develop`/2.0 and is not present in 1.3.x.

## Environment

- `develop`.

nikhiln64:
### ✨ Description

`ClassPluginDocumentation` declared a process-lifetime static cache with no bound:

```java
private static final Map<PluginDocIdentifier, ClassPluginDocumentation<?>> CACHE = new ConcurrentHashMap<>();
```

The key includes plugin class name + **version** + `allProperties`, and the values are large generated documentation objects (JSON schemas, `$defs`, outputs, metrics). The plugin-docs endpoints populate it every time a user opens a task in the editor, and with multiple plugin versions installed the keyspace is effectively unbounded, so heap grows for the entire JVM lifetime.

This PR replaces the raw `ConcurrentHashMap` with a Caffeine cache bounded by `maximumSize(1_000)` + `expireAfterAccess(1h)`, consistent with the existing bounded caches in core (`FlowWithDefaultCache`, `QueueLagPoller`). Caffeine is already a `core` dependency; the `of()` call switches from `computeIfAbsent` to the equivalent `Cache#get(key, mappingFunction)`, so caching semantics for callers are unchanged.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/16983.

### 🛠️ Backend Checklist

- [x] All unit and integration tests pass `./gradlew :core:test --tests ClassPluginDocumentationTest` (5 tests) and `--tests DocumentationGeneratorTest` (8 tests, the other consumer of this class)

### 📝 Additional Notes

- Added `shouldCachePerClassVersionAndAllProperties` covering the caching contract we own: same key returns the cached instance; a different version or `allProperties` flag generates a distinct object. Eviction/bounding itself is Caffeine's contract and is deliberately not re-tested (size-based eviction is async and would make the test flaky).
- Sizing rationale: OSS installs typically have hundreds of plugin classes, so 1,000 entries comfortably covers the hot set; `expireAfterAccess` reclaims memory on idle instances either way.
- Developed with AI assistance (Claude Code), reviewed and driven by a human. Template-mandated cat joke: this cache used to hoard like a cat with cardboard boxes, now it only keeps the 1,000 boxes it actually sits in. 🐱

### DIFF
--- a/core/src/main/java/io/kestra/core/docs/ClassPluginDocumentation.java
+++ b/core/src/main/java/io/kestra/core/docs/ClassPluginDocumentation.java
@@ -1,9 +1,12 @@
 package io.kestra.core.docs;
 
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
-import java.util.concurrent.ConcurrentHashMap;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 
 import io.kestra.core.plugins.PluginClassAndMetadata;
 
@@ -16,7 +19,14 @@ import lombok.ToString;
 @EqualsAndHashCode
 @ToString
 public class ClassPluginDocumentation<T> extends AbstractClassDocumentation<T> {
-    private static final Map<PluginDocIdentifier, ClassPluginDocumentation<?>> CACHE = new ConcurrentHashMap<>();
+    // Bounded: the keyspace (class name + version + allProperties) is effectively unbounded in EE
+    // where multiple plugin versions can be installed, and the generated documentation values are
+    // large (JSON schemas, $defs, outputs). An unbounded map here grows the heap for the whole JVM
+    // lifetime as users browse plugins/versions (see #16983).
+    private static final Cache<PluginDocIdentifier, ClassPluginDocumentation<?>> CACHE = Caffeine.newBuilder()
+        .maximumSize(1_000)
+        .expireAfterAccess(Duration.ofHours(1))
+        .build();
     private String icon;
     private String group;
     protected String docLicense;
@@ -87,7 +97,7 @@ public class ClassPluginDocumentation<T> extends AbstractClassDocumentation<T> {
 
     public static <T> ClassPluginDocumentation<T> of(JsonSchemaGenerator jsonSchemaGenerator, PluginClassAndMetadata<T> plugin, String version, boolean allProperties) {
         //noinspection unchecked
-        return (ClassPluginDocumentation<T>) CACHE.computeIfAbsent(
+        return (ClassPluginDocumentation<T>) CACHE.get(
             new PluginDocIdentifier(plugin.type(), version, allProperties),
             (key) -> new ClassPluginDocumentation<>(jsonSchemaGenerator, plugin, allProperties)
         );
--- a/core/src/main/java/io/kestra/core/docs/ClassPluginDocumentation.java
+++ b/core/src/main/java/io/kestra/core/docs/ClassPluginDocumentation.java
@@ -20,7 +20,7 @@ import lombok.ToString;
 @ToString
 public class ClassPluginDocumentation<T> extends AbstractClassDocumentation<T> {
     private static final int CACHE_MAXIMUM_SIZE = 1_000;
-    private static final Duration CACHE_EXPIRE_AFTER_ACCESS = Duration.ofHours(1);
+    private static final Duration CACHE_EXPIRE_AFTER_ACCESS = Duration.ofDays(1);
 
     // Bounded: the keyspace (class name + version + allProperties) is effectively unbounded in EE
     // where multiple plugin versions can be installed, and the generated documentation values are
--- a/core/src/main/java/io/kestra/core/docs/ClassPluginDocumentation.java
+++ b/core/src/main/java/io/kestra/core/docs/ClassPluginDocumentation.java
@@ -19,13 +19,16 @@ import lombok.ToString;
 @EqualsAndHashCode
 @ToString
 public class ClassPluginDocumentation<T> extends AbstractClassDocumentation<T> {
+    private static final int CACHE_MAXIMUM_SIZE = 1_000;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/kestra/core/docs/ClassPluginDocumentation.java
```diff
@@ -16,7 +19,14 @@
...
 @ToString
 public class ClassPluginDocumentation<T> extends AbstractClassDocumentation<T> {
-    private static final Map<PluginDocIdentifier, ClassPluginDocumentation<?>> CACHE = new ConcurrentHashMap<>();
+    // Bounded: the keyspace (class name + version + allProperties) is effectively unbounded in EE
+    // where multiple plugin versions can be installed, and the generated documentation values are
+    // large (JSON schemas, $defs, outputs). An unbounded map here grows the heap for the whole JVM
+    // lifetime as users browse plugins/versions (see #16983).
+    private static final Cache<PluginDocIdentifier, ClassPluginDocumentation<?>> CACHE = Caffeine.newBuilder()
+        .maximumSize(1_000)
+        .expireAfterAccess(Duration.ofHours(1))
```
nikhiln64: Instead of hardcoding move these constants into static variables.

nikhiln64: Extracted `CACHE_MAXIMUM_SIZE` and `CACHE_EXPIRE_AFTER_ACCESS` constants.

fdelbrayelle:
@MilosPaunovic This issue was reassigned to Core ownership so I'll change the reviewer, thanks.

nikhiln64:
Note on the red `Backend - Tests` check: all tests pass in the run (including `ClassPluginDocumentationTest` 5/5 and `DocumentationGeneratorTest` 8/8) - the failure is the `Report - Java` step, where `action-junit-report` can't create check runs from a fork PR's read-only token (`Resource not accessible by integration`, cf. mikepenz/action-junit-report#23). Test log itself is fully green.

flcarre (approved):

nikhiln64:
@flcarre thanks for the approval! The branch is rebased on the latest develop and the new CI run is waiting for maintainer approval to start. Whenever someone can approve the run this should be good to go.

nikhiln64:
One more gentle nudge on this one since it is approved and rebased but the CI runs are still waiting for maintainer approval to start. After this I will just leave it with the team.

core/src/main/java/io/kestra/core/docs/ClassPluginDocumentation.java
```diff
@@ -16,7 +19,17 @@
 @EqualsAndHashCode
 @ToString
 public class ClassPluginDocumentation<T> extends AbstractClassDocumentation<T> {
-    private static final Map<PluginDocIdentifier, ClassPluginDocumentation<?>> CACHE = new ConcurrentHashMap<>();
+    private static final int CACHE_MAXIMUM_SIZE = 1_000;
+    private static final Duration CACHE_EXPIRE_AFTER_ACCESS = Duration.ofHours(1);
```
loicmathieu:
```suggestion
    private static final Duration CACHE_EXPIRE_AFTER_ACCESS = Duration.ofDays(1);
```

I would cache for a higher value so you would typically never need to reload during a working day more than once.

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/docs/ClassPluginDocumentation.java
nikhiln64: Applied, expiry is now one day. Branch is also rebased onto the latest develop.

MilosPaunovic:
If you've addressed any of the requested changes from the PR review, @nikhiln64, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

nikhiln64:
Thanks for the pointer @MilosPaunovic, done. The one day expiry suggestion from the review is applied on the branch and the conversation is now resolved.

One note on the red Backend Tests check from last night's run. The summary shows tests: 4449, success: 4435, skipped: 14, failed: 0, and the task ended with Timeout has been exceeded, so every executed test passed and the job just ran over the Gradle task timeout on a slow runner. A rerun should come back green.

loicmathieu (approved):
LGTM

PR #17398, opened 2026-07-14
fix(executor): honor parent task logLevel for the subflow creation log

### LINKED ISSUE #16238: Subflow creation log is always emitted as INFO and cannot be filtered by plugin logLevel
#  Describe the issue

We are trying to control the logs stored in the `logs` table so that only WARN or higher level logs are persisted.

We configured the plugin default log level as follows:

```yaml
plugins:
  defaults:
    - type: io.kestra.plugin
      forced: true
      values:
        logLevel: WARN
```
However, when using io.kestra.plugin.core.flow.Subflow, one INFO-level log is still written even though the log level is configured as WARN.
From our investigation, this log is emitted when a subflow execution is created, before writing the related message to the queue.

The relevant code appears to be:
```
jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
```
Inside executionQueue, the log level is directly specified as:
```
.level(Level.INFO)
```
Because the level is hardcoded as INFO, it does not appear to be affected by the configured plugin logLevel.

As a result, even when users configure plugin logs to WARN or higher, this specific Subflow-related INFO log is still persisted in the logs table.

Could this behavior be changed so that this log follows the same log level filtering mechanism as other plugin logs?

Alternatively, if there is another recommended way to filter out this specific INFO log from the logs table, please let us know.

# Describe the issue step by step
1.Configure plugin default log level to WARN.
2.Create a flow that uses io.kestra.plugin.core.flow.Subflow.
3.Execute the parent flow.
4.Check the logs table`.
5.Most plugin INFO logs are filtered out as expected.
6.However, the Subflow creation log is still stored as INFO.

# Expected behavior
When plugin logLevel is set to WARN, INFO-level Subflow creation logs should not be persisted.

# Actual behavior
A Subflow-related INFO log is still persisted because it appears to be emitted with a hardcoded Level.INFO in JdbcExecutor.

# Environment
Kestra version: v1.2.3
Runner: JDBC
Database: PostgreSQL

nikhiln64:
### ✨ Description

When a subflow execution is created, the executor emits a `Created new execution [[link ...]]` log entry through `LogEntryEmitter` with a hardcoded `Level.INFO` bypassing the regular logging pipeline and therefore any configured level filtering. Users who set plugin defaults like:

```yaml
plugins:
  defaults:
    - type: io.kestra.plugin
      forced: true
      values:
        logLevel: WARN
```

to keep the `logs` table lean still get this INFO entry persisted for every single subflow execution.

The fix looks up the parent task on the flow (which `findByExecutionThenInjectDefaults` has already enriched with plugin defaults, so `logLevel: WARN` from configuration is visible) and skips the emit when INFO is below the configured level. Behavior is unchanged when no `logLevel` is configured, and the server log line (`log.info`) is untouched only the persisted LogEntry respects the filter.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/16238.

### 🛠️ Backend Checklist

- [x] All unit and integration tests pass `:executor:test --tests ExecutionEventMessageHandlerSubflowLogTest` (new, 2 tests) and `:jdbc-h2:test --tests H2RunnerTest` per the executor-change guideline (100 tests)

### 📝 Additional Notes

- The new test drives the real handler with a parent flow holding a `Subflow` task: with no `logLevel` the creation log is emitted (guards against regressing the default behavior); with `logLevel: WARN` it is suppressed. Verified red without the fix, green with it.
- While running `H2RunnerTest` three times I observed one-off failures in `flowTriggerWithPause` and `flowTriggerAnyMode` (await-timeout style, a different test each run, all passing on other runs, including a fully green run on clean `develop`). They look like pre-existing timing flakiness in the multiple-condition trigger tests rather than anything related to this change flagging in case it's useful signal.
- Developed with AI assistance (Claude Code), reviewed and driven by a human. Template cat joke: the subflow log was like a cat announcing every door it walks through now it only meows when you've asked to hear it. 🐱

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/RunContextLogger.java
+++ b/core/src/main/java/io/kestra/core/runners/RunContextLogger.java
@@ -224,6 +224,27 @@ public class RunContextLogger implements Supplier<org.slf4j.Logger> {
     }
 
     /**
+     * Emit the given log entry only when its level is enabled for this context's configured
+     * log level. Prefer this over {@link #emitLog(LogEntry)} for manually built entries:
+     * direct emission bypasses the regular logging pipeline, so the configured level filter
+     * (e.g. a task's {@code logLevel} property) must be applied explicitly.
+     *
+     * @see #emitLog(LogEntry)
+     */
+    public void emitLogIfEnabled(LogEntry logEntry) {
+        if (logEntry.getLevel() == null || isLogLevelEnabled(logEntry.getLevel())) {
+            this.emitLog(logEntry);
+        }
+    }
+
+    /**
+     * Returns true when the given level is enabled for this context's configured log level.
+     */
+    public boolean isLogLevelEnabled(org.slf4j.event.Level level) {
+        return this.loglevel == null || Level.toLevel(level.toString()).isGreaterOrEqual(this.loglevel);
+    }
+
+    /**
      * Emit the log lines attached to a dynamically-generated taskrun through the regular logging
      * pipeline, so the standard appender behaviour (secret masking, long-message splitting, level
      * filtering, forwarding to the server log, file vs queue routing) applies natively — the only
--- a/core/src/test/java/io/kestra/core/runners/RunContextLoggerTest.java
+++ b/core/src/test/java/io/kestra/core/runners/RunContextLoggerTest.java
@@ -495,6 +495,32 @@ class RunContextLoggerTest {
             .contains(org.assertj.core.groups.Tuple.tuple(RunContextLogger.PROGRESS_KEY, "pod.created"));
     }
 
+    @Test
+    void shouldFilterDirectlyEmittedLogEntriesWhenLevelIsBelowConfiguredLogLevel() {
+        // Given: a logger whose configured level filter is WARN
+        List<LogEntry> logs = new CopyOnWriteArrayList<>();
+        logQueue.addListener(log -> logs.add(log));
+
+        Flow flow = TestsUtils.mockFlow();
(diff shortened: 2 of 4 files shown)


loicmathieu:
I suspect we have the same issue in more places (at least the new Loop task), so I would create a small helper for that so it can be reused in other places. @nikhiln64 can you handle that?

nikhiln64:
Done @loicmathieu. I extracted the check into RunContextLogger as a new emitLogIfEnabled method backed by an isLogLevelEnabled helper, since that class already carries the configured level filter. The handler now binds its logger to the parent task through the existing task aware factory method and calls emitLogIfEnabled, so the level rule lives in one reusable place, and there is a focused unit test for the helper in RunContextLoggerTest. The branch is also rebased onto the latest develop.

I checked the Loop task while I was there. Its iteration failure log already goes through runContextLoggerFactory.create(parentTaskRun, loop, kind).logger() which honors the configured level natively, and iteration creation does not emit a direct LogEntry at all today, so the subflow creation log was the only direct emission site in the codebase. Any future direct emission can now reuse the helper.

One note on the red Backend Tests check from the previous run. The only failing test is io.kestra.plugin.core.http.RequestTest.head404 which expects a 404 and receives a 403. It fails identically on the latest run of #17397, passes locally in isolation on this branch, and nothing in this diff touches HTTP code, so it looks like cross test interference on the shared embedded server rather than something from this change.

loicmathieu (approved):
LGTM

PR #17404, opened 2026-07-14
fix(flows): align update request with SDK type

hemantmm:
### ✨ Description

What does this PR change?  

This change was extracted from https://github.com/kestra-io/kestra/pull/17131#pullrequestreview-4685785268 following review feedback, since the `flow.ts` SDK type fix is unrelated to the nested file-preview implementation.

It fixes the existing `vue-tsc` error for `FlowsAPI.updateFlow` and mirrors the type cast already used by `FlowsAPI.createFlow`.


hemantmm:
Hi @flcarre! Following your review on #17131, I split the unrelated SDK typing change into a separate PR as suggested.

Could you take a look at this PR when you have a chance? It contains only the `updateFlow` SDK type alignment (matching the existing `createFlow` call) with no functional changes.

Thanks!

elevatebart:
I believe this change is unnecessary.
It will also make the usage of the SDK moot.
SDK is supposed to give hints to the developer to what the API expects and if we specifcally "don't" respect it, we loose this advantage.

PR #17436, opened 2026-07-15
feat(system): add Progressive Web App (PWA) support

flcarre:
## Summary

- Adds a web app manifest (relative `start_url`/`scope`/icon paths), the missing icon sizes (192, 512, maskable, apple-touch-icon), and a manually-registered service worker so Kestra can be installed as a PWA from any browser.
- The service worker's scope is computed at runtime from `window.KESTRA_BASE_PATH` via a small pure `computeSwScope()` function, so the same build artifact stays correct whether served at the root, under a sub-path, or behind a reverse proxy. sw.js is emitted at the root of the `ui/` build output (mapped by the webserver to `<basePath>/ui/**`), which caps the max allowed scope — `computeSwScope` returns `<basePath>/ui/`, not the bare basePath, to stay inside that cap (see the comment in `src/utils/serviceWorker.ts`).
- Precache is restricted to small static app-shell files (manifest, icons, `loader.css`). No JS/CSS bundle is precached: this app's build has no small, cleanly-separable "critical entry chunk" (Monaco, shiki per-language grammars/themes, mermaid, katex, echarts and the topology graph are all reachable from the entry, and the biggest single entry chunks alone are 2-4 MB), so precaching any of it would make the first install multi-MB. All JS/CSS is always fetched from the network.
- `/api/` calls are explicitly `NetworkOnly` in the workbox config for clarity, on top of not being precached.
- No `navigateFallback`: the webserver injects a fresh, request-scoped CSRF meta tag into `index.html` on every request (`StaticFilter`); serving a cached/fallback copy for navigations would ship a stale or missing token and break the next mutating API call. Navigation requests always hit the network. (vite-plugin-pwa defaults `navigateFallback` to `"index.html"`, which is explicitly disabled here.)
- The SW installs in the background and activates via `skipWaiting` + `clientsClaim`, with no reload prompt/toast — matches the "silent update" behavior requested, no mid-session disruption.
- Adds a small, dismissible `KsAlert`/`KsButton` install affordance (`PwaInstallPrompt.vue`) driven by a new `usePwaInstall` composable (`beforeinstallprompt` capture, `canInstall` false when already standalone or on an insecure non-localhost origin).
- New `pwa.*` i18n keys added to `en.json` and translated into all 12 other locales; `check.js`/`check.ts` reports no missing/extra keys.

## Out of scope (deferred)

- Curated mobile UI / mobile-friendly views.
- Push notifications.
- Offline-first data caching or background sync.
- Native/Capacitor wrapper.
- Any webserver/Java change (this is a frontend-only PR).

## Manual QA required

PWA install and service-worker scope must be verified against a served production build behind a sub-path reverse proxy (e.g. served under `/kestra/`); the service worker does not run under the Vite dev server, so this can't be exercised in `npm run dev`.

closes: https://github.com/kestra-io/kestra-ee/issues/7032

## Test plan

- [x] `npm run check:types` (pre-existing, unrelated `src/utils/init.ts` moment-typing error reproduces identically with all my changes stashed — not introduced by this PR)
- [x] `npm run test:unit` - 707 tests passed, including new `computeSwScope` and `usePwaInstall` specs
- [x] `npm run test:lint` - 0 errors (pre-existing warnings only, none in touched files)
- [x] `node src/translations/check.ts` (via `npx tsx`) - no missing/extra keys across all 13 locales
- [ ] Manual QA: install prompt + SW scope behind a sub-path reverse proxy (see above)


flcarre:
## QA — PWA support

Built the production UI and served it via `runLocal`, then browser-tested at the root base path and behind a `/kestra` context-path (reverse-proxy simulation). The service worker was exercised on a real served production build (it does not run under the vite dev server).

**Verdict: PASS**

| # | Check | Result |
|---|-------|--------|
| 1 | Service worker registers, scope = `<base>/ui/` | PASS. `http://localhost:28080/ui/` at root; `http://localhost:28080/kestra/ui/` under a `/kestra` context-path. Active, correct controller, single registration per origin, no `SecurityError`. |
| 2 | Console clean | PASS. No errors on load/reload. (The 401 seen under `/kestra` is the unauthenticated auth-check on a fresh instance, unrelated to the PWA.) |
| 3 | Manifest + icons | PASS. `manifest.webmanifest` served with relative internal URLs (`start_url`/`scope` = `./`); all icons + `sw.js` + workbox return 200 at both base paths; apple-touch-icon linked; precache is shell-only (8 entries, 54.5 KiB). |
| 4 | `navigateFallback` disabled | PASS. Per-request CSRF meta is preserved (index.html is not served from cache); an SW-active hard reload of a rendering route loads fresh with no navigation error. |
| 5 | App shell regression (App.vue) | PASS. Root render intact in light and dark; the install banner is correctly gated to non-anonymous routes (hidden on the setup page); the wiring mirrors the existing root-overlay pattern. |

**Base-path correctness (the load-bearing risk) is verified live at both root (`/ui/`) and a sub-path (`/kestra/ui/`)** — the runtime `computeSwScope(window.KESTRA_BASE_PATH)` resolves the scope correctly, and the service worker is served from a directory that permits that scope. This is the scenario that fails silently behind a reverse proxy, now confirmed working.

**Not exercised live:** the authenticated `DefaultLayout` (left menu + top bar), because a fresh instance gates on first-admin setup (account creation). Covered by the `App.vue` diff (a non-invasive root-overlay addition) and the passing unit tests.

ui/src/utils/serviceWorker.ts
```diff
@@ -0,0 +1,30 @@
+export const SERVICE_WORKER_FILENAME = "sw.js"
+
+// A service worker's max allowed scope is capped by the directory it is served from
+// (unless the server sends a `Service-Worker-Allowed` header, which we don't set - no
+// webserver change). sw.js is emitted at the root of the Vite build output, which the
+// webserver maps to "<basePath>/ui/**" (see cli/src/main/resources/application.yml,
+// micronaut.router.static-resources.ui). The scope must therefore be "<basePath>/ui/",
+// not the bare deploy basePath - registering with the basePath alone would target a
+// directory *above* where sw.js is served and throw a SecurityError.
```
elevatebart: Shorter comments please (2 lines max)

flcarre: its just a poc for now, its not ready for review

elevatebart:
I absolutely love this initiative. 
It makes Kestra a real app on my Phone.

What is the rationale behind starting it now? 
Does it facilitate use of the Camera? other APIs?

flcarre:
Quick context on where this comes from: it started as the PWA piece of the mobile app ticket ([kestra-ee#7032](https://github.com/kestra-io/kestra-ee/issues/7032)), basically making Kestra installable as a standalone app.

Nice bonus though, it's not mobile-only: you can install it on desktop too (Chrome/Edge give you a dedicated window with no browser chrome), which is honestly pretty cool. The actual mobile-optimized UI is a separate follow-up, this PR is just the installable-app plumbing, and it happens to work everywhere.

elevatebart (approved):

PR #17437, opened 2026-07-15
fix(jdbc-mysql): resolve duplicate V1_57 Flyway migration version

AcevedoR:
## What / Why

CI on `releases/v1.3.x` fails every `jdbc-mysql` test (27 tests, all `initializationError`). Root cause:

```
io.micronaut.context.exceptions.BeanInstantiationException:
  Message: Found more than one migration with version 1.57
  -> jdbc-mysql/.../migrations/mysql/V1_57__namespace_file_metadata_case_sensitive.sql
  -> jdbc-mysql/.../migrations/mysql/V1_57__widen_task_id.sql
```

Two migrations landed independently on the release branch with the **same MySQL version `1.57`**:

| File | Origin | Also in postgres? |
|------|--------|-------------------|
| `V1_57__widen_task_id.sql` | #17329 | ✅ yes, at V1_57 |
| `V1_57__namespace_file_metadata_case_sensitive.sql` | Pylon #2018 | ❌ mysql-only |

Flyway refuses to start with duplicate versions, so the `DataSource` bean never instantiates and the whole `jdbc-mysql` module fails to initialize.

## Fix

Rename the **mysql-only** `namespace_file_metadata_case_sensitive` migration from `V1_57` → `V1_58`, keeping `widen_task_id` at `V1_57` so the shared migration stays aligned with its postgres counterpart.

The two migrations are independent, order-agnostic DDL statements. Neither `V1_57` file shipped in any `v1.3.x` tag (latest is `v1.3.27`), so no already-applied migration checksum is affected.

## Scope note

The same CI run also shows 3 `core` HTTP test failures (`HttpClientTest.getProxy`, `RequestRunnerTest.request`/`request_no_options`) — those are environmental: the tests hit live `https://www.google.com/` and received `429` (Google bot captcha) from the runner's IP. Out of scope for this PR; they pass on re-run.


brian-mulier-p (approved):
LGTM, if sure it's only a mysql problem

PR #17438, opened 2026-07-15
feat(flows): VS Code-style find & replace for Source Search

flcarre:
## Summary

- Redesigns the Source Search page around the approved mockup: a collapsible replace row (match-case/whole-word/regex toggles), a namespace + scope (all source/tasks/triggers/inputs) filter row, match navigation, and a results tree with per-flow/per-match checkboxes for bulk replace.
- Extends `GET /flows/source` with `caseSensitive`, `wholeWord`, `regex` and `scope` query params, and adds `POST /flows/source/replace/preview` (computes the diff without persisting), `POST /flows/source/replace/apply` (persists, permission-checked) and `POST /flows/source/replace/line` (single match). Search/replace matching is centralized in `SourceSearchMatcher` (core) so all endpoints share the same semantics; regex mode bypasses the coarse full-text pre-filter since it can't recall patterns spanning multiple lines.
- The preview pane switches to an inline diff (removed/added tinted lines) with a confirm bar once the user previews a replace.
- `SourceSearchService.isEditable(...)` rejects soft-deleted flows and flows carrying the `system.readOnly` label, so the read-only pill, disabled checkboxes and exclusion banner already render in OSS for Git-synced flows. EE overrides the same seam to add `FLOW · UPDATE` permission checks on top (kestra-io/kestra-ee#9381).
- Translates every new string into all 13 locales.

## Addressing the backend review

- **ReDoS.** `SourceSearchMatcher.ensureSafeQuery` rejects an over-long or catastrophically-backtracking user regex via `RegexUtils.isSafeUserRegex`. It is called from `toPattern` — the single choke point every endpoint goes through — plus eagerly in `SourceSearchService.search` so a query matching zero candidates still returns 400 rather than silently succeeding. Matching itself already ran through `RegexUtils.matcher` for the timeout guard. Literal queries are exempt: they are `Pattern.quote`d before compilation.
- **No more catching `IndexOutOfBoundsException`.** `ensureValidReplacement` validates the replacement up-front against the compiled pattern (dangling `\` or `$`, unclosed `${`, unknown named group, group number the pattern does not declare), so `Matcher.appendReplacement`/`replaceAll` can no longer throw. The missing bound is checked too: `column > line.length()` was unguarded before `Matcher.find(int)`.
- **Skips are no longer silent.** `skipped` entries are `SkippedFlow` records carrying a `SkipReason` (`NOT_FOUND`, `READ_ONLY`, `NO_CHANGE`, `NO_MATCH`, `INVALID_FLOW`) and, for `INVALID_FLOW`, the validation message. The UI toast names the distinct reasons instead of a fixed "couldn't be saved".
- **`system.readOnly`.** `isEditable` now checks soft deletion and the `system.readOnly` label rather than returning `true` unconditionally, matching what the frontend flow store already does.
- **Merge damage.** `searchFlowsBySourceCode` had literal conflict markers committed inside its parameter list. Resolved to develop's side, which also restores the `sort` query param and the `@Max(PageableUtils.MAX_PAGE_SIZE)` cap on `size`.
- **Upper caps.** Page size is capped as above; a single replace can rewrite at most 1000 flows (`@Size(max = ...)` on the request), matching the candidate ceiling of one source search. That ceiling, `MAX_SOURCE_SEARCH_CANDIDATES`, now lives on `SourceSearchMatcher` instead of being a private constant of `AbstractJdbcFlowRepository`: every repository has to bound the candidate set it pulls back, since whole-word, regex and scope matching run in Java. Applying the same review to EE caught a real regression there, where the Elasticsearch implementation had replaced a paginated search with an unbounded scroll (fixed in kestra-io/kestra-ee#9381).
- **Dead API.** Dropped the deprecated four-arg `findSourceCode` overload; no caller remained.
- **Store pass-throughs.** `FlowsSearch` calls the generated SDK directly and owns its results, so the store ref and its four pass-throughs are gone. Regenerating the SDK also revealed the committed copy was stale: it predated the `caseSensitive`/`wholeWord`/`regex`/`scope` query params, so it did not describe endpoints the UI already called.

## Deviations from the mockup (with reason)

- The regex-error input no longer gets a literal red border (`aria-invalid` is set instead) — the DS's `KsSearch`/`KsInput` don't expose an error-state prop, and reaching into it would require a `:deep()` selector; the full `KsAlert type="error"` panel already carries the message.
- Each result group's expand/collapse indicator is `KsCollapseItem`'s own disclosure arrow rather than a hand-rotated chevron, to reuse the existing DS affordance instead of duplicating it.
- The collapse/expand-all header button is a single toggle (icon + tooltip swap between collapse-all/expand-all) rather than a static icon, since the mockup doesn't show what it looks like once already collapsed.
- No pagination controls (the mockup doesn't show any either): search fetches up to 200 matching flows per query; a follow-up should add real pagination if an instance can exceed that.
- Replace-all applies at flow granularity: checkboxes let you curate which flows are included, but applying rewrites all of a selected flow's matching occurrences (within the chosen scope). Replacing a single occurrence is done through the per-match replace action.
- `scope` (tasks/triggers/inputs) is resolved via a lightweight heuristic (line range from a top-level YAML key to the next), not a full YAML-AST parse — correct for normal flow structure.

part-of: https://github.com/kestra-io/kestra-ee/issues/8645

**EE counterpart: https://github.com/kestra-io/kestra-ee/pull/9381** — must be merged together, since it implements the new `findSourceCode` signature and the `FLOW · UPDATE` override of `isEditable`.

## Test plan

- [x] Rebased onto `develop` (was 39 commits behind).
- [x] `:core:test --tests SourceSearchMatcherTest` 13/13, `:webserver:test --tests SourceSearchServiceTest` 12/12, `--tests FlowControllerTest` 72/72, `:jdbc-h2:test --tests "*FlowRepositoryTest*"` 95/95
- [x] `cd ui && npm run check:types` — clean
- [x] `npm run test:unit` — 1024/1024 passing
- [x] `npm run test:lint` — 0 errors
- [x] Translations: all 13 locales carry every `source_search` key; the branch introduces no new missing keys (the 117 pre-existing `onboarding.tour.*` gaps are already on `develop`)
- [x] EE compiles and its RBAC tests pass against this branch (kestra-io/kestra-ee#9381: `:webserver-ee:test --tests "*FlowControllerTest*"` 22/22)


flcarre:
**Demo: VS Code-style find & replace for Source Search**

https://github.com/user-attachments/assets/df47e12c-3437-44bc-b75c-4685052ba452

Recorded against a local build of this branch. Search across every flow's YAML source (case, whole-word, regex, scope), jump between matches, and replace with a live inline old → new diff — a single occurrence (two matches on one line each get their own Replace), all matches in one flow, or across every flow. RBAC and secrets are respected: read-only flows are skipped and secret keys are shown key-only.

elevatebart (changes requested):
Please solve merge conflicts first.
Then I would like to avoid using the store as a strong-typer for APIs specially since we now have a full strongly typed client generated on `npm run generate:sdk`.

flcarre:
@elevatebart both addressed: rebased onto `develop` (conflicts resolved) and the store now goes through the generated SDK client (`searchFlowsBySourceCode` / `previewReplaceBySourceCode` / `applyReplaceBySourceCode` / `replaceLineBySourceCode`) instead of raw axios.

ui/src/components/executions/outputs/ExpressionDebugger.vue
```diff
@@ -25,14 +25,6 @@
...
                 <strong>{{ error }}</strong>
             </p>
-            <div v-if="stackTrace" class="my-2">
-                <CopyToClipboard
-                    :text="`${error}\n\n${stackTrace}`"
-                    :label="$t('copy')"
-                    class="d-inline-block me-2"
-                />
-            </div>
-            <pre v-if="stackTrace" class="stack mb-0">{{ stackTrace }}</pre>
```
elevatebart: ??

ui/src/stores/flow.ts
```diff
@@ -441,16 +442,27 @@ export const useFlowStore = defineStore("flow", () => {
             }
         })
     }
-    function searchFlows(options: { [key: string]: any }) {
-        const {sort, ...rest} = options
-        return FlowsAPI.searchFlowsBySourceCode({...rest, sort: sort ? [sort] : undefined}).then(response => {
-            search.value = response.results as unknown as any[]
+    function searchFlows(options: Parameters<typeof FlowsAPI.searchFlowsBySourceCode>[0]) {
+        return FlowsAPI.searchFlowsBySourceCode(options).then(response => {
+            search.value = response.results as SourceSearchResult[]
```
elevatebart: Why do we need to store the search results in a store? Do we have to share it with somewhere?

elevatebart (approved):
Overall good PR, I don't see the changes in `stores/flow.ts` useful. I woudl rather call the API directly since we never use the actual results.

ui/src/components/executions/outputs/ExpressionDebugger.vue
elevatebart: Is this part of the change? It fells like it is not.

Please explain

flcarre: Good catch — that one wasn't mine. #17650 dropped the stacktrace on develop and my branch was just behind, so it only looked like part of this PR. Rebased now, it's gone from the diff.

ui/src/stores/flow.ts
elevatebart: I would even prefer to hit the `FlowsAPI.searchFlowsBySourceCode` in the component instead of using a store passthrough function.

flcarre: Done — FlowsSearch calls `FlowsAPI` directly now and owns its results, so the store ref and the four pass-throughs are gone (-40/+11). Same inlining as your #17632.

Heads up: the develop merge kept the source-search functions in the generated barrel but dropped the matching type re-exports, which turned check:types red. Fixed in a separate commit.

core/src/main/java/io/kestra/core/repositories/FlowRepositoryInterface.java
```diff
@@ -224,7 +224,12 @@ ArrayListTotal<FlowWithSource> findWithSource(
         @Nullable String tenantId,
         @Nullable List<QueryFilter> filters);
 
-    ArrayListTotal<SearchResult<Flow>> findSourceCode(Pageable pageable, @Nullable String query, @Nullable String tenantId, @Nullable String namespace);
+    ArrayListTotal<SearchResult<Flow>> findSourceCode(Pageable pageable, @Nullable String query, boolean caseSensitive, boolean wholeWord, boolean regex, SourceSearchScope scope, @Nullable String tenantId, @Nullable String namespace);
+
+    @Deprecated
+    default ArrayListTotal<SearchResult<Flow>> findSourceCode(Pageable pageable, @Nullable String query, @Nullable String tenantId, @Nullable String namespace) {
```
loicmathieu: Is it still used from the UI? Otherwise better to remove it.

core/src/main/java/io/kestra/core/utils/SourceSearchMatcher.java
```diff
@@ -0,0 +1,153 @@
...
+    private SourceSearchMatcher() {
+    }
+
+    public static Pattern toPattern(String query, boolean caseSensitive, boolean wholeWord, boolean regex) {
+        String base = regex ? query : Pattern.quote(query);
+        String withBoundaries = wholeWord ? "\\b(?:" + base + ")\\b" : base;
+        int flags = caseSensitive ? 0 : (Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE);
+
+        try {
+            return Pattern.compile(withBoundaries, flags);
```
loicmathieu, line 123: There is a risk of Redos here as pattern is user-supplied, so you need to:
1. Check that the regex is safe by using `RegexUtils.isSafeRegex()`. It's usually done inside a controller
2. User RegexUtils.matcher() to have in-depth potection via timeout (l57)

webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
```diff
@@ -259,13 +269,83 @@ public List<Flow> listFlowsByNamespace(
     @ExecuteOn(TaskExecutors.IO)
     @Get(uri = "/source")
     @Operation(tags = { "Flows" }, summary = "Search for flows source code")
-    public PagedResults<SearchResult<Flow>> searchFlowsBySourceCode(
+    public PagedResults<SourceSearchResult> searchFlowsBySourceCode(
         @Parameter(description = "The current page") @QueryValue(defaultValue = "1") @Min(1) int page,
+<<<<<<< HEAD
```
loicmathieu: Loks like a bad merge!

webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
```diff
@@ -259,13 +269,83 @@ public List<Flow> listFlowsByNamespace(
...
         @Parameter(description = "The current page size") @QueryValue(defaultValue = "10") @Min(1) @Max(PageableUtils.MAX_PAGE_SIZE) int size,
         @Parameter(description = "The sort of current page") @Nullable @QueryValue List<String> sort,
+||||||| parent of dda5cda927 (feat(flows): VS Code-style find & replace for Source Search)
+        @Parameter(description = "The current page size") @QueryValue(defaultValue = "10") @Min(1) int size,
```
loicmathieu: Size must have an upper cap

webserver/src/main/java/io/kestra/webserver/services/SourceSearchService.java
```diff
@@ -0,0 +1,236 @@
...
+            return new SourceSearchReplaceApplyResponse(List.of(), List.of(ref));
+        }
+
+        String replacedLine;
+        try {
+            StringBuilder builder = new StringBuilder();
+            matcher.appendReplacement(builder, effectiveReplacement);
+            matcher.appendTail(builder);
+            replacedLine = builder.toString();
+        } catch (IndexOutOfBoundsException | IllegalArgumentException e) {
```
loicmathieu: `IndexOutOfBoundsException` is most likely a bug than an expected error, it's always better to check for array bounds instead of catching it

webserver/src/main/java/io/kestra/webserver/services/SourceSearchService.java
```diff
@@ -0,0 +1,236 @@
...
+        }
+        lines[line - 1] = replacedLine;
+        String newSource = String.join("\n", lines);
+
+        try {
+            GenericFlow genericFlow = GenericFlow.fromYaml(tenantId, newSource);
+            return new SourceSearchReplaceApplyResponse(List.of(flowService.update(genericFlow, current)), List.of());
+        } catch (ConstraintViolationException | FlowProcessingException | DeserializationException e) {
+            log.warn("Skipping flow {}.{} line {} during Source Search replace: {}", namespace, id, line, e.getMessage());
+            return new SourceSearchReplaceApplyResponse(List.of(), List.of(ref));
```
loicmathieu: it might be a good idea to indicate to the UI that this flow has been skiped instead of skipping silently

webserver/src/main/java/io/kestra/webserver/services/SourceSearchService.java
```diff
@@ -0,0 +1,236 @@
...
+        }
+    }
+
+    protected boolean isEditable(FlowInterface flow) {
+        return true;
```
loicmathieu: There is a system.readOnly label for that

webserver/src/main/java/io/kestra/webserver/services/SourceSearchService.java
```diff
@@ -0,0 +1,236 @@
...
+        Pattern pattern,
+        String effectiveReplacement
+    ) {
+        return matches.stream()
+            .map(match -> {
+                String before = stripMarkers(match.snippet());
+                String after;
+                try {
+                    after = RegexUtils.matcher(pattern, before).replaceAll(effectiveReplacement);
+                } catch (IndexOutOfBoundsException | IllegalArgumentException e) {
```
loicmathieu: Again, why catching IndexOutputOfBoundsException?

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/repositories/FlowRepositoryInterface.java
flcarre: Removed in `2d3bf5c64d`. It was an unused back-compat shim I had added; every caller already used the filtered signature.

webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
flcarre: Real conflict markers committed in the parameter list. Fixed in `3f9f09e935`, taking develop's side (restores `sort`).

webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
flcarre: Done, `@Max(PageableUtils.MAX_PAGE_SIZE)`. Also capped the write path: `flows` is `@Size(max = 1000)`.

core/src/main/java/io/kestra/core/utils/SourceSearchMatcher.java
flcarre: Point 2 was already there: `RegexUtils.matcher` is at that line and at both `replaceAll` sites. Point 1 added as `ensureSafeQuery` (`:29`).

I put it in `toPattern` (`:109`) rather than the controller: all four endpoints funnel through it, so nothing can bypass it and there's no duplicated guard. Tell me if you'd still rather have it in the controller.

webserver/src/main/java/io/kestra/webserver/services/SourceSearchService.java
flcarre: Replaced by up-front validation (`ensureValidReplacement`, `SourceSearchMatcher.java:46`). It was hiding a real missing bound: `column` was never checked against the line length, so `find(int)` could throw. Guarded at `:183`.

webserver/src/main/java/io/kestra/webserver/services/SourceSearchService.java
flcarre: Done: `skipped` is now `List<SkippedFlow>` with a `SkipReason` (`NOT_FOUND` / `READ_ONLY` / `NO_CHANGE` / `NO_MATCH` / `INVALID_FLOW`), surfaced in the UI toast.

webserver/src/main/java/io/kestra/webserver/services/SourceSearchService.java
flcarre: Done: `isEditable` now rejects soft-deleted flows and `Label.READ_ONLY` (`:221`). EE composes on top instead of replacing it.

webserver/src/main/java/io/kestra/webserver/services/SourceSearchService.java
flcarre: Same fix as above, the `catch` is gone. No `IndexOutOfBounds` left in either file.

loicmathieu (approved):
LGTM

PR #17443, opened 2026-07-15
ci: reuse EE compile-check webhook in main workflow

AcevedoR:
Reuses the synchronous webhook-based EE compile check introduced in #17115 (which wired it into `pull-request.yml`) inside the main-branch CI (`main-build.yml`).

## What changed

The `trigger-ee` job in `main-build.yml` previously fired an async `oss-updated` `repository_dispatch` at kestra-ee on every `develop` push and forgot about it. It now instead calls the shared composite action `./.github/actions/ee-compile-check` — the same one `pull-request.yml` uses — to run the kestra-ee `compileJava` check for the pushed ref via the Kestra webhook and surface the result inline.

- **`ref`**: `github.ref_name` (the flow resolves the matching kestra-ee ref — same-name branch, else `develop`).
- **`commit-sha`**: `github.sha`, so the flow posts a commit-status callback on the `develop` HEAD — EE-compile health is visible directly on develop.
- **`pr-repo`**: `github.repository` for that callback.
- Requires a `checkout` step (added) so the local composite action can resolve.

The legacy `oss-updated` `repository_dispatch` is **kept but disabled** behind `false &&`, mirroring the rollback pattern in `pull-request.yml` — drop the `false &&` (and disable the webhook step) to revert.

Consumes the existing `KESTRA_CI_EEBUILD_WEBHOOK_URL` repo secret introduced by #17115.

## Notes / open question

Behavior is scoped to `develop` pushes only, exactly as the previous trigger was. The `end` (Slack) job's `needs` were left unchanged, so an EE-compile failure on develop fails the `trigger-ee` job (visible in the Actions UI + as a commit status) but does not currently trigger the Slack failure notification. Happy to wire `trigger-ee` into the `end` job if you'd like develop-breaking EE-compile failures to also ping Slack — left it out to keep scope tight and avoid noise from transient webhook/infra failures.

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -22,13 +22,39 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  # When an OSS ci start, we trigger an EE one
+  # When develop lands a push, run the kestra-ee compile check for the ref (via a
+  # Kestra webhook) and surface the result inline. Replaces the async
+  # "oss-updated" repository_dispatch round-trip.
   trigger-ee:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout  # required so the local composite action below can resolve
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/develop' }}
+        uses: actions/checkout@v4
+
+      # Run the EE compileJava check for the pushed ref on a Kestra instance and
+      # surface the result inline. The flow resolves the matching kestra-ee ref
+      # itself (same-name branch, else develop). commit-sha posts a commit-status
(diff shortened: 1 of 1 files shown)


AcevedoR:
Closing — wrong direction. The main workflow should keep triggering EE **asynchronously** via the existing `oss-updated` repository_dispatch (full EE cross-check, fire-and-forget), not the synchronous compile-only webhook. develop already fires `oss-updated`, so no OSS-side change is needed; the webhook compile-check stays PR-only (#17115). Any follow-up is on the kestra-ee side.

PR #17446, opened 2026-07-15
fix(triggers): bound polling trigger evaluation with a timeout to prevent silent lock leaks

### LINKED ISSUE #17445: Polling trigger silently stuck on JDBC backend when evaluation hangs (evaluation lock never released)
## Describe the issue

On the JDBC/postgres backend, a **polling trigger** (e.g. `io.kestra.plugin.kafka.Trigger`) can get **silently stuck**: its *Next execution date* freezes and never advances, with no error, no failed execution, and nothing in the flow logs. Manually clicking **Unlock** temporarily restores it, but it eventually gets stuck again.

Observed on Kestra 1.3.24, Standalone deployment, `queueType: postgres`, `repositoryType: postgres`. (The trigger being a Kafka trigger is incidental — the scheduler/queue backend is postgres, this is unrelated to the Kafka-*backend* / Kafka Streams scheduler.)

## Root cause

A polling trigger carries an internal *evaluation lock*, `Trigger.evaluateRunningDate`. The scheduler (`AbstractScheduler`) only evaluates triggers whose flag is `null`. The flag is set when the evaluation is dispatched to a worker and cleared when the worker returns a `WorkerTriggerResult`.

Thrown exceptions are handled correctly (the worker's `handleTrigger` → `handleTriggerError` emits a result that releases the lock). The problem is a **missing result** — the worker never produces one, so the flag is never cleared:

1. **Evaluation hangs (primary).** `DefaultWorker.callJob` runs `PollingTriggerInterface.evaluate()` **synchronously with no timeout**. If the evaluation blocks (e.g. a stuck `KafkaConsumer.poll()` during a consumer-group rebalance), the worker thread is wedged forever — no return, no throw, no result → the trigger stays locked.
2. **No-dispatch swallowed.** `AbstractScheduler.sendWorkerTriggerToWorker` persists the lock *before* dispatching and then swallows no-dispatch outcomes (no worker group for the key, `FAIL`/`CANCEL` fallback when no worker is available, queue emission error), leaving the lock set with no job ever running.

There is no automatic recovery on 1.3.x: the flag is cleared only on scheduler restart (`JdbcSchedulerTriggerState.initTriggerEvaluateRunning`) or via manual **Unlock**. `executionMonitor` only logs. (Trigger evaluation has no timeout even on `develop`, which reworked the locking model separately.)

## Expected behavior

A polling trigger whose evaluation hangs or is never dispatched should recover on its own — the worker thread should be freed and the evaluation lock released — instead of silently stalling the schedule until an operator manually unlocks it.

## Environment

- Kestra Version: 1.3.24
- Deployment: Standalone
- Queue/Repository: PostgreSQL

nkwiatkowski:
### What does this PR do?

Fixes a bug where a **polling trigger** on the JDBC backend could get **silently stuck** — its *Next execution date* freezes and never advances, with no error and nothing in the flow logs — until an operator manually clicks **Unlock** (after which it eventually gets stuck again).

Closes #17445

### Root cause

The polling trigger's evaluation lock (`Trigger.evaluateRunningDate`) leaks whenever the worker produces **no result** (as opposed to throwing — thrown exceptions are already handled and release the lock):

1. **Evaluation hangs.** `DefaultWorker.callJob` runs `evaluate()` synchronously with **no timeout**, so a blocking evaluation (e.g. a stuck `KafkaConsumer.poll()` during a consumer-group rebalance) wedges the worker thread forever and never emits a result.
2. **No-dispatch swallowed.** `AbstractScheduler.sendWorkerTriggerToWorker` persisted the lock *before* dispatching, then swallowed no-dispatch outcomes (no worker group for the key, `FAIL`/`CANCEL` fallback, queue emission error), locking the trigger with no job ever running.

There was no automatic recovery — only a scheduler restart or manual Unlock.

### Changes

- **Worker-level trigger-evaluation timeout** (`kestra.worker.trigger-evaluation-timeout`, default `PT10M`): `WorkerTriggerCallable` now bounds `evaluate()` with a Failsafe `Timeout.withInterrupt()`, mirroring `WorkerTaskCallable`. On timeout the evaluation is interrupted (freeing the worker thread) and an error result is emitted, which releases the scheduler's evaluation lock. Applies to polling triggers only; realtime triggers are unaffected. The default is deliberately generous so legitimate long evaluations are never cut — lower it for faster recovery from hangs.
- **Lock only on actual dispatch**: `sendWorkerTriggerToWorker` now returns whether the job was really dispatched, and `handle()` persists `evaluateRunningDate` only when it was.

### How was it tested?

- `WorkerTriggerCallableTest` — a blocking evaluation is interrupted and reported as `FAILED` with a `TimeoutExceededException`; a fast evaluation returns `SUCCESS`.
- `SchedulerTriggerDispatchLockTest` — a no-dispatch outcome leaves the trigger unlocked (verified to fail if the fix is reverted).
- Full `:worker:test` and `:scheduler:test` suites pass (existing `SchedulerPollingTriggerTest` still green — normal dispatch is unaffected).

### Notes

- Targets `releases/v1.3.x`. All changes are in OSS; Enterprise inherits them (EE does not override `DefaultWorker` or the scheduler).
- Trigger evaluation has no timeout on `develop` either (the locking model was reworked separately there); a forward-port will follow.

### DIFF
--- a/worker/src/main/java/io/kestra/worker/WorkerConfig.java
+++ b/worker/src/main/java/io/kestra/worker/WorkerConfig.java
@@ -1,0 +1,26 @@
+package io.kestra.worker;
+
+import java.time.Duration;
+
+import io.micronaut.context.annotation.ConfigurationProperties;
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.bind.annotation.Bindable;
+
+/**
+ * Worker configuration.
+ *
+ * @param triggerEvaluationTimeout The maximum time a polling trigger evaluation is allowed to run on
+ *        a worker before it is interrupted. This bounds evaluations that would otherwise block
+ *        forever (e.g. a stuck {@code KafkaConsumer.poll()} during a consumer-group rebalance): on
+ *        timeout the worker thread is freed and an error result is emitted so the scheduler releases
+ *        the trigger evaluation lock instead of leaving the trigger silently stuck. Applies to
+ *        polling triggers only (realtime triggers run for their whole lifetime by design). The
+ *        default is deliberately generous so that legitimate long evaluations are never interrupted;
+ *        lower it if faster recovery from hangs is desired.
+ */
+@ConfigurationProperties("kestra.worker")
+public record WorkerConfig(
+    @Bindable(defaultValue = "10m")
+    @Nullable Duration triggerEvaluationTimeout
+) {
+}
--- a/worker/src/main/java/io/kestra/worker/WorkerConfig.java
+++ b/worker/src/main/java/io/kestra/worker/WorkerConfig.java
@@ -9,7 +9,7 @@ import io.micronaut.core.bind.annotation.Bindable;
 /**
  * Worker configuration.
  *
- * @param triggerEvaluationTimeout The maximum time a polling trigger evaluation is allowed to run on
+ * @param pollingTriggerTimeout The maximum time a polling trigger evaluation is allowed to run on
  *        a worker before it is interrupted. This bounds evaluations that would otherwise block
  *        forever (e.g. a stuck {@code KafkaConsumer.poll()} during a consumer-group rebalance): on
  *        timeout the worker thread is freed and an error result is emitted so the scheduler releases
@@ -21,6 +21,6 @@ import io.micronaut.core.bind.annotation.Bindable;
 @ConfigurationProperties("kestra.worker")
 public record WorkerConfig(
     @Bindable(defaultValue = "10m")
-    @Nullable Duration triggerEvaluationTimeout
+    @Nullable Duration pollingTriggerTimeout
 ) {
 }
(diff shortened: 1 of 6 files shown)


worker/src/main/java/io/kestra/worker/WorkerConfig.java
```diff
@@ -0,0 +1,26 @@
...
+ * @param triggerEvaluationTimeout The maximum time a polling trigger evaluation is allowed to run on
+ *        a worker before it is interrupted. This bounds evaluations that would otherwise block
+ *        forever (e.g. a stuck {@code KafkaConsumer.poll()} during a consumer-group rebalance): on
+ *        timeout the worker thread is freed and an error result is emitted so the scheduler releases
+ *        the trigger evaluation lock instead of leaving the trigger silently stuck. Applies to
+ *        polling triggers only (realtime triggers run for their whole lifetime by design). The
+ *        default is deliberately generous so that legitimate long evaluations are never interrupted;
+ *        lower it if faster recovery from hangs is desired.
+ */
+@ConfigurationProperties("kestra.worker")
```
fhussonnois, line 21: Could you just change the property name to be `kestra.worker.polling-trigger-timeout.`

fhussonnois (approved):
LGTM, minus the property name. Thanks

PR #17456, opened 2026-07-16
fix(system): widden too small columns and limit trigger.id to 256

loicmathieu:
Add validation for Trigger.id for maximum of 256 char as Task.id.
Widden columns to match the maximum size defined in the code.

This would prevent future issues as https://github.com/kestra-io/kestra/issues/17328


loicmathieu:
cc @fhussonnois for cross-validation as it touch some trigger stufff

fhussonnois (approved):

PR #17457, opened 2026-07-16
feat(filters): add advanced label filter operators

### LINKED ISSUE #16344: Add execution label filter
## Request
Allow users to filter/search executions by execution labels from the Executions page.

## Context
Customer in Pylon #1831 labels executions with a case identifier and wants to find executions by that label. The Dashboard page supports a labels filter, but the Executions page search/filter does not appear to include labels.

## Expected behavior
The Executions page should expose a label filter/search option, ideally consistent with the Dashboard label filter UX.

### Additional Options
- The user should be able to set multiple label filters.

### LINKED ISSUE #16439: [Filters] New filter operators for Labels part 2
### Feature description

Seeing the original https://github.com/kestra-io/kestra/issues/11587 closed we would like to keep the topic of advanced filtering of labels open. Note the AND/OR functionality done is a subset of the originally envisioned broad functionality improving the overall usability.

Relevant part of the #11587 screenshot:

Image

---

I have no idea about relevance of #15952 in this context.

---

There were new operators to be implemented. Currently, there are `EQUALS` and `NOT EQUALS` operators available in the UI. Additionally, following operators were listed in the epic:

* `LIKE` - contains substring
* `NOT LIKE`  - does not contain substring
* `IS NULL` - label key does not exist
* `IS NOT NULL` - label key exists

hemantmm:
closes: #16439
closes #16344

Compation PR: https://github.com/kestra-io/kestra-ee/pull/9551
---

### ✨ Description

What does this PR change?  

Added advanced label filtering operators for flows, executions, and dashboards. Labels now support `has any of`, `has none of`, `has all of`, `contains`, `does not contain`, `is set`, and `is not set`, backed by `IN`, `NOT_IN`, `EQUALS`, `CONTAINS`,` NOT_CONTAINS`, `IS_NOT_NULL`, and `IS_NULL`.

https://github.com/user-attachments/assets/ca4433f5-50b2-4cf1-9d0c-a909bfc4563b


core/src/main/java/io/kestra/core/models/QueryFilter.java
```diff
@@ -173,7 +179,7 @@ public List<Op> supportedOp() {
         LABELS("labels") {
             @Override
             public List<Op> supportedOp() {
-                return List.of(Op.EQUALS, Op.NOT_EQUALS, Op.IN, Op.NOT_IN, Op.CONTAINS);
+                return List.of(Op.IN, Op.NOT_IN, Op.EQUALS, Op.CONTAINS, Op.NOT_CONTAINS, Op.IS_NOT_NULL, Op.IS_NULL);
```
flcarre: `NOT_EQUALS` was dropped from labels support, which wasn't part of the ask and is exactly what's breaking 15 CI tests (`InvalidQueryFiltersException` on H2/MySQL/Postgres). Restoring it should turn everything green:
```suggestion
                return List.of(Op.IN, Op.NOT_IN, Op.EQUALS, Op.NOT_EQUALS, Op.CONTAINS, Op.NOT_CONTAINS, Op.IS_NOT_NULL, Op.IS_NULL);
```

flcarre (changes requested):
Really nice coverage of the four new operators across H2, MySQL and Postgres. One blocker inline (the CI failures), plus two smaller things:

- Reopening a labels contains/is-set chip after a reload shows an empty value in the edit popup: `processFieldValue` returns an array for key-value fields, but the text/null comparator path in `FilterEditPopper` expects a string.
- The new comparator strings are copied in English into the non-English locales, could you translate them like the neighboring keys?

hemantmm:
Thank you, I will address the comments and resolve the issue.

loicmathieu:
This would need to be ported to our EE Elasticsearch repository before being able to be merged.

hemantmm:
Thanks, I'm happy to make any follow-up changes needed on this PR.

flcarre (commented):
Thanks a lot for iterating on this @hemantmm 🙏 The advanced label operators are a really nice addition, and a few things here are genuinely well done:

- The `AbstractJdbcFlowRepository.findLabelCondition` change (`value.isLeft() ? value.getLeft() : value.getRight()`) fixes a real pre-existing bug: the old unconditional `.getLeft()` throws `NoSuchElementException` on an `Either.Right`, so `CONTAINS` on flow labels was already broken. Nice catch, and well covered by the new `containsFilter` / `notContainsFilter` tests.
- The reopen-chip fix (gating the array branch on `KV_COMPARATORS` in `filterChipFactory.ts`) resolves the empty-value-on-reload issue cleanly, and the new `filterChipFactory.test.ts` locks the round-trip down.
- Comparator strings are translated across all 13 locales rather than copy-pasted English. 👍
- IN/NOT_IN semantics (OR across entries for IN, AND for NOT_IN) are consistent across H2/MySQL/Postgres.

I traced the whole path (frontend filter pipeline → `QueryFilter` → the JDBC backends and the dashboard bridge). One thing is worth fixing before merge, plus a small question and a couple of optional nits.

### 🔴 Dashboard global label filter ignores the chosen operator

This PR now exposes `IN / NOT_IN / CONTAINS / NOT_CONTAINS / IS_SET / IS_NOT_SET` for labels on the dashboard filters (`dashboardFilters.ts`, all three composables). But the runtime bridge that turns a global filter into a dashboard query, `IExecutions.whereWithGlobalFilters` (in `core/.../plugin/core/dashboard/data/IExecutions.java`), hardcodes `Contains` for labels and never reads `f.operation()`:

```java
labelFilters.forEach(f -> {
    if (f.value() instanceof Map<?, ?> m) {
        m.forEach((key, value) -> updatedWhere.add(
            Contains.<Fields>builder().field(Fields.LABELS).key(key.toString()).value(value).build()));
    } else {
        updatedWhere.add(Contains.<Fields>builder().field(Fields.LABELS).value(f.value()).build());
    }
});
```

The `NAMESPACE` / `FLOW_ID` / `STATE` / `SCOPE` blocks right next to it already delegate to `f.toDashboardFilterBuilder(...)`, which is why they honor the operator. So on a dashboard, picking "Has none of" / "Does not contain" / "Is not set" for a label produces the exact same query as "Contains", silently returning the wrong rows with no error.

The good news: `AbstractFilter` already carries the `key`, and all the filter classes you need already exist (`EqualTo`, `NotEqualTo`, `In`, `NotIn`, `IsNull`, `IsNotNull`, `Contains`). A quick fix is to switch on the operator, something along these lines:

```java
m.forEach((key, value) -> updatedWhere.add(labelFilter(f.operation(), key.toString(), value)));

// ...

private AbstractFilter<Fields> labelFilter(QueryFilter.Op op, String key, Object value) {
    return switch (op) {
        case EQUALS      -> EqualTo.<Fields>builder().field(Fields.LABELS).key(key).value(value).build();
        case NOT_EQUALS  -> NotEqualTo.<Fields>builder().field(Fields.LABELS).key(key).value(value).build();
        case IN          -> In.<Fields>builder().field(Fields.LABELS).key(key).values(/* split value into a list, like QueryFilter.asValues */).build();
        case NOT_IN      -> NotIn.<Fields>builder().field(Fields.LABELS).key(key).values(/* same */).build();
        case IS_NULL     -> IsNull.<Fields>builder().field(Fields.LABELS).key(key).build();
        case IS_NOT_NULL -> IsNotNull.<Fields>builder().field(Fields.LABELS).key(key).build();
        default          -> Contains.<Fields>builder().field(Fields.LABELS).key(key).value(value).build();
    };
}
```

One caveat: there's no `NotContains` dashboard filter class yet, so `NOT_CONTAINS` can't be represented on dashboards. Simplest for this PR is probably to drop `NOT_CONTAINS` (and any operator without a matching filter class) from the dashboard label comparator list until a `NotContains` filter exists, so the UI never offers an operator the backend can't honor. Relatedly, the new `default -> throw` in `QueryFilter.toDashboardFilterBuilder` removes the previous compile-time exhaustiveness safety net; adding a `NOT_CONTAINS` case (or a `NotContains` class) would restore it.

### ❓ Quick question: `flowId` lost `NOT_EQUALS`

`Comparators.NOT_EQUALS` was removed from the `flowId` field in both `executionFilter.ts` (`useExecutionFilter`) and `dashboardFilters.ts` (`useNamespaceDashboardFilter`). The backend still lists `Op.NOT_EQUALS` for `FLOW_ID`, and it looks unrelated to labels, so I wanted to check whether it was intentional or accidental collateral from one of the develop merges. If unintentional, re-adding `Comparators.NOT_EQUALS` restores "flow ID is not X" on the Executions list and the Namespace dashboard.

### Minor / optional

- `COMPARATOR_DESCRIPTIONS[IS_NULL]` / `[IS_NOT_NULL]` use label-specific wording ("Label key does not exist" / "...exists with any value") in a table shared across all fields. Fine while these operators are label-only, but worth a more generic phrasing if another field ever adopts them.
- The JDBC "Map" branch has no `CONTAINS` / `NOT_CONTAINS` case, so a hand-crafted URL like `filters[labels][CONTAINS][key]=v` would 500 instead of failing cleanly. Not reachable through the normal UI, purely defensive.

### CI

`Backend - Tests` is currently red on `MysqlRunnerTest.waitforNestedThreeLevels()`, a `waitFor` timing test unrelated to anything this PR touches. It looks like a flake rather than a regression, but worth a rerun to get it green before merge.

Thanks again for the solid work here, the label-filtering UX is a great improvement. Happy to re-review as soon as the dashboard label path honors the selected operator. 🙌

hemantmm:
Thanks for the detailed review @flcarre, and for tracing the flow end-to-end! The dashboard label filter issue makes sense—I missed that `IExecutions.whereWithGlobalFilters` was still hardcoding `Contains`. I'll update it to honor the selected operator, remove `NOT_CONTAINS` from the dashboard until it's supported there, and I'll also check the `flowId` `NOT_EQUALS` changes (that wasn't intentional if it came from a merge). Thanks again!

MilosPaunovic:
@hemantmm Can you resolve the conflicts here please?

loicmathieu (approved):
@hemantmm backend size LGTM, thanks a lot, nice PR.
One note for the future (don't update the PR) for multi-line sring you can use texblock instead of string concatenation: easier to read.

@flcarre can you make a last review on the frontend change?
I'll tackle the EE side for Elasticsearch.

flcarre:
Thanks a lot for taking this on, @hemantmm — this is a big, well-structured change and the repository-side filtering is in great shape: the new operators are wired cleanly across H2/MySQL/Postgres for both flows and executions, all user input goes through bound parameters (no injection surface), the JSON/jq escaping is reused correctly, and the new `AbstractFlowRepositoryTest`/`AbstractExecutionRepositoryTest` cases give solid coverage. The frontend side is tidy too — the new comparators are translated in all 13 locales and the NULL-operator input handling in `FilterEditPopper` is correct.

There are two issues that are causing the red `Backend - Tests` run, plus one design question. Once these are sorted I think this is very close.

## 🔴 1. `QueryFilterTest` — the test edit is the problem, not the enum

The failing case is `[28] field=LABELS, operation=NOT_EQUALS, resource=FLOW` in the *invalid* set.

`NOT_EQUALS` on labels is still a real, supported operation: it's listed in `QueryFilter.Field.LABELS.supportedOp()`, implemented in every repository service, exercised by `AbstractFlowRepositoryTest#shouldFilterFlowsWithNotEqualsLabelOperator`, and documented in the `FlowController`/`ExecutionController` OpenAPI parameters. Validation (`collectValidationErrors`) checks support resource-agnostically via `field().supportedOp().contains(op)`, so moving `NOT_EQUALS` into the invalid set means the test expects an exception that will never be thrown.

The UI no longer offering `NOT_EQUALS` for labels is totally fine — the backend can keep supporting an operator the UI doesn't surface. The fix is just to revert the two test hunks in `QueryFilterTest.java`:

- keep `Op.NOT_EQUALS` in the **valid** set for `Field.LABELS, Resource.FLOW`, and
- drop the `Op.NOT_EQUALS` line you added to the **invalid** set for the same field/resource.

No change to `QueryFilter.java` is needed here.

## 🔴 2. Dashboard label filters generate invalid SQL for every operator except CONTAINS

This is what fails `DashboardControllerTest.shouldPreviewChartFilteredByLabels()`:

```
Internal server error: SQL [ ... "labels" = cast(? as varchar) ... ]; Column "labels" not found
```

`IExecutions.labelFilter()` now builds `EqualTo` / `In` / `NotIn` / `IsNull` / `IsNotNull` filters on `Fields.LABELS`. But on the dashboard/JDBC path, only `Contains` on `LABELS` gets the special JSON handling — see `AbstractJdbcExecutionRepository.where()` (it pulls out `Contains<...LABELS>` filters) and `JdbcFilterService` (its `containsCondition` delegates label filters to the per-DB `findLabelCondition`). Every other filter type on `LABELS` falls through to generic scalar handling and resolves `LABELS` to the plain column name `labels`, which doesn't exist as a scalar column (labels live inside the JSON `value` column) — hence the error.

The fix is to give `EqualTo`/`In`/`NotIn`/`IsNull`/`IsNotNull` on `LABELS` the same treatment `Contains` already gets: detect the label field and delegate to the per-DB `findLabelCondition(Either<Map,String>, Op)` (already implemented) instead of building a generic `field(field).op(value)` condition, and generalize the label-extraction in `AbstractJdbcExecutionRepository.where()` so it recognizes the new filter types, not just `Contains`.

Worth noting: `DashboardDataGlobalFiltersTest.shouldMapLabelFiltersForExecutions` only asserts the filter *objects* are constructed correctly — it never runs them through the SQL backend, so it can't catch this. A test that executes the resulting query (like `shouldPreviewChartFilteredByLabels` does) is the reproduction case.

## 🟠 3. `IN` means two different things depending on the surface

- **Repository/API path** (`filters[labels][IN][key]=value` on the Flows/Executions lists): `IN` over a `Map` is an **OR across key/value pairs** — "has any of these labels" (confirmed by your new `inFilter = Map.of("foo","bar","unknown","value")` test expecting `flow-with-label`).
- **Dashboard path** (`IExecutions.asValues()`): `IN` treats the map value as a **comma-separated list of values for a single key** (`Map.of("environment","prod,dev")` → `In(key="environment", values=["prod","dev"])`).

Same operator, same field, different results across surfaces. Right now #2 hides this (the dashboard path doesn't execute), but once #2 is fixed a user will get different results filtering a dashboard vs. a list with an equivalent-looking value. Could you pick one interpretation (or document the divergence intentionally)? The comma-list-per-key reading maps most naturally to the UI's "has any of" chip.

## 🟡 Minor (non-blocking, up to you)

- No SQL-injection-test coverage for the three new operators (`NOT_CONTAINS`, `IS_NULL`, `IS_NOT_NULL`) in `AbstractSQLInjectionTest` — risk is low since everything is bound, but nice to add.
- New exception messages use `+` concatenation; the project convention (AGENTS.md) is `"...: %s.".formatted(op)`. Matches the surrounding legacy style, so purely a fix-forward.
- Tiny: trailing comma in `KV_COMPARATORS`, and `COMPARATOR_LABELS` hardcodes English for the 3 new entries (consistent with the existing entries in that map).

Nothing on the security or performance side — the new code is clean there.

Happy to help verify once you push the fixes. Thanks again for the contribution! 🙏

loicmathieu:
I also think `NOT_CONTAINS` can be wired from the UI but is not supported in the backend...
Anyway, would be a deeper QA before merging.

flcarre:
A quick follow-up focused on the **frontend** side 🎨 — overall it's nicely done, so mostly a few consistency points here.

**What's great:**
- The operator encode/decode needed zero plumbing changes because it's name-based (`Comparators[operation]` on decode, `keyOfComparator` reverse-lookup on encode), so `NOT_CONTAINS`/`IS_NULL`/`IS_NOT_NULL` flow through cleanly — nice that the design held up.
- Input rendering in `FilterEditPopper.vue` is correct on the list pages: `IN`/`NOT_IN`/`EQUALS` → key-value UI, `CONTAINS`/`NOT_CONTAINS` → single text, `IS_NULL`/`IS_NOT_NULL` → key-only (via `NULL_COMPARATORS`).
- The `label_comparators` keys are added across all 13 locales — thanks for not skipping that.

**🟠 The 6 "labels" filter configs have drifted apart.** Same concept, three different setups:

| Config | valueType | Operators |
|---|---|---|
| `useExecutionFilter` / `useFlowFilter` / `useFlowExecutionFilter` | `key-value` | IN, NOT_IN, EQUALS, CONTAINS, **NOT_CONTAINS**, IS_NOT_NULL, IS_NULL |
| `useDashboardFilter` | `key-value` | IN, NOT_IN, EQUALS, CONTAINS, IS_NOT_NULL, IS_NULL (**NOT_CONTAINS missing**) |
| `useNamespaceDashboardFilter` / `useFlowDashboardFilter` | **`text`** | EQUALS, CONTAINS, IS_NOT_NULL, IS_NULL |

Two things worth aligning:
- `NOT_CONTAINS` is offered on the Executions list but not on the Executions dashboard — looks like an oversight since the backend supports it there too.
- `valueType: "text"` on the namespace/flow dashboard label filters means `EQUALS` renders a single value input and encodes `filters[labels][EQUALS]=<string>` with no key, whereas labels are key-value everywhere else. Worth switching these to `key-value` for consistency (and so `EQUALS` actually carries a key).

**🟡 Test coverage — nice start, one gap.** `filterChipFactory.test.ts` round-trips `CONTAINS`/`NOT_CONTAINS`/`IS_NULL`/`IS_NOT_NULL`, but not `IN`/`NOT_IN`, which are the trickiest (multi-value, bucketed by operation in `LabelBucket`). A couple of cases there would lock down the most fragile path.

**🟢 Minor:**
- `COMPARATOR_LABELS` in `filterTypes.ts` hardcodes the English strings for the 3 new entries (`"Does Not Contain"`, `"Is Set"`, `"Is Not Set"`). It's overridden by `labelComparatorLabels()` in the configs and it matches the existing entries in that map, so purely a fix-forward toward the i18n convention — the `COMPARATOR_DESCRIPTIONS` next to it already use i18n keys, for reference.
- Tiny trailing comma in `KV_COMPARATORS`.

Really solid frontend work overall — these are mostly about making the labels filter feel identical across every surface it appears on. 🙌

hemantmm:
Thanks a lot for the detailed review!

I pushed fixes for the backend and frontend points:

- Restored `LABELS + NOT_EQUALS` as a valid backend operation in `QueryFilterTest`.
- Updated dashboard execution label filtering so `EQUALS`, `IN`, `NOT_IN`, `IS_NULL`, `IS_NOT_NULL`, `CONTAINS`, and `NOT_CONTAINS` all go through the DB-specific label JSON condition path instead of falling back to the scalar `labels` column.
- Added dashboard support for `NOT_CONTAINS`.
- Aligned dashboard label filter configs so labels use key-value input consistently, and exposed `NOT_CONTAINS` on execution dashboards.
- Added/updated tests for dashboard label mapping and label filter chip round-trips, including `IN` / `NOT_IN`.

For the `IN` semantics, I kept the repository/API interpretation: `IN` over label maps means “has any of these labels”, and the dashboard mapping now preserves that with an OR.

Verified with:

- `./gradlew :core:test --tests io.kestra.core.models.QueryFilterTest --tests io.kestra.plugin.core.dashboard.data.DashboardDataGlobalFiltersTest`
- `./gradlew :webserver:test --tests io.kestra.webserver.controllers.api.DashboardControllerTest.shouldPreviewChartFilteredByLabels`

Thanks again for the thorough review! Happy to make any further adjustments if needed. 🙏

hemantmm:

flcarre:
Re-checked the **frontend** after your latest push — everything I raised is addressed. Nice work. ✅

- `NOT_CONTAINS` is now offered on the Executions dashboard too.
- `valueType` switched from `text` to `key-value` on the namespace and flow dashboard label filters.
- `filterChipFactory.test.ts` now covers `IN`/`NOT_IN` chip restoration (the array shape `["team:core"]`) — that was the fragile path, good to have it locked down.
- Trailing comma in `KV_COMPARATORS` cleaned up.

All 6 `labels` filter configs are consistent now — same `valueType`, same `comparatorLabels`, same `showComparatorSelection`. The remaining operator difference (namespace/flow dashboards expose 5 operators instead of 7, no `IN`/`NOT_IN`) reads as intentional and matches what those surfaces support, so no concern there.

i18n is complete: the new `label_comparators` and `comparator_descriptions` keys are present in all 13 locales. Both frontend CI jobs are green (`Frontend - Tests` and `Frontend - Design System Tests`).

One leftover I'd explicitly **not** fix here: `COMPARATOR_LABELS` in `filterTypes.ts` still hardcodes the English strings for the 3 new entries. All 16 entries in that map are in the same state and it's always overridden by `labelComparatorLabels()` in the configs, so migrating it to i18n keys is a separate cleanup — out of scope for this PR.

**Frontend side: LGTM** 🎉 (the backend feedback in my other comment still stands.)

flcarre (approved):
pls rebase and fix your conflicts

hemantmm:
Thanks again for the review @flcarre! One question regarding the remaining backend feedback: for the `LABELS` `IN` operator, would you prefer that the dashboard semantics be aligned with the repository/API semantics, or is the current divergence acceptable if it's documented? I can update the implementation either way—I just want to make sure it matches the intended behavior.

PR #17478, opened 2026-07-17
feat: new WorkerTaskResultListener listener

Skraye:
Add a WorkerTaskResultListener extension point so anything can react to joined worker task results without replacing the executor.

  - WorkerTaskResultMessageHandler now notifies registered WorkerTaskResultListener beans when a result is joined (present result); listener exceptions are logged and never fail execution processing.
  - No behavior change for OSS


loicmathieu (approved):
LGTM

PR #17488, opened 2026-07-17
fix(build): renormalize gradlew.bat line endings

elevatebart:
Renormalizes gradlew.bat's git-stored line endings (LF-in-blob per `eol=crlf`) to stop the alien CRLF blob from causing a permanent phantom diff; checked-out file stays CRLF.


loicmathieu (approved):
LGTM

PR #17494, opened 2026-07-17
refactor: migrate execution tabs to router-driven child routes

elevatebart:
### ✨ Description

Migrates the Executions, Flows, and Namespaces detail pages from a flat `:tab?` route param + dynamic component switching to real vue-router child routes — each tab gets its own linkable, ctrl/cmd-clickable URL, and layout (top bar, tab shell) is shared automatically via `<router-view>`.

- `executionTabs.ts`, `flowTabs.ts`, `namespaceTabRoutes.ts` are now the single source of truth for each page's tabs.
- Added `useActiveTab()` (resolves the active tab from the matched child route, falling back to the legacy `:tab` param) and `routeFamily()` (normalizes `<page>/<tab>` back to `<page>` for identity checks), so the rest of the app keeps working whether a page has migrated or not.
- Fixed every navigation site that pushed a `tab` param at a now-migrated route — vue-router silently drops it once the parent's own path no longer declares `:tab?` — to target the child route by name instead.

### 🔗 Related Issue

N/A

### 🎨 Frontend Checklist

- [x] Type-check, lint, and unit tests pass
- [ ] E2E tests pass — one test is currently under investigation (see Additional Notes)

### 📝 Additional Notes

Companion PR: kestra-io/kestra-ee#9380, which applies the same pattern to kestra-ee's own tabbed pages.

One E2E test (`flow.spec.ts` — switching to the Flow's Overview tab via the mobile tab-select dropdown) is failing in kestra-ee's CI; a targeted component test confirmed the core router-view tab-switching mechanism itself is sound, so this needs a live-browser repro to confirm whether it's a real regression or a Monaco-disposal timing flake. Tracked separately, not blocking review of this PR's approach.

🐱


ui/src/routes/routes.ts
```diff
@@ -45,7 +46,18 @@ const routes: RouteRecordRaw[] = [
...
+        path: "/:tenant?/executions/:namespace/:flowId/:id",
+        component: () => import("../components/executions/ExecutionRoot.vue"),
+        // Resolve legacy deep-links `{name: "executions/update", params: {tab}}` and bare
+        // `/:id` URLs to the matching child route, preserving params and query.
+        redirect: (to) => {
+            const tab = (to.params.tab as string) || localStorage.getItem("executeDefaultTab") || "overview"
+            return {name: `${EXECUTION_PARENT_ROUTE}/${tab}`, params: to.params, query: to.query}
+        },
+        children: EXECUTION_TAB_ROUTES,
+    },
```
elevatebart: Could all of this live in `executionTabs.ts`? We could call it `executionRoutes.ts`. It would make each page responsible for its own tabs routes and children. Flow, Tests, Apps would have the same architecture.

elevatebart: Good call — done in 1fb846036: moved the parent route (path/component/redirect/children) into `executionTabs.ts` as `EXECUTION_ROUTE`, so `routes.ts` now just references it and the Executions page owns its full routing structure.

Kept the filename as `executionTabs.ts` rather than renaming to `executionRoutes.ts` for now — `kestra-ee`'s `executionTabs.ts` imports `EXECUTION_PARENT_ROUTE`/`EXECUTION_TAB_ROUTES` from this file via a literal path (`kestra/src/components/executions/executionTabs`), since it composes its own parent route with EE-specific children rather than reusing `EXECUTION_ROUTE` directly. A rename would need a coordinated companion change there, so I'd rather do that (and the broader Flow/Tests/Apps convention) as a separate follow-up than fold it into this pilot.

---

flcarre (approved):

flcarre:
Re-reviewed after your latest push — thanks for the quick turnaround. 🙏

✅ **Fixed:** the namespace *Policies* tab (`plugin-defaults` → `policies`) and the `audit-logs` demo-tab key. Confirmed live: `/namespaces/edit/<ns>/policies` now renders instead of 404-ing.

❌ **Still reproducing — the namespace tab bar is dead when you land on the bare parent URL.**

Steps (live OSS instance built from this branch's HEAD):
1. Open `/namespaces/edit/<ns>` with **no tab segment** — the exact target of the Dashboard namespace-column link (`{name: 'namespaces/update', params: {id}}` in `dashboard/sections/table/columns/Namespace.vue`, unchanged).
2. The tab column renders (Overview … Audit Logs) but **all 13 links resolve to the same bare URL** — clicking *Dependencies*, *Flows*, etc. is a silent no-op (the URL never changes).

Bare namespace parent — every tab link resolves to the same bare URL, so tab clicks are no-ops

Root cause is the one from the first pass: the namespace parent route has no default-child `redirect` (unlike `EXECUTION_ROUTE` / `FLOW_ROUTE`), so `route.meta.tab` is `undefined`, `isRouterDriven` is `false` in `RouteTabsSidebar.routeFor`, and vue-router drops the `tab` param against a path that no longer declares it. Control: entering via a real tab (e.g. `/flows`) works fine — the links are distinct and correct.

Either fix resolves it (the redirect is the robust one):

```ts
// ui/src/routes/routes.ts — mirror EXECUTION_ROUTE / FLOW_ROUTE
{
    name: NAMESPACE_PARENT_ROUTE,
    path: "/:tenant?/namespaces/edit/:id",
    component: () => import("../components/namespaces/Namespace.vue"),
    redirect: (to) => ({
        name: `${NAMESPACE_PARENT_ROUTE}/${(to.params.tab as string) || "overview"}`,
        params: to.params,
        query: to.query,
    }),
    children: createNamespaceTabRoutes(),
},
```

…and/or point the Dashboard column link at a real child (`namespaces/update/overview`), the way `override/components/namespaces/Namespaces.vue` already does.

Also still open from the first review (not re-tested this pass): the flow-tab permission gating is applied to the tab *bar* only, not to what `<router-view>` renders — a user without permission can deep-link `/flows/edit/<ns>/<id>/triggers` and mount the component. Worth a `beforeEnter` guard on the flow tab routes.

<sub>QA'd on a live OSS instance built from this branch.</sub>

elevatebart:
Good catch, confirmed and fixed in `7c54391eb` — added the missing `redirect` to `NAMESPACE_PARENT_ROUTE` in `routes.ts`, mirroring `EXECUTION_ROUTE`/`FLOW_ROUTE`, defaulting to `overview`. Went with the redirect (not the Dashboard-link change) since it fixes every bare-parent entry point at the source, not just that one caller. Verified `check:types`/lint/unit tests clean.

Re: the flow-tab permission-gating gap (bar-only filtering, no route guard on `.../triggers` etc.) — agreed that's real and worth fixing, but it's a separate, larger design decision (what should happen on a denied deep-link: redirect to overview, show a 403, etc.) rather than a one-line fix, and you flagged it as not re-tested this pass. Happy to take it on as a follow-up if you'd like — let me know.

---

PR #17497, opened 2026-07-17
perf(core): load @types/node and Monaco typings registration lazily

elevatebart:
### ✨ Description

`main.ts` eagerly inlined the entire `@types/node` tree as raw strings (`import.meta.glob` with `eager: true`) and kicked off a Monaco import at boot for every user, editor page or not. The typings alone accounted for **2.4 MB of the 2.65 MB entry chunk** (~444 KB gzipped on the critical path).

This PR (updated per review):
- moves the typings glob into a dedicated `monacoNodeTypes.ts` module (eager `?raw` glob) that is **dynamically imported** from `monacoEnvironment.ts`, so Vite bundles the whole `@types/node` tree into a **single lazy chunk** — off the boot critical path, and one request instead of ~60 separate chunks;
- registers the extra libs from `MonacoEnvironment.getWorker` the first time a `javascript`/`typescript` worker is requested — the earliest point a user actually needs node typings;
- **imports the Monaco TypeScript language contribution** (`monaco-editor/esm/vs/language/typescript/monaco.contribution`) in both the lazy loader and `KsEditor`. Monaco 0.52's ESM build leaves `languages.typescript` undefined until this module is imported — without it the typings registration (and `KsEditor`'s `setCompilerOptions`) silently no-op'd. This fixes a **regression** in ambient-type completions introduced by the Monaco update.

**Measured (production build):**

| | before | after |
|---|---|---|
| entry chunk (`index-*.js`) | 2,648 KB (gzip 518 KB) | **~249 KB (gzip ~74 KB)** |
| `@types/node` payload | in entry chunk, fetched at boot | **1 lazy chunk** (`monacoNodeTypes-*.js`, 2,468 KB / gzip 441 KB), fetched on first JS/TS editor use, not preloaded |

### 🔗 Related Issue

None — follows up on an internal frontend performance audit.

### 🎨 Frontend Checklist

- [ ] Screenshots or video recordings attached showing the `UI` changes — no visual changes; verified nothing typings-related appears in `index.html` preloads

### 📝 Additional Notes

- The single-chunk approach keeps the typings off the boot path while collapsing ~60 network requests into one; the chunk only loads when a JS/TS editor is first used.
- The `editor.api` chunk itself is still `modulepreload`ed because app code (`KsEditor`, the monaco language-configurator composables) imports it statically — untangling that is a separate, larger refactor worth its own issue.

https://claude.ai/code/session_014kPcgZ5tmBYQFfZJ69e5Mj

### DIFF
--- a/ui/src/utils/monacoEnvironment.ts
+++ b/ui/src/utils/monacoEnvironment.ts
@@ -3,23 +3,18 @@ import JsonWorker from "monaco-editor/esm/vs/language/json/json.worker?worker"
 import TypeScriptWorker from "monaco-editor/esm/vs/language/typescript/ts.worker?worker"
 import YamlWorker from "../components/inputs/yaml.worker.js?worker"
 
-const nodeTypesRaw = import.meta.glob("/node_modules/@types/node/**/*.d.ts", {query: "?raw", import: "default"}) as Record<string, () => Promise<string>>
-
 let nodeTypesRequested = false
 
-// Registers @types/node into Monaco's TS service on first JS/TS worker request — both
-// the types and the TS language contribution are multi-MB payloads that must not load
-// at boot. Importing the contribution is what populates `languages.typescript` (Monaco's
-// ESM build leaves it undefined otherwise).
+// Lazily registers @types/node on first JS/TS worker request. The TS contribution import
+// is what populates `languages.typescript` (Monaco's ESM build leaves it undefined otherwise).
 function loadNodeTypes() {
     Promise.all([
         import("monaco-editor/esm/vs/language/typescript/monaco.contribution"),
         import("monaco-editor/esm/vs/editor/editor.api"),
-    ]).then(([, {languages}]) => {
-        for (const path in nodeTypesRaw) {
-            nodeTypesRaw[path]().then((content) => {
-                languages.typescript.typescriptDefaults.addExtraLib(content, `file://${path}`)
-            })
+        import("./monacoNodeTypes"),
+    ]).then(([, {languages}, {default: nodeTypes}]) => {
+        for (const path in nodeTypes) {
+            languages.typescript.typescriptDefaults.addExtraLib(nodeTypes[path], `file://${path}`)
         }
     })
 }
--- a/ui/src/utils/monacoEnvironment.ts
+++ b/ui/src/utils/monacoEnvironment.ts
@@ -3,6 +3,28 @@ import JsonWorker from "monaco-editor/esm/vs/language/json/json.worker?worker"
 import TypeScriptWorker from "monaco-editor/esm/vs/language/typescript/ts.worker?worker"
 import YamlWorker from "../components/inputs/yaml.worker.js?worker"
 
+let nodeTypesRequested = false
+
+// Registers @types/node into Monaco's TypeScript language service the first
+// time a JS/TS worker is requested. Both Monaco and the raw typings are
+// multi-megabyte payloads that must not load at boot; extra libs added after
+// worker creation still propagate to open models. The retry loop waits for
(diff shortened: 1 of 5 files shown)


ui/src/utils/nodeTypes.ts
```diff
@@ -0,0 +1,4 @@
+// Several megabytes of raw @types/node typings for the Monaco TypeScript
+// language service. Only ever import this module dynamically — a static
+// import would inline the whole payload into the importer's chunk.
+export default import.meta.glob("/node_modules/@types/node/**/*.d.ts", {eager: true, query: "?raw", import: "default"}) as Record<string, string>
```
elevatebart: Why can't we import this directly in ui/src/utils/monacoEnvironment.ts ? eager:true could be removed.

elevatebart: Two separate reasons:

**Why not in `monacoEnvironment.ts`:** that file is statically imported by `main.ts` (the worker registrations must exist before any editor mounts), so its code lands in the boot-loaded chunk. An `eager: true` glob there would inline the raw typings right back into the entry bundle — the whole point of the dedicated module is to be a boundary that is only ever imported dynamically, so Vite emits the payload as its own lazy chunk.

**Why keep `eager: true`:** dropping it does work mechanically — the glob then returns lazy thunks (tiny), which could indeed live directly in `monacoEnvironment.ts`. The trade-off is that Vite emits one chunk per matched file: `@types/node` currently matches **88 `.d.ts` files**, so first editor use would trigger 88 requests/chunks instead of one 2.4 MB (~400 KB gzip) chunk fetched once.

Happy to switch to the non-eager shape if you prefer fewer files in the repo over fewer chunks in the build — just say the word.

---

ui/src/utils/monacoEnvironment.ts
```diff
@@ -3,6 +3,28 @@ import JsonWorker from "monaco-editor/esm/vs/language/json/json.worker?worker"
 import TypeScriptWorker from "monaco-editor/esm/vs/language/typescript/ts.worker?worker"
 import YamlWorker from "../components/inputs/yaml.worker.js?worker"
 
+let nodeTypesRequested = false
+
+// Registers @types/node into Monaco's TypeScript language service the first
```
elevatebart: Make this a 2 line comment

ui/src/utils/monacoEnvironment.ts
```diff
@@ -3,6 +3,25 @@ import JsonWorker from "monaco-editor/esm/vs/language/json/json.worker?worker"
...
 import YamlWorker from "../components/inputs/yaml.worker.js?worker"
 
+let nodeTypesRequested = false
+
+// Registers @types/node into Monaco's TS service on first JS/TS worker request — both
+// are multi-MB payloads that must not load at boot. Retries until languages.typescript exists.
+function loadNodeTypes(tries = 0) {
+    Promise.all([
+        import("monaco-editor/esm/vs/editor/editor.api"),
+        import("./nodeTypes"),
```
elevatebart:
```suggestion
        import.meta.glob("/node_modules/@types/node/**/*.d.ts", {query: "?raw", import: "default"}),
```

elevatebart: Applied in 584992f — with one adjustment: the non-eager glob returns loader thunks (`() => Promise<string>`), not raw strings, so it lives as a module-level const and the registration loop awaits each file before calling `addExtraLib`. `nodeTypes.ts` is deleted. Verified with a production build: entry chunk stays at ~249 KB and the typings ship as 61 lazy chunks fetched on first JS/TS editor use, none preloaded.

---

Piyush-r-bhaskar (approved):

PR #17500, opened 2026-07-17
fix(scripts): clean up leaked Docker container on interrupt during setup

loicmathieu:
## Summary

`io.kestra.plugin.scripts.runner.docker.DockerTest.interruptAfterResume()` has been `@FlakyTest`-quarantined and fails often in CI — e.g. on [PR #17485](https://github.com/kestra-io/kestra/pull/17485), with `AssertionError: Expected: is <true> but: was <false>` on the test's final "no leftover containers" assertion.

**Root cause:** `Docker.run()` only guaranteed container/volume cleanup (kill + optional delete) via a `try/finally` wrapping the attach-and-wait phase. Everything before that — resume-container lookup, image pull, container creation, volume/file handling, container start, and the resumed-path's volume lookup (`listVolumesCmd()`) — had no cleanup safety net. If the executing thread was raw-interrupted (`Thread.interrupt()`, as opposed to the graceful `TaskRunner.kill()` API) while blocked on a Docker HTTP call in that window, the resulting exception propagated straight out of `run()`, leaking the container. This explains why the sibling test `killAfterResume()` (same file, never flaky) is reliable: it calls the graceful `.kill()` API instead of raw `Thread.interrupt()`, so it never exercised this gap.

## Changes

- `Docker.java`: extracted the cleanup logic into a shared `cleanupContainer(...)` helper, and wrapped the container-resolution/pre-attach setup phase in a `try/catch(Exception)` that runs the same cleanup on any exception before rethrowing — mirroring the existing `finally` used for the attach/wait phase. This closes the leak for both the create path (image pull / create / volume-populate / start) and the resume path (volume lookup), all previously unprotected.
- `DockerTest.java`: removed `@FlakyTest` from `interruptAfterResume()` now that cleanup is deterministic regardless of where the interrupt lands.

## Test plan

- [x] `:script:compileJava` / `:script:compileTestJava` — clean
- [x] `:script:test --tests ...DockerTest.interruptAfterResume` — run 5x against a real Docker daemon (previously flaky) — all pass
- [x] `:script:test --tests ...DockerTest` (full class, 10 tests) — all pass, including `killAfterResume()` (no regression)
- [x] Verified no containers leaked by any test run (`docker ps -a` before/after)
- [x] Code-reviewed; one indentation nit found and fixed via `spotlessApply`

### DIFF
--- a/script/src/main/java/io/kestra/plugin/scripts/runner/docker/Docker.java
+++ b/script/src/main/java/io/kestra/plugin/scripts/runner/docker/Docker.java
@@ -417,123 +417,133 @@ public class Docker extends TaskRunner<Docker.DockerTaskRunnerDetailResult> {
             String filesVolumeName = null;
             var strategy = runContext.render(this.fileHandlingStrategy).as(FileHandlingStrategy.class).orElse(null);
 
-            // pull image only if we will create a new container
-            if (containerId == null) {
-                var renderedPolicy = runContext.render(this.getPullPolicy()).as(PullPolicy.class).orElseThrow();
-                if (!PullPolicy.NEVER.equals(renderedPolicy)) {
-                    pullImage(dockerClient, image, renderedPolicy, logger);
-                }
+            // Everything below that can leave a container behind (create or resume) must clean it up if it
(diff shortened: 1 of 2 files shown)


Malaydewangan09 (approved):
LGTM, thanks!

PR #17495, opened 2026-07-17
feat(execution): possible NoSuchElementException on State.getEndDate

loicmathieu:
…il()

Covers getStartDate(), getEndDate() (terminated/paused/running branches), getDuration(), getDurationOrComputeIt(), humanDuration(), maxDate(), minDate(), and State.Type.fail(Task).

Fixes #17055

PR #17501, opened 2026-07-17
fix(triggers): fix orphan-GC race in unlock and restart, validate flow exists first

loicmathieu:
## Summary

`TriggerControllerTest.shouldUnlockTriggerWhenLocked()` failed often in CI — e.g. on [PR #17495](https://github.com/kestra-io/kestra/pull/17495) — with:
```
io.micronaut.http.client.exceptions.HttpClientResponseException: Not Found: Trigger disappeared after unlock: [tenant=main, namespace=..., flow=..., trigger=...]
```
A prior fix attempt ([#17473](https://github.com/kestra-io/kestra/pull/17473)) tried seeding the test's flow-less trigger with a far-future `nextEvaluationDate`, but it didn't close the race — here's why.

**Root cause:** `TriggerStateService.unlockTriggerById` sends a `ResetTrigger` event, then re-reads the trigger. The scheduler's `onResetTrigger` handler calls `state.reset(clock)`, which *unconditionally* nulls `nextEvaluationDate`, and only recomputes a real date `if` the trigger's flow/definition is found. For a flow-less trigger that recompute is skipped, leaving `nextEvaluationDate=null` the instant it's unlocked — regardless of whatever date the fixture originally seeded (that's why the prior fix didn't help: unlocking itself wipes it). The scheduler's eligibility query treats `NULL` as "eligible now", so the very next poll picks up the now-eligible, still-flow-less trigger, finds no backing flow, and deletes it as an orphan — racing the webserver's own re-read.

## Changes

1. **Test fix** — give the trigger fixture a real backing flow (mirroring the established `shouldDeleteTriggerWhenExists` pattern: create the flow, await the scheduler's own async auto-creation of the matching `TriggerState` row, then lock it). With a real flow, `onResetTrigger` properly recomputes `nextEvaluationDate` from the trigger's own schedule instead of leaving it null — closing the race entirely, not just narrowing it.
2. **Production hardening** — `unlockTriggerById` now validates the flow/trigger-definition still exists (`validateToggleable`) before sending the reset, failing fast with 404 for a genuinely orphaned trigger instead of racing the orphan-GC. Extended the same guard to the batch `unlockAllByIds`/`unlockAllMatching` endpoints for parity (their existing sibling endpoints, `toggleAllByIds`/`toggleAllMatching`, already had this guard). Added a dedicated test for the new 404 behavior.
3. **Restart endpoint parity** — `TriggerStateService.resetTrigger` (the `/restart` endpoint) sends the same `ResetTrigger` event through the identical scheduler handler, so it shares the exact same race and had no guard either. Added the same `validateToggleable` fail-fast check before the kill event is emitted. Fixed `shouldRestartTriggerWhenExists`, whose fixture trigger id didn't correspond to any real trigger definition in its flow (and would now correctly 404 under the guard), to use the flow-backed fixture helper, and added a dedicated test for the new 404 behavior.

Two of the batch-unlock tests used flow-less fixtures expecting a successful unlock count — updated them to use a real flow too, since a flow-less trigger is now correctly recognized as orphaned rather than silently accepted.

## Test plan

- [x] `:webserver:compileJava` / `:webserver:compileTestJava` — clean
- [x] `:webserver:test --tests ...TriggerControllerTest.shouldUnlockTriggerWhenLocked` — run 5x (previously flaky) — all pass
- [x] `:webserver:test --tests ...TriggerControllerTest` (full class, 23 tests) — all pass, including the unlock and restart orphan-404 tests and the fixed batch/restart tests, no regressions

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/services/TriggerStateService.java
+++ b/webserver/src/main/java/io/kestra/webserver/services/TriggerStateService.java
@@ -75,7 +75,7 @@ public class TriggerStateService {
      *
      * @param trigger the trigger identifier.
      * @return the refreshed trigger state.
-     * @throws NotFoundException if the trigger does not exist.
+     * @throws NotFoundException if the trigger, its flow, or the trigger definition within that flow does not exist.
      * @throws ConflictException if the trigger is already unlocked, is a realtime trigger, or the reset failed.
      */
     public TriggerState unlockTriggerById(final TriggerId trigger) throws NotFoundException, ConflictException {
@@ -88,6 +88,9 @@ public class TriggerStateService {
             // scheduler submit a second instance while the first is still running on a worker.
             throw new ConflictException("trigger %s is a realtime trigger, reset it to kill and restart it".formatted(trigger));
         }
+        // Fail fast if the trigger is orphaned otherwise the reset would race the scheduler's orphan-GC.
+        validateToggleable(trigger);
+
         awaitBlockingAction(
             trigger.uid(),
             operationId -> triggerEventQueue.send(new ResetTrigger(trigger).withOperationId(operationId)),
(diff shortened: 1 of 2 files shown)


fhussonnois (approved):
LGTM!

PR #17504, opened 2026-07-17
feat(cases): QueryFilter/JDBC support + UI foundation for Cases (EE)

brian-mulier-p:
- QueryFilter `Resource.CASE` / `Resource.CASE_TEMPLATE` + `Field.SEVERITY` / `Field.ASSIGNEE` for the EE Cases incident-management feature
- Extension-point bridge in `Executions.vue` / `Overview.vue` so EE can inject a Cases column, bulk actions, and a linked-cases panel without forking these components
- `demo/Cases.vue` upsell page + route + menu teaser for OSS users without the EE feature
- Design-system: `KsSelect`/`KsOption` colorMap mechanism (severity/status badge coloring) + `KsSegmented` icon slot forwarding (Cases list/board view toggle)
- Executions `timeRange` filter fix + regression test

Companion to kestra-io/kestra-ee's Cases PR.


ui/src/components/executions/Executions.vue
```diff
@@ -781,7 +804,7 @@
             ? executionFilter.value
             : flowExecutionFilter.value
         const fields = (configuration.keys ?? []).flatMap((entry: {key: string}) =>
-            entry.key === "timeRange" ? ["startDate", "endDate"] : [entry.key],
+            entry.key === "timeRange" ? ["timeRange", "startDate", "endDate"] : [entry.key],
```
MilosPaunovic, line 808: `loadQuery` feeds the **query bulk actions** (`genericConfirmCallback`, line ~870), so this now forwards `timeRange` *alongside* the `startDate`/`endDate` it already sent, on every Select-All bulk action.

The two E2E failures on this PR are exactly the two `execution-bulk-actions.spec.ts` *"when using Select All"* tests, both timing out. That's a strong enough correlation that I'd want it ruled out before merge — could the backend bulk endpoints be rejecting or mis-resolving the redundant `timeRange` + explicit-dates combination?

If the fix is only needed for the *filter display* and not for the bulk-action query, it might be cleaner to keep `timeRange` out of `loadQuery`'s output and add it only where the filter UI reads the supported fields.

ui/src/components/executions/Executions.vue
```diff
@@ -263,6 +274,17 @@
                     </KsTooltip>
                 </template>
             </KsTableColumn>
+
+            <KsTableColumn
+                v-for="col in extraColumns"
```
MilosPaunovic: These extra columns are rendered outside the `useTableColumns` / `visibleColumns` mechanism that every other optional column goes through, so the EE-injected Cases column is permanently visible and is the one column a user can't toggle off in the column selector.

Could `extraColumns` be merged into `optionalColumns` instead (with `default: true`) so it participates in the selector and the `DISPLAY_EXECUTIONS_COLUMNS` persistence like the rest?

ui/packages/design-system/src/components/Form/KsSelect/KsSelect.vue
```diff
@@ -15,18 +15,22 @@
         <template v-if="$slots.label" #label="p">
             <slot name="label" v-bind="p" />
         </template>
+        <template v-else-if="colorMap" #label="p">
+            <span class="kel-select-color-option" :style="{color: colorMap[p.value]}">{{ p.label }}</span>
```
MilosPaunovic: Small asymmetry with `KsOption`: this fallback guards on **`colorMap` existing**, while `KsOption` guards on **the entry existing** (`v-else-if="color"`).

So for a value that isn't in the map, the collapsed label renders `.kel-select-color-option` — which unconditionally draws the `::before` bullet in the inherited text colour — while the matching dropdown row renders a plain label with no bullet. Same value, dot in one place and not the other.

`v-else-if="colorMap?.[p.value]"` would keep the two in sync.

ui/packages/kestra-sdk/src/client-facade.ts
```diff
@@ -133,7 +133,21 @@ export function createClientFacade(
...
+            response = await fetch(request)
+        } catch (networkError) {
+            // A network-level failure (offline, CORS block, abort) rejects here instead of
+            // producing a Response. Without this catch, client.interceptors.error.fns - where
+            // the app wires NProgress's "requestsCompleted" bump - never runs for this request,
+            // so its "requestsTotal" increment is never matched and the loading indicator
+            // never reaches done.
+            let finalError: unknown = networkError
+            for (const fn of client.interceptors.error.fns) {
+                if (fn) finalError = await fn(finalError, undefined as unknown as Response, request, interceptorOptions)
```
MilosPaunovic: The fix itself is right — both currently-registered error interceptors (`kestraHttp.ts` and `hey-api-plugin/runtime.ts`) start with a `if (!response)` branch, so passing no response is handled.

But `undefined as unknown as Response` makes the signature lie to every future interceptor: the type says a `Response` is always there, and the next one written against that type will `response.status` and throw inside the catch block. Since the no-response case is now a real, reachable state, worth widening the interceptor's `response` parameter to `Response | undefined` rather than casting it away here.

ui/packages/kestra-sdk/src/dev-freshness.ts
```diff
@@ -44,16 +44,24 @@ export async function warnIfSdkStale(
...
+        // normal fetches), which would make this staleness check compare against a stale cached
+        // response instead of the live backend.
+        const response = await fetch(url, {credentials: "include", cache: "no-store"})
+        // Unauthenticated (e.g. on the login screen itself, before any session cookie exists), the
+        // spec endpoint redirects to the login page and `fetch` follows it — hashing that HTML would
+        // be a false positive, and a stable one (same page, same build), so it can look like permanent
+        // drift. Don't consume the one-shot check on this: only commit to `alreadyChecked` once we
+        // actually got the spec, so a later call (e.g. after logging in and reloading) can still work.
+        if (!response.ok || response.redirected) return
+        alreadyChecked = true
```
MilosPaunovic: Two side effects of moving the latch below the `await`:

- Concurrent callers can both pass the `alreadyChecked` guard before either reaches this line, so both fetch and both may warn.
- If `fetch` throws (offline, backend not up yet), the latch is never set, so the check re-runs on every subsequent call rather than being genuinely one-shot.

Both are dev-only and harmless in practice — but setting the latch in a `finally`, or right after a non-redirected response is confirmed but before the hashing `await`, would preserve the redirect fix without either.

ui/src/components/layout/empty/links.ts
```diff
@@ -39,6 +39,7 @@ export const links: Record<string, EmptyLinks> = {
         video: "https://www.youtube.com/watch?v=XhICXP_GXic",
         learnMore: "https://kestra.io/docs/enterprise/governance/assets",
     },
+    cases: {},
```
MilosPaunovic: Every other entry here carries at least a `learnMore`. With an empty object the Cases upsell page renders with no "learn more" or video link at all, which is a bit of a dead end for the OSS user this page is aimed at. Is there a docs URL to point at yet, even the general EE governance page?

ui/src/components/demo/Cases.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+    <TopNavBar v-if="topbar" :title="routeInfo.title" />
+    <Empty
+        type="cases"
+        demoCta
+        :title="t(`demos.cases.title`)"
+    >
+        <template #description>
+            {{ $t(`demos.cases.message`) }}
```
MilosPaunovic: Nit: `t()` on the `:title` above and `$t()` here — worth picking one. (The other `demo/*.vue` files do this too, so equally fine to leave it consistent with them.)

MilosPaunovic (commented):
### Frontend review

Two things I'd want resolved before merge (the E2E failure and the skipped frontend jobs), the rest are smaller. Inline comments below.

**CI**

- `E2E - Tests / check` is failing on `executions/execution-bulk-actions.spec.ts` — 2 tests, both the *"when using Select All"* variants, timing out on `locator.click` after 3 retries each. That's the `queryBulkAction` path, which runs through `loadQuery` → `supportedFilterFields` — the one line this PR changes there. See the inline note on `Executions.vue`.
- `Frontend - Tests` and `Frontend - Design System tests` both report **skipping**, so the new `KsSelect` colorMap test added in this PR has not actually run in CI. Worth a re-run / checking the path filter before merging.

**Misc**

- `KsSelect.vue` lost its trailing newline (`\ No newline at end of file` in the diff).
- `unrun@^0.3.1` is added as a devDependency to 5 packages but isn't referenced anywhere in source. I assume it's a tsdown peer that surfaced — worth one line in the PR description so it doesn't read as an unrelated change.

Translations look good — checked all locales, no missing or extra keys.

core/src/main/java/io/kestra/core/models/QueryFilter.java
```diff
@@ -299,7 +299,19 @@ public List<Op> supportedOp() {
         STATUS("status") {
             @Override
             public List<Op> supportedOp() {
-                return List.of(Op.EQUALS, Op.NOT_EQUALS);
+                return List.of(Op.EQUALS, Op.NOT_EQUALS, Op.IN, Op.NOT_IN);
+            }
+        },
```
ammeek, line 304: Field.STATUS.supportedOp() is shared by every resource that declares STATUS. This includes Resource.INVITATION it looks like tests were not added to the AbstractInvitationRepositoryTest to cover this. I would not be surprised if InvitationRepositorys do not support STATUS IN by default.

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -931,6 +935,19 @@ private ExecutorContext handlePausedDelay(ExecutorContext executor, List<WorkerT
             .toList();
 
         if (executor.getExecution().getState().getCurrent() != State.Type.PAUSED) {
+            workerTaskResults
+                .stream()
+                .filter(workerTaskResult -> workerTaskResult.getTaskRun().getState().getCurrent() == State.Type.PAUSED)
+                .forEach(throwConsumer(workerTaskResult ->
+                {
+                    Task task = executor.getFlow().findTaskByTaskId(workerTaskResult.getTaskRun().getTaskId());
```
ammeek: Should this line be moved into the try as a defensive measure?

ammeek (changes requested):
Still reviewing but wanted to allow you to address these comments before the review is done.

ui/packages/kestra-sdk/src/dev-freshness.spec.ts
```diff
@@ -0,0 +1,54 @@
+import {beforeEach, describe, expect, it, vi} from "vitest"
+
+describe("warnIfSdkStale", () => {
+    beforeEach(() => {
```
MilosPaunovic, line 4: **This is what's failing `Frontend - Tests`.** The develop merge brought in the new `no-unrestored-global-stub` lint rule (`ui/lint-rules/`), and this file trips it 3 times — one per `vi.stubGlobal("fetch", ...)` at lines 15, 29 and 46, with no teardown. `Found 42 warnings and 3 errors.`

The rule's reasoning is that the unit suite runs with `isolate: false`, so a stubbed global leaks into every later spec file. One line here clears it:

```ts
beforeEach(() => {
    vi.resetModules()
    vi.restoreAllMocks()
})

afterAll(() => vi.unstubAllGlobals())
```

(`vi.restoreAllMocks()` doesn't undo `stubGlobal` — that needs `unstubAllGlobals` specifically, which is why the rule looks for it by name.)

ui/src/components/executions/Executions.vue
```diff
@@ -601,18 +616,24 @@
...
     )
 
+    const allColumns = computed(() => [
+        ...optionalColumns.value,
+        ...getExtraColumns().map(col => ({...col, label: t(col.label)})),
+    ])
+
     const {visibleColumns: displayColumns, updateVisibleColumns: updateDisplayColumns} = useTableColumns({
-        columns: optionalColumns.value,
+        columns: allColumns.value,
```
MilosPaunovic, line 625: Nice fix on routing the extra columns through the picker. One wrinkle now that the column set is **dynamic**: `useTableColumns` takes a plain `ColumnConfig[]`, not a ref or getter, and computes `defaultOrder` from it once at setup. `getExtraColumns()` on the EE side reads the feature flag and `useAuthStore().user`, so passing `allColumns.value` freezes whatever those happened to be during setup.

Two consequences:

1. **Picker miss on a race** — if features/roles resolve after this component mounts, the Cases column never makes it into `defaultOrder`, so it's absent from the column-picker even though `visibleColumns` would render it.
2. **Users silently lose their saved column order** — `columnOrder`'s localStorage serializer validates with `parsed.length === defaultOrder.length` and falls back to `defaultOrder` otherwise. So any load where the Cases column's presence differs from the load that wrote the value discards the user's custom executions column ordering. That's reachable just by toggling the Cases feature or changing a user's `CASE` permission, not only by the race above.

`ColumnConfig` already has `condition?: () => boolean` for exactly this, and `initializeVisibleColumns` honours it (`c.default && (!c.condition || c.condition())`). Returning the Cases column unconditionally with a `condition` — rather than omitting it from the array — keeps `defaultOrder` stable across loads and fixes both. Happy to be told the race isn't reachable in practice, but the localStorage one looks reachable regardless.

ui/packages/kestra-sdk/src/dev-freshness.ts
MilosPaunovic: Not blocking, and the redirect fix this enables is right — just noting the two side effects of latching below the `await`, in case they're worth a follow-up:

- concurrent callers can both clear the `alreadyChecked` guard at line 46 before either reaches here, so both fetch and both may warn;
- if `fetch` itself throws (offline, backend not up yet), the latch is never set, so this stops being one-shot and re-runs on every later call.

Both are dev-only. Latching in a `finally`, or immediately after the `!response.ok || response.redirected` check but before the hashing `await`, would keep the redirect behaviour without either.

ui/src/components/demo/Cases.vue
MilosPaunovic: Nit, carried over: `t()` on the `:title` above and `$t()` here. The sibling `demo/*.vue` files do the same, so equally fine to leave as-is for consistency.

MilosPaunovic (approved):
## Approving — **frontend scope only**

To be explicit about what this approval covers: I reviewed the **frontend changes only** (`ui/src`, `ui/packages`). I have not reviewed the Java/backend or the SQL migrations, and this approval shouldn't be read as sign-off on those. Anything I found outside the frontend is raised separately, not here.

Thanks for the turnaround on the last round — nearly everything landed:

- extra columns now go through `useTableColumns`, so the EE Cases column participates in the native column-picker and persistence (and extension cells are excluded from sorting) ✅
- `KsSelect`'s colorMap fallback now guards on the entry, so the label and the dropdown row agree ✅
- `client-facade`'s error interceptor signature widened to `Response | undefined` instead of the cast ✅
- `cases` empty-state `learnMore` added, trailing newline restored ✅
- and the frontend test jobs actually run now, which is how the two items below surfaced

### Two things to land before merge

`Frontend - Tests` is currently red, and it's **oxlint, not the tests** — 3 errors, all in this PR's new `dev-freshness.spec.ts`. One line fixes it (inline). The unit and design-system suites are green.

Everything else below is quality, not blocking.

ammeek (approved):
LGTM

PR #17508, opened 2026-07-20
fix(plugins): fall back to plain type resolution instead of excluding plugin types from schema

### LINKED ISSUE #12102: Newly installed plugins don't auto complete for standalone install
### Describe the issue

I am running Kestra executable from https://api.kestra.io/v1/versions/download which comes with no plugins. When I clone plugin repositories and build them with `./gradlew shadowJar` and move the `.jar` file into my `plugins` directory, it shows them in the Plugin page but autocomplete doesn't work. The flow will let me save it and run it successfully, but it shows a warning.

I already had `fs`, `git`, `notifications` and `vmware` installed but have not faced this problem until I added `scripts`, specifically the JAR from `plugin-script` (unclear if this is actually needed - seems to execute flows successfully without it) and `plugin-script-shell`. My `plugins` directory looks like this:

Image

Unclear if this is something with how the scripts repo is setup or if this is down to my setup.

My example specifically:

Image

Image

```yaml
id: ssh_git_sync_openssh
namespace: company.team

tasks:
  - id: test
    taskRunner:
    commands:
      - cd
      - cd .ssh
      - pwd
      - ls

  - id: connect_to_ssh
    type: io.kestra.plugin.fs.ssh.Command
    host: myserver
    authMethod: OPEN_SSH
    openSSHConfigDir: /Users/wrussell/.ssh
    commands:
      - cd ssh-example
      - git pull
      - chmod +x deployment.sh
      - ./deployment.sh

triggers:
  - id: schedule
    cron: "0 7 * * *"
    disabled: true
```

### Environment

- Kestra Version: `1.0.4` running as a standalone installation with H2 db.

fdelbrayelle:
## Summary
- `JsonSchemaGenerator.safelyResolveSubtype()` caught plugin-type resolution failures and silently dropped the type, no logging. A plugin task that fails generic-type resolution (e.g. a manually built jar hitting a classloading conflict) disappears from the JSON schema used for editor autocompletion, while still showing up on the `/plugins` page and running fine (Jackson deserialization doesn't go through this path). Matches #12102: newly installed plugins are listed and executable, but don't autocomplete and show schema-validation warnings, with no error anywhere to diagnose why.
- Broadened the `catch` from `Exception` to `Throwable` — classloader conflicts (`NoClassDefFoundError`/`LinkageError`), plausible when independently built plugin jars with cross-dependencies (e.g. `plugin-script` + `plugin-script-shell`) load in separate isolated classloaders, weren't even caught before.
- **Fallback instead of exclusion**: when `resolveSubtype(declaredType, clz)` throws, fall back to a plain `typeContext.resolve(clz)` — the same call every other plugin category (`TaskRunner`, `Chart`, `Asset`, `LogExporter`, ...) already uses with no subtype-binding walk, so it isn't exposed to this failure mode. The type now still appears in the schema/autocompletion (possibly missing some generic-derived detail) instead of vanishing entirely. Only if the fallback also fails is the type excluded, and now with a `log.warn` naming the class and cause so it's diagnosable.

Fixes #12102

## Test plan
- [x] `./gradlew :core:compileJava`
- [x] `./gradlew :core:test --tests "io.kestra.core.docs.JsonSchemaGeneratorTest"` — 22/22 pass, including:
  - `safelyResolveSubtypeShouldFallBackToPlainResolveWhenSubtypeResolutionFails` — subtype resolution fails, asserts the fallback plain resolve is used and the type is still returned.
  - `safelyResolveSubtypeShouldReturnEmptyWhenBothSubtypeAndPlainResolveFail` — both resolutions fail (mocked), asserts graceful empty result with no exception propagating.

### DIFF
--- a/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
+++ b/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
@@ -987,13 +987,15 @@ public class JsonSchemaGenerator {
     protected static Optional<ResolvedType> safelyResolveSubtype(ResolvedType declaredType, Class<?> clz, TypeContext typeContext) {
         try {
             return Optional.ofNullable(typeContext.resolveSubtype(declaredType, clz));
-        } catch (Throwable e) {
+        } catch (Exception | LinkageError e) {
             // resolveSubtype() walks the full generic hierarchy against declaredType and can throw
             // (e.g. java.lang.TypeNotPresentException on a non-backward-compatible plugin, or a
             // classloading conflict such as NoClassDefFoundError/LinkageError from a plugin jar).
             // Fall back to a plain resolve(), the same call every other plugin category (TaskRunner,
             // Chart, Asset, ...) already relies on, so the type still appears in the schema/autocompletion
             // even if some generic-derived detail is missing, instead of silently disappearing.
+            // Deliberately not catch(Throwable): VirtualMachineError (OutOfMemoryError, StackOverflowError)
+            // must propagate rather than be logged and swallowed here.
             try {
                 ResolvedType fallback = typeContext.resolve(clz);
                 log.warn(
@@ -1004,7 +1006,7 @@ public class JsonSchemaGenerator {
                     e.getMessage()
                 );
                 return Optional.of(fallback);
-            } catch (Throwable fallbackException) {
+            } catch (Exception | LinkageError fallbackException) {
                 log.warn(
                     "Unable to resolve subtype '{}' of '{}' for schema generation, it will be excluded from autocompletion. Cause: [{}] {}",
                     clz.getName(),
--- a/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
+++ b/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
@@ -987,9 +987,18 @@ public class JsonSchemaGenerator {
     protected static Optional<ResolvedType> safelyResolveSubtype(ResolvedType declaredType, Class<?> clz, TypeContext typeContext) {
         try {
             return Optional.ofNullable(typeContext.resolveSubtype(declaredType, clz));
-        } catch (Exception e) {
-            // exception can be thrown when resolving a plugin-type depending on
-            // a non-backward compatible kestra (e.g., java.lang.TypeNotPresentException).
+        } catch (Throwable e) {
+            // Can be thrown when resolving a plugin-type depending on a non-backward compatible kestra
+            // (e.g., java.lang.TypeNotPresentException), or a plugin jar with a classloading conflict
+            // (e.g., NoClassDefFoundError/LinkageError). Log so the excluded type is not silently missing
+            // from the schema used for editor autocompletion.
+            log.warn(
+                "Unable to resolve subtype '{}' of '{}' for schema generation, it will be excluded from autocompletion. Cause: [{}] {}",
+                clz.getName(),
+                declaredType.getErasedType().getName(),
+                e.getClass().getSimpleName(),
+                e.getMessage()
+            );
             return Optional.empty();
         }
     }
--- a/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
+++ b/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
@@ -988,18 +988,32 @@ public class JsonSchemaGenerator {
         try {
             return Optional.ofNullable(typeContext.resolveSubtype(declaredType, clz));
         } catch (Throwable e) {
-            // Can be thrown when resolving a plugin-type depending on a non-backward compatible kestra
-            // (e.g., java.lang.TypeNotPresentException), or a plugin jar with a classloading conflict
-            // (e.g., NoClassDefFoundError/LinkageError). Log so the excluded type is not silently missing
-            // from the schema used for editor autocompletion.
-            log.warn(
-                "Unable to resolve subtype '{}' of '{}' for schema generation, it will be excluded from autocompletion. Cause: [{}] {}",
-                clz.getName(),
-                declaredType.getErasedType().getName(),
-                e.getClass().getSimpleName(),
-                e.getMessage()
-            );
-            return Optional.empty();
+            // resolveSubtype() walks the full generic hierarchy against declaredType and can throw
+            // (e.g. java.lang.TypeNotPresentException on a non-backward-compatible plugin, or a
+            // classloading conflict such as NoClassDefFoundError/LinkageError from a plugin jar).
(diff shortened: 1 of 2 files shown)


fdelbrayelle:
## QA

**Before vs after, code-level:**
- Checked out `JsonSchemaGenerator.java` at the pre-fix commit (`a829fd30f9`) with this PR's new tests applied on top → `safelyResolveSubtypeShouldFallBackToPlainResolveWhenSubtypeResolutionFails` **fails**: old code (`catch (Exception e)`, no fallback, no log) silently drops the type. Matches #12102 exactly — no error anywhere to diagnose it.
- Restored HEAD (fix applied) → both new tests **pass**. Full `:core:test --tests JsonSchemaGeneratorTest`: 22/22 green.

**Does this actually fix #12102?**
Traced which schema path feeds editor autocomplete: `schemas()` (DRAFT_7) — unconditionally goes through `subtypeResolver` → `safelyResolveSubtype`. (`generate()`/DRAFT_2019_09 explicitly excludes `Task`/`AbstractTrigger` from subtype expansion — that's the single-plugin doc-page path, not autocomplete.) So the fix is on the code path that matters.

Two outcomes once `resolveSubtype()` throws:
- **Fallback (`typeContext.resolve(clz)`) succeeds** → type reappears in schema/autocomplete. Both reports in the issue (`plugin-kafka`, `plugin-script`/`plugin-script-shell`) describe plugins that *run fine at runtime* — Jackson already proves the class + hierarchy loads — so the failure was specifically in the generic-binding walk against the declared `Task` type, exactly what the fallback bypasses. Expect this to resolve both reported repros.
- **Fallback also fails** (class genuinely unloadable) → type still excluded, same end-user symptom — but now `log.warn` names the class and cause, closing the "no error anywhere" complaint even if autocomplete can't recover for that harder case.

Wasn't able to get a live end-to-end confirmation with the real jars from the issue (`plugin-kafka`, `plugin-script-shell`) — dev-server boot hit an unrelated port conflict, and a direct `PluginScanner`+`JsonSchemaGenerator` JUnit repro didn't get far enough to register the jar in the time available. So this is code-reading + the PR's own tests, not a live-server confirmation. Worth being explicit in the PR description that the residual "fallback also fails" case is now diagnosable but not auto-recovered.

jymaire:
## Review — fix(plugins): fall back to plain type resolution instead of excluding plugin types from schema

Solid, well-scoped fix with a clear write-up and matching tests. The core idea — recover a type via plain `resolve()` instead of silently dropping it, and log diagnostically when anything goes wrong — is the right call and directly addresses the "installed plugin runs but won't autocomplete, with no error anywhere" symptom in #12102.

### What it does
`safelyResolveSubtype()` (used by `subtypeResolver` for `Task`, `AbstractTrigger` and `Input` subtypes) previously caught `Exception` and returned `Optional.empty()` with no logging, so a subtype that failed generic resolution vanished from the schema. The PR:
- Broadens the catch to `Throwable` (to catch `NoClassDefFoundError`/`LinkageError` from classloader conflicts).
- Falls back to `typeContext.resolve(clz)` so the type still appears (minus some generic detail).
- Excludes the type only if the fallback also fails, and now `log.warn`s the class, declared type, and cause in both branches.

### Strengths
- **Fallback over exclusion** is the correct behavior change — matches what every other plugin category (`TaskRunner`, `Chart`, `Asset`, …) already does.
- **Diagnosability**: both log messages name the class, declared type, exception type and message — exactly what was missing in #12102.
- **Log-spam is bounded**: schema output is cached (`JsonSchemaCache` / `@Cacheable`), so a persistently-failing plugin warns on cache population, not per request.
- **Good test coverage**: both the fallback-success and both-fail paths are exercised.

### Main concern — catching `Throwable` is too wide
`Throwable` also swallows `OutOfMemoryError`, `StackOverflowError`, `VirtualMachineError`, etc. Catching those, logging a warn, and continuing schema generation can mask a genuinely broken JVM state and produce misleading output. Since the actual motivation is `LinkageError`/`NoClassDefFoundError` (both extend `Error`, not `Exception`), consider narrowing to exactly what you intend to recover from:

```java
} catch (Exception | LinkageError e) {
```

This still catches `NoClassDefFoundError` (a subclass of `LinkageError`) and `TypeNotPresentException` (an `Exception`) while letting fatal `VirtualMachineError`s propagate. Same for the inner fallback catch. If there's a concrete non-`LinkageError` `Error` you've seen in the wild, name it explicitly rather than falling back to the whole `Throwable` tree.

### Minor / style
- **Fully-qualified names in tests**: the two new tests inline `com.github.victools...`, `org.mockito.Mockito`, `com.fasterxml.classmate...` everywhere. Adding imports would make them much easier to read and is more consistent with typical test style. (The production file does lean on inline FQNs, so this is a judgment call, but tests read better with imports.)
- **Given-When-Then**: AGENTS.md asks for `// Given / When / Then` comments; worth adding for consistency.
- **First test's implicit assumption**: `safelyResolveSubtypeShouldFallBackToPlainResolveWhenSubtypeResolutionFails` relies on classmate throwing for `resolveSubtype(Task, String)`. It works, but a one-line comment noting *why* classmate throws (unrelated type) would make the intent robust against library behavior changes. (You partly do this already — good.)
- The recovered-but-degraded case and the hard-exclusion case both log at `warn`. That's defensible, but the successful-fallback path arguably fits `info`/`debug` since the outcome is still usable; keeping both at `warn` is fine if you want them visible.

### Correctness / security / perf
- No security impact; no behavioral change to Jackson deserialization or execution paths (schema-generation only).
- Performance impact is negligible and gated by the schema cache.
- Logic is correct: nested try/catch cleanly separates the recover vs. exclude outcomes.

**Verdict:** Approve once the `Throwable` catch is narrowed (or a comment justifies why the full `Throwable` tree must be caught). Everything else is optional polish.

jymaire:
### Concern: catching `Throwable` is too broad

The broadened `catch (Throwable e)` also swallows `OutOfMemoryError`, `StackOverflowError`, and other `VirtualMachineError`s — logging a `warn` and continuing schema generation in those states can mask a genuinely broken JVM and produce misleading output.

The actual targets here (`NoClassDefFoundError`, `LinkageError`, `TypeNotPresentException`) don't require the whole `Throwable` tree. Narrowing to:

```java
} catch (Exception | LinkageError e) {
```

still catches `NoClassDefFoundError` (a `LinkageError` subclass) and `TypeNotPresentException` (an `Exception`), while letting fatal `VirtualMachineError`s propagate. Same for the inner fallback catch.

If there's a concrete non-`LinkageError` `Error` you've observed in the wild, I'd name it explicitly rather than fall back to `Throwable`.

jymaire (approved):

fdelbrayelle:
Good catch, agreed — no concrete non-`LinkageError` `Error` observed here, so no reason to swallow `VirtualMachineError`. Narrowed both catches to `Exception | LinkageError` in 00ee156dbc.

PR #17512, opened 2026-07-20
refactor(executions): consolidate JSON tree components

### LINKED ISSUE #16766: Consolidate the two JSON tree components (KsJsonTree vs VariableTreeView)
## Context

Two recursive JSON-tree components now coexist in the UI after [PR #16580](https://github.com/kestra-io/kestra/pull/16580) (enriched execution logs) and [PR #16648](https://github.com/kestra-io/kestra/pull/16648) (Outputs UI 2.0) both landed:

| | `KsJsonTree` | `VariableTreeView` |
|---|---|---|
| Location | `ui/packages/design-system/src/components/Data/KsJsonTree.vue` | `ui/src/components/executions/outputs/VariableTreeView.vue` |
| Used by | Log lines (inline collapsible JSON) | `ExecutionVariableExplorer` (outputs explorer) |
| Rendering | recursive nested DOM | flattened rows + line-number gutter |
| API | `value / nodeKey / depth / defaultExpanded` | `value / basePath / selectedPath` + `@select(path)` |
| Width handling | ResizeObserver, adaptive truncation | none |

During the #16580 review, @elevatebart noted the JSON-tree felt like it should already exist for Execution Outputs and should live in the design system. At review time Outputs used a Monaco editor (no tree); #16648 then shipped `VariableTreeView` independently — so the duplication the review anticipated is now real.

## Goal

Converge on a single design-system tree component used by both surfaces, without regressing either UX:

- Log lines: compact inline preview, width-aware truncation, no path selection.
- Outputs explorer: line-number gutter, path-based selection (`@select`), collapsed-preview counts (i18n).

Likely shape: extend the DS `KsJsonTree` with optional `gutter` / `selectable` (path emit) / preview modes, migrate `ExecutionVariableExplorer` onto it, then delete `VariableTreeView`. Keep it minimal and reversible; the Outputs explorer is an always-visible execution surface, so verify both surfaces render correctly (light + dark) before removing the old component.

## Notes
- This is intentionally out of scope for #16580 (already approved) to avoid a risky refactor on a core surface inside a feature PR.

hemantmm:
closes: #16766 

---

### ✨ Description

What does this PR change?  

- Extends `KsJsonTree` with an optional row-based mode for execution outputs.
- Added gutter line numbers, selectable paths, and collapsed preview formatter support.
- Migrated `ExecutionVariableExplorer` from `VariableTreeView` to `KsJsonTree`.
- Deleted the duplicate `VariableTreeView` component.
- Added Storybook and unit coverage for the shared JSON tree behavior.

https://github.com/user-attachments/assets/3233da31-40cb-4492-84ee-7c60397c3f06

https://github.com/user-attachments/assets/21f510cc-bbb4-45f4-a640-1c635fef0926


elevatebart (approved):
Great work keeping the PR to a small footprint 💪

I usually do not like to pass a functions as a Prop in VueJs: Their execution context is confusing. You are trying to keep the changes to the codebase minimal so I understand why. Thank you.
 
I would almost prefer that the "count" of sub-element be either a boolean prop or a permanent feature of `KsJsonTree.vue`.

Tell me what you decide, we can merge as is

hemantmm:
Thank you. I would prefer a Boolean prop because the count is mainly needed in rows mode for the Outputs explorer. This way, a prop keeps it optional without affecting the compact inline log preview.

flcarre (approved):

elevatebart:
After digging a little deeper in your code, it really feels like the 2 modes are seriously disjoined and have to be 2 separate components. What I had in mind is a more integrated UX, where both components feature could be used in both situations.

elevatebart (changes requested):
I'd want to check if we cannot use the KsJsonTree as is before adding the whole new mode to it.

elevatebart:
Now that I look at the ksJsonTree, I see a few UX issues:
- expansion of an object is done on the left even for arrays
- no information about the contents of the examded object before is opened
- the curly brackets before and after the content bring noise and very little semantic value
- the objects on the same level are misagned. Expandable rows carets push teh content to the right even if they are on the same level.

I wonder if we should simply replace it entirely with the rows mode.

@flcarre what do you think?

hemantmm:
Thanks, that makes sense. I agree that the two modes are too disjoint.

I can see how the current implementation leads to the UX issues you pointed out. I'll wait for @flcarre's input before reworking on it so that I can align with the direction you both prefer.

MilosPaunovic:
@hemantmm Can you resolve the conflicts here, please?

elevatebart (approved):
NOw it's fixed

PR #17513, opened 2026-07-20
test(logs): make the LogDataStore contract load-once and backend-agnostic

nkwiatkowski:
## What

Reworks `AbstractLogDataStoreTest` — the contract every `LogDataStoreInterface` backend must satisfy (JDBC, Elasticsearch, and external stores such as GCP Cloud Logging) — so it can run against **eventually-consistent, quota-limited** stores as well as read-your-write ones, without a write→read (and a wait) per test.

## Why

The previous contract issued a read per test. Against an eventually-consistent, quota-limited backend (GCP Cloud Logging) that meant either a wait per test or spurious `RESOURCE_EXHAUSTED`. It also seeded fixed 2020/2021 timestamps, which retention-limited backends silently drop — making those fixtures un-queryable there.

## Changes

- **Load-once**: `@BeforeAll seed()` writes each fixture group under its own random tenant via `saveBatch`, then a single `awaitIndexing(BooleanSupplier)` probe. Default is a no-op (read-your-write backends never wait); eventually-consistent backends override it to poll. Read tests assert against the seeded data with no further writes.
- **Exhaustive `find()` matrix**: every field × operation, plus composed AND/OR/nested logical filters (replaces the old existence-only combination matrix).
- **Full round-trip contract**: `save_preservesEveryField` asserts every `LogEntry` field reads back identical, incl. millisecond timestamp fidelity.
- **Relative timestamps**: fixtures anchor to *now*, never fixed historical dates, so retention-limited backends can ingest and query them.
- **Whole interface with real values**: `saveBatch`, all `findBy*` variants (incl. `minLevel` composed with `attempt`), offset + cursor pagination, `fetchData`/`fetchValue` `canAggregate` branch, purge/delete `canPurge` branch, unsupported-filter validation.

## Verification

- **H2** (read-your-write): 112 / 112 green.
- **GCP Cloud Logging** (real backend, via the external log-data-store plugin): 112 / 112 green, **zero quota errors**.

This needs to land first so the `tests` snapshot picks up the new contract; the GCP Cloud Logging log-data-store plugin's CI depends on it.

### DIFF
--- a/tests/src/main/java/io/kestra/core/repositories/AbstractLogDataStoreTest.java
+++ b/tests/src/main/java/io/kestra/core/repositories/AbstractLogDataStoreTest.java
@@ -1,14 +1,18 @@
 package io.kestra.core.repositories;
 
-import java.io.IOException;
 import java.time.Instant;
+import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.function.BooleanSupplier;
 import java.util.stream.Stream;
 
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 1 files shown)


loicmathieu (approved):
LGTM

PR #17515, opened 2026-07-20
fix(core): replace stack-trace matching with a ScopedValue marker for internal I/O

loicmathieu:
## Summary
- `AbstractChecker.isRunContext()` (in kestra-ee's worker sandbox) used to grant a file-access escape hatch by scanning the current thread's stack trace for three trusted class names (`DefaultRunContext`, `InternalStorage`, `InternalNamespace`). That stays `true` for any I/O performed while one of those frames merely lingers on the call stack, not only for the framework's own I/O.
- Adds `io.kestra.core.runners.PriviledgedIo`, a `ScopedValue<Boolean>` guardrail marker bound narrowly around the exact JDK-level I/O calls the framework performs on the plugin's behalf in `InternalStorage`, `InternalNamespace`, and `DefaultRunContext.logFileURI()`.
- This is a defense-in-depth guardrail against a fragile heuristic, not a new adversarial security boundary — see the companion kestra-ee PR for how the sandbox consumes it, and kestra-io/kestra-ee#9014 for full context.

## Companion PR
kestra-io/kestra-ee (branch `fix/workload-isolation-hardening`) updates `AbstractChecker.isRunContext()` to check `PriviledgedIo.isActive()` instead of the stack-trace scan, and hardens the sandbox's `Checkers` dispatch bridge.

## Test plan
- [x] New `PriviledgedIoTest` covering the scoped-value lifecycle (default inactive, active inside `call`/`run`, inactive after, checked-exception propagation)
- [x] `./gradlew :core:test --tests "io.kestra.core.storages.*" --tests "io.kestra.core.runners.DefaultRunContext*" --tests "io.kestra.core.runners.PriviledgedIoTest"` — all green


loicmathieu:
Closing — after further review, this fix doesn't need any OSS core changes. `DefaultRunContext`, `InternalStorage`, and `InternalNamespace` are already public classes, so the EE-side check can reference their `Class` objects directly for identity comparison via `StackWalker`, with no wrapping needed on this side. See kestra-io/kestra-ee#9390 for the final single-repo fix.

PR #17518, opened 2026-07-20
test(logs): make the LogDataStore aggregation fixture backend-agnostic

nkwiatkowski:
## What

The `AbstractLogDataStoreTest` `DISTINCT` read-fixture used three identical `INFO` levels, so the `fetchData` / `fetchValue` `COUNT`-over-`LEVEL` assertions depended on a semantic that differs by backend:

- **JDBC** counts rows → `COUNT(level)` = 3
- **Elasticsearch** counts distinct values via `cardinality` → `COUNT(level)` = 1

This makes the fixture use **three distinct levels** (`INFO`/`WARN`/`ERROR`) so `COUNT(level)` == 3 under both semantics — keeping the contract backend-agnostic. A comment documents why they must stay distinct. No `DISTINCT` filter case filters on level, so nothing else is affected.

## Why

Follow-up to #17513 (load-once LogDataStore contract). Running that contract against the Elasticsearch log data store surfaced this over-assertion. The visibility half of the ES failures is fixed in the companion EE PR (refreshes the ES index in `awaitIndexing`); this PR fixes the aggregation half in the shared contract.

## Verification

- `:jdbc-h2:test --tests "*H2LogDataStoreTest*"` → 112/112 green
- `:repository-elasticsearch-ee:test --tests "*ElasticSearchLogDataStoreTest*"` (EE, against this branch + companion) → 112/112 green

Companion EE PR: refresh ES index in `ElasticSearchLogDataStoreTest.awaitIndexing` (same branch name).

### DIFF
--- a/tests/src/main/java/io/kestra/core/repositories/AbstractLogDataStoreTest.java
+++ b/tests/src/main/java/io/kestra/core/repositories/AbstractLogDataStoreTest.java
@@ -141,10 +141,15 @@ public abstract class AbstractLogDataStoreTest {
         log(Level.INFO, "exec-sink").taskId("sink").taskRunId("tr-sink").attemptNumber(2).build()
     );
 
+    // The three levels are DISTINCT on purpose: fetchData/fetchValue aggregate COUNT over the LEVEL field, and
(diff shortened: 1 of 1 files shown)


loicmathieu (approved):
LGTM

PR #17528, opened 2026-07-21
fix(core): count non-terminated task runs against Parallel/Dag concur…

### LINKED ISSUE #17506: Dag.concurrent / Parallel.concurrent does not strictly enforce the parallelism cap
## Summary

`Dag.concurrent` is documented as *"Number of concurrent parallel tasks that can be running at any point in time"*, and `Parallel.concurrent` similarly. In practice the cap can be exceeded on the initial dispatch burst, because the slot accounting in `FlowableUtils.resolveParallelNexts` only counts taskRuns in state `RUNNING`, ignoring the `CREATED` window between dispatch and worker pickup.

The same defect does **not** affect `ForEach.concurrencyLimit`, because `ForEach` goes through a sibling method (`resolveConcurrentNexts`) that counts every non-terminal taskRun and uses `>=`. So the fix is essentially to bring `resolveParallelNexts` in line with `resolveConcurrentNexts`.

Verified against `v1.3.26`; the code path is unchanged on `master` at the time of writing.

## Reproduction

A minimal DAG with 5 tasks — three of them (`dag_a`, `dag_b`, `dag_d`) are roots with `dependsOn: []`:

```yaml
id: dag_concurrent_bug
namespace: company.team
tasks:
  - id: first
    type: io.kestra.plugin.core.flow.Dag
    concurrent: 2
    tasks:
      - task: { id: dag_a, type: io.kestra.plugin.core.log.Log, message: dag_a }
        dependsOn: []
      - task: { id: dag_b, type: io.kestra.plugin.core.log.Log, message: dag_b }
        dependsOn: []
      - task: { id: dag_c, type: io.kestra.plugin.core.log.Log, message: dag_c }
        dependsOn: [dag_a]
      - task: { id: dag_d, type: io.kestra.plugin.core.log.Log, message: dag_d }
        dependsOn: []
      - task: { id: dag_e, type: io.kestra.plugin.core.log.Log, message: dag_e }
        dependsOn: [dag_b, dag_d]
```

With `concurrent: 2`, the expected initial burst is **2** root taskRuns. In practice, when workers pick up tasks slightly slower than the executor emits dispatch messages, **all 3 roots (`dag_a`, `dag_b`, `dag_d`) are created together**.

### Observed taskRun timeline (real execution, redacted)

Execution id `6bcB8tWwcFCrV72hmNmg0e`, `concurrent=2`:

| taskId | state.CREATED           | state.RUNNING           | Δ from first CREATED |
|--------|-------------------------|-------------------------|----------------------|
| `dag_a`| `08:23:07.420305Z`      | `08:23:07.609602Z`      | +0 ms                |
| `dag_b`| `08:23:07.420344Z`      | `08:23:07.594622Z`      | +0.04 ms             |
| `dag_d`| `08:23:07.481023Z`      | `08:23:07.630443Z`      | **+60.7 ms** ⚠️      |
| `dag_c`| `08:23:58.915724Z`      | `08:23:59.856662Z`      | (later, depends on a)|
| `dag_e`| `08:23:58.915749Z`      | `08:23:59.853200Z`      | (later, depends on b+d)|

`dag_d` was created at `.481`, **129 ms before `dag_a` reached `RUNNING`** (`.609`) and **113 ms before `dag_b` reached `RUNNING`** (`.594`). All three root pods launched concurrently — the `concurrent=2` cap was overshot by 50%.

The exact reproducibility depends on how fast worker consumption catches up with executor dispatch. Slower workers, larger root fan-outs, or bursty message queues (RocketMQ, Kafka, JDBC polling under load) all make it more likely. A subsequent trigger of the same flow may respect the cap correctly — the defect is timing-dependent, not deterministic.

## Root cause

`FlowableUtils.resolveParallelNexts` (used by both `Dag` and `Parallel`), around line 448 in v1.3.26:

```java
long runningCount = taskRuns
    .stream()
    .filter(taskRun -> taskRun.getState().isRunning())
    .count();
// ...
if (concurrency > 0 && runningCount > concurrency) {
    return Collections.emptyList();
}
// ...
if (concurrency > 0) {
    nextTaskRunStream = nextTaskRunStream.limit(concurrency - runningCount);
}
```

A taskRun goes through `CREATED → RUNNING → terminal`. The gap between dispatch and worker pickup can be tens to hundreds of milliseconds. During that window `isRunning()` returns `false`, so `runningCount` under-reports the number of in-flight taskRuns. When several candidates become dispatchable at the same tick (many `dependsOn: []` roots in a Dag, many children in a `Parallel`), the executor emits multiple dispatch batches before the first batch flips to `RUNNING`, and each batch believes the slot budget is still full.

The `if (!notFinds.isEmpty() && lastCreated.isEmpty())` guard slightly further down catches the *same-tick, same-executor* case, but does not help when dispatch messages are consumed by a different executor instance or by the same executor before the earlier CREATED taskRun has been merged back into the execution snapshot. In a distributed executor deployment this is exactly the situation that produces the overshoot above.

Compare with `resolveConcurrentNexts` in the same file (used by `ForEach.concurrencyLimit`), around line 329:

```java
long nonTerminatedCount = taskRuns
    .stream()
    .filter(taskRun -> !taskRun.getState().isTerminated())
    .count();

if (concurrency > 0 && nonTerminatedCount >= concurrency) {
    return Collections.emptyList();
}
```

This one counts every non-terminal state (including `CREATED`) and uses `>=`, so it enforces the cap strictly. `ForEach.concurrencyLimit` therefore does not exhibit this defect.

## Impact

Users configure `Dag.concurrent` / `Parallel.concurrent` to protect a downstream system: rate-limited API, DB connection pool, Kubernetes namespace quota, etc. When a Dag has more ready roots than `concurrent` allows, or a `Parallel` has more children than `concurrent`, the initial burst can exceed the cap by up to *N − concurrent*, where *N* is the number of ready candidates.

- **`Dag`**: Any DAG whose initial layer has more `dependsOn: []` tasks than `concurrent`.
- **`Parallel`**: Any `Parallel` block with `concurrent` set (non-zero) and more children than `concurrent`.
- **`ForEach`**: Not affected — already uses the correct filter.

## Suggested fix

Bring `resolveParallelNexts` in line with its sibling `resolveConcurrentNexts`:

```diff
- long runningCount = taskRuns
-     .stream()
-     .filter(taskRun -> taskRun.getState().isRunning())
-     .count();
+ long nonTerminatedCount = taskRuns
+     .stream()
+     .filter(taskRun -> !taskRun.getState().isTerminated())
+     .count();

- if (concurrency > 0 && runningCount > concurrency) {
+ if (concurrency > 0 && nonTerminatedCount >= concurrency) {
      return Collections.emptyList();
  }
  // ...
  if (concurrency > 0) {
-     nextTaskRunStream = nextTaskRunStream.limit(concurrency - runningCount);
+     nextTaskRunStream = nextTaskRunStream.limit(concurrency - nonTerminatedCount);
  }
```

A regression test can be added under `DagTest` / `ParallelTest`:

1. Use a Dag whose root layer has `N > concurrent` tasks whose worker executor intentionally pauses on the `RUNNING` transition (`Sleep` for a few seconds).
2. Poll `taskRunList` immediately after execution start.
3. Assert that the count of non-terminal taskRuns never exceeds `concurrent`.

## Environment

- Kestra version: verified on `v1.3.26`, code path unchanged on `master`.
- File: `core/src/main/java/io/kestra/core/runners/FlowableUtils.java`.
- Executor: RocketMQ (multi-instance) — but the defect is code-level and reproduces on any executor whose taskRun snapshots are not fully synchronous with dispatch emission (JDBC/Kafka executors have the same characteristic under load).

## Contribution

If the maintainers agree this is a bug, I'm happy to contribute a PR with the fix and the regression tests described above. Just let me know and I'll pick it up.

flow: 

[flow.yml](https://github.com/user-attachments/files/30175098/flow.yml)

executions:

[executions.json](https://github.com/user-attachments/files/30175103/executions.json)

### Environment

- Kestra Version: develop

aulightzxy-web:
### ✨ Description

`Parallel` and `Dag` enforce their `concurrent` limit through
`FlowableUtils.resolveParallelNexts(...)`. The number of busy slots is computed as:

```java
long runningCount = taskRuns.stream()
    .filter(taskRun -> taskRun.getState().isRunning())   // RUNNING || KILLING only
    .count();
```

`State.isRunning()` is `true` only for `RUNNING` and `KILLING`. It therefore does
**not** count child task runs that are in-flight but momentarily in another
non-terminal state — most importantly **`RETRYING`** and **`PAUSED`** (also
`SUBMITTED` / `QUEUED`).

There is a second guard just below — `findLastCreated(...)` — that blocks any further
dispatch while at least one child is still `CREATED`:

```java
Optional<TaskRun> lastCreated = execution.findLastCreated(taskRuns);
if (!notFinds.isEmpty() && lastCreated.isEmpty()) { /* dispatch */ }
```

This `CREATED` guard is what keeps concurrency in check for the common case: after
the first batch is created, no new children are dispatched until those leave the
`CREATED` state. So `CREATED` on its own does **not** breach the cap.

The states that breach the cap are the ones that pass **both** checks:
`RETRYING` and `PAUSED` are **neither `isRunning()` nor `CREATED`**, so they are not
counted as a busy slot **and** they do not trip the `lastCreated` guard →
**over-subscription**.

#### Reproduction (`concurrent: 2`, `RETRYING`)

- child `a` is `RETRYING`, child `b` is `SUCCESS`, children `c`/`d` not yet started;
- `runningCount = 0` (neither `a` nor `b` is `isRunning()`);
- `lastCreated` is empty (no `CREATED` child) → guard passes;
- dispatch does `limit(concurrency - runningCount) = limit(2 - 0) = 2` → **both `c`
  and `d` start**;
- now `a (RETRYING) + c + d = 3` task runs are in flight — **one over the limit of 2**.

`PAUSED` reproduces identically with a `Pause` child instead of a retried one.

#### The fix

Count every non-terminated task run as consuming a slot — the same semantics the
`ForEach` / concurrent-values path (`resolveConcurrentNexts`) already uses:

```java
long runningCount = taskRuns.stream()
    .filter(taskRun -> !taskRun.getState().isTerminated())
    .count();
```

With the fix, the `RETRYING`/`PAUSED` child holds a slot, so only **one** more child
starts, keeping the in-flight count at exactly `concurrent`.

#### Notes on scope

- `CREATED` is now also included in the count, which makes the existing
  `findLastCreated` guard partly redundant. It is intentionally **left in place** to
  keep this change minimal and behaviour-preserving for the batch-dispatch path.
- The `limit(concurrency - runningCount)` clamp is the real throttle; the
  `runningCount > concurrency` early-return is only a short-circuit, so the change is
  self-consistent (covered by the added regression tests).
- Default behaviour is untouched: `Parallel`/`Dag` default to `concurrent: 0`
  (unlimited); the new count only affects dispatch when `concurrent > 0`.
- Blast radius is exactly two call sites — `Dag.resolveNexts` (`resolveDagNexts`) and
  `Parallel.resolveNexts` (`resolveParallelNexts`); no other callers.

> Correction vs. the linked issue: the issue framed the leak around `CREATED` task
> runs. On `develop` the `findLastCreated` guard already prevents a `CREATED`-only
> breach; the states that actually slip past both checks are `RETRYING` and `PAUSED`.
> This PR fixes that, and the `!isTerminated()` count also subsumes `CREATED` for
> good measure. (The `CREATED`-visible symptom the reporter saw came from an
> environment where the `findLastCreated` guard was not present.)

### 🔗 Related Issue

Closes #17506

### 🛠️ Backend Checklist

Verified locally on JDK 25 (Temurin 25.0.3): `./gradlew :core:test --tests
io.kestra.core.runners.FlowableUtilsTest` → **25 tests, 0 failures**. Reverting the
one-line fix back to `isRunning()` makes the two reproduction tests fail
(`Expected size: 1 but was: 2`), confirming they capture the bug.

New unit tests in `core/.../runners/FlowableUtilsTest.java`:

| Test | State under test | Asserts |
|------|------------------|---------|
| `resolveParallelNexts_retryingChildShouldConsumeConcurrencySlot` | one `RETRYING` child, `concurrent=2` | only **1** new child dispatched (fails on `isRunning()`, passes on fix) |
| `resolveParallelNexts_pausedChildShouldConsumeConcurrencySlot` | one `PAUSED` child, `concurrent=2` | only **1** new child dispatched |
| `resolveParallelNexts_runningChildrenAtLimitShouldDispatchNothing` | two `RUNNING` children, `concurrent=2` | **0** dispatched (no regression) |
| `resolveParallelNexts_freshStartShouldDispatchUpToConcurrency` | nothing dispatched, `concurrent=2` | exactly **2** dispatched (no under-dispatch) |

### 📝 Additional Notes

The `RETRYING` slot leak is the easiest way to observe this in a real flow: give a
child a `retry` policy, let its first attempt fail, and while it sits in `RETRYING`
the parent starts more siblings than `concurrent` permits. `PAUSED` reproduces the
same way with a `Pause` child. Happy to also tighten the `> concurrency`
early-return to `>=` to fully mirror the `ForEach` path if reviewers prefer.

### 🤖 AI Authors

🐱 Why did the `RETRYING` task run get past the concurrency bouncer? Because it
wasn't on the `isRunning()` guest list *and* it wasn't wearing a `CREATED` badge —
so it just strolled in like it owned the place. Turns out the bouncer should have
been checking for anyone who isn't `terminated()` yet.


aulightzxy-web:
@loicmathieu, the failing check is a 403 on the auto-comment step, not an actual test failure, all tests passed.

loicmathieu (approved):
LGTM, thanks

PR #17537, opened 2026-07-21
fix(tests): support directory paths in the @LoadFlows metastore wait

nkwiatkowski:
## What
The metastore-wait added to `AbstractLoaderExtension.loadFlows` parsed each annotation path directly as a single `Flow`, which fails for **directory** fixtures (e.g. `@LoadTestResources(flows = {"flows/valids"})`): reading a directory as YAML yields its file listing, not a flow. This broke EE controller tests (`FlowControllerAllTest`, `ExecutionControllerAllTest`, `ExecutionControllerPolicyGateTest`) across all backends with `ParameterResolutionException: ... Illegal flow source`.

## How
- `LocalFlowRepositoryLoader.load(...)` now returns the persisted `List<FlowWithSource>` it created/updated (all overloads; source-compatible — no caller used the return value). `TestsUtils.loads(...)` forwards it.
- `loadFlows` drives the metastore-wait from those persisted flows, so a directory expands to the flows it actually contains — directories and single files behave identically. `deleteFlows` is likewise folder-aware.
- The wait matches the **metastore cache** (the source of truth for flow liveness) rather than re-reading the repository, whose tenant-scoped lookups are eventually consistent on Elasticsearch.
- The wait **settles** rather than requiring completeness: a few valid flows never reach the cache (their queue event fails to deserialize, e.g. task aliases like `alias-task`) yet are still persisted, so it stops once the cache stops catching up (3s quiet period, 30s cap) and warns about stragglers instead of failing the fixture. awaitility is no longer used here (removed from the ArchUnit exclusion).

## Test
- New guard test `FlowLoaderDirectoryTest` loads a directory via `@LoadFlows`.
- Verified green: OSS `FailTest`, `TriggerTest` (`@LoadFlows` + Flow trigger), ArchUnit; and downstream EE `FlowControllerAllTest`/`ExecutionControllerAllTest` (H2/Postgres/Mysql/Elastic).

Companion EE PR: kestra-io/kestra-ee (ES loop TASK_ID mapping) — linked below.

### DIFF
--- a/tests/src/main/java/io/kestra/core/junit/extensions/AbstractLoaderExtension.java
+++ b/tests/src/main/java/io/kestra/core/junit/extensions/AbstractLoaderExtension.java
@@ -9,11 +9,13 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
 
+import org.awaitility.core.ConditionTimeoutException;
 import org.junit.jupiter.api.extension.ExtensionContext;
 
 import io.kestra.core.junit.services.TestTenantLifecycle;
@@ -30,12 +32,11 @@ import io.kestra.core.utils.TestsUtils;
 import io.micronaut.context.ApplicationContext;
 import io.micronaut.data.model.Pageable;
 import io.micronaut.test.extensions.junit5.MicronautJunit5Extension;
-import lombok.extern.slf4j.Slf4j;
 
 import static io.kestra.core.junit.extensions.ExtensionUtils.loadFile;
 import static io.kestra.core.utils.Rethrow.throwFunction;
+import static org.awaitility.Awaitility.await;
 
-@Slf4j
 public abstract class AbstractLoaderExtension {
 
     protected ApplicationContext context;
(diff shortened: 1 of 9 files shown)


tests/src/main/java/io/kestra/core/junit/extensions/AbstractLoaderExtension.java
```diff
@@ -81,48 +90,96 @@ protected void loadFlows(ExtensionContext extensionContext, String tenantId, Str
...
-                    )
+        List<FlowWithSource> pending = new ArrayList<>(loadedFlows);
+        waitForFlowsInMetaStore(tenantId, flowMetaStore, pending);
+    }
+
+    private static final Duration METASTORE_WAIT_BUDGET = Duration.ofSeconds(30);
+    private static final Duration METASTORE_WAIT_QUIET_PERIOD = Duration.ofSeconds(3);
+    private static final Duration METASTORE_WAIT_POLL_INTERVAL = Duration.ofMillis(100);
+
+    private static void waitForFlowsInMetaStore(String tenantId, FlowMetaStoreInterface flowMetaStore, List<FlowWithSource> pending) {
```
loicmathieu: This is a bit complex, we should just await that the metastore allLastVersion() contains all flows.

loicmathieu (changes requested):

loicmathieu (approved):
LGTM

PR #17539, opened 2026-07-21
fix(webserver): stop leaking sensitive instance metadata via /api/v1/configs

loicmathieu:
## Summary

`/api/v1/configs` was unconditionally whitelisted in `AuthenticationFilter` and returned sensitive instance metadata to any unauthenticated caller: exact Kestra version, commit id/date, queue type, system namespace, hidden-label prefixes, plugin hash, and instance URL. `isBasicAuthInitialized: false` also let attackers automate discovery of unsecured instances (kestra-io/kestra-ee#7987).

- Adds a new minimal public endpoint `GET /api/v1/configs/login` returning only `{isBasicAuthInitialized}` — the single field the login/setup UI needs before authentication.
- `/api/v1/configs` now requires authentication like any other endpoint; only `/api/v1/configs/login` is whitelisted in `AuthenticationFilter`, with the same exact-match bypass protection (GHSA-5vc5-wxxq-3fjx class) as the existing `/configs`/`basicAuth` whitelist entries.
- Updates the frontend's pre-authentication flows (router guard in `main.ts`, the login page, the first-run setup wizard) to call the new minimal endpoint instead of the full one. The full config is now only loaded once the user is authenticated (post-login, post-account-creation).
- `TenantAliasingRooter`'s exclusion list is updated to include the new route for consistency with the existing `/api/v1/configs` entry.

Fixes https://github.com/kestra-io/kestra-ee/issues/7987

## Test plan

- [x] `./gradlew :webserver:test --tests "io.kestra.webserver.filter.AuthenticationFilterTest" --tests "io.kestra.webserver.controllers.api.MiscControllerTest"` — 25/25 passing, including new tests asserting `/api/v1/configs` now requires auth and `/api/v1/configs/login` is public and exact-match-checked.
- [x] `cd ui && npx vitest run tests/unit` — 731/731 passing, including new tests for `loadLoginConfig()` and `addBasicAuth()`'s config-loading order.
- [x] `cd ui && vue-tsc --build tsconfig.app.json` — clean.

### DIFF
--- a/ui/src/components/basicauth/BasicAuthLogin.vue
+++ b/ui/src/components/basicauth/BasicAuthLogin.vue
@@ -120,7 +120,7 @@
 
 
     const checkServerInitialization = async () => {
-        const response = await axios.get(`${apiUrlWithoutTenants()}/configs`, {timeout: 10000})
+        const response = await axios.get(`${apiUrlWithoutTenants()}/configs/login`, {timeout: 10000})
         return response.data?.isBasicAuthInitialized
     }
 
--- a/ui/src/components/basicauth/BasicAuthSetup.vue
+++ b/ui/src/components/basicauth/BasicAuthSetup.vue
@@ -160,7 +160,7 @@
     import {useMiscStore} from "override/stores/misc"
     import {useSurveySkip} from "../../composables/useSurveyData"
     import {trackSetupEvent} from "../../composables/usePosthog"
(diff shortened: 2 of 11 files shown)


nkwiatkowski (approved):

elevatebart (approved):
Cool !! you have added a new configs/login endpoint. That makes the process a lot cleaner.

PR #17540, opened 2026-07-21
refactor(core): restrict CrudEventType to CRUD values and use the CrudEvent factories

fhussonnois:
Companion of the kestra-ee AuditableAction refactor:

- remove the authentication values (LOGIN, LOGOUT, IMPERSONATE, LOGIN_FAILURE, ACCOUNT_LOCKED) from CrudEventType — authentication events are EE audit facts and are now published through the EE AuditEvent, not the CrudEvent publisher; nothing references these values anymore
- replace manual new CrudEvent<>(after, before, type) construction with the static of()/create()/delete() factories that derive the event type from the before/after models


Skraye (approved):
LGTM

PR #17543, opened 2026-07-21
fix(design-system): limit KsDurationPicker to fixed-length units

loicmathieu:
## Summary
- Fixes kestra-io/kestra-ee#9369: saving a namespace/tenant Quota fails with `Text cannot be parsed to a Duration` because `KsDurationPicker` lets users compose calendar-based ISO-8601 durations (e.g. `P1Y`, `P2M`, `P3W`), which `java.time.Duration.parse()` cannot handle (`Duration` only supports the fixed-length `PnDTnHnMnS` subset).
- Removes the Years/Months/Weeks fields from `KsDurationPicker` and tightens its custom-input parser/regex to the same fixed-length subset, so the component can no longer emit a value the backend rejects. Typing `P1Y` in the custom box now shows the existing "Invalid ISO 8601 duration" message instead of silently producing an unparseable value.

image

## Test plan
- [x] `npm run unit:test` in `ui/packages/design-system` — updated/added tests pass (field count 7→4, calendar-duration seeds replaced, new regression test asserting `P1Y` is rejected)
- [x] `npm run types:test` — clean
- [x] `npm run lint:test` — clean (pre-existing unrelated warnings only)
- [ ] Manual: Namespace Edit → Quotas → add a quota with days/hours → Save succeeds; Tenant Edit same

### DIFF
--- a/ui/packages/design-system/src/components/Form/KsDurationPicker.vue
+++ b/ui/packages/design-system/src/components/Form/KsDurationPicker.vue
@@ -2,39 +2,6 @@
     <div class="ks-duration-picker" v-bind="$attrs">
         <div class="ks-duration-picker__fields">
             <div class="ks-duration-picker__field">
-                <label for="ks-duration-years">{{ $t('years') }}</label>
-                <KsInputNumber
-                    size="small"
-                    controlsPosition="right"
-                    id="ks-duration-years"
-                    v-model="years"
-                    :disabled="disabled"
-                    :min="0"
-                />
-            </div>
-            <div class="ks-duration-picker__field">
-                <label for="ks-duration-months">{{ $t('months') }}</label>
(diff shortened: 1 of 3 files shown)


loicmathieu:
@florian-create I don't know if you use the same component in the no-code to edit durations or not, but you might check for that issue.

elevatebart:
> @florian-create I don't know if you use the same component in the no-code to edit durations or not, but you might check for that issue.

I believe you meant @flcarre

elevatebart (approved):
One caveat: The ISO norm includes Year and month so erroring or describing this as respecting the ISO norm can be misleading. 
If Year or month are mentioned, the duration will fail to parse.

elevatebart:
Also,... fix the storybook test please?

PR #17567, opened 2026-07-22
fix(jdbc): detect deadlocks nested deeper than one cause level

loicmathieu:
DeadlockPredicate only inspected e.getCause(), one level deep. Under Postgres, once a statement deadlocks (SQLState 40P01) the transaction is poisoned, so a later statement in the same failed retry attempt surfaces a secondary "current transaction is aborted" exception (25P02) wrapping the original deadlock one level deeper. The single-level check missed this shape, so the retry never fired.

Walk the full cause chain instead, guarding against cyclic chains with an identity-based seen-set to avoid circular cause chain.


nkwiatkowski (approved):

loicmathieu:
Backported to 1.0 and 1.3

PR #17573, opened 2026-07-22
feat(core): normalize PostHog event names to surface.object.action

flcarre:
## Summary

- Frontend PostHog events funnel through a single choke point (`posthogEvents()` in `ui/src/stores/api.ts`), so the entire `app.*` rename is implemented with one pure resolver (`ui/src/utils/analytics/eventNaming.ts`) called there, instead of touching every call site.
- Adds `resolvePosthogEventName(type, properties)` implementing the `app.<object>.<action>` mapping agreed in the tracking issue below (flat 1:1 renames, plus `editor_tab_action` / `ossauth` / `onboarding` split by their discriminating property, with `onboarding`'s discriminator read from the nested `onboarding.action`). Unmapped events fall back to the lowercased `type` so the dbt drift monitor (kestra-io/data#428) still catches anything unmapped.
- The backend `/v1/reports/events` payload (`type`) is deliberately left untouched — backend event naming is explicitly out of scope per data#354 Q7. Resolving the PostHog name only inside `posthogEvents()` keeps the two paths decoupled with no call-site changes.
- Two in-tutorial onboarding actions (`flow_saved_during_tutorial`, `flow_executed_during_tutorial`) are folded into `app.onboarding-step.viewed` per product decision, with the original action retained as a property.

part-of: https://github.com/kestra-io/data/issues/354

## Test plan

- [x] `npm run test:unit` — 789/789 tests pass, including new `eventNaming.spec.ts` covering every flat/split mapping, the nested `onboarding.action` discriminator, unmapped-type fallback, and a drift guard asserting the `app.<object>.<action>` shape
- [x] `npm run test:lint` — 0 errors
- [x] `node src/translations/check.js` (now `check.ts`) — no missing/extra keys in any locale (this change adds no i18n keys)

### DIFF
--- a/ui/src/stores/api.ts
+++ b/ui/src/stores/api.ts
@@ -6,6 +6,7 @@ import {useMiscStore} from "override/stores/misc"
 import {capturePosthogEvent, disablePosthog} from "../utils/posthog"
 import {ensureUid} from "../utils/uid"
 import {PendingEventsBuffer} from "../utils/analytics/pendingEvents"
+import {resolvePosthogEventName} from "../utils/analytics/eventNaming"
 
 export const API_URL = "https://api.kestra.io"
 
@@ -186,7 +187,7 @@ export const useApiStore = defineStore("api", () => {
         delete finalData.date
         delete finalData.counter
 
-        const eventName = type === "PAGE" ? "$pageview" : data.type.toLowerCase()
+        const eventName = type === "PAGE" ? "$pageview" : resolvePosthogEventName(data.type, finalData as Record<string, any>)
 
         if (type === "PAGE") {
             const origin = data.page?.origin ?? window.location.origin
--- a/ui/src/utils/analytics/eventNaming.spec.ts
+++ b/ui/src/utils/analytics/eventNaming.spec.ts
@@ -1,37 +1,39 @@
 import {describe, expect, it} from "vitest"
 import {resolvePosthogEventName} from "./eventNaming"
 
+const fixtures: [string, Record<string, any>, string][] = [
+    ["flow_execution", {}, "app.flow.executed"],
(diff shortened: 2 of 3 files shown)


flcarre:
## QA — PASS

Browser QA of the running OSS monorepo from this branch (backend `:cli:runLocal` on :8080, Vite UI on :5399).

**Scenario A — resolver logic (in-browser): ✅ 30/30**
Imported the shipped `ui/src/utils/analytics/eventNaming.ts` live in the running app and asserted `resolvePosthogEventName()` for 30 cases. All produced the expected `surface.object.action` names: flat renames (`flow_execution`→`app.flow.executed`, `error`→`app.error.occurred`, `setup_flow:*`, `survey_*`, `ai_copilot`, `blueprint`), the three catch-all splits (`editor_tab_action`, `ossauth`, `onboarding`), the folded in-tutorial onboarding actions (→ `app.onboarding-step.viewed`), `ossauth` with no action → `app.oss-auth.completed`, and unknown-type / unknown-discriminator fallbacks.

**App boot / regression: ✅**
SPA renders, zero console errors.

**Live emitting path: ✅**
A client-side navigation fires the `PAGE` event through the modified `posthogEvents()` choke point; the path executes cleanly and the posthog modules lazy-load, with no errors — confirming the `api.ts` integration is regression-free.

**Not run:** in-app execute/editor/blueprint surfaces (reaching them needs first-admin account creation). Those call sites are unchanged by this PR and feed the single verified choke point; combined with 34/34 unit tests + green type-check/lint/translations, residual risk is negligible.

**Note:** PostHog emission is intentionally disabled in dev builds (`MODE==="development"`), so live `app.*` emission verification is deferred to staging/prod, backed by the kestra-io/data#428 drift monitor.

elevatebart (approved):

PR #17577, opened 2026-07-22
fix(core): order SubflowExecutionEnd batch locks to avoid executor deadlock

### LINKED ISSUE #17574: Executor shuts down on transient PostgreSQL deadlock during SubflowExecutionEnd batch processing
## Environment

- **Kestra version:** 1.3.26 (also present on current `releases/v1.3.x`)
- **Backend:** PostgreSQL 14.18
- **Topology:** JDBC runner, multiple executor replicas (Docker Swarm), subflow-heavy workload (`*_Master` flows fanning out many `*_Worker` subflow executions with `waitForExecution: true`)

## Symptom

Executor processes shut themselves down under load. The user-visible errors are a shutdown that cannot complete cleanly:

```
WARN  ForkJoinPool.commonPool-worker-5 io.kestra.jdbc.runner.JdbcQueue Error while stopping polling
org.awaitility.core.ConditionTimeoutException: Condition with Lambda expression in io.kestra.jdbc.runner.JdbcQueue was not fulfilled within 30 seconds.
    at io.kestra.jdbc.runner.JdbcQueue.lambda$poll$2(JdbcQueue.java:522)
    at io.kestra.jdbc.runner.JdbcExecutor.close(JdbcExecutor.java:1729)
...
ERROR jdbc-queue-SubflowExecutionEnd_0 io.kestra.jdbc.runner.JdbcQueue Couldn't wait for queue executor to close properly, forcing shutdown
```

These are emitted **during** shutdown; they are not the trigger.

## Trigger (from logs)

The shutdown is initiated by a fatal error while polling the `SubflowExecutionEnd` queue:

```
ERROR jdbc-queue-SubflowExecutionEnd_0 io.kestra.jdbc.runner.JdbcQueue Fatal error while polling the 'io.kestra.core.runners.SubflowExecutionEnd' queue. Initiating shutdown.
org.jooq.exception.DataAccessException: SQL [select "value" from executions where ("key" = ? and "deleted" = ?) for update]; ERROR: current transaction is aborted, commands ignored until end of transaction block
    ...
    at io.kestra.jdbc.repository.AbstractJdbcExecutionRepository.lock(AbstractJdbcExecutionRepository.java:911)
    at io.kestra.jdbc.runner.JdbcExecutor.subflowExecutionEndQueue(JdbcExecutor.java:1057)
    at java.base/java.util.ArrayList.forEach(Unknown Source)
    at io.kestra.jdbc.runner.JdbcExecutor.lambda$run$11(JdbcExecutor.java:347)
    at io.kestra.jdbc.runner.JdbcQueue.lambda$receiveImpl$2(JdbcQueue.java:420)
Caused by: org.postgresql.util.PSQLException: ERROR: current transaction is aborted, commands ignored until end of transaction block
Caused by: org.postgresql.util.PSQLException: ERROR: deadlock detected
  Detail: Process 33005 waits for ShareLock on transaction 1133733317; blocked by process 33503.
Process 33503 waits for ShareLock on transaction 1133732538; blocked by process 33005.
  Where: while locking tuple (27527,15) in relation "executions"
```

The root of the `Caused by` chain is `ERROR: deadlock detected` on `relation "executions"`. The visible top-level `current transaction is aborted` (SQLState `25P02`) is a follow-on error on the already-aborted transaction, not the originating failure. A second log file from a different container shows the identical pattern with a 3-process cycle.

## Sequence of facts

1. The `SubflowExecutionEnd` queue is consumed in-transaction: `JdbcExecutor` registers it via `receiveTransaction(null, Executor.class, (dslContext, eithers) -> eithers.forEach(this::subflowExecutionEndQueue))` (`JdbcExecutor.java:345-348`).

2. `JdbcQueue.receiveImpl` wraps the whole poll in a single transaction and invokes the consumer for the fetched batch inside it (`JdbcQueue.java:413-430`). `receiveFetch` limits the batch to `pollSize`, which defaults to **100** (`JdbcQueue.java:269`, `Configuration.pollSize = 100` at `JdbcQueue.java:581`).

3. For each message in the batch, `subflowExecutionEndQueue` calls `executionRepository.lock(parentExecutionId, ...)` (`JdbcExecutor.java:1057`), which issues `SELECT "value" FROM executions WHERE "key" = ? AND "deleted" = ? FOR UPDATE` on a single parent-execution row (`AbstractJdbcExecutionRepository.lock`, `AbstractJdbcExecutionRepository.java:911-922`).

4. Therefore a single transaction acquires up to 100 `FOR UPDATE` row locks on `executions`, sequentially. With multiple executor replicas each processing such a batch, two transactions can acquire overlapping parent rows in opposite order, producing a circular wait — the deadlock reported by PostgreSQL. (The queue-table fetch itself uses `FOR UPDATE SKIP LOCKED` — `PostgresQueue.receiveFetch` — so queue rows are not part of the cycle.)

5. Deadlocks are retried by `JooqDSLContextWrapper`: its retry predicate matches SQLState `40001`, `40P01`, and MySQL error `1205`, with unlimited attempts capped at 60s (`JooqDSLContextWrapper.java`). However, the deadlock aborts the shared batch transaction; subsequent statements in that transaction fail with SQLState `25P02` (`current transaction is aborted`), which is **not** in the retry predicate. The error that ultimately propagates is the `25P02`, so neither the inner nor the outer retry layer retries it.

6. The propagated exception reaches the poll loop's catch-all at `JdbcQueue.java:495-502`:

```java
} catch (Exception e) {
    log.error("Fatal error while polling the '{}' queue. Initiating shutdown.", queueType(), e);
    running.set(false);
    KestraContext.getContext().shutdown();
}
```

This calls `KestraContext.getContext().shutdown()`, taking down the executor. The `awaitility` timeout and "Couldn't wait for queue executor to close properly" messages are produced by the subsequent shutdown.

## When this behavior appeared

The catch-all that calls `shutdown()` was added in commit `42cb301d67` — `fix(core): keep JDBC queue consumer alive on transient DB errors (#16816)` (2026-06-16), present in the 1.3.26 build (`revision b663834 / 2026-06-30`). Before that commit, an exception escaping the poll loop terminated the queue's poll thread without shutting down the process.

## Reproduction conditions

- PostgreSQL backend, two or more executor replicas.
- A parent flow that fans out many `subflow` tasks with `waitForExecution: true`, run concurrently so that many `SubflowExecutionEnd` messages are processed in the same batch across replicas, generating concurrent `FOR UPDATE` locks on `executions` rows.

## Attached data

Two executor container log files from the reporting user; both show exactly one `deadlock detected` → `Fatal error while polling the 'SubflowExecutionEnd' queue. Initiating shutdown.` sequence followed by the shutdown-stall messages.

nkwiatkowski:
## What & why

Fixes #17574.

Under heavy subflow fan-out, executors on a Postgres backend crash in a
restart loop. Root cause (full analysis in #17574):

1. The `SubflowExecutionEnd` queue consumer processes a batch of up to
   `pollSize` (default 100) messages in **one transaction**, each taking a
   `SELECT … FOR UPDATE` on a parent `executions` row.
2. Two executors locking overlapping parent rows in different order deadlock
   on `executions`. The deadlock (`40P01`) aborts the shared transaction and
   subsequent statements fail with `25P02` (*current transaction is aborted*),
   which the retry predicate does not match.
3. The `25P02` reaches `JdbcQueue.poll`'s catch-all, which (since #16816 made
   poll failures fatal) calls `KestraContext.shutdown()` → executor crash.

## Changes

- **`JdbcExecutor` — group + ordered locking.** The `SubflowExecutionEnd` batch
  is grouped by parent execution id and processed once per parent in sorted,
  deterministic order. Concurrent batches on different executors now acquire
  overlapping parent locks in the same order, so a lock cycle can no longer
  form. Grouping also collapses the repeated read-lock of the same parent row
  during a fan-out. The consumer stays `receiveTransaction`, so the cross-JVM
  duplicate-prevention is unchanged, and it remains one commit per batch.
- **`JdbcQueue` — transient deadlocks are no longer fatal.** The poll-loop
  catch now inspects the whole cause chain for a transient DB error
  (deadlock `40P01`/`40001`, MySQL lock-wait `1205`) — necessary because the
  deadlock hides behind the top-level `25P02`. On a transient error it logs and
  re-polls a fresh batch instead of shutting down; genuinely unrecoverable
  errors still fail fast. This also protects against deadlocks originating from
  the periodic delay/SLA sweeps (other multi-row lockers).

## Tests

- New `JdbcQueueTransientErrorTest`: deadlock-behind-aborted-tx, direct `40P01`,
  `40001`, MySQL `1205`, non-transient → false, null → false.
- `H2QueueTest` (incl. `shouldShutdownWhenConsumerFails`) still green —
  unrecoverable errors still shut down.
- `H2RunnerTest`: 86 tests pass, including all 12 subflow / forEachItem tests
  that exercise the `SubflowExecutionEnd` path.

## Notes

- Targeted at `releases/v1.3.x`; the same fix should be forward-ported to
  `develop`.
- `isTransientDatabaseError` intentionally mirrors the SQLState set in
  `JooqDSLContextWrapper.predicate()` (comment links them); happy to
  consolidate into a shared helper if preferred.

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/runner/JdbcExecutor.java
@@ -344,7 +344,7 @@ public class JdbcExecutor implements ExecutorInterface {
         this.receiveCancellations.addFirst(this.subflowExecutionResultQueue.receive(Executor.class, this::subflowExecutionResultQueue));
         this.receiveCancellations.addFirst(
             ((JdbcQueue<SubflowExecutionEnd>) this.subflowExecutionEndQueue)
-                .receiveTransaction(null, Executor.class, (dslContext, eithers) -> eithers.forEach(this::subflowExecutionEndQueue))
+                .receiveTransaction(null, Executor.class, (dslContext, eithers) -> this.subflowExecutionEndQueue(eithers))
         );
         this.receiveCancellations.addFirst(this.multipleConditionEventQueue.receive(Executor.class, this::multipleConditionEventQueue));
         this.receiveCancellations.addFirst(maintenanceService.listen(new MaintenanceService.MaintenanceListener() {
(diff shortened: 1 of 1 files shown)


loicmathieu (approved):
LGTM

PR #17581, opened 2026-07-22
Refactor FlowsAPI.createFlow call signature

elevatebart:
### ✨ Description

Refactors the `createFlow` function in the flow store to align with the correct API call signature for `FlowsAPI.createFlow`. The change separates the request body parameters from the options parameters, passing them as distinct arguments rather than combining them into a single object with a type assertion.

**Changes:**
- Split the combined options object into two separate arguments: request body (`{body, draft}`) and API options (`{showMessageOnError}`)
- Removed unnecessary type assertion on the combined object
- Maintained functional equivalence while improving type safety and API contract clarity

### 🔗 Related Issue

No related issue.

### 🎨 Frontend Checklist

### 📝 Additional Notes

This is a refactoring change that corrects the API call pattern to match the actual function signature of `FlowsAPI.createFlow`, which expects separate parameters for the request body and options rather than a combined object.

🐱 Why don't cats play poker in the jungle? Too many cheetahs!

https://claude.ai/code/session_013uzpTRfqo6bmCiB2Dcjbdy

### DIFF
--- a/ui/src/stores/flow.ts
+++ b/ui/src/stores/flow.ts
@@ -541,11 +541,10 @@ export const useFlowStore = defineStore("flow", () => {
     }
 
     function createFlow(options: { flow: string, draft?: boolean }) {
-        return FlowsAPI.createFlow({
-            body: options.flow,
-            draft: options.draft ?? false,
-            showMessageOnError: false,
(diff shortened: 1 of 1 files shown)


elevatebart:
no

PR #17582, opened 2026-07-22
fix(plugins): revalidate plugin schemas so newly added plugins autocomplete (#12102)

jymaire:
## What & why

Part of #12102 — newly installed plugins on a standalone install show up on the `/plugins` catalog page and execute fine, but do **not** appear in the flow editor's `type:` autocomplete and are flagged as unknown types.

This PR addresses the **client-side / HTTP caching** facet of that issue. It is complementary to the schema-generation work in **#17508** (and **#17568**, which is stacked into #17508) — see [Related PRs](#related-prs) below. Files are disjoint, so the PRs merge independently.

### Root cause

The editor's `type:` autocomplete and type-validation are driven entirely by the aggregated schema at `GET /api/v1/plugins/schemas/{type}` (monaco-yaml fetches it directly). That endpoint was served with `Cache-Control: public, max-age=3600` and **no ETag**, while the catalog endpoints (`/plugins`, `/plugins/groups/subgroups`) send no cache header and read the live registry.

So after a plugin set change, the browser keeps serving the **stale schema for up to an hour with no way to revalidate** → the new plugin isn't offered in autocomplete and reads as an unknown type, even though `/plugins` already lists it. Reproduced live: with a plugin removed server-side, the editor still validated its type against the cached schema until the HTTP cache was bypassed (hard reload / "Disable cache").

A second, latent contributor: the server-side `JsonSchemaCache` populates via `computeIfAbsent` and its `clear()` had **zero callers**, so a runtime registry change (EE `plugins install`, future hot-reload) would never invalidate it.

### Changes

- **`PluginController`** — `schemas/{type}` and `properties/{type}` now respond with `Cache-Control: no-cache` + a strong **`ETag`** derived from the installed plugin set (`PluginRegistry#hash()`) and the server version. A matching `If-None-Match` is answered `304 Not Modified`, so repeat editor opens still skip re-downloading the ~720 KB schema — freshness without losing the caching win.
- **`JsonSchemaCache`** — injects `PluginRegistry` and drops the cached schemas when the registry hash changes (pull-based self-invalidation; the registry is a static singleton so a push-based `clear()` from it would be fragile and miss paths). `clear()` now also clears the properties cache.

### Tests

- `JsonSchemaCacheTest` (new) — serves from cache while the registry hash is stable; regenerates when it changes.
- `PluginControllerTest#flowSchemaIsRevalidatedWithEtag` (new) — asserts `no-cache` + non-blank `ETag`, and that a matching `If-None-Match` yields `304`.

All pass; `core` and `webserver` compile clean.

### Note for EE

`JsonSchemaCache`'s constructor gained a `PluginRegistry` parameter. It's DI-injected everywhere in OSS (no `new` call sites), so OSS is unaffected — but EE code that constructs or `@Replaces` it will need the matching constructor.

<a name="related-prs"></a>
## Related PRs — #12102 has more than one root cause

This issue has **two distinct failure modes**, fixed at different layers. This PR is the **cache** layer; the **schema-generation** layer is handled separately:

| PR | Layer | What it fixes |
|----|-------|---------------|
| **#17508** (open) + **#17568** (merged into #17508's branch) | Schema **generation** (`JsonSchemaGenerator`) | A plugin built against an incompatible Kestra version fails subtype resolution and **vanishes from the schema entirely** (or crashes the whole flow schema). Adds a plain-resolve fallback, a placeholder schema for unresolved types, and per-plugin generation-failure isolation. |
| **this PR (#17582)** | HTTP **cache** (`PluginController` / `JsonSchemaCache`) | The schema *does* contain the plugin, but the browser serves a **stale cached copy** after a plugin-set change. Adds `no-cache` + ETag revalidation and registry-hash cache invalidation. |

They are complementary and synergistic (both use the plugin registry as source of truth: #17568 reconciles the schema against it, this PR keys the cache on its hash). **Suggested closing PR for #12102: #17508** — hence this PR says *"Part of #12102"* rather than *"Fixes"*, to avoid double-closing the issue.

## How to reproduce (on the buggy version, before this fix)

**Prerequisites:** a Kestra **standalone** on an affected version (e.g. `1.2.x`) and any plugin JAR that is *not* bundled in that standalone — e.g. build one locally: `git clone` a plugin repo and run `./gradlew shadowJar` (the JAR lands in `build/libs/`). Kafka (`plugin-kafka`) is a good example. Run under Java 21 (1.x standalones don't support the Java 25 removed Security Manager).

1. Start the standalone with an **empty** plugins directory:
   `java -jar kestra-<version> server standalone --plugins ./plugins`
2. In the UI, open (or create) a flow in the editor and let it load. This makes monaco-yaml fetch `GET /api/v1/plugins/schemas/flow`, which the browser then caches for one hour (`Cache-Control: public, max-age=3600`, no `ETag`). Keep DevTools **"Disable cache" OFF** — the bug depends on the browser reusing this cache.
3. Stop the server, drop the plugin JAR into `./plugins`, and restart.
4. Confirm the plugin really is installed: it appears on the **Plugins** page, and a flow using one of its task types executes successfully.
5. Back in the flow editor, reload the page normally (**F5**). The new task type (e.g. `io.kestra.plugin.kafka.Produce`) is **not** offered in `type:` autocomplete and is treated as an unknown type — because the browser is still serving the schema cached in step 2, even though the server's schema now contains it. This is the reported symptom: *installed & visible in /plugins, but no editor autocomplete*.
6. Hard-reload (**Ctrl+Shift+R**) or tick DevTools **"Disable cache"**: the schema is re-fetched from the server and the type is recognized immediately — confirming the stale HTTP cache is the cause, not the registry or the schema generation.

**After this fix:** step 5 recognizes the plugin without any hard-reload, because `/plugins/schemas/*` is served with `Cache-Control: no-cache` + an `ETag` that changes whenever the installed plugin set changes. The browser revalidates on every use and gets the fresh schema (or a lightweight `304 Not Modified` when nothing changed). Verified on a standalone built from this branch: `/plugins/schemas/flow` returns `Cache-Control: no-cache` + `ETag`, and a conditional request with the matching `If-None-Match` returns `304`.

> Tip: the effect is symmetric and easy to demonstrate either direction — you can also start *with* the plugin (editor caches a schema that has it), then remove it + restart, and watch the editor keep validating the now-removed type until a hard-reload.

### DIFF
--- a/core/src/main/java/io/kestra/core/docs/JsonSchemaCache.java
+++ b/core/src/main/java/io/kestra/core/docs/JsonSchemaCache.java
@@ -6,11 +6,13 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicLong;
 
 import io.kestra.core.models.dashboards.Dashboard;
 import io.kestra.core.models.flows.Flow;
 import io.kestra.core.models.tasks.Task;
 import io.kestra.core.models.triggers.AbstractTrigger;
+import io.kestra.core.plugins.PluginRegistry;
 
 import jakarta.inject.Singleton;
 
@@ -21,19 +23,27 @@ import jakarta.inject.Singleton;
 public class JsonSchemaCache {
 
     private final JsonSchemaGenerator jsonSchemaGenerator;
+    private final PluginRegistry pluginRegistry;
 
     private final ConcurrentMap<CacheKey, Map<String, Object>> schemaCache = new ConcurrentHashMap<>();
     private final ConcurrentMap<SchemaType, Map<String, Object>> propertiesCache = new ConcurrentHashMap<>();
 
     private final Map<SchemaType, Class<?>> classesBySchemaType = new HashMap<>();
 
+    // Hash of the plugin registry the cached schemas were generated from. When the registry changes
+    // (a plugin is installed or removed at runtime) the cached schemas become stale and must be dropped,
+    // otherwise the editor keeps completing/validating against a schema missing the new plugin (#12102).
+    private final AtomicLong cachedRegistryHash = new AtomicLong(Long.MIN_VALUE);
+
     /**
      * Creates a new {@link JsonSchemaCache} instance.
      *
      * @param jsonSchemaGenerator The {@link JsonSchemaGenerator}.
+     * @param pluginRegistry      The {@link PluginRegistry} whose content the cached schemas depend on.
      */
-    public JsonSchemaCache(final JsonSchemaGenerator jsonSchemaGenerator) {
+    public JsonSchemaCache(final JsonSchemaGenerator jsonSchemaGenerator, final PluginRegistry pluginRegistry) {
         this.jsonSchemaGenerator = Objects.requireNonNull(jsonSchemaGenerator, "JsonSchemaGenerator cannot be null");
+        this.pluginRegistry = Objects.requireNonNull(pluginRegistry, "PluginRegistry cannot be null");
         registerClassForType(SchemaType.FLOW, Flow.class);
         registerClassForType(SchemaType.TASK, Task.class);
         registerClassForType(SchemaType.TRIGGER, AbstractTrigger.class);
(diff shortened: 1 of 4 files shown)


jymaire:
### Companion EE PR

The EE compile check failed because this PR changed the `JsonSchemaCache` constructor signature (added `PluginRegistry`), and the EE `KestraBeansFactory.jsonSchemaCache()` @Primary override still called the old single-arg constructor.

Fixed in **kestra-io/kestra-ee#9464** (branch name matches so the EE compile check resolves it). Merge together.

ℹ️ The remaining `AssetService.asyncUpsert` compile error is **pre-existing on EE `develop`** (from `feat(assets): asset locking via durable leases`) and unrelated to this PR.

fdelbrayelle (approved):

PR #17584, opened 2026-07-22
feat(core): add stable machine-readable error codes to API error responses

fhussonnois:
Error bodies built by ErrorController now carry a stable `code` field (ErrorCode enum) alongside the human-readable message, so API clients can branch on the error type instead of parsing free-text messages.

- ErrorCode enum + HasErrorCode contract in core exceptions; NotFoundException carries NOT_FOUND, FlowNotFoundException carries FLOW_NOT_FOUND
- ApiError extends Micronaut's JsonError with the additive code property
- KestraHttpStatusException: an HttpStatusException carrying an ErrorCode, for filters that respond with a status directly
- code resolution: exception-provided code > handler-specific code >
  HTTP-status default; existing jsonError(...) signatures are unchanged
- TenantValidationFilter now emits TENANT_NOT_FOUND

See: https://github.com/kestra-io/kestra-ee/pull/9436

### DIFF
--- a/core/src/main/java/io/kestra/core/exceptions/ErrorCode.java
+++ b/core/src/main/java/io/kestra/core/exceptions/ErrorCode.java
@@ -1,0 +1,45 @@
+package io.kestra.core.exceptions;
+
+/**
+ * Stable machine-readable codes exposed in API error responses so that clients can branch on the
+ * error type instead of parsing free-text messages.
+ *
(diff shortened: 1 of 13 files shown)


elevatebart:
What can I help you with here?

loicmathieu (approved):
LGTM, but I wonder if switching to the "almost standard" problem standard would not be better...

PR #17585, opened 2026-07-22
feat(executions): render HTML file outputs in a sandboxed iframe preview

### LINKED ISSUE #5332: Render HTML file outputs in the Outputs preview
### Feature description

Adding this inside an iframe with good restrictions will allow us to add it without compromising security

```yaml
id: html
namespace: company.team
tasks:
  - id: get
    uri: https://huggingface.co/datasets/kestra/datasets/resolve/main/html/forecast.html
```

Tanweer-Ahmed-Chiktay:
## Summary

Closes #5332.

- Detects `.html`/`.htm` output files by extension in `FilePreview.vue`
- Renders them in a sandboxed `<iframe>` using `sandbox="allow-scripts"` (no `allow-same-origin`): scripts can run, but the frame cannot access the parent page's cookies, localStorage, or make authenticated requests on the user's behalf
- The file URL is the same `/executions/{id}/file?path=…` endpoint already used for downloads, so the browser's existing session authenticates the request
- Displays filename, file size, and a download button above the iframe — matching the existing button-bar pattern
- Falls back to the unchanged `RawPreview` path for all non-HTML file types
- Adds `file_preview.html_preview_title` translation key to all 13 supported locales (en, de, es, fr, hi, it, ja, ko, pl, pt, pt_BR, ru, zh_CN)

## Test plan

- [ ] Create a Kestra flow that writes an HTML file output (e.g. a Plotly chart or simple HTML page)
- [ ] Open the execution's Outputs tab and click the `.html` file — it should render inline in an iframe
- [ ] Verify clicking a `.txt`, `.json`, `.csv`, or binary file still shows the existing RawPreview / big-file warning
- [ ] Verify the Download button works for the HTML file
- [ ] Verify no console errors or mixed-content warnings in the browser


elevatebart (changes requested):
Thanks for this @Tanweer-Ahmed-Chiktay this is indeed useful.
I specially like the care you took of sandboxing the iframe. 
Good thinking!

### Tests

Can you share the flow you tested this with?
I can't make one that outputs correctly. 
They all return a download prompt instead.

### Assets

Also, all local assets, images and stylesheets will error since there is no server to respond and make them available.
I wonder if we want to do something about it before merging the feature. 
A service worker that acts as a proxy?
A full on backend miniwebsite?

### Performance

One last thing: if the returned html is big, we should avoid rendering it in full and still use the bigfile

Tanweer-Ahmed-Chiktay:
Thanks @elevatebart — pushed a follow-up addressing all three points.

### Tests

The download prompt was the root problem, and it also explains why nothing rendered for you. The `/executions/{id}/file` endpoint sends `Content-Disposition: attachment`, so a browser navigating an `iframe src` to it always downloads instead of rendering. Switched the iframe to `srcdoc`: the file is fetched as text and injected, which bypasses that header entirely.

While fixing it I also found the first `srcdoc` source (the `/file/preview` endpoint) truncates — it caps rows (100 default / 5000 max) and bytes (2 MB in `TextFileRenderer`) — so larger outputs would render as a broken partial document (cut-off tags/scripts). The preview now fetches the **complete** file via a new `fileContent()` store method instead.

Flow I verified with:

```yaml
id: html-output-preview-test
namespace: io.kestra.tests
tasks:
  - id: generate_report
    script: |
      html = """<!DOCTYPE html>
      <html lang="en">
      <head><meta charset="UTF-8"><title>Forecast</title>
      <style>h1 { color: #8405ff; } .bar { display:inline-block; width:40px; background:#8405ff; }</style></head>
      <body>
        <h1>Forecast report</h1>
        <div id="chart"></div>
        <script>
          [40,90,60,120,80,150].forEach(v => {
            const b = document.createElement('div');
            b.className = 'bar'; b.style.height = v + 'px';
            document.getElementById('chart').appendChild(b);
          });
        </script>
      </body></html>"""
      with open("report.html", "w") as f:
          f.write(html)
    outputFiles:
      - "*.html"
```

Open the execution → Outputs → `report.html`; it renders inline and the inline script runs inside the sandbox.

Added unit tests for both the component (srcdoc holds full content, sandbox attributes, big-file guard, empty file, error + retry, undefined path) and the store wiring (hits `/file` not `/preview`, identity `transformResponse` so the HTML isn't JSON-parsed, error propagation).

### Assets

Correct — with `srcdoc` the base URL is `about:blank`, so relative `img`/`link`/`script` paths won't resolve. This preview is scoped to self-contained documents (inline/CDN assets, data URIs — e.g. Plotly with `include_plotlyjs`), which covers the example in the issue. Added a dismissable-free info note in the preview so users know local assets won't load.

For assets I'd go with a backend inline-serving endpoint over a service worker — the SW can't control a sandboxed opaque-origin iframe without dropping the sandbox, and it still needs the backend anyway. A `/executions/{id}/output/{path}` endpoint serving inline with correct content-types (and a path-traversal guard) lets the iframe use a sandboxed `src`, so relative assets resolve while the sandbox stays intact. Happy to build that as a follow-up if you agree on the direction — it's a meaningful chunk of backend + a security review, so it felt better separated from this PR. Let me know if you'd rather it land here.

### Performance

Reused the existing big-file path: `getFileMeta()` runs first and if the file is `>= 10 MB` the preview shows the big-file warning + "Load anyway" instead of loading it — so nothing large is pulled into memory (the `srcdoc` holds the whole document as a string, so this matters more here than for the text preview). "Load anyway" fetches the full content on explicit opt-in.

Tanweer-Ahmed-Chiktay:
CI fix: the frontend `check:types` job failed on a `TS7006` — the `transformResponse` callback in the new `fileContent()` store method had an implicitly-typed parameter. Annotated it (`(data: string) => data`); it's the identity transform, so runtime behavior is unchanged. Pushed in 6ea03864d.

elevatebart (approved):
Let's risk it

MilosPaunovic:
Is this ready to be merged @elevatebart?

PR #17586, opened 2026-07-22
fix(core): surface meaningful 503 errors when AI Copilot provider fails

### LINKED ISSUE #17041: AI Copilot returns unclear error message when unsuccessful at generating flow
### Describe the issue

Users are finding the AI Copilot keeps returning the following:
```
I can only assist with creating Kestra flow
```
This can happen when it does not succeed at generating a flow for whatever reason. The error message comes up even when users have specified a clear flow.

Sometimes it can be due to invalid/expired API keys too.

### Environment

- Kestra Version: `1.3.21` and probably newer.

Tanweer-Ahmed-Chiktay:
## Summary

Closes #17041.

Before this fix, any exception thrown by LangChain4j that was **not** an `AiException` (e.g. an HTTP 401 from the upstream AI provider for an invalid API key, a 429 rate-limit response, a quota exhaustion error, or a transient network failure) would propagate through to Micronaut's generic `Throwable` handler and produce a cryptic `500 Internal Server Error` with no actionable information for the user.

Changes:
- Added try-catch to `generateFlow` and `generateDashboard` in `AiController`:
  - `AiException` is re-thrown so the existing global error handler (which maps it to 422) continues to work
  - All other exceptions are caught, logged, and mapped to **503 Service Unavailable** with a human-readable body
- Added `providerErrorMessage(Exception)` static helper that classifies the exception message by keyword:
  - 401 / unauthorized / api key → authentication failure message
  - 429 / rate limit → rate-limit message
  - quota / billing → quota/billing message
  - anything else → generic "temporary issue" message
- Added two new unit tests in `AiControllerTest`:
  - `shouldReturn503WhenProviderReturnsUnauthorized` — WireMock stubs a 401 from the provider; asserts the client sees 503
  - `shouldClassifyProviderErrorByKeyword` — unit-tests the classification logic directly for auth, rate-limit, quota, and unknown cases

## Test plan

- [ ] Configure an AI provider with an intentionally invalid API key; trigger flow generation — should now see a 503 with an auth-failure message instead of a generic 500
- [ ] Run `./gradlew :webserver:test --tests "io.kestra.webserver.controllers.api.AiControllerTest"` — both new tests should pass
- [ ] Run existing `mTLS` and null-yaml regression tests — should be unaffected

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/AiController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/AiController.java
@@ -7,6 +7,7 @@ import java.util.Map;
 import com.fasterxml.jackson.annotation.JsonCreator;
 
 import io.kestra.core.tenant.TenantService;
+import io.kestra.libs.copilot.exceptions.AiException;
 import io.kestra.webserver.services.ai.AiServiceInterface;
 import io.kestra.webserver.services.ai.AiServiceManager;
 import io.kestra.webserver.services.ai.GenerationResult;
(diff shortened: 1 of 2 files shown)


MilosPaunovic:
Hey @Tanweer-Ahmed-Chiktay,

Thanks for the effort on this, but as this version of `AI Copilot` will be deprecated soon, I'll have to close this PR.

PR #17596, opened 2026-07-23
fix(webserver): mark PagedResults.total as required

elevatebart:
Restores the required `total` contract on `PagedResults` (lost when cursor pagination boxed it to `Long` with no annotation), and splits the log-search cursor path into its own `CursorPagedResults` type so it can still omit `total` legitimately. Fixes 18 generated-SDK TypeScript errors in `kestra-ee`'s `ui-ee` stores.

🐱 Why did the cat sit on the `total` field? Because it heard it was finally required to be *purr*-sistent.

⚠️ https://github.com/kestra-io/kestra-ee/pull/9470 is blocked until this PR and kestra-io/kestra-ee#9463 are both merged.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FHkHL9QBwMrST5bbJj6pLg)_

### DIFF
--- a/.github/workflows/update-generated-sdk.yml
+++ b/.github/workflows/update-generated-sdk.yml
@@ -1,0 +1,182 @@
+name: Update Generated SDK
+
+on:
+  push:
+    branches:
+      - develop
+  workflow_dispatch:
+
+concurrency:
+  group: update-generated-sdk-${{ github.repository }}
+  cancel-in-progress: false
+
+jobs:
+  update-generated-sdk:
+    name: Update Generated SDK
+    runs-on: ubuntu-latest
+    if: github.repository == 'kestra-io/kestra'
+    steps:
+      - uses: kestra-io/actions/composite/checkout-and-auth@main
+        name: Checkout
+        with:
+          gh-app-id: ${{ secrets.GH_BOT_APP_ID }}
+          gh-app-private-key: ${{ secrets.GH_BOT_PRIVATE_KEY }}
+          repository: kestra-io/kestra
+          ref: develop
+
+      # github.event.head_commit is only populated for push events; resolve the commit
+      # subject from history so it's correct regardless of what triggered this run.
+      - name: Resolve develop commit message
+        id: source-commit
+        shell: bash
+        run: echo "message=$(git log -1 --format=%s)" >> "$GITHUB_OUTPUT"
+
+      - uses: kestra-io/actions/composite/setup-build@main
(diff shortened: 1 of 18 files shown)


nkwiatkowski (changes requested):
You can't just change PagedResults to CursorPagedResults because now that we can ahve many log repository backend, some will return pagination with total, and some will return cursur without total. The frontend must be able to handle both.

elevatebart:
Thanks for the review — the dual-shape handling is still fully there, just moved rather than removed. `CursorPagedResults.of(Page<T>)` is byte-for-byte the same branching logic that used to live in `PagedResults.of(Page<T>)`:

```java
if (page.hasTotalSize()) {
    pagedResults.total = page.getTotalSize();
    pagedResults.type = PaginationType.OFFSET;
} else {
    pagedResults.type = PaginationType.CURSOR;
    // ... nextCursor from the CursoredPage
}
```

`LogController.searchLogs` still calls this exact method against whatever `Page<LogEntry>` the active `LogDataStoreInterface` backend returns, so a JDBC/H2 store (offset, with `total`) and an external cursor-based store both still serialize correctly for that one endpoint — nothing about backend-dependent shape was dropped.

The only thing that changed is the *schema name* for that one endpoint (`PagedResultsLogEntry` → `CursorPagedResultsLogEntry`), which was necessary so the ~30 other list endpoints (which never have a cursor mode and always populate `total`) can mark `total` as required in the OpenAPI/generated-SDK schema — that's the actual bug this PR fixes (see the [PR description](https://github.com/kestra-io/kestra/pull/17596) for the 18 TypeScript errors it resolves). Frontend code consuming logs (`ui/src/stores/logs.ts`) doesn't reference the schema type by name, so it's unaffected by the rename.

Happy to add a code comment or rename differently if that would make the split clearer — let me know if you still see a gap I'm missing.

---

elevatebart:
Renamed `CursorPagedResults` → `CursorOrOffsetPagedResults` (states both supported pagination modes explicitly rather than implying cursor is the default).

On whether `total` makes sense here: kept it, optional. The default JDBC/H2 log store always knows it (so dropping it there would regress the UI's result count), it's only unknown for a backend that genuinely can't count without a full scan — the frontend already handles that (`ui/src/stores/logs.ts` does `response.total ?? 0`).

Pushed as `50fea83`. Let me know if you'd still like a different name or approach.

---

nkwiatkowski (approved):

PR #17597, opened 2026-07-23
feat(flows): publish a draft flow into a saved flow

flcarre:
## Summary

- Once a flow's latest revision is a draft, execution silently falls back to the last published revision, and there was no button to publish/promote the draft: the Save split-button's main action already publishes (`draft:false`) but is disabled whenever there are no unsaved changes, so a freshly opened, unchanged draft could not be published from any control.
- Add an explicit **Publish** action, enabled whenever the draft is valid (no flow errors), independent of unsaved changes:
  - In the flow editor top bar (`Actions.vue`), next to the Save split-button (demoted to secondary while a draft exists).
  - In `FlowRun`'s draft warning banner. `FlowRun.vue` is shared between OSS and EE, so this run-panel button ships to both editions from this one OSS change.
- `flowStore.publishDraft()` accepts an optional target flow, used by `FlowRun` (whose own flow reference lives in `executionsStore`, not `flowStore`) so it always publishes the flow actually shown to the user rather than whatever this store's `flow`/`flowYaml` last held from an unrelated page.
- Publishing reuses `save(false)`'s existing "saved" toast and the reactive clearing of the Draft badge/warning as feedback - no new confirm dialog or toast, consistent with normal flow-save semantics.
- Added all 12 non-English translations for the two new keys (`publish`, `publish_draft`); `node ui/src/translations/check.js` is clean for every locale.

Follow-up: an EE PR will add the same Publish affordance to the app-view/EE-only top-bar surfaces if any exist there.

part-of: https://github.com/kestra-io/kestra-ee/issues/9272

## Test plan

- [x] `npm run test:unit` passes (756/756, including the new `tests/unit/override/components/flows/Actions.spec.ts`)
- [x] `npm run test:lint` passes (no new warnings/errors)
- [x] `node ui/src/translations/check.js` reports "No missing keys. No extra keys." for every locale


flcarre:
## QA — PASS ✅

Built and ran the monorepo from this branch locally (OSS, embedded H2), driven through a real browser in **light and dark mode**.

| # | Scenario | Result |
|---|----------|--------|
| 1 | Save-as-draft → `Draft` badge + primary **Publish** button appear; Save split-button demoted to secondary | ✅ |
| 2 | **Publish enabled with zero unsaved changes** (the reported dead-end) → publishes, `draft:false`, badge + Publish disappear, Save returns primary | ✅ |
| 3 | Run-panel inline **Publish draft** → publishes and the draft warning clears | ✅ |
| 4 | Invalid draft → top-bar Publish **disabled** | ✅ |
| 5 | Light + dark rendering, no console errors | ✅ |

### A data-loss regression was caught and fixed during QA

The run panel (`TriggerFlow` → `FlowRun`) is embedded in the same editor top bar and shares the `flowStore` Pinia singleton, so an earlier iteration reloaded flow state into the shared editor buffer, wiping unsaved edits on publish. It now fetches the source with `store:false` and publishes it directly, never touching `flowYaml` / `flowYamlOrigin` / `haveChange`, and this is covered by a regression test. Re-verified live: with unsaved edits in the buffer, run-panel Publish publishes the saved draft, preserves the unsaved edits, and keeps the unsaved-changes indicator.

### Demo

https://github.com/user-attachments/assets/f1b04ea6-9c19-4866-8663-577ec2a08803

ui/src/components/flows/useFlowEditorActions.ts
```diff
@@ -75,6 +76,20 @@ export function useFlowEditorActions() {
         }
     }
 
+    async function publishDraft() {
+        try {
+            const outcome = await flowStore.publishDraft()
+            if (isSuccessfulFlowSaveOutcome(outcome)) {
+                onboardingStore.recordSave()
```
elevatebart, line 83: that is a really "odd" way to handle onboarding. I guess that is how it is done in other places within kestra. But I do not like to find bits and boops of "onboarding" inside the codebase.

elevatebart (approved):

ui/src/components/flows/useFlowEditorActions.ts
flcarre: Agreed, it's scattered. Kept it consistent with the existing pattern here rather than mixing conventions in this PR — opened #17616 to consolidate it into the store methods.

PR #17610, opened 2026-07-23
feat(core): add locked query filter and assetFailureBehavior task property

brian-mulier-p:
- Add `QueryFilter.Field.LOCKED` (EQUALS only), exposed exclusively on `Resource.ASSET`'s supported fields, backing the Assets page's locked/unlocked filter (see companion kestra-io/kestra-ee PR)
- Add a task-level `assetFailureBehavior` property (`IGNORE` default / `FAIL` / `WARN`) so a task can escalate when rendering or emitting its asset declaration fails, instead of the failure being silently logged

Refs kestra-io/kestra-ee#9466, kestra-io/kestra-ee#9467

### DIFF
--- a/executor/src/main/java/io/kestra/executor/ExecutorService.java
+++ b/executor/src/main/java/io/kestra/executor/ExecutorService.java
@@ -1432,72 +1432,63 @@ public class ExecutorService {
 
         if (taskRun.getState().isTerminated()) {
             log.trace("TaskRun terminated: {}", taskRun);
-            metricRegistry
-                .counter(
-                    MetricRegistry.METRIC_EXECUTOR_TASKRUN_ENDED_COUNT,
-                    MetricRegistry.METRIC_EXECUTOR_TASKRUN_ENDED_COUNT_DESCRIPTION,
-                    metricRegistry.tags(workerTaskResult)
-                )
-                .increment();
 
-            metricRegistry
-                .timer(
-                    MetricRegistry.METRIC_EXECUTOR_TASKRUN_ENDED_DURATION,
-                    MetricRegistry.METRIC_EXECUTOR_TASKRUN_ENDED_DURATION_DESCRIPTION,
-                    metricRegistry.tags(workerTaskResult)
-                )
-                .record(taskRun.getState().getDurationOrComputeIt());
+            // may be escalated below by assetFailureBehavior; kept separate from `taskRun` for the metrics in the finally block
+            TaskRun finalTaskRun = taskRun;
+            try {
+                // outputs
+                taskOutputService.saveOutputs(taskRun, workerTaskResult.getOutputs());
 
-            // outputs
-            taskOutputService.saveOutputs(taskRun, workerTaskResult.getOutputs());
-
-            ExecutionKind executionKind = Optional.ofNullable(executor.getExecution().getKind()).orElse(ExecutionKind.NORMAL);
-            if (
-                taskRun.getAssets() != null &&
-                    (!taskRun.getAssets().getInputs().isEmpty() || !taskRun.getAssets().getOutputs().isEmpty())
-                    && executionKind != ExecutionKind.TEST
-            ) {
-                AssetUser assetUser = new AssetUser(
-                    taskRun.getTenantId(),
-                    taskRun.getNamespace(),
-                    taskRun.getFlowId(),
-                    newExecution.getFlowRevision(),
-                    taskRun.getExecutionId(),
-                    taskRun.getTaskId(),
-                    taskRun.getId(),
-                    taskRun.getState().getCurrent(),
-                    taskRun.getState().getStartDate(),
-                    taskRun.getState().getEndDate().orElse(null)
-                );
-
-                List<AssetIdentifier> outputIdentifiers = taskRun.getAssets().getOutputs().stream()
-                    .map(asset -> asset.withTenantId(taskRun.getTenantId()))
-                    .map(AssetIdentifier::of)
-                    .toList();
-                List<AssetIdentifier> inputAssets = taskRun.getAssets().getInputs().stream()
-                    .map(assetIdentifier -> assetIdentifier.withTenantId(taskRun.getTenantId()))
-                    .toList();
-                try {
-                    assetService.assetLineage(
-                        assetUser,
-                        inputAssets,
-                        outputIdentifiers
+                ExecutionKind executionKind = Optional.ofNullable(executor.getExecution().getKind()).orElse(ExecutionKind.NORMAL);
+                if (
+                    taskRun.getAssets() != null &&
+                        (!taskRun.getAssets().getInputs().isEmpty() || !taskRun.getAssets().getOutputs().isEmpty())
+                        && executionKind != ExecutionKind.TEST
+                ) {
+                    AssetUser assetUser = new AssetUser(
+                        taskRun.getTenantId(),
+                        taskRun.getNamespace(),
+                        taskRun.getFlowId(),
+                        newExecution.getFlowRevision(),
+                        taskRun.getExecutionId(),
+                        taskRun.getTaskId(),
+                        taskRun.getId(),
+                        taskRun.getState().getCurrent(),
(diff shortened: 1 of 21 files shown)


core/src/main/java/io/kestra/core/models/tasks/Task.java
```diff
@@ -101,6 +101,14 @@ abstract public class Task implements TaskInterface {
     @Nullable
     private AssetsDeclaration assets;
 
+    @Builder.Default
+    @PluginProperty(hidden = true, group = "advanced")
+    @Schema(
+        title = "Asset failure behavior",
+        description = "Behavior applied to the task state when a declared asset fails to render, emit, or be persisted (e.g. a lock conflict): FAIL escalates it to FAILED, WARN (default) warns it if it would otherwise succeed, IGNORE leaves the state untouched. `allowWarning` still downgrades a WARN-escalated state to SUCCESS, but a FAIL-escalated state is never softened by `allowFailure` — a task's own genuine failure is still governed by `allowFailure` as before."
+    )
+    private AssetFailureBehavior assetFailureBehavior = AssetFailureBehavior.WARN;
```
loicmathieu: Can't it be inside the AssetsDeclaration to avoid adding a top level attribute?

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -1432,72 +1432,122 @@ public void addWorkerTaskResult(ExecutorContext executor, Supplier<FlowWithSourc
...
+            TaskRun finalTaskRun = taskRun;
+            try {
+                // outputs
+                taskOutputService.saveOutputs(taskRun, workerTaskResult.getOutputs());
+
+                ExecutionKind executionKind = Optional.ofNullable(executor.getExecution().getKind()).orElse(ExecutionKind.NORMAL);
+                if (
+                    taskRun.getAssets() != null &&
+                        (!taskRun.getAssets().getInputs().isEmpty() || !taskRun.getAssets().getOutputs().isEmpty())
+                        && executionKind != ExecutionKind.TEST
```
loicmathieu, line 1446: Do you really want to also process assets for PLAYGROUND and LOOP execution?

If it's only for normal execution it's better to check on null or NORMAL

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -1432,72 +1432,122 @@ public void addWorkerTaskResult(ExecutorContext executor, Supplier<FlowWithSourc
...
+                        .map(assetIdentifier -> assetIdentifier.withTenantId(taskRun.getTenantId()))
+                        .toList();
+                    try {
+                        assetService.assetLineage(
+                            assetUser,
+                            inputAssets,
+                            outputIdentifiers
+                        );
+                    } catch (QueueException e) {
+                        log.warn("Unable to submit asset lineage event for {} -> {}", inputAssets, outputIdentifiers, e);
```
loicmathieu:
```suggestion
                        log.error("Unable to submit asset lineage event for {} -> {}", inputAssets, outputIdentifiers, e);
```

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -1432,72 +1432,122 @@ public void addWorkerTaskResult(ExecutorContext executor, Supplier<FlowWithSourc
...
-                } catch (QueueException e) {
-                    log.warn("Unable to submit asset lineage event for {} -> {}", inputAssets, outputIdentifiers, e);
-                }
+                        // escalate this taskRun's own state per assetFailureBehavior, mirroring WorkerTaskProcessor's render/emit-time behavior
+                        Task task = assetUpsertFailures.isEmpty() ? null : flow.get().findTaskByTaskIdOrNull(taskRun.getTaskId());
+                        if (task != null) {
+                            State.Type currentState = taskRun.getState().getCurrent();
+                            State.Type escalatedState = task.getAssetFailureBehavior().apply(currentState);
+
+                            // drop the last attempt so LogEntry.of(TaskRun,...) tags the log with the real last attempt, not the next one
```
loicmathieu: I don't understand that, we never do that anywhere

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -1432,72 +1432,122 @@ public void addWorkerTaskResult(ExecutorContext executor, Supplier<FlowWithSourc
...
+                                contextLogger.logger().warn(
+                                    "An asset failed to be upserted but the task state was not changed (assetFailureBehavior: {}): {}",
+                                    task.getAssetFailureBehavior(), String.join("; ", assetUpsertFailures)
+                                );
+                            }
+
+                            // allowWarning has final say over a WARN-escalated state; a FAILED escalation (assetFailureBehavior: FAIL) is
+                            // never softened by allowFailure — every FAILED reachable here came from that escalation (currentState can't
+                            // already be FAILED, guarded above), and FAILED is a stronger signal than allowFailure's softening
+                            State.Type newState = escalatedState;
```
loicmathieu: You should also check allowFailure

executor/src/main/java/io/kestra/executor/ExecutorService.java
loicmathieu, line 1: Seeing the amount of asset related code, I wonder if they would be better inside the AssetService than directly here.

worker/src/main/java/io/kestra/worker/processors/WorkerTaskProcessor.java
```diff
@@ -289,7 +289,28 @@ private WorkerTaskResult runTask(WorkerTask workerTask, boolean cleanUp, Default
...
+                        "An asset failed to be emitted but the task state was not changed (assetFailureBehavior: {})",
+                        assetFailureBehavior
+                    );
+                }
+                assetEscalatedToFailed = newState != state && newState.isFailed();
+                state = newState;
+            }
+
+            // allowFailure has final say over a genuine task failure, gated on shouldBeRetried as before; a FAILED
+            // escalated by assetFailureBehavior is never softened — it's a stronger signal than allowFailure
```
loicmathieu: Why? allowFailure/allowSuccess are here for exactly that: treat failures are warning/success. It should always be honored.

loicmathieu (changes requested):
Maybe @ammeek want to have a look at the worker changes.

core/src/main/java/io/kestra/core/models/tasks/Task.java
brian-mulier-p: Could be, it's just that it impacts the entire task behavior in the end but why not it makes sense

executor/src/main/java/io/kestra/executor/ExecutorService.java
brian-mulier-p: Yes that's on my refactoring todo list after rereading thoroughly

worker/src/main/java/io/kestra/worker/processors/WorkerTaskProcessor.java
brian-mulier-p: Yes idk Claude got confused, I explicitly told the reverse... Nice catch

executor/src/main/java/io/kestra/executor/ExecutorService.java
brian-mulier-p: Yes it's intended, see https://kestra-io.slack.com/archives/C052XEFFC90/p1771496021292699. Anyway this was already the existing behavior, it's just some refactoring leading to this being shown as a full diff'

executor/src/main/java/io/kestra/executor/ExecutorService.java
brian-mulier-p: It's because I'm adding a LogEntry to an already ran and finished attempt. This is due to the post-ran nature of it. If I let the natural process, the logEntry will be tagged to attempt 2 (1 in term of index) which will not be displayed in the UI preventing any feedback to the user for it.

loicmathieu (approved):
LGTM

PR #17613, opened 2026-07-23
feat(execution): add support for allowed and denied list for HTTP tasks

loicmathieu:
HTTP tasks and functions can be exposed to SSRF as they can access any URLs from the server which include internal URLs.

Add support for configuring globally allowed and denied list of accessible URLs to protect against SSRF.

Fixes https://github.com/kestra-io/kestra-ee/issues/9037

### DIFF
--- a/core/src/main/java/io/kestra/core/http/client/HttpClient.java
+++ b/core/src/main/java/io/kestra/core/http/client/HttpClient.java
@@ -11,6 +11,7 @@ import java.security.KeyManagementException;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.time.Duration;
+import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


loicmathieu:
@aj-emerich this should be a good addition to the security guide for 2.0 (no backport planned)

nkwiatkowski (approved):

loicmathieu:
EE build fail because EE develop is ahead of the base of this branch, passed locally, merging it.

aj-emerich:
@loicmathieu will there be a second PR for `HttpFunction.java`? I didn't see it in the one attached. Or is pebble covered by `HttpClient`?

aj-emerich:
And one more question: since matching is prefix-based on the full URI string, `http://169.254.169.254` does not block `https://169.254.169.254`. Would that require two separate lists to cover?

PR #17614, opened 2026-07-23
fix(core): run migrate CLI commands in a minimal application context

### LINKED ISSUE #17612: kestra migrate CLI starts the full application before migrations run
## Problem

The `kestra migrate {run,plan,unlock,repair}` CLI commands start a **full** `ApplicationContext` before the migration logic runs.

The migration runner was a `@Context @Order(HIGHEST_PRECEDENCE)` bean, and the CLI relied on a static `MigrationRunner.setSkipAutoRun(true)` flag to stop its `@PostConstruct` auto-run. But `new MicronautFactory(applicationContext)` calls `ctx.start()` **before** picocli executes the command, so every other eager `@Context` bean initializes (and connects to the database) before the migration is applied. The `skipAutoRun` flag silences only the runner, not the rest of the context.

Setting `kestra.server-type=false` does not help either: the leaking beans are gated on `kestra.repository.type` (which migrations require), not on `server-type`.

## Expected behaviour

- Server startup keeps auto-applying migrations before any other bean touches the DB (unchanged).
- The pure migration commands should start **only** the migration runner + its dependencies + the `DataSource`, and nothing else.
- The `skipAutoRun` static hack should be removed.

## Proposed fix

- Make `MigrationRunner`/`EeMigrationRunner` plain `@Singleton` services and move the startup trigger into a dedicated `@Context @Order(HIGHEST_PRECEDENCE)` `MigrationStartupRunner` (same `@Requires(kestra.repository.type)` gating, so startup behaviour is unchanged).
- Give the migration commands a minimal `ApplicationContext` whose `resolveBeanDefinitionReferences()` drops the eager `@Context` beans gated on a runtime property (`repository.type` / `queue.type` / `server-type`) — i.e. the startup trigger and any data/server bean — while keeping DI infrastructure. The runner is then resolved lazily and invoked directly.
- Remove `skipAutoRun` entirely.

nkwiatkowski:
Closes #17612

## Problem

`kestra migrate {run,plan,unlock,repair}` started a **full** `ApplicationContext` before the migration ran. `new MicronautFactory(ctx)` starts the context — eagerly initializing every `@Context` bean (which connects to the DB) — *before* picocli executes the command, so the `skipAutoRun` flag only silenced the runner, not the rest.

## Fix

- `MigrationRunner` is now a plain `@Singleton`; the startup trigger moved to a dedicated `@Context @Order(HIGHEST_PRECEDENCE)` `MigrationStartupRunner` with the **same** `@Requires(kestra.repository.type)` gating — so server startup and tests are unchanged.
- The migration commands (new shared `AbstractMigrationCommand` base) run against a minimal context whose `resolveBeanDefinitionReferences()` drops eager `@Context` beans gated on a runtime property (`repository.type` / `queue.type` / `server-type`) — the startup trigger and any data/server bean — while keeping DI infrastructure. The runner is resolved lazily and invoked directly.
- The `skipAutoRun` static is removed entirely.

## Tests

- New `MigrationApplicationContextTest` guards the minimal context (runner + `DataSource` present, startup trigger absent, migrations pending).
- `:cli:test` (63) and core/EE migration suites green.

> Companion EE PR: kestra-io/kestra-ee (references #17612).

### DIFF
--- a/cli/src/main/java/io/kestra/cli/AbstractCommand.java
+++ b/cli/src/main/java/io/kestra/cli/AbstractCommand.java
@@ -15,7 +15,6 @@ import com.google.common.collect.ImmutableMap;
 
 import io.kestra.cli.commands.servers.ServerCommandInterface;
 import io.kestra.cli.services.StartupHookInterface;
-import io.kestra.core.migration.MigrationRunner;
 import io.kestra.core.plugins.PluginManager;
 import io.kestra.core.plugins.PluginRegistry;
 import io.kestra.core.services.FlowAutoLoader;
@@ -66,12 +65,6 @@ public abstract class AbstractCommand extends BaseCommand implements Callable<In
 
     @Option(names = { "-p", "--plugins" }, description = "Path to plugins directory")
     protected Path pluginsPath = Optional.ofNullable(System.getenv("KESTRA_PLUGINS_PATH")).map(Paths::get).orElse(null);
-
-    @SuppressWarnings("unused")
-    public static Map<String, Object> propertiesOverrides() {
-        MigrationRunner.setSkipAutoRun(true);
-        return Map.of();
-    }
 
     @Override
     public Integer call() throws Exception {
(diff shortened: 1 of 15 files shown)


loicmathieu (approved):
LGTM

PR #17617, opened 2026-07-23
feat(core): add ProtectedZipInputStream to guard against ZIP-bomb attacks

loicmathieu:
## Summary
- Adds a reusable `ProtectedZipInputStream extends ZipInputStream` (`core/.../security/`) that enforces a max entry count and a max per-entry uncompressed size, checked *during* decompression rather than after an entry is fully buffered — replacing the duplicated, buggy inline guard that had to be copied at every ZIP-read call site.
- Wires it into the two places that decompress a **user-uploaded** ZIP archive: flow import (`HasSource.readSourceFile`) and namespace file upload (`NamespaceFileController`). Two other ZIP-read sites (`WorkingDirectory` cache, `WorkerTaskProcessor` cache) decompress Kestra's own trusted internal archives, not user input, and were left untouched by design.
- New opt-in config `kestra.security.zip-bomb-protection.{enabled,max-number-of-entries,max-entry-size}` (disabled by default). A violation throws an unchecked `ZipBombDetectedException extends IllegalArgumentException`, which rides the existing `IllegalArgumentExceptionHandler` to an HTTP 422.
- A companion EE PR (kestra-io/kestra-ee#9488) wires the same protection into `BackupService.restore` for user-restorable backups. That PR originally needed a `ProtectedZipFile` (random-access `ZipFile` sibling) since restore looked up entries by name; it was later refactored to a single forward `ZipInputStream` pass instead (backup archives always write their header entry first and resources in a fixed order), so `ProtectedZipFile` was added and then removed in this same PR — only `ProtectedZipInputStream` remains, with a single protection mechanism shared by both consumers.

## Test plan
- [x] New unit tests: `ProtectedZipInputStreamTest` (6 cases — within limits, exactly-at-boundary, too-many-entries, entry-too-large, disabled/null config passthrough)
- [x] `HasSourceTest` updated for the new signature
- [x] New end-to-end webserver tests: `FlowControllerZipBombTest`, `NamespaceFileControllerZipBombTest` (assert HTTP 422 on a bomb upload)
- [x] Existing `FlowControllerTest` (63 tests) and `NamespaceFileControllerTest` (23 tests) re-run with zero regressions
- [x] Code-reviewed (approved, no blocking findings) across two rounds

### DIFF
--- a/core/src/main/java/io/kestra/core/models/HasSource.java
+++ b/core/src/main/java/io/kestra/core/models/HasSource.java
@@ -10,6 +10,8 @@ import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 import java.util.zip.ZipOutputStream;
 
+import io.kestra.core.security.ProtectedZipInputStream;
+import io.kestra.core.security.SecurityConfiguration.ZipBombProtectionConfiguration;
 import io.micronaut.http.multipart.CompletedFileUpload;
 
 /**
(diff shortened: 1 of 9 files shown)


nkwiatkowski (approved):

loicmathieu:
@aj-emerich another thing worth mentionning inside the Kestra security guide: a user can enable zip bomb protection and configure limits on the number of entries and the max entry size. This is disable by default.
It will apply to all ZIP upload (flow, namespace file, backup)

aj-emerich:
@loicmathieu nice, thanks for flagging. Backport 1.x?

PR #16900, opened 2026-06-18
feat(executions): align bulk restart revision picker with bulk replay (#8337)

pathak-satyam:
## Why

Kestra lets you re-run failed executions two ways — **Replay** and **Restart** — each usable on a single execution or in bulk, against the *original* or *latest* flow revision. Three of those four paths already let you pick the latest revision. **Bulk Restart was the exception**: it fired immediately on the original revision with no dialog and no way to choose the latest, unlike every other re-run path. That inconsistency is what #8337 reported — users doing a bulk Restart silently got stale flow logic. This PR closes that gap.

## Acceptance criteria

- [x] Bulk Restart endpoints accept an optional flag to use the latest flow revision
- [x] Bulk Restart in the UI opens a dialog offering original vs. latest, matching bulk Replay
- [x] Tests added for the backend flag (`ExecutionControllerRunnerTest`) and the frontend store (`executionsBulkRestart.spec.ts`, 3 cases)
- [x] All tests passing — `npm run test:unit` ✅, ESLint ✅, translations check ✅
- [x] Follows project conventions (Conventional Commits, `Ks*` components, i18n keys, OpenAPI annotations)
- [x] No breaking changes — default `latestRevision=false` preserves original behavior

## What this does

Closes #8337.

Aligns the **bulk Restart** action with **bulk Replay**: bulk Restart now opens a confirmation dialog letting the user pick the **original** or **latest** flow revision — the same choice bulk Replay already offers. Previously bulk Restart fired immediately with no dialog and always used the original revision.

## Changes

**Frontend**
- `Executions.vue` — the Restart bulk action opens a confirmation dialog (`Cancel` / `Restart latest revision` / `OK`), mirroring the Replay dialog. `restartExecutions(latestRevision)` routes through `genericConfirmCallback` with the revision flag.
- `stores/executions.ts` — `bulkRestartExecution` now forwards options as query params (the `latestRevision` flag was previously dropped on `/restart/by-ids`).

**Backend**
- `ExecutionController.java` — `/restart/by-ids` and `/restart/by-query` accept an optional `latestRevision` query param. When `true`, each execution is restarted against its flow's latest revision, mirroring the replay implementation. Defaults to `false`, so existing behavior is unchanged.

**i18n** — reuses existing keys (`bulk restart`, `restart latest revision`, `executions restarted`); no new keys.

## Testing

[...]

PR #17622, opened 2026-07-24
feat(system): standardize the 404 contract with edition + route-matched headers

elevatebart:

Fixes `client-sdk`&#39;s EE-only-route 404 heuristic misreporting real not-founds as &#34;Enterprise feature not available&#34; (see companion PR in `client-sdk`).

- New `NotFoundHeadersFilter`: sets `X-Kestra-Edition: OSS|EE` and `X-Kestra-Route-Matched: true|false` on every 404 response (mirrors `SecurityHeadersFilter`&#39;s `ServerFilterPhase.FIRST` pattern so it covers 404s that bypass `ErrorController` entirely). `X-Kestra-Route-Matched` is derived from the router&#39;s own match result, so it needs no opt-in from any 404-producing call site.
  - The match result is captured in a `@RequestFilter` rather than read in the response filter. When an `@Error` handler takes over, Micronaut overwrites the `ROUTE_MATCH` request attribute with the error route (`RequestLifecycle#handleErrorRoute` → `RouteExecutor#setRouteAttributes`), and `ErrorRouteMatch` is not a `UriMatchInfo` — so reading it on the way out reports &#34;no route matched&#34; for every 404 raised as an exception, which is Kestra&#39;s dominant not-found path. The request filter is not pre-matching, so it runs after the router matched and before any error route can replace the attribute.
- `ui/packages/hey-api-plugin`: the shared error interceptor skips the EE-only-route heuristic whenever `X-Kestra-Route-Matched: true`; a matched EE-only route on a server whose edition is confirmed EE now throws a new `SdkVersionMismatchError` instead of `EnterpriseFeatureError`. Falls back to today&#39;s behavior when the header is absent (older servers).
- New `NotFoundHeadersCorsCustomizer`: appends both header names to the `exposed-headers` of every CORS configuration a deployment defines. Browsers hide any response header that isn&#39;t listed in `Access-Control-Expose-Headers`, so without this the headers are invisible to cross-origin JavaScript and the SDK silently falls back to the old heuristic — i.e. the fix would not reach the browser `client-sdk` it targets. The customizer creates no configuration and touches no origins, credentials or methods, so it avoids what keeps a named CORS configuration out of the OSS defaults: one declared without `allowedOrigins` matches *any* origin with credentials allowed, and `CorsFilter` uses the first configuration matching the request origin, so an OSS default could shadow a deployment&#39;s own origin-restricted configuration.
- JSON response body is unchanged — purely additive headers, so old/new SDKs and servers stay compatible either direction.

Note the two headers are readable by unauthenticated callers. Both facts are already public: the edition via `/api/v1/configs` (`MiscController.Configuration.edition`), and route existence via the published OpenAPI spec.

An earlier iteration of this PR added an entity-tagging mechanism (`NotFoundException(Enum, String)` + `X-Kestra-Entity` header) as the primary signal, but it required every 404-producing call site to opt in — which nothing in OSS did, so it didn&#39;t actually fix the reported bug. Replaced with the router-derived `X-Kestra-Route-Matched` above, which needs no opt-in and is correct for every 404.

Tests: `NotFoundHeadersFilterTest` (including a case reproducing the error-route attribute overwrite, which fails against the previous implementation), `NotFoundHeadersFilterIntegrationTest` (both the status-route and the thrown-`NotFoundException` 404 paths), `NotFoundHeadersCorsCustomizerTest`, `runtime.spec.ts`.

### ⚠️ Signature change

`@kestra-io/hey-api-plugin` 0.3.0 → 0.4.0: `SdkVersionMismatchError` is a new export extending `Error`, not `EnterpriseFeatureError` — a consumer matching `e instanceof EnterpriseFeatureError` no longer catches the EE-server case. No such consumer exists in the OSS UI or `client-sdk`; the EE UI should be checked.

This branch originally bumped the plugin to 0.3.0; `develop` shipped and released 0.3.0 in the meantime (#19632), so the merge moved this bump to 0.4.0. `client-sdk`#368 pins the tarball, so the **Release hey-api-plugin** workflow has to be dispatched after this merges, before that PR can go green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CYWYDYyKSrq31Z7iu3f3BH)_


jymaire:
Nice, clean approach — purely additive headers with an unchanged JSON body means old/new servers and SDKs interoperate either direction, and the `ServerFilterPhase.FIRST` choice is well-justified in the javadoc.

Three things I'd like to see addressed:

### 1. The interceptor change has no test coverage

Commit a74e397 dropped `runtime.spec.ts` as "redundant coverage-only", but this is the only user-visible behavior in the PR, and it gained three branches:

- entity-tagged 404 → skip the EE heuristic entirely
- `X-Kestra-Edition: EE` → `SdkVersionMismatchError`
- otherwise → `EnterpriseFeatureError` (existing behavior)

Worth noting the spec would actually have run: `ui/vitest.config.unit.js` excludes only `packages/design-system/**`, so a `*.spec.ts` under `packages/hey-api-plugin/` is picked up by the unit project. Could we bring it back?

### 2. Cross-origin clients can't read either header

No `Access-Control-Expose-Headers` is set, and Micronaut's `exposedHeaders` defaults to empty — so `response.headers.get(ENTITY_HEADER)` / `get(EDITION_HEADER)` return `null` for any browser client on a different origin, and the interceptor silently falls back to today's behavior.

That's the documented dev topology (`cli/src/main/resources/application-override.yml` allows `http://localhost:5173`) and the normal case for the external `client-sdk` used from a browser — i.e. exactly the consumer this was built for. Both header names should be added to the CORS `exposed-headers` defaults.

### 3. Duplicate `ErrorControllerTest`

There's already `webserver/src/test/java/io/kestra/webserver/controllers/api/ErrorControllerTest.java` (`@MicronautTest`, same controller). Having two classes with the same simple name for the same controller makes `./gradlew :webserver:test --tests "ErrorControllerTest"` ambiguous. I'd fold the three new cases into the existing class, or rename this one to something like `ErrorControllerNotFoundTest`.

elevatebart:
Thanks for the review, addressed all three in 48c566c:

1. **Test coverage**: re-added `runtime.spec.ts` with the 3 branch cases you listed (entity-present, EE-edition, OSS/legacy fallback), plus 2 more (entity-present with mismatched edition, non-matching route). 5/5 passing.
2. **CORS**: added a new `micronaut.server.cors.configurations.web.exposed-headers` block to `cli/src/main/resources/application.yml` listing both headers. No existing CORS config there before, so this is additive — doesn't change whether CORS itself is enabled for a given deployment, just what's exposed once it is.
3. **Duplicate test class name**: renamed to `ErrorControllerNotFoundTest`.

---

jymaire:
Reviewed this — the goal is sound, the `ServerFilterPhase.FIRST` choice is right (outermost filter ⇒ response processing runs last, so short-circuited/bare 404s are covered), and the headers-absent fallback keeps old servers/SDKs working. Two things I'd rework before merge.

### 1. The entity half of the contract has no producers, so the reported bug isn't actually fixed

Nothing calls the new `NotFoundException(Enum<?>, String)` constructor. In OSS `X-Kestra-Entity` is therefore never emitted, and the other 404 idioms structurally can't carry it: `HttpResponse.notFound()` (4 sites in `webserver/src/main`), plus the `NoSuchElementException` / `FileNotFoundException` / `UnsatisfiedQueryValueRouteException` handlers in `ErrorController`.

Net effect on a real EE server: a genuine "invitation not found" on an EE-only route still has no entity header, so the interceptor now returns `SdkVersionMismatchError` — *"your Kestra SDK and server versions are out of sync"*. That's a different wrong answer, and arguably more confusing than the "talk to sales" one it replaces. Making it correct requires migrating every 404-producing call site behind every EE-only route, forever, with no compiler enforcement — and the failure mode is silent.

The server already knows the answer definitively, so I'd derive it instead of asking call sites to opt in:

```java
// in EditionHeaderFilter, or on ErrorController.notFound(HttpRequest)
boolean matched = request.getAttribute(HttpAttributes.ROUTE_MATCH).isPresent();
if (!matched) response.getHeaders().set("X-Kestra-Route-Matched", "false");
```

One producer, always accurate, zero migration, and it answers exactly the question the client is asking ("does this route exist on this server?"). `X-Kestra-Entity` then becomes optional debugging metadata rather than load-bearing. Worth noting too that edition is *already* public via the `/api/v1/configs` payload (`MiscController.Configuration.edition`), so route-matched is the only genuinely new information the server needs to provide here.

### 2. The CORS block is a latent security regression, and is ineffective for the case it targets

`cli/src/main/resources/application.yml` declares `micronaut.server.cors.configurations.web` with only `exposed-headers`. Checked against `micronaut-http-server` 4.10.20: `CorsOriginConfiguration` defaults to `allowedOrigins = ANY`, `allowCredentials = true`, `allowPrivateNetwork = true`, and `CorsFilter` streams `getConfigurations()` and takes `findFirst()` of any config matching the origin.

So for a deployment that enables CORS **and** defines its own named configuration with an origin allowlist, this shipped `web` config also matches every origin — silently bypassing their allowlist, with credentials allowed. Our docs only document `micronaut.server.cors.enabled: true`, so there's no `web` naming convention for a hardened deployment to land in. And in that same scenario the `web` exposed-headers don't apply to their matched config anyway, so the block delivers no benefit precisely where it does harm.

I'd drop it from the OSS defaults and document the `exposed-headers` requirement in the CORS docs instead.

---

Also FYI on CI: the backend failure is `:jdbc-h2:flakyTest` (unrelated flaky suite) and the Frontend job died in the reviewdog step on a GitHub 500 fetching the PR diff (infra — the frontend tests never reported). E2E and `kestra-ee/backend-tests` are red though; worth confirming those aren't the `jsonError` signature change or EE PR ordering.

elevatebart:
Thanks for the thorough second pass — addressed both in a127441:

1. **No producers for the entity contract**: replaced the entity-absence check entirely with a new `X-Kestra-Route-Matched` header, set by `EditionHeaderFilter` from `BasicHttpAttributes.getRouteMatchInfo(request).isPresent()` — the router's own match result, so it's always accurate with zero opt-in required. The SDK interceptor now skips the EE-only heuristic whenever `X-Kestra-Route-Matched: true` (a matched route ran application code, so it's always a genuine not-found), and falls back to the pre-existing `matchRoute` + edition heuristic when the header is `false` or absent (older servers). `X-Kestra-Entity`/`NotFoundException(entity, ...)` stay as-is, now purely optional debugging metadata as you suggested — no producers needed for correctness. Added two integration test cases (`shouldSetRouteMatchedHeaderToFalseOnAnUnmatchedRoute404`, `shouldSetRouteMatchedHeaderToTrueOnAGenuineNotFoundFromAMatchedRoute`) and updated the 5 `runtime.spec.ts` cases to the new header.
2. **CORS regression**: dropped the `micronaut.server.cors.configurations.web` block from `cli/src/main/resources/application.yml` entirely — replaced with a comment explaining why an OSS-default named CORS config is risky (matches any origin w/ credentials when no `allowedOrigins` is set, and `CorsFilter` takes the first matching config). Documented the `exposed-headers` requirement (now `X-Kestra-Edition` + `X-Kestra-Route-Matched`) in `ui/README.md`'s CORS dev-setup section instead, since that's the in-repo home for this guidance.

Also bumped `@kestra-io/hey-api-plugin` to 0.3.1 for the runtime behavior change.

---

elevatebart:
Looked into the `kestra-ee/backend-tests` failure on a127441. It's a pre-existing CI/infra issue, not something in this PR's diff:

The `trigger-ee` job's webhook-based EE compile check fails at Gradle plugin resolution, before any repo code is touched:
```
Plugin [id: 'io.kestra.gradle.spotless-conventions', version: '1.2.6'] was not found in any of the following sources
BUILD FAILED in 16s
```

- That plugin version (`1.2.6`) is identical between `kestra-ee`'s `develop` and this PR's companion branch (`claude/kestra-404-contract-ydrrcw`) — no version drift introduced here.
- `kestra-ee/settings.gradle` has no `pluginManagement` block, so this plugin resolves through the CI runner's own external Gradle init config (a private plugin repository) — outside both `kestra` and `kestra-ee`.
- The underlying "EE CI via Kestra webhook" mechanism is very new (its workflow run history shows only a handful of runs, all against an internal test branch, with a mixed pass/fail record even there).

Nothing to fix on the `kestra`/`kestra-ee` side for this one — looks like it needs attention on the CI runner's plugin-repository access. Flagging for a maintainer with visibility into that pipeline.

---

elevatebart:
Pushed 0128dba: dropped the entity-tagging mechanism entirely (`NotFoundException(Enum, String)` constructor/`entity()` accessor, `X-Kestra-Entity` header in `ErrorController`, and the now-obsolete `ErrorControllerNotFoundTest`). `X-Kestra-Route-Matched` fully covers the disambiguation without needing any call site to opt in, so the entity mechanism no longer served a purpose. Both `NotFoundException` and `ErrorController` are back to their pre-PR shape aside from that. Updated the PR title/description to match the final design.

The companion `kestra-ee` entity-tagging PR (#9484) has been closed for the same reason.

---

elevatebart:
FYI: `E2E - Tests / check` is failing consistently across commits, unrelated to this PR's diff — same 4 tests fail every run (`execution-bulk-actions.spec.ts` "Labels changed"/"Restart", `flow.spec.ts` "should create and execute the example Flow"/"with input"), same 1 flaky-but-passes-on-retry (`execution-bulk-actions.spec.ts` "Replay"), same 26 passed. I confirmed this exact failure signature was already present on `5a00f9f` — the commit before any of my changes — and reproduces identically on both `a127441` and `0128dba`. None of the failing specs touch 404 handling, CORS, or anything in this PR's scope (they're flow-creation and execution-bulk-action smoke tests). Flagging as a known pre-existing flake rather than something to fix here; won't keep re-flagging this specific one unless it changes.

---

jymaire:
The final design looks right to me — deriving route-matched from the router instead of per-call-site opt-in fixes the gap I flagged last round, the `ServerFilterPhase.FIRST` + `setIfAbsent` shape matches `SecurityHeadersFilter`, and the additive-headers/absent-header fallback keeps both directions compatible. One thing I'd change before merge:

### Revert the `HttpResponse` → `MutableHttpResponse` widening in `ErrorController`

The 3 static `jsonError(...)` signatures plus the `MutableHttpResponse` import were needed by the entity-header iteration that `0128dba` deleted. Nothing in the final diff mutates the response, so the change is now pure leftover — and it widens a public static API that both OSS and EE call.

The reason I'd rather not carry it: the `kestra-ee/backend-tests` / `trigger-ee` checks are red at Gradle plugin resolution, i.e. the EE compile path never ran on this branch. That's exactly the check that would catch a return-type break downstream, so the change is currently unverified against EE. Dropping it removes the risk at zero cost.

Everything else I'd raise is a nit and can be folded in or skipped:
- `EditionHeaderFilterIntegrationTest`: drop `rebuildContext = true` — the sibling test needs it for its per-test `@Property` overrides, this class has none, so it just re-boots the context four times.
- No unit test for the `setIfAbsent` branches (header already present, non-404 status); `SecurityHeadersFilter` has both a unit and an integration test.
- `ui/README.md` snippet mixes `allowedOrigins` (camel) with `exposed-headers` (kebab) in the same block.
- `hey-api-plugin` goes `0.2.0` → `0.3.1`, skipping `0.3.0` — intentional?

elevatebart:
Thanks — pushed b1df74c, which takes the main point and all four nits.

**Reverted the `HttpResponse` → `MutableHttpResponse` widening.** You're right that it was pure leftover from the entity-header iteration; nothing in the final diff mutates the response. The 3 static `jsonError(...)` signatures and the import are back, and both `ErrorController.java` and `NotFoundException.java` are now byte-identical to `develop` (`git diff develop -- <both files>` is empty), so there's no return-type change for EE to trip over — which matters given the EE compile path never ran here.

Nits:
- **`rebuildContext = true`** dropped from `EditionHeaderFilterIntegrationTest`, and its javadoc now points at the unit test the way `SecurityHeadersFilterIntegrationTest` does.
- **Unit test added** — `EditionHeaderFilterTest`, 5 cases: both headers on a 404, route-matched `true` when the router matched, the edition value coming from `EditionProvider`, already-set headers left untouched, and no header at all on a non-404. I also extracted a private `setIfAbsent` helper in the filter so it reads like `SecurityHeadersFilter`. The matched-route case sets the attribute via `RouteAttributes.setRouteMatch(request, mock(UriRouteMatch.class))` rather than `HttpAttributes.ROUTE_MATCH`, since that enum is `@Deprecated(forRemoval = true)` in 4.8+ and `UriRouteMatch extends UriMatchInfo`, so `BasicHttpAttributes.getRouteMatchInfo` picks it up.
- **`ui/README.md`** now uses `exposedHeaders`, consistent with the `allowedOrigins` block right above it.
- **The version skip was not intentional** — good catch. It was two separate bumps inside this PR (`5f4f679` did 0.2.0 → 0.3.0, `a127441` then did 0.3.1 for the reworked runtime), which is pointless for a single unreleased change, and a patch bump understates a runtime behavior change anyway. Collapsed to one `0.2.0` → `0.3.0` minor bump (`package.json` + `package-lock.json`).

One caveat on verification: this container has no JDK 25 and toolchain auto-download is blocked, so I could not run `./gradlew :webserver:test` locally. I verified `EditionHeaderFilter` + the new test by compiling them against the real Micronaut 4.10.17 artifacts in an isolated harness (only `EditionProvider` stubbed) and running the suite — 5/5 pass. CI is the real check.

---

elevatebart:
`kestra-ee/backend-tests` went red on b1df74c again, ~2 minutes after the push — same fast-fail timing and same external blocker I diagnosed on `a127441` in https://github.com/kestra-io/kestra/pull/17622#issuecomment-5104428361 (EE compile dies at Gradle plugin resolution for `io.kestra.gradle.spotless-conventions:1.2.6`, before any repo code is touched; that resolves through the CI runner's own init config, outside both `kestra` and `kestra-ee`). Nothing in b1df74c can affect it — the only main-source change reverts `ErrorController` to `develop`'s exact bytes, which *removes* this PR's one EE-visible signature change rather than adding one.

Still needs a maintainer with visibility into that runner's plugin-repository access. The remaining checks (Backend, Frontend, E2E) are still running on this SHA; I'll follow up on those.

---

elevatebart:
Correction to my comment above — I got the diagnosis wrong. Once the `trigger-ee` job logs became available, the failure on b1df74c is **not** the Gradle plugin-resolution issue from `a127441`. That part is fixed: the EE build now runs (3m38s, 49 tasks executed) and fails at compilation:

```
repository-elasticsearch-ee/.../ElasticSearchAgentThreadRepository.java:15:
  error: package io.kestra.core.ai.agent.models does not exist
repository-postgres-ee/.../PostgresAgentThreadRepository extends AbstractJdbcAgentThreadRepository
:webserver-ee:compileJava > io.micronaut.annotation.processing.PostponeToNextRoundException
```

That's base drift, not the plugin repository. `io.kestra.core.ai.agent.models` / `.repositories` and `jdbc/.../AbstractJdbcAgentThreadRepository` exist on today's `develop` but not on this branch — its last merge from `develop` was `5a00f9f` (Jul 24). The EE companion branch is built on current EE `develop`, which needs those OSS core classes, so EE can't compile against this branch.

Fixed by merging `develop` in — pushed 9825417, no conflicts. This branch had already been maintained with merges (`868a4b9`, `5a00f9f`) rather than rebases, so I kept that pattern rather than rewriting the PR history. Confirmed the PR's own content survived the merge intact, and `ErrorController` / `NotFoundException` are still byte-identical to `develop`. `develop`'s `hey-api-plugin` is at `0.2.0`, so the single `0.3.0` bump from my last commit is still the right baseline.

Waiting on the re-run to confirm the EE compile clears.

---

jymaire:
## Review

Thanks for the thorough write-up in the description — the reasoning about why the previous entity-tagging iteration didn't fix the bug, and the CORS security analysis, are both spot on.

Unfortunately I think the replacement mechanism has the same class of problem: it doesn't fire on the dominant not-found path. I verified this by running against the PR head, details below. **Not ready to merge as-is**, but the fix looks small.

---

### Blocker — `X-Kestra-Route-Matched` reports `false` for 404s raised as exceptions

`BasicHttpAttributes.getRouteMatchInfo(request)` reads the `ROUTE_MATCH` request attribute **coerced to `UriMatchInfo`**:

```java
return request.getAttribute(HttpAttributes.ROUTE_MATCH, UriMatchInfo.class);
```

When an `@Error` handler takes over — which is what `ErrorController` does for `NotFoundException` — Micronaut **overwrites** that attribute with the error route:

* `RequestLifecycle.handleErrorRoute:271` → `RouteExecutor.setRouteAttributes(request, errorRoute)`
* `RouteExecutor:343`, same overwrite
* `ErrorRouteMatch extends AbstractRouteMatch` → **not** a `UriMatchInfo`

So the conversion yields an empty `Optional` and the filter writes `false`. Response filters run *after* the error handler, so they always observe the error route, never the original URI route.

Status routes (a `null` body → 404) do **not** overwrite the attribute — which is precisely why the integration test passes.

#### Verified on the PR head

I added a probe controller and ran `./gradlew :webserver:test`:

```
404 from `throw new NotFoundException(...)`   →  routeMatched=false   ✗
404 from returning `null` (status route)      →  routeMatched=true    ✓
```

`EditionHeaderFilterTest` 5/5 and `EditionHeaderFilterIntegrationTest` 4/4 pass. The integration test hits `/api/v1/{tenant}/flows/{namespace}/{id}`, which returns `null` — so it exercises the one path that works, and the gap is invisible to the suite.

#### Why this matters rather than being an edge case

`ErrorController` maps four exception types to 404: `NotFoundException`, `NoSuchElementException`, `FileNotFoundException`, and `HttpStatusException` with a 404 status. That's the standard "resource not found" path in Kestra, and more so in EE controllers (audit logs, apps, SCIM…), which is exactly where the EE-only routes live.

Concretely, on an EE server, a genuine "this audit-log entry does not exist" on an EE-only route now yields `Route-Matched: false` + `Edition: EE` → the SDK throws `SdkVersionMismatchError` ("your SDK and server versions are out of sync"). One wrong message replaces another.

#### Suggested fix

Either works and both are small:

1. **Stamp a Kestra-owned request attribute in a `@RequestFilter`**, and read that in the response filter. `doRouteMatch` runs at pre-matching, *before* the filter chain, so `getRouteMatchInfo` is still truthful there. This keeps the "derived from the router's own match result" property you're after.
2. **Inject `Router`** and evaluate the match in the response filter (`router.findClosest(request) != null`), independent of the attribute lifecycle.

Plus a test on the thrown-`NotFoundException` path — that one case is what would have caught this.

---

### Design concern — the fix doesn't reach browser clients by default

`response.headers.get(...)` returns `null` cross-origin unless the server lists the headers in `Access-Control-Expose-Headers`, and the interceptor then silently falls back to today's buggy heuristic. Since OSS deliberately ships no CORS defaults, the headline fix ends up opt-in per deployment for exactly the browser `client-sdk` case being targeted.

The reasoning for not shipping a named CORS config is correct — a configuration with no `allowedOrigins` would match any origin with credentials, and `CorsFilter` takes the first match. But you can get the exposure without that risk: a `BeanCreatedEventListener<CorsOriginConfiguration>` that appends the two header names to whatever configurations a deployment *already* defines. It creates no configuration and adds no origin.

At minimum, it's worth stating the limitation in the description — as written it reads as fixed for browser clients.

---

### Smaller points

* **Naming** — `EditionHeaderFilter` / `addEditionHeader` set two headers, only one of which is about edition. Something like `NotFoundHeadersFilter` reads truer.
* **Downstream compatibility** — `SdkVersionMismatchError` doesn't extend `EnterpriseFeatureError`, so consumers doing `e instanceof EnterpriseFeatureError` stop catching the EE-server case. Intended, and `0.2.0 → 0.3.0` on a `0.x` package is the right signal, but worth confirming the companion `client-sdk` PR and the EE UI both handle the new type.
* **Docs inconsistency** — `application.yml` says `exposed-headers`, `ui/README.md` says `exposedHeaders`. Micronaut accepts both; pick one.
* **Duplication** — `setIfAbsent` is copied from `SecurityHeadersFilter`. Fine at this size, but a shared helper would be better if a third filter shows up.
* **Comment weight** — 8 lines in `application.yml` justifying an *absent* setting is a lot for a config file; the new `ui/README.md` section already carries the rationale.
* **Branch hygiene** — the head is a merge of `develop`, plus add-then-revert churn (`a12744123c` → `0128dba4cb`, `b1df74c39c`). `AGENTS.md` asks for rebase over merge; squash-merge makes the net diff fine, but worth rebasing first.
* **Security (conscious accept, not a fix)** — `X-Kestra-Edition` on every unauthenticated 404 is minor OSS/EE fingerprinting, and `X-Kestra-Route-Matched` is a route-existence oracle. Low risk given the OpenAPI spec is public and unauthorized requests get 401/403.

---

### What's good

* Filter phase choice is right and correctly mirrors `SecurityHeadersFilter`, so it covers 404s that bypass `ErrorController` entirely.
* `setIfAbsent` keeps controllers and other filters able to override.
* Constructor injection with `Objects.requireNonNull`, concise Javadoc, `should…When…` test naming, Given/When/Then, AssertJ, no `@Nested` — all in line with `AGENTS.md`.
* Unit/integration test split is the right shape; the frontend spec covers the fallback matrix well.
* Response body untouched, headers purely additive, graceful fallback for older servers.

---

### CI

Frontend tests, `trigger-ee` and EE backend tests are green; backend tests and E2E still running at the time of writing. Since this adds a global response filter, worth a second look once backend tests land.

elevatebart:
You're right, and thanks for tracing it to the exact line — I reproduced it independently before changing anything. Fixed in 6b39374.

### Blocker: confirmed and fixed

I verified your chain against micronaut 4.10.17 sources rather than take it on trust:

- `RequestLifecycle#handleErrorRoute:271` → `RouteExecutor.setRouteAttributes(request, errorRoute)` → `RouteAttributes.setRouteMatch(...)`, i.e. `ROUTE_MATCH` is overwritten.
- `ErrorRouteMatch extends AbstractRouteMatch implements MethodBasedRouteMatch` → `RouteMatch`, which does **not** extend `UriMatchInfo`, so the `UriMatchInfo.class` coercion in `getRouteMatchInfo` yields empty.
- `RouteExecutor#findStatusRoute` never calls `setRouteAttributes` — which is exactly why the status-route case passed and hid the gap.

I took your option 1 (`@RequestFilter` stamping a Kestra-owned attribute), with one addition: the response filter falls back to the live attribute when the stamped one is absent. That's what makes it correct for unmatched routes under either reading of the lifecycle — a route miss short-circuits at `onRouteMiss` before route filters, so either the request filter didn't run (absent → `false`) or it ran and computed `false`. Confirmed our filter is not in `preMatchingAlwaysMatchesFilterRoutes` (that list is gated on `FilterRoute#isPreMatching`, i.e. `@RequestFilter(preMatching = true)`), so it runs after `doRouteMatch`.

I went with option 1 over injecting `Router` because `TenantAliasingRooter` overrides `findClosest` only — a response-filter `findAllClosest` would bypass the tenant-alias fallback and report `false` for URIs Kestra genuinely routes, and `findClosest` itself can throw `DuplicateRouteException`. Stamping the real result avoids re-deriving it.

**On the test you asked for:** `shouldSetRouteMatchedToTrueWhenAnErrorRouteReplacedTheRouteMatch` stamps the match, then swaps in a non-`UriMatchInfo` `RouteMatch` the way the error route does. I checked it actually bites — against the previous implementation it fails (`expected "true" but was "false"`) while the other 5 pass. Integration side now covers both 404 shapes explicitly: the status-route one (`/flows/{namespace}/{id}`) and a thrown `NotFoundException` (`/plugins/{group}/pluginUi/{path}`, which throws for an unknown group).

### Design concern: implemented, not just documented

Your `BeanCreatedEventListener` idea needed one adjustment — `CorsOriginConfiguration` is a plain POJO built by `CorsOriginConverter`, not a bean, so a listener on it never fires. Hooking `HttpServerConfiguration.CorsConfiguration` (which *is* `@ConfigurationProperties("cors")`) works: `NotFoundHeadersCorsCustomizer` appends both header names to the `exposed-headers` of each entry in `getConfigurations()`. Creates no configuration, touches no origins/credentials/methods, no-ops when CORS is disabled, and skips a header a deployment already exposes (case-insensitively). So the browser path works by default now instead of being opt-in.

### Smaller points

- **Naming** — agreed: `NotFoundHeadersFilter` / `addNotFoundHeaders`.
- **`SdkVersionMismatchError` compatibility** — checked: both classes extend `Error` directly, and the only `instanceof EnterpriseFeatureError` references anywhere in the OSS UI or `client-sdk` are inside `runtime.ts` and its own spec. So nothing in either repo breaks; the EE UI is the one place I can't see from here, worth a grep on your side. I've added an `⚠️ Signature change` note to the description either way.
- **Docs inconsistency** — resolved by deletion: the manual `exposed-headers` snippet is gone now that the customizer handles it, so there's no second spelling left. The remaining `ui/README.md` text is prose.
- **Comment weight** — cut the 8 lines in `application.yml` to 2 pointing at the customizer; the rationale lives in the class javadoc and the README.
- **Security** — noted explicitly in the filter javadoc and the PR description as a conscious accept, with the reasoning you gave (edition already public via `/api/v1/configs`, route existence via the OpenAPI spec).
- **`setIfAbsent` duplication** — left as-is per your "fine at this size"; worth extracting if a third filter appears.
- **Branch hygiene** — the `develop` merge in the head is load-bearing: EE couldn't compile against this branch at all (it needed `io.kestra.core.ai.agent.*`, added to `develop` after this branch's last sync), which is what the red `kestra-ee/backend-tests` was — not the Gradle plugin-resolution problem I wrongly attributed it to earlier in this thread. Happy to rebase onto `develop` and force-push if you'd rather review a linear branch; I left it merged since squash-merge flattens it and a rewrite mid-review costs you your place. Say the word.

**Verification caveat:** no JDK 25 in my environment and toolchain download is blocked, so I could not run `./gradlew :webserver:test`. Both new unit test classes (10 cases) were compiled and run against the real micronaut 4.10.17 artifacts in an isolated harness — the only stub was `EditionProvider` — and the pre-fix/post-fix comparison above came from that same harness. The integration cases and the EE compile need CI.

---

elevatebart:
`kestra-ee/backend-tests` went red on 6b39374 — checked whether the `EditionHeaderFilter` → `NotFoundHeadersFilter` rename broke a downstream EE reference, and it didn't. The EE compile fails in a file this PR doesn't touch:

```
core-ee/src/main/java/io/kestra/ee/runners/RunContextFactory.java:99:
  error: cannot find symbol   secretOutputs.addAll(innerBuilder.secretOutputs());
core-ee/src/main/java/io/kestra/ee/runners/RunContextFactory.java:112:
  error: method does not override or implement a method from a supertype
```

Base drift again, one commit further along: `secretOutputs` landed on OSS `develop` in `629c196` ("fix(execution): mask SECRET flow outputs and trigger outputs in logs"), which is *after* the `develop` I merged in 9825417 (`d8ffbbc`). EE `develop` already carries the paired change, so it needs that OSS commit to compile.

Merged `develop` again — pushed fe2f6f1, no conflicts, PR content verified intact.

Worth flagging as a pattern rather than a one-off: every paired OSS/EE change that lands on `develop` breaks this branch's EE compile until it's re-synced, so `kestra-ee/backend-tests` will keep going red on any commit that sits behind `develop` for a while. It says nothing about this diff. If you want the branch rebased rather than carrying a third merge, now is a good moment — say so and I'll rebase and force-push.

---

elevatebart:
CI on fe2f6f1, since you asked for a second look at the backend once it landed:

- **Backend - Tests: green** — 7557 tests, 0 failed (webserver 1108 passed). This is the first run that actually exercises the new integration cases, which closes the verification gap I flagged: `shouldSetRouteMatchedHeaderToTrueOnANotFoundExceptionFromAMatchedRoute` asserts both a 404 status *and* `Route-Matched: true` against a real server on the `ErrorController` path, so it can't pass vacuously — a non-404 or a `false` header both fail it. The equivalent assertion fails against the pre-fix implementation.
- **trigger-ee: green** — EE compiles on this branch again, which confirms the two red rounds were `develop` drift and that the `EditionHeaderFilter` → `NotFoundHeadersFilter` rename broke nothing downstream in EE.
- Frontend, Build Artifacts, Publish Docker, spelling, submit-gradle: green.
- **E2E: still running.** If it comes back red with the same 4 specs (`execution-bulk-actions` labels/restart, `flow` create-and-execute/with-input) that's the pre-existing failure I documented on `5a00f9f`, before any of my changes; I won't re-flag it unless the signature changes.

The OpenAPI diff in the bot comment is unrelated to this PR — policies YAML source, quota limits and `recoverMissedSchedules`, all pre-existing drift between the EE-generated spec and the committed `client-sdk` one. This change adds only response headers, which don't appear in the spec at all.

Nothing outstanding from your review on my side. The one open question back to you is whether you want the branch rebased instead of carrying the `develop` merges.

---

ui/packages/hey-api-plugin/src/errors.ts
```diff
@@ -34,6 +34,28 @@ export class EnterpriseFeatureError extends Error {
     }
 }
 
+/**
+ * Thrown instead of {@link EnterpriseFeatureError} when a matched EE-only route 404s on a server
+ * that confirms it's EE — an SDK/server version mismatch, not a licensing issue.
+ */
+export class SdkVersionMismatchError extends Error {
```
jymaire, line 41: **`ui/packages/hey-api-plugin/README.md` needs updating alongside this new export.** (GitHub won't let me anchor a comment there — that file isn't in this PR's diff — so I'm leaving it on the class it should document.)

The README's *"Signaling Enterprise-only routes (`EnterpriseFeatureError`)"* section (L46–86 on `develop`) is the documented public contract for this behavior, and its worked example is now incomplete: against an EE server the interceptor returns `SdkVersionMismatchError`, so the documented `instanceof EnterpriseFeatureError` check silently misses precisely the case this PR adds. The 0.2.0 → 0.3.0 bump is flagged as a signature change in the PR description, but the package's own docs still describe 0.2.0 behavior.

Suggested fix — retitle the section, extend the catch example, and add a sentence on what decides between the two errors.

Heading (L46):

```diff
-### Signaling Enterprise-only routes (`EnterpriseFeatureError`)
+### Signaling Enterprise-only routes (`EnterpriseFeatureError` / `SdkVersionMismatchError`)
```

Import line (L58):

```diff
-import { createConfigureClient, EnterpriseFeatureError } from "@kestra-io/hey-api-plugin/runtime"
+import {
+    createConfigureClient,
+    EnterpriseFeatureError,
+    SdkVersionMismatchError,
+} from "@kestra-io/hey-api-plugin/runtime"
```

Catch example (L72–81):

```diff
 try {
     await listAuditLogs()
 } catch (e) {
     if (e instanceof EnterpriseFeatureError) {
         // e.feature, e.docsUrl, e.contactSalesUrl are structured data — build your own
         // "Upgrade to unlock X" UI instead of parsing e.message.
     }
+    if (e instanceof SdkVersionMismatchError) {
+        // The server confirmed it is EE (X-Kestra-Edition: EE) yet the route is missing:
+        // an SDK/server version skew, not a licensing problem. e.feature is the route's
+        // feature id. Prompt to align versions — never "upgrade to Enterprise".
+    }
 }
```

And a short paragraph after that block, since the selection rule is otherwise undiscoverable from the SDK side alone:

> Which of the two you get is decided by the `X-Kestra-Edition` and `X-Kestra-Route-Matched` headers Kestra sets on every 404 (server side: `NotFoundHeadersFilter`). A 404 carrying `X-Kestra-Route-Matched: true` came from application code running a real lookup, so it is always a genuine not-found and is never turned into either error, regardless of what `matchRoute` says. Servers predating these headers fall back to the `matchRoute`-plus-edition heuristic, so behavior against them is unchanged.

Worth stating explicitly in that paragraph that `SdkVersionMismatchError` deliberately does **not** extend `EnterpriseFeatureError`, so a consumer matching only the latter stops catching the EE-server case. For the record I checked and there is no such consumer — `EnterpriseFeatureError` appears nowhere in OSS `ui/` outside this package, and `kestra-io/kestra-ee` only mentions the name in its `.github` OpenAPI-tagging scripts — so this is a docs gap, not a migration.

jymaire (approved):

ui/packages/hey-api-plugin/src/errors.ts
elevatebart: Good catch — the package README was the one place still describing 0.2.0 behavior. Applied in 89af56a, taking your diffs as written: retitled heading, three-line import, the `SdkVersionMismatchError` catch branch, and your selection-rule paragraph verbatim (I checked it against the interceptor first — the `ROUTE_MATCHED_HEADER !== "true"` guard sits ahead of `matchRoute`, so "never turned into either error regardless of what `matchRoute` says" is exactly right).

Two small additions beyond your diff:

- Gave the non-inheritance its own paragraph with the *reason*, not just the consequence: the two errors call for opposite advice ("align your versions" vs "upgrade to Enterprise"), so conflating them would send an EE customer to sales over a version skew. That's the bit a future reader is most likely to want to undo without it.
- Fixed the section's intro sentence, which still promised the 404 becomes "a typed, actionable `EnterpriseFeatureError`" — now names both and which server each corresponds to. Retitling the heading alone would have left that line contradicting the example below it.

Thanks also for checking the consumer side in `kestra-ee` — that closes the one gap I couldn't see from here, and it matches what I found in OSS `ui/` and `client-sdk` (no `instanceof EnterpriseFeatureError` outside this package and its spec). So docs gap, not a migration, as you say.

---

elevatebart:
`Frontend - Tests` went red on 89af56a, which was a docs-only commit — so worth saying what it actually was. The UI tests themselves passed ("No UI test failures — nothing to report"); the job died in `check:types`:

```
src/stores/executions.ts(144,82): error TS2353: Object literal may only specify known properties,
and 'latestRevision' does not exist in type 'Omit<{ tenant: string; body: string[]; }, "tenant"> & ...'
```

`ui/src/stores/executions.ts` isn't in this PR's diff. The line comes from `4db903d` on `develop` ("align bulk restart revision picker with bulk replay", #16900), which changed two halves together: the UI now passes `latestRevision` to `restartExecutionsByIds`, and the Java `ExecutionController.restartExecutionsByIds` gained the matching `@QueryValue`. My branch had neither, and the generated SDK isn't committed — it's produced from the spec built off this branch's Java — so the type-check saw develop's newer UI against a spec generated without that parameter.

Merged `develop` again (448dd85). Verified post-merge that both halves are now present: `ExecutionController.restartExecutionsByIds` declares `latestRevision` and the store passes it, so the regenerated spec and the UI agree.

That's the third `develop` re-sync in this PR, all from paired OSS/UI or OSS/EE changes landing while the branch sat behind — none from this diff. Heads-up that it reset your approval; nothing in the code changed, only the merge. Since the branch keeps needing this, the rebase offer stands and I'd now lean toward it if you want to stop the merge commits accumulating.

---

elevatebart:
Correcting myself: the `develop` merge in 448dd85 did **not** fix `check:types`, and my explanation for it was wrong. Same error, byte for byte, on the merge commit. The real cause is on `develop`, and nothing on this branch can fix it.

**What I got wrong:** I said the generated SDK is produced in CI from the spec built off this branch's Java. It isn't. `ui/packages/kestra-sdk/openapi-ts.config.ts` says so explicitly — *"The generated output under ./src/openapi is COMMITTED to git — it is not regenerated in the fast (npm) CI. Regeneration is an explicit, Gradle-bound step."* The npm CI only runs `tsdown` over the committed `src/openapi`, so merging `develop` could never have changed the generated types. That merge was unnecessary; it cost you an approval for nothing, and I'm sorry for the churn.

**The actual state of `develop`:**

| | `develop` tip | |
|---|---|---|
| `ui/src/stores/executions.ts:144` | passes `latestRevision` to `restartExecutionsByIds` | from `4db903d` |
| `ui/packages/kestra-sdk/src/openapi/types.gen.ts:4208` | `RestartExecutionsByIdsData` → `query?: never` | no `latestRevision` |

`latestRevision` *is* in the committed SDK at lines 4149 and 4185 — the two **replay** operations, which had it before. Restart never got it, because `4db903d` ("align bulk restart revision picker with bulk replay", #16900) landed today at 05:17 and is 5 commits from tip, while the most recent `ci: auto-update generated kestra-sdk` (`306988f`, #17677) is 16 commits back — i.e. the regeneration for that change hasn't landed yet.

So `develop` is internally inconsistent right now and `cd ui && npm run check:types` fails on it directly, with no PR involved. Every PR branched from or merged with current `develop` fails at that same line — this isn't specific to #17622.

**Fix, and why not here:** regenerate the committed SDK (`./gradlew :webserver:generateOpenapiSpec` then `npm run generate:sdk` from `ui/`), which is what the auto-update workflow does. I can't produce that in this environment — the spec generation needs a JDK 25 toolchain I don't have — and hand-editing `types.gen.ts` would be wrong (generated, hash-stamped, and overwritten on the next run). It should self-heal on the next `ci: auto-update generated kestra-sdk`; worth a ping to whoever owns #16900 if that doesn't fire soon, since it's blocking the whole queue.

Everything in this PR's own diff is unaffected: backend, EE compile and the frontend **tests** are green ("No UI test failures — nothing to report" in the same job that fails at the type check).

---

elevatebart:
Both red checks on 448dd85 trace to the same `develop` commit, and neither is fixable from this branch. `4db903d` ("align bulk restart revision picker with bulk replay", #16900) changed the OSS `ExecutionController` signatures without its two paired counterparts landing:

```java
// OSS webserver/.../api/ExecutionController.java:1137 (and :1201 for byQuery)
public MutableHttpResponse<ApiAsyncOperationResponse> restartExecutionsByIds(
    @Body List<String> executionsId,
    @Nullable @QueryValue(defaultValue = "false") Boolean latestRevision) throws Exception {
```

**1. `trigger-ee` / `kestra-ee/backend-tests`** — the EE subclass still overrides the old arity:

```
kestra-ee/webserver-ee/.../api/ExecutionController.java:457:
  error: method restartExecutionsByIds in class ExecutionController cannot be applied to given types;
    return super.restartExecutionsByIds(executionsId);
:466: error: method restartExecutionsByQuery in class ExecutionController cannot be applied to given types;
:448, :460: error: method does not override or implement a method from a supertype  (@Override)
```

**2. `Frontend - Tests`** — the committed generated SDK doesn't carry the new query param, so develop's own store fails `check:types` (detailed in my previous comment).

Two things worth stating plainly, since I've been wrong twice in this thread and don't want to add a third guess:

- **This is not drift on my side.** I re-checked: `git log HEAD..origin/develop` is empty — the branch is exactly current with `develop`. So unlike the two earlier rounds, being behind is ruled out.
- **The rename and the CORS customizer are cleared.** `trigger-ee` passed on fe2f6f1, which already contained both (6b39374). The only changes since are a docs-only commit and this `develop` merge, so the EE break arrived with `develop`, not with this PR.

Reverting the merge wouldn't help — it would just put the branch behind again, which is what broke EE in the two earlier rounds. What's needed is on `develop`: the EE-side override update for the new signature, and a `ci: auto-update generated kestra-sdk` run for the query param. Both belong with #16900; I'd suggest looping in its author, since every PR on current `develop` hits both.

This PR's own diff remains green where it can be: backend suite passed on fe2f6f1 (7557 tests, including the new integration cases), the frontend **tests** pass in the same job that fails at the type check, and EE compiled clean on fe2f6f1 with the full change set.

---

PR #17627, opened 2026-07-24
fix(executions): expose saved-filter injection key via tracking util

flcarre:
## What

Re-export `SAVED_FILTER_ANALYTICS_INJECTION_KEY` from `ui/src/utils/savedFilterTracking.ts`, and import it (together with `trackSavedFilter`) from that util in `App.vue` instead of directly from the `@kestra-io/design-system` barrel.

## Why

kestra-ee `develop` CI has been red since kestra-ee#9089 merged. That PR added a direct barrel import to the EE app shell:

```ts
// ui-ee/src/App.vue
import {SAVED_FILTER_ANALYTICS_INJECTION_KEY} from "@kestra-io/design-system"
```

EE's `ui-ee` type-check runs `vue-tsc --build tsconfig.app.json` with `preserveSymlinks: true` and no project-reference redirect for the design-system. A direct **value** import from the design-system barrel in `App.vue` pulls the entire design-system *source* into the ui-ee program, which TS then type-checks under ui-ee's config and fails to resolve the design-system's own transitive dependencies:

```
Cannot find module 'motion-v' | 'mermaid' | 'monaco-editor/...' | 'moment-timezone' | 'remark-*'
```

Importing the injection key through this OSS util — a referenced project consumed via emitted declarations — avoids dragging the design-system source into the EE check. Verified locally: the design-system source errors disappear and the key resolves.

`#17130` (which introduced `SAVED_FILTER_ANALYTICS_INJECTION_KEY`) is already merged; this is the follow-up that lets the EE app shell consume the key safely.

## Companion PR

Must merge **before** the EE side: kestra-ee `App.vue` will import both symbols from `kestra/src/utils/savedFilterTracking`.

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -21,7 +21,6 @@
 
     import {ref, computed, watch, onMounted, provide} from "vue"
     import {useRoute} from "vue-router"
-    import {SAVED_FILTER_ANALYTICS_INJECTION_KEY} from "@kestra-io/design-system"
     import {useApiStore} from "./stores/api"
     import {useLayoutStore} from "./stores/layout"
     import {useCoreStore} from "./stores/core"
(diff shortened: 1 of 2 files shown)


elevatebart (approved):
Safe enough

PR #17630, opened 2026-07-24
fix(core): use a mutable map in RunVariables.computeTasksMap

### LINKED ISSUE #17629: RunVariables.computeTasksMap crashes with UnsupportedOperationException (immutable map) — fatal on SubflowExecutionEnd
## Describe the bug

`RunVariables.DefaultBuilder.computeTasksMap()` builds the `tasks` run-variable by storing an **immutable** map (`Map.of("state", …)`) for a valueless task run, then later calling `.put(...)` on that same map when a *valued* task run for the **same task id** is encountered. The `.put()` throws `java.lang.UnsupportedOperationException`.

When this happens while processing a `SubflowExecutionEnd` message, the executor treats it as a **fatal error and shuts the application down**, causing a crash loop (the poisoned queue message is redelivered on every restart).

```
ERROR [SubflowExecutionEnd] fatal error while consuming messages. Initiating application shutdown.
java.lang.UnsupportedOperationException
	at java.base/java.util.ImmutableCollections.uoe(Unknown Source)
	at java.base/java.util.ImmutableCollections$AbstractImmutableMap.put(Unknown Source)
	at io.kestra.core.runners.RunVariables$DefaultBuilder.lambda$computeTasksMap$0(RunVariables.java:650)
	at io.kestra.core.runners.RunVariables$DefaultBuilder.computeTasksMap(RunVariables.java:641)
	at io.kestra.core.runners.RunVariables$DefaultBuilder.build(RunVariables.java:412)
	at io.kestra.core.runners.RunContextFactory.of(RunContextFactory.java:166)
	at io.kestra.executor.handler.SubflowExecutionEndMessageHandler.lambda$handle$0(SubflowExecutionEndMessageHandler.java:74)
	...
```

## Root cause

`core/src/main/java/io/kestra/core/runners/RunVariables.java`, `computeTasksMap()`:

```java
if (taskRun.getValue() == null) {
    tasksMap.put(taskRun.getTaskId(), Map.of("state", taskRun.getState().getCurrent())); // immutable
} else {
    if (tasksMap.containsKey(taskRun.getTaskId())) {
        Map<String, Object> taskRunMap = (Map<String, Object>) tasksMap.get(taskRun.getTaskId());
        taskRunMap.put(taskRun.getValue(), Map.of("state", ...)); // 💥 .put() on the immutable map above
        ...
```

The failure is **ordering-dependent**: it only triggers when, for the same task id, the *valueless* task run is processed **before** the *valued* one (the valueless branch stores `Map.of`; the valued branch then mutates it). If the valued run comes first, the `else` branch allocates a mutable `HashMap` and everything works — which is why it was latent for so long.

## Why it surfaces now

The bug was introduced in **#5386** (`feat(core): add new prop tasks …`, commit `705b9af`, released in **v0.20.0**, Oct 2024), but was unreachable in practice.

It became reachable via **"feat(flow): Loop should use the execution context of its parent"** (commit `1c759e7`). Since that change, `RunVariables.build()` computes the task list as:

```java
var realExecution = execution.getLoopRun() != null ? execution.getLoopRun().parent() : execution;
var taskRunList = ListUtils.concat(execution.getTaskRunList(), realExecution.getTaskRunList());
```

For a Loop, `realExecution` is the **parent** execution, so the parent's *valueless* loop-task run is concatenated with the child's *valued* iteration runs (same task id) — producing exactly the mixed valueless-then-valued ordering that detonates the immutable map, notably on the `SubflowExecutionEnd` path.

## To Reproduce

Unit-level reproducer (fails before the fix with `UnsupportedOperationException`):

```java
TaskRun valueless = TaskRun.builder()
    .id(IdUtils.create()).taskId("hello").executionId("exec-id")
    .namespace("ns").flowId("flow").state(new State())
    .build();
TaskRun valued = TaskRun.builder()
    .id(IdUtils.create()).taskId("hello").executionId("exec-id")
    .namespace("ns").flowId("flow").value("item-1").state(new State())
    .build();

Execution execution = Execution.builder()
    .id("exec-id").namespace("ns").flowId("flow").state(new State())
    .taskRunList(List.of(valueless, valued)) // valueless BEFORE valued, same task id
    .build();

new RunVariables.DefaultBuilder()
    .withExecution(execution)
    .build(new RunContextLogger(), PropertyContext.create(renderer)); // throws UnsupportedOperationException
```

Functionally this is hit by a Loop/subflow whose parent context is merged in, on the `SubflowExecutionEnd` handler.

## Expected behavior

The `tasks` variable should be built without crashing regardless of task-run ordering; the valueless state and per-value states should coexist under the same task id. A crash here must never take the whole executor down.

## Proposed fix

Use a mutable map in the valueless branch so the later per-value merge can `put()` into it (matches the behavior the value-first ordering already produces):

```java
if (taskRun.getValue() == null) {
    Map<String, Object> stateMap = new HashMap<>();
    stateMap.put("state", taskRun.getState().getCurrent());
    tasksMap.put(taskRun.getTaskId(), stateMap);
} else {
    ...
```

## Environment

- Introduced: v0.20.0 (#5386); made reachable by the "Loop uses parent execution context" change.
- Observed on a `postgres` JDBC queue, standalone server, while processing a `SubflowExecutionEnd` message.

nkwiatkowski:
## What

`RunVariables.computeTasksMap()` stored an **immutable** `Map.of("state", …)` for a valueless task run, then `put()` into that same map when a *valued* task run for the **same task id** followed — throwing `UnsupportedOperationException`. On the `SubflowExecutionEnd` path this is treated as fatal and **crash-loops the executor** (the poisoned queue message is redelivered on every restart).

This PR stores a mutable map in the valueless branch so the per-value merge succeeds regardless of task-run ordering.

## Root cause / why it surfaces now

- Latent since **#5386** (v0.20.0): the immutable-map pattern was harmless because the value-first ordering allocates a `HashMap`.
- Made reachable by *"Loop should use the execution context of its parent"* — `build()` now `concat`s the parent's valueless loop-task run with the child's valued iteration runs (same id), producing the valueless-then-valued ordering that detonates the immutable map.

## Test

Added `RunVariablesTest.shouldBuildTasksMapWhenSameTaskIdHasValuelessAndValuedTaskRuns` — fails on `UnsupportedOperationException` before the fix, green after.

Fixes #17629

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/RunVariables.java
+++ b/core/src/main/java/io/kestra/core/runners/RunVariables.java
@@ -642,7 +642,9 @@ public final class RunVariables {
             {
                 if (taskRun.getState() != null) {
                     if (taskRun.getValue() == null) {
-                        tasksMap.put(taskRun.getTaskId(), Map.of("state", taskRun.getState().getCurrent()));
+                        Map<String, Object> stateMap = new HashMap<>();
+                        stateMap.put("state", taskRun.getState().getCurrent());
+                        tasksMap.put(taskRun.getTaskId(), stateMap);
                     } else {
                         if (tasksMap.containsKey(taskRun.getTaskId())) {
                             @SuppressWarnings("unchecked")
--- a/core/src/main/java/io/kestra/core/runners/RunVariables.java
+++ b/core/src/main/java/io/kestra/core/runners/RunVariables.java
@@ -642,7 +642,7 @@ public final class RunVariables {
             {
                 if (taskRun.getState() != null) {
                     if (taskRun.getValue() == null) {
-                        Map<String, Object> stateMap = new HashMap<>();
+                        Map<String, Object> stateMap = HashMap.newHashMap(2);
                         stateMap.put("state", taskRun.getState().getCurrent());
                         tasksMap.put(taskRun.getTaskId(), stateMap);
                     } else {
--- a/core/src/test/java/io/kestra/core/runners/RunVariablesTest.java
+++ b/core/src/test/java/io/kestra/core/runners/RunVariablesTest.java
@@ -336,6 +336,39 @@ class RunVariablesTest {
     }
 
     @Test
+    @SuppressWarnings("unchecked")
+    void shouldBuildTasksMapWhenSameTaskIdHasValuelessAndValuedTaskRuns() {
+        // Given — the crash order: a valueless taskRun for "hello" recorded first (stored as an
+        // immutable Map.of in computeTasksMap), then another taskRun for the SAME id carrying a
+        // value (as produced by a Loop/iteration whose parent context is concatenated in). This is
+        // the SubflowExecutionEnd queue crash: computeTasksMap must not put() into an immutable map.
+        TaskRun valueless = TaskRun.builder()
+            .id(IdUtils.create()).taskId("hello").executionId("exec-id")
+            .namespace("ns").flowId("flow").state(new State())
+            .build();
+        TaskRun valued = TaskRun.builder()
+            .id(IdUtils.create()).taskId("hello").executionId("exec-id")
+            .namespace("ns").flowId("flow").value("item-1").state(new State())
+            .build();
+
+        Execution execution = Execution.builder()
+            .id("exec-id").namespace("ns").flowId("flow").state(new State())
+            .taskRunList(List.of(valueless, valued))
+            .build();
+
+        // When
+        Map<String, Object> variables = new RunVariables.DefaultBuilder()
+            .withExecution(execution)
+            .build(new RunContextLogger(), PropertyContext.create(renderer));
+
+        // Then — the valueless state and the per-value state coexist under the same task id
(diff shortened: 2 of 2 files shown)


core/src/main/java/io/kestra/core/runners/RunVariables.java
```diff
@@ -642,7 +642,9 @@ private Map<String, Object> computeTasksMap(List<TaskRun> taskRunList) {
             {
                 if (taskRun.getState() != null) {
                     if (taskRun.getValue() == null) {
-                        tasksMap.put(taskRun.getTaskId(), Map.of("state", taskRun.getState().getCurrent()));
+                        Map<String, Object> stateMap = new HashMap<>();
```
loicmathieu:
```suggestion
                        Map<String, Object> stateMap = HashMap.newHashMap(2);
```

loicmathieu (approved):
LGTM, one minor improvements

PR #17634, opened 2026-07-24
feat(triggers): add BatchWebhook trigger for buffered events (#4064)

### LINKED ISSUE #4064: Add a core BatchWebhook trigger plugin
## New Trigger

New Core trigger `io.kestra.plugin.core.trigger.BatchWebhook` with `eventsCount` and `pollingInterval` properties allowing to trigger an execution:
- if at least X number of events specified using the `eventsCount` property have been received
- if the `pollingInterval` has passed AND at least 1 event has been received.

The goal of this new trigger is to allow reacting to 1 or more events at once. Currently, the Webhook trigger will trigger exactly one execution per webhook event. This is sometimes not desirable, e.g. in use cases where the events are ingested into BigQuery — trying to ingest each event separately would lead to rate limits as the service is designed for Batch loading rather than inserting each transaction separately. Using this new trigger, Kestra effectively acts as a buffer, collecting events and processing only when needed (e.g., in batches of 100, or every 15 min if there is at least 1 event received within this time period)

1. One execution per 100 events:
```yaml
triggers:
  - id: webhook
    type: io.kestra.plugin.core.trigger.BatchWebhook
    key: 4wjtkzwVGBM9yKnjm3yv8r
    eventsCount: 100
    conditions: # optional
      - type: io.kestra.plugin.core.condition.ExpressionCondition
        expression: "{{ trigger.body.hello == 'world' }}"
```

2. One execution every 15 minutes if at least 1 event is received within this time period:

```yaml
triggers:
  - id: webhook
    type: io.kestra.plugin.core.trigger.BatchWebhook
    key: 4wjtkzwVGBM9yKnjm3yv8r
    pollingInterval: PT15M       
```

## UI Observability (can be added later)

We want to help users troubleshoot event processing and bring observability into which events have been received and which of them have been associated with a given execution. Ideally, we should add an `Events` tab on the Flow page.

This UI tab can present a table with the following columns:
- `eventSent` = timestamp when the event has been sent by an external system, 
- `eventReceived` = timestamp when the event has been received by Kestra
- Event **payload** aka `trigger.body`
- Execution ID if the event made it to a batch for a specific execution. If not, the user can troubleshoot why by filtering for events without an Execution ID (e.g., only 2 events received and 5 are needed to trigger an execution)

We should add a bulk option to delete multiple events when needed, e.g., if a malformed/test event was sent and shouldn't be processed = the user can delete it.

## Output

ION file with one row per event

rishi-jat:
## Summary
Adds `io.kestra.plugin.core.trigger.BatchWebhook` to buffer webhook events and start one execution when:
- `eventsCount` is reached, or
- `pollingInterval` elapsed and at least one event is buffered

At least one of `eventsCount` / `pollingInterval` is required. Output is an ION file (`trigger.uri`) with one row per event, plus `trigger.count`.

Per-event payload filtering uses `eventFilter` (not `when` — `when` is schedule-time only and does not have event body).

Closes #4064

## Notes / limitations
- Buffering is KV read-modify-write with in-process locks only (best-effort under concurrent webserver/worker or multi-replica).
- UI Events tab from the issue is out of scope for this PR.
- Unit tests cover validation, buffering, count/interval flush, remainder, ION output, and `eventFilter`.


rishi-jat:
CC @loicmathieu

rishi-jat:
Hey @MilosPaunovic, it would be great if anyone from the team could review this PR. I’m not sure if the team is busy right now or if the priority is something else -- in that case, you can ignore this. Thanks!

loicmathieu:

This cannot work on a distributed environment as it runs in-memory.
Overall, I'm not sure we will implement thisfeature at all so I'm putting your PR on-hold, we'll let you know when we decide about this feature.

PR #17640, opened 2026-07-24
fix(dashboards): documentation panel not updating on chart block sele…

### LINKED ISSUE #17579: Custom dashboard - Specific charts documentation isn't displayed
### Describe the issue

Hello.

I would like to notify that clicking or selecting in no code on a chart definition block isn't changing the documentation panel, whatever you may do the documentation displayed is always the "flow level" one.

To reproduce : select "Create a dashboard" from the main overview page -> open the documentation panel and the flow code one (if not already open) -> try selecting different charts definition -> nothing happens -> you reproduced the mentioned problem.

Image

### Environment

- Kestra Version: 1.3.26

Lalithkishore1224:
## Issue

Fixes #17579

In the dashboard editor's **No Code** tab, selecting a different chart
block in the list was supposed to update the **Documentation** panel
to show docs specific to that chart type. Instead, the panel stayed
stuck on the generic dashboard intro no matter which block you
selected.

## Root cause

`TaskEditor.vue` has a reactive watch that's meant to load the correct
documentation whenever the selected block's type changes:

```ts
watch([selectedTaskType, fullSchema], ([task]) => {
    if (task) {
        if (isPlugin.value) {
            pluginsStore.updateDocumentation(taskModel.value)
        }
    }
}, {immediate: true})
```

It was passing the raw parsed block object (`taskModel.value`) straight
into `updateDocumentation()`. That object has a `type` field (e.g.
`{ id: 'kpi_success_ratio', type: 'io.kestra.plugin.core.dashboard.chart.KPI', ... }`),
but `updateDocumentation()` expects a `cls` field:

```ts
async function updateDocumentation(pluginElement) {
    if (!pluginElement?.cls || !allTypes.value.includes(pluginElement.cls)) {
        editorPlugin.value = undefined   // <-- always hit this, cls was always undefined
        ...
        return
    }
    ...
}
```

Since `cls` was always `undefined`, this guard fired every time and
silently reset the panel back to the generic intro — regardless of
which chart block you'd actually selected.

The other two call sites in the same file (`onTaskEditorClick` and
`TaskEditor`'s own `onActivated`) already handle this correctly by
explicitly mapping `type` → `cls`. This fix applies the same mapping
to the third, broken call site.

## Fix

```diff
- pluginsStore.updateDocumentation(taskModel.value as Parameters<typeof pluginsStore.updateDocumentation>[0])
+ pluginsStore.updateDocumentation({cls: task, version: taskModel.value?.version, hash: hash.value})
```

(plus moving the `hash`/`miscStore` declarations earlier in the file,
since the fixed watch now needs `hash` at setup time)

## Testing

Verified by hand against a local build (backend run via
`./gradlew :cli:runLocal`, frontend via `vite`), with the Documentation
panel open alongside the No Code editor:

**Before the fix** — selecting different chart blocks (KPI, Table,
Markdown, TimeSeries) left the Documentation panel stuck on the
generic dashboard intro every time.

image

**After the fix** — selecting a chart block immediately updates the
Documentation panel to that chart type's specific docs, and switching
between different chart types (KPI → Table → Markdown → TimeSeries →
back to KPI) updates it correctly every time, with no extra clicks
needed.

image

I also independently verified the **Code** tab's equivalent
interaction (clicking a `type:` value directly in the raw YAML editor)
was already working correctly and is unaffected by this change — that
code path lives in a separate component (`DashboardCodeEditor.vue`)
with its own correct `{cls: type}` mapping.


Piyush-r-bhaskar:
@flcarre is looking into an issue where all multipanel are showing only Flow documentation doc. It might be related.

flcarre:
Hey @Lalithkishore1224 — thank you so much for this fix, and especially for the crystal-clear write-up of the `cls` vs `type` root cause! 🙏 Spot on.

While investigating a closely related report (the documentation panel falling back to the generic **Flow** intro across the dashboard **and** the Apps/Test Suites editors), I folded your No-Code fix into #17648 so both root causes ship together in one place — you're credited as co-author on that commit. #17648 also restores the feature-specific intro that the doc-panel redesign (#17048) dropped, and covers the EE Apps/Tests editors that reuse the same component.

This PR is effectively superseded by #17648, but your work is in there. Really appreciate the contribution — hope to see more! 🚀

Lalithkishore1224:
Thanks @flcarre! I appreciate you folding the fix in and adding me as a co-author. Glad the write-up was helpful, and I look forward to contributing more to Kestra .

Piyush-r-bhaskar:
So, can we close this one @flcarre

PR #17641, opened 2026-07-25
feat(triggers): support multipart/form-data and binary webhook bodies

### LINKED ISSUE #17619: Webhook trigger: support multipart/form-data and binary request bodies
### Feature description

 The webhook trigger reads the request body as a `String`
`core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java:168` —
 `(String) context.request().getBody().getContent()`) and JSON-parses it. The controller itself accepts any
 content type (`webserver/.../ExecutionController.java:562` — `consumes = { MediaType.ALL }`), but because
 the trigger decodes to a `String`:

 - binary bodies (`application/octet-stream`, images) are corrupted by UTF-8 decoding;
 - `multipart/form-data` file parts are inaccessible — there's no `trigger.parts` / uploaded-file view.

 (Verified on `develop`, `2.0.0-SNAPSHOT`.)

 **Use case.** We're moving an external partner's REST API onto Kestra. One endpoint receives
 `multipart/form-data` with result photos; another receives raw image bytes stored as DB BLOBs. Neither can
 be served today because the bytes don't survive.

 **Proposed.** When the request is `multipart/form-data`, expose the file parts and form fields; for other
 binary bodies, expose the raw bytes base64-encoded — base64 is the natural carrier for bytes through
 Kestra's string/JSON trigger variables:

 ```yaml
 # what the flow would see
 trigger:
   parts:                       # multipart file parts
     - name: photo              # form field name
       filename: result.jpg
       contentType: image/jpeg
       size: 20345
       content: "<base64>"      # bytes intact; decode with base64decode
   formFields:                  # non-file multipart fields
     note: ["looks good"]
   bodyBase64: "<base64>"       # for non-multipart binary bodies (opt-in)
 ```

 A request-size cap would be needed (Jetty defaults to ~1 MB).

 **Acceptance.** A `multipart/form-data` request with a binary file part reaches the flow with the bytes
 **intact** (round-trip a JPEG, compare checksums); part filename + content-type available; text/JSON
 `trigger.body` unchanged.

 **Reference implementation.** Implemented and tested — including a multipart JPEG that round-trips
 byte-identical — in <https://github.com/ulise/kestra-rest-plugin>. Happy to open a PR against the webhook.

ulise:
Closes #17619

### ✨ Description

The webhook trigger read the request body as a `String` and JSON-parsed it, so **binary bodies were corrupted by UTF-8 decoding** and **`multipart/form-data` file parts were inaccessible** — file uploads could not be served by a flow. This is the concrete gap described in #17619.

The webhook routes now take the request as it arrives and read the body themselves, once the trigger it reaches is known. What happens to it is the trigger's `fetchType`:

```yaml
triggers:
  - id: webhook
    key: 4wjtkzwVGBM9yKnjm3yv8r
    fetchType: STORE      # FETCH (default) | STORE | NONE
```

- **`FETCH`** — the default, and **byte-identical to today**: the body reaches the flow on `trigger.body`, deserialized as JSON when it can be. A body whose content type is neither text, JSON, XML, form-urlencoded, YAML nor CSV is base64-encoded, since trigger variables carry text and not bytes.
- **`STORE`** — the body is streamed into the internal storage as it is received, and the flow reaches it as a `kestra://` URI on `trigger.uri`. No part of it is ever held in memory, and none of it travels through the execution record.
- **`NONE`** — the body is read off the connection and dropped.

**File parts are always stored**, whatever the fetch type — a file has no meaningful representation inside an execution. A `multipart/form-data` call reaches the flow as `trigger.parts` (the file parts, each with `name`, `filename`, `contentType`, `size` and its `uri`) and `trigger.formFields` (the parts that are not files):

```yaml
# what the flow sees for a multipart upload
trigger:
  parts:
    - name: photo
      filename: result.jpg
      contentType: image/jpeg
      size: 20345
      uri: kestra:///io/kestra/tests/upload/executions/5cVhZ…/webhook/0/result.jpg
  formFields:
    note: ["looks good"]
```

```yaml
tasks:
  - id: measure_upload
    type: io.kestra.plugin.core.storage.Size
    uri: "{{ trigger.parts[0].uri }}"
```

Stored bytes live under the execution the call creates, so they are purged with it. Parts are numbered (`webhook/{n}/{filename}`) because two parts of one request may carry the same caller-chosen filename, and only `FilenameUtils.getName` of it is kept so a part cannot be written outside its execution's directory. Storing before the execution exists means minting its id up front: the routes do, and pass it through `WebhookContext`, as `RunContextInitializer#configureTrigger` already does for polling triggers.

A call that stores bytes but **creates no execution cleans up after itself** — a 204 from unmet conditions, a 422 from unrenderable inputs, or a veto from `onWebhookMatched`. Nothing else would ever purge those files, since the execution they are scoped to will not exist. Reading the body only once the trigger is known also means a call whose key matches no webhook now stops before a single byte of its body is read.

The two multipart routes are hidden from the generated OpenAPI spec (`@Hidden`): they serve the **same** endpoint as the routes that accept any other content type, and a second operation on the same path + method would shadow those in the spec.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/17619

### 🛠️ Backend Checklist

New tests:

- `WebhookBodyServiceTest` — each fetch type (stored body, dropped body, binary body kept intact under `FETCH`), a stored part under its execution, a filename traversing its directory, parts sharing a filename, a request with no part, a `STORE` request with no body, and the cleanup of everything stored for a call that creates no execution.
- `ExecutionControllerRunnerTest` — drives the real routes over HTTP: multipart parts stored and read back, two files of the same name kept apart, parts not kept when the trigger's conditions are not met, a binary body base64-encoded under `FETCH`, a body stored under `STORE`, `STORE` with no body leaving nothing behind, and `NONE`. The `STORE` and multipart cases send 512 random bytes containing deliberately invalid UTF-8 and assert the content round-trips **byte-identical** out of the storage.
- `MicronautHttpServiceTest` — the text-vs-binary body decision (binary kept as bytes, JSON/form-urlencoded/no-content-type decoded as string, empty body → no body).
- `WebhookTest#shouldFetchBodyAndStayValidWhenTriggerDeclaresNoFetchType` — a flow written before the property existed keeps `FETCH` and stays valid.

The existing `webhook()` test (JSON object/array/string/empty bodies) still passes unchanged over the new route binding, which is the backwards-compatibility guard that mattered most here.

### 📝 Additional Notes

**On the design.** This PR first exposed file parts and binary bodies **base64-encoded in the trigger variables**, as #17619 proposed. @loicmathieu steered it to the internal storage for file parts ([comment](https://github.com/kestra-io/kestra/pull/17641#issuecomment-5101526580)) and to an explicit `fetchType` for the body rather than content-type detection ([comment](https://github.com/kestra-io/kestra/pull/17641#issuecomment-5115801333)); both are what is described above. The description was rewritten to the final shape on 2026-07-31 — the thread reads against an older version of it.

While rewriting I also corrected a claim I had made in the original: that there is **no `base64decode` Pebble filter in the codebase**. That was wrong — Pebble 4.1.2 (`platform/build.gradle`) registers `base64decode`/`base64encode` in its `CoreExtension`, and `ExtensionCustomizer` removes only `escape`, `raw`, `replace`, `numberformat` and `l18n`. It returns `new String(decoded, charset)`, so it round-trips a text body but cannot rebuild a binary file — which is why base64 was the wrong home for an upload, but not for the reason I gave.

Five choices worth your eye, any of which I'll change:

1. **A dedicated three-value enum rather than the shared `FetchType`.** `FETCH_ONE` has no meaning for a body — there is one, never rows — so the property is `AbstractWebhookTrigger.FetchType` with `NONE`/`FETCH`/`STORE`, nested as `Loop` already nests its own. Happy to reuse `io.kestra.core.models.tasks.common.FetchType` and treat `FETCH_ONE` as a synonym of `FETCH` if you'd rather have one enum.
2. **`fetchType` governs the body only** — the file parts of a multipart call are stored whatever it says, per your point above. So `NONE` on a multipart call still stores its parts.
3. **It sits on `AbstractWebhookTrigger`, not on `Webhook`.** The route has to read it off the trigger before it touches the body, and every webhook trigger — EE's included — receives a body.
4. **It is a plain `@PluginProperty`, not a `Property<>`.** Rendering a `Property` needs a run context carrying trigger variables, and those do not exist until the body has been read. `wait` and `returnOutputs` are plain for the same reason. It defaults to `FETCH` on the builder *and* on the getter, so a flow written before the property existed keeps working and stays valid.
5. **`STORE` hides the body from `when`.** A condition can no longer read a body that was never deserialized. It's in the property description; it seemed better than storing *and* parsing.

Two notes for EE:

- `protected ExecutionController.webhook(...)` changes from the Micronaut `HttpRequest` to `io.kestra.core.http.HttpRequest` and gains a `URI storedBodyUri` parameter — a break for any subclass overriding it.
- `onWebhookMatched` still runs after the body and the parts have been stored, since it is keyed on the trigger the body had to be read for. A veto no longer leaks them, though: a call refused after storing deletes what it stored.

**No webhook-specific size cap.** With the body and the parts streamed rather than buffered, `micronaut.server.max-request-size` is the only bound needed, and the base64 inflation I measured earlier in the thread applies only to `FETCH`, which is the mode where the caller is sending a body meant to be read as a variable.

### 🤖 AI Authors

This PR was prepared with Claude Code (Opus 4.8) under my direction and review — flagging it per the template. As requested: I told my flow a UDP joke about multipart uploads, but I'm not sure it got every part of it. The cat, meanwhile, insisted on reviewing the diff `curl`ed up on the keyboard and gave it four paws up. 🐱


core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
```diff
@@ -272,17 +346,51 @@ public static class Output implements io.kestra.core.models.tasks.Output {
...
         )
         @NotNull
         private Object body;
 
+        @Schema(
+            title = "The base64-encoded body for the webhook request",
+            description = "Only set when the request body is binary, i.e. when its content type is neither text, JSON, XML, " +
+                "form-urlencoded, YAML nor CSV; base64 is how bytes travel through the text-based trigger variables."
+        )
+        private String bodyBase64;
```
loicmathieu: I don't think an additional attribute is needed, the same body can be use for base64 encoded value

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
```diff
@@ -272,17 +346,51 @@ public static class Output implements io.kestra.core.models.tasks.Output {
...
+         * A file part of a {@code multipart/form-data} webhook request.
+         *
+         * @param name        the form field name of the part
+         * @param filename    the name of the uploaded file
+         * @param contentType the content type of the part, {@code null} if the caller did not send one
+         * @param size        the size of the part content in bytes, before base64 encoding
+         * @param content     the part content, base64-encoded
+         */
+        @Schema(title = "A file part of a `multipart/form-data` webhook request")
+        public record Part(String name, String filename, String contentType, int size, String content) {
```
loicmathieu: Outputs complex attributes are usually defined as static classes, and all of their attributes should also be documented.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -564,8 +569,40 @@ public Mono<HttpResponse<?>> triggerExecutionByPostWebhook(
...
         @Parameter(description = "Optional additional path segments") @Nullable @PathVariable String path,
-        HttpRequest<String> request) throws IllegalVariableEvaluationException {
-        return this.webhook(namespace, id, key, path, request);
+        HttpRequest<byte[]> request) throws IllegalVariableEvaluationException {
+        return this.webhook(namespace, id, key, path, MicronautHttpService.from(request));
+    }
+
+    // Hidden from the API spec: this is the same endpoint as the route that takes any other content type, which
+    // the spec already describes; a second operation on the same path and method would shadow it.
+    @Hidden
```
loicmathieu, line 605: The caveats are that the OpenAPI will not show that you can pass parts.

As it's an optional attribute I wonder if we should not add the catch-all-content-type method and branch on the content type.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
```diff
@@ -592,16 +629,69 @@ public Mono<HttpResponse<?>> triggerExecutionByPutWebhook(
...
+     */
+    private Mono<HttpResponse<?>> webhookMultipart(
+        String namespace,
+        String id,
+        String key,
+        String path,
+        MultipartBody parts,
+        HttpRequest<?> request) {
+        // Each part is read on the thread it is emitted on: its buffer is released as soon as it has been emitted.
+        return (parts == null ? Flux.<CompletedPart> empty() : Flux.from(parts))
```
loicmathieu:
```suggestion
        return (parts == null ? Flux.<CompletedPart> empty() : Flux.from(parts)).publishOn(Schedulers.boundedElastic())
```

We do this when reading file inputs for execution, it's better to do it also so the virtual thread scheduler will be used.

loicmathieu (changes requested):

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
ulise: Done — the binary body is now base64-encoded on `body` and the extra attribute is gone.

One consequence worth recording: `body` is now a `String` both for a plain-text body and for a base64-encoded binary one, so a flow distinguishes them from the `content-type` header rather than from the presence of an attribute. I've stated that explicitly in the `body` schema description. Happy to revisit if you'd rather have a discriminator later.

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
ulise: Done — `Part` is now a `@Builder`/`@Getter` static class with a `@Schema` on every attribute (`name`, `filename`, `contentType`, `size`, `content`).

AGENTS.md currently says to use Java records for simple data carriers; if Output attributes should be static classes instead, a small update there might help future contributors.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
ulise: I tried a single catch-all route, but I couldn't make Micronaut bind both the raw body and multipart parts reliably on the same method. In one variant the body was claimed twice and requests failed; in the other, multipart worked but non-multipart requests arrived with a missing body. So I've kept the two routes for now.

I documented the multipart form on the catch-all operation so the OpenAPI spec shows the parts without changing the runtime routing.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
ulise: Applied, thank you — and it turned out to be more than a scheduling improvement, because it surfaced a latent bug in my own code.

`NettyCompletedFileUpload#getBytes()` already releases the part (it calls `discard()` in its `finally`), so my explicit `fileUpload.discard()` was a second release. As long as the read happened inside `onNext` on the emitting thread a second reference was still alive, so that extra release went 2→1 and was silently harmless. With `publishOn` the read no longer runs inside `onNext`, that reference is already gone, and the redundant `discard()` decrements from zero — every multipart request then failed with `Illegal state: refCnt: 0, decrement: 1`.

So the change is `publishOn(Schedulers.boundedElastic())` *plus* dropping the explicit `discard()`. Verified with the existing test and, while investigating, with three parts totalling ~1.6 MB repeated five times — all byte-identical. (`FlowInputOutput#readData` is unaffected: it only discards in the branch where it doesn't read the part.)

ulise:
A follow-up on the size-cap question I left open in the description, now that I have concrete numbers.

Kestra ships with both limits raised well above the Micronaut defaults (`cli/src/main/resources/application.yml`):

| setting | Micronaut default | Kestra |
|---|---|---|
| `micronaut.server.max-request-size` | 10 MB | **10 GB** |
| `micronaut.server.multipart.max-file-size` | 1 MB | **10 GB** |
| `micronaut.server.multipart.mixed` | `false` | **`true`** (parts above the 10 MB threshold spill to disk) |

That matters here because both webhook paths materialise the payload in full and then base64-encode it (+33%) into the trigger variables, which travel in the execution record. So `mixed: true` — which exists precisely so that a large part doesn't sit in the heap — is defeated by the `getBytes()` call, and a 100 MB upload becomes roughly 133 MB of base64 inside a single execution. Under the shipped configuration nothing bounds that below the 10 GB ceiling, which I doubt is the intent for a trigger variable.

Which suggests one of:

1. **A webhook-specific cap** — reject above some N MB with a 413. Smallest change; someone needs to pick N.
2. **The internal-storage variant** (`kestra://` URI per part) I raised in the description. This measurement is really an argument for it: it removes the base64 inflation and keeps the bytes out of the execution record entirely.
3. **Leave it** and document that webhook uploads are meant to be small.

I'd lean towards 2 for file parts, with 1 as a guard in the meantime — but it reshapes `WebhookService.newExecution`, so I'd rather have your steer before writing it.

MilosPaunovic:
If you've addressed any of the requested changes from the PR review, @ulise, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
loicmathieu: If we need a discriminator, better to add an enum that can optionally be checked than two attributes like a new `responseType` attribute. WDYT?
Anyway, I'm OK with header check as it's pretty standard in HTTP APIs.

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
loicmathieu: Yes, this didn't account for outputs, we might update the AGENTS.md or start using records here.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
loicmathieu: You should only discard explicitly on errors to force releasing resources.

loicmathieu:
@ulise you are right, a file upload must be stored inside Kestra's internal storage (you can access it easily from the runContext) and not as a base64 encoded part.

webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
loicmathieu: OK then

loicmathieu (changes requested):
Please store multipart file-upload inside Kestra's internal storage.

For the body itself, we cannot do this unless we support a `fetchType` attribute like in the HTTP trigger so the user choose what to do with the  body (default to store it inside the `body` output).

core/src/main/java/io/kestra/plugin/core/trigger/Webhook.java
ulise: Yes, let's go with the headers.

ulise:
Thanks, that settles it — pushed.

File parts are now streamed into the internal storage as they are received, and the part carries its URI instead of its content:

```yaml
tasks:
  - id: measure_upload
    type: io.kestra.plugin.core.storage.Size
    uri: "{{ trigger.parts[0].uri }}"
```

`trigger.parts[].content` is gone; `name`, `filename`, `contentType` and `size` stay. `CompletedFileUpload#getInputStream` goes straight into `StorageInterface#put`, so nothing is held in memory and nothing binary reaches the execution record. That also retires the size question above: `max-request-size` is the only bound needed, so I added no webhook-specific cap.

Two notes on the details, in case you disagree with either:

- Storing a part means knowing the execution id first, so the routes mint it up front and pass it through `WebhookContext`, as `RunContextInitializer#configureTrigger` already does for polling triggers. Files land under `.../executions/{executionId}/webhook/{n}/{filename}`, numbered because two parts may share a filename.
- A call that stores parts but creates no execution cleans up after itself (204 and 422), and the key is matched *before* the body is read, so a wrong key stores nothing.

One thing I left as it was, since your comment is about parts specifically: a binary body sent without multipart (`application/octet-stream`) is still base64-encoded onto `trigger.body`, with the same problem for a large payload. Should it be stored the same way, with `trigger.body` becoming the URI, or is `body` better left as a body? Happy to do it in this PR either way.

loicmathieu:
> One thing I left as it was, since your comment is about parts specifically: a binary body sent without multipart (application/octet-stream) is still base64-encoded onto trigger.body, with the same problem for a large payload. Should it be stored the same way, with trigger.body becoming the URI, or is body better left as a body? Happy to do it in this PR either way.

There are two ways of doing that:
1. Add a new property `FetchType fetchType` we can reuse the existing enum or create a dedicated one:
  - NONE: don't fetch the body
  - FETCH/FETCH_ONE: fetch the body into the existing outputs (base64 is a binary stream)
  - STORE: store the body into the internal storage and add a new `uri` output with the URI
 2. Detect application/octet-stream and automatically store it into the internal storage and add a new `uri` output with the URI.

I think 1 better match what we already have (ex inside the HTTP Request task) and let the user choose. It's not always easy to work with internal storage files.

ulise:
Done — option 1, pushed.

The trigger now carries a `fetchType` deciding what happens to the body:

```yaml
triggers:
  - id: webhook
    key: 4wjtkzwVGBM9yKnjm3yv8r
    fetchType: STORE      # FETCH (default) | STORE | NONE
```

- `FETCH` — the body on `trigger.body`, JSON deserialized, binary base64-encoded. The default, byte-identical to today.
- `STORE` — the body streamed into the internal storage as it is received, its URI on `trigger.uri`. Nothing of it travels through the execution, and no part of it is ever held in memory.
- `NONE` — the body read off the connection and dropped.

Reading the body only once the trigger is known meant it could no longer be bound by the route: the routes now take the request as it arrives and read it themselves through Micronaut's `ByteBody`. One welcome consequence — a call whose key matches no webhook now stops before a single byte of its body is read, as a multipart call already did.

Five assumptions I had to make; say the word on any of them and I'll change it.

1. **A dedicated three-value enum rather than the shared `FetchType`.** `FETCH_ONE` has no meaning for a body — there is one, never rows — so the property is `AbstractWebhookTrigger.FetchType` with `NONE`/`FETCH`/`STORE`, nested as `Loop` already nests its own. The property name stays `fetchType`. Happy to reuse `io.kestra.core.models.tasks.common.FetchType` and treat `FETCH_ONE` as a synonym of `FETCH` if you'd rather have one enum.

2. **It governs the body only.** The file parts of a `multipart/form-data` call are stored whatever the fetch type — a file has no meaningful representation inside an execution, which is your point from above. So `NONE` on a multipart call still stores its parts.

3. **It sits on `AbstractWebhookTrigger`, not on `Webhook`.** The route has to read it off the trigger before it touches the body, and every webhook trigger — EE's included — receives a body.

4. **It is a plain `@PluginProperty`, not a `Property<>`.** Rendering a `Property` needs a run context carrying trigger variables, and those do not exist until the body has been read. `wait` and `returnOutputs` are plain for the same reason. It defaults to `FETCH` on the builder *and* on the getter, so a flow written before the property existed keeps working and stays valid.

5. **`STORE` hides the body from `when`.** A condition can no longer read a body that was never deserialized. It's in the property description; it seemed better than storing *and* parsing.

Two notes for EE:

- `protected ExecutionController.webhook(...)` gained a `URI storedBodyUri` parameter — on top of the `HttpRequest` change already flagged in the description.
- `onWebhookMatched` still runs after the body and the parts have been stored, since it is keyed on the trigger the body had to be read for. A veto no longer leaks them, though: a call that is refused after storing now deletes what it stored, which is also the case the parts had left open.

Tests: `WebhookBodyServiceTest` covers each fetch type and the cleanup; `ExecutionControllerRunnerTest` drives the real routes for `STORE` (512 bytes of deliberately invalid UTF-8, asserted byte-identical out of the storage), for `STORE` with no body at all (no URI, no empty file left behind) and for `NONE`. The existing `webhook()` test still passes unchanged over the new route binding, which is the backwards-compatibility guard that mattered most here.

ulise:
Housekeeping, no code change: I've **rewritten the description to the final shape** of the PR. It still described the base64 design this started from, which was confusing to read against the code as it now stands — parts in the internal storage, `fetchType` on the body. The five open choices, the EE notes and the test list are all in there, updated. Earlier comments in the thread therefore read against an older description.

One correction worth calling out, @loicmathieu, since you read the original: I claimed there is **no `base64decode` Pebble filter in the codebase**. That was wrong. Pebble 4.1.2 (pinned in `platform/build.gradle`) registers `base64decode`/`base64encode` in its own `CoreExtension`, and our `ExtensionCustomizer` removes only `escape`, `raw`, `replace`, `numberformat` and `l18n` — so the filter is available to flows today.

It doesn't change where this PR landed. `Base64DecoderFilter` returns `new String(decoded, charset)` and asks for "a correctly Base64 encoded string containing an UTF-8 string", so it round-trips a text body and mangles a binary one: a flow could still not have rebuilt a JPEG from a base64 part. Your call was right, my reason for agreeing with it wasn't. Apologies for putting a wrong fact in front of you.

loicmathieu:
@ulise thanks a lot, this is all amazing work. I need to take some time to review it and analyse the change and potential implications for sub-plugins as AbstractWebhhok is subclassed in external plugins.

loicmathieu (approved):
LGTM, nice work

loicmathieu:
@Pradumnasaraf can you do a QA of this?
It provides three new functionality:
- The ability to send binary data to a webhook, it will be base64 encoded
- The ability to send the webhook body into an internal storage file (fetchType: STORE)
- The ability to send multipart/form-encoded webhook where each part, if it's a file, will be stored inside the internal storage.

@aj-emerich worth documenting for 2.0

Pradumnasaraf:
@loicmathieu https://github.com/kestra-io/kestra/issues/17619#issuecomment-5263546229

Issue #17355, opened 2026-07-10
[Docs] Sidebar search changes the selected page but not the visible content


pavangudiwada:
### Describe the issue

In the in-app Docs sidebar, clicking pages from search changes the selected value in the **Documentation menu**, but the visible documentation content does not change.

This makes search results look clickable while leaving the user on the previous documentation page.

### Reproducibility and Visuals
1. Open the Kestra in-app Docs sidebar.
2. Search for a documentation page.
3. Click a result from the search results.
4. Observe that the **Documentation menu** value changes.
5. The visible documentation content does not change to the selected page.

Video evidence:

https://github.com/user-attachments/assets/cc4cd02a-4e21-4e02-96a4-548399dacf41

### Expected Behaviour
Clicking a Docs search result should update both the selected menu value and the visible documentation content.

The menu state and page content should stay in sync.

### Environment

- Kestra Version: `1.3.27`
- Observed: `Thu, Jul 9, 2026`
- Edition: Open Source

Siddhant-Kashyap:
i would like to work on this 🙏

Siddhant-Kashyap:
i have raised the PR can someone look into it ,please

MilosPaunovic:
Closed via https://github.com/kestra-io/kestra/pull/17644.

PR #17644, opened 2026-07-26
fix(docs): load searched document in panel instead of opening new tab

Siddhant-Kashyap:

---

### ✨ Description

When you searched for something in the docs panel and clicked a result, the content below never changed — it just stayed on whatever page was already loaded. Turns out the search results were using ContextDocsLink with the full https://kestra.io/docs/... URL, which the component treated as an external link and opened in a new tab. So the panel never actually knew you clicked anything.

Fixed by normalizing the URL to a relative path before handing it to ContextDocsLink, so it navigates inside the panel like it should.

Also fixed a crash where item.icon could be undefined on some docs cards, and swapped out the manual onMounted + watch pattern in ContextChildCard with watchEffect so it re-fetches automatically when the page changes. Removed :key="ast" from MDCRenderer as well — it was causing the whole doc tree to remount twice on every navigation due to the two-stage render.

### 🔗 Related Issue

 Closes https://github.com/kestra-io/kestra/issues/17355

### 🎨 Frontend Checklist

[Screencast from 2026-07-26 15-18-52.webm](https://github.com/user-attachments/assets/b822b8ff-dc54-47f3-9b0d-bba85adbc747)

### 📝 Additional Notes

Only frontend changes. The ContextDocs.vue tweak is a nice-to-have rather than a hard requirement — the real fix is the URL normalization in ContextDocsSearch.vue.

### DIFF
--- a/ui/src/components/docs/ContextChildCard.vue
+++ b/ui/src/components/docs/ContextChildCard.vue
@@ -9,7 +9,7 @@
         >
             <div class="card h-100">
                 <div class="card-body d-flex align-items-center">
-                    <span class="card-icon">
+                    <span v-if="item.icon" class="card-icon">
                         <img
                             :src="docStore.resourceUrl(item.icon.replace(/^\/src\/contents\//, ''))"
                             :alt="item.title"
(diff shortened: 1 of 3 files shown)


Piyush-r-bhaskar (approved):
LGTM

Piyush-r-bhaskar:
Thanks for the contribution @Siddhant-Kashyap 🎉

PR #17650, opened 2026-07-27
Remove stack trace display from expression debugger

elevatebart:
### ✨ Description

Realigns the frontend with the backend security fix in #17600, which removed `stackTrace` from the `/eval` endpoint results (`ExecutionController.EvalResult`) so internal Java stack traces are no longer exposed via the API.

The generated SDK type and its only consumer still carried the now-nonexistent field:
- Removed `stackTrace` from the `ExecutionControllerEvalResult` SDK type.
- Removed the dead stack-trace handling in `ExpressionDebugger` (state, template blocks, unused `CopyToClipboard` import, orphaned `.stack` style). The error message is still shown.

### 🔗 Related

Follow-up to #17600 (fixes kestra-io/kestra-ee#9035).


loicmathieu (approved):
LGTM, thanks

PR #17656, opened 2026-07-27
refactor(executions): use partial indexes for nullable trigger_id and parent_id columns

loicmathieu:
## Summary
- `executions.trigger_id` and `executions.parent_id` are null for the majority of rows (only sub-flow/triggered executions populate them), and no query path filters on `IS NULL`/`IS NOT NULL` for these two columns.
- Adding `WHERE ... IS NOT NULL` to `idx_executions_trigger_id` and `executions_parent_id` in `2.0.01-upgrade-postgres.sql` keeps these indexes smaller and lowers insertion cost, with no impact on existing lookups.

## Test plan
- [ ] Run Postgres-backed JDBC tests to confirm the migration applies cleanly


loicmathieu:
~~This might not be a good idea for `trigger_id` due to https://github.com/kestra-io/kestra-ee/issues/4547~~

https://github.com/kestra-io/kestra-ee/issues/4547 will be done entirely in UI using the `system.from` label.

nkwiatkowski (approved):

PR #17657, opened 2026-07-27
fix(iam): stop routing EE users into the OSS basic-auth setup wizard

elevatebart:
The OSS first-run wizard is reachable in EE at `/ui/setup` and 403s on submit — it posts to `POST /api/v1/{tenant}/basicAuth`, which EE does not implement. Flags the route `ossOnly` (the filter existed, no route ever set it) and makes the pre-auth guard fail closed to login instead of to the wizard.

Pre-auth payload untouched — `/configs/login` still exposes only `isBasicAuthInitialized`, `/configs` still requires auth.

Companion EE PR: kestra-io/kestra-ee#9518
Behaviour introduced by #17539 + kestra-io/kestra-ee#9405

### DIFF
--- a/ui/src/main.ts
+++ b/ui/src/main.ts
@@ -33,10 +33,7 @@ const app = createApp(App)
 // the design system depending on the app's plugin-icon API
 app.provide(TASK_ICON_INJECTION_KEY, TaskIcon)
 
-// Fail closed: an unexpected error while probing the pre-auth endpoints tells us nothing about
-// whether this instance needs the first-run wizard, so send the user to the login page. Setup is
-// only reachable from the positive isBasicAuthInitialized === false signal in beforeResolve —
-// treating an error as "not initialized" parked EE users on the OSS wizard, which then 403s.
+// Fail closed: an error probing the pre-auth endpoints is no evidence that setup is needed.
 const handleAuthError = (to: {fullPath: string}) => {
     BasicAuth.logout()
     const fromPath = to.fullPath !== "/ui/login" ? to.fullPath : undefined
--- a/ui/src/main.ts
+++ b/ui/src/main.ts
@@ -33,13 +33,14 @@ const app = createApp(App)
 // the design system depending on the app's plugin-icon API
 app.provide(TASK_ICON_INJECTION_KEY, TaskIcon)
 
-const handleAuthError = (error: Error, to: {fullPath: string}) => {
-    if (error.message?.includes("401")) {
-        BasicAuth.logout()
-        const fromPath = to.fullPath !== "/ui/login" ? to.fullPath : undefined
-        return {name: "login", query: fromPath ? {from: fromPath} : {}}
-    }
-    return {name: "setup"}
+// Fail closed: an unexpected error while probing the pre-auth endpoints tells us nothing about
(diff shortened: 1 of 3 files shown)


elevatebart:
`trigger-ee` / `kestra-ee/backend-tests` red here is not from this branch — this diff is three UI files with no Java. The EE build fails at configuration time:

```
kestra-ee/build.gradle line 39
Plugin [id: 'io.kestra.gradle.spotless-conventions', version: '1.2.6'] was not found
  Searched in the following repositories: Gradle Central Plugin Repository
```

The internal Kestra plugin repository is missing from the search list. #17656 (unrelated, opened minutes earlier) fails identically, so `develop` is affected too. Will re-check when it resolves.

---

elevatebart:
E2E is red with 4 failures, and I do not believe they come from this branch — but flagging rather than dismissing, since this PR does touch the router guard.

```
[chromium] › tests/e2e/executions/execution-bulk-actions.spec.ts:51 › Restart only on a filtered set … Select All
[chromium] › tests/e2e/executions/execution-bulk-actions.spec.ts:98 › Replay only on a filtered set … Select All
[chromium] › tests/e2e/flow.spec.ts:34 › should create and execute the example Flow
[chromium] › tests/e2e/flow.spec.ts:61 › should create and execute a Flow with input
```

The flow ones fail on `getByRole('button', {name: 'Create', exact: true})` not being found on the flows page, through 5 retries.

Why this diff is an unlikely cause:

- The only runtime change is inside the `catch` of `beforeResolve`, which redirects to `login` instead of `setup`. It runs only when a pre-auth request throws. The happy path (`isBasicAuthInitialized === false` → setup) is untouched.
- `ossOnly: true` on the setup route is inert in OSS — nothing in this repo reads the flag; only `ui-ee/src/routes/routes.ts` filters on it.
- If the guard were misrouting, the app would be stuck on login/setup and *every* test would fail. The bulk-action tests get as far as filtering executions and selecting all, so the app boots and authenticates.

That said, I can't fully rule it out from logs alone — a re-run would settle whether these are flaky. Happy to dig into the trace artifacts if anyone has seen these two specs fail elsewhere recently.

---

loicmathieu (approved):
LGTM

elevatebart:
Upgrading my earlier "probably not this branch" on the E2E failures to something demonstrable.

`9b99351` differs from `f00d6f0` **only in comment text** — a comment-trim commit, no runtime change whatsoever. Across those two SHAs:

| SHA | E2E result |
|---|---|
| f00d6f0 | 4 failed, 1 passed |
| 9b99351 | 5 failed |

The extra failure is `execution-bulk-actions.spec.ts:11 › Labels changed only on a filtered set … Select All` — the test that passed on the previous run. Comments cannot change which tests pass, so the outcome is varying independently of the diff: these specs are flaky (or environment-sensitive) in this window, not broken by this PR.

For completeness, the other four are identical across both runs:

```
execution-bulk-actions.spec.ts:51 › Restart … Select All
execution-bulk-actions.spec.ts:98 › Replay … Select All
flow.spec.ts:34 › should create and execute the example Flow
flow.spec.ts:61 › should create and execute a Flow with input
```

All five are `Select All` bulk actions or flow creation — nothing touching the pre-auth router guard, which is the only runtime change here.

---

PR #17659, opened 2026-07-27
fix(namespaces): reject a file and a directory sharing the same path

ammeek:
✨ Description

Enforces that a namespace file path is either a file or a directory, never both.

Path lookups in the metadata index match <path> and <path>/ alike (AbstractJdbcNamespaceFileMetadataRepository.pathCondition, and the same in the EE ElasticSearch repository), so a directory entry at a file's path was silently accepted as that file's previous revision. putFile then derived the new revision from it and wrote the content to file.sql.v2, while save pinned the persisted entry to revision 1 and kept it a directory. The result was a file that 404s on open, renders as a folder after a refresh, and cannot be recreated — the UI delete is a soft delete, so the offending entry survives every cleanup and re-poisons the next create.

InternalNamespace.putFile and InternalNamespace.createDirectory now pass their path lookup through a single guard:

- a live entry of the other kind is a genuine conflict and raises ConflictException, which ErrorController maps to a 409 carrying the message, instead of silently corrupting the path;
- a soft-deleted entry of the other kind no longer owns the path, so it is purged (index entry and storage object) and the write proceeds from a clean slate.

That second branch means affected instances heal themselves on upgrade rather than needing DB surgery.

The guard sits in InternalNamespace deliberately: it is the only implementation of Namespace in OSS and EE, so the webserver, CLI, flow tasks (UploadFiles) and the git-sync plugins are all covered by one check. #16541 added equivalent guards to the file explorer, but those are client-side and read the rendered tree, so they cannot see a soft-deleted entry and do not apply to any non-UI caller.

### DIFF
--- a/core/src/main/java/io/kestra/core/storages/InternalNamespace.java
+++ b/core/src/main/java/io/kestra/core/storages/InternalNamespace.java
@@ -12,6 +12,7 @@ import java.util.function.Predicate;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 
+import io.kestra.core.exceptions.ConflictException;
 import io.kestra.core.models.namespaces.files.NamespaceFileMetadata;
 import io.kestra.core.namespace.NamespaceFileMetadataStateStore;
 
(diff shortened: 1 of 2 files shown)


Skraye (approved):
LGTM

PR #17678, opened 2026-07-28
test(tests): make retrySubflow load its subflow and use a dedicated tenant

nkwiatkowski:
## What

`AbstractRunnerRetryTest.retrySubflow` only loaded `retry-subflow.yaml` via `@ExecuteFlow`, so the subflow it launches (`failed-first`) was never in the repository.

The test still **passed on develop**, but for the wrong reason: with the subflow missing, the `Unable to find flow 'failed-first'` path also happens to produce 3 attempts and a `FAILED` state, so the assertion held **without the subflow ever running**. (On `releases/v1.0.x`/`v1.3.x` the same test bug surfaces as a hard failure — `expected: 3 but was: 1` — fixed in #17675 / #17676.)

## Fix
- Load **both** flows via `@LoadFlows` and run `retry-subflow` explicitly through `TestRunnerUtils` in `RetryCaseTest`, so the subflow is actually executed — the `Unable to find flow failed-first` log line is now gone — and the test exercises real subflow retry.
- Use a **dedicated tenant** (`retrysubflowtenant`) instead of `MAIN_TENANT`, isolating the test's flows/executions from other tests sharing the main tenant (a flakiness source), consistent with the `retryNewExecution*` tests.

## Verification
`:jdbc-h2:test --tests "...H2RunnerRetryTest.retrySubflow"` → 1 passed, and `Unable to find flow failed-first` no longer appears in the run log (subflow now resolved and executed).

### DIFF
--- a/core/src/test/java/io/kestra/core/runners/AbstractRunnerRetryTest.java
+++ b/core/src/test/java/io/kestra/core/runners/AbstractRunnerRetryTest.java
@@ -110,9 +110,9 @@ public abstract class AbstractRunnerRetryTest {
     }
 
     @Test
-    @ExecuteFlow("flows/valids/retry-subflow.yaml")
-    void retrySubflow(Execution execution) {
-        retryCaseTest.retrySubflow(execution);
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
LGTM

PR #17681, opened 2026-07-28
fix(ai): refresh the copilot empty state logo with the new 3D mark

MilosPaunovic:
The logo above the "Turn your idea into a workflow" empty state looked off — the mark sat off-center in its circle and read as lower-quality, not matching the rest of the page (kestra-io/kestra-ee#9539).

## Fix
- Swapped `copilot-illustration.png` for the designer-provided **3D Kestra mark — the AI variant (with the sparkles)**, per Alejandro on the issue. Resized the 512px source to 256px (a crisp @2x for a 128px display).
- Bumped the display size from 96px (`--ks-spacing-16`) to the **128px** design spec (`8rem` — no spacing token maps to 8rem, so a raw rem per the DS "fall back to rem, never px" rule).

The new artwork is centered within its own circle, so `.copilot-artwork` renders it dead-center and it reads on-brand.

**OSS-only**: `CopilotChat` is a shared OSS component with no EE override, so this fixes it for EE too — no companion PR.

Closes https://github.com/kestra-io/kestra-ee/issues/9539.


Pradumnasaraf (approved):
Looks good!

Image

PR #17682, opened 2026-07-28
fix(sdk generation): load full history so the generated-SDK branch can merge develop

elevatebart:
- `fetch-depth: 0` on the checkout, so develop's history is present and the merge has a merge base
- Fetch `chore/update-kestra-sdk` without `--depth=1`, so its history is present too
- Merge with `-X theirs`, so develop wins any conflict on the files regenerated in the next step
- Pass the develop commit subject to `git commit` via an env var, so a `$(...)` subject cannot run on the runner
- Pass the previous PR body to the `github-script` step via an env var, so a `${...}` subject cannot run as JS

### DIFF
--- a/.github/workflows/update-generated-sdk.yml
+++ b/.github/workflows/update-generated-sdk.yml
@@ -23,6 +23,9 @@ jobs:
           gh-app-private-key: ${{ secrets.GH_BOT_PRIVATE_KEY }}
           repository: kestra-io/kestra
           ref: develop
+          # full history: a shallow clone has no merge base with the chore branch,
+          # so the merge below fails with "refusing to merge unrelated histories"
+          fetch-depth: 0
 
       # github.event.head_commit is only populated for push events; resolve the commit
       # subject from history so it's correct regardless of what triggered this run.
@@ -59,13 +62,14 @@ jobs:
       - name: Checkout the chore/update-kestra-sdk branch
         shell: bash
         run: |
-          # the checkout above is a shallow, single-branch clone of develop, so the
-          # remote chore branch is not present locally. Fetch it explicitly first.
-          git fetch --depth=1 origin chore/update-kestra-sdk || true
+          # the checkout above is a single-branch clone of develop, so the remote chore
+          # branch is not present locally. Fetch it, with its history, explicitly first.
+          git fetch origin chore/update-kestra-sdk || true
           if git rev-parse --verify --quiet origin/chore/update-kestra-sdk; then
             # reuse it, bringing in anything new from develop since it was created
             git switch -c chore/update-kestra-sdk origin/chore/update-kestra-sdk
-            git merge origin/develop --no-edit
+            # the SDK is regenerated right below, so let develop win any conflict
+            git merge origin/develop --no-edit -X theirs
           else
             # otherwise, create it from develop
             git switch -c chore/update-kestra-sdk origin/develop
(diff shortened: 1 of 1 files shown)


elevatebart:
The two EE checks that went red on the previous SHA (`3f54d74`) are not caused by this PR, which changes a single GitHub Actions YAML file and no Java:

- **`trigger-ee`** — EE compilation failed on Gradle plugin resolution: `Plugin [id: 'io.kestra.gradle.spotless-conventions', version: '1.2.6'] was not found in any of the following sources`. That is a plugin-resolution problem in the EE build, unreachable from a workflow file.
- **`kestra-ee/backend-tests`** — the same check is failing on #17680 (SHA `1843e5f`, a different author and diff), so it is failing independently of this branch.

I have not tried to fix either, since both sit in EE build infrastructure rather than in this diff. Happy to dig into the `spotless-conventions` resolution failure separately if it is not already known.

---

.github/workflows/update-generated-sdk.yml
```diff
@@ -56,20 +56,23 @@ jobs:
...
+          # the new commit so the branch keeps its earlier regeneration commits.
+          git fetch --depth=1 origin \
+            +refs/heads/chore/update-kestra-sdk:refs/remotes/origin/chore/update-kestra-sdk || true
+          # check out develop rather than the chore branch, and never merge develop
+          # into it: both fetches are shallow, so the two histories share no merge
+          # base and `git merge origin/develop` failed with "refusing to merge
+          # unrelated histories". Generating from develop's tree also means the SDK
+          # is generated against develop's current backend sources; the "Commit and
+          # push" step below reattaches the branch history without needing a merge.
+          git switch -c chore/update-kestra-sdk origin/develop
```
elevatebart: I would prefer this be done only if the branch does not exist already and if it does, merge develop into it instead.

elevatebart: And keep all comments under 2 lines please

elevatebart: Done — reverted to the original conditional (create from develop only when the branch is absent, otherwise merge develop into it) and instead loaded the history the merge needs: `fetch-depth: 0` on the checkout, and the chore-branch fetch without `--depth=1`. The whole diff is now those two depth changes.

Verified against a fixture that mirrors the CI clone shape (single-branch clone + wildcard refspec):

- depth 1 → `fatal: refusing to merge unrelated histories` (the reported failure)
- full depth → merge succeeds, push fast-forwards with no force, the earlier regeneration commit stays in the branch, and the PR diff is still only `ui/packages/kestra-sdk/...`

One residual worth knowing about, unchanged by this PR: if develop's *generated* SDK files are ever rewritten while this branch is open, the merge conflicts on them (`CONFLICT (content) in types.gen.ts`). In practice they only change when a bot PR merges, and the branch content then matches — so I left it alone rather than adding conflict handling here.

---

elevatebart: Done — every comment in the diff is now two lines or fewer, and the long ones from the earlier revision are gone with it.

---

elevatebart: Handled — the merge is now `git merge origin/develop --no-edit -X theirs`, so develop wins any conflicting hunk. Safe here because the only files this branch ever touches are regenerated in the very next step, so develop's side gets overwritten anyway.

Re-ran the fixture:

- develop rewrote the generated SDK (previously fatal) → resolves to develop's side, then regeneration overwrites it; push fast-forwards, earlier commit kept
- develop left the SDK alone → clean merge, unchanged
- shallow clone → still `refusing to merge unrelated histories`, confirming `fetch-depth: 0` is what fixes the reported failure and `-X theirs` only covers the conflict case

---

MilosPaunovic (approved):

PR #17700, opened 2026-07-29
fix(plugins): upload working-dir files when Docker MOUNT strategy is used

### LINKED ISSUE #17526: TaskRunner Docker with fileHandlingStrategy MOUNT error
### Describe the issue

Using "fileHandlingStrategy: MOUNT" in "io.kestra.plugin.scripts.python.Script" with a dedicated docker task runner don't work.
("python: can't open file '/tmp/kestra-wd/tmp/43zhGVqpWVfcLvi6ZJRZiX/14502749081614047022.py': [Errno 2] No such file or directory").
```
- id: test
    taskRunner:
      fileHandlingStrategy: MOUNT
    script: print("hello")
```
For scripts that process large files, the time spent mounting and unmounting volumes results in significant delays. We therefore tried using "MOUNT" to access the storage point directly, but without success. Even with the simplest of scripts, we still encountered errors.

### Environment

- Kestra Version: 1.3.28

RizwanAhamed13:
### ✨ Description

`fileHandlingStrategy: MOUNT` on the Docker task runner previously relied only on a bind mount of the working directory. When the Docker daemon cannot see the worker filesystem (typical when Kestra runs in Docker with only `/var/run/docker.sock` mounted), that bind is empty, so scripts fail with:

```text
python: can't open file '/tmp/kestra-wd/tmp/.../*.py': [Errno 2] No such file or directory
```

This change uploads working-directory input files into the container before start for both `VOLUME` and `MOUNT`, and still keeps the MOUNT bind for large-file output access without a full download round-trip. Bind paths now use an absolute normalized path for consistency with `workingDir`.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/17526

### 🛠️ Backend Checklist

- [x] Unit tests added/updated for the upload decision (`DockerErrorHandlingTest`)
- [x] Integration coverage added for MOUNT with a worker-created input file (`DockerTest`)

### 📝 Additional Notes

- Extracted `uploadWorkingDirectoryFiles` / `shouldUploadWorkingDirectoryFiles` so VOLUME and MOUNT share the same upload path.
- For `MOUNT`, outputs still rely on the bind mount (no full working-directory download), which keeps the large-file benefit of MOUNT.


MilosPaunovic:
@RizwanAhamed13 Please resolve the merge conflict.

RizwanAhamed13:
Rebased this PR onto the latest develop and resolved the conflict in Docker.java.
The resolution preserves the newer setup-phase container cleanup while keeping the working-directory upload behavior for both VOLUME and MOUNT.
Focused tests pass (DockerErrorHandlingTest: 5/5). The Docker integration test compiles, but I couldn’t execute it locally because the Docker daemon was unavailable; backend and E2E CI checks are currently running.
The PR is now mergeable and ready for another review.

MilosPaunovic:
Also, we'd really appreciate if you don't use LLM to create every single sentence/answer.

RizwanAhamed13:
Fair point — sorry about that. I’ll keep future comments concise and in my own words. Thanks for the feedback.

loicmathieu (commented):
I'm not sure doing it unconditionaly is a good idea, there can be GB files inside the working directory so uploading them inside the container is not a good idea.

If mount didn't work for you, I strongly advise to use VOLUME which exactly does what you are asking for.

thomas-depierre:
@loicmathieu That is precisely why we want to avoid using a volume: transfer overhead is huge at scale (from what I’ve seen, an intermediate local TAR archive is created before pushing everything, and that alone causes losses...).

thomas-depierre:
@MilosPaunovic, could you tell us what is blocking the merge?

loicmathieu:
Decision is not to do it: the VOLUME strategy use a volume, to use a mount, the MOUNT strategy should be used.


Anyway, thanks a lot @RizwanAhamed13 for your contribution

Issue #13134, opened 2025-11-24
Race condition on WorkingDirectory task's `postExecuteTasks`


brian-mulier-p:
### Describe the issue

There is a gap in our handling of the WorkingDirectory task that is possibly causing its `outputFiles` not to be populated properly. I (almost) consistently have a test failure upon executing WorkingDirectoryTest. To reproduce remove the `@org.junit.jupiter.api.parallel.Execution(ExecutionMode.SAME_THREAD)` introduced [here](https://github.com/kestra-io/kestra/commit/2f02abec554a31fcd950e75398931107e9f883b3) (to remove flakiness) and run the whole test class locally (only running this test doesn't work as you need some load for it to occur.

The race is here:
- We execute a WorkingDirectory task with an `outputFiles: ["test.txt"]`
- We process its inner task that creates a file `test.txt`
- The worker task result SUCCESS of the inner task is sent to the Executor which moves the WorkingDirectory (flowable) to SUCCESS state
- The outputs of each taskrun is serialized to the execution
- The `postExecuteTasks`, responsible for creating the `outputs.ion` file containing the outputFiles, is executed after the WorkerTaskResult telling SUCCESS is sent
- Possibly the Execution is moved to SUCCESS with its `outputs.wdir.outputFiles` NOT populated because the `postExecuteTasks` hasn't kicked in yet and the WorkingDirectory.getOutputs() checks existence of `outputs.ion` which is not created yet...

### Environment

- Kestra Version: develop

ameyypawar:
`WorkingDirectoryTest.outputFiles()` never runs: `@FlakyTest` has no `@Test`, so it is undiscoverable. Adding it takes `:core:test` 3286 → 3287. The race is still live after #17701.

https://github.com/ameyypawar/kestra/tree/fix/13134-workingdirectory-outputfiles-race

PR #17701, opened 2026-07-29
fix(executor): join worker task results of the same execution in queue order

loicmathieu:
## Summary

- **`WorkingDirectoryTest.namespaceFilesWithNamespace()`** — a real product bug in `DefaultExecutor`: the `workerTaskResultQueue` batch subscriber processed each `WorkerTaskResult` as an independent future with no grouping by execution id, unlike the sibling `executionEventQueue` subscriber which deliberately groups by `executionId`. A later-queued sibling's result (e.g. a FAILED task) could join before an earlier one, terminating a flowable task (e.g. `WorkingDirectory`) before the earlier sibling's output was ever persisted — silently dropping it. Fixed by mirroring the sibling's grouping pattern.
- **`PostgresRunnerConcurrencyTest.flowConcurrencySlaFailSubflow()`** — a flaky test, not a product bug: the concurrency-limit counter decrement commits in a transaction *after* the terminal execution row, so a single zero-tolerance read taken right after observing all executions terminal could race a still-in-flight release. Fixed by polling with a bounded timeout instead of asserting a single read, and removed the now-unneeded `@FlakyTest` tag (verified stable across 10 repeated runs against Postgres).
- **`@FlakyTest` was silently undiscoverable** — it carried `@Tag("flaky")` but no `@Test` meta-annotation, so a method annotated with `@FlakyTest` alone was never picked up by JUnit under either the `test` or `flakyTest` Gradle tasks. Added `@Test` as a meta-annotation. This revived `WorkingDirectoryTest.outputFiles()` (still blocked by the pre-existing #13134 race, now excluded explicitly via `@Disabled` instead of invisibly) and `VariablesTest.recursiveVars()` (passes cleanly).

## Test plan

- [x] `:executor:test` — 81/81 passing
- [x] `:jdbc-h2:test --tests "*H2RunnerTest*"` — 99/99 passing (mandated by AGENTS.md for executor changes)
- [x] `:jdbc-h2:test :jdbc-mysql:test :jdbc-postgres:test --tests "*RunnerConcurrencyTest*"` — all passing
- [x] Postgres `PostgresRunnerConcurrencyTest` repeated 10x — all passing, no flakes
- [x] `:core:flakyTest :webserver:flakyTest :script:flakyTest :cli:flakyTest` — all clean, no unexpected failures from newly-discoverable tests
- [x] Independent code review (subagent) — no issues found beyond a cosmetic comment, fixed


tests/src/main/java/io/kestra/core/junit/annotations/FlakyTest.java
```diff
@@ -6,13 +6,19 @@
...
+ * Used to document that a test is flaky and needs to be reworked.
+ * <p>
+ * Meta-annotated with {@link Test} so a method carrying only {@code @FlakyTest} is still
+ * discovered by JUnit — without it, such a method is silently skipped by both the {@code test}
+ * and {@code flakyTest} Gradle tasks.
  */
 @Target({ ElementType.TYPE, ElementType.METHOD })
 @Retention(RetentionPolicy.RUNTIME)
 @Tag("flaky")
+@Test
```
nkwiatkowski: We didin't put Test annotation here because a ParametrizedTest can be flaky too, and the annotation must work on it.

nkwiatkowski (approved):

tests/src/main/java/io/kestra/core/junit/annotations/FlakyTest.java
loicmathieu: This is a valid concern, I'll remove it for now  but it hides real testing issues :(

PR #17702, opened 2026-07-29
fix(flows): prevent selecting a past schedule date when executing

### LINKED ISSUE #17348: [Executions] Scheduled execution allows selecting a date and time in the past
### Describe the issue

When running a scheduled execution, the UI allows selecting a date and time in the past. After the execution is submitted, the execution appears to start at the time the user clicked **Execute**, rather than using the past scheduled time.

This makes the scheduling control misleading because the user can choose an invalid schedule value and only later infer that the value was not actually used.

1. Open a flow.
2. Click **Execute**.
3. Choose the scheduled execution option.
4. Select a date and time in the past.
5. Submit the execution.
6. The UI accepts the past value, but the execution start time is the time the user clicked **Execute**.

Video evidence:

https://github.com/user-attachments/assets/bf005662-5cb8-43fb-ba81-91f72e89f0f7

### Expected Behaviour
The scheduler should prevent users from selecting a date and time in the past, or show a clear validation message before submission.

If past values are intentionally allowed, the UI should explain that the execution will start immediately instead of preserving the selected past time.

### Environment

- Kestra Version: `1.3.27`
- Observed: `Thu, Jul 9, 2026 2:10 PM IST`
- Edition: Open Source

RizwanAhamed13:
### ✨ Description

When scheduling an execution from the Execute dialog, the UI allowed selecting a date and time in the past. The executor ignores past `scheduleDate` values and starts immediately, so the control was misleading.

This change:
- disables past calendar days in the schedule date picker
- disables hours/minutes already elapsed when the selected day is today
- blocks submit with a clear error if a past value is still present
- only sends `scheduleDate` when one is set (instead of always sending "now")

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/17348

### 🎨 Frontend Checklist

- [x] Unit tests added for schedule-date helpers
- [x] Translations added for the validation message

### 📝 Additional Notes

- Helpers live in `ui/src/utils/scheduleDate.ts` so the picker rules are unit-tested without mounting the dialog.

### DIFF
--- a/script/src/main/java/io/kestra/plugin/scripts/runner/docker/Docker.java
+++ b/script/src/main/java/io/kestra/plugin/scripts/runner/docker/Docker.java
@@ -291,7 +291,15 @@ public class Docker extends TaskRunner<Docker.DockerTaskRunnerDetailResult> {
         description = """
             How to handle local files (input files, output files, namespace files, ...).
             By default, we create a volume and copy the file into the volume bind path.
-            Configuring it to `MOUNT` will mount the working directory instead."""
+            Configuring it to `MOUNT` will bind-mount the working directory instead, which
+            avoids copying output files back after the task finishes (useful for large files).
+
+            With `MOUNT`, input files are still uploaded into the container before start so
+            scripts keep working when the Docker daemon cannot see the worker filesystem
+            (for example Kestra running in Docker with only the Docker socket mounted).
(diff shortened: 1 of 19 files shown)


MilosPaunovic:
Hey @RizwanAhamed13, as you're neither assigned to the issue nor, apparently, you don't even look the opened PR or just don't care about many, many merge conflicts, I am just outright going to close this.

PR #17704, opened 2026-07-29
fix(ui): regenerate the SDK for the bulk restart latestRevision param

ammeek:
### ✨ Description

Regenerates the committed TS SDK so the bulk restart endpoints expose their
`latestRevision` query param, matching the backend and the UI call site.

`RestartExecutionsByIdsData` / `RestartExecutionsByQueryData` now carry
`query.latestRevision`, exactly like the replay equivalents already do. The store's
`bulkRestartExecution` forwards the flag the revision picker collects, so this is
what lets that choice reach the API.


loicmathieu (approved):
LGTM

PR #17705, opened 2026-07-29
fix(ai): default the namespace when the copilot authors a flow without one

MilosPaunovic:
## What

Asking the AI Copilot to **create a flow without a namespace** — the prompt doesn't mention one and no flow/namespace page is in focus (e.g. opening the copilot from the sidebar / a list page / the full-page `/ai`) — made `author-flow` fail with:

```
Value for the variable 'namespace' is null
```

and then the agent retried the identical `author-flow` call on its own, looping instead of stopping or asking for a namespace.

## Why

The `author-flow` tool declares `namespace` as optional (`required = false`, "omit if unknown"), so the model legitimately calls it with `namespace = null` when nothing supplied one. That null flows into `FlowGenerationPrompt` → the generation pipeline's prompt template, which is rendered by `PebbleSafeTemplate` — and that throws on any **null-valued** variable. The tool errored, and the orchestrator fed the error back to the model, which just re-issued the same call → the loop.

`buildPebbleExpressions` already tolerates a null namespace; it's the (non-editable) library prompt template that hard-requires it.

## Fix

In `AuthorFlowTool`, when the namespace is null/blank, default it to Kestra's canonical example namespace `company.team` so generation always succeeds (no crash, no loop), and report the assumption back via a new `assumedNamespace` field on the tool result. The `@Tool` description instructs the agent to tell the user which namespace was used and that they can change it before applying.

Revisions (a `currentFlowYaml` is provided) still default for the template's sake but do **not** claim an assumption — the real namespace lives in the flow's YAML and is preserved.

This is an OSS change; EE's `EeAuthorFlowTool` extends `AuthorFlowTool` only for RBAC, so it inherits the fix — no EE change needed.

## Tests

`AuthorFlowToolTest` — 9 passing (6 existing + 3 new): create-without-namespace defaults + reports the assumption, revise-without-namespace defaults but does not report, and namespace-provided reports nothing.

Closes https://github.com/kestra-io/kestra-ee/issues/9534.
Closes https://github.com/kestra-io/kestra-ee/issues/9541.


webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/AuthorFlowTool.java
```diff
@@ -49,7 +59,7 @@ public ArtefactKind artefact() {
...
-            + "Returns an object { draftId, kind, valid, validationError, yaml }: `draftId` and `yaml` identify the published draft and `valid` reflects validation (`validationError` holds the constraints when invalid). If generation fails, the tool errors with the reason instead of returning."
+            + "Returns an object { draftId, kind, valid, validationError, yaml, assumedNamespace }: `draftId` and `yaml` identify the published draft and `valid` reflects validation (`validationError` holds the constraints when invalid). When the flow was authored without a namespace, `assumedNamespace` holds the namespace it defaulted to — tell the user the flow was drafted in that namespace and that they can change it. If generation fails, the tool errors with the reason instead of returning."
     )
     public Result authorFlow(
         @P(name = "instructions", value = "What the flow should do, or how the current flow should be changed") String instructions,
```
ammeek, line 65:
```suggestion
public Result authorFlow(
@P(name = "instructions", value = "What the flow should do, or how the current flow should be changed", , defaultValue = DEFAULT_NAMESPACE) String instructions,
```

If we are setting a defaultNamespace we should add it to the tools spec.

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/AuthorFlowTool.java
```diff
@@ -100,12 +119,13 @@ private String validate(final String tenant, final String yaml) {
      * @param valid whether the generated YAML passed validation
      * @param validationError the validation constraints when invalid, else null
      * @param yaml the generated YAML
+     * @param assumedNamespace the namespace the draft defaulted to when the caller provided none, else null
      */
-    public record Result(String draftId, ArtefactKind kind, boolean valid, String validationError, String yaml)
+    public record Result(String draftId, ArtefactKind kind, boolean valid, String validationError, String yaml, String assumedNamespace)
```
ammeek, line 124: do we need to specifically return assumed namespace if it's already returned in the yaml?

ammeek (changes requested):

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/AuthorFlowTool.java
MilosPaunovic: The YAML has the namespace *value*, but not the fact that it was *assumed* vs. chosen. `assumedNamespace` is non-null only when we default it (new flow, no namespace given), so the agent can tell the user "drafted in `company.team`, change if needed" — and only when it was a guess. Without it, the agent can't tell a fallback from a deliberate choice, so it'd either go silent (the bug https://github.com/kestra-io/kestra-ee/issues/9534 fixes) or mention the namespace every time.

ammeek (approved):
LGTM

PR #17713, opened 2026-07-29
feat(flows): add Select All button to MULTISELECT input dropdown

### LINKED ISSUE #17639: Multi-Select Select All icon button request
### Feature description

Hello incredible team,

For multi selects, is it possible to have a "Select All" icon button of some sorts? Would be so helpful for us.

Image

Let me know your thoughts.

Siddhant-Kashyap:

---

### ✨ Description

What does this PR change?  
Adds a selectAll prop to the KsSelect design system component. When passed, a "Select All" and "Deselect All" button appears at the top of the dropdown. The MULTISELECT input in the Execute flow dialog uses this prop to let users select or clear all options in one click.

### 🔗 Related Issue

 Closes https://github.com/kestra-io/kestra/issues/17639

### 🎨 Frontend Checklist

image

### 📝 Additional Notes
Added "Deselect All" button alongside "Select All" as it felt like a natural companion it only appears once at least one item is selected. Let me know if you'd prefer to remove it and keep only "Select All".

### DIFF
--- a/ui/packages/design-system/src/components/Form/KsSelect/KsSelect.vue
+++ b/ui/packages/design-system/src/components/Form/KsSelect/KsSelect.vue
@@ -6,15 +6,17 @@
         <template v-if="$slots.prefix" #prefix>
             <slot name="prefix" />
         </template>
-        <template v-if="(selectAll && multiple) || $slots.header" #header>
-            <div v-if="selectAll && multiple" class="kel-select-all-header">
-                <button type="button" class="kel-select-all-btn" @click="selectAllVisible()">
-                    {{ $t('filter.select all') }}
-                </button>
-                <button v-if="model?.length" type="button" class="kel-select-all-btn" @click="model = []">
-                    {{ $t('filter.deselect all') }}
-                </button>
-            </div>
+        <template v-if="showSelectAll || $slots.header" #header>
+            <button
+                v-if="showSelectAll"
+                type="button"
+                class="kel-select-all-btn"
+                role="checkbox"
+                :aria-checked="allVisibleSelected ? 'true' : (someVisibleSelected ? 'mixed' : 'false')"
+                @click="toggleSelectAll()"
+            >
+                {{ $t('filter.select all') }}
+            </button>
             <slot v-if="$slots.header" name="header" />
         </template>
         <template v-if="$slots.footer" #footer>
@@ -81,10 +83,32 @@
 
     const elSelectRef = ref<InstanceType<typeof ElSelect>>()
 
-    const selectAllVisible = (): void => {
-        const opts: Array<{visible: boolean; value: any}> = (elSelectRef.value as any)?.optionsArray ?? []
-        const values = opts.filter(o => o.visible).map(o => o.value)
-        model.value = [...new Set([...(model.value ?? []), ...values])]
+    // Options passing ElSelect's own filter. `optionsArray` is exposed as a ComputedRef in the
+    // Element Plus types but unwrapped on the instance proxy, hence the cast.
+    const visibleOptions = computed<Array<{visible: boolean; value: any}>>(() =>
+        ((elSelectRef.value as any)?.optionsArray ?? []).filter((o: {visible: boolean}) => o.visible),
+    )
+
+    // Selecting nothing is meaningless, so the action stays hidden until there is something to select.
+    const showSelectAll = computed(() => Boolean(props.selectAll && props.multiple && visibleOptions.value.length))
+
+    const selectedValues = computed(() => new Set(Array.isArray(model.value) ? model.value : []))
+
+    const allVisibleSelected = computed(() =>
+        visibleOptions.value.length > 0 && visibleOptions.value.every(o => selectedValues.value.has(o.value)),
+    )
+
+    const someVisibleSelected = computed(() =>
+        !allVisibleSelected.value && visibleOptions.value.some(o => selectedValues.value.has(o.value)),
+    )
+
+    const toggleSelectAll = (): void => {
+        const values = visibleOptions.value.map(o => o.value)
+        model.value = allVisibleSelected.value
+            ? [...selectedValues.value].filter(v => !values.includes(v))
+            : [...new Set([...selectedValues.value, ...values])]
+        // Closing also clears the filter query, so the next open starts from the full list.
+        elSelectRef.value?.blur()
     }
 
     defineSlots<{
@@ -230,6 +254,33 @@
             box-shadow: none;
         }
 
+        .kel-select-dropdown__header {
+            padding: var(--ks-spacing-1);
+            border-bottom: 1px solid var(--ks-border-default);
+        }
+
+        .kel-select-all-btn {
+            display: block;
+            position: relative;
+            width: 100%;
+            background: none;
+            border: none;
+            border-radius: var(--ks-radius-xs);
+            cursor: pointer;
+            text-align: left;
+            font-family: inherit;
+            font-size: var(--ks-font-size-xs);
+            color: var(--ks-text-primary);
+            /* Mirrors the Element Plus option metrics — including the gutter kept free for the
+               check icon — so the row lines up with the list below. */
+            padding: 0 2rem 0 1.25rem;
+            height: 2.125rem;
+
+            &:hover {
+                background-color: var(--ks-bg-hover-elevated);
+            }
+        }
+
         .kel-select-dropdown__list {
             padding: var(--ks-spacing-1);
 
@@ -259,7 +310,8 @@
             }
         }
 
-        .kel-select-dropdown .kel-select-dropdown__item.is-selected::after {
+        .kel-select-dropdown .kel-select-dropdown__item.is-selected::after,
+        .kel-select-dropdown .kel-select-all-btn[aria-checked="true"]::after {
             content: "";
             position: absolute;
             right: 12px;
@@ -275,28 +327,5 @@
 
     .kel-icon.kel-select__caret.kel-select__icon {
         font-size: var(--ks-font-size-md);
-    }
-
-    .kel-select-all-header {
-        display: flex;
-        align-items: center;
-        justify-content: space-between;
-        padding: var(--ks-spacing-2) var(--ks-spacing-3);
-        border-bottom: 1px solid var(--ks-border-default);
-
-        .kel-select-all-btn {
-            background: none;
-            border: none;
-            padding: 0;
-            cursor: pointer;
-            font-family: inherit;
-            font-size: var(--ks-font-size-xs);
-            font-weight: var(--ks-font-weight-semibold);
-            color: var(--ks-text-link);
-
-            &:hover {
-                text-decoration: underline;
-            }
-        }
     }
 </style>
--- a/ui/packages/design-system/src/components/Form/KsSelect/KsSelect.vue
+++ b/ui/packages/design-system/src/components/Form/KsSelect/KsSelect.vue
@@ -1,13 +1,21 @@
 <template>
-    <ElSelect v-model="model" v-bind="({...filteredProps(), ...$attrs} as any)" :suffixIcon="resolvedSuffixIcon" :class="{'kel-select--fit': fit}" @change="emit('change', $event)">
+    <ElSelect ref="elSelectRef" v-model="model" v-bind="({...filteredProps(), ...$attrs} as any)" :suffixIcon="resolvedSuffixIcon" :class="{'kel-select--fit': fit}" @change="emit('change', $event)">
         <template v-if="$slots.default" #default>
             <slot />
         </template>
         <template v-if="$slots.prefix" #prefix>
             <slot name="prefix" />
         </template>
-        <template v-if="$slots.header" #header>
-            <slot name="header" />
+        <template v-if="(selectAll && multiple) || $slots.header" #header>
+            <div v-if="selectAll && multiple" class="kel-select-all-header">
+                <button type="button" class="kel-select-all-btn" @click="selectAllVisible()">
+                    {{ $t('filter.select all') }}
+                </button>
+                <button v-if="model?.length" type="button" class="kel-select-all-btn" @click="model = []">
+                    {{ $t('filter.deselect all') }}
+                </button>
+            </div>
+            <slot v-if="$slots.header" name="header" />
         </template>
         <template v-if="$slots.footer" #footer>
             <slot name="footer" />
@@ -22,7 +30,7 @@
 </template>
 
 <script setup lang="ts">
-    import {type Component, computed, h, markRaw} from "vue"
+    import {type Component, computed, h, markRaw, ref} from "vue"
     import {ElSelect} from "element-plus"
     import Loading from "vue-material-design-icons/Loading.vue"
     import KsIcon from "../../Basic/KsIcon.vue"
@@ -53,6 +61,7 @@
         suffixIcon?: Component | string
         loading?: boolean
         fit?: boolean
+        selectAll?: boolean
     }>(), {
         placeholder: undefined,
         size: undefined,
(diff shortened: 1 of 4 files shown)


MilosPaunovic (commented):
Thanks for this — the issue is real and the change is refreshingly small. A few things need addressing before merge, though, mostly because the design system **already has this exact control** and this adds a third, weaker copy of it.

### Blocking

**1. `selectAllOptions` is destructive — it drops custom values (data loss)**

[`InputsForm.vue:583-587`](https://github.com/kestra-io/kestra/blob/229f6c10490f808947177e62250bc6ea09c888ae/ui/src/components/inputs/InputsForm.vue#L583-L587) replaces the whole model with only the *predefined* option values:

```ts
const allValues = ((input.values ?? input.options) ?? []).map(toOption).map(o => o.value)
multiSelectInputs[input.id] = allValues
```

MULTISELECT passes `:allowCreate="input.allowCustomValue"`, so a user can have selected values that aren't in `input.values`/`input.options`. Clicking "Select All" silently discards them.

Worse case: `values` is optional on `MultiselectInput` (nothing in `MultiselectInputValidator` requires it), so for a `MULTISELECT` with `allowCustomValue: true` and no static `values`, `allValues` is `[]` — the button renders unconditionally and clicking it **clears the user's entire selection**. A control labelled "Select All" that wipes input is the worst possible failure mode here.

Should be additive, like the existing implementation:

```ts
const values = new Set(multiSelectInputs[input.id] ?? [])
options.forEach(o => values.add(o.value))
```

Also guard the header on there being at least one option.

**2. "Select All" ignores the active filter**

MULTISELECT is always `filterable`. Type `prod` to narrow 50 options down to 2, click Select All → you get all 50. [`FilterMultiSelect.handleSelectAllChange`](https://github.com/kestra-io/kestra/blob/229f6c10490f808947177e62250bc6ea09c888ae/ui/packages/design-system/src/components/Data/KsDataTable/filter/layout/FilterMultiSelect.vue#L280-L285) deliberately targets `filteredOptions`, not `props.options`. Whatever we pick, the two controls in the product shouldn't disagree.

**3. The `#header` slot is silently swallowed**

```vue
<template v-if="selectAll && multiple" #header>   …
<template v-else-if="$slots.header" #header>       …
```

Any consumer passing both `selectAll` and a `header` slot loses their slot content with no warning. Nothing hits this today, but it's a trap baked into a shared DS component. Render both, or compose them.

**4. Missing story + unit test**

`KsSelect` already has [`tests/units/Form/KsSelect.test.ts`](https://github.com/kestra-io/kestra/blob/229f6c10490f808947177e62250bc6ea09c888ae/ui/packages/design-system/tests/units/Form/KsSelect.test.ts) and [`tests/storybook/Form/KsSelect.stories.ts`](https://github.com/kestra-io/kestra/blob/229f6c10490f808947177e62250bc6ea09c888ae/ui/packages/design-system/tests/storybook/Form/KsSelect.stories.ts); neither is touched. `ui/AGENTS.md` requires a story + test in the same PR for a new DS prop. Worth covering: select-all emits, deselect-all clears, header hidden when `multiple` is false, header hidden with zero options.

### Design — please reconsider the approach

The `#header` slot **already existed** before this PR. So the whole feature could live in `InputsForm` with **zero** design-system changes, which sidesteps issues 3 and 4 entirely.

If we do want it in the DS (reasonable — consistency across call sites), then it shouldn't be a third hand-rolled copy. `FilterMultiSelect` already has the markup, the styles (`.select-all-btn` / `.clear-btn`), and better semantics:

- a tri-state `KsCheckbox` with `indeterminate`, `role="checkbox"` and `aria-checked="mixed"` — so the user can *see* whether everything is selected
- one toggle rather than two buttons, so "Select All" isn't a dead click once everything is already selected
- filter-aware, additive set operations

Extracting that header into a small shared component used by both `FilterMultiSelect` and `KsSelect` is the version of this change I'd merge.

### Non-blocking

- **i18n key ownership.** `$t('filter.select all')` resolves only because `registerDesignSystemI18n` globs every `*.locale.ts` into the global scope — the keys are owned by `KsFilter.locale.ts`. `ui/AGENTS.md` rule 10 says a component's keys live with that component, and `filter.*` is the wrong namespace for a generic form select. Move them somewhere shared (or to a `KsSelect.locale.ts`).
- **Asymmetric wiring.** "Select All" emits and defers to the parent; "Deselect All" mutates `model` directly. Both work (deselect propagates via `@update:model-value` → `onMultiSelectChange`), but pick one direction.
- **Silent no-op.** `selectAll` with no `@select-all` listener renders a button that does nothing. Passing options into the DS, or handling it internally, avoids the footgun.
- **`opacity: 0.8` on hover** — prefer a semantic token colour change over an opacity magic number.
- **Duplication.** `((input.values ?? input.options) ?? []).map(toOption)` now appears in both the template and `selectAllOptions` — worth a computed helper.

### Question

The checklist has E2E unrun. Since these buttons live inside the ElSelect popper, could you confirm manually: after clicking Select All, does the dropdown stay open, does the filter input keep focus, and does `Escape` still close it?

On your open question in the description — keep "Deselect All". It's the right companion; `FilterMultiSelect` does the same.

MilosPaunovic:
Hey @Siddhant-Kashyap, are there any updates on this?

Siddhant-Kashyap:
> Hey @Siddhant-Kashyap, are there any updates on this?

hello @MilosPaunovic I'll push the changes in a few hours once I get home. The changes are already ready; I just need to push them. Thanks for your patience.and sorry for delay

MilosPaunovic:
Is this now done @Siddhant-Kashyap?

Siddhant-Kashyap:
> Is this now done @Siddhant-Kashyap?

yes i have done the changes

**Blocking 1** : `selectAllOptions` is now additive (`new Set` + `forEach add`), preserving custom-typed values. Header is guarded when no options exist.

**Blocking 2** : `KsSelect` reads `ElSelect`'s `states.inputValue` via ref at click time and emits it with `select-all`. `selectAllOptions` filters predefined options by that query before adding  matching `FilterMultiSelect` behavior.

**Blocking 3** : `#header` template now activates on `(selectAll && multiple) || $slots.header` with inner `v-if` guards  both render together.

**Blocking 4** : Added `SelectAll` story to `KsSelect.stories.ts` (with `argTypes` entry) and 8 unit tests to `KsSelect.test.ts` covering header visibility, `select-all` emission with filter query, deselect-all, and slot composition.

**Non-blocking** : `opacity: 0.8` → `text-decoration: underline`; `font-weight: 600` → `var(--ks-font-weight-semibold)`.

**Design suggestion** : Happy to extract a shared component with `FilterMultiSelect` if preferred.


i18n / asymmetric wiring / duplication — will follow up if needed.

MilosPaunovic:
And what about 20+ files that are in conflict?

Siddhant-Kashyap:
> And what about 20+ files that are in conflict?

earlier i did git merge develop this :disappointed: 
and i had some changes in my local that also got pushed , now its fixed and updated

MilosPaunovic (commented):
Thanks for the quick turnaround — most of this is resolved. Verified fixed:

- **1** — select-all is additive now, custom-typed values survive. ✅
- **3** — `#header` composes with the select-all header, and there's a test for it. ✅
- **4** — story + 8 unit tests, Design System CI job green. ✅
- Styling — `--ks-font-weight-semibold` is a real token, `text-decoration` over `opacity`. ✅

Also, for what it's worth: `states.inputValue` is part of ElSelect's *typed* exposed surface (`select.vue.d.ts`), so that access is legitimate — and the `as any` on `elSelectRef.value` should be unnecessary. Worth trying without it.

### Still open

**2 is only half-fixed.** The plumbing works, but `selectAllOptions` re-implements ElSelect's filter predicate, and it already disagrees with it:

- ElSelect matches the **label only** — `new RegExp(escapeStringRegexp(query), "i").test(currentLabel)` in `useOption.mjs`. The PR matches label **or value**. So for `{label: "Production", value: "prod-eu-1"}` with query `eu-1`, the dropdown shows nothing but Select All silently adds the option.
- The PR trims the query; ElSelect doesn't. `"prod "` → empty dropdown, Select All still adds matches.
- Any consumer passing `filterMethod` or `remote` makes the re-implementation plainly wrong.
- The predicate is now duplicated a third time, in the story.

**"Header is guarded when no options exist"** — it isn't; the template is still just `v-if="selectAll && multiple"`. Not blocking anymore (additive select-all over zero options is a harmless no-op rather than destructive), but the guard isn't there.

**New:** `const current` at [`InputsForm.vue:591`](https://github.com/kestra-io/kestra/blob/a5011b211a8d298b9e4211b5c495aabc3fc2b997/ui/src/components/inputs/InputsForm.vue#L591) shadows the wizard's `current` ref — oxlint flags it in CI.

### One change that closes all of the above

Handle it **inside `KsSelect`**, symmetric with deselect-all — which already proves the pattern works:

```ts
const selectAllVisible = () => {
    const visible = (elSelectRef.value?.optionsArray ?? []).filter(o => o.visible)
    model.value = [...new Set([...(model.value ?? []), ...visible.map(o => o.value)])]
}
```

`optionsArray` is exposed and each option exposes `visible: Ref<boolean>` — that's ElSelect's *own* filter verdict, so divergence becomes structurally impossible. Deselect-all already does `model = []` and `InputsForm` picks it up via `@update:model-value` → `onMultiSelectChange`; select-all can work the same way.

That deletes the `select-all` emit, all 17 lines of `selectAllOptions`, the story's `onSelectAll`, and the shadowed `current` — and `optionsArray.length` gives you the zero-options guard for free. It also resolves the asymmetric-wiring and duplication points from my first pass.

### CI

Red, but **not your fault** — the hard failure is `FilePreview.spec.ts:27` ("Unnecessary escape character"), which this PR doesn't touch. Your merge base still had the bad escape; develop has since fixed it. Rebase on latest develop and it clears. The only annotation attributable to this PR is the `current` shadow warning above.

Siddhant-Kashyap:
> Thanks for the quick turnaround — most of this is resolved. Verified fixed:
> 
>     * **1** — select-all is additive now, custom-typed values survive. ✅
> 
>     * **3** — `#header` composes with the select-all header, and there's a test for it. ✅
> 
>     * **4** — story + 8 unit tests, Design System CI job green. ✅
> 
>     * Styling — `--ks-font-weight-semibold` is a real token, `text-decoration` over `opacity`. ✅
> 
> 
> Also, for what it's worth: `states.inputValue` is part of ElSelect's _typed_ exposed surface (`select.vue.d.ts`), so that access is legitimate — and the `as any` on `elSelectRef.value` should be unnecessary. Worth trying without it.
> ### Still open
> 
> **2 is only half-fixed.** The plumbing works, but `selectAllOptions` re-implements ElSelect's filter predicate, and it already disagrees with it:
> 
>     * ElSelect matches the **label only** — `new RegExp(escapeStringRegexp(query), "i").test(currentLabel)` in `useOption.mjs`. The PR matches label **or value**. So for `{label: "Production", value: "prod-eu-1"}` with query `eu-1`, the dropdown shows nothing but Select All silently adds the option.
> 
>     * The PR trims the query; ElSelect doesn't. `"prod "` → empty dropdown, Select All still adds matches.
> 
>     * Any consumer passing `filterMethod` or `remote` makes the re-implementation plainly wrong.
> 
>     * The predicate is now duplicated a third time, in the story.
> 
> 
> **"Header is guarded when no options exist"** — it isn't; the template is still just `v-if="selectAll && multiple"`. Not blocking anymore (additive select-all over zero options is a harmless no-op rather than destructive), but the guard isn't there.
> 
> **New:** `const current` at [`InputsForm.vue:591`](https://github.com/kestra-io/kestra/blob/a5011b211a8d298b9e4211b5c495aabc3fc2b997/ui/src/components/inputs/InputsForm.vue#L591) shadows the wizard's `current` ref — oxlint flags it in CI.
> ### One change that closes all of the above
> 
> Handle it **inside `KsSelect`**, symmetric with deselect-all — which already proves the pattern works:
> 
> ```ts
> const selectAllVisible = () => {
>     const visible = (elSelectRef.value?.optionsArray ?? []).filter(o => o.visible)
>     model.value = [...new Set([...(model.value ?? []), ...visible.map(o => o.value)])]
> }
> ```
> 
> `optionsArray` is exposed and each option exposes `visible: Ref<boolean>` — that's ElSelect's _own_ filter verdict, so divergence becomes structurally impossible. Deselect-all already does `model = []` and `InputsForm` picks it up via `@update:model-value` → `onMultiSelectChange`; select-all can work the same way.
> 
> That deletes the `select-all` emit, all 17 lines of `selectAllOptions`, the story's `onSelectAll`, and the shadowed `current` — and `optionsArray.length` gives you the zero-options guard for free. It also resolves the asymmetric-wiring and duplication points from my first pass.
> ### CI
> 
> Red, but **not your fault** — the hard failure is `FilePreview.spec.ts:27` ("Unnecessary escape character"), which this PR doesn't touch. Your merge base still had the bad escape; develop has since fixed it. Rebase on latest develop and it clears. The only annotation attributable to this PR is the `current` shadow warning above.

**Filter predicate mismatch (blocking)** -> Removed `selectAllOptions` from `InputsForm.vue` entirely. Select All logic now lives inside `KsSelect` using ElSelect's own `optionsArray.filter(o => o.visible)`  no re-implementation of filter matching, so divergence with custom `filterMethod` or `remote` is structurally impossible.

**Variable shadowing (oxlint)** -> `const current` is gone along with the removed function.

**Asymmetric wiring (non-blocking)** -> Both Select All and Deselect All now update `model.value` directly inside `KsSelect`. No more emit on one side, direct mutation on the other.

**Code duplication (non-blocking)** -> Removed with `selectAllOptions`, no duplication remains.

**Zero-options guard** -> Handled automatically: if no options are visible, `optionsArray.filter(o => o.visible)` returns `[]` and the Set merge is a no-op.

MilosPaunovic (approved):

MilosPaunovic:
Thanks for sticking with this through three rounds, @Siddhant-Kashyap - the `optionsArray` approach you landed on is exactly right, and moving it into the design system meant `InputsForm` needed just a single prop. Nice result.

I've pushed some polish on top rather than sending you round again:

- **Removed "Deselect All"** — the row is now a single toggle; `clearable` already covers clear-everything.
- **Restyled to match the option list** — option padding/height, `--ks-text-primary`, same hover background, so it reads as the first row of the list rather than a blue link bar.
- **Fixed a doubled separator** — `.kel-select-dropdown__header` already ships `padding` + `border-bottom` from Element Plus, so the inner wrapper was adding a second pair.
- **Clicking closes the dropdown** via `blur()`, which also clears the filter query.
- **Hidden when there's nothing to select** — no options, or a filter matching none.
- **Checked state shows the same right-side checkmark as selected options**; clicking a checked row deselects the visible ones.

Tests 8 → 24, now covering the filtered path — which was the one behavior that had gone wrong twice and the one thing with no test. Thanks again, this is a real improvement!

Siddhant-Kashyap:
> Thanks for sticking with this through three rounds, @Siddhant-Kashyap - the `optionsArray` approach you landed on is exactly right, and moving it into the design system meant `InputsForm` needed just a single prop. Nice result.
> 
> I've pushed some polish on top rather than sending you round again:
> 
> - **Removed "Deselect All"** — the row is now a single toggle; `clearable` already covers clear-everything.
> - **Restyled to match the option list** — option padding/height, `--ks-text-primary`, same hover background, so it reads as the first row of the list rather than a blue link bar.
> - **Fixed a doubled separator** — `.kel-select-dropdown__header` already ships `padding` + `border-bottom` from Element Plus, so the inner wrapper was adding a second pair.
> - **Clicking closes the dropdown** via `blur()`, which also clears the filter query.
> - **Hidden when there's nothing to select** — no options, or a filter matching none.
> - **Checked state shows the same right-side checkmark as selected options**; clicking a checked row deselects the visible ones.
> 
> Tests 8 → 24, now covering the filtered path — which was the one behavior that had gone wrong twice and the one thing with no test. Thanks again, this is a real improvement!

Thanks for your support @MilosPaunovic

PR #17724, opened 2026-07-30
perf: cut redundant work on the JDBC, retry and worker hot paths

loicmathieu:
### Description

Three targeted performance fixes on the JDBC/executor hot path, all found by CPU and allocation profiling a running instance under load.

**`refactor(jdbc): avoid double-binding the upsert payload via DSL.excluded()`**

`persist()` / `buildInsertRequest()` bound the full serialized entity **twice** per upsert — once for the `INSERT` values and again for the `ON CONFLICT` / `ON DUPLICATE KEY UPDATE` set clause. Since the payload is the whole entity as a single JSON/JSONB value, every upsert sends it over the wire twice.

A shared `AbstractJdbcRepository.excluded(...)` helper now maps each column to `DSL.excluded(field)`, so the update clause references the row already bound by the insert. jOOQ renders this per dialect (the `EXCLUDED` pseudo-table on Postgres, `VALUES(column)` on MySQL/MariaDB. H2 is unaffected: it does UPDATE-then-INSERT-if-0-rows and never uses `onConflict`/`onDuplicateKeyUpdate`.

This matters because queue publishing was measured at ~65% wire cost, so payload bytes are the dominant term there.

**`refactor(core): reuse a single retry policy chain instead of rebuilding it per transaction`**

`RetryUtils.Instance.runRetryIf()` rebuilt the entire Failsafe policy chain — retry policy, fallback, and their two listeners — on **every call**. `JooqDSLContextWrapper` calls it on every database transaction, so this was being reconstructed for essentially every query the application ran.

Failsafe policies and executors are immutable and thread-safe once built, so `Instance.retryerIf(predicate)` now builds the chain once and returns a reusable `Retryer`; `JooqDSLContextWrapper` holds a single static instance.

**`refactor(worker): bind the kill handler once instead of per job processed`**

`process()` passed `this::kill` as a fresh capturing lambda to `executionKilledManager.register()` on every job, even though a single processor instance handles many jobs sequentially and `this` never changes between them. Now bound once as a field. Allocation profiling attributed ~1% of all application allocations to this one call site.

### 🛠️ Backend Checklist

Ran targeted suites against a live Postgres 14: `PostgresRepositoryTest`, `RetryUtilsTest`, `JooqDSLContextWrapperDeadlockPredicateTest` and `WorkerTaskProcessorTest` (the last exercising the shutdown-interrupt path that goes through the changed kill registration).

### DIFF
--- a/core/src/main/java/io/kestra/core/repositories/ExecutionRepositoryInterface.java
+++ b/core/src/main/java/io/kestra/core/repositories/ExecutionRepositoryInterface.java
@@ -88,9 +88,8 @@ public interface ExecutionRepositoryInterface extends QueryBuilderInterface<Exec
         @Nullable ZonedDateTime endDate,
         @Nullable List<State.Type> state,
         @Nullable Map<String, String> labels,
-        @Nullable String triggerExecutionId,
-        @Nullable ChildFilter childFilter) {
-        return find(query, tenantId, scope, namespace, flowId, startDate, endDate, state, labels, triggerExecutionId, childFilter, false);
+        @Nullable String triggerExecutionId) {
+        return find(query, tenantId, scope, namespace, flowId, startDate, endDate, state, labels, triggerExecutionId, false);
     }
 
     Flux<Execution> find(
@@ -104,7 +103,6 @@ public interface ExecutionRepositoryInterface extends QueryBuilderInterface<Exec
         @Nullable List<State.Type> state,
         @Nullable Map<String, String> labels,
         @Nullable String triggerExecutionId,
-        @Nullable ChildFilter childFilter,
         boolean allowDeleted);
 
     Flux<Execution> findAllAsync(@Nullable String tenantId);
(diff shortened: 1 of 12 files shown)


nkwiatkowski (approved):

loicmathieu:
Analysis of our official simple benchmark (1) shows an improvement, but it must be taken with caution as it's a single run of a single benchmark.

**Verdict**: **improvement**, especially under saturation: at 4000 req/s avg exec latency drops from ~15s to ~4.1s (-72%) and p99 from 57s to 8.4s (-85%), driven by the candidate handling the
  queue-drain query far more efficiently. Light-to-moderate load (1000–3000 req/s) sees a steady 4–11% latency improvement. Trade-off: kestra CPU runs higher on the candidate at saturation,
  and heap is slightly higher at mid-range rates: it trades for the latency/throughput gains.

PR #17731, opened 2026-07-30
perf(tests): speed up the E2E CI job

elevatebart:
## What

Cuts the E2E CI job's wall-clock time while testing the same things. Measured on a green scheduled run ([30491690152](https://github.com/kestra-io/kestra/actions/runs/30491690152)), the ~12-minute job (16+ when flaky) splits into: build 157s, backend boot ~50s, Playwright 6.5min — of which ~3min were two chronic flaky retries in the bulk-actions specs.

### Build stage
- Skip `./gradlew clean` in CI — the checkout is fresh, so the invocation is pure overhead. Local behavior unchanged.
- Skip the duplicate `npm ci` in CI via a new `SKIP_NPM_CI` make variable — the workflow already runs `npm ci` two steps before `make build-frontend` reinstalls node_modules from scratch (~40s).
- Pre-pull the Docker images needed later (kestra-base, postgres) in the background at script start so the downloads overlap with the Gradle build instead of sitting on the critical path.

### Drop the unused dind sidecar
No E2E flow or fixture exercises a Docker-runner task, so the `docker:dind-rootless` container and its `dind-socket` volume in `docker-compose-postgres.yml` were pure overhead — an extra image to pull, an extra privileged container to boot, and CPU/memory pressure on the already-contended 4-vCPU runner for zero test coverage. Removed the service, its volume, the `init` container's now-unnecessary chown of it, and a stale comment in `playwright.config.ts` that referenced it.

### Test stage (same coverage, less load)
The three bulk-action tests each created 26+1 executions **serially** over the API, then bulk-restarted/replayed all 26 on a 4-vCPU runner that also hosts the JVM, Postgres, dind and the browsers. That contention is exactly what made stragglers miss the 60s reload-poll window and trigger retries.
- The dataset shrinks from 26 executions / 25-per-page to `PAGE_SIZE + 1` = 11 / 10-per-page. The semantics under test — *Select All reaches beyond the visible page* — are identical and now explicit in the constant.
- Executions are created and deleted concurrently.
- Dropped a `waitForLoadState("networkidle")`.

Note: [#17726](https://github.com/kestra-io/kestra/pull/17726) landed on `develop` while this branch was open and independently reworked the same pagination helpers (URL-driven `setPaginationTo`, total-counter-based after-refresh assertion). This branch is now rebased on top of it and keeps its version of those helpers — my earlier dropdown-click race fix is superseded by its simpler URL-based approach.

## Expected result

Job time should drop from ~12min to ~8–9min, and the two chronic bulk-actions flakes (each retry costs ~90s and they have failed most scheduled runs this week) should disappear since the contention that caused them is gone.

## Verified

Ran the actual CI script end to end locally (`build-and-start-e2e-tests.sh`: fresh Gradle build → local Docker image build from source → dind-free `docker compose up` → full Playwright suite): **32 passed**, 260s total (125s of that is the test run itself). `tsc --noEmit` clean on the e2e tsconfig.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,8 +44,10 @@ buildSkipTests: clean
 test: clean
 	./gradlew test
 
+# Set SKIP_NPM_CI=true when dependencies are already installed (e.g. the CI
+# workflow ran `npm ci` beforehand) to avoid a redundant reinstall.
 build-frontend:
-	cd ui && npm ci && npm run build
+	cd ui && { [ "${SKIP_NPM_CI}" = "true" ] && [ -d node_modules ] || npm ci; } && npm run build
 
 build-exec: build-frontend
 	./gradlew -q executableJar --no-daemon --priority=normal
--- a/build-and-start-e2e-tests.sh
+++ b/build-and-start-e2e-tests.sh
@@ -15,8 +15,26 @@ start_time=$(date +%s)
 
 echo ""
 echo "Building the image for this current repository"
-./gradlew clean -q
-make build-docker VERSION=$LOCAL_IMAGE_VERSION
+
+# Pull the images needed later (docker build base + compose services) in the
+# background so the downloads overlap with the Gradle/npm build instead of
+# sitting on the critical path.
+BASE_IMAGE="$(sed -n 's/^ARG BASE_IMAGE="\(.*\)"/\1/p' Dockerfile)"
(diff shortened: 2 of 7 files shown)


MilosPaunovic (approved):
🚀 🚀 🚀

PR #17736, opened 2026-07-30
perf(tests): run the ui unit suite without per-file isolation, and warn on state leaks

elevatebart:
## What

Two changes to the `unit` vitest project:

1. **`isolate: false`** — keep `node_modules` warm in the worker instead of re-importing them per file, with `vi.resetModules()` in the shared setup to keep modules per-file fresh.
2. **`tests/unit/leakGuard.ts`** — warns when a spec leaks shared state into the next one.

CI unit-test step: **129–178s → 45s** (vs recent `develop` runs of the same job). Locally ~48s → ~16s.

The saving is in imports (cumulative 285s → 94s) — the jsdom environment is still rebuilt per file, so that cost is unchanged.

## Why the setup change is required

Sharing a worker shares the module registry, which breaks `vi.mock`: once a module is evaluated while one spec's mock is active, its cached instance keeps that mock, so a later spec mocking the same id transitively gets the earlier spec's factory. `useStateFilter.spec.ts` got `DrillDownDrawer.spec.ts`'s `useRoute()` (no `query`); SDK-mocking specs ran real request functions.

53 of 117 specs call `vi.mock` (27 on `vue-router`, 17 on the SDK), so which pairs collide depends on run order — patching individual specs would leave every future one a landmine. `vi.resetModules()` runs before each spec's own imports, giving every file a fresh module graph without rebuilding the environment 121 times.

## Leak guard

Globals, DOM, storage and timers are still shared. The guard snapshots them per file and reports against **the file that caused the leak**, not the unrelated file that fails later:

```
⚠️  state leak in tests/unit/stores/pluginsLoadIcon.spec.ts:
   - globalThis.Image was replaced and not restored (call vi.unstubAllGlobals() in afterAll)
```

It flags ~20 files today (leftover `localStorage` keys, DOM left attached to `document.body`, unrestored `document.title`, stubbed globals), none of which break the suite — so it is **warn-only**. Cleanup is follow-up work.

- `npm run test:unit:leaks` — turns findings into failures
- `npm run test:unit:shuffle` — randomizes file order (green across 5 seeds)

Reports go to fd 2 because the root `vitest.config.js` swallows console output of passing files.

## Notes

`unstubGlobals: true` was tried and dropped — it only restores between tests, and made the guard misattribute an inherited stub to an innocent file.

Full `--sequence.shuffle` (which also shuffles tests within a file) surfaces a pre-existing failure in `flowAutoCompletionProvider.spec.ts:282` — it asserts `toHaveBeenCalledOnce()` on a mock no `beforeEach` resets. Fails with `--isolate` too, so left alone.

## Test plan

`npm run test:unit` x3, `--isolate`, `--coverage`, and `test:unit:shuffle` across 5 seeds — all green.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -24,7 +24,7 @@
         "test:all": "vitest run --coverage",
         "test:unit": "vitest run --project=unit",
         "test:unit:shuffle": "vitest run --project=unit --sequence.shuffle.files",
-        "test:unit:leaks": "VITEST_LEAK_GUARD=error vitest run --project=unit --sequence.shuffle.files",
+        "test:unit:leaks": "cross-env VITEST_LEAK_GUARD=error vitest run --project=unit --sequence.shuffle.files",
         "test:storybook": "./scripts/run-storybook-tests.sh",
         "test:e2e": "./run-e2e-tests.sh",
         "test:e2e-without-starting-backend": "playwright test --config=tests/e2e/playwright.config.ts",
--- a/ui/package.json
+++ b/ui/package.json
@@ -23,6 +23,8 @@
         "test:lint": "oxlint src packages && eslint",
         "test:all": "vitest run --coverage",
         "test:unit": "vitest run --project=unit",
+        "test:unit:shuffle": "vitest run --project=unit --sequence.shuffle.files",
+        "test:unit:leaks": "VITEST_LEAK_GUARD=error vitest run --project=unit --sequence.shuffle.files",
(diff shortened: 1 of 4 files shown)


elevatebart:
Makes unit tests run in 49s instead of 2m58s (3.6x)

MilosPaunovic (approved):
I'm very happy seeing couple of PRs lately that target FE test speed increases!

PR #17737, opened 2026-07-30
perf(tests): parallelize and slim down the frontend CI

elevatebart:
## What

Companion to **kestra-io/actions#169** — together they cut the frontend CI critical path from a measured **10m20s–11m30s** to an expected **~2.5–3.5 min** per PR.

### Storybook tests (was 5m02s, the biggest cost)
- [.storybook/preview.jsx](../blob/elevate-bot/frontend-ci-performance-a5dd50/ui/.storybook/preview.jsx): removed the eager `import.meta.glob` of **every `@types/node` .d.ts as raw strings** fed to Monaco's TS worker in each browser page — `src/utils/monacoEnvironment` already registers the same types lazily on the first TS worker request (and the eager copy double-registered them under a second `file://` path).
- [vitest.config.js](../blob/elevate-bot/frontend-ci-performance-a5dd50/ui/vitest.config.js): sets `STORYBOOK=true` before the vite config resolves. `storybook dev/build` set it themselves but the vitest addon doesn't, so **Module Federation + dts-plugin + VitePWA/workbox were running inside every test shard**.
- [scripts/run-storybook-tests.sh](../blob/elevate-bot/frontend-ci-performance-a5dd50/ui/scripts/run-storybook-tests.sh): new `SHARD=i/n` mode — CI now runs the 4 shards as a **parallel matrix** (one fresh process + browser each) instead of sequentially. Shards can't be collapsed: even with the fixes above, an unsharded run still crashed with `[birpc] rpc is closed` after ~31 story files, so browser-session degradation (Monaco workers created per-model, never terminated) is real beyond the types payload. Local runs keep the sequential shards + merged report.

### Unit tests
- `pool: "threads"` (cheaper than the forks default). Isolation kept: `--no-isolate` is 3× faster but 4 spec files leak module state — flagged as a follow-up.

### JUnit reports were silently never produced
Vitest ignores project-level `reporters` config, so `test-report.junit.xml` / `test-report.storybook.junit.xml` never existed and the PR failing-tests comment has been a no-op. The junit reporter now comes from CLI flags (`test:unit` script, storybook script); the reporter composite in actions#169 globs the per-shard files.

### Build once per PR, skip what's unneeded
- [pull-request.yml](../blob/elevate-bot/frontend-ci-performance-a5dd50/.github/workflows/pull-request.yml): `e2e-tests` and `generate-pull-request-docker-image` are now gated on ui/backend file changes (docs-only and draft PRs stop burning ~18 runner-minutes), and a hoisted `build-artifacts` job feeds both — the E2E job previously rebuilt the entire product (npm build + Gradle executableJar + uncached docker build) in parallel with the identical build in the docker-image job.
- [Makefile](../blob/elevate-bot/frontend-ci-performance-a5dd50/Makefile) `build-docker-from-exec` + `E2E_USE_PREBUILT_EXE` in [build-and-start-e2e-tests.sh](../blob/elevate-bot/frontend-ci-performance-a5dd50/build-and-start-e2e-tests.sh): E2E builds only the Docker image from the prebuilt executable. Fork PRs (no shared artifacts) keep the build-from-source fallback.
- `ui/tsconfig.test.json`: `"rereferences"` → `"references"` (correctness only; A/B-benchmarked at no measurable timing change).

## Merge order

⚠️ **Merge kestra-io/actions#169 first.** The reusable workflows are pinned `@main`, and this PR passes new inputs (`exe-artifact`, `skip-build`) that don't exist there yet — CI on this PR will fail with "invalid input" errors until actions#169 lands, then re-running validates the full setup live.

## Testing done locally

- `npm run check:types` green (before + after the references fix, A/B ×2 each).
- `npm run test:unit`: 121 files / 983 tests green with threads pool, junit written.
- Storybook: `SHARD=2/4` mode green with per-shard junit; STORYBOOK gating verified as the only two env usages; a full-suite run passed once and crashed once (the `[birpc]` crash), confirming shards must stay.
- Threads + `--no-isolate` trialed: 4 files / 15 tests fail from module-state leaks → follow-up instead.
- All YAML parse-validated, scripts `bash -n` clean, make targets dry-run verified.

## After merge, validate on this PR

1. Re-run CI once actions#169 is on main; check the four frontend jobs' wall times.
2. Push a commit with a deliberately failing unit test + storybook test → the PR comment should list both; revert → comment clears.
3. Confirm the E2E job skips Java/Gradle setup and builds the image from the `exe` artifact.


ui/vitest.config.js
```diff
@@ -21,6 +21,11 @@ const dirname =
 // Instead, disable the scan during merge only — nothing is served here anyway.
 const isMergeReports = process.argv.includes("--merge-reports")
 
+// `storybook dev/build` set STORYBOOK themselves, but the storybook vitest
+// addon does not — without it, vite.config.js enables Module Federation and
+// VitePWA in every test run. Must be set before viteConfig() is called below.
+process.env.STORYBOOK ??= "true"
```
elevatebart, line 30: Will this not leak into unit tests?

ui/vitest.config.js
```diff
@@ -82,18 +87,8 @@ export default defineConfig({
...
-                    // single, whole-suite JUnit report CI reads to list failing tests.
-                    reporters: [
-                        ["default"],
-                        ["junit"],
-                    ],
-                    outputFile: {
-                        junit: "./test-report.storybook.junit.xml",
-                    },
+                    // No reporters here: Vitest ignores project-level `reporters`;
+                    // run-storybook-tests.sh passes them (incl. junit) on the CLI.
```
elevatebart, line 95:
```suggestion
```

elevatebart:
Closing — most of these ideas have landed via other PRs (e.g. #17736 for the unit-test isolation/leak fixes), and the storybook sharding approach here has been abandoned. Keeping the exploration/findings in the plan history for reference.

PR #17739, opened 2026-07-30
fix(core): bind dedicated JDBC log store datasource config

nkwiatkowski:
## What

Fixes a boot crash when a JDBC store (`h2` / `postgres` / `mysql`) is configured as an **external log data store pointed at a dedicated database** via `kestra.logs.<type>.{url,username,password,table}`.

With the documented config, the server failed to start with:

```
Failed to create log store 'postgres'. Error: Unrecognized field "url"
(class io.kestra.repository.postgres.PostgresLogDataStore)
```

Refs kestra-io/kestra-ee#9549

## Root cause

`kestra.logs.<type>.*` is consumed twice:

- `LogJdbcDataSourceProvider` reads `url/username/password/table` from it to build the dedicated Hikari pool and run the log-table migrations — this works.
- `KestraBeansFactory.logDataStore()` passes the **same** map into `LogDataStoreInterfaceFactory.make()` → `AbstractPluginInterfaceFactory.resolve()` → `JacksonMapper.toMap(config, <store>.class)`, which uses the strict mapper. The JDBC store classes declared **none** of those fields, so deserialization threw `UnrecognizedPropertyException` and the bean (and the server) never came up.

The existing `H2LogDataStoreDedicatedTest` didn't catch it because it called `make("h2", Map.of())` with an **empty** map, sidestepping the exact deserialization production performs.

## Fix

Declare `url` / `username` / `password` / `table` as bound properties on `AbstractJdbcLogDataStore` (mirroring how `ElasticSearchLogDataStore` declares its `index`), so the sub-config deserializes cleanly onto the store plugin. `table` is bound via a `tableName` field with `@JsonProperty("table")` to avoid clashing with the inherited jOOQ `getTable()`.

`LogJdbcDataSourceProvider` remains the single authority for the connection — it still reads the keys from `LogsConfig` and owns the dedicated pool + migrations. These fields are just what lets the strict plugin deserialization accept (and document) the same config.

**No behaviour change to the default / fallback path:** with no `kestra.logs.type` (logs stay in the main backend DB), an empty config binds with the new fields left `null`, so the store reuses the primary datasource exactly as before.

## Tests

- `H2LogDataStoreDedicatedTest` now builds the store from the **real** `logsConfig.getLogConfig("h2")` map instead of an empty map, so it exercises and guards the deserialization that regressed.
- New pure-Jackson binding tests: `PostgresLogDataStoreConfigBindingTest`, `MysqlLogDataStoreConfigBindingTest`.
- Existing log-store suite green.

## Manual verification

Built the standalone jar and booted it against two Postgres databases (main + dedicated logs) with the documented config. Server starts, a flow's logs land in the **dedicated** logs DB (`SELECT count(*) FROM logs` → 1) with **0** rows in the main backend DB, and read back correctly through `GET /api/v1/{tenant}/logs/{executionId}`.

## Credentials for the dedicated log database

The dedicated log datasource is fully independent — its credentials are **never** inherited from the main datasource (inheriting could silently write logs to the customer's log database as a privileged backend user). If `kestra.logs.<type>.url` is set without a `username`, the server now **fails fast** at startup with a clear message rather than building a pool with null credentials. `password` stays optional (passwordless / trust auth).

---
Companion EE PR (Elasticsearch log store `index`, same root cause): kestra-io/kestra-ee#9615


jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLogDataStore.java
```diff
@@ -35,15 +38,49 @@
...
+
+    @Schema(title = "Password for the dedicated logs database.")
+    @PluginProperty(hidden = true)
+    @Getter
+    @Setter
+    private String password;
+
+    @Schema(title = "Name of the log table in the dedicated logs database (defaults to 'logs').")
+    @PluginProperty
+    @JsonProperty("table")
```
loicmathieu, line 73: Why?

loicmathieu (changes requested):
There are too much comment and maybe also too much tests...

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLogDataStore.java
nkwiatkowski: Because table is already used by AbstractJdbcCrudRepository, so can't be use as a field name.

loicmathieu (approved):
LGTM

PR #17740, opened 2026-07-31
feat(onboarding): spotlight the part of the page each tour step is about

anna-geller:
## Spotlight

Everything but the part a step is about is dimmed, and that part gets a ring that pulses briefly. The ring follows its target, so a panel that resizes or logs that arrive after the page keep it in the right place, and a target below the fold is scrolled into view. Suppressed while a dialog is open, and static under `prefers-reduced-motion`.

Selector lists are ranked, so a step can fall back to the timeline until its log line renders. Each step lights what its copy describes: the flow in the editor, the failed task run, the revision diff, the **Debug Expression** button. Step 12 keeps the ring without the dim, since reading the payload that button returns is the point of that step.

## AI Copilot copy

The tour uses a ready-made flow rather than prompting a model, so step 1 is now "Flows can be drafted from a prompt" with a **Use the example flow** button, and step 4 no longer credits Copilot for that flow. The first summary card trades the Execute sentence for the agent skills that generate flows from an editor or a terminal.

Five of the strings rewritten here were already translated by #17789, and the generator skips keys that have a value, so those five are removed from the 12 languages for the next scheduled run to fill.

## Review

@RemiBarthe this is the glow effect you asked about, and it builds on your #17763, so step 2 keeps its card on the left and step 12 keeps its card on the right. Could you look at the technical side, in particular the per-frame measurement in `TourOverlay.vue`, and change anything you would rather have done differently?


Piyush-r-bhaskar (approved):
LGTM

PR #17741, opened 2026-07-31
feat: add missing activation

Skraye:
Fixes flow execution tracking, which counted Execute clicks instead of actual runs — a single run was reported twice, a cancelled one still counted.

Adds the missing activation events for flow creation, secrets and KV, plus a section property on page views.

Companion PR: https://github.com/kestra-io/kestra-ee/pull/9605

 Relates to : https://github.com/kestra-io/kestra-ee/issues/9524

### DIFF
--- a/ui/src/components/flows/FlowRun.vue
+++ b/ui/src/components/flows/FlowRun.vue
@@ -398,6 +398,9 @@
                 apiStore.posthogEvents({
                     type: "FLOW_EXECUTION",
                     action: "submit",
+                    // Replay-with-inputs bypasses triggerExecution, so it never emits a matching
+                    // `executed`: flagged so it can be excluded from the submit/executed funnel.
+                    is_replay: Boolean(props.replaySubmit),
                 })
 
                 const mergedInputs = props.selectedTrigger?.inputs
--- a/ui/src/components/flows/FlowRun.vue
+++ b/ui/src/components/flows/FlowRun.vue
@@ -387,10 +387,6 @@
 
     function onSubmit() {
         if (form.value && flowCanBeExecuted.value) {
-            apiStore.posthogEvents({
-                type: "FLOW_EXECUTION",
-                action: "submit",
-            })
             checks.value = []
             executeClicked.value = false
             coreStore.message = undefined
@@ -398,6 +394,11 @@
                 if (!valid) {
                     return
                 }
+
+                apiStore.posthogEvents({
+                    type: "FLOW_EXECUTION",
+                    action: "submit",
+                })
 
                 const mergedInputs = props.selectedTrigger?.inputs
                     ? {...props.selectedTrigger.inputs, ...inputsNoDefaults.value}
--- a/ui/src/components/secrets/SecretsTable.vue
+++ b/ui/src/components/secrets/SecretsTable.vue
@@ -564,12 +564,17 @@
                 ? namespacesStore.createSecrets
                 : namespacesStore.patchSecret
 
-            actionMethod({namespace: secret.value?.namespace as string, secret: secretData})
+            // Snapshot before the request: resetForm() swaps secret.value out when the drawer closes,
+            // and the .then() would then read the flag off a different object.
+            const wasUpdate = secret.value?.update === true
+            const namespace = secret.value?.namespace
+
+            actionMethod({namespace: namespace as string, secret: secretData})
                 .then(() => {
                     apiStore.posthogEvents({
-                        type: secret.value?.update === true ? "SECRET_UPDATED" : "SECRET_CREATED",
+                        type: wasUpdate ? "SECRET_UPDATED" : "SECRET_CREATED",
                         secret_type: "secret",
-                        namespace: secret.value?.namespace,
+                        namespace,
                         has_tags: (secretData.tags?.length ?? 0) > 0,
                     })
 
--- a/ui/src/components/secrets/SecretsTable.vue
+++ b/ui/src/components/secrets/SecretsTable.vue
@@ -241,6 +241,7 @@
     import * as SecretsAPI from "@kestra-io/kestra-sdk/secrets"
     import {useAuthStore} from "override/stores/auth"
     import {useNamespacesStore} from "override/stores/namespaces"
+    import {useApiStore} from "../../stores/api"
     import {useSecretsFilter} from "../filter/configurations"
     import {useTableColumns} from "../../composables/useTableColumns"
     import {useDiscardGuard} from "../../composables/useDiscardGuard"
@@ -294,6 +295,7 @@
     const router = useRouter()
     const authStore = useAuthStore()
     const namespacesStore = useNamespacesStore()
+    const apiStore = useApiStore()
 
     const form = ref<FormInstance>()
 
@@ -564,6 +566,13 @@
 
             actionMethod({namespace: secret.value?.namespace as string, secret: secretData})
                 .then(() => {
+                    apiStore.posthogEvents({
+                        type: secret.value?.update === true ? "SECRET_UPDATED" : "SECRET_CREATED",
+                        secret_type: "secret",
+                        namespace: secret.value?.namespace,
+                        has_tags: (secretData.tags?.length ?? 0) > 0,
+                    })
+
                     secret.value!.update = true
                     toast.saved(secret.value?.key || "")
                     addSecretDrawerVisible.value = false
--- a/ui/src/stores/flow.ts
+++ b/ui/src/stores/flow.ts
@@ -18,6 +18,8 @@ import type {FlowWithSource,  AbstractTrigger, Task as SdkTask} from "@kestra-io
 import * as FlowsAPI from "@kestra-io/kestra-sdk/flows"
 import * as MetricsAPI from "@kestra-io/kestra-sdk/metrics"
 import {defaultNamespace} from "../composables/useNamespaces"
+import {useApiStore} from "./api"
+import {flowTaskStats, isExampleFlow, primaryTriggerType} from "../utils/analytics/activation"
 
 const textYamlHeader = {
     headers: {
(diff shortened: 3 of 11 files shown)


ui/src/stores/flow.ts
```diff
@@ -543,10 +545,29 @@ export const useFlowStore = defineStore("flow", () => {
             localStorage.removeItem(`el-fl-creation-${creationId.value}`)
             creationId.value = undefined
 
+            if (!options.draft) {
+                trackFlowCreated(flow.value)
+            }
```
flcarre, line 550: `restoreFlow()` in `ui/src/override/components/flows/Actions.vue` calls `flowStore.createFlow({flow: YAML_UTILS.deleteMetadata(...)})` with no `draft` key. `options.draft` is therefore `undefined`, `!undefined` is `true`, and this gate fires `FLOW_CREATED` every time a user restores a soft-deleted flow.

Restoring a deleted flow is not a new-flow creation for activation purposes, so this inflates the exact metric the PR is trying to make trustworthy.

The gate keys off the wrong signal: `draft` distinguishes autosave from a real save, not restore from create. Either give `createFlow()` an explicit opt-out that `restoreFlow()` passes, or move the `trackFlowCreated()` call out of the store and into the one genuine "user created a new flow" call site.

ui/src/components/flows/FlowRun.vue
```diff
@@ -399,6 +395,11 @@
                     return
                 }
 
+                apiStore.posthogEvents({
+                    type: "FLOW_EXECUTION",
+                    action: "submit",
+                })
```
flcarre, line 404: This fires unconditionally, before the branch below chooses between `props.replaySubmit` and `executeTask`.

`props.replaySubmit` is wired only from `ReplayWithInputs.vue` to `executionsStore.replayExecutionWithInputs`, which never reaches `triggerExecution` and therefore never emits `executed`. So every replay-with-inputs submission emits an `app.execute-modal.submitted` that can never resolve into a matching `app.flow.executed`.

That is not just a missing event, it silently depresses the submit to executed conversion rate, which is the number this PR exists to make correct. Two ways out: track the replay path as well, or add a discriminator property here (`is_replay`, or reuse `location`) so replay submits can be excluded from the funnel.

ui/src/components/secrets/SecretsTable.vue
```diff
@@ -564,6 +566,13 @@
 
             actionMethod({namespace: secret.value?.namespace as string, secret: secretData})
                 .then(() => {
+                    apiStore.posthogEvents({
+                        type: secret.value?.update === true ? "SECRET_UPDATED" : "SECRET_CREATED",
+                        secret_type: "secret",
+                        namespace: secret.value?.namespace,
+                        has_tags: (secretData.tags?.length ?? 0) > 0,
+                    })
```
flcarre: `secret` is a single ref mutated in place by `updateSecretModal()`, and `resetForm()` replaces `secret.value` wholesale whenever the drawer closes. Nothing disables the drawer's Cancel or X while `actionMethod(...)` is in flight, so if the user closes it before the response lands, this `.then()` reads `update` off a different object than the one that was submitted and reports an update as a create.

`KVTable.vue` in this same PR already does the right thing (`const wasUpdate = kv.value.update === true` before the request). Worth matching so the two files behave identically:

```suggestion
            const wasUpdate = secret.value?.update === true

            actionMethod({namespace: secret.value?.namespace as string, secret: secretData})
                .then(() => {
                    apiStore.posthogEvents({
                        type: wasUpdate ? "SECRET_UPDATED" : "SECRET_CREATED",
                        secret_type: "secret",
                        namespace: secret.value?.namespace,
                        has_tags: (secretData.tags?.length ?? 0) > 0,
                    })
```

flcarre (commented):
Reviewed together with the EE companion. The core mechanism is good: splitting `open_modal` / `submit` / `executed` through `SPLIT_EVENT_RESOLVERS` is the right call, the naming follows `surface.object.action` cleanly, and `activation.spec.ts` is real coverage rather than tautology.

Two things I verified for you, so nobody has to worry about them:

- **Nothing regresses to the un-namespaced fallback.** In production, `flow_execution` / `app.flow.executed` has only ever carried `action` values `open_modal` and `submit`, and both are in the new map. I also grepped both editions: there is no third `action` value emitted anywhere.
- **No saved PostHog insight or dashboard** references `app.flow.executed` or `flow_execution`, so the semantic change breaks nothing downstream.

Below are the things I would like addressed before this lands. Details inline.

### Confirmed bugs

1. **`restoreFlow()` now fires `app.flow.created`.** Restoring a soft-deleted flow calls `createFlow()` without a `draft` key, so the new gate passes. That pollutes exactly the metric this PR is making trustworthy. See the inline note on `ui/src/stores/flow.ts`.
2. **The submit/executed pair is asymmetric for replay-with-inputs**, which makes the funnel conversion rate unreliable. See the inline note on `FlowRun.vue`.
3. **EE: service-account token creation is untracked.** `authStore.createApiToken` is instrumented, but `UserApiTokens.vue` and `ServiceAccountEdit.vue` both go through `serviceAccountsStore.createApiToken` (`ui-ee/src/stores/serviceAccounts.ts`), which emits nothing. Service accounts are a first-class EE feature, so `secret_type: "token"` will undercount. Commented on the EE PR too.
4. **`SecretsTable.vue` reads the created/updated flag after the request resolves**, while `KVTable.vue` in this same PR correctly snapshots it before. One-line fix suggested inline.

### Open question, worth a decision rather than a fix

`triggerExecution` has exactly three callers (`submitTask.ts`, `playground.ts`, `useFlowEditorActions.ts`). Every other path that creates an execution bypasses it and stays invisible to `app.flow.executed`:

- single Replay / Restart (`Restart.vue`)
- all bulk actions (`Executions.vue`: bulk replay, restart, force-run)
- the playground breakpoint-resume branch (`playground.ts`, `replayOrTriggerExecution`), which is precisely the open question the issue raised. Only the fresh-run half is answered.

This may well be deliberate, that only fresh runs count as activation. If so, could you note it in the PR description or the tracking plan? Right now it reads as an oversight, and the next person to touch this will not know which it was.

### Design concerns

- **`app.secret.created` conflates three unrelated objects** (KV entries, namespace secrets, EE API tokens), separated only by `secret_type`. Inline note on `eventNaming.ts`.
- **`primaryTriggerType` leaks unbounded cardinality.** Suggestion inline.
- **`EXAMPLE_NAMESPACE` duplicates the existing `TUTORIAL_NAMESPACE`.** Suggestion inline.

### Not addressed here

The DoD item "the app calls `identify()` with the id carried on the console onboarding deeplink" is in neither PR. `identify()` is still only called with `ensureUid()` at `ui/src/utils/posthog.ts:169`. Follow-up, or out of scope for this one?

No PII issues found. The invite tracker's "never send the email" holds, `role` is an RBAC role name, the rest are booleans.

ui/src/utils/analytics/activation.ts
```diff
@@ -0,0 +1,68 @@
+// Namespace the "Getting Started" blueprints are auto-loaded into (see FlowAutoLoaderService).
+const EXAMPLE_NAMESPACE = "tutorial"
+
+export function isExampleFlow(namespace: string | undefined): boolean {
+    return namespace === EXAMPLE_NAMESPACE
+}
```
flcarre: This constant already exists as `TUTORIAL_NAMESPACE` in `ui/src/utils/constants.ts:70`, consumed by `ui/src/stores/core.ts` and `ui/src/utils/welcomeGuard.ts` for the same "is this the getting-started namespace" check. Two copies will drift.

Separately, the comment does not hold up: I read `FlowAutoLoaderService.java` and it never hardcodes `"tutorial"`. It pulls blueprints tagged `getting-started` from the blueprints API and takes each namespace from the blueprint's own YAML, with the one special-cased blueprint rewritten to `systemFlowsConfiguration.namespace()`. The value is right, the citation is not.

```suggestion
import {TUTORIAL_NAMESPACE} from "../constants"

export function isExampleFlow(namespace: string | undefined): boolean {
    return namespace === TUTORIAL_NAMESPACE
}
```

ui/src/utils/analytics/activation.ts
```diff
@@ -0,0 +1,68 @@
...
+ * How the flow is meant to be started: `manual` with no trigger, otherwise the first trigger's kind
+ * (short class name, mapped to the funnel vocabulary where one exists).
+ */
+export function primaryTriggerType(triggers: {type?: string}[] | undefined): string {
+    const first = triggers?.[0]?.type
+    if (!first) return "manual"
+
+    const shortName = (first.split(".").pop() ?? first).toLowerCase()
+    return TRIGGER_KINDS[shortName] ?? shortName
+}
```
flcarre, line 72: Only `schedule` is bucketed. `webhook` and `flow` happen to come out clean, but that is a coincidence of their class names rather than something the map guarantees, and the tests only cover those three lucky cases.

Everything else falls through to the raw lowercased short class name, so `io.kestra.plugin.aws.sqs.RealtimeTrigger` becomes `"realtimetrigger"` and every plugin trigger in the ecosystem mints a new distinct PostHog property value. That is unbounded cardinality on a property meant to be a curated funnel dimension.

Making the known set explicit and bucketing the rest keeps the existing tests green:

```suggestion
const TRIGGER_KINDS: Record<string, string> = {
    schedule: "cron",
    webhook: "webhook",
    flow: "flow",
}

/**
 * How the flow is meant to be started: `manual` with no trigger, otherwise the first trigger's kind
 * mapped to the funnel vocabulary. Anything outside the known set is bucketed as `other`, so plugin
 * triggers cannot mint unbounded PostHog property values.
 */
export function primaryTriggerType(triggers: {type?: string}[] | undefined): string {
    const first = triggers?.[0]?.type
    if (!first) return "manual"

    const shortName = (first.split(".").pop() ?? first).toLowerCase()
    return TRIGGER_KINDS[shortName] ?? "other"
}
```

ui/src/utils/analytics/eventNaming.ts
```diff
@@ -1,6 +1,9 @@
 // Taxonomy source of truth: kestra-io/data#354
 const FLAT_EVENT_NAMES: Record<string, string> = {
-    flow_execution: "app.flow.executed",
+    flow_created: "app.flow.created",
+    secret_created: "app.secret.created",
+    secret_updated: "app.secret.updated",
```
flcarre: These two names end up covering three unrelated product objects: a KV Store entry (`KVTable.vue`, `secret_type: "kv"`), a namespace Secret (`SecretsTable.vue`, `secret_type: "secret"`), and an EE personal API token (`override/stores/auth.ts`, `secret_type: "token"`). Different pages, different permissions, different mental models.

The practical cost: anyone building "secrets created" on this event silently counts KV entries and API tokens unless they remember the `secret_type` filter every single time. That is the kind of thing that is discovered six months later in a number someone already presented.

This file already has the mechanism to avoid it. `SPLIT_EVENT_RESOLVERS` dispatches `flow_execution` off `action`; the same pattern off `secret_type` gives `app.kv.created`, `app.secret.created`, `app.token.created`. Cheap now, a data migration later.

flcarre:
LGTM ! Just remove the useless comment once the review is done

flcarre (approved):

PR #17744, opened 2026-07-31
fix(worker): use last CREATED history for worker.queued.duration metric

### LINKED ISSUE #17774: worker.queued.duration metric includes retry interval, making it unusable for Worker queue monitoring
## Context

The `worker.queued.duration` metric is described as _"Task queued duration inside the Worker"_. We'd like to use this metric to monitor the **actual scheduling latency inside the Worker** — i.e., the time a task spends waiting in the Worker's internal queue before being picked up for execution.

However, in retry scenarios, the metric reports inflated values (e.g., 5+ minutes) because `State.getStartDate()` always returns the very first CREATED timestamp, not the time the task was most recently enqueued.

## Problem

When a task is retried, the TaskRun's state histories look like:

```
CREATED   T+0s      ← original creation
RUNNING   T+0s
FAILED    T+0s
RETRYING  T+0s
CREATED   T+5m      ← re-enqueued after retry interval (e.g. PT5M)
RUNNING   T+5m+0.1s ← Worker picks up the task
```

The current code:
```java
Duration.between(workerTask.getTaskRun().getState().getStartDate(), Instant.now())
```

`getStartDate()` returns the first CREATED time (T+0s), so the recorded duration is ~5 minutes. But the task only spent ~0.1s actually queued inside the Worker after the retry.

## Why this matters

1. **Metric semantics**: The description says "Task queued duration **inside the Worker**". The retry interval (PT5M) is an Executor-side scheduling delay, not Worker queue backpressure. Mixing them conflates two different signals.

2. **Monitoring & alerting**: We want to use this metric to detect Worker capacity issues (queue saturation, thread starvation). If retried tasks inflate the metric by minutes, it triggers false alerts and makes the metric unreliable for capacity monitoring.

3. **Consistency**: A task that retries 3 times with PT5M intervals would report ~15 minutes of "queue duration" even if the Worker picked it up within milliseconds each time.

## Proposal

Use the last CREATED entry in the TaskRun's state histories as the queue start time:

```java
Instant queuedSince = workerTask.getTaskRun().getState().getHistories().stream()
    .filter(h -> h.getState() == CREATED)
    .reduce((first, second) -> second)
    .map(State.History::getDate)
    .orElse(workerTask.getTaskRun().getState().getStartDate());
```

This way:
- **Normal (first-run) tasks**: single CREATED entry, behavior unchanged
- **Retried tasks**: uses the most recent CREATED timestamp, reflecting actual Worker queue wait time

## Alternative

If keeping the current behavior is intentional for overall task lifecycle visibility, perhaps a separate metric (e.g., `worker.queued.duration.actual`) could be introduced that excludes retry intervals, so users can monitor Worker internal scheduling latency independently.

## Related PR

#17744 (submitted a fix, but happy to discuss the approach first)

CauchyHu:
Closes https://github.com/kestra-io/kestra/issues/17774.

## Summary

Fix `worker.queued.duration` metric reporting inflated values in retry scenarios.

**Root cause:** `State.getStartDate()` always returns the very first CREATED timestamp from when the TaskRun was originally created. In retry scenarios, the TaskRun's state histories contain multiple CREATED entries (one per retry cycle), but the metric was using the original one — causing durations of 5+ minutes instead of the actual queue wait time (seconds).

**Fix:** Find the last CREATED entry in the TaskRun's state histories, which represents when the task was most recently enqueued.

## Reproduction

Flow with retry configured:
```yaml
id: fail_then_retry
namespace: company.test
tasks:
  - id: force_fail
    retry:
      type: constant
      interval: PT5M
      maxAttempts: 3
      maxDuration: PT100H
    errorMessage: "Force fail to trigger retry"
```

After retry, the TaskRun state histories look like:
```
CREATED  T+0s    ← original creation
RUNNING  T+0s
FAILED   T+0s
RETRYING T+0s
CREATED  T+5m    ← re-enqueued after retry interval
RUNNING  T+5m    ← worker picks up
```

**Before fix:** `worker.queued.duration` = ~5 -10min (uses first CREATED at T+0s)  
image
**After fix:** `worker.queued.duration` = ~18s (uses last CREATED at T+5m)
image

## Test plan

- [x] Verified with production data: retry scenario correctly reports queue duration in seconds instead of minutes
- [x] Normal (first-run) scenario: single CREATED entry, behaves identically to before
- [x] Compiles successfully against develop branch


loicmathieu (commented):
This is highly opinionated, if you retry a taskrun, it's normal for me that the duration includes the retry.

By the way, we use the same calculation for retry at the flow level.

I think this is the kind of things that are better discussed first inside an issue so please open an issue with the rational of why we should do that.

CauchyHu:
> This is highly opinionated, if you retry a taskrun, it's normal for me that the duration includes the retry.
> 
> By the way, we use the same calculation for retry at the flow level.
> 
> I think this is the kind of things that are better discussed first inside an issue so please open an issue with the rational of why we should do that.

@loicmathieu I’ve created an issue with detailed explanations of the motivation and rationale for this change. Feel free to review it and leave your comments. link:  https://github.com/kestra-io/kestra/issues/17774

PR #17522, opened 2026-07-20
fix(ui): handle missing dashboard defaults

9904099:
### ✨ Description

- Falls back to the bundled dashboard when the optional default-dashboard settings endpoint returns 404.
- Prevents an expected, locally handled 404 from replacing the Flow Overview with the global 404 page.
- Adds regression coverage for a Flow Overview URL carrying `filters[timeRange][EQUALS]=PT24H` and for centralized 404 handling.

### 🔗 Related Issue

Fixes https://github.com/kestra-io/kestra/issues/17368

### 🎨 Frontend Checklist

- [x] Code builds without errors (`npm run build`)
- [ ] All existing E2E tests pass (`npm run test:e2e`) - not run; this environment has no Kestra backend/browser runtime.
- [ ] Screenshots or video recordings attached showing the `UI` changes - behavior-only error handling change; the issue Jam captures the regression.

### 📝 Additional Notes

Tests run:
- `npm run test:unit -- tests/unit/stores/dashboardDefaults.spec.ts tests/unit/utils/kestraHttp.spec.ts`
- `npm run check:types`
- `npm run test:lint`
- `npm run build`

### 🤖 AI Authors

Cat joke: Why did the cat sit on the dashboard? It wanted to monitor the purr-formance metrics.

PR #17753, opened 2026-07-31
fix(dashboards): fall back to the bundled dashboard when a reference is dangling

### LINKED ISSUE #17752: Overview shows the global 404 page when a dashboard reference is dangling
### Describe the issue

When a dashboard id is still referenced somewhere but the dashboard itself no longer exists, the Flow Overview (and the Namespace Overview, and the Home dashboard) renders the global **"Page not found"** screen instead of falling back to the dashboard bundled in the UI.

A dashboard id outlives its dashboard in two ordinary situations:

- the user selected a dashboard on that page, which persists it in `localStorage` under `userDashboard/<tenant>/<routeFamily>`, and that dashboard is deleted afterwards
- an admin set a default dashboard (`defaultFlowOverviewDashboard` / `defaultNamespaceOverviewDashboard` / `defaultHomeDashboard`) and that dashboard is deleted afterwards

Once that happens the page is stuck: reloading does not help, because the dangling id is read again on every visit.

### Steps to reproduce

1. Start a fresh OSS instance and run the `tutorial/hello-world` flow once, so the Overview has data.
2. Simulate the dangling reference (this is what a deleted dashboard leaves behind):
   ```js
   localStorage.setItem("userDashboard/main/flows/update", "deleted-dashboard-id")
   ```
3. Open `/ui/main/flows/edit/tutorial/hello-world/overview`.

Observed: the whole page is replaced by the 404 screen. The network tab shows `GET /api/v1/main/dashboards/deleted-dashboard-id → 404`.

Expected: the Overview renders, falling back to the dashboard bundled in the UI.

### Root cause

Two independent problems compound.

`useDashboardStore.load()` already catches the 404 and returns `undefined`, but that catch never gets the chance to matter: the global error interceptor in `ui/src/utils/kestraHttp.ts` runs first and sets `coreStore.error = 404`, and `DefaultLayout.vue` then swaps the whole page for the 404 screen. Opting the request out with `showMessageOnError: false` is what lets a locally handled 404 stay local. The same applies to `generate()` (`POST /dashboards/{id}/charts/{chartId}`), which has the same local catch and the same missing opt-out.

Then, even with the interceptor silenced, `Dashboard.vue`'s `load()` only logs a warning and shows an error toast when the dashboard is missing. It never falls back, so the page renders an empty dashboard.

### Environment

- Kestra Version: `develop` (reproduced on the `kestra/kestra:develop` image built 2026-07-31)

flcarre:
### ✨ Description

A dashboard id can outlive the dashboard it points to, either from a stale `localStorage` entry (`userDashboard/<tenant>/<routeFamily>`, written when the user picks a dashboard) or from a tenant default whose dashboard was deleted afterwards. In that case the Flow Overview rendered the global **"Page not found"** screen instead of falling back to the dashboard bundled in the UI, and reloading did not help because the dangling id is read again on every visit.

Two things were needed:

- **`ui/src/stores/dashboard.ts`** — `load()` and `generate()` already caught the 404 locally, but that catch never mattered: the global error interceptor in `ui/src/utils/kestraHttp.ts` runs first and sets `coreStore.error = 404`, and `DefaultLayout.vue` then swaps the whole page. Both requests now pass `showMessageOnError: false`, which is the existing mechanism for an expected, locally handled 404 (same pattern as `useApplyDraft.ts`).
- **`ui/src/components/dashboard/Dashboard.vue`** — with the interceptor silenced, the missing-dashboard branch only warned and showed a toast, leaving an empty dashboard. It now falls back to the bundled definition, keeping the toast so the user still learns the configured dashboard is gone.

### 🔗 Related Issue

Fixes https://github.com/kestra-io/kestra/issues/17752

### 🧪 How it was verified

Against the `kestra/kestra:develop` image built 2026-07-31, with a dangling id planted in `localStorage` and the same `GET /dashboards/deleted-dashboard-id → 404` occurring in both runs:

| | Page body |
|---|---|
| before | `/ Page not found Workspace Dashboards …` |
| after | `/ Flows / tutorial / hello-world Overview Execute …` |

Unit tests cover both the opt-out being passed and non-404 errors still propagating.

### 🎨 Frontend Checklist

- [x] Code builds without errors (`npm run check:types`, `eslint` clean on touched files)
- [x] Targeted unit tests pass (`dashboardDanglingId.spec.ts`, plus the existing dashboard specs, 36 passing)
- [ ] All existing E2E tests pass (`npm run test:e2e`) — not run locally
- [x] Behaviour verified in a browser against a real backend

### 📝 Additional Notes

Found while re-reviewing #17522, which targeted a different endpoint (`/dashboards/settings/default-dashboards`) that always returns 200 in OSS and therefore could not produce this symptom.

### DIFF
--- a/ui/src/components/dashboard/Dashboard.vue
+++ b/ui/src/components/dashboard/Dashboard.vue
@@ -171,6 +171,7 @@
                     title: err,
                     message: err,
                 }
+                await useDefaultDashboardBundledInUI()
             }
         }
 
--- a/ui/src/stores/dashboard.ts
+++ b/ui/src/stores/dashboard.ts
@@ -205,10 +205,12 @@ export const useDashboardStore = defineStore("dashboard", () => {
         return false
     }
 
+    const silent = {showMessageOnError: false} as Parameters<typeof DashboardsAPI.dashboard>[1]
+
     async function load(id: Dashboard["id"]) : Promise<Dashboard | undefined> {
         let data
         try{
-            data = await DashboardsAPI.dashboard({id}) as Dashboard
+            data = await DashboardsAPI.dashboard({id}, silent) as Dashboard
         } catch {
             return undefined
         }
@@ -242,7 +244,10 @@ export const useDashboardStore = defineStore("dashboard", () => {
 
     async function generate(id: Dashboard["id"], chartId: Chart["id"], parameters: ChartFiltersOverrides) {
         try {
-            return await DashboardsAPI.dashboardChartData({id, chartId, ...parameters} as globalThis.Parameters<typeof DashboardsAPI.dashboardChartData>[0])
+            return await DashboardsAPI.dashboardChartData(
+                {id, chartId, ...parameters} as globalThis.Parameters<typeof DashboardsAPI.dashboardChartData>[0],
+                silent as globalThis.Parameters<typeof DashboardsAPI.dashboardChartData>[1],
+            )
         } catch (e: any) {
             if (e.status === 404) return undefined
             throw e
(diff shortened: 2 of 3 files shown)


elevatebart (approved):
LGTM

Issue #16579, opened 2026-06-05
Concurency breaks when subflow being killed


NovytskyiD:
### Describe the issue

Subflow

```
id: termite_707115
namespace: company.team

sla:
  - id: maxDuration
    type: MAX_DURATION
    duration: PT1M
    behavior: FAIL
    labels:
      sla: miss
      reason: durationExceeded

tasks:
  - id: wait5MinForOtherShards
    type: io.kestra.plugin.core.flow.Sleep
    duration: "PT5M"
```

Parent flow

```
id: chinchilla_932032
namespace: company.team

concurrency:
  limit: 1
  behavior: QUEUE

tasks:
  - id: subFlow
    type: io.kestra.plugin.core.flow.Subflow
    flowId: termite_707115
    namespace: company.team
```

Run multiple copies of the parent flow - only the first execution will respect the concurrency limit, then when executions start to fail, more than 1 copy will be in Running status

<img width="1629" height="903" alt="Image" src="https://github.com/user-attachments/assets/28358196-5763-491c-9942-b3d660cf266d" />

<img width="1606" height="896" alt="Image" src="https://github.com/user-attachments/assets/53959d6c-60d6-4a09-8e6e-5b6d70a97914" />

<img width="1675" height="895" alt="Image" src="https://github.com/user-attachments/assets/5b39c41e-9ee4-4165-ab6f-9333c640fa1b" />

### Environment

- Kestra Version: 1.2.6 K8s + PG

loicmathieu:
@NovytskyiD Can you update to the latest 1.3 release as there was improvements in this area?

NovytskyiD:
@loicmathieu I will try to push an update on our side, but it can take some time.

Pradumnasaraf:
Working as expected now. QA done ✅. Cc @NovytskyiD

PR #17759, opened 2026-07-31
Fix/concurrency limit release

loicmathieu:
A flow concurrency limit could get stuck with a non-zero running count while
nothing was actually running, so with the CANCEL behavior every later execution
was cancelled for good. The counter decrement clamps at zero, so releasing one
slot too many heals itself while releasing one too few never does, which is why
the symptom is always a count stuck too high.

Four independent ways a slot was never returned, two of them deterministic:

- ConcurrencySlotReleaseProcessor inferred "this execution never held a slot"
  from its state history. An execution the gate admitted and that then
  terminated while still CREATED, which an EXECUTION_ASSERTION SLA does in the
  very same handler pass, has exactly the state history of an execution the gate
  rejected, so its slot was never given back.
- The kill guard required the state at cycle entry to be KILLING. For a flow
  that combines concurrency with afterExecution, isTerminated() stays false
  while afterExecution task runs are pending, so the KILLING to KILLED cycle
  does not release and the later cycle that truly ends the execution is
  discarded as a duplicate.
- Only the KILLED path was deduplicated, so a duplicate terminal cycle on an
  already FAILED execution released a second slot and admitted one execution too
  many (GitHub #16579).
- A release failing on a database error escaped toExecution, reached the queue
  subscriber fatal error handling and shut the instance down. The redelivered
  terminal execution never reaches the release again, so the slot stayed held.

The decision now rests on two facts the data already carries: the concurrency
scopes stamp that the admission gate writes on the execution metadata, and
whether this cycle is the one that terminated the execution. The former state
history heuristics are kept as a fallback for executions admitted before the
stamp existed. ExecutorContext keeps the execution as loaded at cycle entry only
when it was already terminal then, since any other entry answers the question on
its own and nothing needs to be retained.

The release also runs first in the terminal block, because a QueueException
raised by the subflow or loop emission would otherwise skip it and leak the slot
with no way back, and it never propagates: it logs and increments
executor.concurrency.slot.release.failed.count instead.

Fixed along the way:

- The deadlock predicate never tested the thrown exception itself and ignored
  suppressed exceptions and getNextException(), so a deadlock could escape
  unretried.
- The concurrency limit row is locked by primary key instead of through the
  non-unique flow index, which holds the very same value.
- FlowService deleted a concurrency limit row only on the exact revision that
  dropped the concurrency block, leaving orphan rows that kept a stuck count
  alive. Removal is now absolute, while creation stays a transition so its blind
  upsert cannot reset the count of the executions in flight.
- The flow Concurrency tab rendered nothing at all for a flow that no longer
  declares a concurrency block, which is precisely when a stuck count was
  invisible outside the superadmin pages. It now surfaces a leftover limit that
  still holds slots.

The two new runner tests fail without this change. flowConcurrencySlaFailSubflow
loses its flaky tag: it was failing often in CI on the duplicate terminal cycle
this change now rejects, and it passed 20 consecutive runs on H2 plus a full run
on each of the three backends.

Fixes https://github.com/kestra-io/kestra-ee/issues/9200


nkwiatkowski (approved):

loicmathieu:
Part of it (lock on ID, deadlock predicate) backported to 1.3

PR #17763, opened 2026-07-31
fix(flows): prevent tour dialog from covering highlighted content

RemiBarthe:
closes https://github.com/kestra-io/kestra-ee/issues/9617

This pull request makes minor adjustments to the onboarding tour scenes to improve the UI experience by updating the tooltip placement for certain steps.

- Onboarding Tour Scene Placement Updates:
  * Set the `placement` property to `"left"` for the `flow_generated` tour scene, ensuring the tooltip appears on the left side of the target element.
  * Removed the `placement: "left"` property from a later tour scene targeting the expression debugger, allowing it to use the default placement.


anna-geller (approved):
We can merge as is, but for debug expressions, it was deliberate to put it to the left, but no strong opinion. We can move it to the right. LGTM!

PR #17771, opened 2026-07-31
fix(flows): surface flow editor setup failures instead of a blank page

elevatebart:
### ✨ Description

`FlowCreate` called `setupFlow()` without awaiting or catching it, so a rejection became an unhandled promise rejection and the page stayed empty for good — with no error, no logging path a user would find, and no way to retry.

Two independent gates make that blank page total:

- `<MultiPanelFlowEditorView v-if="flowStore.flow" />` (`ui/src/components/flows/FlowCreate.vue:8`) — no flow, no editor.
- The Save split button is `v-if="isEditTab && editorIsAllowedEdit && !deleted"` (`ui/src/override/components/flows/Actions.vue:65`), and `flowStore.isAllowedEdit` (`ui/src/stores/flow.ts:828`) returns `false` while `!flow.value`.

So a single failure in `setupFlow` removes both the editor and the Save button, while the route has already committed to `/flows/new`.

This PR does **not** fix whatever throws. It makes the failure visible instead of silent: `setupFlow` is now awaited inside an `initialize` wrapper that renders `KsAlert type="error"` with a retry control when it rejects, and logs the error to the console.

Notes on the approach:

- Top-level `await` in `<script setup>` was avoided deliberately — it would make the component async and require a `Suspense` boundary the router view does not provide. `initialize()` awaits `setupFlow()` internally and never rejects.
- The assignment of `flowStore.flow` stays synchronous on the default path (`setupFlow`'s body runs to completion before its first suspension point there), so the happy path and first-paint behaviour are unchanged.
- Added `data-test="flow-create-error"` and `data-test="flow-create-error-retry"` so the error state is addressable from Playwright.

### 🔗 Related Issue

No tracking issue. This came out of investigating intermittently failing E2E runs on `/flows/new`, where `getByRole("button", {name: "Save", exact: true})` and the Monaco editor content both time out on the same page — e.g. [run 30634632394](https://github.com/kestra-io/kestra/actions/runs/30634632394/job/91169610376) and the scheduled develop run [30634723529](https://github.com/kestra-io/kestra/actions/runs/30634723529). Worth stressing that the Save locator is **correct** — it resolves to a real `<button>` from the Element Plus split button inside `KsDropdown`, and identical commits both pass and fail across scheduled runs, so the locator is not the bug.

### 🎨 Frontend Checklist

- [ ] All existing E2E tests pass (`npm run test:e2e`) — **not run**: needs a running Kestra backend + Postgres, which I could not stand up in this environment. Please let CI cover it.
- [ ] Screenshots or video recordings attached showing the `UI` changes — **not attached**, same reason: reproducing the error banner needs the app running against a backend, plus an induced `setupFlow` failure.

Also run: `npx eslint src/components/flows/FlowCreate.vue` (clean) and `npm run check:types` (design-system, app and test projects all pass).

### 📝 Additional Notes

**Translations.** Two keys added, both under the existing `something_went_wrong` group, and propagated to all 13 non-English locale files. `retry` reuses each file's existing translated value for `no_code.fields.general.retry` verbatim; `opening_flow_editor` follows each file's existing `something_went_wrong.loading_execution` phrasing, keeping `flow` untranslated per the reserved-terms rule.

`npm run translations:check` still fails, but it **already fails on `develop`** — the pre-existing gaps are the `onboarding.tour.*`, `test_event.*` and `welcome_copilot.flows.ordersRevenueToSlack.*` keys awaiting the auto-translate workflow. I verified neither of my two new keys appears in the missing-key output.

**No unit test.** There is no component-level Vitest precedent in `ui/src/` (`src/utils/analytics/eventNaming.spec.ts` is the only spec there). Mounting `FlowCreate` would mean mocking the Pinia stores, the router, i18n, the `override/` aliases, `@kestra-io/topology` and `MultiPanelFlowEditorView` (which pulls in Monaco) — a brittle harness well out of proportion to a 20-line error-state change, and closer to a "gas factory" than the guidelines allow. The practical verification is the E2E suite: next time this trips in CI, the DOM snapshot will show the error banner and the console will name the throwing call, instead of an empty page that reads like a stale selector.

### 🤖 AI Authors

Raised by Claude Code. Why did the cat sit on the keyboard? It heard the flow editor had a `catch` block. 🐱

---
_Generated by [Claude Code](https://claude.ai/code/session_01M9RtNPN5yvKivQuRq1RjgK)_


brian-mulier-p (approved):
LGTM

flcarre:
Had a look at this. The direction is right, a permanently blank page with no signal is the worst possible outcome here. But I don't think the catch is placed where the failure actually happens, and there are a couple of things worth sorting before merge.

### The try/catch can't see the rejection on the path the flaky test hits

`setupFlow` calls `flowStore.initYamlSource()` without awaiting it (line 149). That returns `validateFlow(...)`, which is a bare `FlowsAPI.validateFlows(...).then(...)` with no catch anywhere in the chain, and it also fires `fetchGraph()` unawaited. So `setupFlow`'s promise settles as soon as those are invoked and their rejections stay unhandled, which is exactly the failure mode this PR sets out to close.

On the default `/flows/new` route with no query, which is what `tests/e2e/flow.spec.ts` goes through, those two are the only async work in `setupFlow`. Everything else reachable before the `flowStore.flow = {...}` assignment is synchronous and can't throw: `getNamespacesForAction` returns `[]` unconditionally in OSS (`override/stores/auth.ts`), `defaultNamespace()` only reads localStorage, `getRandomID()` is pure math, and the YAML parse is already try/caught locally. So for the CI flake you're chasing, the new banner won't render.

The fix isn't just adding `await` though: a 4xx from `validateFlows` is a soft failure, and escalating it to a page-level error would hide an editor that should have opened. Probably wants its own catch that keeps the editor up and reports inline.

### On the blueprint path the global handler gets there first

`blueprints.ts` uses `useClient()`, so those calls go through `handleErrorCentrally`. A 404 sets `coreStore.error`, and since `flows/create` declares no `meta.layout`, `App.vue` renders it inside `DefaultLayout`, whose slot is the `v-else` of `<Errors v-if="coreStore.error">`. The whole subtree gets replaced, so the banner is unreachable, and a stale or deleted `blueprintId` is probably the most plausible real trigger. On a 5xx with a body it's the opposite: `coreStore.message` fires `ErrorToast` and the inline alert repeats the same string. Passing `showMessageOnError: false` on the blueprint calls made from this path would let the component own the error.

### Loading state

The template only branches on `setupError` and `flowStore.flow`, so the page is blank while the blueprint fetch is in flight, and blank again during a retry since `initialize()` clears `setupError` first. Same symptom, shorter window, and it makes Retry feel dead on exactly the slow network that caused the failure. A `KsSkeleton` branch plus `:loading` on the retry button would cover it.

### Retry can silently swap the onboarding flow

`sessionStorage.removeItem(ONBOARDING_FLOW_PRESET_KEY)` runs at line 111, before the flow is committed at line 142. Anything throwing in between and a retry falls through to the generic template instead of the tour flow, with nothing telling the user. Not reachable today, but it's a trap for whoever touches this function next. Cheaper to consume the preset only once setup has succeeded.

### Smaller things

- `> section.full-container > * { flex: 1 }` in `app.scss` makes `.flow-create-error` full height, so with `align-items: flex-start` the alert pins to the top left of an otherwise empty viewport. For a whole-page dead end the house pattern is a centered `KsEmptyState` with an action slot, which also lets the scoped style block go away.
- Raw `error.message` as the body will read "Network Error" to a user. Worth a translated description with the raw detail demoted, and since at that point they have neither an editor nor a Save button, a "start from a blank flow" escape hatch is probably worth more than Retry on its own.
- The seven line comment above `initialize()` is the kind of narrative comment `AGENTS.md` rules out. One line on the why is enough, the history belongs in the commit message.
- On the "no component test precedent" point: true of `ui/src`, but the unit tests live in `ui/tests/unit`, which has 119 specs and 46 that `mount()`. `tests/unit/override/components/flows/Actions.spec.ts` already mocks `useFlowStore`, `override/stores/auth` and `vue-router` in roughly the shape this would need, so the harness cost is lower than the description suggests. A spec that rejects `getFlowBlueprint`, asserts the error region, then asserts a successful retry renders the editor would also have caught the preset issue above.

Translations look good, all 13 locales got real values and `flow` is left untranslated throughout.

PR #17772, opened 2026-07-31
test: mock fetch layer for Storybook to intercept SDK API calls

elevatebart:
## ✨ Description

Adds a comprehensive fetch-layer API mock for Storybook that intercepts HTTP calls from the generated SDK and axios-based code, eliminating network errors and unhandled promise rejections during story development and testing.

**Problem:** Almost every component tree issues HTTP calls, but there's no backend behind Storybook. The previous `axios.defaults.adapter` stub only covered 4 files in `src/` that import axios directly. The generated SDK (from `packages/kestra-sdk`) and the axios-like facade both use `fetch`, which was escaping to the Vite dev server, causing:
- GET requests returning 200 with non-JSON bodies, leading to `Invalid prop` warnings
- POST requests 404'ing and throwing errors surfaced as `PromiseRejectionEvent`s
- Warnings appearing after stories finished rendering ("unknown test")

**Solution:** 
- Created `ui/.storybook/apiMock.js` that patches `window.fetch` before any SDK or `src/` modules are evaluated
- Provides a `HANDLERS` registry mapping API routes to mock payloads (keyed by `METHOD <path>`)
- Supports pattern matching for parameterized routes (`:name` segments and trailing `*`)
- Returns correctly-shaped empty payloads (arrays for lists, objects for single resources) so components degrade gracefully instead of throwing
- Logs unmocked routes once per story with the story name and URL, making it easy to trace blank-looking stories to missing mocks
- Exports `resolveApiRequest()` and `mockClientFallback()` for story-local mocks to fall back to

**Changes:**
- `preview.jsx`: Imports `apiMock` first, configures SDK client with mocked fetch, uses `useClient()` for stores, moves rejection reporter to `apiMock`
- `vitest.setup.js`: Imports `apiMock` first (install is idempotent)
- Updated story files to use `mockClientFallback()` for unmocked routes instead of returning empty objects
- `LowCodeEditor.stories.jsx`: Removed SDK mock, uses real fixture data, improved source YAML formatting

## 🔗 Related Issue

N/A

## 🎨 Frontend Checklist

- [x] Existing Storybook stories render without network errors
- [x] Unhandled promise rejections are properly reported with context
- [x] Stories can be written cheaply with automatic fallback to empty data

## 📝 Additional Notes

The fetch mock is installed at the earliest possible point (before any SDK or `src/` modules load) to ensure all API calls—whether from generated SDK code, the axios facade, or direct fetch calls—are intercepted. The pattern matching system allows flexible route definitions while maintaining specificity (exact keys win over patterns, more specific patterns win over general ones).

The `beginStoryScope()` function is called by Storybook's `beforeEach` hook to scope warnings to individual stories and clear the de-duplication record, so developers see which story needs a mock added rather than only the first story that hit an unmocked route.

https://claude.ai/code/session_01QJyb6CSvtgVh413xU3Yj6Z

### DIFF
--- a/ui/.storybook/apiMock.js
+++ b/ui/.storybook/apiMock.js
@@ -67,6 +67,7 @@ const HANDLERS = {
 
     // --- executions ---------------------------------------------------------------------------
     "GET /executions/search": {results: [], total: 0},
+    "GET /executions/namespaces/:namespace/flows/:flowId/average-duration": {avgDurationMs: null, count: 0},
     "GET /outputs/:executionId": [],
 
     // --- dashboards & misc --------------------------------------------------------------------
--- a/ui/.storybook/apiMock.js
+++ b/ui/.storybook/apiMock.js
@@ -1,0 +1,287 @@
+// Storybook/Vitest API double.
+//
+// WHY THIS EXISTS
+// ---------------
+// There is no backend behind Storybook, yet almost every component tree issues HTTP calls. Before
+// this module the only stub was `axios.defaults.adapter` (see preview.jsx), which covers just the 4
+// files in src/ that still import axios. Everything else is fetch-based:
+//
+//   * the generated SDK (packages/kestra-sdk/src/openapi/*) resolves
+//     `options.fetch ?? _config.fetch ?? globalThis.fetch` per request (openapi/client/client.gen.ts),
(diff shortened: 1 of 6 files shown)


brian-mulier-p (approved):
Why not, good idea

PR #17778, opened 2026-07-31
fix(ai): bound Copilot token spend and shrink the plugin-schema tool result

ammeek:
✨ Description

Two related cost-control changes to the AI Copilot backend: nothing bounded how much text a turn could push at the model, and one read tool was returning ~5× more than it needed to.

1. get-plugin-schema returns only the plugin's own properties (fix(plugins))

The tool returned the raw generated schema — a $ref into a definitions map holding every transitively referenced class — so the properties every task inherits from its base class (retry variants, assets, worker selector, cache) dominated the payload. For io.kestra.plugin.core.log.Log, a plugin with two properties, that was ~10 KB of JSON with under 10% describing the plugin itself, behind JSON pointers the model had to resolve itself.

It now builds the schema via ClassPluginDocumentation with allProperties = false, the same representation the plugin documentation API serves. Dropping the base-class properties also drops the definitions they pull in: 10,076 → 2,111 characters for that plugin. The result now also exposes the plugin's outputs and example YAML, which the old shape didn't surface at all.

Measured across four core plugins (characters of JSON):

| plugin | before | after |
|---|---|---|
| log.Log | 10,076 | 2,082 |
| flow.Subflow | 11,713 | 3,840 |
| trigger.Webhook | 9,863 | 5,895 |
| trigger.Schedule | 21,064 | 17,000 |


webserver/src/main/java/io/kestra/webserver/controllers/api/AiAgentController.java
```diff
@@ -134,15 +142,19 @@ public Flux<Event<Object>> chat(
...
         // its turn's trace, so confirming a parked action never counts against the cap.
-        if (threadManager.turnCount(tenant, threadId) >= maxTurnsPerThread) {
+        if (AiThreadManager.turnCount(log) >= maxTurnsPerThread) {
             throw new HttpStatusException(
                 HttpStatus.TOO_MANY_REQUESTS,
                 "This thread has reached its maximum of %d turns; start a new thread.".formatted(maxTurnsPerThread)
             );
         }
 
+        ensureContextWithinBudget(log, request.prompt());
```
fhussonnois, line 156: I think you should do the `ensureContextWithinBudget` before and withing the agent-loop on each tool call. Otherwise, a turn can call up to `maxSequentialToolsInvocations` each resending the accumulated messages for the thread.

webserver/src/main/java/io/kestra/webserver/services/ai/agent/internals/ContextSize.java
```diff
@@ -0,0 +1,67 @@
...
+     * projection rather than failing the guard on it.
+     */
+    private static int jsonLength(final Object value) {
+        if (value == null) {
+            return 0;
+        }
+        try {
+            return MAPPER.writeValueAsString(value).length();
+        } catch (Exception e) {
+            return 0;
```
fhussonnois, line 64: Let's log that exception in debug

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/ToolResultTooLargeException.java
```diff
@@ -0,0 +1,37 @@
...
+
+import io.kestra.core.exceptions.KestraRuntimeException;
+
+/**
+ * Thrown by {@link ToolCatalog#dispatch} when a tool returned more than {@code maxToolResultChars}.
+ * The result is deliberately not truncated: a cut-off result would leave the model reasoning on partial
+ * data without knowing it. The orchestrator converts this into a failed tool result, so the model can
+ * retry with a narrower request and the client can tell the user what happened.
+ */
+public class ToolResultTooLargeException extends KestraRuntimeException {
```
fhussonnois, line 11: getter methods seem to be dead-code

webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/ToolCatalog.java
```diff
@@ -187,12 +192,20 @@ public DispatchResult dispatch(final ToolExecutionRequest request, final AgentCa
         // Carry the caller context to the @Tool method as a langchain4j managed argument — the executor
         // injects it by type into the tool's AgentCallContext.Context parameter — not a thread-local.
         ToolExecutionResult executed = entry.executor().executeWithContext(stripEmptyOptionalArgs(entry, request), AgentCallContext.into(context));
+        ensureResultWithinBudget(entry.name(), executed.resultText());
```
fhussonnois, line 195: - `ToolExecutionResult executed = entry.executor().executeWithContext(...); ` // 194 — side effect already happened
- `ensureResultWithinBudget(entry.name(), executed.resultText());  `          // 195 — throws
- `ArtefactDraft artefact = executed.result() instanceof PublishableToolResult `... // 198 — never reached

The check runs after execution. On a write tool whose result exceeds the cap: the write landed, but the artefact is discarded, and `failTool` tells both the model and the user the call failed. The model's likely next move is to retry leading to a duplicate tool execution I guess.

So instead of the model reasoning on partial data as if complete, it reasons on "failed" when the work succeeded.

You can't know the size before producing it, so the fix isn't reordering and it's not routing this through `failTool`. The artefact must survive (sent to the user), and the model must be told the truth: the call succeeded, the result was too large to show you.

fhussonnois (changes requested):

PR #17782, opened 2026-08-02
feat(charts): allow setting progressDeadlineSeconds on Deployments

### LINKED ISSUE #17781: Helm chart: no way to set progressDeadlineSeconds on the Deployment
### Describe the issue

`charts/kestra` has no way to set `progressDeadlineSeconds` on the rendered Deployments.

`templates/deployment.yaml` templates `strategy` and `updateStrategy` from values:

```yaml
  {{- if eq $kind "Deployment" }}
  {{- with (default $common.strategy $content.strategy) }}
  strategy:
    {{- toYaml . | nindent 4 }}
  {{- end }}
  {{- end }}
```

but there is no `progressDeadlineSeconds` key and no generic spec passthrough, so a chart user cannot change it at all — not via `common`, not per deployment, and not from a parent chart that depends on this one.

### Why it matters

`progressDeadlineSeconds` is a watchdog on the current rollout: if no new replica becomes ready within it, the Deployment controller sets `Progressing=False` with reason `ProgressDeadlineExceeded`. It aborts nothing — the rollout continues and eventually succeeds — but the condition is what other tools read.

With a large image, a cold pull on a node without the layers cached routinely runs past the Kubernetes default of **600 s**. Pulling is not "progress" as far as the controller is concerned, so the condition gets set mid-download, and two common things then report a failure that is not one:

- **ArgoCD** derives Deployment health from exactly those conditions, so the application shows `Degraded` for the remainder of the pull, and anything gating on application health sees a failed sync.
- **`kubectl rollout status`** returns immediately instead of blocking, so a deploy script that waits on it concludes the rollout finished while the image is still downloading.

We hit this with a Kestra image of ~3.2 GB (stock Kestra plus a plugin), where a first pull is regularly over ten minutes. The only workaround available to a chart consumer is to patch the field onto the Deployment after the fact, which needs a hook, RBAC and a pod — a lot of machinery for a field the chart could simply template.

### Proposed change

Add it next to `strategy`, following the existing `default $common.X $content.X` pattern so it can be set globally or per deployment, rendered only in the `Deployment` branch since a StatefulSet has no such field, and unset by default so nothing changes for existing users.

Happy to open the PR — the change is three lines in `deployment.yaml` plus the values entry and a test suite.

### Environment

- Kestra chart version: 1.3.30 (also checked against `develop`)

USchulzeEyssingBK:
### ✨ Description

Adds `progressDeadlineSeconds` to the Helm chart's Deployment template. Today `templates/deployment.yaml` templates `strategy` and `updateStrategy` from values and nothing else, with no generic spec passthrough, so a chart user cannot set this field at all.

It matters for large images. The field is a watchdog on the current rollout: if no new replica becomes ready within it, the controller sets `Progressing=False` with reason `ProgressDeadlineExceeded`. It aborts nothing, but a cold image pull that runs past the 600 s default trips it mid-download, and tooling reads that as a failed rollout — ArgoCD derives Deployment health from exactly those conditions and reports the application `Degraded`, and `kubectl rollout status` returns immediately instead of blocking, so anything gating on it concludes the rollout finished while the image is still downloading. We see this with a ~3.2 GB Kestra image where a first pull is regularly over ten minutes.

The change follows the conventions already in the file:

```yaml
  {{- with (default $common.progressDeadlineSeconds $content.progressDeadlineSeconds) }}
  progressDeadlineSeconds: {{ . }}
  {{- end }}
```

- same `default $common.X $content.X` pattern as `strategy` and `revisionHistoryLimit`, so it can be set for all deployments or per deployment;
- inside the `if eq $kind "Deployment"` branch only — a StatefulSet has no such field;
- **unset by default**, so nothing changes for existing users. The diff is additions only.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/17781

### 🛠️ Backend Checklist

### 📝 Additional Notes

Verified with `helm unittest` against `develop`. The new suite covers all four behaviours:

```
 PASS  Deployment progressDeadlineSeconds	tests/deployment_progress_deadline_test.yaml
Tests:       4 passed, 4 total
```

- not rendered when unset (the Kubernetes default stays in place)
- `common.progressDeadlineSeconds` applies globally
- a deployment-level value overrides `common`
- not rendered on a `StatefulSet`

Also confirmed by rendering: `helm template` produces no field when unset, `1800` from `common`, `3600` when the deployment overrides it, and nothing at all with `common.kind=StatefulSet`.

One thing to flag, unrelated to this PR: `tests/jvm_active_processor_test.yaml` currently fails on a clean `develop` checkout (4 tests, expecting `KESTRA_JAVA_OPTS` where the template renders `_JAVA_OPTIONS`). I left it alone rather than fold an unrelated fix into this change — happy to open a separate issue if that is not already known.

`README.md` is updated for the helm-docs values table.

### 🤖 AI Authors

This PR was prepared with Claude Code, so — as the template asks — a cat joke:

> Our cat kept sitting on the keyboard during the rollout, so we raised `progressDeadlineSeconds`. Turns out she wasn't slow, she was just waiting for the purr-fect moment to become ready. 🐱


loicmathieu:
@Skraye I think this one falls into your responsability

charts/kestra/values.yaml
```diff
@@ -40,6 +40,11 @@ common:
   strategy: {}
   # -- StatefulSet update strategy.
   updateStrategy: {}
+  # -- Seconds a Deployment rollout may make no progress before the controller
```
MarcoSaba: Nit: this renders as `string` / `""` in the README table via helm-docs, but the field is semantically an int (seconds) — quoting a value like `"1800"` would work by accident but isn't the intended type. helm-docs supports an explicit type override via `# -- (type)`, which fixes the README without changing the falsy-`""`-skip behavior the template relies on.

```suggestion
  # -- (int) Seconds a Deployment rollout may make no progress before the controller
```

MarcoSaba (approved):
Follows existing chart conventions exactly (`strategy`/`revisionHistoryLimit` pattern), good test coverage, additive/unset-by-default so no behavior change. Nit fixed. LGTM.

PR #17784, opened 2026-08-03
fix(core): bucket metric aggregation in UTC

### LINKED ISSUE #17674: AbstractJdbcRepository.getDate() reconstructs Instant using JVM local timezone instead of UTC
### Description

`AbstractJdbcRepository.getDate(Record, String groupByType)` reconstructs an `Instant` from SQL-extracted date parts (`year`/`month`/`day`/`hour`/`minute`) using the **JVM's default timezone**, even though those parts were extracted from a timestamp column that is stored/parsed as **UTC**. Whenever the JVM's default timezone isn't UTC, this introduces a shift equal to the UTC offset, which can move an aggregation bucket's data into the wrong bucket (or, for narrow bucket widths, drop it out of the requested window's bucket grid entirely).

This affects every caller of `aggregateByFlowId` / metric and dashboard aggregation code paths that use this date-bucketing helper.

### Root cause

- `jdbc/src/main/java/io/kestra/jdbc/AbstractJdbcRepository.java` (`getDate`, currently around lines 296-325) builds the returned `Instant` with `TimeZone.getDefault().toZoneId()`:

  ```java
  case "hour" -> {
      return ZonedDateTime.of(year, month, day, hour, 0, 0, 0, TimeZone.getDefault().toZoneId()).toInstant();
  }
  ```

  The `year`/`month`/`day`/`hour`/`minute` values are extracted from the DB via `DSL.year()`/`month()`/`day()`/`hour()`-style date-part functions applied to a timestamp column.

- That timestamp column, however, is a **UTC-only generated column**. For example, in H2 (`jdbc-h2/src/main/resources/migrations/baseline-h2.sql`):

  ```sql
  "timestamp" TIMESTAMP NOT NULL GENERATED ALWAYS AS (PARSEDATETIME(LEFT(JQ_STRING("value", '.timestamp'), 23) || '+00:00', 'yyyy-MM-dd''T''HH:mm:ss.SSSXXX')),
  ```

  The `+00:00` suffix always parses the stored value as UTC. So the raw `year/month/day/hour/minute` fields SQL hands back are UTC wall-clock components — but `getDate()` re-labels them as being in the JVM's local zone before converting to `Instant`, silently shifting the timestamp by the local UTC offset (e.g. +2h under CEST).

### Impact

- Aggregation buckets (`aggregateByFlowId` and similar dashboard/metric aggregation queries) can attribute a data point to the wrong bucket whenever the JVM's default timezone isn't UTC.
- With narrow bucket widths (hour/day), a data point can even fail to land in any bucket of a requested window, since the reconstructed `Instant` no longer round-trips consistently with the original data.

### Suggested fix

`getDate()` should reconstruct the `Instant` using `ZoneOffset.UTC` (matching how the source timestamp column is generated/parsed), not `TimeZone.getDefault()`.

bishalbera:
closes #17674 

---

### ✨ Description

What does this PR change?  

 - `getDate()` reassembles in `ZoneOffset.UTC` (5 sites).
  - New `groupByTimestampField` hook on `AbstractJdbcRepository`, overridden by
    `PostgresMetricRepository` to extract through `AT TIME ZONE 'UTC'` — so all three dialects now
    feed UTC digits in. H2/MySQL keep the existing cast.
  - Metric gap-filler moves to UTC alongside, or real data points stop matching their filled slots.

 Unit tests assert exact bucket instants under `Asia/Kolkata`, `Europe/Berlin` and                                                                                  
  `America/Los_Angeles`. `PostgresUtcTimestampFieldTest` runs the real expression against real
  Postgres under each zone — reverting the fix makes it fail with `expected: 14, but was: 20`.

  Full suites green: `jdbc` (59), `jdbc-h2` (969), `jdbc-mysql` (850), `jdbc-postgres` metric tests.

### 🤖 AI Authors

I've used claude to generate tests and java docs.

### 1. The tests fail without the fix (negative control)

  Reverting both halves of the fix on this branch — `getDate()` back to `TimeZone.getDefault()`,
  and the Postgres expression back to a plain cast — while keeping the new tests:

  AbstractJdbcRepositoryTest > shouldReassembleBucketsInUtcRegardlessOfDefaultZone  ✘
      expected: 2026-07-29T14:30:00Z
       but was: 2026-07-29T09:00:00Z          ← -5:30, reassembled as Asia/Kolkata

  AbstractJdbcRepositoryTest > shouldReassembleWeekBucketInUtcRegardlessOfDefaultZone  ✘
      expected: 2021-06-14T00:00:00Z
       but was: 2021-06-13T18:30:00Z          ← bucket lands on the previous day

  PostgresUtcTimestampFieldTest > shouldExtractUtcDatePartsRegardlessOfSessionTimeZone  ✘
      [hour, JVM zone Asia/Kolkata]
      expected: 14
       but was: 20                            ← real Postgres, extraction in session tz

  The Postgres failure is the one worth noting: it's against a live `postgres:14.13`, and it's the
  error the plain one-line fix would have shipped.

  ### 2. They pass with it

  AbstractJdbcRepositoryTest          5 passing
  PostgresUtcTimestampFieldTest       2 passing
  
  ### 3. Backend behaviour, measured

  Same source instant `2026-07-29T14:00:00Z`, each dialect's real column definition from the
  migrations, three JVM default zones (UTC / Asia/Kolkata / Europe/Berlin) — extracted `day, hour`:

  | Backend | Column | Extracted | Digits are |
  |---|---|---|---|
  | H2 | `TIMESTAMP` ← `PARSEDATETIME(… \|\| '+00:00')` | `29,14` / `29,14` / `29,14` | UTC, always |
  | MySQL | `DATETIME(6)` ← `STR_TO_DATE(…'%fZ')` | `29,14` / `29,14` / `29,14` | UTC, always |
  | Postgres | `TIMESTAMPTZ` | `29,14` / `29,19` / `29,16` | JVM-local |

  `SHOW TIME ZONE` on those three connections returned `UTC`, `Asia/Kolkata`, `Europe/Berlin` —
  pgjdbc propagates the JVM default into the session.

### DIFF
--- a/jdbc-postgres/src/main/java/io/kestra/repository/postgres/PostgresMetricRepository.java
+++ b/jdbc-postgres/src/main/java/io/kestra/repository/postgres/PostgresMetricRepository.java
@@ -1,5 +1,6 @@
 package io.kestra.repository.postgres;
 
+import java.sql.Timestamp;
 import java.util.Date;
 
 import org.jooq.Field;
@@ -20,6 +21,11 @@ public class PostgresMetricRepository extends AbstractJdbcMetricRepository {
     public PostgresMetricRepository(@Named("metrics") PostgresRepository<MetricEntry> repository,
         JdbcFilterService filterService) {
         super(repository, filterService);
+    }
+
+    @Override
+    protected Field<Timestamp> groupByTimestampField(String dateField) {
+        return PostgresRepositoryUtils.utcTimestampField(dateField);
     }
 
     @Override
--- a/jdbc-postgres/src/main/java/io/kestra/repository/postgres/PostgresRepositoryUtils.java
+++ b/jdbc-postgres/src/main/java/io/kestra/repository/postgres/PostgresRepositoryUtils.java
@@ -1,5 +1,6 @@
 package io.kestra.repository.postgres;
 
+import java.sql.Timestamp;
 import java.util.Date;
 
 import org.jooq.Field;
@@ -10,6 +11,18 @@ import io.kestra.core.utils.DateUtils;
 public final class PostgresRepositoryUtils {
     private PostgresRepositoryUtils() {
         // utility class pattern
+    }
+
+    /**
+     * Renders {@code dateField} as a UTC wall-clock timestamp, for SQL-side date-part extraction.
+     * <p>
+     * Postgres date columns are {@code TIMESTAMPTZ}, which stores an absolute instant. Casting one to
+     * {@code timestamp} resolves it in the session timezone, and pgjdbc sets the session timezone
+     * from the JVM default — so a plain cast yields local, not UTC, date parts. {@code AT TIME ZONE
+     * 'UTC'} pins the conversion instead of inheriting it from the connection.
+     */
+    public static Field<Timestamp> utcTimestampField(String dateField) {
+        return DSL.field("({0} AT TIME ZONE 'UTC')", Timestamp.class, DSL.field(DSL.quotedName(dateField)));
     }
 
     public static Field<Date> formatDateField(String dateField, DateUtils.GroupType groupType) {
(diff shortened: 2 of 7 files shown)


loicmathieu (approved):
LGTM, thanks a lot

loicmathieu:
Backported to 1.3

PR #17797, opened 2026-08-03
fix(ui): preserve repeated label filter values

### LINKED ISSUE #17725: [Filters] Labels "has any of" and "has none of" drop earlier values for the same key
### Describe the issue

On the Flows, Executions and Dashboards pages, the Labels filter with "has any of" or "has none of" only keeps one value per key. If you add environment:production and then environment:staging, the second pair replaces the first, so you cannot filter for a key that matches any of several values. Using two different keys works fine and keeps both pairs.

1. Deploy the three flows below.
2. Go to Flows. Click Add filters, choose Labels, and set the operator to "has any of".
3. Add environment:production, then add environment:staging.
4. Only environment:staging stays, and the list shows only nightly_etl. The two production flows are missing.

```yaml
id: daily_sales_report
namespace: company.reporting
labels:
  team: data
  environment: production
  tier: critical
tasks:
  - id: log
    message: daily sales report
```

```yaml
id: nightly_etl
namespace: company.reporting
labels:
  team: data
  environment: staging
tasks:
  - id: log
    message: nightly etl
```

```yaml
id: platform_healthcheck
namespace: company.platform
labels:
  team: platform
  environment: production
tasks:
  - id: log
    message: platform healthcheck
```

### Environment

Kestra Version: 2.0 develop
Kestra Edition: OSS

9904099:
### ✨ Description

Preserves multiple label values with the same key for the `has any of` (`IN`) and `has none of` (`NOT_IN`) comparators.

The change keeps those pairs through the complete frontend data path:
- the key/value editor retains distinct values instead of replacing the earlier pair;
- route encoding stores repeated label parameters as arrays, preserves colons and percent characters inside values, and route decoding restores each `key:value` pair exactly once;
- malformed or empty key/value pairs are ignored without discarding valid label values;
- reserved label keys such as `constructor`, `toString`, and `__proto__` are accumulated without inheriting object prototype properties;
- SDK query filters use an explicit OR node for `IN` and an explicit AND node for `NOT_IN` when a label key repeats;
- the shared query serializer preserves nested logical groups and comparator-only `IS_NULL` / `IS_NOT_NULL` leaves on the HTTP wire;
- malformed QueryFilter roots and arrays are rejected atomically before dispatch, while null/undefined filters and ordinary non-filter arrays keep their existing serialization.

Single-value comparators and label operations without repeated keys keep their existing behavior.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/17725

### 🎨 Frontend Checklist

- [ ] All existing E2E tests pass (`npm run test:e2e`) — not run locally because the integrated backend/browser environment is unavailable
- [ ] Screenshots or video recordings attached showing the `UI` changes — this is a filter-state/data-path correction with no visual styling change

### 📝 Additional Notes

Validated locally with:
- design-system unit suite: 740 tests passed;
- application unit suite: 1,183 tests passed;
- focused route/query/wire regressions: 72 tests passed;
- TypeScript checks: design system, app, and tests passed;
- changed-file lint: 0 warnings and 0 errors;
- production UI build: passed (8,324 modules transformed);
- backend `QueryFilterFormatBinderTest`: 15 tests passed; no backend code is changed by this PR.

AI author cat joke: The cat filtered for `environment:production` and `environment:staging`—then refused to drop either because both boxes were occupied. 🐱

### 🤖 AI Authors

This PR was prepared with AI assistance and independently reviewed before submission.


elevatebart (approved):
Nice improvement, we still need to figure out "has all of",

When we choose this operator, only one value can be selected. Not what we expect.

Interesting fact: this is technically translated to `EQUALS`. 
See https://github.com/9904099/kestra/blob/fix/17725-label-multi-values/ui/src/components/filter/configurations/labelComparatorLabels.ts#L8. 
I would think this is wrong.

PR #17799, opened 2026-08-03
Feat/jdbc query metrics

### LINKED ISSUE #17721: Unbounded metric cardinality: JDBC query duration tags the raw SQL text
The JDBC query-duration metric (`jdbc.query.duration`, registered in `JooqExecuteListenerFactory.java`) tags every recorded timer with the rendered SQL text of the statement:

```java
metricRegistry.timer(MetricRegistry.METRIC_JDBC_QUERY_DURATION, MetricRegistry.METRIC_JDBC_QUERY_DURATION_DESCRIPTION, tags)
```

```java
private String[] tags(ExecuteContext ctx) {
    var tags = new String[] { "batch", ctx.batchMode().name() };
    if (ctx.batchMode() != ExecuteContext.BatchMode.MULTIPLE) {
        return ArrayUtils.addAll(tags, "sql", ctx.sql());
    }
    return tags;
}
```

The comment on the batch-mode check ("in batch query, the query will be expanded without parameters, and will lead to overflow of metrics") shows the cardinality risk was already recognized for batch statements, but the same risk exists for regular (non-batch) statements and is not guarded there.

**Concretely, this is unbounded, not just large.** `In`/`NotIn` (`core/src/main/java/io/kestra/core/models/dashboards/filters/In.java`, `NotIn.java`) are public API models — `values` is a plain `List<Object>` with no size cap, deserialized directly from a REST filter payload on any list-filter endpoint (executions, logs, flows, etc.). Those values flow into `JdbcFilterService.java` (`.in(filter.getValues())` / `.notIn(...)`), and jOOQ renders one `?` placeholder per list element, so the *same logical query* produces a *different* `ctx.sql()` string depending on how many values the API caller passes.

Since `Micrometer`'s `MeterRegistry` deduplicates internally by `Meter.Id` (name + tags) and there is no `MeterFilter` configured anywhere in the codebase to cap cardinality (`grep -rn "MeterFilter"` across the repo returns nothing), every distinct rendered SQL string ever executed accumulates as a permanent, never-evicted `Timer` in the registry — driven by external API input, not bounded by the size of the codebase. This is the classic "cardinality explosion" failure mode for tag-based metrics and can also affect whatever scrapes/exports these metrics (e.g. a Prometheus backend).

## Decision:

1. Sanitize the queries so that IN and ORDER BY are deterministic (best effort).
2. Only record queries that last more than a configurable threshold, by default 100ms.

loicmathieu:
Closes https://github.com/kestra-io/kestra/issues/17721

1. Sanitized so IN-list, identifiers and sort columns are redacted, only emitted when the query is over 10ms, configurable.
2. Switch from executeStart/executeEnds to start/ends so we record also the fecth type, in conjonction with slow query monitoring on the database itself this might allow to discover connectivity issues or driver level issues
3. Stop formatting SQL queries as this is on itself a source of variance and is not necessary
4. Make the queue INSERT values deterministic


jdbc/src/main/java/io/kestra/jdbc/JooqExecuteListenerFactory.java
```diff
@@ -21,19 +21,28 @@
 @Slf4j
 @Factory
 public class JooqExecuteListenerFactory {
+    private static final Duration SLOW_QUERY_THRESHOLD = Duration.ofMillis(500);
```
Skraye: Would be great to make that configurable

jdbc/src/main/java/io/kestra/jdbc/JooqExecuteListenerFactory.java
```diff
@@ -48,30 +57,34 @@ public void executeStart(ExecuteContext ctx) {
                     }
 
                     @Override
-                    public void executeEnd(ExecuteContext ctx) {
-                        Duration duration = Duration.ofMillis(System.currentTimeMillis() - startTime);
+                    public void end(ExecuteContext ctx) {
```
Skraye: With method that stream, wont the duration be false ?

Skraye (changes requested):

jdbc/src/main/java/io/kestra/jdbc/JooqExecuteListenerFactory.java
loicmathieu: Done

jdbc/src/main/java/io/kestra/jdbc/JooqExecuteListenerFactory.java
loicmathieu: Maybe, but this was pre-existing.

loicmathieu: Or maybe not... I'll switch back to executeStart/executeEnd

nkwiatkowski (approved):

Skraye (approved):
LGTM, but feel like 10ms is a bit short, wont it always be triggered by log sql request by example ?

loicmathieu:
> LGTM, but feel like 10ms is a bit short, wont it always be triggered by log sql request by example

This is always a trade-off; we can increase it later if needed.

PR #17804, opened 2026-08-03
build(core): run e2e tests on pushes to develop

MilosPaunovic:
Closes https://github.com/kestra-io/kestra-ee/issues/9662.

## What

The `push` trigger on `e2e-scheduling.yml` targeted `main` — a branch that does not exist in this repository (the default branch is `develop`). The "and on merge" half of #14594 therefore never fired: every run of this workflow since it was added has been a `schedule` or a manual dispatch, never a push.

- **Fix the trigger** — `main` → `develop`. A merge now gets its own E2E signal with per-commit attribution instead of waiting for the next scheduled window.
- **Add a concurrency group** — now that the suite runs per merge this matters: the scheduled path passes no `exe-artifact`, so `kestra-oss-e2e-tests.yml` builds the whole product from source. Without a group, back-to-back merges each start a full build and stack up. Successive runs on the same ref now supersede each other, matching what `pull-request.yml` and `main-build.yml` already do.
- **Thin the schedule** — every two hours on weekdays (`0 9-21/2 * * 1-5`) down to one nightly run (`0 3 * * *`). That window was exactly working hours, i.e. exactly when merges land, so those runs mostly re-tested a commit the push trigger had already covered. A nightly run keeps what the schedule genuinely adds — flakiness on an unchanged commit, plus external drift such as base images or plugin downloads on days with no merges — and extends it to weekends, which the weekday-only cron never covered.

## Notes

- Net CI volume goes down, not up: ~35 scheduled runs/week become 7, and merges to `develop` are far fewer than that.
- The concurrency group is deliberately keyed on `github.ref` only, so a merge can cancel an in-flight scheduled run on the same ref. That is intended — the newest develop commit is the signal worth having. If we'd rather protect the nightly run, add `-${{ github.event_name }}` to the group key.


elevatebart (approved):
LGTM

PR #17816, opened 2026-08-03
feat: add deprecation support for Pebble functions

### LINKED ISSUE #15101: Provide a way too deprecate a Pebble functions
### Feature description

We sometime deprecate Pebble functions as they have replacements.
But we have no way to provide feedback to the user about the deprecation, so the function is still used.

We need a way, for deprecated function, to display a warning in flow validation and the editor

hemantmm:
closes: #15101

---

### ✨ Description

What does this PR change?  

Adds support for deprecated Pebble functions by exposing deprecation metadata and surfacing it in both flow validation and the editor.

## Changes

- Adds deprecation metadata (`deprecated` and `replacement`) to Pebble function definitions.
- Display validation warnings when deprecated Pebble functions are used in flows.
- Shows the recommended replacement function when available.
- Add/update tests covering deprecation metadata and validation warnings.


hemantmm:
- Drafts are intentionally allowed to be saved even when invalid.
- Before execution, the flow must be valid, so validation happens there instead.

core/src/main/java/io/kestra/core/runners/pebble/PebbleExpressionService.java
```diff
@@ -46,18 +46,16 @@ public PebbleExpressionService(List<Extension> extensions) {
...
-            {
+            .map(entry -> {
                 Function fn = entry.getValue();
                 List<String> argNames = fn.getArgumentNames();
-                if (argNames == null) {
-                    return new PebbleFunction(entry.getKey(), List.of());
-                }
                 Map<String, String> defaults = fn instanceof KestraFunction kf ? kf.getArgumentDefaults() : Map.of();
-                List<PebbleFunction.Argument> arguments = argNames.stream()
+                boolean deprecated = fn instanceof KestraFunction kf ? kf.isDeprecated() : fn.getClass().isAnnotationPresent(Deprecated.class);
```
loicmathieu:
```suggestion
                boolean deprecated = fn.getClass().isAnnotationPresent(Deprecated.class);
```

There is no point of branching and delegating to a function that do exactly that.

core/src/main/java/io/kestra/core/runners/pebble/functions/KestraFunction.java
```diff
@@ -15,4 +15,17 @@ public interface KestraFunction extends Function {
      * Ordering is not required — callers use {@link #getArgumentNames()} to determine order.
      */
     Map<String, String> getArgumentDefaults();
+
+      /**
+     * Returns {@code true} if this Pebble function is deprecated.
+     */
+    default boolean isDeprecated() {
+        return getClass().isAnnotationPresent(Deprecated.class);
+    }
```
loicmathieu: It didn't seem to be needed

core/src/main/java/io/kestra/core/services/FlowService.java
```diff
@@ -591,6 +595,32 @@ public List<String> warnings(Flow flow, String tenantId) {
                 .forEach(msg -> warnings.add("Trigger '" + trigger.getId() + "': " + msg))
         );
 
+        if (pebbleExpressionService != null && flow.getSource() != null) {
+            Map<String, PebbleFunction> deprecatedFunctions = pebbleExpressionService.functions().stream()
+                .filter(PebbleFunction::deprecated)
+                .collect(Collectors.toMap(PebbleFunction::name, f -> f));
+            if (!deprecatedFunctions.isEmpty()) {
+                Pattern functionPattern = Pattern.compile("\\b([a-zA-Z0-9_]+)\\s*\\(");
```
loicmathieu: move it to a constant

loicmathieu (changes requested):
Interesting approach!
Some minor feedback

hemantmm:
Thanks for the feedback @loicmathieu! I've addressed all the review comments:

- removed the unnecessary isDeprecated() method,
- used the annotation check directly,
- moved the Pebble function regex to a class-level constant.

loicmathieu (approved):
LGTM, thanks a lot

loicmathieu:
@hemantmm it didn't compile can you have a look please?

hemantmm:
@loicmathieu, fixed, thanks! The compile issue was in `FlowService`, where I was still using the deprecated helper method. I changed it to check @Deprecated directly as suggested and pushed the fix. Could you please take another look?

loicmathieu:
@hemantmm it still didn't compile in `FlowServiceTest`

hemantmm:
@loicmathieu, I reproduced it locally. `FlowServiceTest` itself compiles successfully (:core:compileTestJava passes), but the test fails at runtime because warnings() returns an empty list. So the remaining issue appears to be in the deprecated-function detection/pattern logic rather than Java compilation.

loicmathieu:
@hemantmm you can clearly see that CI is not working, if you're not willing to fix it i'll close this for now.

hemantmm:
Thanks for pointing that out. I will fix the compilation issue and push the changes. CI should pass with the updated test.

hemantmm:
@loicmathieu, can you review this PR now?

loicmathieu:
@hemantmm now there are conflicts, sorry...

hemantmm:
@loicmathieu resolved.

MilosPaunovic:
Can you give it another look now @loicmathieu?

loicmathieu:
@hemantmm can you rebase and fix the conflict, sorry about that, this time I'll merge it as soon as it's OK

PR #17817, opened 2026-08-03
feat(system): allow configuring the fail-fast mechanism

loicmathieu:
By default queues are fail-fast: if a subscriber fail to process the message, it is not acknowledge and the suscriber throws, usually closing Kestra.

With this change, this is now configurable and the user may choose to opt-out and ignore the message instead (which is usually not a good idea).

Closes https://github.com/kestra-io/kestra-ee/issues/9498

@aj-emerich it migt be documented in the queue configuration section in 2.0 but as it's highly dangerous I'm not sure it's a good idea. We might revisit this later.


nkwiatkowski (approved):

aj-emerich:
@loicmathieu what's your read on this? worth documenting to be completely user facing or only brought up in cases where deemed appropriate?

PR #17826, opened 2026-08-04
fix: cleaner should not be running on worker

Skraye:
Stops the expired KV entries purge from being scheduled on a worker. It needs database access, which a worker doesn't have: every hourly tick failed and logged an error with a stack trace, for work that was never the worker's to do.
 
 companion PR : https://github.com/kestra-io/kestra-ee/pull/9650


loicmathieu (approved):
LGTM

ammeek (approved):
LGTM

Issue #9008, opened 2025-05-27
Investigate improvement on flowable task processing


loicmathieu:
### Feature description

The main entry point of a flowable task for the Executor is the `resolveNexts()` method which is called to compute the list of subtask of a flowable task.

It appears that this method is always called 3 times for an `If`:
- The first time it returns the task to run
- The second time an empty list
- The third time an empty list!

The third call seems useless, we may investigate if we can improve the executor algorithm to avoid it.

PR #17832, opened 2026-08-04
refactor(executions): skip redundant flowable resolution when children cannot progress

loicmathieu:
## Summary

Fixes the redundant executor work reported in #9008: a running flowable's `resolveNexts()`/`resolveState()` were being re-evaluated on every executor cycle, even on cycles where the resolver could only return empty (e.g. an `If` task's `resolveNexts()` was called 4 times per execution, 2 of them wasted). Each call also rebuilds a `RunContext`, which runs an `outputRepository.findByExecution()` query — so every wasted call is a wasted DB query too.

- Adds `ExecutorService.canChildrenProgress(execution, parentTaskRun)`: a flowable can only produce a new child, or terminate, when it has no child task run yet or at least one child has terminated — every `FlowableUtils.resolve*Nexts`/`resolveState` implementation already returns empty/absent otherwise, so this guard never hides real work.
- Skips the whole RUNNING-flowable block in `handleFlowableTasks` (RunContext included) when the guard says no child can progress.

**Known limitation:** once any child has terminated the guard stops helping, so a `Sequential` of 5 tasks still re-resolves on every cycle from task 2 onward. This removes the redundancy at the head of each flowable, not all of it.

Verified with `LOGGER_LEVELS_IO_KESTRA_EXECUTOR=DEBUG` tracing on `IfTest.ifTruthy`: 4 `resolveNexts` calls → 2, executor cycle count unchanged (9).

Companion PR with a larger, riskier change (also collapsing executor cycles, not just resolveNexts calls) for comparison: see the linked PR for `perf/flowable-collapse-executor-cycle`.

## Test plan

- [x] New regression test `IfTest.shouldResolveFlowableOnlyWhenItCanProgress` pins the flowable-resolution count via `METRIC_EXECUTOR_FLOWABLE_EXECUTION_COUNT` (asserts a delta of exactly 2 around one execution)
- [x] `io.kestra.plugin.core.flow.IfTest` — 8/8 passing
- [x] Full `io.kestra.plugin.core.flow.*` suite (Sequential, Parallel, Dag, Switch, Pause, Loop, WorkingDirectory, AllowFailure, Finally, ...) — 123/123 passing
- [x] `H2RunnerTest` (includes `multipleIf`, 3 `If`s inside a concurrent `Parallel`) — 99/99 passing
- [x] `H2RunnerRetryTest` (verifies the retry/errors-branch race fixed in #17538 is unaffected) — 21/21 passing

Refs #9008

### DIFF
--- a/core/src/test/java/io/kestra/plugin/core/flow/IfTest.java
+++ b/core/src/test/java/io/kestra/plugin/core/flow/IfTest.java
@@ -10,6 +10,7 @@ import org.junit.jupiter.api.Test;
 import io.kestra.core.junit.annotations.ExecuteFlow;
 import io.kestra.core.junit.annotations.KestraTest;
 import io.kestra.core.junit.annotations.LoadFlows;
+import io.kestra.core.metrics.MetricRegistry;
 import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.executions.TaskRunAttempt;
 import io.kestra.core.models.flows.State;
@@ -29,6 +30,9 @@ class IfTest {
 
     @Inject
     private ExecutionRepositoryInterface executionRepository;
+
+    @Inject
+    private MetricRegistry metricRegistry;
 
     @Test
     @LoadFlows(value = { "flows/valids/if-condition.yaml" }, tenantId = "iftruthy")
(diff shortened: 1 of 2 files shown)


loicmathieu:
## Benchmark result (scenario 2)

### Baseline

| step | rate | avg ms | p99 ms | failed | kestra cpu | kestra mem | infra cpu | infra mem | heap |
|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| 1 | 100 | 592 | 652 | 0 | 12.2% | 2.3 GiB | 19.2% | 0.9 GiB | 571.8 MiB |
| 2 | 200 | 592 | 733 | 0 | 16.6% | 2.3 GiB | 27.1% | 0.9 GiB | 504.2 MiB |
| 3 | 300 | 633 | 791 | 0 | 23.6% | 2.4 GiB | 37.8% | 0.9 GiB | 388.3 MiB |
| 4 | 400 | 721 | 986 | 0 | 27% | 2.4 GiB | 41.2% | 1 GiB | 476.8 MiB |
| 5 | 500 | 932 | 1273 | 0 | 31% | 2.4 GiB | 45.7% | 1 GiB | 457.2 MiB |
| 6 | 600 | 2002 | 2693 | 0 | 37.6% | 2.5 GiB | 55.7% | 1 GiB | 457.9 MiB |
| 7 | 700 | 12848 | 25032 | 0 | 35.9% | 2.5 GiB | 54.7% | 1 GiB | 522.3 MiB |

### This PR

| step | rate | avg ms | p99 ms | failed | kestra cpu | kestra mem | infra cpu | infra mem | heap |
|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| 1 | 100 | 485 | 554 | 0 | 11.8% | 2.3 GiB | 16.2% | 0.9 GiB | 565.4 MiB |
| 2 | 200 | 470 | 540 | 0 | 15.3% | 2.3 GiB | 22.4% | 0.9 GiB | 451.7 MiB |
| 3 | 300 | 497 | 606 | 0 | 21.9% | 2.3 GiB | 32.5% | 0.9 GiB | 720.4 MiB |
| 4 | 400 | 503 | 600 | 0 | 24.6% | 2.4 GiB | 40.3% | 0.9 GiB | 808.7 MiB |
| 5 | 500 | 553 | 683 | 0 | 30.1% | 2.4 GiB | 43.8% | 1 GiB | 736.9 MiB |
| 6 | 600 | 643 | 848 | 0 | 31.7% | 2.4 GiB | 50.9% | 1 GiB | 580.5 MiB |
| 7 | 700 | 928 | 1308 | 0 | 39.6% | 2.4 GiB | 54.4% | 1 GiB | 400.7 MiB |

### Summary (IA)

Clear improvement, not noise: PR 17832 beats v2.0.0-rc6 at every rate, and the gap widens with load — at 700 req/s the baseline was effectively falling over (avg 12.8s, p99 25s) while the candidate holds steady (avg 928ms, p99 1.3s), a ~93% latency cut. CPU is roughly flat to slightly better on both VMs across the board (kestra cpu +10% only at the 700 req/s tail, likely just because it's finishing far more work per step).

One thing to flag: heap on the candidate runs 27–86% *higher* than baseline at 300–600 req/s, before dropping below baseline at 700 — worth checking there isn't a GC/retention regression in 17832 (a JVM change that trades earlier collection for higher steady-state heap could still be fine, but it's the one metric that moved the wrong way).

Both runs are on the same stack/backend/scenario (`complex`, OSS Postgres, same machine types), so the comparison is solid — the baseline's collapse at 700 req/s looks like a real ceiling in rc6 that 17832 fixes, not a fluke.

loicmathieu:
Closing in favor of https://github.com/kestra-io/kestra/pull/17833

PR #17833, opened 2026-08-04
refactor(executions): dispatch a newly-created root task run in the same executor cycle

### LINKED ISSUE #9008: Investigate improvement on flowable task processing

loicmathieu:
### 1. Skip redundant flowable resolution

Fixes the redundant executor work reported in #9008: a running flowable's resolveNexts()/resolveState() was being re-evaluated on every executor cycle, even on cycles where the resolver could only return empty (e.g. an If task's resolveNexts() was called 4 times per execution, 2 of them wasted). Each call also rebuilds a `RunContext`, which runs an outputRepository.findByExecution() query — so every wasted call is a wasted DB query too.

Adds `ExecutorService.canChildrenProgress(execution, parentTaskRun)`: a flowable can only produce a new child, or terminate, when it has no child task run yet or at least one child has terminated — every FlowableUtils.resolve*Nexts/resolveState implementation already returns empty/absent otherwise, so this guard never hides real work.
Skips the whole RUNNING-flowable block in handleFlowableTasks (RunContext included) when the guard says no child can progress.
Known limitation: once any child has terminated the guard stops helping, so a Sequential of 5 tasks still re-resolves on every cycle from task 2 onward. This removes the redundancy at the head of each flowable, not all of it.

Verified with LOGGER_LEVELS_IO_KESTRA_EXECUTOR=DEBUG tracing on IfTest.ifTruthy: 4 resolveNexts calls → 2,

### 2. Dispatch a newly-created root task run in the same executor cycle

`ExecutorContext.withTaskRun` used to only append new task runs to a separate `nexts` list, merged into the execution by `onNexts()` only after `process()` returned. That meant `handleWorkerTasks` could never see a task run created earlier in the same pass, costing a full executor cycle (execution-store lock + queue round-trip) between e.g. `handleNext` creating the flow's first task and `handleWorkerTasks` dispatching it to RUNNING.

- `withTaskRun` now merges new task runs into the execution as soon as they're produced, and sets `executionUpdated` directly on that merge.
- `onNexts()`'s own merge logic is now dead code (its one caller always passes an execution already merged into) and has been stripped back down to its one remaining job: the CREATED → RUNNING transition.

**Why this PR doesn't also reorder `handleFlowableTasks` before `handleWorkerTasks`** (which would collapse the *second* redundant cycle too, per #9008): It breaks every `Loop`-based test. The `Loop` case inside `handleFlowableTasks` transitions its own CREATED task run straight to RUNNING as it spawns sub-executions. Running `handleWorkerTasks` after that would leave the Loop task run RUNNING with no attempts (`handleWorkerTasks` only attaches the first RUNNING attempt to task runs it finds still CREATED), which crashes `LoopExecutionEventMessageHandler.terminateLoop()` (`attempts.getLast()` on an empty list) the moment the first sub-execution finishes. It would need either splitting `handleFlowableTasks` into a reorderable "compute nexts" half and a position-sensitive "custom executor logic" half, or making `Loop`'s spawn logic not bypass `handleWorkerTasks`'s attempt-injection.

Verified with `LOGGER_LEVELS_IO_KESTRA_EXECUTOR=DEBUG` tracing on `IfTest.ifTruthy`: 9 executor cycles → 8 (the `handleNext`-creates / `handleWorkerTasks`-dispatches pair now collapses into one; the `handleFlowableTasks`-creates / `handleWorkerTasks`-dispatches pair does not, for the reason above).

## Test plan

- [x] Same regression test as #17832 (`IfTest.shouldResolveFlowableOnlyWhenItCanProgress`) — flowable-resolution count unchanged at 2
- [x] `io.kestra.plugin.core.flow.IfTest` — 8/8 passing
- [x] Full `io.kestra.plugin.core.flow.*` suite — 123/123 passing (including the 5 Loop-based classes that failed with the full reorder before it was reverted)
- [x] `H2RunnerTest` — 99/99 passing
- [x] `H2RunnerRetryTest` — 21/21 passing

Closes #9008

### DIFF
--- a/core/src/test/java/io/kestra/plugin/core/flow/IfTest.java
+++ b/core/src/test/java/io/kestra/plugin/core/flow/IfTest.java
@@ -10,6 +10,7 @@ import org.junit.jupiter.api.Test;
 import io.kestra.core.junit.annotations.ExecuteFlow;
 import io.kestra.core.junit.annotations.KestraTest;
 import io.kestra.core.junit.annotations.LoadFlows;
+import io.kestra.core.metrics.MetricRegistry;
 import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.executions.TaskRunAttempt;
 import io.kestra.core.models.flows.State;
@@ -29,6 +30,9 @@ class IfTest {
 
     @Inject
     private ExecutionRepositoryInterface executionRepository;
+
+    @Inject
+    private MetricRegistry metricRegistry;
 
     @Test
     @LoadFlows(value = { "flows/valids/if-condition.yaml" }, tenantId = "iftruthy")
@@ -169,5 +173,33 @@ class IfTest {
         assertThat(execution.getState().getCurrent()).isEqualTo(State.Type.FAILED);
         assertThat(execution.getTaskRunList().getFirst().getState().getCurrent()).isEqualTo(State.Type.FAILED);
         assertThat(execution.getTaskRunList().getFirst().getAttempts().getFirst().getState().getCurrent()).isEqualTo(State.Type.FAILED);
+    }
+
+    // https://github.com/kestra-io/kestra/issues/9008: the Executor used to call resolveNexts() on a
+    // running If for every executor cycle, most of them resolving to nothing. Pin the number of
+    // resolutions so a regression shows up here rather than only in a profiler.
+    @Test
+    @LoadFlows(value = { "flows/valids/if-condition.yaml" }, tenantId = "ifflowablecount")
+    void shouldResolveFlowableOnlyWhenItCanProgress() throws TimeoutException, QueueException {
+        // Given
+        double before = flowableExecutionCount();
+
+        // When
+        Execution execution = runnerUtils.runOne(
+            "ifflowablecount", "io.kestra.tests", "if-condition", null,
+            (f, e) -> Map.of("param", true), Duration.ofSeconds(120)
+        );
+
+        // Then
+        assertThat(execution.getState().getCurrent()).isEqualTo(State.Type.SUCCESS);
+        assertThat(flowableExecutionCount() - before).isEqualTo(2d);
+    }
+
+    private double flowableExecutionCount() {
+        return metricRegistry.counter(
+            MetricRegistry.METRIC_EXECUTOR_FLOWABLE_EXECUTION_COUNT,
+            MetricRegistry.METRIC_EXECUTOR_FLOWABLE_EXECUTION_COUNT_DESCRIPTION,
+            MetricRegistry.TAG_TASK_TYPE, If.class.getName()
+        ).count();
     }
 }
--- a/executor/src/main/java/io/kestra/executor/ExecutorContext.java
+++ b/executor/src/main/java/io/kestra/executor/ExecutorContext.java
@@ -28,13 +28,15 @@ public class ExecutorContext {
 
     private Execution execution;
     private Exception exception;
-    private final List<String> from = new ArrayList<>();
-    private boolean executionUpdated = false;
     private FlowWithSource flow;
+
+    private final List<String> from = new ArrayList<>(8);
+    private boolean executionUpdated = false;
+    private int nextCount = 0;
+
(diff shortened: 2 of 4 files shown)


loicmathieu:
## Benchmark result (scenario 2)

### Baseline

| step | rate | avg ms | p99 ms | failed | kestra cpu | kestra mem | infra cpu | infra mem | heap |
|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| 1 | 100 | 592 | 652 | 0 | 12.2% | 2.3 GiB | 19.2% | 0.9 GiB | 571.8 MiB |
| 2 | 200 | 592 | 733 | 0 | 16.6% | 2.3 GiB | 27.1% | 0.9 GiB | 504.2 MiB |
| 3 | 300 | 633 | 791 | 0 | 23.6% | 2.4 GiB | 37.8% | 0.9 GiB | 388.3 MiB |
| 4 | 400 | 721 | 986 | 0 | 27% | 2.4 GiB | 41.2% | 1 GiB | 476.8 MiB |
| 5 | 500 | 932 | 1273 | 0 | 31% | 2.4 GiB | 45.7% | 1 GiB | 457.2 MiB |
| 6 | 600 | 2002 | 2693 | 0 | 37.6% | 2.5 GiB | 55.7% | 1 GiB | 457.9 MiB |
| 7 | 700 | 12848 | 25032 | 0 | 35.9% | 2.5 GiB | 54.7% | 1 GiB | 522.3 MiB |

### This PR

| step | rate | avg ms | p99 ms | failed | kestra cpu | kestra mem | infra cpu | infra mem | heap |
|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| 1 | 100 | 494 | 570 | 0 | 11.9% | 2.1 GiB | 14.5% | 0.9 GiB | 605.2 MiB |
| 2 | 200 | 492 | 617 | 0 | 16.6% | 2.2 GiB | 20.5% | 0.9 GiB | 636.4 MiB |
| 3 | 300 | 501 | 608 | 0 | 22.1% | 2.2 GiB | 29% | 0.9 GiB | 540.4 MiB |
| 4 | 400 | 511 | 645 | 0 | 24.6% | 2.2 GiB | 33.8% | 0.9 GiB | 512.7 MiB |
| 5 | 500 | 527 | 662 | 0 | 30.6% | 2.2 GiB | 40.6% | 1 GiB | 413 MiB |
| 6 | 600 | 629 | 831 | 0 | 32.8% | 2.2 GiB | 44% | 1 GiB | 378.1 MiB |
| 7 | 700 | 803 | 1136 | 0 | 40.3% | 2.3 GiB | 51.6% | 1 GiB | 390.7 MiB |

### Summary (IA)

**Clear improvement, not noise**: PR #17833 beats v2.0.0-rc6 at every rate on latency, and the gap widens sharply under load. At 100–400 req/s it's a steady 15–35% cut in avg/p99. At 600–700 req/s the baseline falls off a cliff — avg 2002ms→12848ms and p99 2693ms→25032ms from step 6 to 7 — while the candidate stays flat (629→803ms avg, 831→1136ms p99), so the candidate lifts the effective ceiling of this stack well past 700 req/s. Query volume is proportionally lower on the candidate at the same rate (e.g. `update executions` calls down ~11% across steps) with similar or lower per-query mean latency, consistent with a genuine efficiency fix rather than a workload difference. CPU/mem track the same trend (infra CPU down 5–25%, kestra CPU roughly flat), and no failed/killed executions on either run — heap deltas are noisy in both directions and not concerning given zero GC/connection pressure in either report. Caveat: comparing an RC baseline to a PR build against `develop`'s trajectory, not a released version — treat this as "PR fixes a real regression/bottleneck" rather than a version-over-version release comparison.

AcevedoR (approved):

PR #17840, opened 2026-08-04
fix(flows): guard getNamespace against a missing router context

RemiBarthe:
This pull request makes a minor update to the `useFlowStore` store. The change improves the robustness of the `getNamespace` function by adding optional chaining to safely access the `namespace` query parameter from the route object.

* Improved safety when accessing `namespace` from `route.query` in `getNamespace` by using optional chaining.


MilosPaunovic (approved):

loicmathieu (approved):
LGTM

PR #17868, opened 2026-08-05
fix(worker): register every controller gRPC service when the controller is embedded

fhussonnois:
A webserver starts an embedded controller unless `--no-controller` is passed, but every metadata gRPC service was gated on `kestra.server-type` being CONTROLLER or STANDALONE. On a webserver those beans were absent from the context, so the embedded gRPC server silently came up without them: worker registration and job dispatch worked, while KV, namespace files, flow metastore and execution log lookups failed at task runtime with `UNIMPLEMENTED: Method not found`.

Introduce `@RequiresControllerServer`, which also matches WEBSERVER, and apply it to every controller-hosted bean. Instantiation stays lazy, so a `--no-controller` webserver still builds none of them, and `ControllerStorageRegistrar` keeps advertising nothing since no CONTROLLER service instance ever reaches RUNNING.

The controller now logs the gRPC services it registered, as a missing one is otherwise only visible to workers as an UNIMPLEMENTED error at task runtime.

### DIFF
--- a/worker-controller/src/main/java/io/kestra/controller/DefaultController.java
+++ b/worker-controller/src/main/java/io/kestra/controller/DefaultController.java
@@ -2,6 +2,7 @@ package io.kestra.controller;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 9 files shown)


ammeek (approved):
LGTM

PR #17870, opened 2026-08-05
fix(ui): clarify the "can not delete" error on a flow's last task

### LINKED ISSUE #17323: [No-code editor] Deleting an unsaved task from topology shows an unclear error
### Describe the issue

Deleting a task from the topology view is not working as expected when the task has not been saved yet. The UI shows a generic "Can not delete" message, which does not explain what the user needs to do next.

1. Create or edit a flow in the no-code editor.
2. Add a task in the topology view but do not save the flow.
3. Try to delete that unsaved task.
4. The UI shows a generic "Can not delete" notification.

Additional video: 

https://github.com/user-attachments/assets/132aa36b-95f3-4a7f-a803-ad67ff27b90c

### Expected Behaviour
If unsaved tasks cannot be deleted, the notification should explain that clearly, for example "Please save the flow before deleting tasks." If deleting unsaved tasks is intended to be supported, the task should be removable without requiring the flow to be saved first.

### Environment

- Kestra Version: `1.3.27`
- Observed: `Wed, Jul 8, 2026 7:27 PM IST`
- Edition: Open Source

harshcodes17:
Deleting a flow's only task in the topology view showed a vague "Can not delete" toast. The title now states the actual rule (a flow must have at least one task) and the message tells the user what to do about it, instead of leaving them to guess.

Closes #17323

---

### Description

Deleting a flow's only/last task from the Topology view triggers a client-side guard (a flow must have at least one task) that shows an error toast. The toast worked, but its copy didn't explain *why* the delete was blocked or *what to do* about it — just "Can not delete."

This PR only changes the toast copy in `ui/src/translations/en.json`, no behavior change:

| | Before | After |
|---|---|---|
| Title | `Can not delete` | `A flow must have at least one task` |
| Message | `Each flow must have at least one task.` | `Add another task before deleting this one.` |

Screenshot below confirms the new title renders correctly when attempting to delete a flow's only task.

Screenshot 2026-08-05 at 10 29 37 PM

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/17323

### Frontend Checklist

- [ ] All existing E2E tests pass (`npm run test:e2e`) — no existing e2e test covers this delete-task-in-topology flow, so this wasn't run against it specifically; verified manually instead by reproducing the bug and confirming the new copy in the browser.

### Additional Notes

While reproducing this, I found one separate, unrelated bugs I did **not** fix here to keep this PR focused on the wording:
 **The toast never re-appears after the first one.** `ErrorToast.vue` only shows its notification from an `onMounted` hook, which fires once when `coreStore.message` first transitions from empty to set. Nothing resets `coreStore.message` afterward and there's no `watch()` on the message prop, so the first toast never auto-dismisses (`duration: 0`), and any *second* error anywhere in the app while the first is still showing silently fails to display.


RemiBarthe (changes requested):
Thanks for the PR!

1. **The real bug is upstream: the second string was never rendered.** `ErrorToast.vue` rebuilds the object it passes to `ErrorToastContainer` and only copies `content.message`, while `coreStore.message` is populated with `.message`. The toast body is therefore always empty. The reporter was literally only ever seeing "Can not delete". Changing the strings has no visible effect without that fix:

```js
message: h(ErrorToastContainer, {
    message: {
        message: props.message?.message,
        content: {message: props.message?.content?.message ?? ""},
    },
```

2. **On the strings:** the title states the failure, the body carries both the constraint and the next step the issue asks for:

```json
"can not delete": "Cannot delete this task",
"can not have less than 1 task": "A flow must have at least one task. Add another task before deleting this one."
```

harshcodes17:
> Thanks for the PR!
> 
> 1. **The real bug is upstream: the second string was never rendered.** `ErrorToast.vue` rebuilds the object it passes to `ErrorToastContainer` and only copies `content.message`, while `coreStore.message` is populated with `.message`. The toast body is therefore always empty. The reporter was literally only ever seeing "Can not delete". Changing the strings has no visible effect without that fix:
> 
> ```js
> message: h(ErrorToastContainer, {
>     message: {
>         message: props.message?.message,
>         content: {message: props.message?.content?.message ?? ""},
>     },
> ```
> 
> 2. **On the strings:** the title states the failure, the body carries both the constraint and the next step the issue asks for:
> 
> ```json
> "can not delete": "Cannot delete this task",
> "can not have less than 1 task": "A flow must have at least one task. Add another task before deleting this one."
> ```

Thanks for catching that — you're right, the wording change alone was a no-op without the plumbing fix.

Fixed both in 214d76df3:
- `ErrorToast.vue` now forwards `props.message?.message` alongside `content.message`, so the body actually renders.
- Updated the copy to your suggested wording — title states the failure, body carries the rule + next step.

Verified `npm run check:types` and `npm run translations:check` both pass clean with the change.

RemiBarthe:
Thanks for the quick turnaround!

One last thing caused by the `ErrorToast` fix: [`Dashboard.vue:169`](https://github.com/kestra-io/kestra/blob/develop/ui/src/components/dashboard/Dashboard.vue#L169) sets `title` and `message` to the same string. The body used to be dropped, so it was invisible. Now the sentence shows up twice in the toast. It's the only one of the 10 call sites in that situation, the others all pass a distinct body.

```diff
 coreStore.message = {
     variant: "error",
     title: err,
-    message: err,
 }
```

Good to merge after that :rocket:

harshcodes17:
> Thanks for the quick turnaround!
> 
> One last thing caused by the `ErrorToast` fix: [`Dashboard.vue:169`](https://github.com/kestra-io/kestra/blob/develop/ui/src/components/dashboard/Dashboard.vue#L169) sets `title` and `message` to the same string. The body used to be dropped, so it was invisible. Now the sentence shows up twice in the toast. It's the only one of the 10 call sites in that situation, the others all pass a distinct body.
> 
> ```diff
>  coreStore.message = {
>      variant: "error",
>      title: err,
> -    message: err,
>  }
> ```
> 
> Good to merge after that 🚀

Good catch — removed the redundant `message: err` in Dashboard.vue (c89656ed6). Confirmed it was the only call site with that pattern; checked all other `coreStore.message` usages and they each pass a distinct title/message.

RemiBarthe (approved):

PR #17873, opened 2026-08-05
feat(core): parse kotlp OTLP telemetry in TaskLogLineMatcher

tchiotludo:
## What this PR does

Kestra is introducing [kotlp](https://github.com/kestra-io/kotlp) as a process wrapper for task-runner commands. kotlp emits OpenTelemetry telemetry as OTLP/JSON NDJSON records in two shapes:

- **Console (default `kjson` framing):** one record per line as `::{"otlp":<OTLP/JSON record>}::`, flowing through the existing `TaskLogLineMatcher` path.
- **File (`--log-dir`):** bare OTLP NDJSON, one record per line, no framing. In that mode the file is the *only* full copy — kotlp drops the console framing and prints the child's output raw instead.

This PR empowers `TaskLogLineMatcher` to parse both:

- **New `OtlpRecord` POJO model** (`io.kestra.core.models.tasks.runners.otlp`) mirroring the OTLP/JSON wire format (logs, metrics with gauge/sum data points, spans). Plain Jackson records — no new dependency; unknown sections (histograms, …) are ignored by the lenient mapper, and OTLP int64-as-string fields coerce to `Long` automatically.
- **`TaskLogMatch` gains an `otlp` component** so the existing single Jackson parse detects the framed form. A new protected `handleOtlp(...)` step:
  - forwards OTLP **logs** to `runContext.logger()` with the record's `timeUnixNano` as the original timestamp (severity mapped per the OTel log data model, `severityText` fallback, default INFO);
  - forwards OTLP **metrics** to `runContext.metric(...)`: gauge and *cumulative* sum → Kestra `Gauge` (last periodic sample wins, so cumulative totals like `process.cpu.time` are not over-counted), *delta* sum → Kestra `Counter`; data-point attributes become tags;
  - leaves OTLP **traces** untouched — they are only exposed as POJOs via `match.otlp().resourceSpans()` for task-runner implementations, until native trace support lands.
- **New public `parseOtlp(InputStream, Logger, RunContext, Instant)`** ingesting bare OTLP NDJSON files (kotlp `--log-dir` output): forwards logs/metrics identically and returns all parsed records; blank, truncated (file rotation) and non-OTLP lines are skipped with a warning.

All new processing steps are `protected`, preserving the EE subclass/`@Replaces` extension pattern.

## Naming

Both `koltp` and `oltp` were letter transpositions of *k + OTLP*. Upstream renamed the repo and binary to `kotlp` and, in the same commit, started emitting the envelope key as `otlp`. That second half is API: a consumer matching on `::{"oltp":` silently stops matching, with nothing logged. Every mention on this side now follows.

`kotlpVersion` is therefore pinned to **0.1.0**: `v0.0.1` predates the rename, so its release assets are still named `koltp` and its binary still emits the old key. `downloadKotlp` fetches the binary by asset name, so anything older than v0.1.0 fails the build outright rather than embedding a binary Kestra cannot read.

## Tests

`TaskLogLineMatcherTest` (16 cases): severity-number and severity-text mapping, timestamp fallback, non-string bodies, gauge/cumulative/delta metric semantics, attribute-to-tag flattening, span exposure without forwarding, NDJSON stream parsing, unknown-field tolerance, and regression of the classic `::{"outputs":…}::` syntax alone and combined with `otlp` on the same line.

The NDJSON fixture is a **verbatim `--log-dir` recording** of the pinned kotlp v0.1.0 release (six records over three metric batches, only `host.name` replaced), not a hand-written sample — so the parser is checked against the binary we ship rather than against a reading of the spec. It covers metrics the parser had never been fed (`process.cpu.utilization`, `process.memory.virtual`, `process.thread.count`, `process.open_file_descriptor.count`), the same series sampled repeatedly, and the closing rusage-derived batch. The three lines that must be skipped (blank, truncated mid-object, non-OTLP object) are appended after the recording, since kotlp does not produce them on purpose.

`CommandsWrapperKotlpTest` (7 cases) and `KotlpUtilsTest` (3 cases) cover the wrapper end to end. Two assertions exist specifically for the failure mode of the rename — an embedded binary spelling the key the old way produces perfectly valid output that Kestra ingests *none* of, with no error anywhere:

- with kotlp enabled, `process.cpu.time` must reach the run context, which only happens if the binary and the matcher spell the envelope key the same way;
- with `--log-dir` set, the run context must have *no* metrics (the console goes raw) until the file is fed back through `parseOtlp`, which is what a caller setting the option has to do.

## ⚠️ Merge order

kestra-io/kestra-ee#9729 has to go in together with this: the EE re-parse path invokes `TaskLogMatch`'s canonical constructor directly, so either side alone leaves EE uncompilable.

## ⚠️ EE note

`TaskLogMatch` gained a fifth component, named `otlp`; any EE code invoking its canonical constructor directly needs a one-line update. In OSS it is only constructed by Jackson. The companion PR kestra-io/kestra-ee#9729 carries the rename.


tchiotludo:
EE companion PR (same branch name): kestra-io/kestra-ee#9729 — carries the new `oltp` component through the EE `TaskLogMatchWithEncrypted` re-parse path. Merge this OSS PR first.

core/src/main/java/io/kestra/core/models/tasks/runners/otlp/OtlpRecord.java
loicmathieu, line 1: We have the OTLP proto inside the classpath in the package `io.opentelemetry.proto`, I know that proto may not be very easy to use, but did you consider using them instead of maintaining our own record here?

core/src/main/java/io/kestra/core/models/tasks/runners/TaskLogLineMatcher.java
loicmathieu, line 1: Replace everywhere oltp -> otlp

core/src/main/java/io/kestra/core/models/tasks/runners/TaskLogLineMatcher.java
```diff
@@ -98,9 +113,214 @@ protected TaskLogMatch handle(Logger logger, RunContext runContext, Instant inst
...
+     * untouched as they are only exposed to the caller for now.
+     */
+    protected void handleOtlp(Logger logger, RunContext runContext, Instant instant, OtlpRecord record, String data) {
+        ListUtils.emptyOnNull(record.resourceLogs()).stream()
+            .flatMap(resourceLogs -> ListUtils.emptyOnNull(resourceLogs.scopeLogs()).stream())
+            .flatMap(scopeLogs -> ListUtils.emptyOnNull(scopeLogs.logRecords()).stream())
+            .forEach(logRecord ->
+            {
+                try {
+                    runContext
```
loicmathieu, line 196: I wonder if we should not also add the traceId and spanId as log record attributes.
It needs to be checked as you have them inside the OTLP logs so you should be able to add them somewher, maybe inside the MDC.

core/src/main/java/io/kestra/core/models/tasks/runners/TaskLogLineMatcher.java
```diff
@@ -98,9 +113,214 @@ protected TaskLogMatch handle(Logger logger, RunContext runContext, Instant inst
...
+     * @param line one line of bare OTLP/JSON
+     * @return the parsed record
+     */
+    public OtlpRecord parseOtlpLine(String line) throws JsonProcessingException {
+        return MAPPER.readValue(line, OtlpRecord.class);
+    }
+
+    /**
+     * Forwards the logs and metrics of an OTLP record to the {@link RunContext}; traces are left
+     * untouched as they are only exposed to the caller for now.
```
loicmathieu, line 187: Can you open a followup for sending traces?
I'm not sure we can send them as is but we can at least create a single span with the master span you created inside kotlp.

loicmathieu (changes requested):
Please also change everywhere `oltp` -> `otlp`.

core/src/main/java/io/kestra/core/models/tasks/runners/TaskLogLineMatcher.java
loicmathieu: Done, in `0e538b64da` — and the same transposition was in the binary name, so `koltp` → `kotlp` went with it: the class names (`KotlpOptions`, `KotlpUtils`), the `kotlp` Gradle property and download task, the embedded resource path, the `.gitignore` entry and every javadoc mention.

Worth calling out why the `oltp` → `otlp` half is more than cosmetic: upstream changed the emitted envelope key in the same commit as the repo rename, so a consumer still matching `::{"oltp":` gets *valid* records it silently ignores — no exception, no warning, just no telemetry. `262b13bb36` adds an assertion for exactly that: with kotlp enabled, `process.cpu.time` has to reach the run context, which only holds if the embedded binary and the matcher spell the key the same way.

That also forces `kotlpVersion` to `0.1.0` (was `0.0.1`): the v0.0.1 release predates the rename, so its assets are still named `koltp` and its binary still emits the old key. **This PR needs kotlp v0.1.0 released before CI can be green.**

core/src/main/java/io/kestra/core/models/tasks/runners/otlp/OtlpRecord.java
loicmathieu: I looked at it, and I'd rather keep the records — but it is a close enough call that I'll switch if you prefer.

The proto classes are not actually on `core`'s classpath. `io.opentelemetry.proto:opentelemetry-proto` appears once in the build, as a *constraint* in `platform/build.gradle:44` (version alignment for whichever module drags it in transitively); `./gradlew :core:dependencies --configuration runtimeClasspath` lists no `io.opentelemetry.proto` entry. So using them here means adding an `-alpha` artifact as a compile dependency of `core`.

The second problem is the parse itself. Those are protobuf messages: OTLP/JSON does not go into them through Jackson, it needs `protobuf-java-util`'s `JsonFormat.parser()` (plus `ignoringUnknownFields()`, which is what keeps histograms and future sections from throwing). Today the framed line is one `MAPPER.readValue(..., TaskLogMatch.class)` call and the OTLP record is just a component of that record. With proto types, `TaskLogMatch` can no longer hold the parsed value directly — we would have to keep the subtree as a `JsonNode`, re-serialize it, and hand it to `JsonFormat`, i.e. parse the payload twice and take a second dependency to do it.

What we would gain is real: the field set stops being ours to maintain, and it would already cover histograms/exponential histograms/profiles. If we start *sending* traces onward (your other comment) that argument gets stronger, since we would then want the full span model rather than the subset. Happy to do it as a follow-up on that trigger.

core/src/main/java/io/kestra/core/models/tasks/runners/TaskLogLineMatcher.java
loicmathieu: Checked, and there are two blockers today — both fixable, neither in this PR:

1. **kotlp doesn't put them on log records.** I ran the merged binary (`v0.0.1-10-g8072519`) and dumped the record keys: log records carry `timeUnixNano`, `observedTimeUnixNano`, `severityNumber`, `severityText`, `body` and `attributes` — no `traceId`/`spanId`. Only the root span has them (`grep traceId src/*.c` hits `traces.c` and `otlp_pb.c`, never `logs.c`). `OtlpRecord.LogRecord` already models both fields, so a *generic* OTLP producer would populate them, but kotlp itself has nothing to forward yet.

2. **The sink would drop them.** `RunContextLogger.ForwardAppender.transform` rebuilds the event and calls `setMDCPropertyMap(runContextLogger.mdcContext())`, so an MDC entry set on the incoming event is overwritten. Key-value pairs survive that rebuild, but `logEntry(...)` only reads `PROGRESS_KEY` out of them and `LogEntry` has no trace-context field, so nothing would reach the persisted log.

So doing this properly means kotlp stamping the root span's ids onto each log record, plus a `traceId`/`spanId` on `LogEntry`. Both are worth it — it is what would let a user pivot from a task log to the span — but they are a separate change.

core/src/main/java/io/kestra/core/models/tasks/runners/TaskLogLineMatcher.java
loicmathieu: Not filed yet — flagging it back to you rather than opening it unilaterally; say the word and I will.

For the record of what it would cover: kotlp already emits exactly the single span you describe (`exec <command>`, `kind: 1`, `process.exit.code` attribute, status `1`/`2` on success/failure), plus whatever the child exports through kotlp's embedded OTLP receiver, which arrives as additional `resourceSpans`. This PR leaves all of it parsed-but-not-forwarded, reachable as `match.otlp().resourceSpans()`, so the follow-up is purely about the Kestra-side sink.

script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/KotlpUtils.java
```diff
@@ -0,0 +1,65 @@
...
+        return destination;
+    }
+
+    private static void setExecutable(Path path) throws IOException {
+        try {
+            Set<PosixFilePermission> permissions = new HashSet<>(Files.getPosixFilePermissions(path));
+            permissions.add(PosixFilePermission.OWNER_EXECUTE);
+            permissions.add(PosixFilePermission.GROUP_EXECUTE);
+            permissions.add(PosixFilePermission.OTHERS_EXECUTE);
+            Files.setPosixFilePermissions(path, permissions);
```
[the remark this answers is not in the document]
loicmathieu: False positive on the "read, write and execute" wording: the code only ever *adds* execute bits. `Files.copy` creates the file under the umask (typically `rw-r--r--`) and `setExecutable` adds `OWNER_EXECUTE`/`GROUP_EXECUTE`/`OTHERS_EXECUTE` — no write bit is granted to group or others at any point, so the result is `rwxr-xr-x`, the normal mode for an executable.

The others-execute bit is also load-bearing rather than incidental: the binary is copied into the task's working directory, which container-based task runners mount into a container whose user need not be the uid that wrote the file. Restricting to owner-only would break those runners. And the content is not sensitive — it is a public release artifact, byte-identical to the one on the kotlp releases page.

loicmathieu:
kotlp's own PR is merged and v0.1.0 is released, so this one is now aligned with what the wrapper actually emits. Four commits, plus the matching EE one; replies to each review comment are inline above.

| Commit | What |
|---|---|
| `0e538b64da` | `koltp` → `kotlp`, `oltp` → `otlp` everywhere; `kotlpVersion` → `0.1.0` |
| `5b6374c1c8` | the NDJSON fixture becomes a real kotlp recording |
| `262b13bb36` | assertions that the embedded binary and the matcher agree |
| `8c2dc739d1` | re-record from the published release; two doc corrections |
| kestra-ee#9729 `9fe6a55b90` | the same rename through `TaskLogMatchWithEncrypted` |

`:core:test --tests TaskLogLineMatcherTest` and `:script:test --tests '*Kotlp*'` — **25 tests, 0 failures**, running the published v0.1.0 asset (`sha256 7f3cdea4…`).

## The fixture is now a recording, not a sample

It was written by hand from the wire format, so it could only prove the parser agreed with my reading of the spec. It is now a verbatim `--log-dir` capture of the pinned v0.1.0 binary, with only `host.name` replaced — six records over three metric batches.

That turned out to matter, because the recording contains things the hand-written sample did not:

- four metrics the parser had never been fed — `process.cpu.utilization`, `process.memory.virtual`, `process.thread.count`, `process.open_file_descriptor.count`;
- the same series sampled repeatedly, so the cumulative-sum → `Gauge` collapse is exercised *across batches* (23 data points → 10 distinct `(name, tags)` entries) rather than asserted on a single point. The asserted values pin the temporality too: read `process.disk.io` as a delta sum instead and `read` becomes a `Counter` summing its three samples rather than the last one;
- the closing rusage-derived batch, which is where kotlp's own recent fix shows on the consumer side — `process.disk.io{write}` holds at `81788928` instead of restarting at `0`.

The three lines the parser must skip (blank, truncated mid-object as rotation leaves it, well-formed JSON with no OTLP section) are appended after the recording rather than recorded, since kotlp does not produce them on purpose.

## Two gaps the rename exposed

**Nothing asserted that the binary we embed frames records the way the matcher reads them.** That is the failure mode of the envelope-key change and it is silent: valid output, ingested by nobody. `shouldWrapCommandsWithKotlpWhenEnabled` now runs the embedded binary and requires `process.cpu.time` on the run context.

**`--log-dir` does more than mirror.** Upstream made the file the single full copy: with it set, kotlp drops the console framing and prints the child's output raw, so telemetry stops reaching Kestra through the log lines entirely. The javadoc said the option "mirrors every telemetry record", which reads as both copies being live. It now says what happens, and `shouldPassLogDirOptionWhenConfigured` asserts both halves — no metrics on the run context, then metrics once the file is fed back through `parseOtlp`, which is what a caller setting the option has to do.

While correcting that I checked `--log-flush-interval` against the binary: it does not rotate an existing `log.ndjson`, it replaces it. With the option set there is no `log.ndjson` at all, only `log-1.ndjson`, `log-2.ndjson`, … — so a caller collecting one fixed name comes back empty. `KotlpOptions` now says the whole directory has to be shipped back.

## Two things worth a second opinion

- **`0e538b64da` renames a public record component with no `@JsonAlias("oltp")` fallback.** That is deliberate — the component only exists in this unmerged PR series, so nothing outside it can be emitting the old key — but it is worth confirming no build went out with it.
- **`8c2dc739d1` exists because I nearly shipped a bad measurement.** The version marker beside the embedded binary said `0.1.0` while the binary next to it was a local build of kotlp main, and `downloadKotlp` skips the fetch when the marker matches — so the earlier green run was against a binary CI would never use. The behaviours happened to match, but the run proved less than it looked like it did. It is re-recorded and re-run against the published asset.

## Merge order

kestra-ee#9729 has to land together with this one: the EE re-parse path invokes `TaskLogMatch`'s canonical constructor directly, so either side alone leaves EE uncompilable. I could not build EE locally to confirm — `./gradlew` there fails at project evaluation with `rapt_required` on the Artifact Registry credentials, which needs an interactive `gcloud auth application-default login`. The change is a four-line rename, but it is CI-verified only.

loicmathieu (approved):
LGTM

core/src/main/java/io/kestra/core/models/tasks/runners/TaskLogLineMatcher.java
loicmathieu: Opened: #18107.

It covers the sink itself (Kestra has no span equivalent of `runContext.logger()`/`runContext.metric(...)` yet), the trace-stitching problem, and the volume question.

One thing worth flagging from writing it up: it is not only the root span. kotlp points the child at its embedded receiver through `OTEL_EXPORTER_OTLP_ENDPOINT`/`OTEL_TRACES_EXPORTER`, so an instrumented script's own spans are relayed through the same channel with no extra user configuration — a task's internal trace can reach Kestra, not just the fact that a process ran.

The stitching part has a catch: kotlp mints its own trace id per run and does not read `TRACEPARENT`, so its spans are an island. Either kotlp honours an inbound `traceparent` (upstream change, fixes it for every consumer) or Kestra re-parents on ingestion (works today, no release needed). The issue records both.

PR #17874, opened 2026-08-05
fix(ui): re-open error toast when a new error arrives while one is open

### LINKED ISSUE #17871: [UI] Error toast only shows once — silently drops any second error while the first is still open
### Describe the issue

`ErrorToast.vue`'s notification is only ever shown from an `onMounted` hook, which fires once — the first time `coreStore.message` transitions from empty to set (that transition is what makes `<ErrorToast v-if="coreStore.message">` in `App.vue:3` mount the component). After that:

- The toast never auto-dismisses (`duration: 0` in the `KsNotification` call, `ErrorToast.vue:132`), and nothing else ever resets `coreStore.message` back to empty.
- There's no `watch()` on the `message` prop (`ErrorToast.vue`), so when `coreStore.message` is reassigned to a *new* value while the component is still mounted from the first error, nothing re-triggers a new visual notification — the `onMounted` body simply doesn't run again.

Net effect: only the very first error toast shown in a session ever actually appears. Every subsequent one is silently dropped, and the first one sits on screen forever with no way to dismiss it except a page/route change (there's a `watch(route, () => close())` at `ErrorToast.vue:91`, but that only closes it, it doesn't reset `coreStore.message`, so the *next* error still won't show — the component is still mounted, `onMounted` still won't re-fire).

Since `coreStore.message` is used by many different components, this isn't scoped to one feature — any two errors surfaced this way anywhere in the app, with the first one still open, will only show the first.

#### Steps to reproduce

1. Create a new flow (a fresh flow starts with exactly one task by default).
2. Switch to the **Topology** view.
3. Try to delete that only task. An error toast appears (blocked because a flow needs at least one task) — note it never auto-dismisses.
4. Add a second task, then delete one of the two tasks (not the last one). This succeeds silently, as expected.
5. Try to delete the remaining last task again, to trigger the same "can't delete the last task" error a second time.
6. **No toast appears at all** — even though the delete is correctly still blocked.

#### Suggested fix

Add a `watch(() => props.message, ...)` in `ErrorToast.vue` that closes any existing notification and opens a fresh one whenever the message content changes, instead of relying solely on `onMounted`. Might also be worth deciding whether/when `coreStore.message` should reset back to empty (e.g. on manual close), rather than only relying on remount.

Found this while working on #17323 — leaving it as a separate issue since it's a different, broader bug (affects the whole app's error-toast system, not just that flow).

### Environment

- Kestra Version: develop
- Edition: Open Source

### Environment

- Kestra Version: develop

bunnysayzz:
## What does this PR do?

Fixes #17871 — the error toast only ever showed once per session. `ErrorToast.vue` displayed its notification from `onMounted`, which fires only the first time `coreStore.message` transitions from empty to set. Since `coreStore.message` is never reset, every subsequent error while the component stays mounted was silently dropped (and the first toast never auto-dismissed, `duration: 0`).

This affects error surfacing app-wide, not just one feature: any two errors raised through `coreStore.message` while the first toast is still open only ever show the first.

## How was it fixed?

- Extracted the notification logic in `ErrorToast.vue` into a `show()` function.
- Added `watch(() => props.message, show)` — whenever a new error message arrives while the component is mounted, the previously open notification is closed and a fresh one is opened. (`close()` is called first, which is a no-op on the initial show.)
- The initial mount still uses `onMounted(show)` — the watch has no `immediate` flag, so there is no double-show.

## Scope note

This PR intentionally leaves the *auto-dismiss* question (whether `coreStore.message` should reset on manual close, or the toast should get a duration) for a follow-up, keeping this change atomic and focused on the reported bug: subsequent errors now actually surface.

## Tests

Added `ui/tests/unit/components/ErrorToast.spec.ts`:
- **regression:** a second error message while the component is mounted opens a second notification and closes the first
- **existing behavior:** a route change closes the toast without opening a new notification

All UI unit tests for the touched files pass (3/3 incl. the existing `ErrorToastContainer` suite), `vue-tsc` clean, `oxlint` clean.

## Checklist

- [x] I have read the [contribution guidelines](https://kestra.io/docs/developer-guide/contributing)
- [x] Code follows the project's style conventions
- [x] Tests added/updated
- [x] Typecheck and lint pass locally

### DIFF
--- a/ui/src/components/ErrorToast.vue
+++ b/ui/src/components/ErrorToast.vue
@@ -92,7 +92,10 @@
         close()
     })
 
-    onMounted(() => {
+    const show = () => {
+        // Close any previously open notification so a new message can surface it.
+        close()
+
         const error: ErrorEvent = {
             type: "ERROR",
             error: {
@@ -133,7 +136,14 @@
             dangerouslyUseHTMLString: true,
             customClass: isLargeNotification.value ? "error-notification kel-notification__large" : "error-notification",
         })
-    })
+    }
+
+    // The component stays mounted after the first error (coreStore.message is never reset),
+    // so a new message only replaces the prop — onMounted won't fire again. Watch the prop
+    // and re-open a fresh notification each time the message changes.
+    watch(() => props.message, show)
+
+    onMounted(show)
 </script>
 
 <style lang="scss" scoped>
--- a/ui/src/components/filter/configurations/pluginFilter.ts
+++ b/ui/src/components/filter/configurations/pluginFilter.ts
@@ -1,14 +1,16 @@
 import {computed, ComputedRef} from "vue"
 import type {FilterConfiguration} from "@kestra-io/design-system"
 import {useI18n} from "vue-i18n"
+import {usePluginsStore} from "../../../stores/plugins"
 
 export const usePluginFilter = (): ComputedRef<FilterConfiguration> => {
     const {t} = useI18n()
+    const pluginsStore = usePluginsStore()
 
     return computed(() => {
         return {
             title: t("filter.titles.plugin_filters"),
-            searchPlaceholder: t("filter.search_placeholders.search_plugins", {count: 1400}),
+            searchPlaceholder: t("filter.search_placeholders.search_plugins", {count: pluginsStore.plugins?.length ?? 0}),
             keys: [],
         }
     })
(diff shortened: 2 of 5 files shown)


MilosPaunovic:
Hi there, @bunnysayzz! 👋

PR #17875, opened 2026-08-05
fix(worker): declare a real gRPC retry policy for replay-safe RPCs

fhussonnois:
`max-retry-attempts` only capped a retry policy that was never declared, so it had no effect and a controller lost mid-call failed the task. Metadata reads, execution-log reads and the liveness heartbeat now retry on UNAVAILABLE; results, log/metric submissions, the connect handshake and the job stream stay excluded. Health check and retry share one service config, since a second defaultServiceConfig call replaces the first. Default attempts lowered from 10 to 4 now that they are real.


ammeek (approved):
LGTM

PR #17880, opened 2026-08-06
fix(ui): re-trigger the error toast for every new message, not just t…

### LINKED ISSUE #17871: [UI] Error toast only shows once — silently drops any second error while the first is still open

harshcodes17:
ErrorToast.vue only ever built and showed its KsNotification from an onMounted hook, which fires once when coreStore.message first goes from empty to set. Nothing reset coreStore.message afterward and nothing watched the message prop, so any second error while the first toast was still open (duration: 0, never auto-hides) silently never appeared.

Move the notification build into a watch(() => props.message, ..., {immediate: true}) so it re-fires on every new message, closing any previous notification first. Also close the notification on unmount, since coreStore.message can be reset to falsy elsewhere (e.g. FlowRun.vue), which unmounts this component via the parent's v-if without ever telling the notification library to close its popup.

Closes #17871

---

###  Description

`coreStore.message` is used across many components to trigger a toast via `<ErrorToast v-if="coreStore.message">`. The toast itself was only ever built inside `onMounted`, which runs once — the first time `coreStore.message` goes from empty to set. After that:

- The toast never auto-dismissed (`duration: 0`), and nothing ever reset `coreStore.message` back to empty.
- There was no `watch()` on the `message` prop, so reassigning `coreStore.message` to a new value while the component was already mounted did nothing visually — `onMounted` doesn't run twice.

Net effect: only the very first error toast in a session ever actually appeared. Any second error anywhere in the app, while the first was still showing, was silently dropped.

This PR moves the notification-building logic out of `onMounted` into `watch(() => props.message, showNotification, {immediate: true})`, so it re-fires on every new message (closing the previous notification first), and adds `onUnmounted(() => close())` so the visual popup closes if `coreStore.message` is reset elsewhere (e.g. `FlowRun.vue:392`).

###  Related Issue

Closes https://github.com/kestra-io/kestra/issues/17871

### Frontend Checklist

- [ ] All existing E2E tests pass (`npm run test:e2e`) — no existing e2e test covers this toast re-fire behavior; verified manually instead (see below).
Screenshot 2026-08-06 at 2 22 31 PM

### Additional Notes

Manually verified the fix by triggering the same "can't delete the last task" error twice in a row (delete → blocked, add a task, delete a different one, delete the last one again): before this fix, the second attempt showed nothing; after, the toast reappears correctly both times.

**Note for reviewers:** #17870 is still open (not yet merged) and touches the same block in `ErrorToast.vue` (a different, unrelated bug — the toast body text being silently dropped). This branch was created before that fix, so expect a small rebase/conflict here once #17870 merges first.


RemiBarthe (commented):
Thanks a lot for this one, clean fix! 👍

Just one tiny style nit, nothing blocking: I'd probably drop the comments above `showNotification` and `onUnmounted`. They mostly describe what the code is already doing, so I tend to prefer keeping comments for non-obvious context

MilosPaunovic:
Can you handle the comment here @harshcodes17, so we can get this merged today?

harshcodes17:
> Thanks a lot for this one, clean fix! 👍
> 
> Just one tiny style nit, nothing blocking: I'd probably drop the comments above `showNotification` and `onUnmounted`. They mostly describe what the code is already doing, so I tend to prefer keeping comments for non-obvious context

worked on this , please check

RemiBarthe (approved):

PR #17886, opened 2026-08-06
fix(core): accumulate removed task runs during loop iterations for ac…

### LINKED ISSUE #17808: LoopUntil taskrun count is not accurate in execution statistics
### Describe the issue

A LoopUntil release its subtask taskrun list on each iteration, so only the "last LoopUntil subtask taskruns" are included in the execution itself, which is the basis of the statistic computation.

Count each LoopUntil subtask taskrun inside execution statistics.  

### Environment

- Kestra Version: develop

hemantmm:
…curate statistics

closes: #17808

---

### ✨ Description

What does this PR change?  

## What does this PR do?

This PR fixes an issue where the `taskRunCount` in `ExecutionStatistic` under-reports the total number of executed tasks when using `LoopUntil` with nested structures like `Dag`. 

Previously, when a `LoopUntil` triggered a new iteration, `ExecutionService.retryWaitFor()` removed the descendant task runs from the active `taskRunList` to give them a clean state. Because these removed task runs were entirely discarded from the execution, they were lost from the final aggregation in `ExecutionStatistic`, causing metrics dashboards to report an inaccurately low task run count.

**Changes made:**
- **`ExecutionMetadata`**: Added an `accumulatedTaskRunCount` field and an `accumulateRemovedTaskRuns()` method to safely store the count of tasks that are removed during loop iterations.
- **`ExecutionService`**: Updated `retryWaitFor()` to identify descendant task runs before they are removed, and properly accumulate them into the execution's metadata.
- **`ExecutionStatistic`**: Updated the metric aggregation to include both the final active `taskRunList` size and the historically accumulated `accumulatedTaskRunCount`.

image

https://github.com/user-attachments/assets/0b640695-f9b0-4717-a1ce-24aefdefc641


hemantmm:

hemantmm:

loicmathieu:
@hemantmm thanks for your contribution but we decided not to implement this as the solution is to use the `fromJson()` Pebble function to access a loop iteration value as a JSON object.

PR #17907, opened 2026-08-06
feat(flows): add sortable "Last Modified" column to the flows list

### LINKED ISSUE #17872: Add "Last Modified" column to the Flows list page
### Feature request

Add a "Last Modified" (last updated) column/date to the flows list page (the main Flows table), so users can see at a glance when each flow was last changed.

### Context

Currently, the last-modified timestamp is only visible per-flow, on the Revisions tab. To notice that a flow changed, a user has to open each flow individually and check its revision history — there's no way to scan the whole flow list for recently (or unexpectedly) modified flows.

### Why it's useful

- Quickly spot flows that were updated unexpectedly (e.g. by another team member, automation, or CI/CD sync) without digging into each flow's revisions.
- Makes auditing changes across many flows much faster, especially in namespaces with a large number of flows.

### Suggested implementation

- Add a sortable "Last Modified" column to the flows list table, sourced from the latest revision's timestamp.
- Optionally allow sorting/filtering the flow list by this column.

_Reported via customer feedback (Pylon #2140)._

nkwiatkowski:
### What this does

Adds a **"Last Modified"** column to the flows list page, sourced from the flow's `updated` timestamp (the timestamp of the latest revision). The column is:

- **opt-in** — off by default, enabled from the table column picker (persisted per user), like the other optional flows columns;
- **server-sortable** — clicking the header issues `?sort=updated:asc|desc`, handled by the existing sort plumbing.

Closes https://github.com/kestra-io/kestra/issues/17872

### The `toFlow()` fix

While wiring this up, the Elasticsearch backend returned flow rows **without** `updated` even though the value was stored. Root cause: `FlowWithSource.toFlow()` copied `revision` but never `updated` (its sibling factory `of()` does copy it). With `serialization-inclusion: non_null`, the null field was omitted entirely. The JDBC backends were unaffected because they deserialize the stored `value` JSON directly rather than going through `toFlow()`.

Fix: `toFlow()` now copies `updated`, mirroring `of()`.

**Equality safety:** `updated` is not part of `Flow.equals()`/`hashCode()` — `Flow` uses `@EqualsAndHashCode(callSuper=false)` and `updated` is an inherited `AbstractFlow` field — and `equalsWithoutRevision` (the source-change dedup) already excludes `updated`. So this change cannot affect equality/dedup.

### Tests

- `FlowWithSourceTest#toFlowShouldPreserveRevisionAndUpdated` — unit guard on the fix.
- `AbstractFlowRepositoryTest#shouldSortByUpdatedWhenSortRequested` — shared repository test asserting `sort=updated` orders rows in both directions; runs across H2/MySQL/Postgres **and** the Elasticsearch backend (EE inherits this abstract test).

### Enterprise companion

The Elasticsearch index needs the `updated` field mapped + a reindex of existing docs for sorting to work there — handled in the companion EE PR.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/FlowWithSource.java
+++ b/core/src/main/java/io/kestra/core/models/flows/FlowWithSource.java
@@ -39,6 +39,7 @@ public class FlowWithSource extends Flow {
             .retry(this.retry)
             .sla(this.sla)
             .checks(this.checks)
+            .updated(this.updated)
             .build();
     }
 
--- a/core/src/test/java/io/kestra/core/models/flows/FlowWithSourceTest.java
+++ b/core/src/test/java/io/kestra/core/models/flows/FlowWithSourceTest.java
@@ -1,5 +1,6 @@
 package io.kestra.core.models.flows;
 
+import java.time.Instant;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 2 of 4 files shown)


loicmathieu (approved):
LGTM

PR #17908, opened 2026-08-06
fix(system): register the system worker under its own service type

fhussonnois:
The embedded system worker registered as a regular WORKER, so the services view listed it as a second worker. It now registers as SYSTEM_WORKER and liveness handling uses ServiceType#isWorker(), so its jobs are still re-emitted on crash.

Closes https://github.com/kestra-io/kestra-ee/issues/7918. Closes https://github.com/kestra-io/kestra-ee/issues/9718.


ammeek (approved):
lgtm

PR #17910, opened 2026-08-06
fix(executions): bypass ACL for internal execution-streaming lookups

loicmathieu:
## Summary

Fixes kestra-io/kestra-ee#8824 — a `Webhook` trigger with `wait: true` hangs indefinitely in EE's distributed deployment, even though the execution completes successfully.

**Root cause:** `ExecutionStreamingService`'s queue-consumer fan-out and its post-registration catch-up guard both re-read the execution via the ACL-enforcing `executionRepository.findById(tenantId, id)`. That call runs on a queue-polling thread with no authenticated principal. In EE, the ACL override degrades to `DSL.falseCondition()` when there is no allowed namespace for the (absent) user — so the lookup always returns empty on a background thread, the terminal SSE event is silently dropped. This explains every symptom in the issue: standalone works (ACL bypassed off-request-thread there), distributed fails, the errors keep appearing after the execution is terminal and persisted, and `/executions/search` still finds it (that runs inside the authenticated request).

`ExecutionDependenciesStreamingService` (the execution-dependencies SSE panel) has the exact same shape of bug.

## Changes

- `core/.../services/ExecutionStreamingService.java` — switch both internal lookups (queue-consumer fan-out, post-registration catch-up guard) to `findByIdWithoutAcl`. Authorization for who may follow an execution is already enforced where subscribers register (`/follow`'s `isExecutionAllowedOrThrow` in EE, or the webhook's secret key for the deliberately anonymous `wait: true` path), so bypassing ACL on these internal reads is safe.
- `webserver/.../services/ExecutionDependenciesStreamingService.java` — same fix, plus replaces a bare `.orElseThrow()` with a logged early-return (throwing out of a queue-subscriber callback was a separate hazard: it could crash/redelivery-loop the shared queue consumer under `fail-fast=true`).
- `core/.../services/ExecutionStreamingServiceTest.java` — new regression test that mocks `findById` to always deny (simulating the EE ACL fallback) while `findByIdWithoutAcl` returns the real execution. Verified this test fails against the pre-fix code and passes with the fix.

## Test plan

- [x] `./gradlew :core:test --tests "ExecutionStreamingServiceTest"` — both tests pass
- [x] New test verified to fail against the pre-fix code (reproduces `Unable to find the execution id ... for event type TERMINATED`) and pass with the fix
- [x] `:core:compileTestJava` / `:webserver:compileJava` — clean, no new warnings
- [x] Code review via the `code-review` agent — approved, no blocking issues
- [ ] Manual end-to-end verification on a distributed EE stack (postgres + separate webserver/executor, `kestra.server-type: WEBSERVER`) with the flow from the issue — OSS cannot reproduce EE's ACL divergence, so this is the only check that fully closes the loop

### DIFF
--- a/core/src/main/java/io/kestra/core/services/ExecutionStreamingService.java
+++ b/core/src/main/java/io/kestra/core/services/ExecutionStreamingService.java
@@ -75,7 +75,9 @@ public class ExecutionStreamingService {
             Map<String, Pair<FluxSink<Event<Execution>>, Flow>> executionSubscribers = subscribers.get(event.executionId());
 
             if (!MapUtils.isEmpty(executionSubscribers)) {
-                Optional<Execution> execution = executionRepository.findById(event.tenantId(), event.executionId());
+                // This fan-out runs on the queue-polling thread, which has no authenticated principal,
+                // so the ACL-enforcing findById would always deny authorization on EE.
+                Optional<Execution> execution = executionRepository.findByIdWithoutAcl(event.tenantId(), event.executionId());
                 if (execution.isEmpty()) {
                     log.error("Unable to find the execution id {}", event.executionId());
                     return;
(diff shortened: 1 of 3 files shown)


fhussonnois (approved):
LGTM!

PR #17911, opened 2026-08-06
fix(design-system): clear pre-applied filter cache on reset (#17911)

### LINKED ISSUE #17351: [KV Store] Filter reset can restore stale or unexpected values
### Describe the issue

In the KV Store, filter values can reset to inconsistent stale values. If a user clears all filters, creates a new filter, and then uses reset, the filter does not reset to an empty value. Instead, it defaults to another value, which is often not the previous value the user had selected.

This makes the filter reset behavior unpredictable.

1. Open the KV Store page.
2. Apply one or more filters.
3. Clear all filters.
4. Create a new filter.
5. Click reset for that filter.
6. The filter value does not reset to empty; it defaults to an unexpected value.

Video evidence:

https://github.com/user-attachments/assets/1bd9efd9-7a08-4085-8a57-1c08ccf563dc

### Expected Behaviour
After clearing filters, a newly created filter should reset to the empty/default state.

Reset should not repopulate stale or unrelated values.

### Environment

- Kestra Version: `1.3.27`
- Observed: `Thu, Jul 9, 2026 2:24 PM IST`
- Edition: Open Source

harshit-seth:
Closes https://github.com/kestra-io/kestra/issues/17351

   ## ✨ Description
   Fixes stale filter values reappearing after reset in KV Store (#17351).

   When a user clears all filters and adds a new one, clicking the reset (circular arrow) icon restores previously checked values instead of showing an empty list. The issue was caused by the preAppliedFilters cache not being cleared when clearFilters() ran — and by the same cache not being cleared when a filter chip was dismissed directly via its X (removeFilter), which reproduces the same stale-value symptom via a different path.

   This PR adds clearPreApplied() and clearPreAppliedKey() to usePreAppliedFilters, and calls them from clearFilters() and removeFilter() respectively.

   ## 🎨 Frontend Checklist
   - [x] New unit test added covering both clearFilters and removeFilter cache eviction (3/3 passing locally)
   - [x] Existing E2E tests pass (CI)

   ## 🛠️ Backend Checklist
   No backend changes.

   ## 📝 Notes
   - Addressed review feedback from @MilosPaunovic: cache eviction now also happens on removeFilter, not just clearFilters.
   - Unit test added: `ui/packages/design-system/tests/units/Data/KsFilter/useFilterActions.test.ts`


build.gradle
MilosPaunovic, line 1: Please remove the changes made in this file, as they are not related to the issue you're trying to solve.

gradle.properties
MilosPaunovic, line 1: Please remove the changes made in this file, as they are not related to the issue you're trying to solve.

MilosPaunovic (changes requested):

gradle.properties
harshit-seth: thanks for catch — these were accidentally pulled in from a local develop merge on my end. Reverted both files to match develop in 34194493b, so the PR now only touches the filter composables. Let me know if anything else needs a look.

harshit-seth:
Hi @MilosPaunovic,

Update on this PR:

- `build.gradle` and `gradle.properties` changes have been reverted — they were accidentally committed during a merge and are unrelated to this fix.
- The failing `TriggerControllerTest` (H2 "Timeout trying to lock table" on QUEUES) is a pre-existing flaky test, unrelated to this change — it passed 9/9 on retry. Filed separately for tracking: #17958
- The core fix remains scoped to clearing the `preAppliedFilters` cache in `clearFilters()`, as described in #17351

All checks are green except the known flaky test above. This should be ready for review/merge whenever you get a chance. Let me know if anything else is needed!

MilosPaunovic:
There is a merge conflict now @harshit-seth.

build.gradle
MilosPaunovic: @harshit-seth This file was not reverted.

MilosPaunovic (changes requested):

build.gradle
harshit-seth: Hey Milos, resolved the merge conflict in build.gradle (it's now identical to develop) and all checks are green now, including the previously flaky Backend - Tests. PR should be ready for your review whenever you get a chance!

MilosPaunovic (changes requested):
Thanks for digging into this — the root cause is right and the fix is appropriately minimal. `usePreAppliedFilters` is only ever read by `FilterEditPopper.resetState`, so evicting the cache in `clearFilters()` does fix the reported flow without side effects.

Two things before this can go in:

**1. The stale value still comes back without touching Reset.** `removeFilter` also empties the tree but leaves the cache behind:

1. Open `/kv?filters[key][CONTAINS]=foo`
2. Dismiss the chip via its **X** (not the top-level Reset)
3. Re-add a `key` filter, click reset in the popper → `foo` is back

Same symptom, equally likely path. Please also drop the per-key entry in `removeFilter`, right next to the existing `dismissed.dismissDefaultVisibleKey(found.key)`.

**2. Please add a test.** `grep` output isn't verification, and nothing currently covers these composables — so this is silently re-breakable. A Vitest unit test in `ui/packages/design-system/tests/units/` asserting `hasPreApplied(key) === false` after `clearFilters()` is enough.

Nits:

- Group `type PreApplied` with the adjacent `Tree` / `Dismissed` aliases instead of separating it with a blank line.
- Title: scope should be `design-system` (`filter` isn't a valid scope) and the trailing id is the PR number, so `fix(design-system): clear pre-applied filter cache on reset (#17911)`.
- The description is pasted twice — worth trimming, and please put `Closes https://github.com/kestra-io/kestra/issues/17351` as the first line so GitHub links the issue on merge.

harshit-seth:
@MilosPaunovic  Addressed everything:
- `removeFilter` now evicts the per-key pre-applied cache entry via `preApplied.clearPreAppliedKey(found.key)`, right next to `dismissed.dismissDefaultVisibleKey`.
- Added a Vitest unit test (`useFilterActions.test.ts`) covering both `clearFilters` and `removeFilter` — 3/3 passing locally.
- Nits: `PreApplied` type grouped with `Tree`/`Dismissed`, title updated to the correct scope, description trimmed (no duplicate, `Closes` as first line).

Ready for another look whenever you get a chance!

MilosPaunovic (approved):
Both points addressed, thanks for the quick turnaround. I checked that the new test genuinely runs rather than just sitting there - the `unit` vitest project includes `tests/**/*.test.ts` and `Frontend - Design System tests` is green - and `clearPreAppliedKey` is correctly placed next to `dismissed.dismissDefaultVisibleKey(found.key)`.

I pushed one small commit on top (17b14ada3e) rather than send you round-tripping over nits:

- Dropped the trailing whitespace on the blank line between `removeFilter` and `updateFilter` (`.editorconfig` sets `trim_trailing_whitespace=true`).
- Simplified `clearPreAppliedKey` to delete from the Set and Map in place, matching `markAsPreApplied` just above, instead of cloning both collections and reassigning the refs.

Verified locally before pushing: your 3 new tests pass, all 75 `tests/units/Data` tests pass, eslint is clean on the changed files, and `npm run check:types` passes.

Approving - just let CI go green on that commit before merging.

Non-blocking note for later: the cache is keyed by filter key alone, while the tree deliberately allows several chips sharing a key with different comparators. So dismissing one `key` chip now evicts the default a sibling `key` chip would have restored. `markAsPreApplied` already overwrote same-key entries, so this is not a regression, and resetting to empty beats resetting to a sibling's value - but the cache would have to be keyed on `(key, comparator)` to be exact.

PR #17912, opened 2026-08-06
feat(core)!: adopt RFC 9457 problem details as the single API error format

fhussonnois:
### ✨ Description

Adopts [RFC 9457 Problem Details](https://www.rfc-editor.org/rfc/rfc9457.html) as the **single** error representation of the Kestra API, served as `application/problem+json`.

The same logical failure previously came back as a Micronaut HATEOAS `JsonError`, as `{status, message}`, as a bare `text/plain` string, or as an empty body with nothing to act on — so no client could write one error handler. None of it was in the OpenAPI spec, so generated SDKs exposed no error type at all.

`type` is a stable, documented URI and the only member clients branch on. It replaces matching on English prose: the UI checked whether a message contained `"already exists"` to tell a duplicate from a real failure, so rewording a backend message silently broke the frontend. The coupling ran both ways — backend messages were deliberately shaped `"<reason>: <detail>"` so the UI could split on the first colon for a toast title.

**Backend**

- `io.kestra.core.models.errors` holds the model, a closed `ProblemType` catalogue, and a `ProblemMapper` registry that Enterprise contributes to without editing the OSS table.
- `ErrorController` drops from 21 handlers to 6, because the mapping now lives in the registry. A `JsonErrorResponseBodyProvider` covers errors that skip `@Error` routes, and `ProblemResponseFilter` fills a body into failures that reached the client without one — filters returning `HttpResponse.unauthorized()`, routes returning a bare status or a plain string.
- `ProblemFormatExclusion` keeps formats owned by another specification out of it: the MCP JSON-RPC transport here, SCIM in Enterprise.
- Authorization denials that surfaced as `500` with a stack trace are now `403`, so routine permission failures stop polluting error dashboards.
- Server errors carry a fixed `detail` plus a `traceId`; the real message goes to the log under that same id instead of being echoed to the caller.
- `ViolationPaths` reports validation as `errors[]` with both an RFC 6901 JSON Pointer and the friendlier Kestra path, replacing a payload of ten mostly-null fields. It also fixes a path bug — `tasks[0].type` rendered as `my-task.type`, dropping the collection name, so the message and the structured field disagreed.
- The OpenAPI spec documents the contract on every operation, so generated clients get a real error type. Every 4xx/5xx response now has a documented body; previously `403` and `500` were declared nowhere at all.

**Frontend**

- Problem documents are parsed once, in the shared SDK runtime. `response.data` *is* the document, so the two divergent shapes the app used to see converge.
- The toast reads `title` and `detail` directly. The colon split is gone, and error titles are localized for the first time via `errors.problems.<slug>.title`, falling back to the wire `title` so a new server-side type renders correctly without a UI release.
- `errors[]` renders per field, `traceId` is shown on server errors, and both prose matchers are replaced by `type` comparisons.

### 🔗 Related Issue

Part-of: https://github.com/kestra-io/kestra-ee/issues/9671

Deliberately not a closing keyword — the docs deliverable named in that issue (the `type` URI registry) ships separately, and a cross-repo keyword would not close it from here anyway.

### 💥 Breaking Changes

- `Content-Type` on every error response is `application/problem+json`, not `application/json`.
- `message` removed — use `detail` (specific to this occurrence) or `title` (stable per `type`).
- `_links` removed.
- `_embedded.errors[]` removed — use `errors[]`, whose entries are `{detail, pointer, path}`.
- `invalids[]` removed on bulk endpoints — per-item failures are `errors[]` entries carrying `path` and their own `type`.
- Branch on `type`, never on message text.
- Corrected `5xx` → `4xx` (bug fixes; clients stop retrying these):
  - `ResourceAccessDeniedException`, `SecurityException`, `ToolPermissionDeniedException` → `403`
  - `SecretNotFoundException` → `404`
  - `FlowBlockedException`, `KilledException`, `LockException` → `409`
  - `KVStoreException`, `NoMatchingWorkerQueueException`, `InvalidTriggerConfigurationException`, `TypeConversionException`, `WebhookInputRenderException` → `422`
  - migration exceptions → `503`; `TimeoutExceededException` → `504`
- Corrected `4xx` → `4xx`: "already exists" is now `409 entity-already-exists` everywhere. Flow and dashboard creation previously returned `422`, because the conflict was tunnelled through bean validation. It had five different spellings across the codebase spanning three statuses; `AlreadyExistsException` unifies them.
- The one `4xx` → `5xx` change: `IllegalStateException` is no longer mapped to `409` generically. Throw sites that genuinely mean a conflict or an unprocessable entity now raise a typed exception; everything else reports `500`, which is what it is. Nothing in the UI depended on the old `409` — the Copilot's `turnInProgress` comes from `ConflictException`.
- `detail` on a `5xx` is a fixed, non-revealing string. The exception message reaches only the server log, correlated by the response's `traceId`.
- Removed types: `BulkErrorResponse`, `ApiErrorResponse`, everything under `webserver/exceptions/`, and the `ErrorController.jsonError(...)` statics.
- Custom HTTP reason phrases dropped — illegal in HTTP/2, and the reason now lives in `title`.
- `HttpClientResponseException.getMessage()` now returns the raw JSON document, because Micronaut's error decoder only parses exactly `application/json`. Assert on `detail` instead.

### 🎨 Frontend Checklist

- [ ] All existing E2E tests pass (`npm run test:e2e`) — **not run**, needs a live backend
- [ ] Screenshots or video recordings attached showing the `UI` changes — **not attached**; the error toast is a visible change and worth a look during review

### 🛠️ Backend Checklist

- [x] All unit tests pass — see the note below on one environment-dependent failure; `integrationTest` was **not** run

### 📝 Additional Notes

- **Companion PR:** kestra-io/kestra-ee#EE_PR_NUMBER. Same branch name in both repos, which EE CI relies on to resolve the OSS branch — merge this one first.
- **Docs still outstanding:** the `type` URIs will not resolve until the registry lands in `kestra-io/docs`. That is the remaining piece of the issue.
- Verified: `:core:test` 3157 pass, `:webserver:test` 1107 with one failure — `UnknownHostException: dummyjson.com`, an external DNS dependency unrelated to this change. UI: 1047 unit tests, `check:types` and lint clean.
- Only `en.json` carries the new i18n keys; CI generates the other locales.
- A grep-guard unit test fails the build if `_embedded`, `invalids` or `response.data.message` is reintroduced. The closed `ProblemDetail` type alone is not enough, because a `catch (e: any)` call site bypasses it.

### 🤖 AI Authors

Why did the cat sit on the keyboard during the migration? It heard there were `errors[]` to curl up in. 🐱


webserver/src/main/java/io/kestra/webserver/errors/KestraProblemMappings.java
```diff
@@ -0,0 +1,120 @@
...
+import io.micronaut.web.router.exceptions.UnsatisfiedBodyRouteException;
+import io.micronaut.web.router.exceptions.UnsatisfiedQueryValueRouteException;
+import io.micronaut.web.router.exceptions.UnsatisfiedRouteException;
+import jakarta.inject.Singleton;
+import jakarta.validation.ConstraintViolationException;
+
+/**
+ * The Open Source exception to {@link ProblemType} table.
+ *
+ * <p>{@code IllegalStateException} is deliberately absent: it is a JDK programming-error signal, not a domain
```
loicmathieu, line 56: It would be great to open a followup so we stop using this exception.
I'm pretty sure it's used in a lot of plugins too :(

webserver/src/main/java/io/kestra/webserver/errors/KestraProblemMappings.java
```diff
@@ -0,0 +1,120 @@
...
+ */
+@Singleton
+public class KestraProblemMappings extends ExceptionTypeProblemMapper {
+    @Override
+    protected void register(final BiConsumer<Class<? extends Throwable>, ProblemType> to) {
+        // Unreadable or undecodable request payload.
+        to.accept(JsonParseException.class, ProblemTypes.INVALID_JSON);
+        to.accept(InvalidTypeIdException.class, ProblemTypes.INVALID_PLUGIN_TYPE);
+        to.accept(InvalidFormatException.class, ProblemTypes.INVALID_FORMAT);
+        to.accept(JsonMappingException.class, ProblemTypes.INVALID_JSON_MAPPING);
```
loicmathieu: I would also use INVALID_JSON and remove one type

webserver/src/main/java/io/kestra/webserver/errors/KestraProblemMappings.java
```diff
@@ -0,0 +1,120 @@
...
+        to.accept(DeserializationException.class, ProblemTypes.INTERNAL_ERROR);
+
+        // Request could not be bound to the route.
+        to.accept(UnsatisfiedBodyRouteException.class, ProblemTypes.INVALID_REQUEST_BODY);
+        to.accept(UnsatisfiedQueryValueRouteException.class, ProblemTypes.INVALID_QUERY_PARAMETER);
+        to.accept(UnsatisfiedRouteException.class, ProblemTypes.BAD_REQUEST);
+
+        // Entity validation.
+        to.accept(ConstraintViolationException.class, ProblemTypes.VALIDATION_FAILED);
+        to.accept(BulkValidationException.class, ProblemTypes.BULK_VALIDATION_FAILED);
```
loicmathieu, line 87:
```suggestion
        to.accept(BulkValidationException.class, ProblemTypes.VALIDATION_FAILED);
```

Again, I would use less types

webserver/src/main/java/io/kestra/webserver/errors/KestraProblemMappings.java
```diff
@@ -0,0 +1,120 @@
...
+        to.accept(ValidationErrorException.class, ProblemTypes.RESOURCE_VALIDATION_FAILED);
```
loicmathieu: Maybe also just use VALIDATION_FAILED

webserver/src/main/java/io/kestra/webserver/errors/KestraProblemMappings.java
```diff
@@ -0,0 +1,120 @@
...
+        to.accept(IllegalArgumentException.class, ProblemTypes.INVALID_ARGUMENT);
+        to.accept(TypeConversionException.class, ProblemTypes.INVALID_FORMAT);
+        to.accept(KVStoreException.class, ProblemTypes.INVALID_ENTITY);
+        to.accept(NoMatchingWorkerQueueException.class, ProblemTypes.INVALID_ENTITY);
+        to.accept(WebhookInputRenderException.class, ProblemTypes.INVALID_REQUEST_BODY);
+
+        // Authorization denials. 403 rather than a server error, so they are not recorded as incidents and
+        // clients do not retry them.
+        to.accept(ResourceAccessDeniedException.class, ProblemTypes.FORBIDDEN);
+        to.accept(SecurityException.class, ProblemTypes.FORBIDDEN);
```
loicmathieu, line 105: I think we also use it for encryption soit is not always FORBIDEN

webserver/src/main/java/io/kestra/webserver/errors/ProblemFactory.java
```diff
@@ -0,0 +1,143 @@
...
+                instance,
+                status,
+                type.slug(),
+                detail == null ? "" : detail,
+                throwable
+            );
+            return new ProblemDetail(type.uri(), type.title(), status, SERVER_ERROR_DETAIL, instance, List.of(), traceId);
+        }
+
+        if (log.isTraceEnabled()) {
```
loicmathieu:
```suggestion
        if (log.isDebugEnabled()) {
```

webserver/src/main/java/io/kestra/webserver/errors/ProblemFactory.java
```diff
@@ -0,0 +1,143 @@
...
+            log.trace(
```
loicmathieu:
```suggestion
            log.debug(
```

I just changed it to debug!

webserver/src/main/java/io/kestra/webserver/errors/ProblemFactory.java
```diff
@@ -0,0 +1,143 @@
...
+        final int status,
+        final String detail,
+        final List<ProblemError> errors
+    ) {
+        String instance = request == null ? null : request.getPath();
+
+        if (500 <= status) {
+            String traceId = traceId();
+            log.error(
+                "Problem [traceId={}] {} {} -> {} {}: {}",
```
loicmathieu, line 112: If OTEL is enabled, the traceId would be inside the MDC so it's a little redundant.
But I agree that it also enable searching for textual logs as we didn't display the traceId by default, it's up to the user to update our default log format so I'm OK with it anyway.

webserver/src/main/java/io/kestra/webserver/errors/ProblemType.java
```diff
@@ -0,0 +1,48 @@
...
+ * rather than URNs per RFC 9457 §3.1.1, which notes that moving from non-resolvable to resolvable later
+ * would itself be a breaking change.
+ *
+ * <p>The title must never be parameterised: it has to stay identical across every occurrence of the same
+ * type so clients can use it as a translation key. Per-occurrence text belongs in
+ * the problem document's {@code detail} member.
+ */
+public record ProblemType(String slug, String title, int status) {
+    /** Base of every {@code type} URI. Each slug appended to this resolves to that type's documentation. */
+    public static final String BASE_URI = "https://kestra.io/docs/api-reference/problems/";
```
loicmathieu, line 20: @aj-emerich would need to create this page and adds an entry for each problem type!
So a comment inside the problem type mapping code saying the doc must be updated each time a new type is added would be great to avoid asking for it!

loicmathieu: OK, see the comment in the next file but this should be emphasise!

loicmathieu (changes requested):
LGTM, love it, so much clear!

Some nits to discuss...

MilosPaunovic:
Review of the UI part only (`ui/**`).

Overall this is well executed: prose-matching is genuinely gone (the `already exists` regex in `useApplyDraft`), both HTTP paths (generated SDK and the `useClient` facade) now expose one identical problem shape, the frontend slug catalog matches the backend `ProblemTypes` catalog exactly, and the `noLegacyErrorFields` guard test with its anti-vacuous glob check is a nice touch. Findings, ranked:

1. **Bug - translation key mismatch renders a raw i18n key in every 5xx toast.** `ErrorToastContainer.vue` calls `$t("errors.trace_id")`, but `en.json` defines the key as `"trace id"` (space, not underscore). The app's `createI18n` has no `missing` handler, so the toast footer will literally show `errors.trace_id` next to the trace id on every server-error toast.

2. **Only `en.json` was touched - every other locale is now missing all the new keys** (`errors.problems.*`, the new `errors.400/409/422/500/503` blocks, `errors.generic`, `errors.trace id`). Degradation is graceful (`te()` misses and the server's English `title` is shown), but the localized-error-titles feature ships dark for non-English users until the scheduled auto-translate workflow backfills. Running `npm run translations:generate` before merge would close the gap.

3. **Empty toast body when a bulk catch gets a non-problem error.** In `Executions.vue` (both bulk handlers) and `TriggersManage.vue`, `(problem?.errors ?? []).map(...)` produces `[]` for a network failure or a problem without `errors[]`, and `toast.error([], title)` then hands `KsMarkdown` an empty array as `content` (the `message as string` cast hides it) - the user gets a title with a blank body. The old code threw a `TypeError` on `e?.invalids.map`, so this is still an improvement, but a fallback to `problem?.detail ?? <generic message>` when the list is empty would finish the job.

4. **`problem-types.ts` overpromises.** The doc comment says "every constant here is exercised by a unit test" and "only types with a behavioural consequence belong here", but only `ENTITY_ALREADY_EXISTS` and `ENTITY_NOT_FOUND` are branched on; `CONFLICT`, `TOO_MANY_REQUESTS`, `FORBIDDEN`, `UNAUTHENTICATED`, `SERVICE_UNAVAILABLE`, `INTERNAL_ERROR`, `VALIDATION_FAILED`, `BULK_VALIDATION_FAILED` have no branch and no test. Either trim the catalog to its own admission rule or soften the comment.

5. Nits: `noLegacyErrorFields.spec.ts` has two separate `import ... from "node:fs"` statements, and the first `response...message` regex (`/response\??\.?\.?data\??\.\.?message\b/`) looks garbled and is redundant next to the two sane patterns that follow (the EE copy of the spec is cleaner on both counts). Incidental indentation churn in `en.json` (the `executionProgress` block re-indented 8 to 6 spaces).

Verified and fine: `KsId` needs no import (globally registered); no leftover writers of the old `coreStore.message` shape; no leftover `lodash/escape` usage; the 400-now-rejects-like-any-error change is safe since every in-repo reader of the old shape was updated and the guard test blocks regressions; the new `errors.<status>` blocks are consumed by `Errors.vue`.

One cross-repo note: this must merge together with https://github.com/kestra-io/kestra-ee/pull/9754 (OSS first, EE immediately after) - if OSS merges alone, EE develop's `ServiceTriggers.vue` still reads `e?.invalids` and `validatePassword` still dereferences `data.message`, both of which become TypeErrors in the catch.

webserver/src/main/java/io/kestra/webserver/errors/KestraProblemMappings.java
fhussonnois: For encryption we use the `GeneralSecurityException` which is not a subtype of `SecurityException` so that's OK.

webserver/src/main/java/io/kestra/webserver/errors/KestraProblemMappings.java
fhussonnois: I would prefer keeping the BULK_VALIDATION_FAIlED for now mostly because this impacts 14 endpoints that have to be moved from HTTP code 400 -> 422. UI also contains some dedicated logic for bulk errors.

loicmathieu (approved):
LGTM

aj-emerich:
@fhussonnois this requires a breaking change documentation guide then? And what exactly needs to happen on the docs side? Is this auto-generated on release? This link is already being indexed by search and 404ing

http://kestra.io/docs/api-reference/problems/not-found

PR #17941, opened 2026-08-07
fix(executions): show the BOOL input toggle as on when its default is true, and delete the unused normalize twin that let it regress

### LINKED ISSUE #13214: `BOOL` input set to null when no default value set
### Describe the issue

When using `BOOL` with no default value, it shows it in the input menu as `false` and allows you to execute (aka the field is not blank). 

Image

But when you run the workflow, it fails because the `BOOL` input is `null`. The `BOOL` input should return false as default so when you open the execute menu, the boolean is toggled off unless otherwise stated.

Image

```yaml
id: 01_advent
namespace: advent

inputs:
  - id: confirm
    type: BOOL

tasks:
  - id: if
    condition: "{{ inputs.confirm }}"
    then:
      - id: confirm_log
        message: "Hello!"
```

### Environment

- Kestra Version: develop

MilosPaunovic:
Closes https://github.com/kestra-io/kestra-ee/issues/9772.
Closes https://github.com/kestra-io/kestra/issues/13214.

## What

A BOOL input declared with `defaults: true` rendered with its toggle **off**, and the execution then ran with `false`.

`Input.defaults` is a `Property`, which serialises as its expression string — so `defaults: true` reaches the form as the **string** `"true"`. `el-switch` compares `modelValue` against `activeValue` by identity, and `"true" !== true`, so during its own setup it emits `update:modelValue` = `false`. That does two things: it turns the toggle off, and it runs the form's change handler. The change bumps `inputGeneration`, so the in-flight validate response carrying the real default is discarded as stale, and the follow-up round-trip submits `false` — the toggle is off for good.

The fix coerces a BOOL value to a real boolean in `normalize()`, so the switch accepts the default.

This affects the flow execute form and, through the same shared `InputsForm`, the EE Apps `CreateExecutionForm` block.

## Why it came back, and what stops the next one

This is a regression. The identical fix landed on `releases/v1.3.x` as 6265761bfa (for https://github.com/kestra-io/kestra-ee/issues/8978) but **never reached `develop`** — `git merge-base --is-ancestor 6265761bfa develop` is false. There were no tests on either branch, so nothing caught the drift.

There was also a second way for it to come back, which this PR removes. `normalizeForComponents` was a near-duplicate of `normalize` with **no callers anywhere** — not in `ui/src`, `ui/tests`, `ui/packages`, `kestra-ee/ui-ee` or `ui-libs`. It had already silently drifted from the function it mirrors: it checked `value === null` first (so a null BOOL yielded `undefined` rather than `false`) and never grew the `MULTISELECT`, `JSON` and `YAML` branches `normalize` gained later. Keeping it meant writing every normalization rule twice, which is exactly how a fix goes missing again — someone edits the copy they landed on and the two diverge. It is deleted in its own commit, so the fix stays reviewable on its own.

## Tests

Two layers, since the previous fixes had none:

- `tests/unit/utils/inputs.spec.ts` — `normalize` always yields a real boolean for BOOL (`"true"`, `"false"`, real booleans, `undefined`, `null`), and the retired `BOOLEAN` type's `"undefined"` tri-state is left alone.
- `tests/unit/components/inputs/inputsFormBoolDefault.spec.ts` — mounts `InputsForm` **for real** (no `shallow`), because the bug lives inside `el-switch`'s setup and a stubbed switch would happily accept the string. The validate stub mirrors `FlowInputOutput` (a default is resolved only when the form submits no value), which is what makes the stale-response round-trip reproducible. Covers: toggle on for `defaults: true`, `isDefault` not clobbered, off for `defaults: false`, off with no default, the user can still turn it off, and a BOOL nested in a `FORM` (dotted leaf id).

Verified both directions — with the fix reverted, the helper assertions and the 3 "should be on" component tests fail; with it, the full OSS unit suite is green (139 files / 1156 tests). `npm run check:types`, `oxlint` and `eslint` clean.

### DIFF
--- a/ui/src/utils/inputs.ts
+++ b/ui/src/utils/inputs.ts
@@ -165,8 +165,13 @@ export function normalize(type: InputType | undefined, value: any) {
 
     if (type === "BOOLEAN" && value === undefined) {
         res = "undefined"
-    } else if (type === "BOOL" && value === undefined) {
-        res = false
+    } else if (type === "BOOL") {
+        // `defaults` is a Property, which serialises as its expression string, so `defaults: true`
+        // reaches the form as the string "true". KsSwitch (el-switch) only accepts a real boolean:
+        // given anything else it emits `update:modelValue` = false during setup, which both turns the
+        // toggle off and marks the input as user-edited — so the validate response can no longer
+        // restore the default. Coerce here.
+        res = value === true || value === "true"
     } else if (value === null || value === undefined) {
         res = undefined
     } else if (type === "DATE" || type === "DATETIME") {
@@ -200,8 +205,10 @@ export function normalizeForComponents(type: InputType | undefined, value: any) 
         res = JSON.stringify(res).toString()
     } else if (type === "BOOLEAN" && value === undefined) {
         res = "undefined"
-    } else if (type === "BOOL" && value === undefined) {
-        res = false
+    } else if (type === "BOOL") {
+        // See normalize(): a BOOL default arrives as the string "true"/"false" and el-switch resets
+        // anything that isn't a real boolean.
+        res = value === true || value === "true"
     } else if (type === "STRING" && Array.isArray(res)) {
         res = res.toString()
     }
--- a/ui/src/utils/inputs.ts
+++ b/ui/src/utils/inputs.ts
@@ -191,26 +191,3 @@ export function normalize(type: InputType | undefined, value: any) {
     }
     return res
 }
-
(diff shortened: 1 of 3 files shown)


Pradumnasaraf (commented):
Hey @MilosPaunovic, I see your PR is on the OSE, and apps are not available on that! 

So it's not possible to test it.

Piyush-r-bhaskar (approved):
Both Flow and App is good now. 

Image

Image

PR #17942, opened 2026-08-07
feat(dashboards): custom dashboard to enterprise only

Piyush-r-bhaskar:


MilosPaunovic:
Reviewed both this PR and https://github.com/kestra-io/kestra-ee/pull/9775 together. The migration itself is clean — no dangling references to the deleted files in OSS, and the EE copies are faithful moves with only import-path rewrites. A few findings:

**1. Missing generated translations (blocking).** `ui/src/translations/en.json` gains 9 new keys under `demos.dashboards.*`, but the other locales and `fingerprints.json` were not regenerated (checked fr/de/es/ja — all lack `demos.dashboards`). Please run `npm run translations:generate` and verify with `npm run translations:check`.

**2. Merge order — neither PR can merge alone without transiently breaking the EE UI build.** The EE vite config resolves `override/*` only to `ui-ee/src/override` (the OSS fallback exists only in `tsconfig.base.json`, i.e. type-check only):
- *OSS first:* EE develop still imports the deleted `kestra/src/components/dashboard/components/selector/Selector.vue` (flows + namespaces `Actions.vue`) and the deleted editor internals from its override `Edit.vue` → EE build breaks against OSS develop.
- *EE first:* OSS develop's `routes.ts` still contains `() => import("override/components/dashboard/Edit.vue")`, but the EE PR renames `ui-ee/src/override/components/dashboard/Edit.vue` away → unresolvable dynamic import → EE build breaks.

CI is green on both only because the matching branch names got paired. Cheapest fix: keep a one-line re-export stub at `ui-ee/src/override/components/dashboard/Edit.vue` in the EE PR, then merge EE → OSS, and drop the stub in a follow-up. Otherwise the two need to merge back-to-back.

**3. Interim OSS dead-end for copilot dashboard drafts.** Until the backend flips `isCustomDashboardsEnabled` to `false` in OSS, the copilot still offers dashboard drafts with working actions — but "Open in editor" now lands on the Enterprise demo page and silently discards the drafted YAML. The PR description already owns this (edge cases 3/6), but the backend flip really should ship in the same release.

**4. Minor:**
- No test covers the flag-off behavior: `Header.spec.ts` mocks `configs: {}` (enabled), `CopilotArtefactDraft.spec.ts` hardwires `dashboardSupported: ref(true)`, and the new early-return in `Dashboard.vue` is untested. One disabled-state case each would lock in the actual behavior this PR introduces.
- PR description is empty — given the cross-repo merge-order constraint, it should at least link the EE PR and https://github.com/kestra-io/kestra-ee/issues/8372 and state the intended merge order.

CI note: the `trigger-ee` failure here is infra (`io.kestra.gradle.spotless-conventions 1.2.6` not resolvable), unrelated to this PR.

MilosPaunovic (approved):
Re-reviewed after the latest push - remaining open items:

**1. Merge order (must coordinate before merging).** Neither PR can merge alone without transiently breaking the EE UI build, since the EE vite `override` alias has no OSS fallback. Either add a one-line re-export stub at `ui-ee/src/override/components/dashboard/Edit.vue` (safe order then becomes EE first, OSS second, stub removed in a follow-up), or coordinate merging both back-to-back.

**2. Missing fingerprint entries (minor).** The translations tooling landed on develop after this branch's merge base, so the new locale values have no entries in `ui/scripts/translations/fingerprints.json` and the next `translations:generate` run will re-translate and churn all of them. After merging develop into the branch, one `npm run translations:generate` run records the fingerprints and prevents that.

**3. Translation keys going ghost (follow-up, not blocking).** Once this merges, 8 `dashboards.*` keys will be referenced only by the moved EE components: `creation.label`, `creation.confirmation`, `edition.confirmation`, `edition.id readonly`, `deletion.confirmation`, `switch`, `chart_preview`, `labels.singular` - and `dashboards.preview` is dead in both repos already. Nothing breaks (EE merges OSS translations), but per the ghost-translations convention they should move to `ee_translations`. A follow-up issue + PR pair will be opened to sort that after both PRs merge (EE side first, same ordering as the previous key move).

Non-blocking, from the previous review: empty PR description (worth stating the merge order and linking https://github.com/kestra-io/kestra-ee/pull/9775 + https://github.com/kestra-io/kestra-ee/issues/8372), and no test covers the flag-off behavior.

PR #17953, opened 2026-08-07
feat(tests): surface AI-drafted unit tests in the Copilot

ammeek:
Companion PR for https://github.com/kestra-io/kestra-ee/pull/9779  

Closes: https://github.com/kestra-io/kestra-ee/issues/9555


ui/src/override/stores/misc.ts
```diff
@@ -17,20 +17,23 @@ export const useMiscStore = defineStore("misc", () => {
...
-    // points ("Fix with AI", the editor shortcut, …) via `promptCopilot`; consumed and cleared
-    // by CopilotChat. `null` means nothing pending.
-    const copilotPrompt = ref<string | null>(null)
+    // A prompt to hand to the AI Copilot the next time it renders. Set by entry points ("Fix with
+    // AI", "Generate a unit test", the editor shortcut, …) via `promptCopilot`; consumed and cleared
+    // by CopilotChat. `null` means nothing pending. `autoSend` distinguishes the two entry-point
+    // styles: seed the composer for the user to review and send (the default), or send the turn
+    // straight away for a single-purpose action the user already committed to by clicking it.
+    // `newChat` starts the prompt in a fresh conversation instead of continuing the current one.
+    const copilotPrompt = ref<{text: string; autoSend: boolean; newChat: boolean} | null>(null)
```
MilosPaunovic, line 26: This shape change is the cross-repo break described in the review summary.

`override/stores/misc.ts` is the OSS default that EE shadows. EE develop's copy still declares:

```ts
const copilotPrompt = ref<string | null>(null)
const promptCopilot = (prompt: string) => { ... }
```

so once this merges, `seeded.text` / `.autoSend` / `.newChat` in `CopilotChat.vue` no longer type-check in an EE build. The EE-side update is only on https://github.com/kestra-io/kestra-ee/pull/9779 — these need to land back-to-back.

The `promptCopilot` signature itself is fine: `options` is optional, so the three existing OSS callers (`TaskRunActions.vue`, `overview/components/Banner.vue`, `ErrorToastContainer.vue`) keep working unchanged.

ui/src/components/ai/copilot/CopilotChat.vue
```diff
@@ -339,11 +342,27 @@
...
+    // instead of inheriting an unrelated transcript.
     async function consumeSeededPrompt(): Promise<void> {
         const seeded = miscStore.copilotPrompt
         if (!seeded) return
-        composerText.value = seeded
         miscStore.copilotPrompt = null
+
+        // Settle the in-flight restore before acting on either flag: resetting ahead of it would be
+        // undone when it lands, and sending without it would fork a second thread.
+        if (seeded.newChat || seeded.autoSend) await restored
```
MilosPaunovic, line 353: Minor: this awaits `restored`, but not the provider list.

`selectedProvider` is populated by a separate, un-awaited `async onMounted` further up (the `AiApi.providers()` fetch). Those are two independent network calls, so if the provider list resolves *after* `restoreThread()`, the auto-sent turn goes out with `providerId: undefined`.

Impact is small — the server falls back to its own default, and `selectedProvider` is itself seeded from the `isDefault` provider, so it usually resolves to the same one. But the fix is cheap and symmetric with what you already did for `restored`: keep the providers promise too and await it on the `autoSend` path, so the first turn carries the same `providerId` a manually sent one would.

ui/src/components/ai/copilot/CopilotChat.vue
```diff
@@ -238,7 +238,10 @@
     const {thread, messages, status, streaming, error, errorDetail, notice, pendingConfirmation, unavailable, canSend, sendChat, confirm, cancel, reset, retry, retryLastTurn, loadThread, restoreThread, noteContext} = useAiChat()
 
     // Restore the last conversation on open (threads are persisted server-side); harmless no-op if none.
-    onMounted(() => { restoreThread() })
+    // Kept as a promise so an auto-sent prompt waits for it and continues that thread instead of
+    // racing the restore into a second one.
+    let restored: Promise<void> = Promise.resolve()
```
MilosPaunovic: Nit: `restored` starts as an already-resolved promise and only gets the real one inside `onMounted`.

The mount path is safe (this hook is registered before the `onMounted(consumeSeededPrompt)` below, so the assignment happens first). The only window is the watcher firing between setup and mount — there `await restored` is a no-op and an `autoSend` prompt could fork a second thread.

Practically unreachable, since `promptCopilot` opens the tab and the mount hooks run before any store mutation could land. Calling `restoreThread()` during setup rather than in `onMounted` would remove the window entirely and drop the `Promise.resolve()` placeholder.

MilosPaunovic (commented):
Reviewed the whole diff. One finding is merge-blocking, the rest are minor. The `draftApply.ts` extraction and the test coverage both look good — the specs pin the `reset`-before-`sendChat` ordering and the `canSend === false` fallback, which is exactly the regression this design could produce.

### This cannot merge on its own — it breaks EE develop

`copilotPrompt` changes shape here, and it lives under `override/`, so EE shadows it with `ui-ee/src/override/stores/misc.ts`. **EE develop still has `ref<string | null>` and `promptCopilot(prompt: string)`**, and EE compiles against OSS source — so the moment this lands on develop, EE fails to type-check with `Property 'text' does not exist on type 'string'`.

The matching EE change exists only on https://github.com/kestra-io/kestra-ee/pull/9779, which is itself blocked: the EE SDK cannot be regenerated because `:webserver-ee:compileJava` fails against OSS develop (the webhook signature drift fixed by https://github.com/kestra-io/kestra-ee/pull/9819).

There is no safe single order — if https://github.com/kestra-io/kestra-ee/pull/9779 landed first, EE's store would hand an object to OSS develop's `composerText.value = seeded` string assignment and type-error the other way. So the sequence has to be https://github.com/kestra-io/kestra-ee/pull/9819 → https://github.com/kestra-io/kestra-ee/pull/9779 → this PR, landed back-to-back, with EE red in whatever gap is left.

Worth noting in both PR descriptions, since this is the second cross-repo break of this exact shape on develop today.

### Related: no user-visible behaviour ships here alone

`ui-ee/src/override/components/ai/copilot/testSuiteDraftActions.ts` also exists only on the EE branch, not on EE develop. Until https://github.com/kestra-io/kestra-ee/pull/9779 lands, `testSuiteSupported` is false everywhere and the `TEST_SUITE` path is inert. Harmless, just worth knowing when deciding what to verify after merge.

MilosPaunovic (approved):
Re-reviewed the FE after the develop merge - both earlier CopilotChat comments are addressed (provider list is awaited before auto-send, with a unit test; `restoreThread` now starts at setup). Resolving those two threads. The `copilotPrompt` shape change is covered by the EE companion https://github.com/kestra-io/kestra-ee/pull/9779 - these must merge back-to-back, OSS first.

MilosPaunovic:
Hey @ammeek, maybe you can get your hands on this & it's EE counterpart https://github.com/kestra-io/kestra-ee/pull/9779 once you're done with all the higher priority items before the release, as if we keep it opened for long, it will only have more and more conflicts and be harder to untangle in the end.

ui/src/components/ai/copilot/CopilotChat.vue
```diff
@@ -395,20 +394,45 @@
...
+        if (newThread) {
             if (thread.value || messages.value.length > 0) reset()
-            nextThreadTitle.value = miscStore.copilotThreadTitle
+            nextThreadTitle.value = threadTitle
         }
+
+        if (autoSend) {
+            // The provider list has to be in hand before the turn goes out, or it is sent against
+            // the default provider rather than the one the picker restored.
+            await providersLoaded
```
MilosPaunovic, line 428: This `await` sits before the `canSend` check, so it also gates the composer-seeding fallback below: if `/ai/providers` is slow or hangs (no client timeout on that call), the dock opens and the composer stays empty for as long as the request takes - the prompt is neither sent nor visible. Before this fix the prompt was at least handed over immediately. Cheap fix: seed the composer first and clear it on the send path, or bound the wait with a `Promise.race` against a short timer so the prompt is never invisible.

ui/src/components/ai/copilot/CopilotChat.vue
```diff
@@ -395,20 +394,45 @@
...
+            if (canSend.value) {
+                onSubmit(seeded)
```
MilosPaunovic, line 430: The awaits above widened the window where `consumeSeededPrompt` resumes after the component unmounted: `onBeforeUnmount(cancel)` has already run, so this `onSubmit` starts an `SSE` turn nothing can cancel, streaming into refs nobody renders. A `let disposed = false` flag set in `onBeforeUnmount` and checked before `onSubmit` closes it.

ui/src/components/ai/copilot/CopilotChat.vue
```diff
@@ -395,20 +394,45 @@
...
-        if (miscStore.copilotNewThread) {
+        if (newThread) {
             if (thread.value || messages.value.length > 0) reset()
-            nextThreadTitle.value = miscStore.copilotThreadTitle
+            nextThreadTitle.value = threadTitle
         }
+
+        if (autoSend) {
+            // The provider list has to be in hand before the turn goes out, or it is sent against
+            // the default provider rather than the one the picker restored.
```
MilosPaunovic, line 427: Nit: nothing restores a picker selection - `selectedProvider` is initialised to the `isDefault` provider, falling back to the first in the list. The real effect of the wait is that `providerId` is sent explicitly instead of left `undefined` for the backend to default; worth rewording so the next reader does not go looking for persistence that is not there.

MilosPaunovic (commented):
FE re-review of the rebased branch - the provider-carry fix and the reset-before-send ordering check out, and the new specs genuinely pin the behavior (the provider-wait test holding the promise open is nice). Three smaller points inline on `CopilotChat.vue`; the companion `EE` review at https://github.com/kestra-io/kestra-ee/pull/9779 has the blocking findings, so gating this one on those.

ui/src/components/ai/copilot/CopilotChat.vue
```diff
@@ -392,23 +391,63 @@
...
+        if (sendInitialMessage) {
+            // Shown before the wait below, so a slow `/ai/providers` can't leave the prompt
+            // neither sent nor visible; the send path clears it again.
+            composerText.value = seeded
+            // The turn carries `providerId` only once the list is in hand, otherwise the backend
+            // picks its own default — bounded, so a hanging fetch degrades to that rather than
+            // never sending.
+            await Promise.race([providersLoaded, new Promise((resolve) => setTimeout(resolve, PROVIDERS_WAIT_MS))])
+            if (disposed) return
+            if (canSend.value) {
```
MilosPaunovic, line 443: The seeded text sits in a fully interactive composer for up to `PROVIDERS_WAIT_MS`, and a user acting in that window gets clobbered. If they press Enter during the wait, their turn goes out and `canSend` flips false; the race then resolves, this branch is skipped, and the fallback below re-seeds the already-sent prompt into the composer - a second Enter once the turn finishes duplicates it. If they *edit* the seeded text instead, this branch fires `onSubmit(seeded)` with the stale captured string and the clear on the line above discards their edit. A `composerText.value === seeded` check guarding both the clear/send here and the re-seed fallback below closes both paths.

ui/src/components/ai/copilot/CopilotChat.vue
```diff
@@ -418,7 +457,10 @@
         if (value) consumeSeededPrompt()
```
MilosPaunovic, line 457: `consumeSeededPrompt` is re-entrant, and the bounded wait widens the overlap window from a microtask to something reachable by hand - a double-click on "Generate a unit test", or a second entry point firing while the first run is parked on the race. Both runs hold a live `seeded` (the store is cleared synchronously), so run B sends and run A then re-seeds its own prompt into the composer. A module-scope `consuming` flag around the body would do.

ui/src/components/ai/copilot/CopilotChat.vue
```diff
@@ -392,23 +391,63 @@
...
         }
+
+        if (sendInitialMessage) {
+            // Shown before the wait below, so a slow `/ai/providers` can't leave the prompt
+            // neither sent nor visible; the send path clears it again.
+            composerText.value = seeded
+            // The turn carries `providerId` only once the list is in hand, otherwise the backend
+            // picks its own default — bounded, so a hanging fetch degrades to that rather than
+            // never sending.
+            await Promise.race([providersLoaded, new Promise((resolve) => setTimeout(resolve, PROVIDERS_WAIT_MS))])
```
MilosPaunovic, line 441: Two nits on this line and the guard below: the losing `setTimeout` is never cleared, so a timer stays armed per seeded send (and per spec run) - a `clearTimeout` in a `finally` tidies it; and `disposed` covers unmount but not `KeepAlive` deactivation (switching dock tabs deactivates rather than unmounts, so a parked send still fires on a hidden component). The latter is arguably the desired behavior - if so, a one-line comment saying deactivation is deliberately not treated as disposal would stop the next reader from "fixing" it.

MilosPaunovic (changes requested):
Round 4, on the seeded-prompt fix commit. All three earlier `CopilotChat.vue` points are addressed - the seed-first + bounded provider wait, the `disposed` guard, and the comment reword all check out, and the two new specs pin the wait behavior properly. One real regression introduced by the new commit and one smaller race inline; two nits folded in. The `EE` companion is down to minor items, so this is the last blocking ask from my side.

PR #17511, opened 2026-07-20
feat(ui): Implement search box filter in Outputs tab

ILOVEZERI333:
---

### ✨ Description

What does this PR change?  
Implements the search box in the "Inputs/Outputs" section when viewing executions to filter task runs. 

https://github.com/user-attachments/assets/bbdb8614-68fd-45e3-a578-c04336dbfbde

### 🔗 Related Issue

Which issue does this PR resolve? Closes [#5448](https://github.com/kestra-io/kestra/issues/5448)

### 🎨 Frontend Checklist

- [x] Code builds without errors (`npm run build`)
- [ ] All existing E2E tests pass (`npm run test:e2e`)
- [x] Screenshots or video recordings attached showing the `UI` changes

PR #17959, opened 2026-08-07
fix(executions): render expression prefix and debugger updates for flow outputs (#15631)

### LINKED ISSUE #15631: Render Expression on Execution Overview Bugs
### Describe the issue

Related: https://github.com/kestra-io/kestra-ee/issues/7483

- [ ] Flow Level Outputs display invalid expression, despite it not being selected. Doesn't do this for Variables or Inputs.
  - It displays `{{ outputs.myname }}` but should be `{{ execution.outputs.myname }}` where it's flow level
- [ ] UI freezes when you select one of the values (Variables, Inputs, Flow Outputs, Trigger). UI box is highlighted but never stops being highlighted even when other things selected.
- [ ] UI doesn't put that expression in the render box when selected.

https://github.com/user-attachments/assets/e3f26292-76b1-4535-bc8e-0b3d9b614626

Reproducer flow:
```yaml
id: input_check
namespace: company.team

inputs:
  - id: name
    type: STRING
    defaults: Will

variables:
  myname: Will

tasks:
  - id: hello
    message: "{{ inputs.name }}"

outputs:
  - id: myname
    type: STRING
    value: Hello
```

### Environment

- Kestra Version: 1.3.11 OSS (but EE returns 401 - see related issue above.)

mohitjeswani01:
### ✨ Description
This PR fixes the Flow Outputs expression prefix bug in the Execution Variable Explorer UI, reported in #15631.

1. **Correct Flow Outputs Expression Prefix**: Updated the section prefix for `flowOutputs` from `outputs` to `execution.outputs` in `ExecutionVariableExplorer.vue`. Flow-level outputs require the `execution.` prefix in Pebble template evaluation (`{{ execution.outputs.myname }}` instead of `{{ outputs.myname }}`), since the bare `outputs.<name>` path is already used by task outputs and never resolves a flow-level output.

2. **Sticky highlight (Bug 2) — re-verified, not touched**: The original issue reported that clicking a row (Variables/Inputs/Flow Outputs/Trigger) left a stale highlight active. Re-tested this against current `develop` and could not reproduce it — clicking through all sections correctly moves the highlight with no stale/duplicate active items. `SidebarList.vue` was not modified in this PR. This appears to have been fixed independently by earlier refactors (#17242, #17511, #17512).

3. **Removed unnecessary `flush: "sync"`**: An earlier revision of this PR added `{ immediate: true, flush: "sync" }` to the `props.expression` watcher in `ExpressionDebugger.vue`. Per @flcarre's review, this was unnecessary — `useKsEditor.ts` already has a default-flush watcher that updates Monaco, and `ExpressionDebugger` is always mounted before a row can be clicked. Removed both options and reverted the `result`/`resultLang`/`error`/`clear()` hoist that was only needed to avoid a TDZ error caused by `immediate: true`. Manually verified on `localhost:5173` that the expression box still populates immediately on row click without it.

### 🔗 Related Issue
Closes https://github.com/kestra-io/kestra/issues/15631

### 🎨 Frontend Checklist

### 📝 Additional Notes
- All changes are limited to `ExecutionVariableExplorer.vue` and `ExpressionDebugger.vue`, plus regression tests.
- Added a Storybook play-function story (`FlowOutputsDebugger`) in `ExecutionVariableExplorer.stories.tsx` that clicks the Flow Outputs `myname` item and asserts the real (unstubbed) editor shows `{{ execution.outputs.myname }}`. Also added `outputs: {myname: "Hello"}` to `FAKE_EXECUTION` so the Flow Outputs section renders in stories at all.
- Added a unit test in `ExecutionVariableExplorer.spec.ts` asserting the `sections` computed yields `expression: "execution.outputs.myname"` for flow output items — a direct regression guard for the core fix.


ui/src/components/executions/outputs/ExpressionDebugger.vue
```diff
@@ -81,19 +90,11 @@
             editorValue.value = value ?? ""
             clear()
         },
+        {immediate: true, flush: "sync"},
```
flcarre: I traced the update chain and couldn't find a case where the default (`pre`) flush fails to reach the editor:

- `useKsEditor.ts` already has `watch(() => props.modelValue, ...)` with default flush that calls `setValue()` on the Monaco instance whenever the prop changes.
- `<ExpressionDebugger>` is rendered unconditionally by `ExecutionVariableExplorer.vue`, so the editor is always mounted before a row can be clicked (and `initMonaco()` reads `props.modelValue` at creation time anyway, so an early change can't be lost).
- This is the only `flush: "sync"` in the whole `ui/` tree.

`immediate: true` also looks like a no-op: `editorValue` is already initialised to `props.expression ?? ""` on line 74, and `clear()` has nothing to clear at setup time. It's what forces `result`/`resultLang`/`error`/`clear()` to be hoisted above the watcher, which is why this diff has that reordering churn.

If you have a concrete repro where the editor stays empty without it, let's keep it and add a one-line comment explaining why. Otherwise:

```suggestion
        {immediate: true},
```

(or drop the options object entirely if `immediate` isn't needed either — that would also let the `result`/`clear()` hoist be reverted, shrinking the diff).

flcarre (commented):
Thanks for tackling this one, @mohitjeswani01 — and for picking a bug that needed someone to actually understand the Pebble context model rather than just move a string around. 🙌

I verified the core fix against the backend and it's **correct**:

- `RunVariables` registers `execution.outputs` as a real context path ([`RunVariables.java#L62`](https://github.com/kestra-io/kestra/blob/develop/core/src/main/java/io/kestra/core/runners/RunVariables.java#L62)) and populates it from `execution.getOutputs()` (~L260).
- The top-level `outputs` map is built by `TaskOutputService.computeOutputs()` and is keyed by **taskId** — so `{{ outputs.myname }}` never resolved a flow-level output at all. `execution.outputs.myname` is the only correct path.

I also checked the blast radius of the prefix rename and it's clean: the only producer of the `?select=` deep link is the onboarding tour (`tourScenes.ts`, targeting `trigger.variables` — unaffected), `matchesExplorerItem` matches on label/searchText/value and never on `expression` so search behaviour is unchanged, and the `:key` in `SidebarList` is scoped per section so there's no key collision either way.

**One nice thing your PR fixes that you didn't claim:** by deleting `baseExpressionPath` and making the tree path and the debug path the same string, drilling into a *nested* flow output now emits `{{ execution.outputs.myname.sub }}` instead of the previously un-debuggable `{{ outputs.myname.sub }}`. That's a real extra bug gone, and the simplification is genuinely nicer than the special-case it replaces.

Three things I'd like to sort out before this lands 👇

---

### 1. The PR description claims tests that aren't in the diff

> Added comprehensive unit tests covering all 3 reported bugs.

The diff is exactly the two `.vue` files, no test file. No worries — easy to fix, but our contributing guidelines do ask for a test with every change, and right now nothing in the suite would catch a regression on this:

- `ui/tests/storybook/components/executions/ExecutionVariableExplorer.stories.tsx` — the `FAKE_EXECUTION` fixture has **no `outputs` field**, so the Flow Outputs section is never even rendered in any story.
- `ui/tests/unit/components/executions/ExecutionVariableExplorer.spec.ts` — stubs both `KsEditor: true` and `ExpressionDebugger: true`, so the debugger-seeding path is stubbed away entirely.

Concretely, the smallest useful additions would be:

1. Add `outputs: {myname: "Hello"}` to `FAKE_EXECUTION` in the story, plus a play-function story that clicks the Flow Outputs item and asserts the (unstubbed) editor shows `{{ execution.outputs.myname }}`. We prefer Storybook component tests over Vitest unit tests where a story already exists, and this is the only place that exercises the real editor chain.
2. A plain assertion in `ExecutionVariableExplorer.spec.ts` that the `sections` computed yields `expression: "execution.outputs.myname"` for the `flowOutputs` section — a cheap, direct regression guard for the actual fix.

Could you also update the description so it matches the diff? 🙂

---

### 2. `flush: "sync"` — could you share the repro that needs it?

I dug into this one and couldn't find the defect it's compensating for (inline suggestion below). Summary:

- It's the **only** `flush: "sync"` in the entire `ui/` tree, `src` and design-system combined.
- `useKsEditor.ts` already has a `watch(() => props.modelValue, ...)` with default flush that pushes new content into Monaco, and `<ExpressionDebugger>` is mounted unconditionally, so the editor always exists by the time a row is clicked.
- `immediate: true` looks redundant too: `const editorValue = ref(props.expression ?? "")` already seeds it with the same value, and it's what forced hoisting `result`/`resultLang`/`error`/`clear()` above the watcher (TDZ) — so that reordering is churn rather than a real requirement.

Entirely possible you hit something real that I can't see from the code — if so, a quick repro would be great and we'll keep it (ideally with a comment explaining why, since it's the only one in the codebase). Otherwise I'd rather drop it, since sync watchers bypass Vue's normal batching and tend to be load-bearing in ways nobody remembers later.

---

### 3. Bug #2 (sticky highlight) — worth re-checking on current `develop`

The description says the prefix change eliminates highlight collisions "between task outputs and flow outputs sharing duplicate variable names." That's true, but it only applies when a **flow output id equals a task id** — and the reproducer flow in #15631 has task `hello` and output `myname`, so it never hits that case.

`SidebarList.vue` isn't touched here, and its `:class="{'active': item.expression === selectedExpression}"` binding is plainly reactive, so I don't see a mechanism for a stale highlight in the current component. The explorer has been substantially rewritten since the issue was filed against 1.3.11 (#17242, #17511, #17512), so my guess is that one is already gone independently.

Could you re-test that specific symptom against current `develop` and tell us what you see? If it's already fixed, just narrow the claim in the description; if it still reproduces, that's a separate root cause worth chasing.

---

Bug #1 is solidly fixed and I'd like to get it merged — it's really just the test and the `flush: "sync"` question standing between this and a ship. Thanks again for the contribution! 🚀

mohitjeswani01:
@flcarre thanks for your review on this i will update the pr as instructed !👍

mohitjeswani01:
Hey @flcarre, thanks for the thorough review! Here's what I've done to address all three points:

1. flush: "sync" removed Dropped {immediate: true, flush: "sync"} from the watcher options entirely and reverted the result/resultLang/error/clear() hoist back to their original position (after the watcher), since the TDZ concern no longer applies without immediate: true. Tested manually on localhost:5173 against the develop backend — clicking a Flow Outputs item correctly populates the expression debugger with {{ execution.outputs.myname }} (screenshot attached). No regression found.

2. Tests added

Storybook: Added outputs: {myname: "Hello"} to FAKE_EXECUTION so the Flow Outputs section now renders in all stories, plus a new FlowOutputsDebugger play-function story that clicks the myname item and asserts the real (unstubbed) editor shows {{ execution.outputs.myname }}.
Unit test: Added a plain assertion in ExecutionVariableExplorer.spec.ts that the sections computed property yields expression: "execution.outputs.myname" for the flowOutputs section — a direct, cheap regression guard for the core fix.

3. Bug 2 (sticky highlight) — already gone Re-tested on current develop: clicking myname in Flow Outputs then clicking name in Flow Inputs correctly deactivates myname. Only one item is highlighted at a time. The bug no longer reproduces — must have been fixed independently (screenshot attached). Will narrow the PR description claim accordingly.

image

also on localhost 5173:
image

Bug 2 (sticky highlight)
image

mohitjeswani01:
Hey @flcarre sir happy to hear you feedback on the current changes!

MilosPaunovic:
Hi @mohitjeswani01,

mohitjeswani01:
fixed the conflict in  `ExecutionVariableExplorer.vue`

flcarre:
Thanks for this, the core fix is spot on. Making `flowOutputs` items resolve to `execution.outputs.<name>` instead of the bare `outputs.<name>` matches how Pebble actually resolves flow-level outputs on the backend (`RunVariables.java`), and you also cleaned up a related inconsistency where `expressionPath`/tree selection was using the un-prefixed path while only the debugger box got the `execution.` prefix. The `ExpressionDebugger.vue` watcher revert is a clean no-op vs `develop`, no TDZ risk there. The new Vitest test in `ExecutionVariableExplorer.spec.ts` is a solid, non-tautological regression guard too.

One thing to fix before this is mergeable: the new `FlowOutputsDebugger` Storybook story is currently failing in CI (`Frontend - Tests`, the play function times out on `waitFor(() => canvas.getByText(/flow outputs/i))`).

Root cause: `flowOutputs` is populated via `GET /outputs/executions/{executionId}` (see `loadExecutionOutputs` calling `OutputsAPI.executionOutputs()`), but `meta.beforeEach` in this story file only mocks `/outputs/tasks/*`. The `/outputs/executions/:executionId` route falls back to the global default mock (`{}`) from `.storybook/apiMock.js`, so the Flow Outputs section never has any items, and since `SidebarList.vue` filters out empty sections, the header itself never renders. That's why the very first `waitFor` in the play function can't find it.

The `outputs: {myname: "Hello"}` field you added to `FAKE_EXECUTION` (line 57) isn't actually consumed anywhere for this reason, it's dead data as written.

Suggested fix, add the missing route mock in `meta.beforeEach`:

```tsx
beforeEach() {
    mockStoryApiRoutes({
        [`GET /outputs/tasks/${FAKE_EXECUTION.id}`]: OUTPUTS_INFORMATION,
        [`GET /outputs/tasks/${FAKE_EXECUTION.id}/run-http`]: OUTPUTS_BY_TASK_RUN_ID["run-http"],
        [`GET /outputs/tasks/${FAKE_EXECUTION.id}/run-check`]: OUTPUTS_BY_TASK_RUN_ID["run-check"],
        [`GET /outputs/executions/${FAKE_EXECUTION.id}`]: {myname: "Hello"},
    });
},
```

With that in place, `FAKE_EXECUTION.outputs` can either be dropped or kept with a comment noting it's unrelated to what actually feeds the section (the mock above is what does).

Nothing else stood out: the sticky-highlight investigation (bug #2) checks out against the current `SidebarList.vue`, and no design-system/i18n/security/performance concerns in this diff. Once the Storybook story is green this looks good to merge from my side.

mohitjeswani01:
Thanks for the review @flcarre! Good catch on the missing route mock.

Fixed in the latest push — added the GET /outputs/executions/${FAKE_EXECUTION.id} mock to meta.beforeEach so flowOutputs.value gets populated via the API layer (matching how the component actually fetches it via loadExecutionOutputs). Also removed the dead outputs field from FAKE_EXECUTION since it's no longer consumed by anything.

All 5 stories now pass locally, including FlowOutputsDebugger. CI should be green on this run.
image

MilosPaunovic:
Closing, as the issue itself got closed.

Thanks for all the efforts on it @mohitjeswani01, it's much appreciated! 🍀

PR #17974, opened 2026-08-07
fix(ui): preserve basic auth across tabs

### LINKED ISSUE #17950: Opening a new tab on logged instance with asked me to login
### Describe the issue

- Login on the instance 
- Open any page on a new tab 
- You will be redirected to login page 
- same when opening http://localhost:8080/ui/main directly

### Environment

- Kestra Version: develop

n200534:
## Summary
- Persist the credential-free basic-auth marker in local storage.
- Clear the marker on logout.
- Add regression coverage for authentication restored in a new tab.

## Root cause
The session cookie is shared across tabs, but the UI marker was kept in tab-scoped session storage. A new tab therefore treated an authenticated user as logged out.

Fixes #17950

## Validation
- npm run test:unit -- --run src/utils/basicAuth.spec.ts
- npm run check:types


n200534:
i have opened the PR ,Can  You Please review it and merge. Thanks😊

MilosPaunovic:
Hi @n200534,

ui/src/utils/basicAuth.ts
```diff
@@ -7,7 +7,7 @@ import {useClient} from "@kestra-io/kestra-sdk"
 const AUTH_FLAG_KEY = "kestraBasicAuthenticated"
 
 export async function logout() {
-    sessionStorage.removeItem(AUTH_FLAG_KEY)
+    localStorage.removeItem(AUTH_FLAG_KEY)
```
elevatebart, line 10: A problem exists with using `localStorage` instead of `sessionStorage`. If one quits chrome entirely, the cookies are cleaned but the localStorage is not. This means we cannot trust localStorage to convey login information. 

When I tested your branch I followed these steps:
- logged in ✅
- navigated ✅
- Quit Chrome ✅
- opened kestra again 🔄
- loaded forever and failed 💥

elevatebart (changes requested):
Great momentum forward. 
Thank you for taking this one on. ❤️

I have a problem with using localStorage to store the login status as when you quit chrome, the values are kept. Instead, if there is no value, if we do not know wether user is connected or not, could we instead try to ping the server to confirm?

elevatebart:
There is another more efficient solution. 
Problem is, it involves backend changes.
The Basic AUTH cookie is not accesible by JS, but we could set another cookie up at the same time.
This is done in `MiscController.java`.

elevatebart:
After investigation, if I access anything during `isLoggedIn()` I make the function async, which means I have to update a lot of code around for a solution that is partially wrong.
Furthermore, as soon as I ping an unauthenticated route, I get the annoying native login dialog triggered by the BASIC_AUTH challenge. 
I believe we are not on the right path here.
I will draft a PR for the alternative version of the auth I thought of.

[EDIT] https://github.com/kestra-io/kestra/pull/18024

PR #17992, opened 2026-08-09
feat(ai): support custom HTTP headers on every copilot model provider

ammeek:
### ✨ Description

Custom HTTP headers now work on **every** AI Copilot provider. Five already supported them (OpenAI, Azure OpenAI, DeepSeek, Ollama, OpenRouter); the blocker for the rest was langchain4j, which has exposed `customHeaders` on all the remaining chat model builders since 1.15.0. We are on 1.18.1, so this unblocks Gemini here, and Anthropic, Mistral, Bedrock and Vertex AI in the companion EE PR. `customHeaders()` moves onto the `AiConfiguration` interface so every provider shares one contract instead of a per-provider convention.

```yaml
kestra:
  ai:
    providers:
      - id: gemini
        type: gemini
        configuration:
          model-name: gemini-2.5-flash
          api-key: "{{ secret('GEMINI_API_KEY') }}"
          custom-headers:
            X-Api-Key: secret
            X-Gateway-Route: internal
```

**This also fixes a bug that made the existing support unusable.** Header names configured under a `providers` entry never reached the model API intact: Micronaut applies its camel-case key convention at *every* nesting level of the untyped `configuration` map, so `X-Api-Key` was sent as `xApiKey` — the hyphens were eaten. Every hyphenated header name was affected, which is most real ones (`Authorization` survived only by having no hyphen; header names are case-insensitive per RFC 7230). Headers are now re-read from the raw property source with `StringConvention.RAW`, which returns them exactly as written.

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/services/ai/AiConfiguration.java
+++ b/webserver/src/main/java/io/kestra/webserver/services/ai/AiConfiguration.java
@@ -1,5 +1,9 @@
 package io.kestra.webserver.services.ai;
 
+import java.util.Map;
+
+import io.micronaut.core.annotation.Nullable;
+
 public interface AiConfiguration {
     String type();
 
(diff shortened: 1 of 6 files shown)


fhussonnois (approved):
LGTM!

PR #17995, opened 2026-08-10
feat(core): ResumableTaskInterface for reattach after worker restart

Malaydewangan09:
Adds an opt-in `ResumableTaskInterface` and a `ResumableTaskService` helper so a task that triggers a long-running external job can reattach to it after a worker restart instead of triggering a duplicate, with the reattach decision left in each task's `run()`. Shared plumbing only: remember/recall/forget an opaque handle in the flow namespace KV store, keyed by the taskrun id (stable across a restart, unique per execution).
 
- issue kestra-io/plugin-dbt#307) for reuse across plugins.


Malaydewangan09:
Closing as it will be done on plugin level

Issue #17863, opened 2026-08-05
potential flow execution bug when running Kestra in distributed mode


hpgrahsl:
### Describe the issue

TL;DR

* in distributed mode any flow executions via API/UI seem broken as the REST endpoint immediately 404s -> the 
* backend however successfully executes the flow just fine
this doesn't affect standalone mode at all it seems

Whenever I execute flows in the UI right after starting the execution the UI comes up with a NOT found page. Clicking my way back to the executions I see that it actually ran and worked. So initially I thought a glitch in the UI but now I'm seeing strange API endpoint behaviour as well giving me 404 status codes back when triggering executions via API. Also in this case the execution actually ran fine on the backend.

I've attached 2 videos showing that:

* in standalone it's running just fine

https://github.com/user-attachments/assets/9705dfe8-9cae-488b-8151-bc8cf71f787b

* but in distributed mode it's clearly broken

https://github.com/user-attachments/assets/f22e298d-7ef4-47e9-b00f-abe2208caefc

No suspicious logs observed in the kestra components while running the "broken" case. API endpoint response is this:

```
{
    "message": "Not Found: Execution disappeared after create: 3hBunbLkipbbnNDe78KgYG",
    "logref": null,
    "path": null,
    "_links": {
        "self": {
            "href": "/api/v1/default/executions/company.team/flow_1177752?labels=system.from%3Aui&scheduleDate=2026-08-05T12%3A13%3A13.057%2B02%3A00&kind=NORMAL&breakpoints=&revision=1",
            "templated": false,
            "profile": null,
            "deprecation": null,
            "title": null,
            "hreflang": null,
            "type": null,
            "name": null
        }
    },
    "_embedded": {}
}
```

### Environment

- Kestra Version: v2.0.0-rc6
- Deployment: distributed postgres / mysql
- Edition: tested on EE but most likely affects OSS so both

Pradumnasaraf:
Working. QA done ✅

PR #17998, opened 2026-08-10
fix(executions): wait for the execution when a follow subscriber registers

loicmathieu:
`ExecutionStreamingService.registerSubscriber` ends with a guard for the case where a terminal `FollowExecutionEvent` was consumed — or lost during the queue resume — before the subscriber made it into the map. It reads the execution and, if it is already terminal, completes the sink directly. Nothing else will: the event that would have completed it is gone.

The guard used `ifPresent`, so it did nothing at all when the read came back empty.

That is not a rare state. A subscriber can register before the execution has been persisted, because the executor is a separate process in a distributed installation and may not have applied the create command yet. This has been reported in distributed deployments, where the window between emitting the create command and the execution becoming readable is widest.

## Why it matters

`ExecutionController`'s create-with-`wait=true` endpoint registers a subscriber immediately after emitting the Create command, then blocks on `Flux.last()`. A terminal event missed during registration, combined with a guard that silently does nothing, hangs the request until its timeout — precisely the failure the guard was written to prevent, as its own comment describes.

The `/follow` endpoint is unaffected: it already waits for the execution to exist before registering, and re-reads afterwards.

## Change

Wait for the execution to become readable rather than giving up on the first miss, using the same budget `/follow` already applies (10s, polling every 500ms), and log rather than fail if it never appears. Both callers are `@ExecuteOn(TaskExecutors.IO)`, so the bounded wait is consistent with what the controller already does on that path.

Putting it in `registerSubscriber` rather than at the call site means the unprotected caller and any future one are covered. For `/follow` it costs nothing — the execution is already there, so the first poll returns.

## Not a staleness issue

Worth stating explicitly, since it looks adjacent: this is about the execution not existing yet, not about reading a stale copy. The lookup is `findByIdWithoutAcl`, a realtime get, so no index refresh interval is involved.

## Testing

Adds a test where the execution only becomes readable *after* registration and **no follow event is ever delivered**, so the stream can only complete via the catch-up. It fails against the previous `ifPresent` version. It sits alongside the two existing regression tests for kestra-io/kestra-ee#8824.

Compiles clean. The suite has not been run.

## Related, but not fixed here

kestra-io/kestra#17863 is the same family of bug — a read-after-write across processes on the create path — but a different code path, and this PR does not fix it. That 404 is raised by `ExecutionController.awaitBlockingAction`, before any subscriber is registered, so it never reaches the code changed here.

Its cause also looks different. In `ExecutionCommandMessageHandler.handleCreate` the execution *is* persisted before the `finally` emits the processed event, so the ordering is right. The likely culprit is that on `queue-jdbc` the create joins the dispatch-queue poll transaction — `JooqDSLContextWrapper` documents that a nested `transaction(...)` "silently joins a caller-owned transaction … deferring their writes until that transaction commits" — so the write is not committed when the processed event fires, and the webserver reads nothing. That would explain why it is specific to distributed Postgres/MySQL. `requireNewTransaction` is the documented escape hatch.

### DIFF
--- a/core/src/main/java/io/kestra/core/services/ExecutionStreamingService.java
+++ b/core/src/main/java/io/kestra/core/services/ExecutionStreamingService.java
@@ -1,10 +1,13 @@
 package io.kestra.core.services;
 
+import java.time.Duration;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.commons.lang3.tuple.Pair;
+import org.awaitility.core.ConditionTimeoutException;
 
 import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.flows.Flow;
@@ -12,6 +15,7 @@ import io.kestra.core.queues.BroadcastQueueInterface;
 import io.kestra.core.queues.QueueSubscriber;
 import io.kestra.core.repositories.ExecutionRepositoryInterface;
 import io.kestra.core.runners.FollowExecutionEvent;
+import io.kestra.core.utils.Await;
 import io.kestra.core.utils.ListUtils;
 import io.kestra.core.utils.MapUtils;
 
@@ -34,6 +38,9 @@ import reactor.core.publisher.FluxSink;
 @Slf4j
 @Singleton
 public class ExecutionStreamingService {
+    private static final Duration EXECUTION_LOOKUP_TIMEOUT = Duration.ofSeconds(10);
+    private static final Duration EXECUTION_LOOKUP_POLL_INTERVAL = Duration.ofMillis(500);
+
     private final Map<String, Map<String, Pair<FluxSink<Event<Execution>>, Flow>>> subscribers = new ConcurrentHashMap<>();
     private final Object subscriberLock = new Object();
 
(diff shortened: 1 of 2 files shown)


loicmathieu:
Closing this — the premise is wrong and the change breaks an existing test.

**It regresses `ExecutionStreamingServiceTest`.** `shouldCompleteFluxWhenTerminatedEventArrivesAndExecutionIsPersistedInRepository` deliberately registers the subscriber *before* saving the execution. The blocking wait added here stalls inside `registerSubscriber` for its full budget waiting on an execution the caller can only save after `subscribe()` returns, so the Flux's 5s timeout fires first. Verified: the suite is green on `develop` (2 tests) and fails with this commit applied. Putting the wait inside `registerSubscriber` inverts the caller's control flow, which is legitimate to rely on.

**And the state it guards against does not occur.** Both production callers have already read the execution before registering:

- create with `wait=true` — `awaitBlockingAction(...).flatMap(res -> ... Flux.create(emitter -> registerSubscriber(...)))`, where `res.body()` is an execution `awaitBlockingAction` read successfully;
- `/follow` — waits for the execution via `Await.await()` before registering, and re-reads afterwards.

So `registerSubscriber`'s catch-up cannot legitimately read empty, and the `ifPresent` this set out to harden is unreachable in practice. A reduced version that only logged the empty case would log a branch that never fires.

The distributed 404 this was chasing is a different code path and is addressed by #17999: `awaitBlockingAction` read the execution back through the ACL-enforcing `findById` on the event-queue polling thread, which has no authenticated principal, so it failed closed. kestra-io/kestra-ee#9789 adds the regression test pinning that `findByIdWithoutAcl` bypasses the namespace ACL, across H2, Postgres, MySQL and Elasticsearch.

PR #18009, opened 2026-08-10
fix(core): keep per-emit asset lineage bundles instead of flattening

Malaydewangan09:
Fixes a near-cartesian dbt asset lineage graph: the worker flat-mapped every emitted lineage bundle into one `AssetsInOut`, pairing all inputs with all outputs.

- `TaskRun` now carries `List<AssetsInOut> assetEmits` as the single source of truth for a taskRun's lineage bundles (one per emitted pair, plus the manual `assets:` declaration).
- The old single `assets` object is kept as a deprecated backward-compat bridge: `@JsonSetter("assets")` folds pre-existing executions into `assetEmits` on deserialization, and a deprecated `getAssets()` keeps `ApiTaskRun` working. Historical executions still load, no migration, no API break.
- `WorkerTaskProcessor` stops flat-mapping and collects the declaration plus each emit as separate bundles.
- `ExecutorService` triggers asset processing when any bundle has content.

Verified: legacy `{"assets":{...}}` executions fold into `assetEmits` (TaskRunTest), plus WorkerTaskProcessorTest, ExecutorServiceProcessTaskRunAssetsTest, H2RunnerTest.

- part-of: kestra-io/plugin-dbt#306
- companion PR: kestra-io/kestra-ee#9792
- companion PR: kestra-io/plugin-dbt#311

### DIFF
--- a/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
+++ b/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
@@ -298,7 +298,7 @@ public class TaskRun implements TenantInterface {
             ", value=" + this.getValue() +
             ", parentTaskRunId=" + this.getParentTaskRunId() +
             ", state=" + this.getState().getCurrent().toString() +
-            ", assets=" + this.getAssets() +
+            ", assetEmits=" + this.getAssetEmits() +
             ", attempts=" + this.getAttempts() +
             ")";
     }
--- a/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
+++ b/core/src/main/java/io/kestra/core/models/executions/TaskRun.java
@@ -5,7 +5,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonSetter;
 
 import io.kestra.core.models.TenantInterface;
 import io.kestra.core.models.assets.AssetsInOut;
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/kestra/core/models/executions/TaskRun.java
```diff
@@ -55,9 +57,12 @@ public class TaskRun implements TenantInterface {
     @With
     List<TaskRunAttempt> attempts;
 
+    // Lineage bundles, one AssetsInOut per (inputs -> outputs) pair, kept unmerged so each becomes its own
+    // lineage event. Canonical store for a taskRun's assets (the legacy single `assets` folds in via setAssets).
```
loicmathieu:
```suggestion
    // Lineage bundles, one AssetsInOut per (inputs -> outputs) pair, kept unmerged so each becomes its own lineage event.
```

core/src/main/java/io/kestra/core/models/executions/TaskRun.java
```diff
@@ -91,6 +96,35 @@ public TaskRun clearOutputs() {
...
+     * @deprecated use {@link #getAssetEmits()}. Returns the first bundle so callers still expecting a
+     *             single `assets` (e.g. the API layer) keep working. Not serialized.
+     */
+    @Deprecated(forRemoval = true, since = "2.0.0")
+    @JsonIgnore
+    @Nullable
+    public AssetsInOut getAssets() {
+        return this.assetEmits == null || this.assetEmits.isEmpty() ? null : this.assetEmits.getFirst();
+    }
+
```
loicmathieu: Not needed in 2.0 (if you want to backport to 1.3 you can add it there)

loicmathieu (changes requested):
ApiTaskRuns must be adapted to expose the new list of assetEmits and not the old assets. If you plan to backport in 1.3 you can keep the compatibility here.

core/src/main/java/io/kestra/core/models/executions/TaskRun.java
Malaydewangan09: Okay, thanks!

Malaydewangan09:
Okay, fixed it!

loicmathieu (approved):
LGTM

PR #18012, opened 2026-08-10
Replace execSync with execFileSync for git command

elevatebart:
### ✨ Description

Replaces `execSync` with `execFileSync` in the commit plugin to improve security and robustness when executing the `git log` command. This change separates the command from its arguments, preventing potential shell injection vulnerabilities and making the code more maintainable.

### 🔗 Related Issue

No related issue.

### 📝 Additional Notes

- Changed from `execSync(`git log -1 --format=${format}`)` to `execFileSync("git", ["log", "-1", `--format=${format}`])`
- `execFileSync` is safer as it doesn't spawn a shell and treats arguments as separate parameters
- No functional changes to the output or behavior of the plugin

### 🤖 AI Authors

Why did the cat sit on the computer? Because it wanted to keep an eye on the mouse! 🐱

https://claude.ai/code/session_01GCfPpK1pSZmM8iftM7UzaA

### DIFF
--- a/ui/plugins/commit.ts
+++ b/ui/plugins/commit.ts
@@ -1,7 +1,7 @@
 import type {Plugin} from "vite"
-import {execSync} from "child_process"
+import {execFileSync} from "child_process"
 
-const getInfo = (formats: string[]): string[] => formats.map(format => execSync(`git log -1 --format=${format}`).toString().trim())
+const getInfo = (formats: string[]): string[] => formats.map(format => execFileSync("git", ["log", "-1", `--format=${format}`]).toString().trim())
 
 const comment = (message: string, author: string, date: string): string => `
 <!--


elevatebart:
CI status update: two current failures on this PR, neither caused by this diff (which only touches `ui/plugins/commit.ts`, swapping `execSync` for `execFileSync`):

1. **`trigger-ee` / EE compile check** — the Kestra webhook that runs the compile check returned HTTP 500 with an empty verdict, both on the original run and after a manual re-run. The workflow's own comments call this the "transient infra error" case, but two consecutive identical failures suggest the internal `kestra.engineering.ci` endpoint is having an outage right now rather than a one-off blip.
2. **`kestra-ee/backend-tests`** — also red on the unrelated sibling PR #18013 (a ReDoS fix touching only `ui/packages/topology`) at almost the same timestamp, confirming this is a base/infra-wide failure, not something introduced by either diff.

Holding off on further re-runs until the underlying CI infra recovers — will retry once it's healthy and merge once a reviewer approves and checks are green.

---

MilosPaunovic (approved):

PR #18018, opened 2026-08-10
fix(plugins): space a property's title and description apart in the compact docs panel

### LINKED ISSUE #17948: Plugins inline docs are not showing the description (only title)
### Describe the issue

Image

Image

```yaml
id: flow_7895809
namespace: company.team

tasks:
  - id: purge
    type: io.kestra.plugin.core.execution.PurgeExecutions
    endDate: "{{ now() | dateAdd(-1, 'MONTHS') }}"
    states:
      - KILLED
      - FAILED
      - WARNING
      - SUCCESS
```

### Environment

- Kestra Version: develop

SaadAhmed7:
Closes https://github.com/kestra-io/kestra/issues/17948.
Related to https://github.com/kestra-io/kestra/pull/18514.

### ✨ Description

**This `PR` changed shape twice while it was open - the description below reflects what the branch actually does now, not the approach it was opened with.**

The original complaint in the issue - the compact properties view rendering a property's `title` and never its `description` - was fixed in the meantime by https://github.com/kestra-io/kestra/pull/18514, which split the single `#markdown` slot call into one per field:

```vue
<slot v-if="property.title" name="markdown" :content="sanitizeForMarkdown(property.title)" />
<slot v-if="property.description" name="markdown" :content="sanitizeForMarkdown(property.description)" />
```

Both fields render now, but they render flush against each other, because neither spacing rule on `.compact-prop-desc` can reach them:

- `p + p` does not match, because each paragraph is the only child of its own container.
- `.ks-markdown + .ks-markdown` - added by that same `PR` to compensate - does not match either, because the two `.ks-markdown` elements are not siblings. `SchemaToHtml.vue` wraps *every* `#markdown` slot render in its own `div.markdown`, so the real `DOM` is:

```html
<div class="compact-prop-desc">
  <div class="markdown"><div class="ks-markdown"><p>title</p></div></div>
  <div class="markdown"><div class="ks-markdown"><p>description</p></div></div>
</div>
```

Measured in a running instance against real `PurgeExecutions` schema data, the gap between the two lines is **0px**.

This `PR` renders both fields through a **single** slot call instead, joining them as two markdown paragraphs:

```ts
function propertyDoc(property: JSONProperty): string {
    return [property.title, property.description]
        .filter((text): text is string => Boolean(text))
        .map(sanitizeForMarkdown)
        .join("\n\n")
}
```

That puts the two `<p>` back inside one `.ks-markdown`, where `p + p` applies, giving the intended `var(--ks-spacing-2)` gap - **8px** measured. Each field still goes through `sanitizeForMarkdown` individually, and the now-unreachable `.ks-markdown + .ks-markdown` rule is removed.

Properties that define only a `title` or only a `description` are unaffected, and the expanded view (`PropertyDetail.vue`) is untouched.

### 🖼️ Before / After

**Before** - `title` and `description` run together, no gap:

before

**After** - one paragraph gap between them:

after

### 🎨 Frontend Checklist

- [x] `eslint` clean on the changed file
- [x] Screenshots attached showing the `UI` change


SaadAhmed7:
Before and After Screenshot
kestra-17948-before
kestra-17948-after

ui/src/components/plugins/schema/SchemaPropertiesSection.vue
```diff
@@ -86,7 +86,7 @@
                             >
                                 <slot
                                     name="markdown"
-                                    :content="property.title || property.description || ''"
+                                    :content="property.description || property.title || ''"
```
flcarre: Nice catch on the root cause — the precedence really was backwards here.

One thing though: swapping the order trades one missing field for the other. The issue's screenshot boxes the **description** as the missing piece, but the `title` is what tells you *what the property is*:

| | `namespace` |
|---|---|
| `title` | Namespace whose flows need to be purged, or namespace of the flow that needs to be purged |
| `description` | If `flowId` isn't provided, this is a namespace prefix, else the namespace of the flow. |

After this change `namespace` shows only the caveat and never says what it is. Same for `states` in your after-screenshot: "If not set, executions for any states will be purged." — the "what" is gone. Across `PurgeExecutions` the pattern is consistent: title = what it is, description = the caveat.

The expanded view (`PropertyDetail.vue`) already renders both, so doing the same in the compact view is the consistent choice:

```suggestion
                                    :content="[property.title, property.description].filter(Boolean).join('\n\n')"
```

The existing `.compact-prop-desc :deep(p + p)` rule already spaces the two paragraphs. Worth noting that two separate `<slot>` calls would *not* work here — `KsMarkdown` wraps each render in its own `div.ks-markdown`, so `p + p` would not match and the two lines would sit flush.

If you would rather keep the template lean, a small helper in `<script setup>` reads nicer:

```ts
const propertyDoc = (property: JSONProperty) =>
    [property.title, property.description].filter(Boolean).join("\n\n")
```

Either way the `v-if` guard on the line above stays correct as-is.

flcarre (commented):
Thanks for picking this up @SaadAhmed7 — clear write-up, correct diagnosis, and scoping the change to the compact view only is exactly right. The guard on the line above really did already imply the description should be showing.

I left one suggestion inline: rather than swapping the precedence, I think the compact view should render **both** the title and the description, the way the expanded `PropertyDetail` already does. Otherwise properties like `namespace` and `states` lose the sentence that says what they *are* and keep only the caveat.

Two smaller, non-blocking notes:

- **Sanitizer consistency** — `PropertyDetail` passes content through `sanitizeForMarkdown()` and `SchemaToHtml` through `normalizeColons()`; the compact property line does neither. Descriptions carry code fences and colons far more often than titles do, so this pre-existing gap becomes more visible with your change. A nice thing to fold in while you're on this line, but fine as a follow-up too.
- **Screenshot check** — in your after-screenshot the backticks around `` `namespace` `` render literally. The real path goes through `KsMarkdown` (see `PluginDocumentation.vue`), which renders those as inline code, so the Storybook harness likely stubbed the `#markdown` slot. Worth a quick look in a running instance to confirm the final rendering.

Nothing else on my side — no security, performance, i18n, or design-token concerns. Thanks again for the contribution!

SaadAhmed7:
Thanks @flcarre — you're right, and the reasoning is better than mine. Swapping the precedence fixed the reported symptom but dropped the "what": in my own after-screenshot `states` ended up reading only *"If not set, executions for any states will be purged."* I'd actually weighed rendering both before opening the PR and talked myself out of it; the fact that `PropertyDetail` already does it should have settled it.

Pushed your suggestion. I went with the `<script setup>` helper since it keeps the template lean:

```ts
function propertyDoc(property: JSONProperty): string {
    return [property.title, property.description]
        .filter((text): text is string => Boolean(text))
        .map(sanitizeForMarkdown)
        .join("\n\n")
}
```

I also folded in the **sanitizer** note rather than leaving it as a follow-up — it's a one-call change and it makes the compact line consistent with `PropertyDetail`. Your point about two separate slots breaking `p + p` matched what I found; the single joined string keeps the existing `:deep(p + p)` rule working.

On the **screenshot** — good catch, and you diagnosed it exactly. My Storybook harness had stubbed `#markdown` with a plain `<span>`, so nothing went through `KsMarkdown` and the backticks came out literal. The layout was real but the markdown rendering wasn't, and I should have flagged that when I posted them. I've rebuilt the harness to use the real `KsMarkdown`, mirroring `PluginDocumentation.vue`:

```jsx
markdown: ({content}) => <KsMarkdown content={content} />
```

and asserted on the rendered DOM rather than trusting my own eyes:

| | before | after |
|---|---|---|
| `.compact-prop-desc p` | 4 | **7** |
| inline `<code>` tags | 2 | 2 |
| literal backticks in text | none | none |

7 paragraphs is three properties rendering title + description plus `purgeExecution` rendering title only, which is the fallback behaving correctly. Backticks now render as inline code chips in both.

Fresh before/after attached below. `npm run build` passes and `oxlint` is clean on the changed file.

flcarre (commented):
The code here looks good to me. `propertyDoc()` reads cleanly, the `v-if` guard above it already covers exactly the cases the helper can produce so there is no empty-row risk, and rendering both fields lines up with how `PropertyDetail.vue` already does it.

One request before merge, and it is about the PR rather than the code.

The title, the description and both screenshots still describe the first approach ("swaps the precedence to `description || title`"), but what is on the branch now is the later commit that renders *both* fields as two paragraphs, after review feedback. Anyone approving on that "After" screenshot would be approving behaviour that no longer ships. Could you refresh the title, the description and the screenshot to match what the branch actually does?

I checked the rendering path while I was in there and it is safe: the content goes through `KsMarkdown`, which builds VNodes rather than using `v-html`, and `xssProtection` defaults to `true` with neither caller overriding it. So markdown-authored HTML in a plugin description cannot execute.

One open question for the team rather than for you: this roughly doubles the text per property in the *compact* panel, and `.compact-prop-desc` has no clamp. That cuts slightly against the PR's own rationale, that `title` mostly restates the property name shown directly above it. Might be worth a quick word with design on whether the compact view is meant to grow that much, but that is a product call and should not hold up the fix.

Thanks for iterating on this one, the second approach is the better outcome.

flcarre (commented):
Round 3 on `05498582`. The code is finished as far as I am concerned — `propertyDoc()` renders both fields as two paragraphs, and you folded in the sanitizer note from my last round by running each through `sanitizeForMarkdown`. The comment explaining why it is one slot call rather than two (so `KsMarkdown` does not break the `p + p` spacing) is the kind that earns its place. Nothing further on the implementation.

Two things still stand between this and merge, neither of them code.

**1. The title and description still describe the approach that was replaced.** This is the request from my last round, unchanged.

The title reads *"show property description **instead of** title"*, and the body says *"The fix swaps the precedence to `property.description || property.title || ''`"* and *"the `description` is shown **instead**"*. The branch renders **both**. The two screenshots show the old behaviour, and the "After" one in particular shows `flowId` with only the description — not what a user will see.

This is worth more than tidiness: the squash commit inherits the title and body, so the permanent history would describe behaviour that never shipped, and anyone approving off that "After" screenshot is approving the wrong thing. `fix(plugins): show both title and description in compact plugin docs` would match what the branch does.

**2. The PR now conflicts with `develop`** (`mergeable_state: dirty`), which it did not last week. Merging `develop` in should be enough — the diff is a single file and one function.

Once those two are sorted this is good to go from my side. Thanks for sticking with it through three rounds; the second approach really is the better outcome.

---

MilosPaunovic:
Hey @SaadAhmed7, are there any updates on this?

MilosPaunovic (approved):
Round 4 on `39e48b2`. Both items from my last round are closed: the branch merges cleanly again, and the title and description now describe what actually ships.

Worth recording why this change is still needed at all, because its purpose moved while it was open. https://github.com/kestra-io/kestra/pull/18514 landed the both-fields rendering in the meantime, so the original "description never shows" complaint is already fixed on `develop`. What that `PR` left behind is a spacing defect: it added `.ks-markdown + .ks-markdown` to separate the two renders, but `SchemaToHtml.vue` wraps every `#markdown` slot render in its own `div.markdown`, so the two `.ks-markdown` elements are never siblings and the rule cannot match. `p + p` cannot match either, since each paragraph is alone in its own container. I measured it in a running instance against real `PurgeExecutions` schema data: `develop` renders a **0px** gap between a property's title and description, this branch renders **8px**.

So `propertyDoc()` is not a stylistic alternative to what shipped - it is what makes the paragraph spacing reachable again, and the comment above it now says so precisely. Dropping the unreachable `.ks-markdown + .ks-markdown` rule is the right call too.

For transparency: I resolved the `develop` conflict on this branch, and while in there I removed that dead rule and expanded the comment above `propertyDoc()`. The implementation itself is @SaadAhmed7's.

`npm run check:types` and `eslint` are clean, and the net diff against `develop` is one file.

Approving. One item for after the merge: `releases/v2.0.x` carries the same two-slot version, so it needs its own backport - `releases/v1.3.x` and `releases/v1.0.x` do not have `ui/src/components/plugins/schema/` at all.

Issue #14876, opened 2026-03-02
Submitted Task Run State does not have a colour


wrussell1999:
### Describe the issue

<img width="559" height="254" alt="Image" src="https://github.com/user-attachments/assets/0a0b4798-776c-4d8d-8be6-2f3cd4cc0928" />

### Environment

- Kestra Version: develop

wrussell1999:
Seems like a new state. Related: https://github.com/kestra-io/docs/issues/4117

MilosPaunovic:
Introduced in https://github.com/kestra-io/kestra/pull/11764.

MilosPaunovic:
Can you help me out here @wrussell1999 with a flow reproduction? 🙏🏻

wrussell1999:
Should work with any flow.

```yaml
id: submitted
namespace: company.team

tasks:
  - id: hello
    type: io.kestra.plugin.core.log.Log
    message: Hello World! 🚀
```

To reproduce:
1. Execute the flow
2. Go to the Gantt view
3. Click on the bar part of the Gantt view to expand and display logs
4. Hover over the duration on the right side to the left of the taskrun status.

https://github.com/user-attachments/assets/c5b94023-4cde-439d-bf09-202a9d9a5197

<img width="936" height="356" alt="Image" src="https://github.com/user-attachments/assets/fb32365a-bc7d-4b39-840a-0c9c99e13a30" />

MilosPaunovic:
Okay, I can make the temporary fix here, but what we would need for a permanent solution is for @Nico-Kestra to add another variable into Figma, called `--ks-chart-submitted` and export & attach the file here, so we can recreate the variables with it.

PR #18029, opened 2026-08-10
fix(executions): make task duration and state history readable

### LINKED ISSUE #17954: Better ux for status
### Feature description

Image

This status components are useful to understand the behavior, but actually hard to be read

flcarre:
## Summary

- Replaces the shared `Duration` tooltip (topology node chip, Gantt duration column, Logs task-run line) — today it lists four absolute wall-clock instants (`CREATED: 2026-08-07 15:36:15.804`, ...) while the Gantt bar 40px to its left computes and shows a *different*, conflicting queued/running breakdown for the same task on the same row.
- **Tier 1 (hover):** one breakdown card — total duration, a DS state badge, a two-segment queued/running proportion bar, and one absolute anchor instant (`Started`/`Failed at`/`Skipped at`/`Created`, depending on state).
- **Tier 2 (click, "State history"):** the full transition list, one row per entry (`HH:mm:ss.SSS` leading, `tabular-nums`), the delta between consecutive entries labelled on the dashed connector, grouped per attempt, bounded to `max-height: 15rem` with a `Copy` action — replacing today's unbounded, `nowrap` tooltip.
- The trigger is now a real `<button type="button">` (was a `<span>`, unreachable by keyboard), with a dotted-underline affordance and an `aria-label`.

**Behaviour change (deliberate):** the queued/running split is now computed by a pure `computeDurationBreakdown()` from the *full* history, not just `runningState[0]` as `Gantt.vue` did before. The old logic billed *everything after the first RUNNING entry* as running time, so a retried task's inter-attempt waiting (FAILED → RETRYING → next RUNNING) was silently counted as work. The new function treats every non-RUNNING span as queued, proven by a unit test asserting the inter-attempt gap lands in `queued`.

**New DS badge + token:** `SUBMITTED` had no color of its own — `Duration.vue` mapped it to `CREATED`'s blue with an in-code "minor hack" comment (#14876), making the two states visually identical. Added a first-class `SUBMITTED` entry to `KsExecutionStatus` (icon `TrayArrowDown`) plus `--ks-status-submitted` / `--ks-status-border-submitted` / `--ks-status-background-submitted` in `ks-tokens.scss` (not the generated `ks-theme-*.scss` files, which `scripts/generate-palette.mjs` overwrites from Figma).

**Known follow-ups, intentionally out of scope:**
- Pre-existing light-mode contrast failure on the topology status chip (`--ks-text-success` on `--ks-bg-tag` ≈ 2.09:1 at `--ks-font-size-2xs`, under AA) — needs a design decision on a chip-specific token.
- `--ks-tooltip-*` token drift: Figma declares `ks/tooltip-background` / `ks/tooltip-border`, neither exists in code; `KsTooltip.vue` uses `--ks-bg-input` / `--ks-border-default`. This PR follows the code tokens.

## Test plan

- [x] `npm run test:unit` passes (1183 tests, including a new pure-function suite for `computeDurationBreakdown` covering: no history, never-ran/SKIPPED, still-running, multi-attempt with an inter-attempt gap, a midnight-spanning history, and `queued + running === total`) and a `Duration.vue` component suite covering the trigger's disabled/dash state, the tier-1 split rendering, the "did not run" card, attempt grouping, and the day separator
- [x] `npm run test:lint` passes (no new warnings/errors)
- [x] `npm run translations:check` reports "No missing keys. No extra keys." for all 12 locales (new `state_history.*` keys translated in the same PR)
- [x] Verified all three call sites (`TaskNode.vue`, `Gantt.vue`, `TaskRunLine.vue`) still render via the shared component with no prop changes needed

closes: https://github.com/kestra-io/kestra/issues/17954

## Demo

https://github.com/user-attachments/assets/2524749d-c0fc-4106-91a6-5a57e351ba46

Recorded against a local instance on the branch. Scenes: the Gantt duration column, the
queued-vs-running breakdown on a 3-attempt retried task, the state-history panel, a paused
task's own bucket, and the same card in dark mode.


flcarre:
@tchiotludo Does this meet your expectations ?

tchiotludo:
@flcarre more than expected !! thanks a lot !

tchiotludo (approved):

RemiBarthe:
Nice work ! Extracting `computeDurationBreakdown()` into a pure tested function is the right seam, and dropping the Gantt bar's competing tooltip genuinely fixes the inconsistency in the issue. Nothing blocking; three follow-ups.

1. **`"3 Attempt(s)"`**
 `Duration.vue:38` and `:326` concatenate `{{ displayedAttemptCount }} {{ $t('attempts') }}`, and `attempts` is `"Attempt(s)"` in `en.json`. So a retried task's card reads *"12.4s 3 Attempt(s)"* and the header *"2026-08-07 · 3 Attempt(s)"*. `Gantt.vue:119` already dodges this with `.toLowerCase()`. A dedicated key (`"state_history.attempt_count": "{count} attempts"`) fixes both and keeps `ui/AGENTS.md` happy.

2. **Two more call sites than the test plan mentions.** 
`Duration` from `@kestra-io/topology` is also mounted at `Gantt.vue:47` (header `total_duration`) and `overview/components/Banner.vue:137` (banner footer stat) — both execution-level and untouched here, so both now get the dotted-underlined button and the hover card. Worth a screenshot of each, and `state_history.aria_open` ("Show the state history for this task **run**") reads oddly there.

3. **The live counter doesn't resume after a retry.** 
`paint()` (`Duration.vue:236`) only starts the interval when `breakdown.isRunning` is true at that moment, and the watcher only re-calls it when the *first* history entry's date changes — which never happens. So after `FAILED` cancels the tick, the following `RUNNING` never restarts it: the value steps forward at each transition but doesn't tick for the whole second attempt. Same out of `QUEUED`/`RETRYING` (both `isRunning: false` in `state.ts`). This isn't a regression — the old code froze identically — but since this block is being rewritten anyway, one line covers it:

```ts
watch(() => breakdown.value.isRunning, (running) => running ? paint() : cancel(), {immediate: true})
```

replacing `onMounted(paint)` and the `normalizedHistories` watcher.

Nit: the description says the new `SUBMITTED` token fixes it being visually identical to `CREATED`, but both resolve to `--ks-status-info` — only the icon differs. The tokens themselves are needed for the `$statusList` loop, so nothing to change in the code.

RemiBarthe (approved):

PR #18033, opened 2026-08-10
feat(executions)!: store the execution outputs outside of the execution

loicmathieu:
### ✨ Description

Execution-level outputs (the flow `outputs:` block) were rendered at the end of every execution and stored **inside** the `Execution` object. They therefore travelled through the queue, bloated the `executions` table, and were returned in *every* item of the execution search endpoint.

This does for execution outputs what was already done for task run outputs: they now live in a dedicated `execution_outputs` table.

- The executor stores them when an execution ends (`ExecutorService.onEnd`), and a new `ExecutionOutputService` reads them back — spilling to internal storage above `kestra.execution.outputs.limit`, exactly like task outputs. The serialization / offloading logic is now shared through an `AbstractOutputService`.
- A new `GET /api/v1/{tenant}/outputs/executions/{executionId}` endpoint exposes them, and the UI reads it from the execution Outputs tab.
- `Execution.outputs` is kept as a `@Deprecated @Hidden` pre-2.0 compatibility field: it is read first when present and migrated to the new table on ingest, so existing executions keep displaying their outputs with no batch migration.
- Second commit: the two pre-existing task-output endpoints move from `/outputs/{executionId}[/{taskRunId}]` to `/outputs/tasks/{executionId}[/{taskRunId}]`, so every route under the controller has a literal second segment and they can no longer shadow each other.

Runtime payloads that *are* the feature keep serving outputs, loaded from the new table: webhook responses, the `subflow()` function result, and flow-trigger `trigger.outputs`.

### 🔗 Related Issue

None — follow-up to the task-run outputs work.

### 🎨 Frontend Checklist

- [x] Code builds without errors — `npm run check:types` and `eslint` are clean on the changed files (full `npm run build` not run)
- [ ] All existing E2E tests pass — **not run locally**
- [ ] Screenshots or video recordings attached — no visual change; the Outputs tab renders the same data, just fetched from the new endpoint

The generated SDK was regenerated. Endpoint identifiers are unchanged, so the exported client functions keep their names and no UI call site changed for the `/tasks` move.

### 🛠️ Backend Checklist

Verified: full `:core:test` (3179 tests) and `:executor:test`; `H2RunnerTest`; the new repository test 5/5 against **H2, PostgreSQL and MySQL** (so the migration is exercised on all three dialects); `OutputControllerTest`, `ExecutionControllerRunnerTest`, `McpToolServiceTest`, and the trigger/webhook/subflow suites. The only failure in the full run was the pre-existing `HttpClientTest.errorUnreachable`, which asserts on an English OS error string and fails on a French locale — unrelated.

### 🤖 AI Authors

This PR was raised with AI assistance, so, as instructed: our cat tried to help review the outputs refactor, but he insisted every single one belonged **inside** the execution — he's a firm believer in keeping things in the `litter`al object. He was overruled; the outputs now live in their own table, and he has retreated to the storage layer in protest. 🐱


AcevedoR:
Reviewed locally, nice change. Three things:

- **Perf:** `RunContextFactory.of(flow, execution)` always loads the outputs, so `FlowTriggerService` reads the same row 1 + N times per terminated execution (once up front, then once per flow with a flow trigger). An `of(...)` overload taking the already-loaded outputs would avoid it.
- **Bug:** the `PurgeResult` reduce in `ExecutionService` doesn't sum `executionOutputsCount`, so a purge spanning more than one batch reports only the first batch (`taskOutputsCount` has this too). Found by reading the code, not tested.
- **Nit:** `getOutputs` returns an empty map instead of null, so webhook responses now send `"outputs": {}` where they used to omit the key — `Flow.evaluate` already turns empty into null for exactly this reason.

AcevedoR (commented):
see comment above

loicmathieu:
@AcevedoR 

> Perf: RunContextFactory.of(flow, execution) always loads the outputs, so FlowTriggerService reads the same row 1 + N times per terminated execution (once up front, then once per flow with a flow trigger). An of(...) overload taking the already-loaded outputs would avoid it.

This would leak RunContext creation details outside of the RunContextFactory, so I'd rather not do it now. 

Loading all outputs upfront when creating the RunContext is anyway not a good idea so I already have a follow-up (with a PoC) to lazily load outputs when needed. I'm not sure it will be delivered for 2.0, but surely for 2.1. 

Moreover, most executions didn't carry any outputs; what could be done is to check if the flow defines outputs before loading them, I'll open a followup for that later, easy move.

I'll check the other two feedback later.

AcevedoR (approved):

PR #18040, opened 2026-08-11
fix(core): ignore plugin-group types from blueprint includedTasks

### LINKED ISSUE #18031: [2.0] Plugin Groups display as plugin tasks in blueprints
A lot of Blueprints use Plugin Defaults which is changed in 2.0. As some plugins use the plugin group in plugin defaults when using a variety of tasks (AWS, Docker, DB etc), this is causing confusion with blueprints and showing the plugin group as a task, but then it says the task doesn't exist (as it's not a task) and will not let you open the blueprint in the editor. 

In this example (`/blueprints/flow/community/algolia-index-sync`), we have 2 valid postgres tasks but the plugin default for the postgresql plugin as a while shows as invalid.

Image

The Docker example is below is slightly different. This is due to a renaming of the tasks when you actually copy the flow code and paste it into the editor.

Image

Solution:
- Fix blueprints to be valid in 2.0

### Describe the issue

Docker Plugin is not included with the RC7 image of 2.0. Only the Docker task runner is included. This is the image with all plugins.

```
Requires plugin(s) not installed on this instance: docker
```

Image

### Environment

- Kestra Version: RC7 2.0 EE

Dharshika-11:
## What change was made?

This PR fixes an issue where blueprint `pluginDefaults` using plugin group/package names were incorrectly treated as missing task types in the UI.

In Kestra 2.0, blueprints can define default values for a plugin namespace, for example:

* `io.kestra.plugin.jdbc.postgresql`
* `io.kestra.plugin.docker`

These values represent **plugin package/group names**, not actual Java task classes. However, they were included in the `includedTasks` list returned by the Community API.

The UI interpreted these package names as task types and attempted to determine whether they were installed. Since these package/group names are not registered as task types, the blueprint was incorrectly considered to have missing plugins.

This resulted in:

* The **Use** button being disabled.
* Incorrect/missing plugin indicators.
* Broken task icons being displayed for plugin groups.

## How was it fixed?

A new helper function, `isTaskClass(cls: string)`, was introduced to distinguish actual task classes from plugin group/package names.

The helper examines the last segment of the class name:

* If the last segment starts with an **uppercase letter**, it is treated as a task class.
* If the last segment starts with a **lowercase letter**, it is treated as a package/group name and ignored.

The filtering was applied to the following components:

* `useBlueprintPlugins.ts`
* `BlueprintCard.vue`
* `BlueprintOverview.vue`
* `BlueprintListRow.vue`
* `BlueprintIconStack.vue`

This prevents plugin group names from being incorrectly reported as missing tasks or displayed as task icons.

## Tests

Unit tests were added to `useBlueprintPlugins.spec.ts` to verify the new `isTaskClass` behavior and ensure that plugin group/package paths are correctly ignored.

### Verification performed

* All **13 unit tests passed**.
* Lint checks passed successfully.
* Verified that plugin group names are no longer treated as missing task types.
* Verified that broken task icons are no longer displayed for plugin groups.

## Related Issue

Fixes #18031

### DIFF
--- a/ui/src/components/flows/blueprints/BlueprintCard.vue
+++ b/ui/src/components/flows/blueprints/BlueprintCard.vue
@@ -63,7 +63,7 @@
     import TaskIcon from "../../plugins/TaskIcon.vue"
     import AlertCircleOutline from "vue-material-design-icons/AlertCircleOutline.vue"
     import {canCreate} from "override/composables/blueprintsPermissions"
-    import {useBlueprintPlugins} from "../../../composables/useBlueprintPlugins"
+    import {blueprintTaskTypes, useBlueprintPlugins} from "../../../composables/useBlueprintPlugins"
     import type {BlueprintTag, FlowBlueprint} from "../../../stores/blueprints"
 
     const {t} = useI18n()
@@ -101,7 +101,7 @@
     )
 
     const tasks = computed(() =>
-        [...new Set(props.blueprint.includedTasks)],
+        blueprintTaskTypes(props.blueprint.includedTasks),
     )
     
     const visibleTasks = computed(() =>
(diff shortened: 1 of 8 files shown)


Piyush-r-bhaskar (approved):
LGTM

PR #18041, opened 2026-08-11
feat(flows): cross-resource Source Search across flows, files, KV and secrets

flcarre:
## Summary

Extends Source Search from flows-only to a **frontend-only** cross-resource scope covering four types, each reachable from an existing endpoint with no backend change:

| Type | Endpoint | What is searched | Cost |
|---|---|---|---|
| Flows | `GET /flows/source` | File content, with line and column. Replace supported. | 1 request |
| Namespace files | `GET /{namespace}/files/search` | The file **path** only, never the content. | N requests, one per namespace |
| KV keys | `GET /kv` | Key names only, cross-namespace. | 1 request |
| Secret keys | `GET /secrets` | Key names only. Values never read. | 1 request |

- `ui/src/stores/crossResourceSearch.ts` / `ui/src/utils/crossResourceSearch.ts`: the per-type search/state store and pure helpers (from a prior session in this branch). Extended here, additively, with `resourceCountFor`, an `errorMessage` on a fully-failed namespace-files fetch (RBAC), and `creationDate`/`expirationDate` on KV entries for the preview card.
- `FlowsSearch.vue`: a "Search in" pill row (`KsCheckTag`, pill) with live per-type counts and status, "Select all types", a flows-only "Flow section" scope (renamed from the old plain segmented control), a cross-resource summary + match navigation, partial-result banners (per-type failure, namespace-files fan-out progress), and an explicit exclusion breakdown when Replace is open.
- `SourceSearchResults.vue`: results grouped by type then by resource, with three row shapes — flow (line + snippet, replaceable), namespace file (path, directories dimmed), KV/secret (key, never a value) — and a per-namespace retry affordance for failed file searches.
- `SourceSearchPreview.vue`: the flow diff/editor preview is unchanged; a new metadata-card preview covers files/KV/secrets, each stating plainly what was **not** searched, with the value always withheld for KV and secrets.

All four types are always searched (search-in pills are a display/replace filter, not a fetch gate) so a type you've hidden can still surface as "1 KV key matches in a type you haven't selected" in the empty-results hint, and a permission failure on any type shows up immediately as a failed pill.

### Explicit out of scope
No search endpoint exists for these (three are EE-only): Apps, Tests, Assets, Variables, Plugin defaults. Also out of scope: namespace-file **content** (would require one HTTP request per file) and Replace on anything but flows (the replace API is flow-typed).

### Deviations from the mockup
- **RBAC pill (frame 3):** the mockup shows a type the user can't read as pre-disabled with a permission tooltip. The frontend has no way to know that before calling the endpoint, so instead a forbidden type is fetched, fails, and surfaces as a **failed** pill with `errorMessageFor`'s message — functionally equivalent, one round trip later.
- **File preview metadata:** `GET /{namespace}/files/search` returns paths only (no size/modified date), so the file metadata card omits those rows rather than inventing them.
- **"Open in KV store" / "Open in secrets"** link to the respective list page; there's no confirmed deep-link query param to preselect a specific key.
- Per-toggle hint copy (Aa vs ab vs `.*`) is unified into one generic "Flows only" sentence rather than three separate mockup variants, to keep the translation surface bounded.

## Test plan

- [x] `npx vitest run --project=unit` — 1201 passed
- [x] `npm run check:types` — clean
- [x] `npm run test:lint` — 0 errors (pre-existing unrelated warnings only)
- [x] `npm run translations:check` — no missing/extra keys across all 12 locales
- [x] Storybook stories updated for `SourceSearchResults` and `SourceSearchPreview` covering all four types, replace mode, progressive/failed namespace states, and dark mode

## Demo

One query across all four types, then narrowing the scope with the **Search in** pills. Flows match on file content, Namespace files on the path, KV and Secrets on the key name.

https://github.com/user-attachments/assets/c4a058eb-6058-4066-be73-681247782dcb

part-of: https://github.com/kestra-io/kestra-ee/issues/8644


ui/src/components/flows/FlowsSearch.vue
```diff
@@ -265,12 +361,11 @@
     const route = useRoute()
     const router = useRouter()
     const toast = useToast()
+    const store = useCrossResourceSearchStore()
```
elevatebart: You could call it `crossRessourceSearchStore` to make search/replace the code easier

ui/tests/storybook/components/SourceSearchResults.stories.tsx
```diff
@@ -10,14 +11,33 @@ const i18n = createI18n({
...
+                tag_keys_only_values: "Keys only — values are not searched",
+                tag_paths_only: "Paths only — file content is not searched",
+                tag_search_only: "Search only — not replaced",
+                tag_source_code: "Source code",
+                type_files: "Namespace files",
+                type_flows: "Flows",
+                type_kv: "KV keys",
+                type_meta: "{matches} · {resources}",
+                type_secrets: "Secret keys",
             },
```
elevatebart: We repeat the content of this multiple times. Can we do without?

elevatebart (approved):
A few nitpicks, and those pesky little unit tests ><

ui/src/components/flows/FlowsSearch.vue
flcarre: Done, renamed to `crossResourceSearchStore` (40 call sites). I went with the correct spelling of "resource" rather than `crossRessourceSearchStore`, so it matches the composable (`useCrossResourceSearchStore`) and the module (`stores/crossResourceSearch.ts`) and stays greppable as one token.

ui/tests/storybook/components/SourceSearchResults.stories.tsx
flcarre: Yes, and it turned out to be worse than it looked: the same hand-copied `en.json` subset existed in four places (both story files and both specs). It was also dead weight, since Storybook's global preview already installs the real `en.json` via `initApp(app, [], {}, en)`, so the local `createI18n` never applied. Dropped it from both story files, and the two specs now import `en.json` directly (the `Dependencies.spec.ts` pattern) so they assert against real copy. Also collapsed the repeated per-story boilerplate into a `story()` factory plus a shared `allTypesFound` fixture and a `withLoadedFlow` decorator. Net `-444/+164`.

While in there I fixed the four failing `SourceSearchPreview` stories: the preview's "open in ..." buttons resolve named routes (`kv/list`, `secrets/list`, `namespaces/update/files`) that the global preview router does not declare, so `useLink()` threw. Added a `vueRouter` decorator with those routes; Storybook is 244/244.

PR #18057, opened 2026-08-11
test(executions): reproduce kestra-ee#9796 — lost post-join event wedges an execution RUNNING with its slot held

AcevedoR:
## What

A testkit reproducer for kestra-io/kestra-ee#9796 — *"The final task in a flow failed but execution is still marked as Running and still occupies concurrency slots"* — plus eight probes that rule out the pure decision logic.

**Stacked on #17257** (needs the executor testkit) — review only the last commit.

## The mechanism (reproduced in `redeliveredResultOfTheFailedFinalTaskDoesNotHealTheExecution`)

1. The final task's FAILED `WorkerTaskResult` arrives. `WorkerTaskResultMessageHandler` **only joins it** inside the state-store lock: the taskrun is persisted FAILED, the execution is persisted **still RUNNING**. The terminal transition (`handleEnd` → FAILED, slot release) belongs to the follow-up `UPDATED` event.
2. That `UPDATED` event is emitted by `DefaultExecutor#toExecution` **after** the join transaction committed — the two are not atomic. Executor crash/restart in that window (or a dropped message) loses the event.
3. At-least-once delivery redelivers the `WorkerTaskResult` — the execution's only chance to heal. But the redelivery dedup (`hasTaskRunJoinable` → lock returns empty) comes back **empty**, and `DefaultExecutor` calls `toExecution` only on a present context (`.ifPresent(this::toExecution)`).

Result: no event ever targets the execution again. It is wedged **RUNNING forever with its final task FAILED, and its concurrency slot is held forever** — with `behavior: CANCEL` every later execution is cancelled for good (with `QUEUE`, queued forever). Exactly the reported symptom.

The test drives the **real** handler twice over the testkit stores and asserts the wedged end-state (execution RUNNING, taskrun FAILED, `running(flow) == 1`, nothing in flight).

## What is ruled out

Eight sibling probes drive other "final task fails" shapes end-to-end (flow `errors` branch, failing errors task, Sequential / Parallel-with-`concurrent` / WorkingDirectory finals, exhausted retry, retry gated on a failing flowable errors branch) — **all terminate and release correctly**. The pure state-machine decisions are healthy; the wedge is the non-atomic join-then-emit plus the dedup that suppresses recovery.

## Fix directions (not in this PR — maintainer's call)

- **Heal on redelivery**: when the result is not joinable but the stored execution is non-terminal with all task runs terminal, still emit the `UPDATED` event (idempotent re-drive), or return the stored execution's context.
- **Or make it atomic**: emit the post-join `UPDATED` event through a transactional-outbox seam, like the delay / SLA-monitor / slot-release loops fixed in #17257 — this is the same bug class.

## Testing

`FinalTaskFailureReproTest`: 9 tests, 0 failures (the reproducer *passes* by pinning the wedge — once fixed, its final assertions flip and the test should be updated to pin the healed behavior).

### DIFF
--- a/executor/src/test/java/io/kestra/executor/statemachine/FinalTaskFailureReproTest.java
+++ b/executor/src/test/java/io/kestra/executor/statemachine/FinalTaskFailureReproTest.java
@@ -1,0 +1,384 @@
+package io.kestra.executor.statemachine;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
+import java.util.Optional;
+
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import io.kestra.core.models.executions.Execution;
+import io.kestra.core.models.executions.TaskRun;
+import io.kestra.core.models.executions.TaskRunAttempt;
+import io.kestra.core.models.flows.Concurrency;
+import io.kestra.core.models.flows.FlowWithSource;
+import io.kestra.core.models.flows.State;
+import io.kestra.core.models.property.Property;
+import io.kestra.core.models.tasks.retrys.AbstractRetry;
+import io.kestra.core.models.tasks.retrys.Constant;
+import io.kestra.core.runners.ExecutionEvent;
+import io.kestra.core.runners.ExecutionEventType;
+import io.kestra.core.runners.WorkerTaskResult;
+import io.kestra.executor.ExecutorContext;
+import io.kestra.executor.testkit.Executions;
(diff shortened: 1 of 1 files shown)


AcevedoR:
Closing: kestra-ee#9796 is now fixed (CLOSED/COMPLETED), so this reproducer has served its purpose. Happy to reopen if the testkit reproduction is still wanted as a regression guard — it is stacked on #17257.

PR #18058, opened 2026-08-11
fix(core): percent-encode '#' and '%' in InternalStorage URI construction

### LINKED ISSUE #18051: Internal storage: '#' in a filename silently truncates and overwrites objects, '%' makes them unreadable
## Summary

Two characters that `ScriptService` currently treats as valid in internal-storage paths are broken end to end:

- **`#`** — silently truncates the stored object name at the `#`, and makes distinct output files **collide and overwrite each other**.
- **`%`** — a filename containing a literal `%xx` sequence is silently stored under a *different* name, and the resulting URI is then unreadable (`FileNotFoundException`). A bare `%` throws an opaque `Malformed escape pair` error.

## Root cause

The write side builds URIs with the **parsing** `URI` constructors while the read side uses a character allowlist that knows nothing about percent-encoding.

`core/src/main/java/io/kestra/core/storages/InternalStorage.java:185-194` (and `:159-163`, `:253-262`):

```java
URI uri = context.getContextStorageURI();
URI resolved = uri.resolve(uri.getPath() + PATH_SEPARATOR + (name != null ? name : file.getName()));
```

`URI.resolve(String)` *parses* the string, so the raw filename is interpreted as URI syntax rather than escaped as data:

- `#` starts a **fragment**. `report#1.csv` yields path `.../report` + fragment `1.csv`. Every storage backend reads only `uri.getPath()`, so the object is written as `report` and the rest is dropped.
- `%xx` is read as an **existing escape** and decoded. `100%20.txt` is written to disk as `100 .txt`.
- A `%` not followed by two hex digits throws `IllegalArgumentException: Malformed escape pair`.

`#` passes validation on the way in because it is in the allowlist at `core/src/main/java/io/kestra/core/models/tasks/runners/ScriptService.java:36`:

```java
private static final String ALLOWED_PATH_CHARS = "-\\p{Alnum}._\\+~#=/,:;";
```

On the way back out, `LocalStorage` returns the URI built with the **quoting** constructor (`storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java:357-363`), which re-escapes the space back to `%20`. But `%` is *not* in `ALLOWED_PATH_CHARS`, so `INTERNAL_STORAGE_PATTERN` truncates the URI at the `%` when `ScriptService.replaceInternalStorage` scans the command.

Traced through the real code paths:

```
filename report#1.csv
  putFile resolved     : //tenant/ns/exec/task/report#1.csv
  storage path used    : /ns/exec/task/report      fragment=1.csv     <-- ".csv" lost
  URI stored in outputs: kestra:///ns/exec/task/report

filename report#2.csv
  storage path used    : /ns/exec/task/report                          <-- same object, overwrites
  URI stored in outputs: kestra:///ns/exec/task/report

filename 100%20.txt
  storage path used    : /ns/exec/task/100 .txt                        <-- name silently changed
  URI stored in outputs: kestra:///ns/exec/task/100%20.txt
  read-back regex match: kestra:///ns/exec/task/100     [TRUNCATED -> FileNotFoundException]

filename 100%.txt
  putFile -> IllegalArgumentException: Malformed escape pair at index 17
```

### `#` — silent overwrite via `{{ outputDir }}`

This path goes through `ScriptService.uploadOutputFiles`, which uses the relative filename with no unique prefix, so two files collapse onto one object:

```yaml
id: hash_filename_repro
namespace: company.team

tasks:
  - id: generate
    taskRunner:
    commands:
      - "echo 'FIRST'  > {{ outputDir }}/report#1.csv"
      - "echo 'SECOND' > {{ outputDir }}/report#2.csv"
```

**Expected:** two output files with distinct URIs preserving `#1.csv` / `#2.csv`.
**Actual:** `outputFiles` has both keys, but both URIs are `kestra://.../report` — one object, containing `SECOND`. The first file is silently lost, and the `.csv` extension is gone from the stored name.

### `%` — unreadable output file via `outputFiles:`

This path goes through `FilesService.outputFiles`, which does **not** call `validateStoragePath`, so `%` is never rejected:

```yaml
id: percent_filename_repro
namespace: company.team

tasks:
  - id: generate
    taskRunner:
    outputFiles:
      - "100%20.txt"
    commands:
      - "echo 'TEST' > '100%20.txt'"

  - id: read
    taskRunner:
    commands:
      - "cat '{{ outputs.generate.outputFiles[\"100%20.txt\"] }}'"
```

**Expected:** `read` prints `TEST`.
**Actual:** `read` fails with `FileNotFoundException` — the URI was truncated at `.../100`. The object on disk is also named `100 .txt`, not `100%20.txt`.

## Impact

Medium. `#` is explicitly advertised as supported by `validateStoragePath`, and `report#1.csv`-style names are ordinary. Losing an output file with no error at all is the worst failure mode here — same shape as #16896 (`+`/space collision in namespace files).

`%` matters less on its own, but it **blocks the correct fix**: any move to properly escape filenames produces `%xx` in the URI, which today's `INTERNAL_STORAGE_PATTERN` cannot match.

## Suggested fix

1. Build storage URIs with the multi-argument (**quoting**) `URI` constructor in `InternalStorage.putFile` (`InternalStorage.java:161`, `:187`) and the private helpers (`:255`, `:261`), instead of `URI.resolve(String)` / `URI.create(String)`. This is the pattern already used and documented at `NamespaceFileController.toFileUri` (`webserver/src/main/java/io/kestra/webserver/controllers/api/NamespaceFileController.java:369-377`), in `NamespaceFile.of` and in `LocalStorage.kestraUri`.

2. Add `%` to `ScriptService.ALLOWED_PATH_CHARS` so percent-escaped URIs survive `INTERNAL_STORAGE_PATTERN`. Without this, step 1 makes matters *worse* — correctly escaped URIs would start truncating.

3. Decide whether `#` should stay in `ALLOWED_PATH_CHARS`. Until step 1 lands, `validateStoragePath` accepting `#` is actively harmful — it green-lights a name that is then silently mangled.

## Environment

- Kestra `develop` @ `2842554900`
- Storage: `local` (`LocalStorage`); the `#` fragment truncation happens in `InternalStorage` and so is backend-independent

KishoreGhost:
## Summary

Fixes #18051

Two characters in filenames caused silent data corruption in Kestra's internal storage:

- **#** was interpreted as a URI fragment delimiter by URI.resolve(String), silently truncating the stored path and causing distinct output files to **collide and overwrite each other**.
- **%** was interpreted as a percent-escape start, silently decoding the filename to a different name; a bare % threw IllegalArgumentException: Malformed escape pair.

## Root cause

All four putFile helpers in InternalStorage called either URI.resolve(String) or URI.create(String) with a raw filename, which *parses* the string as URI syntax instead of treating the filename as opaque data.

## Fix

Replace the parsing constructors with the **quoting** four-argument URI(scheme, host, path, fragment) constructor — which percent-encodes any character that is not legal in a URI path — in all four affected sites. This follows the same pattern already established in NamespaceFile.of and NamespaceFileController.toFileUri.

Also update ScriptService.ALLOWED_PATH_CHARS:
- **Remove #**: alidateStoragePath was green-lighting a name that the old URI construction silently mangled.
- **Add %**: correctly percent-encoded kestra URIs (e.g. kestra:///path/report%231.csv) now survive INTERNAL_STORAGE_PATTERN matching so that 
eplaceInternalStorage can resolve them.

## Changes

| File | Change |
|------|--------|
| core/.../InternalStorage.java | Replace URI.resolve(String) / URI.create(String) with uildStorageUri() using the quoting 4-arg URI constructor in all 4 putFile helpers |
| core/.../ScriptService.java | Remove # from, add % to ALLOWED_PATH_CHARS; update error message |
| core/.../InternalStorageUriTest.java | New unit test — verifies #, %, and %xx sequences are properly percent-encoded |
| core/.../ScriptServiceTest.java | New tests — # rejected by alidateStoragePath, % accepted, percent-encoded URI matched by INTERNAL_STORAGE_PATTERN |

## Test results

All 18 tests pass (13 existing + 5 new).


core/src/test/java/io/kestra/core/runners/pebble/functions/HttpFunctionTest.java
```diff
@@ -36,16 +36,34 @@ class HttpFunctionTest {
     private VariableRenderer variableRenderer;
 
     @Test
-    void defaultHttpCall() throws IllegalVariableEvaluationException {
-        String rendered = variableRenderer.render("{{ http(url) }}", Map.of("url", "https://dummyjson.com/todos"));
+    void defaultHttpCall(WireMockRuntimeInfo wmRuntimeInfo) throws IllegalVariableEvaluationException {
```
sagarkhandagre998: I am not able to understand this tests, Probably HttpFunctionTest changes are out of scope . 

why did you switched the tests from external dummyjson.com to a local localhost/WireMock server ?

sagarkhandagre998 (changes requested):

core/src/test/java/io/kestra/core/runners/pebble/functions/HttpFunctionTest.java
KishoreGhost: Hey @sagarkhandagre998. You are correct, the HttpFunctionTest changes were out of scope for this PR. I originally switched them to a local WireMock server because the external dummyjson.com API was throwing 429 Too Many Requests / Connection reset errors in CI, which was causing the build to fail randomly.

However, you're totally right that it's completely unrelated to the URI bracket encoding fix. I have just pushed a commit to revert those test changes so we can keep this PR strictly focused on the immediate bug fix. Thanks for the review!

core/src/test/java/io/kestra/core/models/tasks/runners/ScriptServiceTest.java
loicmathieu, line 1: Can you remove most comments, they are useless

core/src/test/java/io/kestra/core/storages/InternalStorageUriTest.java
loicmathieu, line 1: It didn't seems to be worth it: this class is a regression test against a constructor for the Java library itself!
Please, remove it.

loicmathieu (changes requested):
Thanks a lot, some small feedback before I can merge it

KishoreGhost:
> Thanks a lot, some small feedback before I can merge it

Hey @loicmathieu, I have addressed the comments you gave. Kindly give it a look again.

MilosPaunovic:
If you've addressed any of the requested changes from the PR review, @KishoreGhost, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

core/src/test/java/io/kestra/core/models/tasks/runners/ScriptServiceTest.java
loicmathieu: I wanted to say: most comments you added.
As a rule of thomb, never remove comments you didn't wrote.

Can you revert removing existing comments?

loicmathieu (changes requested):

core/src/test/java/io/kestra/core/models/tasks/runners/ScriptServiceTest.java
```diff
@@ -166,9 +192,6 @@ void shouldReplaceInternalStorageWithSpecialChars() throws IOException {
         String tenant = IdUtils.create();
         var runContext = runContextFactory.of("id", "namespace", tenant);
 
-        // Colon (:) is also supported by the regex but can't be tested here:
```
sagarkhandagre998, line 169: Revert exactly at this place , line 169.

sagarkhandagre998 (changes requested):

loicmathieu (approved):

PR #18059, opened 2026-08-11
fix(executor): send WorkerTask event outside execution lock (#18006)

### LINKED ISSUE #18006: Investiguate sending WorkerTask event to the Worker outside of the execution lock
### Feature description

Today, WorkerTask events are sent to the Worker inside the lock.

In JDBC, as Micronaut binds a transaction to the thread, it means the WorkerTask are only sent after the execution has been updated, even if `toExecution()` is called outside the lock.

In non-transactional queue, they are sent immediately, which can lead to races if the Worker processes the taskrun quicker than the Executor commit the lock. With Redis/Elasticsearch this happens and lead to lock contention.

**Investigate** whether we can send the WorkerTask outside the execution lock to reduce lock contention.

Neeraj21216:
### ✨ Description

This PR defers the emission of `WorkerTask` events (`WorkerJobEvent`) to the queue until **after** the `executionStateStore.lock(...)` block has completed and released the database execution lock.

Previously, `workerJobEventQueue.emit(...)` was called inside `ExecutionEventMessageHandler.java` while holding the execution lock. On non-transactional queues (such as Redis, Kafka, In-Memory, or Elasticsearch), fast workers would receive and process the task immediately, returning `WorkerTaskRunning` or `WorkerTaskResult` events while the executor thread was still inside the lock callback. This caused severe lock contention, thread waiting, and race conditions where the task run state in the DB had not yet committed as `SUBMITTED`.

With this change:
- `ExecutorContext` accumulates `WorkerJobEvent` items during the lock block.
- `ExecutionEventMessageHandler` emits all pending `WorkerJobEvent` items post-lock after `executionStateStore.lock(...)` finishes and commits.

Why was the cat sitting on the computer? To keep an eye on the mouse... and to make sure there's no lock contention on the yarn thread! 🐱🐾

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18006

### 🛠️ Backend Checklist

### 📝 Additional Notes

- This change follows the same post-lock event emission pattern already used in `DefaultExecutor.java` for `executionDelayLoop` and `concurrencySlotReleaseProcessor`.


executor/src/main/java/io/kestra/executor/handler/ExecutionEventMessageHandler.java
```diff
@@ -404,6 +405,16 @@ public Optional<ExecutorContext> handle(ExecutionEvent message) {
                 }
             )
         );
+
+        executorContextOpt.ifPresent(executor -> {
```
sagarkhandagre998, line 409: `executorContextOpt` is never declare. The module does not compile. Capture the result first: `Optional<ExecutorContext> executorContextOpt = executionStateStore.lock(...);`

executor/src/main/java/io/kestra/executor/handler/ExecutionEventMessageHandler.java
```diff
@@ -404,6 +405,16 @@ public Optional<ExecutorContext> handle(ExecutionEvent message) {
...
+            if (!executor.getWorkerJobEvents().isEmpty()) {
+                executor.getWorkerJobEvents().forEach(throwConsumer(event -> {
+                    workerJobEventQueue.emit(event.workerQueueId(), event);
```
sagarkhandagre998, line 412: `workerJobEventQueue.emit` moved outside the `try/catch (QueueException)` block. If the emit fails, the events are lost and the task run stays SUBMITTED forever - `ExecutorService` only re-queues CREATED task runs `(ExecutorService.java:1101)`.

Wrap it in try/catch and fail the execution.

executor/src/main/java/io/kestra/executor/handler/ExecutionEventMessageHandler.java
```diff
@@ -284,10 +284,11 @@ public Optional<ExecutorContext> handle(ExecutionEvent message) {
...
                                                 if (workerTask.getTask() instanceof WorkingDirectory) {
                                                     // WorkingDirectory is a flowable so it will be moved to RUNNING a few lines under
-                                                    workerJobEventQueue.emit(workerQueueId, WorkerJobEvent.of(workerTask, workerQueueId));
+                                                    executor.withWorkerJobEvent(WorkerJobEvent.of(workerTask, workerQueueId));
                                                 } else {
                                                     TaskRun taskRun = workerTask.getTaskRun().withState(State.Type.SUBMITTED);
-                                                    workerJobEventQueue.emit(workerQueueId, WorkerJobEvent.of(workerTask.withTaskRun(taskRun), workerQueueId));
+                                                    // WorkerJobEvents are collected here and emitted outside of the execution lock to prevent race conditions on non-transactional queues.
+                                                    executor.withWorkerJobEvent(WorkerJobEvent.of(workerTask.withTaskRun(taskRun), workerQueueId));
```
sagarkhandagre998, line 291: With the event emitted after the lock, the WorkerJobEvent (task run still CREATED) is now sent after addWorkerTaskResults has persisted the flowable task run as RUNNING. Previously it was emitted first. Please confirm this ordering is safe for WorkingDirectory.

sagarkhandagre998 (changes requested):

loicmathieu:
Closing it for inactivity.

PR #18062, opened 2026-08-11
feat: Add full screen button for task logs

### LINKED ISSUE #17966: Add a full screen button for log of a task
### Feature description

Image

Now when a task output a lot of logs, you have only a 300px height to see all the logs. 
Add a button that will open the log in a modal as full screen / full height.

n200534:
## Description

Add a full-screen button to the task logs viewer that opens logs in a modal dialog, resolving the 300px height constraint that limits log visibility.

## Changes

- **Button Added**: New fullscreen button in logs toolbar (between LogDisplaySettings and Download)
- **Modal Implementation**: KsDialog with fullscreen attribute displays logs with full height/width
- **View Modes Supported**: 
  - Compact view: TaskRunDetails component
  - Raw view: Full-height DynamicScroller logs
- **Translations**: Added "fullscreen logs" key to all 12 language files

## Issue

Closes #17966

## Testing

- TypeScript compilation ✓
- Translation consistency check ✓
- Frontend build ✓
- ESLint checks ✓

## Files Modified

- `ui/src/components/executions/Logs.vue` - Added fullscreen button and modal
- `ui/src/translations/*.json` - Added translation keys for all languages

## Screenshots/Demo

The feature provides full-screen viewing of task logs by:
1. Clicking the fullscreen icon button in the logs toolbar
2. Modal opens with full-screen log display
3. All filters, searches, and log level selections remain functional
4. Close via Escape key or X button


n200534:
Thanks — I upgraded @hey-api/openapi-ts to 0.97.3 and updated package-lock.json in this branch; lint, type checks and translations passed. I have resolved the review thread.

n200534:
CI runs were created for this PR but no jobs started (check-suite created with zero check runs). This commonly happens for pull requests from forks when repository Actions settings or reusable-workflow restrictions block job execution until a maintainer approves the workflow run.

Could a repository maintainer please approve and re-run the affected workflow runs? Run IDs:
- Codespell: 31498801945
- Pull Request Workflow: 31498802722

Commands (requires maintainer/admin rights):
- Approve Codespell: gh api -X POST /repos/kestra-io/kestra/actions/runs/31498801945/approve
- Re-run Codespell: gh run rerun 31498801945 --repo kestra-io/kestra
- Approve PR Workflow: gh api -X POST /repos/kestra-io/kestra/actions/runs/31498802722/approve
- Re-run PR Workflow: gh run rerun 31498802722 --repo kestra-io/kestra

If approval cannot be granted, an alternative is to push this branch into the upstream repo (create branch 'n200534-add-fullscreen-task-logs' inside kestra-io/kestra or ask a maintainer to create a branch with these commits) so CI runs without fork restrictions.

Note: I attempted to re-run but do not have admin rights. Separately, I addressed the Plerion security finding by upgrading @hey-api/openapi-ts to 0.97.3 and updated package-lock.json; local lint/type/translation checks passed.

Thanks!

ui/package.json
```diff
@@ -39,7 +39,7 @@
         "@codecov/vite-plugin": "^2.0.1",
         "@esbuild-plugins/node-modules-polyfill": "^0.2.2",
         "@google/genai": "^2.11.0",
-        "@hey-api/openapi-ts": "0.97.0",
+        "@hey-api/openapi-ts": "^0.97.3",
```
flcarre: Two things here, both small.

`@hey-api/openapi-ts` is one of only 4 exact-pinned dependencies out of 93 in this file, and #18002 re-pinned it deliberately two days ago. It drives the generated SDK, so a caret range can let codegen output drift between installs. Worth keeping the bump but restoring the exact pin:

```suggestion
        "@hey-api/openapi-ts": "0.97.3",
```

Separately, `0.97.3` pulls a nested `js-yaml@4.1.1`, which is what Plerion's second comment is flagging (2 High + 1 Medium). `develop` already has the patched `js-yaml@4.3.1` at top level, so adding this to the `overrides` block that already exists in this file forces the nested copy up too:

```json
"js-yaml": "^4.3.1"
```

ui/src/components/executions/Logs.vue
```diff
@@ -93,6 +94,57 @@
                 </template>
             </DynamicScroller>
         </KsCard>
+
+        <KsDialog v-model="fullscreenModalOpen" :title="t('logs')" fullscreen @close="fullscreenModalOpen = false">
```
flcarre: `fullscreen` is not a declared `KsDialog` prop, it falls through `$attrs` to `ElDialog`. That gives you Element Plus's fullscreen box, but not a resolvable height: `.kel-dialog__body` has no height, flex or overflow rules, so the `height: 100%` further down resolves to `auto` and sizes to content. I measured the dialog class list as `kel-dialog is-fullscreen`, with no `is-fill`.

`KsDialog`'s own `fill` prop is what sets up the flex column plus `min-height: 0` cascade. Adding it (and dropping the redundant `@close`, since `v-model` already drives the state):

```suggestion
        <KsDialog v-model="fullscreenModalOpen" :title="t('logs')" fill fullscreen>
```

Please check this in the browser, `fill` + `fullscreen` is not combined anywhere else in the codebase yet. Note it is necessary but not sufficient: the 300px cap in compact view comes from `TaskRunDetails`' own scoped style and needs fixing there too.

Also worth moving the `.logs-toolbar` markup inside this dialog. As a sibling before it, it ends up behind the opaque fullscreen box, so search and the log level filter are not clickable while the modal is open.

ui/src/components/executions/Logs.vue
```diff
@@ -548,4 +606,15 @@
...
   }
+
+  .fullscreen-logs-container {
+    height: 100%;
+    display: flex;
+    flex-direction: column;
+  }
+
+  .fullscreen-logs {
+    max-height: none;
```
flcarre: In compact view this is a no-op: `.log-lines` has no `max-height` in this file, and the 300px cap lives in `TaskRunDetails`' scoped block where this rule cannot reach it.

In raw view it does apply, and it removes the only bound the scroller had. Measured with the modal open in raw view at 1600x1000:

```
.fullscreen-logs   offsetHeight: 4184   scrollHeight: 4184
.kel-dialog__body  height: 4200px   overflow: visible
.kel-dialog        height: 1000px
```

`offsetHeight == scrollHeight` means `DynamicScroller` stops windowing and renders everything into the DOM, and nothing scrolls: `window.scrollTo`, the dialog wrapper and the scroller all stayed at `scrollTop: 0`, with the last line at `top: 4182`, below the viewport. So most of the logs cannot be reached.

With `fill` on the dialog, this block should give the scroller a bounded flex height instead of removing the cap, something like `flex: 1; min-height: 0;` and no `height: 100%`.

flcarre (commented):
Thanks for taking this on, and welcome! Issue #17966 is a real pain point, and the direction you picked (a button in the logs toolbar that opens the logs in a modal) is exactly what was asked for.

A few things you did well and that I want to call out explicitly:

- You reused existing primitives instead of inventing new ones: `KsDialog`, `KsButton`, `KsNoData`, `TaskRunDetails`, `DynamicScroller`.
- The new icon button matches the neighbouring Download/Copy buttons exactly, including `square`, `type`, `:icon`, `:aria-label` and `:tooltip`. The `aria-label` works: my accessibility snapshot picked it up as `[button] "Full screen logs"`.
- The empty state reuses `KsNoData` the same way the existing raw-view branch does, a few lines above.
- You added the new i18n key to all 13 locale files in the same PR rather than leaving it for the translation job. That is the right instinct.
- You responded to the Plerion bot and self-corrected the translation nesting. Appreciated.

I built the branch and ran it locally (OSS from source, embedded H2, a flow producing two log entries of ~150 lines each) to check the behaviour rather than only reading the diff. Below is what I measured. There are a few blocking issues, but none of them are hard to fix, and the overall shape of the change is sound.

## 1. The 300px cap is still there in compact view (the default)

This is the constraint issue #17966 asks to remove, and it survives inside the fullscreen modal.

Measured inside the open modal, in compact view:

| Element | `max-height` | `offsetHeight` | `scrollHeight` |
|---|---|---|---|
| `.log-lines` (tall entry) | `300px` | **300** | 2220 |
| `.log-lines` (second entry) | `300px` | **300** | 1804 |

The cap comes from `TaskRunDetails.vue`:

```scss
.log-lines {
    transition: max-height 0.2s ease-out;
    max-height: 300px;
}
```

That rule lives in `TaskRunDetails`' own `scoped` style block, so the new `.fullscreen-logs { max-height: none }` in `Logs.vue` cannot reach it (different scope id), and it is only applied to the raw-view `DynamicScroller` anyway, not to `TaskRunDetails`. The modal does show more blocks at once because the page chrome is gone, but each log block is still clipped to 300px with an inner scrollbar.

`TaskRunDetails` already has the mechanism you need, right below the cap:

```scss
&.single-line {
    max-height: calc(100vh - 250px);
}
```

So the fix belongs in `TaskRunDetails` (a prop or class that opts into a full-height cap), not in parent CSS.

## 2. Raw view renders past the viewport with no way to scroll

In raw view the `max-height: none` override does apply, and that turns out to be worse than the cap. Measured with the modal open in raw view, viewport 1600x1000:

```
.fullscreen-logs        max-height: none   offsetHeight: 4184   scrollHeight: 4184
.fullscreen-logs-container                  height: 4184px
.kel-dialog__body                           height: 4200px   overflow: visible
.kel-dialog                                 height: 1000px
window.innerHeight                          1000
```

Two problems follow from this:

- **Most of the logs are unreachable.** I tried scrolling every plausible container to the bottom (`window.scrollTo`, the dialog wrapper, the scroller itself). All three stayed at `scrollTop: 0`, and the last rendered line sits at `top: 4182`, below the 1000px viewport. The dialog body is 4200px tall inside a 1000px dialog with `overflow: visible`, and nothing scrolls. The inline card at least scrolled inside its 300px box.
- **Virtualization is defeated.** `offsetHeight === scrollHeight` means `DynamicScroller` believes its viewport is 4184px tall, so it stops windowing. With ~300 lines that is merely wasteful. On a task with tens of thousands of lines, which is the case this feature exists for, it would render the whole list into the DOM.

The root cause is that `height: 100%` never resolves. `.kel-dialog__body` has no height, flex or overflow rules of its own, so a percentage height on a child falls back to `auto` and sizes to content.

`KsDialog` already solves exactly this with its `fill` prop:

```scss
&.is-fill {
    display: flex;
    flex-direction: column;
    max-height: calc(100vh - var(--kel-dialog-margin-top, 15vh) - var(--kel-dialog-fill-gutter));
    overflow: hidden;

    .kel-dialog__body {
        display: flex;
        flex-direction: column;
        flex: 1;
        min-height: 0;
        overflow: hidden;
    }
}
```

My measurement confirms it is not being used: the dialog's class list is `kel-dialog is-fullscreen`, with no `is-fill`. Note that `fullscreen` is not a declared `KsDialog` prop, it falls through `$attrs` to `ElDialog`, so you get Element Plus's fullscreen box without the design system's height cascade. Adding `fill` alongside it is necessary, though on its own it will not fix finding 1. Please verify the combination in the browser, since it is not used anywhere else in the codebase yet.

## 3. The toolbar is not reachable while the modal is open

The PR description says "All filters, searches, and log level selections remain functional". The toolbar is rendered as a sibling *before* `<KsDialog>`, and `is-fullscreen` gives an opaque box (`background-color: rgb(255, 255, 255)`) covering `0,0` to `1600x1000`. Hit-testing the toolbar's own coordinates with the modal open:

```
probe point (447, 208)  ->  element is inside .kel-dialog : true
                            element is inside .logs-toolbar : false
fullscreen button covered : true
```

So search, the log level filter, the level navigators, the expand/collapse control, the raw/compact toggle, and even the fullscreen button itself are all behind the overlay. Worth moving those controls inside the dialog, otherwise a user has to close the modal to change a filter and reopen it.

## 4. `npm ci` fails on this branch, so CI would fail at the install step

This one is quick to fix and worth doing first, because it blocks everything else from being verified:

```bash
cd ui && npm ci --dry-run
```

On `develop` that resolves cleanly (1441 packages). On this branch:

```
npm error code EUSAGE
npm error `npm ci` can only install packages when your package.json and package-lock.json are in sync.
npm error Missing: @emnapi/runtime@1.11.3 from lock file
npm error Missing: tree-sitter@0.22.4 from lock file
```

Regenerating the lockfile from a clean state should clear it.

## 5. The modal's component tree stays mounted after you close it

`KsDialog` does not unmount body content on close by default (`destroyOnClose` is not set). Measured after opening the modal once and pressing Escape:

```
modal still present in DOM : true
.log-lines total           : 8      (4 inline + 4 still inside the closed dialog)
```

So from the first open onwards there are two live `TaskRunDetails` trees for the same data, both reacting to every filter keystroke, log level change and SSE flush. The modal copy also drops `ref="logs"`, `@opened-taskruns-count` and `@log-indices-by-level`, so it keeps independent expand/collapse state and its counts never reach the toolbar. And `TaskRunDetails.onMounted` calls `OutputsAPI.taskRunOutputs(...)` for `Loop` task runs, so that request fires twice for flows using them.

Rather than duplicating the two branches, consider keeping a single `TaskRunDetails`/`DynamicScroller` instance and changing only where it renders, or extracting the `v-if="!raw_view"` / `v-else` pair into one shared child used by both the card and the dialog. As it stands every future change to the logs view has to be made twice, and the two copies have already drifted.

## 6. The dependency bump makes the security posture worse

I can see why you did this (the Plerion finding on `@hey-api/openapi-ts`), but the upgrade traded one Medium CVE for three more, which is what Plerion's second comment is reporting. `0.97.3` swapped its transitive `yaml` for `js-yaml`, and the resolved tree now contains:

```
node_modules/@hey-api/json-schema-ref-parser/node_modules/js-yaml  4.1.1   <- vulnerable
node_modules/js-yaml                                              4.3.1   <- already patched on develop
```

`develop` already carries the patched `js-yaml@4.3.1` at top level. Two suggestions:

- Add `"js-yaml": "^4.3.1"` to the `overrides` block that already exists in `ui/package.json`, which forces the nested copy up to the patched version.
- Keep the pin exact. `@hey-api/openapi-ts` is one of only 4 exact-pinned dependencies out of 93, and #18002 re-pinned it deliberately two days ago. It drives the generated SDK, so a caret can let codegen output drift.

Cleanest of all would be to split the dependency bump into its own PR, so it can be reviewed and reverted independently of the logs feature.

## 7. The 12 non-English locales received the English string

Every non-English file got `"fullscreen logs": "Full screen logs"` verbatim. The translation job skips keys that already have a value, so these would stay frozen in English rather than being picked up later. Note that `log` is on the reserved-terms list and should not be translated. Suggested values, worth a native-speaker check:

| Locale | Value |
|---|---|
| de | `Logs im Vollbild` |
| es | `Logs a pantalla completa` |
| fr | `Logs en plein écran` |
| hi | `Logs पूर्ण स्क्रीन में` |
| it | `Logs a schermo intero` |
| ja | `Logsを全画面表示` |
| ko | `Logs 전체 화면으로 보기` |
| pl | `Logs na pełnym ekranie` |
| pt | `Logs em ecrã inteiro` |
| pt_BR | `Logs em tela cheia` |
| ru | `Логи на весь экран` |
| zh_CN | `全屏日志` |

## Smaller things

- `ui/src/override/components/useLeftMenu.ts:382`: an unrelated trailing-comma edit, probably format-on-save. Worth dropping from the PR.
- `Logs.vue:98`: `@close="fullscreenModalOpen = false"` is redundant, `v-model` already drives the state.
- `Logs.vue` no longer ends with a newline.
- No test was added. Not a hard requirement here since `Logs.vue` has none today, but a small test asserting the button opens the dialog would help.
- The `as any` casts in the new block are copies of casts already in this file, so no new type debt. Not a concern.

## One note on CI

CI has not actually run on this PR. Only `Plerion Code Security` completed; `Codespell` and `Pull Request Workflow` are both in `action_required`, which is the fork-approval gate you flagged. So the lint, type check, translation check and build results in the PR description are local-only for now. Getting `npm ci` green (finding 4) first will make that approval worth spending.

## Where this leaves us

The idea and the component choices are right, and findings 4, 6, 7 and the smaller items are mechanical. The substantive work is findings 1 to 3, and they are really one question: what owns the height inside the dialog. If `TaskRunDetails` gains a full-height mode and the dialog gets a resolvable height via `fill`, with the toolbar moved inside, the feature does what the issue asks. Happy to look again once you have pushed, and happy to go into more detail on any of the above.

ui/src/override/components/useLeftMenu.ts
```diff
@@ -379,7 +379,7 @@ export function useLeftMenu() {
                         },
                         icon: {
                             element: Battery40,
-                        }
+                        },
```
flcarre, line 379: Unrelated to the logs feature, looks like format-on-save. Worth dropping so the diff stays focused:

```suggestion
                        }
```

MilosPaunovic:
If you've addressed any of the requested changes from the PR review, @n200534, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

n200534:
@flcarre Thanks for the detailed review. I’ve addressed the fullscreen sizing and scrolling behavior, kept a single log viewer using Teleport, moved the controls into the fullscreen dialog, added Storybook coverage, and synchronized the frontend lockfile. All CI checks are now passing and the review conversations have been resolved. Ready for re-review.

ui/src/components/executions/Logs.vue
n200534: Addressed. The fullscreen container and scroller now use `flex: 1` with `min-height: 0`, and the raw `DynamicScroller` retains a bounded `max-height` (`calc(100vh - 250px)` in fullscreen). This preserves virtualization and scrolling without using `height: 100%`.

ui/src/override/components/useLeftMenu.ts
n200534: Addressed. I removed the unrelated EOF formatting change from `useLeftMenu.ts`, so the file no longer appears in the PR diff.

n200534:
@MilosPaunovic Please assign someone to review this PR i made the changes in the code

MilosPaunovic:
Hi @n200534,

flcarre (commented):
Thanks for taking this on. A full-screen view for task logs is genuinely useful, and putting the toggle in the existing toolbar next to Download and Copy is the right home for it.

Two things need sorting before this can go green.

### Unrelated dependency churn

`ui/package.json` and `ui/package-lock.json` carry an `@hey-api/openapi-ts` bump from 0.97.0 to 0.97.3, a `yaml` to `js-yaml` swap, and peer-flag rewrites across the `sass-embedded-*` and `@emnapi/*` entries. None of that is needed for this feature.

This is almost certainly a local npm rewriting the lockfile rather than anything you did on purpose. `develop` just landed `e3b71ffc1e` with a `ui/.npmrc` specifically to stop it happening. Rebasing on develop to pick that up, then restoring both files from develop, should clear it.

### Translations will fail the PR gate

You added the key to `en.json` *and* to all twelve locale files, which is the right instinct. The missing piece is `ui/scripts/translations/fingerprints.json`, which was not regenerated and has no entry for `fullscreen logs`. The check treats a missing fingerprint the same as a drifted one, so it reports the key as stale in every locale. `npm run translations:generate` from `ui/` fixes that and the key ordering in one go.

### A design question worth raising before you invest more

There is already a maximize mechanism in `MultiPanelTabs.vue`: `maximizedPanelIndex`, `Fullscreen` / `FullscreenExit` icons, its own Escape handling, and `multi_panel_editor.fullscreen` / `.exit_fullscreen` keys. This PR builds a second, independent one on `KsDialog` plus `Teleport`, and then half-reuses the old namespace, with the exit label reading `multi_panel_editor.exit_fullscreen` while the enter label is a new root-level `fullscreen logs`. One button spanning two namespaces is the part that will confuse the next reader.

Worth deciding with the team whether to reuse the existing mechanism or commit to the dialog approach and give it its own key pair. Not something you should have to guess at on your own.

### Smaller things

* `fullscreen` is not a declared prop on `KsDialog`; it only reaches `ElDialog` through `$attrs`. It also collides with `fill`: `.is-fill` sets a `max-height` and `overflow: hidden` that win at equal specificity, so the result is a roughly 98vh dialog with clipped, non-scrolling content rather than true full screen.
* The Storybook `play` selects on `.kel-dialog.is-fullscreen`, `.logs-toolbar` and `.fullscreen-logs`. Those are Element Plus internals and feature CSS names, which break on the next upgrade with no behaviour change. `data-test` attributes are the stable contract.
* Removing the `Array.isArray(executionsStore.logs)` guard enables a `KsNoData` branch that was previously unreachable, since `logs` is an object and the check was always false. Nothing gates it on `logsLoading`, so "No logs data" now flashes on every raw-view load before results arrive.

The feature itself is worth having, so none of the above is a reason to drop it.

n200534:
Thanks for the detailed review. Logs.vue is also rendered outside MultiPanelTabs, including the execution detail route and Flow Playground, so reusing MultiPanelTabs directly would not cover every logs context. My proposal is to keep the fullscreen behavior local to Logs.vue, give it a dedicated enter/exit translation-key pair, and correct the KsDialog sizing and scrolling behavior. Does that direction fit what you had in mind?

flcarre (commented):
Round 2. **All four findings from my previous round are fixed**, and the toolbar one is solved better than what I suggested.

- `KsDialog` now carries `fill` **and** `fullscreen`, and the redundant `@close` is gone.
- Rather than moving or duplicating the `.logs-toolbar` markup, `Teleport` moves the whole filter + toolbar + logs subtree between an inline target and the dialog target. One instance, so the `TaskRunDetails` component is preserved across the toggle and the expanded task runs and scroll position survive it. Better than the duplication my suggestion would have led to.
- The scroller is bounded by `flex: 1; min-height: 0` instead of `height: 100%`, so `DynamicScroller` keeps windowing.
- The `TaskRunDetails` 300px cap — the "necessary but not sufficient" half — is handled by the new `fullHeight` prop rather than left behind.
- `@hey-api/openapi-ts` is back to an exact pin, and `useLeftMenu.ts` is out of the diff.

One thing worth stating so it does not read as unfinished: the lockfile shows `@hey-api/json-schema-ref-parser@1.4.2` declaring `js-yaml: 4.1.1`, which is the package's own manifest and not something an override rewrites. There is no nested `node_modules/@hey-api/json-schema-ref-parser/node_modules/js-yaml` entry, so it resolves against the hoisted top-level copy, which the new `"js-yaml": "^4.3.1"` override pins. The three CVEs Plerion flagged are genuinely closed.

## One finding

**The `KsNoData` guard change makes the raw-view empty state flash during the initial load.** (`ui/src/components/executions/Logs.vue`)

```diff
-  v-if="Array.isArray((executionsStore.logs as any)) && temporalLogs.length === 0"
+  v-if="temporalLogs.length === 0"
```

Dropping the `Array.isArray` half is right, and it fixes more than you may have intended: the store initialises `logs` as `{total: 0, results: []}` (`stores/executions.ts:113`), so `Array.isArray(executionsStore.logs)` is **always false** in the app. The plain-array shape only comes from tests, which `temporalLogs`' own comment says explicitly. That empty state has therefore never rendered in production — the change is what makes it reachable at all.

The consequence is that `temporalLogs` is also `[]` before the first response lands, so raw view now shows "No logs data" while the logs are still loading, then swaps to content. `ui/AGENTS.md` asks every async surface to render loading separately from empty.

`logsLoading` already exists and is correctly maintained (set at `:411`, cleared at `:288` and `:416`) but is never read in the template, so it is exactly the signal this needs:

```
v-if="!logsLoading && temporalLogs.length === 0"
```

## Nit

`calc(100vh - 250px)` is duplicated in the `Logs.vue` inline style and in `TaskRunDetails.vue`'s scoped `.log-lines.full-height`. Changing the dialog chrome means remembering both. A shared custom property set on the container would keep them in step — not worth holding the PR for.

Nothing else from me. The `Fullscreen` story asserting the toolbar is inside the dialog and that `aria-pressed` round-trips is the right shape of test for this.

---

n200534:
Implemented the requested review changes and updated the existing PR branch.

Changes include:
- made the execution logs fullscreen toggle work reliably using the existing dialog/teleport architecture
- ensured task-run log details use the available fullscreen height
- corrected loading and empty-state behavior
- added translated fullscreen labels across all locales
- added Storybook coverage for entering and exiting fullscreen
- rebased onto the latest develop branch and resolved the PR conflict

Validation completed locally: TypeScript checks, lint, translation checks, production build, Storybook build, and a direct Chromium fullscreen-toggle check all pass.

The PR is now mergeable and CI is running. Could you please review it again when convenient?

flcarre (commented):
Second iteration reviewed — every one of the four inline comments from the previous round is addressed, the unrelated `useLeftMenu.ts` formatting hunk is gone, and CI is green. Nice, careful work.

To record what I checked so it isn't re-litigated: the `ui/package.json` change is exactly two lines — the `@hey-api/openapi-ts` `0.97.0 → 0.97.3` patch bump and `js-yaml` added to the existing **`overrides`** block. That's **no new runtime dependency and no bundle-size impact**; the lockfile is net −15 lines and actually removes a nested `yaml` copy. I verified every lockfile hunk: only the `@hey-api/*` chain moved, registry and integrity hashes are well-formed, no transitive additions.

There's also no design-system duplication. `MultiPanelTabs` has a panel-maximize, but `Logs.vue` is a route tab and a playground pill tab, never mounted inside it — so mirroring the `MultiPanelTabs` a11y pattern rather than reusing a component that doesn't apply is the right call. The icon-only toggle carries `aria-label`, `:tooltip` (the prop, not a wrapper) and `aria-pressed`, and all three flip with state. `Escape` and focus are left to `KsDialog` rather than reimplemented. `data-test` hooks throughout. Zero design-system violations across the 210 added lines in `ui/src`.

Four things I'd like resolved.

## 1. Scroll position is lost on every open and close

`<Teleport :to="logsTarget">` changes its target on toggle, so Vue re-inserts the whole subtree into the dialog. A scrollable element that is detached and re-attached has its `scrollTop` reset by the browser, and `DynamicScroller` then re-windows from index 0.

The user opens fullscreen precisely because they're deep in a long log, so landing back at the top is the one outcome the feature exists to avoid. Could you measure it the way you measured the last round — scroll to the bottom of a ~5k-line execution, toggle fullscreen, read `scrollTop`? If it reproduces, capturing the scroller's `scrollTop` before the toggle and restoring it on `nextTick` is the smallest fix.

There's a larger alternative worth weighing: toggling a `position: fixed; inset: 0` class on the existing root gives the same visual modal, moves no DOM at all, and sidesteps both this and item 3 — at the cost of wiring `Escape` and a focus trap yourself, which `KsDialog` currently gives you free. Genuinely a trade-off; if the scroll measures clean, leave it as is.

## 2. `max-height: calc(100vh - 250px)` can clip the tail of the log

`KsDialog`'s `is-fill` sets `overflow: hidden` on the body, so anything the 250px budget under-estimates is unreachable with nothing to scroll it — the same failure class as the `height: 100%` bug from the last round, at a different breakpoint. That budget has to cover the dialog header (62px) **plus** `KSFilter` **plus** `.logs-toolbar`, and both of the latter are `flex-wrap: wrap`: narrow the window to ~900px with three or four filters applied and they take two rows each.

Letting the flex chain do the work removes the guess — `display: flex; flex-direction: column; flex: 1; min-height: 0` on `.attempt-wrapper` under fullscreen, then `flex: 1; min-height: 0` on the scroller with no `max-height`. That's also what `.fullscreen-logs` was clearly meant to do: it currently applies `flex: 1; min-height: 0` to the scroller whose parent is `KsCard.attempt-wrapper`, which isn't a flex container, so that rule is inert today.

## 3. `fill` and `fullscreen` fight each other in the cascade

`KsDialog`'s `.kel-dialog.is-fill` declares `max-height: calc(100vh - var(--kel-dialog-margin-top, 15vh) - 2vh)` and `margin-bottom: 2vh`; Element Plus's `.is-fullscreen` sets `margin-top: 0; margin-bottom: 0; height: 100%`. Same specificity, and `is-fill` is declared later, so it wins — the "fullscreen" dialog resolves to about 98vh with a 2vh gap at the bottom and none at the top.

You need `fill` (it's what provides the flex body), so this isn't fixable at the call site. It belongs upstream: guard `is-fill`'s `max-height`/`margin-bottom` with `&:not(.is-fullscreen)` in `KsDialog.vue`, plus a `fill + fullscreen` story. The combination is used nowhere else in the repo, which is why nobody has hit it — worth raising with the design-system owners rather than absorbing here.

## 4. "No logs data" now flashes on every running execution

The guard changed from `Array.isArray(executionsStore.logs) && …` to `!logsLoading && temporalLogs.length === 0`. The old condition was effectively always false in the app (the store holds `{total, results}`, not an array), so this empty state never rendered outside tests. Making it real is correct per the four-states rule — but `streamLogs()` never sets `logsLoading`, so a running execution shows "No logs data" from mount until the first SSE batch clears the 200ms buffer. On a slow-starting task that's seconds of a wrong empty state. Setting `logsLoading.value = true` at the top of `streamLogs()` and clearing it on the first message covers it.

## Test coverage

The `Fullscreen` story's decorator sets `LOGS_VIEW_TYPE = "true"`, so it only exercises the **raw** `DynamicScroller` branch. Compact view — the default for real users, and the branch the new `fullHeight` prop exists for — is never rendered in fullscreen by any test. Given the 300px cap in compact view was the actual bug in the linked issue, that's the path most worth pinning: a second story with `LOGS_VIEW_TYPE = "false"` asserting the taskrun log block is no longer capped inside the dialog. A dark-mode variant would be good too.

flcarre (changes requested):
Formalising the gate from my detailed review above. To be clear on what is *not* at issue: the dependency changes are correct (a devDependency patch bump plus an `overrides` entry, no new runtime dependency, lockfile net −15 lines), there's no design-system duplication, and the a11y on the toggle is right. Four things remain:

**1. Scroll position is lost on every toggle.** `<Teleport :to="logsTarget">` changes target, so Vue re-inserts the subtree and the browser resets `scrollTop`; `DynamicScroller` then re-windows from index 0. Since the whole point of fullscreen is being deep in a long log, please measure this (scroll to the bottom of a ~5k-line execution, toggle, read `scrollTop`) and restore the offset on `nextTick` if it reproduces.

**2. `max-height: calc(100vh - 250px)` can clip the tail.** `is-fill` sets `overflow: hidden` on the dialog body, so whatever the budget under-estimates is unreachable — and it has to cover the header plus `KSFilter` plus `.logs-toolbar`, the latter two both `flex-wrap: wrap`. Letting the flex chain size it removes the guess, which is also what the currently-inert `.fullscreen-logs` rule was reaching for.

**3. "No logs data" now flashes on every running execution.** Making that state real is correct, but `streamLogs()` never sets `logsLoading`, so it shows from mount until the first SSE batch. Set it at the top of `streamLogs()` and clear it on the first message.

**4. Compact view is untested in fullscreen.** The `Fullscreen` story sets `LOGS_VIEW_TYPE = "true"`, so only the raw `DynamicScroller` branch is exercised — while compact view is the default and is the branch the new `fullHeight` prop exists for, and the 300px cap there was the original bug. A second story with `LOGS_VIEW_TYPE = "false"` would pin it.

Not blocking on you: the `fill` + `fullscreen` cascade conflict I described belongs upstream in `KsDialog`, since `is-fill`'s `max-height` wins over `.is-fullscreen` at equal specificity. Worth raising with the design-system owners as a separate change.

n200534:
Thanks for the detailed review. I addressed all four blocking points in commit 89d89e7cf:

- Preserved raw and compact log scroll positions across fullscreen open and close, including Escape, by capturing the active scrollers and restoring them after the Teleport move.
- Removed the fullscreen calc-based height budget and completed the flex sizing chain through the card body and raw scroller, keeping the virtualized viewport bounded.
- Set logsLoading while the raw SSE stream starts and clear it when the first real buffered batch is committed, or when the stream closes or fails, so the empty state no longer flashes.
- Added compact-view fullscreen Storybook coverage and stable scroller selectors; the existing fullscreen story now also exercises 5,000 logs and verifies scroll preservation.

Validation completed: TypeScript checks, targeted ESLint and Oxlint, production UI build, diff checks, and manual Storybook verification for raw and compact fullscreen behavior. Raw scroll offset was preserved exactly through Escape; compact view stayed within normal virtual-scroller adjustment.

Could you please take another look when convenient?

flcarre (changes requested):
Three of the four blocking points from my last review are genuinely fixed — scroll preservation, the loading-flash fix, and stable selectors. The fourth is only half done.

**Still blocking:** the calc-based height budget (`max-height: calc(100vh - 250px)`) is untouched in `TaskRunDetails.vue` — that's the compact/default view, and the one where the original linked-issue bug actually lives. Only the raw-log view got the flex-chain treatment; your reply's "completed the flex sizing chain through the card body and raw scroller" is accurate but doesn't mention the compact path is still on the old budget.

**Test-quality issue:** the new `CompactFullscreen` story asserts `maxHeight !== "300px"` — but the height has been `calc(100vh - 250px)` since round 2, never `300px`, so this assertion would pass on the exact code my last review flagged as broken. It doesn't test what it claims to.

Smaller: scroll-restore timing is a heuristic double-restore (rAF + `setTimeout(0)`) rather than being driven by a scroller-ready signal — works in the tested case, could still flake under slower rendering. And neither scroll preservation nor a literal Escape keypress is exercised for the compact view.

Once the compact view actually gets the same sizing treatment as the raw view, and the story asserts something that can fail for the right reason, this is good to go — everything else (dependency bump, i18n, no design-system violations) checks out clean.

MilosPaunovic:
Hey @n200534, are there any updates on this?

n200534:
The compact/default fullscreen path is now bounded by the actual available log viewport. Because the virtual task rows are positioned independently of the flex layout, each task card uses the measured outer viewport height, and its body and log scroller allocate the remaining space with flex sizing. The fullscreen `calc(100vh - 250px)` budget has been removed.

Scroll restoration now uses stable task-run keys and the dialog/scroller lifecycle events, including the dialog transition completing, instead of the rAF + timeout double restore.

The compact Storybook cases cover multiple task runs, a single task run, light/dark themes, button close/reopen, a literal Escape keypress, preserved scroll offsets, and reaching the final log entry with thousands of logs. The test checks viewport bounds: temporarily restoring the previous layout made it fail because the scroller extended to 740px in a 720px viewport.

Validation: all 7 log Storybook cases and 49 related unit tests passed; type checking, targeted lint, and the production build passed. Local Storybook validation needed a temporary dependency-prebundling override for the existing `aria-query` CommonJS import error; no dependency or lockfile changes are included in this update.

Rebased onto the current `develop` to clear the branch conflicts and pushed commit `3ccbf2764` on the same branch. Please take another look at the compact sizing and scroll restoration changes.

flcarre (changes requested):
Round 5. **Both blocking points from my last review are genuinely fixed**, and the fix for the second one is better than what I asked for.

- The compact path no longer carries the `calc(100vh - 250px)` budget. `.log-wrapper .log-lines.full-height` sets `max-height: none`, and since it sits after the nested `.single-line` rule at equal specificity it actually wins — I checked that rather than assuming it.
- The `CompactFullscreen` story now asserts `getComputedStyle(scroller).maxHeight === "none"` **and** that the scroller's bottom stays inside the dialog bounds, with 1700 entries per task run in a 900x720 viewport. That fails on the code I flagged last round, which is exactly what was missing before. The dark variant and the single-task variant are a good addition, and the `queryAllByText(...).length < 300` check pins virtualization.
- Scroll restoration is now driven by the dialog lifecycle and scroller events instead of the rAF + `setTimeout(0)` pair, and a literal `{Escape}` keypress is exercised.

No design-system violations across the change: no `element-plus` import, no new `:deep()`, no hex/`rgb()`/`--el-*`, `data-test` hooks throughout, `aria-label` + `aria-pressed` + `:tooltip` all flipping with state, and Escape/focus left to `KsDialog`. The i18n is complete — both keys under `logs_view.*` in all thirteen files, with `log`/`Logs` correctly left untranslated per the reserved-terms rule, and `fingerprints.json` regenerated.

## Blocking

### 1. CI is red because of this PR, and the fix is one line

`Npm - check ts-any` fails:

```
explicit any went from 6 to 7; type the value instead of raising the baseline
  tests/storybook/components/executions/Logs.stories.tsx: 6 -> 7
```

The new one is in the decorator:

```ts
executionsStore.flow = {tasks: execution.taskRunList.map(task => ({id: task.taskId, type: "io.kestra.plugin.core.log.Log"}))} as any;
```

`ui/AGENTS.md` is explicit that the baseline is never the thing to move. `Flow` comes from `@kestra-io/kestra-sdk`, so typing the literal clears this without touching `baseline.json`. Everything else in the job passed, so this single line is what is holding the PR red.

### 2. The measured height falls back to `0px`

`ui/src/components/logs/TaskRunDetails.vue`:

```scss
.fullscreen-attempt { max-height: var(--task-run-viewport-height, 0px); }
```

The class is bound to `fullHeight`, but the variable is only written once `taskRunViewportHeight` is truthy, and it starts at `0` and is first written by the `useResizeObserver` callback. So between `fullHeight` flipping true and the observer's first delivery, every task card resolves to `max-height: 0px`. The same gap reopens right after the `Teleport` moves the scroller into the dialog.

I would not hold the PR on how visible that is — the dialog's own transition covers part of it — but `0px` is the wrong fallback for a value that is only ever a cap. `none` makes the uninitialised state degrade to today's behaviour instead of to a collapsed card, and costs nothing.

I do think the measured-variable approach itself is justified here, and I want to say so explicitly since it is the third sizing strategy this PR has been through: the task rows are absolutely positioned by the virtual scroller, so a pure flex chain genuinely cannot size them. With the bounds assertion in the story backing it, I am satisfied this one is settled.

### 3. A pending scroll key can outlive its scroller

`ui/src/components/executions/Logs.vue`. A key is only deleted from `pendingLogScrollPositions` if its `[data-scroll-key]` element is found in the current target. In compact view the task rows are virtualized, so a task run the user scrolled and then scrolled past is not in the DOM when fullscreen opens. Its key stays pending indefinitely, and three things follow:

- `rememberLogScroll` has `!pendingLogScrollPositions.has(key)` as a guard, so that task run's scroll position is **never recorded again** for the life of the execution.
- `pendingLogScrollPositions.size` stays non-zero, so every `@update` / `@resize` / `@visible` from any of the three scrollers keeps scheduling a `nextTick` and a `querySelectorAll` that can do nothing useful.
- When the task run is eventually recycled back into view, `restoreLogScroll` finds it and assigns `scrollTop` from an offset captured before the toggle — yanking the user's scroll at an arbitrary later moment.

`finishLogScrollRestore` is the natural place to bound this: once the dialog has finished opening or closing, any key that could be restored has been, so clearing the whole map there rather than per-matched-key removes the class of bug. Worth reproducing first the way you measured the last round — scroll task 1, scroll it out of the virtual window, toggle fullscreen, scroll it back.

## Non-blocking

- **Five triggers now drive one restore**: a `flush: "sync"` watch on `fullscreenModalOpen`, a `flush: "post"` watch on `logsTarget`, the dialog's `@opened`/`@closed`, and `@update`/`@resize`/`@visible` on three scrollers. This is a real improvement on the rAF + timeout pair, but it is now hard to say what the terminating condition is by reading it. If the pending map gets cleared on the lifecycle events per the point above, most of these become redundant.
- **The error state is still missing.** `sse.onerror` calls `closeLogsSSE()`, which clears `logsLoading` and leaves `logsLoaded` false — so a log stream that fails renders nothing at all: no logs, no empty state, no error, no retry. The loading half you added is right; the fourth state is the one still open. This predates your PR, so I am not asking you to fix it here, but it is worth an issue.
- **`waitFor` callbacks that mutate what they assert on.** In `compactFullscreenPlay` the final block sets `lastScroller.scrollTop` and `taskScroller.scrollTop` *inside* the polled callback. The retry loop is producing the state it then checks, which makes a pass weaker evidence than it looks and is a classic source of slow flake. Drive the scroll once before the `waitFor`, then assert only.
- **Nits:** `:bodyStyle` in `TaskRunDetails` is rebuilt per virtualized row on every render, unlike `FULLSCREEN_CARD_BODY_STYLE` which you correctly hoisted in `Logs.vue`. And there are now four near-identical `flex: 1; min-height: 0` blocks across the two files.

## Still not yours

The `fill` + `fullscreen` cascade conflict I described in an earlier round is unchanged and still belongs upstream — `KsDialog`'s `.is-fill` beats Element Plus's `.is-fullscreen` at equal specificity by source order, so the dialog resolves to roughly 98vh with a gap at the bottom. You need `fill`, so there is nothing to do at the call site. I will raise it with the design-system owners separately.

---

Get `check:ts-any` green, put `none` behind that variable, and bound the pending map, and I am happy with this. The sizing question that has taken four rounds is answered, and the test now proves it.

n200534:
Final follow-up: rebased again onto current `develop` to clear the newly introduced conflict and fixed the frontend CI failure by removing the added explicit `any` from the Storybook fixture.

The rebase also exposed a virtual-scroller readiness race: a temporary, smaller scroll range could be treated as a completed restoration. Restoration now completes only when the original offset is actually available, and the inner scroller's resize event provides the readiness retry. The compact tests keep the scroll-preservation assertion away from the bottom boundary; reaching the final log entry is validated separately.

n200534:
Addressed the remaining Round 5 points in `bed78d64a`:

- Changed the fullscreen task-card fallback to `none`. To avoid an unbounded virtual scroller before the first ResizeObserver delivery, the CSS variable is seeded synchronously from the existing scroller height before Teleport, then replaced by the measured dialog viewport.
- Bounded `pendingLogScrollPositions` at dialog lifecycle completion. Cleanup waits until the currently rendered scrollers report a restorable range, restores those rows, and then clears the complete map so recycled/absent task keys cannot survive the transition.
- Added a focused regression with ten task runs: task 1 is scrolled, recycled out of view, fullscreen is toggled, and returning to task 1 verifies that its obsolete offset is not reapplied.
- Moved all scroll-driving operations outside `waitFor`; its final callbacks now only query and assert. The frame-driven helper accounts for virtual-scroller measurement updates without mutating assertion polling.
- Hoisted the fullscreen card body style and consolidated the duplicate full-height flex declaration.

Validation after these changes: 8/8 log Storybook cases passed, including the focused virtualized-task case; 49/49 related unit tests passed; type checks, targeted lint, and the production build passed. The explicit-`any` count remains at the existing baseline (6).

MilosPaunovic:
Can you resolve the conflicts here @n200534?

n200534:
Conflicts are resolved. I rebased the branch onto the latest `develop` (`95261e6fd`) and force-pushed the rebased history with lease protection.

While resolving `TaskRunDetails.vue`, I preserved the newer typed `DynamicScroller` template ref from `develop` and combined it with the compact fullscreen viewport measurement/resize behavior from this PR. I also kept the OpenAPI generator bump in its current `devDependencies` location.

Validated locally after the rebase:

- `npm run check:types`
- `npm run check:ts-any`
- focused unit suite: 49/49 passed
- fullscreen logs Storybook suite: 8/8 passed
- `npm run translations:check`
- `npm run test:lint`
- `npm run build`

GitHub now reports the PR as mergeable; the new CI run is in progress.

flcarre (approved):
Round 6. **All three blocking points from my last review are fixed**, and I checked each rather than taking the reply for it.

- `check:ts-any` is green — the decorator now types the literal as `typeof executionsStore.flow` instead of `as any`, and `baseline.json` is untouched. Every step of `Frontend - Tests` passes, including `check ts-any` and the Storybook run.
- `.fullscreen-attempt` falls back to `max-height: none`, and the `flush: "sync"` watch on `props.fullHeight` seeds `--task-run-viewport-height` from the live scroller before the Teleport, so the gap before the first `ResizeObserver` delivery degrades to today's behaviour rather than to a collapsed card.
- The pending scroll map is bounded at dialog lifecycle: `finishLogScrollRestore` arms `clearPendingAfterRestore`, and `applyPendingLogScrollPositions` clears the whole map once every scroller it can reach is settled — so a key belonging to a task run that was virtualized out no longer survives the toggle. `CompactFullscreenVirtualizedTasks` pins exactly that: scroll task 1, scroll it out of the window, toggle, scroll it back, assert the user's later scroll is not yanked. That is the right reproduction.

Two of the non-blocking notes were taken as well: `FULLSCREEN_CARD_BODY_STYLE` is hoisted in `TaskRunDetails.vue` so it is no longer rebuilt per virtualized row, and `compactFullscreenPlay` drives the scroll before the `waitFor` instead of inside the polled callback.

**On the age of this PR:** it was rebased onto `95261e6f` yesterday, and `develop` has not touched `Logs.vue`, `TaskRunDetails.vue` or `Logs.stories.tsx` since. The only `ui/` overlap risk — `scripts/explicit-any/baseline.json`, which `develop` trimmed by two entries — is not a file this PR modifies. No rebase needed; the green CI still means what it says.

No design-system violations across the change: no `element-plus` import in `ui/src`, no new `:deep()`, no hex/`rgb()`/`--el-*`, `data-test` hooks throughout, and `:bodyStyle` reaching `ElCard` through `KsCard`'s `$attrs` passthrough is the extension point `ui/AGENTS.md` explicitly asks wrappers to leave open.

Approving. The rest is small and none of it needs to hold the merge.

## Non-blocking

- **The label does not match the fullscreen control we already ship.** `multi_panel_editor.fullscreen` / `exit_fullscreen` read "Fullscreen" / "Exit fullscreen" and use the same `Fullscreen` / `FullscreenExit` icons with the same `aria-pressed`. The new pair reads "Full screen logs" / "Exit fullscreen" — so the enter label differs while the exit label is identical, for two controls that are visible at the same time in the flow editor with the playground open. Matching `logs_view.fullscreen` to "Fullscreen" would settle it; one line per locale.
- **`:emitUpdate="true"` is unconditional** on both scrollers in `TaskRunDetails.vue`. It costs an emit per animation frame on every log scroll for every user, to serve machinery that only does work during a fullscreen transition — `restoreLogScroll` returns immediately when nothing is pending, so the cost is small, but it is paid on the default path. If you narrow it, check the close path still restores, since `@closed` is what drives that half.
- **Duplication.** Four near-identical `flex: 1; min-height: 0` blocks across the two stylesheets and a verbatim copy of `FULLSCREEN_CARD_BODY_STYLE` in each file. Fine as is; worth a shared mixin if this grows.
- **Five triggers still drive one restore.** Bounding the map on the dialog lifecycle was supposed to make most of them redundant, and it does — the `flush: "post"` watch on `logsTarget` and some of the scroller events are now belt-and-braces. Not worth churning the file again for.

## Not yours

- **`fill` + `fullscreen` in `KsDialog`.** Unchanged on `develop`: `.is-fill` declares `max-height: calc(100vh - var(--kel-dialog-margin-top, 15vh) - 2vh)` and `margin-bottom: 2vh` with no `:not(.is-fullscreen)` guard, and wins over Element Plus's `.is-fullscreen` by source order at equal specificity. So this dialog is about 98vh with a gap at the bottom. `fullscreen` is also not a declared prop — it only reaches `ElDialog` via `$attrs`, so nothing types or stories the combination. I will raise both with the design-system owners.
- **The error state.** `sse.onerror` calls `closeLogsSSE()`, which clears `logsLoading` and leaves `logsLoaded` false, so a failed log stream renders nothing: no logs, no empty state, no error, no retry. Predates this PR; the loading half you added is the part that was in scope. I will file it.

Thanks for the persistence through six rounds — the sizing question took a while to land, and the measured-viewport approach plus the bounds assertion in the story is the right answer to it.

PR #18068, opened 2026-08-11
fix(cli): use logger instead of stdout for the startup ready message

### LINKED ISSUE #17946: Replace StdOut startup message but logger message
### Describe the issue

When Kestra start, we display this message: 
```
14:20:24.207 INFO  grpc-default-executor-0 i.k.c.g.services.WorkerJobDispatcher Resumed subscription for Worker Queue 'default'

✅ Kestra is ready! Open the UI at: https://kestra.ee:8443
```
We should never output log message without logger, it will break all log parser that people applied, replace by a real `logger.info`

### Environment

- Kestra Version: develop

aetherloop:
## What

The `✅ Kestra is ready!` startup banner was written straight to `System.out`, bypassing the logging framework and breaking log parsers that expect every line to come from the configured logger.

Route it through `log.info` in both server commands:

- `StandAloneCommand` - also gains `@Slf4j`, since Lombok generates `log` as `private static final`, so the one on `AbstractServerCommand` is not inherited.
- `WebServerCommand` - already had a logger; just swaps the `System.out.println` call.

The leading `\n` is dropped, as the logger owns line formatting, and the URL becomes a `{}` parameter instead of string concatenation.

## Notes

`grep -rn "Kestra is ready" --include=*.java` confirms these were the only two occurrences.

Closes #17946

### DIFF
--- a/cli/src/main/java/io/kestra/cli/commands/servers/StandAloneCommand.java
+++ b/cli/src/main/java/io/kestra/cli/commands/servers/StandAloneCommand.java
@@ -21,6 +21,7 @@ import io.kestra.core.utils.Await;
 import jakarta.annotation.Nullable;
 import jakarta.inject.Inject;
 import jakarta.inject.Provider;
+import lombok.extern.slf4j.Slf4j;
 import picocli.CommandLine;
 import picocli.CommandLine.Option;
 
@@ -28,6 +29,7 @@ import picocli.CommandLine.Option;
     name = "standalone",
     description = "Start the standalone all-in-one server"
 )
+@Slf4j
 public class StandAloneCommand extends AbstractServerCommand {
     @CommandLine.Spec
     CommandLine.Model.CommandSpec spec;
@@ -142,7 +144,7 @@ public class StandAloneCommand extends AbstractServerCommand {
                 fileWatcher.startListeningFromConfig();
             }
 
-            embeddedServer.ifPresent(server -> System.out.println("\n✅ Kestra is ready! Open the UI at: " + server.getURL()));
+            embeddedServer.ifPresent(server -> log.info("✅ Kestra is ready! Open the UI at: {}", server.getURL()));
 
             Await.await().forever().until(() -> !this.applicationContext.isRunning());
         }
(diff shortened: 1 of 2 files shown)


aetherloop:
@MilosPaunovic Please review the PR! Thank you

loicmathieu:
I agree that it's not perfect but it was specifically asked so it's more clearly seen...
We'll need to discuss this internaly.

AcevedoR:
still need discussing @loicmathieu or are we good to merge ?

cli/src/main/java/io/kestra/cli/commands/servers/StandAloneCommand.java
```diff
@@ -142,7 +144,7 @@ public Integer call() throws Exception {
                 fileWatcher.startListeningFromConfig();
             }
 
-            embeddedServer.ifPresent(server -> System.out.println("\n✅ Kestra is ready! Open the UI at: " + server.getURL()));
+            embeddedServer.ifPresent(server -> log.info("✅ Kestra is ready! Open the UI at: {}", server.getURL()));
```
loicmathieu, line 147:
```suggestion
            embeddedServer.ifPresent(server -> log.info("\n✅ Kestra is ready! Open the UI at: {}", server.getURL()));
```

cli/src/main/java/io/kestra/cli/commands/servers/WebServerCommand.java
```diff
@@ -95,7 +95,7 @@ public Integer call() throws Exception {
         }
 
         log.info("Webserver started");
-        embeddedServer.ifPresent(server -> System.out.println("\n✅ Kestra is ready! Open the UI at: " + server.getURL()));
+        embeddedServer.ifPresent(server -> log.info("✅ Kestra is ready! Open the UI at: {}", server.getURL()));
```
loicmathieu, line 98:
```suggestion
        embeddedServer.ifPresent(server -> log.info("\n✅ Kestra is ready! Open the UI at: {}", server.getURL()));
```

loicmathieu (changes requested):
I agree to merge it as your rational for this line to goes into whatever appender is configured is accurate.

But please add back the newline so it's more obvious to see on the console as it's why we did it in the first place.

tchiotludo:
the new line is the main reason of the issue, it break log parser! 
Perfect like this, thanks @aetherloop

PR #18071, opened 2026-08-11
fix(ui): default logs and editor font size to 12px at medium mode

### LINKED ISSUE #17963: Default font size for editor / logs
### Describe the issue

Image

Default font size for log and editor are 14px (like the interface font size). 
But monospace font are heavily taller than sans serif one, making logs highly visible .

Image

Move the default to be 12px 

### Environment

- Kestra Version: develop

Lalithkishore1224:
Fixes #17963

## Summary

Logs and the code editor now derive their base font size from a dedicated **monospace** scale (`11 / 12 / 14` for small / medium / large) instead of the interface scale (`12 / 14 / 16`). The medium mode — the default — now renders logs and editor text at **12px** instead of 14px.

## Why

Monospace fonts are visually taller than sans-serif ones, so rendering them at the same size as the interface (`14px`) makes logs and editor content look oversized and hard to scan. Keeping monospace content one notch below the interface scale restores visual balance. See the screenshots in #17963.

## Changes

- Add `MONO_BASE_PX` (`11 / 12 / 14`) in `ui/src/utils/appFontSize.ts` and `ui/packages/design-system/src/utils/fontScale.ts`.
- Use `MONO_BASE_PX` for the effective logs font size (`useLogDisplay`) and the editor font size (`useKsEditor`).
- Keep the one-time localStorage migration at `_fontMigratedV2` unchanged. Since V2 already shipped, a stored `logsFontSize=14` is an explicit user choice, not a leftover default, and must not be silently cleared. The existing V2 migration still clears the legacy defaults for users who never ran it, landing them on the new mono values.

## Behavior after this change

| Mode   | Before | After |
|--------|--------|-------|
| Small  | 12px   | 11px  |
| Medium | 14px   | 12px  |
| Large  | 16px   | 14px  |

Existing explicit user overrides are preserved; the existing one-time migration only clears the legacy defaults (`logsFontSize=14`, `editorFontSize=12`) for users who never ran it.

## Tests

- Updated `ui/tests/unit/composables/useLogDisplay.spec.ts` for the new mono defaults and the existing V2 migration behavior.
- Added `MONO_BASE_PX` coverage in `ui/tests/unit/utils/appFontSize.spec.ts`.
- 21 unit tests pass; `check:types` is clean.


RemiBarthe (changes requested):
Thanks for the contribution ! Good call on giving mono its own scale

One change before merge: please drop the migration hunk entirely and keep MIGRATION_FLAG = "_fontMigratedV2" with both legacy clears as they are on develop.

V2 already shipped, so for those users a stored logsFontSize === 14 isn't a leftover default. It's a value they explicitly picked in Settings. Bumping to V3 re-runs the migration and silently deletes it.
Keeping V2 handles both cases: users who never ran it get their old defaults cleared and land on the new mono values; already-migrated users keep their choices.

Lalithkishore1224:
Addressed the review: dropped the V3 migration bump. `MIGRATION_FLAG` stays at `_fontMigratedV2` with both legacy clears (`logsFontSize=14`, `editorFontSize=12`) exactly as on develop, so already-migrated users keep their explicit choices while never-migrated users still get cleared onto the new mono defaults. Pushed in 4409b90eba.

RemiBarthe (approved):
Great work, thanks for your contribution! 
LGTM! 🚀

PR #18075, opened 2026-08-12
Feat: Implemented taskrun-level statistics

### LINKED ISSUE #17810: Add taskrun statistics
### Feature description

In 2.0, we compute execution statistics.

Add taskrun-level statistics, as it might be costly and use a lot of database space; it should be disabled by default.

This would open the door for finer-grained dashboards (task-level dashboards) and SLAs.

MoMoUsama:
closes #17810

### Description
Introduces task-run level statistics stored in a separate table for fast retrieval and long-term retention. In JDBC, raw entries are periodically compacted into 1-minute aggregate buckets to reduce resource usage and table bloat.

### Key Points
- Disabled by Default: `Enable using kestra.task-run-statistics.enabled`.

- Configurable Compaction Interval: 1-minute compaction runs are configurable via `kestra.jdbc.task-run-statistics.compactor.fixed-delay` (default: 1m).

- Bean Registration Strategy: For simplicity, feature related beans are initialized even when disabled except the Compactor not created if the flag is disabled, avoiding background polling overhead.


MoMoUsama:
Hi @loicmathieu, @MilosPaunovic,
Just checking in, is there anything I should improve on my PR before it can be merged?

PR #18089, opened 2026-08-12
fix(core): keep per-emit asset lineage bundles instead of flattening (1.3 backport)

Malaydewangan09:
Backport of #18009 to `releases/v1.3.x`.

With `assets.enableAuto: true` on a task that emits many models (for example dbt), the asset lineage graph collapsed into a near-cartesian product: every emit was merged into one `assets` bundle, so USED_BY was drawn from every input to every output.

This keeps each emit as its own bundle:
- `TaskRun`: the single `assets` field is replaced by a per-emit `assetEmits` list, mirroring #18009.
- `DefaultWorker` builds one bundle per emit (plus the manual `assets:` declaration as its own bundle) instead of flattening.
- `ExecutorService` reads `assetEmits` and passes the bundles to `assetLineage`, so USED_BY edges are written per pair.

Backward compatibility (1.3 is a minor). Following Loic's review guidance for the backport:
- A deprecated `@JsonSetter("assets") setAssets(...)` folds the old single `assets` object into `assetEmits` on deserialization, so old executions in the DB still load. This is the same deserialize bridge kept on 2.0.
- The deprecated `getAssets()` getter is kept (2.0 drops it). Loic asked to keep it here so callers expecting a single `assets` keep working.
- `assetEmits` is `@JsonInclude(NON_EMPTY)`.

- part-of: kestra-io/plugin-dbt#306
- companion PR: kestra-io/kestra-ee#9853

### DIFF
--- a/core/src/main/java/io/kestra/core/assets/AssetService.java
+++ b/core/src/main/java/io/kestra/core/assets/AssetService.java
@@ -5,6 +5,7 @@ import java.util.List;
 import io.kestra.core.models.assets.Asset;
 import io.kestra.core.models.assets.AssetIdentifier;
 import io.kestra.core.models.assets.AssetUser;
+import io.kestra.core.models.assets.AssetsInOut;
 import io.kestra.core.queues.QueueException;
 
 import io.micronaut.context.annotation.Secondary;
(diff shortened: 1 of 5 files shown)


loicmathieu (approved):
LGTM

PR #18099, opened 2026-08-12
fix(ui): offer docs instead of a mismatched video on empty states

anna-geller:
Several empty states offered a "Watch the video" button pointing at a recording about a different feature. The Policies state, for instance, linked to the Plugin Defaults video.

Empty states without a video now show a **"Learn more"** button on their documentation page instead of the smaller arrow link below it. States with a matching video are unchanged. Four videos covering a different feature were removed, and a missing `groups` entry was added.

@wrussell1999 let's replace the `learnMore` URLs with [short.io](https://short.io) links as discussed

Two things that aren't visible in the diff:
- No Enterprise Edition change is needed
- `cases`, `policies`, `quotas` and `mcpToolFlows` return 404 on kestra.io/docs until the 2.0 docs are released, and they all resolve on the 2.0 preview


elevatebart (approved):
Good job!!!
NItpick: Too verbose comments in the code.

Think of the dev looking at this file in a month. 
Does he have time for that much reading?
Is everything valuable to him?

wrussell1999:
Videos are gone from every empty state, not just the mismatched ones, since the docs pages already embed them.

The Secrets page now renders Empty with demoCta like Policies does, replacing around 90 lines of bespoke markup and CSS. Two side effects worth a look: the Enterprise Edition tag is gone (no other demo page shows one) and the primary button now reads "Contact Sales". That page also had a layout bug, where its v-show hidden table still matched `:has(.ks-data-table-wrapper)` in `app.scss` and pushed the promo to the bottom of an empty page. The table unmounts now.

Its upsell copy went from 628 to 236 characters, in line with the other pages. The twelve translations are hand-written since no GEMINI_API_KEY was available, so they are worth a skim.

Still on the old utm_source=app scheme, if we want them here: the demo CTAs, and the inline doc links inside the translated copy.

wrussell1999:
Secrets are more inline with the other enterprise pages. I think this is ready to be merged in but would be good to have a second pair of eyes check it out.

Screenshot 2026-08-12 at 16 52 43

PR #18103, opened 2026-08-12
fix(flows): lock id and namespace in the editor instead of reverting them

### LINKED ISSUE #17615: Flow & dashboard editor: make id/namespace read-only but still copyable
### Context

Follow-up from #16627 (review by @elevatebart).

`id` and `namespace` are immutable for an existing flow. Today the YAML editor lets the user type into those lines, then **silently reverts** the change and shows a "Read-only property" notice. #16627 softened the symptom (warning instead of error, deduplicated to once per session), but the underlying interaction is still confusing: the editor accepts input into a field it will immediately undo.

Production analytics over a single week: this notice fired for **49 users across 59 distinct flows** — real friction on one of the highest-time-on-page screens.

### Proposed improvement

Make the top of the flow YAML (the `id:` and `namespace:` lines) **visually read-only but still selectable and copyable** for existing flows, so that:

- users can't accidentally edit them → no revert, no toast,
- they can still select and copy the values,
- the constraint is self-explanatory up front rather than surfaced after the fact.

Apply the same treatment to the dashboard editor (`id`).

### Notes / direction

- Monaco supports read-only regions / editable ranges (constrained-editor pattern) and read-only decorations — the `id`/`namespace` lines can be marked non-editable while the rest of the document stays editable. Read-only ranges stay selectable and copyable by default.
- Existing flows only (`isCreating = false`); creating a new flow keeps both fields editable.
- Once the region is truly read-only, the revert-and-toast logic in `onEdit` (`ui/src/stores/flow.ts`, `ui/src/stores/dashboard.ts`) can be removed.

### Acceptance criteria

- [ ] Opening an existing flow: `id:` / `namespace:` render as read-only in the editor.
- [ ] Those values can still be selected and copied.
- [ ] Editing the rest of the document works normally.
- [ ] Creating a new flow: both fields fully editable.
- [ ] No "Read-only property" toast and no silent revert during normal editing.
- [ ] Same behavior for the dashboard editor (`id`).

ameyypawar:
## Summary

`id` and `namespace` are immutable once a flow exists, but the YAML editor accepted typing into those lines and then, one `onEdit` debounce later, showed a "Read-only property" toast and put the old value back — 49 users across 59 flows in a single week, per the issue.

This locks the lines instead, so the keystroke never lands. They stay selectable and copyable, the rest of the document edits normally, and creating a flow leaves both fields editable — matching what the no-code panel already does via `useFlowFields`.

Notes on the implementation:

- Monaco ships no read-only-range API of its own (the constrained-editor pattern the issue mentions is a third-party plugin), so `useReadOnlyYamlKeys` undoes an offending change on the tick it arrives. Only the locked lines are rewritten, so a paste or find-and-replace spanning them keeps everything else. An inline comment on a locked line is carried over.
- The correction goes through `model.pushEditOperations` with no `pushStackElement()`, so it joins the undo element already open for the user's keystroke. `executeEdits` opened a second one, and undoing that restored the violating text and re-entered the listener.
- `onEdit`'s warning is suppressed only for this editor, via a `metadataGuarded` flag gated on a code editor having actually mounted. The no-code editor and applied Copilot drafts still surface it.

**Scope:** the dashboard half is no longer reachable here — `DashboardCodeEditor.vue` was deleted by #17942 ("custom dashboard to enterprise only"). The no-revert/no-toast part of that acceptance criterion was separately satisfied by #17652 in July; only the read-only decoration is left, with no OSS dashboard editor to carry it.

Closes https://github.com/kestra-io/kestra/issues/17615, follow-up to #16627

## Test plan

- [x] `useReadOnlyYamlKeys.spec.ts` — 47 cases, driven through an editor double that models Monaco's undo stack. Confirmed to **fail against the previous implementation**, not just pass against this one.
- [x] `npx vitest run --project unit` — 217 files, 2091 tests, 0 failures.
- [x] `oxlint` clean. `eslint` 0 errors, 1 warning: `vue/enforce-style-attribute` on the unscoped `<style>` — the same rule as the review's point about moving that CSS into `useEditorDecorations`.
- [x] `vue-tsc --build tsconfig.app.json` — no errors in any file this PR touches.
- [ ] `flow-metadata-readonly.spec.ts` is included but **not run here**; it needs a live dev server and backend.


ameyypawar:
Noted, thank you for your patience.

flcarre (commented):
Thanks for this. Locking the keystroke is the right call over the accept-then-revert loop, and the edge cases you covered (inline comments, quoted values, paste spanning a locked and an unlocked line, CRLF, key deletion) are more thorough than most fixes in this area. I pulled the branch and the lock itself works exactly as advertised: typing on `id:` and `namespace:` is refused cleanly, and the decorations land on the right two lines.

One behaviour worth fixing before this lands.

### Undo stops working after any refused keystroke

`useReadOnlyYamlKeys` calls `editor.executeEdits(...)` inside `onDidChangeModelContent` without grouping the correction with the edit that triggered it, so the undo stack ends up in a state it can't step past.

Repro on this branch:

1. Open any flow in the YAML editor.
2. Edit a normal line, for example append something to `message:`.
3. Click the `id:` line and type a character. It is correctly refused.
4. Press Cmd+Z (Ctrl+Z). Nothing happens, and the edit from step 2 can no longer be undone at all, no matter how many times you press it.

I ran the control separately to be sure the keystrokes were reaching Monaco: without step 3, Cmd+Z reverts the step 2 edit normally. It is only after a refused keystroke that undo goes dead for the rest of the session, and the user's earlier legitimate work becomes unreachable.

Wrapping the correction in `pushStackElement()` boundaries, or applying it through `model.pushEditOperations` so it merges with the user's own edit, should be enough.

### Design system

* `FlowFileEditorTab.vue:390` and `DashboardCodeEditor.vue:113` both hardcode `rgba(128, 128, 128, 0.14)`. A fixed 14% grey is not the same affordance over a near-white and a near-black surface, which is the thing the token system exists to solve.
* More to the point, that rule is an unscoped global block reaching into `.monaco-editor`, which is DOM owned by `KsEditor`. The design system already owns this seam in `packages/design-system/src/composables/useEditorDecorations.ts` (`highlightLinesRange`). Moving the rule up there, and ideally the decoration call too as a `readOnlyLines` option, would give both editors the same treatment from one place and remove the duplication that the dashboard file's own comment calls out.

### Accessibility

The lock is announced only through a Monaco `hoverMessage`, which is mouse-only and invisible to assistive tech. A keyboard or screen-reader user types on the `id:` line and the keystroke vanishes with no feedback at all. A gutter lock glyph with an accessible label, or an inline message when an edit is refused, would cover that.

### One smaller thing

The correction rewrites the whole line as `key: expected`, so `id: my_flow # keep` loses its comment the first time any violating keystroke lands, including the comment the user was allowed to edit. `readTopLevelValue` already strips comments and quoting when reading, so the information is there to preserve the original suffix.

Nice work on the edge cases. The undo one is the only thing I would call blocking.

flcarre (commented):
Round 2 on `7673c24a`.

**The blocker is properly fixed.** `applyCorrection` now goes through `model.pushEditOperations` with no intervening `pushStackElement()`, and the doc comment explains the mechanism accurately. What I appreciate more is that the test double models the undo stack — `executeEdits` pushes a snapshot, `pushEditOperations` does not — so `leaves undo working after a keystroke is refused` and `does not grow the undo stack for a correction` would both fail against the previous implementation. That is a test that can fail for the right reason, not one that documents the fix.

Three of my other four points are unchanged, though. Taking them in order of how much I care.

## 1. The comment on a locked line is still destroyed

`restoreLines` writes `text: \`${key}: ${options.expected.value[key]}\``, so `id: my_flow # keep` becomes `id: my_flow` the first time any violating keystroke lands on that line.

The new test reads like it covers this but does not:

```js
it("does not freeze a document whose locked line carries a comment", () => {
    const edited = commented.replace("message: hello", "message: changed")
    double.type(edited)
    expect(double.current()).toBe(edited)
})
```

That edit is on `message:`, so `violatedKeys` returns `[]` and `restoreLines` never runs. It proves a comment does not *provoke* a correction — which was never the concern — not that a correction *preserves* one. `scalarOf` already separates the value from the suffix, so the suffix is available to keep.

## 2. The design-system point — engaging with your reasoning, not repeating mine

You answered the duplication with a chunking argument: an SFC style block ships in that component's CSS chunk, and a dashboard route need never mount `FlowFileEditorTab`. That is true, but it argues *for* moving the rule rather than against: `packages/design-system/src/composables/useEditorDecorations.ts` already owns this exact seam, and it ships with `KsEditor` — which both routes mount by definition, since neither can show an editor without it. One copy, loaded exactly when the editor is, and no rule shipped to a route that has no editor at all.

The `rgba(128, 128, 128, 0.14)` half is the part I would still push back on. The new comment says the tint "is theme-neutral so it reads as locked on both the light and dark editor themes" — that restates the premise rather than answering it. A fixed 14% grey composited over a near-white surface and over a near-black one gives two different contrast ratios against two different backgrounds; "neutral in hue" is not "equivalent in effect". That is the specific thing `--ks-*` exists to resolve, and `ui/AGENTS.md` rules out `rgba()` in feature code without exception.

If no existing token expresses "locked line tint", that is a design-system gap worth raising rather than routing around — and it would be the natural companion to a `readOnlyLines` option on `useEditorDecorations`.

## 3. Accessibility is unchanged

The lock is still announced only through Monaco's `hoverMessage`, which is mouse-only. Your own E2E spec is the clearest statement of the problem: `refuses a keystroke aimed at the id line` asserts the character is *never inserted*, with nothing asserting the user is told why. For a keyboard or screen-reader user the keystroke vanishes silently.

Not a merge blocker in my view, but it should not close silently either — a gutter glyph with an accessible label, or an inline message on refusal, would cover it.

## 4. New — the guard re-attaches more often than the comment claims

`useReadOnlyYamlKeys.ts`, the watcher at the bottom:

```js
// ... `expected` is a computed whose identity already changes exactly when the locked values do.
watch([options.editor, options.enabled, options.expected], ...)
```

That is not what the call site produces. `FlowFileEditorTab` passes:

```js
expected: computed(() => props.flow ? {id: flowStore.flow?.id, namespace: flowStore.flow?.namespace} : {})
```

A computed returning a fresh object yields a new reference whenever any dependency invalidates it — and `flowStore.flow` is *replaced* on every save, revision load and refetch, not only when the id or namespace changes. Each of those fires the watcher, which runs `attach()` → `detach()`, clearing the decorations and resetting `lastValid`.

Mostly self-healing, since `attach` repaints and recomputes from the current model. The narrow hole is that `attach` sets `lastValid = violatedKeys(initial).length ? undefined : initial`, so if the buffer happens to violate at that instant, `lastValid` becomes `undefined` and the whole-document fallback — the only thing that can recover a *deleted* locked key — silently does nothing until the next clean edit.

This is the trap `ui/AGENTS.md` documents under "The deep-watch / computed-spread trap": an object-returning computed as a watch source needs to be a primitive instead. Keying on the values rather than the container fixes both the churn and the wrong comment:

```js
watch([options.editor, options.enabled, () => JSON.stringify(options.expected.value)], ...)
```

---

To be clear about where this leaves things: the thing I called blocking is fixed, and the edge-case coverage on the composable is genuinely good. Items 1 and 4 are small and concrete. Item 2 is a convention call I would not merge around, but it is not mine alone to settle — worth @MilosPaunovic's or a design-system owner's read on whether the rule belongs in `useEditorDecorations`.

---

ameyypawar:
Merged `develop` in, and updated the description — the dashboard half of this PR is gone with the file: `DashboardCodeEditor.vue` was deleted by #17942 ("custom dashboard to enterprise only").

On that acceptance criterion in #17615: the no-revert/no-toast part was already satisfied by #17652 back in July. Only the read-only decoration is left over, and there is no OSS dashboard editor to put it on — whether that reopens elsewhere or closes is your call, so I have left the `Closes` link alone.

Also fixed since your last pass: `metadataGuarded` suppressed the warning without checking a code editor had actually mounted, so a revision-preview diff editor left the lock absent and the warning gone — the silent revert this PR exists to remove.

`--project unit` is 2091 passing on the merged branch.

flcarre (changes requested):
This is strong work. Refusing the edit on the tick it arrives instead of undoing it a second later is the right call, and `pushEditOperations` without an intervening `pushStackElement()` — so one undo reverts both halves — is exactly the subtlety that makes this kind of guard usable rather than a trap. The 39 unit tests over a hand-rolled model cover the parts that actually matter (undo stack, paste spanning a locked line, comment preservation, re-attach identity). A few things to fix before merge.

### Hardcoded colour, and Monaco styling in feature code

```scss
.monaco-editor .ks-readonly-yaml-line {
    background-color: rgba(128, 128, 128, 0.14);
}
```

`ui/AGENTS.md` rules this out: colours come from `--ks-*` tokens, never `rgb()`/`rgba()`, and a raw colour is an explicit reject criterion in the UI review checklist. "Theme-neutral so it reads as locked on both themes" is the reasoning a token is supposed to carry — hand-picking a grey means it will be almost right in one theme and slightly wrong in the other, with nothing to grep for later.

Since this styles Monaco's internals and `KsEditor` is a design-system component, the rule belongs in the design system alongside the existing `--ks-editor-*` family rather than in an unscoped `<style>` block in a feature component. If no suitable token exists, please add one to `ks-theme-light.scss` / `ks-theme-dark.scss` / `ks-theme-dark-2.scss` with design, and keep the selector in the design system.

### The hover tooltip reuses the toast copy, which no longer describes what happened

`hoverMessage: computed(() => t("namespace and id readonly"))` resolves to:

> The properties `namespace` and `id` cannot be changed — they are now set to their initial values. If you want to rename a flow or change its namespace, you can create a new flow and remove the old one.

That is a past-tense explanation of a revert. On hover over a locked line, nothing has been reverted — the whole point of this change is that the revert no longer happens. It needs its own short key, something like:

```json
"flow metadata locked": "`id` and `namespace` cannot be changed after a flow is created."
```

plus propagation to the twelve locales (`npm run translations:generate`, then `translations:check` clean).

### The whole-document fallback discards a paste with no feedback at all

When a locked key is removed outright, `restoreLines` returns `false` and the listener replaces the entire buffer with `lastValid`. `keeps the rest of a paste that also changed a locked line` covers the good path, but this one is different: a paste that simply does not carry an `id:` line — a partial snippet, a document copied from elsewhere — is silently thrown away in full.

That is worse than the behaviour being replaced, because `metadataGuarded` now suppresses the warning on the grounds that "the user was never able to make the change". True for a keystroke on a locked line; not true here, where the edit was accepted and then wholesale reverted. Please surface a message on that specific path (the fallback branch), and add a test for it.

### `enabled` flipping true while the buffer already violates

`attach()` leaves `lastValid` undefined when the initial buffer violates, which disables the only recovery path — the code comment on the `watch` acknowledges this. Then a paste dropping `id` does nothing at all and the violation persists with no feedback. Small enough to leave, but worth a line in the composable doc saying it is a known dead end rather than leaving the next reader to work it out.

### Comment density

`AGENTS.md`: the default is no comment, one sentence when a reviewer would otherwise have to ask *why*, and no JSDoc on internal functions. As written there are paragraph-length blocks on `splitScalar`, `applyCorrection`, `restoreLines`, a twelve-line block on `useReadOnlyYamlKeys`, and a ten-line block on the `watch`. The *reasons* are genuinely worth keeping — the undo-element merge and the serialised watch key especially — but each should be one sentence. The rest (what the two halves of `splitScalar` are for, how the recursion works) is visible in the code.

### Nits

- `restoreLines` writes `${key}: ${expected[key]}`, so a line the user had written as `id: "my_flow"` comes back unquoted. Cosmetic for ids and namespaces, which never need quoting, but it is a silent rewrite of something the user typed — the same argument the code already makes for preserving inline comments.
- `reports a changed id` / `reports a changed namespace` are the same test with one property changed; a couple of others in `violatedKeys` are close. Worth trimming per `AGENTS.md` ("one test per behavior, rather than one per property").
- Neither the e2e nor the unit suite covers navigating from one flow to another, or restoring a revision, while the guard is attached. The `expected` object and the editor buffer are updated by different code paths, so it would be good to have one test pinning that a whole-document replacement with fresh expectations is not read as a violation. The composable's design argues it cannot happen; a test would make that argument load-bearing.

ameyypawar:
All six addressed. Three worth calling out.

**Colour** — used the existing `--ks-bg-inactive` rather than adding a token: `ks-theme-*` is generated from `scripts/Figma.json`, so a new one has to come from design. Selector moved into `KsEditor` next to `highlight-lines`. If that token reads wrong, easy to swap.

**Translations** — `translations:generate` needs a `GEMINI_API_KEY` I don't have, so the twelve locales are hand-written, with the fingerprint from the repo's own `fingerprintOf`. `translations:check` and `check-translations.mjs --scope oss` pass with no stale keys, but the wording wants a second pair of eyes.

**Navigation test** — it passes with the expectations updated before the buffer; in the reverse order the guard reverts the incoming document. So "cannot happen" rests on the composable's watch running before KsEditor's `modelValue` watcher, which is my reading of watcher creation order rather than something I observed running. Worth a check — happy to guard it explicitly instead.

flcarre (changes requested):
Round 3 on `b1c14c1`. Every one of my six points is addressed, and several are addressed better than I asked for.

* The Monaco rule now lives in `KsEditor.vue` on `var(--ks-bg-inactive)`. I checked the token: it is declared in all three of `ks-theme-light.scss`, `ks-theme-dark.scss` and `ks-theme-dark-2.scss`, so it is a real token and not a plausible-looking name.
* `flow metadata locked` is its own key, present in the thirteen locale files with real translations rather than English copies, with `flow`, `id` and `namespace` correctly left untranslated. Fingerprints updated.
* `onReverted` fires only on the fallback branch, and `says nothing when the locked line was corrected in place` pins that it does not fire on the ordinary path. That is the distinction I was worried about, tested from both sides.
* The already-violating-buffer dead end is documented where the next reader will hit it.
* The comment blocks are down to the reasons that are genuinely load-bearing.
* Quoting is preserved, the duplicated `violatedKeys` test is gone, and the navigation swap now has a test.

Nothing above needs another round. What follows is new, and it came in with this round's restructuring.

## The two lines dropped from the change listener were doing work

The listener used to end with:

```ts
lastValid = model.getValue()
paint(editor, lastValid)
```

Both are gone, replaced by `if (fallback !== undefined) options.onReverted?.(violations)`. The clean-edit path above still does `lastValid = current; paint(editor, current)`, so this only bites on the correction path, but it bites in two ways.

### `lastValid` now lags behind an in-place correction

`restoreLines` deliberately keeps the rest of the edit, which means the buffer after a correction can hold work that `lastValid` does not.

1. Paste a document that changes `id:` **and** adds a task.
2. `restoreLines` puts `id:` back and keeps the new task. Buffer: original flow plus the task. `lastValid`: original flow, without the task.
3. Delete the `id:` line. `restoreLines` returns false, the fallback restores `lastValid`, and the task from step 1 is gone.

`keeps the rest of a paste that also changed a locked line` covers step 2, which is why this passes. Re-recording `lastValid` after a successful `restoreLines` closes it.

### The decorations are not repainted after the whole-document fallback

The fallback applies an edit over `model.getFullModelRange()`. Monaco's decorations are tracked ranges, so a replacement spanning them collapses them rather than moving them, and the locked-line highlight and its hover end up on line 1 instead of on the `id:` and `namespace:` lines. `paint()` used to rebuild them right after; now nothing does until the next clean edit.

To be straight about the epistemics: I read this off the code and the Monaco range-tracking semantics, I did not watch it happen in a browser. The unit double models the undo stack but not decorations, so no test here can see it either way. If you would rather confirm before changing anything, expanding the double to record decoration ranges would make it visible, and would be the same kind of test as the undo one.

## Unrelated to your change

Neither `7e2af818` nor the commit before it has a single check run against it. CI has not executed on this head at all, so the PR cannot merge regardless of review state. Worth a nudge to get the workflow triggered.

flcarre (approved):
Confirmed both fixes from my last review are real, not just claimed — `lastValid` is re-recorded after an in-place correction and decorations repaint unconditionally after the fallback, both pinned by new tests that model Monaco's actual decoration-collapse behavior. Token usage, translations (all 13 locales, fingerprints match), and the deep-watch fix all check out. Approving.

One non-blocking note carried over from before: refusal feedback is still hover-only, no keyboard/screen-reader affordance — you already called this non-blocking, just flagging it's still there. Separately: this branch has never run CI (zero check-runs on HEAD) — worth a rebase/re-push to actually get a signal before merge.

MilosPaunovic:

PR #18109, opened 2026-08-12
fix(core): make GraphUtils topology walk iterative and cycle-safe

AcevedoR:
## Problem

A production instance exited with code 1 on a `StackOverflowError` in `GraphUtils.recursiveEdge()` after a user replayed a failed task.

Two task runs for one task build two nodes sharing a UID, turning the chain edge between them into a **self-loop**. `recursiveEdge()` had no visited set and followed it forever — depth is irrelevant, four tasks suffice. Being an `Error` it escaped `catch (Exception e)` in the queue handlers, so `ThreadUncaughtExceptionHandler` exited the JVM: one replay took down executor, scheduler, worker and webserver together.

## Change

`recursiveEdge` becomes an explicit-stack DFS: each node is visited once, and a node reached twice on the current path throws `CyclicGraphException` naming the cycle. Edges are indexed by source rather than rescanned at every step. Output is unchanged — `successors()` only consumes the set of reachable edge sources. Replay now fails diagnosably instead of killing the instance.

Tests are TDD; both new cases fail with `StackOverflowError` on the unfixed code.

## Follow-ups

- #18112 stops the same-UID nodes being built at all.
- An `Error` in a queue handler still exits the JVM.

### DIFF
--- a/core/src/main/java/io/kestra/core/exceptions/CyclicGraphException.java
+++ b/core/src/main/java/io/kestra/core/exceptions/CyclicGraphException.java
@@ -1,0 +1,19 @@
+package io.kestra.core.exceptions;
+
+import java.io.Serial;
+import java.util.List;
+
+public class CyclicGraphException extends KestraRuntimeException {
+    @Serial
+    private static final long serialVersionUID = 1L;
+
+    public CyclicGraphException(String message) {
+        super(message);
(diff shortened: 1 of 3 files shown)


loicmathieu (approved):
LGTM

Issue #18097, opened 2026-08-12
Command palette "Go to <section>" leaves the target hidden at the canvas edge


flcarre:
### Describe the issue

In the no-code Blocks editor, running a `Go to <section>` command from the command palette selects the right block — the status bar shows `Selected · Errors` and the keyboard focus ring does move — but the canvas barely scrolls: the destination lands flush against the bottom edge of the viewport, where the editor's own status bar sits on top of it. The jump reads as "nothing happened", and you have to scroll manually to find where the selection went.

Spotted while recording a demo of the Blocks editor (#17096) on `develop`.

### Steps to reproduce

1. Open a flow with enough blocks to overflow the canvas height, in the flow editor's **No-code** tab (viewport 1920x1080)
2. Click the canvas so the selection ring sits on the first block
3. Press <kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd>, type `errors`, activate **Go to Errors** (clicking the entry and pressing <kbd>Enter</kbd> behave the same)

### Expected behaviour

The target section scrolls into comfortable view — centred, or at least clear of the sticky status bar — so the jump is visible.

### Actual behaviour

The selection is correct (`__section:errors` carries the ring, the status bar reads `Selected · Errors`) but the section is left at the very bottom edge of the canvas. Measured immediately after the command, in a 1080px-high viewport:

| Command | Target's `getBoundingClientRect().top` after the jump |
| --- | --- |
| Go to Errors | 1028px (≈50px visible, partly behind the status bar) |
| Go to Finally | 881px |

The same jump performed by walking there with the arrow keys ends up looking fine, because each step only moves between neighbours.

### Probable root cause

`ui/src/components/no-code/blocks/useCanvasFocus.ts:32`:

```ts
cardFocusTarget(card).focus({preventScroll: true})
card.scrollIntoView({block: "nearest"})
```

`block: "nearest"` scrolls the minimum distance needed. That is the right call for arrow-key stepping between adjacent cards, but for a palette jump across the whole canvas it parks the destination against the viewport edge, and it does not account for the canvas status bar overlaying the bottom of the scroll container.

Worth considering: `block: "center"` for jumps that are not adjacent moves (or `scroll-padding-bottom` on the scroll container matching the status-bar height), keeping `nearest` for arrow-key navigation.

### Environment

- Branch: `develop` (post-#17096), OSS, local instance
- Reproduced in Chromium 1920x1080, both light theme and modal task-edit mode

[...]

PR #18110, opened 2026-08-12
fix(no-code): centre the canvas on a command-palette jump

ameyypawar:
## Summary

- `Go to <section>` from the command palette selected the right block but barely scrolled: the destination landed flush against the bottom edge of the viewport, where the status bar is painted over the scroll container. The jump read as "nothing happened".
- `useCanvasFocus` scrolled every focus move with `scrollIntoView({block: "nearest"})`. That is the right call for arrow-key stepping between neighbours, but a palette jump crosses the whole canvas and `nearest` moves the minimum distance needed.
- `focusCanvasCard` now takes an optional alignment, still defaulting to `nearest`. Only `goToSection` asks for `center`, so keyboard navigation is untouched.
- `.block-editor-main` also declares `scroll-padding-bottom` matching the status bar it overlays, so `nearest` can no longer stop with a card underneath it either. That is a separate symptom from the one reported — arrow-stepping to the last block hit it too.

Related: #18097

## Test plan

- [x] `ui/tests/e2e/blocks/blocks-navigation.spec.ts`, on a canvas tall enough to scroll (new `blocks-tall.yaml` fixture). Two of its three cases were **confirmed to fail against a build without this change**: the jump lands 230px from the scrollport's centre against a 129px bound, and arrow-stepping to the last block leaves its bottom at 719.5 while the status bar starts at 684 — 35px underneath it. The third case (jumping back up to Tasks) passes either way and is a directional guard only.
- [x] The jump assertion measures distance from the scrollport's centre rather than "not at the bottom", so a revert to either `nearest` or `start` fails it.
- [x] `ui/tests/unit/components/no-code/blocks/useCanvasFocus.spec.ts` pins the contract: the default stays `nearest`, `moveFocus` stays `nearest`, and only an explicit `align` opts into `center`.
- [x] `npm run test:unit`: 1540 passing, 0 failing (+4, exactly the new spec).
- [x] `npm run check:types` passes for the design-system, app and test projects.
- [x] `npm run test:lint`: 0 errors on the touched files.
- [x] Full `tests/e2e/blocks` suite against the change: 46 passing. The suite is flaky under parallelism on `develop` as well — two baseline runs there failed different tests — and the two failures that persist (`blocks-edit-forms.spec.ts:151` and `:173`) reproduce on `develop` in isolation.

> `useTaskPicker.ts` has a similar `scrollIntoView({block: "nearest"})` for section heads when inserting. Left alone as out of scope — happy to fold it in if you would rather.


MilosPaunovic:
Hi there, @ameyypawar! 👋

ameyypawar:
Noted, thank you for your patience.

ui/tests/e2e/blocks/blocks-navigation.spec.ts
```diff
@@ -1,128 +1,88 @@
...
+import {expectRing, login, openBlockEditor, walkTo} from "./blocks.helpers"
 
-// Keyboard-only navigation across the whole block editor canvas: the flow is
-// the blocks-canvas fixture (trigger, Sequential with two children plus its
-// own empty errors/finally lanes, two leaf tasks, empty top-level
-// errors/finally sections).
-test.describe("Block editor — keyboard navigation", () => {
+// Moving around a canvas that is taller than the viewport. Both behaviours here
+// only exist once the canvas actually scrolls, hence the tall fixture.
+test.describe("Block editor — canvas navigation", () => {
```
flcarre: **This is the one blocking point.** The PR rewrites this spec wholesale instead of adding to it, and eight pre-existing keyboard-navigation e2e tests are deleted with nothing replacing them:

- walk order + wrap-around
- ArrowUp
- step into a group / step back out
- collapse / expand
- the roving-tabindex contract on Tab
- click syncs the focus ring
- Enter opens the dock
- the help overlay

None of that is related to the centring fix, and all of it is coverage we lose. On top of that the fixture switches to one with no flowable, so nested-lane traversal isn't even reachable from this file any more.

Could you keep the existing tests as they are and add the new centring cases alongside them? If the new fixture is genuinely needed for the tall-canvas scenario, the cleanest split is to leave `blocks-navigation.spec.ts` on its current fixture and put the centring tests in their own spec with `blocks-tall.yaml`.

---

ui/tests/e2e/blocks/blocks-navigation.spec.ts
```diff
@@ -1,128 +1,88 @@
...
 
     test.beforeEach(async ({page, request, baseURL}) => {
         flowsApi = new FlowsApi(request, baseURL)
-        flowId = await flowsApi.generateFlowViaApi("blocks-canvas.yaml", "blocks-canvas-fixture")
+        flowId = await flowsApi.generateFlowViaApi("blocks-tall.yaml", "blocks-tall-fixture")
         await login(page)
         await openBlockEditor(page, flowId)
+
+        // Everything below measures a scroll, so prove there is one to measure.
+        const overflows = await page.locator(".block-editor-main").evaluate(
```
flcarre: The scrollport is selected by `.block-editor-main`, which is a styling class — here and again at lines 51 and 82. `ui/AGENTS.md` is explicit that e2e selectors must be `data-test` attributes and never `.el-*` / `.ks-*` / style class names, precisely because those aren't a stable contract.

There's no `data-test` on the scroll container yet, so this needs adding one in `BlockEditor.vue` and pointing the three call sites at it.

---

ui/src/components/no-code/blocks/BlockEditor.vue
```diff
@@ -729,6 +735,10 @@
         height: 100%;
         overflow-y: auto;
         padding: var(--ks-spacing-6) var(--ks-spacing-4) calc(2.25rem + var(--ks-spacing-6));
+        /* The status bar is painted over the bottom of this scroll container, so
+           scrollIntoView has to stop short of it — otherwise a card scrolled to
+           the bottom edge ends up underneath it. */
+        scroll-padding-bottom: calc(2.25rem + var(--ks-spacing-6));
```
flcarre: The `scroll-padding-bottom` is the right mechanism — `nearest` otherwise treats the absolutely-positioned footer as part of the visible scrollport — but this is now the fourth place the `2.25rem` status-bar height is hardcoded in this file.

If the footer ever grows, the reserved strip silently stops matching it and the bug this PR fixes comes back, with nothing failing to tell us. Worth pulling it into a single local custom property (`--status-bar-height`) and having all four sites, this one included, read from it.

---

ui/tests/e2e/blocks/blocks-navigation.spec.ts
```diff
@@ -1,128 +1,88 @@
...
-        await page.keyboard.press("ArrowLeft")
-        await expectRing(page, "seq_group")
+    async function goTo(page: Page, section: string) {
+        await page.keyboard.press("ControlOrMeta+Shift+P")
+        const menuInput = page.getByPlaceholder("Type a command or search a task…")
+        await expect(menuInput).toBeFocused()
+        await menuInput.fill(section)
+        // The same term also matches "Insert <section>", so activate the goto
+        // entry itself rather than pressing Enter on whatever sorted first.
+        await page.getByText(`Go to ${section}`, {exact: true}).click()
```
flcarre: `getByText("Go to …")` here is page-wide. The existing equivalent in `blocks-after-execution.spec.ts:114` scopes the same lookup to the menu, and `blocks.helpers.ts` documents why an unscoped text match is dangerous on this canvas — the block labels themselves can contain the same text, so this can match a card instead of the menu entry.

The `goTo` helper below is also a third copy of logic that already exists twice; better to move it into `blocks.helpers.ts` and have all three specs share it.

---

ui/tests/e2e/blocks/blocks-navigation.spec.ts
```diff
@@ -1,128 +1,88 @@
...
-    test("ArrowLeft on an expanded group collapses it, ArrowRight re-expands", async ({page}) => {
-        await walkTo(page, "seq_group")
-
-        await page.keyboard.press("ArrowLeft")
-        await expect(page.locator("[data-block-id='child_one']")).toBeHidden()
-
-        await page.keyboard.press("ArrowRight")
-        await expect(page.locator("[data-block-id='child_one']")).toBeVisible()
-    })
+    test("arrow-stepping to the last block keeps it clear of the status bar", async ({page}) => {
```
flcarre: This test is named after "the last block", but `err_handler` is stop 18 of 20 in the fixture. The max-scroll case — where the target can't be centred because there isn't half a scrollport of content below it, which is the interesting boundary and the one quoted in the PR description — is never actually exercised.

Either point this at the genuinely last block, or add a second case for it.

Related, on the assertion just above at line 59: the centring bound needs roughly half a scrollport of content below the target, and the fixture only barely provides that. It'll go red on a viewport change or if the sentinel card height changes. Asserting the target is within some tolerance of the scrollport centre, rather than against a bound derived from the fixture's total height, would hold up better.

---

ui/tests/unit/components/no-code/blocks/useCanvasFocus.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+
+function setup(ids: string[]) {
+    const root = buildCanvas(ids)
+    const editorEl = ref<HTMLElement | undefined>(root)
+    // useCanvasFocus registers no lifecycle hooks, so it runs outside a component.
+    return {root, focus: useCanvasFocus(editorEl, () => [])}
+}
+
+describe("useCanvasFocus — scroll alignment", () => {
+    it("scrolls the minimum distance by default", async () => {
```
flcarre, line 45: This pins a one-line pass-through — that `align` reaches `scrollIntoView` — which is behaviour the e2e tests already cover end to end. Per `AGENTS.md` ("don't test that a prop is passed through"), it doesn't earn its place: it can only fail on a refactor that changed nothing a reviewer would care about.

I'd drop this case and keep whichever ones in this file assert actual focus behaviour.

---

flcarre (commented):
**The production change is sound — I found no correctness defect in it.** `focusCanvasCard(id, {align})` defaults to `nearest`, so arrow-stepping is untouched; `goToSection` is the only caller opting into `center`; `scroll-padding-bottom` on `.block-editor-main` correctly stops `nearest` short of the absolutely-positioned footer, which `nearest` would otherwise treat as visible scrollport. And the ordering works out: `then()` in `blockCommandMenu.ts` closes the menu before the action runs, and `focusCanvasCard` defers to `nextTick`, so the measurement happens after the overlay is gone.

My comments are all on the test changes, and one of them is blocking: this PR deletes eight existing keyboard-navigation e2e tests while rewriting the spec, with no replacement. Details inline.

**One finding I couldn't anchor inline** because the file isn't in the diff: `ui/tests/e2e/blocks/PLAN.md:41` still documents the deleted cases as covered. Whichever way the test question is resolved, PLAN.md needs to match reality in the same PR.

Once the deleted tests are back I'm happy to approve — the rest is small stuff.

---

flcarre (commented):
The diagnosis here is better than most, and worth spelling out because it isn't obvious from the diff: `.block-editor-main` already had `padding-bottom`, but `scrollIntoView({block: "nearest"})` on the *last* card aligns its bottom edge to the scrollport's bottom edge and never scrolls into that padding — which is why the card ends up under the status bar. `scroll-padding-bottom` is the right fix, and it's genuinely a separate problem from the `center` alignment, since with the padding alone a palette jump would still park the target flush against the edge. Both changes are needed.

The `{align?: ScrollLogicalPosition}` option defaulting to `nearest` is three lines for one call site — proportionate, nothing to simplify. Design-system compliance is clean across every added line: no `element-plus`, no hex/`rgb()`/`--el-*`/`--bs-*`/SCSS `$`, no `:deep()`, no raw `px`, and the tokens used (`--ks-spacing-3/4/6`) are all declared. No listener, observer or timer is added, so there's nothing to clean up on unmount, and the canvas isn't virtualized so the `getBoundingClientRect`/`scrollIntoView` path is safe.

The e2e coverage is real, too — `Math.abs(targetCentre - scrollportCentre) < scrollport.height * 0.25` is two-sided so both `nearest` and `start` fail it, and `target.bottom <= statusBar.y` is exactly the regression. Selectors are all `data-test=` / `data-block-id=`.

## The unit spec doesn't test the change

`useCanvasFocus.spec.ts` stubs `card.scrollIntoView = vi.fn()` and then never asserts on it. Both tests only check that `focusedId` echoes the id just passed in, which is the pattern `ui/AGENTS.md` names explicitly ("don't test that a prop is passed through, that a computed returns its own input"). Net effect: the one behaviour this PR changes has no unit coverage, while the PR description says the spec pins that contract.

I saw the comment explaining that alignment is asserted in e2e because jsdom has no layout — that's fair for *alignment*, but asserting the **argument** needs no layout at all:

```ts
expect(card.scrollIntoView).toHaveBeenCalledWith({block: "nearest"})   // default
expect(card.scrollIntoView).toHaveBeenCalledWith({block: "center"})    // {align: "center"}
expect(card.scrollIntoView).toHaveBeenCalledWith({block: "nearest"})   // after moveFocus
```

That's the spec the description already promises. The two echo tests can go.

## The fallback re-duplicates the magic number the extraction removes

`develop` had `2.25rem` written 4 times. This PR declares `--status-bar-height: 2.25rem` and then writes `var(--status-bar-height, 2.25rem)` at all five use sites — 6 occurrences of the literal, up from 4. Change the height and five stale-but-plausible fallbacks silently survive.

The comment justifies them as keeping `BlockEditorStatusBar` correct standalone, but there's no story or standalone mount for it — it's only ever rendered inside `.block-editor`. Dropping to `var(--status-bar-height)` at all five sites removes the trap.

## Comment volume

Roughly 24 of the ~40 added source lines are comments: an 8-line JSDoc for one optional field, a 5-line paragraph explaining that CSS custom properties inherit, the same point restated a third time in `BlockEditorStatusBar`, and a prose restatement of what `{align: "center"}` already says. Our guideline is one sentence, only where a reviewer would otherwise have to ask *why*.

The one worth keeping is the `scroll-padding-bottom` rationale — the status-bar-overlaps-the-scrollport constraint is genuinely invisible from the code. Trimmed to a sentence, the others removed.

## Observation, not a request

`focusCanvasCard` isn't the single choke point the new option implies — there are five `scrollIntoView` sites in `no-code/blocks/`, and `useBlockOperations.ts` calls `focus.focusedCard()?.scrollIntoView(...)` directly, bypassing it. All pre-existing and this PR doesn't worsen it; worth a follow-up issue rather than scope creep here.

flcarre (approved):
Re-reviewed at `328ef607`. All three points from my earlier comment are addressed:

- **The unit spec now tests the change.** It asserts `scrollIntoView` was called with `{block: "nearest"}` by default, `{block: "center"}` under `{align: "center"}`, and `nearest` after `moveFocus` — plus a case proving nothing scrolls when there's nothing to focus. The two tests that only echoed `focusedId` back are gone. That's the contract the PR description promised, and it now has teeth.
- **The magic number is declared once.** `2.25rem` appears exactly once, and all five use sites read `var(--status-bar-height)` with no fallback — down from six literals. Changing the height can no longer leave stale-but-plausible values behind.
- **The comments are trimmed** to the one that earns its place.

The fix itself was already right, and worth restating for whoever reads this later: `.block-editor-main` had `padding-bottom`, but `scrollIntoView({block: "nearest"})` aligns the last card's bottom edge to the scrollport's bottom edge and never scrolls into that padding — so `scroll-padding-bottom` and the opt-in `center` alignment are two separate, both-necessary changes. The e2e coverage is real: the centring assertion is two-sided, and `target.bottom <= statusBar.y` is exactly the regression.

Approving. One thing before merge: the rebase produced a new SHA, which reset the fork workflow gate — I've just approved the runs, so please wait for them to come back green. Expect a red `OpenTelemetry - Export Trace`; it fails at *Set up job* on a missing `kestra-io/actions/actions/otel-collect@main` and has nothing to do with this change.

Thanks for turning this around so precisely.

MilosPaunovic:

MilosPaunovic:
Is this ready to be merged @flcarre?

PR #18118, opened 2026-08-13
fix(topology): align task runner badge with the task title

### LINKED ISSUE #17955: Alignement when a taskrunner is displayed on the Topology
### Describe the issue

Image

The runner should on the title area 

Image

### Environment

- Kestra Version: develop

Lkx-JY:
## What does this PR do?

Fixes the misalignment of the task runner badge on Topology nodes. When a task
had a `taskRunner` configured, its badge was rendered as a separate row *above*
the task title instead of sitting inline with it, pushing the title down and
breaking the vertical alignment of the node.

## Changes

- `ui/packages/topology/src/nodes/BasicNode.vue`: moved the `#badge` slot into
  the `.node-title` row so the badge renders on the same line as the task title
  (between the title and the status/actions).
- `ui/packages/topology/src/nodes/TaskNode.vue`: removed the now-unneeded
  `align-self: flex-start` from `.runner-badge` — it was written for the previous
  column layout; the title row's `align-items: center` now handles vertical
  centering.

## How to verify

Create a flow with a task that has a `taskRunner` (e.g.
`io.kestra.plugin.scripts.runner.docker.Docker`) and open the Topology view.
The runner badge should appear inline with the task title, on the same line,
instead of stacking above it.

Fixes #17955

### DIFF
--- a/ui/packages/topology/src/nodes/BasicNode.vue
+++ b/ui/packages/topology/src/nodes/BasicNode.vue
@@ -18,11 +18,11 @@
                             {{ displayTitle }}
                         </KsTooltip>
                     </div>
-                    <slot name="title-status" />
-                    <slot name="title-actions" />
                 </div>
                 <slot name="content" />
             </div>
+            <slot name="title-status" />
+            <slot name="title-actions" />
         </div>
         <slot name="details" />
     </div>
@@ -185,6 +185,7 @@
         flex-direction: column;
         justify-content: center;
         margin-left: 0.7rem;
+        margin-right: var(--ks-spacing-1);
         flex: 1;
         min-width: 0;
 
--- a/ui/packages/topology/src/nodes/BasicNode.vue
+++ b/ui/packages/topology/src/nodes/BasicNode.vue
@@ -144,6 +144,7 @@
             padding: var(--ks-spacing-2);
             padding-right: var(--ks-spacing-4);
             align-items: center;
+            gap: var(--ks-spacing-1);
             width: 218px;
             height: 56px;
         }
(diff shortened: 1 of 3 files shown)


RemiBarthe (changes requested):
Thanks for digging into this !

It's not quite the result we're after though. The badge staying above the title is intentional and matches the design. What's misaligned in the screenshot is the status row: 1m, 11.97s sits ~6px lower than 0.37s and 0.03s on the nodes. That's what the red line points at.

Moving the badge into the title row also costs too much width: `.node-title` is only ~220px in an execution and already carries the title, the status tag and the kebab menu. The title drops from 130px to 43px, so produce becomes prod… and the badge itself clips to GCP Cloud….

Suggested direction: leave the badge where it is, and move the` title-status `/ `title-actions` slots out of `.node-content` so they become direct children of `.main-content`. It's already `display: flex; align-items: center`, so the status tag and menu centre on the node's full height and stop caring whether a badge sits above the title. Two lines, and the title loses nothing.

One thing to watch out for: the current spacing around the title comes from the status tag's 0.4rem padding inflating .node-title to ~23px, plus that row's gap. Take the status tag out and both disappear: the badge ends up flush against the title, and the title's ellipsis ends up flush against the status tag. Worth restoring explicitly, e.g. a `margin-bottom: var(--ks-spacing-1)` on `.runner-badge` and a` margin-right: var(--ks-spacing-1) `on` .node-content`.

Thanks again!

Lkx-JY:
@RemiBarthe thanks for the detailed feedback — I've reworked the fix to follow your suggested
  direction:

  - The badge now stays above the title (reverted the "badge inline" change).
  - `title-status` / `title-actions` slots are now direct children of `.main-content`, so they center
  on the node's full height.
  - Restored the spacing with `margin-right: var(--ks-spacing-1)` on `.node-content` and
  `margin-bottom: var(--ks-spacing-1)` on `.runner-badge`.
  - Added a Vitest test (`BasicNode.test.ts`) asserting the slot placement.

  Could you take another look when you have a chance?

RemiBarthe (approved):
LGTM, thanks again for the contribution! 🚀

PR #18119, opened 2026-08-13
fix(executions): build every execution from the flow processed for runtime

fhussonnois:
### ✨ Description

Every route that creates an execution now builds it from the flow **processed for runtime** — plugin defaults injected and, in the Enterprise Edition, governance policies applied — instead of the flow as authored.

An `Execution` snapshots the flow's `labels` and `variables` at creation time. Resolving the authored definition therefore dropped everything a runtime parse adds, on every route except the ones a trigger drives. And where authored labels *were* carried, they won: `ExecutionService.create` appends the trigger's labels last, `Label.deduplicate` keeps the last occurrence, and `LabelService.fromTrigger` folded the flow's authored labels into that list — so the authored value overrode the value plugin defaults or governance had resolved.

**Changes**

| Area | Change |
|---|---|
| `core` — `FlowMetaStoreInterface`, `DefaultFlowMetaStore` | Adds `findByIdThenInjectDefaults` and its `findByIdFromTask` sibling (which keeps the namespace-allowed check). The meta-store could only hand back a runtime-processed flow once an execution existed, while every path creating one needs that flow *before* it exists. Both memoize on the resolved flow UID rather than on the requested identifier, which is what makes one cache safe for every entry point. |
| `executor` — `ExecutionCommandMessageHandler` | A `Create` command resolves its flow with `findByIdThenInjectDefaults`, on the creation and the replay paths alike. |
| `core` — `ExecutableUtils`, `SubflowFunction` | A subflow task resolves its child with `findByIdFromTaskThenInjectDefaults`. The Pebble `subflow()` function goes through it too, which also surfaces a governance rejection there rather than as a created-then-failed execution. |
| `executor` — `FlowTriggerService`, `plugin.core.trigger.Flow` | Both flow-trigger routes evaluated the flow as authored. The standard-condition one sends the execution straight to the queue, so it carried neither an injected label nor a flow variable at all; the `dependsOn` one goes through a `Create` command but hand-built its labels. Both now resolve the processed flow — only once a trigger is about to fire, so matching still runs on the authored definition — and build through `Execution.newExecution` rather than setting the field. |
| `webserver` — `ExecutionController`, new `FlowLabelsResolver` | `parseLabels` takes the flow and delegates to `FlowLabelsResolver`, the single place deciding whether a label passed when starting an execution applies to the target flow. Identity in OSS — callers keep overriding flow labels and adding their own; editions replace it to restrict. The webhook route resolves the processed flow for its labels, variables and trigger evaluation (the key is still matched on the authored flow, so an unknown key stays a 404). |
| `core` — `LabelService` | Adds `fromTriggerOnly`, returning just the trigger's labels, and deprecates `fromTrigger` for removal: `Execution.newExecution` is left as the single place a flow contributes its own labels. |
| `core`/`scheduler` — `TriggerService`, `TriggerEvaluationResult`, `MetricRegistry`, `TriggerScheduler` | `TriggerService` added the flow's labels a second time on top of the trigger's, so `TriggerEvaluationResult` and every realtime trigger send carried them twice. Removing that dropped them from the `scheduler.trigger.count` and `scheduler.trigger.delay.duration` tags, so those are now passed explicitly — trigger labels first, so a trigger still wins. |
| `worker` — `GrpcWorkerFlowMetaStore` | A dedicated worker resolves flow metadata over gRPC and has no parsing service, so the new lookups throw `UnsupportedOperationException` there, as the existing runtime-processed lookup already does. |

**Tests**

- `DefaultFlowMetaStoreTest` — the new lookups return the processed flow and memoize on the resolved UID.
- `FlowTriggerServiceTest` — both flow-trigger routes carry the processed flow's labels and variables.
- `LabelServiceTest`, `MetricRegistryTest`, `FlowTest`, `ScheduleTest`, `WebhookPluginTest`, `ExecutionCommandMessageHandlerTest` — updated or extended for the above.

### 🔗 Related Issue

Refs https://github.com/kestra-io/kestra-ee/issues/9611 — this is the OSS half of the fix for the first bug reported there (label injection skipped on executions that are not trigger-driven). The `labels: {}` behaviour reported as the second bug in that issue is untouched here, so the issue stays open.

### 🛠️ Backend Checklist

### 📝 Additional Notes

- **Companion PR:** https://github.com/kestra-io/kestra-ee/pull/9873 — same branch name. The Enterprise Edition side adds the policy-aware `FlowLabelsResolver` and the end-to-end coverage of each creation route under governance. Please merge them together: without the EE PR the new `FlowLabelsResolver` seam is simply identity, and without this PR the EE tests fail.
- `LabelService.fromTrigger` is deprecated rather than removed since it is public API reachable from plugins; it now delegates to `fromTriggerOnly` and is unused in this repo.
- The webhook route deliberately keeps its response body reporting the flow's labels — that body is public-facing, whereas policy-injected labels are internal.

### ⚠️ Known gap — flow triggers evaluate the authored trigger

Every other route now evaluates the trigger as the flow processed for runtime carries it: schedule, polling and realtime take it from the parsed flow in `DefaultSchedulableTriggerFetcher`, and the webhook route re-resolves it through `ExecutionController#processedForRuntime`. `FlowTriggerService` does not — it passes the resolved flow to `evaluate`, but the trigger instance itself still comes from the authored definition, so a policy changing a flow trigger's own configuration is ignored there. Only its labels and variables are governed, which is what this PR fixes.

A blocked flow is also handled three different ways: the scheduler skips the trigger (`TriggerFlowParser#parseOrSkip` returns `null`), a flow trigger still creates the execution and lets the executor fail it, and the webhook route falls back to the authored trigger for the edition's gate to reject. Both are follow-ups, not addressed here.

### DIFF
--- a/core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
+++ b/core/src/main/java/io/kestra/core/metrics/MetricRegistry.java
@@ -19,6 +19,7 @@ import io.kestra.core.runners.SubflowExecutionResult;
 import io.kestra.core.runners.WorkerTask;
 import io.kestra.core.runners.WorkerTaskResult;
 import io.kestra.core.runners.WorkerTrigger;
+import io.kestra.core.utils.ListUtils;
 import io.kestra.core.worker.WorkerGroups;
 import io.kestra.core.worker.WorkerQueues;
 
@@ -557,13 +558,23 @@ public class MetricRegistry {
         return ArrayUtils.addAll(ArrayUtils.addAll(baseTags, labelTags), tenantTag);
     }
 
-    public String[] tags(TriggerEvaluationResult evaluationResult, TriggerId triggerId) {
+    /**
+     * Return tags for a trigger evaluation, tagged like the execution it produces.
+     *
+     * @param evaluationResult the evaluation result, carrying the trigger's own labels
+     * @param triggerId the evaluated trigger
+     * @param flowLabels the labels of the flow owning the trigger, which the evaluation result does not carry
+     * @return tags to apply to metrics
+     */
+    public String[] tags(TriggerEvaluationResult evaluationResult, TriggerId triggerId, @Nullable List<Label> flowLabels) {
         var baseTags = new String[] {
             TAG_FLOW_ID, triggerId.getFlowId(),
             TAG_NAMESPACE_ID, triggerId.getNamespace(),
             TAG_STATE, evaluationResult.stateType().name(),
         };
-        var labelTags = getLabelTags(evaluationResult.labels() != null ? evaluationResult.labels() : List.of());
+        // trigger labels first: getLabelTags keeps the first match, which is how the trigger overrides the
+        // flow on the execution these metrics describe (there, flow labels come first and dedup keeps the last)
+        var labelTags = getLabelTags(ListUtils.emptyOnNull(ListUtils.concat(evaluationResult.labels(), flowLabels)));
         var tenantTag = getTenantTag(triggerId.getTenantId());
         return ArrayUtils.addAll(ArrayUtils.addAll(baseTags, labelTags), tenantTag);
     }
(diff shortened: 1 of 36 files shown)


ammeek (approved):
LGTM

AcevedoR (commented):
pluginDefaults codebase was already a bit messy in the past, that's why I think this seems a bit all over the place ^^'

AcevedoR (approved):
LGTM, because I know it will be improved in follow up PRs

Issue #18094, opened 2026-08-12
[2.0] Page goes blank after deleting a revision


wrussell1999:
### Describe the issue

Steps to reproduce:
1. Create a flow and make a number of changes, saving in between each change
2. Go to the revisions tab
3. Select the drop down to change revisions
4. Delete one of the revisions
5. Screen goes blank. Needs to be refreshed to see revisions again. 

https://github.com/user-attachments/assets/c1349c56-6fa0-4101-b2cb-6f7321d9708a

### Environment

- Kestra Version: 2.0 RC7 EE

wrussell1999:
Seems to only break when you delete from the right side. If you start from the left, it didn't happen?

https://github.com/user-attachments/assets/bdc133bb-5345-4610-9ec3-b52c33998b24

n200534:
Hi! I'm interested in working on this issue. If no one is currently working on it, could you please assign it to me?

MilosPaunovic:
Hi @n200534,  

You're already assigned to another issue in our repository, and we usually assign only one at a time to new contributors. Let's make sure to wrap that one up before jumping onto another.  🚀

ameyypawar:
Would like to take this on once my current PRs are wrapped up :)

n200534:
i have assigned another task ,currently i am working on that but i have interest in working with this issue if maintainers assigns to me , i will take on this :)

MilosPaunovic:
@ameyypawar You still have one PR opened, until that one is closed, I will not be able to assign you to new issues.

Tejaskrishnan:
Hi @MilosPaunovic , 
I'd like to work on this issue if it's still open.

MilosPaunovic:
Absolutely, go for it @Tejaskrishnan, thanks! 🚀

MilosPaunovic:
Hey @Tejaskrishnan, are there any updates on this?

Tejaskrishnan:
Hi @MilosPaunovic ,
Yes, I’m working on it. I’ll create a pull request at the earliest.

jasperjonkhans:
Is this issue still available? If yes, then I'd like to work it. Thanks!

Pradumnasaraf:
Working. QA done ✅

<img width="1278" height="1080" alt="Image" src="https://github.com/user-attachments/assets/0382808d-29cf-4cd6-a2d0-197ceeab300c" />

Pradumnasaraf:
[...]

PR #18125, opened 2026-08-13
fix(flows): keep the revisions view alive after deleting a revision

ameyypawar:
## Summary

- Deleting a revision blanked the page until a reload. Both sides of the diff are stored as **positions in a derived, sorted list**, so removing an entry shortens that list and the stored positions no longer mean what they did.
- Where a position fell past the end, `revisionNumber()` read `.revision` off `undefined`. That happened during render — it is called for the `crud` slot — so Vue tore the subtree down and left an empty page. Where a position stayed in range it silently repointed at a different revision, which is the same bug without the crash.
- Both sides are now re-anchored **by revision number** whenever the list changes. A side whose revision survived follows it; a side whose revision was the one deleted holds its position, which after the removal belongs to the neighbour that shifted into it.
- `onDelete` no longer calls `load()`. The parent refreshes asynchronously (`await router.push` → `await fetchRevisions`), so that call only ever saw the pre-delete list.
- `revisionNumber()` returns `undefined` out of range rather than throwing, so no future caller can blank the view the same way.

Related: #18094

## Test plan

- [x] `ui/tests/unit/components/layout/Revisions.spec.ts` drives the list changing under the diff. Three of its eight cases were **confirmed to fail against a build without this change**, with the production error `TypeError: Cannot read properties of undefined (reading 'revision')`.
- [x] Covers removing an unrelated revision, removing the one on show, losing the newest, dropping to a single revision (empty state), opening from `revisionLeft`/`revisionRight` in the URL, and a revision being **added** — the same watcher fires on save, where the newest must still win.
- [x] `npm run test:unit`: 1544 passing, 0 failing (+8, exactly the new spec).
- [x] `npm run check:types` passes for the design-system, app and test projects.
- [x] `npm run test:lint`: 0 errors on the touched files.
- [x] `npm run test:storybook`: 65 files / 239 tests green, including this component's own `Revisions.stories.tsx`.

> Two notes for review. The reporter observed this "only when you delete from the right side" — the mechanism here is symmetric, and what decides whether you get a blank page rather than a silently wrong diff is whether the stored position lands past the end of the list; the right selector defaults to the last entry, which fits, but I did not reproduce the exact asymmetry. And there is no e2e for the revisions view to extend, so the coverage above is unit plus the existing stories.

### DIFF
--- a/ui/src/components/layout/Revisions.vue
+++ b/ui/src/components/layout/Revisions.vue
@@ -40,7 +40,7 @@
                         <KsButtonGroup>
                             <KsButton
                                 :icon="Restore"
-                                :disabled="revisionLeftText === currentRevisionWithSource.source"
+                                :disabled="revisionLeftText === currentRevisionWithSource?.source"
                                 @click="restoreRevision(revisionLeftIndex, revisionLeftText)"
                                 data-testid="restore-left"
                             >
@@ -82,7 +82,7 @@
                         <KsButtonGroup>
                             <KsButton
                                 :icon="Restore"
-                                :disabled="revisionRightText === currentRevisionWithSource.source"
+                                :disabled="revisionRightText === currentRevisionWithSource?.source"
                                 @click="restoreRevision(revisionRightIndex, revisionRightText)"
                                 data-testid="restore-right"
                             >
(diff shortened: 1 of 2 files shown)


MilosPaunovic:
Hi there, @ameyypawar! 👋

PR #18127, opened 2026-08-13
feat(logs): forward-only cursor pagination for the logs UI

nkwiatkowski:
## What

Makes the logs list UI handle **cursor-based (forward-only) pagination** when the backing log data store advertises `type: CURSOR`, while leaving the existing offset/numbered-page behavior untouched for normal stores (JDBC/Elasticsearch).

External log data stores such as GCP Cloud Logging and Datadog expose only an opaque next-page cursor — no random-access offset and no exact total — so the log page must adapt:

- The store decides the mode from the **first response** (it cannot know beforehand, since the UI may deep-link/restore page N).
- In cursor mode the UI shows a forward-only **Next** control instead of the numbered pager (`KsDataTable` hides its numbered pager on its own when `total` is absent), and a refresh/filter-change returns to the first page.
- The per-level count badges and the top timeseries chart are hidden in cursor mode for non-aggregating stores (there is no total to count against).

## How

Frontend-only; all mode logic lives in the Pinia logs store and is unit-tested:

- `ui/src/stores/logs.ts` — capture `type` + `nextCursor` from every response; expose `paginationType`, `nextCursor`, `isCursorMode`, `hasNextCursor`; thread an optional `cursor` through `findLogs`.
- `ui/src/components/logs/LogsWrapper.vue` — render the Next control in cursor mode; skip the total-dependent level badges and timeseries chart.
- `ui/tests/unit/stores/logsCursor.spec.ts` — covers offset capture, cursor capture, cursor pass-through, cursor omission on first page/refresh, `nextCursor` clearing on the last page, and the cursor→offset mode-transition reset.

No new i18n keys (reuses `next` / `refresh`).

## Notes

- The **backend** for this (`PaginationType`, `LogDataStoreInterface.paginationType()`, `CursorOrOffsetPagedResults`, and the webserver log-search override in #9463) is already on `develop`; this PR is the UI counterpart.
- Enterprise `ui-ee` reuses these OSS files unchanged, so no companion EE change is needed.
- Validated live against a real Datadog (CURSOR) store: forward-only Next walk, `total` absent, filters/prefix-scoping intact.

Part of https://github.com/kestra-io/kestra-ee/issues/7054


MilosPaunovic:
Reviewed the UI. Two things I'd want changed before merge, plus some smaller notes.

**1. "Next" on the last page of data dumps you into an empty state with no way back.**

The backend contract is explicit - `CursorOrOffsetPagedResults.java:43`: *"A non-empty page always carries a `nextCursor`, so the client keeps paging until it gets back an empty page (not until `nextCursor` is null)."* And `of()` only leaves `nextCursor` null when the page is empty.

So the last page *containing logs* still carries a cursor, `hasNextCursor` is true, and Next renders. Clicking it sets `logs.value = []`, the template falls into the `v-else-if="!isLoading"` branch and shows `KsNoData`, and `nextCursor` clears so Next disappears too. The logs are gone, there is no Previous, and `loadNext` never touches the URL so browser Back leaves the page entirely. Only escape is refresh, which restarts at page 1.

Suggestion: when a cursor `Next` returns zero results, keep the current rows and just drop the Next control instead of replacing `logs` with `[]`.

**2. No way back to the previous page at all.**

Forward-only is a backend constraint, not a client one - the client already knows every cursor it has used. A small stack (`[undefined, "tok-1", "tok-2", …]`) gives you a working Previous plus a "page N" indicator for ~10 lines in the store. Without it, and combined with #1, the only recovery from any misstep is a full reset to page 1.

**Should fix**

3. Cursor mode kills the level quick-filters, not just the counts. `refreshLevelCounts` early-returns, so `serverLevelCounts` stays `{}`, so `presentLevels` (`count > 0`) is empty, so the whole `LogLevelNavigator` row disappears - and those chips are the level *filter* affordance (`@select="selectLevel(level)"`), not just badges. Rendering them without counts needs `totalCount?: number` in `LogLevelNavigator`; if that's out of scope, worth stating explicitly.

4. The chart toggle stays visible in cursor mode and does nothing - `showStatChart()` is gated on `!isCursorMode` but `chart: {shown: true, …}` isn't.

5. No component test. Every behaviour that can go wrong (Next visibility, the dead end above, chart/badge hiding) lives in `LogsWrapper.vue` and is untested; all six specs cover the store. One mounted spec would have caught #1, #3 and #4.

**Minor**

- `loadNext` bypasses `KsDataTable`, so `@loaded` never fires - the scroll-to-bottom that numbered pages get doesn't happen on Next.
- URL `page` becomes a lie in cursor mode: `loadData` still sends it on first load, `loadNext` never updates it, so `?page=3` renders page 1.
- `response.type ?? "OFFSET"` is dead - `type` is `@NotNull` and non-optional in the generated type.
- `:aria-label="t('next')"` just duplicates the button's visible text.
- `t("next")` collides with #18223 (`t()` → `$t()` sweep) - merge-order touch-up for whichever lands second.
- Description nit: `total` is never absent, the store coerces it to `0`; the pager hides because `showPagination` requires `total > minPageSize`.
- `downloadLogs` fetches one page with no cursor, so downloads silently truncate for cursor stores.

**Verified good:** `cursor` is a real declared query param so it survives `buildClientParams`; `PaginationType` matches the store's ref type; pager correctly hidden at `total: 0`; filter-change and refresh both correctly re-fetch without a cursor; EE reuse really is a no-op (`Actions.vue:250` reads only `logsStore.logs`).

**CI:** `Frontend - Tests` is red but not from this PR - 133/133 passed including `logsCursor.spec.ts`; the job died on an unrelated storybook browser flake (`ErrorToastContainer.stories.tsx`, "Browser connection was closed"). Needs a rerun.

MilosPaunovic:
Re-reviewed at `fad54ad4ea`. CI is all green and the blocking issues are fixed.

**One thing still open:**

The `LogLevelNavigator` row disappears in cursor mode - `refreshLevelCounts()` early-returns on `isCursorMode` (line 394), so the entire level filter affordance vanishes. The level filter is useful in cursor mode too (filters don't care about pagination backend). Either enable the counts in cursor mode (one backend call per filter change, data's already there) or render the navigator without counts (`totalCount?: number` on `LogLevelNavigator`). The component test doesn't cover this, so it didn't surface.

**Minor:** URL `page` becomes a lie, dead fallback `?? "OFFSET"`, redundant `aria-label`, downloads truncate to first page, no scroll-to-bottom on Next.

Ship-ready if you're comfortable skipping the level navigator refinement; one more touch to render them without totals would polish it.

MilosPaunovic (approved):
Approved at `093fa9f79a`. Both round-1 blockers are fixed (cursor back-stack with `Previous`, empty last page keeps the current rows), the chart toggle is gated, and the level-chip hiding in cursor mode is now explicitly locked by tests with a sound rationale - per-level counts do not exist on cursor stores, and level filtering stays available via the main filter bar. No backport needed (`CursorOrOffsetPagedResults` exists on `develop` only) and no `EE` companion required.

PR #18131, opened 2026-08-13
fix(core): resolve pre-existing serialization and lifecycle issues

### LINKED ISSUE #18122: Pre-existing serialization, lifecycle, and error-handling issues surfaced by the Micronaut 5 / Jackson 3 migration
While migrating to Micronaut 5 (which moves the HTTP layer to Jackson 3), five pre-existing problems surfaced that are **independent of that migration**. They were deliberately left out of its scope; this issue records them so they are not lost.

None of them is a regression introduced by the Micronaut 5 work — each is reachable on `develop` today.

---

## 1. `MarkdownSource` serializes a duplicate `type` property

`core/src/main/java/io/kestra/plugin/core/dashboard/chart/mardown/sources/MarkdownSource.java` declares:

```java
@JsonTypeInfo(
    use = JsonTypeInfo.Id.NAME,
    property = "type",
    visible = true
)
public class MarkdownSource {
    private String type;   // ← also serialized, by @Getter
}
```

`include` is not set, so it defaults to `As.PROPERTY`, meaning Jackson **writes the type id as its own `type` property** — while the class *also* exposes a `type` field. A serialized markdown source therefore contains `type` twice.

It is the only `@JsonTypeInfo` site in `core` that both declares its own `type` member and uses the default `As.PROPERTY`; the other ~20 hierarchies use `include = As.EXISTING_PROPERTY` with `visible = true`, which makes Jackson reuse the existing field instead of emitting a second one.

**Why it matters.** Jackson 2 tolerates the duplicate silently (last value wins), so this has been invisible. Jackson 3 takes a duplicate-property code path instead — during the migration it manifested as an opaque client-side failure decoding a dashboard response:

```
Error decoding HTTP response body: Cannot invoke
"tools.jackson.core.util.JacksonFeatureSet.isEnabled(...)" because "this._readCapabilities" is null
 ... (through reference chain: DashboardController$DashboardResponse["charts"]->ArrayList[0])
```

Emitting a duplicate key is also not strictly valid JSON for consumers that reject it.

**Reproduce.** `DashboardControllerTest.shouldRejectExportOfMarkdownChart` exercises the path (a dashboard with `type: io.kestra.plugin.core.dashboard.chart.Markdown` and a nested `source: {type: Text}`). Reproduced live against a running instance too: `POST /api/v1/{tenant}/dashboards` with that exact shape returns `200` with a raw response body literally containing `"source":{"type":"Text","type":"Text","content":"..."}` — a genuine duplicate JSON key, confirmed by counting occurrences in the raw bytes (a JSON parser that keeps last-value-wins, like Python's or Jackson 2's, hides it). `GET` and `charts/preview` on the same chart do not crash, confirming the migration's read-side robustness fix holds; the write-side duplicate itself is unfixed, as intended.

**Suggested fix.** Add `include = JsonTypeInfo.As.EXISTING_PROPERTY` to `MarkdownSource`, matching the convention used by every other polymorphic model. Worth asserting the serialized form contains `type` exactly once.

---

## 2. The triggers CSV export includes `tenantId`, unlike the executions and flows exports

`TenantSerializer` strips `tenantId` from anything implementing `TenantInterface` on its way out of the API. `TriggerState` (`core/src/main/java/io/kestra/core/scheduler/model/TriggerState.java:32`) has a `tenantId` field but implements `TriggerId`, **not** `TenantInterface` — so the modifier does not apply to it.

Because `CSVUtils.toCSVFlux` derives the header from the first row's key set, `GET /api/v1/{tenant}/triggers/export/by-query/csv` emits a `tenantId` column, whereas the executions and flows exports (whose entities do implement `TenantInterface`) do not.

**Why it matters.** This is primarily an **inconsistency** between the three exports rather than a cross-tenant disclosure: the rows returned are already scoped to the caller's tenant, so the value exposed is a tenant id the caller is operating within. Still worth deciding deliberately, since:

- the three CSV exports should plausibly agree on whether `tenantId` is a column, and
- the exports are downloadable artifacts that may be shared beyond the person who generated them.

**Suggested fix.** Decide whether `TriggerState` should implement `TenantInterface` (and so be stripped like the others), or whether the column is intentional here. Either way, a test asserting the chosen behaviour would pin it — none of the three exports asserted anything about `tenantId` until recently.

---

## 3. `KestraContext.getPluginRegistry()` throws `NullPointerException` after the context is closed

`core/src/main/java/io/kestra/core/contexts/KestraContext.java` (`Initializer#getPluginRegistry`):

```java
public PluginRegistry getPluginRegistry() {
    return this.applicationContext.getBean(PluginRegistry.class);
}
```

A `KestraContext.Initializer` left behind by a closed application context has a null `applicationContext`, so this throws `NullPointerException` rather than a meaningful `IllegalStateException`. `getStorageInterface()` immediately below has the same shape.

Callers reasonably guard on the documented failure mode instead — `PluginDeserializer.checkState()` catches `IllegalStateException | NoSuchBeanException` to fall back to a default registry, with the comment *"This error can only happen if the KestraContext is not initialized (i.e. in unit tests)"*, which is precisely the case that NPEs.

**Why it matters.** It produces an ordering-dependent test failure: `TaskRunTest` failed inside the full `:core:test` run while passing in isolation, because another test in the same JVM had closed a context first. The NPE surfaced as a confusing `JsonMappingException ... (was java.lang.NullPointerException)` several frames from the cause.

**Current state.** The Micronaut 5 branch widened both plugin deserializers' catch to include `NullPointerException` as a stopgap, so the flake is suppressed. That is a workaround at the call site, not a fix.

**Suggested fix.** Have `KestraContext.Initializer` null-check `applicationContext` and throw `IllegalStateException` (both in `getPluginRegistry()` and `getStorageInterface()`), then narrow the deserializers' catch back to the documented exception types. See also the related hazard around static `KestraContext` lifetime across test contexts.

---

## 4. `ApiExecution`, `ApiLightExecution` and `WebhookResponse` leak `tenantId`, unlike the domain `Execution` model

`TenantSerializer` strips `tenantId` from anything implementing `TenantInterface` on its way out of the API —
this existed as a Jackson 2 `BeanSerializerModifier` on `develop` already, and was ported to a Jackson 3
`ValueSerializerModifier` for the Micronaut 5 branch with identical semantics.

`webserver/src/main/java/io/kestra/webserver/models/api/ApiExecution.java`,
`webserver/src/main/java/io/kestra/webserver/models/api/ApiLightExecution.java` and
`core/src/main/java/io/kestra/plugin/core/trigger/WebhookResponse.java` are separate DTOs that wrap `Execution`
for the public API. All three declare their own `tenantId` component and populate it directly:

```java
public record ApiExecution(@NotNull String tenantId, @NotNull String id, ...) {
    public static ApiExecution of(Execution execution) {
        return new ApiExecution(execution.getTenantId(), execution.getId(), ...);
    }
}
```

Neither implements `TenantInterface`, so the modifier's `TenantInterface.class.isAssignableFrom(beanDesc.getBeanClass())`
check never matches, and `tenantId` is serialized with its real value (confirmed live: `GET
/api/v1/main/executions/{id}` returns `"tenantId":"main"`).

This is a pre-existing gap, not a Micronaut 5 regression: `ApiExecution` and `ApiLightExecution` are both
untouched by the migration branch (`git log origin/develop..HEAD` on either file is empty), and the Jackson 2
`TenantSerializer` on `develop` has the exact same `TenantInterface`-only scope.

**The convention is documented elsewhere in the same package.** `ApiTriggerState`
(`webserver/src/main/java/io/kestra/webserver/models/api/ApiTriggerState.java`) has no `tenantId` field at all,
and its class javadoc says so explicitly:

> Excludes internal scheduler fields (`tenantId`, `vnode`, `lastEventId`).

So `ApiExecution`/`ApiLightExecution` are the outliers against the team's own stated intent, not an
alternative convention.

**Reachable via:**
- `GET /api/v1/{tenant}/executions/{executionId}` → `ApiExecution`
- `GET /api/v1/{tenant}/executions/search` and the flow-scoped execution search (`ExecutionController.searchExecutions`
  / `searchExecutionsByFlowId`) → `ApiLightExecution`, so every execution list/search response is affected too.
- `POST /api/v1/{tenant}/executions/webhook/{namespace}/{id}/{key}` → `WebhookResponse`, confirmed live including
  through the `multipart/form-data` variant. This one is public and unauthenticated by design (the webhook key is
  the credential), so it is the most exposed of the three.

**Why it matters.** In OSS, with a single `main` tenant, this is closer to an inconsistency than a disclosure —
similar in character to item 2 above. In EE, with real multi-tenancy, every execution response broadcasts which
tenant it belongs to, which is at minimum inconsistent with `ApiTriggerState`'s documented exclusion and with
every other tenant-scoped API response.

**Suggested fix.** Simplest: drop the `tenantId` field from all three records, matching `ApiTriggerState`'s
convention, if no consumer needs it. Alternatively, have them implement `TenantInterface` — note a record's
canonical accessor is `tenantId()`, not `getTenantId()`, so satisfying the interface needs an explicit bridging
method, not just `implements TenantInterface`. Either way, a test asserting `tenantId`'s absence (or documented
presence) would pin the decision, the way item 2 suggests for the CSV exports.

---

## 5. A missing required multipart `Part` returns `500` instead of `4xx`

`io.micronaut.web.router.exceptions.UnsatisfiedPartRouteException` — the multipart sibling of
`UnsatisfiedBodyRouteException` and `UnsatisfiedQueryValueRouteException`, both already handled by
`webserver/src/main/java/io/kestra/webserver/controllers/ErrorController.java` — has no handler of its own, so
it falls through to the generic `Throwable` handler:

```java
@Error(global = true)
public HttpResponse<JsonError> error(HttpRequest<?> request, UnsatisfiedBodyRouteException e) {
    return jsonError(request, e, HttpStatus.UNPROCESSABLE_ENTITY, "Invalid route params");
}
// no handler for UnsatisfiedPartRouteException
```

Confirmed live: `POST /api/v1/{tenant}/namespaces/{namespace}/files?path=...` with no `fileContent` multipart
part returns:

```
500 {"message":"Internal server error: Required Part [fileContent] not specified", ...}
```

`io.micronaut.web.router.exceptions.UnsatisfiedPartRouteException` already exists, unhandled, in Micronaut
4.10.26 — this is not a Micronaut 5 change, and `develop`'s `ErrorController` has the same gap (it only imports
`UnsatisfiedBodyRouteException` and `UnsatisfiedQueryValueRouteException`).

Any `@Part`-annotated required multipart parameter across the webserver is affected, not just namespace file
upload — e.g. flow validation's multipart path (`FlowController.validateFlows`), execution inputs, and app
imports.

**Suggested fix.** Add a handler mirroring the existing two:

```java
@Error(global = true)
public HttpResponse<JsonError> error(HttpRequest<?> request, UnsatisfiedPartRouteException e) {
    return jsonError(request, e, HttpStatus.UNPROCESSABLE_ENTITY, "Invalid route params");
}
```

hemantmm:
closes: #18122

---

### ✨ Description

What does this PR change?  

This PR resolves three pre-existing bugs that surfaced during the Micronaut 5 / Jackson 3 migration. 

- Fixed duplicate `type` serialization in `MarkdownSource` by using `JsonTypeInfo.As.EXISTING_PROPERTY`.
- Added a regression test ensuring the serialized Markdown source contains the `type` property only once.
- Make `TriggerState` implement `TenantInterface` so `tenantId` is handled consistently with other tenant-scoped API exports.
- Added a regression test ensuring trigger CSV exports do not contain `tenantId`.


core/src/main/java/io/kestra/core/scheduler/model/TriggerState.java
```diff
@@ -29,7 +30,7 @@
 @EqualsAndHashCode
 @AllArgsConstructor
 @Builder
-public final class TriggerState implements TriggerId {
+public final class TriggerState implements TriggerId, TenantInterface {
```
loicmathieu: @fhussonnois without this, the TriggerState contains the tenantId when read from the API, no other object contains it. But I'm not sure if it has any implication inside the Scheduler WDYT?

loicmathieu (approved):
LGTM if @fhussonnois is OK for the change inside the `TriggerState`

core/src/main/java/io/kestra/core/scheduler/model/TriggerState.java
fhussonnois: The `TriggerState` is not exposed by API, All controllers API only expose an `ApiTriggerState` object which doesn't contain the tenant-id. So as long as the  `TenantSerializer` is only used by Micronaut for the API it's should ok.

fhussonnois (changes requested):
Implementing `TriggerInterface` on `TriggerState` is valid. That said, the Export API shouldn't expose `TriggerState` because it leaks internal state. We should expose `ApiTriggerAndState` instead.

hemantmm:
Thanks for the clarification. That makes sense. I'll revert the `TenantInterface` change from `TriggerState` and update the export to expose `ApiTriggerAndState` instead, keeping the regression test to verify that `tenantId` is not included in the CSV.

loicmathieu (approved):
LGTM, thanks

PR #18132, opened 2026-08-13
build(script): resolve kotlp as a Maven dependency instead of downloading it

AcevedoR:
## What

Deletes the hand-rolled `downloadKotlp` task and resolves the binary as an ordinary dependency:

```groovy
// platform/build.gradle
api 'io.kestra:kotlp:0.1.2!!'

// script/build.gradle
implementation 'io.kestra:kotlp'
```

`KotlpUtils` is unchanged: the artifact carries the APE at `/kotlp/kotlp` — the exact classpath resource it already reads — plus `/kotlp/kotlp.version`, so the binary now comes from the dependency jar instead of `script/src/main/resources/`. Also gone: the `kotlpVersion` property, the version marker file and its `.gitignore` entry.

## Why

`downloadKotlp` fetched the GitHub release asset at build time, and GitHub throttles **anonymous** release-asset downloads by closing the connection with no HTTP response — 3 of 10 cold JVM attempts failed from one IP, and `HttpURLConnection` doesn't retry a fresh connection, so it became a hard build failure (`Unexpected end of file from server`). It hit kestra-ee constantly because its self-hosted runners share one egress IP.

As a dependency, Gradle handles it: caching (a warm cache means **no download at all** — the old output sat outside every cache, so every CI job re-fetched it and burned the anonymous budget), checksums, resolution retries, offline builds, and Central's CDN. #18126 added a retry loop to survive the throttle; this removes the failure mode instead and deletes that code.

It also drops the `kotlpVersion` stopgap kestra-ee had to add to its own `gradle.properties` just to configure this OSS module — the version is BOM-managed now.

## Pinning

Strict (`!!`) because the binary's OTLP envelope key must match what `TaskLogLineMatcher` parses — a mismatch emits valid telemetry that Kestra ingests none of, silently. Verified against a locally published `0.2.0`: a competing `implementation 'io.kestra:kotlp:0.2.0'` is **downgraded** to the pinned version and the build succeeds (strict wins; it does not fail on conflict).

## Testing

Verified against a locally published artifact, then end to end on running servers:

- `:script:test` + `TaskLogLineMatcher`: **63 tests, 0 failures**, including the real-process kotlp wrap test (binary present and executable in the working dir, `process.cpu.time` reaching the run context) and the disabled case leaving the command unwrapped;
- the module jar no longer embeds the binary; **both distributions still do** — `shadowJar` for OSS and kestra-ee each contain `kotlp/kotlp` (`sha256 7f3cdea4…`, byte-identical to the released asset) and `kotlp/kotlp.version`;
- a **live OSS server** resolved it on its real classpath, and a **licensed local EE server** extracted and executed the binary straight from its fat jar (`kotlp --version → kotlp 0.1.0`);
- published POM: `io.kestra:kotlp`, `scope=runtime`, versionless via the imported `io.kestra:platform` BOM — same shape as `docker-java` and every other dependency.

Known coverage gaps found while testing, both pre-existing and worth a follow-up: no test asserts the child's stdout survives the wrap, and none asserts a non-zero exit code is proxied. I verified both by hand (`exit 42` propagates; the stdout line arrives as a LogRecord with `log.iostream=stdout`).

## Status

`io.kestra:kotlp:0.1.2` is **published on Maven Central** (kestra-io/kotlp `v0.1.2`). Verified by wiping the local artifacts and resolving from Central: `io.kestra:kotlp:{strictly 0.1.2} -> 0.1.2`, with the kotlp tests green against it.

Ready to merge. #18126 (the interim retry loop) becomes moot once this lands — it deletes the code that PR touches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,7 +57,6 @@ docker/app/secrets/*.yml
 
 ### Build
 core/src/main/resources/gradle.properties
-script/src/main/resources/kotlp/
 .plugins.override
 
 # H2 Database
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,4 @@
 version=2.0.0-SNAPSHOT
-kotlpVersion=0.1.0
 
 org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError
 org.gradle.parallel=true
--- a/platform/build.gradle
+++ b/platform/build.gradle
@@ -119,6 +119,8 @@ dependencies {
         api 'org.postgresql:postgresql:42.7.13'
         api 'com.github.docker-java:docker-java:3.7.1'
         api 'com.github.docker-java:docker-java-transport-httpclient5:3.7.1'
+        // Carries no classes: just the kotlp APE, read off the classpath as /kotlp/kotlp.
+        api 'io.kestra:kotlp:0.1.0'
         api "org.opensearch.client:opensearch-java:$opensearchVersion"
         api "org.opensearch.client:opensearch-rest-client:$opensearchRestVersion"
         api "org.opensearch.client:opensearch-rest-high-level-client:$opensearchRestVersion" // used by the elasticsearch plugin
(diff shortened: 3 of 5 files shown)


loicmathieu (approved):
LGTM

PR #9907, opened 2025-07-06
fix(cli): prevent NPE on commands defaulting to help

yuri1969:
### What changes are being made and why?

Prevent NPE on commands defaulting to help. For example, `./kestra server` used to fail with:

```text
java.lang.NullPointerException: Cannot invoke "jakarta.inject.Provider.get()" because "this.pluginRegistryProvider" is null
	at io.kestra.cli.AbstractCommand.call(AbstractCommand.java:101)
	at io.kestra.cli.commands.servers.ServerCommand.call(ServerCommand.java:29)
	at io.kestra.cli.commands.servers.ServerCommand.call(ServerCommand.java:10)
	at picocli.CommandLine.executeUserObject(CommandLine.java:2045)
	at picocli.CommandLine.access$1500(CommandLine.java:148)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2465)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2457)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2419)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2277)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2421)
	at picocli.CommandLine.execute(CommandLine.java:2174)
	at io.kestra.cli.App.execute(App.java:70)
	at io.kestra.cli.App.main(App.java:52)
```

---

### How the changes have been QAed?

Sadly, I was unable to reproduce that NPE using unit tests - I tried a similar approach to `io.kestra.cli.AppTest`.

PR #18134, opened 2026-08-13
fix(flows): gate the Edit flow fallback on an existing, non-creating flow

flcarre:
On `/flows/new`, `flowStore.isCreating` is `true` but `flowStore.flow` is not set until the flow exists, so the save/publish control is skipped and the `v-else-if` fallback falls through to **Edit flow**. Clicking it pushes `flows/update/edit` with an undefined id and namespace, so nothing happens.

The fallback is now gated on an existing flow and on creation not being in progress, which leaves it where it belongs: the overview of an already-saved flow.

```
v-else-if="flow && canEdit && !deleted && !flowStore.isCreating"
```

Closes https://github.com/kestra-io/kestra-ee/issues/9897

## Why this PR shrank

This PR originally fixed three dead ends on the flow creation page. [#18303](https://github.com/kestra-io/kestra/pull/18303) landed on `develop` in the meantime and dropped the flow-creation funnel entirely — `NewFlowLanding.vue`, `ImportYaml.vue` and their tests are deleted; `/flows/new` now initialises the flow and mounts the editor directly, with an auto-generated id/namespace. That superseded two of the three fixes here:

- **"Open editor greyed out with no sign why" (kestra-ee#9898)** — moot, the required-field landing form it applied to no longer exists.
- **"Create a system flow lands on Overview" (kestra-ee#9907)** — already fixed independently by #18303's namespace-aware routing (its description calls out and fixes the exact same query-param-vs-child-route bug).

Only the **Edit flow** button fix (kestra-ee#9897) still applies on top of the new direct-editor flow and is kept here. kestra-ee#9898 and #9907 are left open rather than closed by this PR, since they should be verified against #18303's fix and closed there.

## Tests

`Actions.spec.ts`: no Edit flow on the create page, where there is no flow to edit yet.

## Verification

- `npm run test:unit` — 191 files / 1816 tests, all green
- `npm run test:lint` — 0 errors
- `npm run check:types` — clean

### DIFF
--- a/ui/src/override/components/flows/Actions.vue
+++ b/ui/src/override/components/flows/Actions.vue
@@ -61,7 +61,7 @@
                 </template>
             </KsDropdown>
             <NavBarAction
-                v-else-if="canEdit && !deleted"
+                v-else-if="flow && canEdit && !deleted && !flowStore.isCreating"
                 :icon="Pencil"
                 :label="$t('edit flow')"
                 @click="editFlow"
--- a/ui/tests/unit/override/components/flows/Actions.spec.ts
+++ b/ui/tests/unit/override/components/flows/Actions.spec.ts
@@ -7,7 +7,8 @@ import KestraDesignSystem from "@kestra-io/design-system"
 const publishDraft = vi.fn().mockResolvedValue("saved")
 
 const routeState = {tab: "edit"}
-const flowState = {deleted: false}
+const flowState = {deleted: false, exists: true, isCreating: false}
+const editorState = {isAllowedEdit: true}
 
 vi.mock("vue-router", () => ({
     useRoute: () => ({params: {tab: routeState.tab}, query: {}}),
@@ -20,8 +21,10 @@ vi.mock("override/stores/auth", () => ({
 
 vi.mock("../../../../../src/stores/flow", () => ({
     useFlowStore: () => ({
-        flow: {id: "f", namespace: "ns", draft: true, deleted: flowState.deleted, source: "id: f\nnamespace: ns\n"},
-        isCreating: false,
+        flow: flowState.exists
+            ? {id: "f", namespace: "ns", draft: true, deleted: flowState.deleted, source: "id: f\nnamespace: ns\n"}
+            : undefined,
+        isCreating: flowState.isCreating,
         createFlow: vi.fn(),
     }),
 }))
@@ -51,7 +54,9 @@ vi.mock("../../../../../src/components/flows/useFlowEditorActions", () => ({
         canSave: false,
         hasErrors: false,
         isReadOnly: false,
-        isAllowedEdit: true,
+        get isAllowedEdit() {
+            return editorState.isAllowedEdit
+        },
         // The real composable returns computed refs; `isDraft` is read from script (not just
         // auto-unwrapped in a template), so the mock has to be a ref for that read to work.
         isDraft: computed(() => true),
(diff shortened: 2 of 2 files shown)


MilosPaunovic (commented):
All three fixes check out against the code:

- `isAllowedEdit` returns `false` whenever `flowStore.flow` is unset, which is exactly the `/flows/new` state, so the old fallback condition did render a dead Edit flow button - the new `flow && canEdit && !deleted && !flowStore.isCreating` gate is right, and the overview tab of an existing flow keeps the fallback as before.
- The `namespaces/update` parent redirect only reads `params.tab`, so the old `query: {tab: 'blueprints'}` was indeed dropped; the new child-route target exists via `createNamespaceTabRoutes` and `blueprints` is in `NAMESPACE_TAB_NAMES`. Good catch replacing the unit test that had locked in the wrong assertion.
- Required markers, the always-mounted `role="status"` hint, and the translation changes (2 keys x 13 locales + 2 fingerprint entries, nothing else) all look clean.

One real gap:

**The Edit flow dead-end fix does not reach the EE build, but this PR closes https://github.com/kestra-io/kestra-ee/issues/9897.** EE replaces this component with `ui-ee/src/override/components/flows/Actions.vue`, whose `secondaryAction` computed falls back to `"editFlow"` whenever `editorIsAllowedEdit` is false and `canEditFlow` is true - precisely the create-page state (same `useFlowEditorActions` composable, no flow yet). Its `editFlow` then pushes `flows/update/edit` with undefined id/namespace, the identical dead end. Since the issue was reported against EE, this needs an EE companion PR (gate the `editFlow` fallback in `secondaryAction` on `flowStore.flow && !flowStore.isCreating`), or the `Closes .../9897` line should move to that companion so the issue is not closed while EE still exhibits it. The other two issues are genuinely fixed for EE, since `NewFlowLanding.vue` and the routes are shared OSS code with no EE override.

Two non-blocking notes:

- The hint only covers empty fields; with both filled but one invalid, the button is disabled and the hint is blank. The field-level error carries the explanation then, so this is fine, just noting the asymmetry.
- Pre-existing and out of scope: old-style bookmarks like `/namespaces/edit/<id>?tab=blueprints` still land on overview, since the parent redirect never reads the query.

flcarre:
Thanks for the review, @MilosPaunovic.

Fixed the real gap: opened the EE companion PR kestra-io/kestra-ee#10096, gating the `editFlow` fallback in EE's own `secondaryAction` computed the same way this PR gates the OSS one (`flow.value && canEditFlow.value && !deleted.value && !flowStore.isCreating`), with a regression test that fails without the fix.

Also moved the `Closes .../9897` line from this PR's description to the EE companion PR, so the issue only auto-closes once EE is actually fixed rather than when this OSS half merges.

Left the two non-blocking notes as-is, per your comment.

MilosPaunovic:
Companion https://github.com/kestra-io/kestra-ee/pull/10096 reviewed and approved. With the `Closes` line moved there, the issue wiring is right: https://github.com/kestra-io/kestra-ee/issues/9898 and https://github.com/kestra-io/kestra-ee/issues/9907 close with this PR, https://github.com/kestra-io/kestra-ee/issues/9897 closes with the EE half. Nothing else open on my side.

MilosPaunovic (approved):
Approving - the one open finding from my review is resolved: the EE companion https://github.com/kestra-io/kestra-ee/pull/10096 is approved and now carries the `Closes` line for https://github.com/kestra-io/kestra-ee/issues/9897, so this PR correctly closes only the two issues the shared OSS code fixes for both builds. All checks green, nothing else open.

PR #18138, opened 2026-08-13
build(deps): migrate to Micronaut 5, keeping Kestra's Jackson on 2.x

### LINKED ISSUE #15745: Upgrade to Micronaut 5.0
See https://micronaut.io/2026/05/20/micronaut-framework-5-0-0-released/

The following PR prevent Dependabot to update to Micronaut 5, it should be reverted.
https://github.com/kestra-io/kestra/pull/16384

loicmathieu:
Migrates to Micronaut 5.0 while keeping Kestra's own serialization on Jackson 2.

Closes https://github.com/kestra-io/kestra-ee/issues/6963.
Closes https://github.com/kestra-io/kestra/issues/15745

## Why the split

`JacksonMapper` is reached statically from ~90 files and is pinned to Jackson 2 by
`com.github.fge:json-patch`, so it cannot follow Micronaut onto Jackson 3 yet. Micronaut 5 uses
Jackson 3 (`tools.jackson`) for the HTTP layer, so the two majors now coexist deliberately:

- **`io.kestra.core.serializers.JacksonMapper` stays on Jackson 2** — storage, queue, JDBC, YAML flow
  parsing, worker, tests.
- **Micronaut's own mapper moves to Jackson 3** — request and response bodies in the webserver.

This is less of a change than it sounds, because the Micronaut mapper was already a separate mapper
from the hub; it only had `PluginModule` bolted on. Jackson 3 ships no annotations module of its own
(JSTEP-1) and depends on `jackson-annotations:2.x`, so every shared annotation — `@JsonProperty`,
`@JsonTypeInfo`, `@JsonSubTypes`, `@JsonCreator`, `@JsonAnyGetter`/`@JsonAnySetter` — keeps working
under both. Micronaut 5's `Jackson2AnnotationSupport` covers `as=`, `builder=`, `typing=` and
`@JsonPOJOBuilder`; it does **not** cover `using=`, which is why exactly two serde pairs needed
Jackson 3 twins (`ListOrMapOfLabel*` and `Property.Property*`).

## Behaviour guardrails

Jackson 3 flips several defaults, so `ObjectMapperFactory` now sets them explicitly rather than relying
on the defaults matching: `SORT_PROPERTIES_ALPHABETICALLY` off, `FAIL_ON_TRAILING_TOKENS` off,
`ALLOW_FINAL_FIELDS_AS_MUTATORS` on. That last one is load-bearing and was not obvious — without it
Jackson 3 silently ignores final fields carrying a `@Builder.Default` initializer, i.e. `Property<T>` on
every plugin task.

`TenantSerializer` needed porting to `ValueSerializerModifier`: it is picked up purely by Micronaut
injecting modifier beans, with no explicit registration anywhere, so a Jackson 2 modifier would have
been silently ignored and `tenantId` would have started leaking into API payloads. There was no test
covering that; there is one now.

## Follow-up

kestra-io/kestra#18122 tracks pre-existing issues found and deliberately left alone.
https://github.com/kestra-io/kestra/issues/19554
https://github.com/kestra-io/kestra/issues/19549


graemerocher:
anything we can do to make you guys consider migrating to Micronaut Serialization. It offers significant performance improvements which we anticipate will grow bigger over time.

loicmathieu:
@graemerocher, now that I have restarted working on this, I see your comment about Micronaut Serialization.

My main concern is that we rely heavily on Jackson today, both in Kestra itself and in its 2k plugins: annotations, databind, modules (ION, CSV, Joda, ...), ... We also use JSON Patch and JSON Schema. Jackson is not only used for our HTTP layer but also to store entities in the database (we mostly use key/value, where each entity is persisted as JSONB).

In fact, what this migration shows is that we have 2 serialization hub:
- JacksonMapper (still on Jackson 2) for manual serialization (DB-serialization, plugins, ad-hoc serialization in hundred of places)
- HTTP-layer serialization, now on Jackson 3

We had to duplicate serializers from Jackson 2 to Jackson 3 for the HTTP-layer, maybe using Micronaut Serialization here would work but it would make it mandatory to duplicate custom serializers. We did it for migrating to Micronaut 5 because we don't have the choice but I expect it to be ephemeral until our own ecosystem or plugins and third-party library support Jackson 3.

I'll open an issue to evaluate switching the HTTP layer to Micronaut Serialization during our next release cycle (3 months cycle) - https://github.com/kestra-io/kestra/issues/19565.

AcevedoR (approved):

graemerocher:
@loicmathieu we would be interested to hear your requirements. We have JSON Schema support and JSON Patch is something on the radar. In the upcoming version we added support for YAML, TOML and other formats. If performance and memory matters for you the migration may be worth it https://micronaut-projects.github.io/micronaut-serialization/snapshot/guide/#why

Especially in upcoming versions. What would be good is if you could find gaps and report them to https://github.com/micronaut-projects/micronaut-serialization/issues so we can action them.

loicmathieu:
@graemerocher JSON schema, JSON draft, YAML (with anchoring). HTTP layer + manual serialization (for persisting in the DB). We have a task to evaluate Micronaut Serialization most probably in a few month, I'll get in touch, thank you.

PR #18139, opened 2026-08-13
fix(topology): resolve task icons missing from the plugin icons index

### LINKED ISSUE #18129: [2.0] Topology Icons don't load unless plugin docs loaded first
### Describe the issue

Steps to reproduce:

1. Go to a flow with multiple tasks - I used the `data-engineering-pipeline` tutorial flow
2. Open the topology view - no icons displayed
3. Go to plugin docs and click on each task
4. Each icon will load after that tasks plugin docs has been loaded
5. Refresh page and all the icons disappear again unless you open each plugin doc again

https://github.com/user-attachments/assets/9e4215a2-923a-45c2-8495-fff05a7420d0

### Environment

- Kestra Version: 2.0 RC7 OSS. **I can't reproduce on EE RC7**

elevatebart:
### ✨ Description

Topology nodes were the only icon surface fed **solely** by the bulk `GET /api/v1/plugins/icons` index. `Topology.vue` passed `icons` down to `TaskNode` / `TriggerNode` / `BasicNode` but never `loadIcon`, so a class that index doesn't carry had no way to ever get an icon — no per-class lookup, no ecosystem fallback, no retry. Every other surface (plugin docs, blueprints, task edit, gantt, no-code blocks, plugin select) passes both props.

That is exactly why opening a task's plugin documentation made its topology icon appear: the doc page resolved the class through `loadIcon` and filled the *shared* store map the topology reads. A reload emptied the store and the icons were gone again.

The change threads `loadIcon` through the topology so its nodes resolve icons the same way the rest of the app does, plus two adjacent defects in `stores/plugins.ts`:

- `fetchIcons` memoized a **rejected** promise — one failed catalog request left `iconsLoaded` false forever and every later call returned the same rejection, making every icon in the app unresolvable for the rest of the session.
- `loadIcon` now waits for a catalog fetch already in flight rather than racing it. Without this, threading `loadIcon` into the topology fires one per-class request per rendered node on first paint; with it, a flow's nodes are served by the bulk response that was already on its way.

A fourth commit is test hygiene rather than behaviour: the new icons tests and the pre-existing layout tests each declared their own `mountBasicNode` helper in `BasicNode.test.ts`. The redeclaration made the whole file fail to parse — `SyntaxError: Identifier 'mountBasicNode' has already been declared` — so none of its tests ran, the two older layout ones included. The two helpers are now one that takes prop overrides and slots.

**How this was verified** — built OSS `develop`, ran `server local` + the Vite dev UI against it, and drove the flow topology with Playwright:

| scenario | before | after |
|---|---|---|
| class present in the bulk index | icon renders | icon renders, no extra requests |
| one class removed from the bulk response | placeholder, **zero** recovery attempts | resolved through the per-class / ecosystem path |
| bulk request fails (500) | every node blank, zero recovery attempts | nodes resolve individually |

The middle row reproduces the issue exactly: with the class dropped, the `download` node showed the generic placeholder while its four siblings rendered fine; opening that class's plugin doc in the side panel made the topology icon appear live, without a reload.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18129.

### 🎨 Frontend Checklist

- [ ] All existing E2E tests pass (`npm run test:e2e`) — not run here. On the current head: `npm run check:types` clean, `eslint` clean on the changed files, the full `packages/topology` unit suite green (129 tests across 6 files), and `tests/unit/stores` + `tests/unit/components/plugins` at 203 passed / 2 failed. Both failures are timeouts in `tests/unit/stores/logsCursor.spec.ts`, which this PR doesn't touch and which fails identically on `develop`.
- [ ] Screenshots or video recordings attached showing the `UI` changes — behaviour is covered by the table above and by the new tests; happy to attach captures if useful

### 📝 Additional Notes

Two things a reviewer may want to weigh in on:

1. **Why EE didn't reproduce.** Whether a class is in the bulk index depends on which plugins are registered on that instance, so the OSS/EE split is a symptom of instance state, not of EE-specific code. The topology gap is a bug either way — with the EE plugin manager, any instance can run a task type the index doesn't cover.
2. **A follow-up I deliberately left out of scope.** `loadIcon` still skips the local per-class endpoint entirely once the bulk index has loaded (`iconsLoaded ? Promise.resolve(undefined) : …`), treating the index as authoritative — there is an existing test asserting that. It isn't quite authoritative: the server's per-class route also consults the group-icon map, and the index is `@Cacheable`, so a plugin installed after it was built won't be in the client's copy. Those classes can then only be resolved from `api.kestra.io`, which won't work on an air-gapped install. Worth a separate discussion rather than flipping a tested behaviour inside a bug fix.

### 🤖 AI Authors

Why did the cat sit on the topology graph? It heard the nodes had no icons and figured a blank square was clearly a spot reserved for a cat. 🐱


ui/src/stores/plugins.ts
```diff
@@ -107,10 +107,15 @@ function usePluginsIcons() {
...
 
         iconsPromiseLocal.value =
-            axios.get<Record<string, RawPluginIcon>>(`${apiUrlWithoutTenants()}/plugins/icons`, {}).then(async response => {
+            axios.get<Record<string, RawPluginIcon>>(`${apiUrlWithoutTenants()}/plugins/icons`, {}).then(response => {
                 pluginsIcons.value = toPluginIconDataMap(response.data)
                 iconsLoaded.value = true
                 return icons.value
+            }).catch(() => {
+                // Drop the memoized promise so a later caller retries; keeping a rejected one
+                // cached leaves every icon in the app unresolvable for the whole session.
```
elevatebart:
```suggestion
                // rejected request -> try again
```

elevatebart: Applied in 0f999c2. Your suggestion anchored to the second line of the two-line comment, so I replaced the whole comment with it rather than committing it as-is — otherwise the first line would have been left dangling.

---

elevatebart:
@flcarre this one is ready for your review when you have a moment — CI is green on `83fc7f4aa` (15 check runs, 9 success, 6 skipped by the path filter for a frontend-only diff), no merge conflict against `develop`, and no open review threads.

One thing changed since the review was requested: a fourth commit merges two duplicate `mountBasicNode` helpers in `ui/packages/topology/tests/units/nodes/BasicNode.test.ts`. The redeclaration made the whole file fail to parse (`SyntaxError: Identifier 'mountBasicNode' has already been declared`), so none of its tests ran — the two pre-existing layout tests included. `Frontend - Tests` was passing before only because the job's reporting didn't surface that suite as a blocker. The file now loads and its 4 tests pass; the full `packages/topology` suite is 129 tests green.

The reviewable behaviour is unchanged from the original three commits: `loadIcon` threaded through the topology, plus the rejected-promise memoization and the catalog race in `stores/plugins.ts`. The description's test line has been updated to the numbers verified on this head.

---

elevatebart:
Correcting myself on one sentence in my previous comment. I wrote that `Frontend - Tests` was green before "only because the job's reporting didn't surface that suite as a blocker". I hadn't verified that, and I can't from here — the step is defined in `kestra-io/actions` (`kestra-oss-frontend-tests.yml`), outside what I can read. Please disregard that explanation; the fix itself is unaffected.

What is verified:

- The `unit` vitest project does collect the file — `vitest list --project=unit --filesOnly` includes `packages/topology/tests/units/nodes/BasicNode.test.ts`, and only `packages/design-system/**` is excluded.
- With the duplicate helper in place, that file fails to parse and vitest exits `1`: `Test Files 1 failed`, `Tests no tests`, `SyntaxError: Identifier 'mountBasicNode' has already been declared`.
- On `0f999c22` (the Aug 13 head, which had the duplicate), run [31706758501](https://github.com/kestra-io/kestra/actions/runs/31706758501) shows step 12 "Run front-end unit tests" concluding **success**, and the `report-ui-test-failures` step logging "No UI test failures — nothing to report".

So a file that fails to *collect* passed through that job green on Aug 13. Whether that's `continue-on-error` on the step, a narrower test command, or the JUnit report not carrying collection errors, I can't tell from this repo — worth a look by someone who can read the shared workflow, since it would hide any future parse or import error in a unit test file. Happy to open a separate issue if that's useful.

---

flcarre (approved):

PR #18140, opened 2026-08-13
fix(flows): validate flows as they will run

fhussonnois:
Flow validation parses the flow **as authored**, not as it will run. Any governance an edition applies at runtime is invisible to `ModelValidator`, so a task whose required property is supplied by governance is reported invalid and cannot be saved.

This is the seam that makes the fix possible; the behaviour change lives in the companion PR kestra-io/kestra-ee#9909 (same branch name, needs to be merged together).

### Changes

- `FlowParsingService.parseForValidation(flow)` — parses the flow as it will run. Defaults to the existing lenient parse, so **OSS behaviour is unchanged**. Unlike `parseForRuntime` it never rejects the flow: blocking is owned by the save and execution gates.
- Route the validation paths through it: `FlowService.create` / `update` / `validate` / `validateForExecution`, and `AbstractJdbcFlowRepository.update` (which re-validates independently of the service — the existing `// FIXME validation is done both here and in the repo`).

The `create`/`update` double-parse was the old plugin-defaults injection point, which is why the javadoc still claimed validation happened "after injecting all plugin default values". Updated.

### Tests

`:core:test --tests "io.kestra.core.services.*"` 95 passing, `H2FlowRepositoryTest` 122 passing. No behaviour change here, so no new OSS test; the regression is covered in the EE PR.

### DIFF
--- a/core/src/main/java/io/kestra/core/services/FlowParsingService.java
+++ b/core/src/main/java/io/kestra/core/services/FlowParsingService.java
@@ -130,6 +130,26 @@ public class FlowParsingService {
     }
 
     /**
+     * Parses the given abstract flow for the validation paths (save, validate endpoint, pre-execution check),
+     * returning a parsed {@link FlowWithSource}. May be overridden to apply the same governance mutations the
+     * flow will run with, so constraints are checked against the effective configuration rather than the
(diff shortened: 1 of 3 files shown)


ammeek (approved):
LGTM

PR #18141, opened 2026-08-13
build(design-system): consolidate UI assets and lazy-load Monaco/Shiki

elevatebart:
## What changed

The production UI build shipped **898 asset files**, and every page — including the flows list — preloaded **141 chunks**, among them the entire Monaco toolchain (editor.api, the TypeScript language service) and every Shiki grammar. This PR consolidates the build to **404 files** with **6 modulepreloads** on first paint, and Monaco/Shiki are now fetched only when an editor or markdown surface actually renders.

Three root causes, three fixes:

1. **Shiki full bundle** — `shikiHighlighter.ts` used `createHighlighter` from `shiki`, whose bundled-language registry emits one lazy chunk per language (~350 files). It now uses `createHighlighterCore` from `shiki/core` with 28 statically registered grammars, and every other grammar is fetched on demand from a single `shiki-langs` fallback chunk (5 MB, loaded only when a fence uses one of them). No loss of language support; fence aliases (`js`, `sh`, …) still resolve, and languages Shiki doesn't know still render as plain text.
2. **Eager design-system barrel** — `index.ts` statically imported every component, so any `import {KsButton}` dragged KsEditor→Monaco and KsMarkdown→Shiki into the eager graph of every page. Both are now exported as `defineAsyncComponent` wrappers (template usage and `h()` callers are unaffected; prop types preserved via a type-only cast).
3. **No chunk strategy** — a new `ui/plugins/consolidateChunks.js` defines rolldown `codeSplitting` groups: `monaco` (4.1 MB lazy), `markdown` (2.4 MB lazy), `design-system` (one unit incl. element-plus, shareable by federated plugin UIs), and single eager `vendor`/`app` chunks.

## Numbers

| Metric | Before | After |
|---|---|---|
| Asset files | 898 (789 JS + 90 CSS) | 404 (310 JS + 75 CSS) |
| `modulepreload` on first paint | 141 | 6 |
| Total assets size | 41 MB | 33 MB |
| Monaco/Shiki on the flows list | eager | lazy |

Cold boot through `vite preview` + proxied API: 47 JS/CSS requests, ~1.3 MB transferred, no failures.

## Reviewer notes

- **The plugin must be registered after `@module-federation/vite`**: MF deletes user-declared `codeSplitting.groups` in its `config` hook, so ours are appended to the groups MF installs, at negative priority so MF's wrapper groups always win their modules.
- **Grouping mistakes crash at boot, not at build** (TDZ errors from chunk cycles). The constraints are documented in the plugin: MF shared singletons (vue/@vue, kestra-sdk in both its `@kestra-io/...` and `packages/kestra-sdk/dist/...` id forms) keep their own chunks; `?worker` stubs stay with their importer; `@kestra-io/topology` stays ungrouped (it imports the DS barrel and is eagerly reachable via `flowYamlUtils` imports in stores); no `maxSize` splitting (it can cut cyclic libs like zrender across chunks).
- EE consumes the same barrel, so KsEditor/KsMarkdown become async there too; EE only gets the chunk groups if its vite config adds the plugin.
- Verified: full design-system unit tests (773), KsMarkdown tests, `check:types` in both `ui` and the design-system package, eslint, and a browser smoke test of the built bundle. Worth a manual pass on a logged-in session: flow editor (async KsEditor), plugin docs (markdown chunk), and a federated plugin UI if available.
- The on-demand grammar bundle must never become a *static* import of another chunk (grammars embed each other, e.g. `html` → `css`), so its exclusion set is derived from the module graph in `buildEnd` and a `generateBundle` guard fails the build if any chunk gains a static edge to it.
- Possible follow-up: a dedicated subpath export for `flowYamlUtils` in `@kestra-io/topology` would shrink the eager graph further.


elevatebart:
Restored on-demand language support (`4727a2a78c`), so the compatibility caveat in the PR description no longer applies.

`KsMarkdown` again falls back to Shiki's bundled registry for a fence whose language isn't pre-registered. The difference from the original code is packaging: a `shiki-langs` chunk group collapses those ~320 grammars into **one 5 MB chunk**, fetched only when such a fence actually appears, instead of the ~350 per-language chunks Shiki's registry emits by default.

The subtle part is keeping that chunk off the critical path. Grammars embed each other (`html` statically imports `css` and `javascript`, and `css` is not in the pre-registered set), so a name-based exclusion list would leave an embedded grammar in the fallback chunk and turn it into a *static* import of the markdown chunk — every markdown render would then pull 5 MB. So the exclusion set is derived from the module graph in `buildEnd`: any grammar a module statically imports, plus its transitive grammar imports. That also picked up a second static consumer I'd have missed by hand, `src/components/plugins/schema/shikiToolset.ts`.

As a backstop, a `generateBundle` guard fails the build if any chunk gains a static import of the fallback bundle. I verified it isn't vacuous by temporarily removing the exclusion — the build failed with the intended message pointing at the leaking chunk.

Measurements unchanged where it matters: **405 asset files, 8 modulepreloads**, markdown chunk still 2452 KB (byte-identical size, so no grammar data leaked in), fallback emitted as one separate 5112 KB chunk referenced only through a dynamic `import()`.

Tests: two new cases in `KsMarkdown.test.ts` cover an on-demand grammar (elixir highlights) and an unknown language (plain-text fallback). Full suites green — design-system 775, app 1704, plus typecheck and lint.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

MilosPaunovic (approved):

PR #18150, opened 2026-08-13
feat(charts): expose disabling the webserver's embedded controller as a values option

### LINKED ISSUE #18149: Helm chart: expose disabling the webserver's embedded controller as a values option
Hey Kestra team, thanks for the great chart — small ergonomics suggestion from setting up distributed mode! 🙂

**Context**

In distributed mode, `deployments.webserver` starts an embedded controller alongside the webserver process. When also running a dedicated `deployments.controller` Deployment (the recommended distributed-mode topology), this leaves two controllers running for the same release — the embedded one on the webserver, and the dedicated one.

The fix is passing `--no-controller` to the webserver's own CLI invocation. Today the only way to do that via the Helm chart (`kestra-io/kestra`, chart `kestra`, currently `1.3.33`) is through the generic `deployments.webserver.extraArgs` list:

```yaml
deployments:
  webserver:
    enabled: true
    extraArgs:
      - "--no-controller"
```

This works, but it means every operator running distributed mode with a dedicated controller has to independently discover this exact flag — it's not mentioned in the chart's `values.yaml` comments (which describe `extraArgs` generically as "Extra arguments to pass to the container"), and nothing in the chart surfaces the webserver/controller overlap at all.

**Request**

Would love to see this exposed as a first-class boolean value, e.g. `deployments.webserver.controller: false` (mirroring the existing `deployments.controller.enabled` toggle), mapping internally to appending `--no-controller`. This would make the recommended distributed-mode topology (dedicated controller + webserver with its embedded controller disabled) discoverable directly from `values.yaml`, instead of requiring tribal knowledge of an undocumented CLI flag — and keep `extraArgs` free for its intended purpose rather than overloading it for a documented, common topology choice.

Thanks for considering it!

PhillypHenning:
## Summary
- Adds `deployments.webserver.controller` (boolean, default `true`) to the Helm chart, mapping `false` to `--no-controller` on the webserver's CLI invocation.
- In distributed mode, the webserver starts an embedded gRPC worker controller by default. When also running a dedicated `deployments.controller` Deployment (the topology this chart's README recommends), that leaves two controllers running for one release. Today the only way to disable the embedded one is passing `--no-controller` through the generic `deployments.webserver.extraArgs`, which isn't surfaced anywhere in `values.yaml`.
- Purely additive and backward-compatible: default (`true`) preserves today's behavior exactly.

Closes #18149

## Changes
- `charts/kestra/values.yaml`: new `deployments.webserver.controller` value.
- `charts/kestra/templates/deployment.yaml`: conditionally appends `--no-controller` to the webserver's command when `controller` is `false`, scoped only to the webserver Deployment.
- `charts/kestra/README.md.gotmpl` / `README.md`: updated the existing distributed-mode note to document the new option (regenerated via `helm-docs`).
- `charts/kestra/tests/deployment_webserver_controller_test.yaml`: new helm-unittest suite covering the default (no-op), disabled (`--no-controller` appended), and isolation from other deployment types (e.g. `standalone`).

## Test plan
- [x] `helm lint charts/kestra`
- [x] `helm template` with `deployments.webserver.controller` unset/`true`/`false` — confirmed rendered command matches expectations in each case
- [x] `helm template` with `deployments.standalone.enabled=true` and `deployments.webserver.controller=false` — confirmed the flag does not leak into the standalone Deployment
- [x] `helm unittest charts/kestra -f tests/deployment_webserver_controller_test.yaml` — 3/3 passing
- [x] Confirmed the pre-existing failures in `tests/jvm_active_processor_test.yaml` are unrelated (present identically on `develop` before this change)

### DIFF
--- a/charts/kestra/README.md
+++ b/charts/kestra/README.md
@@ -308,13 +308,12 @@ configurations:
 > **Note:** The webserver pod also starts an embedded controller by default. In distributed mode
 > workers only connect to `<release-name>-controller`, so this embedded controller is unused.
 > This is harmless — Kestra supports multiple controllers for load balancing — but wastes resources.
-> To disable it, add `--no-controller` to the webserver's `extraArgs`:
+> Disable it with:
 >
 > ```yaml
 > deployments:
 >   webserver:
->     extraArgs:
->       - --no-controller
+>     controller: false
 > ```
 
 ### Upgrade prerequisite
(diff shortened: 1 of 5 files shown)


PhillypHenning:
Reopening from the Bitovi org fork instead of my personal fork.

PR #18177, opened 2026-08-14
fix(design-system): bundle component locales instead of chunking them

elevatebart:
### ✨ Description

Eager-globs the design-system `*.locale.ts` files so they compile into the design-system chunk, and makes `registerDesignSystemI18n` synchronous.

The lazy glob made every locale file a dynamic-import boundary: one chunk per file, fetched serially before `app.mount()`. Today only the `consolidateChunks` design-system group keeps them out of the shipped build — remove that group and all 8 requests come back. Eager makes it structural, and the synchronous merge also fixes the fire-and-forget `void registerDesignSystemI18n(i18n)` on the plugin install path.

Per-language splitting was measured and rejected: all 13 languages are 46 KB gzipped total, and the authoring format (one file, all languages) is what the translation generator/checker and `fingerprints-design-system.json` consume.

Production build, cold boot in Chromium up to mount (uncompressed, static server):

| | chunks | modulepreload | first-paint requests | first-paint bytes |
|---|---|---|---|---|
| before | 296 | 8 | 55 | 5,399,966 |
| after | 296 | 8 | 55 | 5,399,432 |

Unchanged, as expected — the group already merged them. Mechanism check with `consolidateChunks()` disabled: lazy emits `KsEmpty.locale-*.js`, `KsFilter.locale-*.js`, … (8 chunks, 185 first-paint requests, 6 locale fetches before mount); eager emits none.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra-ee/issues/9948.

### 🎨 Frontend Checklist

- [x] `npm run check:types`, `npm run test:unit` (187 files / 1766 tests), design-system `unit:test` (96 files / 778 tests) pass

### 📝 Additional Notes

New test `ui/tests/unit/designSystem/i18n.spec.ts` mounts `KsEmpty` right after `registerDesignSystemI18n` with no await and asserts the label is translated in `en` and `fr`; it fails against the old lazy implementation. Verified in the built app too: `$t("no_data")` at mount resolves to the French and Japanese strings.

### 🤖 AI Authors

Why did the cat sit on the bundler? It heard there were too many chunks and only one of them was tuna. 🐱

---
_Generated by [Claude Code](https://claude.ai/code/session_01DkqRmhps1ycqPuFUvg7m79)_

### DIFF
--- a/ui/packages/design-system/src/i18n/index.ts
+++ b/ui/packages/design-system/src/i18n/index.ts
@@ -8,13 +8,17 @@ export function setDesignSystemLocale(locale: string) {
     designSystemLocale.value = locale
 }
 
+// Eager: a lazy glob makes every locale file its own dynamic-import boundary, so the bundler
+// emits one chunk per file and the app fetches them serially before it can render anything.
 const localeModules = import.meta.glob<{default: Record<string, object>}>(
     "../components/**/*.locale.ts",
+    {eager: true},
 )
 
-export async function registerDesignSystemI18n(i18n: I18n) {
-    for (const loadMod of Object.values(localeModules)) {
-        const mod = await loadMod()
+/** Merges the component locale files into the app messages. Synchronous, so callers cannot
+ * render before the design-system keys exist. */
+export function registerDesignSystemI18n(i18n: I18n) {
+    for (const mod of Object.values(localeModules)) {
         for (const [lang, messages] of Object.entries(mod.default)) {
             i18n.global.mergeLocaleMessage(lang, messages)
         }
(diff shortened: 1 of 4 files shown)


elevatebart:
Closing: the symptom this targets is already gone on `develop`, so the change buys ~0.01% of first-paint bytes.

`plugins/consolidateChunks.js`'s `design-system` group matches everything under `design-system/src/`, locale modules included, so they already compile into `design-system-*.js` and cost no extra requests. Measured cold boot to mount, before vs. after: 55 → 55 requests, 5,399,966 → 5,399,432 bytes, 296 → 296 chunks.

Measurements kept on https://github.com/kestra-io/kestra-ee/issues/9948.

---

Issue #2, opened 2019-11-27
Properly handle outputs for EachSequential  


tchiotludo:
On `EachSequential`, the output is unreachable with handlebars : 
https://github.com/tchiotludo/floworc/blob/cb93c3d93622193835a466fa0e0b63d7286f10a6/core/src/main/java/org/floworc/core/models/executions/Execution.java#L200
since we add a `uid()` with `TaskRunId`.
Find a proper way to add a the value of all childs

Issue #1, opened 2019-11-27
Sequential Task with errors flow are not handle


tchiotludo:
```yaml
id: errors
namespace: org.floworc.tests

tasks:
- id: failed
  type: org.floworc.core.tasks.scripts.Bash
  commands:
  - 'exit 1'
  errors:
  - id: 2nd
    type: org.floworc.core.tasks.debugs.Echo
    format: second {{task.id}}
```

The errors will not be triggers, need to be global to work or on `FlowableTask`

PR #18178, opened 2026-08-14
feat(ui): suggest alternate source search separator

### LINKED ISSUE #18061: Source Search misses resources written with a different separator (us-east-1 vs US_EAST_1)
Source Search matches literally, so a query written with one separator never finds a resource written with another. The sharpest case is secrets: `us-east-1` returns **zero** secret keys on an instance where `US_EAST_1_ACCESS_KEY` exists.

This is not a regression. It is the existing matching semantics, made visible by the cross-resource scope added in kestra-io/kestra#18041, which puts all four resource types on one screen for the first time.

### Why it happens

Flow matching quotes the query when regex is off (`SourceSearchMatcher.toPattern`, `core/src/main/java/io/kestra/core/utils/SourceSearchMatcher.java`), and is case-insensitive by default. So case is already handled; the separator is not.

OSS secrets are sourced from `SECRET_<NAME>` environment variables, so their keys are always uppercase with underscores. A user searching in the kebab-case form that appears in their flow YAML will never match them.

### Measured on a local instance

Seeded secrets: `US_EAST_1_ACCESS_KEY`, `LANDING_BUCKET_US_EAST_1`.

| Query | Secret keys found |
|---|---|
| `us-east-1` | **0** |
| `us_east_1` | 2 |
| `US_EAST_1` | 2 |
| `east` | 2 |

### Why it matters

The feature's stated purpose is answering "where is this value referenced across my Kestra estate". A user chasing a region, a bucket or a credential name silently misses an entire resource type, with nothing on screen indicating the query form is the reason.

### Options, roughly in increasing cost

1. **Hint only.** When a type returns zero and the query contains a separator, suggest the alternative form. Additive, changes no matching semantics.
2. **Normalise separators in the query.** Treat `-`, `_` and `.` as interchangeable. Catches the real-world case, but also changes flow search behaviour, which is already shipped and in use.
3. **A "loose match" toggle.** Explicit user control, more UI surface.

Option 2 changes the semantics of a shipped feature, so it needs a product call rather than being folded into a frontend PR.

### Not in scope here

The related inconsistency, where KV and namespace-file search split the query into terms while flows and secrets match the whole query, is handled in kestra-io/kestra#18041 by highlighting per term so a row never renders unmarked.

yunuschoudhurywork-lang:
### ✨ Description

Adds a suggestion for Source Search when no results are found and the query uses a different separator format.

For example:
- Searching `us-east-1` with no results can suggest `us_east_1`.
- Searching `us_east_1` with no results can suggest `us-east-1`.
- The suggestion is shown only when the alternate query actually returns results.
- Clicking the suggested query triggers the search again.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18061

### 🎨 Frontend Checklist

- [x] Relevant frontend tests pass
- [x] UI behavior tested for the Source Search suggestion

### 🛠️ Backend Checklist

None.

### 📝 Additional Notes

The existing Source Search matching behavior is unchanged. The suggestion is handled on the frontend after the original search returns no results.

### 🤖 AI Authors

None.


flcarre (commented):
Thanks for this, and welcome! The idea is a good one: when a source search comes back empty, offering the `-`/`_` variant is exactly the kind of small touch that saves people a retype.

I pulled the branch to try it and hit a few things that would stop it running as submitted, so I'm flagging them together rather than one at a time.

### The component doesn't compile yet

Four identifiers in `FlowsSearch.vue` aren't declared in scope:

* `:753` uses `response.results`, but `crossResourceSearchStore.search(...)` returns `Promise<void>` and never assigns a `response`. The only `response` in the file is the parameter of `reportReplaceResult` at `:681`, which is a different scope. The store already populates its own state, so this local `results` mirror can go away entirely.
* `:769` assigns to `errorMessage`, which isn't declared anywhere. The store already exposes `errorMessageFor(type)` for exactly this.
* `:500` and `:753` use the type `SourceSearchResult`, but only the plural *component* `SourceSearchResults` is imported (`:341`).
* `:507` uses `selected.value`, but the ref is called `selection`. That line is also worth dropping from the diff entirely: it rebuilds the key by hand in a format that no longer matches `crossSearchResultKey()`, which `goToMatch` compares against at `:597`/`:606`, so next/prev match navigation would quietly stop working. It looks unrelated to the feature.

### Two behavioural things

* `:761` does `alternativeQuery.value`, but `getSeparatorVariant` returns `string | null`, so `.value` is `undefined` and the suggestion query goes out empty.
* `:758` drops `namespace: namespaceFilter.value`, which the primary search passes at `:750`. Inside a namespace-scoped search the suggestion would be computed globally, so clicking it could land on another empty result.

### Translations

The new `did_you_mean` key is in `en.json` only. Nothing to undo here, the generator just needs a run: `npm run translations:generate` from `ui/`, then commit the twelve locale files and `fingerprints.json` alongside it.

### Rendering

The copy interpolates the suggestion into `did_you_mean` and then repeats it in the button plus a trailing `?`, so it reads as "Did you mean us_east_1? us_east_1 ?". Using `<i18n-t keypath="source_search.did_you_mean">` with the clickable part as a named slot fixes that in one go. The plain `<button>` also wants to be a `KsButton` or `KsLink` so it picks up theming in both light and dark.

None of this is a problem with the idea itself. One suggestion for the shape: `getSeparatorVariant` is a nice pure function, and it would be at home in `ui/src/utils/crossResourceSearch.ts` where it can carry a unit test. A test there would have caught the `.value`-on-a-string one immediately.

Happy to take another look once it builds.

MilosPaunovic:
Hey @yunuschoudhurywork-lang, are there any updates on this?

yunuschoudhurywork-lang:
> Hey @yunuschoudhurywork-lang, are there any updates on this?

Hi @MilosPaunovic! I’m still working on it. The work is in progress, and I expect to have it completed by tomorrow. I’ll share an update once it’s ready for review.

yunuschoudhurywork-lang:
@flcarre Hey👋, I’ve pushed the requested post-review updates.

* Updated the source search flow to use the correct store results and preserve the active namespace/search filters for the alternate query.
* Standardized selection, navigation, toggling, and validation around crossSearchResultKey().
* Added the “Did you mean” behavior so the separator variant is suggested only when it produces actual results, with the suggestion updating the search query when selected.
* Moved getSeparatorVariant() into sourceSearchDiff.ts and added unit coverage for hyphen/underscore conversion and the no-separator case.
* Removed the unused computed values and resolved the related TypeScript issues.

I also ran the focused unit test for the new utility (3/3 passing), type-checking, and the full unit test suite. The full suite still has unrelated existing/environment failures, including the @kestra-io/hey-api-plugin/dist/runtime.js module issue. The translation check also requires the generated locale files to be regenerated through the project's translation-generation setup.

The requested implementation changes are now pushed to the PR. Please take a look when you get a chance.
@MilosPaunovic as Promised sir!🙌

yunuschoudhurywork-lang:
@MilosPaunovic status?

flcarre (changes requested):
There's more here than the CI flakiness you flagged — one of the failures is a real regression this PR introduces, not an environment issue.

**CI (blocking):**
- Frontend tests fail because `SourceSearchSelectionGroup` was removed from `sourceSearchDiff.ts`, but `sourceSearchDiff.spec.ts` (an untouched, pre-existing file) still imports it. That's not the hey-api-plugin issue — it's this PR's own type removal. Either keep a compatible export or update that spec to use `SourceSearchResult`.
- Translations check fails because `did_you_mean` was only added to `en.json` — needs `npm run translations:generate` for the other 12 locales, as you already know.

**Not yet addressed from my first review:**
- The clickable suggestion in the empty state is still a raw `<button>`, not `KsButton`/`KsLink` — no theming, no light/dark, no focus states.
- The "Did you mean" text still renders garbled: `t()` already appends the `?`, so the button repeats the suggestion and there's a stray literal `?` after it — reads as "Did you mean us_east_1? us_east_1 ?". Use `<i18n-t>` with the clickable part as a named slot, like I suggested last time.

**New, smaller:**
- The local `results` ref only mirrors `crossResourceSearchStore.flows.results` at one call site and is read nowhere else — drop it, check the store directly.
- Dead CSS left behind (`.source-search__no-results-title`/`-hint`, including a `:deep(code)` inside) after the markup that used those classes was replaced.

Good news: the actual crash fixes from before (undefined `response`, missing `errorMessage`, `crossSearchResultKey()` everywhere) are genuinely in, and the new `getSeparatorVariant` util is clean and well tested. Let's get the regression and the two design-system items closed out and this is close.

yunuschoudhurywork-lang:
@flcarre Thanks! I've addressed the actionable feedback from the review and pushed the changes in commit .

Implemented:

* Restored the SourceSearchSelectionGroup export as a compatibility alias to SourceSearchResult, so the existing sourceSearchDiff.spec.ts / frontend references are not broken.
* Replaced the raw HTML <button> for the suggestion with the Kestra KsButton component.
*Updated the “Did you mean” rendering to use <i18n-t> with the named suggestion slot, removing the duplicated suggestion text and extra ?.
* Removed the redundant local results ref and now read the results directly from crossResourceSearchStore.flows.results.
Removed the obsolete empty error catch that only cleared the removed local results state, while retaining the finally block so searchPending is reset correctly.
* Removed the dead .source-search__no-results-title / .source-search__no-results-hint CSS and associated :deep(code) styling.
* Kept the previously addressed crash/behavior fixes intact.
* Kept getSeparatorVariant() in sourceSearchDiff.ts with its existing focused tests.

Verification performed:

* npx vitest run src/utils/sourceSearchDiff.spec.ts → 1 test file passed
* Targeted vue-tsc check for FlowsSearch.vue, sourceSearchDiff.ts, and sourceSearchDiff.spec.ts → no errors
* git diff --check → clean
* Final diff reviewed; only the two intended files were changed.

Translations:
I also ran npm run translations:check. It correctly reports source_search.did_you_mean as stale in en.json and missing from the generated locale files. Running npm run translations:generate requires the project's GEMINI_API_KEY, which I don't currently have access to. I therefore did not manually modify the generated locale files or fingerprints. This part remains pending until the project translation generation can be run with the appropriate credentials.

The implementation changes are committed and pushed. Once the translation generation is run with the project credentials, the remaining translation check should be rerun.

flcarre (changes requested):
Thanks for pushing the fixes — the `<i18n-t>` + `KsButton` cleanup, dropping the redundant local `results` ref, removing the obsolete empty `catch`, and clearing the dead `.source-search__no-results-*` CSS all land as described and look correct.

This still can't merge as-is though — CI is red for two independent, real reasons.

**1. Blocker — the "compatibility alias" broke an existing test (the actual cause of the Frontend Tests failure)**

`ui/src/utils/sourceSearchDiff.ts:6`:
```ts
export type SourceSearchSelectionGroup = SourceSearchResult
```
This isn't backward compatible. Before this commit `SourceSearchSelectionGroup.matches` was `{line: number; column: number}[]`. Now it's `SourceMatch[]` = `Required<ApiSourceMatch>[]`, which also requires `snippet`. `ui/tests/unit/utils/sourceSearchDiff.spec.ts` (pre-existing, untouched by this PR) builds fixtures as `{line, column: 0}` and no longer compiles:

```
tests/unit/utils/sourceSearchDiff.spec.ts(28,5): error TS2322: Type '{ line: number; column: number; }[]' is not assignable to type 'Required<SourceMatch>[]'.
```

That's why "Frontend Tests" is red — `npm run check:types` fails before vitest even runs. This contradicts "targeted vue-tsc check clean" — please re-run `npm run check:types` locally and you'll see it fail. Give `SourceSearchSelectionGroup` back a `matches` shape that doesn't require `snippet`, rather than a plain alias to `SourceSearchResult`.

**2. Blocker — translation drift**

`translations:check` reports `source_search.did_you_mean` as stale/missing across all 12 locale files, exactly as you flagged. Per this repo's rules, that has to be fixed by running `npm run translations:generate` and committing the regenerated locale files + `fingerprints.json` — never by hand-editing them. If nobody on the team can run it with `GEMINI_API_KEY` for you, this PR stays blocked until someone does.

**Minor**
- `FlowsSearch.vue:257` — `<KsButton text size="small">`: `text` isn't a declared `KsButton` prop (only `type="text"` is). It currently falls through to the underlying `ElButton` via `$attrs`, but please use `type="text"` for consistency with the component's typed API.
- The empty-results state no longer shows the "N matches found in {type}, which you haven't selected" hint — `source_search.no_results_hidden_type` and `source_search.no_results_in_types` are now unused in all 13 language files. Not introduced by this latest commit, but worth cleaning up (or restoring) while you're in this component.
- The new alternate-query fetch block has 4 `vue/script-indent` lint warnings and a couple of missing-space `if(...)"` nits — run `npm run lint` before the next push.

**Not this PR's fault**
- E2E failed on `java.net.SocketException: Connection reset` during `make build-exec` — looks like a CI infra hiccup unrelated to this diff. Worth a rerun once the two blockers above are fixed.

Please fix #1 and get #2 regenerated, then I'll take another pass.

flcarre (changes requested):
Progress since the last round: the type-compatibility blocker is genuinely fixed — `SourceSearchSelectionGroup` is back to its original `{line, column}` shape instead of the `SourceSearchResult` alias, so `sourceSearchDiff.spec.ts` compiles again and Frontend Tests are green. The `KsButton type="text"` fix also landed as asked.

This still can't merge — CI is red for the same reason flagged twice already.

**Blocker — translation drift (unresolved)**

`source_search.did_you_mean` is still only in `en.json`. The "Translations - Key parity" job fails on exactly this key in all 12 other locales. Per repo rules this needs `npm run translations:generate` (requires `GEMINI_API_KEY`) and the regenerated locale files + `fingerprints.json` committed — never hand-edited. If you don't have access to that key, flag it and a maintainer can run it and push directly, since the actual code is otherwise close.

**Minor — business logic in the component**

`FlowsSearch.vue:733` (`fetchResults`) calls `FlowsAPI.searchFlowsBySourceCode(...)` directly for the "did you mean" alternate query, instead of going through `crossResourceSearchStore` like the primary search does. Since `fetchResults`'s `try` has no `catch` (only `finally`), a failure here becomes an unhandled rejection instead of a handled error state.

**Minor — leftover unused translation keys**

`source_search.no_results_hidden_type` and `source_search.no_results_in_types` are unused everywhere now. Not new this round, but worth removing while touching this component.

**Nit — lint**

Worth a `npm run lint` pass before the next push if not already clean.

yunuschoudhurywork-lang:
@flcarre I’ve addressed the remaining code-level items.


The only remaining blocker is the translation parity step. I ran npm run translations:generate, but the generation fails locally because the required Google credentials are not available (Could not load the default credentials). I therefore did not commit the generated locale/fingerprint changes from the failed generation.

Could a maintainer run the translation generation with the project credentials and push the generated locale/fingerprint updates? The implementation is otherwise ready.

flcarre (changes requested):
Thanks for sticking with this through four rounds. The things I asked for last time genuinely landed — `SourceSearchSelectionGroup` is back to its original shape, `KsButton type="text"` is in, the `<i18n-t>` rendering is correct, and the optional lookup no longer produces an unhandled rejection. The feature is a good one and `getSeparatorVariant` is clean.

Two blockers remain before this can merge.

### 1. CI: translation parity (unchanged for four rounds)

`Translations - Key parity, placeholders and drift` fails with 13 errors:

```
[OSS] 1 key(s) have no up-to-date translation ...: source_search|did_you_mean
[OSS] Translation "de" is missing key "source_search.did_you_mean"
... and the same for es, fr, hi, it, ja, ko, pl, pt, pt_BR, ru, zh_CN
```

The fix is `npm run translations:generate` from `ui/`, then commit the 12 regenerated locale files plus `fingerprints.json`. Never hand-edit them.

You've said twice you don't have `GEMINI_API_KEY`, and that's fair — this one is on us, not you. **@MilosPaunovic** — can you run the generator and push the result to this branch? That clears the last hard blocker.

While it's running, note that removing `no_results_hidden_type` and `no_results_in_types` from `en.json` leaves them orphaned in all 12 other locales. The regeneration should clean those up too.

### 2. The suggestion probe is scoped to flows, but the empty state isn't

`FlowsSearch.vue:730`:

```ts
if (crossResourceSearchStore.flows.results.length === 0) {
```

The empty state is decided by `summaryMatchCount` across every *selected* resource type, but this gate looks at flows alone. When the user has deselected "flows", `crossResourceSearchStore.search()` sets `flows.value = {status: "idle", results: []}` unconditionally (`stores/crossResourceSearch.ts:292`), so the condition is trivially true and we fire a flows-only probe for a search that isn't looking at flows. Click the resulting suggestion and it re-runs against the user's actual types — landing on another empty result, which is the exact failure mode we discussed for `namespace` in the first round. Gate on the same signal the empty state uses, and probe the same types the search used.

### 3. The direct API call also bypasses the store's staleness guard

`FlowsSearch.vue:734` still calls `FlowsAPI.searchFlowsBySourceCode(...)` directly. Beyond the layering point from last time, the concrete cost is that `crossResourceSearchStore.search()` threads a `nextGeneration()` token through every request precisely so a late response can be discarded. The probe has no such guard: type `us-east-1` (empty, probe in flight), keep typing to a query that's also empty, and the first probe can resolve afterwards and set `suggestedQuery` to the variant of the *previous* query. Routing this through the store gets the guard for free.

### 4. The empty state lost its typography

`FlowsSearch.vue:250-251`:

```html
<h3>{{ t('source_search.no_results_title', {query}) }}</h3>
<p>{{ t('source_search.no_results_description') }}</p>
```

Deleting `.source-search__no-results-title` / `.source-search__no-results-hint` took their `--ks-font-size-sm` / `--ks-font-weight-medium` / `--ks-text-secondary` rules with them, and nothing replaced them. `KsEmpty` only styles the description container, which a bare `<h3>` overrides with its own heading size — so the title now renders much larger and darker than it did. Use `KsText`, or reapply the tokens.

Related: dropping `hiddenTypeHint` removed the "N matches found in {type}, which you haven't selected" line, but the `Select all types` button it explained is still there (`:264`). That button now appears with no stated reason. This is also outside the stated scope of a separator-suggestion PR — I'd restore the hint rather than expand the diff further.

### Smaller

- `sourceSearchDiff.ts:124` — not gated on regex mode. With "Use regular expression" enabled, `flow-[0-9]+` becomes `flow_[0_9]+`. Skip the suggestion when `regexEnabled` is on.
- `sourceSearchDiff.ts:125` — mixed separators only flip one way: `my-flow_name` -> `my_flow_name`, never `my-flow-name`. Worth a decision (and a test) either way; the empty-string case is untested too.
- `sourceSearchDiff.spec.ts` — this module already has a spec at `ui/tests/unit/utils/sourceSearchDiff.spec.ts`. Adding a second under `ui/src/utils/` splits the tests across two conventions (239 specs under `tests/unit/**` vs 25 under `src/`). Please move the `getSeparatorVariant` cases into the existing file.
- `en.json:2426` — `did_you_mean` is inserted after `no_results_title`; the rest of the block is alphabetical.
- `FlowsSearch.vue:736` — `size: 200` duplicates the store's page size as a literal.
- `FlowsSearch.vue:744` — drop the `// Ignore failures from the optional suggestion lookup.` comment; the empty catch says it.

Fix #2 and #3 (and ideally #4), and I'll get the translation generation sorted on our side so this can land.

flcarre:
Picking this back up — you pushed after my `CHANGES_REQUESTED` and GitHub didn't re-request me, so it dropped off my list. Re-reviewing the alternate-separator suggestion now.

The second failure, `Translations - Key parity, placeholders and drift`, *is* yours — a new user-facing string needs its key in all thirteen locale files with `fingerprints.json` regenerated (`cd ui && npm run translations:generate && npm run translations:check`). Worth clearing before the re-review so I'm reading a green branch.

yunuschoudhurywork-lang:
hey @flcarre thanks again! I’ve addressed all the requested code-side changes, including the search/probe conditions, store-based staleness guard, empty-state typography, regex and mixed-separator handling, test coverage, translation key ordering, and the remaining cleanup items. The translation generation/parity changes have been left for you to handle as mentioned.

flcarre (changes requested):
Round 6. Let me lead with the part that is our fault, not yours.

## The Frontend Tests failure is new, and it's a rule we changed under you

The red job is no longer the type error from earlier rounds. It is now:

```
FAIL  tests/unit/i18n/i18nGuard.spec.ts > i18n > follows the i18n conventions across ui/src and ui/packages
AssertionError: Fix the i18n usage, see ui/AGENTS.md:
src/components/flows/FlowsSearch.vue:247 (render the string with $t() and named arguments, <i18n-t> is not used)
src/components/flows/FlowsSearch.vue:238 (templates call $t(), the t() from useI18n() belongs in <script>)
src/components/flows/FlowsSearch.vue:241 (templates call $t(), the t() from useI18n() belongs in <script>)
```

`tests/unit/i18n/i18nGuard.ts` and the rewritten `ui/AGENTS.md:100` landed on develop *after* your last merge, and `ui/AGENTS.md` now reads:

> Pass plurals and interpolation as named arguments to `$t(...)`, never string-concatenate. `<i18n-t>` is not used, and `tests/unit/i18n/i18nGuard.spec.ts` fails on it.

So the `<i18n-t>` rendering I asked you for in rounds 1-3, and confirmed as correct in rounds 4 and 5, is now the thing CI rejects. Sorry — that one is on us.

Your branch is 36 commits behind develop, so merging develop is step one; the guard isn't even in your tree yet. Then three edits:

`FlowsSearch.vue:253` and `:256` — swap `t(...)` for the global `$t(...)`:

```html
<KsText tag="h3" size="large">
    {{ $t('source_search.no_results_title', {query}) }}
</KsText>
<KsText tag="p">
    {{ $t('source_search.no_results_description') }}
</KsText>
```

`FlowsSearch.vue:262` — the replacement for `<i18n-t>` when a string embeds a component is `splitTranslation(t, key, slot)` from `src/utils/splitTranslation`, which hands back the text either side of the placeholder. `did_you_mean` is already `"Did you mean {suggestion}?"`, so it slots straight in. Keep the two halves and the button on one physical line — Vue turns a line break between an interpolation and an element into a space, and `AGENTS.md:101` calls that out specifically.

## Translation parity — still the same 13 errors, still not yours to fix

```
[OSS] 1 key(s) have no up-to-date translation ...: source_search|did_you_mean
[OSS] Translation "de" is missing key "source_search.did_you_mean"
... es, fr, hi, it, ja, ko, pl, pt, pt_BR, ru, zh_CN
```

@MilosPaunovic — this still needs `npm run translations:generate` from `ui/` with `GEMINI_API_KEY`, plus the 12 locale files and `fingerprints.json` committed to this branch. While it runs, note that `no_results_in_types` was removed from `en.json` but is still present in all 12 other locales (I checked de, fr, ja, zh_CN) — the regeneration should clear those out too.

## What landed from the last round

All of it, and cleanly:

- **Probe scoping** — the gate is now `selectedTypes.includes("flows") && !anyCountingSelected && summaryMatchCount === 0`, the same signal `searchViewState` uses to decide the empty state. That closes the flows-only-probe-on-a-non-flows-search hole.
- **Store routing and the staleness guard** — `searchFlowSuggestion(params, gen)` lives in the store, `search()` returns its generation, and `isCurrent(gen)` is applied before the result counts. No more direct `FlowsAPI` call from the component.
- **Regex** — `if (!params.query || params.regex) return null`, so `flow-[0-9]+` is left alone.
- **Mixed separators** — now returns `null` rather than flipping one way, with a test for it and for the empty string.
- **Typography** — `KsText tag="h3" size="large"` / `tag="p"` replaces the bare `<h3>`/`<p>`, and `hiddenTypeHint` is back, so the `Select all types` button has its explanation again.
- **Spec location** — the `getSeparatorVariant` cases are in the existing `ui/tests/unit/utils/sourceSearchDiff.spec.ts`.
- **`size: 200` literal and the narrative comment** — both gone.

The heuristic itself reads well now. It only suggests when the variant search actually returns flows, so it can't produce a dead-end suggestion, and the `-` XOR `_` condition means it can never echo back the string the user typed. That was the thing I was most worried about, and it's handled.

## One real issue left in the code

**`FlowsSearch.vue:749` — a stale probe can still wipe a fresher suggestion.**

```ts
suggestedQuery.value = await crossResourceSearchStore.searchFlowSuggestion({...}, gen)
```

The store's guard makes a stale probe return `null`, but the component assigns that `null` regardless. Two overlapping runs:

1. Run A starts (gen 1), search in flight.
2. Run B starts (gen 2), resolves, probe resolves, `suggestedQuery = "us_east_1"`.
3. Run A's search resolves late, its probe sees `isCurrent(1) === false`, returns `null` — and A assigns `null`, clearing B's suggestion.

The generation token protects the store's own state; it doesn't protect a ref the component writes from the return value. Either have the component check the generation before assigning, or move `suggestedQuery` into the store behind the same `isCurrent` gate the other setters use. Worth noting the same run also evaluates `summaryMatchCount.value === 0` against whatever is in the store *now*, so a stale run reads a newer run's counts.

## Smaller

- `crossResourceSearch.ts:167` — the probe asks for `size: SEARCH_PAGE_SIZE` (200 results with source snippets) to answer "are there any?". `size: 1` does it.
- `crossResourceSearch.ts:172` — `response.results.length` is unguarded, while `searchFlows` at `:148` writes `(response.results ?? [])`. The `catch` turns the TypeError into a silent `null`, so it hides rather than crashes, but it should match its sibling.
- `FlowsSearch.vue:755` — `searchPending` is cleared in `finally`, after the probe. That avoids a pop-in, which I like, but it also means every empty search containing a separator waits a second serial round-trip before "No matches" appears at all. A `size: 1` probe makes it cheap enough that I'd leave it.
- `FlowsSearch.vue:744` — because the gate requires flows to be selected, someone searching only files/KV/secrets never sees a suggestion even when the variant would match there. Fine as a scope limit, but say so in the description so it isn't read as a bug later.
- `en.json` — `did_you_mean` is still sitting after `no_results_title`; the rest of the block is alphabetical.
- `FlowsSearch.vue:249` — the empty-state container has no `aria-live`, so the suggestion appearing after the probe isn't announced. The `KsButton` is a real `<button>`, so focus and Enter/Space already work; it's just the announcement that's missing.

Merge develop, do the three `$t`/`splitTranslation` edits, fix the stale-assign at `:749`, and I'll chase the translation generation on our side so this can finally land.

MilosPaunovic:
@yunuschoudhurywork-lang Translations are handled - I ran the `Translations - Generate for a pull request` workflow on our side and it pushed `64388bcb0b` onto this branch.

It filled `source_search.did_you_mean` in all twelve locales and refreshed `fingerprints.json`. The same pass also dropped `source_search.no_results_in_types`, which had been removed from `en.json` earlier on this branch but was still sitting in every other locale. `Translations - Key parity, placeholders and drift` is green now.

One thing to know: if you move `did_you_mean` into its alphabetical slot in `en.json`, that marks the key stale again - ping me and I will re-run the workflow.

yunuschoudhurywork-lang:
@MilosPaunovic Thanks for taking care of the translation generation and parity fixes. 

@flcarre I’ve addressed the remaining Round 6 code changes, including the $t()/splitTranslation updates, stale suggestion assignment, and the size: 1/results guard cleanup. I’ve pushed the changes to the branch. Appreciate the continued review and guidance!

flcarre (approved):
Round 7, and this one is green. Every check passes.

You found the right fix for the guard failure on your own, which I appreciate — `splitTranslation(t, "source_search.did_you_mean", "suggestion")` with both halves and the `KsButton` on one physical line is exactly what `ui/AGENTS.md:101` asks for, and the two `t(...)` -> `$t(...)` swaps are correct. `tests/unit/i18n/i18nGuard.spec.ts` passes locally here too. No suppression, no key moved, no feature dropped to get there.

The stale-assign from last round is closed as well: the store returns `undefined` on a stale generation and the component only assigns when it isn't. `size: 1` instead of `SEARCH_PAGE_SIZE`, and `(response.results ?? [])` to match `searchFlows` — both in.

Translations are sorted on our side and this branch left them alone, which is the right call. To correct myself from last round: **leave `did_you_mean` where it is in `en.json`.** Moving it back to its alphabetical slot re-marks the key stale and forces another generation run. The alphabetical nit is withdrawn.

I'm approving. Nothing below blocks the merge — fix what you like in a final push, or let it ride.

### Small things

**`crossResourceSearch.ts:155` / `:158`** — the two early returns fire before any `await`, so they aren't generation-guarded:

```ts
if (!params.query || params.regex) return null
const alternativeQuery = getSeparatorVariant(params.query)
if (!alternativeQuery) return null
```

A run whose query has no separator can still resolve late and hand back `null`, clearing a newer run's suggestion. Narrow — it needs an overlap with one of the two non-debounced paths — but `return isCurrent(gen) ? null : undefined` on both lines closes it, same as the `catch` already does.

**`FlowsSearch.vue:720`** — `searchPending.value = true` is now only set in the debounce watcher. `searchViewState` computes `searching = searchPending || anyCounting`, and the store clears `counting` before the probe runs, so on the two paths that skip the debounce (`:668` after a replace-all, `:781` on mount) the empty state paints first and the suggestion pops in a round-trip later. Setting it at the top of `fetchResults` makes all three paths behave the same; the watcher line stays useful for the debounce window.

**`tests/unit/stores/crossResourceSearch.spec.ts`** — `searchFlowSuggestion` has no store-level test. That spec already covers generation staleness ("keeps the newest results when an older search resolves last"), so the regex gate, the empty-results `null`, and the stale-`undefined` would slot right in alongside it. `getSeparatorVariant` is well covered; this layer isn't.

**`FlowsSearch.vue:247`** — the suggestion line is a bare `<p>` while the three siblings in the same `KsEmpty` description slot are `KsText tag="p"`. `KsText` there keeps the typography tokens consistent. Same line: the `KsButton` could use a `data-test` so this has an E2E handle.

**`sourceSearchDiff.ts`** — still no trailing newline at EOF, and `SourceSearchSelectionGroup` carries an unrelated semicolon reformat plus a stray double blank line after it. Worth reverting so the diff stays about the feature.

**`FlowsSearch.vue:234`** — no `aria-live` on the empty-state container, so a suggestion arriving after the probe isn't announced. The `KsButton` is a real `<button>`, so focus and Enter/Space already work.

Thanks for staying with this through seven rounds and a rule we changed under you mid-flight. Good feature.

PR #18180, opened 2026-08-14
perf(ui): keep echarts and vue-flow off the boot path

elevatebart:
### ✨ Description

Gets `echarts` and `@vue-flow` off the boot path, the follow-up #18141 left behind.

- **echarts.** `design-system/src/index.ts` statically imported the five chart components, and `init.ts` registers the whole barrel globally, so `echarts` was eager on every page. Converted `KsEchart`/`KsGraph`/`KsLine`/`KsBar`/`KsPie` to the existing `asyncComponent` helper (same shape as `KsEditor`/`KsMarkdown`, type-only cast kept). This needs a matching lazy chunk group — without it the eager `vendor` and `design-system` groups recapture the toolchain and nothing moves.
- **flowYamlUtils.** Added a `@kestra-io/topology/flow-yaml-utils` subpath, backed by a module reaching neither vue-flow nor the design system, and repointed `stores/flow.ts`, `stores/dashboard.ts`, `utils/flowableBlockOps.ts`.
- **vue-flow.** `FlowGraph` was only ever used in type positions, so it is now a `type` import. `stores/playground.ts` genuinely calls two `vueFlowUtils` functions, both inside `async` functions, so they load on demand.
- **One extra eager path the issue did not name.** `App.vue → DrillDownDrawer → LogsWrapper → Sections → dashboard-types → TimeSeries → echarts/charts`. `LogsWrapper` only renders behind a `v-if` in that drawer (and with `:withCharts="false"`), so it is now async too. Without this the barrel change alone leaves echarts eager.
- **A dashboard chunk group**, because the previous point has a cost: once `Sections`/`TimeSeries` leave the eager `app` chunk they scatter into a dozen small chunks. Grouping them recovers most of that (see below).

`KsPie`'s story imported the barrel rather than the component, unlike the other four chart stories, so it started rendering the async wrapper; aligned it with its siblings.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra-ee/issues/9952.

Companion EE PR repointing the three equivalent `ui-ee` stores: kestra-io/kestra-ee#9954. That one depends on this PR for the new export, so merge this first.

### 📊 Measurements

Production build, cold boot in Chromium against a static server, uncompressed.

**Login route** (`/login`, the anonymous route from the issue):

| | modulepreload | requests | transferred | echarts | @vue-flow |
|---|---|---|---|---|---|
| before | 27 | 43 | 5,399,429 | `vendor-*.js` | `vendor-*.js` |
| after | 27 | 43 | **4,288,068** | none | none |

−1,111,361 bytes (−20.6%) at an identical request count. Detection is by chunk *contents*, not filename — in the baseline both libraries sat inside `vendor`, whose name says nothing. `vendor` itself drops 2,557 kB → 1,586 kB.

**Dashboard route** (`/` → `home`), the one route that pays for this:

| | requests | transferred |
|---|---|---|
| before | 51 | 5,404,612 |
| after, no dashboard group | 66 | 5,054,359 |
| **after** | **59** | **5,052,200** |

So the dashboard makes **8 more requests and downloads 352 kB less**. Two of those eight are structural — `charts` has to be its own chunk or echarts is back on the boot path, and `dashboard` is the merge that replaced six files. The other six are small modules shared between the eager `app` chunk and the lazy `dashboard` chunk, which rolldown hoists rather than duplicates; squeezing them out would mean duplicating them or letting a group straddle eager and lazy, which is the cycle hazard #18141 documents. Not worth it.

Every route that does not render charts — login included — is a flat −1.1 MB at no request cost.

### 🧪 Chunk-cycle check

The named risk is a boot-time TDZ / chunk cycle, not a build failure, so cycles are checked directly:

- Static-import cycles between emitted chunks: **1 before, 1 after** — the pre-existing `app → index → app`. No new cycle, and no boot chunk statically imports a lazy toolchain chunk.
- In-browser force-evaluation of the lazy chunks after boot: `charts` 10/10 exports defined, `dashboard` 12/12, `vue-flow-utils` 28/28, zero `pageerror`s. A cycle would surface here as a TDZ `ReferenceError`.

Both new groups match narrowly for the same reason: `ksChartUtils` is re-exported eagerly by the design-system barrel, and `DrillDownDrawer`/`dashboard/composables` are eager via `App.vue`, so a directory-wide pattern in either group would pull an eager module into a lazy chunk and cycle the two.

### 🔍 consolidateChunks topology exclusion — still required

The issue asked whether it can go. Measured: removing it saves 2 requests and 76 bytes, and **reintroduces the `vendor → design-system → vendor` cycle its comment warns about** (2 cycles vs 1). Left in place, comment unchanged.

### 🎨 Frontend Checklist

- [x] All existing E2E tests pass (`npm run test:e2e`) — green in CI on this PR
- [x] `npm run check:types`; `npm run test:unit` (186 files / 1764 tests); `npm run test:storybook` (68 / 244); design-system `unit:test` (96 / 778) and `storybook:test` (80 / 479); topology `unit:test` (4 / 104); `npm run test:lint` (0 errors)

### 📝 Additional Notes

**Please land after #18177** (kestra-io/kestra-ee#9948) or rebase this on it — both touch the design-system package, and #18177 also reasons about the `consolidateChunks` groups this PR adds to.

**What is and is not covered by a real browser.** CI's E2E job is green, which exercises the logged-in app against a real backend — login, the flow editor and the no-code `blocks` suites, executions and the AI copilot. That is the strongest evidence here that nothing boots broken.

Still **not** covered by any automated browser run, and worth a human pass before merge — these are the three surfaces the issue names, and none of the E2E specs touch them:

- the **topology graph** (the `@vue-flow` render),
- a **dashboard with charts** (the `echarts` render),
- the **playground** (whose store now loads `vueFlowUtils` on demand).

Locally I could not run the backend (Java 21 vs the required 25; the PR Docker image cannot be pulled because the blob host `pkg-containers.githubusercontent.com` is denied by this environment's network policy, though `ghcr.io` itself is reachable), so instead I booted the real production bundle in Chromium on `/` and `/login` — app mounts, no page errors, only backend-absence console noise — force-evaluated the `charts`, `dashboard` and `vue-flow-utils` chunks in that page, and rendered all five chart components in a real browser through their Storybook stories.

### 🤖 AI Authors

Our cat reviewed the bundle and found it acceptable: fewer chunks to sit on, but she insists the 671 kB charts chunk is now the perfect size for a nap. 🐱


elevatebart:
Closing: the maintainers decided not to take this change. The dashboard route's request-count regression was judged not worth the boot-path saving.

Recording the measurements so nobody re-attempts this blind. Production build, cold boot in Chromium, uncompressed:

| route | requests | transferred |
|---|---|---|
| login, before | 43 | 5,399,429 |
| login, after | 43 | 4,288,068 |
| dashboard, before | 51 | 5,404,612 |
| dashboard, after | 59 | 5,052,200 |

The dashboard cost is the sticking point and it is largely structural. Getting `echarts` off the boot path requires `App.vue → DrillDownDrawer → LogsWrapper → Sections → dashboard-types → TimeSeries → echarts` to stop being a static chain; once `Sections`/`TimeSeries` leave the eager `app` chunk they scatter, and a `dashboard` chunk group only recovered 7 of the 15 extra requests. Two of the remaining eight cannot be removed at all — `charts` must be its own chunk or echarts is eager again, and `dashboard` is the merge itself. The last six are modules shared between the eager `app` chunk and the lazy `dashboard` chunk, which rolldown hoists rather than duplicates.

Two findings worth keeping regardless of this PR's fate:

1. **The `consolidateChunks` topology exclusion is still required.** Removing it saves 2 requests and 76 bytes and reintroduces the `vendor → design-system → vendor` cycle its comment warns about. kestra-io/kestra-ee#9952 asked whether it could go; the answer is measured no.
2. **`FlowGraph` in `stores/flow.ts` and `stores/playground.ts` is only ever used in type positions.** Making those `import type` removes `@vue-flow/core` from those stores' boot path with no behavioural change and no dashboard cost at all. That part is separable from everything else here if anyone wants a zero-risk subset later.

The branch `claude/eager-graph-chart-loading-7mkg3b` is left in place, so the work is recoverable.

---

Issue #9951, opened 2025-07-07
Created & Updated dates should be displayed accurately instead of <x mins ago>


shruti-0508:
### Describe the issue

Since we already capture the Flow creation date & update date.
So, we can simply show the exact date or timestamp on Kestra UI

{"id":"bU6V6jB4QxT3TkRuumVzK","namespace":"tutorial","flowId":"hello-world","flowRevision":3,"inputs":{"user":"Shruti-test1"},"labels":[{"key":"system.correlationId","value":"bU6V6jB4QxT3TkRuumVzK"}],"state":{"current":"CREATED","histories":[{"state":"CREATED","**date":"2025-07-07**T11:25:13.239616461Z"}],"duration":"PT0.022245167S","startDate":"2025-07-07T11:25:13.239616461Z"},"originalId":"bU6V6jB4QxT3TkRuumVzK","deleted":false,"metadata":{"attemptNumber":1,"originalCreatedDate":"2025-07-07T11:25:13.239779003Z"}

![Image](https://github.com/user-attachments/assets/21bee9d5-c4f4-4467-8525-b635f4158e00)

### Environment

- Kestra Version: develop

anna-geller:
This is a deliberate choice that we show it as X minutes/hours ago. When you hover over this, you can get the exact date IIRC. 

@Nico-Kestra @Ben8t can you check if there's opportunity to nicely combine relative and absolute dates here and close the issue if everything is as expected?

Ben8t:
DMed @Nico-Kestra and he is agreeing with you @anna-geller: everything expected here.

I'm closing the issue for now. @shruti-0508 is there any high-level use case where you actually need the absolute date? Can you give us more details here on what use of the absolute date you will do?

shruti-0508:
@Ben8t Currently I dont have any use case for the absolute date. I was doing testing on Kestra, so I thought it can be a bug.

Issue #9947, opened 2025-07-07
Duration mismatch in API response vs Executions overview


shruti-0508:
### Describe the issue

There is a duration mismatch I found when I executed the flow using curl request on terminal and validated the every field of the response on Kestra UI.

**STEPS**
1. Copy the API from flow and directly execute curl command on the terminal.
2. Check the id in response and look for the same on Kestra Executions window.
3. Validate each field, duration of flow execution on curl request is **PT0.012238875S** which is approximately 12ms.  and same execution summary on Kestra UI shows **5.67s**

Here is the snapshot of Kestra UI and CURL REQUEST

![Image](https://github.com/user-attachments/assets/523a5bef-f54a-4da1-a1bd-95768669e884)

curl -v -X POST -H 'Content-Type: multipart/form-data' -F 'user=Shruti ' 'http://localhost:8080/api/v1/executions/tutorial/hello-world'

Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> POST /api/v1/executions/tutorial/hello-world HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.4.0
> Accept: */*
> Content-Length: 157
> Content-Type: multipart/form-data; boundary=------------------------5eUewXTflx2fr1oA5nodjq
> 
* We are completely uploaded and fine
< HTTP/1.1 200 OK
< date: Mon, 7 Jul 2025 10:27:29 GMT
< content-type: application/json
< content-length: 580
< 
* Connection #0 to host localhost left intact
[...]

PR #18181, opened 2026-08-14
fix(core): serve precompressed UI assets instead of per-request jar inflation

elevatebart:
### ✨ Description

Every `/ui/**` request re-inflated the deflated entry from inside the shadow jar, re-compressed it from scratch, and `StaticFilter` did a blocking read of `index.html` on the Netty event loop per request.

- The UI build now emits precompressed `.gz`/`.br` siblings for every compressible output (`ui/plugins/precompressAssets.js`, brotli q11 / gzip 9); `UiController` replaces the `classpath:ui` static mapping and `VueStaticResourceResolver`, streaming the `Accept-Encoding`-negotiated variant straight from the classpath on the IO executor — no request ever inflates or compresses content, and no bytes are held in memory. ETags derive from zip-entry metadata (CRC + size), so `If-None-Match` answers 304 without opening a stream; hashed `/ui/assets/**` get `immutable, max-age=31536000`, `index.html` stays `no-cache`, SPA history fallback preserved. Explicit `Content-Encoding` also bypasses Netty's dynamic compressor for the whole UI asset path.
- `index.html` is rewritten once into an immutable template (`UiIndexService`); per request only the CSRF meta substitution runs. `Content-Length` is the UTF-8 byte length, cache headers survive, no body conversion for non-HTML. `StaticFilter` deleted.
- `PluginController.getPluginUi` streams third-party plugin files with a weak source-hash ETag (`no-cache` revalidation costs no read) — resolves the `//todo add front cache later`.

An earlier iteration used a bounded in-memory cache with runtime compression; review moved it to build-time precompression, which also removes the first-visitor compress-everything tax (benchmark in the comments).

JVM flags and the auth path are untouched (kestra-io/kestra-ee#9950, #9951).

Companion EE PR: kestra-io/kestra-ee#9955.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra-ee/issues/9949.

### 🛠️ Backend Checklist

### 📝 Additional Notes

Benchmarks in the comments: browser-level A/Bs (k6 + Chromium, 30 VUs — TTFB p75 826ms → 26ms vs develop; precompressed streaming vs the cache iteration: cold first load 5.3 → 1.7 CPU-s) and a server-side cold-load A/B (burst CPU −84%, liveness p50 825ms → 32ms during the burst, repeat visit collapses to 304s).

https://claude.ai/code/session_01Y74HcfQRK8wzBF41PNTTic


AcevedoR:
## Plain-language summary of this PR (what was wrong, what changes)

### The problem

The web UI is a few hundred files (JS, CSS, fonts…) shipped **inside the Kestra jar**, which is itself a zip archive. Before this PR, the webserver did zero preparation and zero remembering:

1. **Every request unzipped the file again.** When a browser asked for `app.js`, the server dug into the jar, decompressed that entry from scratch, and streamed it out. Ask for the same file 1,000 times → it unzips it 1,000 times.
2. **Every response was re-compressed again.** The bytes were then gzip-compressed on the fly — from scratch, on every request, even for the exact same file.
3. **Both happened on Netty's event-loop threads** — the small set of threads whose job is to *dispatch* network traffic, not do heavy work. While they compress JavaScript, **nothing else gets served, including `/health`**. That's why the pod in kestra-io/kestra-ee#9947 stalled and was killed by its Kubernetes liveness probe: it wasn't out of memory, its network threads were choked doing zip work.
4. **`StaticFilter` made `index.html` worse**: it intercepted every `/ui/**` response and, for the HTML page, read the whole file as a string and did find-and-replace (base path, title, CSRF token) — per request, on the event loop.
5. **Weak browser caching**: no ETags, no `immutable` headers — browsers re-fetched files they already had.

One login page ≈ 300 requests × (unzip + recompress + blocking reads) = a CPU bomb on the worst possible threads.

### The fix

The mental model: **do the work once, remember the result, serve from memory forever after.**

- **A bounded in-memory cache (64MB default, `kestra.webserver.ui-resource-cache.max-size`).** On first request a file is read from the jar *once*, compressed *once* (gzip, and brotli when brotli4j is present), fingerprinted (ETag) *once*, and kept in memory. Every later request is a copy from RAM — the jar is never touched again. Least-used entries are evicted if the cache fills.
- **Everything runs on IO worker threads, not the event loop** (`@ExecuteOn(TaskExecutors.IO)`), so `/health` always answers — the liveness-kill scenario can't be triggered by asset serving anymore.
- **Proper browser caching.** Vite content-hashes filenames (`app-abc123.js`), so those get `max-age=31536000, immutable` — a returning browser makes almost no asset requests. Everything else revalidates via ETag and gets a `304` answered straight from memory.
- **`index.html` is rewritten once at startup** into an immutable template (`UiIndexService`); the only per-request work is inserting the user's CSRF meta tag — a string concatenation. `StaticFilter` and the `VueStaticResourceResolver` hack are deleted; a plain controller (`UiController`) replaces them, keeping the SPA history-mode fallback.
- **Plugin UI files go through the same cache** (resolving the old `//todo add front cache later`), keyed by plugin UI source hash so upgrades never serve stale bytes.

AcevedoR:
## A/B benchmark: this PR vs its base commit (30-VU cold-boot stress)

Ran the `browser:stress` scenario from `benchmark-internal` (k6 + headless Chromium, 30 virtual users, each a fresh uncached browser session logging in and browsing the UI — the same protocol that reproduced the kestra-io/kestra-ee#9947 incident profile).

**Methodology** (per the A/B runbook): both sides source-built on the same machine with the same toolchain (`build-kestra-image.sh`, no plugins) — **A** = merge-base on develop (`846dfbc8`), **B** = this PR's head (`e47a091b`). Identical lifecycle per side: fresh VM deploy → cold JVM → wiped Postgres → seed → scenario, same `n2-standard-32` browser VM for both. Stack: `oss/postgres`, kestra node = 4 vCPU.

| metric | A (base) | B (this PR) | change |
|---|---|---|---|
| TTFB p75 | 826 ms | **29 ms** | −96% |
| browser HTTP avg / p95 | 0.9 s / 4.5 s | **0.1 s / 0.6 s** | −87% |
| LCP p75 | 9 s | **2.2 s** | −76% |
| FCP p75 | 4 s | **1.8 s** | −55% |
| login avg / p95 | 18.6 s / 32.1 s | **6.0 s / 19.4 s** | −68% / −40% |
| iterations in same wall-clock | 104 | **158** | +52% |
| sessions past login | 42 | **55** | +31% |
| checks passed | 261/261 | **367/367** | 100% both |

The TTFB collapse (826 ms → 29 ms) is the direct signature of the fix: assets served from memory instead of per-request jar inflation + recompression. 100% checks passing on both sides means the new ETag/gzip/304 plumbing broke nothing functionally — every UI journey (login, flow create, execute, search, logs) completed correctly.

Per-action latencies (avg, survivors only): `open_flows` 10.6 s → 2.2 s, `open_dashboard` 7.0 s → 2.9 s, `create_flow_wizard` 16.8 s → 5.4 s, `save_flow` 5.3 s → 1.8 s.

**Caveat:** session abandonment under this extreme protocol isn't gone (55/158 attempts past login vs 42/104 — note B absorbed 52% more session attempts in the same wall-clock because iterations completed faster). The residual matches what this PR deliberately leaves to the companion issues: baseline/direct memory (kestra-io/kestra-ee#9950) and per-request auth cost (kestra-io/kestra-ee#9951), plus 30 simultaneous cold Chromium boots being genuinely CPU-hard for a 4-vCPU node. This PR fixes the asset layer (kestra-io/kestra-ee#9949), and the numbers show it does exactly that.

Reports: `benchmark-internal` `results/browser/roman-oss-postgres-20260817-130832` (A) and `results/browser/roman-oss-postgres-20260817-131523` (B).

elevatebart:
## Server-side cold-load benchmark (complements the browser A/B above)

Same shadow-jar deployment layout as production: a synthetic Vite-shaped UI dist (22MB raw, 38 files, ~4.7MB gzipped — the post-#18141 profile) shaded into two jars, **A** = merge-base `846dfbc8`, **B** = this PR. `server local` (H2), JDK 25, default JVM flags, 4 vCPU. Client: 30 VUs, each fetching `index.html` + all 38 assets (`Accept-Encoding: gzip, deflate, br, zstd`, 6 connections/VU, no client cache), then a second burst re-sending the captured `If-None-Match`. Liveness probed every 200ms during the bursts (`/ping`; the management port isn't exposed under `server local`, and `/ping` rides the same event loops whose saturation caused the kestra-io/kestra-ee#9947 probe kills).

| metric (1170 requests, cold JVM) | A (base) | B (this PR) | change |
|---|---|---|---|
| burst wall time | 20.9 s | **4.7 s** | −77% |
| per-request p50 / p95 / p99 | 2.0 / 9.2 / 15.7 s | **0.44 / 1.7 / 3.0 s** | p95 −81% |
| server CPU for the burst | 62.3 s | **10.0 s** | −84% |
| bytes on wire | 153.1 MB | 147.5 MB | −4% (precompressed gzip −9 vs per-request level 6) |
| liveness during burst p50 / max | 825 / 1450 ms | **32 / 337 ms** | stays responsive |
| RSS after burst | 884 MB | **667 MB** | −25%, incl. the resident asset cache |
| **revalidation burst** (repeat visit) | 1170×200, 153.1 MB, 44.7 s CPU | **1140×304 + 30×200, 9.6 KB, 1.5 s CPU** | ~all bytes and CPU gone |

The repeat-visit row understates production impact: hashed `/ui/assets/**` now ship `immutable, max-age=31536000`, so a returning browser doesn't even send those 1140 requests. The 30 index.html fetches stay 200 here only because each simulated user starts without its CSRF cookie; with the cookie the index revalidates to a 304 too.

Errors: 0 on both sides, all bodies byte-identical after decompression.

---

AcevedoR (approved):

webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java
```diff
@@ -533,7 +540,10 @@ private boolean isDistributionAllowed(PluginDistribution distribution) {
     @Get(value = "/{group}/pluginUi/{path:.*}")
     @ExecuteOn(TaskExecutors.IO)
     @Operation(tags = { "Plugins" }, summary = "Get plugins group by subgroups")
-    public HttpResponse<StreamedFile> getPluginUi(
+    // The response is raw file bytes, not base64: keep the OpenAPI schema as binary like the previous StreamedFile signature.
+    @ApiResponse(responseCode = "200", description = "Ok", content = @Content(schema = @io.swagger.v3.oas.annotations.media.Schema(type = "string", format = "binary")))
+    public HttpResponse<byte[]> getPluginUi(
```
loicmathieu: This was previously streamed and is now a big byte[] which is concerning if some artifacts are big.

webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java
```diff
@@ -545,21 +555,33 @@ public HttpResponse<StreamedFile> getPluginUi(
...
+            .get(cacheKey, mediaType, () -> readPluginUiResource(plugin, resourcePath))
+            .map(resource -> (HttpResponse<byte[]>) uiResourceCacheService.respond(request, resource, REVALIDATE_CACHE_DIRECTIVE))
+            .orElseThrow(NotFoundException::new);
+    }
 
-        //todo add front cache later
-        return HttpResponse.ok(streamedFile);
+    private static Optional<byte[]> readPluginUiResource(RegisteredPlugin plugin, String resourcePath) {
+        try (InputStream in = plugin.getClassLoader().getResourceAsStream(resourcePath)) {
+            return in == null ? Optional.empty() : Optional.of(in.readAllBytes());
```
loicmathieu: Again, we previously stream.

I think this is too naive and we need to be able to:
1. stream the file as before
2. bypass the cache if the file is too big so a buggy plugin artifact cannot fill the cache letting no cache space available for more useful files

webserver/src/main/java/io/kestra/webserver/services/UiIndexService.java
```diff
@@ -0,0 +1,187 @@
...
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+
+/**
+ * Serves the UI {@code index.html}: the file is read from the classpath and rewritten (base path, analytics,
+ * title, custom head) only once into an immutable template; per request only the CSRF meta tag is substituted.
+ */
+@Singleton
+@Requires(property = "kestra.webserver.ui.enabled", notEquals = "false", defaultValue = "true")
+public class UiIndexService {
```
loicmathieu, line 34: I don't understand why we can't just read and return the index, and why we had to update some many things inside it.

webserver/src/main/java/io/kestra/webserver/services/UiResourceCacheService.java
```diff
@@ -0,0 +1,159 @@
...
+import io.micronaut.http.HttpHeaders;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.http.HttpResponse;
+import io.micronaut.http.MediaType;
+import io.micronaut.http.MutableHttpResponse;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+
+/**
+ * Bounded in-memory cache of static UI resources (main UI and plugin UI), each entry holding the raw bytes,
```
loicmathieu: And what happen if a browser disable compression?
We might implement a slow-path with decompression from GZIP.

webserver/src/main/java/io/kestra/webserver/services/UiResourceCacheService.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private static <T> MutableHttpResponse<T> applyCacheHeaders(MutableHttpResponse<T> response, String etag, String cacheControl) {
+        return response
+            .header(HttpHeaders.ETAG, etag)
+            .header(HttpHeaders.CACHE_CONTROL, cacheControl)
+            .header(HttpHeaders.VARY, HttpHeaders.ACCEPT_ENCODING);
+    }
+
+    private static CachedUiResource build(MediaType mediaType, byte[] raw) {
```
loicmathieu: We store 3 byte arrays for each resources, which means big resources can occupies a large fraction of the cache.

I think we should have a safety guardrails that avoids putting inside the cache too big resources.
We might also avoid storing GZIP/BROTLI if the compression ratio is not good (for example if it's less than 50%).

webserver/src/main/java/io/kestra/webserver/utils/UiResourceCompression.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private UiResourceCompression() {
+    }
+
+    public static boolean isBrotliAvailable() {
+        return BROTLI_AVAILABLE;
+    }
+
+    public static byte[] gzip(byte[] raw, int level) {
+        ByteArrayOutputStream out = new ByteArrayOutputStream(Math.max(64, raw.length / 3));
```
loicmathieu: It should be properly closed.

webserver/src/main/java/io/kestra/webserver/utils/UiResourceCompression.java
```diff
@@ -0,0 +1,72 @@
...
+                raw,
+                // Quality 5 keeps the one-time compression cost per resource low while still beating gzip on size.
+                new com.aayushatharva.brotli4j.encoder.Encoder.Parameters().setQuality(5)
+            );
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private static boolean detectBrotli() {
```
loicmathieu: Move this inside a static block so it's done only once at class loading time

webserver/build.gradle
```diff
@@ -2,6 +2,9 @@ configurations {
     implementation.extendsFrom(micronaut)
 }
 
+// Not managed by the Micronaut BOM; used compileOnly in main and on the test classpath to exercise the brotli path.
+def brotli4jVersion = "1.18.0"
```
loicmathieu: Move the version inside the platform

webserver/build.gradle
```diff
@@ -24,6 +27,10 @@ dependencies {
 
     implementation "io.micronaut.security:micronaut-security-csrf"
 
+    // Optional brotli precompression for the UI asset cache: only active when brotli4j and its
+    // platform natives are present on the runtime classpath (see AssetCompression).
+    compileOnly "com.aayushatharva.brotli4j:brotli4j:${brotli4jVersion}"
```
loicmathieu: I'm not sure I understand compileOnly here, it would not be integrated into the final artefact so, well, it cannot work?

loicmathieu (changes requested):

webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java
elevatebart: Done in 8f91347: both points. Files above `kestra.webserver.ui-resource-cache.max-entry-size` (default 8MB) never enter the cache and are streamed straight from the jar via `StreamedFile`, exactly as before — for plugin UI files and `/ui/**` alike. The size comes from the zip entry metadata (`URLConnection.getContentLengthLong()`), so the check itself doesn't inflate anything.

---

webserver/build.gradle
elevatebart: You're right that `compileOnly` made brotli opt-in only (the availability probe fell back to gzip when the jar was absent, so it worked but never activated in our own distribution). 8f91347 makes it real: `implementation` for brotli4j plus `runtimeOnly` per-platform natives (linux/osx/windows, x86_64 + aarch64), versions managed in the platform BOM. When a native can't load, `Brotli4jLoader.isAvailable()` is false and we serve gzip only.

---

webserver/src/main/java/io/kestra/webserver/services/UiIndexService.java
elevatebart: The rewrites aren't new — the old `StaticFilter` did all of them (base path, `KESTRA_GOOGLE_ANALYTICS`, `html-title`, `html-head`) with string replaces on every single request, plus the CSRF meta. This class does the config-driven ones once into an immutable template; the only per-request work left is the CSRF meta concatenation, which can't be precomputed because the token is per-user. So it's the same behavior as before, moved from per-request to once. "Just read and return the index" would drop the base-path rewrite (broken assets behind a context path), the html-title/html-head/GA options, and the CSRF token the login flow needs.

---

webserver/src/main/java/io/kestra/webserver/services/UiResourceCacheService.java
elevatebart: Both done in 8f91347: resources above `kestra.webserver.ui-resource-cache.max-entry-size` (default 8MB) never enter the cache and are streamed from the jar, and a compressed variant is only stored when it at least halves the raw size (cutoff moved from 0.9 to 0.5), so at most ~2x the raw bytes are ever resident per entry.

---

webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java
AcevedoR: this was done

webserver/src/main/java/io/kestra/webserver/utils/UiResourceCompression.java
AcevedoR: done

webserver/src/main/java/io/kestra/webserver/utils/UiResourceCompression.java
AcevedoR: done

AcevedoR:
## Benchmark: precompressed streaming (final) vs the in-memory cache (first iteration)

Validating that replacing the cache with build-time precompression (commit 9d1cba9) keeps the gains. Same A/B protocol as the earlier benchmark (k6 + headless Chromium, 30 cold-boot VUs, `stress` scenario, both sides source-built on the same machine, fresh VM + cold JVM + wiped Postgres per side, same browser VM): **A** = this PR at the in-memory-cache iteration (`74582e7`), **B** = this PR final (precompressed streaming, `9d1cba9`).

| metric | A (in-memory cache) | B (precompressed streaming) |
|---|---|---|
| sessions past login | 57 | 54 (run-to-run noise) |
| login avg / p95 | 7.5 s / 21.4 s | **6.3 s** / 21.3 s |
| TTFB p75 | 88 ms | **26 ms** |
| FCP p75 / LCP p75 | 2.4 s / 3.3 s | **1.8 s / 2.2 s** |
| browser HTTP avg / p95 | 0.2 s / 0.9 s | **0.1 s / 0.6 s** |
| checks passed | 369/369 | 356/356 |

Every per-action latency (dashboard, flows, logs, search, save) also improved or held; 100% functional checks on both sides.

**Single-user cold-start** (the #9947 kill scenario — one uncached browser, ~390 asset requests, first traffic a fresh JVM sees; measured as server CPU-seconds consumed):

| build | cold first load | warm repeat |
|---|---|---|
| develop (compress on the fly) | 18.0 CPU-s | 13.6 CPU-s |
| in-memory cache iteration | 5.3 CPU-s (first visitor fills the cache) | 0.75 CPU-s |
| **precompressed streaming (final)** | **1.7 CPU-s** | 1.0 CPU-s |

The cache variant still paid a ~5 CPU-second compress-everything tax on the very first page load — precisely the moment a CPU-limited pod is also class-loading and JIT-ing. Precompression removes that tax entirely, which is why B beats A on TTFB/FCP/LCP under cold-boot stress: there is no first-visitor penalty left.

Also verified live: `Accept-Encoding` negotiation serves `br` to browser-like clients, `gzip` to gzip-only clients, raw bytes to clients with no `Accept-Encoding`, and `If-None-Match` answers `304` without opening a stream. Since responses carry an explicit `Content-Encoding`, Netty's dynamic compressor (zstd on EE, with its per-response direct-memory buffering) is bypassed for the whole UI asset path.

Reports: `benchmark-internal` `results/browser/roman-oss-postgres-20260819-171903` (A) / `-175922` (B).

webserver/src/main/java/io/kestra/webserver/controllers/api/PluginController.java
elevatebart: Resolved by the rework in 9d1cba9: plugin UI files are never buffered at all. `getPluginUi` always returns a `StreamedFile`, as it did before this PR — the entity tag now comes from the plugin UI source hash, so revalidation costs no read either. There is no `byte[]` on this path any more, and no server-side cache to size.

---

webserver/src/main/java/io/kestra/webserver/services/UiResourceCacheService.java
elevatebart: Superseded — please ignore my earlier reply about `max-entry-size` and the 0.5 ratio cutoff, that machinery is gone. 9d1cba9 removed the in-memory cache entirely: the UI build now emits `.gz`/`.br` siblings (`ui/plugins/precompressAssets.js`), and a request streams the negotiated variant straight from the classpath. Nothing is stored in memory, so there is no cache to bound and no compression ratio to judge at runtime — and the first visitor no longer pays a compress-everything tax (~5 CPU-s in the cache iteration, per the benchmark above).

---

webserver/src/main/java/io/kestra/webserver/controllers/api/UiController.java
```diff
@@ -0,0 +1,193 @@
...
+    @ExecuteOn(TaskExecutors.IO)
+    public HttpResponse<?> serve(HttpRequest<?> request, @PathVariable String path) {
+        String normalized = path == null ? "" : path;
+        if (normalized.contains("..") || normalized.contains("\\") || normalized.contains("\0")) {
+            return HttpResponse.notFound();
+        }
+
+        // Only paths whose last segment has an extension can be static resources; SPA routes never resolve here.
+        if (isStaticResourceCandidate(normalized) && !INDEX_HTML.equals(normalized)) {
+            URL identityUrl = UiController.class.getClassLoader().getResource("ui/" + normalized);
```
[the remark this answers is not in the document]
elevatebart: Hardened in c3d4abb. The inline blacklist is replaced by `FileUtils.isSafeRelativePath`, a sibling of the existing `isParentTraversal` in core: it is segment-aware, treats backslashes as separators (as a directory-based classpath entry would on Windows), and refuses absolute paths and NUL bytes. Anything that passes it cannot escape the `ui/` prefix it is appended to, so `getResource` only ever sees a name inside the UI tree.

`FilenameUtils.getName(...)` isn't usable here — the UI legitimately serves subdirectories (`assets/app-abc123.js`), so stripping the directory part would break every asset.

Covered by unit tests on the helper (traversal, backslash, absolute, NUL, plus a file name that merely contains `..`, which the old check wrongly rejected) and by end-to-end cases asserting `/ui/**` traversal attempts return 404 rather than falling back to `index.html`.

---

AcevedoR:
this pr was a bit too complex, we ended up only keep the UiController and re-tough the cache + some optimizations and the perf were good again

PR #18193, opened 2026-08-16
feat(core): bring the AI Copilot free tier back to OSS

ammeek:
### ✨ Description

**A fresh OSS install has a working Copilot again.**

AI Copilot currently requires an LLM API key. Without one there is no Copilot at all — the endpoints answer
`503` and the feature reads as broken. The previous free tier (one-shot flow authoring relayed through
api.kestra.io) was removed during the Copilot v2 agentic work and nothing replaced it, so a fresh install, a
demo instance and a self-serve trial have had nothing.

This restores it, and restores the *real* Copilot rather than a reduced one. When no provider is configured,
Copilot registers a hosted provider pointed at `https://api.kestra.io/v1/ai/relay/gemini`. The instance keeps
the prompts, the tools and the agent loop — it is the same `GeminiAiService` with a different base URL and an
identity header — so the free tier cannot lag the product. The key, the model and the allowance live at the
far end.

> ⚠️ **Data flow, please read.** On this path prompts, the flow YAML being authored, and tool results that
> can include flow source and execution logs are sent to `api.kestra.io`. It is **enabled by default**, and
> only when the instance has configured no provider of its own — an explicitly configured provider always
> wins and is never displaced, even if it fails to build. Set `kestra.ai.free-tier.enabled: false` to turn
> it off.

**Showing a user where they stand.** A budget is only usable if it is visible, so the composer shows the
share of the allowance left, warns as it runs down, explains itself in a tooltip, and disables send once it
is exhausted. The percentage is computed server-side and rendered as returned — a figure computed on both
sides is one that eventually disagrees with the one being enforced. `GET /api/v1/{tenant}/ai/usage` is the
endpoint behind it, so an exhausted allowance is visible *before* a turn is spent discovering it.

**Usage accounting.** Making any of the above true needed per-model-call token accounting on the instance:
prompt tokens, the cached share of the prompt, generated tokens and thinking tokens, recorded against the
provider that served the call. Counts are stored as reported rather than pre-weighted, so changing the
weights re-prices history instead of invalidating it. New `ai_usage` table, migration `2.0.23-ai-usage`
(H2/Postgres/MySQL), token counts as generated columns so a window's total is one indexed `SUM`.

**Spend limits for *any* provider — a deliberate extension of scope.** Once the instance could measure and
display spend for the hosted provider, restricting that to the hosted provider would have meant special-casing
it throughout. Generalising it instead was comparable work, so any provider may now declare a `usage-limit`
block and get the same accounting, the same display and the same enforcement against its own key.

Ceilings are in *weighted* units, where one unit is one cold input token — cached input bills at a fraction
of cold and output at a multiple of it, so a raw token sum can be wrong by an order of magnitude. Two axes:
installation-wide (`max-weight`) and per-caller (`user-max-weight`), both counted across every tenant, since
a provider key belongs to the installation and is billed to it as one bill. Windows are calendar periods
(`DAILY`/`WEEKLY`/`MONTHLY`, UTC), so a refused caller can be told a date. Enforcement is check-then-charge
per model call: refused up front with `429`, and a ceiling reached mid-turn ends the turn gracefully rather
than failing it, with the reason streamed and persisted.

Nothing is shown or enforced for a provider that declares no limit, which is the default. Recording happens
either way, so switching a limit on reports against history that already exists.

**Metrics.** Counters for model calls and each token class, tagged by provider, model and tenant; gauges for
weighted spend, the ceiling and the remaining share; counters for turns refused for want of allowance and
for usage that could not be stored. Written from the same counts that go to the database, so a dashboard and
the stored history cannot disagree.

### 🔗 Related Issue

Part of https://github.com/kestra-io/kestra-ee/issues/9832


webserver/src/main/java/io/kestra/webserver/services/ai/AiFreeTierLimitProvider.java
```diff
@@ -0,0 +1,73 @@
...
+     * <p>Filtered on {@code enabled} as a declared provider's own ceiling is, so a relay reporting a
+     * switched-off limit reads as no ceiling rather than an active one with nothing behind it.
+     */
+    public Optional<AiUsageLimitConfiguration> limit() {
+        return Optional.ofNullable(fetched.get()).filter(AiUsageLimitConfiguration::enabled);
+    }
+
+    /** Picks up a re-price or a re-size without a restart. Delayed past startup, since no turn can have run yet. */
+    @Scheduled(fixedDelay = "1h", initialDelay = "30s")
+    public void refresh() {
```
fhussonnois: I would prefer to not use a scheduled thread here; best would be to refresh the limit on read i.e., when `limit()` is called check whether a a refresh should be trigger asynchronously.

fhussonnois (changes requested):

PR #18208, opened 2026-08-17
fix(executions): stop a cleared input being refilled with its default

### LINKED ISSUE #17897: Input fields for UI to enter the input is not proper
### Describe the issue

Steps to Reproduce:

-   Create a work flow which requires input and has a default value and  execute the flow
- It will give a prompt to enter teh input when we try to enter the value...the default value 
- When we try to delete and enter teh input it is overriding the input value.

Priority: Normal

Please check the below attached Screen Recording
https://drive.google.com/file/d/1OUq5mCtM0i8Ibp3bhDD3grAR5dI5qCaa/view?usp=sharing

### Environment

- Kestra Version: develop

RaunaqDavidNath:
Closes https://github.com/kestra-io/kestra/issues/17897.

An execution input that has a `defaults` value refills itself after you clear it.
Clearing the field submits nothing on the validate round-trip, because
`normalizeInputValues` drops empty strings from the payload, so the backend does the
right thing for its view of the world and answers `isDefault: true` with the resolved
default. `metadataCallback` then replaces the whole metadata array with that response,
throwing away the `isDefault false` that `onChange` had set, and `updateDefaults` writes
the default back into the field. The form has no way to tell "never filled in" from
"deliberately cleared". Since STRING inputs render in Monaco, the refill moves the cursor
too, which is what mangles a value while it is being typed.

InputsForm now records which inputs the user has edited and skips those when applying
defaults. Inputs with an `expression` or `dependsOn` are left out of the guard on purpose,
since the server still renders their values and a blanket skip would freeze them, and the
set is cleared alongside the validation cache so that selecting another flow can apply its
own defaults.

Two things worth knowing if you review this.

Reproducing it needs a value on the wire first. `validateInputs` skips a round-trip whose
payload signature has not changed, and for a flow with a single defaulted input the
untouched state and the cleared state both submit nothing, so clearing the field and
waiting shows nothing at all. You have to type a value, let the 500ms debounce fire, and
only then clear it. Typing and clearing back to back coalesces into a single validate of
the empty payload and is skipped.

A spurious `onChange` costs more after this change than before it. Previously a wrong
`isDefault false` healed itself on the next validate response; now the input stays
protected for the life of the component. Given the el-switch history documented in
`inputsFormBoolDefault.spec.ts` I am happy to gate the recording on the `ready` emit
instead if you would prefer that.

Tests: a Vitest spec covering the two clearing regressions (plain and FORM-nested) plus
three guards, and a `ClearedDefault` Storybook story that drives the real Monaco editor in
Chromium. The two exercise different round-trip paths, the executions store for the spec
and the `emit("validation")` branch for the story. Reverting the fix fails both; reverting
only the cache clear fails only the flow-swap test. Full unit suite 1774 passing, Storybook
245 passing, lint and `check:types` clean. Also verified by hand against a local server,
video below.

https://github.com/user-attachments/assets/fa3a0867-a0af-4c34-9a78-3966dfe44def


elevatebart (approved):

RaunaqDavidNath:
Updated the branch. The earlier Codespell failure was from 'useDependencies.ts'
on the old base, fixed upstream since. @elevatebart mind approving the workflow
run?

RaunaqDavidNath:
Gentle ping on this one. Approved and green since Aug 25, no conflicts. @MilosPaunovic @loicmathieu anything else needed from me, or is this good to merge?

PR #18209, opened 2026-08-17
feat(core): Add OTLP trace forwarding from kotlp

### LINKED ISSUE #18107: Forward kotlp OTLP traces to Kestra
Follow-up to #17873, which parses kotlp's OTLP telemetry and forwards **logs** and **metrics** to the `RunContext`, but deliberately leaves **traces** untouched. They are parsed and exposed as POJOs (`match.otlp().resourceSpans()`, and the records returned by `TaskLogLineMatcher#parseOtlp`), and nothing consumes them yet.

## What kotlp already gives us

Two kinds of spans arrive on the same channel as the logs and metrics:

1. **A root execution span**, emitted once when the wrapped process exits — `name: "exec <command>"`, `kind: 1` (INTERNAL), start/end in unix-nanos, `process.exit.code` (plus `process.exit.signal` when killed, and `kotlp.log.file.count` when `--log-flush-interval` is set), and `status.code` `1`/`2` for success/failure.
2. **Whatever the child itself exports.** kotlp runs an embedded OTLP/HTTP receiver and points the child at it through `OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_EXPORTER_OTLP_PROTOCOL`, `OTEL_TRACES_EXPORTER=otlp` and `OTEL_SERVICE_NAME`. So an instrumented script — a Python task using the OTel SDK, say — has its spans relayed through kotlp as ordinary `resourceSpans` records, with no extra configuration from the user.

The second half is the interesting one: it means a task's *internal* trace can reach Kestra, not just the fact that a process ran.

## What is missing on the Kestra side

- **A sink.** Logs go to `runContext.logger()` and metrics to `runContext.metric(...)`; there is no equivalent for spans. Deciding what that is — emit onto Kestra's own OTel tracer, persist alongside the task run, or both — is the bulk of this issue.
- **Trace stitching.** kotlp mints its own trace id per run (`kotlp_rand_hex` in `src/main.c`) and does not read `TRACEPARENT` from its environment, so its spans form an island disconnected from the execution's trace. Two ways out, not exclusive:
  - teach kotlp to honour an inbound W3C `traceparent` and use it as the parent of the root span, so the child's spans nest under the task run for free (an upstream change in kestra-io/kotlp);
  - re-parent on ingestion in Kestra, rewriting the trace id and setting the root span's parent to the current task-run span.

  The first is cleaner and also fixes any consumer other than Kestra; the second works without an upstream release.
- **A decision on volume.** A chatty instrumented task can export a lot of spans. Whether they are all forwarded, sampled, or gated behind an opt-in on the task-runner options is worth settling before this ships.

## Related

- Also worth considering as part of this: forwarding `traceId`/`spanId` onto log records so a task log can be pivoted to its span. That needs a change on both ends — kotlp does not currently stamp trace context onto log records (only spans carry it), and `LogEntry` has no field for it, so a key-value on the logging event would be dropped by `RunContextLogger`. See the discussion on #17873.
- Depending on how far the span model has to go, this may be the trigger to swap the hand-rolled `OtlpRecord` POJOs for the generated `io.opentelemetry.proto` classes — the trade-off is written up on #17873.

hemantmm:
closes: #18107

---

### ✨ Description

What does this PR change?  

This PR implements OTLP trace forwarding from `kotlp` to Kestra's tracing backend. It builds upon previous work that parses `kotlp`'s OTLP telemetry (logs and metrics) by forwarding the instrumented spans (traces) into Kestra while preserving the original nanosecond timestamps, attributes, and span hierarchy.

## Changes

- Added `traceId` and `spanId` to `LogEntry` and propagated them through the logging pipeline.
- Updated `TaskLogLineMatcher` and `RunContextLogger` to preserve OTLP trace context.
- Added `OtlpSpanForwarder` to:
  - Recreate OTLP spans in Kestra's OpenTelemetry context.
  - Preserve span start/end timestamps with nanosecond precision.
  - Map span kinds, attributes, events, links, and status.
- Added `forwardTraces` to `KotlpOptions` as an opt-in feature.
- Propagate the `forwardTraces` option through the script execution logging pipeline.
- Added tests for OTLP trace context propagation and trace processing.


hemantmm:

hemantmm:

PR #18214, opened 2026-08-17
fix(worker): surface gRPC message size rejections in the execution logs

fhussonnois:
### ✨ Description

When a worker job or a task result is rejected because it exceeds a gRPC inbound message size limit, the failure was only ever logged server-side. The user got a `FAILED` task run with **no log line at all**, so nothing explained why the execution failed or how to fix it.

Both directions of the worker/controller exchange are fixed, and each now emits an actionable ERROR log attached to the task run (or the trigger):

1. **Controller → worker** (`WorkerJobDispatcher.rejectOversizedJob`): the job payload exceeds the worker's advertised inbound limit, so the job is failed before dispatch. The reason went to the controller's own logger only, and the class had no access to the log queue at all. It now emits a `LogEntry` through `RunContextLoggerFactory` for both task and trigger jobs.

2. **Worker → controller** (`GrpcWorkerIOSenderFactory.taskResultSender`): the serialized task result exceeds the controller's inbound limit, so the fallback resends a failed result without outputs. The reason went through `Logs.logTaskRun`, which — despite the task-run MDC it prints — is a *server* logging utility that never puts a `LogEntry` on the worker log queue. It now also emits through `LogEntryEmitter`.

3. While verifying (2) in the UI, the failed task run showed a phantom **"Attempt 1/2 Success"** with the error log on the successful attempt. Cause: `TaskRun.fail()` *appends* an attempt, and the worker's result already carries its own. The fallback now uses `withStateAndAttempt(FAILED)`, which fails the attempt that actually ran. As a side effect the attempt count stays accurate, so a task with a `retry` policy no longer burns an extra attempt on this path.

The message the user now sees:

> Failed to send the task result to the worker controller: its serialized size exceeds the maximum inbound gRPC message size configured on the controller. The task run is failed and its outputs are dropped. Reduce the size of the task outputs, or increase `kestra.grpc.max-inbound-message-size` on the worker controller.

### 🔗 Related Issue

No linked issue — found while investigating a misconfigured `kestra.grpc.max-inbound-message-size` (the default is 10 MB).

### 🛠️ Backend Checklist

### 📝 Additional Notes

- Log entries are anchored to the zero-based index of the attempt they belong to (`attempts.size() - 1`), which is how the UI groups them (`TaskRunDetails.vue` uses `taskRun.attempts.length - 1`). Using the raw attempt count would have attributed the line to a non-existent attempt and it would have stayed invisible — the tests pin this.
- Verified: `WorkerJobDispatcherTest` (62 tests, 2 new), `GrpcWorkerIOSenderTest` (5 tests, extended), `H2RunnerTest` (99 tests) green.
- The oversized-result assertions were folded into the existing `shouldSendFailedResultWithoutOutputsWhenTaskResultIsTooLarge` rather than added as a second test: two 2 MB sends in one class break each other, since the server-side deframer failure tears down the shared channel for whichever runs second.
- Still not covered: the `LogEntry` batch sender has no fallback, so a log batch that exceeds the controller limit is dropped silently. There is no non-recursive way to report that through the log path, so it is left as a separate problem.

### 🤖 AI Authors

Why did the cat get thrown off the gRPC stream? It exceeded the maximum inbound *meow*-sage size. 🐱

### DIFF
--- a/worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
+++ b/worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
@@ -1043,8 +1043,6 @@ public class WorkerJobDispatcher {
 
         // Fail the job cleanly so the execution reaches a terminal state.
         if (job instanceof WorkerTask workerTask) {
-            // The worker never receives the job so it never logs the reason: without this the user
-            // only sees a FAILED task run with no logs.
             emitJobLog(
                 runContextLoggerFactory.create(workerTask),
                 LogEntry.of(workerTask.getTaskRun(), workerTask.getExecutionKind()),
--- a/worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
+++ b/worker-controller/src/main/java/io/kestra/controller/grpc/services/WorkerJobDispatcher.java
@@ -2,6 +2,7 @@ package io.kestra.controller.grpc.services;
 
 import java.io.IOException;
 import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


Skraye (approved):
LGTM

PR #18216, opened 2026-08-18
fix(ui): preserve plugin documentation sections after refresh

### LINKED ISSUE #17967: [2.0] Plugin Docs - tabs don't show up when first clicking on task
### Describe the issue

Steps to reproduce:
- Open a flow (refresh page to make sure it's a fresh load)
- Click on a task like Python which has loads of docs
- Overview shows, but Properties, Outputs and Examples don't

https://github.com/user-attachments/assets/0274107f-48e8-4cf5-9e79-8761776fb449

Missing tabs:
Image

What it should look like:

Image

### Environment

- Kestra Version: RC7 2.0 OSS

dhanusharer:
## What does this PR do?
Fixes #17967, where the plugin documentation panel could show only the
**Overview** section after a full page refresh. Sections such as
**Properties**, **Outputs**, and **Examples** would disappear until the
user switched to another task and back.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change

## Root cause
During the initial task resolution, the same plugin can be assigned to the
documentation state more than once:
- first with an undefined version
- later with the resolved plugin version

`PluginDocumentation.vue` was watching the entire `currentPlugin` object.
Because the store creates a new object during these updates, the watcher
reset `sectionCounts` even when the plugin itself had not changed.
At that point, `SchemaToHtml.vue` had already emitted the section counts, so
the reset could leave the documentation navigation with only **Overview**.

## Fix
The watcher now tracks the plugin `cls` and `version` values instead of the
whole object reference.

The documentation state is reset when:
- the plugin class changes
- the version changes from one defined version to another

It is not reset when:
- the same plugin is represented by a new object reference
- the plugin version changes from `undefined` to a resolved version for the
  same plugin class

This preserves the section counts during the initial refresh lifecycle while
still resetting the documentation state when the actual plugin changes.

## Tests
Added regression tests covering:
- same `cls` and `version` with a new object reference
- plugin class changes
- defined plugin version changes
- `undefined` → defined version transition for the same plugin class

Focused test result:
```text
Test Files  1 passed
Tests       4 passed
```

## Checklist
- [x] Self-reviewed the diff
- [x] Added/updated tests
- [ ] Updated docs (n/a — internal component behavior only)

### DIFF
--- a/ui/src/components/plugins/PluginDocumentation.vue
+++ b/ui/src/components/plugins/PluginDocumentation.vue
@@ -375,10 +375,21 @@
         sectionCounts.value = counts
     }
 
-    watch(currentPlugin, () => {
-        activeSection.value = "overview"
-        sectionCounts.value = {}
-    })
+    watch(
+        [() => currentPlugin.value?.cls, () => currentPlugin.value?.version],
+        ([cls, version], [oldCls, oldVersion]) => {
+            const clsChanged = cls !== oldCls
+            const versionChanged =
+                oldVersion !== undefined &&
+                version !== undefined &&
+                version !== oldVersion
+
+            if (clsChanged || versionChanged) {
+                activeSection.value = "overview"
+                sectionCounts.value = {}
+            }
+        },
+    )
 
     const selectSection = (id: string) => {
         activeSection.value = id
--- a/ui/src/components/plugins/PluginDocumentation.vue
+++ b/ui/src/components/plugins/PluginDocumentation.vue
@@ -47,13 +47,14 @@
                     <p v-if="pluginSummary" class="dp-summary">{{ pluginSummary }}</p>
                 </div>
 
-                <nav class="dp-nav" aria-label="Documentation sections">
+                <nav class="dp-nav" aria-label="Documentation sections" data-test="plugin-doc-nav">
                     <button
                         type="button"
                         v-for="chip in navChips"
                         :key="chip.id"
                         class="dp-navchip"
                         :class="{active: activeSection === chip.id}"
+                        :data-test="`plugin-doc-navchip-${chip.id}`"
                         @click="selectSection(chip.id)"
                     >
                         {{ chip.label }}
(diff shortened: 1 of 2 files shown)


dhanusharer:
Hello @Piyush-r-bhaskar , kindly review the PR when you have a chance. Thank you!

MilosPaunovic:
Hi @dhanusharer,

Piyush-r-bhaskar:
hey @dhanusharer 

I am unable to reproduce this issue. It seems to be transient.

Can you share how you reproduced and was there anything in devtools console ?

dhanusharer:
Hi @Piyush-r-bhaskar, thanks for checking.

I was able to reproduce the issue locally with the following sequence:

1. Start the Kestra UI and open the Flow Editor.
2. Create or open a flow containing `io.kestra.plugin.scripts.python.Script`.
3. Open the plugin documentation by clicking the task `type`.
4. Before refreshing, the documentation panel shows `Overview`, `Properties`, `Outputs`, and `Examples`.
5. Perform a full page refresh while staying on the same flow/editor page.
6. Select the same Python task again.

Under this sequence, the documentation panel can come back with only `Overview`. Switching to another task and then back to the Python task causes the other sections (`Properties`, `Outputs`, and `Examples`) to appear again.

During the investigation, I found that the plugin documentation state can be updated multiple times during the initial task-resolution lifecycle. In particular, the same plugin class can transition from an undefined version to its resolved version. The previous watcher on the whole `currentPlugin` object cleared `sectionCounts` during that update, while `SchemaToHtml` did not necessarily emit the section counts again, leaving the panel with only the `Overview` section.

I added regression coverage for this lifecycle, including:
- the same plugin being assigned through a new object reference
- plugin class changes
- defined plugin version changes
- the `undefined -> defined` version transition for the same plugin class

The focused test suite passes 4/4.

Piyush-r-bhaskar (approved):
Thanks for the contribution 🎉

PR #18222, opened 2026-08-18
fix(core): batch of small UI polish fixes across executions, flows, design-system, and topology

loicmathieu:
## Summary

A batch of small, independent UI fixes reported on GitHub. Each is a minimal, surgical change scoped to its own commit.

- **Chevron spacing in execution logs** — the expand/collapse chevron had no margin of its own and sat flush against the task name whenever the task-icon column collapsed. Added spacing.
image

- **Duration/status/menu vertical alignment in execution logs** — the duration text's default line-height made its box taller than its icon-based siblings, so it looked higher even though all three are flex-centered. Normalized with `line-height: 1`.
image

- **Wrong toast label on Restart** — clicking "Restart" showed "Execution is Replayed" because the success toast was hardcoded to the `replayed` i18n key regardless of which action (Restart vs Replay) triggered it. Now derives the key from the existing `replayOrRestart` computed.
image

- **KsInputNumber increment/decrement border visibility** — the +/- buttons drew their border via a token mapped to `--ks-border-default`, while the input wrapper uses a stronger token (`--ks-border-strong`), making the buttons' edges nearly invisible next to the input. Aligned the tokens.
image

- **Draft badge alignment in revision dropdown** — `.revision-label` had no `align-items`, defaulting to `stretch`, misaligning the "Draft" `KsTag` badge against its sibling text.
image

- **Runner-badge overlap in topology task node** — the badge had `align-self: flex-start` inside a flex-column parent, so it was never constrained to the parent's shrunk width — its own `overflow:hidden`/ellipsis never had a box to clip against. Once a long-running task's duration text grew wide enough to squeeze the row, the badge rendered at full width and overlapped the task name. Added `max-width: 100%`.
image

- **Double border on the toggled expression editor** — a field's own wrapper drew a real `border`, stacked on top of `KsEditor`'s own inset `box-shadow` "border" on the same element, producing a visible double ring once a string field (e.g. `lateMaximumDelay`) was switched to expression/pebble mode. Removed the redundant border and exposed a `--ks-editor-single-line-border-color` CSS variable from `KsEditor` so the focus-state color change still reaches the correct inner ring.
image

## Test plan

- [x] Each fix reviewed via a dedicated code-review pass (no blocking issues found)
- [x] Manual/visual verification in a running browser (not performed in this environment — see individual commit messages for reasoning)

Closes kestra-io/kestra-ee#9305
Closes kestra-io/kestra-ee#9429
Closes kestra-io/kestra-ee#9606
Closes kestra-io/kestra-ee#8960
Closes kestra-io/kestra-ee#9273
Closes kestra-io/kestra-ee#9462
Closes kestra-io/kestra-ee#9482

### DIFF
--- a/ui/packages/design-system/src/components/Form/KsEditor.vue
+++ b/ui/packages/design-system/src/components/Form/KsEditor.vue
@@ -233,7 +233,7 @@
                 background-color: var(--ks-bg-input);
                 border-radius: var(--kel-input-border-radius, var(--kel-border-radius-base));
                 transition: var(--kel-transition-box-shadow);
-                box-shadow: 0 0 0 1px var(--ks-border-default) inset;
+                box-shadow: 0 0 0 1px var(--ks-editor-single-line-border-color, var(--ks-border-default)) inset;
 
                 &.custom-dark-vs-theme {
                     background-color: var(--ks-bg-input);
--- a/ui/packages/design-system/src/components/Form/KsInputNumber.vue
+++ b/ui/packages/design-system/src/components/Form/KsInputNumber.vue
@@ -52,6 +52,7 @@
 
         .kel-input-number__increase, .kel-input-number__decrease {
             background: var(--ks-bg-surface);
+            border-color: var(--ks-border-strong);
         }
 
         &.kel-input-number--small {
--- a/ui/packages/topology/src/nodes/TaskNode.vue
+++ b/ui/packages/topology/src/nodes/TaskNode.vue
@@ -481,6 +481,7 @@ button.playground-button {
 
 .runner-badge {
     align-self: flex-start;
+    max-width: 100%;
     margin-bottom: var(--ks-spacing-1);
     padding: 0 var(--ks-spacing-2);
     border-radius: var(--ks-radius-base);
--- a/ui/src/components/executions/TaskRunLine.vue
+++ b/ui/src/components/executions/TaskRunLine.vue
@@ -243,6 +243,7 @@
         .task-duration small {
             white-space: nowrap;
             color: var(--ks-text-secondary);
+            line-height: 1;
         }
 
     }
--- a/ui/src/components/executions/TaskRunLine.vue
+++ b/ui/src/components/executions/TaskRunLine.vue
@@ -4,7 +4,7 @@
         class="taskrun-header"
         :style="{'--depth': depth}"
     >
-        <div>
+        <div class="me-1">
             <KsIcon
                 v-if="!taskRunId && shouldDisplayChevron(currentTaskRun)"
                 type="default"
--- a/ui/src/components/executions/overview/components/actions/Restart.vue
+++ b/ui/src/components/executions/overview/components/actions/Restart.vue
@@ -374,7 +374,7 @@
 
         const newExecution = response
 
-        toast.success(t("replayed"))
+        toast.success(t(props.isReplay ? "replayed" : "restarted"))
 
         if (newExecution.id !== props.execution.id) {
             window.location.href = router.resolve({
--- a/ui/src/components/layout/Revisions.vue
+++ b/ui/src/components/layout/Revisions.vue
@@ -441,6 +441,7 @@
 
         .revision-label {
             display: flex;
+            align-items: center;
             gap: var(--ks-spacing-2);
         }
     }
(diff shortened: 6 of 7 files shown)


MilosPaunovic (commented):
Reviewed all 7 commits. Verified each root-cause claim against the source. Nothing blocking, but two things I would change before merge.

### Verified correct

- **`KsInputNumber`** - the premise checks out. `ui/packages/design-system/src/components/Form/KsInput.vue:78` sets `--kel-input-border-color: var(--ks-border-strong)` globally on `.kel-input`, which `ElInputNumber`'s inner `ElInput` inherits, while the +/- buttons were still on `--el-border-color`. Real mismatch. Specificity is safe: element-plus's `@include when(disabled)` resolves to `.el-input-number.is-disabled .el-input-number__increase` (0,3,0) and still beats the new rule (0,2,0), so the disabled state is unaffected. One thing worth eyeballing: the reported issue (Display settings) uses `controlsPosition="right"`, where this border is the divider between the input and the stacked buttons rather than an outer edge. That variant is also used in EE at `ui-ee/src/components/cluster/WorkerGroupQueueSubscription.vue`.
- **`TaskNode` runner badge** - correct diagnosis. `align-self: flex-start` makes the item shrink-to-fit in a column flex parent, so `overflow: hidden` + ellipsis had no box to clip against. `max-width: 100%` is the right one-liner.
- **`TaskString` / `KsEditor` CSS-var hook** - sound. Custom properties inherit into the child component subtree regardless of `scoped`, and the `.wrapper--toggle > .string-editor` selector demonstrably already matched (it is what drew the outer half of the double border), so the new declaration lands. `single-line` comes from `props.inline` (`ui/packages/design-system/src/composables/useKsEditor.ts:163`), which `TaskString` passes, so the inner ring is always present.
- **`Revisions.vue`** - fine.
- No EE counterparts exist for any of the 7 files, so no EE companion PR is needed, and the closing keywords correctly use the cross-repo `kestra-io/kestra-ee#N` form.

### Worth changing

**1. `Restart.vue` - the dynamic i18n key is a regression in greppability**

```js
toast.success(t(`${replayOrRestart.value}ed`))
```

The behaviour is right, I confirmed `replayed` and `restarted` both exist in all 13 locales. But building the key by string concatenation makes both keys invisible to any literal-key scan, which is exactly what the ghost-translations tooling keys off. We already maintain a manual never-remove list for dynamic keys because of this. Suggest:

```js
toast.success(t(props.isReplay ? "replayed" : "restarted"))
```

Same behaviour, both keys stay greppable, no new entry on the exceptions list.

**2. PR title does not match the commit convention.** `fix(ui):` - `ui` is not in the allowed scope list, and the title enumerates four areas where the convention is one scope only. The individual commits are all correctly scoped (`executions`, `flows`, `design-system`, `topology`), so the PR title just needs to pick one, e.g. `fix(core): batch of small UI polish fixes`.

### Non-blocking notes

**3. The chevron spacing fix layers a third spacing mechanism.** `.taskrun-header` already has `gap: var(--ks-spacing-2)`, and the reason the chevron looked flush is that `.task-icon` carries `margin-left: calc(-1 * var(--ks-spacing-2))`, which cancels that gap. Adding `me-1` on top means the icon column now shifts 4px right on `md+`, and the margin also applies on rows where the chevron is absent, since the wrapper div renders unconditionally. It works, but reducing the negative margin would be the more direct fix.

**4. `--ks-editor-single-line-border-color` is new public design-system API with no docs or story.** `--ks-*` is otherwise the design-token namespace, so a component-level override hook sitting in it is a little muddy. At minimum worth a line in the `KsEditor` story so the next person knows it exists.

**5. Removing the wrapper border shrinks the toggle-mode editor by 2px** in a layout where it is a `flex: 1` sibling of `KsDatePicker` / `TaskDuration`. Probably an improvement, since they should now match at `min-height: var(--kel-component-size)`, but the `date-time` toggle case is worth checking too, not just `duration` / `lateMaximumDelay`.

**6. Verification gap.** 7 visual fixes, no automated coverage, and the test plan says browser verification was not performed. The `KsEditor` change touches shared design-system CSS used at every inline editor site, so that is the one I would actually run before merging.

loicmathieu:
@MilosPaunovic I updated the PR title and the restarted/replayed label code.

MilosPaunovic (approved):

PR #18232, opened 2026-08-18
fix(executions): put the bulk actions overflow menu last in the action bar

RemiBarthe:
closes https://github.com/kestra-io/kestra-ee/issues/9975
Related to https://github.com/kestra-io/kestra-ee/pull/10039.

Capture d’écran 2026-08-18 à 13 07 26

The bulk-actions bar rendered the "more actions" dropdown before the slot that extensions fill, so the EE Cases actions landed after it. An overflow menu reads as the terminator of an action row, and anything to its right reads as forgotten. While "Create case from selection" is a first-class action.

Render the extension components before the dropdown so it closes the bar.


MilosPaunovic (commented):
Verified the diff is a pure block move: 11 added / 11 removed, byte-identical multiset. The fix is correct, with a few notes below.

### Root cause confirmed

The reason this works isn't just aesthetic ordering - element-plus hard-codes the assumption that a dropdown inside a button group is its **last** member. From `element-plus/theme-chalk/src/button-group.scss`:

```scss
& > .#{$namespace}-dropdown {
  & > .#{$namespace}-button {
    border-top-left-radius: 0;
    border-bottom-left-radius: 0;
    border-left-color: getCssVar('button', 'divide-border-color');
  }
}
```

Only the *left* radii get zeroed - the right stays fully rounded. So with EE's `bulkActionComponents` rendered after `KsDropdown`, the dots button sat mid-group with square-left / rounded-right, which is exactly the mismatch reported in the issue. Moving it last makes the CSS's assumption true.

Worth noting for future changes in that slot: the `#bulk-actions` content lands inside `KsBulkSelect`'s `KsButtonGroup` (`ui/packages/design-system/src/components/Data/KsDataTable/KsBulkSelect.vue`), so all of element-plus's `:first-child` / `:last-child` / `:not(:last-child) { margin-right: -1px }` rules apply to whatever a page puts there.

### Scope checks

- **OSS-only PR for an EE issue is right.** `bulkActionComponents` is `[]` in `ui/src/override/components/executions/executionsExtensions.ts`, so OSS already had the dropdown last - zero visual change there. EE has no `Executions.vue` override (only its own `executionsExtensions.ts`), so this is the complete fix and no EE companion PR is needed.
- **No backport.** `releases/v1.3.x` `Executions.vue` has no `bulkActionComponents` at all, so the bug can't occur on 1.3.
- **E2E unaffected** - `ui/tests/e2e/pages/executions.page.ts` locates the overflow menu by `.kel-dropdown`, not by position.
- `Flows.vue` also injects a `component v-for` in its `#bulk-actions` slot but has no `KsDropdown`, so it doesn't need the same treatment.

### Latent fragility on the EE side (follow-up, not blocking)

`ui-ee/src/components/cases/CreateCaseModal.vue` is a `KsDialog` **without** `appendToBody`, unlike its siblings (`AddToExistingCaseButton` and the labels dialog both pass `:appendToBody="true"`). Element Plus's dialog only teleports when that prop is set, and the overlay is `v-show`-ed rather than `v-if`-ed, so a hidden `.kel-overlay` div is rendered as a direct child of `.kel-button-group`, between the two Cases buttons. It is also unconditional, while the button above it is `v-if="canCreate"`, so it renders even for users who cannot create cases.

Harmless today (`display: none`, and it isn't last-child), but it *is* a structural sibling that `:first-child`, `:last-child` and `:not(:last-child) { margin-right: -1px }` all count. The next reorder in that slot will silently steal the group's edge rounding and reintroduce this exact class of bug. Adding `:appendToBody="true"` there would keep the group's children to buttons plus the dropdown only.

### Housekeeping

1. **The PR title is truncated:** `fix(executions): put the bulk actions overflow menu last in the actio…`. The commit subject is complete (`...in the actions bar`), only the PR title got cut - worth restoring.
2. **The description is inaccurate:** it says the components moved "from inside a dialog template", but they were never inside the dialog, they were siblings after the labels `KsDialog`. The `diffhunk://` links don't resolve either, and "likely to improve user experience or fix a rendering issue" hedges on the actual reason. One sentence about the element-plus dropdown-last assumption would serve the next reader much better. The closing keyword at the top is correct.
3. **CI:** `trigger-ee` is red for infra reasons, not code - `Illegal state: Cannot execute a disabled flow`, then `EE compile check inconclusive for ref ... HTTP 409`. The EE CI webhook flow is disabled on the CI instance. Everything else is green (Frontend - Tests, E2E check, Plerion, build artifacts), so it just needs a re-run once that flow is re-enabled.

### Adjacent pre-existing issue (optional)

All five `KsDropdownItem`s in that overflow menu are `v-if="canUpdate"`. A user with read + delete but not update sees a dots button that opens an empty menu. Not introduced here, but now that the button is the visual terminus of the group it is more conspicuous - gating the whole `KsDropdown` on `canUpdate` would be a one-liner if you want to fold it in.

MilosPaunovic (approved):
Approving the code: re-verified after the develop merge that the net diff vs the merge base is still only the 11-line reorder in `Executions.vue`, and the element-plus `button-group` dropdown-last rule makes it the right fix.

Non-blocking leftovers from my earlier comment, none of which need another review round:
- The truncated PR title (`...in the actio…`) and the inaccurate description are worth fixing before merge.
- `trigger-ee` is red for infra reasons on both SHAs (`Illegal state: Cannot execute a disabled flow` on `run_backend_ee_ci_webhook`, HTTP 409), so it needs that flow re-enabled rather than a re-run.
- The `CreateCaseModal` `appendToBody` cleanup is an EE follow-up.

MilosPaunovic (approved):
Re-approved after the branch updates - the net diff vs the merge base is still only the 11-line block move in `Executions.vue`, which is the right fix: element-plus's `button-group.scss` zeroes only the *left* radii of `> .el-dropdown > .el-button` (it assumes a dropdown in a button group is the last member), so mid-group the dots button rendered square-left / rounded-right - the exact border mismatch in the issue.

**Needs an EE follow-up** to keep this fixed:

`ui-ee/src/components/cases/CreateCaseModal.vue` is a `KsDialog` without `appendToBody`, unlike its siblings (`AddToExistingCaseButton` and the labels dialog both pass `:appendToBody="true"`). Element Plus only teleports the dialog when that prop is set, and the overlay is `v-show`-ed rather than `v-if`-ed, so a hidden `.kel-overlay` div is rendered as a direct child of the `.kel-button-group`, between the two Cases buttons. It also renders unconditionally, while the button above it is `v-if="canCreate"`.

Harmless today (`display: none`, and it isn't last-child), but it is a structural sibling that `:first-child`, `:last-child` and `:not(:last-child) { margin-right: -1px }` all count - the next reorder in this slot will silently steal the group's edge rounding and reintroduce this exact class of bug. Adding `:appendToBody="true"` there keeps the group's children to buttons plus the dropdown only.

MilosPaunovic:
The EE follow-up is up: https://github.com/kestra-io/kestra-ee/pull/10039 - it adds `:appendToBody="true"` to `CreateCaseModal.vue` so the non-teleported overlay stops sitting inside the bulk actions button group, which keeps this fix safe against future reorders of the slot content.

MilosPaunovic:
@RemiBarthe Feel free to merge the EE pull request as well, once you merge this one.

PR #18270, opened 2026-08-19
fix(logs): stop a stale search from replacing the logs list

AcevedoR:
Related to kestra-io/kestra-ee#8205.

Entering `/logs` fired **three** page-sized searches: `KsDataTable` loads on mount, so the first ran against the bare query (no level, no time window — a scan of the whole log table), then the default time range landed, then the default level. Each round also fired 5 level-count probes.

`findLogs` published whatever came back, without checking the response still belonged to the newest search — so the last one to land won. When the broad search is the slow one, its late answer overwrites the correct result: the list shows rows the filters exclude, or nothing when the stale search matched nothing, while every request looks fine in the network tab.

Fix:
- `useRouteFilterPolicy` exposes `isRouteSettled`; `LogsWrapper` gates its first load on it → one search per page entry instead of three.
- The logs store drops superseded responses → covers the races the gate doesn't (filter changes, pagination, refresh).

Two unit specs, both failing on `develop` and passing here. Reproduced in a browser by delaying only the discarded searches: URL says *at or above INFO* (6 lines), list rendered all 9 including DEBUG; now stays at 6.

### DIFF
--- a/ui/AGENTS.md
+++ b/ui/AGENTS.md
@@ -148,6 +148,10 @@ const pageSize = ref(25)
 
 **Never** maintain a separate `internalPage` / `pageNumber` ref *and* bind the prop to a different value — that re-introduces the drift bug (URL says page 2, UI shows page 1) that this contract exists to prevent.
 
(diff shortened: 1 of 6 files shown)


RemiBarthe (approved):

AcevedoR:
I added a commit to handle more cases, requesting re review

RemiBarthe (approved):
LGTM ! 🚀

PR #18274, opened 2026-08-19
fix(execution): Kill switched CANCEL or KIL execution should unlock t…

loicmathieu:
…rigger

In both cases, the execution should at least fly to the DefaultExecutor.toExecution() methods so terminal operations are still done.

Fixes https://github.com/kestra-io/kestra-ee/issues/10023


AcevedoR (approved):
LGTM but still need a tweak for the tests to pass

Issue #18231, opened 2026-08-18
Internal server 500 error when running flow


RWaltersMA:
### Describe the issue

The following script will be sucessfully run, however the UI will show an internal server error 500

```
id: do-flow
namespace: test

tasks:
  - id: create_droplet
    type: io.kestra.plugin.digitalocean.droplet.Create
    name: "runner-{{ execution.id | lower }}"
    region: nyc3
    size: s-2vcpu-4gb
    image: ubuntu-24-04-x64
    apiToken: "xxxxxxx"

  - id: wait_until_active
    type: io.kestra.plugin.core.flow.LoopUntil
    condition: "{{ outputs.poll.status == 'active' }}"
    checkFrequency:
      interval: PT10S
      maxDuration: PT10M
    failOnMaxReached: true
    tasks:
      - id: poll
        type: io.kestra.plugin.digitalocean.droplet.Get
        dropletId: "{{ outputs.create_droplet.id }}"
        apiToken: "xxxxxxx"

finally:
  - id: delete_droplet
    type: io.kestra.plugin.digitalocean.droplet.Delete
    apiToken: "xxxxxxx"
    dropletId: "{{ outputs.create_droplet.id }}"
```

Results:

<img width="1075" height="610" alt="Image" src="https://github.com/user-attachments/assets/22d925ac-243a-41b0-91f3-cb2ccec6d39f" />

### Environment

- Kestra Version: 2.0 RC8

Pradumnasaraf:
Working. QA done ✅

```yaml
id: poll_until_iteration_target
namespace: qa.loopuntil

tasks:
  - id: loop
    type: io.kestra.plugin.core.flow.LoopUntil
    condition: "{{ outputs.return_iteration.value == '4' }}"
    checkFrequency:
      interval: PT1S
    tasks:
      - id: return_iteration
        type: io.kestra.plugin.core.debug.Return
        format: "{{ outputs.loop.iterationCount }}"

```

<img width="1360" height="1080" alt="Image" src="https://github.com/user-attachments/assets/38aa9577-a614-42e0-a646-1ba47728715b" />

PR #18280, opened 2026-08-19
fix(core): clean up orphaned TaskOutput records when pruning TaskRuns

### LINKED ISSUE #18238: Investigate deleting TaskOutput rows when LoopUntil prunes a task run
### Context

Follow-up from #18231 / #18237.

`ExecutionService.retryWaitFor()` (`core/src/main/java/io/kestra/core/services/ExecutionService.java`) prunes every descendant task run of an iterating `LoopUntil` between iterations so the next iteration starts with a clean state:

```java
// Remove all descendants (not just direct children) of the iterating LoopUntil so that nested
// LoopUntil tasks start the next iteration with a clean state and don't inherit stale outputs.
```

It does not delete the corresponding `TaskOutput` rows for those pruned task runs. #18237 fixed the resulting symptom (`GET /outputs/tasks/{executionId}` throwing a 500 when it hit one of these now-orphaned rows) by having `OutputController` skip rows whose task run no longer exists, but the orphaned rows themselves are never cleaned up.

### Problem

A long-running `LoopUntil` (e.g. `checkFrequency: {interval: PT10S, maxDuration: PT10M}` ≈ 60 iterations) leaves one orphan `TaskOutput` row per iteration per child task that produces outputs. These rows are never referenced again and are only ever cleaned up when the whole execution is purged — an unbounded, execution-lifetime storage leak for any flow that loops a lot.

### To investigate

- Whether `retryWaitFor()` should delete the outputs of the task runs it prunes, in the same transaction, via a new `TaskOutputRepositoryInterface#deleteByTaskRunIds(...)` (today the interface only exposes `purgeByExecutionIds`, see `core/src/main/java/io/kestra/core/repositories/TaskOutputRepositoryInterface.java`).
- JDBC implementation in `jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcTaskOutputRepository.java`, plus the equivalent EE (Elasticsearch) repository implementation — both need to stay in sync.
- Whether deleting eagerly is safe: outputs referenced via `outputs.loop.*` / nested `LoopUntil` iteration counts should not be affected, and the internal-storage-backed outputs (`shouldStoreInInternalStorage`, see `core/src/main/java/io/kestra/core/services/TaskOutputService.java`) also need their storage objects removed, not just the DB row.
- Whether this should be scoped only to `LoopUntil`/`retryWaitFor()`, or generalized to any place that currently prunes task runs without a matching output cleanup (e.g. `Loop`).

### Out of scope here

This is a storage-growth/cleanup concern, not a correctness bug — the read-side 500 is already fixed in #18237.

hemantmm:
closes: #18238

---

### ✨ Description

What does this PR change?  

This PR fixes a storage leak caused by in-place pruning of task runs during execution state changes (such as `LoopUntil` retries via `ExecutionService.retryWaitFor()`). 

Previously, when descendant task runs were removed from the in-memory `Execution` object to start a new iteration with a clean state, their corresponding `TaskOutput` database records were left behind, causing an unbounded storage leak for long-running loops.

- Added `deleteByTaskRunIds` to `TaskOutputRepositoryInterface` and its JDBC implementation.
- Added orchestration in `TaskOutputService` to also delete associated internal storage files before dropping the DB records.
- Rather than changing the `ExecutionService` APIs to return pruned task run IDs, the executor (`DefaultExecutor` and `ExecutionCommandMessageHandler`) compares the `TaskRunList` before and after execution mutations and derives the removed task run IDs.
- The cleanup is performed only when the execution ID remains the same, avoiding treating newly created executions (e.g. from `replay()`) as pruned task runs.

### Notes:
- **EE Counterpart Required:** I don't have access to the EE repository, so the corresponding Elasticsearch repository implementation for `TaskOutputRepositoryInterface#deleteByTaskRunIds` will need to be added by the Kestra team before this can be merged.

Related to: #18231, #18237


hemantmm:

loicmathieu:
@hemantmm this PR have a lot of unrelated changes, I suspect you have some unrelated merge commits, can you rebase on develop?

hemantmm:
@loicmathieu, I have rebased the branch.

loicmathieu:
@hemantmm there are still a lot of unrelated changes, maybe you can close this one and re-create the PR from develop.

hemantmm:
closing in favor of https://github.com/kestra-io/kestra/pull/19394

Issue #9755, opened 2025-06-27
UX: Everytime a fetch request returns 404 we redirect the whole page.


elevatebart:
### Describe the issue

When a fetch request fails because of a 404, the system redirects users to a "page not found" page. 
Unfortunately, this hides the core issue that the fetch failed or 404ed. 
It gets super hard to know what is wrong for the final user ... and obviously for developers.

### Environment

- Kestra Version: develop

starlightknown:
<img width="1232" alt="Image" src="https://github.com/user-attachments/assets/4417cac7-3ddf-467d-b098-b2a2fc3e6c34" />

something like this? it used to go to 404 but now I added a check to show message

PR #18282, opened 2026-08-19
fix(ui): name the request that 404ed on the not-found page

AcevedoR:
Any fetch returning 404 swaps the whole page for a generic "Page not found", hiding which call actually failed (#9755).

The failed request now travels through the core store and is rendered under the message as `GET /api/v1/main/flows/ns/missing` plus the server message — no behaviour change to *when* the page takes over, and no new translation keys.

Unit tests cover both the recorded 404 and the `ignoreNotFound` opt-out.

### DIFF
--- a/ui/src/components/errors/Errors.vue
+++ b/ui/src/components/errors/Errors.vue
@@ -8,6 +8,11 @@
             <span v-html="$t('errors.' + code + '.content')" />
         </p>
 
+        <p v-if="failedRequest" class="failed-request">
+            <code>{{ failedRequest.method }} {{ failedRequest.url }}</code>
+            <span v-if="failedRequest.message">{{ failedRequest.message }}</span>
+        </p>
+
         <KsButton v-if="!isFullScreen()" tag="router-link" :to="{name: 'home'}" type="primary" size="large">
             {{ $t("back_to_dashboard") }}
         </KsButton>
@@ -33,6 +38,8 @@
     const coreStore = useCoreStore()
     const route = useRoute()
 
+    const failedRequest = computed(() => coreStore.failedRequest)
+
     const routeInfo = computed(() => ({title: t("errors." + props.code + ".title")}))
 
     useRouteContext(routeInfo)
@@ -45,6 +52,7 @@
         () => route.fullPath,
         () => {
             coreStore.error = undefined
+            coreStore.failedRequest = undefined
         },
     )
 </script>
@@ -64,5 +72,17 @@
     p {
         line-height: 22px;
         font-size: var(--ks-font-size-sm);
+    }
+
+    .failed-request {
+        display: flex;
+        flex-direction: column;
+        gap: 0.25rem;
+        color: var(--ks-content-secondary);
+
+        code {
+            font-family: var(--bs-font-monospace);
+            word-break: break-all;
+        }
     }
 </style>
--- a/ui/src/components/flows/blueprints/BlueprintsBrowser.vue
+++ b/ui/src/components/flows/blueprints/BlueprintsBrowser.vue
@@ -306,7 +306,10 @@
             emit("loaded")
         } catch {
             if (props.embed) error.value = true
-            else coreStore.error = 404
+            else {
+                coreStore.error = 404
+                coreStore.failedRequest = undefined
+            }
         }
     }
 
--- a/ui/src/stores/core.ts
+++ b/ui/src/stores/core.ts
@@ -5,9 +5,17 @@ import {Message} from "../components/ErrorToast.vue"
 import {TUTORIAL_NAMESPACE} from "../utils/constants"
 import {Flow} from "./flow"
 
+export interface FailedRequest {
+    status: number
+    method: string
+    url: string
+    message?: string
+}
+
 export const useCoreStore = defineStore("core", () => {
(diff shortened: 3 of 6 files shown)


AcevedoR:
because Bart is OOO, maybe @Piyush-r-bhaskar you could give me your though on this PR so I know if I'm going in the right direction or not :)

ui/src/components/errors/Errors.vue
```diff
@@ -8,6 +8,11 @@
             <span v-html="$t('errors.' + code + '.content')" />
         </p>
 
+        <p v-if="failedRequest" class="failed-request">
+            <code>{{ failedRequest.method }} {{ failedRequest.url }}</code>
+            <span v-if="failedRequest.message">{{ failedRequest.message }}</span>
```
Piyush-r-bhaskar, line 13: Nothing tells the user what this URL is, we just print `GET /api/v1/main/flows/company/hello` under the picture. Could we add a short label, with a translation key?

Also `FailedRequest` carries `status`, but we only print `method`, `url` and `message`, so that field never reaches the screen. Show it or drop it.

ui/src/components/errors/Errors.vue
```diff
@@ -65,4 +73,16 @@
...
+        display: flex;
+        flex-direction: column;
+        gap: 0.25rem;
+        color: var(--ks-content-secondary);
+
+        code {
+            font-family: var(--bs-font-monospace);
+            word-break: break-all;
+        }
+    }
```
Piyush-r-bhaskar, line 87: `--ks-content-secondary` and `--bs-font-monospace` don't exist, so neither the colour nor the monospace font applies here. The bootstrap vars were renamed to `--kbs-*`. The gap can use the spacing scale too.

```scss
.failed-request {
    display: flex;
    flex-direction: column;
    gap: var(--ks-spacing-1);
    color: var(--ks-text-secondary);

    code {
        font-family: var(--ks-font-family-mono);
        word-break: break-all;
    }
}
```

Piyush-r-bhaskar (commented):
Printing the failed URL helps, but this only covers half of #9755.

The issue is two things: the 404 page doesn't say what failed, and one background call 404ing kills the whole page. This PR fixes the first. The second is still there, we just print the URL on top of it. Can we do both here?

Easy repro on develop. Save this in `company` and run it once:

```yaml
id: hello
namespace: company

tasks:
  - id: log
    message: hello
```

Now delete the flow and open that execution from the list. The execution is fine, logs and outputs and all, but you get "Page not found". One background call went looking for the flow we just deleted.

The cause is that in `kestraHttp` we treat "a request came back 404" and "this URL isn't a page in Kestra" as the same thing. I'd flip it: a 404 shows a toast, using the payload you already collect here, and only the page's own main request gets the full error screen. So many call sites already opt out with `ignoreNotFound`, which says the default is wrong.

Left a few smaller things on the diff.

ui/src/stores/core.ts
```diff
@@ -5,9 +5,17 @@ import {Message} from "../components/ErrorToast.vue"
...
+    status: number
+    method: string
+    url: string
+    message?: string
+}
+
 export const useCoreStore = defineStore("core", () => {
     const message = ref<Message>()
     const error = ref<any>()
+    const failedRequest = ref<FailedRequest>()
```
Piyush-r-bhaskar, line 18: Can `error` and `failedRequest` be one value here? They always have to be set and cleared together and nothing enforces it. Miss the clearing once and the page shows an old URL from a failure that has nothing to do with what just broke. Either `error = {code, request}` or a small `setError()` / `clearError()` pair.

ui/src/components/flows/blueprints/BlueprintsBrowser.vue
```diff
@@ -302,7 +302,10 @@
             emit("loaded")
         } catch {
             if (props.embed) error.value = true
-            else coreStore.error = 404
+            else {
+                coreStore.error = 404
+                coreStore.failedRequest = undefined
```
Piyush-r-bhaskar, line 311: This extra line is what I mean over in `core.ts`. Every place that sets `coreStore.error` now has to remember to clear `failedRequest` as well, and at some point one of them won't.

ui/tests/unit/utils/kestraHttp.spec.ts
```diff
@@ -65,3 +65,53 @@ describe("setupKestraHttp router NProgress hooks", () => {
...
+    const notFoundRequest = {
+        method: "get",
+        url: missingFlowUrl,
+    }
+
+    it("records the request that 404ed so the error page can name it", () => {
+        const coreStore: Record<string, unknown> = {}
+        const onError = captureErrorInterceptor(coreStore)
+
+        onError(Object.assign(new Error("Not Found"), {status: 404}), notFoundResponse, notFoundRequest)
```
Piyush-r-bhaskar, line 97: This builds the error by hand, so it skips the step where the SDK puts the status in front of the message. In the real app `message` is `404 Page Not Found`, not `Not Found`, so we end up printing the status right under a heading that already says the same thing.

Piyush-r-bhaskar (commented):
.

AcevedoR:
fixed by Bart in another pr :)

PR #18283, opened 2026-08-19
fix(core): mask a logged exception's message and stacktrace, not just the log statement

### LINKED ISSUE #14637: Secrets with value {# -  Unclosed comment
### Describe the issue

If you create a secret:
```bash
echo -n 'test_secret_with_pebble_comment_{#asd' | base64
dGVzdF9zZWNyZXRfd2l0aF9wZWJibGVfY29tbWVudF97I2FzZA==
```

and use this secret:
```yml
id: pebble_comment_secret_test
namespace: company

tasks:
  - id: secret_test
    type: "io.kestra.plugin.fs.sftp.List"
    from: "~/test_folder/"
    host: localhost
    username: test_user
    password: " {{ secret('PEBBLE_COMMENT') }} "
    regExp: "^\\S+json$"
```

following error occures:
```
2026-02-16 12:05:08.468io.pebbletemplates.pebble.error.ParserException: Unclosed comment. ( ****** :1)
```

leaked secret in trace:
```
TRACE 2026-02-16T11:05:08.468309Z io.kestra.core.exceptions.IllegalVariableEvaluationException: io.pebbletemplates.pebble.error.ParserException: Unclosed comment. ( test_secret_with_pebble_comment_{#asd :1)
	at io.kestra.core.runners.VariableRenderer.properPebbleException(VariableRenderer.java:52)
	at io.kestra.core.runners.VariableRenderer.renderOnce(VariableRenderer.java:108)
	at io.kestra.core.runners.VariableRenderer.render(VariableRenderer.java:79)
	at io.kestra.core.runners.VariableRenderer.render(VariableRenderer.java:64)
	at io.kestra.core.runners.VariableRenderer.render(VariableRenderer.java:56)
	at io.kestra.core.runners.DefaultRunContext.render(DefaultRunContext.java:238)
	at io.kestra.plugin.fs.vfs.VfsService.uri(VfsService.java:50)
	at io.kestra.plugin.fs.vfs.AbstractVfsTask.uri(AbstractVfsTask.java:57)
	at io.kestra.plugin.fs.vfs.List.run(List.java:51)
	at io.kestra.plugin.fs.vfs.List.run(List.java:13)
	at io.kestra.worker.WorkerTaskCallable.doCall(WorkerTaskCallable.java:81)
	at io.kestra.worker.AbstractWorkerCallable.call(AbstractWorkerCallable.java:64)
	at io.kestra.worker.WorkerSecurityService.callInSecurityContext(WorkerSecurityService.java:10)
	at io.kestra.worker.DefaultWorker.lambda$callJob$19(DefaultWorker.java:980)
	at io.kestra.core.trace.NoopTracer.inCurrentContext(NoopTracer.java:15)
	at io.kestra.worker.DefaultWorker.callJob(DefaultWorker.java:976)
	at io.kestra.worker.DefaultWorker.runAttempt(DefaultWorker.java:934)
	at io.kestra.worker.DefaultWorker.run(DefaultWorker.java:728)
	at io.kestra.worker.DefaultWorker.handleTask(DefaultWorker.java:379)
	at io.kestra.worker.DefaultWorker.lambda$run$8(DefaultWorker.java:294)
	at io.micrometer.core.instrument.internal.TimedRunnable.run(TimedRunnable.java:49)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: io.pebbletemplates.pebble.error.ParserException: Unclosed comment. ( test_secret_with_pebble_comment_{#asd :1)
	at io.pebbletemplates.pebble.lexer.LexerImpl.tokenizeComment(LexerImpl.java:386)
	at io.pebbletemplates.pebble.lexer.LexerImpl.tokenize(LexerImpl.java:195)
	at io.pebbletemplates.pebble.PebbleEngine.getPebbleTemplate(PebbleEngine.java:164)
	at io.pebbletemplates.pebble.PebbleEngine.lambda$getTemplate$0(PebbleEngine.java:152)
	at io.kestra.core.runners.pebble.PebbleLruCache.computeIfAbsent(PebbleLruCache.java:29)
	at io.kestra.core.runners.pebble.PebbleLruCache.computeIfAbsent(PebbleLruCache.java:12)
	at io.pebbletemplates.pebble.PebbleEngine.getTemplate(PebbleEngine.java:152)
	at io.pebbletemplates.pebble.PebbleEngine.getLiteralTemplate(PebbleEngine.java:134)
	at io.kestra.core.runners.VariableRenderer.renderOnce(VariableRenderer.java:99)
	... 22 more

```

### Environment

- Kestra Version: 1.1.15

rsd-darshan:
### ✨ Description

Narrowed after review (see discussion below): this PR now only fixes the log-masking gap.

`RunContextLogger`'s `ContextAppender` was masking a log statement's own message/args, but passing an attached exception's message and stack trace through unmasked — so if a secret ever ends up embedded in an exception's text (e.g. via a render failure that quotes the offending value), it leaked into the execution logs unredacted. `ContextAppender.append()` now applies the same secret-masking pass to those entries too.

Dropped the earlier approach of wrapping `secret()`'s return value in raw-tag markers to prevent the parser crash itself: per @loicmathieu, that only special-cases one function, while the same class of issue can occur with any function whose output gets re-parsed under recursive rendering (e.g. `kv()`), and hardening against that generally isn't wanted.

### 🔗 Related Issue

Closes #14637 (this addresses the secret-leak half of that report; the parse-error/crash itself is not fixed here per maintainer guidance above)

### 🛠️ Backend Checklist

### 📝 Additional Notes

- `RunContextLoggerTest.secrets()` had a stale, non-asserting check (`assertThat(...)` never terminated with `.isTrue()`/etc.) and a comment documenting the leak as accepted behavior — replaced with a real assertion that the exception's stack-trace entry is now masked.


core/src/main/java/io/kestra/core/runners/RunContextLogger.java
```diff
@@ -374,7 +374,7 @@ protected BaseAppender(RunContextLogger runContextLogger, Logger logger) {
             this.logger = logger;
         }
 
-        private String replaceSecret(String data) {
+        protected String replaceSecret(String data) {
```
loicmathieu, line 377: Why is it protected?

loicmathieu (changes requested):
I don't understand why you have to wrap a secret inside a raw tag?

core/src/main/java/io/kestra/core/runners/RunContextLogger.java
rsd-darshan: Needed it so `ContextAppender.append()` (a sibling nested class) can reuse this exact masking logic when masking an attached exception's message/stacktrace, instead of duplicating the loop there. Happy to make it package-private instead if you'd prefer a narrower scope.

rsd-darshan:
@loicmathieu on the raw-tag wrapping: `VariableRenderer` can re-parse its own already-rendered output as Pebble source (recursive rendering, or a value being rendered again elsewhere). If a secret's literal value contains `{#`, `{{` or `{%`, that later parse blows up with an error that quotes the offending text — i.e. the secret itself — into the logs. Wrapping the returned value in Kestra's own `{% raw %}...{% endraw %}` markers makes any later pass treat it as opaque literal text, reusing the exact same mechanism a user could already use themselves for raw blocks, instead of adding new special-casing to the renderer for secrets specifically.

The failing test was a real gap in that: a typed render (e.g. `{{ [secret('X')] }}` for a MULTISELECT input) can return the wrapped value nested inside a List/Map rather than as the top-level String, so the existing marker-stripping (which only handled a top-level String) missed it. Pushed a fix that strips markers recursively through Map/List/Set results, plus a regression test for that exact case.

loicmathieu:
@rsd-darshan 

> VariableRenderer can re-parse its own already-rendered output as Pebble source (recursive rendering, or a value being rendered again elsewhere).

This seems unrelated to the secret leakage. Does this PR tries to fix two unrelated things?
We strongly discourage using recursive rendering and the same class of issues can occurs on other functions as well (as `kv()` for ex) so I don't think we would want to fix it.

People using secrets with reserved characters and recursive rendering should make sure their secrets are escaped.

rsd-darshan:
@loicmathieu fair point, and it applies generally rather than just to `secret()` — dropped the raw-tag wrapping in `SecretFunction`/`VariableRenderer` and the tests that went with it. This PR now only fixes the masking gap: `ContextAppender` was masking a log statement's own message/args, but passing an attached exception's message/stacktrace through unmasked, so a secret embedded in one still leaked. That part's unrelated to recursive rendering and stands on its own.

Updated the title/description accordingly.

loicmathieu (approved):
LGTM, thanks

PR #18288, opened 2026-08-19
fix(worker): exclude the dialect JDBC repositories from a worker context

fhussonnois:
### ✨ Description

A worker deployed as a dedicated service owns no repository, and its application context already drops every datasource. Domain repositories carry `@RepositoryBean`, which keeps them out of a worker, but the three `@EachBean(JdbcTableConfig)` dialect repositories — `PostgresRepository`, `H2Repository`, `MysqlRepository` — are gated by their own `@Requires(condition = …)` and stayed registered. Anything resolving one on a worker reached `JooqDSLContextWrapper` and died on the missing `DataSource`.

They cannot carry `@RepositoryBean` itself: it bundles `@Singleton`, which conflicts with `@EachBean`, so they get the server-type condition directly.

### 🔗 Related Issue

Related to https://github.com/kestra-io/kestra-ee/issues/10067 — closed by the companion EE PR, see the note below.

### 🛠️ Backend Checklist

- [x] All unit and integration tests pass — `WorkerServerContextTest`, `ServerCommandValidatorTest`

### 📝 Additional Notes

- **Companion PR:** kestra-io/kestra-ee, branch `fix/worker-repository-beans` (same name).
- `WorkerServerContextTest.shouldNotRegisterARepositoryOnWorker` was sanity-checked by removing the condition again: it fails on the `H2Repository` assertion, so it does guard the change.
- **Scope:** on its own this is hygiene rather than the fix for the linked issue. With `kestra.secret.type=jdbc` it turns the reported `DependencyInjectionException` into a missing-bean condition; what makes that actionable, and what makes the configuration work, lives in the EE PR. Its value here is that a worker context is now free of repository beans whatever the backend, so `containsBean(...)` guards behave uniformly.

### DIFF
--- a/cli/src/test/java/io/kestra/cli/commands/servers/WorkerServerContextTest.java
+++ b/cli/src/test/java/io/kestra/cli/commands/servers/WorkerServerContextTest.java
@@ -10,6 +10,8 @@ import org.junit.jupiter.api.io.TempDir;
 
 import io.kestra.cli.Kestra;
 import io.kestra.core.migration.MigrationStartupRunner;
+import io.kestra.core.repositories.FlowRepositoryInterface;
+import io.kestra.repository.h2.H2Repository;
 
 import io.micronaut.context.ApplicationContext;
 import io.micronaut.context.env.Environment;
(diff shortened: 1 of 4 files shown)


Skraye (approved):
LGTM

PR #18298, opened 2026-08-20
fix(ui): add bottom margin to new flow landing page (#18285)

### LINKED ISSUE #18285: [UI] Lack of bottom margin/padding in the new flow landing page
### Describe the issue

Image

As you can see in the screenshot above, the last card action at the bottom is touching the bottom of the screen and has no bottom margin/padding to give it some space. This is on a 15.6" screen (laptop). App zoom is set at 100%.

### Environment

- Kestra Version: v2.0 RC8

SANTHOSHKUMAR-303:
✨ Description
Fixed an issue on the new flow landing page (NewFlowLanding.vue) and YAML import page (ImportYaml.vue) where the bottom card action touched the bottom edge of the screen without bottom margin/padding when scrolling on smaller screens (e.g. 15.6" laptop screens at 100% zoom).

Added margin-bottom: var(--ks-spacing-8) to .landing-body and adjusted .new-flow-landing container padding in NewFlowLanding.vue.
Added margin-bottom: var(--ks-spacing-6) to .import-actions and adjusted .import-yaml container padding in ImportYaml.vue.
This ensures that block-level margin is preserved during flexbox overflow scroll in WebKit/Blink browsers, guaranteeing proper bottom spacing below the last card element.
🔗 Related Issue
Closes https://github.com/kestra-io/kestra/issues/18285

🎨 Frontend Checklist
 Code builds without errors (npm run build)
 All existing E2E tests pass (npm run test:e2e)
 Screenshots or video recordings attached showing the UI changes
📝 Additional Notes

- **Technical Context**: Modern WebKit/Blink browsers collapse flex container `padding-bottom` when the parent container (`main`) scrolls overflow content. Shifting the bottom spacing to a `margin-bottom` on the child block elements (`.landing-body` and `.import-actions`) ensures the bottom spacing is explicitly included in the scrollHeight calculation across all browsers.
- **Layout Compatibility**: On larger viewports where content fits without scrolling, `min-height: 100%` on `.new-flow-landing` continues to stretch smoothly without double padding.

🤖 AI Authors
Why was the cat sitting on the computer keyboard?
Because it wanted to keep an eye on the mouse! 🐱

### DIFF
--- a/ui/src/components/flows/create/ImportYaml.vue
+++ b/ui/src/components/flows/create/ImportYaml.vue
@@ -134,7 +134,7 @@
         display: flex;
         flex-direction: column;
         gap: var(--ks-spacing-4);
-        padding: var(--ks-spacing-6) var(--ks-spacing-4);
+        padding: var(--ks-spacing-6) var(--ks-spacing-4) 0;
         width: 100%;
         max-width: 48rem;
         margin: 0 auto;
@@ -210,5 +210,6 @@
     .import-actions {
         display: flex;
         justify-content: flex-end;
+        margin-bottom: var(--ks-spacing-6);
     }
 </style>
--- a/ui/src/components/flows/create/NewFlowLanding.vue
+++ b/ui/src/components/flows/create/NewFlowLanding.vue
@@ -197,7 +197,7 @@
         display: flex;
         flex-direction: column;
         align-items: center;
-        padding: var(--ks-spacing-8) var(--ks-spacing-4);
+        padding: var(--ks-spacing-8) var(--ks-spacing-4) 0;
         min-height: 100%;
     }
 
@@ -223,6 +223,7 @@
         gap: var(--ks-spacing-4);
         width: 100%;
         max-width: 36rem;
+        margin-bottom: var(--ks-spacing-8);
     }
 
     .primary-card-body {


MilosPaunovic (changes requested):
Thanks for digging into this one! I tested the change locally, and unfortunately it does not fix the reported issue - the root cause is elsewhere.

**Why the margin does not help on the landing page:** `.new-flow-landing` has `min-height: 100%` and gets `flex: 1` from `main > section.full-container > *`. The section's height is pinned to the visible height of the `main` scroller, and `min-height: 100%` replaces the default `min-height: auto`, so the flex item cannot grow beyond the section. When the content is taller than the viewport it overflows the element's box, and neither the old `padding-bottom` nor the new `margin-bottom` contributes to the scrollable overflow that propagates to `main` (only border boxes propagate through `overflow: visible` ancestors; margins do not). Measured before and after applying this PR: `main.scrollHeight` is identical and the last card still touches the viewport bottom when fully scrolled.

**The actual fix:** remove `min-height: 100%` from `.new-flow-landing` and keep the original padding. With the default `min-height: auto` the flex item grows to its content, so its `padding-bottom` sits below the last card and counts toward the scroll height. The non-scrolling case is unaffected: `flex: 1` from `section.full-container > *` already stretches the element to fill the page, so `min-height: 100%` is redundant. Verified locally: 48px gap at full scroll, and the element still fills the section on tall viewports.

**ImportYaml.vue:** this page never had the bug - `.import-yaml` has no `min-height: 100%`, so its box already grows with the content and the existing `padding-bottom` is honored when scrolling. Please revert the changes to this file.

So the whole PR can shrink to a one-line deletion in `NewFlowLanding.vue`. Could you update it accordingly? One more small thing: the commit scope should be `flows` rather than `ui`.

MilosPaunovic:
As no response on this, closing.

Thanks for the efforts so far! 🍀

PR #18307, opened 2026-08-20
fix(triggers): remove the Flow trigger window.fireOnce property

anna-geller:
`window.fireOnce` defaults to `false`, so the dependency results stored in the window are not reset after the trigger has fired. The trigger is re-evaluated on every terminal execution state change in the tenant, so another execution is then created by any flow that finishes, including flows not listed in `dependsOn`. The 1.x equivalent `preconditions.resetOnSuccess` defaulted to `true`, so migrated flows get the opposite behavior.

The property is removed and the stored results are always reset once the trigger has fired. `mode: ANY` covers creating an execution as soon as any single upstream flow succeeds. `getResetOnSuccess()` goes with it, since `DependsOnMultipleCondition` was its only implementer.

`flowTriggerDependsOnResetsAfterFiring` covers both halves on H2, with a `window` on the test flow so it exercises the shape that used to loop.

Closes https://github.com/kestra-io/kestra-ee/issues/10105. Paired with kestra-io/kestra-ee#10106 and kestra-io/kestra2-flow-migration#2; docs changes prepared separately.


loicmathieu (approved):
LGTM, I'm surprised we have more test and not less as we remove a functionality but I can have a look later if some tests are redondant.

You can merge it as is.

PR #18308, opened 2026-08-20
refactor(iam): rename the SUPER_ADMIN query filter to INSTANCE_OWNER

nkwiatkowski:
Renames the `SUPER_ADMIN` query filter field (wire value `super_admin`) to `INSTANCE_OWNER` (`instance_owner`) and the `AbstractJdbcRepository.getSuperAdminCondition` hook to `getInstanceOwnerCondition`, aligning the generic filter framework with the Enterprise `superAdmin` → `instanceOwner` rename.

The filter field is only used by the EE user / service-account / invitation listings; the UI (which ships in lockstep) is the only caller of the wire value.

## Companion
Companion to the Enterprise PR kestra-io/kestra-ee#10109, which renames the instance-ownership flag `superAdmin` → `instanceOwner` (issue kestra-io/kestra-ee#10069). This OSS change must merge together with it.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/QueryFilter.java
+++ b/core/src/main/java/io/kestra/core/models/QueryFilter.java
@@ -496,8 +496,8 @@ public record QueryFilter(
                 return List.of();
             }
         },
-        @JsonProperty("super_admin")
-        SUPER_ADMIN("super_admin") {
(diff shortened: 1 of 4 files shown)


loicmathieu (approved):
LGTM

PR #18309, opened 2026-08-20
fix(chart): correct worker group HPA target name

### LINKED ISSUE #18101: [BUG] Helm chart: HorizontalPodAutoscaler for a workerGroup targets a non-existent Deployment
## Description

Enabling `autoscaler` on a `workerGroups` entry renders a `HorizontalPodAutoscaler` whose `scaleTargetRef.name` points at a Deployment that does not exist. The HPA is created successfully and then silently never scales anything.

The Deployment block names worker-group Deployments `<fullname>-workergroup-<name>`, but the HPA block names both the HPA and its `scaleTargetRef` `<fullname>-<name>`, omitting the `workergroup-` infix.

## How to reproduce

`values.yaml`:

```yaml
configurations:
  application:
    kestra:
      repository:
        type: h2

deployments:
  standalone:
    enabled: false
  webserver:
    enabled: true

workerGroups:
  gpu:
    enabled: true
    workerThreads: 8
    autoscaler:
      enabled: true
      minReplicas: 2
      maxReplicas: 5
```

```bash
helm template kes kestra/kestra --version 1.3.30 -f values.yaml
```

## Actual result

```yaml
kind: HorizontalPodAutoscaler
metadata:
  name: "kes-kestra-gpu"
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: "kes-kestra-gpu"
  minReplicas: 2
  maxReplicas: 5
```

While the Deployment in the same release is named:

```yaml
kind: Deployment
metadata:
  name: "kes-kestra-workergroup-gpu"
```

## Expected result

`scaleTargetRef.name` should be `kes-kestra-workergroup-gpu`, matching the rendered Deployment.

## Impact

The failure is silent from Helm's point of view. `helm install` and `helm upgrade` both succeed, and the HPA object is created. Since the referenced scale target does not exist, the HPA cannot resolve it and never scales the worker group, which stays pinned at its static replica count. The mismatch is only visible if someone inspects the HPA's `AbleToScale` condition or compares `scaleTargetRef.name` against the Deployment list.

Note: the above is from rendered manifests (`helm template`), not from a deployed cluster, so the exact controller condition text is not quoted here.

## Root cause

In `charts/kestra/templates/deployment.yaml`, the Deployment `metadata.name` branches on whether the entry is a worker group:

```gotemplate
  {{- if and (eq $type "worker") (ne $name "worker") }}
  name: "{{ include "kestra.fullname" $ }}-workergroup-{{ $name }}"
  {{- else }}
  name: "{{ include "kestra.fullname" $ }}-{{ $name }}"
  {{- end }}
```

The HPA block further down the same file does not apply that branch (at `2.0.0-rc5` these are lines 341 and 346):

```gotemplate
kind: HorizontalPodAutoscaler
metadata:
  name: "{{ include "kestra.fullname" $ }}-{{ $name }}"
spec:
  scaleTargetRef:
    ...
    name: "{{ include "kestra.fullname" $ }}-{{ $name }}"
```

## Suggested fix

Compute the resource name once near the top of the `range` body and reuse it in both the Deployment and the HPA, so the two cannot drift again:

```gotemplate
{{- $resourceName := printf "%s-%s" (include "kestra.fullname" $) $name }}
{{- if and (eq $type "worker") (ne $name "worker") }}
{{- $resourceName = printf "%s-workergroup-%s" (include "kestra.fullname" $) $name }}
{{- end }}
```

Note that fixing only `scaleTargetRef` would leave the HPA object itself named `<fullname>-<group>`. That is functional, but it also means a worker group named after a built-in component (for example `workerGroups.executor`) would collide with that component's HPA name. Deriving both fields from one variable avoids this.

## Affected versions

Verified by rendering the published charts: chart `1.0.54` (app v1.3.15), `1.3.22`, `1.3.30`, and the current chart source at `2.0.0-rc5`. All produce the mismatched name.

## Workaround

Scale worker groups with a static `replicas` value instead of `autoscaler`, or attach an externally managed HPA that targets the correct `<release>-kestra-workergroup-<group>` Deployment.

AradhyaSharma31:
### Description

Fixes the HorizontalPodAutoscaler generated for worker groups so that it targets the correct Deployment.

Worker group Deployments are named using the `<fullname>-workergroup-<name>` pattern, while the HPA previously used `<fullname>-<name>` for both its own name and `scaleTargetRef.name`.

This change introduces a shared `$resourceName` variable and uses it for both the Deployment and HPA, keeping the generated resource names consistent.

### Related Issue

Closes https://github.com/kestra-io/kestra/issues/18101

### Backend Checklist

### Additional Notes

Verified the rendered Helm manifests using `helm template` with a worker group configured with an autoscaler. The HPA now correctly references the same `kes-kestra-workergroup-gpu` Deployment.

### DIFF
--- a/charts/kestra/templates/deployment.yaml
+++ b/charts/kestra/templates/deployment.yaml
@@ -13,6 +13,10 @@
 {{- $type = $name }}
 {{- end }}
 {{- end }}
+{{- $resourceName := printf "%s-%s" (include "kestra.fullname" $) $name }}
+{{- if and (eq $type "worker") (ne $name "worker") }}
+{{- $resourceName = printf "%s-workergroup-%s" (include "kestra.fullname" $) $name }}
+{{- end }}
 {{- $dind := false }}
 {{- if and ($global.dind.enabled) (eq $type "worker") }}
 {{- $dind = true }}
@@ -24,11 +28,7 @@
 apiVersion: apps/v1
 kind: {{ $kind }}
 metadata:
-  {{- if and (eq $type "worker") (ne $name "worker") }}
-  name: "{{ include "kestra.fullname" $ }}-workergroup-{{ $name }}"
-  {{- else }}
-  name: "{{ include "kestra.fullname" $ }}-{{ $name }}"
-  {{- end }}
+  name: "{{ $resourceName }}"
   labels:
     {{- include "kestra.labels" $ | nindent 4 }}
     app.kubernetes.io/component: {{ $type }}
@@ -173,7 +173,7 @@ spec:
           command:
             - "sh"
             - "-c"
-            - | 
+            - |
                 exec /app/kestra server {{ $type }} \
                 {{- if $content.workerThreads }}
                 {{- if eq $type "standalone" }}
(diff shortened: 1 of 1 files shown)


AradhyaSharma31:
Hi @Skraye,

could you take a look at this PR? It’s been open for over two weeks and has been sitting idle. Would appreciate your review so we can move this forward. Thanks!

AradhyaSharma31:
Hi @Skraye, just following up on this PR. Would appreciate a review when you get a chance. Thanks!

gperreymond (approved):

PR #18311, opened 2026-08-20
fix(design-system): stop registering all element-plus components globally

flcarre:
## Summary
- `KestraDesignSystem.install()` called `app.use(ElementPlus, {namespace: "kel"})`, which references element-plus's internal array of all ~96 components — this defeats tree-shaking for the ~37 (`ElTableV2`, `ElCascader`, `ElCalendar`, `ElTransfer`, `ElTreeV2`, `ElImage`, `ElMention`, `ElTour`, `ElWatermark`, `ElSelectV2`, `ElAnchor`, ...) that no `Ks*` wrapper actually uses.
- Every `Ks*` component already imports its own `El*` dependency directly in `<script setup>`, so global string-based registration is unnecessary — only `provideGlobalConfig` (the `"kel"` CSS namespace) is still needed, and is now called directly.
- Verified via a real production build: `design-system-*.js` drops from 1,140.54 kB to 929.48 kB (gzip 334.87 kB → 260.04 kB), ~18-22% smaller. `vendor-*.js` size is unchanged, confirming no duplication.

## Test plan
- [x] `vue-tsc --noEmit` clean
- [x] `oxlint` / `eslint` clean on the changed file
- [x] Design-system unit suite: 97 files / 780 tests passing
- [x] Design-system Storybook component-test suite (real Chromium rendering of every story): 80 files / 479 tests passing
- [x] Static verification that no template anywhere in `ui/src`, `ui/packages/design-system`, `ui/packages/topology` (and the local EE checkout) references an `El*` tag without a matching local import — confirms nothing relies on the removed global registration
- [x] Independent code review (kestra-core-code-reviewer agent): APPROVE, 0 findings


flcarre:
## Live QA follow-up

Ran the actual app in a browser against a running OSS backend to sanity-check this change beyond static analysis + Storybook. Found a real regression in the first commit and fixed it here (61fff91).

**What broke:** dropping `app.use(ElementPlus, {namespace: "kel"})` entirely also skipped installing Element Plus's own service plugins (`ElMessage`, `ElMessageBox`, `ElNotification`, `ElLoading`, `ElInfiniteScroll`, `ElPopoverDirective`). Their `.install(app)` is what wires `_context` to the app instance — without it, their detached render trees (e.g. an `ElNotification`'s content) can't resolve globally-registered `Ks*` components anymore.

**Repro:** triggered any error toast (`ErrorToastContainer.vue`, which renders `<KsButton>`/`<KsMarkdown>` and relies on global registration, not a local import) — console showed:
```
[Vue warn]: Failed to resolve component: KsButton
[Vue warn]: Failed to resolve component: KsMarkdown
```
Both silently failed to render inside the toast.

**Fix:** re-install just those 6 plugin entries via `app.use()` — they were already in the bundle anyway (`KsMessage.ts`/`KsNotification.ts`/etc. import them directly), so this adds ~2.5 kB back, not the ~211 kB the tree-shaking fix removed. Confirmed fixed live (toast now renders correctly with icon/title/close button, no console warnings) and re-verified: `vue-tsc` clean, 780/780 unit tests, 479/479 Storybook tests, final `design-system-*.js` = 932.01 kB (gzip 269.77 kB) vs the original 1,140.54 kB.

ui/packages/design-system/src/index.ts
```diff
@@ -455,7 +464,16 @@ export {
...
     install(app: App) {
         if (!(app as any)[INSTALLED_KEY]) {
-            app.use(ElementPlus, {namespace: "kel"})
+            // Every Ks* component imports its own El* dependency directly, so global registration
+            // is unneeded and only defeats tree-shaking of the ~96 Element Plus components. The
+            // services below still need app.use(): it's what wires their _context to this app, so
+            // their detached render trees (e.g. an ElNotification's content) can still resolve
+            // globally-registered Ks* components like KsButton/KsMarkdown.
+            (app as any)[INSTALLED_KEY] = true
+            provideGlobalConfig({namespace: "kel"}, app, true)
```
MilosPaunovic, line 473: Not blocking, worth a note for future readers: this hand-replicates `makeInstaller` from element-plus 2.14.2 (set `INSTALLED_KEY`, install the plugin list, `provideGlobalConfig` with `global = true`). If element-plus ever adds a 7th entry to its `plugin.ts` or changes the installer contract, this needs to be revisited on upgrade - might be worth pinning that expectation with a small unit test that compares this list against `element-plus/es/plugin.mjs`'s default export.

MilosPaunovic (changes requested):
Nice work - the approach is sound and well-verified. I checked the mechanics against element-plus 2.14.2's own installer: the 6 reinstalled plugins (`ElInfiniteScroll`, `ElLoading`, `ElMessage`, `ElMessageBox`, `ElNotification`, `ElPopoverDirective`) exactly match its `plugin.ts` list, and the manual `INSTALLED_KEY` + `provideGlobalConfig(..., app, true)` sequence reproduces `makeInstaller`'s semantics one-for-one, minus the component registrations. I also swept the `EE` checkout and dynamic resolution paths (`resolveComponent("el-...")`, `h("el-...")`) - all clean.

One blocking issue, and it is what makes the `Frontend - Tests` check red: the static verification only covered `ui/src` and the packages, but `ui/tests/storybook/theme/ShowCase.vue` renders ~42 raw `el-*` tags (`el-button` x25, `el-select`, `el-table`, `el-alert`, `el-date-picker`, ...) with no element-plus imports - it was the one remaining consumer of the removed global registration, and its story now fails with `Failed to resolve component: el-button`.

Fix: import the needed `El*` components locally in `ShowCase.vue`. Raw element-plus usage is intentional there (the story exists to exercise the `kel` theme), and the `elementPlusGuard` ban only applies to feature code in `ui/src`, so local imports in a test file are fine.

Happy to approve once that story is green.

flcarre:
@MilosPaunovic Good catch, fixed in 968ea6b.

Imported the needed \`El*\` components locally in \`ShowCase.vue\` and rewrote the ~40 raw tags to PascalCase (\`el-button\` → \`ElButton\`, etc.) to match the project's \`vue/component-name-in-template-casing\` convention — the kebab-case tags were silently exempt from that lint rule while unregistered (\`registeredComponentsOnly: true\`), but trip it once locally imported.

That also surfaced a real pre-existing type gap: with \`ElAlert\` now actually resolved, \`vue-tsc\` correctly flagged \`:type=\"type.toLowerCase()\"\` as a plain \`string\`, not the literal union \`ElAlert\` expects. Replaced the inline array with a typed \`alertTypes\` const + a small \`capitalize()\` helper for the title text.

Re-verified: \`vue-tsc\`, \`eslint\`, \`oxlint\` all clean on the file; full \`ui/\` storybook suite 68/68 files (244/244 tests) including the \`ElementPlusPlayground\` story itself; full \`ui/\` unit suite 191/191 files (1826/1826 tests).

MilosPaunovic (approved):
Re-review clean. ✅ ShowCase.vue is now locally importing the 15 El* components it renders (`ElButton`, `ElAlert`, `ElTag`, `ElSwitch`, `ElInput`, `ElDatePicker`, `ElTable`, `ElTableColumn`, `ElSelect`, `ElAutocomplete`) - converted to PascalCase tags, added template logic helpers (`alertTypes`, `capitalize`). Both `Frontend - Tests` and `Frontend - Design System tests` now pass. Code looks solid; bundle savings verified as described in the PR. Ready to merge.

Issue #17569, opened 2026-07-22
Allow to interrupt a task run from the task run detail menu


loicmathieu:
### Feature description

From the UI, you can do several actions on a task run but not _interrupting it_.

<img width="2145" height="760" alt="Image" src="https://github.com/user-attachments/assets/828b9bb2-054b-478d-9b78-4866f714aa21" />

Add a new action _interrupt_ (or stop run) for taskrun in the RUNNING state, allowing to interrupt their processing inside the Worker and end them in a specific state (FAILED/CANCELLED).

This will reuse the current ExecutionKill event to ask the worker to kill the task run in a specific state.

**Why isn't it called KILLED**: you can only kill an execution as kill is shortcircuiting any processing tasks (if there are multiple in parallel) and any _normal execution processing_: allow failure is not respected for example. Here interrupting the task run gives more power: you decided if you want to continue (CANCELLED) or fail the task (FAILED). But the execution continue processing as for any standard failure.

jaiten-made:
Hi maintainers,

I'd like to investigate this feature and work on the implementation. Could you please assign this issue to me?

Thanks!

MilosPaunovic:
Absolutely, go for it @jaiten-made, thanks! 🚀

jaiten-made:
Hi @MilosPaunovic and @loicmathieu,

The implementation is ready for review in PR #17592

I have added the Interrupt option in the task run detail actions, allowing users to choose either FAILED or CANCELLED as target states. I've also implemented the backend changes so the worker terminates only the targeted task run (allowing the rest of the execution flow to continue normally).

Please let me know if you have any feedback or if you'd like any adjustments!

PR #18314, opened 2026-08-20
feat(execution): allow flowable parallel tasks to fail-fast they chil…

loicmathieu:
…d tasks

Parallel and Dag can have one failing tasks and other still processing concurrenctly which would lead to those tasks being run un-necessary.

You can now configure `onChildFailure`  on both so when one child task fail, the worker can be asked to interrupt any still running child tasks in FAILED or CANCELLED.

https://github.com/kestra-io/kestra/issues/15365

## Breaking change

The default is `FAILED` so now Dag and Parallel will fail concurrently running tasks when one task fail.
Use `CONTINUE` to keep the previous behavior.

### DIFF
--- a/executor/src/main/java/io/kestra/executor/ExecutorService.java
+++ b/executor/src/main/java/io/kestra/executor/ExecutorService.java
@@ -559,7 +559,14 @@ public class ExecutorService {
             if (taskRun.getState().isRunning() && task instanceof FlowableTask<?> && this.canChildrenProgress(executor.getExecution(), taskRun)) {
                 RunContext runContext = runContextFactory.of(executor.getFlow(), task, executor.getExecution(), taskRun);
                 nextTaskRuns.addAll(this.childNextsTaskRun(executor, taskRun, runContext));
-                this.childWorkerTaskResult(executor.getFlow(), executor.getExecution(), taskRun, runContext).ifPresent(list::add);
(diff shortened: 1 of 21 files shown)


loicmathieu:
@fhussonnois FYI this PR adds a new capacity to the Worker to interrupt individual tasks (and not the whole execution) in a target state. I plan to allow doing that for any task run from the API later in 2.1 as described some time ago in https://github.com/kestra-io/kestra/issues/17569

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -841,6 +852,59 @@ else if (executor.getFlow().getRetry() != null) {
         return executor;
     }
 
+    /**
+     * Interrupts every still-running task run in the task subtree so they don't keep executing on an already-failed branch.
+     * Send an {@link ExecutionKilledTaskRuns} event to the Worker to interrupt the non-terminated task runs.
+     */
+    private void interruptOnChildFailure(ExecutorContext executor, OnChildFailureInterface onChildFailure, TaskRun parentTaskRun, RunContext runContext) throws InternalException {
```
AcevedoR, line 859: this method might be a good candidate for a unit test suite that covers edge cases

AcevedoR (approved):

loicmathieu:
@Ben8t it contains a new (last minute) small breaking change

PR #18316, opened 2026-08-20
fix(flows): prevent blank screen crash when deleting a revision

### LINKED ISSUE #18094: [2.0] Page goes blank after deleting a revision

Tejaskrishnan:
### ✨ Description

Fixes a crash when deleting a revision in the revisions dropdown comparison view. 

Previously, when a revision was deleted, the indices representing the left and right revisions compared (`revisionLeftIndex` and `revisionRightIndex`) pointed to invalid or out-of-bounds positions because the array shifted. Accessing those shifted indices directly threw a `TypeError` (e.g. `Cannot read properties of undefined (reading 'revision')`), which crashed the Vue rendering lifecycle and broke the page.

This PR fixes it by:
1. Making index-to-revision helper functions (`revisionNumber`, `revisionIndex`) safe against `undefined` inputs and boundary issues.
2. Changing key watcher hooks to observe computed, stable revision numbers (`selectedLeftRevision`, `selectedRightRevision`) instead of raw index values.
3. Adding a watcher on `sortedRevisions` to trigger index loading updates when the revision list is updated.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18094

### 🎨 Frontend Checklist

https://github.com/user-attachments/assets/bac57292-8b24-42f2-9f21-6bdfa7a0e553


ui/src/components/layout/Revisions.vue
```diff
@@ -250,13 +264,17 @@
...
-            router.push({
-                query: {
-                    ...route.query,
-                    revisionLeft: sortedRevisions.value[revisionLeftIndex.value].revision,
-                    revisionRight: sortedRevisions.value[revisionRightIndex.value].revision,
-                },
-            })
+            const leftRevision = revisionNumber(revisionLeftIndex.value)
+            const rightRevision = revisionNumber(revisionRightIndex.value)
+            if (leftRevision !== undefined && rightRevision !== undefined) {
```
flcarre: **This turns the crash into a state the user can't get out of.** Guarding on "both sides defined" means the `router.push` is skipped entirely, so a stale `revisionRight` / `revisionLeft` query param pointing at a deleted revision can never be cleared.

Concretely: open a shared link like `?revisionLeft=2&revisionRight=7` where revision 7 has since been deleted. `revisionNumber(revisionRightIndex)` returns `undefined`, the push is skipped, and the right column *and* the editor stay hidden — permanently. Changing the left dropdown no longer updates the URL either, because that path also goes through `addQuery`, so there's no way back short of hand-editing the address bar.

The guard should push whatever *is* resolvable and drop the param that isn't, rather than bailing out. Something like: build the query from the defined side(s), and explicitly set the stale key to `undefined` so `router.push` removes it.

---

ui/src/components/layout/Revisions.vue
```diff
@@ -227,18 +235,24 @@
         }
     }
 
-    function revisionIndex(revision: string) {
-        const revisionInt = parseInt(revision)
+    function revisionIndex(revision: string | number | undefined) {
+        if (revision === undefined || revision === null) return undefined
```
flcarre: Re-deriving indices from stable revision numbers is the right direction, and returning `undefined` for a revision that no longer exists is correct here. But `load()` doesn't finish the repair.

At line 233 the left-index fallback is `else if (revisionRightIndex.value !== undefined && revisionRightIndex.value > 0)`. When the right index resolves to `undefined` — exactly the deleted-revision case this PR is about — that branch is skipped, so a stale out-of-range `revisionLeftIndex` from the previous render survives untouched.

The template still renders that column, since it only checks `!== undefined`. The result is a `KsSelect` with no matching option, `revision: undefined` handed to the `crud` slot, and a Restore button that now silently no-ops through the new `revNum === undefined` early return at line 254 — a dead button with no feedback.

Worth handling the two indices independently: validate each against `sortedRevisions` and repair whichever one is stale, instead of making the left branch conditional on the right one having resolved.

---

ui/src/components/layout/Revisions.vue
```diff
@@ -374,12 +401,16 @@
...
 
-    watch(() => currentRevisionWithSource.value.revision, (newRevision, oldRevision) => {
-        if (revisionNumber(revisionRightIndex.value) === oldRevision) {
+    watch(() => currentRevisionWithSource.value?.revision, (newRevision, oldRevision) => {
+        if (newRevision && oldRevision && revisionNumber(revisionRightIndex.value) === oldRevision) {
             revisionRightIndex.value = revisionIndex(newRevision.toString())
         }
     })
 
+    watch(sortedRevisions, () => {
```
flcarre: This watcher fires on array *identity*, not content. `sortedRevisions` is a `computed` returning a fresh `toSorted` array, so every re-evaluation produces a new reference and this callback runs — and `load()` resets **both** selections.

Two consequences:

- For `editRouteQuery: false` consumers (the namespace-file history in FileExplorer) there is no URL to restore the selection from, so `load()` falls back to the defaults. Restoring a file therefore discards the pair the user was comparing — 1 vs 3 silently becomes 5 vs 6.
- For flows it re-runs on every navigation, because FlowRevisions' route watcher refetches on each query change, so this fires even when the revision list is identical.

Keying the watcher on the joined revision numbers rather than the array (`() => sortedRevisions.value.map(r => r.revision).join(",")`) and then only repairing *invalid* selections instead of calling the full `load()` would fix both. That's the same trap documented under "The deep-watch / computed-spread trap" in `ui/AGENTS.md` — a computed source that returns an object needs to return a primitive instead.

---

flcarre (commented):
The diagnosis is right, and so is the direction. The blank screen came from `revisionNumber()` dereferencing `sortedRevisions[index]` with an index that went out of range once the list shrank, called straight from the render path (`revisionNumber(item.value)` and the `crud` slot), which took down the render tree. Guarding `revisionNumber` / `revisionIndex` plus the `currentRevisionWithSource?.` optional chaining does stop that, and re-deriving indices from stable revision numbers instead of positions is the correct fix rather than a patch.

What I'd like resolved before merging is that the crash is currently replaced by a stuck state rather than a recovered one — three points inline, the first being the one that matters most (a shared link with a deleted `revisionRight` becomes permanently half-blank with no way back).

Two other things I looked into and deliberately am **not** asking you to change, so you don't spend time on them:

- The double content load caused by the left/right watchers running before the `sortedRevisions` watcher in the same flush. Harmless for flows — `listFlowRevisions` returns `FlowWithSource`, so `loadRevisionContent` is a cache hit and the correct value still lands last.
- The missing `revisionSource` check in `restoreRevision`. The `:disabled` comparison against `currentRevisionWithSource?.source` happens to block the reachable window already.

Neither has a scenario that actually misbehaves today.

---

Tejaskrishnan:
Thanks for the thorough review and clear breakdown, @flcarre.

I'm implementing the changes based on your three points:
1. **URL params & Stuck state:** Building the query from valid sides and setting missing/stale keys to `undefined` so `router.push` cleans up the URL.
2. **Independent index validation in `load()`:** Decoupling left and right index fallbacks so a stale index is repaired regardless of the other's state.
3. **Watcher source:** Keying `sortedRevisions` by joined revision numbers (`() => sortedRevisions.value.map(r => r.revision).join(",")`) and repairing only invalid selections without resetting the whole state.

Before pushing the update, I mapped out the test cases below to ensure we don't leave any stuck states or stale queries. Could you confirm if this test matrix covers all the scenarios you had in mind, or if there are other edge cases you'd like added?

---

### Test Matrix

#### 1. Basic & URL Load
* **No URL params:** Right shows latest/current revision; left shows the one immediately prior.
* **Both valid (`?revisionLeft=2&revisionRight=4`):** Both load and display correctly.
* **Both missing/invalid (`?revisionLeft=99&revisionRight=98`):** Both stale query params are dropped; defaults load cleanly.
* **Left missing, right valid (`?revisionLeft=99&revisionRight=4`):** `revisionLeft` dropped from URL; right stays `4`.
* **Left valid, right missing (`?revisionLeft=2&revisionRight=99`):** `revisionRight` dropped from URL; left stays `2`.
* **Only `?revisionRight=3` provided:** Right loads `3`; left defaults to prior revision.
* **Only `?revisionLeft=2` provided:** Left loads `2`; right defaults to current/latest.

#### 2. Deletion Scenarios
* **Selected revision deleted (e.g., L=6, R=7 -> delete 7):** Left stays 6; right updates to the new current revision.
* **Selected revision deleted (e.g., L=6, R=8 -> delete 6):** Left falls back to oldest available; right stays 8.
* **Non-selected revision deleted (e.g., L=4, R=8 -> delete 6):** Selections, content, and URL remain unchanged.
* **Content Verification:** YAML content re-syncs accurately for both panels after any deletion (no stale cache/content).
* **Double delete:** Second delete triggers an error toast gracefully; UI remains interactive and stable.

#### 3. Dropdowns & UI Controls
* **Dropdown updates:** Changing left or right updates the respective YAML content and the URL.
* **Restore button state:** Disabled if showing current revision; enabled with confirmation prompt when viewing an older revision.
* **Option exclusion:** Selecting the exact same revision on both sides is prevented in dropdown options.

#### 4. Navigation & Reactivity (`editRouteQuery: false` & Edge Cases)
* **Browser history:** Back/Forward buttons cleanly restore previous query/selection states.
* **Minimal revisions:** Works when only 2 revisions exist and one is deleted.
* **External list updates:** When `sortedRevisions` changes externally (or via non-URL consumers like FileExplorer), only invalid selections get repaired—preserving valid custom comparisons.

flcarre (changes requested):
The root cause is correctly identified, which is the hard part here. `revisionNumber()` doing `sortedRevisions.value[index].revision` on an index left pointing past the end of the array throws during render — reached from the `#crud` slot bindings at `Revisions.vue:52` and `:94`, and from the delete-icon `v-if` at `:36` and `:78` — and a render-time throw is what blanks the page. Guarding that function fixes it. A few things to address before merge.

### There is no test

This is a crash fix with a single file changed and nothing pinning it. Per `AGENTS.md`, a bug fix should come with a test that fails on the commit before yours. #18827 does this for the same class of bug in a sibling component, so there is a pattern to follow: mount `Revisions` with a small revision list, delete the revision the indices point at, and assert the component still renders. That is the one change I would insist on — everything else below is refinement.

### Four of the optional chains are dead code

The template body is already gated by `v-if="revisions && revisions.length > 1"` at line 2, so inside it `sortedRevisions` is never empty and `currentRevisionWithSource` is never undefined. That makes these unnecessary:

- `:disabled="revisionLeftText === currentRevisionWithSource?.source"` (line 43)
- the same on line 85
- `currentRevisionWithSource.value?.revision` in `formatRevisionText`
- `currentRevisionWithSource.value?.revision` in `options`

The one that *is* needed is the watcher getter:

```ts
watch(() => currentRevisionWithSource.value?.revision, …)
```

because a watch getter runs regardless of the `v-if`. Keeping the other four costs nothing at runtime but tells the next reader that `currentRevisionWithSource` can be undefined in the template, which is not true, and hides the fact that the watcher is the real hazard.

### `watch(sortedRevisions, () => load())` overlaps with what is already there

`onDelete` already calls `load()`, and this new watcher will fire for the same prop update, so the reset now happens twice by two different routes. If the explicit call in `onDelete` was the broken one — it runs before the parent has refreshed `props.revisions`, so it reads a stale list — then delete it and keep the watcher, rather than leaving both.

The watcher also has a reach beyond the bug: it re-runs `load()` on *any* change to `props.revisions`, and `load()` re-derives the indices from `route.query`. For `FileExplorer.vue`, which passes `editRouteQuery: false` and therefore never writes the selection to the URL, that means any refresh of the revisions array silently resets the user's chosen comparison to latest-vs-previous. After a restore that is arguably what you want, but it is a behaviour change that is not mentioned anywhere — please either scope the watcher to the case it needs to handle, or state the intent in the description so it is a decision rather than a side effect.

### The index → revision → index round trip

`selectedLeftRevision` converts an index to a revision number, and the watcher then converts it straight back with `revisionIndex(newValue)` before calling `loadRevisionContent`. `loadRevisionContent` already returns `undefined` for an undefined index, so the two new computeds, the `undefined` early-returns in the watchers, and the `typeof revision === "number"` branch added to `revisionIndex` are all carrying a value that is thrown away again.

If the intent is "re-fire when the array shifts under a stable index", say so in a one-line comment — it is a real and non-obvious reason. If not, `watch(revisionLeftIndex, …)` plus the existing guard in `loadRevisionContent` is the smaller change.

### Pre-existing, but this PR adds triggers to it

Both revision watchers are async and both flip the shared `isLoadingRevisions` flag, with no sequence guard: whichever `loadRevisionContent` resolves last wins, and the faster one clears the loading flag while the other is still in flight. `ui/AGENTS.md` calls for dropping superseded responses in exactly this shape. Not yours to fix, but this PR adds two more paths that can start a load concurrently, so it is closer to the surface now than it was.

### Nits

- `if (revision === undefined || revision === null)` — the signature says `string | number | undefined`, so the `null` arm is guarding against something the types say cannot arrive.
- `addQuery()` silently does nothing when either index is undefined. Correct as a crash guard, but it means the URL and the visible selection can drift apart with no indication.
- With the guard in place, `revisionNumber(revisionLeftIndex)` can now hand `undefined` to the `#crud` slot. Worth a quick check that `Crud` renders sensibly with `revision: undefined` rather than producing a broken link.

Tejaskrishnan:
Hi  @flcarre , thanks again for the detailed review! I've updated the PR with all feedback:

1. **Added Regression Test**: Added `DeleteSelectedRevision` in `Revisions.stories.tsx` testing the exact crash scenario (mount with revision list -> delete selected revision -> confirm deletion -> assert component renders safely without blanking the page).
2. **Removed Dead Optional Chains**: Removed the unnecessary `?.` on `currentRevisionWithSource.value` inside `formatRevisionText` and `options()`.
3. **Reactivity & URL Query**: Ensured both left and right revision selections resolve independently, repairing stale selections and removing missing query params from the URL via `addQuery()`.

flcarre (approved):
All three points from my last review are properly fixed — regression test reproduces the exact crash, dead optional chains removed, and left/right revision resolution is now independent with stale query params actually dropped from the URL (confirmed against vue-router's stringifyQuery). Approving.

One small thing for later: the new delete icon has no `data-test` hook, so the regression test falls back to a generic selector — worth adding `data-test="delete-revision"` at some point. And a heads up, unrelated to this PR: `FileExplorer.vue` uses `Revisions` for namespace-file history without `canDelete=false` or a `@deleted` handler, so deleting there would hit the flow-revision-delete API instead — pre-existing, worth its own issue.

MilosPaunovic:

MilosPaunovic:
Another ping @flcarre, is this now in a state to be merged?

PR #18318, opened 2026-08-20
fix(secrets): stop sending decrypted secrets to the worker

fhussonnois:
### ✨ Description

Secrets were decrypted when the run context was **built**, and `WorkerTaskData.from()` then copied that variables map onto the worker job queue. `SECRET` inputs, `SECRET` flow outputs and encrypted trigger variables therefore travelled in cleartext — and for the JDBC and Kafka queue implementations they were **persisted** that way.

Decryption now happens when the variables are **read** instead of when they are built, so the encrypted form is what exists at the serialization boundary:

- `RunVariables.DefaultBuilder.build()` no longer decrypts anything.
- `DefaultRunContext` decrypts the `inputs`, `outputs` and `trigger` subtrees once, memoized, and registers each value it decrypts with the logger for masking.
- `WorkerRunContextData.filterVariables` reads the pre-decryption map, so the wire carries `EncryptedString` envelopes.
- `WorkerTaskData.secretOutputs` is removed — it carried plaintext secret values on the job envelope purely so the worker could mask them. The worker now derives them while decrypting.

This **moves** the existing single decryption pass rather than adding one, so a run context costs no more to build than before, and nothing at all when it is built and never read. Everything the eager pass needed and the lazy one does not is deleted with it: the builder's `decryptVariables` flag, secret key, `secretInputs` and `secretOutputs`, `RunContext.getSecretOutputs()`, and the `EncryptionConfig` the worker-side initializer no longer uses. Net **−100 lines** of production code.

Note that a plain "build the worker context undecrypted" fix would have been wrong: the executor renders `runIf` (`ExecutionEventMessageHandler`, `ExecutorService`) and subflow input values (`createSubflowExecutions`) against this same context, so it genuinely needs plaintext locally. Decrypt-on-read keeps those working while the serialized form stays encrypted.

**Two further bugs were found on the way, both fixed here because this change relies on them being correct:**

1. **Plaintext was being written into the live `Execution`.** `decodeInput` mutated the nested input map it was handed, and that map is shared with `Execution.getInputs()` through a shallow copy — so decoding a `FORM`-nested `SECRET` input replaced the ciphertext on the `Execution` the executor goes on to persist. Adding `_context` to the trigger variables mutated the `Execution` the same way.
2. **Log masking was order-dependent.** Secrets were replaced in registration order, so a secret that is a prefix of another left the remainder visible — `pass` registered before `password` produced `******word`. Secrets are now masked longest-first, and registering the same secret twice no longer grows the list.

`Secret.decrypt` also now handles both shapes an encrypted value can take: an `EncryptedString` instance when it never left the JVM, and a `Map{type, value}` once it has been through serialization. Only the latter was handled, so in-process paths silently skipped decryption.

### 🔗 Related Issue

Refs https://github.com/kestra-io/kestra-ee/issues/6624. Companion PR: kestra-io/kestra-ee on the same branch name (`fix/worker-job-encrypted-secrets`), which carries the EE builder cleanup and an EE-side regression test.

**Three commits, and the order is load-bearing:**

1. `fix(secrets): stop sending decrypted secrets to the worker` — the fix above.
2. `fix(secrets): encrypt a SECRET input resolved from its flow default` — closes a leak the first commit exposed. `FlowInputOutput` encrypts every `SECRET` value it reads, but the defaults fallback in `RunVariables` (*"useful for triggers"*) fills in inputs missing from the execution via `FlowInputOutput.resolveDefaultValue`, which routes `SECRET` through the same branch as `STRING` and returns **plaintext**. Since commit 1 makes the variables map the wire format, a `SECRET` resolved from its default was still serialized in cleartext: a run context for an execution with unresolved inputs produced `"inputs":{"secret":"password","nested":{"key":"password"}}`. Defaults are Pebble-rendered, so such a value can also be an expression resolving to a real secret rather than a literal. The fallback now encrypts, restoring the invariant commit 1 depends on — *every secret in the variables map is encrypted until it is read*.
3. `refactor(secrets): drop the secret-input list used for log masking` — with that invariant actually holding, the id list becomes dead. Masking was index-driven (`RunContext` carried the flow's `SECRET` input ids; `setLogger` walked the map to those positions with `findSecret`); it is now structural, since the decryptor reports each value it decrypts wherever it sits in the tree. Removes `RunContext.getSecretInputs()`, the field, builder and `clone()` plumbing, the `setLogger` block, `findSecret`, the `WorkerTaskData` / `WorkerTriggerData` wire components, `secretInputsFromFlow` with its four call sites, and the `ReusableInputsExpander` injection it needed. It also retires `findSecret`'s unguarded cast to `String`, which would have thrown `ClassCastException` on a context built with `decryptVariables=false`.

Commit 3 is only safe because of commit 2: beforehand a `SECRET` resolved from a default sat in the map as plaintext, nothing existed for the decryptor to report, and `findSecret` was the only thing masking it.

### 🛠️ Backend Checklist

### 📝 Additional Notes

**Regression tests.** `WorkerTaskDataTest` asserts on the *serialized* payload, which is the assertion that would have caught this originally — one case for supplied inputs, one for inputs resolved from flow defaults, plus one confirming `{{ inputs.secret }}` still renders plaintext. Both wire assertions were verified to fail when the fix is neutered. The fixture is `input-log-secret.yaml` because it declares a top-level *and* a dotted `SECRET` input, so one payload covers both code paths.

**Verification.**

| Suite | Result |
|---|---|
| `:core:test` | 3321 passing |
| `:worker:test` / `:executor:test` / `:worker-controller:test` | 113 / 159 / 81 passing |
| `:jdbc-h2:test --tests *H2RunnerTest*` | 99 passing |

3773 passing, zero failures. `InputsTest.shouldNotLogSecretInput` — which catches the prefix-masking bug fixed in commit 1 — passes with the id-driven registration removed, which is the check that mattered for commit 3. Companion EE branch: 1227 passing.

**Wire-format break.** `WorkerTaskData` loses both `secretOutputs` and `secretInputs`, and the variables encoding changes, so an old executor and a new worker are mutually incompatible mid-upgrade; a worker without `kestra.encryption.secret-key` can no longer decrypt what it receives. Accepted deliberately for 2.0. Not backported — the exposure exists back to at least 1.0, but there the whole `RunContext` is serialized into `WorkerTask` with `@JsonInclude` on `getVariables()`, so it needs a different patch.

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
+++ b/core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
@@ -77,7 +77,6 @@ public class DefaultRunContext extends RunContext {
     private String triggerExecutionId;
     private Storage storage;
     private Map<String, Object> pluginConfiguration;
-    private List<String> secretInputs;
     private List<String> secretOutputs;
     private String traceParent;
 
@@ -162,15 +161,6 @@ public class DefaultRunContext extends RunContext {
      */
     @Override
     @JsonInclude
-    public List<String> getSecretInputs() {
-        return secretInputs;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    @JsonInclude
     public String getTraceParent() {
         return traceParent;
     }
@@ -232,18 +222,6 @@ public class DefaultRunContext extends RunContext {
     void setLogger(final RunContextLogger logger) {
         this.logger = logger;
 
-        // this is used when a run context is re-hydrated so we need to add again the secrets from the inputs
-        if (!ListUtils.isEmpty(secretInputs) && getVariables().containsKey("inputs")) {
-            @SuppressWarnings("unchecked")
-            Map<String, Object> inputs = (Map<String, Object>) getVariables().get("inputs");
-            for (String secretInput : secretInputs) {
-                String secret = findSecret(secretInput, inputs);
-                if (secret != null) {
-                    logger.usedSecret(secret);
-                }
-            }
-        }
-
         // this is used when a run context is re-hydrated so we need to add again the decrypted SECRET flow outputs
         if (!ListUtils.isEmpty(secretOutputs)) {
             secretOutputs.forEach(logger::usedSecret);
@@ -259,18 +237,6 @@ public class DefaultRunContext extends RunContext {
         List<String> updated = new ArrayList<>(ListUtils.emptyOnNull(secretOutputs));
         updated.add(secret);
         secretOutputs = updated;
-    }
-
-    @SuppressWarnings("unchecked")
-    private String findSecret(String secretInput, Map<String, Object> inputs) {
-        if (secretInput.indexOf('.') > 0) {
-            String prefix = secretInput.substring(0, secretInput.indexOf('.'));
-            String suffix = secretInput.substring(secretInput.indexOf('.') + 1);
-            Map<String, Object> subInputs = (Map<String, Object>) inputs.get(prefix);
-            return findSecret(suffix, subInputs);
-        }
-
-        return (String) inputs.get(secretInput);
     }
 
     void setPluginConfiguration(final Map<String, Object> pluginConfiguration) {
@@ -301,7 +267,6 @@ public class DefaultRunContext extends RunContext {
         runContext.logger = this.logger;
         runContext.storage = this.storage;
         runContext.pluginConfiguration = this.pluginConfiguration;
-        runContext.secretInputs = this.secretInputs;
         runContext.secretOutputs = this.secretOutputs;
         runContext.decryptVariables = this.decryptVariables;
         if (isInitialized.get()) {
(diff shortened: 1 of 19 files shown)


core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
```diff
@@ -107,25 +112,48 @@ public String getTriggerExecutionId() {
     @Override
     @JsonInclude
     public Map<String, Object> getVariables() {
-        return variables;
+        if (!decryptVariables) {
```
loicmathieu, line 115: This is safe today because the runContext is never shared between threads.
It may be better to add a comment about that, or use already a more safe pattern (synchronized methods of DCL idiom).

core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
```diff
@@ -65,13 +69,14 @@ public class DefaultRunContext extends RunContext {
     private SDK sdk;
 
     private Map<String, Object> variables;
+    private boolean decryptVariables = true;
+    private Map<String, Object> decryptedVariables;
```
loicmathieu, line 73: This will double the size of the run context in memory :(

loicmathieu (approved):
LGTM

PR #18324, opened 2026-08-21
fix(core): display Reverse separator default in plugin docs

### LINKED ISSUE #14220: The default `\n` separator is not displayed in the Reverse task’s default value in the “Docs” tab.
### Describe the issue

For Revere Task, the default value for separator should show `\n` in the docs tab in editor but instead is blank.

Image

https://github.com/kestra-io/kestra/blob/e32819a1da94a4dbe5d2c4f19bdd93519bcf1d87/core/src/main/java/io/kestra/plugin/core/storage/Reverse.java#L57

I can't work out why it might be blank as the same `\n` is used for Split and ForEachItem and is displayed correctly. Thought it might have been trying to render the new line instead of showing the string but does not seem to be the case.

https://github.com/kestra-io/kestra/blob/e32819a1da94a4dbe5d2c4f19bdd93519bcf1d87/core/src/main/java/io/kestra/plugin/core/storage/Split.java#L83

Image

https://github.com/kestra-io/kestra/blob/e32819a1da94a4dbe5d2c4f19bdd93519bcf1d87/core/src/main/java/io/kestra/plugin/core/flow/ForEachItem.java#L647

Image

### Environment

- Kestra Version: develop

thurdass:
### ✨ Description

Expose the Reverse task separator default as the escaped `\\n` value in the generated plugin schema, so it is displayed correctly in the Docs tab. The runtime value remains a real newline.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/14220

### 🛠️ Backend Checklist

- [x] Code changes are limited to the backend
- [x] Added a regression test for the generated schema default

### 📝 Additional Notes

`git diff --check` passes. The targeted Gradle test could not start locally because the environment has Java 21 while the project requires Java 25; the failure occurs during Gradle toolchain configuration.

### 🤖 AI Authors

The cat tried to fix the newline, but it only managed to make one long line. 🐱

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/storage/Reverse.java
+++ b/core/src/main/java/io/kestra/plugin/core/storage/Reverse.java
@@ -72,7 +72,8 @@ public class Reverse extends Task implements RunnableTask<Reverse.Output> {
     private Property<String> from;
 
     @Schema(
-        title = "The separator used to join the file into chunks. By default, it's a newline `\\n` character. If you are on Windows, you might want to use `\\r\\n` instead."
+        title = "The separator used to join the file into chunks. By default, it's a newline `\\n` character. If you are on Windows, you might want to use `\\r\\n` instead.",
+        defaultValue = "\\n"
     )
     @Builder.Default
     private Property<String> separator = Property.ofValue("\n");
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
LGTM, thanks

PR #18344, opened 2026-08-21
fix(dashboards): correct execution chart durations and show time gaps

### LINKED ISSUE #18290: Executions timeline chart shows inconsistent time-bucket granularity and incorrect duration (0s) on hover
### Describe the issue

The "Total Executions" chart (visible on the Executions page and Dashboard Overview) has two related problems with its time-axis data:

1. **Inconsistent bucket granularity**: Data points are bucketed at different time granularities depending on age/density, but rendered as equal-width bars on the same axis. An execution that actually ran at 11:38 PM is bucketed and labeled as "23:00" (rounded to the hour), while nearby execution clusters are bucketed to the minute ("00:00", "00:01"). This makes the timeline visually misleading, it looks like executions are evenly spaced in time when the underlying bucket widths are wildly different.

2. **Incorrect duration on hover**: Hovering over the oldest bucket shows "duration = 0s", but the actual execution in that bucket (id `SZUV7xfy`, ran Wed Aug 19 2026 11:38 PM) had a real recorded duration of 1.02s per the Executions table.

**Steps to reproduce**
1. Execute several flows at different times, at least one more than an hour before the others (to span two different auto-bucket granularities)
2. Go to the Executions page, with the default `timeRange=PT24H` filter
3. Hover over the oldest/sparsest bar in the "Total Executions" chart

Image

4. Compare the tooltip's timestamp and duration against the actual execution's Start Date and Duration shown in the Executions table

Image

**Expected behavior**
- Time buckets should use a consistent granularity across the visible range, or the axis should clearly communicate when granularity changes.
- The duration shown in the tooltip should match the actual execution duration recorded in the Executions table.

**Actual behavior**
- Bucket granularity silently shifts from hourly to per-minute across the same chart.
- Tooltip shows `duration = 0s` for execution `SZUV7xfy`, which actually took 1.02s (Wed, Aug 19 2026, 11:38 PM, hello_world flow).

**Related context**
A similar complaint about this chart's duration computation was raised previously in #6242 ("Executions graph mean duration isn't computed as supposed"), reported on v0.19.12 in Dec 2024 and closed as "not planned." This report confirms the underlying problem still reproduces on the current 1.0.56, so it may be worth reopening or tracking as a fresh regression rather than a resolved issue.

### Environment

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/18290.

## What

The `Total Executions` chart (Executions page, default dashboard) showed wrong durations and a misleading time axis. This PR fixes both, across backend and frontend.

### Backend - duration values were wrong on all three JDBC databases

- **H2**: the `state_duration` generated column stored **seconds** (the raw `.state.duration` JSON value) while Postgres and MySQL store milliseconds, so every dashboard duration on H2 was 1000x too small (the reported `duration = 0s`). A new migration (`2.0.24-execution-duration-millis`) regenerates the column in milliseconds. The frozen baseline is untouched (checksum validation).
- **Postgres**: the query wrapped the column in `state_duration / 1000`, which Postgres evaluates as **integer division**, truncating every sub-second execution to 0. The division now goes through a decimal literal (`/ 1000.0`).
- **MySQL**: the aggregation layer wrapped every aggregated value in a jOOQ `cast(Double.class)`, which renders as `CAST(... AS DECIMAL)` on MySQL - scale 0, so every value was rounded to a whole number (0.5s + 1.5s summed to 3s). The cast is now a `coerce`, which keeps the Java-side type without emitting any SQL cast.

New shared repository test `dashboard_fetchData_durationAggregationInSeconds` runs against all three databases and guards all three defects (verified green locally on H2, Postgres and MySQL; the full execution + metric repository suites also pass on all three).

### Frontend - tooltip and axis

- The `TimeSeries` chart tooltip printed the duration line's raw number; it now formats it with `humanDuration` via the standard ECharts per-series `tooltip.valueFormatter` option, which `KsEchart`'s external tooltip honors from now on (covered by a new `KsEchart` unit test and shown in the existing Storybook story).
- The category axis only contained buckets that had data, so sparse executions looked evenly spaced no matter how far apart they ran - the "visually misleading" part of the issue. The chart now zero-fills the buckets between the earliest and latest returned dates (`fillTimeBucketLabels`, unit-tested), keeping real time gaps visible.
- The duration line is built on the same axis labels as the bars so both always stay aligned, and the previously hardcoded `date`/`duration` column keys now come from the chart definition.

Note: the reporter's "inconsistent bucket granularity" observation was the axis reading `23:00` / `00:00` / `01:00` - consistent hourly buckets that only looked mixed because empty buckets were missing; the zero-fill addresses the underlying confusion.

### Verified locally

Live on develop + H2 (`server local`), 3 executions totalling 0.768s: the preview endpoint returned `duration: 0.000767794` before the fix and `0.767794` after (migration applied on an existing database at boot). The rendered chart now shows 28 minute buckets spanning the gap between two execution groups, and the tooltip prints `0.76s` instead of a raw float.

A follow-up backend issue will be filed separately: the Postgres `state_duration` generated column uses `EXTRACT(MILLISECONDS FROM ...)`, which only returns the seconds field of the interval, so durations of 1 minute and longer lose their minutes/hours part (the new shared test deliberately stays sub-minute until that is fixed).


ammeek (approved):
LGTM

PR #18352, opened 2026-08-21
feat(assets): asset editable type

RemiBarthe:
Part of https://github.com/kestra-io/kestra-ee/issues/9231

This pull request introduces improvements to asset type handling and label display customization. The main changes include enhancing the `Asset` model to allow conditional type updates, adding tests to verify this behavior, and updating the UI to support a display value for labels.

### Asset update logic enhancements

* Modified the `toUpdated` method in `Asset.java` to accept a new `allowTypeChange` boolean parameter, enabling or preventing changes to the asset's `type` field based on this flag.

### Test coverage improvements

* Added a new test class `AssetTest.java` with tests to ensure `toUpdated` correctly keeps the previous type when type changes are not allowed, replaces the type when allowed, and falls back to the previous type if the incoming asset has no type.

### UI Enhancements

* Updated `Labels.vue` to support an optional `display` property for labels, allowing a custom display value in the UI while preserving the underlying value. [[1]](diffhunk://#diff-7e5e0bbde1df3a61d0cc63fe001dee757a6fa4382d936c7e39681a93f4794dedL11-R12) [[2]](diffhunk://#diff-7e5e0bbde1df3a61d0cc63fe001dee757a6fa4382d936c7e39681a93f4794dedR27)


MilosPaunovic:
Reviewed together with https://github.com/kestra-io/kestra-ee/pull/10154 (main feedback there). On the `OSS` side the `toUpdated` change is solid: `firstNonNull` on both branches means a `PUT` without a `type` can never null out the stored one, creation keeps the incoming `type` regardless of the flag, and `AssetTest` pins all four cases down nicely.

Two small things:

- The `Labels.vue` `display` prop is not mentioned in the `PR` description - worth a line there, since it is the only `UI` change in this `PR` and its consumer lives in `EE`.
- Title could use a scope per the convention (`feat(assets): ...`).

One cross-cutting point raised on the `EE` `PR`: since the previous metadata map is merged unconditionally, a `type` change keeps the old `type`'s metadata keys forever - see the discussion there before merging.

**Merge order:** this `PR` removes the 2-arg `toUpdated` without an overload, so it must land immediately before the `EE` `PR` or `EE` `develop` will not compile in between.

MilosPaunovic:
Re-reviewed - nothing left open in the code. The red `trigger-ee` / `kestra-ee/backend-tests` checks are a stale cross-branch pairing, not a real failure: the run started at 16:17 UTC and compiled against the `EE` branch before its `develop` merge landed at 16:30 UTC, so it hit the `instanceOwner` rename mismatch in `IAM` code (`AbstractJdbcUserRepository`, `ElasticSearchInvitationRepository`) - unrelated to `assets`. Both branches now contain both halves of the rename, so a plain re-run should be green.

Reminder for whoever merges: this `PR` removes the 2-arg `toUpdated` outright, so land it immediately before https://github.com/kestra-io/kestra-ee/pull/10154 or `EE` `develop` will not compile in between.

MilosPaunovic (approved):

PR #18363, opened 2026-08-21
fix: prevent NPE when Switch task has only defaults and no cases

### LINKED ISSUE #18357: [Bash2.0] Switch task with only 'defaults' (no 'cases') passes validation then NPEs at runtime
## Summary

A `io.kestra.plugin.core.flow.Switch` task that declares only `defaults` (no `cases`) passes validation but throws `NullPointerException` in the executor at runtime, failing the execution. Verified on `2.0.0-rc10`.

## Expected result

Either validation rejects a `Switch` with no `cases` at save time, or (better, since `defaults` alone is meaningful — "always run the default branch") the executor runs the `defaults` branch. It should not NPE.

## Repro steps

1. Create this flow (validates and saves — `SwitchTaskValidator` only rejects when *both* `cases` and `defaults` are empty):

```yaml
id: switch_npe
namespace: company.team
tasks:
  - id: sw
    value: "anything"
    defaults:
      - id: d
        message: default
```

2. Execute it.

## Actual result

Execution ends `FAILED` with, in the server logs:

```
java.lang.NullPointerException: Cannot invoke "java.util.Map.entrySet()" because "this.cases" is null
	at io.kestra.plugin.core.flow.Switch.childTasks(Switch.java:167)
	at io.kestra.plugin.core.flow.Switch.resolveNexts(Switch.java:194)
```

## Root cause

`core/src/main/java/io/kestra/plugin/core/flow/Switch.java`:
- `childTasks` (line ~167) and `outputs` (line ~206) dereference `cases.entrySet()` unconditionally.
- `cases` is null when only `defaults` is provided.
- `SwitchTaskValidator` allows this (it only fails when both are empty), so the mismatch surfaces at runtime as an NPE rather than a validation error.

## Suggested fix

Null-guard `cases` in `childTasks`/`outputs` (treat null as empty map), so a `defaults`-only `Switch` runs the default branch.

## Impact

Common flowable task; valid-looking config; runtime NPE contained to the one execution. Verified on `2.0.0-rc10` (commit `b48d8d1`, Docker H2).

Fayupable:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18357

### ✨ Description

A `Switch` task with only `defaults` (no `cases`) passes validation — the validator only rejects a Switch when both `cases` and `defaults` are missing — but crashes at runtime with `NullPointerException: Cannot invoke "java.util.Map.entrySet()" because "this.cases" is null`, because `childTasks()` and `outputs()` dereferenced `cases` without a null check.

`childTasks()` now treats a null `cases` as an empty map, so a defaults-only Switch falls through to the default branch instead of NPE'ing. `outputs()` short-circuits `defaults` to `true` when `cases` is null, matching the semantic meaning (the default branch was taken).

### 🛠️ Backend Checklist

### 📝 Additional Notes

Added `switch-defaults-only.yaml` and a `switchDefaultsOnly()` test in `SwitchTest.java` verifying a defaults-only Switch executes successfully and reports `defaults: true`. This reproduces the exact bug scenario: the flow passes validation (only rejected when *both* `cases` and `defaults` are empty) but previously NPE'd at runtime.

I commented on the issue asking to be assigned but haven't heard back yet — opening this PR in the meantime since the fix is small and ready. Happy to hold off or adjust if a maintainer prefers otherwise.

### 🤖 AI Authors

Why did the null `cases` map break up with the Switch task? It said there was no *entry* point to the relationship. 🐱

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/flow/Switch.java
+++ b/core/src/main/java/io/kestra/plugin/core/flow/Switch.java
@@ -164,7 +164,7 @@ public class Switch extends Task implements FlowableTask<Switch.Output> {
     @Override
     public List<ResolvedTask> childTasks(RunContext runContext, TaskRun parentTaskRun) throws IllegalVariableEvaluationException {
         final String value = rendererValue(runContext);
-        return cases.entrySet()
+        return (this.cases == null ? Map.<String, List<Task>>of() : this.cases).entrySet()
             .stream()
             .filter(throwPredicate(entry -> entry.getKey().equals(value)))
             .map(Map.Entry::getValue)
@@ -203,7 +203,7 @@ public class Switch extends Task implements FlowableTask<Switch.Output> {
         return Output.builder()
             .value(rendererValue(runContext))
             .defaults(
-                cases
+                this.cases == null || this.cases
                     .entrySet()
                     .stream()
                     .noneMatch(throwPredicate(entry -> entry.getKey().equals(rendererValue(runContext))))
--- a/core/src/main/java/io/kestra/plugin/core/flow/Switch.java
+++ b/core/src/main/java/io/kestra/plugin/core/flow/Switch.java
@@ -201,14 +201,10 @@ public class Switch extends Task implements FlowableTask<Switch.Output> {
 
     @Override
     public Switch.Output outputs(RunContext runContext) throws IllegalVariableEvaluationException {
+        final String value = rendererValue(runContext);
         return Output.builder()
-            .value(rendererValue(runContext))
-            .defaults(
-                MapUtils.isEmpty(this.cases) || this.cases
-                    .entrySet()
-                    .stream()
-                    .noneMatch(throwPredicate(entry -> entry.getKey().equals(rendererValue(runContext))))
-            )
+            .value(value)
+            .defaults(MapUtils.emptyOnNull(this.cases).keySet().stream().noneMatch(value::equals))
             .build();
     }
 
--- a/core/src/main/java/io/kestra/plugin/core/flow/Switch.java
+++ b/core/src/main/java/io/kestra/plugin/core/flow/Switch.java
@@ -27,6 +27,7 @@ import io.kestra.core.models.tasks.Task;
 import io.kestra.core.runners.FlowableUtils;
 import io.kestra.core.runners.RunContext;
 import io.kestra.core.utils.GraphUtils;
+import io.kestra.core.utils.MapUtils;
 import io.kestra.core.validations.SwitchTaskValidation;
 
 import io.swagger.v3.oas.annotations.media.Schema;
@@ -164,7 +165,7 @@ public class Switch extends Task implements FlowableTask<Switch.Output> {
     @Override
     public List<ResolvedTask> childTasks(RunContext runContext, TaskRun parentTaskRun) throws IllegalVariableEvaluationException {
         final String value = rendererValue(runContext);
-        return (this.cases == null ? Map.<String, List<Task>>of() : this.cases).entrySet()
+        return MapUtils.emptyOnNull(this.cases).entrySet()
             .stream()
             .filter(throwPredicate(entry -> entry.getKey().equals(value)))
             .map(Map.Entry::getValue)
@@ -203,7 +204,7 @@ public class Switch extends Task implements FlowableTask<Switch.Output> {
         return Output.builder()
             .value(rendererValue(runContext))
             .defaults(
-                this.cases == null || this.cases
+                MapUtils.isEmpty(this.cases) || this.cases
                     .entrySet()
                     .stream()
                     .noneMatch(throwPredicate(entry -> entry.getKey().equals(rendererValue(runContext))))
--- a/core/src/test/java/io/kestra/plugin/core/flow/SwitchTest.java
+++ b/core/src/test/java/io/kestra/plugin/core/flow/SwitchTest.java
@@ -100,6 +100,22 @@ class SwitchTest {
     }
 
     @Test
+    @LoadFlows(value = { "flows/valids/switch-defaults-only.yaml" }, tenantId = "switchdefaultsonly")
+    void switchDefaultsOnly() throws TimeoutException, QueueException, io.kestra.core.exceptions.InternalException {
+        Execution execution = runnerUtils.runOne(
+            "switchdefaultsonly",
+            "io.kestra.tests",
+            "switch-defaults-only",
+            null,
+            (f, e) -> ImmutableMap.of("string", "ANYTHING")
+        );
+
+        assertThat(execution.getState().getCurrent()).isEqualTo(State.Type.SUCCESS);
+        assertThat(execution.getTaskRunList().get(1).getTaskId()).isEqualTo("default");
+        assertThat((Boolean) taskOutputService.getOutputs(execution.findTaskRunsByTaskId("parent-seq").getFirst()).get("defaults")).isTrue();
+    }
+
+    @Test
     @LoadFlows(value = { "flows/valids/switch-impossible.yaml" }, tenantId = "switchimpossible")
     void switchImpossible() throws TimeoutException, QueueException {
         Execution execution = runnerUtils.runOne(
--- a/core/src/test/resources/flows/valids/switch-defaults-only.yaml
+++ b/core/src/test/resources/flows/valids/switch-defaults-only.yaml
@@ -1,0 +1,15 @@
+id: switch-defaults-only
+namespace: io.kestra.tests
+
+inputs:
+  - id: string
+    type: STRING
+
+tasks:
+  - id: parent-seq
+    type: io.kestra.plugin.core.flow.Switch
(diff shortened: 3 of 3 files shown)


Fayupable:
Note: I commented on the issue asking to be assigned but haven't heard back yet — opening this PR in the meantime since the fix is small and ready. Happy to hold off or adjust if a maintainer prefers otherwise.

loicmathieu:
~PRs should always target `develop`~

Sorry, PR mismatch

Fayupable:
Hi @loicmathieu, just checking in on this whenever you have a chance. Is there anything i should change or fix on my end? happy to update it

core/src/main/java/io/kestra/plugin/core/flow/Switch.java
```diff
@@ -164,7 +164,7 @@ public GraphCluster tasksTree(Execution execution, TaskRun taskRun, List<String>
     @Override
     public List<ResolvedTask> childTasks(RunContext runContext, TaskRun parentTaskRun) throws IllegalVariableEvaluationException {
         final String value = rendererValue(runContext);
-        return cases.entrySet()
+        return (this.cases == null ? Map.<String, List<Task>>of() : this.cases).entrySet()
```
flcarre: Nit, no runtime impact — both new guards hand-roll helpers that already exist in `io.kestra.core.utils.MapUtils`:

- here: `MapUtils.emptyOnNull(this.cases).entrySet()` instead of the `Map.<String, List<Task>>of()` ternary
- line 206: `MapUtils.isEmpty(this.cases) || …` instead of `this.cases == null || this.cases.entrySet()…`

`AGENTS.md` asks for those utilities to be reused rather than reimplemented. Beyond the convention, the current form states the null contract for `cases` twice in two different idioms, and the `outputs()` variant relies on `||` short-circuit ordering for its null safety — `MapUtils.isEmpty` says it once and doesn't.

---

flcarre (commented):
The fix is correct and complete — only one optional nit below. What I checked:

- **The two patched sites were the only unguarded ones.** `getCases()` / `this.cases` is dereferenced in exactly four places in `Switch.java` (133, 152, 167, 206) plus `SwitchTaskValidator:25`. `allChildTasks()` and `tasksTree()` already had `!= null` ternaries, and the validator uses `== null ||`.
- **Both changed expressions are semantically right.** With `cases == null`, `childTasks()` falls through `findFirst()` → `orElse(resolveTasks(this.defaults, …))`, i.e. the default branch. `outputs()` reporting `defaults = true` matches the existing `cases == {}` behaviour (`noneMatch` on an empty stream is `true`), so null and empty now behave identically. Operator precedence on line 206 is fine — the `==`/method chain binds tighter than `||`.
- **No new failure mode when `cases` *and* `defaults` are both null** (a validation-bypassed flow): `childTasks()` returns `null`, which `FlowableUtils.resolveState` (218) and `innerResolveSequentialNexts` (82) already handle. That is the pre-existing `switch-impossible.yaml` path, which is expected to end `FAILED`.
- **The new flow file doesn't desync the suite.** `Helpers.FLOWS_COUNT` is derived dynamically from the `flows/valids/` directory listing (`tests/src/main/java/io/kestra/core/Helpers.java:38`), so `FlowControllerTest`'s assertions are unaffected, and no test hardcodes a flow count or counts `switch`-matching flows.
- **The test earns its place and would fail before the fix** (the execution ended `FAILED` on the NPE), and it covers both changed lines — the `defaults: true` assertion goes through `outputs()`, the `default` task-id assertion through `childTasks()`.

One thing I noticed but that is **not** for this PR: `outputs()` re-renders `value` with `.skipCache()` inside the `noneMatch` predicate, once per case entry, so a non-deterministic `value` expression (e.g. `{{ now() }}`) can make the reported `defaults` flag disagree with the branch `childTasks()` actually selected. Pre-existing and untouched here — worth a separate issue if anyone cares.

---

flcarre (commented):
Thanks for this — it's a well-scoped first contribution: correct diagnosis, minimal diff, a sanctioned utility, and a test that genuinely reproduces the bug (rarer than it should be).

I checked the fix is at the right layer rather than papering over a validation gap. `SwitchTaskValidator` rejects a `Switch` only when `cases` **and** `defaults` are both empty, so a defaults-only Switch is a deliberately valid configuration and guarding at runtime is the coherent resolution — tightening the validator would break flows it has always accepted. This also matches what the issue asks for.

A few things I verified while reading:

- On `develop`, `cases.entrySet()` is unguarded in exactly two places, `childTasks()` and `outputs()`. `allChildTasks()` and `tasksTree()` already null-guard it, so graph and topology building are untouched by this change, and both `resolveState()` and `resolveNexts()` route through `childTasks()` — the single guard covers every execution path.
- `MapUtils.emptyOnNull` returns the same instance when non-null, so there's no behaviour change for existing flows, and `cases: {}` keeps its previous result.
- The new test covers both changed lines: had only `childTasks()` been fixed, the `defaults` assertion would have hit the `outputs()` NPE instead.

## One suggestion

`outputs()` and `childTasks()` now express the same null case two different ways in one file — `MapUtils.emptyOnNull(...)` in one, `MapUtils.isEmpty(this.cases) || this.cases...` in the other. Since you're already rewriting that expression, it's also worth hoisting `rendererValue(runContext)` out of the predicate: it's `runContext.render(this.value).skipCache()`, so a 3-case Switch currently re-renders the Pebble `value` up to 4 times per `outputs()` call.

```java
    @Override
    public Switch.Output outputs(RunContext runContext) throws IllegalVariableEvaluationException {
        final String value = rendererValue(runContext);
        return Output.builder()
            .value(value)
            .defaults(MapUtils.emptyOnNull(this.cases).keySet().stream().noneMatch(value::equals))
            .build();
    }
```

Behaviour is identical, the null case still yields `defaults = true`, and the file uses one idiom throughout. Entirely optional — what's there now is correct.

## Before merge

- **CI has never run on this PR.** `gh pr checks` reports no checks at all and the status rollup is empty — that's the first-time-contributor gate, so a maintainer needs to approve the workflow run. Worth flagging because an empty rollup reads as "nothing failing" when it actually means the change is unverified.
- **Please rebase rather than merge.** The branch tip is a merge commit and contains a second one; our guidelines ask for a rebase on `develop`. The branch is well behind `develop`, but neither `Switch.java`, `SwitchTest.java` nor `switch.yaml` has changed there since the merge base, so the rebase should be textually free.
- Since `Switch` is a flowable resolved by the executor, please run `./gradlew :jdbc-h2:test --tests "H2RunnerTest"` alongside `:core:test`.

Nothing here blocks the change on its merits.

Fayupable:
Thanks for the detailed review!
Rebased onto develop (conflict-free, as you expected). Applied the outputs() suggestion as-is.
Test results:
```
./gradlew :core:test
3496 total · 3489 passed · 1 failed · 6 skipped in 6m 02s

The 1 failure is HttpClientTest.initializationError, unrelated to this
change — it needs a running Docker daemon (Testcontainers), which
wasn't available in my local environment.

./gradlew :jdbc-h2:test --tests "H2RunnerTest"
100 total · 100 passed · 0 failed · 0 skipped in 2m 08s
```

flcarre (approved):
Re-reviewed at `85260363`. Both points from my earlier comment are addressed, and the change is tighter than when I first read it:

- **Rebased** — the merge commits are gone and the branch is now 5 commits behind `develop` instead of 123. The cumulative diff is down to three files, `+5/-8` in `Switch.java`.
- **The idiom is unified** — `childTasks()` and `outputs()` both go through `MapUtils.emptyOnNull(this.cases)`, and `rendererValue(runContext)` is hoisted out of the `noneMatch` predicate, so the Pebble `value` is rendered once per call instead of once per case entry (it's `skipCache()`, so that repetition was real). `throwPredicate` is still used by `childTasks()`, so no import went stale, and the null path still yields `defaults = true`.

The fix remains correctly placed: `SwitchTaskValidator` rejects a `Switch` only when `cases` **and** `defaults` are both empty, so defaults-only is a deliberately valid configuration and guarding at runtime is right rather than tightening the validator. The two previously-unguarded `cases.entrySet()` dereferences are the only ones in the class — `allChildTasks()` and `tasksTree()` already null-guarded — so graph and topology building are untouched, and both `resolveState()` and `resolveNexts()` route through `childTasks()`.

CI on the previous head was green, including `Backend - Tests`; the only red job was `OpenTelemetry - Export Trace`, which fails at *Set up job* on a missing `kestra-io/actions/actions/otel-collect@main` and is unrelated to this change (it's currently reddening every PR that reaches it).

Approving. One thing before merge: I've just approved the workflow runs for this new head, so please wait for them to come back green — and since `Switch` is resolved by the executor, `./gradlew :jdbc-h2:test --tests "H2RunnerTest"` is worth running alongside `:core:test`.

Nice work, and thanks for picking up the refactor suggestion.

loicmathieu (approved):
LGTM, thanks

loicmathieu:
Backported to 2.0 and 1.3

PR #18378, opened 2026-08-21
fix(core): avoid duplicate value in INT/FLOAT parse error messages

### LINKED ISSUE #18289: Error message for invalid INT input/output redundantly repeats the same value twice
### Describe the issue

When a value fails to parse as INT (either as a flow input or a flow output), the resulting error message repeats the exact same string twice, in a format that reads as if it should be showing two different pieces of information (e.g. "attempted value" vs "actual received value") but isn't.

Example 1 (input): "Invalid input for `retry_count`, For input string: '3.5', but received '3.5'"
Example 2 (output): "Invalid output for `invalid`, For input string: 'this is not a number', but received 'this is not a number'"

This is minor but reduces the diagnostic value of the message and can look like a templating/substitution bug to users debugging a failed execution.

**Reproducer (input case)**

```yaml
id: bad_input_test
namespace: dev
inputs:
  - id: retry_count
    type: INT
    required: true
tasks:
  - id: show_value
    message: "Retry count is {{ inputs.retry_count }}"
```

Execute and enter `3.5` for retry_count.

Image

**Reproducer (output case)**

```yaml
id: bad_output_test
namespace: dev
tasks:
  - id: hello
    format: "this is not a number"
outputs:
  - id: invalid
    type: INT
    value: "{{ outputs.hello.value }}"
```

Execute this flow.

Image
Image

**Expected behavior**

The message should either show two genuinely distinct values, or drop the redundant clause entirely (e.g. "Invalid input for `retry_count`: could not parse '3.5' as INT").

**Actual behavior**

Both clauses show the identical string, adding no extra diagnostic information.

### Environment

riddhij-7:
### Related Issue

Fixes #18289

### Description

`NumberFormatException`'s own message already embeds the value being parsed (`For input string: "3.5"`), and the error-message builder appends the value again separately (`, but received '3.5'`), producing a duplicated value in the final error for INT/FLOAT inputs and outputs.

This wraps the INT and FLOAT parsing in `FlowInputOutput.parseType()` to throw a clean message instead of letting the JDK's self-describing exception text leak through.

Before:
Invalid input for `retry_count`, For input string: "3.5", but received "3.5"

After:
Invalid input for `retry_count`, not a valid `INT`, but received `3.5`

Kept the existing `Invalid X for Y, <reason>, but received <value>` template.
Verified both the input and output repro cases from the issue.

###  Backend Checklist

###  Additional Notes

A companion PR applies the same fix to releases/v1.3.x: #<https://github.com/kestra-io/kestra/pull/18380>

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
@@ -475,9 +475,27 @@ public class FlowInputOutput {
                     String encrypted = EncryptionService.encrypt(secretKey.get(), current.toString());
                     yield EncryptedString.from(encrypted);
                 }
-                case INT -> current instanceof Integer ? current : Integer.valueOf(current.toString());
+                case INT -> {
+                    if (current instanceof Integer) {
+                        yield current;
+                    }
+                    try {
+                        yield Integer.valueOf(current.toString());
+                    } catch (NumberFormatException e) {
+                        throw new IllegalArgumentException("not a valid `INT`");
+                    } 
+                }
                 // Assuming that after the render we must have a double/int, so we can safely use its toString representation
-                case FLOAT -> current instanceof Float ? current : Float.valueOf(current.toString());
+                case FLOAT -> {
+                    if (current instanceof Float) {
+                        yield current;
+                    }
+                    try {
+                        yield Float.valueOf(current.toString());
+                    } catch (NumberFormatException e) {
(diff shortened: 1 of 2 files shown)


loicmathieu:
PRs should always target `develop`

MilosPaunovic:
@loicmathieu The bug was reported on `v1.0.x`, why the PR targeting that branch is wrong if the bug does not exist anymore on `develop`?

loicmathieu:
Duplicated with https://github.com/kestra-io/kestra/pull/18380

PR #18380, opened 2026-08-21
fix(core): avoid duplicate value in INT/FLOAT parse error messages

### LINKED ISSUE #18289: Error message for invalid INT input/output redundantly repeats the same value twice

riddhij-7:
### Related Issue

Fixes #18289

### Description
Same root cause and fix as #<https://github.com/kestra-io/kestra/pull/18378> (releases/v1.0.x), applied here to releases/v1.3.x the unguarded Integer.valueOf()/Float.valueOf() calls in FlowInputOutput.parseType() let NumberFormatException's self-describing message duplicate the value already present in the outer error wrapper.

Before:
Invalid value for input `retry_count`. Cause: For input string: "3.5"

After:
Invalid value for input `retry_count`. Cause: not a valid `INT`

Verified both input and output repro cases and added a regression test covering the message.

### Backend Checklist

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
@@ -515,9 +515,27 @@ public class FlowInputOutput {
                     String encrypted = EncryptionService.encrypt(secretKey.get(), current.toString());
                     yield EncryptedString.from(encrypted);
                 }
-                case INT -> current instanceof Integer ? current : Integer.valueOf(current.toString());
+                case INT -> {
+                    if (current instanceof Integer) {
+                        yield current;
+                    } 
+                    try {
+                        yield Integer.valueOf(current.toString());
+                    } catch (NumberFormatException e) {
+                        throw new IllegalArgumentException("not a valid `INT`");
+                    }
+                }
                 // Assuming that after the render we must have a double/int, so we can safely use its toString representation
-                case FLOAT -> current instanceof Float ? current : Float.valueOf(current.toString());
+                case FLOAT -> {
+                    if (current instanceof Float) {
+                        yield current;
+                    }
+                    try {
+                        yield Float.valueOf(current.toString());
+                    } catch (NumberFormatException e) {
+                        throw new IllegalArgumentException("not a valid `FLOAT`");
+                    }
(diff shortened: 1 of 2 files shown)


loicmathieu:
To be sure I understand why did you target 1.3? If the issue exists in develop it should target the develop branch.

riddhij-7:
I targeted 1.3.x (and 1.0.x in #18378) because the bug doesn't actually exist on develop anymore; an unrelated refactor there already changed the message-building path, and I verified both repro cases no longer duplicate the value on develop. So there's no develop-side fix to open a PR for, the bug is only present on these older branches, which is why Milos asked me to check and fix them directly.

If backports like this should go through a different process on your end rather than a direct PR, happy to close this and defer to that, just let me know what works best. @loicmathieu

loicmathieu (approved):
LGTM, thanks

PR #18391, opened 2026-08-21
fix(flows): use pencil-lock icon for search-only source search markers

flcarre:
## Summary

- The "Search in" type pills (Namespace files, KV keys, Secret keys) and the results-list "Search only" tag both used `PencilOff` to mean "this type is search-only, can't be replaced" - at pill size (14px icon / 12px label) its diagonal slash reads as an "x" remove control, and since the one type that actually is removable from replace scope ("Flows") has no icon at all, the signal looked inverted.
- Replaced `PencilOff` with `PencilLockOutline` in both places (`FlowsSearch.vue` pills and `SourceSearchResults.vue`'s results tag), so the same "can't be replaced" concept uses one glyph everywhere it appears, including when both markers are visible together in replace mode.
- `Lock` was tried first, matching the per-flow-group read-only tag, but it collides with the Secrets pill's own `LockOutline` type-icon (double padlock) and with `Lock`'s separate, already-established meaning in `SourceSearchResults.vue`: an RBAC read-only marker for individual non-editable flows (`group.editable === false`) and secret-value chips. `PencilLockOutline` avoids both collisions while staying legible at pill size.
- No copy change: existing tooltips/labels are untouched.

Closes: https://github.com/kestra-io/kestra-ee/issues/10004

## Test plan

- [x] `cd ui && rtk npm run check:types` - clean
- [x] `npm run test:unit` - 1848/1848 tests passed (one unrelated `KsMarkdown`/shiki teardown-timing error, pre-existing, unaffected by this change)
- [x] `npm run test:lint` - 0 errors, only pre-existing warnings unrelated to the touched files
- [x] `npm run translations:check` - clean, no i18n changes needed

### DIFF
--- a/ui/src/components/flows/FlowsSearch.vue
+++ b/ui/src/components/flows/FlowsSearch.vue
@@ -115,7 +115,7 @@
                         <span v-if="query && crossResourceSearchStore.statusFor(type) !== 'idle'" class="source-search__pill-count">{{ crossResourceSearchStore.countFor(type) }}</span>
                         <Loading v-if="crossResourceSearchStore.statusFor(type) === 'counting'" class="source-search__pill-spin" />
                         <AlertCircleOutline v-else-if="query && crossResourceSearchStore.statusFor(type) === 'failed'" :title="crossResourceSearchStore.errorMessageFor(type)" />
-                        <Lock v-else-if="type !== 'flows'" />
+                        <PencilLockOutline v-else-if="type !== 'flows'" />
                     </KsCheckTag>
                     <KsButton class="source-search__pill-outline" size="small" @click="selectAllTypes">
                         {{ $t('source_search.select_all_types') }}
@@ -340,7 +340,7 @@
     import Magnify from "vue-material-design-icons/Magnify.vue"
     import InformationOutline from "vue-material-design-icons/InformationOutline.vue"
     import AlertCircleOutline from "vue-material-design-icons/AlertCircleOutline.vue"
-    import Lock from "vue-material-design-icons/Lock.vue"
+    import PencilLockOutline from "vue-material-design-icons/PencilLockOutline.vue"
     import Loading from "vue-material-design-icons/Loading.vue"
     import FileTreeOutline from "vue-material-design-icons/FileTreeOutline.vue"
     import FolderOpenOutline from "vue-material-design-icons/FolderOpenOutline.vue"
--- a/ui/src/components/flows/FlowsSearch.vue
+++ b/ui/src/components/flows/FlowsSearch.vue
@@ -115,7 +115,7 @@
                         <span v-if="query && crossResourceSearchStore.statusFor(type) !== 'idle'" class="source-search__pill-count">{{ crossResourceSearchStore.countFor(type) }}</span>
                         <Loading v-if="crossResourceSearchStore.statusFor(type) === 'counting'" class="source-search__pill-spin" />
                         <AlertCircleOutline v-else-if="query && crossResourceSearchStore.statusFor(type) === 'failed'" :title="crossResourceSearchStore.errorMessageFor(type)" />
-                        <PencilOff v-else-if="type !== 'flows'" />
+                        <Lock v-else-if="type !== 'flows'" />
                     </KsCheckTag>
                     <KsButton class="source-search__pill-outline" size="small" @click="selectAllTypes">
                         {{ $t('source_search.select_all_types') }}
@@ -340,7 +340,7 @@
     import Magnify from "vue-material-design-icons/Magnify.vue"
     import InformationOutline from "vue-material-design-icons/InformationOutline.vue"
     import AlertCircleOutline from "vue-material-design-icons/AlertCircleOutline.vue"
-    import PencilOff from "vue-material-design-icons/PencilOff.vue"
+    import Lock from "vue-material-design-icons/Lock.vue"
     import Loading from "vue-material-design-icons/Loading.vue"
     import FileTreeOutline from "vue-material-design-icons/FileTreeOutline.vue"
     import FolderOpenOutline from "vue-material-design-icons/FolderOpenOutline.vue"
--- a/ui/src/components/flows/SourceSearchResults.vue
+++ b/ui/src/components/flows/SourceSearchResults.vue
@@ -22,7 +22,7 @@
                 </KsTag>
                 <KsTag v-else-if="type !== 'flows'" size="small" round>
                     <template #icon>
-                        <PencilOff />
+                        <PencilLockOutline />
                     </template>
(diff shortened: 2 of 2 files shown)


flcarre:
## QA Report — PR #18391 (kestra-ee#10004)

**Verdict: PASS**

Tested against the real running app this time (`http://localhost:53101`, OSS backend on `:8180`) — this fix only involves rendering (no write/save operations), so it wasn't affected by the proxied-POST/CSRF quirk noted on the previous two PRs in this batch.

### Summary

| Scenario | Result |
|---|---|
| 1. Non-Flows pills show a clear lock icon instead of the ambiguous pencil-slash | ✅ SUCCESS |
| 2. Tooltip copy unchanged, still explains the read-only/no-replace restriction | ✅ SUCCESS |
| 3. Non-regression — Flows pill still has no icon; pill selection/counts still work | ✅ SUCCESS |
| 4. Dark mode | ✅ SUCCESS |

### Scenario 1 & 3: Pill icons (Flows → Source search)

Screenshot of the "Search in" row:

`Flows` (no icon) · `Namespace files 🔒` · `KV keys 🔒` · `Secret keys 🔒`

All three read-only types now show an unambiguous padlock at the trailing edge instead of the pencil-with-a-slash that read as a remove/×  control. `Flows` — the one type that's actually selectable/removable — correctly still shows no icon.

### Scenario 2: Tooltip copy

Read the `title` attribute directly off each pill (native tooltips don't screenshot reliably in headless automation):

```json
Namespace files: "Searches the file path only, not the file content. Cannot be replaced."
KV keys:         "Searches key names only. Cannot be replaced."
Secret keys:     "Searches key names only. Cannot be replaced."
```

Matches pre-fix copy exactly — only the glyph changed, as intended.

### Scenario 4: Dark mode

Re-screenshot after switching `localStorage.theme` to `dark` and reloading — same clean lock icons, good contrast, no regression.

flcarre:
## QA re-verification — PR #18391 (after review follow-up)

**Verdict: PASS** (supersedes my earlier PASS above, which validated the first icon attempt before it was found to have a double-lock defect on the Secret keys pill)

The final code review caught a real regression in the original `Lock` icon choice: the Secret keys pill already has `LockOutline` as its own leading type-icon, so adding `Lock` as the trailing "can't replace" marker put two padlocks on one small pill. The follow-up commit switches the trailing marker to `PencilLockOutline` instead. Re-verified live against the running app:

- **No double-lock**: Secret keys pill now shows one plain padlock (leading, unchanged) + one pencil-with-lock-badge (trailing marker) — visually distinct glyphs, not a repeated icon.
- **Still unambiguous**: `PencilLockOutline` doesn't read as a remove/× control (the original bug this issue reports).
- **Tooltips unchanged**: re-confirmed the `title` attributes on all three pills still read correctly ("Searches the file path only... Cannot be replaced." / "Searches key names only. Cannot be replaced.").
- **Dark mode**: re-checked, same result, good contrast.

### CI note

One check, `trigger-ee` (`kestra-ee/backend-tests`), is red. I independently pulled the raw GitHub Actions job log rather than taking this at face value: it's a genuine `kestra-ee` Java compile failure (`jdbc-ee` / `repository-elasticsearch-ee`, "method does not override or implement a method from a supertype", "cannot find symbol") — a pre-existing OSS/EE integration break unrelated to this frontend-only diff. All other checks are green.

MilosPaunovic (commented):
Verified the glyph swap at real pill size (14px icon, 12px label) rather than from the description - `PencilLockOutline` no longer reads as an `×`, so the reported problem on the pills is gone.

Three items still open:

**1. The same glyph is still on the results tag.** `SourceSearchResults.vue:25` renders `PencilOff` for `source_search.tag_search_only` - the same "this type cannot be replaced" meaning, on a `size="small"` round `KsTag`, in the same view (`SourceSearchResults` is rendered from `FlowsSearch.vue:279`). In replace mode both markers are visible at once, so one concept now has two glyphs and the ambiguous one survives at exactly the size where it misreads. Same one-line swap there would close the loop.

**2. The `PR` description no longer matches the code.** It says:

> Replaced `PencilOff` with `Lock`, the icon already used for the same read-only semantic on the per-flow-group tag in `SourceSearchResults.vue`

The shipped icon is `PencilLockOutline`, not `Lock`; and `Lock` in `SourceSearchResults.vue` marks a non-editable individual `flow` (`group.editable === false`, line 67) plus secret chips (lines 119, 293), which is a different semantic from the type-level restriction. Since the description becomes the squash body, it would land in history describing something the diff does not do.

**3. `CI` is red for a stale base.** `trigger-ee` / `kestra-ee/backend-tests` are failing while everything else is green. A frontend-only two-line icon swap cannot break `EE` `Java` compilation, and the comparable open `PR`s https://github.com/kestra-io/kestra/pull/18566, https://github.com/kestra-io/kestra/pull/18564, https://github.com/kestra-io/kestra/pull/18562 and https://github.com/kestra-io/kestra/pull/18561 all report `kestra-ee/backend-tests` green now, so the develop-side break has since been fixed. This branch is 29 commits behind `develop` - a rebase and re-run should clear it rather than merging with a red required check.

Two notes, neither blocking:

- **No backport needed.** `source_search` does not exist on either release branch - `ui/src/components/flows/FlowsSearch.vue` on `releases/v1.3.x` and `releases/v1.0.x` contains no `source_search` markup at all, so this is `develop`-only.
- **Nothing asserts the marker icon** and there is no `FlowsSearch` spec at all, so the safety net here is manual review. Not worth standing up a spec for a glyph, just flagging it.

flcarre:
Thanks for the thorough second pass. All three addressed:

1. Swapped the results-list "Search only" tag's icon (`SourceSearchResults.vue`) from `PencilOff` to `PencilLockOutline`, matching the pills — both markers now use the same glyph, including when they're visible together in replace mode.
2. Rewrote the PR description to match what actually shipped: `PencilLockOutline`, not `Lock`. Also added the "why not `Lock`" reasoning — it collides with the Secrets pill's own `LockOutline` icon (double padlock) and with `Lock`'s separate existing meaning in this file (RBAC read-only marker on individual flows/secret chips).
3. Confirmed — CI is fully green now including `trigger-ee`/`kestra-ee/backend-tests`, so that was indeed a transient develop-side break. Rebased onto latest `develop` anyway per your suggestion and re-pushed; still green.

Left the icon-assertion/spec question as-is per your own note — agreed it's not worth a dedicated spec for a glyph.

MilosPaunovic (approved):
Both items from the previous round are addressed - `PencilLockOutline` is now used in both `FlowsSearch.vue` and `SourceSearchResults.vue`, so the "search only" marker is one glyph everywhere, and the description matches what the diff ships. All checks green. `LGTM`.

PR #18465, opened 2026-08-21
fix(core): reject invalid BOOL input values

### LINKED ISSUE #18342: [Bash2.0][Inputs] BOOL input silently coerces any non-true string to false instead of rejecting it
### Describe the issue

A `BOOL` flow input silently coerces any non-`true` string to `false` at execution time, instead of rejecting invalid boolean values. Every other typed input validates strictly and returns 422 on bad input, so `BOOL` is the odd one out.

A user who passes `yes` or `1` expecting true gets `false` with no error, so the flow runs with the wrong value and nothing signals the mistake.

The docs are explicit that this should be rejected: [Inputs > Input types](https://kestra.io/docs/workflow-components/inputs) states **"BOOL: Must be `true` or `false` passed as strings."** So a value that is neither should fail validation, not be silently coerced.

Flow:

```yaml
id: typed_inputs
namespace: qa.inputs
inputs:
  - id: active
    type: BOOL
tasks:
  - id: echo
    message: "active={{ inputs.active }}"
```

Trigger with various values for `active` and read the resolved input on the execution:

```
active=true   -> 200, resolved True
active=false  -> 200, resolved False
active=maybe  -> 200, resolved False   (accepted, silently false)
active=yes    -> 200, resolved False   (accepted, silently false)
active=1      -> 200, resolved False   (accepted, silently false)
active=xyz    -> 200, resolved False   (accepted, silently false)
active=       -> 422  (missing required input)
```

Only an exact case-insensitive `true` resolves to true; everything else is accepted and coerced to false. Contrast with the other typed inputs, which all reject bad values:

```
INT=abc               -> 422 Invalid value for input `count`
DURATION=banana       -> 422 Invalid value for input `window`
SELECT=platinum       -> 422 (not in allowed values)
JSON={bad             -> 422
```

Expected: an invalid boolean value should be rejected with 422, the same as the other input types.

### Environment

Kestra Version: 2.0.0-rc10
Kestra Edition: reproduced on OSS (also on EE)

AradhyaSharma31:
### Related Issue

Closes https://github.com/kestra-io/kestra/issues/18342

### Description

Fixes BOOL input validation so that invalid string values are rejected instead of being silently coerced to `false`.

Previously, values such as `maybe`, `yes`, or `1` were accepted and resolved to `false`.

After this change:
- `true` → accepted
- `false` → accepted
- `maybe` → rejected with `422`
- `yes` → rejected with `422`
- `1` → rejected with `422`

This makes BOOL input validation consistent with the other typed input types.

### Additional Notes

The behavior was verified locally with valid and invalid BOOL input values.

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
@@ -593,7 +593,18 @@ public class FlowInputOutput {
             case STRING, EMAIL, SELECT -> current.toString();
             case INT -> TypeConverter.toInteger(current);
             case FLOAT -> TypeConverter.toFloat(current);
-            case BOOL -> TypeConverter.toBoolean(current);
+            case BOOL -> {
+                if (current instanceof Boolean b) {
+                    yield b;
+                }
+
+                if (!(current instanceof String s &&
+                    (s.equalsIgnoreCase("true") || s.equalsIgnoreCase("false")))) {
+                    throw new IllegalArgumentException("Unable to parse `" + current + "` as a boolean");
+                }
+
+                yield TypeConverter.toBoolean(current);
+            }
             case DATETIME -> TypeConverter.toInstant(current);
             case DATE -> TypeConverter.toLocalDate(current);
             case TIME -> TypeConverter.toLocalTime(current);
--- a/core/src/test/java/io/kestra/core/runners/FlowInputOutputTest.java
+++ b/core/src/test/java/io/kestra/core/runners/FlowInputOutputTest.java
@@ -24,6 +24,7 @@ import io.kestra.core.encryption.EncryptionService;
 import io.kestra.core.exceptions.InputOutputValidationException;
 import io.kestra.core.models.executions.Execution;
 import io.kestra.core.models.flows.*;
+import io.kestra.core.models.flows.input.BoolInput;
 import io.kestra.core.models.flows.input.EmailInput;
 import io.kestra.core.models.flows.input.FileInput;
 import io.kestra.core.models.flows.input.FloatInput;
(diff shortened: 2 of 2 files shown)


core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
```diff
@@ -599,7 +599,12 @@ private Object parseType(Execution execution, Type type, String id, Type element
                 case INT -> TypeConverter.toInteger(current);
                 // Assuming that after the render we must have a double/int, so we can safely use its toString representation
                 case FLOAT -> TypeConverter.toFloat(current);
-                case BOOL -> TypeConverter.toBoolean(current);
+                case BOOL -> {
```
loicmathieu: I think we should also support real boolean value, it's convenient at least for testing?.

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
AradhyaSharma31: Good point, I've updated the validation to support Boolean values as well. I've also added test coverage for both cases. Please take another look when you get a chance.

MilosPaunovic:
If you've addressed any of the requested changes from the PR review, @AradhyaSharma31, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

AradhyaSharma31:
> If you've addressed any of the requested changes from the PR review, @AradhyaSharma31, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

Thanks for letting me know, I’ve done that.

loicmathieu (approved):
LGTM, thanks a lot

loicmathieu:
Backported to 2.0

PR #18483, opened 2026-08-21
fix(flows): render a readable message for the quota duration constraint

### LINKED ISSUE #18480: [Bash2.0] Quota duration validation returns raw template code instead of a message
## Describe the issue

Setting a quota window shorter than one minute returns a validation error made of raw template code instead of a readable message. The user sees about 500 characters of expression syntax rather than "must be longer than or equal to 1 minute". It happens on both editions.

## Steps to reproduce

Create a flow with a quota window under a minute:

```yaml
id: quotabad
namespace: qa.quota
quotas:
  - duration: PT10S
    limit: 5
    behavior: FAIL
tasks:
  - id: a
    message: x
```

## Actual behaviour

The save fails with `422` and this message:

```
Invalid entity: quotas[0].duration: must be longer than$inclusive == true ? ' or equal to' : ''$days == 0 ? '' : days == 1 ? ' 1 day' : ' ' += days += ' days'$hours == 0 ? '' : hours == 1 ? ' 1 hour' : ' ' += hours += ' hours'$minutes == 0 ? '' : minutes == 1 ? ' 1 minute' : ' ' += minutes += ' minutes'$seconds == 0 ? '' : seconds == 1 ? ' 1 second' : ' ' += seconds += ' seconds'$millis == 0 ? '' : millis == 1 ? ' 1 milli' : ' ' += millis += ' millis'$nanos == 0 ? '' : nanos == 1 ? ' 1 nano' : ' ' += nanos += ' nanos'
```

The quota is correctly rejected, only the message is unusable. The neighbouring `limit: 0` check reads fine ("quotas[0].limit: must be greater than 0"), and the dashboard time-window constraints read fine too, because those set their own message text.

Image

## Expected behaviour

- Show a readable message, for example "quotas[0].duration: must be at least 1 minute".

## Environment

- Kestra Version: rc-10
- Edition: Enterprise

ManojTauro:
## Summary

`Quota.duration` used `@DurationMin(minutes = 1)` with no custom message, so it fell back to Hibernate Validator's default message for that constraint, which is built from EL expressions that don't get evaluated in Kestra's validator setup. This caused raw EL template text to show up instead of a readable message.

Added an explicit static `message`, referenced from the same pattern already used by `TimeWindow`'s duration constraints.

## Test plan

- [ ] Create a flow with `quotas[0].duration: PT10S` and confirm the error now reads "must be longer than or equal to 1 minute" instead of raw EL syntax.

Fixes #18480

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/quota/Quota.java
+++ b/core/src/main/java/io/kestra/core/models/flows/quota/Quota.java
@@ -21,7 +21,7 @@ import lombok.experimental.SuperBuilder;
 @EqualsAndHashCode
 public class Quota {
     @NotNull
-    @DurationMin(minutes = 1)
+    @DurationMin(minutes = 1, message = "must be longer than or equal to 1 minute")
     private Duration duration;
 
     @NotNull


loicmathieu (approved):
LGTM, thanks

Issue #18373, opened 2026-08-21
[Bash2.0] concurrency is all lower case on the Flow Editor page


wrussell1999:
### Describe the issue

Should have a capital C at the start: Concurrency

<img width="2628" height="1483" alt="Image" src="https://github.com/user-attachments/assets/26b9cc15-b896-45e5-8f37-5cd855fd85ab" />

### Environment

- Kestra Version: 2.0 EE RC10

SANTHOSHKUMAR-303:
i want to work on this issue can assign this issue for me

jjj-n:
I'd like to work on this. The `concurrency` property in `Flow.java` is missing a `@Schema` title, so the generated JSON schema falls back to the lowercase field name — I'll add the missing title.

MilosPaunovic:
Hi there, @jjj-n! 👋

Thanks for your interest in contributing and for opening this PR!

As a general open-source convention, it’s best to **first comment on the issue to express your interest** and wait until you're **assigned** before starting any work. This helps avoid overlapping efforts and keeps contributions coordinated.  

We really appreciate your enthusiasm and look forward to future contributions! 🙌

PR #18484, opened 2026-08-22
fix(core): add missing schema title for the flow concurrency property

jjj-n:
### What changes?

The `concurrency` field in `Flow.java` had no `@Schema` annotation, so the generated JSON schema fell back to the raw lowercase field name and the no-code Flow Editor displayed it as "concurrency" instead of "Concurrency".

This adds the missing `@Schema` title (and a short description), matching how other flow-level properties such as `retry` declare theirs.

### Why

related to #18373.

### How to test

Open the no-code Flow Editor for any flow and check that the concurrency property label now reads "Concurrency" with a capital C.


MilosPaunovic (approved):

MilosPaunovic:
Thanks for the contribution @jjj-n, it's much appreciated! 🚀

PR #18519, opened 2026-08-24
fix(assets): allow an output asset to be declared by its id alone

RemiBarthe:
closes https://github.com/kestra-io/kestra-ee/issues/10084

This pull request introduces improvements to asset deserialization and emission, particularly around handling assets without explicit types and ensuring robust processing when asset outputs fail to render. The changes enhance the reliability of asset lineage tracking and clarify the behavior when asset declarations are incomplete or erroneous.

**Asset Deserialization and Type Handling:**
- Assets declared without a type are now automatically deserialized as `External` assets, making asset referencing as permissive for `assets.outputs` as it already was for `assets.inputs`. The deserialization logic has been updated to use a new `defaultClass` mechanism, and an asset that already exists keeps its stored type, so a task can update the metadata of a typed asset without repeating its type.

**Namespace Handling:**
- The namespace of an existing asset is now retained rather than cleared when a declaration omits it, which previously left the asset with no namespace and therefore invisible to every user whose asset permission is scoped to namespaces. A declared namespace no longer overwrites the stored one either, so the flow-driven write matches what the REST update path already enforces.
- An asset emitted without a namespace now inherits the namespace of the flow emitting it, which also covers the `External` assets auto-created from `assets.inputs`.

**Asset Emission Robustness:**
- The worker now renders declared asset inputs and outputs independently, so if rendering the outputs fails, the declared inputs are still preserved and emitted. This prevents loss of input asset lineage information due to unrelated output rendering errors. An invalid declaration is still classified as `DECLARATION_INVALID` and still fails the task, unchanged. Logging and error handling have been improved for better traceability.
- The isolation is between inputs and outputs, not per asset: a single malformed entry still discards the rest of its own list, since the property is rendered as a whole.

**Testing and Verification:**
- New and updated tests verify the deserialization of assets declared without a type, the retention of the previous type and namespace, the namespace inherited from the flow, and the emission of asset inputs even when outputs cannot be rendered.

**Supporting Refactors and Documentation:**
- The deserializer infrastructure was refactored to introduce a `defaultClass` method, providing a cleaner and more extensible approach to default plugin type resolution.

https://github.com/kestra-io/kestra-ee/pull/10636


brian-mulier-p (approved):
LGTM

PR #18521, opened 2026-08-24
fix(core): render execution-updating task properties for each execution

### LINKED ISSUE #18518: SetVariables reuses the first execution's rendered value for all later executions of the same flow
### Describe the issue

`SetVariables` (and the other `ExecutionUpdatableTask`s) render their `Property` **once per flow revision per executor JVM**, and every subsequent execution of that flow reuses the first execution's rendered value — regardless of the inputs of the current execution.

The result is silently wrong data: a flow that computes a boolean from an input keeps the boolean of whichever execution happened to run first after the executor loaded the flow.

Hit in production on our release flow `docker-publish`, where `IS_LTS` / `IS_LATEST` stayed `true` for a version that does not match the LTS prefix.

### Reproducer

```yaml
id: debug-inputs-lts-and-latest
namespace: company.team

inputs:
  - id: VERSION_TO_RELEASE
    type: STRING
    defaults: '1.0.56'

variables:
  LTS_VERSION_PREFIX: "1.3."
  IS_LTS: false

tasks:
  - id: computeIsLTS
    type: io.kestra.plugin.core.execution.SetVariables
    variables:
      IS_LTS: "{{ inputs.VERSION_TO_RELEASE | startsWith(vars.LTS_VERSION_PREFIX) }}"

  - id: log
    message: "version {{ inputs.VERSION_TO_RELEASE }} — lts: {{ vars.IS_LTS }}"
```

1. Execute with `VERSION_TO_RELEASE=1.3.9` → logs `lts: true` ✅
2. Execute the **same flow** with `VERSION_TO_RELEASE=1.0.56` → logs `lts: true` ❌ (expected `false`)

The reverse order is just as sticky: first run with `1.0.56` → `false`, and every later run with `1.3.9` also yields `false`. Verified deterministic (3/3 runs each way) on `kestra/kestra:v2.0.0-rc8` with the Postgres backend, and the same code path is present on `develop`.

Two things confirm the cache is keyed on the flow object, not the execution:
- An **identical flow deployed under a different id** returns the correct value on its own first run, then gets stuck the same way.
- **Bumping the flow revision** resets it — then the first execution of the new revision poisons it again.

The `startsWith` filter is not at fault: a standalone `Log` task evaluating the very same expression is correct on every run. The stale value is already wrong in the execution's `variables` map written by `SetVariables`.

### Root cause

`Property` memoizes the rendered value on the `Property` instance itself — `core/src/main/java/io/kestra/core/models/property/Property.java`:

```java
public static <T> T as(Property<T> property, PropertyContext context, Class<T> clazz, Map<String, Object> variables) throws IllegalVariableEvaluationException {
    if (property.skipCache || property.value == null) {
        String rendered = context.render(property.expression, variables);
        property.value = deserialize(rendered, clazz);
    }
    return property.value;
}
```

For a regular worker task this is harmless — the task is deserialized fresh for each task run. But `SetVariables` is an `ExecutionUpdatableTask`, run inside the executor at `executor/src/main/java/io/kestra/executor/ExecutorService.java:1374`:

```java
if (!(workerTask.getTask() instanceof ExecutionUpdatableTask executionUpdatingTask)) {
    return false;
}
...
var newExecution = executionUpdatingTask.update(executor.getExecution(), executorTask.runContext());
```

`workerTask.getTask()` is the task instance owned by the executor's **cached flow**, shared by every execution of that flow revision — so the first render's value is pinned for the lifetime of that cache entry.

Same exposure applies to the other `ExecutionUpdatableTask`s: `Labels`, `Exit`, `UnsetVariables`.

There is precedent for the fix: commit `424a6cb43` ("fix(execution): skip the render cache in flowable for properties used to compute next tasks") applied `Property#skipCache()` for flowables. Executor-side updatable tasks were never covered.

### Suggested fix

Render with `skipCache()` in the `update()` of the `ExecutionUpdatableTask` implementations (or make `ExecutorService.handleExecutionUpdatingTasks` operate on a copy of the task rather than the cached flow's instance).

### Environment

- Kestra Version: v2.0.0-rc8 (code path unchanged on `develop`)
- Backend: Postgres
- Also observed on Kestra Cloud production

AcevedoR:
Closes #18518

### What

`SetVariables`, `UnsetVariables` and `Exit` are `ExecutionUpdatableTask`s: the executor runs them itself (`ExecutorService.handleExecutionUpdatingTasks`) by calling `update()` on `workerTask.getTask()`, which is the task instance of the executor's **cached flow**. That single instance is shared by every execution of the flow revision.

`Property` memoizes its rendered value on the instance:

```java
if (property.skipCache || property.value == null) {
    String rendered = context.render(property.expression, variables);
    property.value = deserialize(rendered, clazz);
}
return property.value;
```

So the value rendered by the *first* execution was silently reused by all the following ones, whatever their inputs. Reproduced on `v2.0.0-rc8` and hit in production: a release flow computing `IS_LTS` / `IS_LATEST` from its version input kept the booleans of the first execution — deploying an execution to the wrong docker tags.

```yaml
tasks:
  - id: computeIsLts
    type: io.kestra.plugin.core.execution.SetVariables
    variables:
      IS_LTS: "{{ inputs.VERSION | startsWith('1.3.') }}"
  - id: log
    message: "{{ inputs.VERSION }} — lts: {{ vars.IS_LTS }}"
```

Run it with `1.3.9` (→ `true`), then with `1.0.56`: it still logs `true`. Deploying the same flow under another id, or bumping its revision, resets the cache — then the first run of the new revision pins it again.

### How

Render those properties with `Property#skipCache()` inside `update()`, exactly like `424a6cb43` already did for the flowable properties used to compute the next tasks. `Labels` is not affected — it renders raw `Object` values through the run context, not through `Property`.

### Tests

Written test-first, both fail on `develop` and pass with the fix:

- `SetVariablesTest#shouldRenderVariablesForEachExecution` — one deserialized task instance, `update()` called twice with different execution contexts, asserts the second render reflects the second context.
- `UnsetVariablesTest#shouldRenderVariablesForEachExecution` — same for the list property.

`io.kestra.plugin.core.execution.*` and `io.kestra.core.models.property.*` are green (35 tests).

### Backport

Worth backporting — the bug is present in every version shipping `Property`-based `SetVariables`.

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/execution/Exit.java
+++ b/core/src/main/java/io/kestra/plugin/core/execution/Exit.java
@@ -114,7 +114,9 @@ public class Exit extends Task implements ExecutionUpdatableTask {
     }
 
     private State.Type executionState(RunContext runContext) throws IllegalVariableEvaluationException {
-        return switch (runContext.render(this.state).as(ExitState.class).orElseThrow()) {
+        // the executor reuses the task instance of its cached flow for every execution of the flow,
+        // so the rendering cache of the property must be skipped to render with the current execution context
+        return switch (runContext.render(this.state.skipCache()).as(ExitState.class).orElseThrow()) {
             case ExitState.SUCCESS -> State.Type.SUCCESS;
             case WARNING -> State.Type.WARNING;
             case KILLED -> State.Type.KILLED;
--- a/core/src/main/java/io/kestra/plugin/core/execution/SetVariables.java
+++ b/core/src/main/java/io/kestra/plugin/core/execution/SetVariables.java
@@ -66,8 +66,10 @@ public class SetVariables extends Task implements ExecutionUpdatableTask {
 
     @Override
     public Execution update(Execution execution, RunContext runContext) throws Exception {
-        Map<String, Object> renderedVars = runContext.render(this.variables).asMap(String.class, Object.class);
-        boolean renderedOverwrite = runContext.render(overwrite).as(Boolean.class).orElseThrow();
+        // the executor reuses the task instance of its cached flow for every execution of the flow,
+        // so the rendering cache of the properties must be skipped to render with the current execution context
+        Map<String, Object> renderedVars = runContext.render(this.variables.skipCache()).asMap(String.class, Object.class);
+        boolean renderedOverwrite = runContext.render(this.overwrite.skipCache()).as(Boolean.class).orElseThrow();
 
         Map<String, Object> currentVariables = execution.getVariables() == null ? Collections.emptyMap() : execution.getVariables();
 
--- a/core/src/main/java/io/kestra/plugin/core/execution/UnsetVariables.java
+++ b/core/src/main/java/io/kestra/plugin/core/execution/UnsetVariables.java
@@ -69,8 +69,10 @@ public class UnsetVariables extends Task implements ExecutionUpdatableTask {
 
     @Override
     public Execution update(Execution execution, RunContext runContext) throws Exception {
-        List<String> renderedVariables = runContext.render(variables).asList(String.class);
-        boolean renderedIgnoreMissing = runContext.render(ignoreMissing).as(Boolean.class).orElseThrow();
+        // the executor reuses the task instance of its cached flow for every execution of the flow,
+        // so the rendering cache of the properties must be skipped to render with the current execution context
+        List<String> renderedVariables = runContext.render(this.variables.skipCache()).asList(String.class);
+        boolean renderedIgnoreMissing = runContext.render(this.ignoreMissing.skipCache()).as(Boolean.class).orElseThrow();
         Map<String, Object> variables = execution.getVariables();
         for (String key : renderedVariables) {
             removeVar(variables, key, renderedIgnoreMissing);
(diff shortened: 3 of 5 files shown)


loicmathieu (approved):
LGTM

AcevedoR:
Alternative, broader fix opened as #18529: it scopes the `Property` render cache to the context it was rendered with, so every executor-side rendering is covered instead of each task having to call `skipCache()`. Same issue (#18518) — pick whichever you prefer, they are mutually exclusive.

AcevedoR:
Closing in favour of #18529, which fixes this at the root instead of per task.

Field evidence made the narrow approach untenable: four more tasks were reproduced with the same defect — `Exit.state` (execution ends SUCCESS despite `FAILED`), `Parallel.concurrent` (limit stuck at the first run's value, so `concurrent: 4` ran serially — 12s for four 3s sleeps), `Pause.pauseDuration` (asked `PT10S`, paused 2.6s) and `UnsetVariables.variables` (asked to drop `b`, dropped `a`).

This PR would have fixed only two of those four: `Parallel.concurrent` is rendered at `Parallel.java:171` and `Pause.pauseDuration` at `ExecutorService.java:994`, neither of which it touches — so it would have closed the reported case and left the neighbours live, which is what the previous rounds of `skipCache()` patches did (see #18518 for the history: 8 fixes in 12 months, same root cause each time).

#18529 scopes the render cache to the context a value was rendered with, which covers all of these at once. The regression tests from this PR are carried over there.

PR #18528, opened 2026-08-24
fix(executions): respect the Timezone setting in the Gantt scale and Run timeline

### LINKED ISSUE #18453: [Bash2.0] Gantt tab ignores the Timezone setting and shows an ambiguous 12-hour clock
## Describe the issue

The time scale on an execution's **Gantt** tab ignores the Timezone you picked in Settings. It always uses the timezone of the machine running the browser, so the times on the Gantt do not match the times shown everywhere else in Kestra (executions list, logs), which do respect the setting. Anyone who sets a timezone different from their laptop's (a common case for teams working across regions, or anyone who prefers UTC) reads the wrong clock time for when tasks ran. The times are also printed on a 12-hour clock with no AM/PM, so `4:41:19` could mean morning or afternoon.

## Steps to reproduce

1. Go to **Settings** and set **Timezone** to `America/Los_Angeles` (any zone that is not your machine's).
2. Run any flow with a couple of tasks, then open the execution and its **Gantt** tab.
3. Compare the times on the Gantt scale with the same execution's start time on the Executions list or Logs tab.

## Actual behaviour

The Gantt scale shows the machine's local time, not the selected timezone. For one execution that started at `16:41:19` UTC, with the Kestra setting fixed to `America/Los_Angeles` (so `9:41:19` is correct), the Gantt showed a different time for each machine timezone:

| Machine timezone | Gantt scale shows | Correct (per setting) |
|---|---|---|
| `Asia/Kolkata` | `10:11:19` | `9:41:19` |
| `UTC` | `4:41:19` | `9:41:19` |
| `America/Los_Angeles` | `9:41:19` | `9:41:19` (right only by luck) |

The Logs tab for the same execution shows the correct `9:41` in every case, so the Gantt is the odd one out.

## Expected behaviour

- The Gantt time scale should use the timezone from Settings, matching the executions list and the logs.
- The times should not be ambiguous: either use a 24-hour clock or include AM/PM.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise

### LINKED ISSUE #18456: [Bash2.0] Execution Run timeline ignores the Timezone setting
## Describe the issue

The **Run timeline** on an execution's Overview tab ignores the Timezone set in Settings and uses the browser machine's timezone instead. The executions list and logs respect the setting, so the timeline disagrees with them for the same execution.

## Steps to reproduce

1. In **Settings**, set **Timezone** to `America/Los_Angeles` (any zone that is not your machine's).
2. Run any flow, open the execution, stay on the **Overview** tab.
3. Click **See timeline** and read the timestamps.
4. Compare with the same execution's start time in the Executions list.

## Actual behaviour

The timestamps follow the machine's timezone. For an execution started at `17:07:16` UTC, with the setting on `America/Los_Angeles` (correct: `10:07:16`):

| Machine timezone | Timeline shows |
|---|---|
| `Asia/Kolkata` | `2026-08-21 22:37:16.803` |
| `UTC` | `2026-08-21 17:07:16.803` |

The Executions list shows `10:07:16` correctly in both cases.

## Expected behaviour

- The Run timeline should use the Timezone from Settings, matching the executions list and logs.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise

flcarre:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18453.
Closes https://github.com/kestra-io/kestra/issues/18456.

### ✨ Description

The Gantt time scale and the execution Overview **Run timeline** formatted their timestamps with a bare `moment(...)`, so both followed the browser machine's timezone. The executions list and the logs format through `utils/filters#date`, which applies the **Timezone** from Settings — so the same execution showed a different clock time depending on which surface you looked at.

For an execution started at `16:41:19` UTC with the setting on `America/Los_Angeles` (`9:41:19` is correct), the Gantt showed `10:11:19` on a machine set to `Asia/Kolkata` and `4:41:19` on one set to `UTC`, while Logs showed `9:41` in both.

Both surfaces now go through `utils/filters#date`, so the stored timezone and date format apply the way they do everywhere else.

The Gantt ticks also move from `h:mm:ss` to an explicit 24-hour `HH:mm:ss`. The scale has no room for AM/PM, so a 12-hour clock left `4:41:19` genuinely ambiguous between morning and afternoon.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit` — 1835 passed)

/demo-video

### 📝 Additional Notes

`moment` had no remaining use in either file, so the import is replaced rather than kept alongside.

Deliberately not done: the two surfaces are the ones the issues name. I did not sweep the rest of the app for other bare `moment().format(...)` calls — that is worth a follow-up pass, but widening this PR would make it much harder to review.

Two pre-existing unit failures (`KsIconButton > uses router-link tag when to prop is provided` and `KsEmptyState > renders Learn more as a button when learnMore is set`) reproduce on a clean `develop` checkout and are unrelated to this change.

### 🤖 AI Authors

I asked the cat why the Gantt chart was in the wrong timezone. He said he'd already moved to Pacific Standard Nap Time and the rest of us should catch up. 🐱

---


ui/src/components/executions/Gantt.vue
flcarre, line 1: **Known limitation worth a reviewer's call: changing the Timezone setting does not refresh an open Gantt.**

`utils/filters#date` reads the timezone from `localStorage` on every call, and `localStorage` is not reactive. `dates` is filled by `computeDates()`, which only runs when its watcher fires on execution/task-run changes — so if you change the Timezone in Settings while a Gantt is open, the ticks keep the old zone until something else re-triggers the computation.

`startTime` / `endTime` are `computed`, so they have the same issue: nothing in their dependency graph changes when the setting does.

This is pre-existing behaviour across the whole app rather than something this PR introduces — every `dateFilter` caller reads the same non-reactive key, and the settings page has always relied on a reload. I did not want to introduce a reactive timezone store in a bug-fix PR touching two files. Flagging it so it is a conscious decision rather than an oversight; happy to open a follow-up for a reactive `useTimezone()` if you think it is worth it.

---

ui/src/components/executions/overview/components/RunTimeline.vue
flcarre, line 1: **Why `"iso"` rather than the literal format string.**

`dateFilter(date, "iso")` is not a generic ISO-8601 formatter — `utils/filters#date` special-cases the string `"iso"` to mean exactly `YYYY-MM-DD HH:mm:ss.SSS`, which is byte-for-byte the format this component used before. So the rendered output is unchanged apart from the timezone now being applied.

I used the sentinel rather than passing the literal so this stays tied to whatever the shared helper considers the ISO-ish display format, instead of drifting from it. The trade-off is that `"iso"` reads like a magic string at the call site; the alternative reads like a duplicated format constant. Say the word if you prefer the explicit literal here.

---

flcarre (commented):
## Self-review

Re-read the diff adversarially. It is 9 lines across two files and I found no correctness defect, but three things a reviewer should decide on rather than take on trust.

**What I verified**
- `moment` had no other use in either file, so replacing the import leaves nothing dangling. `npm run check:types` and `npm run build` both pass.
- `dateFilter(endDate.toISOString(), …)` / `dateFilter(new Date(timestamp).toISOString(), …)`: the helper takes a string, and `Date` values had to be serialised. `toISOString()` emits UTC with an explicit `Z`, so `moment` parses it unambiguously and the subsequent `.tz(...)` conversion is correct. Passing a `Date` straight through would also have worked but would have relied on an undeclared overload.
- `computeDates()` allocates one `Date` per tick and the helper does one `localStorage` read per call — 5 ticks per recompute, so not worth optimising.

**Two inline notes** on the non-reactive timezone read and on the `"iso"` sentinel.

**Third point, on scope:** I deliberately did not sweep the rest of the app for other bare `moment().format(...)` calls that have the same defect. The two issues name these two surfaces; widening the PR would have made it much harder to review and would have mixed verified fixes with unverified ones. If a broader audit is wanted, it should be its own PR.

**Not verified:** I could not run the app or see the linked screenshots — the sandbox proxy blocks GitHub `user-attachments` (404 from the API). So the reasoning above is from the code and the issue text, and the `/demo-video` output is the first actual visual confirmation. Please sanity-check the Gantt scale renders sensible 24-hour ticks before merging.

---

MilosPaunovic (changes requested):
Thanks Florian - the direction is right and the `RunTimeline` half is clean, but the `Gantt` half crashes on a state the component already knows about, and that is what is turning `Frontend - Tests` red.

**Blocker: `computeDates()` throws `RangeError: Invalid time value`**

```js
const formatDate = (timestamp: number): string => dateFilter(new Date(timestamp).toISOString(), TICK_FORMAT)
```

For a terminal `execution` with an empty `taskRunList`, `stop()` is `Math.max(...[])` → `-Infinity`, so `delta()` is non-finite and the tick timestamps come out `NaN` / `-Infinity`. `new Date(NaN).toISOString()` **throws**, whereas the old `moment(NaN).format("h:mm:ss")` returned the harmless string `"Invalid date"`. That is the whole regression.

`CI` reproduces it - `Gantt.stories.tsx` already covers exactly this state and 3 of its 6 tests fail:

```
× Cancelled   × Failed   × Queued
RangeError: Invalid time value
 ❯ formatDate   src/components/executions/Gantt.vue:550:89
 ❯ computeDates src/components/executions/Gantt.vue:555:26
 ❯ compute      src/components/executions/Gantt.vue:468:8
 ❯ watch.immediate src/components/executions/Gantt.vue:605:16
```

The state is not synthetic - `emptyStateHint` in the same file documents it ("cancelled or failed by a concurrency limit before any task started"), and `hasValidDate` (`isFinite(delta())`) exists precisely to hide the header for it. `compute()` just never consults it. There is no app-level `errorHandler` in `ui/src`, so at runtime this is an uncaught exception on every `execution` update for those `executions` rather than a visible break, but it should not ship either way.

Minimal fix in `computeDates()`:

```js
if (!hasValidDate.value) {
    dates.value = []
    return
}
```

That restores `develop`'s effective behaviour, where the ticks were `"Invalid date"` strings the `v-if="hasValidDate"` header never rendered.

**Same root cause, currently latent: `endTime`**

`dateFilter(endDate.toISOString(), TICK_FORMAT)` with `endDate = new Date(stop())` → `new Date(-Infinity).toISOString()` throws for the same `executions`. It only survives today because the `computed` is lazy and the header sits behind `v-if="hasValidDate"` - one template change away from being the same crash. Worth an early `return ""` while you are in there.

**Nit: drop the `number` → `Date` → `string` roundtrip**

`date()` in `utils/filters` is typed `dateString: string`, but `moment` accepts `string | number | Date`. Widening the parameter lets both `Gantt` call sites become `dateFilter(timestamp, TICK_FORMAT)`, removes the `Date` and `toISOString()` allocations from a function that re-runs on a 40 ms interval for running `executions`, and makes the invalid-value case degrade to `"Invalid date"` instead of throwing. That fixes the class of bug rather than this one instance, so it is my preferred shape of the two.

**Nit: nothing locks the fix in**

The stories caught the crash, but no test asserts the actual behaviour you are fixing - that a tick label follows `TIMEZONE_STORAGE_KEY` instead of the machine zone. One assertion with that key set would stop a future bare `moment()` from quietly reintroducing this.

**Checklist**

"Unit tests pass (`npm run test:unit` - 1835 passed)" is accurate but incomplete: the `storybook` component tests are a separate step in the same job and are red from this change. The two `KsIconButton` / `KsEmptyState` failures you flagged as pre-existing are unrelated to these three.

**What is correct**

- `dateFilter(date, "iso")` in `RunTimeline` is right - `filters#date` special-cases `"iso"` to exactly `YYYY-MM-DD HH:mm:ss.SSS`, byte-identical to the old literal, so only the timezone changes. Good call documenting the sentinel inline.
- Your non-reactive-timezone caveat checks out: `onTimezone` in `BasicSettings.vue` persists without reloading, unlike `onLang` which does `document.location.assign(...)`. Genuinely pre-existing, fine out of scope - a follow-up issue for a reactive `useTimezone()` would be welcome.
- `HH:mm:ss` over `h:mm:ss` is well justified, the axis has no room for AM/PM.

**Backport**

The bug reproduces on both older lines, so this wants backports once `develop` is green:

- `releases/v1.3.x`: `Gantt.vue` has `moment(timestamp).format("h:mm:ss")`, and the `RunTimeline` predecessor `overview/components/sidebar/Timeline.vue` has `moment(date).format("YYYY-MM-DD HH:mm:ss.SSS")` - both halves apply.
- `releases/v1.0.x`: `Gantt.vue` has `this.$moment(ts).format("h:mm:ss")` - `Gantt` only, no `Timeline` component there.

MilosPaunovic (approved):
Nothing left open from my side - the whole check set is green and the backport call is recorded. Approving.

The two inline notes are yours to keep or drop; neither needs a change. Happy to review the follow-ups if you raise them: the reactive `useTimezone()` store, and the sweep for the remaining bare `moment().format(...)` callers.

flcarre:
Backports raised, both branches as you called it:

- `releases/v1.3.x` — Gantt scale + `overview/components/sidebar/Timeline.vue`: #18719
- `releases/v1.0.x` — Gantt only: #18724

Neither cherry-picks. Two notes from the 1.0 one: `computeDates` runs from a watcher, where `getCurrentInstance()` is null, so `filters#date` had to stop resolving `moment` through the app instance or the fix would have thrown on every execution update. And I did **not** backport the `hasValidDate` early return — 1.0 has no such flag and renders the ticks unconditionally, so the no-task-run execution shows five "Invalid date" labels before and after, unchanged. Happy to add the guard if you'd rather have it there too.

---

PR #18529, opened 2026-08-24
fix(core): only reuse a rendered property for the context it was rendered with

### LINKED ISSUE #18518: SetVariables reuses the first execution's rendered value for all later executions of the same flow

AcevedoR:
Closes #18518 — **alternative to #18521**, pick one. Same bug, fixed at the root instead of per task.

### The bug

`Property` memoizes its rendered value on the instance:

```java
if (property.skipCache || property.value == null) {
    String rendered = context.render(property.expression, variables);
    property.value = deserialize(rendered, clazz);
}
return property.value;
```

That is fine while a `Property` instance belongs to a single rendering, which is the case for a worker task — it is deserialized for each task run. It is wrong as soon as the instance outlives it, and the executor does exactly that: it renders the tasks of the flow **it keeps in its cache**, so one instance is shared by every execution of the flow revision and the first rendering is pinned for all the following ones.

Reproduced on `v2.0.0-rc8` and hit in production: a release flow deriving `IS_LTS` / `IS_LATEST` from its version input kept the booleans of the first execution — run it with `1.3.9` (→ `true`), then with `1.0.56`, and it still says `true`.

### Why fix it here rather than in the tasks

#18521 adds `skipCache()` to `SetVariables`, `UnsetVariables` and `Exit`. It is small and backport-friendly, but it is the second time we patch a symptom of the same root cause — `424a6cb43` already did it for the flowable properties used to compute the next tasks — and it stays opt-in: the next `ExecutionUpdatableTask`, or the next executor-side rendering, has to remember the incantation or silently serves stale values again.

This PR makes the cache correct by construction instead: **a rendered value is only reused for the context it was rendered with.**

```java
private boolean shouldRender(PropertyContext context) {
    if (this.literal) {
        return false;
    }

    if (this.skipCache || this.value == null) {
        return true;
    }

    return this.renderedWith == null || this.renderedWith.get() != context;
}
```

The context is held through a `WeakReference` and compared by identity, so nothing is retained and a cache entry can never be attributed to the wrong rendering. The rendering stays cached for the repeated renders it exists for — a plugin calling `runContext.render(prop)` several times uses the same `RunContext`, so it still renders once — and a mismatch degrades to rendering again, never to a wrong value.

`ofValue()` keeps its documented contract (the value is returned as-is, never rendered) through an explicit `literal` flag rather than by relying on `value != null`, which the new check would otherwise have confused with a cached rendering.

### Scope, deliberately kept small

- The additional-variables case is **left as it is**: `as(clazz, variables)` still reuses the rendering when only the variables differ, as `RunContextPropertyTest#asShouldReturnCachedRenderedProperty` explicitly asserts, with `skipCache()` as the documented escape hatch. It is arguably the same class of staleness, but changing it is a behavior change, not a bug fix — happy to do it in a follow-up if you want it.
- The existing `skipCache()` call sites keep working and become redundant; they can be cleaned up later.

### Tests

Written test-first, all failing on `develop` and passing with the fix:

- `PropertyTest#shouldRenderAgainWhenTheRenderContextChanges`
- `PropertyTest#shouldRenderListAndMapAgainWhenTheRenderContextChanges` — the `asList` / `asMap` paths cache too
- `SetVariablesTest#shouldRenderVariablesForEachExecution` — the end-to-end symptom, on the task instance reuse the executor does

Plus two tests locking the behavior that must not change: `shouldCacheTheRenderingOfTheSameContext` (still one rendering per context) and `shouldNeverRenderAValueSetAtBuildTime`.

`io.kestra.core.runners.*`, `io.kestra.plugin.core.flow.*`, `io.kestra.core.models.property.*` and `io.kestra.plugin.core.execution.*` are green (905 tests), and `H2RunnerTest` passes as required for an executor-affecting change.


AcevedoR:
Review pass found three defects in the first commit; all three are fixed in `0ecfdaf112`, each with a test that fails without it.

**HIGH — the render methods returned the shared field, not the value they rendered.** `DefaultExecutor` processes executions on `maxCachedThreadPool(numberOfThreads)`, so N executions of the same flow revision render the *same* `Property` instance concurrently — the very sharing this PR is about. Interleaved, execution A could receive B's value, and the instance could be left claiming A's context had rendered B's value, so a later render on that same context served it again. The rendered value and the context it was made for are now published together as one immutable `volatile` field, and each call returns the value it rendered rather than re-reading shared state. Covered by `PropertyTest#shouldGiveEachConcurrentRenderItsOwnValue` (500 renders over 8 threads, red before the fix).

**MEDIUM — the Lombok builder path lost its value.** `Property.<T>builder().value(v).build()` yielded `literal=false, expression=null`, so it took the render branch: `as()` rendered a null expression and overwrote the value with null, and `asList`/`asMap` NPE'd on `expression.trim()`. That path is reachable from EE and out-of-tree plugins. A value is now recognized as set at build time by the absence of any rendering, which covers `ofValue()` and the builder alike — the `literal` flag is gone rather than patched. Covered by `shouldNeverRenderAValueSetThroughTheBuilder`.

**LOW — `skipCache()` discarded the build-time value**, so `ofValue(...)` read through `RunContextProperty#skipCache` rendered its serialized form and could throw on an unknown variable. It now returns itself when there is no expression evaluation to skip. Covered by an added assertion in `shouldNeverRenderAValueSetAtBuildTime`.

Also noted in review and worth recording: `PropertyContext.create()` call sites (`FileInputValidator`, `RunContextFactory`, `FlowInputOutput.resolveDefaultValue`) now get a fresh identity per call, turning previously-pinned input-default renders into per-request renders — a fix, at negligible cost.

Verification: `io.kestra.core.models.property.*`, `io.kestra.core.runners.*`, `io.kestra.plugin.core.flow.*` and `io.kestra.plugin.core.execution.*` are green, and `H2RunnerTest` passes 100/100.

`Labels` is the one `ExecutionUpdatableTask` without a case in `SharedTaskInstanceRenderTest` — it renders raw `Object` values through the run context rather than through `Property`, so it is not exposed to this bug; happy to add a case anyway if you want the symmetry.

AcevedoR:
Added `PropertyRenderCacheBehaviourTest` (`be4c6c72a6`) to make the discussion about this cache concrete. Every behaviour is written twice — as it runs by default, and with `skipCache()`, which is how a property behaves when its rendering is not reused — so the pairs read side by side:

| Behaviour | Cached (default) | Not cached |
|---|---|---|
| A non-deterministic expression (`{{ uuid() }}`) rendered twice | one value | two values |
| Number of evaluations of one expression | one per context | one per call |
| Runtime validation of the rendered value (`@Min(0)`) | applied | silently skipped |
| Two executions rendering the same task instance | a value each — the cache is per context | a value each |

Only the last row is purely a cache. The first three are behaviours plugins depend on:

- **Stability** — a task rendering one property to name a file and again to read it back gets one name. Without the cache it gets two, and the same applies to `randomPort()`, `encrypt()` and `http()`, which have side effects per evaluation.
- **Cost** — measured with a counting `PropertyContext` rather than asserted: 1 evaluation vs 3 for three calls.
- **Validation** — `shouldNotValidateTheRenderedValueWhenNotCached` passes with `-2` for a `@Min(0)` property, showing the constraint never fires. This is pre-existing (`RunContextProperty.getProperty()` renders into a copy, so the extractor reads an empty value on the original), and it means `Switch`, `LoopUntil` and `FlowInputOutput` defaults currently get no runtime property validation. Worth a separate issue.

The intent is documentation: anyone proposing to remove the memoization can run this class and see exactly which behaviours they need a plan for first — and the last row shows the cross-execution staleness this PR fixes is the part that really was just a cache.

PR #18530, opened 2026-08-24
fix(core): stop silently handing out truncated log exports and empty chart exports

### LINKED ISSUE #18334: [Bash2.0][Dashboards] Chart Export is offered on an empty chart and downloads a 0 byte file
### Describe the issue

On a dashboard where a chart has no data, the card still offers **Export** on hover. Clicking it downloads a 0 byte file, not even a CSV header row.

The same endpoint is fine when the chart has data, so this is specific to the empty case:

- chart with no rows: `HTTP 200`, **0 bytes**
- same chart with rows: `HTTP 200`, 10629 bytes, starting with `namespace,id,state,flow,start_date`

So a user on an empty dashboard gets an unusable file with no indication anything was wrong. Either the export should be disabled while the chart has nothing to export, or it should still emit the header row so the download is at least a valid CSV.

Image

https://jam.dev/c/df631dae-2088-456b-aeac-061cedfc0136

1. Open **Dashboards** on an instance with no executions in the selected interval.
2. Every card shows "There are no results to be shown."
3. Hover the **Total Executions** card (or **Executions**) and click **Export**.

A 0 byte file downloads.

The same through the API, using the built-in chart definition from `GET /dashboards/defaults/definitions`:

```
POST /api/v1/{tenant}/dashboards/charts/export?format=CSV
body: {"chart": "<total_executions_timeseries definition>"}

no data      -> HTTP 200, 0 bytes
with data    -> HTTP 200, 10629 bytes, "namespace,id,state,flow,start_date" + rows
```

### Environment

Kestra Version: 2.0.0-rc10
Kestra Edition: reproduced on EE, and the endpoint behaves the same way on OSS

### LINKED ISSUE #18479: [Bash2.0] Download logs silently writes only the first 1000 lines
## Describe the issue

On the Logs page, **Download logs** silently writes only the first 1000 lines. If your filtered view has more, the rest are dropped with no warning and nothing in the file to say it is incomplete, so you can end up investigating an incident with a log file that quietly stops. The dialog only says "Filters default to your current view", which suggests you get everything the filters match.

## Steps to reproduce

1. Run a flow that produces more than 1000 log lines.
2. Open **Logs**, filter to that namespace, and note the total at the bottom of the table.
3. Click **Download logs** and confirm.
4. Count the lines in the downloaded file.

## Actual behaviour

With a filter matching 1205 lines:

- The page shows **Total 1205**.
- The downloaded file contains exactly **1000** lines, so 205 are missing.
- The download requests `logs/search?page=1&size=1000&...`: the size is fixed at 1000 no matter how many lines match.
- Nothing warns the user, and the file gives no hint that it was cut short.

## Expected behaviour

- Download everything the filters match, or tell the user the export was limited and how many lines were skipped.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise

### LINKED ISSUE #18545: [BATCH 2.0][UI] A failed request during a log download leaves the user with no file and no message
## Describe the issue

**Download logs** has no error handling. If the request fails, the dialog's spinner clears, no file is written and nothing tells the user why — the rejection is unhandled.

`ui/src/components/logs/LogsWrapper.vue`, in `downloadLogs()`:

```ts
downloading.value = true
logsStore.downloadLogs(params)
    .then((result) => { ... })
    .finally(() => (downloading.value = false))
```

No `.catch(...)`. `finally` resets the button, so the UI looks like the download simply did nothing.

## Why it matters more now

This is pre-existing, but the odds of hitting it went up sharply with #18479 ([#18530](https://github.com/kestra-io/kestra/pull/18530)): the export used to issue a single request, and now pages through the whole result set. A 40 000-line export makes ~40 requests, and any one of them failing loses the download.

## Steps to reproduce

1. Open **Logs** and filter to something with a few thousand matching lines.
2. Click **Download logs** → confirm.
3. Kill the network (or make the backend return a 5xx) partway through the paging.

## Actual behaviour

The spinner stops, no file downloads, no toast. Nothing in the UI distinguishes this from "there was nothing to export".

## Expected behaviour

Tell the user the export failed. A partial file is deliberately *not* the answer here — silently handing over an incomplete log file is the bug #18479 was about — so an error message with no download is the right outcome.

## Notes

Raised in the review of [#18530](https://github.com/kestra-io/kestra/pull/18530) and left out of it: choosing between "no file + error" and "partial file + warning" is a product call, and that PR was already changing what the download returns.

## Environment

- Kestra Version: `develop`
- Edition: OSS

flcarre:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18479.
Closes https://github.com/kestra-io/kestra/issues/18334.
Closes https://github.com/kestra-io/kestra/issues/18545.

### ✨ Description

Two exports produced a file that looked complete but was not, with nothing telling the user. Grouped because they are the same class of bug and the same fix shape: detect the incomplete case and say so instead of writing the file quietly.

**Download logs was capped at the first 1000 lines (#18479)**

`downloadLogs` issued a single `logs/search?page=1&size=1000` request and wrote whatever came back, so a filter matching 1205 lines produced a file with exactly 1000 — 205 dropped, no warning, and nothing in the file to say it stopped early. The dialog's "Filters default to your current view" suggests you get everything the filters match.

The store now pages through the whole result set, following `nextCursor` in cursor mode and incrementing `page` otherwise. A 50 000-line ceiling still guards the tab, since the file is assembled in memory — but hitting it is now surfaced as a warning naming how many lines were skipped, rather than passing silently.

**How each paging mode knows it has reached the end**

The two modes do not share an end-of-data signal, and conflating them re-introduced the bug this PR is about. Offset paging ends on a page shorter than `size`. Cursor paging does not: a cursor page may come back short while more remain, and the backend marks the end with an *empty* page — `nextCursor` is documented as present on every non-empty page. An earlier revision of this PR broke out of the loop on any short page, so an export against a cursor-paginated log store stopped early and reported itself `complete`. The short-page break is now offset-only, and two tests pin the difference.

**A refused page keeps what was already collected (#18545)**

Deep offset paging can be refused outright rather than returning a short page — Elasticsearch caps `from + size` at `index.max_result_window` (10 000 by default), so page 11 fails. The pages already fetched are kept and the outcome is reported as incomplete, with the line count, instead of the whole export being discarded.

This is the one place where the implementation and #18545 disagreed. That issue asks for "an error message with no download", on the grounds that handing over an incomplete log file silently is the bug #18479 was about. The deciding word is *silently*: the export is no longer silent, so a partial file plus an explicit "some matching lines are missing" warning gives the user more than an error alone, and the pages already paid for are not thrown away. #18545 is closed by this PR on that basis — if a reviewer prefers no file at all, it is a two-line change in `LogsWrapper.downloadLogs`.

**Chart export on an empty chart downloaded a 0 byte file (#18334)**

A chart with no rows answers its export with `HTTP 200` and an empty body, which was written out as a 0 byte file — not even a valid CSV with a header row. `downloadHandler` now reports the empty body instead of downloading it, and the card tells the user the chart has nothing to export.

The issue offered two options: disable Export while the chart is empty, or emit the header row. Emitting the header is a backend change, and disabling the button needs each chart component to report its row count up to `Sections.vue`. Reporting at download time fixes the user-visible symptom — an unusable file with no indication anything was wrong — without either of those, so it is what this PR does.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit`; `logsDownload.spec.ts` — 11 passed, the two cursor-paging cases failing before the fix)
- [x] Translations are complete if `en.json` changed (`npm run translations:check` reports no missing, extra or stale keys)

/demo-video

### 📝 Additional Notes

**Six new keys, not two** — an earlier version of this description said two, and was not updated as the outcomes were split apart: `logs_download_truncated`, `logs_download_capped`, `logs_download_partial`, `logs_download_failed`, `logs_download_empty` and `dashboards.exportEmpty`. Each names a distinct outcome the caller can tell apart (a known skipped count, the ceiling with no total to quantify it, a refused page, no file at all, nothing matched).

They are translated by hand across all twelve locales with matching `fingerprints.json` entries, since `npm run translations:generate` needs a `GEMINI_API_KEY` I do not have — so they are worth a reviewer's eye, and ideally a regeneration by someone who has the key. `translations:check` verifies the placeholders are identical to the English source, but not the register: the German strings first shipped here with "Ihren Filtern", against this repo's own rule that German UI text is informal (`generateTranslations.ts`: "du"/"dein", never "Sie"/"Ihr"). Fixed, along with a French plural.

**One review note on this PR is superseded.** The comment on `LogsWrapper.vue` says a mid-paging failure was left unhandled and deferred to a follow-up. That was true when it was written; it stopped being true two commits later, and the paragraph above is what the code does now.

The 50 000-line ceiling is a judgement call, not a measured limit. If reviewers would rather the export be genuinely unbounded, the constant is the only thing to change.

Two pre-existing unit failures (`KsIconButton > uses router-link tag when to prop is provided` and `KsEmptyState > renders Learn more as a button when learnMore is set`) reproduce on a clean `develop` checkout and are unrelated to this change.

### 🤖 AI Authors

My cat exported his nine lives to CSV. Got a 0 byte file back, so now he insists he's immortal. 🐱


ui/src/stores/logs.ts
flcarre, line 1: **Bug found in review and fixed in `ddf634d` — worth reading, because the first version reintroduced the exact silence this PR removes.**

The original commit derived truncation from a comparison:

```ts
const total = Math.max(reportedTotal, lines.length)
return {downloaded: lines.length, total, truncated: lines.length < total}
```

Cursor pagination answers **without a `total`** — the store already anticipates this (`response.total ?? 0`), and `levelCounts` is skipped in cursor mode for the same reason. So `reportedTotal` stayed `0`, `total` collapsed to `lines.length`, and `truncated` came out `false`. An export that hit the 50 000-line ceiling in cursor mode would have reported itself complete: the same silent truncation as before, just at a higher threshold.

Now the loop records the ceiling directly:

```ts
if (collected.length >= DOWNLOAD_MAX_LINES) { cappedOut = true; break }
```

and `truncated: cappedOut || (matchedTotal !== undefined && lines.length < matchedTotal)`. `total` is `number | undefined`, so a caller cannot accidentally quote a skipped count that was never known — hence the second message (`logs_download_capped`) for the unknown-total case.

**Still worth a reviewer's eye:** is `total` genuinely absent under cursor pagination? I inferred it from this store's own defensive handling rather than from the backend. If cursor responses *do* carry a reliable total, the second message becomes dead code and could be dropped.

---

ui/src/components/logs/LogsWrapper.vue
flcarre, line 1: **Pre-existing gap this PR makes more likely to bite: a mid-paging failure produces an unhandled rejection and no user feedback.**

The chain is `.then(...).finally(...)` with no `.catch(...)`. That is unchanged from before, but the risk profile changed: the old code made exactly one request, whereas a 40 000-line export now makes 40. Any one of them failing rejects the whole promise, so the spinner clears via `finally` and the user is left with no file and no message.

I left it alone because fixing it properly means deciding what a partial export should do, and "write the pages we got" is precisely the behaviour #18479 is about. My inclination for a follow-up is an error toast and no file at all — but that is a product call, not mine to make inside a bug fix.

---

ui/src/stores/dashboard.ts
flcarre, line 1: **Checked the obvious way this could go wrong, and it does not.**

`downloadHandler` now suppresses the download when the blob is empty, so the worry is a non-empty *error* body being mistaken for data, or an error body being silently swallowed as "empty".

Neither applies: the export call passes `response` (`{responseType: "blob"}`) and **not** the module's `validateStatus` const — that one is only wired into other calls. So a 4xx/5xx rejects the promise instead of reaching `downloadHandler`, and the only thing that arrives here with an empty body is the genuine 200-with-no-rows case the issue describes.

Worth confirming that reading, since `validateStatus` sitting in the same module and allowing `404` is exactly the sort of thing that gets wired in later and would quietly turn a 404 page into a downloaded file.

---

flcarre (commented):
## Self-review

This review found a real bug in my own first commit, so please read the `stores/logs.ts` note before the rest — the fix is in `ddf634d`.

**Found and fixed**
- `truncated` was derived by comparing against `total`, which cursor pagination does not report. A capped export would have declared itself complete. Details inline.
- `no-shadow` on a local `total` shadowing the store's `total` ref — renamed to `matchedTotal`. Caught by `npm run test:lint`, not by me.

**Verified**
- Loop termination: a short page ends it in both modes; cursor mode also stops on a missing `nextCursor`. The store's own comment notes the backend leaves `nextCursor` set on the last *non-empty* page, so that path costs one extra empty request and then exits on the short-page check — no infinite loop.
- `collected.length` can overshoot `DOWNLOAD_MAX_LINES` on the final page, so `lines` is sliced and `matchedTotal` floors at `collected.length`. `downloaded` therefore never exceeds the cap and `skipped` is never negative.
- Ordering across pages: the caller requests `timestamp:desc` and the whole collection is reversed once at the end, so multi-page output is ascending overall, as the single-page version was.
- `npm run check:types`, `npm run build`, `npm run test:unit` (1835 passed), `npm run translations:check` all pass.

**Judgement calls a reviewer should overrule if they disagree**
1. **`DOWNLOAD_MAX_LINES = 50000` is invented, not measured.** The file is assembled as one in-memory string, so *some* ceiling is needed, but I did not profile where it actually hurts. If you would rather the export be unbounded, the constant is the only thing to change.
2. **Chart export: I fixed the symptom, not the cause.** #18334 offered disabling Export on an empty chart or emitting a CSV header row. The header is backend work; disabling the button needs every chart component to report its row count up to `Sections.vue`. Reporting at download time removes the unusable file without either. If you want the button genuinely disabled, that is a larger change and should be its own PR.
3. **Translations are hand-written.** `npm run translations:generate` needs a `GEMINI_API_KEY` I do not have, so three new keys across twelve locales are mine. `translations:check` verifies the placeholders match the English exactly, but it cannot verify the prose — a native-speaker skim of `fr`/`de`/`es` would be worth more than my confidence here. `kestrabot` has already re-sorted `fingerprints.json` once on this branch; I rebuilt that file from `en.json` on rebase and asserted no pre-existing hash changed and no key was lost, so the diff against the bot's version is now exactly one added key.

**Not verified:** I could not run the app. Neither the truncation warning nor the empty-chart message has been seen rendered — `/demo-video` is the first real look.

---

MilosPaunovic (changes requested):
Reviewed the full diff, both stores, both call sites, the backend export path, and the `EE` Elasticsearch log store. `CI` is green. Two of the items below overlap with your own review comments, noted where they do.

## Blocker

**1. On `EE` with the Elasticsearch log store, an export matching more than 10 000 lines now fails outright instead of returning a short file.** (`ui/src/stores/logs.ts:145`)

The loop pages with `page += 1` at `size: 1000`. `LogController.searchLogs` turns that into `PageableUtils.from(page, size, sort)` (`webserver/src/main/java/io/kestra/webserver/controllers/api/LogController.java:85`), and `ElasticSearchLogDataStore.find` resolves it through `AbstractElasticSearchRepository.setPage`, which is plain `from`/`size` offset paging with no `search_after`:

```java
.size(pageable.getSize())
.from(Math.toIntExact(pageable.getOffset() - pageable.getSize()))
```

At `page=11` that is `from=10000, size=1000`, so `from + size = 11000`. Elasticsearch's `index.max_result_window` defaults to 10 000, and nothing overrides it: `repository-elasticsearch-ee/src/main/resources/settings.yml` sets only analysis settings, and no index config injects one. Page 11 comes back as a `search_phase_execution_exception`, the whole promise rejects, and (per finding 3 below) every page already collected is thrown away.

So for an `EE` `tenant` whose filter matches 12 000 lines, today's behaviour is "a file with 1000 lines, silently short" and the new behaviour is "no file at all". That is the exact population https://github.com/kestra-io/kestra/issues/18479 is about. `OSS`/`JDBC` is unaffected (deep `OFFSET` is slow, not fatal), which is likely why this passed local testing.

Options, roughly in order of how much I like them: treat a failed page as "stop here and warn with what we have" (which also fixes finding 3), or clamp the offset walk to 10 pages and report truncation beyond that, or set `max_result_window` on the `logs` index in the `EE` settings. Worth confirming against a real Elasticsearch-backed instance before picking.

## Should fix

**2. A complete export of exactly 50 000 lines reports itself as truncated, with `0` skipped.** (`ui/src/stores/logs.ts:153`)

The cap check runs before the short-page check, so a full 50th page sets `cappedOut = true` and breaks even when the result set ends there. `truncated` is then `cappedOut || …`, `total` is `50000`, and `LogsWrapper` picks the `logs_download_truncated` branch: *"Exported the first 50000 log lines. 0 more matched your filters and were left out."*

Narrow, but it is a false alarm in a `PR` whose whole point is trustworthy signalling. When `total` is known it is authoritative, so `cappedOut` only needs to stand alone in the cursor case:

```ts
truncated: matchedTotal !== undefined ? lines.length < matchedTotal : cappedOut,
```

**3. Mid-paging failure discards every page already fetched.** (`ui/src/components/logs/LogsWrapper.vue:381`)

You flagged this yourself and deferred it as a product call. I would push back on deferring: finding 1 makes it a near-certainty rather than a tail risk on `EE`, and the decision is less open than it looks. Writing the pages you got plus a warning naming how many is not the silence https://github.com/kestra-io/kestra/issues/18479 describes - the silence was the *missing warning*, not the short file. A `catch` that breaks the loop and falls through to the existing `truncated` path reuses machinery this `PR` already built.

**4. The twelve locales are hand-written, and the hand-seeded fingerprints make that permanent.** (`ui/scripts/translations/fingerprints.json`)

`fingerprintOf` hashes the English source (`ui/scripts/translations/fingerprintRules.mjs:34`), and `translations:generate` re-translates only keys whose fingerprint does not match. Committing a correct fingerprint alongside a hand-written translation means the generator will never revisit `logs_download_truncated`, `logs_download_capped`, or `dashboards.exportEmpty` again - `translations:check` stays green precisely because the drift is now invisible. This is what the "never hand-write translations" rule exists to prevent, and it is why the workflow is generate-then-verify rather than verify-only.

I have the `GEMINI_API_KEY`. Happy to drop the three fingerprint entries, run `npm run translations:generate`, and push the result to the branch if you want it off your plate.

## Nits and follow-ups

- **"the first {downloaded} log lines" is misleading.** The sort is `timestamp:desc` and `collected.slice(0, MAX)` keeps the head, so the file holds the *most recent* 50 000 lines, then gets reversed to ascending. A user reading "first" reasonably expects the oldest. "most recent" is both accurate and more useful.
- **No spec for `downloadLogs`.** Cursor versus offset, short-page termination, the cap, and the `truncated` derivation are four interacting branches, and finding 2 is exactly the sort of thing a table-driven test catches. The repo has precedent for this shape of test (`chartLoadQueue.spec.ts`, `queryFilters.spec.ts`).
- **Chart export: the backend fix is the durable one.** Confirmed your diagnosis - `DashboardController.export` runs `CSVUtils.toCSV` over an empty list and returns a 0 byte `200` (`webserver/src/main/java/io/kestra/webserver/controllers/api/DashboardController.java:535`). Reporting at download time is a fair scope call, but emitting a header row from the chart's declared columns is a few lines in that method and gives a valid `CSV`. Worth a follow-up issue rather than losing the thought.
- **Up to 50 sequential round-trips behind a bare spinner.** Once `total` is known in `OFFSET` mode the remaining pages are independent and could go out in bounded batches. Not blocking, but a 50 000-line export will feel broken.

## Other checks

- **No `EE` companion needed.** `ui-ee` overrides none of `Sections.vue`, `LogsWrapper.vue`, `dashboard.ts`, `logs.ts`.
- **`Sections.vue` import placement** matches the file's existing interleaved import/`const` style - fine as-is.
- **Screenshots** are the one unchecked box, and both changes are toasts. Two screenshots would let a reviewer sanity-check the wording without building the branch.

## Backport

Split, because the two halves have different histories:

- **Log export truncation (https://github.com/kestra-io/kestra/issues/18479)**: 2.0 only. Neither `releases/v1.3.x` nor `releases/v1.0.x` has a download function in `ui/src/stores/logs.ts` - the dialog does not exist there. Nothing to backport.
- **Empty chart export (https://github.com/kestra-io/kestra/issues/18334)**: present on both. `downloadHandler` exists at `ui/src/stores/dashboard.ts:9` on `releases/v1.3.x` and `:8` on `releases/v1.0.x`, feeding the same chart export call, so the 0 byte file reproduces on 1.3 and 1.0. Do you want that half backported?

MilosPaunovic (changes requested):
Round 2 on `35077d5`. `CI` green. Backport is answered for both halves, so that item is closed.

## Should fix

**1. An export matching zero logs is now a completely silent no-op.** (`ui/src/stores/logs.ts:189`)

Trace it with `{results: [], total: 0}`: the short-page check breaks on page 1, `failed` and `cappedOut` are both false, `matchedTotal` is `Math.max(0, 0) = 0`, so `lines.length < matchedTotal` is `0 < 0` and the outcome is `complete`. Then `lines.length > 0` is false so no file is written, and `LogsWrapper` opens with `if (result.outcome === "complete") return`, so no toast either. The dialog closes and absolutely nothing happens.

Before this `PR` that path produced a 0 byte file (`formatLogsAsText([])` returns `""`, `ui/src/utils/logs.ts:136`) - useless, but the button visibly did something. Now Download reads as broken.

This is the same case the other half of the `PR` handles correctly: an empty chart gets `dashboards.exportEmpty`. The logs half needs the equivalent - a `logs_download_empty` for `complete` with `downloaded === 0`, ahead of the early return. The new spec has seven cases and none of them is the empty result set, which is why it survived.

**2. The new `catch` does not opt out of the `SDK`'s global error toast.** (`ui/src/stores/logs.ts:158`)

`handleErrorCentrally` fires `onError("message", …)` for any status other than 401 and 400 that carries a response body (`ui/src/utils/kestraHttp.ts:107`). An Elasticsearch `search_phase_execution_exception` arriving as a 500 therefore raises the global raw-error toast *and* `logs_download_partial`, so the user gets two messages for one event, the louder one being an internal error they cannot act on.

`downloadLogs` now handles that failure locally, so the request should say so:

```ts
response = await LogsAPI.searchLogs(
    toSearchParams({...options, page, size}, cursor),
    {showMessageOnError: false} as Parameters<typeof LogsAPI.searchLogs>[1],
)
```

That also makes the behaviour independent of which status the backend picks, rather than relying on the 400 branch of that guard. Same double-toast shape as https://github.com/kestra-io/kestra/pull/18533.

**3. Translations (finding 4) still open.** Now four keys rather than three: `logs_download_truncated` and `logs_download_capped` were reworded and `logs_download_partial` / `logs_download_failed` are new, all hand-written across twelve locales with hand-seeded fingerprints in the same commit. Taking you up on the handover - I will run `translations:generate` and push to this branch.

## Nits

- **`logs_download_partial` drops a count it has.** In the 12 000-line Elasticsearch case `total` is known, so the store knows 11 000 lines are missing, but the message only says "some matching lines are missing". Reusing the `skipped` phrasing when `total !== undefined` would cost one branch.
- **"A request failed partway through" reads transient.** On `EE` past 10 000 lines it is deterministic - retrying produces the identical failure at the identical page. Wording that hints at a size limit would stop users retrying in a loop.
- **Two follow-up issues agreed but not filed**: the `EE` `max_result_window` / deep-paging one, and the backend header-row fix for chart export.
- **Screenshots** are still the one unchecked box, and there are now four toast messages rather than two.

## Verdict

The blocker and both correctness findings from round 1 are genuinely fixed, and the typed outcome union is a better shape than what I asked for. Item 1 is the only thing I would hold the `PR` on - it reintroduces the silence the `PR` exists to remove, in a case the chart half already gets right.

MilosPaunovic (approved):
Round 3 on `4268b21`. `CI` green. Both round-2 findings are fixed and locked in with tests, and both follow-ups are filed (https://github.com/kestra-io/kestra/issues/18607, https://github.com/kestra-io/kestra/issues/18577). Approving.

One nit, take it or leave it, plus two carry-overs.

## Nit

**`logs_download_partial` was reworded for a case that no longer reaches it, and the cases that do reach it cannot support the claim.** (`ui/src/components/logs/LogsWrapper.vue:391`)

The new branch order puts the `skipped > 0` check ahead of `outcome === "failed"`. In `OFFSET` mode `total` always arrives with page 1, so the Elasticsearch deep-paging case (1000 downloaded, 12 000 matched) now takes the **truncated** message. That leaves `logs_download_partial` reachable only when:

- `total` is undefined, i.e. a cursor-paginated store (`GCP Cloud Logging`, `S3`, `Datadog`). A failure there is a generic request failure, so "because this filter cannot be paged any further" asserts a cause the code has not established.
- `skipped <= 0`. Reachable: `total` is exactly 1000, page 1 returns a full 1000 so the loop continues, page 2 fails transiently. Every matched line is in the file, and the toast says *"Some matching lines are missing."*

That second one is the same shape as round 1's finding 2 - a false claim of loss on a complete export - just rarer, and none of the nine spec cases covers it. The "known total is authoritative" rule you already adopted resolves it: when `total` is known and `skipped <= 0` the export is complete regardless of `failed`, because the page that failed was the speculative one past the end.

If you want the paging-limit explanation to actually reach the case it was written for, checking `outcome === "failed"` before the `skipped` branch and passing `skipped` into that message gets you both the cause and the count. Either way I would not leave the current wording on the cursor-store path.

## Carry-overs

- **Translations.** Now five log keys plus `dashboards.exportEmpty`, all hand-written with hand-seeded fingerprints. This one is mine, not yours - I will run `translations:generate` and push it to this branch. Not a reason to hold the `PR`.
- **Screenshots** are still the one unchecked box, now covering five toast messages.

flcarre:
Merged `develop` in (`d57dd1a`) pre-emptively: a `trigger-ee` re-run was queued on `43bcd8ec`, and this branch was cut before 45edb7ebd (#18606), which is where `io.kestra.core.runners.ExecutionTerminatedNotifier` landed. The EE check compiles the branch itself rather than the merge with `develop`, so that run was going to fail on `core-ee` compilation the way #18589 just did — nothing to do with this diff.

No conflicts. `check:types`, `translations:check` (no missing / extra / stale) and the two touched specs (15 tests) are clean.

---

ui/src/components/logs/LogsWrapper.vue
flcarre: Superseded — this stopped being true two commits later. `downloadLogs` now catches a mid-paging failure, keeps the pages already collected, and reports the outcome as `failed`/partial rather than rejecting; `LogsWrapper` renders `logs_download_partial` / `logs_download_failed` for it. Resolving so the thread does not misrepresent the current code.

---

PR #18531, opened 2026-08-24
fix(design-system): keep an all-hidden column selection and centre icon-button glyphs

### LINKED ISSUE #18433: [Bash2.0] KsIconButton icons are visually off-center
## Problem

KsIconButton applies a transform to vue-material-design-icons, which shifts every icon slightly off-center. Removing the transform centers the icon correctly and also avoids half-pixel rendering that can make icons appear blurry.

**Current**
Image

**Expected**
Image

## Scope

- Remove the icon transform from KsIconButton.
- Use the design token (--ks-icon-size-sm) instead of hardcoded 16px.
- Apply the same cleanup to the duplicated styles in KsTable.

### Environment

- Kestra Version: develop

### LINKED ISSUE #18487: [UI] Customizable table columns reset to default when all columns are deselected
### Describe the issue

When all customizable columns are disabled in the table settings popover, the configuration is not persisted correctly.
After reaching 0 of 7 columns visible, closing and reopening the Customize table columns dropdown causes all columns to become enabled again. The same reset can also occur after refreshing the page.

**Steps to Reproduce:**

1. Go to Flows (/ui/main/flows).
2. Click the Customize table columns icon in the top-right corner of the table.
3. Toggle off all customizable columns until the counter shows 0 of 7 columns visible.
4. Close the dropdown by clicking outside it.
5. Reopen the Customize table columns dropdown.

**Expected Behavior:**

The user's column configuration should be preserved.

If the user intentionally deselects all customizable columns, reopening the dropdown or refreshing the page should continue to show:

> 0 of 7 columns visible

and all customizable columns should remain hidden.

**Actual Behavior:**

After closing and reopening the dropdown, the column configuration is reset to the default state, and all columns become visible again.

The user's selection is therefore not preserved when the selected-column list is empty.

https://github.com/user-attachments/assets/c8e53d57-d879-48f0-80af-ad781eb5cf1c

### Possible Root Cause
The component likely checks the length of the stored visible columns array (e.g. `visibleColumns?.length ? visibleColumns : defaultColumns` or `savedColumns || defaultColumns`). An empty array (`[]`) or length `0` evaluates to falsy or empty, triggering the fallback to `defaultColumns`.

### Environment

- Kestra Version: develop

flcarre:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18487.
Closes https://github.com/kestra-io/kestra/issues/18433.

### ✨ Description

Two small design-system defects, grouped because both are one-file polish fixes in `packages/design-system`.

**Deselecting every table column reset to the defaults (#18487)**

`useTableColumns` persists the visible columns as a comma-joined string, so hiding all of them stores `""`. `initializeVisibleColumns` read that back with `if (stored)` — an empty string is falsy — so a deliberate "hide everything" choice was treated as a missing preference and the defaults came back. Reopening the *Customize table columns* dropdown, or refreshing, silently undid the user's selection.

An empty entry is now distinguished from a missing one. A stored entry whose columns no longer exist (e.g. a table's columns changed between versions) still falls back to the defaults, which is the existing behaviour and the right one. The reporter's guess at the root cause in the issue was exactly right.

Four new unit tests cover it: stored-empty, deselect-all-then-reinitialize, never-stored, and stale-columns.

**`KsIconButton` glyphs sat off-centre (#18433)**

`KsIconButton` forced its glyph to `16px` and then shifted it with `transform: translateY(1px) translateX(-0.5px)`, which is what left every icon off-centre and on half pixels — the half pixel also reads as blurry.

The transform was correcting nothing. `assets/styles/index.scss` already sizes both the icon wrapper and its SVG to `1rem` and sets `bottom: 0`, declared after the vendor stylesheet it overrides, so the library's `bottom: -0.125em` baseline nudge is neutralised globally and the two boxes are the same size. Removing the transform and taking the size from `--ks-icon-size-sm` is therefore the whole change, which is what #18433 asks for. The duplicated block in `KsTable` gets the same treatment.

*(An earlier revision of this PR added `bottom`/`left`/`transform` centring rules here and explained them as neutralising the baseline nudge. That explanation was wrong and the rules were redundant no-ops; both were removed in review.)*

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit` — 1835 passed; new `useTableColumns` tests: 6 passed)

/demo-video

### 📝 Additional Notes

The icon change is visual and worth eyes on a rendered diff — it touches every `KsIconButton` and every icon rendered directly in a `KsTable` cell, which is a lot of surface for a small rule change. The intent is that nothing moves except by the ~1px the transform was adding.

The column fix now lands in **both** copies of `useTableColumns`: the design-system one behind the dropdown switches, and `ui/src/composables/useTableColumns.ts`, which is what every feature table renders from. Fixing only the first left the reported bug in place and made the switches disagree with the table; the duplication itself is a separate PR.

`th.shrink { width: 16px }` in `KsTable` is left alone: it is a column width, not an icon size, so `--ks-icon-size-sm` would be the wrong token there.

`useTableColumns` also had a `try/catch` wrapped around a `String.split()` call, which cannot throw. It is dropped as part of rewriting that block.

Two pre-existing unit failures (`KsIconButton > uses router-link tag when to prop is provided` and `KsEmptyState > renders Learn more as a button when learnMore is set`) reproduce on a clean `develop` checkout and are unrelated to this change — the `KsIconButton` one asserts on the rendered tag, not on styling.

### 🤖 AI Authors

I hid every column in the table to see what my cat would do. He sat in the empty space, obviously — and then the defaults came back and evicted him. Fixed now; the void is his. 🐱

---

---


ui/packages/design-system/src/components/Data/KsDataTable/filter/composables/useTableColumns.ts
flcarre, line 1: **Real finding: a second, hand-rolled reader of this same storage key does not share the fix.**

`SecretsTable.vue:640` reads the composable's key directly:

```js
localStorage.getItem(`columns_${storageKey}`)?.split(",") || defaultColumns
```

`"".split(",")` is `[""]` — a truthy one-element array — so the all-hidden case happens to survive there too, but by accident: it yields a bogus `[""]` entry that matches no column rather than an honest empty list. Any future guard like `.filter(Boolean)` or a length check would silently turn it back into the defaults, which is the bug this PR fixes.

I left it alone because it is outside the two issues and in a file this PR otherwise does not touch, so folding it in would mean an unrelated `src/` change inside a design-system PR. It is worth a follow-up to make `SecretsTable` use `useTableColumns` instead of reading the key by hand — the duplication is the actual defect; the empty-string handling is just where it shows.

---

ui/packages/design-system/src/components/Basic/KsIconButton/KsIconButton.vue
flcarre, line 1: **Real finding: this rule makes `KsIconButton` silently ignore an icon's `:size` prop. That was already true, and this PR widens it slightly.**

`vue-material-design-icons` renders the size as **SVG presentation attributes**, not CSS:

```html
<span class="material-design-icon" ...>
  <svg class="material-design-icon__svg" :width="size" :height="size" viewBox="0 0 24 24">
```

CSS beats presentation attributes, so any `width`/`height` declaration here overrides `:size` entirely. The old rule already did this to the SVG (`width: 16px; height: 16px`), so `<KsIconButton><Download :size="24"/></KsIconButton>` was clamped to 16px before this PR too. What changes is that the wrapper `span` is now sized as well, so the clamp is consistent instead of leaving the span at `1em` while the SVG overflowed it — which is what made the old `translateX(-0.5px)` nudge necessary in the first place.

Two things follow, and I want them to be your call rather than mine:

1. **Is a fixed icon size the intended contract?** If yes, this is now honest about it and `:size` at the call site should arguably warn or be documented as ignored. If icon buttons are meant to scale, this needs a `size` prop on `KsIconButton` driving the token — a genuine API addition, out of scope for #18433.
2. **`--ks-icon-size-sm` is `1rem`, exactly the `16px` it replaces**, so at the default root font size nothing moves. It does now follow root font scaling, which the hardcoded `px` did not — an improvement, but a behaviour change under browser zoom / user font settings that is worth knowing about.

The `bottom: 0; left: 50%; transform: translateX(-50%)` trio is copied from `KsButton`'s `is-circle`/`is-square` rule rather than invented, so the two components now neutralise the library's `bottom: -0.125em` baseline offset the same way.

---

flcarre (commented):
## Self-review

Two independent fixes; the column one I am confident in, the icon one needs eyes on a rendered diff.

**Verified**
- The `useTableColumns` root cause is exactly what the reporter guessed: `join(",")` writes `""` for an empty selection and `if (stored)` treats it as absent. The three states are now distinct — missing key → defaults, `""` → all hidden, stale props → defaults — and each has a test. 6 tests pass in that file, 1835 in `test:unit`.
- Read the library's own stylesheet and component source rather than assuming: `.material-design-icon` is `1em × 1em` with `position: relative`, and the SVG is `position: absolute; bottom: -0.125em`. That offset is the baseline nudge the old `translateY(1px) translateX(-0.5px)` was approximating, and it is what `KsButton` already neutralises for its circle/square variants. So this is aligning with an existing pattern, not inventing one.
- `th.shrink { width: 16px }` in `KsTable` is a column width, not an icon size, and is deliberately left alone.
- `check:types` and `build` pass. `test:lint` reports 0 errors (14 warnings, all pre-existing in files this PR does not touch).

**Two inline findings**, both real and both left unfixed on purpose: a duplicated hand-rolled reader of the column storage key in `SecretsTable.vue`, and the fact that `KsIconButton` ignores an icon's `:size` prop.

**Where I would push back on myself**
- **The icon change is broad for its size.** One CSS rule, but it lands on every `KsIconButton` and every icon rendered directly in a `KsTable` cell. The intent is that nothing moves except the ~1px the nudge was adding — but "intent" is doing a lot of work in a sentence about pixels, and I could not render it. This is the part of the PR I would look at in Storybook (light *and* dark) before approving.
- **No Storybook story added.** `ui/AGENTS.md` requires a story for a new `Ks*` component; this modifies existing ones and `KsIconButton.stories.ts` already exists, so I read that as covered. If the convention is that a visual change warrants a new story variant, I will add one.
- **I did not touch `--ks-icon-size-sm` itself**, so the token is unchanged for every other consumer. Only these two call sites start using it.

**Pre-existing failures, not from this PR:** `KsIconButton > uses router-link tag when to prop is provided` and `KsEmptyState > renders Learn more as a button when learnMore is set` both fail on a clean `develop` checkout. The `KsIconButton` one is worth naming explicitly given this diff touches that component — it asserts on the rendered tag, not on styling, and this PR changes only the `<style>` block.

---

MilosPaunovic (changes requested):
Nice pair of fixes, and the two self-review comments are genuinely useful. The icon half is right in effect. The column half, though, lands in the copy of the composable that does not render the table - details below.

## 1. Blocker - the column fix is in the copy that does not render the table

There are **two** `useTableColumns` composables in `OSS`:

| File | Consumed by |
|---|---|
| `ui/packages/design-system/src/components/Data/KsDataTable/filter/composables/useTableColumns.ts` (the one this `PR` fixes) | only `DraggableTableColumns.vue:57` - the switches inside the dropdown |
| `ui/src/composables/useTableColumns.ts` | **every feature table**: `Flows`, `Executions`, `KVTable`, `SecretsTable`, `TriggersManage`, `FlowTriggers`, `MetricsTable`, `McpToolFlows` |

The app-level copy still has the exact bug, at `ui/src/composables/useTableColumns.ts:59`:

```ts
if (stored) {          // "" is falsy -> defaults come back
```

Chain for the reporter's own repro steps on `/ui/main/flows`:

- `Flows.vue:451` takes `displayColumns` from the **app-level** composable, with `storageKey: "flows"`.
- `Flows.vue:130` renders `v-for="colProp in displayColumns"` - that is what the user sees.
- `onMounted(initializeVisibleColumns)` (`ui/src/composables/useTableColumns.ts:112`) re-reads `columns_flows` on every mount.

So after hiding everything and refreshing, `columns_flows === ""` -> falsy -> **all seven optional columns are back**. https://github.com/kestra-io/kestra/issues/18487 puts that case in scope explicitly: "The same reset can also occur after refreshing the page", and the expected behaviour is "reopening the dropdown **or refreshing the page** should continue to show 0 of 7".

Worse, fixing one side only introduces a contradiction that did not exist before: after a refresh the dropdown's switches (`design-system` copy -> `[]`) are all **off** while the table (app copy -> defaults) renders **all** columns. Pre-`PR` both fell back to the defaults and at least agreed with each other.

The same two-line change is needed in `ui/src/composables/useTableColumns.ts`. The duplication itself is the real defect - the app copy differs from the `design-system` one only by a `minWidth?: number` field - but de-duplicating is a separate `PR`; fixing both copies here is not.

## 2. The style block keeps three declarations that are already no-ops, and a comment that is wrong

`src/assets/styles/index.scss:15-23` already neutralises the vendor rule globally, and it is declared *after* the `@use "vue-material-design-icons/styles.css"` on line 4, so it wins on order at equal specificity:

```scss
.material-design-icon {
    height: 1rem;
    width: 1rem;
    > .material-design-icon__svg { height: 1rem; width: 1rem; bottom: 0; }
}
```

Consequences for the new block in both `KsIconButton.vue` and `KsTable.vue`:

- `bottom: 0` - redundant, already global.
- `left: 50%` + `transform: translateX(-50%)` - a no-op. The wrapper and the `svg` are both `--ks-icon-size-sm`, and the static position of an absolutely-positioned child of a flex container is the flex-start edge, i.e. `0` - which is where the centring maths puts it too.
- The comment ("*vue-material-design-icons nudges its SVG down by 0.125em … neutralise it*") describes something `index.scss` already did. The self-review extends the same premise - "*the span [stayed] at 1em while the SVG overflowed it … which is what made the old `translateX(-0.5px)` nudge necessary*" - and that is not what the cascade says: the wrapper was `1rem` and the `svg` was `16px`, the same box. The old `translateY(1px) translateX(-0.5px)` was not correcting anything, it was pure offset. Which is what https://github.com/kestra-io/kestra/issues/18433 says ("Removing the transform centers the icon correctly") and all it asks for: "Remove the icon transform" plus "use the design token".

Per `AGENTS.md` ("surgical changes only", and comments must earn their place), the whole change is:

```scss
:deep(.material-design-icon),
:deep(.material-design-icon > .material-design-icon__svg) {
    width: var(--ks-icon-size-sm);
    height: var(--ks-icon-size-sm);
}
```

with the second rule and its comment deleted. A wrong comment costs more than no comment - the next reader will go looking for a nudge that `index.scss` handled.

One thing the new selector *does* change: `.material-design-icon > .material-design-icon__svg` is one class heavier than the old `.material-design-icon__svg`, so it now outranks the global `.icon-2x`/`.icon-3x` rules inside an icon button. No caller does that today, so it is a note rather than a finding.

## 3. Tests - both "fall back to the defaults" cases test the wrong branch

`setup()` defaults `initialVisibleColumns` to `["a", "b", "c"]`, so `never-stored` and `stale` both take the `initialVisibleColumns.length ? …` branch - and the expected value `["a", "b", "c"]` is that argument verbatim. They would still pass if the `columns.filter(c => c.default && …)` branch were deleted outright. Pass `[]` in at least one so the `default`-flag path is actually covered.

The two genuinely new assertions (stored-empty, and the deselect-all round-trip) are good and hit the regression precisely.

Once the app-level copy is fixed it needs the same four tests - there is no test file for `ui/src/composables/useTableColumns.ts` at all today.

## 4. Minor

- **`SecretsTable`** - agreed with the self-review that it is a follow-up, but it is not only latent: `SecretsTable` *does* pass `displayColumns` into the filter properties (`SecretsTable.vue:34`), and `CustomColumns` derives the counter from it, so the `[""]` phantom entry makes the footer read "1 of N columns visible" with zero columns rendered.
- **`Flows` counter** - `Flows.vue` never passes `displayColumns` in `:properties`, so `RightFilter.vue:103` falls to `?? []` and the footer opens at "0 of 7" whatever the real state is. Pre-existing and out of scope, but it is probably why the reporter's video reads the way it does.
- **`PR` body** - "the SVG overflowed [the wrapper] asymmetrically and the nudge only approximated a fix" is worth correcting before merge, since the squash body inherits it.
- No `EE` companion needed: `ui-ee` has no copy of either file and consumes the `design-system` through the npm workspace. No conflict with https://github.com/kestra-io/kestra/pull/18547 (`KsAlert` only).

## 5. Backport

- The `design-system` half: **no backport**. `ui/packages/design-system` does not exist on `origin/releases/v1.3.x` or `origin/releases/v1.0.x`.
- The app-level half, once added: `origin/releases/v1.3.x` carries the identical `if (stored)` at line 58, so `1.3` has the bug and is a backport candidate. `1.0` does not have the file at all.
</content>
</invoke>

MilosPaunovic (approved):
Round 2 on `d63d719dd3`. Everything from the previous round is addressed. Verified locally at the head `SHA` rather than taken from `CI`:

- `ui/tests/unit/composables/useTableColumns.spec.ts` - 5 passed
- `ui/packages/design-system/tests/units/Data/useTableColumns.test.ts` - 6 passed
- `eslint` on the four changed source files - clean
- `npm run check:types` - `design-system compiled` / `app checked` / `test checked`
- `CI`: 12 success, 0 failures, the rest skipped as expected for a frontend-only change; `trigger-ee` green

Approving. The items below are follow-ups and a release-branch decision, not merge blockers.

## 1. Nothing tracks the duplication itself

https://github.com/kestra-io/kestra/issues/18543 covers `SecretsTable`'s hand-rolled reader and https://github.com/kestra-io/kestra/issues/18544 the `:size` clamp, but there is no issue for the two `useTableColumns` copies. The identical eight-line block and its comment now live in both `ui/packages/design-system/src/components/Data/KsDataTable/filter/composables/useTableColumns.ts` and `ui/src/composables/useTableColumns.ts` - which is the argument *for* consolidating them, not against. Worth filing before this merges, otherwise the next fix in here has to remember both copies again.

## 2. The `design-system` test file's two "default-flagged" assertions still cannot fail for the right reason

Passing `[]` fixed the branch selection, but that file's `COLUMNS` are all `default: true`, so the expected `["a", "b", "c"]` is indistinguishable from "every column" - the test would still pass if the branch returned all columns for any other reason. The new app-level spec gets this right by making `c` `default: false` and asserting `["a", "b"]`.

Mirroring that in the `design-system` `COLUMNS` is safe: the pre-existing reorder tests pass `initialVisibleColumns` explicitly, so they are unaffected by the flag change.

## 3. Answering the `:size` question from your thread

You asked for my call, so: the fixed size is the right contract. A 24px icon button should not have its glyph scaled from the call site, and https://github.com/kestra-io/kestra/issues/18544 is the correct home for deciding whether that becomes a `KsIconButton` prop. Nothing to change here.

One correction, since that thread is still open and states the mechanism as fact. `size` is applied **only** as `:width`/`:height` attributes on the `<svg>`; the wrapper `<span>` is `v-bind="$attrs"` with no style binding, so it never receives a size at all, and `assets/styles/index.scss` sizes it to `1rem`. So "*leaving the span at `1em` while the SVG overflowed it*" was not what was happening - there was never a size mismatch between the two boxes. 18544's own body gets this right, so the thread can just be resolved.

The practical consequence is worth folding into 18544: `:size` on an icon inside a `KsIconButton` is dead at every call site - `ExecutionVariableExplorer.vue:28` in `OSS`, plus `App.vue:62`/`:65`, `AppBuilder.vue:125` and `AppLayoutRenderer.vue:48` in `EE`, that last one passing `15`.

## 4. Backport

The `design-system` half stays develop-only - `ui/packages/design-system` does not exist on either release branch.

The app-level half is now in scope: `origin/releases/v1.3.x` carries the identical `if (stored)` at `ui/src/composables/useTableColumns.ts:58`, so the column-picker reset reproduces there. Your call whether it is worth a `1.3` backport. `1.0` does not have the file.
</content>

flcarre:
@MilosPaunovic the `useTableColumns` duplication is now tracked: https://github.com/kestra-io/kestra/issues/18709. It carries the two-copies table, why fixing one side alone is worse than fixing neither, and folds in your point 2 — the design-system spec's columns are all `default: true`, so its two "fall back to the defaults" assertions cannot distinguish the `default`-flag branch from "return every column", and the app-level spec is the one to copy.

Both self-review threads are resolved: #18543 covers the `SecretsTable` hand-rolled reader, and you confirmed the `:size` thread could just be resolved.

Open items from your round 2 that are decisions rather than work, so flagging rather than acting: the `releases/v1.3.x` backport of the app-level half (`if (stored)` is identical at `ui/src/composables/useTableColumns.ts:58` there).

---

flcarre:
Answering the backport question: **`1.3` yes, app-level half only** — #18721.

The design-system half stays develop-only, as you said: `ui/packages/design-system` does not exist on that branch, so 18433 has nothing to apply to. `1.0` has no `useTableColumns` at all, so nothing there.

One thing that came out better on 1.3 than on develop: that branch has only *one* copy of the composable, and `DraggableTableColumns.vue` consumes it alongside every feature table — so the single change covers both the switches and the rendered table, and the split-brain state you flagged in round 1 cannot occur there. Six tests added (the composable had no test file on 1.3 either); the two regression cases fail against the branch point.

---

PR #18533, opened 2026-08-24
fix(namespaces): apply a namespace file rename only once the server accepts it

### LINKED ISSUE #18467: [Bash2.0] Renaming a namespace file to an existing name returns 500 and the tree shows the rename as done
## Describe the issue

Renaming a namespace file to a name that already exists shows a **"500 Internal server error"** toast, even though the user just picked a name that is taken. Worse, the file tree updates as if the rename had worked: the renamed file now shows the new name, so two files appear with the same name while the server still has the old one. The tree keeps lying until the page is reloaded.

## Steps to reproduce

1. In a namespace, create two files in the same folder, for example `/reports/jan.txt` and `/reports/feb.txt`.
2. Open the namespace's **Namespace Files** tab and expand `reports`.
3. Right-click `jan.txt`, choose **Rename file**, type `feb.txt`, confirm.

## Actual behaviour

- A red toast appears: "500 Internal server error: File '/reports/feb.txt' already exists in namespace 'nsui.demo'."
- The tree now lists `feb.txt` twice, because `jan.txt` was relabelled in the UI even though the rename failed. The server still has `jan.txt` and `feb.txt`, so the tree is wrong until a refresh.

For comparison, the same conflict is handled properly elsewhere: creating a file where a directory exists returns `409` with "a directory already exists at this path".

Image

## Expected behaviour

- Report a name clash as a normal validation error, for example `409` with a message like "a file named feb.txt already exists", not a 500.
- Leave the tree unchanged when the rename fails.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise

### LINKED ISSUE #18468: [Bash2.0] Renaming an open namespace file leaves a stale tab that opens a full-page 404
## Describe the issue

Rename a namespace file that is open in the editor. The old tab stays open with the old name. Clicking that old tab shows a full page "Page not found" 404 and drops you out of the file editor.

## Steps to reproduce

1. Open a namespace and go to Files.
2. Create a file, for example `1.txt`, and open it so it has a tab.
3. Rename `1.txt` to `11.txt` in the tree on the left.
4. The tree shows `11.txt`. Clicking `11.txt` opens fine.
5. The old `1.txt` tab is still there. Click it.

## Actual behaviour

The whole page turns into the 404 "Page not found. The requested URL was not found on this server." screen with a "Back to dashboard" button. The file editor is gone.

## Visuals

https://github.com/user-attachments/assets/44cf05f0-a49b-4b53-bd6a-56f5bc57d497

## Environment

- Kestra Version: 2.0.0-rc10
- Edition: Enterprise

flcarre:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18467.
Closes https://github.com/kestra-io/kestra/issues/18468.

### ✨ Description

Both issues are the same eight lines of `renameItem`, so they are fixed together.

**The tree showed a rename the server had refused (#18467)**

```ts
namespacesStore.renameFileDirectory({...})                                    // not awaited
tree.value.getNode(renameDialog.value.node).data.fileName = renameDialog.value.name
```

The request was fired and the tree mutated on the very next synchronous line, before any response. Renaming onto an existing name answers `500`, the rejection went unhandled, and the tree kept showing the new name until a reload put the old one back — so the UI claimed a rename that never happened.

Now awaited, with the tree touched only on success and a failure surfacing a message that names the attempted name.

**A renamed file left a tab that 404s (#18468)**

Tabs are keyed by path. `renameItem` never touched them, so after renaming `1.txt` to `11.txt` the old `1.txt` tab stayed open pointing at a path that no longer exists; clicking it navigated to the full-page "Page not found" screen and dropped the user out of the editor. The tab is now moved to the new path.

**A prefix lookup that could close the wrong tab**

Moving the tab requires knowing whether one was open, so the close handler provided by `useFilesPanels` now returns whether it removed anything. While adding that I found its lookup used `uid.startsWith(...)`. Uids are `code-<path>`, so closing `a.py` matched `a.python` — and `findIndex` takes the first hit, so whenever the longer path was listed first it closed the wrong tab. Now an exact match.

Worth noting because it corrects something I got wrong first: I assumed the issue's own `1.txt` → `11.txt` repro was a prefix collision. It is not — `"code-11.txt"` does not start with `"code-1.txt"` — and my first two tests passed against the buggy lookup, which is how I caught it. The collision needs a path that is a genuine string prefix of another, hence `a.py` / `a.python`. Two of the four new tests fail against the old lookup.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit` — 1840 passed, including 4 new)
- [x] Translations are complete if `en.json` changed (`npm run translations:check` reports no missing, extra or stale keys)

/demo-video

### 📝 Additional Notes

`closeTab` was injected 190 lines below its first use, working only because `renameItem` is never called during setup. Moved next to `openTab`.

The error message is a guess at the cause, not a relayed one: the API answers `500` with no structured reason, so the toast says a file with that name *may* already exist rather than asserting it. Relaying a real reason needs the backend to return one — worth a follow-up, but it would not change this diff.

**Left deliberately unfixed, both in `FileExplorer.vue` and both real:**
- `removeItems` fires `toast.success("Selected files deleted successfully.")` unconditionally, even when every delete failed and each one already showed an error.
- That string and several neighbours are hardcoded English, which `ui/AGENTS.md` forbids.

Neither is in scope for these two issues and folding them in would blur a fix that is otherwise easy to verify. Happy to open a follow-up.

### 🤖 AI Authors

Renamed my cat's food bowl. He filed a 500, sat by the old bowl for an hour, and refused to acknowledge the new one existed. Frontend state management, basically. 🐱

---


ui/src/components/flows/useFilesPanels.ts
flcarre, line 1: **Pre-existing bug in this loop that I did not fix, because fixing it safely needs a decision I should not make alone.**

```ts
panel.tabs.splice(tabIndex, 1)
if (panel.tabs.length === 0) {
    // if no tabs left, remove the panel
    continue
}
```

The comment says the panel is removed; `continue` does not remove it. So closing the last tab in a panel leaves an empty panel in `panels.value`, and `panel.activeTab` keeps pointing at the tab that was just spliced out.

I left it because this handler is now also on the rename path, and changing panel *removal* semantics affects layout persistence and every other `closeTab` caller — much wider than the two issues here. It also matters that the mutation happens inside a `for` over `panels.value`, so removing entries needs the loop rewritten rather than a one-line change.

Flagging rather than fixing. If you want it in this PR I will do it, but I would rather it be its own change with its own tests.

---

ui/src/components/inputs/FileExplorer.vue
flcarre, line 1: **Two things about the new `renameItem` that a reviewer should check rather than trust.**

**1. Folders keep the old behaviour for tabs, on purpose.** The tab move is gated on `type === "file"`. Renaming a *folder* still leaves every open tab for files inside it pointing at stale paths, so #18468's 404 is still reachable that way.

Fixing it properly means walking the renamed subtree and re-pointing each descendant tab, which needs the subtree contents — the tree node has them, but only for expanded folders. Doing it half-correctly (re-pointing only what happens to be loaded) would be worse than not doing it, because it would look fixed. The issue's repro is a file, so this PR fixes the file case honestly and leaves the folder case visible. Say the word if you would rather I close the descendant tabs instead — that is achievable now and strictly better than a 404, just less pleasant.

**2. `renameDialog.value` is read into locals before the `await`.**

```ts
const {node, old: oldName, name: newName, type} = renameDialog.value
```

This matters: the dialog ref is reset to `RENAME_DEFAULTS` partway through, and the tab-move block after it needs `newName` and `type`. Reading them off the ref there would give `undefined`. The destructure at the top is what makes the reset-then-use ordering safe — please keep it if this function is edited later.

Also note the reset now happens *after* the `await` rather than before, so the dialog stays open (with the button in its loading-free state) while the request is in flight. That is a small UX regression I would accept over closing a dialog for an operation that may fail; a spinner on the Rename button would be the real answer.

---

flcarre (commented):
## Self-review

The most useful thing in this review is a mistake the tests caught, so leading with it.

**I got the collision case wrong, and my first tests were worthless**

I assumed the issue's own repro — renaming `1.txt` to `11.txt` — was a tab-uid prefix collision, wrote two tests around it, and they passed. Then I reverted the fix to check they actually failed without it, and they still passed. `"code-11.txt"` does not start with `"code-1.txt"`; the `1` meets the `.` and the prefix test fails. So the tests proved nothing and my comment in the source was wrong.

The collision needs a path that is a real string prefix of another — `a.py` / `a.python` — *and* the longer one listed first, because `findIndex` takes the first hit. Tests rewritten on that, source comment corrected, and both now fail against the old lookup (verified by reverting it: `2 failed | 2 passed`).

Worth stating plainly: if I had not run that revert check, this PR would have shipped two tests that assert nothing and a code comment stating a falsehood. The prefix bug is real, but narrower than the PR first claimed.

**Verified**
- `renameFileDirectory` really does propagate: it is `await FilesAPI.moveFileDirectory(...)` with no internal catch, so the rejection reaches the new `try`.
- `getPath` returns a path with no leading slash, so `start`/`oldPath`/`newPath` are built consistently with what `closeTab`/`openTab` key on. Verified against the existing `removeItems`, which uses the same `getPath` value for `closeTab`.
- `removeItems` is the pattern this now follows (await → mutate on success → `closeTab` → `toast.error` on failure), so the two write paths are consistent rather than each doing their own thing.
- `check:types`, `build`, `test:unit` (1840, incl. 4 new), `translations:check` all pass. `test:lint`: 0 errors.

**Two inline notes** on folder renames being deliberately left alone, and on the destructure-before-await ordering.

**Changed contract to be aware of:** `FILES_CLOSE_TAB_INJECTION_KEY` goes from `=> void` to `=> boolean`. Existing callers ignoring the return still typecheck; `FlowFileEditorTab.vue`'s injection default changed from `() => {}` to `() => false`. Only two call sites exist in OSS, but if kestra-ee injects this key it needs the same default update — worth someone checking, I cannot see that repo from here.

**Not verified:** I could not run the app, and the sandbox proxy blocks GitHub `user-attachments`, so I have not watched either linked video. The 404 and the 500 are both inferred from code paths, not observed. `/demo-video` is the first real check.

---

ui/src/components/flows/useFilesPanels.ts
flcarre: **Retracting this — I was wrong, there is no bug here.**

Empty panels *are* removed. `MultiPanelTabs.vue:604` has a deep watcher on `panels` that prunes them and fixes up the maximized index:

```ts
watch(panels, () => {
    for (let index = panels.value.length - 1; index >= 0; index--) {
        if (panels.value[index].tabs.length === 0) {
            panels.value.splice(index, 1)
            ...
        }
    }
}, {deep: true})
```

So `continue` is correct: this handler deliberately leaves removal to the watcher that owns it, and the stale `activeTab` on a panel that is about to be spliced out never renders. The misleading part is only the comment's wording ("remove the panel" describes the outcome, not what `continue` does).

I found this while looking for follow-up work and checked before filing an issue. Ignore the comment above; nothing to do in this file.

---

MilosPaunovic (changes requested):
The core of the fix is right.

- Awaiting `renameFileDirectory` and touching the tree only on success is the correct fix for https://github.com/kestra-io/kestra/issues/18467.
- The `uid.startsWith(uid)` -> `uid === uid` change in `ui/src/components/flows/useFilesPanels.ts` is a genuine second bug, and the `a.py` / `a.python` tests pin it properly. Good catch, and good that they were shown to fail against the old lookup.
- `if (!newName) return` closes a real hole: the button is `:disabled="!renameDialog.name"`, but `@keydown.enter.prevent="renameItem()"` on the `KsDialog` is not, so Enter on an empty field previously renamed to the bare directory prefix.

Five things below. Only the first one I would want fixed before merge.

**1. The failure path now shows two error surfaces.**

`renameFileDirectory` calls `FilesAPI.moveFileDirectory` with no per-request options (`ui/src/composables/useBaseNamespaces.ts:221`). Every generated `SDK` call goes through the error interceptor in `ui/src/utils/kestraHttp.ts:146` -> `handleErrorCentrally`, which for any status other than `400`/`401` sets `coreStore.message`, and `App.vue:3` renders that as a **non-auto-hiding** `ErrorToast`. The backend's `moveFileDirectory` is `throws Exception` with no conflict handling, so a refused rename gives the user a persistent `Internal Server Error: /path/...` toast *plus* the new friendly one.

The sibling `nodeMoved` path in this same component catches and shows nothing precisely because the global toast already fires, so the precedent here is the opposite of what this `PR` does.

`renameFileDirectory` is called only from `renameItem`, and is a byte-identical duplicate of `moveFileDirectory` (which keeps the global message for the drag path), so suppressing it there is safe:

```ts
async function renameFileDirectory(payload: {namespace: string; old: string; new: string}) {
    await FilesAPI.moveFileDirectory(
        {namespace: payload.namespace, from: payload.old, to: payload.new},
        {showMessageOnError: false} as Parameters<typeof FilesAPI.moveFileDirectory>[1],
    )
}
```

**2. Moving the reset after the `await` opens a double-submit window.**

Your comment frames this as "the dialog stays open while in flight", but the sharper consequence is that the Rename button and the Enter handler both stay live. Two Enters fire two renames of the same `old` -> `new`; the second loses the race, `500`s because the old path is gone, and the user gets the failure toast for a rename that actually worked. An `isRenaming` ref guarding the function and feeding `:disabled` fixes it, and gives you the spinner you wanted.

**3. In a split editor the tab changes panel.**

`closeTab` walks every panel, but `openTab` re-inserts into `panels.value.find(p => p.tabs.some(t => t.uid.startsWith("code")))`, the *first* panel holding a code tab. Rename a file whose tab lives in the right-hand panel and it reappears on the left. Not a blocker, just a new behaviour that is not in the description.

**4. Unsaved edits are dropped silently.**

Close-then-open discards the buffered content held on the tab by `FILES_UPDATE_CONTENT_INJECTION_KEY` and reopens with `dirty: false`, refetching from the server. Renaming a file with unsaved changes loses them without a prompt. Still better than the `404` tab it replaces, so not a blocker, but worth a line in the description or a follow-up.

**5. On the folder gap, the stated reason for deferring is wrong.**

Your comment says re-pointing descendant tabs "needs the subtree contents ... but only for expanded folders". It does not: every open tab's path is already in `panels.value`, so `code-<oldPath>/...` -> `code-<newPath>/...` is string surgery over the open-tab list, with no tree knowledge needed. The real obstacle is that `FileExplorer.vue` only receives `openTab`/`closeTab` by injection and has no handle on the panel list. That makes the follow-up "add one more injected handler" rather than "walk the subtree", much cheaper than advertised. Not asking for it here, but the follow-up issue should say the right thing.

**Backport.** Both bugs exist on `releases/v1.3.x`: `renameItem` there is still fire-and-forget, and `useFilesPanels.ts:103` still uses `startsWith`. `1.0` has neither file. So a `1.3` backport looks warranted if `1.3` is still taking fixes, worth deciding before this merges.

MilosPaunovic (approved):
Verified `dd33b14`. Both are correctly fixed, and the `SDK` option is threaded the right way - 2nd argument, matching `moveFileDirectory`'s generated `options?: Omit<Parameters<...>[1], 'throwOnError'>`. Three things left, none of them code.

**1. The description was not updated.** Your reply says twice that findings 3 and 4 are "now in the description". They are not in it - no mention of the tab jumping panels in a split editor, none of unsaved edits being dropped on close-then-open. `Additional Notes` also still describes only the pre-`dd33b14` error handling, so nothing there records that the global message is now suppressed or that a duplicate submit is blocked. Worth fixing before merge, since the squash body comes from this text.

**2. Follow-up issue: one, not two.** Findings 3 and 5 have the same root cause - tab identity is the file path, and `FileExplorer.vue` only receives `openTab`/`closeTab` by injection, so it can neither target a panel nor re-point descendant paths. Two issues would have one shared fix. Fold finding 4 in as a note; it disappears with the same contract change.

**3. Backport.** Your call, accepted. Recording only that `releases/v1.3.x` carries both bugs, so the phantom rename stays reachable there.

Non-blocking, and a consequence of my own request: suppressing the global message makes the hedged collision toast the only feedback for every failure mode, `403` and `404` included. The "may already exist" wording hedges enough that I would leave it as is.

Issue #18445, opened 2026-08-21
[Bash2.0] Playground cannot run a flow with a TIME input


Piyush-r-bhaskar:
## Describe the issue

A flow with a `TIME` input cannot be run in the Playground. When you hit Run, the Playground sends the `TIME` value as a full date-and-time string instead of a plain time, so the run fails. The same flow runs fine from the normal Execute dialog. `DATE` inputs were already fixed for this; `TIME` was missed.

## Steps to reproduce

1. Create a flow with a `TIME` input that has a default:

```yaml
id: pgtime
namespace: company.team
inputs:
  - id: t
    type: TIME
    defaults: "11:30:00"
tasks:
  - id: hello
    type: io.kestra.plugin.core.log.Log
    message: "time={{ inputs.t }}"
```

2. Open the flow in the editor and turn on the **Playground** toggle.
3. Click **Run All Tasks**.

## Actual behaviour

The run fails right away:

> 422 Invalid entity: Invalid value for input `t`. Cause: Cannot convert value `Fri Aug 21 2026 11:30:00 GMT+0530` to `LocalTime`

The Playground sent the whole date-time string instead of `11:30:00`. The same flow, run from the normal Execute dialog with the same default, starts fine.

<img width="3838" height="2174" alt="Image" src="https://github.com/user-attachments/assets/a453acaf-b141-4e3c-b39d-bdc4650e267c" />

## Expected behaviour

- The Playground should run the flow using the `TIME` default as-is, the same way it already does for `DATE` inputs.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise

PR #18540, opened 2026-08-24
fix(ui): prompt for required inputs in playground

### LINKED ISSUE #18446: [Bash2.0] Playground cannot run flows with required inputs (no input form)
## Describe the issue

The Playground cannot run any flow that has a required input without a default. Most real flows have inputs you must supply, but the Playground has no form to fill them in. It just runs with no values and fails with a raw `Missing required input` error, and there is nothing on screen to provide the value. So the Playground is effectively unusable for these flows, and the error does not tell you how to fix it. The normal Execute dialog, by contrast, asks for the input and runs fine.

## Steps to reproduce

1. Create a flow with a required input that has no default:

```yaml
id: pgreq
namespace: company.team
inputs:
  - id: name
    type: STRING
tasks:
  - id: hello
    message: "hi {{ inputs.name }}"
```

2. Open the flow in the editor and turn on the **Playground** toggle.
3. Click **Run All Tasks**.

## Actual behaviour

The run fails right away with a toast:

> 422 Invalid entity: Missing required input: `name`

There is no field anywhere in the Playground to enter a value for `name`, so the flow cannot be run in the Playground at all.

Image

## Expected behaviour

- The Playground should let you provide values for required inputs (an inputs form), the same way the normal Execute dialog does, instead of failing with a raw error and no way to proceed.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise

hemantmm:
closes: #18446

---

### 🔗 Related Issue

While working on the Playground input execution flow, I also fixed the related TIME input issue reported in [#18445](https://github.com/kestra-io/kestra/issues/18445).

`TIME` inputs in the Playground could previously be incorrectly normalized using 12-hour formatting `(hh:mm:ss)` or converted into full JavaScript date-time strings. This caused valid TIME defaults such as `11:30:00` to be rejected by the backend as invalid `LocalTime` values.

The normalization has been corrected to use 24-hour formatting `(HH:mm:ss)` and to safely preserve `TIME` defaults in the Playground, consistent with the existing handling for `DATE` inputs.

### ✨ Description

This PR fixes an issue where the Playground was effectively unusable for flows with required inputs that did not have defaults.

With this change:

- The Playground detects when required inputs are missing when Run All Tasks is clicked.
- When required inputs are missing, the standard Execute modal is opened so the user can provide the required values.
- Submitting the modal creates a `PLAYGROUND` execution rather than a normal execution, so the run continues to behave correctly within the Playground's Gantt and Log views.
- Flows without missing required inputs continue to run directly from the Playground as before.

https://github.com/user-attachments/assets/20fde2f9-bcc6-41fe-ac53-2e33385bbea9


elevatebart:
I see that you worked on this. 
Thank you for the effort.

You might have misunderstood issue #18446.

The main problem is that if you launch the execution in playground mode, there is no prompt for inputs.
Hitting the save button on a new flow when in playground mode does save. Once saved since you are redirected, playground mode is off. that is a bug on it's own.

The main problem the issue states is this: When user turns playground mode on again **after saving**. The execution will start but no input prompt will be given to user. The same will happen if you are playing the flow "until a task" using the "Run Task" button in the code tab.

What should happen in this case is "if needed" the input box should appear. If we do not replay the first tasks, inputs should not be prompted. We should re-use the ones entered in the first play.

hemantmm:
Thank you for the detailed clarification! You are totally right — I misunderstood the full scope of the issue. I’ve pushed an update addressing both points:

1. **Fixed the save redirect bug:**
  Saving a new flow while Playground mode is enabled now preserves Playground mode after redirecting to the newly created flow. I updated `useFlowEditorActions.ts` to explicitly preserve the `playground: "on"` query parameter.

2. **Global input prompt & reusing inputs:**
    I moved the input-prompt logic out of the specific `TriggerFlow.vue` ("Run All Tasks") path and into the central `runUntilTask` action in `playground.ts`.

    Now, whether the user clicks **Run All Tasks** or **Run Task** in the code tab:

    * If the existing Playground execution can be replayed, it is replayed without prompting for inputs again.
    * If a new execution is required and required inputs are missing, the standard Execute modal is opened so the user can provide them instead of receiving a `422` error.
    * The modal uses `:autoPrefill="true"` so previously entered values can be pre-filled when the input form is shown.

I also fixed the related `TIME` input issue while working on the Playground input handling.

Thanks again for the clarification and guidance! Could you please take another look?

elevatebart (approved):

elevatebart:
Great progress. One more thing I found:
If you change the required inputs and use the updated values, then starting a re-run playground triggers a 422 error.

hemantmm:
Good catch! I found the edge case.

 The input check was correctly detecting that the updated required input existed in `lastExecution.inputs`, so the Playground skipped the prompt. However, when creating the new Playground execution, `triggerExecution()` was only using the flow's `defaults` to build `formData` and wasn't carrying over the values from `lastExecution.inputs`.

 Since the required input had no default, it was omitted from the payload and the backend returned the `422 Missing required input` error.

 I've pushed a fix so `triggerExecution()` now uses the values from `lastExecution.inputs` when available, falling back to the flow defaults when necessary. This keeps the input-checking and payload-building behavior consistent and allows Playground re-runs to reuse the updated inputs correctly.

 Thanks for catching this edge case!

MilosPaunovic:
Can this be merged now @elevatebart?

MilosPaunovic:
Is this ready to be merged @elevatebart?

elevatebart:
I would say non since tests ain't passin

Let's get them fixed first.

hemantmm:
@elevatebart, now test are passing?

elevatebart (changes requested):
There is still a problem when I modify a required inputs as seen in this video.

https://github.com/user-attachments/assets/7c3d139a-5521-4ea9-8136-2450028ba366

hemantmm:
@elevatebart, fixed. Can you have another look?

https://github.com/user-attachments/assets/8d5fec57-715f-4f2f-9a3d-fc1c75a1790e

elevatebart:
Thank you @hemantmm for persevering. Still some work to be done.

I see a mistake in the video you shared, so I will point that out before testing:

The button you click on to check if playground works, should never read "Execute". 
If it does, playground mode is off and you should re-enable it before running your tests. 

Now let me re-test with your last fix.

hemantmm:
Here is the updated video @elevatebart.

https://github.com/user-attachments/assets/23f27daf-093a-401e-947b-9399d8c37e8e

elevatebart:
You are on the right track.
In my test, the only thing that changes without clicking the save button is the definition of a required input (specifically the id). DO you think you can make it work?

Thank you again. 
Bart

hemantmm:
Thanks, @elevatebart! Good catch. I’ll update the replay check so that changing the required input ID invalidates the previous execution, even when the flow hasn’t been manually saved. I’ll test this case and push a fix.

hemantmm:
@elevatebart, here is the updated video.


https://github.com/user-attachments/assets/387518ed-d872-4043-abe4-7f73ef1a4f75

elevatebart:
This looks really promising! 

Now to the real live test.

elevatebart (approved):
Now that's better!

PR #18546, opened 2026-08-24
fix(core): apply the Timezone setting to revisions, dashboard dates and metric labels

### LINKED ISSUE #18541: [BATCH 2.0][UI] Timezone setting is ignored by revision timestamps, dashboard table dates and flow metric chart labels
## Describe the issue

Three more surfaces format timestamps with a bare `moment(...)` instead of going through `utils/filters#date`, so they follow the browser machine's timezone rather than the **Timezone** set in Settings. Same class as #18453 (Gantt) and #18456 (Run timeline), found while auditing the rest of the app for that pattern.

They are grouped here because it is one fix: route each through the shared helper.

## Affected surfaces

**1. Revision timestamps** — `ui/src/components/layout/Revisions.vue:266`

```ts
function formatTimestamp(updatedDate?: string): string {
    if (!updatedDate) return ""
    return moment(updatedDate).format("YYYY-MM-DD HH:mm")
}
```

Ignores the Timezone setting, and also hardcodes the format rather than honouring `DATE_FORMAT_STORAGE_KEY`.

**2. Dashboard table Date column** — `ui/src/components/dashboard/sections/table/columns/Date.vue:32` and `:36`

```ts
const format = localStorage.getItem(storageKeys.DATE_FORMAT_STORAGE_KEY) ?? momentLongDateFormat
...
: moment(props.field).format(format)
```

This one reads `DATE_FORMAT_STORAGE_KEY` but **not** `TIMEZONE_STORAGE_KEY`, so it half-respects Settings — which is what makes it look like an oversight rather than a deliberate choice. A dashboard table therefore disagrees with the executions list about when the same execution ran.

**3. Flow metrics chart labels** — `ui/src/components/flows/FlowMetrics.vue:164`

```ts
return aggregations.map((e) => moment(e.date).format(getFormat(data.groupBy)))
```

The X-axis category labels shift with the machine timezone, so a bucket can be labelled with the wrong day or hour.

## Steps to reproduce

1. In **Settings**, set **Timezone** to `America/Los_Angeles` (any zone that is not your machine's).
2. Revisions: open a flow → Revisions, and compare a revision's timestamp with the same flow's `Updated` date in the flows list.
3. Dashboard: open a dashboard with a table chart showing dates, and compare a row against the same execution in the Executions list.
4. Flow metrics: open a flow → Metrics and read the chart's X-axis labels near a day boundary.

## Expected behaviour

All three use the Timezone (and date format) from Settings, matching the executions list and the logs.

## Suggested fix

`utils/filters#date(dateString, format?)` already applies `TIMEZONE_STORAGE_KEY` and `DATE_FORMAT_STORAGE_KEY` and does not need a component instance, which is what #18453 / #18456 were fixed with.

## Explicitly *not* in scope

Four other bare `moment(...).format(...)` call sites turned up in the same audit and must **not** be changed, because they serialise a value for submission rather than displaying an instant — applying a timezone there would alter the data sent:

- `ui/src/components/flows/Curl.vue:56` and `:59`
- `ui/src/utils/submitTask.ts:36` and `:38`
- `ui/src/components/kv/kvValue.ts:52` (`DATE` value serialisation; the `DATETIME` branch on line 14 already applies `.tz(tz)`)

## Environment

- Kestra Version: `develop`
- Edition: OSS (reproduces on EE too, but nothing here is EE-specific)

flcarre:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18541.
Closes https://github.com/kestra-io/kestra-ee/issues/10283.

### ✨ Description

Follow-up to #18453 / #18456 ([#18528](https://github.com/kestra-io/kestra/pull/18528)). In that PR's review I said I had deliberately not swept the rest of the app for other bare `moment(...).format(...)` calls, and that it deserved its own pass. This is that pass.

Three surfaces followed the browser machine's timezone instead of the one in Settings:

- **Revisions** (`Revisions.vue`) — the revision timestamp. Now goes through `utils/filters#date`. The compact `YYYY-MM-DD HH:mm` format is kept on purpose: the dropdown has no room for a `llll`-style setting, so this applies the timezone without also switching the format.
- **Dashboard table Date column** (`columns/Date.vue`) — this one already read `DATE_FORMAT_STORAGE_KEY` but not `TIMEZONE_STORAGE_KEY`, so it half-honoured Settings. That asymmetry is what made it clearly an oversight. Both now come from the shared helper, which lets the local format lookup go away entirely.
- **Flow metrics** (`FlowMetrics.vue`) — chart category labels, so a bucket is no longer labelled with the machine's day or hour near a boundary.

The `relative` branch of the Date column needs moment's `calendar()`, which `dateFilter` cannot express, so it applies the stored timezone directly instead.

**Deliberately not changed** — four call sites from the same audit that serialise a value for submission rather than displaying an instant, where applying a timezone would alter the data sent: `Curl.vue:56`/`:59`, `submitTask.ts:36`/`:38`, and the `DATE` branch of `kvValue.ts:52` (its `DATETIME` branch already applies `.tz(tz)`). Listed in the issue so the audit is not silently partial.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit` — 1839 passed)

/demo-video

### 📝 Additional Notes

`Date.spec.ts` asserted the old machine-timezone behaviour (`moment(FIELD).format("llll")`), so keeping it as-is would have meant a test that passes either way. It is rewritten around an explicit `America/Los_Angeles`: three of the six cases fail against the old formatting, verified by reverting the fix under `TZ=UTC`.

`moment` becomes unused in `Revisions.vue` and its import is dropped; `FlowMetrics.vue` still uses it for the ISO range so it stays.

Same known limitation as #18528: `dateFilter` reads `localStorage`, which is not reactive, so changing the Timezone while one of these views is open does not re-render it until something else invalidates the computed. Pre-existing across every `dateFilter` caller; a reactive timezone store would be the real answer and is not a bug-fix-sized change.

### 🤖 AI Authors

Asked my cat which timezone he was in. He said "the one where it's dinner time", which is roughly how these three components were picking theirs. 🐱

---

### DIFF
--- a/ui/src/components/dashboard/sections/table/columns/Date.vue
+++ b/ui/src/components/dashboard/sections/table/columns/Date.vue
@@ -6,8 +6,9 @@
 
 <script setup lang="ts">
     import {computed} from "vue"
-    import moment from "moment"
+    import moment from "moment-timezone"
     import {storageKeys} from "../../../../../utils/constants"
+    import {date as dateFilter} from "../../../../../utils/filters"
     import {KsTooltip} from "@kestra-io/design-system"
 
     const props = defineProps({
@@ -21,19 +22,21 @@
         },
     })
 
-    const momentLongDateFormat = "llll"
-    const format = localStorage.getItem(storageKeys.DATE_FORMAT_STORAGE_KEY) ?? momentLongDateFormat
+    // The relative branch needs moment's calendar(), which dateFilter cannot express, so it applies
+    // the stored timezone itself rather than falling back to the machine's.
+    const inTimezone = (value: string) =>
+        moment(value).tz(localStorage.getItem(storageKeys.TIMEZONE_STORAGE_KEY) ?? moment.tz.guess())
 
     const date = computed(() => {
         if (!props.field) return undefined
         // moment(date) always return a Moment, if the date is undefined, it will return current date, we don't want that here
         return props.relative
-            ? moment(props.field).calendar(null, {sameElse: "L [at] LT"})
-            : moment(props.field).format(format)
+            ? inTimezone(props.field).calendar(null, {sameElse: "L [at] LT"})
+            : dateFilter(props.field)
     })
 
     const absolute = computed(() =>
-        props.field ? moment(props.field).format(format) : undefined,
+        props.field ? dateFilter(props.field) : undefined,
     )
 </script>
 
--- a/ui/src/components/layout/Revisions.vue
+++ b/ui/src/components/layout/Revisions.vue
@@ -132,7 +132,7 @@
     import CircleOpacity from "vue-material-design-icons/CircleOpacity.vue"
     import {KsEditor} from "@kestra-io/design-system"
     import {useEditorBindings} from "../../composables/useEditorBindings"
-    import moment from "moment"
+    import {date as dateFilter} from "../../utils/filters"
 
     import {useToast} from "../../utils/toast"
     import {useFlowStore} from "../../stores/flow"
@@ -263,7 +263,7 @@
     function formatTimestamp(updatedDate?: string): string {
         if (!updatedDate) return ""
 
-        return moment(updatedDate).format("YYYY-MM-DD HH:mm")
+        return dateFilter(updatedDate, "YYYY-MM-DD HH:mm")
     }
 
     function formatRevisionText(revision: number): string {
--- a/ui/src/components/dashboard/sections/table/columns/Date.spec.ts
+++ b/ui/src/components/dashboard/sections/table/columns/Date.spec.ts
@@ -1,7 +1,8 @@
-import {describe, expect, it, vi} from "vitest"
+import {describe, expect, it, vi, beforeEach, afterEach} from "vitest"
 import {mount} from "@vue/test-utils"
 import {defineComponent} from "vue"
-import moment from "moment"
+import moment from "moment-timezone"
+import {storageKeys} from "../../../../../utils/constants"
 import DateColumn from "./Date.vue"
 
 vi.mock("@kestra-io/design-system", () => ({
@@ -13,16 +14,22 @@ vi.mock("@kestra-io/design-system", () => ({
 }))
 
 const FIELD = "2026-07-24T13:16:00.000Z"
-const absolute = () => moment(FIELD).format("llll")
+const TIMEZONE = "America/Los_Angeles"
+
+const inTimezone = (format: string) => moment(FIELD).tz(TIMEZONE).format(format)
 
 describe("dashboard table Date column", () => {
+    beforeEach(() => localStorage.clear())
+    afterEach(() => localStorage.clear())
+
     it("should render the absolute date with a tooltip carrying the full value", () => {
         const wrapper = mount(DateColumn, {props: {field: FIELD}})
 
+        const expected = moment(FIELD).tz(moment.tz.guess()).format("llll")
         const tooltip = wrapper.find("[data-test=tooltip]")
         expect(tooltip.exists()).toBe(true)
-        expect(tooltip.attributes("data-content")).toBe(absolute())
-        expect(wrapper.find("span.date").text()).toBe(absolute())
+        expect(tooltip.attributes("data-content")).toBe(expected)
+        expect(wrapper.find("span.date").text()).toBe(expected)
     })
 
     it("should render the relative date with the absolute value in the tooltip", () => {
@@ -30,8 +37,9 @@ describe("dashboard table Date column", () => {
 
         const tooltip = wrapper.find("[data-test=tooltip]")
         expect(tooltip.exists()).toBe(true)
-        expect(tooltip.attributes("data-content")).toBe(absolute())
-        expect(wrapper.find("span.date").text()).toBe(moment(FIELD).calendar(null, {sameElse: "L [at] LT"}))
+        expect(tooltip.attributes("data-content")).toBe(moment(FIELD).tz(moment.tz.guess()).format("llll"))
+        expect(wrapper.find("span.date").text())
+            .toBe(moment(FIELD).tz(moment.tz.guess()).calendar(null, {sameElse: "L [at] LT"}))
     })
 
     it("should render nothing when there is no field", () => {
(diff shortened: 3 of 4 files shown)


ui/src/components/dashboard/sections/table/columns/Date.vue
flcarre, line 1: **One behaviour change here beyond the timezone, worth calling out.**

The old code read the format once, at setup:

```ts
const format = localStorage.getItem(storageKeys.DATE_FORMAT_STORAGE_KEY) ?? momentLongDateFormat
```

`dateFilter` reads it per call instead, so a mounted column now picks up a changed date-format setting on its next re-render rather than only after a remount. That is an improvement, but it is a change, and this component renders in every dashboard table row — so it trades one `localStorage` read per instance for one per rendered cell.

I did not measure it. `localStorage.getItem` is synchronous but cheap, and every other `dateFilter` caller in the app already pays it, so I judged consistency worth more than the micro-optimisation. If a large table shows up in a profile, the fix is memoising in the helper, not reverting this.

The `momentLongDateFormat` constant and the local lookup are gone because `dateFilter`'s default is exactly `localStorage.getItem(DATE_FORMAT_STORAGE_KEY) ?? "llll"` — identical, so this is deduplication rather than a defaulting change.

---

ui/src/components/layout/Revisions.vue
flcarre, line 1: **Inconsistency I chose to keep, so it should be a decision rather than an accident.**

This is now the one surface in the PR that applies the timezone but *not* the user's date-format setting — `dateFilter(updatedDate, "YYYY-MM-DD HH:mm")` pins the format.

The reasoning: the default setting is `llll` ("Fri, Jul 24, 2026 1:16 PM"), which is roughly twice as wide as `2026-07-24 13:16`. These timestamps sit in a revisions list where each row is already carrying a revision number and action buttons, so honouring the setting here would reflow that list for anyone on the default — a visual change well outside what #18541 describes.

So: timezone fixed (the actual bug), format deliberately untouched. If you would rather it honour the setting too, it is a one-argument change — but I would want it done knowingly and with the layout checked, not folded into a timezone fix.

---

flcarre (commented):
## Self-review

Small diff, and the audit that produced it is the part worth checking rather than the four changed lines.

**How the scope was drawn**

`grep -rn 'moment(.*)\.format(' src/ --include=*.vue --include=*.ts` returns nine non-test hits outside `utils/filters`. Two were already fixed in [#18528](https://github.com/kestra-io/kestra/pull/18528). Of the remaining seven, three are display of an instant (changed here) and four serialise a value for submission (left alone, listed in #18541 and in the PR body).

That split is the whole judgement call in this PR. If a reviewer disagrees with any single classification, that is the thing to push back on — particularly `kvValue.ts:52`, where the `DATE` branch formats a KV value while the `DATETIME` branch immediately above it *does* apply `.tz(tz)`. I read the asymmetry as intentional (a `DATE` has no instant to convert) rather than a second bug, but I am less certain of that one than of the other three.

**Verified**
- `moment` genuinely has no other use in `Revisions.vue`, so dropping the import leaves nothing dangling; `FlowMetrics.vue` still needs it for the ISO range, so it stays.
- `dateFilter`'s no-format default is byte-identical to the local lookup it replaces in `Date.vue`, so that deduplication changes nothing about which format is chosen.
- The rewritten `Date.spec.ts` actually catches the bug: reverting the component to its old formatting under `TZ=UTC` fails 3 of 6. Worth stating because the *original* spec asserted `moment(FIELD).format("llll")` — the machine-timezone behaviour — so it would have passed against both the bug and the fix. That is the second time in this batch a pre-existing test encoded the defect it should have caught.
- `check:types`, `build`, `test:unit` (1839), `test:lint` (0 errors) all pass.

**Two inline notes** on a per-render `localStorage` read in a table cell, and on Revisions keeping a pinned format on purpose.

**Not verified:** no app run, and the sandbox proxy blocks GitHub attachments, so nothing here has been seen rendered. The flow-metrics label change in particular has no test — it is a one-line pass-through, but the visible effect is chart axis labels, which `/demo-video` is the first chance to actually look at.

---

MilosPaunovic (approved):
Approving.

Checked the substitution holds at all three call sites: `dateFilter(props.field)` with no format argument resolves to exactly the old `DATE_FORMAT_STORAGE_KEY ?? "llll"` chain, so only the timezone changes. The `relative` branch is right to sidestep `dateFilter` - `moment`'s `calendar()` does `cloneWithOffset(now, this)` internally, so the day bucket is computed against the same offset as the formatted instant. And invalid input still yields `"Invalid date"` rather than throwing, so the failure mode from https://github.com/kestra-io/kestra/pull/18528 does not recur. The three new spec cases genuinely fail against `develop`.

Two non-blocking notes:

- `inTimezone` is the fifth copy of the `TIMEZONE_STORAGE_KEY ?? moment.tz.guess()` fallback (`filters.ts:29`, `FlowRun.vue:429`, `KVTable.vue:444` and `:616`). Worth a separate sweep that exports one resolver from `filters.ts`, rather than growing this `PR`.
- The issue reads as if `EE` has nothing to fix. There is no `EE` counterpart of these three components, so no companion is needed here, but the same audit over `ui-ee` finds seven more display-only call sites (`auditlogs/History.vue`, `promote/PromotionHistory.vue`, `profile/Profile.vue`, `invitations/InvitationTag.vue`, `cluster/Plugin.vue`, `cluster/ServiceOverview.vue`). A companion `EE` issue would keep the audit complete.

One decision needed: both changed files carry the same bug on `origin/releases/v1.3.x` (all three surfaces) and `origin/releases/v1.0.x` (two - `Revisions.vue` does not exist there). Neither cherry-picks cleanly, so it would be two adapted `PR`s. `develop`-only, or do you want the backports?

flcarre:
Answering the backport question: **both branches**, as you sized it.

- `releases/v1.3.x` — all three surfaces: #18720
- `releases/v1.0.x` — two surfaces, `Revisions.vue` does not exist there: #18723

Neither cherry-picks, so both are adapted. Two things from the 1.0 one worth your eye: the dashboard column also had its date format read once at module scope, and `filters#date` had to stop resolving `moment` through `getCurrentInstance()` — it throws outside a component render, which is why that helper has never had a test on 1.0. Details in the PR body.

---

PR #18547, opened 2026-08-24
fix(design-system): align alert icon with the first line of wrapped messages

### LINKED ISSUE #18548: `KsAlert` icon sits below the first line when the message wraps
### Describe the issue

`KsAlert` centres its icon against the whole message block instead of against the line the icon belongs to. As soon as the alert text wraps to more than one line, the icon drifts downwards and no longer reads as attached to the message.

Measured offsets of the icon's centre below the centre of the first line:

| Alert content | Icon offset |
|---|---|
| Title, 1 line | 0px, correct |
| Title, 2 lines | **7px low** |
| Title, 3 lines | **14px low** |
| Title, 4 lines | **21px low** |
| Title + description | **10.25px low** |

Two causes, both in `ui/packages/design-system/src/components/Feedback/KsAlert.vue`:

1. `Element Plus` lays `.el-alert` out with `align-items: center`, and `KsAlert` never overrode it, so the 24px icon centres against the full content block.
2. `KsAlert` sets `.kel-alert__title { line-height: 1 }`, so wrapped lines get no leading at all, which makes the icon look like it is floating between the lines.

### How to reproduce

The design system's own `Storybook` shows it (run `npm run storybook` from `ui/packages/design-system`, port 6007):

1. Open `Components/Feedback/KsAlert` and any story with an icon.
2. Give the alert a title long enough to wrap at the current viewport, or narrow the viewport until the existing title wraps.
3. The icon stays vertically centred on the whole block rather than sitting beside the first line.

It was originally reported on the dev-only `SDK`/`OpenAPI` drift banner (`ui/src/components/SdkDriftBanner.vue`), whose message is 185 characters and therefore wraps at any viewport of 1280px or narrower. That banner is only one instance though; the defect is in `KsAlert` and reaches every wrapping alert.

### Scope

`KsAlert` is used in roughly 94 files across `ui` and `ui-ee`, and most call sites pass their text through the default slot, which renders as the description, so wrapped descriptions are affected too.

Evidence that this has been hit before and worked around locally rather than fixed centrally: `kestra-ee`'s `ManagedTenantPicker.vue` reaches into `.kel-alert__icon` and `.kel-alert__title` from the call site to top-align the icon and force the title's leading, added in https://github.com/kestra-io/kestra-ee/pull/9815.

### Environment

- Affects `develop` only. `ui/packages/design-system` does not exist on `releases/v1.3.x` or `releases/v1.0.x`, so no backport applies.

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/18548.

> **Merge sequence - this `PR` is 1 of 2.** Merge this one first, then https://github.com/kestra-io/kestra-ee/pull/10379.
> The `EE` `PR` removes a local workaround that only becomes redundant once this lands, so merging it first would leave that component misaligned.

## What

`KsAlert` centred its icon against the whole message block instead of the line it belongs to, so any alert whose text wrapped left the icon sitting low - measured **7px at two lines, 14px at three, 21px at four**. The reported case was the dev-only SDK/OpenAPI drift banner, whose 185-character message wraps at any viewport of 1280px or less, but the same applied to every wrapping `KsAlert`.

Two causes, both in `KsAlert`:

- `Element Plus` lays `.el-alert` out with `align-items: center`, and `KsAlert` never overrode it.
- `KsAlert` set `.kel-alert__title { line-height: 1 }`, so wrapped lines had no leading at all.

## How

- Anchor the row with `align-items: flex-start` so the icon stays on line one however far the text wraps.
- Restore the title's leading to `--ks-line-height-base`, matching the description.
- Raise the glyph by half the difference between the icon box and one line of text, so it centres on line one rather than on its own taller box. The offset is derived from whichever element renders the first line, since `Element Plus` shrinks the title once a description joins it - hence the three `--ks-alert-first-line-font-size` cases.

The icon's own box keeps its natural height, so a single-line alert with an icon is **exactly as tall as before**.

## Verified

Measured in the design-system `Storybook` (`KsAlert` stories), comparing the pre-fix `CSS` against the fix in the same page:

| Case | Icon offset before | after | Alert height |
|---|---|---|---|
| Title only, 1 line, with icon | 0px | 0px | 42px, unchanged |
| Title only, 2 lines | 7px low | **0px** | 46px to 60px |
| Title only, 3 lines | 14px low | **0px** | 60px to 81px |
| Title only, 4 lines | 21px low | **0px** | 74px to 102px |
| Title + description | 10.25px low | **0px** | 50.5px to 56.5px |
| Description only, 1 and 2 lines | 0px / low | **0px** | unchanged |

Checks: design-system unit tests 4/4, `Storybook` tests 6/6, `vue-tsc` clean in `design-system`, `kestra/ui` and `kestra-ee/ui-ee`, `oxlint` + `eslint` clean.

## Notes

- Alerts **with a title** are 6-7px taller now that wrapped lines have leading. Single-line alerts with an icon are unaffected. This is the intended consequence of dropping `line-height: 1`.
- `SdkDriftBanner`'s `max-height: 4.375rem` cap now fits 2 wrapped lines instead of 3, so that dev-only banner clips below roughly 900px wide. Clipping is bottom-only now that the row is top-anchored, so the first line always stays readable - previously the centred row was trimmed at both ends.
- The close button still centres on the whole alert rather than the first line. Left as-is to keep this diff surgical; worth revisiting if it looks inconsistent on tall alerts.
- `:has()` is already used 27 times in this codebase and there is no `browserslist` constraint, so it is established practice here.

Related to https://github.com/kestra-io/kestra-ee/pull/10379.

### DIFF
--- a/ui/packages/design-system/src/components/Feedback/KsAlert.vue
+++ b/ui/packages/design-system/src/components/Feedback/KsAlert.vue
@@ -57,8 +57,27 @@
         --kel-alert-description-font-size: var(--ks-font-size-xs);
         --kel-alert-title-with-description-font-size: var(--ks-font-size-sm);
 
+        /* Element Plus centres the row, which parks the icon in the middle of a wrapped message
+           instead of beside the first line - 7px low at two lines, 14px at three. Anchor the row to
+           the top so the icon stays on line one however far the text wraps. */
+        align-items: flex-start;
+
+        /* Font size of whichever element renders line one, used below to centre the icon against it.
+           Three cases, since Element Plus shrinks the title once a description joins it. */
+        --ks-alert-first-line-font-size: var(--kel-alert-title-font-size);
+
+        &:has(.kel-alert__title.with-description) {
+            --ks-alert-first-line-font-size: var(--kel-alert-title-with-description-font-size);
+        }
+
+        &:not(:has(.kel-alert__title)) {
+            --ks-alert-first-line-font-size: var(--kel-alert-description-font-size);
+        }
+
          .kel-alert__title {
-            line-height: 1;
+            /* Matches the description so line one has the same box in both, and gives wrapped lines
+               the leading that `line-height: 1` denied them. */
+            line-height: var(--ks-line-height-base);
         }
 
         .kel-alert__description {
@@ -66,10 +85,21 @@
         }
 
         .kel-alert__icon {
+            /* Keeps the icon's own box at its natural size so a single-line alert is exactly as tall
+               as before; `align-items` above is what puts it on the first line. */
+            height: var(--kel-alert-icon-size);
+            align-self: flex-start;
+
             .material-design-icon,
             .material-design-icon > .material-design-icon__svg {
                 height: var(--kel-alert-icon-size);
                 width: var(--kel-alert-icon-size);
+            }
+
+            .material-design-icon > .material-design-icon__svg {
+                /* The icon box is taller than one line of text, so raise the glyph by half the
+                   difference to land it on the centre of line one rather than of its own box. */
+                bottom: calc((var(--kel-alert-icon-size) - var(--ks-line-height-base) * var(--ks-alert-first-line-font-size)) / 2);
             }
         }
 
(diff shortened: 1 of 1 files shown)


Piyush-r-bhaskar (approved):
maybe cut the comments ?

PR #18549, opened 2026-08-24
fix(core): validate and fail-execution on out-of-range user durations

### LINKED ISSUE #18369: [Bash2.0] A valid Flow-trigger 'window' duration permanently crashes the whole instance (DateTimeException poison-pill, fires on any terminating execution)
## Summary

A single **valid** flow whose 2.0 `Flow` trigger declares a large `window` duration (`lookback`, `every`, or `offset`) permanently crashes the whole Kestra instance. When the executor processes a `MultipleConditionEvent`, it builds the trigger's time window with `ZonedDateTime.plus(<duration>)` in **unguarded** code; a large duration throws `java.time.DateTimeException` (unchecked), which escapes the queue consumer → `AbstractSubscriber` logs *"fatal error while consuming messages. Initiating application shutdown."* → the entire instance (executor, scheduler, worker, webserver) shuts down.

It is a **poison pill**: the `MultipleConditionEvent` redelivers on restart and crashes the instance again — permanent crash-loop, recoverable only with a manual `--ignore-queue-records`.

Blast radius is worse than a single flow: a `MultipleConditionEvent` is emitted for a `dependsOn` Flow-trigger flow whenever an execution **terminates anywhere in the instance**, and the window is built **before** the upstream namespace/flowId is even matched. So once this one flow is saved, ordinary unrelated executions finishing across the instance keep re-triggering the crash. Verified live on `2.0.0-rc10`.

## Expected result

A trigger `window` duration that can't form a valid time range should be rejected at save time (validation), or fail just that trigger evaluation. It must never throw an uncaught exception into the queue consumer and shut the whole instance down.

## Repro steps

1. Create an upstream flow:

```yaml
id: upstream
namespace: company.team
tasks:
  - id: hello
    message: hi
```

2. Create the poison flow — it **passes validation and saves cleanly** (`WindowValidator` only checks field mutual-exclusivity, never bounds the duration):

```yaml
id: downstream_sliding
namespace: company.team
tasks:
  - id: noop
    message: triggered
triggers:
  - id: on_upstream
    window:
      lookback: P400000000000D
    dependsOn:
      - namespace: company.team
        flowId: upstream
        states: [SUCCESS]
```

3. Cause **any** execution to terminate (e.g. run `upstream`, or any other flow in the instance).
4. Watch the server: it shuts down. Restart it → it crashes again on the redelivered message.

The same crash occurs via the more common `every` field (`window: { every: P400000000000D }`) or `offset`.

## Actual result

```
java.time.DateTimeException: Invalid value for EpochDay (valid values -365243219162 - 365241780471): 400000020686
	at java.base/java.time.LocalDate.ofEpochDay(...)
	at io.kestra.core.models.triggers.multipleflows.MultipleConditionStateStore.create(MultipleConditionStateStore.java:60)
	at io.kestra.jdbc.runner.AbstractJdbcMultipleConditionStateStore.process(AbstractJdbcMultipleConditionStateStore.java:32)
	at io.kestra.executor.handler.MultipleConditionEventMessageHandler.handle(MultipleConditionEventMessageHandler.java:36)
	at io.kestra.executor.DefaultExecutor.multipleConditionEventQueue(DefaultExecutor.java:490)
...
ERROR io.kestra.queue.AbstractSubscriber [MultipleConditionEvent] fatal error while consuming messages. Initiating application shutdown.
INFO  i.kestra.core.contexts.KestraContext Kestra server - Shutdown completed
```

On restart (poison pill):

```
ERROR ... [MultipleConditionEvent] failed to process message with key 'main_company.team_downstream_sliding_...'. Message will be redelivered. You can ignore this message by starting Kestra with `--ignore-queue-records main_company.team_downstream_sliding_...`.
java.time.DateTimeException: Invalid value for EpochDay ...
ERROR ... [MultipleConditionEvent] fatal error while consuming messages. Initiating application shutdown.
```

## Root cause

- `core/src/main/java/io/kestra/core/models/triggers/multipleflows/MultipleConditionStateStore.java:60` (SLIDING_WINDOW: `now.plus(timeWindow.getWindow())`), and lines 52/55 in the DURATION_WINDOW branch (`now.plus(getWindowAdvance())` for `offset`, `startWindow.plus(window)` for `every`). A large user duration overflows `ZonedDateTime`/`LocalDate.ofEpochDay` → `DateTimeException` (a `RuntimeException`).
- The call chain is entirely unguarded: `MultipleConditionEventMessageHandler.handle` (line 36) invokes `computeExecutionsFromFlowTriggerDependsOn` **outside** its only try/catch (which wraps just the later `emit`); `AbstractJdbcMultipleConditionStateStore.process` runs `create()` inside a JDBC transaction with no try/catch.
- The escaped `DateTimeException` reaches `queue/.../AbstractSubscriber` (`failFast()` defaults true) → full application shutdown, and the message stays on the queue → poison pill. (Same fatal-path family as #18355, different, more easily-hit trigger.)

## Suggested fix

- Bound the `window` duration in `WindowValidator` (reject at save time).
- Wrap `MultipleConditionStateStore.create()` / the window computation and fail just this trigger evaluation on error instead of letting `DateTimeException` escape to the queue consumer.

## Impact

- **Blast radius:** entire instance; and it re-fires on any terminating execution instance-wide, so the instance cannot stay up while this flow exists.
- **Persistence:** poison pill — survives restart; permanent outage until manual `--ignore-queue-records`.
- **Privilege:** any user who can save a flow. Validation does not stop it.

Verified live on `2.0.0-rc10` (commit `b48d8d1`, Docker H2): saved the flow, ran one execution, instance shut down; restart re-crashed on the redelivered message.

### LINKED ISSUE #18421: [Bash2.0] A valid SLA MAX_DURATION with a large duration crashes the whole instance on execution create (Instant overflow poison-pill)
## Summary

A single **valid** flow that declares an SLA of type `MAX_DURATION` with a large `duration` permanently crashes the whole Kestra instance the moment an execution is **created**. The executor computes the SLA monitor deadline as `execution.startDate.plus(sla.duration)`; a large `duration` overflows `Instant` and throws an unchecked `DateTimeException`, which escapes the execution-queue consumer → `AbstractSubscriber` → *"fatal error while consuming messages. Initiating application shutdown."* The instance shuts down, and the execution-create command redelivers on restart → permanent crash-loop. Verified live on `2.0.0-rc10` (Postgres).

SLA is a Kestra 2.0 feature. No task needs to run — the crash fires on execution creation.

## Expected result

`sla.duration` should be bounds-checked (at save time, or when computing the deadline) so it can never overflow the deadline computation. A bad value fails that execution's SLA setup, not the whole instance.

## Repro steps

1. Create this flow (validates and saves cleanly — `MaxDurationSLA.duration` is only `@NotNull`, no bound):

```yaml
id: crash_sla_overflow
namespace: company.team
sla:
  - id: boom
    type: MAX_DURATION
    behavior: FAIL
    duration: PT9000000000000H
tasks:
  - id: hello
    message: hi
```

2. Execute it: `POST /api/v1/main/executions/company.team/crash_sla_overflow`.

`PT9000000000000H` (9×10¹² hours ≈ 3.24×10¹⁶ s) parses as a valid `Duration` (well under `Long.MAX` seconds) so it passes validation, but `startDate + duration` exceeds `Instant.MAX`.

## Actual result

Whole instance shuts down:

```
ERROR queue-postgres_7 io.kestra.queue.AbstractSubscriber [ExecutionCommand] fatal error while consuming messages. Initiating application shutdown.
INFO  i.kestra.core.contexts.KestraContext Kestra server - Shutdown completed
```

On restart (poison pill), the execution-create command redelivers and it crashes again:

```
ERROR ExecutionCommandMessageHandler Unable to process Create command for execution 7khO6mdNR0Yxx...
  at ExecutionCommandMessageHandler.handleCreate(ExecutionCommandMessageHandler.java:157)
ERROR AbstractSubscriber [ExecutionCommand] fatal error while consuming messages. Initiating application shutdown.
```

## Root cause

- `executor/src/main/java/io/kestra/executor/handler/ExecutionEventMessageHandler.java:170`:
  `.deadline(execution.getState().getStartDate().plus(sla.getDuration()))` — `getStartDate()` is `Instant`; a large `Duration` overflows `Instant.plus` → `DateTimeException` (unchecked).
- `core/src/main/java/io/kestra/core/models/flows/sla/types/MaxDurationSLA.java:24-25`: `duration` is `@NotNull` with no bound.
- The SLA-monitor block runs on execution `CREATED` inside the execution-queue consumer; its try/catch handles only `QueueException`/`FlowNotFoundException` (`ExecutionEventMessageHandler.java:388-403`), so the `DateTimeException` escapes → `AbstractSubscriber` fatal shutdown (same fatal-path family as #18355 / #18369).

## Suggested fix

Bound `sla.duration` (validation), and/or guard the deadline computation and fail just this execution instead of letting `DateTimeException` reach the queue consumer.

## Impact

- Whole-instance shutdown from a valid flow, on execution create (no task run needed).
- Poison pill: survives restart → permanent outage until manual `--ignore-queue-records`.
- Any user who can save a flow.

Verified live on `2.0.0-rc10` (commit `b48d8d1`, Docker, Postgres backend). Same defect class (unbounded user `Duration` → `Instant/ZonedDateTime.plus` overflow → unguarded executor consumer) also affects `Pause` duration and task/flow `retry` interval — filed separately.

### LINKED ISSUE #18423: [Bash2.0] A Pause task with a large pauseDuration puts the executor into an infinite retry-storm (Instant overflow livelock)
## Summary

A `Pause` task (a very common flow-control task in Kestra 2.0) with a large `pauseDuration` (or `timeout`) puts the executor into an **infinite retry-storm**: computing the resume deadline does `taskRun.maxDate().plus(pauseDuration)`, which overflows `Instant` and throws an unchecked `DateTimeException`. The execution-event processing fails and is **re-processed endlessly** — every executor thread spins on the same execution forever (I saw all 16 `executor-execution-event-executor` threads looping, the reported per-attempt time climbing 0.1s → 23s+ and still going), flooding logs and pinning CPU. The execution never reaches a terminal state and the executor's throughput collapses. Verified live on `2.0.0-rc10`.

## Expected result

An out-of-range `pauseDuration`/`timeout` fails that one execution cleanly (or is rejected at validation). It must not put the executor into an unbounded reprocessing loop.

## Repro steps

```yaml
id: crash_pause_overflow
namespace: company.team
tasks:
  - id: pause
    pauseDuration: PT9000000000000H
```

1. Save (validates cleanly — `pauseDuration` is only a `Property<Duration>`, no bound).
2. Execute it.

`PT9000000000000H` parses as a valid `Duration` but `maxDate + pauseDuration` exceeds `Instant.MAX`. (`timeout: PT9000000000000H` instead of `pauseDuration` hits the same line via the `timeout` branch.)

## Actual result

The execution never completes; the executor logs, repeating forever across all executor threads:

```
WARN i.k.core.models.executions.Execution [flow: crash_pause_overflow] [execution: ...] Flow failed from executor in 00:00:NN with exception 'Instant exceeds minimum or maximum instant'
java.time.DateTimeException: Instant exceeds minimum or maximum instant
	at java.time.Instant.plusSeconds(...)
	at java.time.Instant.plus(...)
```

The per-attempt elapsed time climbs without bound (0.1s → 23s+ observed) and every executor thread is consumed — an executor-wide livelock / log-and-CPU flood from one execution. The instance stays up but its executor is degraded/saturated.

## Root cause

`executor/src/main/java/io/kestra/executor/ExecutorService.java:939`:
```java
.date(workerTaskResult.getTaskRun().getState().maxDate().plus(duration != null ? duration : timeout))
```
`maxDate()` is `Instant`; `duration`/`timeout` come from the `Pause` task's `Property<Duration>` fields with no bound. The overflow throws `DateTimeException` inside `handlePausedDelay` (called from `executorService.process`); the execution-event consumer treats the processing as failed and reprocesses, so the same overflow recurs every attempt — an infinite loop rather than a single failure.

(Same root defect class — unbounded user `Duration` → `Instant/ZonedDateTime.plus` overflow — as the SLA `MAX_DURATION` bug, which instead causes a full-instance shutdown, and the flow-trigger `window` bug. The task/flow `retry` interval field (`Constant.nextRetryDate`) is a third instance of the same overflow.)

## Suggested fix

Bound `pauseDuration`/`timeout` (validation), and guard the deadline computation so an overflow fails the execution once instead of looping. Apply the same bound to all user `Duration` fields (SLA duration, retry interval, Pause) that feed `Instant.plus`.

## Impact

Executor-wide livelock (retry storm) from a valid config on a super-common task (`Pause`). Instance stays up but executor throughput collapses and logs/CPU flood until the execution is manually killed. Verified live on `2.0.0-rc10` (commit `b48d8d1`, Docker H2).

nkwiatkowski:
## What & why

An unbounded user `Duration` fed into `Instant`/`ZonedDateTime.plus()` overflowed with an unchecked `DateTimeException` that escaped the queue consumer. Under `kestra.queue.fail-fast=true` that shut the whole instance down and left a poison-pill message that crash-looped on restart.

This **replaces the earlier saturation approach**. Saturation silently reinterpreted an out-of-range duration as "effectively forever" (an SLA that never fires, a Pause that waits ~forever), hiding the user's mistake. The new behaviour: **reject bad config where we can (validation), and where we can't, fail the individual execution with a clear error — never crash or stall the executor.**

## How

Two layers, both required:

**Layer A — save-time validation.** A new `@DurationMax` constraint (single global sanity cap, default 10 years / `P3650D`) on the plain-`Duration` fields, rejecting absurd literals with a `422`:
- `MaxDurationSLA.duration`
- retry `Constant.interval`, `Exponential.interval`/`maxInterval`, `Random.minInterval`/`maxInterval`, `AbstractRetry.maxDuration`
- `TimeWindow.window`/`windowAdvance`

**Layer B — runtime guard.** `DateUtils.plusOrThrow(Instant, Duration)` throws a handler-caught `InternalException` on overflow so the single execution fails cleanly (via the existing `handleFailedExecutionFromExecutor` path) and the poison-pill message drains, instead of crashing the executor:
- SLA — `ExecutionEventMessageHandler`
- Pause — `ExecutorService` (`RESUME_FLOW` delay)
- retry arithmetic — `ExecutorService.handleFlowableTasks`
- flow-trigger window — `MultipleConditionEventMessageHandler` wraps the window computation to **log and skip** (no execution to fail)

Layer B is the only fix that covers `Property<Duration>` fields (`Pause`, `LoopUntil`) whose values only exist at runtime, and existing bad flows already sitting on the queue. `fail-fast` stays `true` — this is all handler-level, no queue-layer change.

## Tests
- `DateUtilsTest` — `plusOrThrow` overflow behaviour
- `DurationMaxValidationTest` — `TimeWindow` window/windowAdvance cap + boundary
- `ConstantRetryValidationTest` — retry interval cap boundary
- Executor module compiles; `H2RunnerTest` executor gate green

Closes #18369
Closes #18421
Closes #18423

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/sla/types/MaxDurationSLA.java
+++ b/core/src/main/java/io/kestra/core/models/flows/sla/types/MaxDurationSLA.java
@@ -11,6 +11,7 @@ import io.kestra.core.models.flows.sla.ExecutionMonitoringSLA;
 import io.kestra.core.models.flows.sla.SLA;
 import io.kestra.core.models.flows.sla.Violation;
 import io.kestra.core.runners.RunContext;
+import io.kestra.core.validations.DurationMax;
 
 import jakarta.validation.constraints.NotNull;
 import lombok.Getter;
@@ -22,6 +23,7 @@ import lombok.experimental.SuperBuilder;
 @NoArgsConstructor
 public class MaxDurationSLA extends SLA implements ExecutionMonitoringSLA {
     @NotNull
+    @DurationMax
     private Duration duration;
 
     @Override
--- a/core/src/main/java/io/kestra/core/models/tasks/retrys/AbstractRetry.java
+++ b/core/src/main/java/io/kestra/core/models/tasks/retrys/AbstractRetry.java
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.annotation.JsonTypeInfo;
 
 import dev.failsafe.RetryPolicy;
 import dev.failsafe.RetryPolicyBuilder;
+import io.kestra.core.validations.DurationMax;
 import jakarta.validation.constraints.Min;
 import lombok.Builder;
 import lombok.Getter;
@@ -28,6 +29,7 @@ import lombok.experimental.SuperBuilder;
 public abstract class AbstractRetry {
     abstract public String getType();
 
+    @DurationMax
     private Duration maxDuration;
 
     @Min(1)
--- a/core/src/main/java/io/kestra/core/models/tasks/retrys/Constant.java
+++ b/core/src/main/java/io/kestra/core/models/tasks/retrys/Constant.java
@@ -7,6 +7,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 
 import dev.failsafe.RetryPolicyBuilder;
+import io.kestra.core.validations.DurationMax;
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.constraints.AssertTrue;
 import jakarta.validation.constraints.NotNull;
@@ -26,6 +27,7 @@ public class Constant extends AbstractRetry {
     protected String type = "constant";
 
     @NotNull
+    @DurationMax
     private Duration interval;
 
     @Override
--- a/core/src/main/java/io/kestra/core/models/tasks/retrys/Exponential.java
+++ b/core/src/main/java/io/kestra/core/models/tasks/retrys/Exponential.java
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 
 import dev.failsafe.RetryPolicyBuilder;
+import io.kestra.core.validations.DurationMax;
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.constraints.AssertTrue;
 import jakarta.validation.constraints.NotNull;
@@ -27,9 +28,11 @@ public class Exponential extends AbstractRetry {
     protected String type = "exponential";
 
     @NotNull
+    @DurationMax
     private Duration interval;
 
     @NotNull
+    @DurationMax
     private Duration maxInterval;
 
     private Double delayFactor;
--- a/core/src/main/java/io/kestra/core/models/tasks/retrys/Random.java
+++ b/core/src/main/java/io/kestra/core/models/tasks/retrys/Random.java
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 
 import dev.failsafe.RetryPolicyBuilder;
+import io.kestra.core.validations.DurationMax;
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.constraints.AssertTrue;
 import jakarta.validation.constraints.NotNull;
@@ -27,9 +28,11 @@ public class Random extends AbstractRetry {
     protected String type = "random";
 
     @NotNull
+    @DurationMax
     private Duration minInterval;
 
     @NotNull
+    @DurationMax
     private Duration maxInterval;
 
     @Override
--- a/core/src/main/java/io/kestra/core/models/triggers/TimeWindow.java
+++ b/core/src/main/java/io/kestra/core/models/triggers/TimeWindow.java
@@ -9,6 +9,7 @@ import java.time.temporal.ChronoUnit;
 import org.apache.commons.lang3.tuple.Pair;
 
 import io.kestra.core.models.annotations.PluginProperty;
+import io.kestra.core.validations.DurationMax;
 import io.kestra.core.validations.TimeWindowValidation;
 import io.kestra.core.validations.TimezoneId;
 
@@ -48,6 +49,7 @@ public class TimeWindow {
     )
     @PluginProperty
     @With
+    @DurationMax
     private Duration window;
 
     @Schema(
@@ -60,6 +62,7 @@ public class TimeWindow {
     )
     @PluginProperty
     @With
+    @DurationMax
     private Duration windowAdvance;
 
     @Schema(
--- a/core/src/main/java/io/kestra/core/utils/DateUtils.java
+++ b/core/src/main/java/io/kestra/core/utils/DateUtils.java
@@ -39,6 +39,18 @@ public class DateUtils {
         }
     }
 
+    /**
+     * Adds {@code duration} to {@code instant}, throwing a checked {@link InternalException} when the
+     * result is out of range instead of the unchecked exception raised by {@link Instant#plus}.
+     */
+    public static Instant plusOrThrow(Instant instant, Duration duration) throws InternalException {
+        try {
+            return instant.plus(duration);
+        } catch (ArithmeticException | DateTimeException e) {
+            throw new InternalException("The duration '%s' is out of the supported range when added to '%s'.".formatted(duration, instant), e);
+        }
+    }
+
     public static GroupType groupByType(Duration duration) {
         if (duration.toDays() > GroupValue.MONTH.getValue()) {
             return GroupType.MONTH;
--- a/core/src/main/java/io/kestra/core/validations/DurationMax.java
+++ b/core/src/main/java/io/kestra/core/validations/DurationMax.java
@@ -1,0 +1,25 @@
+package io.kestra.core.validations;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+import io.kestra.core.validations.validator.DurationMaxValidator;
+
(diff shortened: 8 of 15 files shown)


loicmathieu (approved):
LGTM

PR #18551, opened 2026-08-24
fix(core): mask kv() on read-only render paths, fix namespace boundary check

fdelbrayelle:
## Summary
This issue was filed against `kestra-io/kestra-ee#10244`, but its two root causes live here in OSS `kestra`. A companion kestra-ee PR adds the same mask to the EE Cases renderer, which has its own independent mask list — both are needed to fully close the issue.

- `SecureVariableRendererFactory`'s masked-functions list covered only `secret()`, not `kv()`. A caller-supplied expression like `{{ kv('SOME_KEY') }}` on a read-only path (e.g. the execution `eval` action) resolved to the plaintext K/V value — no `KVSTORE` permission needed anywhere on that call path.
- `KVStoreService.isNotParentNamespace` compared namespaces with a plain `String.startsWith` and no dot boundary, so `"development".startsWith("dev")` treated the sibling namespace `dev` as an ancestor and skipped both the cross-namespace ACL check and the namespace-existence check.

Fixes:
- Adds `KvFunction.NAME` to the masked-functions list.
- Requires an exact match or a `parent.` prefix for namespace ancestry, matching the dot-boundary convention every other ACL prefix check in the codebase already uses.

## QA
Real before/after test run:
1. Reverted only the two fixes (kept the new tests) → `./gradlew :core:test --tests SecureVariableRendererFactoryTest --tests KVStoreServiceTest` → **2 of 20 tests fail** (kv() returns the plaintext value; the sibling-namespace access silently succeeds with no exception).
2. Restored the fix → re-ran → **20/20 pass**.

Full report: https://claude.ai/code/artifact/ff7c4cf2-dbcd-418a-b30b-02e1e41969f8

Part of kestra-io/kestra-ee#10244

## Test plan
- [ ] CI build/tests pass
- [x] `./gradlew :core:test --tests SecureVariableRendererFactoryTest --tests KVStoreServiceTest` — 20/20 pass locally
- [x] Added `shouldCreateDebugRendererThatMasksKv` and `shouldNotTreatSiblingNamespaceSharingPrefixAsChild` regression tests
- [x] Confirmed via revert-and-rerun that the new tests catch both vulnerabilities (fail without the fix)

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/SecureVariableRendererFactory.java
+++ b/core/src/main/java/io/kestra/core/runners/SecureVariableRendererFactory.java
@@ -3,6 +3,7 @@ package io.kestra.core.runners;
 import java.util.List;
 
 import io.kestra.core.runners.pebble.PebbleEngineFactory;
+import io.kestra.core.runners.pebble.functions.KvFunction;
 import io.kestra.core.runners.pebble.functions.SecretFunction;
 
 import io.micronaut.context.ApplicationContext;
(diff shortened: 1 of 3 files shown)


fdelbrayelle:
## QA report

Real before/after test run: reverted just the two fixes and re-ran SecureVariableRendererFactoryTest + KVStoreServiceTest — 2/20 tests failed (kv() returned plaintext; sibling namespace access silently succeeded). Restored the fix, re-ran — 20/20 pass.

https://claude.ai/code/artifact/ff7c4cf2-dbcd-418a-b30b-02e1e41969f8

nkwiatkowski (approved):

PR #18554, opened 2026-08-24
fix(design-system): apply custom date-range filter with a single bound set

### LINKED ISSUE #18320: [Bash2.0] Custom date-range picker silently discards typed date, substitutes "now"
**Expected result:** Typing a valid date into the Start/End Date field of the Custom Range filter and clicking OK should apply that typed date.

**Repro steps:**
1. On any Executions list, open the interval filter -> Custom Range.
2. Click Start Date, type `2026-06-01 00:00` (field shows it correctly).
3. Click OK without touching the calendar grid.

**Actual result:** The Start Date field reverts to the current date/time - the typed value is silently discarded, no error or warning shown. Only clicking the exact day cell in the calendar grid works.

fdelbrayelle:
## Summary
- `FilterEditPopper.vue`'s live-apply logic for `time-range`/`timeRange` filters required **both** Start and End dates to be set before emitting an update in Custom Range mode. Setting only a Start Date and clicking OK was a silent no-op — the popover reinitializes its local state from the (unchanged) applied filter next time it opens, so the typed value doesn't even survive as a draft.
- The missing bound now defaults to `new Date()` (now) instead of blocking the apply, matching the "apply that typed date" behavior described in the issue. Switching to Custom Range with *neither* bound touched still applies nothing.
- This is shared by every Custom Range time filter in the design system (Executions, Logs, Flows dashboards, metrics), not just the Executions list from the repro.

Closes #18320

## QA
Reproduced live in a browser against a local Kestra OSS instance (fresh H2 admin, Playwright-driven Chromium) before writing any fix, then proved the fix with a failing→passing Vitest run on the exact same assertion. Full writeup with before/after screenshots and logs: https://claude.ai/code/artifact/07d2df48-a925-4f0e-8f5b-af1503a884c0

## Test plan
- [x] New Vitest tests in `FilterEditPopper.test.ts`: applying with only Start Date set (End defaults to now), and switching to Custom Range with neither bound set still applies nothing
- [x] `npx vitest run --config vitest.unit.config.ts tests/units/Data/KsFilter/FilterEditPopper.test.ts` — 6/6 pass with the fix, 1/6 fails without it (revert-and-rerun verified)
- [x] Live browser reproduction before and after the fix (Playwright against local dev server)

### DIFF
--- a/ui/packages/design-system/src/components/Data/KsDataTable/filter/layout/FilterEditPopper.vue
+++ b/ui/packages/design-system/src/components/Data/KsDataTable/filter/layout/FilterEditPopper.vue
@@ -294,12 +294,11 @@
             return {value: state.textValue, label: state.textValue}
         case "select":
             if (props.filterKey?.key === "timeRange" && state.timeRangeMode === "custom") {
+                const startDate = state.startDateValue ?? new Date()
+                const endDate = state.endDateValue ?? new Date()
                 return {
-                    value: {
-                        startDate: state.startDateValue!,
-                        endDate: state.endDateValue!,
-                    },
-                    label: `${state.startDateValue!.toLocaleDateString()} - ${state.endDateValue!.toLocaleDateString()}`,
+                    value: {startDate, endDate},
+                    label: `${startDate.toLocaleDateString()} - ${endDate.toLocaleDateString()}`,
                     meta: state.dateFilterMode ? {dateFilter: state.dateFilterMode} : undefined,
                 }
             }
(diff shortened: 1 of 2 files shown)


fdelbrayelle:
**QA evidence:** https://claude.ai/code/artifact/07d2df48-a925-4f0e-8f5b-af1503a884c0

Live-browser reproduction (Playwright against a local Kestra OSS dev server) of the reported bug before any fix, root cause in `FilterEditPopper.vue`, the fix, live-browser re-verification after the fix, and a Vitest unit test that fails without the fix and passes with it — before/after screenshots and logs included.

nkwiatkowski (approved):

Issue #18365, opened 2026-08-21
[Bash2.0] Pebble nindent/indent filter has no upper bound on amount → unbounded allocation / OOM crash


fdelbrayelle:
## Summary

The Pebble `indent` / `nindent` filter takes a user-controlled `amount` and does `prefix.repeat(amount)` with **no upper bound**. A single ordinary task expression can therefore allocate an arbitrarily large string and exhaust the JVM heap. On a normally-sized deployment this OOM-kills the worker (or, in a single-JVM deployment, the whole instance). Verified on `2.0.0-rc10`.

## Expected result

`amount` is bounded to a sane maximum; an out-of-range value fails that one task with a clear error. A user expression must not be able to allocate multi-GB strings and OOM the process.

## Repro steps

```yaml
id: oom_nindent
namespace: company.team
tasks:
  - id: boom
    type: io.kestra.plugin.core.debug.Return
    format: "{{ 'x' | nindent(2000000000) }}"
```

Execute it and watch the JVM/container memory.

## Actual result

On my box (container with ~8 GB `MaxHeapSize`, 30 GB host) executing this single task drove memory from **1.8 GB to 17.5 GB** in a few seconds at 300%+ CPU:

```
t=2s  RUNNING  mem=11.41GiB  cpu=306%
t=4s  RUNNING  mem=14.32GiB
t=6s  RUNNING  mem=17.56GiB  cpu=364%
t=8s  FAILED   mem=17.58GiB
```

Here it survived only because the host had 30 GB free and the ~2 GB output was ultimately rejected by the gRPC max-inbound-message-size cap **after** the string was already built (`Return` had even logged `SUCCESS`):

```
ERROR ... Failed to send the task result to the worker controller: its serialized size exceeds the maximum inbound gRPC message size ...
```

Memory stayed pinned at 17.5 GB afterwards (not reclaimed until restart). On a normal deployment (e.g. 2–4 GB heap), or with two such tasks running concurrently, the allocation OOM-kills the JVM before any downstream cap can catch it — i.e. an instance crash triggerable by any user who can run a flow. (For comparison, the `range(a,b)` function *is* bounded and fails cleanly at large sizes; `indent`/`nindent` is not.)

## Root cause

[...]

PR #18555, opened 2026-08-24
fix(core): bound Pebble's range() function to prevent heap exhaustion

### LINKED ISSUE #18418: [Bash2.0] Execution eval endpoint OOMs the webserver on an unbounded range() expression
## Summary

`POST /api/v1/{tenant}/executions/{executionId}/actions/eval` (the debug "evaluate a Pebble expression against this execution" endpoint) evaluates the expression with **no allocation bound**. A body like `{{ range(0, 2000000000) }}` materializes ~2 billion objects and exhausts the JVM heap — the **webserver** returns 500 `Java heap space` and heap stays consumed (~16 GB here). On a normally-sized deployment this OOM-kills the webserver. Verified on `2.0.0-rc10`.

(Related but distinct endpoint: a render-OOM was reported for `POST /expressions/render`. This is a different controller method — `/executions/{id}/actions/eval` — that evaluates against a live execution context; a fix to the render endpoint would not automatically cover this one. Noting the relationship.)

## Expected result

The eval endpoint bounds the work / output size (as ordinary Pebble errors are already handled — `{{ 1/0 }}` returns `200` with an `error` message). An unbounded `range()`/loop should fail that one request with a bounded error, not OOM the server.

## Repro steps

```bash
# get any execution id
EID=$(curl -s -u admin@kestra.io:Kestra42! \
  'http://localhost:8080/api/v1/main/executions/search?size=1' | python3 -c 'import sys,json;print(json.load(sys.stdin)["results"][0]["id"])')

  "http://localhost:8080/api/v1/main/executions/$EID/actions/eval" \
  -H 'Content-Type: text/plain' \
  --data-raw '{{ range(0,2000000000) }}'
```

## Actual result

HTTP 500:

```
{"message":"Internal server error: Java heap space", ...}
```

Server log: `java.lang.OutOfMemoryError: Java heap space`. Container memory jumped to ~16 GB and stayed there. (For comparison, `{{ 1/0 }}` and `{{ notafunc() }}` return `200` with an `error` field — Pebble *errors* are caught, but heap *allocation* is not bounded.)

## Root cause

`webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java` (`evalExpression`, `@Post "/{executionId}/actions/eval"`, line ~383) renders the user expression with no limit on the size of the produced structure; `range(a,b)` eagerly builds `b-a` elements. The `OutOfMemoryError` is not (and cannot cleanly be) caught, so it surfaces as a 500 and destabilizes the webserver heap.

## Suggested fix

Bound `range()` / loop output size in the rendering path, and/or cap the eval result size, so a single eval request cannot exhaust the heap. Apply the same bound across all expression-evaluation endpoints (eval, render).

## Impact

Any authenticated user can OOM the webserver via the debug eval endpoint; on a normal heap this is an instance crash. Verified live on `2.0.0-rc10` (commit `b48d8d1`, Docker H2).

fdelbrayelle:
## Summary
- Pebble's built-in `range()` (`io.pebbletemplates.pebble.extension.core.RangeFunction`) eagerly materializes an `ArrayList` of every element between `start` and `end` before returning — `{{ range(0, 2000000000) }}` allocates ~2 billion boxed entries and exhausts the JVM heap. Any endpoint that renders a caller-supplied Pebble expression against a real `RunContext` is exposed, including the execution `eval` debug endpoint (`POST /executions/{id}/actions/eval`) called out in the issue.
- Added `BoundedRangeFunction`, registered in Kestra's `Extension` bean under the same `range` key. Kestra's function map is merged over Pebble's built-in `CoreExtension` (last write wins), so this override applies everywhere Pebble renders, not just the reported endpoint.
- It computes the requested element count from `start`/`end`/`increment` *before* delegating to Pebble's real `RangeFunction`, and throws a bounded `PebbleException` once it would exceed 1,000,000 elements — surfaced by `VariableRenderer` as a normal `IllegalVariableEvaluationException`/render error, exactly like the existing `{{ 1/0 }}` handling the issue points to as the expected behavior.

Closes #18418

## Out of scope
The issue also flags a related-but-distinct render-OOM report for `POST /expressions/render` and a sibling nindent/indent unbounded-allocation issue (#18365) — both are separate code paths, not touched here.

## QA
Full writeup with before/after test logs: https://claude.ai/code/artifact/843ea13d-8485-455c-a925-8aeed3611b74

## Test plan
- [x] New `BoundedRangeFunctionTest`: normal small range still renders correctly, both ascending and descending oversized ranges throw
- [x] `./gradlew :core:test --tests io.kestra.core.runners.pebble.functions.BoundedRangeFunctionTest` — 3/3 pass with the fix, 2/3 fail without it (revert-and-rerun verified)
- [x] Checked for existing tests calling Pebble's `range()` — none found (unrelated `IntStream.range` matches only)

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
+++ b/core/src/main/java/io/kestra/core/runners/VariableRenderer.java
@@ -104,13 +104,14 @@ public class VariableRenderer {
             PebbleTemplate compiledTemplate = this.pebbleEngine().getLiteralTemplate((String) result);
 
             try {
-                OutputWriter writer = stringify ? new JsonWriter() : new TypedObjectWriter();
+                long maxOutputSize = this.variableConfiguration.getMaxOutputSize();
+                OutputWriter writer = stringify ? new JsonWriter(maxOutputSize) : new TypedObjectWriter(maxOutputSize);
                 compiledTemplate.evaluate(writer, variables);
                 result = writer.output();
             } catch (IllegalArgumentException e) {
                 //can happen in case of mixed type in string
                 if (!stringify) {
-                    JsonWriter fallbackWriter = new JsonWriter();
+                    JsonWriter fallbackWriter = new JsonWriter(this.variableConfiguration.getMaxOutputSize());
                     compiledTemplate.evaluate(fallbackWriter, variables);
                     Object rendered = fallbackWriter.output();
 
--- a/core/src/main/java/io/kestra/core/runners/configuration/VariableConfiguration.java
+++ b/core/src/main/java/io/kestra/core/runners/configuration/VariableConfiguration.java
@@ -9,12 +9,14 @@ import lombok.Getter;
 @ConfigurationProperties("kestra.variables")
 public class VariableConfiguration {
     public static final int DEFAULT_MAX_RENDER_DEPTH = 16;
+    public static final long DEFAULT_MAX_OUTPUT_SIZE = 10 * 1024 * 1024L; // 10 MiB
 
     public VariableConfiguration() {
         this.cacheEnabled = true;
         this.cacheSize = 1000;
         this.recursiveRendering = false;
         this.maxRenderDepth = DEFAULT_MAX_RENDER_DEPTH;
+        this.maxOutputSize = DEFAULT_MAX_OUTPUT_SIZE;
         this.redactedEnvVars = List.of(
             "KESTRA_PLUGINS_PATH", "KESTRA_CONFIGURATION_PATH", "KESTRA_CONFIGURATION", "KESTRA_JAVA_OPTS"
         );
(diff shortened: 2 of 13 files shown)


fdelbrayelle:
**QA evidence:** https://claude.ai/code/artifact/843ea13d-8485-455c-a925-8aeed3611b74

Root cause traced by decompiling Pebble's `RangeFunction`/`ExtensionRegistry` to confirm the eager-allocation behavior and the function-override merge semantics, then a Gradle unit test that fails without the fix (unbounded, no exception) and passes with it (bounded `PebbleException`) — before/after logs included.

loicmathieu (approved):
LGTM

PR #18556, opened 2026-08-24
fix(dashboards): treat null filters as empty on chart-data endpoint

### LINKED ISSUE #18366: [Bash2.0] Dashboard chart-data returns 500 NPE when request body has filters:null
## Summary

`POST /api/v1/{tenant}/dashboards/{id}/charts/{chartId}` (chart-data) returns HTTP 500 with a `NullPointerException` when the request body contains `"filters": null`. The sibling `charts/preview` path guards this; the chart-data path does not. Verified on `2.0.0-rc10`.

## Expected result

A `null` (or omitted) `filters` is treated as "no filters" and the chart data is returned, or a `422` is returned — not a 500.

## Repro steps

```bash
  'http://localhost:8080/api/v1/main/dashboards/_default/charts/x' \
  -H 'Content-Type: application/json' \
  -d '{"filters":null}'
```

(`_default` is the built-in dashboard, so no dashboard row is needed.)

## Actual result

HTTP 500:

```
Internal server error: Cannot invoke "java.util.List.stream()" because "filters" is null
```

## Root cause

`webserver/src/main/java/io/kestra/webserver/controllers/api/DashboardController.java` chart-data path calls `TimeLineSearch.extractFrom(filters)` unconditionally, which does `filters.stream()` on a null list. The `charts/preview` path handles the null case; chart-data does not.

## Suggested fix

Null-guard `filters` (treat null as empty) before `extractFrom`, matching the `preview` path.

## Impact

Confirmed 500 on a core dashboard endpoint. Verified on `2.0.0-rc10` (commit `b48d8d1`, Docker H2).

fdelbrayelle:
## Summary
- `POST /dashboards/{id}/charts/{chartId}` called `TimeLineSearch.extractFrom(filters)` unconditionally in `buildDashboardChardDataQuery`, which does `filters.stream()` and NPEs (500) when the request body has `"filters": null`. The sibling `charts/preview` path (`buildChartPreviewDataQuery`) already guards `filters != null` before calling `extractFrom`.
- Fixed by passing `filters != null ? filters : List.of()` — `TimeLineSearch.extractFrom` already treats "no date filters present" as a first-class case (it defaults `startDate` to 8 days ago when nothing sets it), so this lets that existing default-timeline logic apply naturally instead of duplicating a second hardcoded default.

Closes #18366

## QA
Full writeup, including a test-construction gotcha caught and corrected mid-QA: https://claude.ai/code/artifact/14b54014-c2c8-4658-9464-c13c89e7ce8b

## Test plan
- [x] New `DashboardControllerTest.shouldTreatNullFiltersAsNoFiltersOnChartData` — sends a raw `{"filters":null}` JSON body (matching the issue's curl repro exactly; a builder-constructed request would drop the null key under Jackson's `NON_NULL` inclusion and never exercise the bug)
- [x] `./gradlew :webserver:test --tests io.kestra.webserver.controllers.api.DashboardControllerTest` — 20/20 pass with the fix, exact reported NPE reproduced without it (revert-and-rerun verified)

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/DashboardController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/DashboardController.java
@@ -337,7 +337,7 @@ public class DashboardController {
             return null;
         }
 
-        TimeLineSearch timeLineSearch = TimeLineSearch.extractFrom(filters);
+        TimeLineSearch timeLineSearch = TimeLineSearch.extractFrom(filters != null ? filters : List.of());
         validateTimeline(timeLineSearch.getStartDate(), timeLineSearch.getEndDate());
 
         ZonedDateTime endDate = timeLineSearch.getEndDate();
--- a/webserver/src/test/java/io/kestra/webserver/controllers/api/DashboardControllerTest.java
+++ b/webserver/src/test/java/io/kestra/webserver/controllers/api/DashboardControllerTest.java
@@ -877,6 +877,18 @@ class DashboardControllerTest {
     }
 
     @Test
+    void shouldTreatNullFiltersAsNoFiltersOnChartData() {
+        // Raw JSON body with an explicit "filters":null, matching the reported repro exactly -
+        // going through ChartFiltersOverrides' builder would drop the null key (NON_NULL inclusion)
+        // and never exercise the bug.
+        List<Map> chartData = client.toBlocking().retrieve(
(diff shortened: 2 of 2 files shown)


fdelbrayelle:
**QA evidence:** https://claude.ai/code/artifact/14b54014-c2c8-4658-9464-c13c89e7ce8b

Before/after Gradle test run reproducing the exact reported NPE and confirming the fix, plus a note on a test-construction mistake caught and corrected mid-QA (a builder-based request body silently didn't exercise the bug at all).

ammeek (approved):

PR #18557, opened 2026-08-24
fix(core): enforce allowed-paths on host paths given to FILE inputs

AcevedoR:
A `FILE` input given a plain host path (`d: /etc/passwd`) read it off the server and copied it into internal storage with no check, while `read('file:///etc/passwd')` was correctly refused — so the input was a way around `kestra.local-files.allowed-paths`. Reported against 2.0.0-rc10.

`parseType`'s `case FILE` fell through to `storageInterface.from(..., new File(value))` without ever consulting `LocalPath`, which owns that check. It now goes through `LocalPathFactory`, using a new `LocalPath.authorizedPath()` so the path we read is the one that was authorized — otherwise a symlink could be re-pointed between check and read.

⚠️ **Behaviour change:** `kestra flow test --inputs file=/path/x.csv` now needs that path in `allowed-paths`.

Three regression tests, all failing on `develop`: plain host path, symlink escaping an allowed directory, and the missing-file message. The `kestra://` pass-through is already covered by `InputsTest.fileInputWithFileDefault`, so no test was added for it.

Fixes kestra-io/kestra-ee#10308 (different repo, so it needs closing by hand).

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowInputOutput.java
@@ -4,6 +4,8 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.URI;
+import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDate;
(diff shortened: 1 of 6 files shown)


loicmathieu (approved):
LGTM

PR #18558, opened 2026-08-24
fix(core): reject legacy flat filter parameters instead of ignoring them

AcevedoR:
Closes kestra-io/kestra-ee#10326

### What

Only `filters[field][OP]=value` is read by the query-filter binder. The pre-2.0 flat parameters (`state=RUNNING`, `namespace=x`, `q=*`) were dropped silently, so the query degraded to *match everything*. On the by-query bulk endpoints — kill, delete, restart, change-status, set-labels — a request the caller believed was scoped to one namespace or one state ran against **every execution in the tenant**.

They are now rejected with `422` instead:

```
GET /api/v1/main/executions/search?state=RUNNING
→ 422 Illegal argument: Legacy filter parameter(s) [state] are no longer supported and would
  have been silently ignored. Use the bracket format instead, e.g. filters[state][EQUALS]=value.
```

### How

`QueryFilterFormatBinder` is the single choke point for every filter-accepting endpoint in OSS *and* EE (EE declares no binder of its own), so the check lives there and no controller is touched. A query parameter is rejected when all three hold:

1. it is not in `filters[...]` form;
2. its name is a `QueryFilter.Field` value;
3. the matched route does not declare it as a real argument (`MethodBasedRouteMatch#getRequiredArguments`, plus any `@QueryValue("alias")`).

Condition 3 is what keeps this safe. Several routes legitimately declare parameters that collide with filter-field names — `IAMBindingController.searchBindings` takes `type`, `id` and `namespace`; `NamespaceController.searchNamespaces` takes `existingOnly` (bound as `existing`) — and they keep working. A hand-written denylist would have broken all of them.

The rejected set is derived from `QueryFilter.Field` rather than hand-listed, so it cannot drift as fields are added or removed. When the route arguments cannot be read at all, the check is skipped rather than guessing.

### Escape hatch

```yaml
kestra:
  webserver:
    query-filter:
      reject-legacy-params: false   # default: true
```

For an operator with a stuck integration to buy time on upgrade. Expected to stay enabled.

### Behaviour change

`FlowControllerTest.searchFlowsAll` asserted that `?q=*` still returned every flow "with the old parameters". It passed because the parameter was ignored and the search matched everything — not because `q` was honoured. It now asserts the rejection.

### ⚠️ Companion EE change required

The same pattern is encoded in EE tests, which will fail until they are updated (each of these currently passes only *because* the parameter is ignored):

- `webserver-ee/.../FlowControllerTest.java` — `flows/search?q=*`, and `flows/{enable,disable}/by-query?namespace=…`, which today enables/disables **every flow in the tenant**
- `webserver-ee/.../IAMBindingControllerTest.java` — `bindings/search?q=*`
- `webserver-ee/.../IAMGroupControllerTest.java` — `groups/{id}/members?q=*`
- `webserver-ee/.../ImpersonateTest.java` — `flows/search?q=*`

Verified unaffected: namespace-file search, flow source search, apps preview, promotion flow-source, testsuite search, policy detail — those routes either declare the parameter or take no filters at all.

### Tests

- `QueryFilterFormatBinderTest` — 4 unit tests over the rejection rule (flags legacy, spares route-declared, spares bracket/unknown params, set derived from `Field`)
- `ExecutionControllerTest.shouldRejectLegacyFlatFilterParams` — end-to-end 422 on `?state=RUNNING&namespace=…`, bracket format still 200
- `NamespaceControllerTest.shouldKeepRouteDeclaredParamsWorkingAlongsideFilters` — a route-declared param on a filters endpoint still works
- Full `:webserver:test` green (1183 passing)

### DIFF
--- a/cli/src/main/java/io/kestra/cli/commands/flows/FlowExportCommand.java
+++ b/cli/src/main/java/io/kestra/cli/commands/flows/FlowExportCommand.java
@@ -1,5 +1,7 @@
 package io.kestra.cli.commands.flows;
 
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
@@ -40,7 +42,11 @@ public class FlowExportCommand extends AbstractApiCommand {
 
         try (DefaultHttpClient client = client()) {
             MutableHttpRequest<Object> request = HttpRequest
-                .GET(apiUri("/flows/export/by-query", tenantService.getTenantId(tenantId)) + (namespace != null ? "?namespace=" + namespace : ""))
+                // Bracket format: the flat `namespace=` param was never read by the endpoint, so `--namespace` silently
+                // exported every namespace in the tenant, and is now rejected outright (kestra-io/kestra-ee#10326).
+                // PREFIX keeps the documented meaning of the option - the namespace and its children.
+                .GET(apiUri("/flows/export/by-query", tenantService.getTenantId(tenantId))
+                    + (namespace != null ? "?filters[namespace][PREFIX]=" + URLEncoder.encode(namespace, StandardCharsets.UTF_8) : ""))
                 .accept(MediaType.APPLICATION_OCTET_STREAM);
 
             HttpResponse<byte[]> response = client.toBlocking().exchange(this.requestOptions(request), byte[].class);
--- a/ui/src/components/executions/Executions.vue
+++ b/ui/src/components/executions/Executions.vue
@@ -450,7 +450,7 @@
     import TriggerFlow from "../../components/flows/TriggerFlow.vue"
     import TriggerAvatar from "../../components/flows/TriggerAvatar.vue"
 
-    import {filterValidLabels, keepSupportedFilters, FILTER_FIELD_PATTERN} from "./utils"
+    import {filterValidLabels, keepSupportedFilters, onlyBracketFilters, FILTER_FIELD_PATTERN} from "./utils"
     import {useToast} from "../../utils/toast"
     import {storageKeys} from "../../utils/constants"
     import * as Utils from "../../utils/utils"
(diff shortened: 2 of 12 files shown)


AcevedoR:
### Review round applied (3 commits)

**Binder**
- An explicit `@QueryValue("alias")` now **replaces** the argument name in the declared set instead of adding to it. `NamespaceController.searchNamespaces` declares `@QueryValue(value = "existing") Boolean existingOnly` and `existingOnly` is itself a filter field name — Micronaut never reads it, so sparing it spared exactly the silent widening this guard exists to close. Covered by a test.
- The remediation hint no longer hardcodes `EQUALS`, which several fields don't support (`tags` is `IN/NOT_IN/PREFIX/CONTAINS/STARTS_WITH/ENDS_WITH`) and which would have sent the caller into a second rejection.
- Corrected the javadoc: Micronaut excludes arguments already satisfied by a URI variable, which only makes the check stricter.

**Two more live instances of the bug, found by turning the guard on**

- `kestra flow export --namespace X` built `/flows/export/by-query?namespace=X`. The option was dropped, so the command exported **every flow in the tenant**. Now `filters[namespace][PREFIX]`. `FlowExportCommandTest` was passing *on the widened result* — it asserts "3 flows, or 4 when run in a suite", the fourth being a flow from another namespace that leaked in. This is what the first CI run caught.
- The onboarding tour navigated to the executions list with flat `namespace=`/`scope=`, so the tour list was never scoped to the tour namespace. Now bracket format.

**UI**

`exportExecutionsAsCSV` forwarded the raw route query to `/executions/export/by-query/csv` on plain axios, so a bookmarked pre-2.0 URL turned "Export CSV" into a 422. It now sends only bracket-format keys (`onlyBracketFilters`, with unit tests).

Checked and clean: the executions list and every by-query bulk action go through `routeQueryToQueryFilters`, which drops flat keys client-side, so they never reach the API and are unaffected.

PR #18559, opened 2026-08-24
fix(flows): make the flows table row navigable on single click

### LINKED ISSUE #18319: [Bash2.0] Flows table row shows pointer cursor but clicking empty row area does nothing
**Expected result:** Since each Flows-table row has a pointer cursor (the standard "click anywhere in this row to open it" affordance), clicking anywhere in the row should navigate to the flow.

**Repro steps:**
1. Go to Flows (`/ui/main/flows`).
2. Hover over the empty "Execution statistics" / "Last execution date" area of a row with no execution history (cursor shows pointer).
3. Click there.

**Actual result:** Nothing happens - no navigation, no feedback. Only the flow-id link, namespace link, label chip, and Execute button actually respond to clicks; the rest of the row is a dead click zone despite the pointer cursor.

fdelbrayelle:
## Summary
- The flows table row has `cursor: pointer` on the whole row (the standard "click anywhere to open" affordance), but navigation only fired on `@row-dblclick`. A single click over any non-interactive area — e.g. the empty execution-statistics/last-execution-date cells of a flow with no run history — was a complete no-op.
- `KsDataTable` forwarded `KsTable`'s `row-dblclick` event but never `row-click`; added that pass-through and wired `Flows.vue` to it.
- Every other still-interactive cell in the row (selection checkbox, namespace link, chart drill-down, Execute button) now guards against also triggering row navigation, checked against Element Plus's actual `handleClick` source (it fires `row-click` unconditionally for any cell, checkbox included).

Closes #18319

## QA
Reproduced live against a real flow with no execution history (created via the API to work around an unrelated pre-existing 403 on this local dev instance's browser session), confirmed the single click now navigates, and checked the checkbox/namespace-link/Execute-button don't regress. Full writeup: https://claude.ai/code/artifact/f558f435-53c8-4d05-b1f2-500d877ad41e

## Test plan
- [x] New `KsDataTable` test: `row-click` is forwarded when a table row is clicked
- [x] `npx vitest run --config vitest.unit.config.ts tests/units/Data/KsDataTable.test.ts` — 41/41 pass with the fix, 1/41 fails without it (revert-and-rerun verified)
- [x] Live browser: clicking the empty execution-statistics cell of a real flow with no history navigates to its edit page
- [x] Live browser: clicking the row's selection checkbox still selects it for bulk actions ("1 selected"), does not navigate away

### DIFF
--- a/ui/packages/design-system/src/components/Data/KsDataTable/KsDataTable.vue
+++ b/ui/packages/design-system/src/components/Data/KsDataTable/KsDataTable.vue
@@ -46,6 +46,7 @@
                     @selection-change="selectionChanged"
                     @select="onSelect"
                     @sort-change="onSortChange"
+                    @row-click="(row, column, event) => emit('row-click', row, column, event)"
                     @row-dblclick="(row, column, event) => emit('row-dblclick', row, column, event)"
                 >
                     <KsTableColumn v-if="selectable && showSelection" type="selection" reserveSelection :selectable="rowSelectable" />
@@ -143,6 +144,7 @@
         "update:pageSize": [size: number]
         "sort-change": [sort: SortItem]
         "selection-change": [selection: any[]]
+        "row-click": [row: any, column: any, event: Event]
         "row-dblclick": [row: any, column: any, event: Event]
         "ready": []
         "loaded": []
--- a/ui/packages/design-system/tests/units/Data/KsDataTable.test.ts
+++ b/ui/packages/design-system/tests/units/Data/KsDataTable.test.ts
@@ -5,6 +5,7 @@ import KestraDesignSystem from "../../../src/index"
 import KsDataTable from "../../../src/components/Data/KsDataTable/KsDataTable.vue"
 import KsBulkSelect from "../../../src/components/Data/KsDataTable/KsBulkSelect.vue"
 import KsTableColumn from "../../../src/components/Data/KsTable/KsTableColumn.vue"
+import KsTable from "../../../src/components/Data/KsTable/KsTable.vue"
 
 const globalConfig = {plugins: [createI18n({legacy: false, locale: "en"}), KestraDesignSystem]}
 
@@ -37,19 +38,17 @@ describe("KsDataTable", () => {
         expect(wrapper.find(".kel-table").exists()).toBe(true)
     })
 
-    test("emits row-click when a table row is clicked", async () => {
+    test("forwards row-click from the underlying table", () => {
         const wrapper = mount(KsDataTable, {
             props: {data: SAMPLE_DATA, total: 3},
-            slots: {
-                default: "<ks-table-column prop=\"id\" label=\"ID\" />",
-            },
-            global: {...globalConfig, components: {KsTableColumn}},
+            global: globalConfig,
         })
 
-        await wrapper.find(".kel-table__row").trigger("click")
+        const column = {type: "default", property: "id"}
+        const event = new MouseEvent("click")
+        wrapper.findComponent(KsTable).vm.$emit("rowClick", SAMPLE_DATA[0], column, event)
 
-        expect(wrapper.emitted("row-click")).toBeTruthy()
-        expect(wrapper.emitted("row-click")?.[0]?.[0]).toMatchObject({id: "flow-001"})
+        expect(wrapper.emitted("row-click")?.[0]).toEqual([SAMPLE_DATA[0], column, event])
     })
 
     test("does not render pagination when total is 0", () => {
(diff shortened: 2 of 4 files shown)


fdelbrayelle:
**QA evidence:** https://claude.ai/code/artifact/f558f435-53c8-4d05-b1f2-500d877ad41e

Live-browser reproduction against a real flow with no execution history, root cause in `KsDataTable.vue` (row-click never forwarded, only row-dblclick), the fix, live re-verification, and a Vitest unit test that fails without the fix and passes with it. Also includes a regression check confirming the selection checkbox still selects instead of navigating away.

flcarre (commented):
Pulled this and drove it locally. The core change does what it says: the dead zone in a row is navigable now, and I confirmed the case from #18319 is genuinely fixed.

### One thing to fix: modifier-clicks are hijacked

`RouterLink`'s handler returns early on a cmd/ctrl-click *without* calling `preventDefault` (that is how the browser gets to open the new tab), so the click still bubbles to the `<tr>` and `onRowClick` fires.

Verified on the branch: cmd-clicking the flow-ID link at `Flows.vue:99` leaves `defaultPrevented === false`, so the browser opens its new tab, *and* the current tab navigates to `/flows/edit/<ns>/<id>/edit`. You get the new tab and lose your place in the list at the same time. Under `@row-dblclick` this could not happen, so it is new here.

Two related gaps: `TriggerAvatar` at `:237` has no guard, so clicking the icon to dismiss its own popover leaves the page; and the EE-injected `extensionColumn` cells at `:262` cannot be guarded from this repo at all, since EE has no way to know it must `.stop`.

A single guard in the handler covers all three and lets the per-cell `.stop` wrappers go away:

```ts
const onRowClick = (item: any, column: any, event: MouseEvent) => {
    if (column?.type === "selection") return
    if (event.metaKey || event.ctrlKey || event.shiftKey || event.altKey || event.button !== 0) return
    if ((event.target as HTMLElement)?.closest("a, button, input, canvas, [role='button']")) return
    router.push({...})
}
```

### On the chart-cell guard at `:216-218`: I went looking for a problem and there isn't one

Recording this so nobody else spends time on it. I had wondered whether a flow with executions older than the chart's 24h window would get its dead zone back, since `lastExecutionByFlowReady && getLastExecution(row)` reads a different data source from the one `TimeSeries` renders on (`generated.total > 0`).

Measured on the branch, with a flow that has no executions in the window: `TimeSeries` renders nothing, the wrapper `div` collapses to 0px tall (its `innerHTML` is just `<!--v-if-->`), and a hit-test at the centre of the cell lands on the `td`, not the wrapper. The click reaches the row and navigates. So the `stopPropagation` is simply unreachable in that state and cannot recreate the dead zone.

If you keep the wrapper, a plain `@click.stop` would match the sibling `state.startDate` / `state.current` columns and get the control flow out of the template.

### Minor

The new `KsDataTable` test selects on `.kel-table__row`. Asserting the forwarding at the component seam instead would survive the next Element Plus bump, and it is closer to what the test is actually about.

flcarre (approved):
Pulled `5535db2466` and drove it locally against a fresh `develop` backend. The modifier-click hijack I reported is gone, and the per-cell `.stop` wrappers that were removed do not regress. Approving.

### What I measured

Handler isolated by suppressing the anchor's own default action, so any navigation observed comes from `onRowClick`:

| Click on the flows row | Result |
|---|---|
| plain click, dead zone | navigates to the flow, so the fix for https://github.com/kestra-io/kestra/issues/18319 still works |
| cmd / ctrl / shift / alt-click on the ID link | stays on `/flows` |
| middle click | stays on `/flows` |

Real clicks, per cell:

- ID link, plain click: exactly **one** `router.push` (instrumented on `history.pushState`), so the row handler does not double-fire
- namespace link: goes to `/namespaces/edit/tutorial/overview`, not to the flow
- label tag: stays put, the `.prevent.stop` on `Labels` still holds
- selection checkbox: row checks, bulk bar shows `1 selected`, no navigation
- Play button: execute modal opens, no navigation
- trigger icon: `Trigger details` popover opens, no navigation
- chart canvas: drill-down drawer opens, no navigation
- **chart cell of a flow with no history**: navigates, so the fall-through the wrapper `<div>` was aiming for is now covered by `closest("canvas")`

`TriggerAvatar` is shared with the executions list, and row double-click still navigates there, so the added `.stop` does not break it.

`check:types`, eslint on the four touched files, and the 41 `KsDataTable` unit tests all pass, and the diff has no design-system violations.

### Two non-blocking notes

**The guard itself has no test.** The `KsDataTable` test was rewritten into a re-emit assertion (`findComponent(KsTable).vm.$emit("rowClick", ...)`), which is the pass-through case `ui/AGENTS.md` asks us not to test. The test worth having is the one on the guard: a modifier-click on the ID cell must not push a route. It would fail on `7fcf8df041` and pass on `5535db2466`, which is the bar AGENTS.md sets for a bug fix.

**`TriggerAvatar` gained a `@click.stop`** even though the point of a single guard was to let the per-cell wrappers go. Giving the trigger icon a `role="button"` would fold it into the existing `NON_NAVIGATING_TARGETS` selector, keeping the rule in one place and fixing the accessibility of an interactive element at the same time.

PR #18561, opened 2026-08-24
fix(executions): fail fast on follow-dependencies for an unknown execution

### LINKED ISSUE #18358: [Bash2.0] follow-dependencies pins an IO thread for 10s + 500 on unknown execution (webserver DoS)
## Summary

`GET /api/v1/{tenant}/executions/{executionId}/follow-dependencies` blocks a server IO thread for a full **10 seconds** and then returns HTTP 500 whenever the `executionId` does not exist. Because it pins a `TaskExecutors.IO` thread synchronously (before returning the reactive stream), repeatedly calling it with bogus IDs cheaply exhausts the IO thread pool — a low-effort denial-of-service against the webserver. Verified on `2.0.0-rc10`.

## Expected result

An unknown execution id returns `404` (or an empty stream) immediately, without blocking a server thread.

## Repro steps

1. Call the endpoint with any non-existent execution id:

```bash
curl -sS -u admin@kestra.io:Kestra42! --max-time 15 \
  'http://localhost:8280/api/v1/main/executions/does-not-exist/follow-dependencies'
```

2. Observe it hangs ~10s, then returns 500.
3. Fire several such requests concurrently to see IO-thread starvation.

## Actual result

- Response time ~10.1s, then HTTP 500.
- Server log:

```
ERROR io-executor-thread-6 io.kestra.webserver.controllers.ErrorController Server error: Lambda expression in io.kestra.webserver.controllers.api.ExecutionController expected the predicate to return <true> but was null within 10 seconds.
org.awaitility.core.ConditionTimeoutException: ... expected the predicate to return <true> but was null within 10 seconds.
```

## Root cause

`webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java` (~line 2805, `followDependenciesExecutions`) eagerly runs:

```java
Await.await().atMost(Duration.ofSeconds(10)).pollInterval(...).until(() -> executionRepository.findById(...).orElse(null), Objects::nonNull);
```

on the IO thread **before** returning the Flux. For a missing execution the predicate never becomes true, so the call blocks the thread for the full 10s and then the awaitility timeout escapes as an unhandled `ConditionTimeoutException` → 500 (it isn't in `ErrorController`'s mapped types, so it hits the `Throwable` catch-all).

## Suggested fix

Look the execution up once; if absent return `404` immediately. Do not block an IO thread on a busy-wait; drive the "wait for it to appear" behavior reactively, and map the timeout to a proper status.

## Impact

- Any authenticated user can pin server IO threads 10s at a time and exhaust the pool with a handful of concurrent requests → webserver DoS.
- Every miss is a 10s hang + 500 even in normal use (stale/typo'd id).

Verified on `2.0.0-rc10` (commit `b48d8d1`, Docker H2).

fdelbrayelle:
## Summary
- `followDependenciesExecutions` busy-waited up to 10s on an IO thread via `Await.await().atMost(10s).pollInterval(500ms).until(...)` to look up the execution, then let the resulting `org.awaitility.core.ConditionTimeoutException` escape as an unmapped 500 for a missing id. A handful of concurrent bogus-id requests cheaply exhausts the IO thread pool.
- Replaced the polling lookup with a single `findById`, throwing `NoSuchElementException` (already mapped to 404 in `ErrorController`) immediately when the execution doesn't exist — matching the issue's own suggested fix.

Closes #18358

## QA
Full writeup with before/after test logs (timing-based proof — the "before" run measurably took 10s longer): https://claude.ai/code/artifact/c1816641-b715-4dd0-b3ce-c3c9e414a7e5

## Test plan
- [x] New test `followDependenciesOnUnknownExecutionReturnsNotFoundImmediately`: asserts 404 (not 500) and elapsed time under 5s for the exact repro from the issue
- [x] `./gradlew :webserver:test --tests io.kestra.webserver.controllers.api.ExecutionControllerRunnerTest.followDependenciesOnUnknownExecutionReturnsNotFoundImmediately` — passes in ~1s with the fix, fails (500, ~10s wait) without it (revert-and-rerun verified)
- [x] Existing follow-dependencies tests against real executions (`triggerExecutionAndFollowDependencies`, `shouldEndOnlyTerminatedDependenciesWhenFollowedExecutionIsStillRunning`) still pass unchanged

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
@@ -2801,14 +2801,10 @@ public class ExecutionController {
         //  But as we need the correlationId to unsubscribe, we have no choice but to do it eagerly.
         //  This should not be an issue as long as it executes on an IO thread.
 
-        // Check if execution exists
-        Execution current = Await.await()
-            .atMost(Duration.ofSeconds(10))
-            .pollInterval(Duration.ofMillis(500))
-            .until(
-                () -> executionRepository.findById(tenantService.resolveTenant(), executionId).orElse(null),
-                Objects::nonNull
-            );
+        // Check if execution exists; a missing id must fail fast instead of pinning this IO thread
+        Execution current = executionRepository
+            .findById(tenantService.resolveTenant(), executionId)
+            .orElseThrow(() -> new NoSuchElementException("Unable to find execution '" + executionId + "'"));
 
         String correlationId = current.getLabels().stream().filter(label -> label.key().equals(CORRELATION_ID)).findAny().map(label -> label.value()).orElseThrow();
 
--- a/webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionControllerRunnerTest.java
+++ b/webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionControllerRunnerTest.java
@@ -3127,6 +3127,25 @@ class ExecutionControllerRunnerTest {
     }
 
     @Test
+    void followDependenciesOnUnknownExecutionReturnsNotFoundImmediately() {
+        long start = System.nanoTime();
+
+        HttpClientResponseException e = assertThrows(
+            HttpClientResponseException.class, () -> client.toBlocking().retrieve(
+                HttpRequest.GET("/api/v1/main/executions/does-not-exist/follow-dependencies")
+            )
+        );
+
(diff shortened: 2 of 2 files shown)


fdelbrayelle:
**QA evidence:** https://claude.ai/code/artifact/c1816641-b715-4dd0-b3ce-c3c9e414a7e5

Before/after Gradle test run with a timing-based proof — without the fix the test takes ~10s longer and returns 500; with it restored, an immediate 404. Existing real-execution follow-dependencies tests checked for regressions.

PR #18562, opened 2026-08-24
fix(tasks): coerce non-string Log message list elements instead of crashing

### LINKED ISSUE #18356: [Bash2.0] Log task crashes with ClassCastException when a message-list element is not a string
## Summary

The `io.kestra.plugin.core.log.Log` task — one of the most-used tasks in Kestra — accepts a **list** of messages (documented feature). But if any element of that list is not a string (a number, boolean, or map), the task throws `ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String` at runtime. The flow passes validation and save cleanly; it only blows up when executed. Verified on `2.0.0-rc10`.

## Expected result

Each list element is coerced to a string and logged, exactly as a single scalar message is. A YAML author writing `- 42` should see `42` logged, not a `ClassCastException`.

## Repro steps

1. Create this flow (validates and saves with no error):

```yaml
id: log_cce
namespace: company.team
tasks:
  - id: log
    message:
      - "hello"
      - 42
```

2. Execute it.

## Actual result

The task fails with:

```
class java.lang.Integer cannot be cast to class java.lang.String
(java.lang.Integer and java.lang.String are in module java.base of loader 'bootstrap')
```

Execution ends `FAILED`.

## Root cause

`core/src/main/java/io/kestra/plugin/core/log/Log.java`:
- ~line 98: `Collection<String> messages = (Collection<String>) collectionValue;` — unchecked cast; the element type is never validated.
- ~line 102: `runContext.render(message)` binds to `render(String)`, forcing a per-element checkcast to `String`.

A non-string element (very natural in YAML — `- 42`, `- true`, a map) triggers the cast failure. It is not caught at save/validation time, so the author gets a runtime failure instead of either a helpful validation error or (better) automatic string coercion.

## Suggested fix

Coerce each element with `String.valueOf(...)` (or render `Object` and stringify) instead of casting to `String`, matching how a scalar message is handled.

## Impact

Most-used core task; trivial, natural trigger; fails only at runtime. Verified on `2.0.0-rc10` (commit `b48d8d1`, Docker H2).

fdelbrayelle:
## Summary
- `Log.run()` cast each collection element to `String` before rendering it (`Collection<String> messages = (Collection<String>) collectionValue;` then `runContext.render(message)` binding to the `String` overload), so a natural, valid-looking flow like `message: ["hello", 42]` validates and saves cleanly but throws a `ClassCastException` at execution time.
- Elements are now stringified with `String.valueOf(...)` before rendering — a no-op for elements that were already strings (preserving Pebble-templating for those), and a safe conversion for numbers/booleans/maps instead of a hard cast. Removed the now-unused `@SuppressWarnings("unchecked")` on `run()` along with the unchecked cast it existed for.

Closes #18356

## QA
Full writeup with before/after test logs: https://claude.ai/code/artifact/55f767b7-bb92-4101-ae01-5e8de2cc627f

## Test plan
- [x] New `LogTest.shouldLogNonStringMessageListElementsInsteadOfCrashing`: the exact repro from the issue (`["hello", 42, true]`) now logs all three elements instead of crashing after the first
- [x] `./gradlew :core:test --tests io.kestra.plugin.core.log.LogTest` — passes with the fix, fails with the exact reported `ClassCastException` without it (revert-and-rerun verified)
- [x] Checked other test usages of the `Log` task — all incidental (filler task in unrelated tests), no conflicts with the message-handling change

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/log/Log.java
+++ b/core/src/main/java/io/kestra/plugin/core/log/Log.java
@@ -84,7 +84,6 @@ public class Log extends Task implements RunnableTask<VoidOutput> {
     @Builder.Default
     private Property<Level> level = Property.ofValue(Level.INFO);
 
-    @SuppressWarnings("unchecked")
     @Override
     public VoidOutput run(RunContext runContext) throws Exception {
         Logger logger = runContext.logger();
@@ -95,11 +94,9 @@ public class Log extends Task implements RunnableTask<VoidOutput> {
             String render = runContext.render(stringValue);
             this.log(logger, renderedLevel, render);
         } else if (this.message instanceof Collection<?> collectionValue) {
-            Collection<String> messages = (Collection<String>) collectionValue;
-            messages.forEach(throwConsumer(message ->
+            collectionValue.forEach(throwConsumer(message ->
             {
-                String render;
-                render = runContext.render(message);
+                String render = runContext.render(String.valueOf(message));
                 this.log(logger, renderedLevel, render);
             }));
         } else {
--- a/core/src/test/java/io/kestra/plugin/core/log/LogTest.java
+++ b/core/src/test/java/io/kestra/plugin/core/log/LogTest.java
@@ -1,0 +1,49 @@
+package io.kestra.plugin.core.log;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+import io.kestra.core.junit.annotations.KestraTest;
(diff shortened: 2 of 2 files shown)


fdelbrayelle:
**QA evidence:** https://claude.ai/code/artifact/55f767b7-bb92-4101-ae01-5e8de2cc627f

Before/after Gradle test run using the exact repro from the issue — without the fix, the task logs "hello" then throws the exact reported ClassCastException on the Integer element; with the fix restored, all three elements ("hello", "42", "true") log correctly.

nkwiatkowski (approved):

PR #18564, opened 2026-08-24
fix(executor): fail only the execution when a Loop's values shrinks between iterations

### LINKED ISSUE #18355: [Bash2.0] A single valid Loop flow permanently crashes the whole instance (poison-pill full shutdown)
## Summary

A single, ordinary, **fully valid** flow can permanently take down an entire Kestra instance. If a `io.kestra.plugin.core.flow.Loop` task iterates over a dynamically-rendered `values` expression, and a child task shrinks the underlying source between iterations, the executor throws an uncaught `ArrayIndexOutOfBoundsException` **inside the `LoopExecutionEvent` queue consumer**. The queue subscriber treats any uncaught exception as fatal and **initiates a full application shutdown** — executor, scheduler, worker and webserver all stop.

Worse, the offending `LoopExecutionEvent` message stays on the queue, so on the next startup it is **redelivered and crashes the instance again**. The instance is stuck in a permanent crash-loop and cannot recover on its own — an operator has to manually start Kestra with `--ignore-queue-records <id>`.

This is a full-instance outage (not a single failed execution) triggerable by any user who can create a flow. Verified on `2.0.0-rc10` (Docker, H2).

## Expected result

An iteration whose value can no longer be resolved should fail (or handle) **that execution only**. The executor queue consumer must never let a `RuntimeException` from user data escape and shut down the whole server, and it must never leave a poison-pill message that re-crashes the instance on restart.

## Repro steps

1. Start Kestra `2.0.0-rc10`.
2. Create this flow (it passes validation cleanly):

```yaml
id: loop_shrink
namespace: company.team
tasks:
  - id: seed
    key: items
    kvType: JSON
    value: '["a","b","c"]'
  - id: loop
    concurrencyLimit: 1
    values: "{{ kv('items') }}"
    tasks:
      - id: shrink
        key: items
        kvType: JSON
        value: '["a"]'
```

3. Execute `loop_shrink` once.
4. Watch the server logs / the process.

The `values` expression `{{ kv('items') }}` initially resolves to a 3-element list, so `iterationCount = 3`. Iteration 0 rewrites the KV `items` to a 1-element list. On completion the handler computes `nextIndex = 1`, sees `1 < iterationCount(3)`, **re-resolves** `values` (now size 1), and does `list.get(1)`.

Any pattern where the loop source is re-read and can shrink reproduces it — e.g. a KV, a namespace file, or a subflow output that a child task modifies. It does not require concurrency; `concurrencyLimit: 1` is used here only to make the ordering deterministic.

## Actual result

The whole instance shuts down:

```
ERROR queue-h2_11 io.kestra.queue.AbstractSubscriber [LoopExecutionEvent] fatal error while consuming messages. Initiating application shutdown.
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
...
INFO i.kestra.core.contexts.KestraContext Kestra server - Shutdown completed
```

The container exits. On **restart**, the same message is redelivered and the instance crashes again immediately (verified — it never finishes booting):

```
ERROR queue-h2_11 io.kestra.queue.AbstractSubscriber [LoopExecutionEvent] failed to process message with key '7HJr8fkRSl6YVnim0LAi3B'. Message will be redelivered. You can ignore this message by starting Kestra with `--ignore-queue-records 7HJr8fkRSl6YVnim0LAi3B`.
...
ERROR queue-h2_11 io.kestra.queue.AbstractSubscriber [LoopExecutionEvent] fatal error while consuming messages. Initiating application shutdown.
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
```

## Root cause

`executor/src/main/java/io/kestra/executor/handler/LoopExecutionEventMessageHandler.java`:

- `iterationCount` is fixed at loop init (line 111), but `values` is **re-resolved every iteration** (line 142: `FlowableUtils.resolveValues(runContext, loop.getValues())`).
- `nextIndex` is guarded only against the stale `iterationCount` (line 127: `if (nextIndex < iterationCount)`), **not** against the size of the freshly-resolved list.
- Line 144 `String value = either.getLeft().get(nextIndex);` (and line 148 for the key/value branch) therefore does an out-of-bounds `get()` when the list shrank → `ArrayIndexOutOfBoundsException`.
- The surrounding `catch` (line 172) only handles `InternalException | QueueException | IOException` (plus `FlowNotFoundException`). `IndexOutOfBoundsException` / `NoSuchElementException` are plain `RuntimeException`s and escape. Notably `IllegalVariableEvaluationException` *is* handled, which suggests the omission of `RuntimeException` is an oversight.
- The escaped exception reaches `queue/src/main/java/io/kestra/queue/AbstractSubscriber.java:254`, which logs `fatal error while consuming messages. Initiating application shutdown.` and shuts the process down — and leaves the poison-pill message on the queue.

The URI branch (line 135, `getFirst()` on the re-read values) has the same class of problem and can throw `NoSuchElementException`.

## Suggested fix

- Bound `nextIndex` against the freshly-resolved list size and fail just this execution (via `handleFailedExecutionFromExecutor`) if the source shrank, rather than indexing blindly.
- Broaden the `catch` in the loop handler to convert any `RuntimeException` from user-data resolution into a failed execution instead of letting it escape.
- Separately, the queue subscriber shutting the whole application down on a single un-handled message is a very sharp edge — a poison message should be dead-lettered/skipped, not turned into a permanent, cross-restart, instance-wide outage.

## Impact

- **Blast radius:** entire instance (all services), not one execution.
- **Persistence:** survives restart (poison pill) — permanent outage until manual `--ignore-queue-records`.
- **Privilege required:** any user who can create and run a flow.
- **Likelihood in prod:** high — looping over a mutable source (KV/file/DB) that a child task updates is a natural pattern.

Verified on `2.0.0-rc10` (commit `b48d8d1`, Docker image `kestra:v2.0.0-rc10`, H2).

fdelbrayelle:
## Summary
- `LoopExecutionEventMessageHandler` re-resolves a `Loop`'s `values` expression on every iteration but only bounded the next index against the stale `iterationCount` captured at loop start, not the freshly-resolved list's actual size. If a child task shrinks the underlying source (KV, namespace file, subflow output) between iterations, `list.get(nextIndex)` throws an `IndexOutOfBoundsException`/`NoSuchElementException`.
- The handler's `catch` only covered `InternalException | QueueException | IOException` (plus `FlowNotFoundException` separately) — a plain `RuntimeException` escaped `handle()` entirely, hit `AbstractSubscriber`, which treats any uncaught exception as fatal and shuts the whole instance down, then redelivers the same poison-pill message on every restart.
- Bounds each re-resolved list/URI read against its actual size, failing with a clear `IllegalStateException` when it shrank. Broadened the catch to `RuntimeException` (with `FlowNotFoundException`, itself a `RuntimeException`, caught first to preserve its distinct infinite-loop guard) so any such user-data issue fails only this execution instead of escaping to the queue subscriber.

Closes #18355

## Out of scope
The issue also separately flags that `AbstractSubscriber` shutting the whole instance down on *any* uncaught exception from *any* handler is itself a sharp edge worth hardening globally (dead-lettering, per-handler isolation). This PR removes the concrete trigger for this handler, so the poison-pill scenario is fully prevented for this code path, but a broader queue-consumer resiliency change is a separate, larger architectural decision left for a dedicated follow-up.

## QA
Full writeup with before/after test logs plus the full H2RunnerTest suite: https://claude.ai/code/artifact/287bb3dd-35be-4b2f-8efd-88dedb3acce3

## Test plan
- [x] New test `shouldFailOnlyThisExecutionWhenLoopValuesShrinkBetweenIterations`: seeds a KV-backed loop source, shrinks it between iterations exactly like the issue's repro, asserts the loop task run ends `FAILED` instead of throwing
- [x] `./gradlew :executor:test --tests io.kestra.executor.handler.LoopExecutionEventMessageHandlerTest` — 8/8 pass with the fix, the new test fails with the exact unhandled exception without it (revert-and-rerun verified)
- [x] `./gradlew :core:test --tests io.kestra.plugin.core.flow.LoopTest` — unaffected
- [x] `./gradlew :jdbc-h2:test --tests io.kestra.runner.h2.H2RunnerTest` (per AGENTS.md, run whenever the executor changes) — 100/100 pass

### DIFF
--- a/executor/src/main/java/io/kestra/executor/handler/LoopExecutionEventMessageHandler.java
+++ b/executor/src/main/java/io/kestra/executor/handler/LoopExecutionEventMessageHandler.java
@@ -130,10 +130,10 @@ public class LoopExecutionEventMessageHandler implements ExecutorMessageHandler<
                             // URI mode: seek to stored offset and read the next value
                             long nextOffset = ((Number) outputs.get(Loop.NEXT_OFFSET_OUTPUT)).longValue();
                             String valuesUri = FlowableUtils.resolveLoopValuesUri(runContext, loop.getValues())
-                                .orElseThrow(() -> new IllegalStateException("Loop has a nextOffset output but values did not resolve to a URI"));
+                                .orElseThrow(() -> new InternalException("Loop has a nextOffset output but values did not resolve to a URI"));
                             var valuesAndOffset = FlowableUtils.readLoopValuesFromUri(runContext, valuesUri, nextOffset, 1);
                             if (valuesAndOffset.getLeft().isEmpty()) {
-                                throw new IllegalStateException(
+                                throw new InternalException(
                                     "Loop 'values' has no value left at offset %d for iteration %d; the underlying source likely changed between iterations.".formatted(nextOffset, nextIndex)
                                 );
                             }
@@ -148,7 +148,7 @@ public class LoopExecutionEventMessageHandler implements ExecutorMessageHandler<
                             if (either.isLeft()) {
                                 List<String> values = either.getLeft();
                                 if (nextIndex >= values.size()) {
-                                    throw new IllegalStateException(
+                                    throw new InternalException(
                                         "Loop 'values' resolved to %d element(s), but iteration %d was expected; the underlying source likely changed between iterations."
                                             .formatted(values.size(), nextIndex)
                                     );
@@ -159,7 +159,7 @@ public class LoopExecutionEventMessageHandler implements ExecutorMessageHandler<
                             } else {
                                 List<Pair<String, String>> values = either.getRight();
                                 if (nextIndex >= values.size()) {
-                                    throw new IllegalStateException(
+                                    throw new InternalException(
                                         "Loop 'values' resolved to %d element(s), but iteration %d was expected; the underlying source likely changed between iterations."
                                             .formatted(values.size(), nextIndex)
                                     );
@@ -195,10 +195,7 @@ public class LoopExecutionEventMessageHandler implements ExecutorMessageHandler<
                 }
 
                 return null;
-            } catch (InternalException | QueueException | IOException | RuntimeException e) {
-                // RuntimeException covers user-data issues re-resolving 'values' (e.g. the underlying
-                // source shrank between iterations) - this must fail only this execution, never escape
-                // to the queue subscriber, which would treat it as fatal and shut the whole instance down.
+            } catch (InternalException | QueueException | IOException e) {
                 return executorService.handleFailedExecutionFromExecutor(new ExecutorContext(execution), e);
             }
         });
--- a/executor/src/main/java/io/kestra/executor/handler/LoopExecutionEventMessageHandler.java
+++ b/executor/src/main/java/io/kestra/executor/handler/LoopExecutionEventMessageHandler.java
@@ -132,6 +132,11 @@ public class LoopExecutionEventMessageHandler implements ExecutorMessageHandler<
                             String valuesUri = FlowableUtils.resolveLoopValuesUri(runContext, loop.getValues())
                                 .orElseThrow(() -> new IllegalStateException("Loop has a nextOffset output but values did not resolve to a URI"));
                             var valuesAndOffset = FlowableUtils.readLoopValuesFromUri(runContext, valuesUri, nextOffset, 1);
+                            if (valuesAndOffset.getLeft().isEmpty()) {
+                                throw new IllegalStateException(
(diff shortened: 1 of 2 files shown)


fdelbrayelle:
**QA evidence:** https://claude.ai/code/artifact/287bb3dd-35be-4b2f-8efd-88dedb3acce3

Before/after Gradle test run reproducing the exact reported scenario (KV-backed loop source shrinking between iterations) — without the fix the exact unhandled IndexOutOfBoundsException escapes; with it restored the execution fails cleanly instead. Full H2RunnerTest suite (100/100) and LoopTest checked for regressions per AGENTS.md's executor-change guidance.

nkwiatkowski (approved):

PR #18566, opened 2026-08-24
fix(flows): deduplicate the flow editor's Add-a-trigger panel

### LINKED ISSUE #18321: [Bash2.0] Flow editor side panels are not deduplicated and never reset, degrading to unusable slivers
**Expected result:** Reopening an already-open editor panel (e.g. clicking "Add a trigger" again) should focus the existing panel, not create a new one. Panel state should reasonably reset across full page navigations.

**Repro steps:**
1. Open a flow's Triggers tab, click "Add a trigger" (opens a panel alongside any other open panels like Docs/Blueprints/No-code).
2. Do a full page reload back to the same Triggers tab (fresh navigation, not in-app).
3. Click "Add a trigger" again.

**Actual result:** All previously opened panels are still open after the reload, and a second, duplicate "triggers / Add a trigger" tab is created instead of focusing the existing one. With 5 panels open, the editor becomes a strip of icon-only, barely usable columns.

fdelbrayelle:
## Summary
- Clicking "Add a trigger" navigates with `?createTrigger=true`, and `MultiPanelFlowEditorView.vue`'s `onMounted` handled it by calling `actions.openAddTaskTab` directly — bypassing the dedup check `useNoCodeHandlers.onCreateTask` already provides for the same tab type from the no-code editor's own UI. Every click opened a fresh tab with a new cache-busted uid, and since panel state persists across full page reloads by design, stale tabs never went away either.
- Routing the call through `onCreateTask` surfaced a **second, latent bug**: its dedup-key computation didn't default `position` the same way `openAddTaskTab` does (`"after"`), so a caller omitting `position` (like this one) computed a key that never matched the tab actually created, duplicating it every time regardless. Both defaults now agree.

Closes #18321

## Scope
This fixes the concrete "duplicate on every click" defect the issue's repro demonstrates. The softer "panel state should reasonably reset across full page navigations" expectation is not addressed — panels still persist across reloads by design; they just no longer accumulate duplicates of the same action, which is what actually degrades the editor into "unusable slivers."

## QA
Live browser reproduction against a real local dev server, exact repro steps from the issue, plus a failing→passing unit test. Full writeup: https://claude.ai/code/artifact/af577f7a-0a8a-433f-b790-f81d4c48428f

## Test plan
- [x] New unit test `useNoCodeHandlers.onCreateTask`: repeated calls with the same parentPath/blockSchemaPath (position omitted, matching the createTrigger call site) focus the existing tab instead of duplicating
- [x] `npx vitest run --project=unit tests/unit/components/flows/useNoCodePanels.spec.ts` — 2/2 pass with both fixes, 1/2 fails without them (revert-and-rerun verified)
- [x] Live browser (Playwright against a real dev server): clicked "Add a trigger", reloaded, clicked again — 2 real tab elements confirmed via bounding boxes/localStorage without the fix, 1 with it restored
- [x] Broader flows/no-code Vitest suite (259 tests, 24 files) — unaffected

### DIFF
--- a/ui/src/components/flows/MultiPanelFlowEditorView.vue
+++ b/ui/src/components/flows/MultiPanelFlowEditorView.vue
@@ -109,7 +109,7 @@
                     "triggers",
                     "items",
                 ].join("/")
-                actions.openAddTaskTab({panelIndex, tabIndex: 0}, "triggers", blockSchemaPath)
+                noCodeHandlers.onCreateTask({panelIndex, tabIndex: 0}, "triggers", blockSchemaPath)
             }
         }
     })
@@ -169,7 +169,7 @@
         }))
     }
 
-    const {panels, actions} = useNoCodePanelsFull({
+    const {panels, actions, noCodeHandlers} = useNoCodePanelsFull({
         RawNoCode,
         editorView,
         editorElements: EDITOR_ELEMENTS,
--- a/ui/src/components/flows/useNoCodePanels.ts
+++ b/ui/src/components/flows/useNoCodePanels.ts
@@ -178,7 +178,10 @@ export function setupInitialNoCodeTab(Comp: any, tab: string, handlers: Handlers
 
 export function useNoCodeHandlers(openTabs: Ref<string[]>, focusTab: (tab: string) => void, actions: ReturnType<typeof useNoCodePanels>) {
     const noCodeHandlers: Handlers = {
-        onCreateTask(opener, parentPath, blockSchemaPath, refPath, position){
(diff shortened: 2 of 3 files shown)


fdelbrayelle:
**QA evidence:** https://claude.ai/code/artifact/af577f7a-0a8a-433f-b790-f81d4c48428f

Live-browser reproduction against a real local dev server (Playwright), following the issue's exact repro steps including a real page reload, plus a Vitest unit test that fails without the fix and passes with it. Also documents a second, latent dedup-key bug that a first draft of the unit test's mock missed and the live-browser run caught.

nkwiatkowski (approved):

PR #18267, opened 2026-08-19
Fix dependencies view graph 18164

seshan18:
Closes https://github.com/kestra-io/kestra/issues/18296.

✨ Description
Removes the embedded execution `TimeSeries` chart, duration selector, and associated background computations/styles from the `Dependencies` view so that the tab is exclusively focused on rendering the flow dependency topology graph and table.
🔗 Related Issue
Closes https://github.com/kestra-io/kestra/issues/18164
🎨 Frontend Checklist
- [x] Code builds without errors (`npm run build`)
- [x] All existing E2E tests pass (`npm run test:e2e`)
- [ ] Screenshots or video recordings attached showing the UI changes
📝 Additional Notes
- Cleaned up unused imports (`TimeSeries`, `ChartDurationSelect`, `YAML_UTILS`) and lifecycle hooks/computeds.
- Removed dead `.dependencies-wrapper` and `.chart-header` SCSS rules.
- Prevents unnecessary execution metrics queries from firing on the Dependencies tab.
🤖 AI Authors
Why was the cat sitting on the computer?
To keep an eye on the *mouse*! 🐱

PR #10209, opened 2025-07-18
fix(flows): remove flows date filters UI, they do not exist

AcevedoR:
- fixes https://github.com/kestra-io/kestra/issues/10201

Issue #10207, opened 2025-07-18
[No-code] Enhanced search for optional properties


wrussell1999:
### Feature description

Some tasks have a lot of properties which can be difficult to navigate in the no-code editor. It would almost be worth implementing some sort of list of properties at the side you can use to navigate quickly, or some sort of search.

I think we did a good job of this with the documentation by letting you expand what you want to see:
<img width="1149" height="999" alt="Image" src="https://github.com/user-attachments/assets/05acab74-a943-4b04-9403-d646c35af784" />

I kept finding myself missing properties because there's a lot for the Python Script task:

https://github.com/user-attachments/assets/aa784160-b770-4c99-a010-f54d78aa0bdd

wrussell1999:
Duplicate: https://github.com/kestra-io/kestra/issues/6534

PR #18567, opened 2026-08-24
feat(assets): add a layered DAG view and freshness to asset dependencies

soyasis:
closes https://github.com/kestra-io/kestra-ee/issues/10207

closes https://github.com/kestra-io/kestra-ee/issues/10210

Part of https://github.com/kestra-io/kestra-ee/issues/10031

Companion PR https://github.com/kestra-io/kestra-ee/pull/10393

## What

A Tree / DAG toggle on the assets Dependencies tab. DAG ranks nodes by longest path from the roots into left-to-right columns, ties broken by id, so the same graph yields the same coordinates on every reload; cycles are cut at the lowest id rather than stalling. A flow that writes at least one asset takes the leading column, since the flow triggers the lineage. Its consume edges therefore run right-to-left and are drawn as dashed back-edges, detected from the laid-out coordinates rather than from producer sets. No edge is dropped to achieve the layout — dropping edges into flows is what previously made "what depends on this table" under-report with no warning.

Cards carry freshness as colour, icon and age, with a one-line summary above the canvas so the graph answers "is anything wrong here" before any single node is read. Hovering or selecting a node lights its transitive ancestors and descendants; flow nodes are reachable but not traversable, because one flow writes many unrelated assets and walking through it would light the whole graph from any starting point. Selecting a node swaps the side table for a panel listing the runs that wrote the asset, each linking to its execution.

DAG renders through vue-flow on its own canvas with an explicit id, layered over the hidden chart rather than replacing it: remounting re-runs the force simulation and loses the settled positions, and a `display: none` chart measures 0x0, so a later resize would corrupt the Tree view. Tree stays on ECharts.

Freshness values come from EE — see the companion PR.

Two commits, split by blast radius:

1. `fix(topology)` — only code the four dependency views share. **This changes behaviour in all four: selecting a node no longer moves the camera.** It also pins the series box with `preserveAspect` + `roamTrigger`, without which capturing force positions and switching to `layout: "none"` lets ECharts stretch-fit the bounding rect with independent scaleX/scaleY and every node renders as an ellipse.
2. `feat(assets)` — the DAG canvas, freshness and the side panel, all gated on a `dagView` prop.

## Changes shared by all four views, called out deliberately

- Selecting a node no longer re-centres the graph (commit 1).
- The side table keeps its search, namespace filter and flows switch across selection, and its result count now renders in every view rather than only the asset view.
- The graph pane regains a flex wrapper so the toggle bar can dock above the splitter. `#18267` removed that wrapper along with the chart; the toolbar needs it back.
- The table splitter panel gains `min="320px"`, so at very narrow widths the detail column stops wrapping values one character per line.

## Why not a smaller change

The force layout is the problem being fixed, so a new layout is the minimum. It cannot simply replace the old one: Tree is still the better tool for a dense graph, so both coexist, which is what the toggle and the two canvases are for.

This branch started as a prototype and was trimmed from 4,159 added lines to ~2,400 before review: group-by split out, the demo-only dbt manifest overlay deleted, the level-of-detail machinery dropped, and roughly 760 lines of commentary removed.

## Testing

`npm run test:unit -- tests/unit/dependencies` — 21 tests across three files. Four of them assert the negative case for the flow, execution and namespace views, which share `Dependencies.vue` and `useDependencies.ts` and have regressed repeatedly by inheriting asset-view behaviour. The camera specs pin that only the initial auto-selection moves the viewport.

`npm run check:types` and `npm run test:lint` clean. Manually exercised on a dbt project: both layouts, selection, hover-trace, table filtering, splitter resize, and the three sibling dependency tabs.

## Out of scope

- Group-by / sections — https://github.com/kestra-io/kestra-ee/issues/10209
- Level of detail when zoomed out. https://github.com/kestra-io/kestra-ee/issues/10207 calls the prototype's version inadequate (it dropped the status row but still drew the asset name), so it was removed rather than shipped as-is; it needs its own issue.
- dagre. `dagLayout.ts` is kept because it provides the three things #10207 notes dagre does not give for free: deterministic tie-breaking by id, the pinned leading column, and a stable intra-rank order for #10209 to reorder.
- Unit tests beyond the gating and camera ones, by agreement — `dagLayout`, `dagTrace` and `assetStatus` are pure functions and untested here.

## Known gate

`translations:check` fails: the new `dependency.dag.*` keys are English-only and `translations:generate` needs a `GEMINI_API_KEY` that was not available. Needs regenerating before merge.


soyasis:
Two things worth flagging for review, both deliberate.

**Multi-flow pinning.** The leading-column rule pins any flow that writes an asset, so in a graph with several producing flows a mid-stream flow (one consuming another flow's output) is pinned too, which reads as less truthful than leaving it in rank order. Fine for the single-producing-flow case, which is the common one; noted in `DagCanvas.vue` and worth revisiting when a second producing flow is a real case. Note this differs from what https://github.com/kestra-io/kestra-ee/issues/10207 currently says — the issue text predates the decision and needs updating.

**Prototype lineage.** This branch is the trimmed version of `prototype/assets-dag-view`, which remains as a design reference (and is where the demo video came from). The prototype's frontend dbt-manifest overlay is deleted here rather than ported, so asset-to-asset edges come from Kestra's own lineage — meaning the seed-to-model hop is absent until https://github.com/kestra-io/plugin-dbt/issues/319 lands.

soyasis:
Follow-up PR for group-by, stacked on this one: https://github.com/kestra-io/kestra/pull/18569 (closes https://github.com/kestra-io/kestra-ee/issues/10209). It is based on this branch so its diff stays at 7 files; it needs retargeting to `develop` once this merges.

Piyush-r-bhaskar (approved):
LGTM

PR #18570, opened 2026-08-24
fix(system): let button icons inherit their button's text colour on the license bar

### LINKED ISSUE #18449: [Bash 2.0] [UI] Book a demo icon color button should be white
### Describe the issue

http://localhost:8080/ui/main/admin/stats
Icon color inside button should be white! 

Image

### Environment

- Kestra Version: develop

jaybamroliya:
### ✨ Description

The `.bar :deep(svg)` rule painted every icon in the license bar with `--ks-icon-muted`, including the one on the primary "Get a Demo" button, where it should match the white label.

Removing the rule lets each icon inherit its own button's text colour — muted on the secondary button, white on the primary. It also drops a `:deep()` selector, which `ui/AGENTS.md` disallows.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18449

### 🎨 Frontend Checklist

### DIFF
--- a/ui/src/components/admin/stats/Pricing.vue
+++ b/ui/src/components/admin/stats/Pricing.vue
@@ -55,10 +55,6 @@
             border-radius: 8px;
             background: var(--ks-bg-surface);
             box-shadow: 0px 1px 4px 0px var(--ks-shadow-element);
-
-            :deep(svg) {
-                color: var(--ks-icon-muted);
-            }
         }
 
         .edition {


MilosPaunovic (approved):
Approving - the fix is correct and I verified the colour chain end to end.

`vue-material-design-icons` renders its `svg` with `fill="currentColor"`, and `KsButton` maps `.is-plain.kel-button--primary` to `--ks-btn-primary-text`, so the `Get a Demo` icon now resolves to `#ffffff`. Confirmed against computed styles in the design-system Storybook. Dropping the `:deep()` is the right call too, since `ui/AGENTS.md` disallows it.

One note for the record: the `Pricing` button's icon changes as well, from `--ks-icon-muted` (`#7e7e9b`) to `--ks-text-primary` (`#08090a`), because the plain default button never overrides `--kel-button-text-color`. The description says it stays muted, which is not the case. That is fine by me: nothing in the design system mutes icons inside buttons, so this actually makes the license bar consistent with every other icon button in the `UI`.

No `EE` companion is needed - the `EE` `Stats.vue` override renders its own license section rather than `Pricing`. No backport either: `Pricing.vue` is 2.0-only and does not exist on `releases/v1.3.x` or `releases/v1.0.x`.

I have also corrected the `PR` title scope from `apps` to `system`, since `apps` is the `EE` Apps feature while this is the instance overview page.

Thanks @jaybamroliya!

PR #18572, opened 2026-08-25
fix(namespaces): return 422 instead of 500 on a non-hierarchical path

### LINKED ISSUE #18407: [Bash2.0] Namespace file move returns 500 NPE on an opaque URI in from= (no path validation)
## Summary

`PUT /api/v1/{tenant}/namespaces/{namespace}/files?from=...&to=...` (move/rename a namespace file) returns HTTP 500 with a `NullPointerException` when `from` is an opaque URI such as `a:b`. The `from`/`to` params bind directly to `java.net.URI`, and `URI("a:b").getPath()` is `null` for opaque URIs, so the path-guard regex runs `Pattern.matcher(null)`. Verified on `2.0.0-rc10`.

## Expected result

A malformed / non-path `from` value returns a clear `422` (bad request), not a 500.

## Repro steps

```bash
curl -sS -u admin@kestra.io:Kestra42! -X PUT \
  'http://localhost:8080/api/v1/main/namespaces/company.team/files?from=a:b&to=/x.txt'
```

## Actual result

HTTP 500:

```
Internal server error: Cannot invoke "java.lang.CharSequence.length()" because "this.text" is null
```

(the null is `URI.getPath()` fed into a `Pattern.matcher`).

## Root cause

`webserver/src/main/java/io/kestra/webserver/controllers/api/NamespaceFileController.java` (~line 384, `forbiddenPathsGuard`, reached from `moveFileDirectory`): `from`/`to` bind as `java.net.URI` with no scheme/shape validation. For an opaque URI (`a:b`), `getPath()` returns null, and `pattern.matcher(path.getPath())` → NPE. The NPE falls through `ErrorController`'s `Throwable` catch-all → 500.

## Suggested fix

Validate that `from`/`to` are hierarchical (non-null path) before the guard; return 422 otherwise.

## Impact

Confirmed 500 on a core namespace-files endpoint from trivial malformed input. Verified on `2.0.0-rc10` (commit `b48d8d1`, Docker H2).

### LINKED ISSUE #18417: [Bash2.0] Namespace create-directory returns 500 NPE on an opaque URI in path= (no path validation)
## Summary

`POST /api/v1/{tenant}/namespaces/{namespace}/files/directory?path=...` (create a namespace directory) returns HTTP 500 with a `NullPointerException` when `path` is an opaque URI such as `a:b`. `URI("a:b").getPath()` is null for opaque URIs, and the path-guard runs `Pattern.matcher(null)`. Verified on `2.0.0-rc10`.

(Related: the file *move* endpoint has the same defect via `from=` — reported separately. This one is the directory-create endpoint via `path=`; same guard method, different route, so noting both.)

## Expected result

A malformed / non-hierarchical `path` returns `422`, not a 500.

## Repro steps

```bash
  'http://localhost:8080/api/v1/main/namespaces/company.team/files/directory?path=a:b'
```

## Actual result

HTTP 500:

```
Internal server error: Cannot invoke "java.lang.CharSequence.length()" because "this.text" is null
```

(the null is `URI.getPath()` passed into a `Pattern.matcher`).

## Root cause

`webserver/src/main/java/io/kestra/webserver/controllers/api/NamespaceFileController.java` — the `path` query param binds as `java.net.URI` with no scheme/shape validation; `forbiddenPathsGuard` calls `pattern.matcher(path.getPath())` and `getPath()` is null for an opaque URI → NPE → `Throwable` catch-all → 500.

## Suggested fix

Validate that `path` is hierarchical (non-null path) before the guard; return 422 otherwise. Fix the shared guard so all namespace-file endpoints (create, move, directory) are covered.

## Impact

Confirmed 500 from trivial malformed input on the namespace directory-create endpoint. Verified on `2.0.0-rc10` (commit `b48d8d1`, Docker H2).

fdelbrayelle:
## Summary
- `forbiddenPathsGuard` (shared by every namespace-file endpoint: create, list, move, directory) called `path.getPath()` unconditionally and fed it into `Pattern.matcher(...)`. For an opaque-shaped input like `a:b`, the resulting URI's path is null, and `Matcher.matches()` NPEs trying to read its length — caught only by the generic `Throwable` handler → 500.
- Added a null-path check that throws a proper 422 before the guard ever reaches the regex match.

Closes #18417
Closes #18407

## Scope
Fixing the shared guard covers create, list, move, and directory uniformly, matching the issue's own suggested fix — and directly benefits the sibling file-move issue (#18407), which hits the exact same guard via a different query param (`from=`). Added its own regression test to confirm.

## QA
Full writeup with before/after test logs: https://claude.ai/code/artifact/abf180bd-c7d0-4d5b-80d3-f2603c547a61

## Test plan
- [x] New test `createNamespaceDirectoryWithOpaqueUriPathReturns422`: the exact repro from #18417 (`path=a:b`) now returns 422
- [x] New test `moveFileDirectoryWithOpaqueUriFromReturns422`: the exact repro from #18407 (`from=a:b`) now returns 422
- [x] `./gradlew :webserver:test --tests io.kestra.webserver.controllers.api.NamespaceFileControllerTest` — 25/25 pass with the fix, exact reported NPE reproduced without it (revert-and-rerun verified)

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/NamespaceFileController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/NamespaceFileController.java
@@ -31,8 +31,10 @@ import io.micronaut.core.annotation.Nullable;
 import io.micronaut.data.model.Pageable;
 import io.micronaut.http.HttpHeaders;
 import io.micronaut.http.HttpResponse;
+import io.micronaut.http.HttpStatus;
 import io.micronaut.http.MediaType;
 import io.micronaut.http.annotation.*;
+import io.micronaut.http.exceptions.HttpStatusException;
 import io.micronaut.http.multipart.CompletedFileUpload;
 import io.micronaut.http.server.types.files.StreamedFile;
 import io.micronaut.scheduling.TaskExecutors;
@@ -379,6 +381,10 @@ public class NamespaceFileController {
     private void forbiddenPathsGuard(URI path) {
         if (path == null) {
             return;
+        }
+
+        if (path.getPath() == null) {
+            throw new HttpStatusException(HttpStatus.UNPROCESSABLE_ENTITY, "The path '%s' is not a valid hierarchical path.".formatted(path));
         }
 
         if (forbiddenPathPatterns.stream().anyMatch(pattern -> pattern.matcher(path.getPath()).matches())) {
--- a/webserver/src/test/java/io/kestra/webserver/controllers/api/NamespaceFileControllerTest.java
+++ b/webserver/src/test/java/io/kestra/webserver/controllers/api/NamespaceFileControllerTest.java
@@ -249,6 +249,26 @@ class NamespaceFileControllerTest {
     }
 
     @Test
+    void moveFileDirectoryWithOpaqueUriFromReturns422() {
+        String namespace = TestsUtils.randomNamespace();
+        HttpClientResponseException e = assertThrows(
+            HttpClientResponseException.class,
+            () -> client
+                .toBlocking()
+                .exchange(
+                    HttpRequest.PUT(
+                        "/api/v1/main/namespaces/" + namespace + "/files?from=a:b&to=/x.txt",
+                        null
+                    )
+                )
+        );
+
+        assertThat(e.getStatus().getCode())
+            .as("a non-hierarchical 'from' is a validation error, not an unhandled 500")
+            .isEqualTo(HttpStatus.UNPROCESSABLE_ENTITY.getCode());
+    }
+
+    @Test
     void createGetFileContent() throws IOException {
         String namespace = TestsUtils.randomNamespace();
         MultipartBody body = MultipartBody.builder()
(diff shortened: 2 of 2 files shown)


fdelbrayelle:
**QA evidence:** https://claude.ai/code/artifact/abf180bd-c7d0-4d5b-80d3-f2603c547a61

Before/after Gradle test run using the exact repro from the issue — without the fix, the exact reported NPE reproduces (500); with it restored, a clean 422. Full 24-test NamespaceFileControllerTest suite checked for regressions.

fdelbrayelle:
## QA report (updated)

Added regression coverage for the sibling issue #18407 to this same PR — the shared `forbiddenPathsGuard` fix covers it too. Full before/after evidence for both #18417 and #18407: https://claude.ai/code/artifact/abf180bd-c7d0-4d5b-80d3-f2603c547a61

- Reverted just the null-path check, re-ran → both repros reproduce the exact reported 500 NPE
- Restored the fix, re-ran → `./gradlew :webserver:test --tests NamespaceFileControllerTest` — 25/25 pass

ammeek (approved):

PR #18585, opened 2026-08-25
fix(core): shut down mcp keep-alive schedulers on context close

ammeek:
MCP servers were never shut down when the application context closed.

McpServerHandlerTransport lazily builds one KestraFluxStreamableServerTransportProvider per (tenantId, serverId), and each one starts a KeepAliveScheduler in its constructor. Only evictAndNotify — called when a server is deleted or disabled — ever stopped one. On a normal context shutdown the schedulers were left running.

Each leaked scheduler holds a Flux.interval on the shared boundedElastic pool and pings every session it can see every 30 seconds. Sessions that were initialised over POST but never opened an SSE stream reject the ping immediately, producing a steady stream of:

WARN boundedElastic-4 i.m.util.KeepAliveScheduler Failed to send keep-alive ping to
session McpStreamableServerSession@...: Stream unavailable for session <uuid>

In production this leaks a scheduler and the objects it retains per webserver restart within a JVM. It is most visible in tests, where many contexts are created in one JVM and the warnings continue for the rest of the run.

Two changes:

- McpServerHandlerTransport gets a @PreDestroy that evicts every built handler and server, bounded by a 10s block.
- KestraFluxStreamableServerTransportProvider.closeGracefully() releases the scheduler in doFinally instead of doOnSuccess, so an error while closing sessions cannot leave the interval running.


brian-mulier-p (approved):
LGTM

PR #18586, opened 2026-08-25
fix(plugins): stop inlining icon bytes in the bulk plugin icon indexes

fdelbrayelle:
## Summary

Cold-loading the Plugins page downloaded ~28 MB, 24 MB of it the icons index, and parsing that on the main thread froze the sidebar router, made the hover border trail the cursor, and left the page sluggish.

**The bytes were never used.** Both bulk indexes inline every icon as base64, but the frontend keeps only the metadata (`flowable`, `monochrome`, `hash`) and renders each icon from the per-class, browser-cacheable `GET /plugins/icons/{cls}/icon.svg`. On a local instance the index is 6.36 MB, of which 105 KB (1.7%) is that metadata.

So the indexes now answer metadata only. `hash` is non-null exactly when a class has an icon (both derive from the same `Optional`, verified across all 707 entries), which is what the frontend now derives `hasIcon` from — the response type is unchanged, so no caller loses information.

**A second bug fell out of it.** `GET /plugins/icons/groups` was answering the *task-class* index — keyed by task FQCN instead of by plugin group. Micronaut derives a cache key from the method arguments alone (`DefaultCacheKeyGenerator` returns the shared `ParametersKey.ZERO_ARG_KEY` for any no-arg method), so the two no-arg `@Cacheable("default")` index methods shared one key and whichever populated first served both. Each index now gets its own cache.

| Endpoint | Before | After | Change |
|---|---:|---:|---:|
| `/plugins/icons` | 6,347,375 | 99,073 | −98.4% |
| `/plugins/icons/groups` | 6,347,375 | 24,346 | −99.6% |
| **Plugins page total** | **14,038,367** | **1,466,636** | **−89.6%** |

Closes https://github.com/kestra-io/kestra-ee/issues/10128.

## Test plan

- [x] `PluginControllerTest` 27/27, including two new tests: no index entry carries icon bytes; the groups route is keyed by group, not class
- [x] Sanity-checked — reverting only the cache split makes `shouldAnswerGroupsRatherThanClassesForGroupIcons` fail, and pass again when restored
- [x] `npm run check:types` and eslint clean
- [x] Manual, cold load of `/ui/default/plugins` on EE: sidebar navigation 127–148 ms over three runs, hover applies to exactly the hovered tile in 18–28 ms, all tile icons render via `icon.svg`
- [ ] Generated SDK: the OpenAPI descriptions changed, so `update-generated-sdk` needs to run after merge

Full QA report with screenshots and the measurement method is in the [first comment](https://github.com/kestra-io/kestra/pull/18586#issuecomment-5409268388).


fdelbrayelle:
## QA report

Run on Kestra EE `0.24.0-rc0-SNAPSHOT` (standalone + Postgres) with OSS on `6fac1a1`. Local catalogue: 707 plugin classes / 184 groups.

Full report with screenshots: https://claude.ai/code/artifact/f786ffe3-5297-4b9b-bd05-dfdd7bc0aedf

### Payload

| Endpoint | Before | After | Change |
|---|---:|---:|---:|
| `/plugins/icons` | 6,347,375 | 99,073 | −98.4% |
| `/plugins/icons/groups` | 6,347,375 | 24,346 | −99.6% |
| `/plugins` | 358,760 | 358,760 | — |
| `/plugins/groups/subgroups` | 984,457 | 984,457 | — |
| **Plugins page total** | **14,038,367** | **1,466,636** | **−89.6%** |

Of the 6.36 MB index, 105 KB (1.7%) is the metadata the frontend keeps (`flowable`, `monochrome`, `hash`). The rest is base64 SVG that `toPluginIconData` discards, because each tile renders from the per-class, immutably-cached `GET /plugins/icons/{cls}/icon.svg`. `hash` is non-null exactly when a class has an icon — 0 violations across all 707 entries — so `hasIcon` is derived from it and no caller loses information.

### The three reported symptoms, re-tested

Each on a genuine cold load of `/ui/default/plugins`:

1. **Sidebar navigation** — 127/128/148 ms across three cold loads, clicking *Namespaces* as soon as the sidebar painted. A plugin tile (`router.push`) navigated in 264 ms.
2. **Hover tracking** — `.plugin-card:hover` matched exactly one element (the tile under the cursor) in 18–28 ms across three tiles. No stale highlight.
3. **Responsiveness** — no stall; the largest remaining JSON on the page is 984 KB.

### Second bug found on the way

`GET /plugins/icons/groups` was answering the **task-class** index — keyed by task FQCN (`io.kestra.plugin.core.log.Log`) instead of by plugin group. Both routes returned byte-identical responses.

Cause: Micronaut derives a cache key from the method arguments alone — `DefaultCacheKeyGenerator` returns the shared `ParametersKey.ZERO_ARG_KEY` for any no-arg method — so the two no-arg `@Cacheable("default")` index methods shared one key and whichever populated first served both. Each index now has its own cache. The groups route answers 184 real group/subgroup keys.

### Checks

- 27/27 `PluginControllerTest` pass, including two new tests (no icon bytes in the index; groups route keyed by group).
- Sanity-checked: reverting only the cache split makes `shouldAnswerGroupsRatherThanClassesForGroupIcons` fail, and pass again when restored.
- Browser: cold load issues 564 requests, 152 of them per-class `icon.svg`, all 200 and `immutable`, so a warm reload re-fetches none.

### Limits of this run

- **No broken baseline was measured.** Early readings that looked like a frozen sidebar turned out to be the dev-only "SDK out of date" banner overlaying the page and swallowing clicks; once dismissed, navigation was 127–148 ms. The symptoms verify good on the fixed build, but this harness never reproduced them on the unfixed one — so the before/after evidence here is the payload, not the latency.
- **Smaller catalogue than the report:** 707 classes / 6.36 MB here vs the reporter's 24 MB. Same shape, smaller absolute numbers.
- **Generated SDK not committed** — the OpenAPI descriptions changed, so the committed SDK hash drifts and dev mode warns; `update-generated-sdk` regenerates it after merge.
- **Pre-existing failures:** `pluginsFetchIcons` / `pluginsLoadIcon` fail in this environment on an unmodified checkout too (Node localStorage unavailable to the leak guard).

MilosPaunovic (commented):
The core change is correct and the diagnosis of the second bug is right. `DefaultCacheKeyGenerator` does return the shared `ParametersKey.ZERO_ARG_KEY` for any no-arg method, so the two no-arg `@Cacheable("default")` index methods genuinely collided, and they collided silently because both return `Map<String, PluginIcon>`, so no `ClassCastException` ever surfaced. `inputDocumentation(Type)` is the only other `@Cacheable("default")` in `OSS` or `EE` and it takes an argument, so no collision remains. There is no `@CacheInvalidate` anywhere for these caches, so the split introduces no missed invalidation hook.

I also checked that no consumer loses anything: `hasIcon` is the only thing the frontend derives from the bulk indexes (`TaskIcon` renders from `icon.svg`), and `ui-ee/src` never calls these routes at all.

Findings below, most significant first.

### 1. The groups regression test is order-dependent and can silently stop guarding

`shouldAnswerGroupsRatherThanClassesForGroupIcons` only fails against the old code if something has already populated `ZERO_ARG_KEY` with the *class* index. If `JUnit` happens to run it before any `GET /plugins/icons`, `loadPluginsIcon()` takes the cache miss itself, returns real group data, and the test passes with the bug fully present. It works today, as your revert check shows, but the ordering is `JUnit`'s deterministic-yet-arbitrary method order: adding or renaming a test in `PluginControllerTest` can reshuffle it and defang the guard without anyone noticing.

Worth making the poisoning explicit inside the test:

```java
client.toBlocking().exchange(HttpRequest.GET(PATH + "/icons"));   // populate the shared key first
Map<String, PluginIcon> groups = client.toBlocking().retrieve(...);
```

### 2. Iconless plugin groups now lose the ecosystem-catalog fallback

This is a consequence of fixing the groups route, not of the byte-stripping. Both catalog helpers test key *presence*, not the `hasIcon` flag - `PluginCatalog.vue` (`icons.value[cls] !== undefined`) and `PluginList.vue` (`!!icons.value?.[cls]`) - and `loadPluginsIcon()` puts an entry for *every* group and subgroup, including ones where `iconAndMonochrome("plugin-icon")` is empty (`hash: null`, `hasIcon: false`).

Before this `PR`, `groupIcons` held the class index, so group keys were simply absent, `TaskIcon`'s `providedIcon` stayed undefined, and the lazy path ran `loadEcosystemIcon()` against `api.kestra.io`. Now the entry exists, `TaskIcon` skips `loadIcon` because `providedIcon` is truthy, and `hasIcon: false` renders the generic file icon, so the ecosystem probe never happens.

Impact is small: across the 203 jars in my local plugin directory, only `storage-cloudflare` and `storage-seaweedfs` ship no `icons/plugin-icon.svg`, and every real subgroup package I checked (121 of 121, plus all 14 `io.kestra.plugin.core` subgroups) ships its own. So it is 2 tiles, not a page-wide regression, but they would go from a real logo to a generic one. Cheapest fix is to drop hashless entries in `fetchGroupIcons` so groups keep the lazy/ecosystem path, leaving the class index untouched since "every class gets an entry, iconless ones fall back to the generic icon" is deliberate there.

### 3. `withoutIconBytes` rebuilds the whole response on every request

It allocates a `HashMap` plus one `PluginIcon` per entry (707 on your instance) on each call, while the thing it strips is itself cached. Given `CACHE_DIRECTIVE` is `public, max-age=3600` this is rarely hit, so it is not a defect, but stripping once at index-build time, or a second small `@Cacheable` holding the metadata-only view, would make the served map a constant. Worth it if you expect this route to be hit by many concurrent cold clients.

### 4. `icon` is now permanently `null` on both bulk routes while staying in the schema

The response type is unchanged, which is what keeps the frontend working, but for any third party calling `GET /api/v1/plugins/icons` this is a behavioural break with no signal in the `API` contract: the field is still advertised, just always empty. Consider `@Schema(deprecated = true)` on it for those routes, or a `BREAKING CHANGE:` footer on the commit, so it shows up in the release notes rather than being discovered at runtime.

### 5. Nits

- The new `@Operation` descriptions carry the rationale for the fix ("Inlining them here made this index two orders of magnitude larger... parsing it blocked the main thread"). Consumer-facing `OpenAPI` docs are better off stating current behaviour only; the history reads well in the `PR` but will age oddly in the generated `SDK` comments, which is where it ends up verbatim.
- `HttpResponse.<T> notModified()` is unrelated formatting churn: the added space is not required and the rest of the diff is otherwise tight.
- No frontend unit test covers the exact contract this `PR` introduces: `{icon: null, hash: "..."}` -> `hasIcon: true`. `pluginsLoadIcon.spec.ts` covers the single-icon route only. One case in the `fetchIcons` path would lock it in.
- `maximum-weight` on the two new caches is entry *count*, not bytes: with no `Weigher` bean in `OSS` or `EE`, `micronaut-cache-caffeine` falls back to `Weigher.singletonWeigher()` (verified in the 5.3.0 bytecode). So `10485760` means ~10.5M entries, i.e. effectively unbounded. Harmless and consistent with the existing `default` entry, just not the 10 MB memory bound it looks like.

`CI` is green apart from `submit-gradle`, which failed on a `Connect Timeout Error` uploading the dependency snapshot: infra flake, unrelated to this change.

elevatebart (approved):

fdelbrayelle:
Thanks @MilosPaunovic — the cache-collision cross-check (`inputDocumentation(Type)` being the only other `@Cacheable("default")`, and no `@CacheInvalidate` to miss) and the plugin-jar survey on finding 2 were exactly the parts I couldn't verify myself. All five addressed, rebased onto develop, one commit.

### 1. Order-dependent regression test — fixed

You're right that it was leaning on JUnit's method order. The test now poisons the shared key itself before asserting:

```java
// Ask for the class index first: both indexes are keyed by no argument, so before they were split apart
// this call is what filled the shared cache key that the group index then answered from. Without it the
// assertions below would pass against the unsplit code whenever this test happened to run first.
client.toBlocking().retrieve(HttpRequest.GET(PATH + "/icons"), Argument.mapOf(String.class, PluginIcon.class));
```

Re-ran the neuter check with the explicit ordering in place: pointing `loadPluginsIcon()` back at `plugin-icons` fails only `shouldAnswerGroupsRatherThanClassesForGroupIcons`, and it passes again when restored.

### 2. Iconless groups losing the ecosystem fallback — fixed

Confirmed both call sites test presence, not the flag (`PluginCatalog.vue:262`, `PluginList.vue:259`). Took your cheapest fix: hashless entries are dropped in the group path only, so those groups keep falling through to `loadEcosystemIcon()`. The class index is untouched, since "every class gets an entry, iconless ones fall back to the generic icon" is deliberate there.

```ts
// The group index holds an entry for every group and subgroup, iconless ones included. Callers treat any entry as
// "this group has an icon", so keeping the iconless ones would render the generic icon instead of letting
// TaskIcon fall through to its lazy lookup and the ecosystem catalog.
function toGroupIconDataMap(raw) { … .filter(([, icon]) => icon.icon != null || icon.hash != null) … }
```

Covered by a new test asserting an iconless group is absent from `groupIcons` while a hashed one is present.

### 3. Rebuilding the response per request — fixed

The metadata view is now cached in its own right, so the served map is a constant:

```java
@Cacheable("plugin-icons-metadata")
protected Map<String, PluginIcon> pluginIconsMetadataIndex() {
    return withoutIconBytes(pluginIconsIndex());
}
```

Same for the groups index. Self-invocation still goes through the interceptor here (Micronaut's generated `$Intercepted` subclass overrides the method, unlike a JDK-proxy setup), which the existing `resolvePluginIcon()` → `pluginIconsIndex()` call already relied on.

### 4. `icon` permanently null while staying in the schema — commit footer

Marking the field `@Schema(deprecated = true)` would have been wrong, since `GET /plugins/icons/{cls}` still legitimately returns bytes and the model is shared. So it's a `BREAKING CHANGE:` footer instead, to land in the release notes:

> `GET /api/v1/plugins/icons` and `GET /api/v1/plugins/icons/groups` no longer return icon bytes. The `icon` field stays in the response schema but is always null on these two routes; fetch the bytes per class from `GET /api/v1/plugins/icons/{cls}/icon.svg`, or read `hash` to tell whether a class has an icon at all.

### 5. Nits

- **`@Operation` rationale** — trimmed to current behaviour only. Fair point about it ending up verbatim in the generated SDK comments; the history belongs here, not in the contract.
- **Frontend test for the new contract** — added: `{icon: null, hash: "abc123"}` → `hasIcon: true`, in the `fetchIcons` path.
- **`maximum-weight`** — good catch, and it changed the fix: since every one of these methods takes no argument, each cache holds exactly one entry. So they're now `maximum-size: 1` with a comment saying why they're separate, rather than a number that reads like a 10 MB bound and isn't.
- **`HttpResponse.<T> notModified()`** — this one I'm leaving, because it isn't mine: it's `spotlessApply` from the pre-commit hook. Removing the space fails `./gradlew :webserver:spotlessJavaCheck`. Running `spotlessApply` across the module reformats 39 other files, so develop is broadly non-compliant and any file touched picks up its formatting — I reverted those 39 and kept only this one line.

One correction to my own QA report while I'm here: I reported `pluginsFetchIcons` / `pluginsLoadIcon` as failing in my environment. That was my invocation missing `--localstorage-file`, not a repo problem. With it, both specs pass (15 tests). The four store specs that do fail locally — `dashboardDanglingId`, `mcp`, `dashboardFavouriteCleanup`, `logsCursor` — fail identically on a stashed clean tree, so they're unrelated to this PR.

ammeek (approved):
LGTM

PR #18587, opened 2026-08-25
feat(design-system): let KsSegmented options carry an icon

elevatebart:
---

### 🔗 Related Issue

Companion to https://github.com/kestra-io/kestra-ee/issues/5369 — the EE PR carries the closing keyword. Same branch name in both repos so CI pairs them.

### ✨ Description

`KsSegmented` object options may now declare an `icon` component, rendered before the label:

```ts
const options = [
    {label: t("user"), value: "USER", icon: Account},
    {label: t("group"), value: "GROUP", icon: AccountGroup},
]
```

The custom item rendering only kicks in when an icon is actually present, so every existing usage keeps Element Plus's own markup untouched.

Needed by the EE Access dialog, whose **Access Type** control has an icon per access type in [the design](https://www.figma.com/design/ew0uXk0NRXJ2NBBJTNe2n1/01_UI?node-id=7837-101788). The alternative was a `:deep()` or a cast at the call site, which the design-system rules rule out — extend, don't fork.

_Why did the segmented control adopt a cat? It had exactly one option and wanted a `purrfect` fit._ 🐱

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit` — 196 files, 1844 tests; plus design-system `unit:test` and `storybook:test` — 80 files, 481 tests)
- [ ] Translations are complete if `en.json` changed — n/a, no translation change in this PR

### 📸 Screenshots

_To be attached — no running instance in the environment this was built in. The `WithIcons` story added in `KsSegmented.stories.ts` shows the change on its own._

### 📝 Additional Notes

- Element Plus types its own slot item as `Option`, which knows nothing of our `icon`, hence the small `asObjectOption` narrowing helper rather than a cast.
- `WithIcons` story, a unit test, and the `ui/AGENTS.md` component table are updated alongside.

### 🔗 Related PRs

- [feat(iam): grant access to a service account from the Access page](https://github.com/kestra-io/kestra-ee/pull/10408) — the EE PR this unblocks

https://claude.ai/code/session_01UwMqcHd2fDzChLSwsfngPH


flcarre (approved):
Looks good. I walked all eight existing `KsSegmented` call sites in `ui/src` and none passes an `icon` or overrides the default slot, so `hasIcons` stays false everywhere and Element Plus's stock rendering path is untouched. Story and unit test both present, spacing via `--ks-spacing-1`, no `:deep()`.

Worth flagging for sequencing: kestra-io/kestra-ee#10408 passes `icon:` in its `KsSegmented` options and depends on this landing first. Because the options array there is a typed variable rather than a fresh object literal, TypeScript does no excess-property check, so if EE merges first it goes green and silently ships a text-only control with the icons missing. Order is this PR, then the design-system bump, then EE.

Two small things, neither blocking:

* `labelOf` hardcodes `.label`, so it would ignore an Element Plus `props` alias mapping (`:props="{label: 'name'}"`). No current caller does that, so it is theoretical, but worth a thought if `props` is ever exposed on the wrapper.
* The icon is fixed at `size="sm"` regardless of the control's own `size`, so an icon inside a `size="small"` segmented is proportionally oversized. The `WithIcons` story only shows the default size, so it does not show up in review. Adding icons to the `Sizes` story would surface it.

elevatebart:
Thanks — both points were fair, and the second was a real defect. Pushed `12ee981`:

**Icon size.** Fixed. The icon now maps to the control's own size (`large` → `base`, `default` → `sm`, `small` → `xs`) instead of being pinned at `sm`. Your suggestion about the story was the right call: the `Sizes` story now carries an icon row alongside the plain one, so a future mismatch shows up in review rather than in production, and a unit test locks the mapping.

**`labelOf` hardcoding `.label`.** Leaving this as is, deliberately. The wrapper doesn't forward a `props` alias mapping today and no caller sets one, so supporting it would mean adding a prop nothing uses — which the design-system rules specifically warn against ("only expose props that are actually used somewhere in the codebase; speculative props rot"). If `props` is ever exposed on the wrapper, `labelOf` and `iconOf` should both be revisited together, since both hardcode their key.

**Sequencing.** Agreed, and thanks for spotting it independently — that ordering is exactly what I'd concluded, including the reason EE would go green while shipping a text-only control. I'm merging this one first; EE waits on it.

---

PR #18588, opened 2026-08-25
feat(dashboards): recycle offscreen dashboard charts and cap their canvas pixel ratio

### LINKED ISSUE #14234: [Dashboard] Lazily load dashboard graphs and limit concurrent loading
### Problem Statement
All dashboard graphs are currently loaded eagerly, which can cause unnecessary load and slow initial render times, especially on dashboards with many visualizations.

### Possible Solution

We should improve performance by lazily loading graphs only when they enter the viewport and limiting the number of concurrent graph loads.

- Load graphs only when they become visible in the viewport
- Limit concurrent graph loading (e.g. max 2 graphs at a time - tdb)
- Queue remaining graphs and load them as earlier ones complete

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/14234.
Follows up on https://github.com/kestra-io/kestra/pull/18199.

## What

https://github.com/kestra-io/kestra/pull/18199 made dashboard charts load lazily and capped concurrent chart requests at 3, which fixed the request burst. Users still report out-of-memory crashes on large screens, because that `PR` throttled **requests** but never bounded **memory**:

- **Nothing was ever released.** `Sections.vue` unobserved a block on first intersection and kept the chart mounted for good, so scrolling a long dashboard grew the number of live `ECharts` canvases monotonically. A `FIFO` queue spreads the allocations over time; the peak at the end is identical.
- **Canvas memory scales with the square of the device pixel ratio.** Every dashboard chart used the default `CANVAS` renderer at full device resolution, so the same chart costs several times more on a wide high-DPI monitor than on a laptop. That is why the reports are large-screen-only.
- **The lazy gate barely fires at that size.** `rootMargin: "200px 0px"` against a 1400px-plus viewport means the first observer callback activates most of the visible page, and 200px of lookahead is a rounding error.

Lowering the concurrency cap further would only make big dashboards slower to fill, so this `PR` goes after the retained memory instead.

## Changes

**Recycle canvas charts that scroll far away** - new `useLazyChartBlocks` composable. Two `IntersectionObserver`s per dashboard: activation at `200px` (unchanged), retention at `1200px`. A chart mounts when it enters the activation band and unmounts again once it leaves the retention band, so memory becomes O(visible charts) rather than O(total charts). The gap between the two bands is the hysteresis that stops ordinary scrolling from thrashing.

- Only `Bar`, `Pie` and `TimeSeries` are recycled (`isCanvasChart`). `KPI`, `Table` and `Markdown` cost almost nothing, and `Table` holds its own page, size and tab state, so they stay mounted once activated.
- A recycled chart's `KsSkeleton` placeholder inherits the height the chart had, so collapsing a block never shifts the content below it.
- Re-entering the viewport refetches through the existing `chartLoadQueue`. That is a deliberate trade: fresh data on re-entry, at the cost of a repeat request if a user scrolls past the same chart twice.

**Cap the canvas pixel ratio** - new `maxPixelRatio` prop on `KsEchart` (forwarded through `KsBar` and `KsPie`), set to `1.5` for dashboard charts via `DASHBOARD_CHART_MAX_PIXEL_RATIO`. Opt-in, so no other chart surface changes behaviour.

**Batch activation across frames** - at most 4 charts mount per frame, so a tall viewport no longer mounts its whole first screen in one long task. Smoothness rather than memory, but it is what makes the first paint of a 4K dashboard incremental.

## Interaction with chart export

https://github.com/kestra-io/kestra/pull/18661 landed while this was open and made `exportChart` read a chart's `exportParameters()` out of a `chartsComponents` map. Recycling means a chart can be absent from that map, which would silently drop pagination and quick-filter from an export - but only `Table` implements `exportParameters`, and `Table` charts are never recycled, so the export path cannot be affected. The map is also a better fit for recycling than the array it replaced, since it self-cleans on unmount.

## Measurements

30-chart dashboard (the `YAML` from the issue), `2560x1400` viewport, `devicePixelRatio` 2, headless Chromium, scrolled top to bottom:

| | `develop` | this `PR` |
|---|---|---|
| Canvas backing store, after full scroll | 57.2 MB | **10.7 MB** |
| Live canvases, after full scroll | 12 | **4** |
| Canvas backing store, at first paint | 19.1 MB | **7.3 MB** |
| `JS` heap, after full scroll | 130.4 MB | 129.4 MB |

The `JS` heap is deliberately unremarkable: canvas backing stores are allocated outside the `JS` heap, and that off-heap memory is what the tab actually runs out of. Rendering is pixel-identical between the two branches at `devicePixelRatio` 2.

Independently confirmed by @Pradumnasaraf, who compared `develop` against this branch and saw canvas memory drop after a full scroll with no behaviour change.

## Not in scope

Nothing bounds points-per-series on the backend, so a single `TimeSeries` over `P365D` with a high-cardinality `colorByColumn` can still be a very large payload on its own. Worth a separate issue rather than folding in here.

## Tests

- `useLazyChartBlocks.spec.ts` - 12 cases covering activation, retention, hysteresis, per-frame batching, non-recyclable types, placeholder height, the no-`IntersectionObserver` fallback and observer teardown.
- `KsEchart.test.ts` - 4 cases for `maxPixelRatio` (capping, pass-through below the cap, absent prop, missing `devicePixelRatio`).
- New `CappedPixelRatio` Storybook story.
- On the merged tree: `npm run check:types`, `npm run test:unit` (1952), the design-system unit suite (811) and the `Charts` Storybook tests all pass; `oxlint` and `eslint` clean on the changed files.


MilosPaunovic:
@Pradumnasaraf @Skraye Could you help me out by testing the `Dashboard` from this PR, how it loads and ohw memory intensive it is, does the browser breaks if there are many chats, or the screen is large?

Pradumnasaraf (commented):
Hey @MilosPaunovic, it's better actually. I compared both develop vs this. Canvas memory after a full scroll dropped. Everything is working like previously!

PR #18592, opened 2026-08-25
fix(jdbc): stop the parent execution when a subflow execution is killed

AcevedoR:
Killing a subflow's child execution left the parent `RUNNING` when the `KILLED` taskrun was joined, so `handleNext()` still scheduled the successor tasks — they ran to `SUCCESS` on an execution that then landed in `KILLED`.

The existing cascade can't catch this: it's guarded on `taskRun.getParentTaskRunId() != null`, and `killParentTaskruns` only walks up a taskrun tree, so a top-level `Subflow` task is never reached.

Fix: transition the execution to `KILLING` when a `KILLED` subflow taskrun is joined, so the existing `KILLING` guard in `ExecutorService.process()` engages before next tasks are computed. Keying on the taskrun state keeps it `transmitFailed`-aware — with `transmitFailed: false`, `guessState` returns `SUCCESS` and the branch never fires.

New test `JdbcRunnerTest.killSubflowExecutionStopsParent` fails on the unpatched branch and passes with the fix.

Paired EE PR (Kafka executor has its own copy of this join): kestra-io/kestra-ee#10416. Targets `releases/v1.3.x` only; 2.0 not addressed here.

Related: #12557

### DIFF
--- a/core/src/test/resources/flows/valids/subflow-kill-child-of-parent.yaml
+++ b/core/src/test/resources/flows/valids/subflow-kill-child-of-parent.yaml
@@ -1,0 +1,7 @@
+id: subflow-kill-child-of-parent
+namespace: io.kestra.tests
+
+tasks:
+  - id: sleep
+    type: io.kestra.plugin.core.flow.Sleep
+    duration: PT60S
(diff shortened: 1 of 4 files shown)


AcevedoR:
wrong solution, will be studied later

PR #18597, opened 2026-08-25
fix(executor): do not resume a Pause twice when its pauseDuration expires after a manual resume

### LINKED ISSUE #18507: [Bash2.0] A manually-resumed Pause is resumed again by its pauseDuration timer, wiping onResume and failing the run
## Describe the issue

When a Pause task with `pauseDuration` is resumed manually and the execution is still running when `pauseDuration` elapses, the Pause is resumed a second time by the leftover timer. The second resume carries no inputs, so it wipes the `onResume` inputs from the manual resume and the execution fails.

## Steps to reproduce

1. Create this flow (the `after` sleep is longer than `pauseDuration`, so the run outlives the timer):
   ```yaml
   id: pause_clobber_test
   namespace: company.team
   tasks:
     - id: wait_for_approval
       pauseDuration: PT10S
       onResume:
         - id: reason
           type: STRING
     - id: after
       duration: PT30S
     - id: show
       format: "reason={{ outputs.wait_for_approval.onResume.reason }}"
   ```
2. Execute it. It pauses at `wait_for_approval`.
3. Resume it within a few seconds with `reason=approved`.
4. Let it keep running past the 10s `pauseDuration` (the 30s `after` is still sleeping).

## Actual behaviour

At the 10s mark the Pause is resumed a second time. `outputs.wait_for_approval.onResume.reason` is gone, `show` fails with `Unable to find 'onResume' used in the expression`, and the execution ends FAILED, even though it was approved and nothing failed.

If `after` is short enough that the execution finishes before `pauseDuration` elapses (e.g. `PT2S`), the run succeeds and `reason` is kept.

## Visuals

Image

## Environment

- Kestra Version: 2.0.0-rc10

AcevedoR:
Closes #18507

### What

The pause `ExecutionDelay` row is only removed when it expires — nothing deletes it when the Pause is resumed manually. When the execution outlives `pauseDuration`, `DefaultExecutor.executionDelayLoop()` resumed the Pause a **second** time: it only checked that the execution was not terminated/KILLING, never that the task run was still `PAUSED`. `executionService.markAs(..., RUNNING)` then re-generated the Pause outputs via `generateOutputs(onResumeInputs = null, ...)`, wiping the `onResume` inputs from the manual resume and failing every task referencing them.

This also affected `behavior: FAIL/WARN/CANCEL`, where a manually-resumed Pause would later be failed/cancelled by its own leftover timer.

### Fix

Skip an expired `RESUME_FLOW` delay whose task run is no longer `PAUSED`, returning `null` from the lock callback so nothing is persisted or emitted.

Chosen over deleting the delay on resume: `ExecutionDelayStateStore` has no delete API, and a delete-on-resume would still race the delay loop. This guard is idempotent regardless of ordering.

### Repro (before the fix, real instance)

```
--- t=5s  (after manual resume with reason=approved)
exec: RUNNING
   wait_for_approval SUCCESS outputs= {"onResume":{"reason":"approved"},"resumed":{"on":"...33:20","to":"SUCCESS"}}
--- t=15s (past the PT10S pauseDuration, the next task is still sleeping)
exec: RUNNING
   wait_for_approval SUCCESS outputs= {"resumed":{"on":"...33:27","to":"SUCCESS"}}   <- onResume gone
--- t=40s
exec: FAILED  (show: Unable to find `onResume`)
```

The `resumed.on` timestamp moving from the manual resume to the `pauseDuration` expiry confirms the double resume.

### Tests

Added `pauseRunDurationManuallyResumed` to `AbstractRunnerTest`/`PauseTest.Suite` with `flows/valids/pause-duration-manual-resume.yaml` (`Pause` PT1S with an `onResume` input → `Sleep` PT5S → a task reading `outputs.pause.onResume.reason`), so it runs on H2, Postgres and MySQL. It asserts the Pause has exactly one `PAUSED`/`RUNNING` history entry and that the `onResume` value survives.

- Without the fix it fails with the reported ``Unable to find `onResume` used in the expression `{{ outputs.pause.onResume.reason }}` ``.
- With the fix it passes, and the full `H2RunnerTest` (101 tests) plus `core PauseTest` stay green.

Also verified on a rebuilt local instance: the issue's flow now ends `SUCCESS` with `reason` preserved.

### EE

No EE-side change is needed: `kestra-ee` includes this `:executor` module and has no executor of its own, so the guard covers every backend. `ElasticSearchExecutionDelayStateStore` has the same delete-on-expiry-only semantics but feeds this same consumer.

### DIFF
--- a/core/src/test/java/io/kestra/core/runners/AbstractRunnerTest.java
+++ b/core/src/test/java/io/kestra/core/runners/AbstractRunnerTest.java
@@ -416,6 +416,12 @@ public abstract class AbstractRunnerTest {
     }
 
     @Test
+    @LoadFlows(value = { "flows/valids/pause-duration-manual-resume.yaml" }, tenantId = "pause-manual-resume")
+    public void pauseRunDurationManuallyResumed() throws Exception {
+        pauseTest.runDurationManuallyResumed("pause-manual-resume", runnerUtils);
+    }
+
+    @Test
     @LoadFlows({ "flows/valids/pause-timeout.yaml" })
     public void pauseRunTimeout() throws Exception {
         pauseTest.runTimeout(runnerUtils);
--- a/core/src/test/java/io/kestra/plugin/core/flow/PauseTest.java
+++ b/core/src/test/java/io/kestra/plugin/core/flow/PauseTest.java
@@ -17,6 +17,7 @@ import io.kestra.core.junit.annotations.ExecuteFlow;
 import io.kestra.core.junit.annotations.KestraTest;
 import io.kestra.core.junit.annotations.LoadFlows;
 import io.kestra.core.models.executions.Execution;
+import io.kestra.core.models.executions.TaskRun;
 import io.kestra.core.models.flows.Flow;
 import io.kestra.core.models.flows.State;
 import io.kestra.core.queues.QueueException;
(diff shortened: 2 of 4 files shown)


loicmathieu (approved):
LGTM

PR #18608, opened 2026-08-25
fix(core): reject negative or zero MAX_DURATION SLA durations at save time

AcevedoR:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra-ee/issues/10262.

### ✨ Description

A flow SLA of type `MAX_DURATION` with a negative `duration` (e.g. `PT-1H`) was accepted at save time and the flow showed as Valid. At execution the SLA was breached on the very first evaluation, so with `behavior: FAIL` every execution failed immediately. Zero (`PT0S`) had the same effect.

The `duration` field already had an upper bound (`@DurationMax`, 10 years) but no lower bound. This PR adds a `@DurationPositive` constraint — a new annotation + Micronaut validator mirroring the existing `@DurationMax` pair — and applies it to `MaxDurationSLA.duration`. Saving the repro flow from the issue now fails validation with `duration 'PT-1H' must be positive` instead of producing a flow whose every execution fails.

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched modules (`./gradlew :core:test --tests "io.kestra.core.validations.*" --tests "io.kestra.core.models.flows.*"`)

### 📝 Additional Notes

Evidence: `MaxDurationSLAValidationTest` has four cases — positive duration valid, negative rejected, zero rejected, and the issue's exact flow YAML rejected through `FlowService.validate` (the save-time path). Three of the four fail on the commit before this one; all four pass with the fix.

Scope kept to the SLA field. `@DurationPositive` is reusable, but I did not retrofit it onto retry intervals or trigger windows in this PR.

🐱 Why did the cat's SLA always fail? It set `duration: PT-9L` — it wanted its nine lives *back*.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/sla/types/MaxDurationSLA.java
+++ b/core/src/main/java/io/kestra/core/models/flows/sla/types/MaxDurationSLA.java
@@ -12,6 +12,7 @@ import io.kestra.core.models.flows.sla.SLA;
 import io.kestra.core.models.flows.sla.Violation;
 import io.kestra.core.runners.RunContext;
 import io.kestra.core.validations.DurationMax;
+import io.kestra.core.validations.DurationPositive;
 
 import jakarta.validation.constraints.NotNull;
 import lombok.Getter;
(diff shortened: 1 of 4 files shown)


AcevedoR:
Superseded by the same `@DurationMin` change on develop that closes kestra-ee#10262 (see #18648, closed alongside this one — the two overlapped).

PR #18609, opened 2026-08-25
fix(core): apply plugin defaults to the task inside a TemplatedTask spec

### LINKED ISSUE #8807: TemplatedTask does not apply Plugin Default settings
### Describe the issue

When using `TemplatedTask`, none of the PluginsDefault settings defined in Kestra's config are applied

For example, I have this in my kestra config file:

```
plugins:
    defaults:
      - type: io.kestra.plugin.scripts.runner.docker.Docker
        host: "tcp://127.0.0.1:2375"
```

However, when rendering the `TemplatedTask`, it fails because Kestra tried to connect to the unix:///var/run/docker.sock default socket.

If I define the host: "tcp://127.0.0.1:2375 again inside the TemplatedTask it works fine, however, it defeats the purpose of the Plugins Defaults and I'd also argue it totally bypasses any `force: true` I define at the kestra config level, which can be used to evade security measures.

Can we have `TemplatedTask` inherit whatever is defined in Plugin Defaults?

Related to https://github.com/kestra-io/kestra/issues/8793

### Environment

- Kestra Version: v0.22.6

ammeek:
Closes #8807.

## Problem

With a namespace default of `message: "{{ vars.message }} xyz"` on `io.kestra.plugin.core.log.Log`:

```yaml
tasks:
  - id: test1
    type: io.kestra.plugin.core.log.Log          # logs "abc xyz"
  - id: test2
    type: io.kestra.plugin.core.templating.TemplatedTask
    spec: |
      type: io.kestra.plugin.core.log.Log        # NPEs
```
PluginDefaultService.innerInjectDefault walks the flow as a Map and applies defaults to
nodes carrying a type key. TemplatedTask.spec is a string, so the walk stops there, and
TemplatedTask.run then deserializes its task with no defaults layer. Global, namespace and
flow defaults all feed that one walk, which is why all three fail identically. forced: true
defaults are skipped too, so a templated task escapes enforcement.

Changes
- PluginDefaultService: innerInjectDefault returns the defaulted flow and the defaults it
  resolved; the pure map transform is split out as applyDefaults and exposed as
  applyResolvedDefaults(Map, List<PluginDefault>) for a single plugin definition.
- parseFlowWithAllDefaults stamps the resolved defaults onto every TemplatedTask, reusing
  the task-tree walk the Template tenant injection already does. Skipped on the
  version-defaults-only path so nothing enters persisted flow source, and always overwritten so
  a value written in a flow is never honoured.
- TemplatedTask.run parses the rendered spec to a map, applies the defaults, then builds and
  validates the task via runContext.validate.

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/templating/TemplatedTask.java
+++ b/core/src/main/java/io/kestra/plugin/core/templating/TemplatedTask.java
@@ -1,23 +1,32 @@
 package io.kestra.plugin.core.templating;
 
+import java.util.List;
+import java.util.Map;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 import io.kestra.core.exceptions.IllegalVariableEvaluationException;
 import io.kestra.core.models.annotations.Example;
 import io.kestra.core.models.annotations.Plugin;
+import io.kestra.core.models.flows.PluginDefault;
 import io.kestra.core.models.property.Property;
 import io.kestra.core.models.tasks.Output;
 import io.kestra.core.models.tasks.RunnableTask;
 import io.kestra.core.models.tasks.Task;
+import io.kestra.core.runners.DefaultRunContext;
 import io.kestra.core.runners.RunContext;
 import io.kestra.core.serializers.JacksonMapper;
+import io.kestra.core.services.PluginDefaultService;
+import io.kestra.core.utils.ListUtils;
 
+import io.swagger.v3.oas.annotations.Hidden;
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.constraints.NotNull;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
+import lombok.Setter;
 import lombok.ToString;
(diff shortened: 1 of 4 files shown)


nkwiatkowski (approved):

core/src/main/java/io/kestra/plugin/core/templating/TemplatedTask.java
```diff
@@ -52,11 +63,27 @@ public class TemplatedTask extends Task implements RunnableTask<Output> {
...
+     * after the flow was parsed and where the defaults can no longer be resolved. They are therefore resolved
+     * with the rest of the flow's defaults and carried here — see
+     * {@link PluginDefaultService#applyResolvedDefaults(Map, List)}.
+     * </p>
+     */
+    @Hidden
+    @Setter // we have no other option here as we need to update the task inside the flow when parsing it
+    @EqualsAndHashCode.Exclude
+    @ToString.Exclude
+    private List<PluginDefault> resolvedPluginDefaults;
```
fhussonnois, line 80: I'm really not a fan to use the data model class for holding such runtime info, but I guess there is no other choice.

fhussonnois (approved):
LGTM!

PR #18613, opened 2026-08-25
fix: Setup Docker host access within the devcontainer.

### LINKED ISSUE #18612: [Configuration] Setup Docker host access inside the devcontainer
### Describe the issue

Currently, we cannot run any docker commands within the devcontainer such as end-to-end tests for example using the command: `npm run test:e2e` and others that require Docker since the devcontainer doesn't have access to the docker host via a socket mount and docker installation within the devcontainer config.

This issue proposes updating the environment to support that which is quite simple as we just need to mount the docker socket in the devcontainer config then also install the docker engine within the Dockerfile. 

### Environment

- Kestra Version: develop

m-t-a97:
### 🔗 Related Issue

Closes #18612 

### ✨ Description

Updates the devcontainer environment to support accessing the Docker host so that any Docker related commands and actions can be ran directly from within the devcontainer. This is needed to run e2e tests and other commands that require access to Docker.

---

Instructions to test this PR:

- Open the Kestra project in VSCode then run the command from the command palette to open the project in a devcontainer
- Verify you can access Docker via the shell by running: `docker ps`


m-t-a97:
## Screenshots

image

Shows successfully running Docker commands within the devcontainer

Pradumnasaraf (commented):
The Docker install itself builds fine and `docker ps` will work. But I don't think the socket mount gets us to e2e tests, which is the stated goal.

elevatebart (commented):
Thanks for picking this up — the devcontainer not reaching Docker is a real gap.

I tested the branch on Docker Desktop 4.75.0 / macOS arm64 (engine 29.5.2). `docker ps` does work, as your screenshot shows. But the socket mount gives *docker-outside-of-docker*: containers you start are siblings on the host, not children (`docker info --format '{{.Name}}'` → `docker-desktop`, and `docker ps` lists unrelated containers from my host). That breaks the e2e suite in two ways `docker ps` can't reveal.

**1. Compose bind mounts resolve against the host filesystem.** `docker-compose-postgres.yml` mounts `./data/entrypoint.sh` etc., which inside the container is `/workspaces/kestra/...` — a path the host daemon doesn't have.

```
 Container kestra-e2e-backend  Starting
Error response from daemon: mounts denied:
The path /workspaces/kestra/ui/tests/e2e/data/entrypoint.sh is not shared from the host…

$ docker ps -a
kestra-e2e-backend   Created      ← never starts
```

`npm run test:e2e` therefore dies after 31s with zero tests run. (That's the Docker Desktop symptom; on native Linux there's no file-sharing gate, so the daemon silently creates empty *directories* at those paths and `/app/entrypoint.sh` becomes a directory instead. Same outcome.)

**2. Published ports land on the host's loopback.** Compose publishes `9011:8080`, but `start-e2e-tests-backend.sh` polls `127.0.0.1:9011` and `playwright.config.ts` uses `baseURL: http://localhost:9011` — different netns. Isolated repro:

```
$ docker run -d --name porttest -p 9011:80 nginx     # says 0.0.0.0:9011->80/tcp
from devcontainer  127.0.0.1:9011 -> 000 (connection refused)
from host          127.0.0.1:9011 -> 200
```

### Fix: nested daemon instead of the host socket

Drop the `mounts` entry in `devcontainer.json` and add:

```json
"features": {
  "ghcr.io/devcontainers/features/docker-in-docker:2": {}
},
```

Then delete the Docker apt block from the `Dockerfile` — the feature supplies daemon, CLI, buildx and compose, so those 23 lines go away.

I rebuilt with that and re-ran everything: both problems disappear, no change needed to any e2e script, and `npm run test:e2e` gives **100 passed (2.6m)**. (One failure, `blocks-flow-properties.spec.ts:107` — an unrelated flake, passes in isolation.)

Worth calling out in the description: the feature runs the devcontainer `--privileged`. That's a genuine trade-off against the socket mount, not a free win — but it's what makes the suite work.

m-t-a97:
Thanks guys, great breakdown and thanks for taking the time to look into it further @elevatebart .

I initially thought of a DIND approach but I thought that for the sake of development maybe the docker-outside-docker approach would've been enough for development since Docker Desktop usually runs on the host anyways. Also I'm on Windows using WSL so it's not as restrictive in terms of permissions as MacOS for example.

But as you've nicely outlined above the causes and issues of this mounting approach and the list of things that are not working properly.

The DIND approach does work very nicely using the VSCode feature you've outlined. It essentially sets up and creates the docker environment specifically for the container with its own docker daemon etc. So now when you run docker commands, those containers will be isolated from the host like they belong to the devcontainer instead of sibling containers on the host which is why this docker-outside-docker approach showed you unrelated containers.

Thanks again, I'll make the changes now.

m-t-a97:
Pasting the most recent commit message here:

chore: Replaced docker-outside-docker with docker-in-docker approach.
- Removed the socket mount and added the `docker-in-docker:2` feature to vscode devcontainer setup
- Added postCreateCommand to simplify installing npm dependencies and also installing browsers via playwright
- Removed docker installation from devcontainer Dockerfile
- Changed host.docker.internal:5432 → localhost:5432 since now the docker daemon in the devcontainer owns the containers instead of the user's host machine.

elevatebart (approved):
Great job

m-t-a97:
## Screenshots

image

Containers now isolated within the devcontainer and owned by the devcontainer's docker daemon.

image

e2e tests now successfully running using `npm run test:e2e`. Just to note regarding this on my end, 73 tests pass instead of 100 and it seems like it's the AI copilot tests that are failing. Do you know why that is @elevatebart ?

m-t-a97:
Screenshots here regarding the AI copilot tests:

image

m-t-a97:
image

image

m-t-a97:
image

Okay now the tests are passing locally in the devcontainer on this branch. I think maybe my branch was previously outdated but I pulled in some updates onto my branch earlier today from the main repo's branch and it seems to have fixed the issue. I think all is good now and this PR can be safely merged.

PR #18614, opened 2026-08-26
docs: add RepoCloud as a deployment option

cosark:
Adds [RepoCloud](https://repocloud.io) as a one-click deployment option in the Quick Start section,
alongside the existing AWS CloudFormation and Google Cloud Terraform options.

RepoCloud provides managed cloud hosting for open-source applications with one-click deployment.

- Adds new "Deploy on RepoCloud" subsection to Quick Start in `README.md`
- Uses the same markdown image button format (no explicit height) as the existing AWS Launch Stack button
- Links to: https://repocloud.io/details/Kestra/

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,8 @@ Deploy Kestra on Google Cloud Infrastructure Manager using [our Terraform module
 
 ### Deploy on RepoCloud
 
+Deploy Kestra on RepoCloud with one click:
+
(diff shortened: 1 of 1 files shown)


aj-emerich (commented):
We don't include deployment options like this in the README. Thanks for contributing, but for now this is not something we include.

PR #18639, opened 2026-08-26
fix(flows): scope task-property autocomplete to the current task type

### LINKED ISSUE #18393: [Bash2.0] Flow autocompletion shows all tasks properties not only the current task ones
### Describe the issue

When editing a flow and asking for autocompletion on a property name, all properties of all tasks are shown instead of only the properties of the current task.

Image

### Environment

- Kestra Version: develop

nkwiatkowski:
## What & why

In the flow YAML editor, triggering property-key autocompletion inside a task offered properties from **every** installed task type, not just the ones valid for the task's `type`. The Monaco YAML completion is schema-driven and models `tasks[]` as a union of all types; yaml-language-server does not narrow that union by the `type:` discriminator, so it returns the merged property set.

Fixes #18393.

## How

In the completion wrapper (`yamlLanguageConfigurator.ts`), when the cursor is **not** in a `type:` value context:
1. Resolve the task's `type` at the cursor (YAML node localization + walk to the enclosing `id`+`type` map).
2. `pluginsStore.load({cls: type, commit: false})` to get that type's schema.
3. Restrict the property-key suggestions (`CompletionItemKind.Property`) to that type's own keys; non-property suggestions are untouched.

**Fails open** — if the type can't be resolved or its schema can't be loaded, the original unscoped list is returned, so no context ever loses completion.

The cursor/type resolver and the scoping filter live in a new pure, monaco-free module (`taskCompletionScoping.ts`) so they can be unit-tested directly. The pre-existing required-property helper moved there too and now reuses the shared resolver (net −44 lines in the configurator).

## Tests
- New `tests/unit/composables/taskCompletionScoping.spec.ts` (5): resolves the type inside a task body, returns undefined outside a task, drops foreign properties, leaves value suggestions alone, fails open on empty/unknown keys. Written test-first (verified failing before the fix).
- `npm run check:types` ✓ · `eslint` on changed files ✓ · 20 related editor specs (pebble, subflow-link, flowAutoCompletion, FlowFileEditorTab) ✓

## Note for reviewers
The filter keys off `CompletionItemKind.Property` (what monaco-yaml assigns to object keys). If that ever changed, the filter no-ops (fail-safe, no regression) rather than misbehaving.


ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
```diff
@@ -391,10 +324,33 @@ export class YamlLanguageConfigurator extends AbstractLanguageConfigurator {
...
+            // body offers keys from unrelated plugins. Restrict them to the task's resolved `type`.
+            let scopedSuggestions = suggestions
+            if (!isTypeValueContext) {
+                const type = taskTypeAtCursor({
+                    source: model.getValue(),
+                    cursorIndex: model.getOffsetAt(position),
+                })
+                if (type) {
+                    try {
+                        const pluginDoc = await pluginsStore.load({cls: type, commit: false})
```
flcarre: **This drops most of the valid properties for every task.** `load()` is called without `all: true`, and `/plugins/{cls}` with `all=false` deliberately excludes every property declared on the base `Task` class — `JsonSchemaGenerator.generate` ignores base-declared fields, and all of `Task`'s are either unannotated or `@PluginProperty(hidden = true)`.

So for `io.kestra.plugin.core.log.Log` the returned key set is just `["message", "level"]`. Everything else that `flow-schema.json` declares and that we suggest today — `retry`, `timeout`, `description`, `disabled`, `runIf`, `allowFailure`, `logLevel`, `taskCache` — is now filtered out of the completion list. Same story for triggers.

Passing `all: true` here is the fix.

Worth flagging that the unit test masks this: `taskCompletionScoping.spec.ts` hardcodes `validPropertyKeys: ["id", "type", "message"]`, which is a set the API never actually returns for a real task. A test built from a real `/plugins` response (or at least one that includes the base-`Task` keys) would have caught it.

---

ui/src/composables/monaco/languages/taskCompletionScoping.ts
```diff
@@ -0,0 +1,101 @@
...
+        typeof (value as TaskLike).id === "string" &&
+        typeof (value as TaskLike).type === "string"
+    )
+}
+
+/**
+ * Resolves the task map enclosing the cursor by localizing the YAML node at (and just before) the
+ * cursor and walking its ancestry outwards to the nearest node that has both an `id` and a `type`.
+ */
+export function findTaskLikeAtCursor({source, cursorIndex}: {source: string; cursorIndex: number}): TaskLike | undefined {
```
flcarre, line 83: **Nested sub-objects get scoped against the wrong key set.** `isTaskLike` requires both `id` and `type`, so a cursor inside a nested object walks back out to the enclosing task, and the filter then keeps only that task's *top-level* keys:

- typing under `retry:` → `maxAttempts`, `maxDuration`, `behavior` are all dropped
- typing under `taskRunner:` → `pullPolicy` and friends are dropped

There's a second variant that's worse: a nested child task that already has an `id` but no `type` yet fails `isTaskLike`, so it resolves to the parent `Sequential`, whose own keys are `tasks` / `errors` / `finally`. `type:` itself is therefore never suggested at exactly the moment the user needs it.

The walk needs to stop at the nearest mapping rather than the nearest node carrying both `id` and `type` — and when that mapping isn't the task root, either resolve the sub-schema for it or fail open instead of filtering against the task's top-level keys.

---

ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
```diff
@@ -391,10 +324,33 @@ export class YamlLanguageConfigurator extends AbstractLanguageConfigurator {
                     return suggestion
                 })
 
+            // Monaco YAML derives property suggestions from the union of every task type, so a task
+            // body offers keys from unrelated plugins. Restrict them to the task's resolved `type`.
+            let scopedSuggestions = suggestions
```
flcarre, line 331: `ignoreNotFound: true` only suppresses the toast — the promise still rejects, so nothing gets cached. Combined with `incomplete: true` on the completion result, Monaco re-queries on every character, which means one GET → 404 per keystroke for a misspelled or not-installed task type.

This also fires while the user is still typing the `type:` value itself: `isTypeValueContext` goes false as soon as any text follows `type:`, and `getWordUntilPosition` stops at the last dot, so each character of `io.kestra.plugin.core.log.Lo…` produces its own failed request.

A negative cache (remember the types that 404'd for the session) or dropping `incomplete: true` once the type fails to resolve would both fix it.

---

ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
```diff
@@ -391,10 +324,33 @@ export class YamlLanguageConfigurator extends AbstractLanguageConfigurator {
...
+            if (!isTypeValueContext) {
+                const type = taskTypeAtCursor({
+                    source: model.getValue(),
+                    cursorIndex: model.getOffsetAt(position),
+                })
+                if (type) {
+                    try {
+                        const pluginDoc = await pluginsStore.load({cls: type, commit: false})
+                        const properties = pluginDoc?.schema?.properties?.properties
```
flcarre, line 352: Lower confidence / narrower impact than the other findings, but: the task's pinned `version` isn't forwarded to `load`, so a version-pinned task gets scoped against the latest schema. Where a property was added or removed between versions, the completion list won't match the version the flow actually runs.

---

flcarre (commented):
Thanks for this — scoping task-property completion to the actual task type is the right goal, and extracting `taskCompletionScoping.ts` is behaviour-preserving for the required-property helper that moved.

The problem is that the new scoping filter has two ways of removing valid completions, so as it stands the list ends up narrower than the schema allows wherever a type *does* resolve. Details inline; the two blocking ones are:

1. `load()` is missing `all: true`, so every base-`Task` property (`retry`, `timeout`, `runIf`, `allowFailure`, …) is filtered out of every task.
2. `isTaskLike` requires `id` + `type`, so a cursor inside a nested object (`retry:`, `taskRunner:`) is scoped against the enclosing task's top-level keys.

The "fails open when the type can't be resolved" design does hold, and I checked it — no context loses completion *entirely*. But that safety net doesn't help the two cases above, because there the type resolves fine and the filter runs with a key set that's simply incomplete.

**One finding is outside the diff**, so I couldn't anchor it inline — `ui/src/stores/plugins.ts:333`. The cache-hit branch of `load()` assigns `plugin.value = cachedPluginDoc` without checking `options.commit`, so the `commit: false` you're passing is silently ignored on every cache hit. Since the new call now runs per keystroke, it overwrites the shared `pluginsStore.plugin` — and `TaskEdit.vue:365` reads that store value *in preference to* `editorPlugin`, so the no-code "declared outputs" panel starts showing another task's outputs while you type. Either honour `commit` in the cache-hit branch, or read the cache directly here without going through `load()`.

Happy to look again once those are addressed.

---

ui/src/composables/monaco/languages/yamlLanguageConfigurator.ts
flcarre: The endpoint defaults `all` to `false`, which strips every inherited `Task` property from the returned schema, so the filter below keeps only the leaf type's own keys. For `core.log.Log` that is `level` and `message` — `retry`, `timeout`, `description` and the rest are dropped.

```suggestion
                        const pluginDoc = await pluginsStore.load({cls: type, commit: false, all: true})
```

Note this alone makes the call uncached (`plugins.ts:364`/`:383` skip the cache when `all` is set), so pair it with including the flag in the store's cache key.

flcarre (dismissed):
Thanks for taking this on, and for pulling the cursor/type resolver out into a pure monaco-free module — that part is a real improvement and it made the rest easy to reason about.

I QA'd this on a local instance (backend from `develop`, UI built from `95c39654af`) and hit two blocking problems. Both are reproducible; evidence below. I'm only listing the blockers here, not style-level remarks.

## 1. The scoping drops the current task's own inherited properties

The scoping lookup omits `all`, and the endpoint defaults it to `false`. With `all=false`, `ClassPluginDocumentation` passes `plugin.baseClass()` as `base` into `JsonSchemaGenerator.generate`, whose `ignoreCheck` excludes every field declared on the base class unless it carries `@PluginProperty(hidden = false)`. In `Task`, all inherited fields are `@PluginProperty(hidden = true)` and `id`/`type` are unannotated, so they are all excluded from `schema.properties.properties`.

Measured against the running instance:

| Type | keys the filter keeps | with `all=true` |
|---|---|---|
| `core.log.Log` | 2 — `level`, `message` | 17 |
| `core.debug.Return` | 1 — `format` | 16 |
| `core.flow.Sequential` | 3 — `errors`, `finally`, `tasks` | 18 |
| `core.trigger.Schedule` | 9 | 20 |

So inside a `Log` task the only property-key suggestion left is `level`: `retry`, `timeout`, `description`, `disabled`, `runIf`, `allowFailure`, `logLevel`, `workerSelector`, `taskCache` and `version` are all gone. At the same cursor position on `develop`, `retry` is offered.

The same code path also scopes `triggers:` bodies, since trigger entries carry `{id, type}` too — so `AbstractTrigger`'s inherited properties are dropped the same way. Worth deciding explicitly whether triggers are in scope for this PR, because they're currently affected without being mentioned or tested.

One caveat on the fix: adding `all: true` on its own reintroduces a network request per keystroke, because the store deliberately skips its cache when `all` is set (`ui/src/stores/plugins.ts:364` and `:383`). Including the flag in the cache key fixes both at once:

```ts
const cacheKey = (options.all ? "all:" : "") + (options.hash ? options.hash + id : id)
```

## 2. The scoping doesn't apply while typing, so the original problem is still there

On the branch:

1. New flow, default template (a `Log` task).
2. Cursor on an empty line inside the task body, `Ctrl+Space` → exactly 1 suggestion, `level`. Scoping is active here (and over-filtered, per #1).
3. Type one character, `r` → the list becomes `range`, `rangePartitioningEnd`, `rangePartitioningField`, `rangePartitioningInterval`, `rangePartitioningStart`, `rateLimitDelay`, `readTimeout`, `realm`, … — the unscoped cross-plugin union. This is the settled state, not a transient one; I polled it for 12s and it does not change.

Typing `t` in the same spot gives `table`, `tableDefinition`, `tableId`, `tag`, `tagging` rather than `type`, `timeout`, `taskCache`.

Likely mechanism: a bare partial key makes the document invalid YAML — the editor's validity badge flips to `1 Error(s)` at exactly that moment — so `findTaskLikeAtCursor` can't localize the enclosing task, `taskTypeAtCursor` returns `undefined`, and the filter fails open to the full list. Failing open is the right default, but the consequence is that the scoping only takes effect on an explicit trigger on an empty line, and not in the path where people actually use autocomplete. Worth confirming that reading before picking a fix; resolving the type against the document with the partial token stripped, or from the last known-good parse, would cover it.

Happy to re-run both checks once you've pushed.

flcarre:
I pushed a commit to this branch (ee179f8) fixing both blockers from my review, rather than leaving you to re-derive them. Shout if you'd rather I had left it as review notes and I'll happily drop the commit.

**Inherited properties.** The scoping call now passes `all: true`. Because the store skips its cache whenever `all` is set, that alone would have turned every completion into a request, so the cache key now carries the flag and the two response shapes are cached side by side (`ui/src/stores/plugins.ts`).

**Scoping while typing.** This one wasn't in my original review, I found it while QA'ing. Your resolver localizes the YAML node at the cursor, and a half-typed key localizes to nothing, so it fell through to the fail-open path and returned the unscoped union. Concretely: typing `t` inside a `Log` task offered `table`, `tableDefinition`, `tag`. The interesting part is that `r:` fails too, so it isn't only about invalid YAML: the empty-line case only worked because the whitespace walk-back reaches into the previous line, and any non-whitespace character pins the probe onto the partial token. `findTaskLikeAtCursor` now retries with the cursor's line blanked out (same length, so offsets still line up).

Verified in a running instance, cursor inside the default template's `Log` task:

| | before | after |
|---|---|---|
| empty line + `Ctrl+Space` | `level` only | `level`, `retry`, `runIf`, `description`, `disabled`, `logLevel`, `taskCache`, … |
| type `r` | `range`, `rangePartitioning*`, `rateLimitDelay`, `readTimeout` | `retry`, `runIf`, `policyRefs` |
| type `t` | `table`, `tableDefinition`, `tableId`, `tag` | `taskCache`, `timeout`, `logToFile` |

Tests: 5 added to `taskCompletionScoping.spec.ts` (the four half-typed-key cases plus a nested `Sequential` > `Log` case asserting the *inner* type wins) and a new `pluginsDocumentationCache.spec.ts` for the cache-key split. All five fail on 95c3965 and pass now. `check:types`, `eslint` on the changed files, and the 23 store/composable spec files (165 tests) are green.

Two things I deliberately did **not** touch, both yours to call:

- `isTaskLike` matches on `{id, type}` alone, so it also matches `inputs:`/`outputs:` entries, whose `type` is `STRING`/`INT`. That resolves to a type name that 404s (I confirmed `/api/v1/plugins/STRING` → 404). It fails open so nothing breaks, but it is a wasted request while editing inputs.
- The same path scopes `triggers:` bodies, since trigger entries also carry `{id, type}`. That now works correctly with `all: true` (`Schedule` goes from 9 keys to 20), but it's untested and arguably outside this PR's stated scope.

flcarre:
Pushed 0dad44f, which closes the three remaining threads from the earlier review. I'd missed them on my first pass, so my previous "both blockers fixed" comment was only two thirds of the story.

**Nested sub-maps were scoped against the task's schema.** This was the important one, and it was a regression this PR introduced rather than a pre-existing issue: with the cursor inside `retry:`, completion offered `description`, `disabled`, `policyRefs`, `retry`, `workerSelector` and hid retry's own `behavior`, `interval`, `maxAttempts`, `maxDuration`, `warningOnRetry`. The `all: true` change made it more visible, since the wrong key set went from 2 keys to 17. Resolution now rejects a task when a map sits between it and the cursor, so a sub-map falls open to the unscoped list instead of the wrong one. Verified: inside `retry:` the suggestions are the unscoped union again, while the task body stays scoped.

**Pinned `version` is forwarded** to the documentation lookup, so a version-pinned task is no longer scoped against the latest schema.

**Failed lookups are remembered.** `ignoreNotFound` only suppresses the toast, so the promise still rejected, nothing cached, and with `incomplete: true` the provider re-requested on every keystroke. Types that are not plugin FQCNs are now skipped outright too, which covers the `inputs:`/`outputs:` case where `type` is `STRING`.

That last part overlaps what I filed as https://github.com/kestra-io/kestra/issues/18749 before fixing it here, so the first half of that issue is now moot. I've left it open for the trigger-scoping half, which is still untested.

One process note, since it affects how much of this you should trust: I built the sub-map fix twice. The first attempt passed its unit tests but broke the task-body case in the browser (12 suggestions dropped to 5), because taking the innermost map from a fuzzy probe index can land on an outer node. The version here walks outward to the task as before and only rejects when an intervening map is present. Worth a look at that function specifically.

Checks on my side: `check:types` clean, eslint clean, resolver spec 13 tests, full unit suite green. Two specs (`logsCursor`, `posthog`) time out intermittently on my machine under load but pass in isolation and passed in CI on the previous push.

flcarre (approved):

PR #18640, opened 2026-08-26
perf(ui): give ECharts its own lazy chunk

elevatebart:
> **Stacked on https://github.com/kestra-io/kestra/pull/18638.** Merge #18638 first — this one only pays off on top of it (details below).
>
> **Reviewing note:** this PR targets `develop`, so its diff currently also contains #18638's commit. That is deliberate: the PR-docker-image workflow derives the base image tag from the base branch (`FROM kestra/kestra:$KESTRA_DOCKER_BASE_VERSION`), and a branch name containing `/` is not a valid Docker tag, so basing this PR on `claude/ee-dynamic-imports-warnings-pikloa` failed `Generate PR docker image / Publish Docker` with `invalid reference format`. Once #18638 merges, this diff collapses to just its own two commits. **The commits to review here are `perf(ui): give ECharts its own lazy chunk` and `test(design-system): import KsPie from its SFC in its stories`.**

### 🔗 Related Issue

None — follow-up to the `INEFFECTIVE_DYNAMIC_IMPORT` investigation in #18638.

### ✨ Description

The design-system barrel statically imported `KsEchart`, `KsGraph`, `KsLine`, `KsBar` and `KsPie`, so `main.ts` → `design-system/src/index.ts` pulled ECharts (493 modules) into the eager bundle on behalf of every consumer, including the many that never render a chart. `KsEditor` (Monaco) and `KsMarkdown` (Shiki) already avoid exactly this by exporting async components — the charts were simply never given the same treatment.

**Making them async is not sufficient on its own**, which is the interesting part. Two further things were needed:

1. The dashboard chart wrappers and `Dependencies.vue` import `echarts/core` **directly**, not through the design system. Without a group of its own, ECharts gets scattered across the page chunks the eager graph reaches and stays eager. So it needs a `LAZY_GROUPS` entry *and* a `GROUPS` chunk entry, claimed at priority `-18` — ahead of `design-system` (`-20`), which would otherwise take the `Charts/` directory and pull ECharts back in with it.
2. `Dependencies.vue` is the only ECharts consumer outside the planned pages, so its chunk was left to automatic chunking and statically imported the lazy `echarts` chunk. `consolidateChunks`'s own guard rejects that, with a genuinely excellent error message:

   > Chunk 'Dependencies' statically imports the lazy 'echarts' chunk, so every page reaching 'Dependencies' now downloads it. Its modules should be reported as reaching 'echarts' by collectLazyReach.

   Adding `dependencies` to `PAGE_GROUPS` gives its chunk the `-echarts` suffix and the invariant holds.

Measured on top of #18638, on the eager (preloaded) chunk closure:

| | chunks | raw | gzip |
|---|---|---|---|
| #18638 only | 16 | 3875 KB | 1013 KB |
| + this PR | 16 | 3203 KB | 787 KB |
| **saved** | 0 | **672 KB** | **226 KB** |

ECharts now lands in a single `echarts` chunk, and no eager chunk contains any ECharts module (verified against the build's own module→chunk map, not by grepping output).

Combined with #18638, the eager path goes from **1088 KB to 787 KB gzip**.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit`) — 197 files, 1851 tests
- [x] Storybook tests pass — `ui/` suite (66 files, 242 tests) **and** the design-system workspace suite the `Frontend - Design System tests` job runs, `vitest run --project 'storybook:*'` in `ui/packages/design-system` (80 files, 481 tests)
- [x] Translations are complete if `en.json` changed — no translation changes
- [ ] Screenshots or video recordings attached showing the `UI` changes — no visual change; chunk graph only

### 📝 Additional Notes

**One test change was needed, and it is worth understanding rather than skimming.** `KsPie.stories.ts` was the only chart story importing through the barrel — `KsBar`, `KsLine`, `KsGraph` and `KsEchart` all import the SFC directly. Once the barrel started exporting async components, its `play()` asserted on the DOM before the component had resolved and three stories failed. It now imports the SFC like its four siblings, so the story exercises the component rather than the barrel's export strategy. Nothing was weakened to get green — the assertions are unchanged.

**Worth a reviewer's eye:** the charts now resolve asynchronously, so a template ref into one is `null` for longer than before. Every access I found is optional-chained, and `useDependencies.captureAndFocusWhenReady` already polls for the chart instance — but it bails after ~2s (120 frames), and on a slow connection the `echarts` chunk fetch now sits inside that window. Worst case is the dependency graph rendering without its initial centring, not failing to render. I could not verify this against a running instance from here, so it's worth a click through the flow **Dependencies** tab and a dashboard before merging.

**On `moment`, the other half of what I was asked to chase:** it is eager through exactly one edge —
`App.vue` → `AppTopNavBar.vue` → `stores/playground.ts` → `utils/inputs.ts` → `moment`.
`inputs.ts` uses it for two calls in `normalize()` (`moment(res).toISOString()` and `moment().startOf("day").add(res, "seconds").toString()`). I deliberately did **not** change it here:

- `normalize()` is synchronous, so the import cannot simply be deferred.
- The native equivalents are not behaviourally identical — `moment().toString()` and `Date.prototype.toString()` differ (the latter appends the timezone name in parentheses), and that string feeds `TIME` input defaults.
- `moment` is imported by 35 other files, so removing this one edge moves it into a shared chunk that nearly every page fetches anyway. The win is small and the risk is real.

The genuinely valuable version of that work is migrating off `moment` (it is in maintenance mode), which is its own project rather than a line in this PR.


ui/src/routes/routes.ts
```diff
@@ -114,19 +102,19 @@ const routes: KestraRouteRecord[] = [
...
-    {name: "assets/list", path: "/:tenant?/assets", component: DemoAssets},
-    {name: "cases/list", path: "/:tenant?/cases", component: DemoCases},
-    {name: "admin/iam", path: "/:tenant?/admin/iam", component: DemoIAM},
-    {name: "admin/tenants/list", path: "/:tenant?/admin/tenants/list", component: DemoTenants},
-    {name: "admin/auditlogs/list", path: "/:tenant?/admin/auditlogs", component: DemoAuditLogs},
-    {name: "admin/quotas/list", path: "/:tenant?/admin/quotas", component: DemoQuotas},
-    {name: "admin/policies", path: "/:tenant?/admin/policies", component: DemoPolicies},
-    {name: "admin/instance", path: "/:tenant?/admin/instance", component: DemoInstance},
-    {name: "promote/targets", path: "/:tenant?/promote/targets", component: DemoPromote},
+    {name: "dashboards/create", path: "/:tenant?/dashboards/new", component: () => import("../components/demo/Dashboards.vue")},
```
Piyush-r-bhaskar, line 105: isn't this being done here https://github.com/kestra-io/kestra/pull/18638/changes#diff-8cb906bfa008106f846dd452e72b6a4f81e17fb5269577562ab98cada552a22a ?

Piyush-r-bhaskar (approved):

ui/src/routes/routes.ts
elevatebart: Good catch, and sorry for the confusion — it is the same commit, not duplicated work. This PR is stacked on #18638, so it contains `cb53870` (`perf(ui): keep the logs surface and the demo pages off the eager path`), and that is the hunk you are looking at here.

It targets `develop` rather than the #18638 branch because `Generate PR docker image / Publish Docker` derives the base image tag from the base branch (`FROM kestra/kestra:$KESTRA_DOCKER_BASE_VERSION`). A branch name containing `/` is not a valid Docker tag, so basing it on `claude/ee-dynamic-imports-warnings-pikloa` failed with `invalid reference format`. Targeting `develop` is the trade-off: CI works, at the cost of this extra commit showing in the diff. Once #18638 merges it disappears on its own.

The two commits that are actually this PR's are `perf(ui): give ECharts its own lazy chunk` and `test(design-system): import KsPie from its SFC in its stories`.

---

PR #18643, opened 2026-08-26
fix(core): keep an asset's creation date across updates

anna-geller:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra-ee/issues/10288.

### ✨ Description

An asset's `created` timestamp was overwritten on every update, so it was always identical to `updated` and the original registration time was lost. This happened both when an asset was edited through the API or the UI and when it was re-emitted by a flow.

`created` is defaulted to now by the `Asset` constructor, and the field is not included in a flow-emitted asset or in the YAML sent to `PUT /assets/{id}` (it is `@Hidden`). So a fresh timestamp was always present on the incoming asset, and it was kept by `toUpdated` over the stored one. The stored date is now preferred, and only `updated` is moved forward.

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched modules (`./gradlew :core:test --tests "io.kestra.core.models.assets.*"`)

### 📝 Additional Notes

`AssetTest.shouldKeepCreationDateWhenUpdating` is red on `develop` and green with the fix. Create and update are both routed through `Asset.toUpdated` by the JDBC and the Elasticsearch repositories, so every write path is covered.

For assets updated before this fix, the date they were last stamped with is kept, since the original was never stored anywhere else.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/assets/Asset.java
+++ b/core/src/main/java/io/kestra/core/models/assets/Asset.java
@@ -90,7 +90,7 @@ public abstract class Asset implements HasUID, SoftDeletable<Asset>, Plugin {
      * @return this asset, merged.
      */
     public <T extends Asset> T toUpdated(T previousAsset, boolean allowTypeChange) {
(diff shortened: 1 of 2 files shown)


brian-mulier-p (approved):
LGTM, ty

PR #18645, opened 2026-08-26
fix(core): reject a Flow trigger whose minSatisfied exceeds its dependsOn count

AcevedoR:

Closes https://github.com/kestra-io/kestra-ee/issues/10271

## What was wrong

A `Flow` trigger in `AT_LEAST` mode accepted a `minSatisfied` larger than the number of `dependsOn` conditions. The flow saved as **Valid** (HTTP 200), but the threshold could never be reached, so the trigger silently never fired.

```yaml
triggers:
  - id: f
    mode: AT_LEAST
    minSatisfied: 99      # only one dependency below
    dependsOn:
      - namespace: company.team
        flowId: upstream_flow
        states: [SUCCESS]
```

The lower bound was already guarded — `@Positive` on the field, and `FlowTriggerValidator` requires `minSatisfied` whenever the mode is `AT_LEAST`. Only the upper bound was missing.

## The fix

`FlowTriggerValidator` now also rejects `minSatisfied > dependsOn.size()`. The early return for non-`AT_LEAST` modes keeps the two `AT_LEAST` checks from nesting.

The new check is deliberately scoped to `dependsOn != null`: the field has no `@NotNull`, and whether an `AT_LEAST` trigger should be allowed to omit `dependsOn` entirely is a separate question I did not want to change here — `shouldBeValidWhenModeIsAtLeastWithMinSatisfied` covers exactly that shape and still passes untouched.

## Tests

Two tests added to `FlowTriggerValidationTest`:

- `shouldNotBeValidWhenMinSatisfiedExceedsTheNumberOfDependencies` — the issue's repro (`minSatisfied: 99`, one dependency). **Verified failing before the fix**, passing after.
- `shouldBeValidWhenMinSatisfiedEqualsTheNumberOfDependencies` — pins the `==` boundary as valid (it passed before the fix too, so the new check is not off-by-one).

All of `io.kestra.core.validations.*` plus `MultipleConditionTriggerCaseTest` pass: 125 tests, 0 failures.

### DIFF
--- a/core/src/main/java/io/kestra/core/validations/validator/FlowTriggerValidator.java
+++ b/core/src/main/java/io/kestra/core/validations/validator/FlowTriggerValidator.java
@@ -19,9 +19,22 @@ public class FlowTriggerValidator implements ConstraintValidator<FlowTriggerVali
             return true; // nulls are allowed according to spec
         }
 
-        if (MultipleCondition.Mode.AT_LEAST == value.getMode() && value.getMinSatisfied() == null) {
+        if (MultipleCondition.Mode.AT_LEAST != value.getMode()) {
+            return true;
+        }
+
+        if (value.getMinSatisfied() == null) {
             context.disableDefaultConstraintViolation();
             context.buildConstraintViolationWithTemplate("`minSatisfied` must be set when mode is AT_LEAST")
+                .addConstraintViolation();
+            return false;
+        }
+
+        if (value.getDependsOn() != null && value.getMinSatisfied() > value.getDependsOn().size()) {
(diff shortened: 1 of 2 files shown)


AcevedoR:
Superseded: the same fix landed on develop in 94e1019719 (`fix(flow): Flow trigger minSatisfied should not be above dependsOn`). That version is broader than this one — it uses `ListUtils.emptyOnNull(dependsOn)`, so it also rejects a `minSatisfied` set against a null `dependsOn`, which this PR deliberately skipped. Closing in favour of it.

PR #18648, opened 2026-08-26
fix(core): reject a negative duration on a MAX_DURATION SLA

AcevedoR:

Closes https://github.com/kestra-io/kestra-ee/issues/10262

## What was wrong

A flow SLA with a negative `duration` saved as **Valid**, then breached on the first evaluation of every execution — with `behavior: FAIL`, every run of the flow failed immediately.

```yaml
sla:
  - id: dur
    type: MAX_DURATION
    duration: PT-1H     # accepted at save
    behavior: FAIL
```

`MaxDurationSLA.duration` already carried `@DurationMax`; there was simply no lower bound.

## The fix

Added `@DurationMin` — a direct mirror of the existing `io.kestra.core.validations.DurationMax` (same annotation shape, same validator shape, same message style), defaulting to `PT0S` — and applied it to `MaxDurationSLA.duration` next to `@DurationMax`.

I used Kestra's own annotation rather than Hibernate's `org.hibernate.validator.constraints.time.DurationMin`. Hibernate's `DurationMax` does appear once in the codebase (`core/models/dashboards/TimeWindow`), but I verified by probe that Kestra's `@DurationMax` is definitely enforced by `ModelValidator`, and I did not want to depend on the Hibernate one being wired.

**One reviewable judgement call:** zero is treated as *valid* (inclusive boundary). A `PT0S` SLA is arguably just as degenerate as a negative one, but that is not what the issue reports, so I left it alone and pinned it with a test. Happy to flip it if you'd rather `PT0S` were rejected too.

## Tests

New `MaxDurationSLAValidationTest`:

- `shouldNotBeValidWhenDurationIsNegative` — the reproducer. **Verified failing before the fix** (`Expecting Optional to contain a value but it was empty`), passing after, and it asserts on the `duration` property path specifically.
- `shouldBeValidWithAPositiveDuration` / `shouldBeValidWithAZeroDuration` — pin both boundaries.
- `shouldStillRejectADurationAboveTheMaximum` — proves the pre-existing `@DurationMax` still fires.

> Note on method: my first version of the negative test asserted only `assertThat(valid).isPresent()` and **passed before the fix** — the SLA under test had null `type`/`behavior`, so unrelated violations made it green. The committed tests build a fully valid SLA and assert on the `duration` violation specifically.

`io.kestra.core.validations.*` and `io.kestra.core.models.flows.sla.*` pass: 135 tests, 0 failures.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/sla/types/MaxDurationSLA.java
+++ b/core/src/main/java/io/kestra/core/models/flows/sla/types/MaxDurationSLA.java
@@ -12,6 +12,7 @@ import io.kestra.core.models.flows.sla.SLA;
 import io.kestra.core.models.flows.sla.Violation;
 import io.kestra.core.runners.RunContext;
 import io.kestra.core.validations.DurationMax;
+import io.kestra.core.validations.DurationMin;
 
 import jakarta.validation.constraints.NotNull;
 import lombok.Getter;
@@ -23,6 +24,7 @@ import lombok.experimental.SuperBuilder;
 @NoArgsConstructor
 public class MaxDurationSLA extends SLA implements ExecutionMonitoringSLA {
     @NotNull
+    @DurationMin
     @DurationMax
     private Duration duration;
 
--- a/core/src/main/java/io/kestra/core/validations/DurationMin.java
+++ b/core/src/main/java/io/kestra/core/validations/DurationMin.java
@@ -1,0 +1,25 @@
+package io.kestra.core.validations;
+
+import java.lang.annotation.Retention;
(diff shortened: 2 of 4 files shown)


AcevedoR:
Superseded: develop now carries `@DurationMin(millis = 1, message = "must be a positive duration")` on `MaxDurationSLA.duration`, closing kestra-ee#10262. I verified against develop that it is actually enforced by `ModelValidator` — negative and zero both rejected, positive valid — so the Hibernate annotation works here and the custom `@DurationMin` this PR added is not needed. Note it also rejects `PT0S`, which this PR had deliberately left valid.

PR #18649, opened 2026-08-26
fix(webserver): validate label keys on the execution-trigger and bulk set-labels paths

### LINKED ISSUE #18332: [Bash2.0] Invalid label keys are accepted when triggering executions and bulk-setting labels
## Describe the issue

Kestra enforces a strict format for label keys — a key must start with a lowercase letter and contain only lowercase letters, digits, `.`, `_`, or `-`. This validation is applied inconsistently across the label-writing paths. It is enforced when you set a label on a single execution and when you define labels on a flow, but **not** when you trigger an execution with labels (`POST /executions/{namespace}/{flow}?labels=key:value`) nor when you use the bulk set-labels action (`POST /executions/labels/by-ids`). Those two paths accept keys that violate the format (e.g. `Team`, `1team`) and store them on executions.

The result is an inconsistency: you can create or bulk-label an execution with an invalid key, but then cannot edit that same label from the execution's own label editor (it is rejected). It also lets malformed label keys leak into the system, where other tooling expects the documented format.

## Steps to reproduce

The invalid key `Team` (starts with an uppercase letter) is accepted by some label paths and rejected by others. In the UI:

1. **Accepted** — Open any flow, click **Execute**, switch to the **Other properties** tab, click **Add label**, enter key `Team` value `x`, and click **Execute**. The execution runs and carries the label `Team: x`.
2. **Accepted** — On the executions list, select one or more executions, use the bulk **Set labels** action, add `Team` = `x`, and confirm. The label is set on every selected execution.
3. **Rejected** — Open a single execution's overview and add the label `Team` = `x` there. It fails with *"Invalid label key … must begin with a lowercase letter."*

So the same label can be added from the Execute form and the bulk action, but not from the single execution's own label editor.

Equivalent API behaviour:

```bash
# Trigger with the invalid key           -> 200  (accepted, stored)
curl -u user:pass -X POST 'http://localhost:8080/api/v1/main/executions/demo/myflow?labels=Team:x'
# Bulk set-labels by ids                  -> 202  (accepted, stored)
curl -u user:pass -H 'Content-Type: application/json' -X POST \
    'http://localhost:8080/api/v1/main/executions/labels/by-ids' \
    -d '{"executionsId":["<id>"],"executionLabels":[{"key":"Team","value":"x"}]}'
# Single set-labels on one execution      -> 422  Invalid label key ...
curl -u user:pass -H 'Content-Type: application/json' -X POST \
    'http://localhost:8080/api/v1/main/executions/<id>/actions/labels' -d '[{"key":"Team","value":"x"}]'
```

## Actual behaviour

The execution-trigger path and the bulk set-labels-by-ids path both accept invalid keys (`Team`, `1team`, …) and store them, while the single set-labels endpoint and flow definitions reject the identical label with *"Invalid label key … must begin with a lowercase letter."* A malformed label is fully stored and filterable, but cannot be edited from the execution's own label editor. Reproduces on both OSS and Enterprise.

## Expected behaviour

All label-writing paths should validate keys the same way. The execution-trigger `labels` parameter and the bulk set-labels-by-ids endpoint should reject keys that violate the label-key format with the same `422` the single set-labels endpoint returns, so a key like `Team` is either accepted everywhere or rejected everywhere.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise (shared core execution API)
- Reproduced via: `POST /api/v1/main/executions/{namespace}/{flow}?labels=...` vs `POST .../executions/{id}/actions/labels` on `http://localhost:8080`

AcevedoR:

Closes https://github.com/kestra-io/kestra/issues/18332
Closes https://github.com/kestra-io/kestra-ee/issues/10310

## What was wrong

The label-key format (`^[\p{Ll}][\p{L}0-9._-]*$`, declared on `Label`) was enforced on flow definitions and on `POST /executions/{id}/actions/labels`, but **not** on two other write paths:

| Path | Before |
|---|---|
| `POST /executions/{ns}/{flow}?labels=Team:x` | 200, stored |
| `POST /executions/labels/by-ids` with key `Team` | 202, stored |
| `POST /executions/{id}/actions/labels` with key `Team` | 422 ✅ |

So a label could be created but not subsequently edited from the execution's own editor, and malformed keys leaked into storage.

## The fix

**1. `parseLabels`** (the `?labels=` query path) built `Label` records by hand and never validated them. It now runs each caller-supplied label through the injected `ModelValidator`, which throws the same `ConstraintViolationException` that the single set-labels endpoint already maps to 422.

Validation happens on the caller-supplied labels only — before the method appends `system.from`, and before the EE subclass appends `system.username` — so neither internal label is subject to it (both are well-formed regardless).

**2. `setLabelsOnTerminatedExecutionsByIds`** took a plain `@Body SetLabelsByIdsRequest`. The body is now `@Valid` and the record declares `List<@Valid Label> executionLabels`, so element-level constraints actually run.

## Also fixes kestra-ee#10310

Making the body `@Valid` makes the record's existing `@NotNull executionsId` actually fire. Omitting `executionsId` previously produced:

```
500 Internal server error: Cannot invoke "java.util.List.iterator()" because the return value of
"...SetLabelsByIdsRequest.executionsId()" is null
```

and now answers **422 naming the field**. That issue asked for a 400, but 422-naming-the-field is what the sibling `labels/by-query` endpoint already returns for the same mistake (the issue notes this itself), so I matched the sibling rather than introducing a third shape.

## EE impact

`parseLabels` is `protected` and EE's `ExecutionController` overrides it. The override calls `super.parseLabels(...)` and then appends `system.username`, which happens after validation and is a valid key — so EE behaviour is unchanged apart from the intended rejection.

## Tests

Three tests added to `ExecutionControllerRunnerTest`, each **verified failing before the fix**:

- `shouldRejectAnInvalidLabelKeyWhenTriggeringAnExecution` — `?labels=Team:x` → 422 (`nothing was thrown` before).
- `shouldRejectAnInvalidLabelKeyWhenSettingLabelsByIds` — by-ids with key `Team` → 422 (`nothing was thrown` before).
- `shouldRejectAMissingExecutionsIdWhenSettingLabelsByIds` — by-ids with no ids → 422 (`expected: 422 but was: 500` before).

Full `ExecutionControllerRunnerTest` passes: **113 tests, 0 failures**. That class already covers the accepted-label paths (`system.from:ui`, `system.correlationId`, multi-label triggers, by-query set-labels), which is what makes the regression risk here measurable rather than assumed.

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
@@ -55,6 +55,7 @@ import io.kestra.core.models.topologies.FlowTopology;
 import io.kestra.core.models.topologies.FlowTopologyGraph;
 import io.kestra.core.models.triggers.AbstractTrigger;
 import io.kestra.core.models.validations.ManualConstraintViolation;
+import io.kestra.core.models.validations.ModelValidator;
 import io.kestra.core.plugins.PluginRegistry;
 import io.kestra.core.preview.FilePreview;
 import io.kestra.core.preview.FileRenderer;
@@ -250,6 +251,9 @@ public class ExecutionController {
 
     @Inject
     private WebhookBodyService webhookBodyService;
+
+    @Inject
+    private ModelValidator modelValidator;
 
     @Inject
     private AsyncOperationWaiter asyncOperationWaiter;
@@ -1154,6 +1158,10 @@ public class ExecutionController {
             : RequestUtils.toMap(labels).entrySet().stream()
                 .map(entry -> new Label(entry.getKey(), entry.getValue()))
                 .collect(Collectors.toList());
+
+        // the label key format is enforced on the flow definition and on the single set-labels
+        // endpoint, so enforce it here too rather than letting a malformed key in through this door
+        parsedLabels.forEach(modelValidator::validate);
 
         // system labels can only be set by Kestra itself; the only exceptions accepted from a
         // client are system.correlationId and system.from=ui (sent by the UI). Any other
@@ -2417,7 +2425,7 @@ public class ExecutionController {
     @ApiResponse(responseCode = "202", description = "Accepted", content = { @Content(schema = @Schema(implementation = ApiAsyncOperationResponse.class)) })
     @ApiResponse(responseCode = "400", description = "Validation errors", content = { @Content(schema = @Schema(implementation = BulkErrorResponse.class)) })
     public MutableHttpResponse<ApiAsyncOperationResponse> setLabelsOnTerminatedExecutionsByIds(
-        @RequestBody(description = "The request containing a list of labels and a list of executions") @Body SetLabelsByIdsRequest setLabelsByIds) throws QueueException {
+        @RequestBody(description = "The request containing a list of labels and a list of executions") @Body @Valid SetLabelsByIdsRequest setLabelsByIds) throws QueueException {
         List<Execution> executions = new ArrayList<>();
         Set<ManualConstraintViolation<String>> invalids = new HashSet<>();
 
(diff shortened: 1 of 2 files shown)


AcevedoR:
Superseded: the same fix landed on develop in 2311df1dae (`fix(executions): validate label keys on every execution label API`) — `parseLabels` now runs each label through `ModelValidator` and the by-ids body is `@Valid`, which is the approach this PR took. Closing as redundant.

PR #18652, opened 2026-08-26
fix(core): stop LoopUntil after maxIterations instead of maxIterations + 1

AcevedoR:
Closes https://github.com/kestra-io/kestra-ee/issues/10323


loicmathieu (approved):
LGTM

PR #18672, opened 2026-08-26
fix: remove all compilation warning

tchiotludo:
relate to kestra-io/kestra-ee#10453


jdbc-h2/src/test/java/io/kestra/queue/h2/H2BroadcastQueueTest.java
```diff
@@ -17,8 +17,8 @@ class H2BroadcastQueueTest extends AbstractBroadcastQueueTest {
     JdbcTestUtils jdbcTestUtils;
 
     @BeforeEach
+    @SuppressWarnings("deprecation")
     protected void init() {
         jdbcTestUtils.drop();
-        jdbcTestUtils.migrate();
```
AcevedoR, line 22: is it on purpose ? whats the catch here ?

tchiotludo: just an hard generated PR for now, that I don't look a lot. 
Overall it's not bad, but we should probably remove some that are unsafe before merging (like this one)

tchiotludo: Good catch to double-check, but this one's safe — `migrate()` was *already* a no-op in the codebase before any of today's changes, not something introduced by my edits.

Confirmed. Here's the evidence:

1. **`JdbcTestUtils.java` itself was never touched** by any of my edits — `git diff --stat` shows zero changes to that file. Only the ~40 test files that *called* `.migrate()` were edited.

2. **`migrate()` was already an empty no-op method** in the committed codebase, since commit `b869c0ff1c` ("automate all migration scripts"). `git show HEAD:.../JdbcTestUtils.java` confirms the body is literally `public void migrate() {}` — it does nothing and never did, before any of this session's work.

3. **Migrations now run automatically** via `MigrationStartupRunner`, a `@Context` bean that calls `MigrationRunner.autoRun()` unconditionally at Micronaut context startup, before any repository/service bean can query the database (see the class javadoc in `core/src/main/java/io/kestra/core/migration/MigrationRunner.java:23-25`). So every test's DB schema is already migrated by the time the test context is up, regardless of whether `jdbcTestUtils.migrate()` is called.

So removing those call sites doesn't change test behavior at all — it deletes a line that was silently doing nothing, matching what the deprecation javadoc says explicitly. The `H2RunnerTest` run, plus the full `jdbc`/`jdbc-h2` test suites I ran afterward, all passed, which is consistent with this being a safe no-op removal rather than a behavioral change.

tchiotludo:
Companion EE PR: https://github.com/kestra-io/kestra-ee/pull/11020 (same branch name, `fix/compilation-warnings`). They need to land together — the EE branch drops the `JdbcTestUtils.migrate()` calls that this PR removes.

One warning is left on this side: `:webserver:compileJava` reports `The following options were not recognized by any processor: '[micronaut.openapi.expand.version]'`, from `webserver/build.gradle`.

core/src/main/java/io/kestra/core/cache/NoopCache.java
```diff
@@ -78,7 +76,7 @@ public CacheStats stats() {
 
     @Override
     public ConcurrentMap<K, @NonNull V> asMap() {
-        return (ConcurrentMap<K, V>) EMPTY_MAP;
```
loicmathieu, line 81: This will initiate a map on each call which is not a good idea.
Please revert

core/src/main/java/io/kestra/core/queues/GenericQueueInterface.java
```diff
@@ -18,4 +18,7 @@ public interface GenericQueueInterface<T extends Event> extends AutoCloseable {
      * Get the name of the queue
      */
     String queueName();
+
+    @Override
+    void close();
```
loicmathieu, line 23: I don't understand

executor/src/main/java/io/kestra/executor/DefaultExecutor.java
```diff
@@ -60,6 +60,7 @@
 
 import static io.kestra.core.utils.Rethrow.*;
 
+@SuppressWarnings("this-escape")
```
loicmathieu, line 63: I'm not comptforable with it as it hides a probable bug

loicmathieu (changes requested):
I'm not comptfortable of suppressing varagrs, this-escape and deprecation as it will hides stuff that we need to take action on.
It will suppressin both compilation logs and IDE which is not a good idea.

If you really want to hide them in log, you can disable warning in compilation log instead.

tchiotludo:
I want you to evaluate if the suppress is normal and it's an intended behavior. 
We have a lot of cast, varagrs, etc ... that is mandatory by design so suppress is the only option.
I want developers to have strong opinion if it's a normal behavior (and suppress) or a potential bug and fix it. 

So definitively the solution is not to reduce the log verbosity and hide it, developers must assume.

Then we force on the CI as check to avoid going on this battle all the times. 

Feel free to remove some suppress and remove the "bug" or keep the suppress.

You are free to take over this PR and to finished it yourself and merged it.

PR #18673, opened 2026-08-26
fix(core): parse a flow whose namespace is digits only

### LINKED ISSUE #18472: [Bash2.0] Creating a flow with a digits-only namespace shows a 500 with a raw internal error
## Describe the issue

Create a new flow and set the namespace to digits only. When the editor opens, a 500 error pops up with a raw internal message and the flow shows 1 error.

## Steps to reproduce

1. Go to Flows and click Create.
2. Set Flow id to anything, for example `3132`.
3. Set Namespace to a number only, for example `132312312`.
4. Click Open editor.

## Actual behaviour

A "500 Internal server error" toast appears with the message:

`class java.lang.Integer cannot be cast to class java.lang.String (java.lang.Integer and java.lang.String are in module java.base of loader 'bootstrap')`

The editor also shows 1 error.

## Visuals

Image

## Extra context

A digits-only namespace is a supported value elsewhere. You can create and save a namespace like `4232312312` on the Namespaces page and it saves fine. The create-flow form also accepts a digits-only namespace as valid, with no warning. The flow editor is the only place that breaks on it.

## Environment

- Kestra Version: 2.0.0-rc10
- Edition: Enterprise

AcevedoR:

Closes https://github.com/kestra-io/kestra/issues/18472

## What was wrong

`FlowParsingService.readFlowAsMap` read the namespace back out of the parsed YAML map with a direct cast:

```java
return injectPluginVersions(tenantId, namespace == null ? (String) mapFlow.get("namespace") : namespace, mapFlow);
```

YAML types an unquoted digits-only scalar as an `Integer`, so `namespace: 132312312` blew up with

```
class java.lang.Integer cannot be cast to class java.lang.String
```

which reached the user as a 500 toast when opening the flow editor.

## Why the value is legitimate

`AbstractFlow.namespace` is constrained by `^[a-z0-9][a-z0-9._-]*` — digits-only matches. The issue also notes a digits-only namespace saves fine from the Namespaces page. So this is purely a parsing defect, not a namespace that should have been rejected; fixing the cast lets the value reach normal validation instead of dying before it.

## The fix

`Objects.toString(mapFlow.get("namespace"), null)` — the same pattern `FlowController` already uses when pulling the namespace out of a raw map:

https://github.com/kestra-io/kestra/blob/develop/webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java#L384

Null-safety is preserved: `Objects.toString(null, null)` returns `null`, matching what the cast produced for an absent namespace.

## Tests

`shouldParseAFlowWhoseNamespaceIsDigitsOnly` added to `FlowParsingServiceTest`, parsing a real YAML source through the public `parse(tenantId, source, strictParsing)` entry point. **Verified failing before the fix** with exactly the reported exception:

```
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String
	at io.kestra.core.services.FlowParsingService.readFlowAsMap(FlowParsingService.java:222)
```

- `io.kestra.core.services.*` → 145 tests, 0 failures
- `*FlowController*` (webserver) → 75 tests, 0 failures

### DIFF
--- a/core/src/main/java/io/kestra/core/services/FlowParsingService.java
+++ b/core/src/main/java/io/kestra/core/services/FlowParsingService.java
@@ -1,6 +1,7 @@
 package io.kestra.core.services;
 
 import java.util.Map;
+import java.util.Objects;
 
 import org.slf4j.event.Level;
 
@@ -219,7 +220,9 @@ public class FlowParsingService {
         @Nullable final String namespace,
         final String source) throws JsonProcessingException, FlowProcessingException {
         Map<String, Object> mapFlow = YAML_MAPPER.readValue(source, JacksonMapper.MAP_TYPE_REFERENCE);
-        return injectPluginVersions(tenantId, namespace == null ? (String) mapFlow.get("namespace") : namespace, mapFlow);
+        // the namespace is read back from the parsed YAML, where a digits-only value such as
+        // `namespace: 132312312` arrives as an Integer even though it matches the namespace pattern
+        return injectPluginVersions(tenantId, namespace == null ? Objects.toString(mapFlow.get("namespace"), null) : namespace, mapFlow);
     }
 
     /**
--- a/core/src/test/java/io/kestra/core/services/FlowParsingServiceTest.java
+++ b/core/src/test/java/io/kestra/core/services/FlowParsingServiceTest.java
@@ -24,6 +24,25 @@ class FlowParsingServiceTest {
     private FlowParsingService flowParsingService;
 
     @Test
+    void shouldParseAFlowWhoseNamespaceIsDigitsOnly() throws FlowProcessingException {
+        // Given - YAML parses an unquoted digits-only namespace as an Integer
+        String source = """
+            id: digits
+            namespace: 132312312
+            tasks:
+              - id: log
+                type: io.kestra.plugin.core.log.Log
+                message: hello
+            """;
+
+        // When
+        FlowWithSource parsed = flowParsingService.parse(null, source, false);
+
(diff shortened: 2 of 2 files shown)


core/src/main/java/io/kestra/core/services/FlowParsingService.java
```diff
@@ -219,7 +220,9 @@ protected Map<String, Object> readFlowAsMap(@Nullable final String tenantId,
         @Nullable final String namespace,
         final String source) throws JsonProcessingException, FlowProcessingException {
         Map<String, Object> mapFlow = YAML_MAPPER.readValue(source, JacksonMapper.MAP_TYPE_REFERENCE);
-        return injectPluginVersions(tenantId, namespace == null ? (String) mapFlow.get("namespace") : namespace, mapFlow);
+        // the namespace is read back from the parsed YAML, where a digits-only value such as
+        // `namespace: 132312312` arrives as an Integer even though it matches the namespace pattern
+        return injectPluginVersions(tenantId, namespace == null ? Objects.toString(mapFlow.get("namespace"), null) : namespace, mapFlow);
```
loicmathieu, line 225:
```suggestion
        return injectPluginVersions(tenantId, namespace == null ? Objects.toString(mapFlow.get("namespace")) : namespace, mapFlow);
```

Can't be null right?

loicmathieu (approved):
LGTM

AcevedoR:
Superseded by #18750, which landed this exact change in `FlowParsingService` (including the two-argument `Objects.toString(..., null)`, so an absent namespace still resolves to null rather than the string "null"). Closing as redundant.

PR #18674, opened 2026-08-26
fix(core): explain that Loop values must be a list instead of leaking a Jackson error

### LINKED ISSUE #18454: [Bash2.0] Loop with a string values leaks a raw Jackson exception into the user logs
## Describe the issue

If a `Loop` task's `values` is a plain string instead of a list, the execution fails with a raw Java exception in the user-facing logs: a Jackson `JsonParseException` about an "Unrecognized token", plus internal detail such as `StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION`. Nothing tells the user the real problem, which is simply that `values` needs to be a list. This is an easy mistake to make (a `values` built from an expression that renders to a single value hits it too), and the error gives no clue how to fix it.

## Steps to reproduce

1. Create a flow whose `Loop` gets a plain string as `values`:

```yaml
id: loopstr
namespace: company.team
tasks:
  - id: lp
    values: "hello world"
    tasks:
      - id: l
        message: "v={{ item.value }}"
```

2. Execute it and open the execution's Logs tab.

## Actual behaviour

The execution fails and the log line shown to the user is a raw Java stack message:

> Failed to process flowable task 1wLSgQ3iBD48udmrQbk97W: com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'hello': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false') at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1 ...]

The flow does get rejected, which is right, but the message names Jackson internals and an internal task id instead of the actual mistake.

## Expected behaviour

- The failure should say what is wrong in the user's terms, e.g. that `values` on task `lp` must be a list (or a map, or an expression that renders to one) and that a plain string was given. The Jackson exception belongs in the debug log, not in the user-facing error.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise

AcevedoR:

Closes https://github.com/kestra-io/kestra/issues/18454

## What was wrong

A `Loop` whose `values` is a plain string failed with Jackson internals in the **user-facing** execution log:

> Failed to process flowable task 1wLSgQ3iBD48udmrQbk97W: com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'hello': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false') at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1 …]

## Why the message looked like that

`FlowableUtils.resolveValues` wrapped the parse failure as:

```java
} catch (IOException e) {
    throw new IllegalVariableEvaluationException(e);
}
```

`Exception(Throwable)` sets the message to `cause.toString()`, and `ExecutorService` logs `e.getMessage()` — so the Jackson text *was* the message.

## The fix

Pass an explicit message and keep the parse error as the cause:

```java
throw new IllegalVariableEvaluationException(notAListOrMapMessage(renderValue), e);
```

The cause is still handed to `logger().error(..., e)`, so the Jackson detail remains in the stack trace — which is what the issue asked for.

Two smaller decisions:

- The neighbouring non-array/non-object branch threw `"Unknown value type: NUMBER"`, which is the same mistake described in internal terms. It now shares the same wording, so `values: "hello world"` and `values: "123"` explain themselves identically.
- The rendered value is included (it is the useful part when the value came from an expression) but **truncated to 100 chars**, so a `values` expression rendering to a large payload cannot flood the execution logs.

## Tests

`resolveValues_withPlainString_shouldFailWithAUserFacingMessage` added to `FlowableUtilsTest`, asserting both the new wording *and* the absence of `JsonParseException` / `Unrecognized token` / `StreamReadFeature`. **Verified failing before the fix.**

- `FlowableUtilsTest` + all `*Loop*` tests → 46 tests, 0 failures
- Full `H2RunnerTest` → 100 tests, 0 failures (run because `resolveValues` is on the live Loop/ForEach execution path, not just a parsing helper)

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowableUtils.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowableUtils.java
@@ -415,6 +415,20 @@ public class FlowableUtils {
      * @return a list of String with no duplicates if the values were a list, or a list of pairs of String/String if the values were a map.
      * @throws IllegalVariableEvaluationException in case of JSON error, unsupported value type or duplicate values.
      */
+    private static final int MAX_REPORTED_VALUE_LENGTH = 100;
+
+    /**
+     * Builds the user-facing message for a `values` that did not render to a list or a map. The rendered value is
+     * truncated so a large rendered payload cannot flood the execution logs.
+     */
+    private static String notAListOrMapMessage(String renderValue) {
+        String reported = renderValue == null ? "null"
+            : renderValue.length() > MAX_REPORTED_VALUE_LENGTH
+                ? renderValue.substring(0, MAX_REPORTED_VALUE_LENGTH) + "…"
+                : renderValue;
+        return "The `values` property must be a list, a map, or an expression that renders to one, but it rendered to: " + reported;
+    }
+
     public static Either<List<String>, List<Pair<String, String>>> resolveValues(RunContext runContext, Object values) throws IllegalVariableEvaluationException {
         switch (values) {
             case String stringValue -> {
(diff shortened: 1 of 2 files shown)


AcevedoR:
Superseded: develop now catches `JsonProcessingException` in `FlowableUtils.resolveValues` and reports "The `values` must be a list, a map, or an expression that renders to one", which is what this PR added. Closing as redundant.

PR #18675, opened 2026-08-26
fix(core): return 400 for a malformed query filter

### LINKED ISSUE #18646: Query-filter errors return inconsistent HTTP status codes (400 vs 422)
### Describe the issue

Different kinds of invalid `filters[...]` input on search endpoints (e.g. `flows/search`, `executions/search`) currently return different HTTP status codes for what a caller would consider the same class of error — "the query filter I sent is invalid":

- An unsupported operation for a given field, or another semantic validation failure caught by `QueryFilter.validateQueryFilters` / `AbstractJdbcRepository`, throws `InvalidQueryFiltersException`, which `ErrorController` maps to **400 Bad Request** (`webserver/src/main/java/io/kestra/webserver/controllers/ErrorController.java`, the `InvalidQueryFiltersException` handler).
- A bogus operator name (e.g. `filters[labels][BOGUS_OP]=x`) fails Jackson enum conversion into `QueryFilter.Op`, which surfaces as `IllegalArgumentException` and is mapped to **422 Unprocessable Entity** by the same controller.

Both are "the request is malformed", but callers get two different codes depending on which validation layer happens to catch the problem, which makes client-side error handling for search filters unreliable.

### Context

Surfaced while investigating https://github.com/kestra-io/kestra-ee/issues/10150 (labels filter returning 500 instead of a 4xx) — the fix for that issue standardizes on 400 via `InvalidQueryFiltersException`, matching the existing convention for `InvalidQueryFiltersException`, but does not attempt to unify it with the 422 case, since that would change the status code of every existing invalid-query-filter response across the API (a breaking change for any client parsing status codes) and deserves its own deliberate decision rather than being folded into a bug fix.

### Suggested direction (not decided)

Pick a single status code (400 seems the better fit — the request is syntactically/semantically wrong, not "understood but unprocessable") and either:
- change `IllegalArgumentException` handling for the enum-conversion case specifically, or
- make `QueryFilter`'s Jackson binding fail in a way that raises `InvalidQueryFiltersException` instead of a raw `IllegalArgumentException`.

Needs a decision on backward compatibility (some clients may already depend on 422 for this case).

loicmathieu:
### Related Issue

Closes https://github.com/kestra-io/kestra/issues/18646.

### Description

**Problem:** `filters[...]` query params on search endpoints (`executions/search`, `flows/search`, etc.) returned inconsistent status codes for what a caller sees as the same class of error — "the filter I sent is invalid". An unsupported field or operation for a resource already returned 400 via `InvalidQueryFiltersException` (`QueryFilter.validateQueryFilters` and the JDBC repositories). But `QueryFilterFormatBinder`, the hand-written Micronaut binder that parses `filters[field][op]=value` before any of that validation runs, let a raw `IllegalArgumentException` escape for an unknown field name, an unknown operator name, or a malformed bracket shape — and `ErrorController`'s catch-all handler mapped that to 422. A client couldn't reliably branch on status code to detect a bad filter. The 422 has nothing to do with Jackson (the issue's original hypothesis) — `filters[...]` never goes through Jackson; it's this custom binder calling `Op.valueOf`/`Field.fromString` directly.

A related defect fell out of the same code: a `filters[...]` key whose bracket shape didn't parse was silently dropped, so the caller got a successful *unfiltered* response instead of an error.

**Fix, scoped to the filter's structure, not its values:**
- A filter expression the server **cannot parse** — unknown field/operator name, malformed bracket shape, nesting/width past the configured cap, a malformed key — now throws `InvalidQueryFiltersException` → **400**, matching the ~40 existing structural-validation throw sites elsewhere in the codebase.
- A filter that's understood but whose **value** can't be processed (bad `scope` value, unparseable `timeRange`, inverted `startDate`/`endDate`) is unchanged and still returns **422** — the request was understood, just not processable. This is deliberately narrower than "make every filter error 400": a blanket catch around the binder's parse step would have also caught `RequestUtils.toFlowScopes`, a value error, and pulled it into 400 incorrectly.
- `QueryFilter.Op` gained a `fromString` factory (mirroring `Field` and `Logical`, per the repo's enum convention), so the binder no longer leaks an internal class name in the response body (previously: `"Illegal argument: No enum constant io.kestra.core.models.QueryFilter$Op.BOGUS_OP"`).
- A malformed `filters[...]` key is now rejected (400) instead of silently ignored.

### DIFF
--- a/core/src/main/java/io/kestra/core/exceptions/InvalidQueryFiltersException.java
+++ b/core/src/main/java/io/kestra/core/exceptions/InvalidQueryFiltersException.java
@@ -28,4 +28,14 @@ public class InvalidQueryFiltersException extends KestraRuntimeException {
     public InvalidQueryFiltersException(final String invalid) {
         super(INVALID_QUERY_FILTER_MESSAGE.formatted(invalid));
     }
+
+    /**
+     * Creates a new {@link InvalidQueryFiltersException} instance.
+     *
+     * @param invalid the invalid filter.
+     * @param cause the underlying cause.
+     */
+    public InvalidQueryFiltersException(final String invalid, final Throwable cause) {
+        super(INVALID_QUERY_FILTER_MESSAGE.formatted(invalid), cause);
+    }
 }
--- a/core/src/main/java/io/kestra/core/models/QueryFilter.java
+++ b/core/src/main/java/io/kestra/core/models/QueryFilter.java
@@ -110,7 +110,12 @@ public record QueryFilter(
         IS_NULL,
         IS_NOT_NULL,
         REGEX,
-        PREFIX
+        PREFIX;
+
+        @JsonCreator
+        public static Op fromString(String value) {
+            return Enums.getForNameIgnoreCase(value, Op.class);
+        }
     }
 
     @SuppressWarnings("unchecked")
--- a/webserver/src/main/java/io/kestra/webserver/converters/QueryFilterFormatBinder.java
+++ b/webserver/src/main/java/io/kestra/webserver/converters/QueryFilterFormatBinder.java
@@ -8,6 +8,7 @@ import java.util.regex.Pattern;
 
 import com.google.common.annotations.VisibleForTesting;
 
+import io.kestra.core.exceptions.InvalidQueryFiltersException;
 import io.kestra.core.models.QueryFilter;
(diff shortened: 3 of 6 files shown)


nkwiatkowski (approved):

loicmathieu:
Ci never run, this PR is doomed

PR #18676, opened 2026-08-26
fix(core): reject non-positive durations and counts on user-settable fields

nkwiatkowski:
Closes https://github.com/kestra-io/kestra-ee/issues/10261.
Closes https://github.com/kestra-io/kestra-ee/issues/10262.
Closes https://github.com/kestra-io/kestra-ee/issues/10274.

## Problem

Several user-settable duration and count fields accepted zero or negative values when a flow was saved. The flow showed as *Valid*, but at execution the value was nonsensical: a `PT0S` task `timeout` or a `PT-1H` SLA `duration` failed the run immediately, a negative `Parallel`/`Pause` value ran as a silent no-op, etc. This mirrors the duration-overflow class fixed earlier (kestra#18549), on the other bound.

## Fix

Add positivity validation, reusing Hibernate `@DurationMin` (already used on `Quota.duration`) and `@Positive`:

- **Plain `Duration` fields — validated at save (422):** `MaxDurationSLA.duration`, `Schedule.lateMaximumDelay`, retry `interval`/`maxInterval`/`minInterval`/`maxDuration`, `TimeWindow.window` (strictly positive) and `windowAdvance` (non-negative), `Cache.ttl`.
- **`Property<...>` fields — annotation on the generic type argument** (`Property<@DurationMin Duration>`), so the `PropertyValueExtractor` validates the rendered value at runtime and any literal at save: `Task.timeout`, `Sleep.duration`, `Pause.pauseDuration`, `kv.Set.ttl`, `WorkingDirectory.ttl`, `Split.partitions`/`rows`.

`Pause.pauseDuration` additionally keeps a runtime guard in `ExecutorService.handlePausedDelay`: that render site is not task-bound, so an expression rendering to a non-positive value would otherwise bypass validation — it now fails the execution with a clear message instead of resuming immediately.

## Evidence

`PositiveDurationValidationTest` covers save-time (SLA, Schedule) and runtime `Property` (Sleep, accept + reject) cases. `H2RunnerTest` and the `validations`/retry suites pass; no existing flow fixture trips the new bounds.

## Notes

- `Parallel.concurrent`, `Dag.concurrent` and `LoopUntil.*` are intentionally **not** included: their runtime render happens inside a `catch` in `resolveNexts` that swallows exceptions, so a validation failure cannot cleanly fail the run yet. They are tracked for the follow-up that narrows that catch.
- Adding `@DurationMin` to the widely-referenced `Task.timeout` produces `classfile` warnings (`Cannot find annotation method 'nanos()' in type 'DurationMin'`) in downstream modules that don't carry hibernate-validator on their annotation classpath — the same pattern `Quota.duration` already uses, just more visible.


loicmathieu (approved):
Overal LGTM.

I wonder if `nanos = 1` is not a little too aggressive, most of the time a duration of less of 1s is not realistic.
But it may also be too optimistic.

Your call but I'll at least mandate 1ms et not 1ns.

PR #18686, opened 2026-08-26
build(ui): split the shiki-langs chunk into one chunk per grammar

tchiotludo:
### 🔗 Related Issue

_No related issue._

### ✨ Description

The `shiki-langs-*.js` build chunk was 5.2 MB (968 kB gzip) — Shiki's entire on-demand grammar registry (~215 non-pre-registered languages) collapsed into a single lazy chunk by `ui/plugins/consolidateChunks.js`, so highlighting one code fence in an unusual language downloaded every language's grammar.

`KsMarkdown` pre-registers 29 grammars statically (`shikiHighlighter.ts`) and falls back to `shiki/langs`'s full bundle for anything else. The chunk-group `name` is changed from a constant `"shiki-langs"` to a function that names each grammar its own chunk (`shiki-lang-<name>`), plus one small chunk for the registry index that maps a language to its module. Pre-registered grammars are unaffected — they stay bundled into the `markdown` chunk as before.

Verified with a full `npm run build`: largest chunk dropped from 5.2 MB to 780 kB (`emacs-lisp`, 198 kB gzip), 214 grammar chunks + 1 registry chunk emitted, and the existing build-time invariants in `consolidateChunks.js` (no static edge into a lazy toolchain, no import cycle, no static edge into an on-demand grammar chunk) all held without changes to their logic — only their naming pattern was widened from `shiki-langs` to `shiki-lang(-|s-registry)`.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit` — ran the full `KsMarkdown` suite, including the on-demand-grammar test)
- [ ] Translations are complete if `en.json` changed — not applicable, no translation keys touched
- [ ] Screenshots or video recordings attached showing the `UI` changes — not applicable, build-output-only change with no rendered UI difference; verified via build output file sizes instead

### 📝 Additional Notes

This trades one large file for ~215 small ones in the served `assets/` directory (and in the shipped jar). If that file-count cost is unwanted, an equally small change buckets grammars into a handful of chunks instead of one-per-grammar — happy to switch if preferred.

### 🤖 AI Authors

This PR was raised by an AI. Why did the cat sit on the shiki chunk? Because it wanted to load *purr*-grammar on demand, one byte at a time. 🐱


ui/plugins/consolidateChunks.js
```diff
@@ -128,6 +128,14 @@ const matchesWithLazySubtree = (name, pattern) => (id) =>
 const LANG_MODULE = /node_modules[\\/](@shikijs[\\/]langs|shiki[\\/]dist[\\/]langs)[\\/]/
 const LANG_REGISTRY = /node_modules[\\/]shiki[\\/]dist[\\/]langs(-bundle-full-[^\\/]+)?\.mjs$/
 
+/**
+ * True for the on-demand grammar chunks: one per language, plus the registry
+ * index mapping a language to its chunk. Matches both a bare chunk name and the
+ * hashed file it is emitted as.
+ * @param {string} name
+ */
+const isLangChunkName = (name) => /(^|[\\/])shiki-lang(-|s-registry)/.test(name)
```
elevatebart: Could we exclude some language from this lazy loading? 
I was thinking the most common languages would be Docker/JS/TS/Python/Go/Java.
I would like them to be included in the main Shiki chunk. 
This way, common docs pages do not hit the WebServer with 6 more  requests.

ui/plugins/consolidateChunks.js
```diff
@@ -365,13 +373,19 @@ const GROUPS = [
...
     },
-    // Every grammar that is not pre-registered, in one chunk fetched only when a
-    // code fence uses an exotic language (see loadLanguageOnDemand). Without
-    // this group Shiki's registry emits one chunk per language (~350 files).
+    // One chunk per grammar that is not pre-registered, so an exotic code fence
+    // fetches only the language it names (see loadLanguageOnDemand) instead of
+    // the whole registry, plus the registry index that maps a language to its
+    // chunk. Claimed here rather than left to automatic chunking because the
+    // markdown group's pattern also matches @shikijs/langs and would otherwise
+    // pull every grammar into the chunk that every markdown render downloads.
```
elevatebart: Keep the comments under a sentence as per AGENTS.md.

elevatebart (commented):
Keep comments under check please

ui/plugins/consolidateChunks.js
elevatebart: Good instinct, but those six were already excluded — Docker, JS, TS, Python, Go and Java are six of the 29 grammars `shikiHighlighter.ts` pre-registers statically, and this group's `test` subtracts them: `staticLangs` is derived from the module graph in `collectStaticLangs()`, so anything a module statically imports (plus what it transitively embeds) stays in the `markdown` chunk and never becomes a per-language chunk. So no page was going to gain six requests.

Verified rather than assumed, on the baseline build before my commit — none of the six has a chunk:

```
grammar chunks: 214 + 1 registry
pre-registered grammars with their own chunk: none
  (no shiki-lang-dockerfile, -javascript, -typescript, -python, -go, -java)
```

Alias spellings hold too, which is the part that could have bitten us: `KsMarkdown.vue:434` gates on `getLoadedLanguages().includes(lang)`, and `@shikijs/primitive`'s registry returns resolved grammar names **plus** alias keys (`_alias` is filled from `lang.aliases` on load), so I ran the real highlighter with the actual pre-registered set to confirm. ` ```docker `, ` ```js `, ` ```ts `, ` ```py `, ` ```sh `, ` ```yml `, ` ```jsx `, ` ```tsx ` — all zero-request. 61 spellings resolved for free, well beyond the 29 modules, because `html` embeds `css`/`js` and `markdown` embeds a dozen more.

**What I did add**, since your underlying point stands for languages that genuinely were missing. Selected on evidence, not taste — a language Kestra ships a task runner for will appear in that plugin's docs, which is exactly the markdown `KsMarkdown` renders:

| Added | Why | Cost (gzip) |
|---|---|---|
| `powershell` (`ps`, `ps1`) | `io.kestra.plugin.scripts.powershell` | 4.1 kB |
| `julia` (`jl`) | `io.kestra.plugin.scripts.julia` | 4.3 kB |
| `toml` | config examples in script-plugin docs | 1.3 kB |
| `diff` | near-free | 0.7 kB |
| `shellsession` (` ```console `) | near-free | 0.4 kB |

Measured on a real build: zero-request spellings 61 → 70, grammar chunks 214 → 209, and the `markdown` chunk 338.7 → 348.0 kB gzip (**+9.3 kB**, +59 kB raw). `julia` costs only 4.3 kB despite embedding C, GLSL, SQL, C++, Python, JS and R because every one of those is already resident.

**Held back deliberately**, because each is downloaded by every markdown render whether or not anyone opens that doc page — say the word and I'll add any of them:

- `csharp` — real evidence (`io.kestra.plugin.scripts.dotnet` runs inline C#), but 10.9 kB gzip on its own, more than all five above combined. The one I'd most like your call on.
- `kotlin` 2.2 kB — JBang mentions it, but its examples are Java.
- `bat`/`cmd` 3.3 kB — our Windows docs use PowerShell.
- `nushell` 5.2 kB, `vue` 3.2 kB — no plugin; `vue` fences only appear in repo-internal READMEs, not rendered surfaces. `scss` needed nothing, it is already free via `html`.
- ` ```env ` cannot be fixed this way at all — `env` is not a Shiki language or alias, so it renders as plain text no matter what we pre-register. `dotenv` is the only spelling that works, and it never appears in our docs, so I left it out.

One honest correction to the PR description while I'm here: an exotic fence costs **two** requests, not one — the registry index chunk, then the grammar — and more if that grammar embeds others that aren't resident.

---

ui/plugins/consolidateChunks.js
elevatebart: Trimmed in `9b0ab8d` — three of the comments this branch added, one sentence each:

- the one you flagged, above the grammar group: four sentences down to one, keeping the "why claimed here rather than left to automatic chunking" half, since that is the part a reader cannot get from the code
- the `isLangChunkName` JSDoc: three lines of prose down to one line
- the static-edge check in `generateBundle`: already one sentence, now fits on two lines instead of three

I left the pre-existing multi-line comments elsewhere in the file alone, per the AGENTS.md rule that existing comments stay unless the code they describe changed.

---

ui/plugins/consolidateChunks.js
elevatebart: Numbers above are now stale — correcting them, since three things landed after that reply.

**`csharp` and `kotlin` added** (`23358be`), the two I had held back. That was the whole of my reservation about them, so with the call made there is nothing further to argue: C# had the better evidence of the two (`io.kestra.plugin.scripts.dotnet` runs inline `.csx`), Kotlin comes via JBang.

**The registry index now rides in the `markdown` chunk** (`5e35999`). It reaches its grammars only through dynamic imports, so leaving it to the markdown group adds no static edge and the invariant still holds — an exotic fence now costs **one** request instead of two, which also retires the correction I made at the end of that reply. `LANG_REGISTRY` went with it and `isLangChunkName` narrowed to `shiki-lang-`, as no registry chunk is emitted any more.

**Merged `develop` in** (`9177845`). It had gone `dirty` against [#18679](https://github.com/kestra-io/kestra/pull/18679), which reworked the same `generateBundle` region; both checks are kept — mf-shared's cycle check, then the grammar static-edge check. Merge rather than rebase, since this is not my branch to force-push.

Measured on the merged tree:

| | at `30dd742` | now |
|---|---|---|
| `markdown` chunk | 338.7 kB gzip | **368.2 kB gzip** (+29.5) |
| grammar chunks | 214 + 1 registry | **207 + 0** |
| zero-request spellings | 61 | **74** |
| exotic fence | 2 requests | **1 request** |

+29.5 kB gzip on every markdown render is the honest headline, and it is a deliberate trade for round trips rather than a saving: 11.4 kB of it is `csharp`/`kotlin`, 8.8 kB the index, 9.3 kB the first five grammars. The split itself still does its job — the largest grammar chunk is unchanged at 191.4 kB gzip, against the 968 kB single chunk this PR replaced.

One inaccuracy left that is not mine to edit: the PR description still says "plus one small chunk for the registry index", which is no longer true.

`npm run build` clean with all three invariants holding, `check:types` clean, `test:lint` clean (3 warnings, all pre-existing and elsewhere), 215 files / 2039 unit tests pass.

---

elevatebart (approved):

PR #18691, opened 2026-08-27
fix(core): reject empty switch branches

### LINKED ISSUE #18500: [Bash2.0][Flow] Switch with an empty matched case hangs forever in RUNNING
## Describe the issue
A `Switch` whose matched case has an empty task list never terminates. The execution stays in `RUNNING` forever with no child task, and nothing after the Switch ever runs.

## Steps to reproduce
1. Create and run this flow (the matched case `A` is empty):
   ```yaml
   id: switch_empty_case
   namespace: company.team
   tasks:
     - id: switch
       value: "A"
       cases:
         A: []
         B:
           - id: b
             message: b
     - id: after
       message: after
   ```
2. Observe the execution never leaves `RUNNING`.

## Actual behaviour
The execution stays in `RUNNING` indefinitely. Only the `switch` task is present, in `RUNNING`, with no child task. `after` never runs, and the execution cannot resolve on its own (it has to be killed).

## Visuals

Image

## Environment
- Kestra Version: 2.0.0-SNAPSHOT (develop)
- Edition: Open Source & Enterprise (core flowable)

ME-Massine:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18500.

### ✨ Description

Rejects flows containing empty `Switch` cases or explicitly empty `defaults` during validation.

Previously, these flows could pass validation and remain indefinitely in the `RUNNING` state. Regression tests cover both empty cases and empty defaults.

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched module
- [x] New behavior is covered by tests

### 📝 Additional Notes

Validated with Java 25.0.3:

- `FlowTest`: 15 passing
- `SwitchTest`: 6 passing
- Core build: 3,388 passing, 6 pending

The repository-wide Spotless check still reports unrelated pre-existing violations; neither changed Java file is affected.


core/src/main/java/io/kestra/core/validations/validator/SwitchTaskValidator.java
```diff
@@ -32,6 +35,31 @@ public boolean isValid(
...
+                .map(Map.Entry::getKey)
+                .findFirst();
+
+        if (emptyCase.isPresent()) {
+            context.disableDefaultConstraintViolation();
+            context.buildConstraintViolationWithTemplate("The case '%s' must define at least one task.".formatted(emptyCase.get()))
+                .addConstraintViolation();
+
+            return false;
+        }
```
loicmathieu: I think this can be replaced by a `@NotEmpty` inside the `Switch` class at the declaration side.

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/validations/validator/SwitchTaskValidator.java
loicmathieu: Inside the Switch class you have:

```java
// @FIXME: @Valid break on io.micronaut.validation.validator.DefaultValidator#cascadeToOne with "Cannot validate java.util.ArrayList"
    // @Valid
    @Schema(
        title = "The map of keys and a list of tasks to be executed if the conditional `value` matches the key"
    )
    @PluginProperty(additionalProperties = Task[].class)
    private Map<String, List<Task>> cases;
```

I think it should be something like that:

```java
    @Schema(
        title = "The map of keys and a list of tasks to be executed if the conditional `value` matches the key"
    )
    @PluginProperty(additionalProperties = Task[].class)
    private Map<String, @NotEmpty List<@Valid Task>> cases;
```

loicmathieu (approved):
LGTM

loicmathieu:
@ME-Massine tests are failing, we changed our validation message it can explain the failure.
Can you have a look please?

loicmathieu:
Hearing no feedback from the author I'm closing.

PR #18696, opened 2026-08-27
fix(core): redact encryptedOutputs markers from task runner logs

### LINKED ISSUE #14144: `encryptedOutput` must not be visible in DEBUG logs
### Feature description

Describe the issue
Sample Flow
```
id: encryped_output
namespace: company.team

tasks:
  - id: set
    values:
      val: "my secret value"

  - id: hello
    script: |
      echo '::{"outputs":{"plaintext":"plaintext_value"}}::'
      echo '::{"encryptedOutputs":{"encrypted":{{ outputs.set.values.val }}}}::'

  - id: print
    format: "{{ outputs.hello.vars.encryptedOutputs }}"
```

the `encryptedOutputs` must not be visible in the logs ideally
But instead it is available in DEBUG logs

Image

loicmathieu:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/14144.

### ✨ Description

**Problem:** an inline `shell.Script` task that declares an output sensitive via the `::{"encryptedOutputs":{"key":"value"}}::` marker had that value stored encrypted in the execution outputs — but the runner logged the fully rendered command (the whole `/bin/sh -c <script>` payload) at DEBUG before executing it, leaking the secret in clear text into the execution log, persisted in the DB and shown in the UI. Existing secret masking (`RunContextLogger.usedSecret`) couldn't catch this, since the value comes from a plain task output, never registered as a secret.

**Fix:** added `TaskLogLineMatcher.redactEncryptedOutputs(String)`, a textual, single-line replacement of any `::{...encryptedOutputs...}::` frame with `******` (the payload may not be valid JSON, so this can't be a JSON-based redaction). It's called before the "Starting command" DEBUG line in both the `Docker` and `Process` task runners, and in `TaskLogLineMatcher`'s own warn/OTLP-forwarding paths, which echoed the same raw frame. Plain `outputs` frames are left untouched — they're already shown in clear text in the Outputs tab, so redacting them would only remove debugging value.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/tasks/runners/TaskLogLineMatcher.java
+++ b/core/src/main/java/io/kestra/core/models/tasks/runners/TaskLogLineMatcher.java
@@ -57,6 +57,13 @@ public class TaskLogLineMatcher {
 
     protected static final ObjectMapper MAPPER = JacksonMapper.ofJson(false);
 
+    // The key is EE-only, but the marker protocol is shared, so OSS has to redact it wherever a raw command or log line is emitted.
+    private static final String ENCRYPTED_OUTPUTS_KEY = "encryptedOutputs";
+
+    private static final Pattern ENCRYPTED_LOG_DATA = Pattern.compile("::\\{.*?" + ENCRYPTED_OUTPUTS_KEY + ".*?}::");
+
+    private static final String REDACTED = "******";
+
     /**
      * Attempts to match and extract structured data from a given log line.
      * <p>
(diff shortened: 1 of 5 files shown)


loicmathieu:
@fdelbrayelle it fixes a real security issue (secret leak in DEBUG log), and it would most probably need to be used by all taskrunners as they might also log the command in debug as did Process and the Docker.

fdelbrayelle:
@loicmathieu Is it backportable to Kestra 1.3 :pray: ?

fdelbrayelle (approved):

loicmathieu:
Backported to 1.3 (not to 1.0 as there was too many changes here)

PR #18701, opened 2026-08-27
fix(core): surface swallowed evaluation errors instead of hiding them

### LINKED ISSUE #18454: [Bash2.0] Loop with a string values leaks a raw Jackson exception into the user logs

nkwiatkowski:
Closes https://github.com/kestra-io/kestra/issues/18454.
Closes https://github.com/kestra-io/kestra-ee/issues/10309.
Closes https://github.com/kestra-io/kestra-ee/issues/10297.

## Problem

Three places caught an evaluation error and silently discarded it, so a broken flow looked healthy while doing the wrong thing.

## Fix

- **Loop `values` (#18454)** — a plain string (or an expression rendering to a scalar) made `FlowableUtils.resolveValues` leak a raw Jackson `JsonParseException` into the user log. It now throws a clear `IllegalVariableEvaluationException` — *"The `values` must be a list, a map, or an expression that renders to one, but got: '…'"* — keeping the Jackson detail as the cause for debug logs.
- **SLA assertion (#10309)** — a broken `EXECUTION_ASSERTION` expression was caught and the SLA skipped, letting the run pass as if it had no SLA. `SLAService` now treats an unevaluable SLA as violated with the **SLA's own declared behavior** (fail-closed), consistent with how flow `checks` already stop on a broken gate.
- **Schedule trigger input (#10297)** — a broken input expression made the trigger silently produce no execution, with nothing on the server log. `SchedulableEvaluator` now logs the failure at **ERROR on the server log** so operators can see why the trigger isn't running.

## Evidence

`FlowableUtilsTest` (+2: plain string asserts no Jackson leak; non-container JSON), new `SLAServiceTest` (broken assertion → violation with declared behavior), and the `H2RunnerTest` gate all pass.

## Scope note

`#10297` is fixed as server-log visibility only — surfacing the error on the Triggers page would need a new trigger-state error field, and creating a synthetic failed execution was deliberately excluded, both out of scope for this bugfix.

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/FlowableUtils.java
+++ b/core/src/main/java/io/kestra/core/runners/FlowableUtils.java
@@ -447,8 +447,15 @@ public class FlowableUtils {
                         }
                         return Either.right(resolvedValues);
                     } else {
-                        throw new IllegalVariableEvaluationException("Unknown value type: " + valuesNode.getNodeType());
+                        throw new IllegalVariableEvaluationException(
+                            "The `values` must be a list, a map, or an expression that renders to one, but got: '" + renderValue + "'."
+                        );
                     }
+                } catch (JsonProcessingException e) {
+                    throw new IllegalVariableEvaluationException(
+                        "The `values` must be a list, a map, or an expression that renders to one, but got: '" + renderValue + "'.",
+                        e
+                    );
                 } catch (IOException e) {
                     throw new IllegalVariableEvaluationException(e);
                 }
--- a/core/src/test/java/io/kestra/core/runners/FlowableUtilsTest.java
+++ b/core/src/test/java/io/kestra/core/runners/FlowableUtilsTest.java
@@ -316,6 +316,30 @@ class FlowableUtilsTest {
     }
 
     @Test
+    void resolveValues_withPlainString_shouldThrowClearMessageNotJacksonError() {
+        // Given
+        RunContext runContext = runContextFactory.of();
+
+        // When/Then
+        assertThatThrownBy(() -> FlowableUtils.resolveValues(runContext, "hello world"))
+            .isInstanceOf(IllegalVariableEvaluationException.class)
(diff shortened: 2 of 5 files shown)


loicmathieu (approved):
LGTM

PR #18708, opened 2026-08-27
fix(triggers): keep a single catalog entry per trigger type on the triggers add page

### LINKED ISSUE #18419: [Bash2.0] Triggers Add page: re-applying a filter accumulates and duplicates the list
## Describe the issue

On the Triggers Add page, re-applying a filter keeps adding to the list instead of replacing it. The trigger list grows and duplicates each time a filter is re-applied.

## Steps to reproduce

1. Tenant -> Triggers -> Add.
2. Click "All", then "Realtime" (shows a set of triggers).
3. Click "All" again, then "Realtime" again. Repeat.

## Actual behaviour

Each time the filter is re-applied, more triggers appear and entries duplicate. (See attached video.)

## Visuals

https://github.com/user-attachments/assets/2cbfe5b2-1d8e-4c26-9e96-9710d5cdc0bd

---

## Environment

- Kestra Version: rc-10
- Edition: Enterprise

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/18419.

### ✨ Description

On the Triggers Add page, re-applying a category filter (All -> Realtime -> All -> ...) made the trigger list visibly grow and duplicate on every round.

Root cause: a plugin installed in several versions registers one `RegisteredPlugin` per version, so `GET /api/v1/plugins/triggers` returned the same trigger `type` once per installed version. The grid keys its cards by `type`, and duplicate keys corrupt Vue's keyed diff: every filter re-apply leaks stale DOM nodes, so the list keeps growing (reproduced locally against a duplicated `API` response: 11 -> 19 -> 27 -> 35 cards after three toggles).

Fix, in both layers:

- webserver: `PluginController#listTriggerPlugins` now keeps a single catalog entry per trigger `type`, matching its own `OpenAPI` description ("one entry per trigger class").
- `UI`: the plugins store drops duplicate `type`s defensively, so the grid stays correct even against a backend that still returns duplicates.

With the fix, the same duplicated `API` response renders a stable grid (7 cards on All, 3 on Realtime) across any number of filter toggles.

Realtime filter applied after three All/Realtime rounds, against an `API` response that repeats the TCP and UDP `RealtimeTrigger` types:

| Before | After |
|---|---|
| before | after |

### 🎨 Frontend Checklist

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched modules (`./gradlew :webserver:test`)

### 📝 Additional Notes

The duplication is easiest to hit on `EE`, where plugin management installs several versions of the same plugin, but any registry that registers the same trigger class twice (for example two versions of a jar in the plugins directory) triggers it on `OSS` too.

Why did the cat sit on the trigger catalog? Because it heard every card had a `type` and it refused to be filtered out twice. 🐱


Piyush-r-bhaskar (approved):
Image

@Malaydewangan09 when you get a chance for BE part

PR #18714, opened 2026-08-27
fix(executor): fail the flowable task run on invalid config instead of looping

### LINKED ISSUE #18702: [Bash2.0] Flowable config validation (Parallel/Dag concurrent, LoopUntil) must fail the task run, not the execution
### Describe the issue

Some flowable-task config fields accept invalid values (negative counts / non-positive durations) that are silently misbehaved at runtime instead of failing the execution:

- `Parallel.concurrent` and `Dag.concurrent` — a negative value runs as a silent no-op `SUCCESS`.
- `LoopUntil.checkFrequency.interval` / `maxIterations` / `maxDuration` — non-positive values are accepted.

This is the remaining, harder part of #10274 (the `Parallel.concurrent` bullet), split out because it cannot be fixed the same way as the plain / non-flowable `Property` fields.

### Why the simple fix does not work

These fields are `Property<...>`, so a bean-validation constraint on the generic type argument (`Property<@PositiveOrZero Integer>`) is the right tool, and it correctly:
- rejects a **literal** bad value at save time (via `ModelValidator` + `PropertyValueExtractor`), and
- throws a `ConstraintViolationException` at runtime when an **expression** renders to a bad value.

The problem is the runtime path. `concurrent` is rendered inside `FlowableTask.resolveNexts`, called from `ExecutorService.childNextsTaskRun`. Letting the `ConstraintViolationException` propagate so it fails the execution (`process()` → `withException`) **re-introduces a poison-pill**: the flowable's child task run is still `RUNNING`, so the execution state is recomputed back to `RUNNING` and the executor reprocesses it, re-throwing every cycle. Verified with a runner test — the execution logged `Flow failed from executor ... 'concurrent: must be greater than or equal to 0'` repeatedly and never reached a terminal state.

### Correct approach

Fail the flowable **task run** (mark it `FAILED` with the constraint message), not the execution — reusing the existing "panic mode" pattern in `ExecutorService.childWorkerTaskResult` (~L266), where a `resolveState` failure sets the parent flowable task run to `FAILED` and the execution terminates cleanly. The `resolveNexts`/validation failure needs to be routed through that same task-run-failure path rather than thrown to the execution level.

Once that mechanism is in place, add the validation annotations:
- `Parallel.concurrent`, `Dag.concurrent`: `Property<@PositiveOrZero Integer>`
- `LoopUntil.checkFrequency`: `maxIterations` `@Positive`, `interval` / `maxDuration` `@DurationMin(nanos = 1)`

### Acceptance

- A flow with a negative/non-positive value in any of the above **fails the execution** with a clear message (both literal-at-save and expression-at-runtime).
- No reprocessing loop / poison-pill — the execution reaches terminal `FAILED`.
- `H2RunnerTest` green (this touches the flowable state machine and is the mandatory gate).

### Context

Follows the bug-bash "don't swallow evaluation errors" work: PR #18701 (three localized fixes) and PR #18676 (#10274 plain-field validation). The `resolveNexts` swallow was deliberately left unchanged in #18701 to avoid this poison-pill.

nkwiatkowski:
Closes https://github.com/kestra-io/kestra/issues/18702.

## Problem

Flowable-task config fields accepted invalid values that were silently misbehaved at runtime rather than failing the execution: `Parallel.concurrent` / `Dag.concurrent` ran a negative value as a no-op `SUCCESS`, and `LoopUntil.checkFrequency` (`maxIterations` / `interval` / `maxDuration`) accepted non-positive values.

This is the remaining, harder part of #10274 (the `Parallel.concurrent` bullet), split out because the naive fix re-introduces a poison-pill.

## Fix

Validate the fields with bean-validation on the `Property` type argument — `Property<@PositiveOrZero Integer>`, `@Positive`, `@DurationMin(nanos = 1)` — which rejects a **literal** at save time and throws a `ConstraintViolationException` when an **expression** renders to a bad value at runtime. `LoopUntil.checkFrequency` gets `@Valid` so the nested constraints cascade.

The runtime failure is routed through the flowable **task-run-failure** path — the same pattern as `b08c9bfa07` ("gracefully handle failing flowable tasks"). `childNextsTaskRun` re-throws the `ConstraintViolationException`, and the caller marks the flowable **task run** FAILED:

```java
} catch (ConstraintViolationException e) {
    runContext.logger().error("Failed to process flowable task {}: {}", taskRun.getId(), e.getMessage(), e);
    executor.withExecution(executor.getExecution().withTaskRun(taskRun.withState(State.Type.FAILED)), "flowableValidation");
}
```

Failing the **execution** directly (letting the exception propagate to `process()` → `withException`) does not work here: the flowable child task run is still `RUNNING`, so the execution state is recomputed back to `RUNNING` and reprocessed every cycle — an infinite loop. Failing the **task run** lets the normal state machine terminate the execution cleanly.

## Evidence

- `ParallelTest.parallelWithNegativeConcurrentShouldFailExecution` and `DagTest.dagWithNegativeConcurrentShouldFailExecution` — an expression-rendered negative `concurrent` fails the execution (execution + flowable task run both `FAILED`) in seconds, no loop.
- `FlowableConcurrencyValidationTest` — literal negatives are rejected at save; `0` (unlimited) is accepted.
- `H2RunnerTest` green — no regression across the flowable suite.

## Context

Follows the bug-bash "don't swallow evaluation errors" work (PR #18701). The `resolveNexts` swallow was left untouched there specifically to avoid this poison-pill; this PR routes it through the correct task-run-failure path instead.


core/src/test/java/io/kestra/plugin/core/flow/FlowableConcurrencyValidationTest.java
```diff
@@ -0,0 +1,61 @@
...
+import io.kestra.core.models.property.Property;
+import io.kestra.plugin.core.log.Log;
+
+import io.micronaut.validation.validator.Validator;
+import jakarta.inject.Inject;
+import jakarta.validation.ConstraintViolation;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@KestraTest
```
loicmathieu:
```suggestion
@MicronautTest
```

loicmathieu (approved):
LGTM

PR #18735, opened 2026-08-27
feat(topology): hand topology-details the tenant, the task source and lazy fetchers

elevatebart:
### 🔗 Related Issue

Part of https://github.com/kestra-io/kestra-ee/issues/10487 — the props asked for here; the rest of that issue (`renderedTask`, `taskRun`, `openTaskModal`) is not in this PR.

### ✨ Description

A `topology-details` artifact today calls the API for things the host already knows, so opening a topology node fires requests that the host could have answered: `/metrics/*` for the task's metrics, `/outputs/*` for its outputs, `/flows/*` to re-read the task, and — in `plugin-ai` — a `localStorage.getItem("selectedTenant")` read to work around a bundled SDK copy whose `globalTenant` stays at `"main"`.

Four props are added to the slot contract in `ui/packages/slot-contracts/src/topology-details.ts` (shared as-is by `topology-task-drawer` and `topology-task-modal`) and bound at every host call site in `LowCodeEditor.vue` — the topology node slot, the task drawer, and the modal context built by `showCustomAction`:

- **`tenant`** — the current tenant, from the route.
- **`task`**, now the graph node's task with the same task parsed out of `source` filling in behind it. An execution graph is served through `forExecution()`, which keeps only `id`, `type`, `version` and children, so plugins parse the flow themselves to get the rest back. Source fills those stripped keys in but never overrides what the node carries: it is the flow's latest revision, or an unsaved draft, so it must not restate what an older revision actually ran with. The index is seeded from the flow's task collections, so a task nested in a flowable resolves while an `inputs`/`sla`/`triggers` entry sharing an id cannot shadow one.
- **`fetchOutputs(query?)`** — that task run's outputs, replacing `OutputsAPI.taskRunOutputs({executionId, taskRunId})` and the `ExecutionAPI.execution(id)` fallback.
- **`fetchMetrics(query?)`** — `MetricsAPI.searchByExecution` with the execution, the tenant *and this task* already bound.

The last two are functions rather than values deliberately: an artifact that renders neither outputs nor metrics costs no request, which is why no `outputs` prop is added. An artifact is rendered for one task, so neither asks it which one; both take an optional `taskRunId` to address one iteration of a looped task, and `fetchMetrics` also takes paging and `sort`. Both resolve the execution and task run **when called, not when bound** — an artifact is handed its props once, when the graph is generated, and the task run it should read may only come into existence later (a playground run, a replay). Outside an execution they resolve to `{}` and an empty page rather than being absent.

**`metrics` is removed** — from the contract, from the three host call sites, and from the executions store (`metrics` and `loadMetrics` both). Nothing ever wrote it: `executionsStore.metrics` was only filled by `loadMetrics({store: true})`, and the sole caller in OSS and EE passes `store: false`. So the topology paid for it on every node — an always-empty array filtered per task, plus a watcher that regenerated the whole graph whenever the ref changed — and an artifact reading `props.metrics` got `[]`, which is why `plugin-gcp` fetches metrics itself. `fetchMetrics` covers that case, on demand. The execution Metrics tab (`MetricsTable.vue`) already kept its results in local refs, so it now calls `MetricsAPI.searchByExecution` directly and the store loses an indirection it was the only user of. `flowStore.metrics`, which backs the flow Metrics tab and its filter dropdown, is untouched.

One fix came along the way: `generate-flat-dts` counted the `>` of an arrow as a closing bracket, so the first function-typed prop left its depth counter negative and swallowed every prop after it into one line — and dropped their names from the `KnownSlotsPropNames` declaration. This is type-only: the runtime array comes from `Object.keys(props.shape)` in `src/index.ts` and was always complete, but a plugin compiling against the published package would not see `fetchMetrics` typed. Its two near-identical scanners are now one function that carries the exception, covered by a unit test that fails on the old scanner.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit` — 221 files, 2090 tests)
- [x] Translations are complete if `en.json` changed (no strings changed)

No screenshot: this PR renders nothing new. It hands existing data to plugin artifacts, and the visible effect belongs to the plugin that consumes the props (per the issue, a BigQuery artifact rendering with zero requests in the network tab) — verifiable only against an instance with `plugin-gcp` installed and a plugin build that reads them.

### 📝 Additional Notes

- **Dropping `metrics` is a contract break for a published artifact** that reads `props.metrics` — it now gets `undefined` where it got an empty array. Nothing in-tree reads it and it carried no data anyway, but plugin authors move to `fetchMetrics`.
- The `taskDetailsVersion` bump now watches `progressEvents` alone, and was deliberately not extended to the flow source: the graph already regenerates on `flowGraph`, and watching the source would regenerate every node on each keystroke.
- `runnerTypeByTaskId` / `augmentedFlowGraph` still do their own top-level-only version of the source lookup. They read `flowStore.flowParsed` with a different fallback than the new index does, so folding them together is a behavior change to the taskRunner re-injection rather than a rename — left for a follow-up.

### 🤖 AI Authors

Why did the cat sit on the topology graph? It heard there were plenty of *nodes* to nap on — and it refused to fetch anything. 🐱


MilosPaunovic (changes requested):
Reviewed the diff and the files at the `PR` head, plus a real `packages/slot-contracts` build to check the codegen.

**Verified**

- The `generate-flat-dts` fix is correct. Rebuilding `dist/index.d.ts` at this head emits all 10 props, one per line, with all 10 in `KnownSlotsPropNames`. With the `previous !== "="` exception removed (`develop`'s scanner), `fetchOutputs` and `fetchMetrics` collapse onto one line and `fetchMetrics` disappears from `KnownSlotsPropNames`. Bug and fix both confirmed.
- Dropping `metrics` / `loadMetrics` from the executions store is safe: `executionsStore.metrics` had two readers, both in `LowCodeEditor.vue`, and `loadMetrics` one caller (`MetricsTable.vue`, with `store: false`). `ui-ee` references neither, and no `EE` code touches the three slot contracts. `FlowMetrics.vue`'s `loadMetrics` is a local function of its own.
- `searchByExecution` does accept `taskId`, and `PagedResultsMetricEntry` is exactly `{results, total}`, so the `{results: [], total: 0}` fallback matches the declared return type. `route.params.tenant` matches how the rest of `OSS` reads the `tenant`. `eslint` is clean on all five changed files.

**1. The current flow source overrides the executed task, so an artifact can be handed config that never ran.** `taskWithSource` merges `{...task, ...fromSource}`, source winning. In the execution topology the source is `flowStore.flow?.source` from `flowStore.loadFlow({namespace, id})` with no revision (`Topology.vue:156`), i.e. the flow's latest source, and `flowSource` prefers `flowStore.flowYaml`, which can be an unsaved editor draft from earlier in the session. Open an execution of revision 3 after the `task` was edited and the artifact receives revision 4's values as the `task` that ran, silently replacing what the graph node carried from the executed revision. The stated goal is only to restore what `forExecution()` strips, which is achieved with source as the *fallback* instead: `{...fromSource, ...task}` (stripped keys are absent on the graph node, so they still come through), or keep the current direction only when there is no `execution`.

**2. `collectTasksById` indexes objects that are not `tasks`, and can shadow a real one.** The walk starts at the document root and indexes anything with a string `id` and a string `type`, first-wins. A flow's `inputs`, `outputs`, `sla` and `triggers` entries all qualify (`Flow.java`'s root task collections are only `tasks`, `errors`, `finally`, `afterExecution`; `SLA` and `Input` both carry `id` and `type`). Root key order is document order, so with the very common `inputs:` before `tasks:` layout, a flow with `inputs: [{id: query, type: STRING}]` and a `task` `id: query` indexes the input, and `taskWithSource` then merges `type: STRING` over the `task`: the artifact gets a `task` that is not a `task`, with `task.type` contradicting the `taskType` prop bound next to it. Seeding the walk from the root's task collections fixes it and shrinks the walk.

**3. `fetchOutputs` cannot address an iteration, unlike `fetchMetrics`.** `fetchMetrics` takes a `taskRunId` precisely so a looped `task` can be narrowed to one iteration, but `fetchOutputs` takes nothing and resolves through `currentTaskRunId`, which returns the last matching `taskRun`. Today an artifact calls `OutputsAPI.taskRunOutputs` with a `taskRunId` it picks itself; after this `PR` it gets one arbitrary iteration and no way to ask for another. Since this is a published contract, `fetchOutputs(query?: {taskRunId?: string})` now avoids a second contract change for plugin authors later.

**4. The codegen fix ships without a test.** Pure function, silent failure mode: no build error, a prop just vanishes from `KnownSlotsPropNames`. `splitTopLevelMembers` is not exported so it cannot be tested as it stands; exporting it plus one spec with an arrow-typed member would pin it. Specs under `packages/` are already picked up by the runner (`packages/kestra-sdk/src/dev-freshness.spec.ts`).

**5. The description overstates the codegen impact.** It says `KnownSlotsPropNames` "is what declares the props on the remote component. Without the fix, `fetchMetrics` would never have reached a plugin." The value `useFederatedModule.ts:145` declares props from is the runtime `KnownSlotsPropNames`, built in `packages/slot-contracts/src/index.ts:14` from `Object.keys(props.shape)`, and the flat-dts plugin only rewrites the `.d.ts` chunk (`tsdown.config.ts:13-18`). The runtime array was always complete; the breakage was type-only, so a plugin compiling against the published package would not see `fetchMetrics` typed. Worth correcting so nobody hunts for a runtime bug later.

**6. Same section, smaller.** `MetricsTable` was not overwriting the shared `total`: the removed `loadMetrics` returned early on `store === false` before assigning `total.value`, and `MetricsTable` was its only caller and passed `store: false`. Dropping the store indirection is still right, just not for that reason.

**7. Nit.** `metricsTotal.value = response.total ?? 0` - `PagedResultsMetricEntry.total` is a required `number`, so the `?? 0` is dead.

**8. Two source-reinjection paths now coexist.** `runnerTypeByTaskId` and `augmentedFlowGraph` do a narrow, top-level-only version of what `sourceTaskById` now does generally, and they are a third parse of the same `YAML` (`flowStore.flowParsed` is a fourth). Still load-bearing for `hasExtraDetails` and the graph nodes, so not dead code, but folding them into the new index would leave one path instead of two. Related: `sourceTaskById` re-parses and deep-walks the whole document on every keystroke in the editor, since the node slot reads it from the template. Bounded, and finding 2's whitelist would cut most of the walk.

**9. Contract break sequencing.** Dropping `metrics` is harmless in-tree (nothing read it, it was always `[]`). The only thing to weigh is that it lands on the published package before the `artifact-sdk` sync noted as still open, so a plugin author hits `props.metrics === undefined` before `fetchMetrics` is documented anywhere.

**Backport:** not applicable, `ui/packages/slot-contracts` does not exist on `releases/v1.3.x` or `releases/v1.0.x`, so this is 2.0-only.

elevatebart:
Thanks — all nine check out. Four are fixed in af64b16, two description claims are corrected, and one I'd rather leave for a follow-up.

**1. Source overriding the executed task — fixed.** `taskWithSource` is `{...fromSource, ...task}` now. `TaskForExecution` keeps only `id`, `type`, `version`, `tasks` and `inputs` and *drops* the other keys rather than nulling them, so source as the fallback still fills in everything `forExecution()` strips, while the executed revision wins wherever the node carries a value. The comment says that instead of the old claim.

**2. `collectTasksById` indexing non-tasks — fixed.** Seeded from the root task collections (`tasks`, `errors`, `finally`, `afterExecution`, per `Flow.java`) instead of the document root, so an `inputs` / `outputs` / `sla` / `triggers` entry can no longer shadow a task of the same id. `Input` and `SLA` do both carry `id` and `type`, so the `inputs: [{id: query, type: STRING}]` case was real. It also cuts most of the walk you flag in 8.

**3. `fetchOutputs` iteration — fixed.** `fetchOutputs(query?: {taskRunId?: string})`, defaulting to `currentTaskRunId(taskId)`. Symmetric with `fetchMetrics`, and no second contract change for plugin authors later.

**4. Codegen test — added.** `splitTopLevelMembers` is exported and pinned by `scripts/generate-flat-dts.spec.ts`: an arrow-typed member stays its own member, and a semicolon nested inside a member's type still doesn't split. It fails on develop's scanner. For context, that spec was in an earlier revision of this PR and was dropped at the author's request; re-added on your ask.

**5. Description overstated — corrected, you're right.** `useFederatedModule.ts:145` reads the runtime `KnownSlotsPropNames` built from `Object.keys(props.shape)` in `src/index.ts`, and the flat-dts plugin only rewrites the `.d.ts` chunk, so the runtime array was always complete and the breakage was type-only for a consumer of the published package. The body now says exactly that.

**6. `total` — corrected too.** The early `store === false` return did precede `total.value = ...`, and `MetricsTable` was the only caller, so nothing was being overwritten. Claim removed; dropping the store indirection stands on the plainer ground that the store was left with one caller and no state.

**7. Nit — fixed.** `?? 0` gone; `PagedResultsMetricEntry.total` is a required `number`.

**8. Two reinjection paths — agreed, but not here.** `runnerTypeByTaskId` reads `flowStore.flowParsed` and falls back to `props.source` only when flowParsed carries no runner types; `sourceTaskById` parses `flowStore.flowYaml || props.source`. Those two inputs diverge exactly in the case the existing comment was written for (execution view with a stale or absent `flowYaml`), so folding them is a behavior change to the taskRunner re-injection rather than a de-duplication, and I'd rather not slip it into a props PR. Happy to take it as a follow-up if you want it. The per-keystroke walk is much cheaper after 2.

**9. Sequencing.** The `artifact-sdk` contract is updated by its sync workflow once this merges, so the published contract and the SDK move together rather than the SDK trailing.

**Backport:** agreed, 2.0-only.

Re-verified on the new head: `check:types`, `npm run build`, full `test:unit` (221 files / 2090 tests) and eslint all clean.

---

MilosPaunovic (changes requested):
Re-reviewed at `af64b16e03`. `CI` is green, and locally the new spec passes with the function-typed case failing on the old scanner, `dist/index.d.ts` regenerates with all 10 props (all 10 in `KnownSlotsPropNames`), and `eslint` is clean on the six changed files.

Three things still open.

**1. Node-wins hands an artifact the reduced value for the keys `forExecution()` shrinks rather than drops.** `TaskForExecution.of()` does not only omit properties. For a `FlowableTask` it sets `tasks(flowable.allChildTasks().stream().map(TaskForExecution::of).toList())`, so the node's `tasks` is the children reduced recursively to `id` and `type`, and flattened through `allChildTasks()` so the shape does not even match the source's nesting. Separately, `augmentedFlowGraph` re-injects `taskRunner: {type: runnerType}` into the node before the slot ever sees it, so the node carries a `taskRunner` holding only its `type`. With `{...fromSource, ...task}` the node wins on both keys, so an artifact rendering a flowable's children, or a `taskRunner` (a `Docker` runner's `image`, `pullPolicy`, `cpu`, `memory`), receives the skeleton and the source definition is discarded.

The genuinely-stripped keys are fine: serialization is `non_null` (`cli/src/main/resources/application.yml:124`, `JacksonMapper:155`), so absent keys really are absent and source fills them, and `version` / `inputs` / `subflowId` are never non-null for a normal `task`, so there is no null-clobbering either. The problem is only the keys carried in reduced form. Merging per key instead of shallow-spreading, or exempting `tasks` and `taskRunner` from node precedence, covers it.

**2. `runnerTypeByTaskId` is now a top-level-only duplicate of `sourceTaskById`.** It parses the same `YAML` a third time (`flowStore.flowParsed` is a fourth) and only walks the root `tasks`, `errors` and `finally` lists, so a `taskRunner` on a task nested in a flowable is still missed, both for the `taskType` binding and for `hasExtraDetails`. Feeding the augmentation from `sourceTaskById` instead, and injecting the whole `taskRunner` object rather than just its type string, resolves the `taskRunner` half of finding 1, fixes nested flowables, and drops a parse. Worth noting `sourceTaskById` still re-parses and walks on every keystroke in the editor, although scoping it to the task sections made that much cheaper.

**3. Contract sequencing, unchanged from the last round.** Dropping `metrics` still lands on the published package ahead of the `artifact-sdk` sync, so a plugin author hits `props.metrics === undefined` before `fetchMetrics` is documented anywhere. Your call, not a blocker.

elevatebart:
Both code findings are in, one commit each.

**1. Reduced keys — 26a60d1.** You're right that `TaskForExecution.of()` shrinks rather than drops for `tasks`, and that `augmentedFlowGraph` puts a `{type}`-only `taskRunner` on the node before the slot sees it. `taskWithSource` now excludes those two from node precedence, so a flowable keeps its real nested children and a Docker runner keeps its `image` / `cpu` / `memory`. Everything else still lets the executed node win.

**2. Dedup — 63c330e.** `runnerTypeByTaskId` is gone; `augmentedFlowGraph` reads `sourceTaskById` and injects the whole `taskRunner` object. That drops a parse, fixes runners on tasks nested in a flowable (missed before by both the `taskType` binding and `hasExtraDetails`), and covers `afterExecution` too. The fallback moved with it: when no task in the index carries a runner, the index is rebuilt from `props.source`, since `flowYaml` can be a stale draft in execution view. The augmentation returns the same graph object when it injects nothing, so the topology doesn't regenerate on every keystroke.

**3.** Leaving as is — `artifact-sdk` is synced by its workflow on merge.

`check:types`, `npm run build`, `test:unit` (223 files / 2098 tests) and eslint clean on the merge head.

---

MilosPaunovic (changes requested):
Re-reviewed at `63c330e55b`. `CI` is all green or skipped and `eslint` is clean on the changed files. Two things the new commits introduced, plus the one carry-over.

**1. Which document fills the `task` prop is decided by whether any task happens to declare a `taskRunner`.** `sourceTaskById` returns the `flowSource` index if some task in it has `taskRunner.type`, and otherwise - whenever `flowSource !== props.source` - throws that index away and returns `indexTasks(props.source)`. So a flow with no `taskRunner` anywhere has its `task` prop filled from `props.source`, and adding a `taskRunner` to any unrelated task in the same flow switches it to `flowSource`. The `source` prop handed to the same artifact is always `flowSource`, so in the first case `task` and `source` come from two different documents.

That state is reachable and sticky rather than theoretical: `Topology.vue`'s `loadFlowSource()` returns early when `flowStore.flow` already matches the execution's flow, and nothing clears `flowStore.flowYaml` on navigation (`FlowFileEditorTab.vue:297` writes it per keystroke). Edit a flow's source, then open one of its executions: `loadFlow` is skipped, `flowYaml` is still the unsaved draft while `props.source` is the saved source.

`runnerTypeByTaskId` carried the same heuristic, but there it only chose which document a `taskRunner` *type string* came from; now it chooses the document that fills every key `forExecution()` drops. Deciding by context would be coherent - index `props.source` when there is an `execution` and `flowSource` otherwise - or index `flowSource` only and let the augmentation keep a narrow runner-specific fallback of its own.

**2. `taskWithSource` drops the node's `tasks` and `taskRunner` even when source carries neither.** `const {tasks: _tasks, taskRunner: _taskRunner, ...executed} = task` removes both keys unconditionally, so when `fromSource` has no `tasks` the result has no `tasks` at all: a flowable whose children exist on the node but not in the indexed document loses them entirely, where the previous revision of this code at least handed over the reduced ones. The divergent state in finding 1 reaches this, and so does an executed revision whose children the latest source no longer has. Taking a key from source only when source actually has it closes both.

**3. Carry-over, unchanged.** Dropping `metrics` still lands on the published package ahead of the `artifact-sdk` sync, so a plugin author hits `props.metrics === undefined` before `fetchMetrics` is documented anywhere. Your call.

One observation rather than a finding: the same runner-reconciliation heuristic still lives twice in this file, in the `props.flowGraph` and `props.source` watchers (`LowCodeEditor.vue:506-536`), each with its own `YAML_UTILS.parse` and its own `flowParsed` runner check. The refactor removed one copy; feeding those two from `sourceTaskById` would drop the remaining parses and keep the three in step. Pre-existing, so a follow-up is fine.

elevatebart:
Both fixed in 90dac5f.

- `sourceTaskById` is `indexTasks(flowSource)` unconditionally, so `task` and `source` always come from the same document.
- The runner fallback moved to a `taskRunnerById` computed used only by the graph augmentation, which is the whole of what it covered before the refactor.
- `taskWithSource` now takes `tasks` and `taskRunner` from source only when source has them, so a flowable whose children exist only on the node keeps the reduced ones instead of none.
- Everything else is unchanged: the node still wins on every other key, the augmentation still returns the same graph object when it injects nothing, and `metrics` sequencing stays as is.
- The two watcher copies of the runner heuristic (`:506-536`) are untouched — follow-up, as you suggested.

`check:types`, `npm run build`, `test:unit` (223 files / 2098 tests) and eslint clean.

---

MilosPaunovic (approved):
`90dac5f17c` closes both findings.

`sourceTaskById` is one document now, the same one the `source` prop carries, and the runner fallback survives as `taskRunnerById` scoped to the graph augmentation alone - which is the reach it had before this `PR`. `taskWithSource` is back to node-wins with source reclaiming `tasks` and `taskRunner` only where it has them, so a flowable whose children exist only on the node keeps them. The fallback interaction holds up too: when the draft has no runner and the node's was injected from `props.source`, `fromSource.taskRunner` is `undefined` and the injected value survives.

Nothing else open. The one thing still on your plate is `metrics` leaving the published package ahead of the `artifact-sdk` sync, which is a sequencing call rather than a defect. Worth landing after `Frontend - Tests` and `trigger-ee` finish.

PR #18740, opened 2026-08-27
fix(plugins): resolve auto-install artifacts from real class packages, not the manifest group

fdelbrayelle:
Part of https://github.com/kestra-io/kestra-ee/issues/6145 (KIP-45 QA follow-up; reported in Slack during develop-slim QA, no standalone issue).

### Problem

Reported during develop-slim QA (see [QA report](https://github.com/kestra-io/kestra-ee/issues/6145#issuecomment-5439129183)): saving a flow with `io.kestra.plugin.transform.jsonata.TransformValue` (or any `io.kestra.plugin.transform.*` type) auto-installs **plugin-transform-records** instead of the right artifact, and the save keeps failing with 422 on the still-missing type after the wrong jar is installed.

Root cause: some plugins declare an `X-Kestra-Group` manifest value that does not match their class packages — `plugin-transform-json` declares `io.kestra.plugin.json` while its types live under `io.kestra.plugin.transform.jsonata`, `plugin-transform-grok` declares `io.kestra.plugin.grok`. Both the hosted catalog and the schema bundle mirror that manifest value, so the longest-prefix type→artifact match only ever finds `io.kestra.plugin.transform` (plugin-transform-records) for those types.

### Fix

- `PluginsSchemaCommand` now derives each bundle entry's package group from the longest common package prefix of the plugin's scanned classes — the packages flow `type` values are actually matched against — falling back to the manifest group when a plugin registers no classes.
- `PluginCatalogService.withBundleEntries` keeps the hosted entry on a `groupId:artifactId` conflict but takes the bundle's group, since the hosted value mirrors the same wrong manifest while staying authoritative for license/version metadata.

The manifest values themselves should still be corrected in `kestra-io/plugin-transform` (and the hosted catalog data refreshed) — this change makes the resolution robust against that class of bad metadata either way. Note the fix takes effect on slim images once a new bundle is baked by release CI; the currently published bundles carry the manifest-derived groups.

### Evidence

Reproduced on `kestra/kestra:develop-slim`:

```
$ curl -X POST .../api/v1/plugins/auto-install/detect -d 'type: io.kestra.plugin.transform.jsonata.TransformValue ...'
{"artifacts":[{"groupId":"io.kestra.plugin","artifactId":"plugin-transform-records", ...}]}   # wrong
```

The jar's own content vs its manifest:

```
X-Kestra-Group: io.kestra.plugin.json
classes: io/kestra/plugin/transform/jsonata/**
```

Covered by `PluginsSchemaCommandTest` (common-prefix derivation) and `PluginCatalogServiceTest#shouldPreferBundleGroupOverHostedGroupForSameArtifact` (replacing the test that pinned the old hosted-group-wins behavior, which this bug proved wrong).

### DIFF
--- a/cli/src/main/java/io/kestra/cli/schema/PluginsSchemaCommand.java
+++ b/cli/src/main/java/io/kestra/cli/schema/PluginsSchemaCommand.java
@@ -2,6 +2,7 @@ package io.kestra.cli.schema;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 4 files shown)


fdelbrayelle:
Real fix: https://github.com/kestra-io/plugin-transform/commit/6ce9fa6f2cb8b10f0021d71652c62d8796fefa64

PR #16054, opened 2026-05-15
feat(plugins): auto-install missing plugins on flow save

fdelbrayelle:
## Summary

Adds on-demand plugin acquisition: when a saved flow references a plugin type that isn't installed, Kestra resolves, downloads, and registers the missing artifact automatically — no restart required.

Design background: [KIP-45](https://www.notion.so/kestra-io/KIP-45-Plugin-Management-Decoupling-On-Demand-Acquisition-35736907f7b580bfba09fe7a15e624ce#56270bcd283c4b3ea75c09da1aa48ce2).

- **`PluginCatalogService`**: adds a `group` field (Java package, e.g. `io.kestra.plugin.http`) to `PluginManifest` so FQCN → Maven-artifact resolution works via longest-prefix matching.
- **`PluginAutoInstallService`** (new in `core`): scans a flow YAML for task/trigger types absent from the `PluginRegistry`, resolves their Maven artifact from the catalog, downloads and installs via `PluginManager`, then clears `JsonSchemaCache`. **Gated exclusively to the OSS local persona**: on by default only when `edition == OSS && storage.type == local` (the signal that distinguishes `server local` from a generic standalone deployment on S3/GCS), off everywhere else. An explicit `kestra.plugins.auto-install.enabled` (`true`/`false`) always overrides the computed default, so distributed/EE keep the previous opt-in behaviour. Nothing to change in `kestra-ee` — its `EditionProvider` returns `EE`, so the default resolves to off there automatically.
- **`PluginController`**: new `POST /api/v1/plugins/install` endpoint for explicit artifact installation, plus `PluginManager` injection.
- **`FlowController`**: calls `installMissingPlugins()` before `createFlow()` and `updateFlow()` so missing plugins are transparently installed at save time.

## Architecture

Two independent features, at two different moments:

- **Auto-completion without the JARs** — a pre-baked, per-release *schema bundle* (`plugins-schema.json`, built by the new `plugins-schema` CLI command and published to GCS) is merged into the editor schema on demand (`GET /api/v1/plugins/schemas/{type}?includeCatalog=true`). The YAML editor therefore autocompletes **every plugin in the catalog even when it isn't installed** — no JAR download, just a few MB of JSON.
- **Auto-download / install on save** — missing plugin types are detected, resolved to Maven artifacts, downloaded and registered **in-process (no container restart)** when the flow is saved.

[...]

PR #18743, opened 2026-08-27
fix(core): keep the base definition key when collapsing discriminator wrappers

fdelbrayelle:
Unblocks the EE build (failing `io.kestra.ee.docs.JsonSchemaGeneratorTest`, reported in Slack after https://github.com/kestra-io/kestra/pull/16054 merged).

### Problem

The single-use discriminator-wrapper collapse from #16054 merged the plain `<Class>-1` definition into its `<Class>-2` wrapper and kept the **wrapper's** key — renaming every collapsed subtype definition. Downstream consumers address subtype definitions by the plain key: the EE no-code editors and `io.kestra.ee.docs.JsonSchemaGeneratorTest` (apps, policies, reusable-inputs cases) broke, blocking EE merges.

### Fix

Collapse in the other direction: merge the wrapper's discriminator addition into the plain definition, keep the plain key, and rewrite every `$ref` to the dropped wrapper. Same definition count, same validation semantics, stable names.

A companion EE PR fixes the remaining reusable-inputs failure (EE's `stripReusableInputs` predates the collapse and only handled the inline-`allOf` shape) — that one is EE-side regardless of this change.

### Evidence

- `:core:test --tests JsonSchemaGeneratorTest --tests PluginSchemaBundleServiceTest` green (collapse test updated to pin base-key survival and no dangling wrapper refs).
- `:webserver:test --tests PluginControllerTest` green.
- EE `:core-ee:test --tests io.kestra.ee.docs.JsonSchemaGeneratorTest` green against this branch (with the companion EE fix).

### DIFF
--- a/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
+++ b/core/src/main/java/io/kestra/core/docs/JsonSchemaGenerator.java
@@ -212,8 +212,11 @@ public class JsonSchemaGenerator {
      * {@code <Class>} object, and a {@code <Class>-2} wrapper ({@code allOf: [{$ref: <Class>-1}, {
      * required: ["type"], ...}]}) carrying the discriminator {@code required} needed to use it as
      * an {@code anyOf} branch. When the plain definition is referenced from nowhere else, keeping
-     * it separate only adds an indirection with no reuse benefit — inline it into its wrapper and
-     * drop it, cutting the schema's definition count without changing what it validates.
+     * it separate only adds an indirection with no reuse benefit — merge the wrapper's addition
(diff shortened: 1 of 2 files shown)


fdelbrayelle:
Closing: no longer needed after kestra-io/kestra-ee#10509, which made the EE tests suffix-agnostic and fixed the strip regression EE-side. EE runtime never addressed definitions by key (it matches on properties.type.const), the editor follows $refs, and bundle/local stay aligned since both run the same collapse. Merging this now would only rename the surviving key again (-2 → -1), which would mismatch already-baked bundles against a new runtime and duplicate a few editor stubs during the transition — churn before the RC for no functional gain.

PR #18744, opened 2026-08-27
refactor(plugins): share one shiki highlighter between KsMarkdown and plugin schemas

elevatebart:
### 🔗 Related Issue

No related issue — this came out of a question about why two Shiki highlighters existed.

### ✨ Description

The plugin documentation page built **two** `HighlighterCore` instances. `SchemaToHtml` created one from `ui/src/components/plugins/schema/shikiToolset.ts` (3 grammars: yaml, python, javascript); `KsMarkdown` created another from the design system's `shikiHighlighter.ts` (29 grammars + on-demand loading). `PluginDocumentation.vue` mounts both components, so both were live at once — two registries, two JS regex engines, two compiled grammar and theme sets built from the same modules.

Rolldown's `markdown` chunk group already captured `node_modules/shiki/**`, so this was never a payload problem. It was a duplicated factory and a duplicated runtime instance.

`shikiToolset.ts` is deleted. The design system's highlighter is now exported as `@kestra-io/design-system/shiki` and `SchemaToHtml` uses it, so both consumers share the one module-level singleton. The design-system file had to be the survivor: `KsMarkdown` lives inside the package and cannot import from `ui/src`.

Two things follow from the merge:

- **Plugin schema examples go from 3 grammars to 29.** `example.lang` is plugin-authored, so it was never restricted to the three that were registered — a `json` or `sql` example threw inside a `computed`, during render. The second commit wires in `loadLanguageOnDemand` so anything outside the 29 is fetched from Shiki's full bundle, falling back to plain text when the language is unknown. The synchronous path is unchanged for pre-registered grammars, so nothing renders empty while waiting.
- **`ui` no longer has an undeclared dependency on `shiki`.** It is declared only in `packages/design-system/package.json`; `shikiToolset.ts` was resolving it through workspace hoisting.

Evidence, from clean production builds of this branch and its parent commit:

| | before | after |
|---|---|---|
| `createHighlighterCore` call sites in the bundle | 2 (`markdown`, `flow-markdown`) | 1 (`markdown`) |
| chunk count / chunk-name set | 422 | 422, identical set |
| total assets | 39,689,115 B | 39,689,138 B |

The new `SchemaToCode.spec.ts` cases fail on the parent commit with `Error: Language \`sql\` not found` and `Error: Language \`nonexistent\` not found` — the render-time throw being fixed.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit` — 2047 passed, 217 files)

No screenshot: there is no visual delta. The rendered HTML for existing yaml examples is byte-identical — the highlighter is built from the same grammars and themes either way. The only user-visible change is a case that previously threw during render, which is covered by the tests above rather than by a picture.

### 📝 Additional Notes

- **The bundle does not get smaller.** Total assets are 23 bytes larger, and the chunk graph is unchanged. Anyone reading "collapse two highlighters into one" as a size win will be disappointed; the win is one runtime instance instead of two, plus the language coverage and the dependency hygiene.
- The `markdown` chunk stays lazy and `shiki-langs` (~5 MB, all ~350 grammars) still has no static importer, which `consolidateChunks.js` enforces with a build error. Verified on the rebased base.
- `loadLanguageOnDemand` lives in `SchemaToCode` rather than `SchemaToHtml` because `SchemaToHtml`'s top-level `await` runs once and it does not remount when the plugin changes — the `:key="pluginType"` is on the child, so pre-loading in the parent would go stale.
- Comments in `consolidateChunks.js` that named `shikiToolset.ts` are updated; the chunking logic itself is untouched.

### 🤖 AI Authors

Raised by Claude. Why did the cat sit on the keyboard during the build? It heard there were two highlighters and wanted to make sure only one of them got the byte. 🐱

---
_Generated by [Claude Code](https://claude.ai/code/session_018PUrDMvUhP3SsRTXxGtpaH)_


RemiBarthe (commented):
Really nice cleanup! Two things before merge:

**1. `SchemaToCode.vue:54` — the early-out only matches `"text"`**

Shiki also treats `""`, `"plaintext"`, `"txt"`, `"plain"` and `"ansi"` as no-grammar languages, and none of them are in `getLoadedLanguages()` or in `bundledLanguages` (332 keys — `js`/`yml`/`sql` yes, `txt`/`plaintext`/`plain`/`ansi` no). 
So `lang = "plaintext"` now downloads the ~5 MB `shiki/langs` chunk just to fall back to the plain text the old highlighter rendered for free. `ansi` is worse: it used to come out ANSI-coloured, now it's unstyled after the same download. Gate on Shiki's `isSpecialLang` instead of `=== "text"`.

**2. `shikiHighlighter.ts:115` — a failed `shiki/langs` fetch is cached forever**

The `try`/`catch` wraps only `loadLanguage`; the `await bundledLanguages` on line 117 isn't covered. A 404 on that chunk makes `loadLanguageOnDemand` reject instead of returning `false`, and the rejected promise stays memoized — every later grammar load fails for the rest of the session. From `watchEffect(async …)` it also surfaces as an unhandled rejection rather than the fallback the comment on line 59 promises.

    bundledLanguages ??= import("shiki/langs")
        .then((module) => module.bundledLanguages)
        .catch(() => { bundledLanguages = null; return {} })

elevatebart:
Both confirmed and fixed in ac1f6e7. Thanks — the `ansi` case in particular was a regression I introduced, not just wasted bytes.

I checked your numbers against the installed Shiki before changing anything:

- `isSpecialLang` is exported from **`shiki/core`**, so gating on it costs no bundle weight — no need to reach into the full `shiki` entry.
- 332 bundled languages; `js`/`yml`/`sql` present, `text`/`txt`/`plain`/`plaintext`/`ansi`/`""` all absent.
- `getLoadedLanguages()` on a highlighter built the way `shikiHighlighter.ts` builds it returns `["yaml","yml"]` — real grammars only, never the special ones. So the old `getLoadedLanguages().includes(...)` check missed every one of them.
- On the ANSI point: `codeToHtml(sample, {lang: "ansi"})` with no grammar loaded yields three distinct colours (`#24292e`, `#e1e4e8`, `#ea4a5a`); falling back to `"text"` yields two. The red was genuinely being dropped.

**1.** `SchemaToCode.vue` now gates on `isSpecialLang(requested)`, re-exported from `@kestra-io/design-system/shiki` so the app still has no direct `shiki` import.

**2.** The bundle fetch now catches and clears the memo:

```ts
bundledLanguages ??= import("shiki/langs")
    .then((module) => module.bundledLanguages)
    .catch(() => {
        bundledLanguages = null
        return {}
    })
```

**One thing worth flagging, since it reaches beyond the two lines you pointed at.** I put the special-language short-circuit inside `loadLanguageOnDemand` itself rather than only at the `SchemaToCode` call site, because `KsMarkdown.vue` has the identical `getLoadedLanguages().includes(lang)` gate and so had the same 5 MB fetch. The visible consequence: a ` ```ansi ` fence in markdown now renders ANSI-coloured instead of plain. I read that as the same fix rather than a new behaviour, but say the word and I'll narrow it to `SchemaToCode` and raise KsMarkdown separately.

Tests, both verified failing on the parent commit:

- `shikiHighlighter.test.ts` (new) — a failed bundle fetch resolves `false` instead of rejecting, and a later call retries; a no-grammar language never touches the bundle. Pre-fix both fail with `promise rejected ... instead of resolving`.
- `SchemaToCode.spec.ts` — a no-grammar language passes straight through without calling `loadLanguageOnDemand`.

Green locally: `check:types`, app unit 2048 passed, design-system unit 817 passed, `lint` 0 errors, `npm run build` clean with still 1 `createHighlighterCore` call site and 0 static edges into `shiki-langs`.

---

RemiBarthe (approved):
LGTM 🚀

elevatebart:
`kestra-ee/backend-tests` went red on ac1f6e7. Pushed 062644b (a merge of `develop`) to fix it — flagging it since it lands after the approval, and it is not a code change.

The failure was cross-repo staleness, not anything in the diff. OSS #18741 (`41f39db`, *"make the input-type exclusion a per-schema seam"*) replaced the seam that `core-ee`'s `JsonSchemaGenerator` extends, and EE `develop` was updated in lockstep to override the new `excludedInputTypes(Class<?>)`. This branch predated `41f39db`, so that override matched nothing on the OSS side:

```
core-ee/src/main/java/io/kestra/ee/docs/JsonSchemaGenerator.java:89: error: method does not override or implement a method from a supertype
2 errors
```

Line 89 is the `@Override` above `excludedInputTypes`. Any OSS head older than `41f39db` hits this; it was already showing up as the "Failed to generate EE OpenAPI spec" section in the bot comment above, before it reached the test job.

After the merge all three EE overrides resolve against `core/JsonSchemaGenerator`:

| EE override | OSS supertype |
|---|---|
| `includeInputSubtype(Class<?>)` | line 1093 |
| `schemas(Class<? extends T>, boolean)` | line 116 |
| `excludedInputTypes(Class<?>)` | line 181 ← added by `41f39db` |

Merged rather than rebased, to avoid rewriting the `develop` merge already on this branch and dropping the approval.

Re-validated after the merge: `check:types` clean, app unit 2057 passed, design-system unit 817 passed, `lint` 0 errors, `npm run build` exit 0.

---

PR #18745, opened 2026-08-27
fix(triggers): smooth the enabled toggle animation when switching a trigger off

MilosPaunovic:
Closes https://github.com/kestra-io/kestra-ee/issues/10279.

### ✨ Description

On the Triggers Manage tab, switching a trigger's **Enabled** toggle off made the knob bulge slightly to the right, snap back, hesitate, and only then slide off. Two causes, two fixes:

- **`KsSwitch` press-stretch desync**: the knob's `left` transition ran at 0.28s with an overshoot bezier while `width` ran at 0.18s ease. During the pressed stretch of a checked switch the two animate out of sync, so the knob's outer edge bulges right and snaps back instead of staying pinned. Both properties now share the same duration and easing, which keeps the outer edge exactly pinned throughout the stretch (`left` and `width` change by opposite amounts, so any shared easing cancels out).
- **Row state only changed after the API roundtrip**: the switch is prop-controlled, so the knob only slid off once `set-disabled` returned - a full roundtrip after the click. `TriggersManage` now flips the row optimistically before the call and reverts it if the call fails. The enable dialog flow is unchanged: it still opens before any state change, and cancelling it leaves the knob where it was.

The stutter is an animation-timing bug, so stills cannot show it; instead it was verified against a running EE backend with in-page instrumentation (MutationObserver on the switch class + resource timing):

| | knob starts moving | transitions (`left`/`width`) |
|---|---|---|
| before | ~672 ms after the click (43 ms after the 629 ms API response) | 0.28s bezier / 0.18s ease (desynced) |
| after | same millisecond the request starts, i.e. immediately on click | 0.28s / 0.28s, same bezier |

Also re-checked on the patched build: cancelling the enable dialog leaves the knob off, and confirming Enable flips it with the server state matching afterwards.

### 🎨 Frontend Checklist

- [x] Unit tests pass (KsSwitch unit + Storybook tests, eslint on the changed files)
- [x] Evidence attached (measured timings above; the stutter is not capturable as a still image)


Pradumnasaraf (approved):
Looks good!

PR #18748, opened 2026-08-27
refactor(ui): consolidate useTableColumns into design system

### LINKED ISSUE #18709: Two copies of `useTableColumns` have to be fixed in lockstep
There are two `useTableColumns` composables in OSS, and they differ only by a `minWidth?: number` field:

| File | Consumed by |
|---|---|
| `ui/packages/design-system/src/components/Data/KsDataTable/filter/composables/useTableColumns.ts` | `DraggableTableColumns.vue` — the switches inside the *Customize table columns* dropdown |
| `ui/src/composables/useTableColumns.ts` | every feature table: `Flows`, `Executions`, `KVTable`, `SecretsTable`, `TriggersManage`, `FlowTriggers`, `MetricsTable`, `McpToolFlows` |

Because the dropdown reads one copy and the table renders from the other, a fix applied to only one of them makes the two disagree rather than just leaving the bug in place. https://github.com/kestra-io/kestra/pull/18531 hit exactly that: the first revision fixed the design-system copy alone, which would have left the dropdown showing every switch **off** while the table rendered **all** columns — a contradiction that did not exist before, since previously both fell back to the defaults and at least agreed. It was resolved by landing the same eight-line block in both copies, so that block and its comment now live in two places.

That duplication is the defect. Consolidating on a single composable would remove the class of bug rather than this one instance.

Worth deciding as part of it:

- Which copy survives. The design-system one is the natural home, but it would need to carry `minWidth?: number` for the app-level callers.
- `SecretsTable.vue` reads `columns_${storageKey}` from `localStorage` by hand instead of going through either composable (tracked separately in https://github.com/kestra-io/kestra/issues/18543) — it should consume the surviving composable in the same pass.
- `ui/packages/design-system/tests/units/Data/useTableColumns.test.ts` declares every column with `default: true`, so its two "fall back to the defaults" assertions cannot distinguish the `default`-flag branch from "return every column". `ui/tests/unit/composables/useTableColumns.spec.ts` gets this right by making one column `default: false`. Whichever spec survives should follow the app-level one.

Context: https://github.com/kestra-io/kestra/pull/18531#pullrequestreview-5017069337

hemantmm:
closes: #18709

---

### 🔗 Related Issue

Follow-up to #18531

### ✨ Description

Consolidated the two duplicate implementations of `useTableColumns` into the design-system package (`@kestra-io/design-system`) to eliminate drift and prevent configuration discrepancies between the column customizer dropdown and feature tables.

**Changes:**

1. Design System Enhancement:

- Exported `useTableColumns`, `ColumnConfig`, and `UseTableColumnsOptions` from `ui/packages/design-system/src/index.ts`.
- Added `minWidth?: number` to `ColumnConfig` in the design system to satisfy all feature table requirements.

2. App Refactoring:

- Updated imports across all feature tables (`Flows`, `Executions`, `KVTable`, `TriggersManage`, `FlowTriggers`, `MetricsTable`, and `McpToolFlows`) to consume `useTableColumns` from `@kestra-io/design-system`.
- Removed duplicate `ui/src/composables/useTableColumns.ts`.


flcarre (changes requested):
Thanks for taking this on — the duplication was real, and consolidating on the design-system copy is the right direction. Two things need to be resolved before this can merge, and one is outside the OSS repo so it is easy to miss.

### 1. Deleting `ui/src/composables/useTableColumns.ts` breaks `ui-ee`

`ui-ee` npm-workspaces the OSS UI and imports this composable by that exact path:

- `ui-ee/src/components/users/Users.vue:174`
- `ui-ee/src/components/groups/GroupMembers.vue:156`
- `ui-ee/src/components/invitations/Invitations.vue:168`
- `ui-ee/src/components/bindings/Bindings.vue:268`
- `ui-ee/tests/unit/components/Bindings.spec.ts:8` — `vi.mock("kestra/src/composables/useTableColumns", …)`

OSS CI cannot see this, so it stays green while the EE frontend fails to type-check and build. Either option works for us:

- keep `ui/src/composables/useTableColumns.ts` as a one-line re-export of the design-system composable, so the EE import paths keep resolving; or
- leave the file deleted and we land a paired EE PR migrating those five call sites at the same time. Say the word and we will prepare it — it just has to merge in lockstep.

The re-export shim is the lower-risk option for a first pass, and the EE call sites can be migrated separately afterwards.

### 2. The test file changes break two tests that currently pass

The branch is stale, so the diff GitHub renders is misleading: the whole `describe("useTableColumns persistence", …)` block, with all four of its tests, already exists on `develop` (added in #18531). The only real delta the diff makes to that file is this one line:

```diff
-    {label: "C", prop: "c", default: true},
+    {label: "C", prop: "c", default: false},
```

With `c.default: false`, the fallback branch in `initializeVisibleColumns` returns `["a", "b"]`:

```ts
visibleColumns.value = initialVisibleColumns.length
    ? initialVisibleColumns
    : columns.filter(c => c.default && (!c.condition || c.condition())).map(c => c.prop)
```

Both of these call `setup(key, [])`, so `initialVisibleColumns` is empty and that branch is taken, and both assert `["a", "b", "c"]`:

- `should fall back to the default-flagged columns when no selection was ever stored`
- `should fall back to the default-flagged columns when the stored columns no longer exist`

They pass on `develop` and fail here. Please revert that line and rebase on `develop` — the diff should then be limited to what the PR is actually about: the `minWidth` field, the `index.ts` exports, and the import rewrites.

Worth flagging that no CI has run on this PR yet (only the `welcome` job), so nothing has been validated automatically.

### Nit

The eight import rewrites are all good, and `reorderColumns` was the only export lost in the move — it had no call sites, so nothing regresses there. `setColumnOrder` in the design-system copy already covers `DraggableTableColumns.vue`.

Once 1 and 2 are sorted this is a clean consolidation. Happy to re-review quickly.

hemantmm:
Thanks for catching the `ui-ee` dependency! I went with the lower-risk option and added `ui/src/composables/useTableColumns.ts` back as a one-line re-export so the EE imports keep resolving cleanly. 

I've also reverted the test file change and rebased onto `develop`. The tests are passing locally now, and the diff is strictly limited to the `minWidth` addition, the exports, and the import rewrites.

flcarre (commented):
Re-reviewed against the current head. **All four items from the previous round are resolved** — thanks for working through them:

- The EE breakage is fixed by the re-export shim in `ui/src/composables/useTableColumns.ts`, so kestra-ee's direct-path imports keep resolving (there are 16 of them, more than the earlier review counted).
- The `default: true → false` regression is gone.
- The branch was rebased past https://github.com/kestra-io/kestra/pull/18531, so the misleading diff has cleared.
- CI is green, 8 passed / 0 failed.

I diffed the deleted implementation against the design-system one line by line and it is behaviourally identical: same `orderStorageKey` (`ks-column-order-v2-${storageKey}`), same `visibilityStorageKey` (`columns_${storageKey}`), same pre-v2 key cleanup, same `useLocalStorage` serializer and `writeDefaults: false`, every branch of `initializeVisibleColumns` unchanged, same `toggleColumn` insertion maths. **No storage key changed, so nobody loses their saved column preferences** — that was the main risk in a move like this and it's clean. The only delta is `reorderColumns(from, to)` → `setColumnOrder(order)`, and the placement under `KsDataTable/filter/composables/` matches `useRouteFilterPolicy` and `useDismissedKeys`. `ui/AGENTS.md` already listed `useTableColumns` as a design-system filter composable, so the docs were ahead of the code and this makes them true.

Two things before merge.

## 1. Two behaviours lose their only test

The deleted `ui/tests/unit/composables/useTableColumns.spec.ts` had 7 tests; the existing design-system spec covers 5 of them. These two have no equivalent anywhere after the move:

- **`should keep the stored columns that still exist and drop the rest`** — the only coverage of the partial-match branch in `initializeVisibleColumns` (stored `"gone,a,c"` → `["a","c"]`). The DS spec only exercises the all-invalid case.
- **`should prefer explicit initialVisibleColumns over the default flags`** — the DS fixture declares all three columns `default: true` and its `setup` defaults `initialVisibleColumns` to all three, so the two branches are indistinguishable in every DS test.

Both port over as-is against the existing fixture:

```ts
test("keeps the stored columns that still exist and drops the rest", () => {
    localStorage.setItem("columns_partly-stale", "gone,a,c")
    const table = setup("partly-stale", [])
    expect(table.visibleColumns.value).toEqual(["a", "c"])
    expect(table.visibleCount.value).toBe(2)
    expect(table.totalCount.value).toBe(COLUMNS.length)
})

test("prefers explicit initialVisibleColumns over the default flags", () => {
    const table = setup("explicit", ["c"])
    expect(table.visibleColumns.value).toEqual(["c"])
})
```

## 2. The shim is invisible dead code from OSS's point of view

`ui/src/composables/useTableColumns.ts` now has **zero** consumers inside OSS — it exists solely for kestra-ee's 16 direct-path imports, and OSS CI cannot see them. The next person grepping for unused files deletes it and silently breaks the EE frontend build.

This is exactly the "deliberate omission, so the next person does not fix it back" case our guidelines reserve a comment for:

```ts
// Kept for kestra-ee, which imports this path directly; remove only with the paired EE migration.
export {useTableColumns, type ColumnConfig, type UseTableColumnsOptions} from "@kestra-io/design-system"
```

Worth running `npm run check:types` in `ui-ee` once before merge, since a re-export through the design-system barrel is the same shape as the `vue-tsc` trap we hit in https://github.com/kestra-io/kestra/pull/17627 — I don't expect it to fire here (these are ordinary components, not `App.vue`), but it's a cheap check.

hemantmm:
Thanks for the detailed review! I've just pushed a commit that addresses both of your points:

1. The two missing tests (`keeps the stored columns that still exist and drops the rest` and `prefers explicit initialVisibleColumns over the default flags`) have been added to the design-system spec to restore full coverage.
2. I've added the explanatory comment to the top of the `ui/src/composables/useTableColumns.ts` shim to ensure it isn't accidentally swept up in future dead-code cleanups before the EE migration happens.

flcarre (changes requested):
Formalising the gate from my detailed review above. The consolidation itself is correct — I verified the composable is behaviourally identical to the deleted one, storage keys included, so no user loses their saved column preferences — and all four items from the previous round are resolved. Two things remain before this can merge:

**1. Two behaviours lose their only test.** The deleted `ui/tests/unit/composables/useTableColumns.spec.ts` had 7 tests; the design-system spec covers 5. The partial-match branch of `initializeVisibleColumns` (stored `"gone,a,c"` → `["a","c"]`) and `initialVisibleColumns` taking precedence over the `default` flags now have no coverage anywhere — the DS fixture marks every column `default: true`, so those branches are indistinguishable there. Both port over as-is against the existing fixture; the snippets are in my comment above.

**2. The shim needs a comment saying why it exists.** `ui/src/composables/useTableColumns.ts` has zero consumers inside OSS and exists only for kestra-ee's 16 direct-path imports, which OSS CI cannot see. Without a line explaining that, the next cleanup pass deletes it and silently breaks the EE frontend build.

Both are small. Happy to re-review as soon as they're in.

hemantmm:
I've already pushed a commit that addresses both of these!

- The two missing tests were ported over to the design-system spec.
- The `ui/src/composables/useTableColumns.ts` shim now has the comment explaining why it's kept for `kestra-ee`.

flcarre (approved):
Approving. This is a clean consolidation and both rounds of requested changes are resolved in the current head.

Verified against the current commit:

- The re-export shim at `ui/src/composables/useTableColumns.ts` now carries the comment explaining it exists only for kestra-ee's direct-path imports, so it survives the next dead-code sweep.
- Both previously-missing test cases are ported into `ui/packages/design-system/tests/units/Data/useTableColumns.test.ts`: the partial-match branch of `initializeVisibleColumns` (`"gone,a,c"` to `["a","c"]`), and `initialVisibleColumns` taking precedence over the `default` flags, the latter now actually distinguishable since the explicit `["c"]` differs from the fallback `["a","b","c"]`.
- The `minWidth?: number` field added to the DS `ColumnConfig` is genuinely consumed (`FlowTriggers.vue` reads `col.minWidth`), not a speculative addition.
- `reorderColumns`, the one export dropped in the move, has no callers anywhere in OSS outside the file being deleted, confirmed by a full-repo grep. Nothing regresses.
- No element-plus imports, hex colors, `:deep()` or raw px introduced; the diff is otherwise pure import-path rewrites.

Two non-blocking nits: a stray blank line left in the `MetricsTable.vue` import block, and the shim file still has no trailing newline (pre-existing, not introduced here). Neither needs a follow-up commit.

One thing to line up before this lands: the paired EE migration of the direct-path imports needs its own tracked follow-up, otherwise the shim becomes permanent.

hemantmm:
Thanks for the approval! Even though they were non-blocking, I pushed a quick commit to fix the stray blank line and the trailing newline just to keep things completely clean.

MilosPaunovic:
Is this ready to be merged @flcarre?

PR #18750, opened 2026-08-27
fix(flows): validate input defaults, trigger inputs, webhook key and namespace at save

### LINKED ISSUE #18333: [Bash2.0] Flow input defaults are not validated against the input's own type or allowed values
## Describe the issue

A flow input's `defaults` value is not checked against that input's own type and constraints when the flow is saved or validated. You can define a `SELECT` input whose default is not one of its allowed values, or an `INT` input whose default is not a number, and the flow saves successfully and is reported as **valid** by the editor. The problem only surfaces later, at execution time, when running the flow with that default fails with a `422`.

This gives false confidence: the flow editor shows a green/valid state for a flow that is guaranteed to fail whenever it runs with its own defaults.

## Steps to reproduce

In the UI, create/save this flow — the editor reports it as **valid** and saving succeeds:

```yaml
id: demo
namespace: company.team
inputs:
  - id: color
    type: SELECT
    values: [red, green, blue]
    defaults: purple        # not one of the allowed values
tasks:
  - id: r
    format: "{{ inputs.color }}"
```

Then click **Execute** and run it without changing `color` (using the default).

## Actual behaviour

The flow is created (`200`) and `/flows/validate` returns no constraints and no warnings, so the editor marks it valid. Executing it with the default fails with `422 Invalid value for input 'color'. Cause: it must match the values [red, green, blue]`. An `INT` input with `defaults: "notanumber"` is likewise accepted and validated as OK.

## Expected behaviour

Flow validation (save and the `/flows/validate` endpoint) should check that a literal input default satisfies that input's own type and constraints, and reject or flag it — so the failure is caught at edit time, not deferred to a `422` at execution.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise (shared core flow validation)
- Reproduced via: `/flows/validate` and `/flows` vs `/executions/{namespace}/{flow}` on `http://localhost:8080`

### LINKED ISSUE #18472: [Bash2.0] Creating a flow with a digits-only namespace shows a 500 with a raw internal error

nkwiatkowski:
### What this PR does

Bug-bash batch: bad flow / input / trigger configuration was accepted at save time (or at request time) and only surfaced later at runtime, or as a raw `500`. This tightens validation at the point of entry so the failure is caught where it is authored.

Six fixes, one theme (reject bad input up front):

- **Input defaults not validated** (kestra#18333 / kestra-ee#10259) — a literal `defaults` is now checked against the input's own type and allowed values at save. A `SELECT` default outside `values`, or an `INT` default of `notanumber`, is rejected instead of validating green and failing on first run.
- **Trigger inputs not type-checked** (kestra-ee#10298) — the values a Schedule/Webhook trigger supplies to an input are validated against the target input's type at save (e.g. `"abc"` for an `INT`), so a typo no longer silently stops a schedule.
- **Optional SELECT accepted out-of-list value** (kestra-ee#10258) — the values check was gated on `required: true`; an optional `SELECT` now rejects a value outside its list too (`allowCustomValue` still bypasses it).
- **Webhook blank key** (kestra-ee#10272) — `key` is now `@NotBlank`; a blank/whitespace key (which produced an un-triggerable `403` webhook) is rejected at save.
- **Digits-only namespace → 500** (kestra#18472) — a numeric namespace read from YAML no longer throws a raw `Integer cannot be cast to String`; it is coerced to its string form.
- **Malformed request body leaked internals** (kestra-ee#10273) — a scalar body sent to an endpoint expecting an object returned `422` labelled `Internal server error` and leaked the DTO class name. It now returns a clean `422 Invalid request body`; the raw cause is kept in the server debug log only.

The save-time literal check covers scalar and `SELECT` inputs; infrastructure-backed and structured types (`FILE`, `SECRET`, `JSON`/`ION`/`YAML`, `URI`, `ARRAY`, `MULTISELECT`) are skipped because their literal cannot be resolved without execution-time machinery, and Pebble expressions are left for runtime.

All fixes are in OSS core/webserver; EE inherits them (its `FlowValidator` only overrides the EE-specific hooks).

### Evidence

- `FlowValidationTest` — SELECT default out-of-list, INT default `notanumber`, expression default (skipped), trigger INT input `abc`, webhook blank key.
- `SelectInputTest` — optional SELECT rejects out-of-list.
- `FlowParsingServiceTest` — digits-only namespace parses to `"132312312"`.
- `ErrorControllerTest` — scalar body → clean `422`, no `ApiTriggerId` / `Internal server error` in the response.

Closes https://github.com/kestra-io/kestra/issues/18333.
Closes https://github.com/kestra-io/kestra/issues/18472.

### EE issues (same fixes, cross-repo — close manually on merge)

GitHub only auto-closes issues in this repo, so these are linked for reference and must be closed by hand once this merges:

- https://github.com/kestra-io/kestra-ee/issues/10259 (dup of #18333)
- https://github.com/kestra-io/kestra-ee/issues/10258
- https://github.com/kestra-io/kestra-ee/issues/10298
- https://github.com/kestra-io/kestra-ee/issues/10272
- https://github.com/kestra-io/kestra-ee/issues/10273

### DIFF
--- a/core/src/main/java/io/kestra/core/validations/validator/FlowValidator.java
+++ b/core/src/main/java/io/kestra/core/validations/validator/FlowValidator.java
@@ -9,6 +9,7 @@ import java.util.stream.Stream;
 import io.kestra.core.models.flows.Data;
 import io.kestra.core.models.flows.Flow;
 import io.kestra.core.models.flows.Input;
+import io.kestra.core.models.flows.Type;
 import io.kestra.core.models.flows.input.EeOnly;
 import io.kestra.core.models.flows.input.FormInput;
 import io.kestra.core.models.tasks.ExecutableTask;
@@ -16,6 +17,8 @@ import io.kestra.core.models.tasks.Task;
 import io.kestra.core.models.triggers.AbstractTrigger;
 import io.kestra.core.services.NamespaceService;
 import io.kestra.core.utils.ListUtils;
+import io.kestra.core.utils.PebbleUtil;
+import io.kestra.core.utils.TypeConverter;
 import io.kestra.core.validations.FlowValidation;
 import io.kestra.plugin.core.trigger.AbstractWebhookTrigger;
 import io.kestra.plugin.core.trigger.Schedule;
@@ -27,6 +30,8 @@ import io.micronaut.validation.validator.constraints.ConstraintValidator;
 import io.micronaut.validation.validator.constraints.ConstraintValidatorContext;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
+import jakarta.validation.ConstraintViolation;
+import jakarta.validation.ConstraintViolationException;
 
 import static io.kestra.core.models.Label.READ_ONLY;
 import static io.kestra.core.models.Label.SYSTEM_PREFIX;
@@ -113,6 +118,8 @@ public class FlowValidator implements ConstraintValidator<FlowValidation, Flow> 
         }
 
         findMissingInputsForTriggers(value).forEach(violations::add);
+
+        validateDeclaredInputValues(value, violations);
 
         // system labels
         ListUtils.emptyOnNull(value.getLabels()).stream()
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/kestra/core/validations/validator/FlowValidator.java
```diff
@@ -314,6 +321,103 @@ private Stream<String> findMissingInputsForTriggers(Flow value) {
...
+            case INT -> TypeConverter.toInteger(rawValue);
+            case FLOAT -> TypeConverter.toFloat(rawValue);
+            case BOOL -> TypeConverter.toBoolean(rawValue);
+            case DATETIME -> TypeConverter.toInstant(rawValue);
+            case DATE -> TypeConverter.toLocalDate(rawValue);
+            case TIME -> TypeConverter.toLocalTime(rawValue);
+            case DURATION -> TypeConverter.toDuration(rawValue);
+            case FILE, URI, SECRET, JSON, ION, YAML, ARRAY, MULTISELECT, FORM, REUSABLE_INPUTS -> null;
+        };
+    }
```
loicmathieu: This is duplicated with what we already have in FlowInputOutput, reuse that

webserver/src/main/java/io/kestra/webserver/controllers/ErrorController.java
```diff
@@ -90,7 +90,10 @@ public HttpResponse<JsonError> error(HttpRequest<?> request, ConversionErrorExce
             }
         }
 
-        return jsonError(request, e, HttpStatus.UNPROCESSABLE_ENTITY, "Internal server error");
+        // A conversion error is always caused by bad client input, so it surfaces as a clean client error and the
+        // raw exception message is not sent back to the caller: it carries the target DTO's internal class name.
+        log.debug("Unable to convert request input", e);
```
loicmathieu: there is already a debug log inside the `jsonError` method, it would then be duplicated

loicmathieu (changes requested):

loicmathieu (approved):
LGTM

PR #18751, opened 2026-08-27
fix(kv): preserve plain text values with spaces

### LINKED ISSUE #18335: [Bash2.0] KV values with spaces are truncated at the first space when saved via the API
## Describe the issue

Saving a KV value that contains a space through the API keeps only the first word. The save returns success, but everything from the first space onward is silently dropped, so any multi-word value (a name, a sentence, a path with spaces) loses data.

The web UI and the `Set` KV task are fine because they wrap the value in quotes. It's the plain API (and the `kestra namespace kv update` command without `--type STRING`) that drops the text.

## Steps to reproduce

Save a two-word value, then read it back:

```bash
$ curl -u user:pass -H 'Content-Type: text/plain' \
    -X PUT 'http://localhost:8080/api/v1/main/namespaces/demo/kv/greeting' \
    --data-binary 'hello world'

$ curl -u user:pass 'http://localhost:8080/api/v1/main/namespaces/demo/kv/greeting'
{"type":"STRING","value":"hello"}          # expected "hello world"
```

## Actual behaviour

Only the first word is stored, and the save still reports success:

| Value sent | Value stored |
|---|---|
| `hello world` | `hello` |
| `John Doe Smith` | `John` |
| `a description with several words` | `a` |
| `"hello world"` (quoted) | `hello world` ✓ |

## Expected behaviour

The value should be stored exactly as sent: `hello world` in, `hello world` out. Nothing should be dropped.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise

NimeshXTV:
All PRs submitted by external contributors must follow this template (proper description, related issue, and checklist sections). If you plan to work on a specific issue, comment on the issue first and wait to be assigned before starting any actual work - this avoids duplicated work and ensures a smooth contribution process. PRs that skip either rule **may be automatically closed**.

---
### 🔗 Related Issue

Closes #18335

### ✨ Description

Fix KV values with spaces being truncated when saved through the API.

The parser could read the first part of the value as an Ion value and ignore the rest. This change checks for trailing content and falls back to storing the value as plain text.

Added tests for values containing spaces.

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched modules

### 📝 Additional Notes

None.

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/KVController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/KVController.java
@@ -4,6 +4,8 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.*;
 
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 
@@ -80,7 +82,7 @@ public class KVController {
     );
 
     private String sortMapper(String key) {
-        return key == null ? null : SORT_FIELDS.get(key);
+	return key == null ? null :SORT_FIELDS.get(key);
     }
 
     @ExecuteOn(TaskExecutors.IO)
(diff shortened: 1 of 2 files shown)


NimeshXTV:
Hi @fhussonnois  ! Whenever you have time, could you please take a look at this PR?

PR #18754, opened 2026-08-27
fix(ui): clear the build's INEFFECTIVE_DYNAMIC_IMPORT warnings

elevatebart:
### 🔗 Related Issue

None — from a `npm run build` log. **15 → 0** `INEFFECTIVE_DYNAMIC_IMPORT` warnings, fixed at source, nothing suppressed.

### ✨ Description

Three unrelated causes:

1. **`KsBulkSelect` locale** — the only design-system component reading its locale via a static import into a local i18n scope, while `i18n/index.ts` globs and registers every `*.locale.ts` globally. Now namespaced keys through the global scope, like its seven siblings. Side benefit: its un-namespaced `selected`/`all` keys no longer merge over the app's own (`en.json` has `"selected": "Selected"`).
2. **No-code block cards** — `BranchLane` declared `FlowableClusterCard`/`LeafBlockCard` async, but both are statically imported elsewhere, so nothing could split. `LeafBlockCard` has no cycle at all; `FlowableClusterCard` is mutually recursive with `BranchLane`, and only one side needs to be lazy — that side is unchanged, now commented.
3. **SDK root barrel** — the root entry is a module-federation share, so MF pins every name it exports for remotes; re-exporting the operations there held all per-tag modules in the initial graph. Root now exports the client facade + types only; the barrel moves to a new `@kestra-io/kestra-sdk/all`. `./shared` is `eager` (tenant state every operation imports statically).

**Breaking:** no operations on the SDK root — use its tag subpath or `/all`. Types unaffected; package is private and unpublished.

### 🎨 Frontend Checklist

- [x] `npm run check:types`
- [x] `npm run build` — and 0 warnings of any category
- [x] `npm run test:unit` — 218 files, 2065 tests
- [x] `npm run translations:check` — clean
- [ ] Screenshots — **no visual change**; `KsBulkSelect`'s strings are identical, asserted in `en` and `fr`

### 📝 Additional Notes

- Suppressing the warnings via `onLog` was tried and thrown away.
- `eager` on all SDK subpaths also reaches 0 warnings in two lines, but keeps the per-tag modules in the initial graph; the entry-point split is worth the ~228 KB.
- `packages/design-system/tests/units/Data/KsEmptyState.test.ts` fails on pristine `develop` too — pre-existing, untouched.
- `6363a65` fixes an `es` string the translations job generated as `**{count}**` instead of `<strong>` (rendered via `v-html`).

### 🤖 AI Authors

Why wouldn't the barrel tree-shake? It had committed to all 1220 exports — even a cat with nine lives would call that overcommitting. 🐱

## Related PRs

- [kestra-io/kestra-ee#10528](https://github.com/kestra-io/kestra-ee/pull/10528) — EE half; must land with this one
- [kestra-io/client-sdk#396](https://github.com/kestra-io/client-sdk/pull/396) — same split on the published SDK

https://claude.ai/code/session_01UvrhcR8x2kPEDVaf1E6XJS


elevatebart:
`Frontend - Tests` (storybook `Executions > Bigger Data`) was mine — moving the operations off the SDK root made a page's first subpath import a new dep mid-run, and the re-optimize reload tore the story down; `f2f9df4` excludes the subpaths from dep optimization (full suite: 67 files / 248 tests green locally).

`trigger-ee` is unrelated: the Kestra webhook returned HTTP 502 Bad Gateway, which the action itself treats as infra rather than a compile verdict — this push re-runs it.

---

elevatebart:
My last diagnosis was wrong. The real error is `No match for {name: "namespaces/update"}` — this story's router never declared that route, but its namespace column renders a `KsEntityLink` to it, so `RouterLink` setup threw and cascaded into the unmount TypeError; `c802429` declares the route and reverts the optimizeDeps exclusion, which only shifted the timing of the race that decides whether the row renders at all.

---

AcevedoR (approved):

PR #18756, opened 2026-08-27
fix(ui): blank table when every column is hidden in KV Store and Metrics

### LINKED ISSUE #18579: [UI] Hiding all columns in KV Store table causes broken layout and floating cell icons
### Describe the issue

Hiding all columns in the KV Store table causes the table headers and data columns to disappear while row-level interactive elements, such as checkboxes and edit/delete actions, remain visible and become misaligned/floating in the empty table area.

**Steps to Reproduce**

1. Navigate to KV Store at /ui/main/kv.
2. Open the Customize table columns panel.
3. Toggle off each visible column one by one.
4. Continue until the panel shows 0 of 5 columns visible.
5. Close the panel, or leave it open and inspect the underlying table.

**Actual Behavior:**

Table headers disappear, but row-level interactive elements (checkboxes and action icons like edit/delete) remain rendered and float in empty space without grid alignment.

**Expected Behavior:**

1. Either the table customization should enforce a minimum of 1 visible column (e.g., disabling the toggle for the last remaining column or making the primary identifier like `Key`/`Namespace` non-hideable).

2. Or, if 0 columns are permitted, the table should display a graceful empty/fallback state instead of fragmented floating elements.

https://github.com/user-attachments/assets/05d7d0e2-2415-4625-8d2e-f644a001ba89

### Environment

- Kestra Version: develop

Nishant891:
All PRs submitted by external contributors must follow this template (proper description, related issue, and checklist sections). If you plan to work on a specific issue, comment on the issue first and wait to be assigned before starting any actual work - this avoids duplicated work and ensures a smooth contribution process. PRs that skip either rule **may be automatically closed**.

---

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18579.

### ✨ Description

Hiding all columns in the KV Store table/Metrics section causes the table headers and data columns to disappear while row-level interactive elements, such as checkboxes and edit/delete actions, remain visible and become misaligned/floating in the empty table area.

Both KVTable.vue and MetricsTable.vue define all their data columns inside optionalColumns, making the 0-column state possible.

Th fix include addition of hasVisibleColumns state computed from the length of the orderedVisibleColumns and displayColumns in both the components. If the number of selected columns is 0 it gracefully falls back and displays the "no_data" fingerprint. 

### 🎨 Frontend Checklist

- [ ] Type checking passes (`npm run check:types`)
- [ ] Unit tests pass (`npm run test:unit`)
- [ ] Translations are complete if `en.json` changed (`npm run translations:check` reports no missing, extra, or stale keys)

### 📝 Additional Notes

An additional fingerprint can be added to display a fallback specifically for this case.


MilosPaunovic:
@Nishant891, could you share a screenshots of befrore/after or a quick screen recording showing the current behavior?

MilosPaunovic:
Any updates on this @Nishant891?

MilosPaunovic:
Thanks @Nishant891 - the approach here is the right one. Hiding every column is a deliberate, persisted choice as of https://github.com/kestra-io/kestra/pull/18531, so forcing a minimum of one visible column would have regressed that; falling back to an empty state is the correct half of the two options in the issue.

I pushed a follow-up commit rather than sending you round another review cycle:

- The generic `no_data` message told the user to adjust their filters, which are untouched in this state, and its second line promised the rows would turn up on their own. Replaced with a dedicated `no_results.all_columns_hidden` pair that names the real cause and points at `Customize table columns`, generated across all 12 locales.
- `:total` still carried the real row count, so the pagination footer read `Total 24` underneath an empty table.
- `MetricsTable` fell through to its parent's `empty` slot. `KsDataTable` replaces its entire wrapper when that slot exists and `data` is empty, and the `execution-metrics` storage key is shared with the `MetricsTable` in `LowCodeEditor.vue`, so hiding the columns on the `execution` Metrics tab made the topology task-detail panel report no metrics at all.
- `KsDataTable` gained a `noDataDescription` prop so the empty state can carry that second line.

Also merged `develop` in to clear the conflict.

`check:types`, `eslint` and both unit suites (`design-system` 822, app 2171) are green locally. Screenshots captured against a mocked `API`:

**`develop` - the reported bug**

before

**This `PR` as submitted**

as submitted

**After**

after

MilosPaunovic (approved):
Approach confirmed as the right one, and the follow-up commit addresses the empty-state copy, the stale pagination total and the `MetricsTable` `empty`-slot fallthrough. Thanks for the fix, @Nishant891.

Nishant891:
@MilosPaunovic Sorry was away so couldn't respond. Thanks for the detailed review. I am glad I was able to contribute...

PR #18761, opened 2026-08-27
fix(flows): remove the empty gap between the Context panel sections

### LINKED ISSUE #18495: [Bash2.0] Flow editor Context panel: large empty gap between the KV Pair and Secrets sections
## Describe the issue

In the flow editor Context panel, the sections (Variables, KV Pair, Secrets) are spread apart with a large empty vertical gap between them. Even when the sections are collapsed, KV Pair sits at the top and Secrets is pushed to the very bottom of the panel, leaving a big blank space in the middle that you have to scroll through.

## Steps to reproduce

1. Open a flow in the editor and open the Context tab (the panel on the right).
2. Look at the Variables / KV Pair / Secrets sections, collapsed or expanded.

## Visuals

Image

## Environment

- Kestra Version: 2.0.0-rc10
- Edition: Enterprise

arikaran-rajamanickam:
Closes https://github.com/kestra-io/kestra/issues/18495.

## Problem

In the flow editor `Context` panel, the `Variables`, `KV Pair` and `Secrets` sections were spread apart by a large empty gap, and the gap stayed there when the sections were collapsed.

`app.scss` stretches a listing section to full height when its data table renders an empty state, so the empty state fills the page instead of hugging the top. The inner rule was an unqualified descendant selector, `:has(.ks-data-table-wrapper) { height: 100% }`, which matches **every** ancestor of a `.ks-data-table-wrapper`. In the `Context` panel that includes the `<details>` elements of the `KV Pair` and `Secrets` collapsibles, so each one took the full height of the panel (629px in a 693px panel) even while closed. `Variables` has no data table, so it stayed at its 29px summary height, which is why only the two lower sections drifted apart.

## Fix

Keep the stretch from descending into a `<details>` subtree. A collapsible has to hug its summary when closed, so it is never a candidate for the full-height treatment.

The only two `<details>` in the codebase are the plugin-docs `SchemaSection.vue`, which contains no data table, and the `EE` `Collapsible.vue` used by the `Context` panel. Measured against the running app, the new selector drops exactly the nine elements inside those two collapsibles and leaves every other match untouched; the listing pages match the same set as before.

## Evidence

Local `EE` instance, `Context` panel with all three sections collapsed. Before, `Secrets` sits 629px below `KV Pair`; after, the three sections stack normally.

| Before | After |
|---|---|
|  |  |

Section heights measured in the same instance:

| | `variables` | `kv-pair` | `secrets` |
|---|---|---|---|
| before | 29px | 629px | 629px |
| after | 29px | 29px | 29px |

Expanding `KV Pair` still works and now sizes to its content (358px) instead of being forced to 629px.

## Note on the branch history

This `PR` originally changed `.task-edit-col` in `TaskEdit.vue`. The `Context` panel is not in that component's tree, and every element carrying `.task-edit-col` already resolves to `display: flex; flex-direction: column`, so that change was a no-op. It is reverted on the branch and replaced with the `app.scss` fix above, verified against a running Enterprise instance. See the discussion below.

### DIFF
--- a/ui/src/components/flows/TaskEdit.vue
+++ b/ui/src/components/flows/TaskEdit.vue
@@ -599,8 +599,6 @@
         min-width: 0;
         min-height: 0;
         height: 100%;
-        display: flex;
-        flex-direction: column;
     }
 
     .task-edit-col-inputs {
--- a/ui/src/components/flows/TaskEdit.vue
+++ b/ui/src/components/flows/TaskEdit.vue
@@ -599,6 +599,8 @@
         min-width: 0;
         min-height: 0;
         height: 100%;
+        display: flex;
+        flex-direction: column;
     }
 
     .task-edit-col-inputs {
--- a/ui/src/styles/app.scss
+++ b/ui/src/styles/app.scss
@@ -102,7 +102,9 @@ main {
             flex: 1;
             min-height: 0;
 
-            :has(.ks-data-table-wrapper) {
+            // Every ancestor of the table is matched here, so keep the stretch out of
+            // collapsible sections, which have to hug their summary when closed.
+            :has(.ks-data-table-wrapper):not(details, details *) {
                 height: 100%;
             }
         }


MilosPaunovic (approved):
Hi @arikaran-rajamanickam, thanks for the thorough write-up, and for being upfront that you could not verify against a real `EE` build. I took that as the cue to reproduce it on a running Enterprise instance, and I have pushed the outcome onto your branch so the contribution stays yours. Here is what came out of it.

**The original change was a no-op.** `.task-edit-col` is applied to a child component, and Vue merges that class onto the child's root element rather than onto a wrapper. So `.task-edit-col-inputs` *is* `.task-edit-data`, `.task-edit-col-output` when collapsed *is* `.task-edit-data-rail`, and `.task-edit-col-params` declares it directly. All three already resolve to `display: flex; flex-direction: column`, so the two added lines changed no computed style anywhere. On top of that, the `Context` panel is an Enterprise-only surface rendered as a separate editor tab, never a descendant of `.task-edit-panel-body`, and `.task-edit-col` sits in a `<style scoped>` block that cannot reach it.

**The real cause was in `OSS` after all, just in a different file.** `ui/src/styles/app.scss` stretches a section to full height when its data table shows an empty state, via `:has(.ks-data-table-wrapper) { height: 100% }`. That is an unqualified descendant selector, so it matches *every* ancestor of a data table, including the `<details>` of the `KV Pair` and `Secrets` collapsibles. Each took the full 693px panel height (629px content box) while closed, which is exactly the gap you were chasing. `Variables` has no data table, which is why only the lower two drifted apart.

I reverted the `TaskEdit.vue` change on your branch and pushed the `app.scss` fix in its place, with before/after evidence in the updated description. Your commit stays first in the history, and the merge is credited to you.

Your CSS reasoning was sound, it was just aimed at the wrong element, and an Enterprise-only surface is not something you had any way to inspect. That part is on us for leaving the issue open to contributors. Thanks for the push on this one, and please do pick up another.

PR #18772, opened 2026-08-28
fix(plugins): show metrics and definitions in the flow editor docs panel

### LINKED ISSUE #17968: [2.0] Plugin Docs - Metrics and Definitions don't show anything
### Describe the issue

No Metrics or Defintions display when you click on a task that should have them like Python. Can't click on it either:

Image

Python does have metrics as can be seen from the plugin docs on the website: https://kestra.io/plugins/plugin-script-python/io.kestra.plugin.scripts.python.script#metrics-body.

If I go to a definition from within a property, and click on it, it then expands the Definitions but the arrows aren't clickable. So I can't expand or collapse any of them. It continues to display across other tabs and even after a refresh!

https://github.com/user-attachments/assets/ac3f2e0a-4f2f-424e-9233-b6aa6cf9009f

### Environment

- Kestra Version: 2.0 RC7 OSS

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/17968.

### ✨ Description

In the flow editor docs panel, the `Metrics` and `Definitions` sections of a plugin's documentation showed nothing and could not be clicked. The compact panel redesign gave every section its own nav chip and tab, but left `Metrics` and `Definitions` behind: they had no chips, rendered as permanently collapsed headers on every tab, and a compact-mode CSS rule (`pointer-events: none`, meant to turn the top-level section headers into static labels) also hit their headers and the nested definition arrows, so nothing could be expanded or collapsed - including after navigating to a definition from a property, which then stuck across tabs and refreshes.

This PR makes `Metrics` and `Definitions` first-class tabs of the panel:

- both sections get nav chips with counts, and their content renders on their own tab like the other sections
- the `pointer-events` rule is scoped to the top-level section headers only, so the nested definition arrows expand and collapse again
- clicking a definition reference inside a property switches the panel to the `Definitions` tab and scrolls to the target definition, instead of leaking an expanded `Definitions` block under every tab
- the full plugin docs page (non-compact mode) is unchanged

| Before | After |
|---|---|
| before | metrics |

After - definitions tab, with a definition expanded through its now-clickable arrow:

### 🎨 Frontend Checklist

### 📝 Additional Notes

- The new `plugins.nav_metrics` key is allow-listed in `translationRules.mjs` for the untranslated-keys check: the translation model keeps "Metrics" in English for Hindi on every reroll, matching its sibling nav labels ("Outputs", "Tasks") that stay English through the reserved-terms rule.
- Covered by a new Vitest case in `PluginDocumentation.spec.ts` asserting the chips show their sections and hide the others; it fails on the previous commit, where the chips do not exist.

### 🤖 AI Authors

Why did the cat sit on the docs panel? Because even it could not click the `Definitions` arrows - now it gets its own tab to nap on. 🐱


Pradumnasaraf (approved):
Everything is working; all pointers are working as expected. LGTM

Image
Image
Image

PR #18780, opened 2026-08-28
fix(core): inspect getDeadline() in DAILY_TIME_WINDOW branch of TimeWindowValidator

### LINKED ISSUE #18763: TimeWindowValidator: DAILY_TIME_WINDOW check inspects the wrong field — causes a 500 error and a false validation message
### Describe the issue

The validator for `preconditions.timeWindow` on the `Flow` trigger has a bug in its `DAILY_TIME_WINDOW` branch: the check meant to catch an invalid `deadline` inspects the `startTime` field instead, a field that's required for this same type. This produces two distinct, broken outcomes depending on what else is set.

**Steps to reproduce**

Scenario 1 — `deadline` set alone, no `startTime`:

```yaml
id: repro_timewindow_silent
namespace: qa.repro
tasks:
  - id: log_message
    message: test
triggers:
  - id: flow_trigger
    preconditions:
      id: precond_a
      timeWindow:
        type: DAILY_TIME_WINDOW
        deadline: "09:00:00"
      flows:
        - namespace: qa.repro
          flowId: some_upstream
          states: [SUCCESS]
```

Scenario 2 — a different invalid field (`window`) alongside the legitimately-required `startTime`/`endTime`:

```yaml
id: repro_timewindow_wrongmsg
namespace: qa.repro
tasks:
  - id: log_message
    message: test
triggers:
  - id: flow_trigger
    preconditions:
      id: precond_b
      timeWindow:
        type: DAILY_TIME_WINDOW
        window: PT1H
        startTime: "06:00:00"
        endTime: "09:00:00"
      flows:
        - namespace: qa.repro
          flowId: some_upstream
          states: [SUCCESS]
```

**Expected**

Scenario 1: a clear validation error naming the actual problem (`deadline` isn't valid for `DAILY_TIME_WINDOW`).
Scenario 2: only the errors for fields actually set incorrectly.

**Actual**

Scenario 1 — saving in the UI editor fails with a raw internal exception, not a normal validation message:

```
Unable to validate the flow: HV000033: At least one custom message must be created if the default error message gets disabled.
```

Same via the API — `HTTP 500`:

```json
{"message":"Internal server error: HV000033: At least one custom message must be created if the default error message gets disabled."}
```

Scenario 2 — the editor shows two errors, one correct and one false:

```
Validation error: triggers[0].preconditions.timeWindow: Time window of type `DAILY_TIME_WINDOW` cannot have a deadline. triggers[0].preconditions.timeWindow: Time window of type `DAILY_TIME_WINDOW` cannot have a window.
```

`deadline` was never set in this flow.

Both scenarios reproduced two ways :
Directly in the UI editor and via the API, with identical results each time.

**Root cause**

In `TimeWindowValidator.java` (confirmed at commit `c4e86ac`, v1.3.35), the `DAILY_TIME_WINDOW` case inspects the wrong getter:

```java
if (value.getStartTime() != null) {
    context.buildConstraintViolationWithTemplate("Time window of type `DAILY_TIME_WINDOW` cannot have a deadline.").addConstraintViolation();
}
```

This should check `value.getDeadline() != null`. Since `startTime` is required for `DAILY_TIME_WINDOW` (validated further down in the same method), this wrong check fires on ordinary, correctly-configured triggers whenever any other field is invalid (scenario 2), and never fires when `deadline` alone is the actual problem (scenario 1). In that case `disableDefaultConstraintViolation()` is called but no replacement message is ever added, and the flow fails with the raw `HV000033` error instead of a normal validation result.

### Environment

- Kestra 1.3.35 OSS, commit `c4e86ac`
- Docker, local, macOS 26.5.2 arm64

ssdalai:
Closes https://github.com/kestra-io/kestra/issues/18763

### What changed

The `DAILY_TIME_WINDOW` case in `TimeWindowValidator` was inspecting the wrong field. The check meant to flag an invalid `deadline` was looking at `startTime`, which is itself required for this type. Two failure modes flowed from that:

- When a user set a `deadline` with no `startTime`, the validator would call `disableDefaultConstraintViolation()` and then never add a replacement message, producing an `HV000033` 500 in the UI/API.
- When a user set a legitimate `startTime` + `endTime` + an unrelated invalid field like `window`, the same wrong getter fired and the user got a false "cannot have a deadline" message alongside the real one.

This swaps the check to `getDeadline()` and updates the existing test that was passing only because of the bug. A new regression-guard test covers the exact scenario the reporter hit (valid `startTime` + `endTime` + invalid `window`, expecting exactly one violation, not two).

### Why now

Both repro scenarios are reachable through the regular flow authoring path, so any user editing a trigger that uses `DAILY_TIME_WINDOW` either gets a server error or a confusing false-positive. The fix is one line; the test coverage is what makes the change safe to merge.

### Notes

- I did not add an explicit test for scenario 1 ("deadline alone, no startTime"). That path was already covered indirectly by the existing `shouldNotValidateDailyTimeWindowWhenInvalidParam` once I added a `deadline` to its inputs — the message it asserts ("cannot have a deadline") can only be produced by the `getDeadline() != null` branch after the fix.
- No changes outside the validator and its test. Spotless left those files alone.

### Backend checklist

- [x] `./gradlew :core:test --tests io.kestra.core.validations.TimeWindowValidationTest` passes (14/14)
- [x] `./gradlew :core:spotlessApply` ran; only the two files in this PR were reformatted, the rest were reverted

🐱 Why did the cat sit on the keyboard? Because it wanted to check the getter.

### DIFF
--- a/core/src/main/java/io/kestra/core/validations/validator/TimeWindowValidator.java
+++ b/core/src/main/java/io/kestra/core/validations/validator/TimeWindowValidator.java
@@ -59,7 +59,7 @@ public class TimeWindowValidator implements ConstraintValidator<TimeWindowValida
                     if (value.getWindowAdvance() != null) {
                         context.buildConstraintViolationWithTemplate("Time window of type `DAILY_TIME_WINDOW` cannot have a window advance.").addConstraintViolation();
                     }
-                    if (value.getStartTime() != null) {
+                    if (value.getDeadline() != null) {
                         context.buildConstraintViolationWithTemplate("Time window of type `DAILY_TIME_WINDOW` cannot have a deadline.").addConstraintViolation();
                     }
                     yield false;
--- a/core/src/test/java/io/kestra/core/validations/TimeWindowValidationTest.java
+++ b/core/src/test/java/io/kestra/core/validations/TimeWindowValidationTest.java
@@ -77,13 +77,40 @@ class TimeWindowValidationTest {
 
     @Test
     void shouldNotValidateDailyTimeWindowWhenInvalidParam() {
-        var sla = TimeWindow.builder().type(TimeWindow.Type.DAILY_TIME_WINDOW).startTime(LocalTime.now()).endTime(LocalTime.now()).window(Duration.ofHours(1)).build();
+        var sla = TimeWindow.builder()
+            .type(TimeWindow.Type.DAILY_TIME_WINDOW)
+            .startTime(LocalTime.now())
+            .endTime(LocalTime.now())
+            .window(Duration.ofHours(1))
+            .deadline(LocalTime.now())
+            .build();
 
         Optional<ConstraintViolationException> valid = modelValidator.isValid(sla);
         assertThat(valid.isEmpty()).isFalse();
         assertThat(valid.get().getConstraintViolations()).hasSize(2);
         assertThat(valid.get().getMessage()).contains(": Time window of type `DAILY_TIME_WINDOW` cannot have a window.\n");
         assertThat(valid.get().getMessage()).contains(": Time window of type `DAILY_TIME_WINDOW` cannot have a deadline.\n");
+    }
+
+    @Test
+    void shouldNotValidateDailyTimeWindowWhenStartTimeLooksLikeADeadline() {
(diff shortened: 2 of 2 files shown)


loicmathieu (approved):
LGTM, thanks

Issue #10360, opened 2025-07-27
Strange behavior - when selecting a namespace file from a folder, the folder is highlighted too


anna-geller:
### Describe the issue

https://jam.dev/c/3a83fd0d-d86b-4ac3-842d-a626c1b1edee

<img width="3510" height="2302" alt="Image" src="https://github.com/user-attachments/assets/2024545f-5e0a-4959-9c0a-1a1f3549e905" />

OSS develop from Friday 25.07 10 AM 

### Environment

- Kestra Version: develop

MilosPaunovic:
Hi @deveshpandee,

I see that you've opened a PR for this issue, and if you add a comment here I can assign it to you (otherwise I can't, due to GitHub privacy policies).

deveshpandee:
@MilosPaunovic, Thanks for the heads-up! 
The issue was closed after my PR was merged, but if possible, please assign it to me. Thanks!

PR #18783, opened 2026-08-28
fix(core): strict boolean KV validation, If truthiness docs, and allowFailure on execution-updating tasks

### LINKED ISSUE #18499: [Bash2.0] "If" task evaluates the string "false" as false, contradicting its docs
## Describe the issue

The If task docs say: "Renders condition and coerces it to boolean (empty string/0/null is false, everything else true)." By that rule the string `"false"` should be true, but the code treats `"false"` as false, so a condition rendering to `"false"` runs `else`, not `then`. Docs and behaviour disagree.

## Steps to reproduce

1. Set an If `condition` that renders to the string `"false"`.
2. Run the flow.

## Actual behaviour

`"false"` evaluates to false and the `else` branch runs, even though the docs say everything except empty string / 0 / null is true.

## Visuals

Image

## Environment

- Kestra Version: 2.0.0-rc10
- Edition: Enterprise

nkwiatkowski:
### What this PR does

Three core task-semantics bugs from the bug bash, each accepting or hiding something it should have surfaced:

- **kv.Set BOOLEAN accepted anything** (kestra-ee#10324) — `kvType: BOOLEAN` coerced any non-`"true"` string to `false` and still succeeded, so `"yes"`, `"1"`, `"maybe"` all silently stored `false`. Every other `kvType` (NUMBER, DATETIME, DATE, DURATION, JSON) fails the task on a bad value. BOOLEAN now accepts only `true`/`false` (case-insensitive) and fails the task on anything else, matching the others.
- **If docs contradicted behaviour** (kestra#18499) — the `If` task documented that only an empty string / `0` / `null` are false, but the code (`TruthUtils`) also treats the string `"false"` (and `"-0"`) as false. A condition rendering to `"false"` correctly runs `else`; the docs were the thing that was wrong. Behaviour is unchanged — only the two `@Schema` descriptions now state the real coercion rule.
- **allowFailure ignored by SetVariables / UnsetVariables / Labels** (kestra-ee#10360) — these three are inline *execution-updating* tasks, not worker tasks, so when they failed the executor hard-failed the whole execution and ignored `allowFailure` (unlike every runnable task). The executor now honours `allowFailure` for them — the task ends `WARNING` (or `SUCCESS` with `allowWarning`) and the flow continues — and logs the failure against the task run so it carries its `taskId`/`taskRunId` instead of only surfacing at execution level (the issue's secondary complaint).

### Evidence

- `SetTest` — `booleanTypeRejectsInvalidValue` (`"yes"` → task fails), `booleanTypeAcceptsTrueAndFalseCaseInsensitive`.
- `SetVariablesTest.shouldWarnAndContinueWhenAllowFailureIsSet` (new flow `set-variables-allow-failure.yaml`) — the failing `SetVariables` ends `WARNING`, the next task runs `SUCCESS`, execution ends `WARNING`.
- `H2RunnerTest` green (executor change).

### Note on scope

Kept as one PR since each fix is small; they share the "core task semantics" theme. #10360 touches the executor failure path — it was earmarked for 2.1, but turned out to be a contained change to a single catch block, so it's included here.

Closes https://github.com/kestra-io/kestra/issues/18499.

### EE issues (same fixes, cross-repo — close manually on merge)

GitHub only auto-closes issues in this repo:

- https://github.com/kestra-io/kestra-ee/issues/10324
- https://github.com/kestra-io/kestra-ee/issues/10360

### DIFF
--- a/executor/src/main/java/io/kestra/executor/ExecutorService.java
+++ b/executor/src/main/java/io/kestra/executor/ExecutorService.java
@@ -1517,29 +1517,17 @@ public class ExecutorService {
 
                     // State.Type.fail resolves the failure against allowFailure/allowWarning: FAILED, WARNING or SUCCESS.
                     State.Type failState = State.Type.fail(task);
+                    // Fail the execution only when the task itself ends FAILED; WARNING/SUCCESS let it continue.
                     if (failState == State.Type.FAILED) {
-                        // No allowFailure: the task fails and the execution fails with it.
-                        workerTaskResults.add(
-                            WorkerTaskResult.builder()
-                                .taskRun(workerTask.getTaskRun().fail())
-                                .build()
-                        );
                         executor.withException(e, "handleExecutionUpdatingTasks");
-                    } else if (task.isAllowWarning()) {
-                        // allowFailure + allowWarning: promote the failed task to SUCCESS and let the execution continue.
-                        workerTaskResults.add(
-                            WorkerTaskResult.builder()
-                                .taskRun(workerTask.getTaskRun().fail().withState(State.Type.SUCCESS))
-                                .build()
-                        );
-                    } else {
-                        // allowFailure: downgrade the failed task to WARNING and let the execution continue.
-                        workerTaskResults.add(
-                            WorkerTaskResult.builder()
-                                .taskRun(workerTask.getTaskRun().fail().withState(State.Type.WARNING))
-                                .build()
-                        );
                     }
+
+                    TaskRunAttempt failedAttempt = TaskRunAttempt.builder().state(new State().withState(failState)).build();
+                    workerTaskResults.add(
+                        WorkerTaskResult.builder()
+                            .taskRun(workerTask.getTaskRun().withAttempts(List.of(failedAttempt)).withState(failState))
+                            .build()
+                    );
                 }
                 return true;
             });
--- a/executor/src/main/java/io/kestra/executor/ExecutorService.java
+++ b/executor/src/main/java/io/kestra/executor/ExecutorService.java
@@ -1511,12 +1511,25 @@ public class ExecutorService {
                             .build()
                     );
                 } catch (Exception e) {
-                    workerTaskResults.add(
-                        WorkerTaskResult.builder()
-                            .taskRun(workerTask.getTaskRun().fail())
-                            .build()
-                    );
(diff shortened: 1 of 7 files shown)


executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -1511,12 +1511,25 @@ private ExecutorContext handleExecutionUpdatingTasks(final ExecutorContext execu
...
-                        WorkerTaskResult.builder()
-                            .taskRun(workerTask.getTaskRun().fail())
-                            .build()
-                    );
-                    executor.withException(e, "handleExecutionUpdatingTasks");
+                    Task task = workerTask.getTask();
+                    // Log against the task run so the failure carries its taskId/taskRunId, not only the execution.
+                    executorTask.runContext().logger().error("Failed to process task: {}", e.getMessage(), e);
+
+                    if (task.isAllowFailure()) {
```
loicmathieu: You should also honor allowWarning

loicmathieu (changes requested):

executor/src/main/java/io/kestra/executor/ExecutorService.java
```diff
@@ -1511,12 +1511,35 @@ private ExecutorContext handleExecutionUpdatingTasks(final ExecutorContext execu
...
-                            .build()
-                    );
-                    executor.withException(e, "handleExecutionUpdatingTasks");
+                    Task task = workerTask.getTask();
+                    // Log against the task run so the failure carries its taskId/taskRunId, not only the execution.
+                    executorTask.runContext().logger().error("Failed to process task: {}", e.getMessage(), e);
+
+                    // State.Type.fail resolves the failure against allowFailure/allowWarning: FAILED, WARNING or SUCCESS.
+                    State.Type failState = State.Type.fail(task);
+                    if (failState == State.Type.FAILED) {
```
loicmathieu, line 1521: This can be simplified as you have the correct state returned by State.Type.fail() so you don't need to branch on allowFailure/allowWarning.

The only thing to do is to attach the exception inside the executor if it should FAILED then you update the taskRun state and last attempt with the new state

loicmathieu (changes requested):

loicmathieu (approved):
LGTM

PR #18791, opened 2026-08-28
fix(design-system): filter bar wrapped-row alignment and chip overlap

RemiBarthe:
closes https://github.com/kestra-io/kestra-ee/issues/10342
closes https://github.com/kestra-io/kestra-ee/issues/10341

Capture d’écran 2026-08-28 à 13 34 12

This pull request refactors the layout and styling of the `KsFilter` and related components to simplify flexbox usage, remove dynamic JS-based layout calculations, and ensure consistent spacing using design tokens. It also updates Storybook tests to verify correct chip alignment when filters wrap to multiple rows.

**Layout and Styling Simplification:**

* Replaces the flexbox-based `.filter-container` with `display: contents` to simplify the layout and removes the `.global-filters` wrapper and related CSS, eliminating the need for JavaScript-based row alignment and wrapping calculations. [[1]](diffhunk://#diff-0000018ce479baf11c73b530d121c58105fe360eb98f6a5a3a76efce640cebf6L2-R2) [[2]](diffhunk://#diff-0000018ce479baf11c73b530d121c58105fe360eb98f6a5a3a76efce640cebf6L90-L96) [[3]](diffhunk://#diff-0000018ce479baf11c73b530d121c58105fe360eb98f6a5a3a76efce640cebf6R105) [[4]](diffhunk://#diff-0000018ce479baf11c73b530d121c58105fe360eb98f6a5a3a76efce640cebf6L124-L125) [[5]](diffhunk://#diff-0000018ce479baf11c73b530d121c58105fe360eb98f6a5a3a76efce640cebf6L170-L173) [[6]](diffhunk://#diff-0000018ce479baf11c73b530d121c58105fe360eb98f6a5a3a76efce640cebf6L269-R263) [[7]](diffhunk://#diff-0000018ce479baf11c73b530d121c58105fe360eb98f6a5a3a76efce640cebf6L330-L343)
* Updates gap and spacing values in `KsFilter.vue` and `RightFilter.vue` to use the design token `var(--ks-spacing-2)` for consistency. [[1]](diffhunk://#diff-09899b61f3cc5c47dad4c890d6ca13e35dd05192b8dfce3882fc09ca525dabcaL295-R295) [[2]](diffhunk://#diff-bd82a97ebf889c0943f361a132999e3ce00c28c946e01b217077ea195264bdbaL192-R195)

**Code Cleanup:**

* Removes unused refs, variables, and lifecycle hooks related to measuring and observing filter chip wrapping, since layout is now handled by CSS. [[1]](diffhunk://#diff-0000018ce479baf11c73b530d121c58105fe360eb98f6a5a3a76efce640cebf6L148-R140) [[2]](diffhunk://#diff-0000018ce479baf11c73b530d121c58105fe360eb98f6a5a3a76efce640cebf6L170-L173) [[3]](diffhunk://#diff-0000018ce479baf11c73b530d121c58105fe360eb98f6a5a3a76efce640cebf6L269-R263)

**Component Behavior:**

* Adjusts the `.refresh-btn` style to ensure vertical alignment within the new layout.

**Testing Improvements:**

* Adds a Storybook story (`WrappedRowsShareOneLeftEdge`) that asserts when filter chips wrap to multiple rows, each row starts at the same left edge, ensuring visual consistency without JS layout logic. Also introduces a global key configuration for this test. [[1]](diffhunk://#diff-650836553fcac7d56485a271e5345a7c4bbcd4b4799622a576c9e0a0f4d2e8d9R4) [[2]](diffhunk://#diff-650836553fcac7d56485a271e5345a7c4bbcd4b4799622a576c9e0a0f4d2e8d9R52-R66) [[3]](diffhunk://#diff-650836553fcac7d56485a271e5345a7c4bbcd4b4799622a576c9e0a0f4d2e8d9R406-R463)


flcarre (commented):
Really like the direction: trading a `ResizeObserver` plus a `translateX` correction for `display: contents` is the right call, and it is a net deletion of the exact kind of code that is impossible to reason about later. My guess is that the transform is also what caused ee#10341 — a transform does not affect layout, so the translated `.global-filters` group could visually collide with the heading below while its box claimed to be elsewhere. Nice find.

Four things before I approve, one of which is a real gap.

### The new story asserts nothing

The description says it "adds a Storybook story (`WrappedRowsShareOneLeftEdge`) that asserts when filter chips wrap to multiple rows, each row starts at the same left edge". At `26da8fd` the story is `WithWrappedRows` and has no `play` function, so `npm run test:storybook` renders it and checks nothing. The invariant the PR exists to establish — every wrapped row shares one left edge — is currently pinned by nothing, which also means a revert to the old `translateX` approach would keep the suite green.

Since the geometry is the whole point here, a `play` comparing `getBoundingClientRect().left` across the first element of each row would earn its place, and it runs in real Chromium so the measurement is meaningful. Grouping the children of `.main-filters` by rounded `top` and asserting one distinct `left` per group is enough.

### `searchInputFullWidth` is half dismantled

The PR drops `:class="{'filter-grow': …}"` and the `.filter-grow` rule, which carried `flex-wrap: nowrap; flex-grow: 1`. But the prop itself survives and still drives its other two halves:

- `KsFilter.vue:72,90,227` — declared, defaulted, provided through the filter context
- `MainFilter.vue:46,382` — `search-grow` on `.search-container`, `flex: 2 1 auto`
- `RightFilter.vue:4,196` — `filter-shrink`

So the search input still stretches, but the "keep everything on one line" half is gone: those children now sit in `.main-filters`, which is `flex-wrap: wrap`. `BlueprintsFilterBar.vue` is the live consumer of the prop — could you check that bar before/after? Either it still looks right and the `nowrap` was dead weight worth mentioning in the description, or `searchInputFullWidth` needs to keep expressing "one line" somewhere.

### The separator between global and regular filters disappears

`.global-filters` is removed wholesale, and with it the `padding-left` and `border-left: 1px solid var(--ks-border-default)` that visually separated global chips from the rest. Neither ee#10341 nor ee#10342 asks for that — both are about alignment and overlap — so this reads as a side effect of flattening rather than a decision.

Given how visible it is, worth an explicit OK from design, plus a screenshot of a bar that actually has a global chip in it (the one in the description does not obviously show that case). If the separator should stay, it can live on the first global chip rather than on a wrapper.

### `min-width` goes from 7rem to 14rem

`.filter-container` had `min-width: 7rem`; `.main-filters` now has `14rem` (`KsFilter.vue:306`). That doubles the width at which the main group stops shrinking and starts squeezing `RightFilter` instead. Deliberate, or carried over from a draft? Worth a line either way, since it changes the narrow-viewport behaviour the linked issues are about.

### Nit

After this change `MainFilter`'s `.filter-container` is not a container at all, and `RightFilter.vue:3` uses the same class name for a real flex container. Scoped styles keep them from colliding, but two sibling components meaning different things by one class name is a small trap for the next reader — worth renaming the `MainFilter` one now that it is only there to carry the `--ks-box-shadow` value.

Good change overall. Add the `play`, confirm the blueprints bar and the separator, and I am happy.

flcarre (commented):
This is a genuine root-cause fix and worth saying so: the wrapped rows didn't share a left edge because the `{ }` toggle sat *outside* MainFilter's wrapping container, and the old code compensated with a measured `transform: translateX(-Npx)` on the global group — which is what produced the chip overlap in the first place. Moving the toggle inside one `.main-filters` wrap container and deleting the transform plus the inline `:style` is the right shape, and replacing `watch([globalFilters, conditionalFilters, isComplex], …)` with `watch(separatorPair)` on a string moves in the direction `ui/AGENTS.md` asks for.

Design-system compliance is clean: no new `element-plus` import, no new `:deep()`, no new `.el-*` override, no hex/`rgb()`/`--el-*`/`--bs-*`, flat literal class selectors, and every `--ks-*` token used is declared. I also checked the things this kind of change usually breaks and they're fine — `display: contents` doesn't disturb popover/tooltip positioning or the `.customize-button` / `.rules-pill` anchors, `align-items: flex-start` is safe given the 32px chip/button pinning plus the new `align-self: center`, `chipRefs[id].$el` is safe since `FilterChip` has a single root, and renaming `.filter-container` → `.filter-tokens` breaks no external selector or test.

Two things to resolve before merge.

## 1. The group separator disappears entirely in complex mode

`ends-conditional-group` / `shows-group-separator` are only ever applied inside the conditional-chip `v-for`, and that `v-for` is `v-else` to the `rules-pill` (`v-if="isComplex"`). So once `isComplex` is true — two applied filters on Executions or Logs — no element carries the class and **no divider is drawn between the rules pill and the `timeRange` chip**.

The base drew one unconditionally via `.global-filters { border-left: 1px solid var(--ks-border-default) }`, with `is-wrapped` clearing it only on a detected wrap. And `timeRange` with `groupable: false` is the global key on every list configuration (`execution`, `log`, `trigger`, `metric`, `flowExecution`, `dashboard`), so this is the high-traffic path rather than an edge case.

Fix: also apply the classes to `.rules-pill` when `isComplex && globalFilters.length > 0`, and add a story covering it (two applied filters plus `timeRange`). If dropping the divider in complex mode is deliberate, could you say so in the PR body? It isn't mentioned.

## 2. The separator can go stale on a same-height reflow

`ResizeObserver` fires on `.main-filters`' own size, and `separatorPair` only changes when the two *boundary* chip ids or `isComplex` change. A reflow that moves the boundary chips to different rows without changing the container height, and without changing those two ids, updates neither trigger.

Concrete case: two rows, `[namespace][state]` / `[Started][reset]`. Widen the `namespace` chip's value so `state` moves down next to `Started` — row count and height unchanged, last-conditional id still `state`, first-global id still `Started`. The separator should appear and stays hidden.

The previous `watch([globalFilters, conditionalFilters, …])` fired on every array recompute, so this staleness class is new. The clean fix is to have the same observer watch the chip elements too — re-`observe` all `.filter-chip-wrap` when `separatorPair` changes, since a chip resizing is the actual cause of these reflows. A cheap partial hardening is to fold the two lengths into the key, which closes the mid-list add/remove variants.

Also, one correctness note on the `min-width: 14rem` comment: it claims the group can never be squeezed narrower than its own content, but `FilterChip`'s `.chip` is capped at `max-width: 300px` (18.75rem). Since `.filter-chip-wrap` is `flex-shrink: 0`, a container under ~300px overflows instead — the overlap symptom this PR fixes, reappearing in narrow panels. `min-width: fit-content` or a value at/above the chip cap would make the claim true.

flcarre (approved):
Re-reviewed at `1fd391b1`. Both findings from my earlier comment are resolved, and I checked the source rather than the commit message:

- **The separator now renders in complex mode.** The `rules-pill` is wrapped in a `.filter-chip-wrap` that carries `ends-conditional-group` / `shows-group-separator` when `globalFilters.length > 0`, so the divider between the rules pill and the `timeRange` chip is drawn again. That was the high-traffic path — `timeRange` is the global key on every list configuration.
- **The stale-measure gap is closed.** The observer now observes every `.filter-chip-wrap` in addition to the bar, so a chip resizing — the actual cause of these reflows — retriggers the measurement even when the container height doesn't change.
- `WithFoldedRulesSeparator` pins the folded case with a real `play` assertion.

You also corrected the `min-width` comment to state the 300px-chip limitation honestly instead of asserting an invariant that didn't hold. Documenting the constraint rather than papering over it is the right call.

The underlying fix remains the right one: moving the `{ }` toggle inside the wrapping container and deleting the measured `transform: translateX(-Npx)` addresses the cause of the overlap rather than compensating for it. Design-system compliance is clean, and CI is green on this head.

Approving.

PR #18799, opened 2026-08-28
Fix/webserver openurls route identity

loicmathieu:


loicmathieu:
Copy of https://github.com/kestra-io/kestra/pull/18788 which targets the wrong branch :(

PR #18803, opened 2026-08-28
fix(executions): map oversized payloads to 413 instead of a raw 500

nkwiatkowski:
### What this PR does

Two bug-bash issues where a request whose payload exceeds the queue message-size limit (`kestra.queue.message-protection.limit`, 1MB by default) surfaced as an opaque `500` instead of a clear `4xx`:

- **Large flow input** (kestra-ee#10263) — executing a flow with an input value ≥1MB returned `500 Internal server error: io.kestra.core.queues.MessageTooBigException: …`.
- **Large webhook body** (kestra-ee#10260) — posting a body >1MB to a Webhook trigger returned an empty `500`.

Both are the same underlying `MessageTooBigException` thrown when the `ExecutionCommand` is enqueued.

### Fix

A single global `ErrorController` handler maps `MessageTooBigException` to **413 Request Entity Too Large**, so any endpoint hitting the queue guard behaves consistently. The callers that enqueue the command previously erased the exception's type before it could reach that handler, and are updated to let it through:

- `ExecutionController.awaitBlockingAction` and `WebhookService.startExecution` wrapped every `QueueException` in a plain `RuntimeException`; they now propagate `MessageTooBigException` with `Exceptions.propagate` so the reactive chain delivers it as its real type.
- `Webhook.evaluate` ended with a blanket `onErrorReturn(INTERNAL_SERVER_ERROR)` that masked every error; it now re-raises `MessageTooBigException` (mapped to 413 upstream) and keeps the 500 for other errors.

Not raising the queue limit (cluster-wide implications) — only turning the breach into a clean client error.

### Evidence

New `ExecutionControllerMessageProtectionTest` enables `message-protection` at the production 1MB limit and posts ~2MB payloads: a flow input and a webhook body both return **413**. Existing `ExecutionControllerTest`, `ErrorControllerTest`, `WebhookRoutingTest` and `WebhookPluginTest` stay green.

### EE issues (same fix, cross-repo — close manually on merge)

GitHub only auto-closes issues in this repo:

- https://github.com/kestra-io/kestra-ee/issues/10260
- https://github.com/kestra-io/kestra-ee/issues/10263

### DIFF
--- a/core/src/main/java/io/kestra/core/services/WebhookService.java
+++ b/core/src/main/java/io/kestra/core/services/WebhookService.java
@@ -23,6 +23,7 @@ import io.kestra.core.models.flows.FlowInterface;
 import io.kestra.core.models.flows.State;
 import io.kestra.core.models.triggers.AbstractTrigger;
 import io.kestra.core.queues.DispatchQueueInterface;
+import io.kestra.core.queues.MessageTooBigException;
 import io.kestra.core.runners.FlowInputOutput;
 import io.kestra.core.runners.FlowMetaStoreInterface;
 import io.kestra.core.runners.FlowMetaStores;
@@ -44,6 +45,7 @@ import io.opentelemetry.context.propagation.ContextPropagators;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 import lombok.extern.slf4j.Slf4j;
+import reactor.core.Exceptions;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/kestra/core/services/WebhookService.java
```diff
@@ -234,6 +236,9 @@ public Mono<AsyncOperationProcessedEvent> startExecution(Execution execution) {
                 try {
                     executionCommandQueue.emit(command.withOperationId(operationId));
                     eventPublisher.publishEvent(CrudEvent.create(execution));
+                } catch (MessageTooBigException e) {
```
loicmathieu: Can't this be a single catch with an instance of?
If it's a runtime re throw else encapsulate in a runtime

loicmathieu (approved):

PR #18807, opened 2026-08-28
fix(core): report a fetch 404 as a toast, and resolve a page's entity in a guard

### LINKED ISSUE #9755: UX: Everytime a fetch request returns 404 we redirect the whole page.

elevatebart:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/9755.

### ✨ Description

Any fetch that 404ed set `coreStore.error`, which swaps the whole page for the not-found screen. One failed sub-request hid which request failed, on a screen there was no way back from.

A 404 now takes the same toast path as every other error status, and the failing method and URL are logged to the console. Since #17912 that toast renders the RFC 9457 problem document, so a 404 shows `errors.problems.not-found.title` and the server's own `detail`. The not-found screen stays for what it is meant for: an unmatched route, and pages that set `coreStore.error` themselves.

That alone would leave a genuinely missing flow, execution or namespace rendering a half-baked page — the flow page renders nothing at all, since its whole template is gated on the flow. So detail pages now declare their entity as `meta.entity`, and one guard resolves it before the page mounts: a missing entity renders the not-found screen at the requested URL, and every other failure is still left to the page and its toast.

Five details worth a reviewer's attention:

- **The guard is global, not a per-record `beforeEnter`**, which vue-router skips when only the params change — navigating straight from one flow to another (subflow links, parent/child executions) would otherwise go unchecked.
- **One request, not two.** Each resolver goes through the same store loader the page uses, and each page fetches only what the store does not already hold. The pages therefore mount with their entity in hand: the flow page no longer renders nothing on first paint, and the execution page no longer waits for its first SSE event.
- **Only the newest navigation decides what is on screen.** The guard numbers its navigations and drops a verdict whose navigation has been superseded, and `loadFlow` / `loadExecution` / the namespace `load` number their loads and drop a superseded resolution, the way `stores/logs.ts` already does for a superseded search. Without both, opening a slow flow and clicking a second one before it landed left the first flow's source rendering under the second flow's URL. One race is left, in the guard's early-return path, and is written up in #18861 with the two ways out of it.
- **A resolver puts the navigation behind a network round trip**, with no timeout of its own. That is the trade for not mounting a page around data it never got, and it is why the guard resolves through the store rather than adding a request: the page that mounts is the one that would have made the same call. Worth weighing before the pattern spreads to more routes — see #18861.
- **Dashboards are deliberately excluded.** An unknown dashboard id already falls back to the bundled default with a toast, which is what an admin's deleted default dashboard needs; a not-found screen there would be a regression.

Namespaces only report a missing entity in EE — the OSS endpoint echoes any id back, since an OSS namespace is whatever its flows declare rather than a stored entity.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit`) — 233 files, 2180 tests; Storybook 68 files, 254 tests

### 📝 Additional Notes

- **Merged with #17912 (RFC 9457 problem details), which rewrote this same error path.** `kestraHttp.ts` keeps develop's problem machinery (`asProblem`, `problem` on the error, `legacyData` for non-API bodies, a 400 that rejects like anything else) and this branch's routing (one-argument `onError`, a 404 reported as a toast rather than `coreStore.error`, the hoisted `showMessageOnError` opt-out), with develop's message payload. `ErrorToast.vue` is develop's version outright: its `problemTitle` / `problemDetail` already localize a 404 and show the server's detail, which supersedes the title and body this branch was assembling by hand — so the `errors.requestNotFound` keys and the spec pinning them are dropped rather than merged, and the translation files are develop's.
- **A 404 now toasts, so the editor's own report became a second toast.** Saving a flow deleted under the editor showed `Error / Flow 'x' was not found.` from `useFlowEditorActions.reportSaveError` beside the global `Resource not found` toast, which carries the problem title and the Fix with AI action. The interceptor's decision is now a predicate it uses itself, `isReportedCentrally`, so a caller that reports its own failures can ask instead of restating the condition. The editor's save handler asks — a 400 validation error and a lost connection still get their own toast, since the interceptor leaves those alone — and so does `onDelete`, whose `Failed to delete flow …` no longer stacks on the global toast when the flow is already gone. `FlowsSearch`'s replace and `FlowTriggers`'s actions can double the same way and are left to #18849, which is about collapsing those reporting paths into one.
- **`flex-direction: column` on `#app` is a drive-by fix, kept here because the toast work is what surfaced it.** `#app` is a flex container, and in row direction every in-flow child shares a row with `#app-shell` instead of stacking above it: `DocIdDisplay` (under `?showDocId`), the `PwaInstallPrompt` alert and the dev-only `SdkDriftBanner` each took width from the shell. `#app-shell` already carries `flex: 1; min-height: 0` (`ui/src/styles/app.scss:240`), which only makes sense in a column, so the declaration was missing rather than new. It is one line and global; @elevatebart, who wrote it, checked it against a running instance — happy to lift it into its own PR with a before/after if you would rather it not ride along.
- **The other by-id pages are out of scope, and #18861 tracks them.** Removing the blanket 404 redirect also removes it from every edit page whose entity is missing — `admin/mcp-servers/update` in OSS, and around fifteen EE routes (`apps/update`, `cases/update`, `credentials/update`, `admin/iam/*`, the policy and service-account routes), where an empty form beside a live Save is the worst of the cases. Extending `meta.entity` to them is mechanical, but it is fifteen resolvers through fifteen store loaders, so it is a follow-up rather than an addendum here. `Overview.vue`'s "Execution {id} is not found." copy is recorded there too: the guard now intercepts before that branch can mean what it says.
- **Screenshot still missing** — it needs a running instance, and the namespace case an EE one. Flagging it rather than hiding it: everything else is ready for review.
- `allowDeleted: true` in the flow resolver is load-bearing: without it a deleted flow's page (a real page) would start 404ing.
- A store signature must not name the SDK's own option type — `Parameters[1]` in `loadFlow` broke the EE typecheck, since EE compiles these sources against its own SDK build. Hence the local `KestraRequestOptions`.
- A route comparison inside the store loaders would have covered more of the staleness than the load sequence does, but `stores/playground.ts`, `executions/Topology.vue`, `ExecutionPending.vue` and the product tour all load a flow that is deliberately not the route's, so it would have silently stopped storing for them.
- Tests: `routeEntityGuard.spec.ts`, `entityLoadRace.spec.ts` and the two `useFlowRoot` cases each fail on the commit before the one that introduced them; `flowDraftSave.spec.ts` changed only because `loadFlow` gained a second argument.
- **Both red checks from the first review round are green on this head**, along with everything else. `trigger-ee` returns a verdict again, and `OpenTelemetry - Export Trace` was a missing `otel-collect` in `kestra-io/actions@main`, since fixed upstream — an earlier note here claimed it was red on develop, which no longer holds.

### 🤝 Related PRs

- EE companion, same branch name: https://github.com/kestra-io/kestra-ee/pull/10543 — merge this one first; it carries the guard, the shared route meta and the store loaders.


elevatebart:
Two checks failed on `8ed9afb` (the develop merge commit). Neither is this PR's, and there is no fix in this repo to port for either.

**`OpenTelemetry - Export Trace` — repo-wide, from `kestra-io/actions`.**

```
Can't find 'action.yml', 'action.yaml' or 'Dockerfile' for action
'kestra-io/actions/actions/otel-collect@main'.
```

`otel-collect` is gone from `kestra-io/actions@main` (currently `af2c949`), so every workflow referencing it dies during *"Prepare all required actions"*, before any test body runs. It reproduces on **develop**: [run 33403487604](https://github.com/kestra-io/kestra/actions/runs/33403487604) (head `6f8d38c`) fails five jobs on the identical error — Backend tests, Design System tests, Frontend tests and the OTel job. The same reference exists in `main-build.yml:177`, `pre-release.yml:123`, `e2e-scheduling.yml:39` and `pull-request.yml`, so it is not specific to any branch. The fix belongs in `kestra-io/actions` (restore the action or update the four references); nothing here can carry it.

**`trigger-ee` — inconclusive EE verdict, self-labelled transient.** The webhook returned HTTP 200 with every field empty, and the job's own message says so:

```
EE compile check inconclusive for ref … (no compile verdict; HTTP 200).
Likely a transient runner error — re-run the job.
```

It passed on the previous head (`5e24e8e`), so this is not a new compile break. The companion EE PR is kestra-io/kestra-ee#10543 and its own CI is green.

Both are already being re-run: pushing `1b1e4aa` started [run 33405102952](https://github.com/kestra-io/kestra/actions/runs/33405102952) on the current head, so no manual re-run was spent. If `trigger-ee` fails a second time there I will treat it as real and dig in; the OTel job will keep failing until `kestra-io/actions` is fixed.


---

elevatebart:
Correcting myself on `trigger-ee` — it is **not** transient, and it is specific to this ref.

It failed again on the current head `1b1e4aa` ([job](https://github.com/kestra-io/kestra/actions/runs/33405102952/job/99530788998)) with the identical signature: webhook HTTP 200, `{"testFailed":"","testReport":"","compileFailed":"","compileError":""}`, after ~6 minutes. Meanwhile a contemporaneous run on another branch — `issue-18760-fix`, [run 33403068576](https://github.com/kestra-io/kestra/actions/runs/33403068576) at 14:30 — had `trigger-ee` **green**, its webhook step taking 8m34s. The EE check service is up; it just returns no verdict for this ref.

**There is no compile error to fix here.** The response carries an empty `compileError`/`compileFailed`, i.e. the EE-side flow never produced a verdict — it is not reporting that this branch fails to compile.

**And the verification it performs is already green by another route.** CI pairs OSS and EE branches by name, so the EE companion kestra-io/kestra-ee#10543 builds EE against *this* OSS branch — its `Build Artifacts / Build Jar` passed on `8c44c4f`, as did its frontend tests. That is the same OSS×EE compile this check exists to prove.

The one structural difference between this ref and the refs where `trigger-ee` passes is that this one **has a same-named branch in `kestra-io/kestra-ee`**. That makes the EE-side compile-check flow the place to look — it needs someone with access to it on prd.kestra-tech.kestra.cloud; I can't inspect or fix it from this repo, and the diff here is frontend-only so it cannot affect a Java compile.

Not spending another re-run: two attempts on two different heads produced identical empty verdicts, so a third would tell us nothing new. My note above about `OpenTelemetry - Export Trace` stands unchanged — that one is repo-wide and reproduces on develop.


---

MilosPaunovic (changes requested):
Reviewed at `feaeee489243105be21cc1204d80cca55167fe18`, together with the `EE` companion https://github.com/kestra-io/kestra-ee/pull/10543. Findings 1 and 2 are reproduced with throwaway specs against this branch.

## Blocking

### 1. `Errors.vue` wipes `coreStore.error` immediately, so the not-found screen never survives the navigation that sets it

`entityNotFoundGuard` writes `coreStore.error = 404` in `beforeResolve`, which is before `vue-router` commits the new route. The reactivity flush queued by that write runs before the route object updates, so `Errors.vue` mounts against the old `route.fullPath`, registers its own watcher, and the route commit then fires it:

```ts
// ui/src/components/errors/Errors.vue:44
watch(() => route.fullPath, () => { coreStore.error = undefined })
```

Repro with a real router, the parent `redirect` plus tab child, `DefaultLayout`'s `v-if`/`v-else` shape and `Errors.vue`'s watcher verbatim, navigating to a missing `flow`:

```
entity resolver -> 404
console.error: No entity behind /flows/edit/io.kestra/missing/edit, showing the not-found screen instead.
Errors mounted, route=/
Errors.vue fullPath watcher -> coreStore.error = undefined
!! FlowRoot MOUNTED - it will fetch the flow itself
FINAL: coreStore.error=undefined html=<main> <div>flow page (blank...)</div> </main>
```

End state for a missing `flow`: a one-frame flash of the not-found screen, then `FlowRoot` mounts and `useFlowRoot.load()` refetches. That second call carries no `ENTITY_REQUEST_OPTIONS`, so the 404 toasts, and the page's template is gated on the flow, so it renders nothing. A blank page plus a 404 toast is close to the state this `PR` sets out to remove. Same for `executions` and for `EE` `namespaces`, and same through `OnlyLeftMenuLayout.vue`.

Either fix works: drop the `route.fullPath` watcher from `Errors.vue`, since the guard already clears `coreStore.error` on every navigation and the watcher is now redundant, or move the write out of `beforeResolve` and apply it once the route has committed.

### 2. A superseded navigation's probe still writes `coreStore.error`

The guard awaits the probe and then assigns unconditionally, with no check that the navigation it belongs to is still current. Opening a missing `flow` and immediately clicking an existing one lets the slower 404 resolve last:

```
landed on: /flows/exists   coreStore.error: 404
```

The store writes have the same exposure: `useFlowStore().loadFlow(...)` and `loadExecution(...)` both commit their result for a navigation that may already be abandoned. Comparing `to.fullPath` against `router.currentRoute` before assigning, or applying the verdict after the navigation commits, covers this and finding 1 together.

## Worth addressing before merge

### 3. The safety net is removed globally but restored for three entities

`handleErrorCentrally` no longer sets `coreStore.error` for any 404, but only `flow`, `execution` and `namespace` declare `meta.entity`. Every other detail page loses the not-found screen and renders around data it never got. In `EE` that is around fifteen routes in `ui-ee/src/routes/routes.ts`: `apps/update`, `apps/view`, `cases/update`, `credentials/update`, `admin/iam/roles/update`, `admin/iam/groups/update`, `admin/iam/users/update`, the service-account routes, the policy routes, `admin/instance/worker-queues/update`. In `OSS`, `admin/mcp-servers/update`.

Concretely, `ui-ee/src/components/roles/Role.vue` renders its form off `item`, so a missing `role` id gives an empty role editor with a live Save button where the not-found screen used to be. Either extend `meta.entity` to those routes, or state in the description that they are out of scope and file the follow-up.

### 4. Navigation now blocks on a network round trip

`beforeResolve` returns only once the entity request resolves, so clicking a `flow` row leaves the user on the previous page with just the `NProgress` bar until the response lands, with no timeout. The old behaviour painted the page shell immediately. The one-request tradeoff is real, but it is paid on every `flow`, `execution` and `namespace` open, so it is worth calling out in the description either way.

### 5. Executions already had an in-page not-found state

`ui/src/components/executions/overview/Overview.vue` renders `KsNoData` with `execution not found` when the store holds no execution. With the guard in front of it, that branch is effectively unreachable for a genuinely missing `execution` and the global screen wins instead. Two treatments for one case: worth picking one and dropping or keeping the translation key accordingly.

### 6. Guard store writes land before the old page tears down

For `executions` the guard writes `execution.value` while the previous execution's `SSE` stream is still open, since `followExecution` only closes it once `ExecutionRoot` reacts to the committed route. An event arriving in that window overwrites the store with the previous execution, `followExecution` then sees a mismatched id, clears it and refetches, which loses the optimization the change is for. Low severity, but worth a comment on the ordering at least.

## Nits

- `ui/src/composables/useBaseNamespaces.ts:46` says a missing namespace is reported through `existing`. That ref has one consumer, `ui-ee/src/override/components/namespaces/components/buttons/DetailsActions.vue:68`, and it is never reset to `true` after a miss, so it behaves as a sticky flag rather than a report channel. Pre-existing, but the new comment leans on it.
- The description says the red `OpenTelemetry - Export Trace` check is red on `develop` too. The current `develop` tip is green for that check. The cause is still external (`Can't find 'action.yml' ... for action 'kestra-io/actions/actions/otel-collect@main'`), so a rerun should clear it, but the claim as written no longer holds.
- The branch is 5 commits behind `develop`.

Everything else reads well: the `onError` signature change is safe since neither `main.ts` passes a custom `onError`, the `EE` companion really is just the two `meta:` lines, `allowDeleted: true` in the `flow` resolver is correctly load-bearing, the local `KestraRequestOptions` sidesteps the `EE` `SDK` typecheck problem, and the `console.error` breadcrumbs are a clear improvement on the old silent page swap.

One question for after this lands: does the 404 behaviour need backporting to `releases/v1.3.x` and `releases/v1.0.x`, or is it 2.0-only given it builds on the colocated route tables and https://github.com/kestra-io/kestra/pull/17912?

elevatebart:
Thanks — both blocking findings were real, and both are fixed in d6cb369. Point by point:

**1. The `Errors.vue` watcher.** Dropped, rather than reordered. The guard already writes a verdict on every navigation that declares `meta.entity` and clears it on every one that does not, so the watcher was redundant *and*, as you found, destructive: it fired on the route commit right after the guard's write, so the missing entity's page mounted and refetched without `ENTITY_REQUEST_OPTIONS` — blank page plus toast. I checked the other writers before removing it: `BlueprintsBrowser` sets `coreStore.error = 404` itself and is cleared by the next navigation through the same path (blueprint routes declare no entity), and `BasicAuthLogin` clears explicitly.

**2. The superseded navigation.** Each navigation now carries a counter, and a verdict is dropped if another navigation started while the resolver was in flight:

```ts
const navigation = ++latestNavigation
const verdict = record ? await probe(resolverOf(record)!, to) : undefined

// The user has navigated again while this one was resolving, so its verdict is about a page
// they are no longer on: applying it would show the not-found screen over the newer one.
if (navigation !== latestNavigation) return true

useCoreStore().error = verdict
```

I went with the counter rather than a `router.currentRoute` comparison because `beforeResolve` runs before the commit, so `currentRoute` is still the *previous* route for both the superseded navigation and the winning one — it identifies neither. Covered by `discards the verdict of a navigation the user has already left`, which fails on the previous commit (verified by removing the guard line: 1 failed, 6 passed).

**3. The routes that lost the safety net.** Out of scope here, and filed as #18861 with your list — `admin/mcp-servers/update` in OSS plus the fifteen EE routes, with `ui-ee/src/components/roles/Role.vue` named as the clearest case. Not narrowed silently: it is now a paragraph in the description too. The reason it is a follow-up rather than an addendum is that each of those routes needs its own resolver through its own store loader, half of them in the EE repo, which is a second PR pair either way.

**4. The round trip.** Called out in the description now, as one of the details a reviewer should weigh, along with the fact that the guard has no timeout of its own. The mitigation that is already in the change is that the resolver goes through the store loader the page would have called anyway, so a successful navigation costs the same one request it did before; what is new is that the navigation waits for it instead of the page waiting.

**5. `Overview.vue`'s `KsNoData`.** I decided *not* to touch it, and the key stays. The guard is now the treatment for a missing execution, and it intercepts before the page mounts, so that `v-else` is no longer the not-found branch — it is the "the store has no execution for this route" branch, which after this PR means the fetch failed for a reason other than 404. "Execution {id} is not found." is imperfect copy for a 500, but the toast beside it says what actually happened, and a blank page would be worse than slightly wrong copy. Rewording it is a twelve-locale translation change for a branch that is now rarely reached, which I would rather not couple to this PR — happy to flip it either way (reword, or delete branch and key) if you prefer.

**6. The SSE ordering.** Fixed rather than commented. The previous execution's stream stays open until the new page mounts and follows its own, and `throttledExecutionUpdate` has a 500 ms trailing edge, so an event from the execution the user left could land on top of the guard's load. `loadExecution` now cancels the pending throttled update before writing, since it is the authoritative value:

```ts
// A trailing event from the previous execution's stream, still open until the page
// mounts and follows this one, would otherwise land on top of this load.
throttledExecutionUpdate.cancel()
execution.value = data
```

The narrower fix — dropping events whose id is not the route's — is not available: `FlowPlayground` follows `playgroundStore.latestExecution`, which is not the route's execution.

**Nits.** `existing` now resets to `true` on a successful namespace load. Worth noting it was sticky before this PR too: EE's namespace `DetailsActions` gates the delete button on it, so one missing namespace disabled deletion for every namespace visited afterwards. The guard made that easy to reach, which is how it surfaced. The OTel claim is corrected: `otel-collect` was fixed in `kestra-io/actions@main` (EE develop carries `build: fix otel-collect action name`), and develop's own run went green at 18:06 on Aug 31, after the run on this PR — so it should be green on this push rather than red on develop. `trigger-ee` came back green on its own. The branch is merged with develop as of this push.

**Backport.** 2.0 only, in my view. The toast half depends on #17912, a breaking 2.0 change: on 1.x there is no problem document to render, so a 404 toast would show a bare status with no title or detail, and the `errors.problems.*` keys do not exist there. The guard half depends on the colocated route tables this cycle introduced (`flowTabs.ts` / `executionTabs.ts` exporting route meta, the namespace parent route), so it would have to be rewritten rather than cherry-picked. And the behaviour it replaces is not a regression — a 404 has swapped the page for the not-found screen for years — so it is a UX change for the next major rather than a fix a patch release owes anyone. Say the word if a release manager disagrees and I will scope what a 1.3 version would actually involve.

---

MilosPaunovic (commented):
Round 2 at `d6cb369c12dd09ca6895d378d40b11c87a736a69`. Both blockers verified fixed against this head. Two things are still open, both in the same mechanism.

### 1. The counter guards the verdict, but not the resolver's store write

The verdict half of finding 2 is fixed. The store half is not: a superseded navigation's resolver still runs to completion and commits its entity.

```
url=/flows/fast  store.flow=slow
```

Open a slow `flow`, click a second one before it lands, and the abandoned `loadFlow` writes `flowStore.flow` after the second page has already mounted. `useFlowRoot`'s route-params watcher has fired by then and `previousFlow` is the second flow, so nothing re-fetches: the page renders the first flow's source and graph under the second flow's `URL`, and the debounced `loadDependencies` fires for the wrong one. Same shape for `executions`, where the new `throttledExecutionUpdate.cancel()` makes it slightly worse, since the late load also drops the pending update for the execution the user is actually on.

No clean one-liner. Either the resolvers return the entity and the guard commits it to the store after the freshness check, or the store loaders compare against the route before writing, the way `isFlowLoaded()` and `Overview.vue` already do. Your call which.

### 2. The early-return path does not bump `latestNavigation`

```ts
if (record && from.matched.includes(record) && hasSameParams(to, from)) return true

const navigation = ++latestNavigation
```

The early return lands before the increment, so a tab change on the flow you are already on does not invalidate an in-flight verdict for a different flow. Starting a slow navigation to a missing `flow` and switching tab on the current one before it resolves:

```
at=/flows/a/tab2   error=404
```

The not-found screen lands over the `flow` the user is still on. One line: move `const navigation = ++latestNavigation` above the early return.

### On the question you asked

`Overview.vue`'s `KsNoData`: keeping it is the right call, but the copy is now wrong for the only case that reaches it. "Execution {id} is not found." under a 500 toast contradicts the toast beside it. If a twelve-locale change is not worth coupling to this `PR`, it is worth carrying into https://github.com/kestra-io/kestra/issues/18861 rather than leaving it unrecorded.

---

The `EE` companion https://github.com/kestra-io/kestra-ee/pull/10543 is unchanged at `a2447446f3f912e70878e36e25ff3828db54fe34` and needs nothing. Agreed on 2.0 only for the backport, for the reasons you gave.

elevatebart:
Both open points are fixed in 15af106.

**2 — the early return skipping the increment.** Moved above it, as you said. Also covered now: `lets a tab change invalidate a verdict still resolving for another entity`, which fails with the increment back in its old place.

**1 — the resolver's store write.** Went with the store half, but keyed on a load sequence rather than on the route:

```ts
const load = options.store === false ? undefined : ++latestFlowLoad
…
// A load the user has navigated away from must not become the flow on screen, the same way a
// superseded search is dropped in `stores/logs.ts`.
if (options.store === false || load !== latestFlowLoad) {
    return data
}
```

Same in `loadExecution` — where the check has to sit *above* `throttledExecutionUpdate.cancel()`, exactly for the reason you named — and in the namespace `load`, which has the same shape through the guard and would otherwise leave `existing` false for the namespace the user did land on.

I did not compare against the route, even though `isFlowLoaded()` and `Overview.vue` do it at their level: `stores/playground.ts`, `executions/Topology.vue`, `ExecutionPending.vue` and the product tour all call `loadFlow` for a flow that is deliberately *not* the route's, so a route check inside the loader would silently stop storing for them. The sequence is also what `stores/logs.ts` already does for the same class of bug, so it is one pattern rather than two.

`drops a flow load the user has navigated away from` and `drops an execution load the user has navigated away from` in `tests/unit/stores/entityLoadRace.spec.ts` reproduce your `url=/flows/fast store.flow=slow` case and both fail without the change.

**What that does not cover, and I would rather say so than imply otherwise.** The sequence closes entity-to-entity navigation, not the early-return path: start a slow navigation to flow B, change tab on flow A before it lands, and B's resolver is still the newest *load*, so it writes `flowStore.flow` while the user is on A. The guard drops B's verdict but cannot un-write its commit. Closing that needs either the resolvers returning the entity for the guard to commit after its freshness check — a second function in the `meta` contract — or the pages rendering only an entity whose id matches the route. Both are bigger than this PR, and both are now written up in #18861 along with the route-comparison caveat above.

**`Overview.vue`'s copy** is in #18861 too, as its own point: with the guard intercepting before mount, that branch is only reached on a non-404 failure, where "Execution {id} is not found." contradicts the toast beside it — reword, or drop the branch and the key.

Verified on this head: OSS `check:types`, 233 files / 2179 unit tests, Storybook 68 / 254, lint clean (3 pre-existing warnings elsewhere); EE `check:types`, 127 / 985, Storybook 12 / 40. The EE companion also picked up e1732b9, which makes the reusable-inputs delete-confirm story survive a retry — its failure was a 15 s wait on a ~6 s chunk load, measured identical on develop, with numbers on kestra-io/kestra-ee#10543.

---

elevatebart:
@MilosPaunovic ready for round 3 at `15af106`.

Both points from round 2 are fixed — the load-sequence guard in `loadFlow` / `loadExecution` / the namespace `load`, and the navigation counter moved above the early return — with a test per case that fails without the change. Details in [my previous comment](https://github.com/kestra-io/kestra/pull/18807#issuecomment-5490723339); the one race that survives (the early-return path) and `Overview.vue`'s copy are both recorded in #18861 rather than left implicit.

CI is green on this head, and the EE companion kestra-io/kestra-ee#10543 is green on `e1732b9` — its `Frontend - Tests` failure was the reusable-inputs delete-confirm story, fixed there with the measurements behind it.

---

MilosPaunovic (approved):
Round 3 at `9d0529d55d`. Both items from the last round are fixed and verified against this head, so approving. One thing left, and it is scope rather than correctness.

### `fix(styles): set flex-direction to column for #app`

This is a global rule affecting every page in `OSS` and `EE`, with no explanation in the commit body or the description, on a `PR` about 404 handling that still has no screenshots.

It does fix something real. The in-flow children of `#app` besides `#app-shell` are `DocIdDisplay` (only under `?showDocId`), `PwaInstallPrompt` (a `KsAlert`) and the dev-only `SdkDriftBanner`, and in row direction each of those takes width from the shell rather than stacking above it. `#app-shell`'s own `flex: 1; min-height: 0` says the column was the intent all along. But that is a different bug, and a one-line global layout rule is exactly the kind of change that wants its own `PR` and its own before/after. Split it out, or name what it fixes in the description and confirm a page was actually looked at.

`fix(flows): report a save failure once, not twice` I have no objection to: it is a direct consequence of a 404 now toasting, and `isReportedCentrally` being the interceptor's own predicate rather than a second copy of the condition is the right shape. Worth a line in the description too, since it also changes what `onDelete` reports.

elevatebart:
Thanks for the approval. Both scope notes are now in the description rather than left implicit.

**`flex-direction: column`** — your reading matches ours, and the description now says it: `#app` is a flex container, so in row direction `DocIdDisplay`, the `PwaInstallPrompt` alert and the dev-only `SdkDriftBanner` each take width from `#app-shell` instead of stacking above it, and `#app-shell`'s own `flex: 1; min-height: 0` (`ui/src/styles/app.scss:240`) only makes sense in a column — the declaration was missing rather than new. @elevatebart wrote it and checked it against a running instance; I have not, and the description says so rather than implying otherwise. It stays here as a named drive-by, and the offer to lift it into its own PR with a before/after stands if you would rather it not ride along.

**The double-toast fix** has its own paragraph now, including the `onDelete` change you flagged: `Failed to delete flow …` no longer stacks on the global toast when the flow is already gone. The one thing worth saying beyond the diff is what `isReportedCentrally` deliberately leaves alone — a **400** and a **401** still get the editor's own toast, since the interceptor never reported those, so the fix removes a duplicate rather than a report.

The two other sites that can double the same way, `FlowsSearch`'s replace and `FlowTriggers`'s actions, are named in the description and left to #18849; the predicate is exported for exactly that cleanup.

---

MilosPaunovic (approved):
Re-approving. Still at `9d0529d55d`, unchanged since the last round, so nothing new to look at here — the `EE` companion is where this round's changes landed and it is approved there too.

The one point still open from round 3 is the `#app { flex-direction: column }` commit: not blocking, but it would be worth either splitting it out or naming in the description what it fixes, since it is a global layout rule riding along in a 404-handling `PR`.

PR #18810, opened 2026-08-28
fix(iam): move MCP management API off the tool namespace to close aut…

ammeek:
Fix path collision on mcp auth


loicmathieu (approved):
LGTM

PR #18811, opened 2026-08-28
fix(core): reject unsatisfiable cron expressions

### LINKED ISSUE #18732: Schedule trigger accepts an impossible-date cron (e.g. `0 0 30 2 *`), creating an enabled trigger that silently never fires
### Describe the issue

The `Schedule` trigger validates cron **syntax** but not cron **satisfiability**.
A cron expression that is syntactically well-formed yet can never match a real
calendar date: `0 0 30 2 *` (February 30th), `0 0 31 4 *` (April 31st),
`0 0 31 6 *` (June 31st) — passes `flows/validate` with no error, creates
successfully, and produces a trigger that is **enabled** (`disabled: false`)
with **`nextExecutionDate: null`**. The schedule therefore never runs, and
nothing tells the user their flow will never be triggered.

This behavior is inconsistent with how the same trigger handles its timezone field.
When you set timezone: "Not/AZone", validation rejects it with the message timezone '(Not/AZone)' is not a valid time-zone ID, and the create request returns HTTP 422. The cron field gets syntax validation only, so a
guaranteed-dead schedule slips through the same gate that correctly stops a bad
timezone.

## Steps to reproduce

1. Validate a flow whose Schedule trigger uses an impossible-date cron:

   ```sh
   curl -s -u 'USER:PASS' -X POST \
     'http://localhost:8080/api/v1/flows/validate' \
     -H 'Content-Type: application/x-yaml' \
     --data-binary 'id: qa-feb30
   namespace: qa.cron
   tasks:
     - id: t
       message: x
   triggers:
     - id: sched
       cron: "0 0 30 2 *"'
   ```

2. Create the same flow, then read its trigger back:

   ```sh
   # (POST the same YAML to /api/v1/flows)
   curl -s -u 'USER:PASS' \
     'http://localhost:8080/api/v1/triggers/search?size=200' \
     | jq '.results[] | select(.triggerContext.flowId=="qa-feb30")
           | {cron: .abstractTrigger.cron,
              next: .triggerContext.nextExecutionDate,
              disabled: .triggerContext.disabled}'
   ```
## Actual behavior

- `flows/validate` → `constraints: null` (accepted, no warning)
- `POST /flows` → HTTP 200
- Trigger row:

  ```json
  {"cron": "0 0 30 2 *", "next": null, "disabled": false}
  ```

  The trigger is enabled but `nextExecutionDate` is `null`; it never fires.

- Same result for `0 0 31 4 *` and `0 0 31 6 *`.

Control — the trigger's timezone field **is** semantically validated:

- `timezone: "Not/AZone"` → validate `constraints`:
  `triggers[0].timezone: timezone '(Not/AZone)' is not a valid time-zone ID`,
  create → 422.

Control — a plainly malformed cron is caught too:

- `cron: "not a cron"` → validate `constraints`: `invalid cron expression 'not a cron': Cron expression contains 3 parts but we expect one of [5]`.

## Expected behavior

An unsatisfiable cron expression should be rejected at validation time (422),
or at minimum surface a `warning` on `flows/validate` and on the created
trigger, the same way an invalid timezone is rejected. A schedule that can
provably never fire must not be silently accepted as a healthy, enabled
trigger.

### Environment

- Kestra version: 1.3.35 (OSS)

thurdass:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18732.

### ✨ Description

Rejects cron expressions that are syntactically valid but can never produce an execution date, such as `0 0 30 2 *`.

The validator reuses cron-utils `ExecutionTime.nextExecution(...)` with Kestra’s existing current-time convention. This keeps valid cases such as February 29 on leap years accepted without introducing custom calendar rules.

### 🛠️ Backend Checklist

- [ ] Code compiles and tests pass for the touched module (`./gradlew :core:test --tests "io.kestra.core.validations.ScheduleValidationTest"`)
- [x] New behavior is covered by unit tests

### 📝 Additional Notes

Only `ScheduleValidator.java` and `ScheduleValidationTest.java` were changed. `Schedule.java`, frontend files, generated files, OpenAPI/YAML, and dependencies were left untouched.

The narrow test command could not run because the environment does not have Java 25 installed.

Why did the cron cat get rejected? It had no valid date to land on. 🐱

### DIFF
--- a/core/src/main/java/io/kestra/core/validations/validator/ScheduleValidator.java
+++ b/core/src/main/java/io/kestra/core/validations/validator/ScheduleValidator.java
@@ -1,5 +1,9 @@
 package io.kestra.core.validations.validator;
 
+import com.cronutils.model.Cron;
+import com.cronutils.model.time.ExecutionTime;
+
+import io.kestra.core.scheduler.SchedulerClock;
 import io.kestra.core.validations.ScheduleValidation;
 import io.kestra.plugin.core.trigger.Schedule;
 
@@ -23,9 +27,14 @@ public class ScheduleValidator implements ConstraintValidator<ScheduleValidation
 
         if (value.getCron() != null) { // if null, the standard @NotNull will do its job
             try {
-                // parseCron() now parses + validates on first call and caches the result,
-                // so repeated validation runs on the scheduler hot path are O(1).
-                value.parseCron();
+                Cron parsed = value.parseCron();
+                if (ExecutionTime.forCron(parsed).nextExecution(SchedulerClock.now()).isEmpty()) {
+                    context.disableDefaultConstraintViolation();
+                    context.buildConstraintViolationWithTemplate(
+                        "invalid cron expression '%s': no valid execution date exists".formatted(value.getCron())
+                    ).addConstraintViolation();
+                    return false;
+                }
             } catch (IllegalArgumentException e) {
                 context.disableDefaultConstraintViolation();
                 context.buildConstraintViolationWithTemplate("invalid cron expression '" + value.getCron() + "': " + e.getMessage())
--- a/core/src/main/java/io/kestra/core/validations/validator/ScheduleValidator.java
+++ b/core/src/main/java/io/kestra/core/validations/validator/ScheduleValidator.java
@@ -1,5 +1,9 @@
 package io.kestra.core.validations.validator;
 
+import java.time.DateTimeException;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+
 import com.cronutils.model.Cron;
 import com.cronutils.model.time.ExecutionTime;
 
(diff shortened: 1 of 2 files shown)


dhanusharer:
Could we use the schedule's configured `timezone` when calling `nextExecution(...)` here?

The validator currently uses `SchedulerClock.now()` directly, so the reference `ZonedDateTime` is in the scheduler/system zone. However, `Schedule` supports its own `timezone`, and the actual scheduling path evaluates the schedule with that timezone.

Using the configured timezone here would keep the satisfiability check consistent with the timezone in which the trigger will actually execute.

thurdass:
Good point, thanks! I updated the satisfiability check to evaluate nextExecution(...) using the schedule's configured timezone, matching the timezone conversion used by the runtime scheduling path.

loicmathieu (approved):
LGTM, thanks

loicmathieu:
Backported to 2.0

PR #18816, opened 2026-08-29
feat(core): provide PurgeMetrics task and bulk metric purging

### LINKED ISSUE #18711: Provide a `PurgeMetrics` task
### Feature description

Currently, purge is done via `PurgeExecutions` which is advise to not purge logs because it's more effective to purge logs separatly via the `PurgeLogs` task.

The same issue exist with metrics, using `PurgeExecutions` to purge metrics would mean emitting a `DELETE metrics` query for each batch of executions which is not very performant.

A separate `PurgeMetrics` task would allow purging the metrics table in a single `DELETE` query, as not all executions have metrics it would avoid doing a lot of queries for nothing.

dhanusharer:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18711.

### ✨ Description

Currently, cleaning up old metrics requires running `PurgeExecutions`. This is pretty inefficient because it batches through executions and issues a `DELETE FROM metrics` query for every chunk, even when those executions don't have any metrics stored.

This PR introduces a dedicated `PurgeMetrics` task (`io.kestra.plugin.core.metric.PurgeMetrics`) and adds a direct `purge(...)` method on `MetricRepositoryInterface` / `AbstractJdbcMetricRepository`. This allows old metrics to be purged in bulk with a single `DELETE` query filtered by timestamp and namespace/flow.

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched modules (`./gradlew :core:test`, `./gradlew :jdbc-h2:test`)

### 📝 Additional Notes

- Supports namespace tree/prefix filtering when no `flowId` is given, and exact matching when `flowId` is provided (consistent with `PurgeLogs` and `PurgeStorage`).
- Standard ACL checks are in place for cross-namespace and global purges.
- Kept the existing `purge(Execution)` methods untouched so existing callers don't break.

### 🤖 AI Authors

Why was the cat sitting on the computer? To keep an eye on the mouse! 🐱


dhanusharer:
Hi @loicmathieu and @MilosPaunovic , could you please take a look and review this when you get a chance? Thanks!

core/src/main/java/io/kestra/plugin/core/metric/PurgeMetrics.java
```diff
@@ -0,0 +1,136 @@
...
+                tasks:
+                  - id: purge
+                    type: io.kestra.plugin.core.metric.PurgeMetrics
+                    namespace: company.team
+                    endDate: "{{ now() | dateAdd(-1, 'MONTHS') }}"
+                """
+        )
+    }
+)
+public class PurgeMetrics extends Task implements RunnableTask<PurgeMetrics.Output>, SystemTask {
```
loicmathieu, line 67: Can you add support for purging a single execution ID as it exists for the PurgLogs task?

```java
    @Schema(
        title = "The Execution ID of the metrics to be purged"
    )
    private Property<String> executionId;
```

I think a batchSize should also be added for MySQL:

```java
    @Schema(
        title = "The number of metric rows deleted per batch",
        description = "Only applies on MySQL. When set, deletion runs in batches of this size using `DELETE ... LIMIT n` to stay within `group_replication_transaction_size_limit`. When not set, all matching rows are deleted in a single transaction."
    )
    private Property<Integer> batchSize;
```

loicmathieu (changes requested):

dhanusharer:
Addressed the requested changes:

Added executionId support to purge metrics for a specific execution.
Added MySQL batchSize support using DELETE ... LIMIT n, following the existing PurgeLogs pattern.
Added @Min(1) validation for batchSize.
Added tests covering execution ID filtering, batch deletion, and validation.

PurgeMetricsTest (7 tests) and the relevant H2 repository tests pass successfully.

core/src/main/java/io/kestra/plugin/core/metric/PurgeMetrics.java
MilosPaunovic: If you've addressed any of the requested changes from the PR review, @dhanusharer, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

dhanusharer:
@loicmathieu could you please take a look and review this when you get a chance? Thanks!

core/src/main/java/io/kestra/plugin/core/metric/PurgeMetrics.java
```diff
@@ -0,0 +1,152 @@
...
+
+        if (renderedNamespace == null) {
+            runContext.acl().allowAllNamespaces().check();
+        } else if (!renderedNamespace.equals(flowInfo.namespace())) {
+            runContext.acl().allowNamespace(renderedNamespace).check();
+        }
+
+        String renderedStartDate = runContext.render(this.startDate).as(String.class).orElse(null);
+        ZonedDateTime rStartDate = renderedStartDate != null ? TypeConverter.toZonedDateTime(renderedStartDate) : null;
+        ZonedDateTime rEndDate = TypeConverter.toZonedDateTime(runContext.render(this.endDate).as(String.class).orElseThrow());
```
loicmathieu, line 126: Why do you use the TypeConverter? As far as I know we never do that in a plugin we either use runContext.renader().as(ZonedDateType) or parse it.

core/src/main/java/io/kestra/core/repositories/MetricRepositoryInterface.java
```diff
@@ -34,6 +34,26 @@ MetricAggregations aggregateByFlowId(String tenantId, String namespace, String f
 
     Integer purge(List<Execution> executions);
 
+    default int purge(
```
loicmathieu, line 37: I don't think this method provides any benefits as it's used in any call site

loicmathieu (changes requested):

PR #18819, opened 2026-08-30
fix(triggers): trigger state for webhook, MCP, flow and asset event triggers

### LINKED ISSUE #18379: [Bash2.0] MCP Trigger does not display under Tenant -> Triggers
### Describe the issue

Steps to reproduce:
1. Add the flow below
2. Go to Tenant ->Triggers -> Manate
3. It's blank

https://github.com/user-attachments/assets/7daa7130-4e9d-4ffa-8e34-3bb3b5a4fcdb

I added a schedule trigger to a new flow, and that does appear. The MCP trigger did not though.

Image

```yaml
id: hello_world
namespace: company.team

inputs:
  - id: user
    type: STRING
    defaults: John Doe
    description: "The name of the user to greet."

tasks:
  - id: greet
    values:
      greeting: "Helloooooo, {{ inputs.user }}!"

outputs:
  - id: greeting
    type: STRING
    value: "{{ outputs.greet.values.greeting }}"

triggers:
  - id: mcp
    type: io.kestra.plugin.core.trigger.McpToolTrigger
    toolName: hello_world
    title: Hello World greeting tool
    toolDescription: Returns a personalised greeting. Call this when the user asks for a greeting.
    mcpServer: default
```

### Environment

- Kestra Version: 2.0 RC10 EE

ammeek:
Closes https://github.com/kestra-io/kestra/issues/18379.

The problem

A flow declaring an McpToolTrigger shows it on the flow's own Triggers tab, but Tenant → Triggers → Manage lists nothing for it. The same is true of Webhook and Flow triggers. Only Schedule and polling triggers appear.

The cause

GET /triggers/search lists rows from the scheduler's trigger-state table. TriggerScheduler.onStart only created a TriggerState for triggers implementing WorkerTriggerInterface, and FlowService.recomputeTriggers only emitted TriggerCreated/TriggerUpdated events for those. Webhook, MCP and flow triggers therefore had no row at all, and nothing could reach the endpoint.

The flow page looks correct because it works the other way round: the frontend starts from flowStore.flow.triggers — the definition — and merges any state onto it.

This is not a 2.0 regression. AbstractScheduler in 1.3.x carried the same filter.

The fix

The scheduler creates a state for every declared trigger, both at startup and on flow changes — so triggers on flows that already exist become visible on upgrade, without needing the flow re-saved. Triggers the scheduler does not evaluate are typed TriggerType.UNSCHEDULED and excluded from the scheduling-eligibility queries, so they are listed but never evaluated.

The eligibility predicate is IS DISTINCT FROM 'UNSCHEDULED' rather than != 'UNSCHEDULED', and the Elasticsearch equivalent is a must_not term, because V2_0_03TriggerMigration leaves type unset on rows migrated from 1.x. Those rows are scheduler-evaluated triggers and must stay eligible — requiring a known type instead of excluding one would stop every schedule firing after an upgrade.

No frontend change: the page already renders whatever the endpoint returns.

What implementing it uncovered

Each of the following is a direct consequence of these triggers now having a state row, and each would have shipped as a visible defect if the listing had gone in alone.

The enable/disable switch would have lied. The page renders a toggle per row, writing TriggerState.disabled. The scheduler honours that flag for what it evaluates, but nothing consulted it for these four kinds — they only ever read the flow source's disabled. Each now resolves it at the point it fires: ExecutionController returns 409 for a webhook, McpToolService filters the tool from tools/list and re-checks per call (the tool specification is cached and only refreshed on flow changes, so the listing check alone would leave a held tool callable), FlowTriggerService filters flow triggers, and EE's AssetTriggerService filters asset event triggers. Every read sits after the condition and server filters, so it costs a lookup only for triggers that would actually fire.

Backfill became silently accepted and unrunnable. Backfilling a webhook trigger used to 404 for want of a state. With a state it was accepted and stored, but the eligibility query excludes UNSCHEDULED, so it could never run. createBackfill now rejects with 409, mirroring the existing realtime guard in unlockTriggerById.

stopAfter began silently disabling MCP and asset triggers. DefaultExecutor.sendTriggerExecutionTerminated excluded Webhook, Flow and Subflow by class name but never McpToolTrigger, and EE's asset EventTrigger was never in the list either. That was harmless while those triggers had no state — the handler logged "state not found" and returned — but with a state it runs updateOnExecutionTerminated, which applies stopAfter. The hardcoded list is replaced by resolving the trigger from the flow and checking its TriggerType, which covers both gaps and cannot go stale as trigger kinds are added.

That type check nearly stranded polling and realtime triggers. Its first form treated an unresolvable trigger as "do not release the lock". FlowWithException carries no triggers, so a polling or realtime plugin trigger whose execution terminated while its flow no longer parsed — an uninstalled plugin, say — would never have released its lock, leaving it permanently unschedulable. It falls back to releasing and skips only on a positive identification as unscheduled, since a lock left held is far worse than a redundant release.

The new rows reported "never fired". lastTriggeredDate and executionId are written only by the scheduler, so a listed row that never updates tells the user a webhook has never fired no matter how often it has. A new UnscheduledTriggerFired event carries the firing back from every place these triggers create an execution — including the webhook path that creates a FAILED execution when evaluation throws, which otherwise left the state reporting a stale firing. It is an event rather than a direct write because TriggerState has a single writer, TriggerEventHandler, which enforces event ordering via lastEventId and is routed to the scheduler owning that trigger's vNode. The existing TriggerEvaluated could not be reused: its handler publishes an execution, so it would have created a second one.

Two reads got hot enough to matter. The disabled check now runs per flow-trigger candidate on every terminated execution and on every webhook and MCP call, so AbstractJdbcTriggerRepository overrides the interface default — a full row fetch and Jackson deserialize — with a select of the generated disabled column. And unscheduled rows have next_evaluation_epoch and locked null, so they satisfy the leading predicates of idx_trigger_scheduler and were rejected only by the new, unindexed type predicate: one heap fetch per unscheduled row per scheduling loop. Migration 2.0.25-trigger-scheduler-index recreates that index as (vnode, next_evaluation_epoch, locked, type) so the rejection is index-only. It replaces the index rather than adding one, so the secondary-index count is unchanged.

Scope

Listing these triggers is what exposed the toggle, backfill, the stopAfter path, the empty columns and both hot reads; none were reachable before. Shipping only the listing would have shipped a page with controls that silently do nothing and two columns that permanently read "never".


fhussonnois (changes requested):
I'm sorry; I can't accept this PR; 

**Why**

* By design: The responsibility of the `Scheduler` is to manage `Schedulable`, `Polling`, and `RealTime`  triggers which all have a dedicated state; But Webhook, Flow, and MCP triggers are of a completely different nature; e.g. there is no concept of `next_evaluation_date` or `locked` state, and thus should not be managed by the Scheduler.

The GitHub issue defines something as a “bug,” when in fact it is a complete redesign purpose of the page; the “Tenant > Trigger” page was designed to display only triggers with a status; however, “FlowTrigger” and “Webhook” triggers were not listed on this page even before version 2.0.

Well, so now we're trying to put everything into a single model, but it's not quite that simple.

**Criticals**
- The Executor cannot afford to have to look up the trigger table to determine whether a flow trigger is disabled or not.
 
- The Scheduler can't handle the new created `UnscheduledTriggerFired` event. For example, too many webhook requests could unnecessarily overload the Scheduler's event loop, affecting all triggers.

**Proposal**

* `FlowTrigger` / `Webook` / `MCPTrigger` can't be disabled via the trigger state repository; The user must edit the flow YAML and we must have a WARN/Tooltip in the UI - so no change on the Executor.
 
* The `FlowService` must create/delete the TriggerState for `UNSCHEDULED` trigger type directly on flow mutations.

* The migration script should create the missing states for existing Flow/MCP/webhooh trigger trigger.

* Drop the `UnscheduledTriggerFired` concept; those triggers are comparable to the realtimetrigger and we don't update the state for each execution submittion;

loicmathieu:
This is not a good idea to do that for webhooks and flow triggers, as they are performance-sensitive and you added one database roundtrip.

Especially for flow triggers that can be evaluated literally thousands of times per second!

However, I must admit that this whole disabled-in-DB and in the YAML is a pain, IMHO we should not have done that, but now that it's here, see https://github.com/kestra-io/kestra-ee/issues/10422

webserver/src/main/java/io/kestra/webserver/services/TriggerStateService.java
```diff
@@ -179,11 +181,16 @@ public TriggerState resetTrigger(final TriggerId triggerId) throws NotFoundExcep
...
      *                                  accept and then immediately discard, or if the trigger cannot be backfilled.
      * @throws NotFoundException if the trigger does not exist.
-     * @throws ConflictException if the backfill cannot be created.
+     * @throws ConflictException if the trigger is one the scheduler does not evaluate, or if the backfill
+     *         cannot be created.
      */
     public TriggerState createBackfill(TriggerId triggerId, CreateBackfillTrigger.Backfill backfill) throws NotFoundException, ConflictException {
         validateBackfillWindow(backfill);
-        validateBackfillable(triggerId, getTriggerState(triggerId));
+        TriggerState state = getTriggerState(triggerId);
```
fhussonnois: The role of the `validateBackfillable` is to perform the check so I don't understand the code you added, probably an issue while rebasing - your code is not needed anymore here

webserver/src/main/java/io/kestra/webserver/services/TriggerStateService.java
```diff
@@ -317,10 +324,14 @@ public ApiAsyncOperationResponse deleteAllMatching(String tenant, List<QueryFilt
      * @param recoverMissedSchedules when {@code true}, missed schedules are recovered on enable according to the
      *                               trigger's own configuration; {@code null} or {@code false} means they are skipped.
      * @throws NotFoundException if the flow or trigger does not exist.
-     * @throws ConflictException if the change failed.
+     * @throws ConflictException if the trigger is one the scheduler does not evaluate, or if the change failed.
      */
     public TriggerState toggleTriggerById(TriggerId trigger, boolean disabled, @Nullable Boolean recoverMissedSchedules) throws NotFoundException, ConflictException {
-        validateToggleable(trigger);
+        if (!isEvaluatedByScheduler(findDeclaredTrigger(trigger))) {
```
fhussonnois: Same; why did you drop `validateToggleable` method ? Add you check into that method.

webserver/src/test/java/io/kestra/webserver/controllers/api/TriggerControllerTest.java
```diff
@@ -74,34 +75,41 @@ class TriggerControllerTest {
...
-    @Inject
-    JdbcTestUtils jdbcTestUtils;
-
     @Inject
     Scheduler scheduler;
 
     @Inject
     SchedulerConfiguration schedulerConfiguration;
 
+    // Every test here scopes itself to a namespace of its own, so the tables are deliberately not
```
fhussonnois: drop this comment - low-value

fhussonnois (changes requested):
For me the changes on the Executor are not required by your fix; revert changes on ExecutorCore

PR #18820, opened 2026-08-30
fix(apps): reject out-of-range pagination on search and catalog with a 422

ammeek:
Closes: https://github.com/kestra-io/kestra-ee/issues/10206

### What

`/apps/search` and `/apps/catalog` answered HTTP 500 on out-of-range pagination, and the `/apps/search` body carried the raw SQL statement that failed ("OFFSET must not be negative", with table and column names). Reported against rc-10.

### Root cause

Three findings, one of them already resolved:

1. **`page=0` is already fixed on develop.** Commit `f80bb93165` added a `page < 1` guard to `PageableUtils.from` when closing the sibling issue #10223, so the exact reproducer in the ticket returns 422 today. The ticket is stale on that point only.
2. **`size` was never validated.** `AbstractJdbcRepository` maps a pageable to `.limit(size).offset(offset - size)`, so `?size=-10` becomes a negative `LIMIT` on any list endpoint and the query throws; `/apps/catalog?size=-10` dies earlier still, in `ArrayListTotal.of` on `subList(0, -10)`. Same defect as the ticket, reached through the other parameter.
3. **The leak is one layer up, and not specific to pagination.** `ErrorController.jsonError` appends `e.getMessage()` to the response, and a jOOQ `DataAccessException` embeds the whole statement in its message — so *any* SQL failure exposed schema to the caller.

Apps was also the odd one out: Assets, Cases, Policy, TestSuite, and PromotionTarget all annotate `page`/`size` with `@Min(1)`, which is why the ticket's siblings answered a clean 422 and Apps did not.

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/ErrorController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/ErrorController.java
@@ -2,7 +2,6 @@ package io.kestra.webserver.controllers;
 
 import java.io.FileNotFoundException;
 import java.lang.reflect.Field;
-import java.sql.SQLException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedList;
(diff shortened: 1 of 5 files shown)


brian-mulier-p (approved):
LGTM

PR #18821, opened 2026-08-30
fix(namespaces): clean 404 on directory reads and 409 for writes under a file

ammeek:
Closes [https://github.com/kestra-io/kestra-ee/issues/10325](https://github.com/kestra-io/kestra-ee/issues/10325).

### ✨ Description

Two namespace-file error responses were wrong, and both are reachable from the Files panel:

1. **Directory read error:** `GET /namespaces/{ns}/files?path=/` asked the storage layer for the content of a directory. The JDK's own error came back to the caller as `404 Not Found: /app/storage/{tenant}/{ns}/_files (Is a directory)`, exposing the storage root, the tenant ID, and the namespace-to-directory mapping.
2. **Path collision error:** Creating a file whose parent path is an existing file (`/t.txt/child.txt`) reached the storage layer too, where `Files.createDirectories` cannot make a directory out of a file — returning a 500 carrying the same absolute path for what is actually a client error.

`InternalNamespace` now answers both cases itself:

* A directory read raises the same namespace-relative not-found used for a missing file, keeping the status unchanged while updating only the message.
* A write (file or directory) whose path runs through an existing file raises a `ConflictException` (409). The ancestor check uses a single `findByPaths` lookup of the path's ancestors without a trailing slash (since only directories are stored with one, any hit is a file). This also covers deeper paths like `/t.txt/a/b.txt`, which would otherwise bypass the immediate parent check and fail later in `mkDirs`.

`LocalStorage` no longer lets an absolute path escape in its own exceptions either, providing defense in depth for every other caller: the existing sanitization now covers both reads and parent-directory creation on write. The `FileAlreadyExistsException` from #17093 is preserved, naming the storage URI instead of the server path.

### DIFF
--- a/core/src/main/java/io/kestra/core/storages/InternalNamespace.java
+++ b/core/src/main/java/io/kestra/core/storages/InternalNamespace.java
@@ -260,6 +260,12 @@ public class InternalNamespace implements Namespace {
         // Throw if file not found OR if it's deleted
         NamespaceFileMetadata namespaceFileMetadata = findByPath(normalizedPath, revision).orElseThrow(() -> fileNotFound(normalizedPath, revision));
 
+        if (namespaceFileMetadata.isDirectory()) {
+            throw new FileNotFoundException(
+                "'%s' is a directory in namespace '%s', it has no content to read.".formatted(normalizedPath, namespace)
+            );
+        }
+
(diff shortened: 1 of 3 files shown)


Skraye (approved):
LGTM

PR #18822, opened 2026-08-31
fix(core): mask secrets in log-to-file exception traces

### LINKED ISSUE #18737: Secrets not hidden or replaced while using logToFile: true
### Describe the issue

When using logToFile then previewing the content of the logs filed secrets are hidden sometimes while not at other. In my case, hidden in the error log but not in the trace.

Image

here's a very simple flow that may reproduce this issue :
``` yaml
id: failling_flow
namespace: company.team

inputs:
  - id: some_secret
    type: SECRET
    defaults: "should not be visble"

tasks:
  - id: hello
    logToFile: true
    message: Hello World! 🚀

  - id: fail
    logToFile: true # <---
    errorMessage: "Task failure {{ inputs.some_secret }}" 
```

### Environment

- Kestra Version: 1.3.33

ZHCHJ888:
Closes https://github.com/kestra-io/kestra/issues/18737.

---

### 🔗 Related Issue

When `logToFile: true` is enabled, the log statement is masked before formatting, but Logback appends the attached throwable during `PatternLayout.doLayout(e)`. Secrets contained in exception messages or stack traces can therefore reach the generated log file unmasked.

### ✨ Description

Applies the existing secret replacement pass to the final rendered file-log output, following the same post-render masking approach used by `ContextAppender` in #18283. This preserves the exception trace while preventing registered secrets from being written to the log file.

Adds a regression test covering a secret in both the formatted log message and the attached throwable output.

### 🛠️ Backend Checklist

- [ ] Code compiles and tests pass for the touched modules (`./gradlew :module-name:test`, or `./gradlew build` for cross-module changes)

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/RunContextLogger.java
+++ b/core/src/main/java/io/kestra/core/runners/RunContextLogger.java
@@ -517,7 +517,7 @@ public class RunContextLogger implements Supplier<org.slf4j.Logger> {
             e = this.transform(e);
 
             try {
-                String line = PATTERN_LAYOUT.doLayout(e);
+                String line = replaceSecret(PATTERN_LAYOUT.doLayout(e));
                 fileOS.write(line.getBytes());
             } catch (IOException ex) {
                 // silently do nothing, the message will still be forwarded to the server log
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
LGTM, thanks

loicmathieu:
Backported to 2.0 and 1.3

PR #17480, opened 2026-07-17
refactor(topology): remove drag and drop task swapping functionality

RemiBarthe:
closes https://github.com/kestra-io/kestra-ee/issues/8883

This pull request removes the drag-and-drop task swapping functionality from the topology editor, along with related UI feedback and event handling. It also cleans up unused imports and event listeners, and ensures that nodes are no longer draggable in the graph. These changes simplify the codebase and user interactions by eliminating the ability to swap tasks via drag-and-drop.

**Removal of drag-and-drop task swapping:**
- Removed all logic, event handlers, and UI feedback related to dragging nodes to swap tasks in `Topology.vue`, including the `dragging` state, drag event listeners, and swap validation logic. [[1]](diffhunk://#diff-d4db08ff92e481eaba44480fd7e89fa6e2883ee7cbe6ec74c0f407006b410431L220-R218) [[2]](diffhunk://#diff-d4db08ff92e481eaba44480fd7e89fa6e2883ee7cbe6ec74c0f407006b410431L383-L482)
- Removed the `swapped-task` and `message` events and their handlers from `LowCodeEditor.vue` and `LowCodeEditorWrapper.vue`, as well as the related code for updating expanded subflows after a swap. [[1]](diffhunk://#diff-4e12324e4c8758c01268ce4076a02c7cf9bc95fbb4fba4116c9b8bb1ccc38da5L39-L40) [[2]](diffhunk://#diff-4e12324e4c8758c01268ce4076a02c7cf9bc95fbb4fba4116c9b8bb1ccc38da5L483) [[3]](diffhunk://#diff-4e12324e4c8758c01268ce4076a02c7cf9bc95fbb4fba4116c9b8bb1ccc38da5L796-L808) [[4]](diffhunk://#diff-6b8fc3e246a6b6578f234473000b0ff16fe3ea1dadd5d1d033e44400c968e2fbL15) [[5]](diffhunk://#diff-6b8fc3e246a6b6578f234473000b0ff16fe3ea1dadd5d1d033e44400c968e2fbL57-L84)

**UI and style adjustments:**
- Removed the `topology-node-drop-target` CSS class and its visual feedback from `topology.scss`.
[...]

PR #18827, opened 2026-08-31
fix(topology): refresh graph when expanding subflows

### LINKED ISSUE #18818: Flow Editor Topology "Expand" action does not display expanded subflow until YAML modification
### Summary

In the Flow Editor Topology tab (`Flows -> Edit -> Topology`), clicking the "Expand" action on a `Subflow` node does not expand the subflow or display the child flow's inlined tasks. The canvas remains collapsed until the user modifies the flow YAML, which triggers an indirect graph refetch.

---
### Steps to Reproduce
1. Create and save a child flow:
   ```yaml
   id: subflow-child-demo
   namespace: company.team

   tasks:
     - id: hello
       message: Hello World
   ```

2. Create and save a parent flow referencing the child flow:
   ```yaml
   id: subflow-parent-demo
   namespace: company.team

   tasks:
     - id: child
       namespace: company.team
       flowId: subflow-child-demo
   ```
3. Open Flows -> Edit for `subflow-parent-demo` and navigate to the Topology panel.
4. Click the context menu on the `child` node and select Expand.
---

### Expected Behavior

Clicking "Expand" should query the backend graph API (`/api/v1/{tenant}/flows/graph/source`) with `subflows=child` and render the expanded child tasks inside a cluster on the canvas, matching the behavior of the Execution Topology page.

---

### Actual Behavior

Clicking "Expand" emits the `@expand-subflow` event and updates `flowStore.expandedSubflows`, but never triggers a graph reload. The UI graph remains static. 

If the user modifies any character in the YAML editor, `onEdit` fires and calls `flowStore.fetchGraph()`, which finally queries the server with `?subflows=child` and renders the inlined cluster.

---

### Root Cause

In `ui/src/components/inputs/LowCodeEditorWrapper.vue`:

```typescript
const onExpandSubflow = (subflows: string[]) => {
    flowStore.expandedSubflows = subflows
}
```

The handler updates the store but forgets to call `flowStore.fetchGraph()`.

In contrast, `ui/src/components/executions/Topology.vue` handles this correctly by calling `loadAugmentedGraph()` whenever `expandedSubflows` changes.

---

### Proposed Fix

Update `onExpandSubflow` in `LowCodeEditorWrapper.vue` to fetch the updated graph with loading state management:

```typescript
const onExpandSubflow = async (subflows: string[]) => {
    isLoading.value = true
    flowStore.expandedSubflows = subflows
    try {
        await flowStore.fetchGraph()
    } catch (error) {
        console.error("Failed to fetch expanded subflow graph:", error)
    } finally {
        isLoading.value = false
    }
}
```

### Environment

- Kestra Version: develop

ayush-singh-0601:
Closes https://github.com/kestra-io/kestra/issues/18818.

### 🔗 Related Issue

The Flow Editor updated its expanded-subflow selection but did not request a new topology graph, so the child tasks only appeared after an unrelated YAML edit triggered a refresh.

### ✨ Description

This change exposes the flow store's existing graph refresh action and awaits it immediately after the Expand event updates the selected subflows. The topology keeps its loading state active for the request and clears it whether the request succeeds or fails.

A focused regression test emits Expand, verifies the graph refresh starts with the new selection, and checks that loading remains active until the request settles.

### 🎨 Frontend Checklist

- [x] Type checking passes
- [ ] Full unit test command passes locally
- [ ] Screenshot or video recording attached

### 📝 Additional Notes

The focused regression test passes. The default full unit run reached 2,088 of 2,089 passing tests before an unrelated PluginListWrapper test exceeded its five-second timeout; that exact test passes in isolation. Other existing dynamic-import tests also exceed their per-test budgets on this Windows host when run separately. Lint and the production build pass.

I could not capture honest live UI evidence on this machine because Docker is unavailable and the installed JDK is 20 while this checkout requires Java 25. CI and a reviewer running the normal development stack remain the full-suite and live-UI gates.

### 🤖 AI Authors

Cat review note: the topology now reacts on Expand instead of waiting for someone to edit the meow-ML. 🐱


ayush-singh-0601:
Hi Kestra team — this is ready for review. The fix keeps the existing expanded-subflow state flow, then refreshes the source graph immediately while the topology shows its loading state. I also added a focused regression test for the Expand event-to-refetch path.

When you have a chance, could you please take a look? Thanks!

ayush-singh-0601:
@flcarre can you please tell what to do with mentioned pull request, does it mean it was fixed in the pr or do i need to do something

flcarre (commented):
Thanks for tracking this down — the diagnosis is right and the fix is minimal. I confirmed with `git log -L` that `onExpandSubflow` has **never** called `fetchGraph`: it survived both https://github.com/kestra-io/kestra/pull/17480 and https://github.com/kestra-io/kestra/pull/15984 untouched, so this is a genuine omission rather than a deliberate removal for performance.

The mechanism works — `fetchGraph()` → `loadGraphFromSource()` → `flowGraph.value = data` → the `flowGraph` computed → the `:flowGraph` prop. The blast radius is small: `fetchGraph` was previously private to the store, and this handler is its only new consumer. No watcher was added, so the deep-watch trap doesn't apply, and no loop is possible since `fetchGraph` doesn't write `expandedSubflows`. Design-system compliance is clean (the change is script-only).

Three things I'd like resolved before merge.

## 1. A failed expand leaves the store claiming the subflow is expanded

The `catch` only logs. `flowStore.expandedSubflows` still contains the subflow, and it's passed back down as `:expandedSubflows`, so the node reports itself expanded while nothing rendered.

This matters more than it looks: `expandedSubflows` is never reset anywhere in `stores/flow.ts`, and the editor path only ever appends, so every later `fetchGraph` in the session re-sends the failing value and re-fires the toast on each debounced YAML edit.

We already have this handler done fully, in `ui/src/components/executions/Topology.vue` — it restores `previousExpandedSubflows` in the `catch`. Same shape here:

```ts
const onExpandSubflow = async (subflows: string[]) => {
    const previous = flowStore.expandedSubflows
    isLoading.value = true
    flowStore.expandedSubflows = subflows
    try {
        await flowStore.fetchGraph()
    } catch (error) {
        flowStore.expandedSubflows = previous
        console.error("Failed to fetch expanded subflow graph:", error)
    } finally {
        isLoading.value = false
    }
}
```

## 2. No error state for anything that isn't a 404/422

`loadGraphFromSource` only surfaces a toast for `[404, 422]` *with* subflows, and unlike `loadGraph` it doesn't set `invalidGraph`. So on a 500 or a dropped connection the spinner flashes and nothing is explained. `ui/AGENTS.md` asks for all four async states. The revert above handles correctness; this needs a user-visible error for the remaining cases — with an i18n key, please, not a literal.

## 3. The loading overlay has never actually rendered before

`LowCodeEditor.vue` declares a `"loading"` emit that nothing emits — there are zero `emit("loading")` call sites — so `isLoading` has been permanently `false` until now. This PR lights up `v-ks-loading` on `#topologyWrapper` for the first time, and that directive forces `position: relative` on the `.vue-flow` container, whose children are absolutely positioned.

That may well be fine, but it's exactly the kind of thing only a screenshot proves. I saw your note about not being able to build (JDK 20 vs Java 25) — that's understood, and an honest thing to write down. If the toolchain is the blocker, the nightly Docker image is enough to reproduce and record this without building the backend. Please check dark mode too.

## Worth separate issues, not yours to fix here

- `LowCodeEditor.vue` declares a `"loading"` emit that nothing emits, so `loadingState()` in the wrapper is dead code.
- `flowStore.expandedSubflows` is never reset, including on flow navigation, and there's no collapse path from the editor.
- `fetchGraph`'s `config.validateStatus` is silently ignored — `loadGraphFromSource` reads only `config.params.subflows`.

flcarre (changes requested):
Formalising the gate from my detailed review above. The diagnosis and the fix are correct — `onExpandSubflow` genuinely never called `fetchGraph`, and the blast radius is small. Three things before merge:

**1. A failed expand must revert `expandedSubflows`.** Right now the `catch` only logs, so the store still lists the subflow as expanded while nothing rendered. Because `expandedSubflows` is never reset anywhere in `stores/flow.ts` and the editor only appends, that bad value re-poisons every later `fetchGraph` in the session and re-fires the toast on each debounced edit. `ui/src/components/executions/Topology.vue` already does this correctly — restore the previous value in the `catch`.

**2. An error state for anything that isn't a 404/422.** `loadGraphFromSource` only toasts for `[404, 422]` with subflows and doesn't set `invalidGraph`, so a 500 or a dropped connection gives a spinner flash and no explanation. Needs a user-visible error, via an i18n key rather than a literal.

**3. A screenshot, in light and dark mode.** This PR is the first thing ever to set `isLoading` to `true` — `LowCodeEditor.vue` declares a `"loading"` emit that nothing emits, so the value has been permanently `false` until now. `v-ks-loading` forces `position: relative` on the `.vue-flow` container, whose children are absolutely positioned, so the overlay needs to be seen rather than reasoned about. I understand the JDK constraint you noted; the nightly Docker image is enough to record this without building the backend.

CI is green on this head, incidentally — the only red job is `OpenTelemetry - Export Trace`, which fails on a missing action in `kestra-io/actions` and is unrelated to your change.

ayush-singh-0601:
Addressed review items 1 and 2 in `0fab99d`:

- Failed subflow expansion now restores the previous `expandedSubflows` value.
- Failures outside the store-handled 404/422 cases now show a translated retry error; 404/422 keep the existing store notification without a duplicate toast.
- The focused regression spec now covers loading/refetch, rollback plus generic error feedback, and the handled-404 path.

Validation completed:
- `npm run check:types`
- `npm run test:unit -- src/components/inputs/LowCodeEditorWrapper.spec.ts` (3 passed)
- `npm run translations:check`
- `npm run lint` (0 errors)

I could not capture the requested light/dark screenshots in this session because no browser connection is available, so that visual evidence is still outstanding. The fork branch is at `0fab99d`; GitHub is currently still showing the previous PR head while it synchronizes.

ayush-singh-0601:
@flcarre I have addressed the requested rollback and translated error-state changes in `0fab99d`, with focused regression coverage. Could you please re-review the update when you have a chance? The requested light/dark screenshots are still pending.

ayush-singh-0601:
@flcarre Requested visual evidence for the loading overlay, captured with Playwright in a component harness using the real `LowCodeEditorWrapper.vue` and Kestra `vKsLoading` directive. The graph request is deliberately held pending, and the graph fixture retains absolutely positioned nodes to exercise the layout concern.

**Light mode**

Subflow expansion loading overlay in light mode

**Dark mode**

Subflow expansion loading overlay in dark mode

The loading mask and spinner remain centered in both themes, the graph nodes stay aligned under the overlay, and the dark-theme surface and mask render correctly. Could you please re-review when convenient?

flcarre (approved):
Approving on `631f43d`.

Items 1 and 2 are done, and done in the shape I was hoping for.

* The `catch` restores the previous value, and I checked the emit site: `Topology.vue` sends `[...props.expandedSubflows, expandData.id]`, a fresh array, so the captured reference really is the old value and the rollback is not a no-op.
* The error path toasts through an i18n key, propagated to the thirteen locale files, and suppresses itself on 404/422 so the store's own notification is not doubled.
* The three specs cover the loading window, the rollback plus generic error, and the handled 404. The rollback one fails on the previous head.

One thing I flagged and checked myself: `(error as {status?: number}).status` looked like it might be reading a property that only exists on the axios response, but `kestraHttp.ts` flattens `status` onto the error at the top level, and `loadGraphFromSource` already branches on `error.status` the same way. It is correct and consistent with the surrounding code.

## Withdrawing my third point

I asked for a screenshot on the grounds that this PR is the first thing ever to set `isLoading` true, and that `v-ks-loading` forcing `position: relative` might disturb `.vue-flow`, whose children are absolutely positioned.

That concern was wrong, and I should have checked before asking. `#topologyWrapper` carries `class="vue-flow"`, and `@vue-flow/core`'s own `style.css` opens with:

```css
.vue-flow {
  position: relative;
  ...
}
```

The element is already positioned, so the directive changes nothing. There was no layout risk to photograph. Sorry for sending you after it.

For the record, since it may matter on a future PR: the captures you posted come from a synthetic harness rather than the running editor, with hand-built FLOW/SUBFLOW/TASK cards and no Kestra chrome, so they would not have settled the question either. When a visual point does come up later and the toolchain is in the way, the nightly Docker image runs the real editor without a local backend build. Not needed here.

## Not blocking

CI has not run on this head: `631f43d` has zero check runs. That needs to be triggered before the merge button does anything, but it is unrelated to the change.

The three follow-ups from my first pass (the dead `loading` emit, `expandedSubflows` never being reset, `fetchGraph`'s ignored `validateStatus`) are still worth separate issues, and still not yours to carry here.

ayush-singh-0601:
Thanks @flcarre for the detailed follow-up and for re-checking the overlay concern. I re-verified the current 631f43d head locally: the focused LowCodeEditorWrapper suite passes (3/3), translations are clean, and lint completes with no errors. Since there are still no check runs on this head, could a maintainer please trigger or approve the CI workflows when convenient?

PR #18829, opened 2026-08-31
fix(variables): show and keep the current expiration when editing a KV pair

### LINKED ISSUE #14825: KV Store Update with existing expiry is not displayed
### Describe the issue

This KV Pair has an expiry
Image

When I go to update it, the expiration box suggests it has no expiry. I would assume it would mention there is an expiry that could be overwritten.
Image

Steps to reproduce:
- Create a new KV Pair with an expiry
- Save
- Update it and see the expiration box is set to "Never"

### Environment

- Kestra Version: rc-0 1.3 EE

MilosPaunovic:
Closes https://github.com/kestra-io/kestra/issues/14825.
Closes https://github.com/kestra-io/kestra-ee/issues/6301.

### ✨ Description

Editing a `KV` pair that has an expiration gave no sign of it: the update `API` only takes a `TTL` duration, so the drawer showed the raw remaining ISO duration with no absolute date, and saving recalculated the expiration from the moment the drawer was opened - or silently removed it on older versions.

The edit drawer now shows a hint under the expiration field with the stored absolute expiration date (formatted in the user's timezone and date format), stating that leaving the field untouched keeps it, picking a duration recalculates it from now, and clearing it removes it. On save, an untouched `TTL` is recomputed from the stored expiration date instead of reusing the value prefilled at drawer-open time, so updating the value or description of a pair no longer shifts its expiration by however long the drawer stayed open - the expiration is preserved to the second.

Covered by a new `Playwright` e2e spec: editing an entry with an expiration shows the hint and keeps the expiration on an untouched save, and editing an entry without one shows no hint.

| State | Screenshot |
|-------|------------|
| Before | before |
| After | after |
| After (dark) | after dark |

### 🎨 Frontend Checklist


Pradumnasaraf (approved):
Looks good!

Image

PR #18831, opened 2026-08-31
fix(executions): labels column overflow

RemiBarthe:
closes https://github.com/kestra-io/kestra-ee/issues/10200

Capture d’écran 2026-08-31 à 13 05 34

This pull request enhances the display of labels in the UI, particularly in tables and the reusable `Labels` component. The main improvement is the addition of a "label overflow" mechanism that limits the number of visible labels and consolidates extra labels into a popover, making tables cleaner and more user-friendly.

**Label display improvements:**

* The `Labels` component now accepts a `max` prop to limit the number of visible labels; excess labels are hidden and displayed in a popover, improving readability in dense table views. A single extra label is shown rather than collapsed, since a `+1` popover costs more room than the label it hides.
* Long label values are now truncated individually with an ellipsis. Capping the count alone was not enough: the table runs with `table-layout: auto`, so one ~100-character value still drove the column past the viewport whenever it landed in the first three.
* The label rendering logic has been updated to show a tooltip with the full label text and to handle labels with and without keys more consistently.

**Table integration:**

* The `Executions.vue` and `Flows.vue` tables now use the `max=3` prop for their label columns, so only up to three labels are shown directly, with any additional labels accessible via the overflow popover.

**Accessibility:**

* The overflow chip is a real `<button>` and the popover opens on click, so keyboard users can reach the hidden labels. `KsTag` and `KsCheckTag` both render a non-focusable `<span>`, and the hidden labels are interactive filter controls rather than decoration.

**UI/UX improvements:**

* Added CSS so long values wrap inside the overflow popover instead of bleeding out of it, scoped to the popover itself rather than to every `wrap` caller — the EE audit-log Details column is the only other one, and it is left untouched.
* The overflow chip now shares the label chip styling, so it renders in the same shade, and no longer shrinks, so it is not the first thing clipped when the cell is constrained.

**Table column configuration:**

* The "state" column now comes right after "id", so it stays visible whatever the other columns contain.
* The column order is only persisted once the user actually reorders a column. `useLocalStorage` was writing the default order on first load, which made a stored order indistinguishable from a default and pinned every existing user to the old order; the picker then read that stale order and rewrote it on any column toggle. The table now renders from the column order rather than from the visibility list, so the two can no longer disagree.
* **Migration, user-visible:** the order key moves from `ks-column-order-*` to `ks-column-order-v2-*`, so the entries the old code wrote on its own are ignored, and the stale ones are removed rather than left orphaned. This resets the saved drag-and-drop column order once for **every table with a column picker** — 8 in OSS (executions, flows, flow triggers, execution metrics, secrets, triggers, MCP tool flows, KV) and 19 in EE — not only executions. Hidden and shown column choices are preserved everywhere.

**Tests:**

* A Storybook story pins the slicing: `max=3` with 5 labels renders 3 chips and a `+2`, exactly 4 labels renders all 4 with no chip, and no `max` renders everything.

`Labels` is shared, so the individual truncation also applies to the secrets table, the dashboard drill-down drawer, and the EE assets table.


MilosPaunovic (changes requested):
Thanks @RemiBarthe, the overflow mechanism itself is solid - no existing caller regresses, the nested `Labels` in the popover cannot recurse, and clicks on overflowed labels still apply the filter without bubbling into the row. Two issues block this for me though, plus a few smaller points.

**1. (major) The `state` column reorder never reaches existing users, and the column picker reverts it**

`useTableColumns` persists the column order in `ks-column-order-displayExecutionsColumns` on first load (vueuse `useLocalStorage` writes defaults), and its validity check only compares membership, not order - so every existing user keeps the old order forever. Worse: the *Customize columns* panel reads that stale order, and the moment a user toggles any column it rewrites `columns_displayExecutionsColumns` in the old order, snapping `state` back permanently, with table and picker disagreeing in the meantime. Suggested fix: bump the storage keys in `ui/src/utils/constants.ts` (e.g. `displayExecutionsColumns` -> `displayExecutionsColumnsV2`, same for `displayFlowExecutionsColumns`) or add a one-time migration - and note there are two `useTableColumns` copies sharing the same localStorage keys (`ui/src/composables/useTableColumns.ts` and the `design-system` one), so both are affected.

**2. (major) Long label values are still not truncated individually**

The issue asks to cap the count *and* truncate long values individually, but only the count is capped. Chips keep `white-space: nowrap` with no `max-width`, and `KsDataTable` runs with `table-layout: auto`, so a single ~100-char label value still drives the column past the viewport. The issue's reproducer happens to place its long values 8th and 10th, so it will look fixed - reorder the same labels and it reproduces. This also makes the new `:title` dead weight today, since it duplicates text that is always fully visible. Suggested fix: cap the chip in non-wrap mode, e.g. `max-width` + `overflow: hidden; text-overflow: ellipsis` on the chip content (that is also what makes the tooltip earn its place).

**3. (minor) The `+N` chip can be clipped and is styled off-shade**

It sits last in the `flex-wrap: nowrap; overflow: hidden` container with default `flex-shrink: 1`, so whenever the cell is actually constrained the first thing clipped is the only affordance for the hidden labels - it needs `flex-shrink: 0`. It also renders in a different shade than the chips it summarises: `KsTag` resolves the global `--ks-bg-tag`, while the label chips override it locally in `.label.kel-check-tag`. Giving it the `label` class (or using `KsCheckTag` with `readOnly`) fixes both the shade and keeps the vocabulary consistent.

**4. (minor) Overflowed labels are unreachable by keyboard**

`trigger="hover"` on a `KsTag` renders a non-focusable `<span>`, so a keyboard user has no way to see or click the hidden labels - and these are interactive filter controls, not decoration. Suggest `trigger="click"` (or hover + focus) with the reference rendered as a real button carrying an `aria-label`.

**5. (minor) The new `wrap` CSS silently changes the `EE` audit-log Details column**

`EE` `AuditLogs.vue` is the only existing `wrap` consumer and now inherits `white-space: normal; overflow-wrap: anywhere; max-width: 100%`. `overflow-wrap: anywhere` reduces the min-content contribution, so under `table-layout: auto` that column can collapse to near-minimum width with values broken mid-token. Possibly an improvement, but worth a before/after screenshot on the `EE` audit-logs page - or scope the new rule to the popover content only instead of every `wrap` caller.

**6. (minor) No test for the new behavior**

Three `data-test` hooks were added (`label`, `labels-overflow`, `labels-overflow-content`) and nothing asserts them. A single cheap case pins the slicing: `max=3` with 5 labels renders 3 chips and a `+2` chip; `max` unset renders all and no chip.

**7. (nit)** `max=3` with exactly 4 labels collapses one chip into a `+1` popover that costs roughly the room the hidden label would have taken. Consider collapsing only when `labels.length > max + 1`.

**8. (nit)** Native `title` instead of `KsTooltip` - every chip now shows a browser tooltip even when nothing is truncated. The existing `+N` precedent (`TaskTenants.vue`) uses `KsTooltip`.

**9. (nit)** `:width="360"` and `:hideAfter="150"` are unexplained constants, and the new wrap rule sits flat at the bottom of the style block instead of nested under the existing `&.wrap`.

Also checked and fine: the `text()` refactor is behavior-preserving (including empty-key labels); callers without `max` are unaffected; hidden/system label filtering runs before the slice so `+N` never counts labels the user should not see; `+N` needs no i18n key per existing convention; the reorder breaks no e2e tests; the `EE` executions table (`OSS` component + appended `cases` column) is not disturbed by the reorder - though since the closing issue is https://github.com/kestra-io/kestra-ee/issues/10200, verification should happen on an `EE` instance.

MilosPaunovic (approved):
All previous blockers are addressed, approving. A few non-blocking notes from this round:

**1. (minor) The `PR` description is stale and omits a user-visible migration.** It still says the extra labels show "in a popover when hovered" (it is click-only now), and it does not mention the `ks-column-order-` to `ks-column-order-v2-` key bump - which resets the saved column order for **every** table using the picker (`Flows`, `SecretsTable`, `TriggersManage`, `MetricsTable`, `KVTable`, and the `EE` tables), not just `executions`. Worth updating before merge since the squash body comes from it. The old `ks-column-order-<key>` entries are also left orphaned in `localStorage` (nit).

**2. (minor, a11y) No accessible name on the `+N` button.** A screen reader announces "+2, button" with no hint of what it opens. An `:aria-label` (i18n key with `{count}`) would fix it; also worth confirming the focus ring is visible in both themes, since the rule sets `border: 0` on a custom background.

**3. (minor) `:title` is now unconditional** - chips show a native tooltip even in `wrap` mode and inside the popover, where the full text is already visible (affects `EE` audit logs / assets, `OSS` `SecretsTable` / `DrillDownDrawer`). `:title="wrap ? undefined : text(label)"` would limit it to where truncation can actually happen.

**4. (minor) The column min-content is bounded but can still be large** - with `max=3` the labels column can still demand roughly 3 x 12rem plus the `+N` under `table-layout: auto`. One screenshot with three genuinely long labels would confirm the rest of the table is no longer pushed.

**5. (minor, test gap) Nothing opens the popover.** The three new stories all assert the closed state, so the recursive `Labels`, the popover rendering, the `.labels-popover` scoping and the keyboard activation are unexercised. A fourth story with `userEvent.click` + `waitFor` on `labels-overflow-content` (plus an `{Enter}` variant) would cover the feature's main path.

**6. (minor, test gap) The new unit test drives a code path the app never takes.** It exercises `reorderColumns`, which nothing in `ui/src` calls - the picker goes through the design-system copy's `setColumnOrder`, which got the same change with no test. Mirroring the assertion there, plus a test that a pre-existing old-key `localStorage` entry is ignored, would pin the actual migration.

Nits: the `max + 1` collapse semantics deserve a one-line comment or a `collapseThreshold` name; `:max="3"` is duplicated as a literal in `Executions.vue` and `Flows.vue`; `:width="360"` / `max-width: 12rem` are unexplained constants; `.label-more` inherits the hex `--ks-bg-tag` override rather than a token (pre-existing debt).

`EE` side looks clean this round: the audit-logs `wrap` regression is gone (`overflow-wrap` now scoped to `.labels-popover`), and no `EE` code references `ks-column-order`. `AssetsTable` metadata still has no `+N` since it does not pass `max` - natural `EE` follow-up, not blocking.

PR #18836, opened 2026-08-31
fix(storage-local): create destination parent on move and move/delete companion metadata

### LINKED ISSUE #18435: [Bug]: LocalStorage.move() fails on missing destination directories and mishandles companion .metadata files
### Describe the issue

I found three related issues in `LocalStorage` while testing file move and metadata handling on the `develop` branch.

### 1. `move()` fails when the destination directory does not exist

`LocalStorage.move()` calls `Files.move()` directly without creating the destination parent directory first.

For example, moving:

```text
kestra:///input.csv
```

to:

```text
kestra:///archive/2026/08/input.csv
```

fails if `archive/2026/08/` does not already exist.

The source file exists, but `Files.move()` throws `NoSuchFileException`. This is then caught by `LocalStorage.move()` and reported as:

```text
FileNotFoundException: File not found for URI: kestra:///input.csv
```

The source file remains intact.

**Expected:** The destination parent directory should be created when necessary, or the error should accurately identify the missing destination path.

### 2. `move()` does not move the companion `.metadata` file

When an object is stored with metadata, `LocalStorage` creates a companion file:

```text
payload.txt
payload.txt.metadata
```

After moving `payload.txt` to `payload-moved.txt`, only the payload is moved:

```text
payload-moved.txt
payload.txt.metadata
```

The metadata file remains at the original location.

As a result, `getWithMetadata()` at the new location returns the payload without its metadata.

**Expected:** The companion `.metadata` file should move together with the payload.

### 3. `delete()` leaves the companion `.metadata` file behind

`deleteFromPath()` removes the payload using `Files.deleteIfExists(path)`, but does not remove the corresponding `.metadata` file.

For example:

```text
Before:
file.txt
file.txt.metadata

After delete():
file.txt.metadata
```

**Expected:** Deleting a file should also delete its companion metadata file.

### Minimal reproduction

The following cases reproduce the observed behavior using the existing storage API.

#### Case 1 — Missing destination directory

```java
String tenantId = IdUtils.create();

URI source = storageInterface.put(
    tenantId,
    null,
    new URI("/input.csv"),
    new ByteArrayInputStream("payload".getBytes())
);

URI destination = new URI("/archive/2026/08/input.csv");

storageInterface.move(tenantId, null, source, destination);
```

With `archive/2026/08/` absent, this results in:

```text
java.io.FileNotFoundException:
File not found for URI: kestra:///input.csv
```

The source file exists before and after the failed move.

#### Case 2 — Metadata is not moved

```java
String tenantId = IdUtils.create();

Map<String, String> metadata = Map.of("key", "value");

URI source = storageInterface.put(
    tenantId,
    null,
    new URI("/payload.txt"),
    new StorageObject(
        metadata,
        new ByteArrayInputStream("payload".getBytes())
    )
);

URI destination = new URI("/payload-moved.txt");

storageInterface.move(tenantId, null, source, destination);

StorageObject result =
    storageInterface.getWithMetadata(tenantId, null, destination);
```

Observed:

```text
result.metadata() == null
```

The original `payload.txt.metadata` remains on disk, while `payload-moved.txt.metadata` does not exist.

#### Case 3 — Metadata is not deleted

After storing an object with metadata:

```text
file.txt
file.txt.metadata
```

calling:

```java
storageInterface.delete(
    tenantId,
    null,
    new URI("/file.txt")
);
```

removes `file.txt` but leaves:

```text
file.txt.metadata
```

on disk.

### Impact

These behaviors can:

* report an existing source file as missing;
* cause metadata to be lost when an object is moved;
* leave orphaned `.metadata` files on local storage over time.

### Relevant code

* `storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java` — `move()`
* `storage-local/src/main/java/io/kestra/storage/local/LocalStorage.java` — `deleteFromPath()`
* `storage-local/src/main/java/io/kestra/storage/local/LocalFileAttributes.java` — metadata retrieval

### Existing test coverage

The existing storage tests cover basic move and metadata operations independently, but do not cover:

* moving to a destination with an uncreated parent directory;
* moving an object together with its companion metadata;
* deleting an object together with its companion metadata.

Adding regression tests for these cases would help prevent the behavior from recurring.

### Environment

* Kestra version: `develop`
* Module: `storage-local`
* Storage backend: `local`
* Reproduced locally on the current development branch

suhanrain:
## Related issue

Closes #18435

## Changes

1. **`move()` creates the destination parent directory before moving** — moving an object to a destination whose parent directory does not exist used to throw `NoSuchFileException`, which was then misreported as a missing source file (`FileNotFoundException: File not found for URI: ...`).
2. **`move()` carries the companion `.metadata` file along with the object** — previously the metadata was silently lost after a move and the source `.metadata` file was left orphaned.
3. **`delete()` removes the companion `.metadata` file as well** — previously orphaned `.metadata` files accumulated on disk over time.

## Tests

Added 3 regression tests to `LocalStorageTest`:

- `shouldMoveToDestinationWithUncreatedParentDirectory`
- `shouldMoveObjectWithCompanionMetadata`
- `shouldDeleteObjectWithCompanionMetadata`

All 71 tests in `storage-local` pass.


loicmathieu (changes requested):
Can you remove most added comments or at least restrict to what's really useful

suhanrain:
Sorry about that. I removed the redundant comments and kept only the one that explains why the test checks the file directly.

PR #18841, opened 2026-08-31
fix(plugins): load task icons only once they are on screen

elevatebart:
### 🔗 Related Issue

Part of https://github.com/kestra-io/kestra-ee/issues/10075.

### ✨ Description

Opening the plugin catalog fires hundreds of icon requests in one burst: one `GET /api/v1/plugins/icons/{cls}/icon.svg` per card, plus an `api.kestra.io` probe for every group with no local icon.

`TaskIcon` renders an icon two ways, and only one of them was deferred:

- the `` branch already carried `loading="lazy"`;
- the monochrome branch paints through `mask-image`, which no attribute defers;
- and the per-class lookup (`loadIcon`, which falls back to probing the ecosystem catalog with `new Image()`) ran from a watcher on mount, wherever the icon sat.

Both uncovered paths now wait until the icon comes within 200px of the viewport, tracked by `useOnScreenOnce` — a small latching `IntersectionObserver` composable. It reports `true` from the start where `IntersectionObserver` is unavailable, so a missing API degrades to today's eager loading rather than to icons that never appear.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit` — 229 files, 2131 tests)
- [x] No `en.json` change, so no translation work
- [ ] Screenshots — see the note below

### 📝 Additional Notes

Two tests, both failing on `develop`: the composable's latch and its no-`IntersectionObserver` fallback, and `TaskIcon` leaving a monochrome icon's `mask-image` unset until the element intersects.

No screenshot is attached because there is nothing to see: the icons render exactly as before, they are just requested when they come into view. The behaviour that changed is visible in the network panel rather than on the page, and this session had no running instance to capture one from.

🐱 Our office cat has the same policy as this PR — she will not load until you scroll all the way over to her.

## Related PRs

- [feat(plugins): paginate the plugin catalog](https://github.com/kestra-io/kestra/pull/18842) — the other half of the same issue, independent of this branch.

### DIFF
--- a/ui/src/components/plugins/TaskIcon.spec.ts
+++ b/ui/src/components/plugins/TaskIcon.spec.ts
@@ -1,0 +1,65 @@
+import {afterAll, afterEach, describe, expect, it, vi} from "vitest"
+import {nextTick} from "vue"
+import {mount} from "@vue/test-utils"
+
+import TaskIcon from "./TaskIcon.vue"
+
+type ObserverCallback = (entries: {target: Element; isIntersecting: boolean}[]) => void
+
+const observers: {callback: ObserverCallback}[] = []
+
+class FakeIntersectionObserver {
+    constructor(callback: ObserverCallback) {
+        observers.push({callback})
+    }
+
+    observe() {}
+    unobserve() {}
+    disconnect() {}
+}
+
+const MONOCHROME_CLS = "io.kestra.plugin.core.log.Log"
+
+function mountMonochromeIcon() {
+    return mount(TaskIcon, {
+        props: {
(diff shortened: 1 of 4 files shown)


elevatebart:
Closing: lazy icon loading already landed in https://github.com/kestra-io/kestra/pull/18669, and the extra `IntersectionObserver` layer here goes beyond what was intended.

The remaining half of https://github.com/kestra-io/kestra-ee/issues/10075 is in https://github.com/kestra-io/kestra/pull/18842 (plugin catalog pagination), which is independent of this branch.

---

PR #18843, opened 2026-08-31
fix(dashboards): duration filters take an ISO-8601 duration or a number of seconds

ammeek:
Closes [https://github.com/kestra-io/kestra-ee/issues/10327](https://github.com/kestra-io/kestra-ee/issues/10327).

### ✨ Description

A dashboard chart whose `data.where` filtered `DURATION` with `value: PT1S` answered HTTP 500, and the response body carried the SQL that failed (`operator does not exist: bigint > character varying`). `value: 1000` returned 200, but meant milliseconds — while the DURATION column is rendered in seconds, so there was no format that both worked and matched what the chart displayed.

Three things were wrong:

1. **The value reached the store untouched.** `DURATION` maps to `state_duration` and `JdbcFilterService.greaterThan` binds the filter value as-is, so the query became `state_duration > 'PT1S'`. Every JDBC backend fails on this; Postgres is just the one whose message names the column.
2. **Units disagreed.** `state_duration` is milliseconds on all three JDBC backends, and `columnToField` divides by 1000 for display — so filtering used a different unit from the chart. ElasticSearch indexes `state.duration` in seconds, so a numeric filter also meant different things per edition.
3. **Bad input was a 500.** Nothing validated the value, at save time or at query time.

**How it is addressed**

* A data type declares its duration-typed fields — `IData.durationFields()`, overridden in `IExecutions` — so the rule lives in one place instead of being restated per store.
* New `DurationFilters` (core) rewrites the filters on those fields into the unit the store persists, recursing into `OR` and covering the comparison, `IN` and `NOT_IN` types; `IS_NULL`/`IS_NOT_NULL` pass through. Values are parsed through `JacksonMapper`, so a duration filter accepts exactly what every other duration in Kestra accepts: an ISO-8601 duration, or a number of seconds.
* `AbstractJdbcExecutionRepository.where()` normalizes with `Duration::toMillis`, which covers `fetchData`, KPI `fetchValue` and every EE JDBC repository, since those extend this class. The EE PR does the same with seconds for ElasticSearch.
* An unparseable value or an inapplicable filter type raises `InvalidQueryFiltersException`, which `ErrorController` already maps to 400, and `DataChartValidator`/`DataChartKPIValidator` report it as a chart violation — so a bad duration is rejected with a 422 when the dashboard is saved rather than surfacing when the chart renders.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/dashboards/filters/DurationFilters.java
+++ b/core/src/main/java/io/kestra/core/models/dashboards/filters/DurationFilters.java
@@ -1,5 +1,6 @@
 package io.kestra.core.models.dashboards.filters;
 
+import java.math.BigDecimal;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
@@ -7,8 +8,8 @@ import java.util.Set;
 import java.util.function.Function;
 
 import io.kestra.core.exceptions.InvalidQueryFiltersException;
-import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.utils.ListUtils;
+import io.kestra.core.utils.TypeConverter;
 
 /**
  * Rewrites the filters set on a duration field so that their values are expressed in the unit the store persists.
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/kestra/core/models/dashboards/filters/DurationFilters.java
```diff
@@ -0,0 +1,130 @@
...
+            case NotIn<?> f -> NotIn.<F> builder().field(field).key(key).values(values(f.getValues(), toStoreUnit)).build();
+            case IsNull<?> f -> (AbstractFilter<F>) f;
+            case IsNotNull<?> f -> (AbstractFilter<F>) f;
+            default -> throw new InvalidQueryFiltersException(
+                "filter type `%s` cannot be used on the duration field `%s`.".formatted(filter.getType(), filter.getField())
+            );
+        };
+    }
+
+    private static Duration parse(Object value) {
```
fhussonnois: Can you use our `TypeConverter#toDuration` utility class here ?

fhussonnois (changes requested):
LGTM, minus the parse static method!

fhussonnois (approved):

PR #18844, opened 2026-08-31
fix(executions): reject negative concurrency running count and bind the limit to its path

nkwiatkowski:
Companion PR: kestra-io/kestra-ee#10598 (same branch name).
Fixes kestra-io/kestra-ee#10294 (fix lives in OSS; the EE issue is closed by the companion EE PR).

### Problem

The flow concurrency-limit admin endpoint (`PUT /concurrency-limit/{namespace}/{flowId}`) had two input-handling holes:

- A **negative running count** was accepted with `200` and silently *disabled* the limit — a flow with `limit: 2` would then run every execution at once. `running: -5` went straight through.
- The **path was ignored**: the handler read `namespace`/`flowId` from the request *body*, not from the URL, so a request aimed at one flow could update a different flow named in the body.

### Fix

- `ConcurrencyLimit.running` now carries `@PositiveOrZero`; combined with the existing `@Body @Valid`, a negative count returns **422** instead of disabling the limit.
- The PUT handler binds `{namespace}`/`{flowId}` from the path and rebuilds the record from `path + resolved tenant + body's running`, so the URL is authoritative and a mismatched body can no longer retarget another flow.

### Evidence

`ConcurrencyLimitControllerTest` gains two tests — `shouldRejectNegativeRunningCount` (422) and `shouldIdentifyConcurrencyLimitFromPathNotBody` (path wins, spoofed flow is never created). Both fail on `develop` and pass here; the full class is green (5/5).

### DIFF
--- a/core/src/main/java/io/kestra/core/runners/ConcurrencyLimit.java
+++ b/core/src/main/java/io/kestra/core/runners/ConcurrencyLimit.java
@@ -4,6 +4,7 @@ import io.kestra.core.models.HasUID;
 import io.kestra.core.utils.IdUtils;
 
 import jakarta.validation.constraints.NotNull;
+import jakarta.validation.constraints.PositiveOrZero;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Value;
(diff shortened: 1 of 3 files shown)


loicmathieu (approved):
LGTM

PR #18848, opened 2026-08-31
fix(system): sync slot contracts as kestrabot, pre-formatted for artifact-sdk

elevatebart:
### 🔗 Related Issue

Follow-up to https://github.com/kestra-io/kestra/pull/18847, which got this workflow running for the first time. Its first sync — https://github.com/kestra-io/artifact-sdk/pull/82 — exposed the two problems fixed here.

### ✨ Description

**The PR went up under a maintainer's name.** The branch, the commit and the PR in artifact-sdk are all attributed to whoever owns the token the job pushes with, and `GH_PERSONAL_TOKEN` is a personal one, so artifact-sdk#82 reads as authored by `tchiotludo`. This now mints a `kestrabot` token from `GH_BOT_APP_ID` / `GH_BOT_PRIVATE_KEY` scoped to `artifact-sdk` — the same `actions/create-github-app-token` pattern `update-generated-sdk.yml` and `publish-hey-api-plugin.yml` already use — and sets the git identity to `kestrabot[bot]` so the commit links to the bot too.

**The PR failed Lint & Format.** artifact-sdk formats with oxfmt (2 spaces) and checks it repo-wide, while slot-contracts emits kestra's 4-space style, so every sync would land red and need a manual fix-up commit. The job now runs artifact-sdk's *own* pinned oxfmt over the two files it writes, installing only that repo's root devDependencies (`npm ci --workspaces=false --ignore-scripts`, 8s).

### 📝 Additional Notes

Verified against a real artifact-sdk checkout at `bd6d9dc`, with the contracts built from this branch: `npm run format` flags exactly `packages/artifact-sdk/src/contracts/index.d.ts` and `index.js` before the new step and reports *all matched files use the correct format* after; `npm run lint` (oxlint) is clean; and the formatted output differs from the committed contract only where the contract genuinely changed (`metrics` → `progress`, plus the new `tenant`, `source`, `fetchOutputs` and `fetchMetrics` props).

One prerequisite I cannot check from CI config alone: the kestrabot app installation has to cover `artifact-sdk` with contents and pull-requests write. If it does not, the token step fails first in the job, before anything is pushed — so the failure mode is loud and empty-handed rather than half-done. kestrabot has no recorded activity in that repo yet, so it is worth confirming the installation before the next merge that touches `slot-contracts/src`.

The changeset path moved to a job-level `env` because two steps now reference it.

### 🤖 AI Authors

Why did the cat sit on the keyboard? It heard the contracts needed a paws-itive format check. 🐱

https://claude.ai/code/session_01DNjv3eUzwze4UHiewvMnEQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01DNjv3eUzwze4UHiewvMnEQ)_

### DIFF
--- a/.github/workflows/sync-slot-contracts.yml
+++ b/.github/workflows/sync-slot-contracts.yml
@@ -13,6 +13,8 @@ jobs:
     name: Sync contracts to kestra-io/artifact-sdk
     runs-on: ubuntu-latest
     timeout-minutes: 10
+    env:
+      CHANGESET_FILE: .changeset/sync-slot-contracts.md
     steps:
       - name: Checkout kestra
         uses: actions/checkout@v7
@@ -88,7 +90,6 @@ jobs:
         working-directory: artifact-sdk
         run: |
           mkdir -p .changeset
-          CHANGESET_FILE=".changeset/sync-slot-contracts.md"
           TODAY=$(date -u +%Y-%m-%d)
           if [ -f "$CHANGESET_FILE" ]; then
             # Update the date line in-place (line starting with "Last synced:")
@@ -108,6 +109,14 @@ jobs:
           Last synced: $TODAY
           EOF
           fi
+
+      # artifact-sdk's CI checks formatting repo-wide, so run the pinned oxfmt from its own root
+      # devDependencies (--workspaces=false keeps the install to those) over what we just wrote.
+      - name: Format the synced files
+        working-directory: artifact-sdk
+        run: |
+          npm ci --workspaces=false --ignore-scripts
+          npx oxfmt --write packages/artifact-sdk/src/contracts "$CHANGESET_FILE"
 
       - name: Commit and push
         id: commit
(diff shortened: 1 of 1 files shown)


elevatebart:
`trigger-ee` is red on `aa5690c`, and it isn't this PR's failure: the job posted its payload to the EE compile-check webhook and got `HTTP 502 Bad Gateway` back after ~2 minutes, before any build started. Nothing in this diff is reachable from there — it changes one GitHub workflow file (`.github/workflows/sync-slot-contracts.yml`), and the EE compile check is dispatched for every PR regardless of what it touches.

There is no fix to port for a 502 from that service. I tried re-running the failed job and GitHub refused with `This workflow is already running` (the docker-image build is still going), so I'll re-run it once the run finishes and report back. If it 502s a second time, the dispatch service itself is down and this needs someone with access to it rather than a change here.

---

fdelbrayelle (approved):

PR #18851, opened 2026-08-31
fix(design-system): icon size prop, dead preview formatter, interval filter and textarea overflow

### LINKED ISSUE #18441: [Bash 2.0][UI] Save filter input has a horizontal scrollbar
### Describe the issue

Save filter **input** has a horizontal scrollbar
Image

### Environment

- Kestra Version: develop

### LINKED ISSUE #18544: [BATCH 2.0][design-system] KsIconButton silently ignores the icon's size prop
## Describe the issue

An icon inside `KsIconButton` always renders at 16px, whatever `size` the caller passes. `<KsIconButton><Download :size="24" /></KsIconButton>` looks identical to `:size="12"`, with no warning.

`vue-material-design-icons` renders its size as **SVG presentation attributes**:

```html
<span class="material-design-icon" ...>
  <svg class="material-design-icon__svg" :width="size" :height="size" viewBox="0 0 24 24">
```

CSS beats presentation attributes, so `KsIconButton`'s own `width`/`height` rule on that SVG wins and the prop has no effect. `KsTable`'s in-cell action icons have the same rule and the same behaviour.

## Steps to reproduce

1. Render `<KsIconButton aria-label="x"><Download :size="24" /></KsIconButton>` next to a bare `<Download :size="24" />`.
2. The one inside the button is 16px; the bare one is 24px.

## Expected behaviour

Either the prop is honoured, or the fixed size is a documented part of the contract with a `size` prop on `KsIconButton` itself driving the token — so callers are not silently overridden.

Both are defensible; picking one is a design-system API decision, which is why this is filed rather than patched.

## Notes

Pre-existing — the old rule hardcoded `16px` on the SVG and clamped it the same way. [#18531](https://github.com/kestra-io/kestra/pull/18531) (fixing #18433) replaced the hardcoded `px` with `--ks-icon-size-sm` and sized the wrapper span too, so the clamp is now consistent instead of leaving the span at `1em` with the SVG overflowing it — but it is still a clamp. Raised in that PR's review; filing so the API question does not get lost with it.

One behaviour change worth knowing either way: `--ks-icon-size-sm` is `1rem`, so the size now follows root font scaling (browser zoom, user font settings), which the hardcoded `px` did not.

## Environment

- Kestra Version: `develop`
- Edition: OSS (`ui/packages/design-system`)

### LINKED ISSUE #18635: [BATCH 2.0] ExecutionVariableExplorer passes a previewFormatter prop KsJsonTree does not declare
Found while reviewing https://github.com/kestra-io/kestra/pull/18602.

## The dead prop

`ui/src/components/executions/outputs/ExecutionVariableExplorer.vue:67` passes:

```vue
:previewFormatter="treePreviewFormatter"
```

`KsJsonTree` declares no `previewFormatter` prop. With no matching declaration it falls through as a plain attribute onto the component's root element, so the function is never called and the explorer's own `treePreviewFormatter` is dead code.

## The duplication behind it

`KsJsonTree` has its own private `collapsedPreview` (`ui/packages/design-system/src/components/Data/KsJsonTree.vue`), which is what actually renders collapsed rows. So there are two implementations of "summarise a collapsed container":

- `KsJsonTree.collapsedPreview` — the one that runs
- `ExecutionVariableExplorer.treePreviewFormatter` — the one that does not

They can drift silently, and #18602 is a small example of why that matters: fixing the empty-container rendering meant touching the tree's own preview and leaf paths, while the explorer's unused formatter sat there looking authoritative.

## Suggested resolution

Either:

1. **Delete `treePreviewFormatter`** and let the tree own its preview. Simplest, and correct if the explorer has no formatting need the tree does not already meet.
2. **Declare `previewFormatter` on `KsJsonTree`** as a real prop that overrides `collapsedPreview`, and have the explorer pass it deliberately. Right if the explorer genuinely wants a different summary — e.g. i18n'd item counts — in which case that intent should be expressed rather than accidentally inert.

Deciding between them needs a look at what `treePreviewFormatter` was meant to do differently, which is why this is an issue rather than a drive-by fix.

## Not urgent

Nothing is broken today: the tree's own preview is the one users see and it works. This is dead code plus a duplicated concept, so the cost is the next person's time rather than a defect.

**Scope:** frontend only, `ui/src/components/executions/outputs/` and `ui/packages/design-system/src/components/Data/`.

---

### LINKED ISSUE #18775: Interval filter Reset to default result in blank value
### Describe the issue

Resetting the interval filter's value by the _Reset to default_ button leads to no option being selected.

For example, the executions view uses the 24h predefined interval by default.

### Environment

### LINKED ISSUE #18776: Interval filter does not reject start date after end date
### Describe the issue

The interval filter does allow entering an interval whose start date value is actually after its end date value. This leads to an API validation error (HTTP 422) which should be prevented at the UI side.

### Environment

### LINKED ISSUE #18777: Interval filter does not highlight selected Apply to value
### Describe the issue

The interval filter does not highlight the selected _Apply to_ value (_Start date_, _End date_, or _Start or end date_), although you can see the value as the `dateFilter` query parameter.

### Environment

flcarre:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18544.
Closes https://github.com/kestra-io/kestra/issues/18635.
Closes https://github.com/kestra-io/kestra/issues/18775.
Closes https://github.com/kestra-io/kestra/issues/18776.
Closes https://github.com/kestra-io/kestra/issues/18777.
Closes https://github.com/kestra-io/kestra/issues/18441.

### ✨ Description

Six design-system defects, grouped because they all live in `packages/design-system` and none of them needs more than a file or two.

**The icon size prop was silently ignored (#18544)**

`KsIconButton` clamped the glyph to `--ks-icon-size-sm` through a `:deep()` rule, and CSS beats the SVG presentation attributes `vue-material-design-icons` renders, so a caller's `:size` had no effect and no warning. It now takes a `size` prop on the token scale (`xs` | `sm` | `base` | `lg` | `xl`, default `sm`) driving both the glyph and the button box. Honouring an arbitrary caller size was the other option and was rejected: it would put raw pixel values back into feature code, which `ui/AGENTS.md` forbids.

`sm` still renders a 16px glyph in a 24×24 box, so every existing call site is unchanged. `BackfillBanner` already passed `size="small"`, which fell through as an inert attribute; those three are dropped since `sm` is the default.

**A prop that was never declared (#18635)**

`ExecutionVariableExplorer` passed `:previewFormatter` to `KsJsonTree`, which declares no such prop, so it fell through as a plain attribute and the function never ran. The formatter it passed is identical to the tree's own `collapsedPreview`, i18n keys included, so it is dead code rather than a missing feature: both it and the binding are removed instead of the prop being declared.

**The Apply-to selector ignored the URL (#18777)**

The route decoder merges the two buckets of a custom range into one chip in a pre-pass. That pre-pass built the chip without `meta` while marking both buckets consumed, so the later branch that carries the route's `dateFilter` never ran and the selector fell back to its first option. `createCustomRangeFilter` now takes the meta, the way `createTimeRangeFilter` already did.

**Reset cleared the filter instead of restoring its default (#18775)**

`resetState` restored a default only when a pre-applied filter existed, and otherwise wrote `""`. On a view that defaults to an interval, resetting left nothing selected. It now seeds from `resolveDefaultVisibleValue`, the helper that already resolves a key's configured `defaultValue`.

**An inverted range reached the API (#18776)**

A start date after the end date was accepted and came back 422. The date panels no longer offer days that would invert the range, and an inverted range is not applied, which covers the residual case of two times on the same day. Prevented at the input rather than reported, so no new user-facing string is introduced.

**A horizontal scrollbar under the Save filter description (#18441)**

The textarea allowed horizontal resizing and scrolled a long word sideways. Fixed on `KsInput`, so it applies to every design-system textarea rather than this one dialog.

### 📸 Verified on a running instance

OSS `kestra/kestra:develop-slim` on :9011 with `npm run dev` pointed at it, `develop` and this branch in turn, Chromium 1440x900, light theme. Three of the six have a visible before/after and are captured below. The other three have no visual difference, for three different reasons, so they are written out rather than padded with identical images.

**#18544 — fixed.** Button boxes go from 24/24/24/24/24 to 20/24/28/32/40, and with a real icon the glyph follows: 12/16/20/24/32.

18544-iconbutton-size

One caveat on the evidence: the new `Sizes` story cannot show the glyph, since its icons are inline stub SVGs with `width="16"` and no `.material-design-icon` class, so the `:deep()` rule never matches them. The capture above uses a real `vue-material-design-icons` `Download`. The story's `play()` does fail on `develop` with `expected 24 to be less than 24` and passes here, so it still guards the box.

**#18775 — fixed.** Shown on the Logs page Level filter, the one key in OSS with a configured `defaultValue`. Before, reset leaves "Select an option" and the chip falls back to "Log Level in any"; after, it restores INFO.

18775-reset-to-default

**#18776 — fixed.** The panels refuse the inverting days: with the end date at Sep 7, 23 of the 30 September days are disabled.

18776-inverted-range

The screenshot above was taken before the last commit, when one entry path still got through: from an empty custom range, picking only a start date later than now left `state.endDateValue` null, the `startDateValue > endDateValue` guard was skipped, `getFilterValue` fell back to `endDateValue ?? new Date()`, and the search came back 422. The guard now compares the bounds `getFilterValue` will actually send, which covers a start later than now and an end earlier than it alike. Replayed on the same instance: picking a start three days out with no end date leaves the URL on `filters[timeRange][EQUALS]=PT24H` and raises no request.

**#18777 — no visual change, and still broken.** The state is correct on both branches: the chosen option carries `.active` and the chip label switches to "Ended". The highlight is invisible because `.date-filter-option.active` in `FilterSelect.vue` paints with `--ks-primary` and `--ks-background-card`, and neither token is declared anywhere under `packages/design-system/src/assets/styles`. Computed styles for the active and the inactive options are identical on both branches, `color rgb(8, 9, 10)` and `border-color rgb(8, 9, 10)` over a transparent background, so all three options render as plain text.

The decoder path fixed here needs one key carrying `valueType: "time-range"` **and** `customDateMode: "range"` **and** `dateFilterOptions`. No configuration has all three today: `clusterServiceFilter` (EE) has the first two, while `executionFilter` and `casesFilter` use `valueType: "select"`. So `createCustomRangeFilter` carrying the meta is a latent correctness fix worth keeping, but the issue closes on the tokens, not here.

**#18441 — no visual change.** On `develop` the textarea already resizes vertically only and wraps with `overflow-wrap: break-word`. Neither an empty field nor a long unbroken word produces a horizontal scrollbar: `scrollWidth == clientWidth == 326` on both branches, with and without content. The new rule does apply (`overflow-x` moves from `auto` to `hidden`), so it reads as a guard rather than as the fix for the reported symptom, which no longer reproduces on this dialog.

**#18635 — no visual change by design.** The collapsed preview and the expanded tree render identically before and after on the execution Input/Output tab, which is the claim: the deleted formatter never ran.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit`) — design-system: 826 passed / 100 files; storybook: 490 passed / 81 files
- [x] Translations are complete if `en.json` changed — `en.json` is untouched, no new user-facing string

### 📝 Additional Notes

**Replayed on a running instance**, see the section above. #18544, #18775 and #18776 are confirmed fixed, the last of them after an extra commit closing the one-sided range case. #18777 is not fixed by this branch and needs the two missing design tokens instead.

**Two API decisions taken here** that the issues deliberately left open: the token-scale `size` prop rather than a free-form caller size (#18544), and deleting the dead formatter rather than declaring the prop (#18635). Both are argued above and are easy to reverse if you disagree.

**Regression cover.** Each of the three logic fixes has a test that was checked to fail on the code before it, with the assertion naming the actual defect:

| Test | Fails before the fix with |
|---|---|
| `carries the route's dateFilter onto the merged range chip` | `expected undefined to deeply equal { dateFilter: 'END_DATE' }` |
| `restores the configured default on reset instead of clearing the value` | `expected '' to be 'PT24H'` |
| `does not apply a custom range whose start is after its end` | `expected 1787220000000 to be less than or equal to 1786356000000` |

A new `Sizes` story measures the real button boxes, asserting `sm` is exactly 24px and the scale is strictly increasing, so a future clamp cannot come back unnoticed.

#18441 is CSS only and has no test: there is no visual-regression baseline in the repo to hang one on, and a unit assertion on `resize`/`overflow-x` would only restate the stylesheet.

### 🤖 AI Authors

This PR was written by Claude at @flcarre's request. Why did the cat sit on the design system? It heard the tokens were the only thing it wasn't allowed to knock off the table. 🐱

### DIFF
--- a/ui/packages/design-system/src/components/Basic/KsIconButton/KsIconButton.vue
+++ b/ui/packages/design-system/src/components/Basic/KsIconButton/KsIconButton.vue
@@ -37,6 +37,8 @@
     import KsButton from "../KsButton/KsButton.vue"
     import KsTooltip from "../../Feedback/KsTooltip.vue"
 
+    type IconSize = "xs" | "sm" | "base" | "lg" | "xl"
+
     defineOptions({inheritAttrs: false})
 
     const props = withDefaults(defineProps<{
@@ -47,6 +49,7 @@
         to?: string | Record<string, unknown>
         replace?: boolean
         filled?: boolean
+        size?: IconSize
     }>(), {
         tooltip: "",
         placement: "left",
@@ -55,6 +58,7 @@
         to: undefined,
         replace: false,
         filled: false,
+        size: "sm",
     })
 
     defineSlots<{
@@ -64,7 +68,12 @@
     const attrs = useAttrs()
     const buttonAttrs = computed(() => ({
         ...attrs,
-        class: ["ks-icon-button", {"ks-icon-button--filled": props.filled}, attrs.class],
+        class: [
+            "ks-icon-button",
+            `ks-icon-button--${props.size}`,
+            {"ks-icon-button--filled": props.filled},
+            attrs.class,
+        ],
     }))
 
     const buttonTag = computed(() => (props.to ? "router-link" : undefined))
@@ -73,10 +82,12 @@
 
 <style scoped lang="scss">
     .ks-icon-button {
+        --ks-icon-button-glyph: var(--ks-icon-size-sm);
+
         color: var(--ks-text-primary);
-        width: 24px;
-        height: 24px;
-        min-width: 24px;
+        width: calc(var(--ks-icon-button-glyph) + var(--ks-spacing-2));
+        height: calc(var(--ks-icon-button-glyph) + var(--ks-spacing-2));
+        min-width: calc(var(--ks-icon-button-glyph) + var(--ks-spacing-2));
         border-radius: var(--kel-border-radius-base);
         text-align: center;
         display: inline-flex;
@@ -87,9 +98,25 @@
 
         :deep(.material-design-icon),
         :deep(.material-design-icon > .material-design-icon__svg) {
-            width: var(--ks-icon-size-sm);
-            height: var(--ks-icon-size-sm);
+            width: var(--ks-icon-button-glyph);
+            height: var(--ks-icon-button-glyph);
         }
+    }
+
+    .ks-icon-button--xs {
+        --ks-icon-button-glyph: var(--ks-icon-size-xs);
+    }
+
+    .ks-icon-button--base {
+        --ks-icon-button-glyph: var(--ks-icon-size-base);
+    }
+
+    .ks-icon-button--lg {
+        --ks-icon-button-glyph: var(--ks-icon-size-lg);
+    }
+
+    .ks-icon-button--xl {
+        --ks-icon-button-glyph: var(--ks-icon-size-xl);
     }
 
     .ks-icon-button:not(.ks-icon-button--filled) {
--- a/ui/packages/design-system/src/components/Data/KsDataTable/filter/layout/FilterEditPopper.vue
+++ b/ui/packages/design-system/src/components/Data/KsDataTable/filter/layout/FilterEditPopper.vue
@@ -46,6 +46,7 @@
         NULL_COMPARATORS,
     } from "../utils/filterTypes"
     import {FILTER_CONTEXT_INJECTION_KEY} from "../utils/filterInjectionKeys"
+    import {resolveDefaultVisibleValue} from "../utils/filterChipFactory"
     import FilterText from "./FilterText.vue"
     import FilterRadio from "./FilterRadio.vue"
     import FilterFooter from "./FilterFooter.vue"
@@ -285,10 +286,14 @@
             return
         }
 
+        // Falling back to a blank value left the filter with nothing selected, rather than with
+        // the default the page configured (the executions view's 24h interval, for instance).
+        const defaultValue = resolveDefaultVisibleValue(props.filterKey)
+
         Object.assign(state, {
-            textValue: "",
-            selectValue: "",
-            keyValuePair: [],
+            textValue: typeof defaultValue === "string" ? defaultValue : "",
+            selectValue: typeof defaultValue === "string" ? defaultValue : "",
+            keyValuePair: Array.isArray(defaultValue) ? [...defaultValue] : [],
             radioValue: "ALL",
             dateValue: null,
             timeRangeMode: "predefined",
@@ -383,6 +388,14 @@
         // defaults to now (getFilterValue) so typing just a Start or End date applies immediately.
         if (isTimeRange.value && state.timeRangeMode === "custom"
             && !state.startDateValue && !state.endDateValue) {
+            return
+        }
+
+        // An inverted range is rejected by the API with a 422, so it is never applied: the date
+        // panels already rule out the day-level case, this covers two times on the same day.
+        if (isTimeRange.value && state.timeRangeMode === "custom"
+            && state.startDateValue && state.endDateValue
+            && state.startDateValue > state.endDateValue) {
             return
         }
 
--- a/ui/packages/design-system/src/components/Data/KsDataTable/filter/layout/FilterEditPopper.vue
+++ b/ui/packages/design-system/src/components/Data/KsDataTable/filter/layout/FilterEditPopper.vue
@@ -391,12 +391,13 @@
             return
         }
 
-        // An inverted range is rejected by the API with a 422, so it is never applied: the date
-        // panels already rule out the day-level case, this covers two times on the same day.
-        if (isTimeRange.value && state.timeRangeMode === "custom"
-            && state.startDateValue && state.endDateValue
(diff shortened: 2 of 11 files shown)


elevatebart:
Can I haz Screenshotz?

elevatebart (approved):

PR #18856, opened 2026-08-31
fix(executions): render one metrics action per row

### LINKED ISSUE #18396: [Bash2.0] Metrics icon is shown multiple time on the metrics list
### Describe the issue

Use the following flow:

```yaml
id: metrics
namespace: company.team

tasks:
  - id: metrics
    type: io.kestra.plugin.core.metric.Publish
    metrics:
      - name: counter
        type: counter
        value: "{{randomInt(lower=0, upper=10)}}"
      - name: gauge
        type: gauge
        value: "{{randomInt(lower=0, upper=10)}}"
```

After execution, if you go to the metrics page of that execution you see the list of metrics with the metrics action icon duplicated on each row.

Image

### Environment

- Kestra Version: develop

jasperjonkhans:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18396.

### Description

The metrics table rendered its action column inside the visible-columns loop, producing one action icon per visible column.

This moves the action column outside the loop so every metric row renders exactly one action. A Storybook regression test covers the four-column case.

### Frontend Checklist

- [x] Translations unchanged (`en.json` was not modified)
- [x] Screenshots or video recordings attached showing the UI change

### Additional Notes

The regression test failed before the fix with four actions instead of one and passes after the fix.

### Before and After

before:

image

after: 

image


MilosPaunovic (approved):
Reviewed - `MetricsTable.vue`'s action column was rendering once per visible `column` because it lived inside the `v-for` loop over `displayColumns`; moving it outside so it's a direct child of `KsDataTable` fixes the duplication. The Storybook regression test correctly locks in the fix (verified the `aria-label` selector matches the `view metrics` translation string). No `kestra-ee` counterpart to check, and no other `row-action` usage in the `OSS` `UI` has this same bug pattern.

LGTM.

PR #18857, opened 2026-08-31
fix(docker): surface Docker unavailability before execution (#18739)

### LINKED ISSUE #18739: No warning that Docker access is missing until a script task is run
The Docker socket mount is part of the quickstart command, and it is sometimes left out on purpose.

A flow whose only task is an `echo` command cannot be run on an install where the socket has not been mounted. The Docker task runner is the default for script tasks, so Docker access is needed even for `echo`. That is not indicated anywhere in the flow, and no warning is given when it is saved, so the missing access is only discovered from a failed execution.

Two different socket paths are named in the error: `/var/run/docker.sock` in the suggested fix, and `unix:///dind/docker.sock` as the host that was tried.

### Reproduction

Start Kestra without the socket mount:

```bash
docker run --rm -it -p 8080:8080 --user=root --name kestra kestra/kestra:develop server local
```

Run this flow:

```yaml
id: hello_shell
namespace: company.team
tasks:
  - id: hello
    commands:
      - echo "Hello World"
```

The execution is failed with:

```
Docker execution failed because Docker socket is not accessible.

Fix:
- Mount docker socket:
  -v /var/run/docker.sock:/var/run/docker.sock
- OR use Process runner:
  taskRunner:

Tried Docker host: unix:///dind/docker.sock
```

The same task with `taskRunner: io.kestra.plugin.core.runner.Process` is run without error, and so is a flow with only a `io.kestra.plugin.core.log.Log` task, so only the default task runner is involved.

### To do

- Missing Docker access should be surfaced before the execution. When no Docker daemon can be reached from the instance, a warning should be shown when a flow with a script task is saved or run, so that this can be found out without a failed execution and a stack trace.
- One socket path should be named in the error. Currently the path suggested in the fix and the path that was tried are different, so the wrong socket is looked for by anyone who follows the message.

mohitjeswani01:
### 🔗 Related Issue
Closes https://github.com/kestra-io/kestra/issues/18739

### ✨ Description
Script tasks default to the Docker task runner, but when the Docker socket isn't accessible, this was only ever discovered after a failed execution — nothing warned the user when the flow was saved. Separately, the resulting error message named two different socket paths: a hardcoded `/var/run/docker.sock` in the suggested fix, and the actually-attempted host in the "Tried Docker host" line, which could point users to mount the wrong path.

This PR addresses both:

- Added a new `unavailabilityWarning()` hook on `TaskRunner` (core), which task runners can optionally override to report pre-flight unavailability. `Docker` overrides it to check whether the resolved socket path exists on disk when no custom `host` is configured (a templated/custom host can't be safely evaluated without a `RunContext`, so that case is skipped rather than guessed at).
- `FlowService.warnings()` now surfaces this per-task, so a flow using the Docker runner without Docker access shows a warning in the editor as soon as it's saved or validated — before it's ever run.
- Fixed `dockerSocketNotAccessibleMessage()` so the suggested `-v` mount path is derived from the actually-resolved Docker host instead of a hardcoded default, so the "Fix" suggestion and "Tried Docker host" line always agree.

### 🛠️ Backend Checklist
- [x] Code compiles and tests pass for the touched modules

### 📝 Additional Notes
Manually verified against a locally built image with and without the socket mounted: the warning appears in the flow editor before execution when Docker is unreachable, disappears once the socket is mounted, and the execution failure message now shows matching paths in both the "Fix" and "Tried Docker host" lines. No warning appears for flows using an explicit `Process` runner.

### 📷 Screenshots

**Warning shown before execution (socket not mounted), fix and tried-host paths now match:**
image
**No warning once the socket is mounted — flow is valid:**
image


mohitjeswani01:
Hey @loicmathieu any suggestions from your side?

mohitjeswani01:
Resolved the merge conflict @loicmathieu !

loicmathieu:

The original issue raises concern for which we need to take a decision, sorry about that.
Reading this PR, I think using reflection to find taskrunners clearly indicate that our execution model is not ready for this, it is too brittle.

mohitjeswani01:
Fair call on the reflection , it was there to avoid a hard dependency from core on the script module's Docker runner, but needing it to wire this in is a good sign the check doesn't belong in FlowService. I'll revert the TaskRunner/FlowService warning additions and leave this PR with just the dockerSocketNotAccessibleMessage() path-consistency fix. Pushing that shortly.

mohitjeswani01:
@loicmathieu Dropped the reflection-based pre-flight warning entirely — `TaskRunner.java` and `FlowService.java` are back to identical with `develop` (confirmed byte-for-byte). `Docker.java` now only contains the `dockerSocketNotAccessibleMessage()` / `socketPathFromHost()` fix for the path mismatch.

If the "Files changed" diff shows anything outside those two areas, it's unrelated `develop` commits (`getFlowIfExecutableOrThrow` tombstone logic) surfacing because this PR's merge-base hasn't been refreshed since I last synced — happy to rebase to clean that up if it'd help review.

Ready for another look whenever you have time.

PR #18858, opened 2026-09-01
fix(plugins): show installed script plugins on the catalog with proper titles

### LINKED ISSUE #17888: Plugins installed using CLI don't show up on plugins page
### Describe the issue

Steps to reproduce:
- Set up Kestra using `-no-plugins` docker image
- Attach to the container and install plugin using command: `kestra plugins install io.kestra.plugin:plugin-script-python:LATEST` or `kestra plugins install io.kestra.plugin:plugin-dbt:LATEST`
- Restart Kestra and refresh the plugins page.
- The new plugin does not show up on the main page, but if you click onto an existing plugin, you will see it in the menu down the left side.

After installing through CLI and restarting:

Image

Image

### Environment

- Kestra Version: `latest` but also saw this on 2.0 RC too I think

Piyush-r-bhaskar:
closes #17888

## What was fixed

- **Missing plugin cards on the Plugins page.** Script plugins (e.g. python) bundle the legacy deprecated `io.kestra.core.tasks.scripts` package, which the API reports as a subgroup. The catalog rule "subgroups replace the main entry" then dropped the real card, and the phantom subgroup was hidden as deprecated-only, so the plugin vanished. The main card now survives when a group has no visible subgroup.
- **Bogus "scripts" card on the plugin group page.** The same phantom subgroup rendered as its own card (with a wrong icon). Deprecated-only subgroups are now skipped and the tasks grid shows instead.
- **Lowercase titles ("debug", "agent", "mcp", "s3").** Display titles now come from the public catalog (`api.kestra.io/v1/plugins/subgroups`) via the enrichment store, on: catalog cards, group page cards + header + breadcrumb, and the task page breadcrumb. Falls back to the local title when the public API is unreachable.
- **Cleanup.** Dead CSS rule, redundant `.slice()` and unreachable fallback in the catalog, `elementCountFor` now reuses `extractPluginElements`.
- **Tests.** New specs for PluginCatalog and PluginGroup (fail on pre-fix code) plus a title test in the enrichment store spec.

## QA the original issue

1. `docker run -d --name qa17888 -p 8090:8080 kestra/kestra:latest-no-plugins server local`
2. `docker exec qa17888 sh /app/kestra plugins install io.kestra.plugin:plugin-script-python:LATEST`
   (the `sh` matters: `/app/kestra` has no shebang, calling it directly gives "exec format error")
3. `docker restart qa17888`, open `http://localhost:8090/ui`, finish the setup login.
4. Go to **Plugins**:
   - broken: no Python card, searching "python" finds nothing
   - fixed: a "Python" card with 4 tasks; search finds it
5. Click the Python card:
   - broken: a single "scripts" card with a wrong icon
   - fixed: Tasks grid with Script, Commands, CommandsTrigger, ScriptTrigger (no deprecated legacy task)
6. Titles everywhere (cards, header, breadcrumb) are properly cased, e.g. "Python", not "python".

### DIFF
--- a/ui/src/components/plugins/PluginCatalog.vue
+++ b/ui/src/components/plugins/PluginCatalog.vue
@@ -86,7 +86,7 @@
                     :iconCls="hasIcon(plugin.subGroup) ? plugin.subGroup : plugin.group"
                     :icons
                     :loadIcon="pluginsStore.loadIcon"
-                    :title="plugin.title.capitalize()"
+                    :title="enrichmentStore.getEnrichment(plugin)?.title ?? plugin.title.capitalize()"
                     :description="plugin.description"
                     :categories="plugin.categories"
                     :taskCount="taskCount(plugin)"
@@ -162,9 +162,10 @@
             return acc
         }, {})
 
-        const filtered = Object.values(grouped).flatMap(group =>
-            group.filter(p => p.subGroup).length ? group.filter(p => p.subGroup) : group.filter(p => !p.subGroup),
-        )
+        const filtered = Object.values(grouped).flatMap(group => {
+            const subGroups = group.filter(p => p.subGroup && isVisible(p))
+            return subGroups.length ? subGroups : group.filter(p => !p.subGroup)
+        })
 
         return filtered
             .filter((plugin, index, self) =>
@@ -235,8 +236,7 @@
         return baseList.value
             .filter(plugin => isPluginMatched(plugin, searchInput.value))
             .filter(plugin => matchesSelectedCategories(plugin))
-            .slice()
-            .sort(comparators[sortBy.value] ?? nameAsc)
+            .sort(comparators[sortBy.value])
     })
 
     const loadPluginIcons = async () => {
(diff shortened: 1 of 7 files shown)


RemiBarthe (approved):
LGTM 🚀 

**Two non-blocking notes.**

- Titles could come from the local API. 
`Plugin.of()` already returns the declared `@PluginSubGroup(title = ...)` verbatim `(core/.../docs/Plugin.java:58)`, which is where api.kestra.io gets them from too — that's why the public catalog says "KV" and "HTTP". The lowercase ones are just the 12 of 15 `plugin/core package-info.java` that declare no title and fall back to the bare package segment. So title = "Debug" etc. on those, plus dropping the `.capitalize()` in `PluginCatalog.vue` (it's also what turns "dbt CLI" into "Dbt CLI"), gets the same result offline and for private plugins — where the enrichment fallback still shows the lowercase title today.

- Tests. They select on `article` / `h5`, which are `PluginCard's` internal markup — a `data-test` on the card and its title would make them survive a refactor. `expect(pythonCard.text()).toContain("4") `also matches any "4" in the card. And the ~45-line fixture is duplicated verbatim across both specs, worth extracting.

PR #18860, opened 2026-09-01
fix(core): shiki background in markdown code blocks

Piyush-r-bhaskar:
Markdown code blocks overrode the shiki theme background with a surface token, so the syntax theme and the block background did not match. The overrides are removed and the shiki theme now paints the block.

To address comment from https://github.com/kestra-io/kestra-ee/issues/10477#issuecomment-5478635470


Pradumnasaraf (commented):
Hey, the Apps are EE only, and this PR is open on OSE, so it's not possible to test this thing

PR #18869, opened 2026-09-01
feat(variables): add an expression eval endpoint alongside the display rendere

ammeek:
### ✨ Description

**The problem.** The flow editor's Context tab cannot render `{{ namespace.hey }}`. It returns "Expression not found in current namespace or parent namespaces" even when the variable is defined, while `{{ kv('mykey') }}` works.

The cause is that the tab does not evaluate expressions on the server at all. It fetches the namespace's variables and KV keys up front, then resolves the expression in the browser with a hand-rolled partial reimplementation of Pebble. That evaluator is EE code, but the reason it exists is here: OSS has no endpoint that evaluates a single expression against a flow context. `/expressions/render` is deliberately display-only — it drops `kv()` and every other side-effecting function, and returns anything it cannot resolve as raw text. That is right for topology labels and wrong for an interactive eval box, which has to show either the real value or a real error. With nothing to call, the frontend grew its own engine, and it diverges from the real one on more than `namespace.*`: filters and operators are unsupported, `vars.*` resolves namespace variables instead of the flow's, and inherited variables are shallow-merged where the executor deep-merges them.

**The fix.** Add `POST /api/v1/{tenant}/expressions/eval`. It renders one expression with the full engine through `SecureVariableRendererFactory`, so `secret()` stays masked while everything else — filters, operators, `kv()` — resolves exactly as it would during an execution, and it returns `{result, error}` the way `POST /executions/{id}/actions/eval` already does.

Context resolution is unchanged and now shared: both routes take an `ExpressionContext` and go through the same `variablesFor()`, so they resolve against the same execution / draft source / persisted flow priority order, and an edition can authorize that context in one place instead of duplicating the order per route.


MilosPaunovic (commented):
The approach is right: reusing `variablesFor()` for both routes, running the real engine through `SecureVariableRendererFactory.createOrGet()` (the same renderer `POST /executions/{id}/actions/eval` already uses), and returning `{result, error}` matches the existing eval contract. `IllegalVariableEvaluationException` is also the correct single catch - `VariableRenderer.renderOnce` wraps `IOException | RuntimeException` (including `PebbleException`) into it, so nothing escapes as a 500.

A few things below, of which 2 and 3 are the ones I would want resolved before merge.

### 1. `EE` compile break - merge order

The public `ExpressionController` constructor gains a 7th argument, and `webserver-ee`'s `@Replaces` subclass calls `super(...)` with 6. Merging this alone breaks `kestra-ee` `develop` compilation until https://github.com/kestra-io/kestra-ee/pull/8998 lands. `OSS` first, `EE` immediately after - worth stating on both descriptions.

### 2. The authorization hook the description promises isn't there

The `Javadoc` says a subclass "can authorize that context in one place", but `OSS` exposes no hook - so https://github.com/kestra-io/kestra-ee/pull/8998 has to override each route and re-implement the exact same priority chain in `authorizeContext()`. Two consequences:

- **Drift risk.** If the `OSS` order ever changes (say persisted `flow` before draft source), `EE` authorizes one context while `OSS` resolves another - a silent authorization bypass, with nothing in either repo to catch it.
- **Triple lookup.** On the `executionId` path the same `execution` is fetched three times per call: inside `isExecutionAllowedOrThrow`, again in `authorizeContext` to get the `namespace`, and a third time in `variablesFor`.

Suggestion: resolve the context once in `OSS` into a small record (the `namespace` plus the resolved `flow`/`execution`) and call a `protected void authorize(ExpressionContext, ResolvedContext)` no-op from `variablesFor()`. `EE` then overrides one method, gets the `namespace` handed to it, and cannot fall out of sync.

### 3. `operationId` collision leaks into the `SDK`

`ExecutionController.evalExpression` already owns `evalExpression`, so the spec-diff comment on this `PR` shows the new route generated as `operationId: evalExpression_1` (and `evalExpression_1 200 response`). That becomes a meaningless generated function name in `kestra-sdk` - and is likely why the `EE` frontend hand-builds `${apiUrl()}/expressions/eval` through `useClient()` instead of calling a typed endpoint. Rename the method (e.g. `evalExpressionInContext`) or pin `@Operation(operationId = "...")`.

### 4. No test for the masking claim

The endpoint's core security property is that `secret()` is masked, and nothing asserts it - while the `/render` test does assert `[secret: MY_KEY]`. Also uncovered on `/eval`: the `executionId` path and the `namespace` + `flowId` path, which are exactly the two paths `EE` authorizes differently. Note the mask here is `******` (`MaskingFunction`), not `/render`'s `[secret: KEY]` - a test would pin that difference down.

### 5. `shouldResolveKvWhenEvaluating` leaves state behind

It writes `eval-key` into `io.kestra.tests` and never removes it; that `namespace` is shared with the rest of the suite. Delete it afterwards, or use a dedicated `namespace`.

### 6. Worth confirming: what stays unblocked

`createOrGet()` masks only `secret()`. `PebbleEngineFactory.createWithMaskedFunctions` explicitly documents a `functionsToBlock` list because masking "still invokes the real function", which is unsafe for side-effecting ones such as `http()` - and that list is empty here. So a caller with `FLOW` `EXECUTE` on a single `namespace` can make the server issue arbitrary outbound requests with no `execution` record or audit trail. It matches the `/executions/{id}/actions/eval` precedent so it is probably intentional, but the endpoint `description` should say so: it calls out `secret()` masking while `/render`'s description is explicit that `env()` is kept raw, and here `env()` resolves for real.

### 7. Nits

- The title has a typo - "display rendere" → "renderer", and it becomes the squash subject.
- `namespace` without `flowId` silently falls through to an empty context rather than a 400, so a caller who forgets `flowId` gets a confusing "unable to find" error surfaced straight into the `UI`.

### Backport

`ExpressionController` exists on `releases/v2.0.x` but not on `releases/v1.3.x` or `releases/v1.0.x`, so 2.0 is the only candidate - should the `Context` tab fix ship there too?

MilosPaunovic:
Following up on my review from 2026-09-01. The branch has taken four `develop` merges since, and something landed on `develop` in between that removes this `PR`'s premise.

**The `PR` is currently red on its own new test:**

```
ExpressionControllerTest > shouldResolveKvWhenEvaluating()
expected: "eval-value" but was: "******"
```

That is not a flake. https://github.com/kestra-io/kestra/pull/18551 ("fix(core): mask `kv()` on read-only render paths", merged 2026-09-03) added `KvFunction.NAME` to `SecureVariableRendererFactory`'s masked list, for this exact reason in its own words: `{{ kv('SOME_KEY') }}` on a read-only path resolved to the plaintext `KV` value with no `KVSTORE` permission anywhere on the call path. `develop` now reads:

```java
createWithMaskedFunctions(secureVariableRenderer, List.of(SecretFunction.NAME, KvFunction.NAME))
```

So the endpoint's stated contract no longer holds. The `PR` description says `kv()` "resolves exactly as it would during an execution", and the `@Operation` description says "functions with side effects such as `kv()` are all honoured" - through this renderer it returns `******`.

Two ways forward, and they are quite different decisions:

- **Accept the mask.** Drop the `kv()` claim from the description and the `@Operation` text, and change `shouldResolveKvWhenEvaluating` to assert the mask instead of the value. This also removes the reason for the `KVSTORE` ancestor-chain authorization in https://github.com/kestra-io/kestra-ee/pull/8998, which is currently what makes `{{ namespace.hey }}` return a `403` for anyone without `KVSTORE` at the namespace root - the very expression https://github.com/kestra-io/kestra-ee/issues/8887 is about.
- **Un-mask `kv()` for this route.** That reverses a three-week-old security fix and re-opens the hole https://github.com/kestra-io/kestra/pull/18551 closed, so it needs an explicit decision with @fdelbrayelle rather than a quiet change to the mask list.

Worth noting the bug fix itself is unaffected either way: `namespace.*` comes from `RunVariables`, not from a function, so it resolves normally.

Separately, the seven points from my 2026-09-01 review are all still open on the current head - the four merges are `develop` merges with no changes to the endpoint.

PR #6144, opened 2024-11-27
feat(*): introduce custom dashboards

brian-mulier-p:
closes kestra-io/kestra-ee#1711

PR #18871, opened 2026-09-01
feat(dashboards): move dashboard actions into a header kebab menu

Ben8t:
Closes https://github.com/kestra-io/kestra-ee/issues/10625.

## Problem

Based on feedback and discussion with @soyasis and @anna-geller.

On the dashboard home header, a standalone **Edit Dashboard** button sat next to **Create Flow**, and only for a non-default dashboard. There was no way to create a dashboard from the header, and the default-dashboard view offered no dashboard action at all.

## Fix
Group the dashboard actions into a kebab (⋮) menu placed to the left of **Create Flow**, which stays the only primary CTA:
- **Non-default dashboard** → kebab has **Edit Dashboard** + **Create Dashboard** (the standalone button is removed).
- **Default dashboard** → kebab has **Create Dashboard** only.

The kebab is gated on dashboard-create permission and `isCustomDashboardsEnabled`, so in OSS (custom dashboards disabled) it stays hidden and only surfaces in EE, where the create/edit routes are functional. The breadcrumb and dashboard selector are unchanged.

Adds the `dashboards.creation.label` string; the twelve other locales are filled by the translations CI job.

## Evidence
Verified against a running EE instance (H2). `npm run check:types` and `npm run lint` pass.

| Default dashboard | Non-default dashboard |
|---|---|
| kebab → Create Dashboard only | kebab → Edit Dashboard + Create Dashboard |

Both menu items navigate correctly (`/dashboards/new`, `/dashboards/:id/edit`).

kestra-10625-kebab-default

kestra-10625-kebab-nondefault


anna-geller (approved):
Nice! so fast

MilosPaunovic (changes requested):
Two real findings, one nit.

## 1. The `OSS` gate doesn't gate anything - the kebab does render in `OSS`

https://github.com/kestra-io/kestra/blob/288dc872a21b4acee9cc5cf9d51ebb87bfe5d770/ui/src/components/dashboard/components/Header.vue#L89-L93

`isCustomDashboardsEnabled` can never be `false` in any current deployment:

- [`AbstractJdbcDashboardRepository.isEnabled()`](https://github.com/kestra-io/kestra/blob/288dc872a21b4acee9cc5cf9d51ebb87bfe5d770/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcDashboardRepository.java#L170-L173) returns a hardcoded `true`
- [`ElasticSearchDashboardRepository.isEnabled()`](https://github.com/kestra-io/kestra-ee/blob/b22f40294f2dabb39a48873b5d2a2b4ba580ba04/repository-elasticsearch-ee/src/main/java/io/kestra/ee/repository/elasticsearch/ElasticSearchDashboardRepository.java#L103-L106) returns a hardcoded `true`
- those are the only two implementations, and neither has changed since `feat(*): introduce custom dashboards (#6144)` added them
- [`MiscController.getConfiguration()`](https://github.com/kestra-io/kestra/blob/288dc872a21b4acee9cc5cf9d51ebb87bfe5d770/webserver/src/main/java/io/kestra/webserver/controllers/api/MiscController.java#L144) sets the flag straight from `dashboardRepository.isEnabled()`, and `EE`'s `@Replaces` subclass only calls `super.getConfiguration()` and appends `EE` fields - it never touches this one

So `isCustomDashboardsDisabled` is always `false`, and `canManageDashboards` collapses to `isAllowedDashboard`, which in `OSS` is [`Me.isAllowed()`](https://github.com/kestra-io/kestra/blob/288dc872a21b4acee9cc5cf9d51ebb87bfe5d770/ui/src/override/stores/auth.ts#L17-L19) returning an unconditional `true`.

Net effect in `OSS`: the kebab **does** render on the dashboard home, holding a single **Create Dashboard** item that routes to [`dashboards/create` → `components/demo/Dashboards.vue`](https://github.com/kestra-io/kestra/blob/288dc872a21b4acee9cc5cf9d51ebb87bfe5d770/ui/src/routes/routes.ts#L105), the Enterprise upsell page. And it does so with no lock affordance, immediately next to the dashboard selector, which the `OSS` stub deliberately renders as a locked [`EnterpriseBadge`](https://github.com/kestra-io/kestra/blob/288dc872a21b4acee9cc5cf9d51ebb87bfe5d770/ui/src/override/components/dashboard/Selector.vue#L2). Two adjacent controls for the same `EE`-only feature, one marked locked and one not.

The config flag is a backend-capability signal, not an edition signal. The idiom already used across the `UI` for this is:

```ts
const isOSS = computed(() => miscStore.configs?.edition === "OSS")
```

(`useValues.ts:31`, `Namespaces.vue:133`, `McpServerList.vue:106`, `TaskVersion.vue:17`, among others.) Either gate on that, or move the actions block into an `override/` component the way `Selector.vue` does, so `OSS` ships a locked stub instead of a working-looking menu. The description's claim that "in `OSS` (custom dashboards disabled) it stays hidden and only surfaces in `EE`" needs correcting either way.

## 2. `dashboards.creation.label` already exists in `EE` and shadows the new `OSS` key

[`ee_translations/en.json#L2368-L2372`](https://github.com/kestra-io/kestra-ee/blob/b22f40294f2dabb39a48873b5d2a2b4ba580ba04/ui-ee/src/translations/ee_translations/en.json#L2368-L2372) already defines `dashboards.creation.label: "Create Dashboard"` (alongside `dashboards.creation.confirmation`), used by `Create.vue` and `Content.vue`.

`EE` composes its locales with `lodash/merge(en, en_ee)` in `ui-ee/src/translations/en.ts`, `EE` last - so in `EE`, the only edition where this menu is functional, the label resolves from the `EE` file rather than from the key this `PR` adds. The new `OSS` key and its twelve generated translations are dead there, and a later edit or re-translation of the `OSS` string will silently have no effect in `EE`.

Now that the string is needed by a shared component, `OSS` is the right home for it. Worth dropping the `label` entry from `ee_translations/en.json` (and the twelve `EE` locales, via `translations:generate`) in a companion `EE` `PR`, keeping `confirmation` where it is.

## Nit - bypassing `NavBarActions`

`NavBarActions` already implements exactly this shape: collapse the default slot into a `NavBarActionsDropdown`, keep `#primary` inline. Going straight to `NavBarActionsDropdown` is needed only because `MAX_INLINE_ACTIONS = 1` would render the default-dashboard case (one item) as an inline button rather than a kebab. `Executions.vue` sets the same precedent, so this is fine - but worth a line in the description or a short comment, otherwise the next person reasonably "fixes" it back to `NavBarActions` and quietly loses the kebab on the default dashboard.

## Other checks

- `translations:check` (dependency-free gate) passes: no missing, extra or stale keys. `fingerprints.json` was written by the `GitHub Action` commit rather than by hand, which is correct.
- Layout is unchanged: `NavBarActions` was renderless, so the `#actions` slot emits the same sibling pair (kebab, then primary) in the same order.
- `isNonDefaultDashboard` correctly guards the edit item, so `props.dashboard?.id` in the `dashboards/update` params is always defined at that point.
- No tests. The gating logic (edition, default vs non-default dashboard) is precisely the kind that regresses invisibly - finding 1 exists because nothing exercises it. A small `Vitest` on the two computeds, or a `Storybook` story covering both dashboard states, would earn its place here.

## Backport

`develop` is `2.1.0-SNAPSHOT`, so this does not reach `2.0`. It reads as a feature rather than a fix, so my assumption is no backport - but `releases/v2.0.x` is a live target, so worth confirming whether it should also land there.

Ben8t:
Thanks for the thorough review, @MilosPaunovic — both findings were spot on. Addressed in `d564d0c`:

**1. OSS gate did nothing → now gated on edition.** You're right that `isCustomDashboardsEnabled` is hardcoded `true`, so the kebab rendered in OSS with a working-looking Create Dashboard item next to the locked selector. Switched the gate to the edition signal you pointed at (`miscStore.configs?.edition === "OSS"`, as in `Namespaces.vue`/`useValues.ts`): `canManageDashboards = isAllowedDashboard && !isOSS`. In OSS the actions are now hidden entirely, so the description's "only surfaces in EE" is now accurate. Added a comment noting the capability flag is not an edition signal.

**2. EE key shadows the OSS one.** Correct — `ee_translations/en.json` defined `dashboards.creation.label` and EE merges last. Companion EE PR removes it (keeping `creation.confirmation`) so EE inherits the shared OSS key: kestra-io/kestra-ee#10626. Same branch name; it should merge after this one lands on `develop`.

**Nit — bypassing NavBarActions.** Added a two-line comment explaining that `MAX_INLINE_ACTIONS = 1` would render the single-item default-dashboard case as an inline button and drop the kebab, so the direct `NavBarActionsDropdown` is deliberate (same as `Executions.vue`).

**Tests.** Added `Header.spec.ts` covering the gating that finding 1 slipped through: OSS hides the kebab; EE shows Create-only on the default dashboard and Edit+Create on a non-default one.

**Backport.** Agreed it reads as a feature on `2.1.0-SNAPSHOT` - I would rather backport to`releases/v2.0.x`. It's about adding a label right?

MilosPaunovic (approved):
Looks good, approving. Everything from the last round is addressed, and I verified the fix rather than taking it on faith:

- Ran `Header.spec.ts` in a review worktree: 3 passed.
- Mutation-checked it: reverting `canManageDashboards` to the `isCustomDashboardsDisabled` form makes the `OSS` case fail at `Header.spec.ts:65`. It is a real regression test for the gate, not a vacuous one.
- `oxlint` and `eslint` clean on both changed files; `vue-tsc --project tsconfig.test.json` reports nothing `Header`- or `dashboards`-related.
- The spec is picked up by `vitest --project=unit`: that project sets no `include` override, so the default `**/*.spec.ts` glob applies, and eleven other colocated `src/**` specs already run that way.
- `kestra-ee#10626` reviewed too and clean: removes only `dashboards.creation.label` across all thirteen `ee_translations` files plus its `fingerprints.json` entry, keeps `creation.confirmation`.

Two small things to fold in before merging, neither blocking:

1. **The description still states the old mechanism** - "gated on dashboard-create permission and `isCustomDashboardsEnabled`, so in `OSS` (custom dashboards disabled) it stays hidden". The conclusion is right now, but the mechanism named is the one that did not work, and "(custom dashboards disabled)" is the wrong premise. Worth rewording to the edition gate.

2. **`isCustomDashboardsDisabled` now survives only inside `showSelector`** (`Header.vue:107`), two lines below the new comment declaring that flag hardcoded `true` and unfit to gate anything. Since it can only ever be `false`, the `|| isCustomDashboardsDisabled.value` branch is dead and the file argues both ways. Optional and pre-existing, but dropping it is behaviour-identical and would leave the comment consistent with its surroundings.

Optional test gap: the spec covers edition and default-vs-non-default, but not `isAllowedDashboard === false`. One case with `allowed = false` would close the permission half of the gate.

## Backport

> It's about adding a label right?

Not quite - it also removes the standalone **Edit Dashboard** button and moves that entry point behind two clicks. That is a change to an existing control on a released minor, so whether it ships in `2.0` is @anna-geller's or @soyasis's call rather than a mechanical yes.

If they say yes, the mechanics are clean - I checked:

- `ui/src/components/dashboard/components/Header.vue` is byte-identical between `origin/releases/v2.0.x` and `origin/develop`, so the commit cherry-picks without conflict
- `NavBarActionsDropdown`, `NavBarActions`, `NavBarAction` and `navBarActionsContext.ts` all exist on `2.0`
- the `configs?.edition === "OSS"` idiom exists on `2.0` (`useValues.ts:31`, `McpServerList.vue:106`, among others)
- `dashboards.creation.label` is absent from `2.0`'s `ui/src/translations/en.json`, so the backport has to carry the `en.json` addition and let the translations job regenerate the twelve locales plus `fingerprints.json`, same as here
- `kestra-ee`'s `releases/v2.0.x` still carries the duplicate `dashboards.creation.label` in `ee_translations/en.json`, so `kestra-io/kestra-ee#10626` needs its own `2.0` backport too, merged after the `OSS` `2.0` backport

MilosPaunovic:
Should this be merged @Ben8t?

Ben8t:
@MilosPaunovic yes, let's go. All good for merge on your side?

PR #18875, opened 2026-09-01
fix(namespaces): parent folders not created when a folder name contains a slash

MilosPaunovic:
Closes https://github.com/kestra-io/kestra-ee/issues/6553.

### ✨ Description

Creating a folder named `a/a` from the namespace files editor left a hidden entry behind. `POST /namespaces/{ns}/files/directory?path=/a/a` persisted only `/a/a/`, without the `/a/` entry above it, so the root listing had nothing to show and the folder was gone after a refresh - while still occupying the path in the index. Creating `a` afterwards then made a phantom `a` child appear underneath it.

`Namespace.createDirectory` now creates the missing parent directories, the way `putFile` already does for the ancestors of a new file (`mkDirs`), so a slash in a folder name creates the whole chain like it does in an editor. Nothing changed in the `UI`: the file tree already splits the name on `/` and posts the nested path.

Verified against a local instance, same namespace, same request, before and after the change:

| | before | after |
|---|---|---|
| `POST .../files/directory?path=/a/a` | `200` | `200` |
| `GET .../files/directory` (root) | `[]` | `[{"fileName": "a", "type": "Directory"}]` |
| `GET .../files/directory?path=/a` | `404 Directory not found: /a` | `[{"fileName": "a", "type": "Directory"}]` |

On the unpatched build, creating `a` after that left `/a` holding a child `a` nobody asked for - that is the phantom folder from the issue, and it no longer happens.

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched modules (`./gradlew :core:test --tests "InternalNamespaceTest"`, `./gradlew :webserver:test --tests "NamespaceFileControllerTest"`, `./gradlew :core:test --tests "*Namespace*"`)

### 📝 Additional Notes

No screenshots: the change is backend-only and the `UI` is untouched, so the evidence above is the `API` the file tree calls, captured on a running instance rather than in a test.

One thing deliberately left alone: `createNamespaceDirectory` still publishes a single audit event for the requested path in `EE`, so the parents created underneath it are not audited individually. The file endpoint does audit implicitly created parents (`innerCreateNamespaceFile` returns them), so closing that gap means returning the created list from the directory endpoint too - a small cross-repo change that does not belong in this fix.

Why did the folder cross the road? To get to the other `/side` - it had no parent to stop it. 🐱

### DIFF
--- a/core/src/main/java/io/kestra/core/storages/InternalNamespace.java
+++ b/core/src/main/java/io/kestra/core/storages/InternalNamespace.java
@@ -501,7 +501,7 @@ public class InternalNamespace implements Namespace {
             (maybeParentPath = Optional.ofNullable(NamespaceFileMetadata.parentPath(maybeParentPath.map(Path::toString).orElse(path))).map(Path::of)).isPresent()
                 && !this.exists(maybeParentPath.get())
         ) {
-            this.createDirectory(maybeParentPath.get());
+            this.createSingleDirectory(maybeParentPath.get());
             createdDirs.add(NamespaceFile.of(namespace, maybeParentPath.get().toString().endsWith("/") ? maybeParentPath.get().toString() : maybeParentPath.get() + "/", 1));
         }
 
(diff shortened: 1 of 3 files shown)


MilosPaunovic:
Could you give this a test please @Pradumnasaraf? 🙏🏻

Pradumnasaraf:
Hey, @MilosPaunovic, sure! Do you think this issue https://github.com/kestra-io/kestra-ee/issues/10582 can be tackled at the same time, since it is somewhat related?

MilosPaunovic:
There is already a PR from Piyush for that so I'd keep it separated.

Pradumnasaraf:
Ah, okay, I did not see that. I will test this then

Pradumnasaraf (approved):
Working now! LGTM

https://jam.dev/c/4291d861-0c88-4041-96e2-2eb02be6e413

PR #18877, opened 2026-09-01
fix(flows): no-code enum default preselection, default hint in field title

### LINKED ISSUE #14003: [No-code editor] shows default concurrency type Queue and YAML misses it until toggled
### Describe the issue

In the YAML configuration, there is no concurrency property with type `Queue`, but in the no-code editor, it shows a default type `Queue`. By default, it should be empty/none. When we increase the concurrency limit in the no-code editor, it adds the concurrency property in the YAML config, but without the type `Queue`. To make type `Queue` appear, we have to change to another type and then change back to `Queue`.

Jam: https://jam.dev/c/321c8b1b-6837-42b6-ad09-d6e91afa9055

### Expected Behaviour
By default, concurrency should be empty/none unless set. When setting concurrency in the no-code editor, the YAML should include the correct type `Queue` without needing to change it to another type and back.

### Environment

- Image hash and time: cbc4690 `Tue, Jan 6, 2026 12:48 PM`
- Edition: Open Source

elevatebart:
Closes https://github.com/kestra-io/kestra/issues/14003.

### ✨ Description

**Problem.** In the no-code editor, an enum field whose schema carries a default rendered that default as if the user had selected it. On a new flow, `Flow → Configure → concurrency → behavior` showed `QUEUE` even though the YAML held no value; picking `QUEUE` on the first try emitted nothing (the select believed it was already selected), so the only way to write it was to switch to another value and back — and clearing just re-displayed the default.

**Fix.**
- `TaskEnum` no longer falls back to `schema.default` for display: the select stays empty (placeholder visible) until the YAML actually holds a value.
- The schema default is shown as a hint on the field's title line (`default: QUEUE`), reusing the existing pluginDefaults hint; when a flow pluginDefault overrides the schema default, the pluginDefault wins the hint since it is what applies at runtime.
- The clear affordance on the select is only offered when the field is not required.

**Evidence.** Two new Storybook play tests, both verified to fail on the commit before this one and pass after:
- `TaskEnum → WithSchemaDefault`: nothing is preselected, and picking the default value on the first try writes it.
- `TaskObjectField → EnumFieldWithSchemaDefault`: the title line shows `default: QUEUE` while the control holds no value.

### 🎨 Frontend Checklist

- [x] Translations are complete (`npm run translations:check` reports no missing, extra, or stale keys)
- [x] Screenshots of the new field state (empty select + `default: QUEUE` title hint) attached below; the two play-tested stories above cover the same states in CI

### 🏞️ Screenshots

cc2openlight

cc3setlight

### 📝 Additional Notes

The new `block_editor.schema_default_tooltip` key was translated by hand for the twelve languages since no `GEMINI_API_KEY` was available in this environment; the fingerprint entry matches the English source, so the generator will not re-touch it.

The `default: …` hint is not enum-specific: it renders on every field carrying a schema default, across the no-code forms. That is a deliberate trade — removing the preselection removes information, so the hint puts it back — but the scope was decided in this PR rather than in a design review. The screenshots above went to the PR author; the issue owner has not reviewed them. Worth one look if that breadth is not wanted.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UTUS42o8Zi8ZmqhyDe5Yum)_


flcarre (approved):
Approving. The diagnosis is exactly right and the fix is the small one.

`values = props.modelValue ?? schema.default` made `ElSelect` display a value the model did not hold, and `ElSelect` emits nothing when the user picks the value it already shows. Hence the symptom in #14003: `QUEUE` visible, YAML empty, and the only way to write it was to pick something else and come back. Dropping the fallback so the placeholder shows until the model holds a value is the correct fix rather than a workaround on the emit side.

I traced the rest of the chain rather than taking the description for it:

* `required` really does reach `TaskEnum`: `TaskObjectField`'s `componentProps` passes `required: isRequired.value`, so `:clearable="!required"` is live and not decorative.
* The new `data-test="field-${fieldKey}"` lands in the DOM. `KsFormItem` sets `inheritAttrs: false` but binds `$attrs` explicitly onto `ElFormItem`, so the E2E selector will resolve.
* `schemaDefault` handles the falsy cases correctly: `String(false)` and `String(0)` still produce a hint, and only `undefined`, `null` and objects are dropped. That is the bug I expected to find and it is not there.
* Reusing `block_editor.plugin_default` for both sources is fine, since its English value is the neutral `default: {value}`, and the tooltip key is what distinguishes them.
* Both stories fail on the parent commit for the right reason: `WithSchemaDefault` on the click-then-write assertion rather than on the initial one (the model was already empty before the fix, only the display lied), and `EnumFieldWithSchemaDefault` because the hint did not exist at all.
* CI: the only red job is `trigger-ee`, with an empty verdict, which is the usual infrastructure flake and unrelated. `Run storybook component tests` passed in `Frontend - Tests`, so the two new play tests genuinely ran.

Three non-blocking remarks.

## `defaultHintFor` is coupled to the label markup

```ts
.find(hint => hint.previousElementSibling?.textContent?.trim() === field)
```

This is better than the bare text match it replaces, but it now depends on the label being the immediate previous sibling and on its text being exactly the field name. A required marker, an icon or a wrapper added to the label breaks both F9 stories, and the failure will read as "hint missing" rather than "markup moved".

This PR adds the hook that solves it. Since the form item carries `data-test="field-level"` and the hint carries `data-test="field-default-hint"`:

```ts
function defaultHintFor(taskEdit: HTMLElement, field: string) {
    return taskEdit.querySelector(`[data-test='field-${field}'] [data-test='field-default-hint']`)
}
```

Same assertion, no dependency on sibling order.

## The hint is now product-wide, not enum-specific

Every field carrying a schema default gets a `default: …` on its title line, not just enums. Your own story comment says as much, and the reasoning is sound: removing the preselection removes information, so putting it back as a hint is the honest trade. But the issue asked only for the preselection, and this lands on every no-code form in the product.

The description says the screenshots went to the issue owner, so this may already be settled. If it is, a line in the PR saying so would save the next reader the same question; if it is not, a long task form with a hint on most rows is worth one look before it ships.

## zh_CN keeps `value` in English

> 字段未设置时使用的默认 value

`fr`, `es`, `it`, `pt` and the rest all translate it. The reserved-terms rule is aimed at `value` as a domain term, as in a key/value pair, not at the ordinary word inside "default value". Minor, but it reads as an oversight rather than as the rule being applied.

Nice work on the E2E change too: `[data-test='field-level']` instead of `.kel-select` filtered by the displayed text is the selector we are supposed to be writing, and it would have survived this change without being touched.

PR #18878, opened 2026-09-01
fix(secrets): stop presenting the OSS secrets page as locked

### LINKED ISSUE #18348: [Bash2.0] OSS Secrets page is greyed and locked but still display useful information
### Describe the issue

On OSS, on the left menu, the Secret menu item is greyed out with a lock icon, which is for features restricted to EE so users would not click on it.

But it still exposed the OSS secrets: the secret env vars.

Image

## Proposal

Don't display the left menu entry as greyed out with a lock icon, this convey the wrong meaning.
But add to the page an explaination that secrets are **read only** in OSS and EE provides secret manager integrations.

### Environment

- Kestra Version: develop

flcarre:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18348.

### ✨ Description

The Secrets entry in the sidebar carried `locked: true` unconditionally, so it rendered with the same greyed-out treatment and lock badge as Apps and Tests. Those two really are unavailable in OSS. Secrets is not: the page behind it documents the supported OSS path, declaring secrets through `SECRET_<KEY>` environment variables, alongside the EE upsell.

The lock therefore told users "not for you" about a page that works for them. The entry keeps its padlock icon, which belongs to the concept of a secret rather than to availability.

Measured on a running OSS instance, before and after:

| Item | before | after |
|---|---|---|
| Apps | `is-locked`, badge, `rgb(126,126,155)` | unchanged |
| Tests | `is-locked`, badge, `rgb(126,126,155)` | unchanged |
| **Secrets** | `is-locked`, **badge**, `rgb(126,126,155)` | no `is-locked`, **no badge**, normal palette |
| KV Store | none, `rgb(8,9,10)` | unchanged |

### 🎨 Frontend Checklist

- [x] Translations are complete — no string changed

### 📝 Additional Notes

The issue also proposes adding a note on the page that secrets are read-only in OSS. That copy already exists there ("Create a New Secret", "Encode the secret value in Base64", "Set an environment variable named `SECRET_<MY_SECRET_KEY>`"), so this PR only removes the misleading lock. Say the word if you want the wording strengthened as well.

Only the OSS `override/` copy of the menu is touched, so the EE menu, where Secrets is a full feature, is unaffected.

### 🤖 AI Authors

This PR was written by Claude at @flcarre's request. The cat inspected the padlock, found the door open the whole time, and left without commenting. 🐱

### DIFF
--- a/ui/src/override/components/useLeftMenu.ts
+++ b/ui/src/override/components/useLeftMenu.ts
@@ -302,9 +302,6 @@ export function useLeftMenu() {
                         icon: {
                             element: LockOutline,
                         },
-                        attributes: {
-                            locked: true,
-                        },
                     },
                     {
                         id: "triggers",


MilosPaunovic (approved):
Verified the `locked` attribute on `KsSideBarItem` only drives the visual `is-locked` class and padlock badge (`ui/packages/design-system/src/components/Navigation/KsSideBarItem.vue`) - it doesn't gate navigation, so removing it from the `secrets` entry can't change access, only the misleading greyed-out treatment. Confirmed only `secrets` is touched; `apps`, `tests`, and the other `EE`-gated entries keep `locked: true` unchanged, matching the `before`/`after` table. The `EE` sidebar has its own separate `useLeftMenu.ts` override, so it's genuinely unaffected. No translation strings changed, and no existing test covers this file.

Approving - clean, correctly scoped fix.

PR #18908, opened 2026-09-02
fix(executions): validate label keys in the label editor before submit

### LINKED ISSUE #18798: Labels not validated by bulk by-ids action
### Describe the issue

Unlike other labels-related endpoints, the `labels/by-ids` API endpoint does not properly validate set labels.

Reproduction steps:

1. Execute a flow
2. Navigate to the Executions view
3. Select an execution
4. Hit the bulk _Set labels_ button
5. _Add labels_
6. Enter an invalid label - e.g. an uppercase key `FOO` and value `bar`
7. The `labels/by-ids` API updates the labels with the invalid label

### Environment

loicmathieu:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18798.

### ✨ Description

The bulk *Set labels* action on the Executions view accepted an invalid label key (e.g. uppercase `FOO`), unlike the other label endpoints.

The backend side of this was already fixed on `develop` in `2311df1dae` ("fix(executions): validate label keys on every execution label API"), shipped in `v2.0.0-rc12` — the original report was against `rc8`. That commit made `labels/by-ids` reject an invalid key with a 422, with test coverage in `ExecutionControllerTest`.

What was still missing is client-side validation: the shared `LabelInput.vue` editor (used by the bulk dialog, the single-execution popover, and the flow-run form) only checked for empty key/value, so an invalid key silently round-tripped to the server for an opaque error.

This PR adds `isValidLabelKey`/`hasInvalidLabelKeys` to `executionLabels.ts`, mirroring the backend's `Label` key pattern, and surfaces an inline error under the offending row in `LabelInput`. All three callers now disable their confirm/execute button while any key is invalid, and the two dialogs that already double-guard the empty-key case (toast + disabled button) do the same for invalid keys.

Along the way, a real (unrelated but adjacent) bug was found and fixed: the "Select all" bulk path (`labels/by-query`) had no `.catch` at all, so a server rejection there produced no feedback whatsoever. It now shares the same error handler as the `labels/by-ids` path.

### 🎨 Frontend Checklist

### 📝 Additional Notes

- No screenshot attached: this PR was prepared in a background session without a running instance to capture one against. The new inline error (`KsText type="danger"`, `data-test="label-key-error"`) is covered by two Storybook stories (`WithValue`, `InvalidKey`) in `LabelInput.stories.tsx`, both passing (`npm run test:storybook`).
- Manual verification steps for a reviewer: open the Executions view, bulk *Set labels*, enter key `FOO` / value `bar` — expect an inline error under the row and a disabled confirm button; correct to `foo` and confirm it saves. Same check applies to the single-execution label popover and the flow-run form's label inputs.
- No backend change in this PR — that half already shipped in `v2.0.0-rc12`.

### 🤖 AI Authors

Why did the cat refuse to type an uppercase label key? Because it only walks in lowercase on the keyboard. 🐱


elevatebart (approved):
Looks legit, let me test this out

elevatebart:
Works, though I do not like the UX of having the same error repeated on each line.

Screenshot 2026-09-03 at 11 45 05

I would prefer the error to be close to the save button (at the bottom) and the faulty filed to get an icon and a color change. 
Design validation would be needed. Bug fixed.

elevatebart (approved):

loicmathieu:
@elevatebart I thought having the error next to the offending field is a best  practice.
Imagine you have 10 labels and the 5th one is in error, it's more obvious to have the error next to the 5th one instead of at the bottom of the form no?

PR #18911, opened 2026-09-02
fix(scheduler): include the schedule occurrence at the backfill start date

### LINKED ISSUE #18504: [Bash2.0] A backfill runs the occurrence at its end date but not at its start date
## Describe the issue

A backfill runs the occurrence at its end date but skips the one at its start date.

## Steps to reproduce

1. Create a flow with a Schedule trigger, cron `* * * * *`.
2. Open the flow -> Triggers -> run a Backfill from `1:00:00` to `1:05:00`.
3. Check the `trigger.date` of the created executions.

## Actual behaviour

Executions are created for `1:01` through `1:05`, but not for `1:00`.

## Visuals

Image

## Environment

- Kestra Version: 2.0.0-rc10

rishi-jat:
## Summary

A backfill on a `Schedule` trigger treated the range as `(start, end]`: the occurrence at the end date ran, the one at the start date did not.

A backfill from `01:00` to `01:05` on `* * * * *` now creates executions for `01:00` through `01:05`.

Fixes #18504

## Root cause

Creating a backfill set `currentDate = start`, then computed the first evaluation with exclusive `nextExecution(start)`. That jumps to the next cron tick, so `01:00` was skipped.

End was already inclusive (`isAfter(end)`). Only start was exclusive.

`eval()` was already inclusive. Making `nextEvaluationDate()` inclusive in general would re-fire the same tick.

## Change

In `TriggerState.backfill()`, when `currentDate` is unset, seed the cursor at `start.minusNanos(1)` instead of `start`.

The existing exclusive `nextExecution()` then lands on start if start is a cron tick. After that, `currentDate` is advanced to that first tick and later steps stay exclusive.

`currentDate` is left alone when it is already set (pause/resume, in-progress backfill).

## Tests

- Create path: start on a tick, start between ticks, start == end on a tick
- Empty range (`start == end` between ticks) still completes immediately (fixed clock)
- Scheduler loop: first execution `trigger.date` is the start tick
- Existing tests that assumed exclusive start were updated

## How to verify

1. Flow with `cron: "* * * * *"`
2. Backfill `01:00:00` → `01:05:00` (past window)
3. `trigger.date` values: `01:00`, `01:01`, `01:02`, `01:03`, `01:04`, `01:05`

### DIFF
--- a/core/src/main/java/io/kestra/core/scheduler/model/TriggerState.java
+++ b/core/src/main/java/io/kestra/core/scheduler/model/TriggerState.java
@@ -231,7 +231,9 @@ public final class TriggerState implements TriggerId {
             backfill = backfill
                 .toBuilder()
                 .end(backfill.getEnd() != null ? backfill.getEnd() : ZonedDateTime.now(clock))
-                .currentDate(backfill.getCurrentDate() != null ? backfill.getCurrentDate() : backfill.getStart())
+                // Exclusive nextExecution(start) skips a cron tick that falls on start.
+                // Seed just before start so the first evaluation is at-or-after start.
+                .currentDate(backfill.getCurrentDate() != null ? backfill.getCurrentDate() : backfill.getStart().minusNanos(1))
                 // captured once, on backfill creation: pausing re-enters this method while
                 // nextEvaluationDate points inside the backfill window.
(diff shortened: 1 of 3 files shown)


fhussonnois (approved):
LGTM! Thank you for your contribution!

PR #18913, opened 2026-09-02
feat(executions): harmonize task run action menu between gantt and to…

### LINKED ISSUE #18390: [UI] Harmonize task run action menu between Gantt and topology
### Feature description

On the Gantt page and on the Overview page inside the topology, there is a menu to do actions on a taskrun.

Both menu diverge on functionality: we should use a shared component so both offers the same action.

### Gantt

Image

### Topology

Image

hemantmm:
…pology

closes: #18390

---

### ✨ Description

Exposed a `#task-actions` slot in the `@kestra-io/topology` package's `TaskNode.vue` to allow consumers to override the node menu. In `LowCodeEditor.vue`, the `TaskRunActions.vue` component is now injected into this slot when viewing an active execution. Topology-specific actions are passed down into `TaskRunActions` (filtering out duplicates like `outputs` and `replay` in favor of the execution-native ones) so they can be rendered seamlessly in a single unified dropdown menu.

https://github.com/user-attachments/assets/53904e19-e83c-4056-9e49-01edfac8e655


flcarre (changes requested):
Thanks for this. Exposing a slot on `TaskNode` and letting the consumer decide what goes in the node menu is the right seam: the topology package keeps owning the node, and the execution-specific menu stays in `ui/src` where its dependencies live. The diff is small for what it achieves, CI is green across all fourteen checks, and filtering `outputs`/`replay`/`edit` in favour of the execution-native equivalents is the correct call, since `TaskRunActions` already renders `Outputs`, `Restart isReplay` and `TaskEdit`.

Three things I would like resolved before merge.

## 1. The playground gets this menu too, and it puts two very different "Delete" actions in one list

Issue #18390 is about the Gantt and the topology on the execution Overview page, both execution views. But the slot fires on `taskProps.execution && taskProps.taskRun`, and that condition is also true in the flow editor: `LowCodeEditorWrapper` passes `isReadOnly` from `flowStore.isReadOnly`, and the playground sets `executionsStore.execution` in `stores/playground.ts:276`. So as soon as a playground run produces a task run, the editor's node menu goes through this branch.

The filter only removes `outputs`, `replay` and `edit`. `delete` and `add-error` survive, because they are gated on `!readOnly` and the editor is not read-only. The merged menu then renders, in this order:

1. the node actions, `delete` among them, `danger: true`, at the **top** of the list
2. the execution actions, `delete logs` among them, further down

"Delete" there removes the task from the flow. "Delete logs" removes log lines. They now sit in the same dropdown, and the destructive one is the first thing under the cursor. `delete` also carries `divided: true` precisely because it used to be last in the node menu; leading the list, that divider no longer separates anything meaningful.

Two ways out, both fine by me: add `delete` and `add-error` to the excluded keys, or gate the whole slot override on the read-only execution context this PR actually targets, so the editor keeps the plain `NodeMenu` it has today. The second is closer to the issue's scope.

## 2. `taskRuns[0]` collapses a looping task to its first iteration

`TaskNode` deliberately exposes both `:taskRuns="taskRunsWithDynamicChildren"` and `:taskRun="taskRuns[0]"` on the slot. The call site binds only the singular one, so for a task with dynamic children every action in `TaskRunActions` (outputs, restart, change status, download and copy logs) targets iteration zero, silently.

That is the case where the two menus diverge most today: the Gantt renders one `TaskRunActions` per task run, so the user picks the iteration. Harmonising the menu while dropping the iteration is a partial answer to the issue. Either use the plural slot prop to offer the iterations, or say in the PR that the node menu is deliberately first-iteration only and open a follow-up.

## 3. `nodeActions?: any[]`

The type already exists, right next to the component you are now importing:

```ts
// packages/topology/src/nodes/NodeMenu.vue
export interface NodeAction {
    key: string;
    label: string;
    icon: unknown;
    ...
}
```

It is just not re-exported from the package barrel. Since this PR already edits `packages/topology/src/index.ts` to export `NodeMenu`, one more line makes the prop typeable:

```ts
export type {NodeAction} from "./nodes/NodeMenu.vue"
```

and then `nodeActions?: NodeAction[]`. That also makes the `.filter(a => ...)` at the call site type-checked, so a renamed action key stops being a silent no-op.

## Smaller

* The `v-for` block added to `TaskRunActions.vue` and the `.node-action--danger` rule are copied verbatim from `NodeMenu.vue`, style block included. Two copies of the same menu-item rendering will drift. Rendering `NodeMenu`'s items, or extracting the item template, would keep one.
* The slot is `task-actions` in `TaskNode.vue`, `taskActions` in `Topology.vue` and `#taskActions` at the call site. The sibling slot next to it is `taskDetails`, so camel case looks like the house style; worth aligning so the three names grep as one thing.
* The list of filtered keys is three string literals in a template expression. Once the prop is typed, hoisting them to a named constant next to the template would make the intent ("these are already covered by TaskRunActions") readable without cross-referencing `TaskNode`.

## Not blocking

No test covers the new behaviour. `packages/topology/tests/units/nodes/TaskNode.test.ts` already mounts the node, so a case pinning that the slot replaces `NodeMenu` when provided, and that the filtered keys do not reach the merged menu, would fail on the commit before this one. The filtering is behaviour, not plumbing, and it is the part most likely to rot when an action key is renamed.

hemantmm:
Thanks for the review and the detailed feedback! I’ve implemented your requested changes:

1. **Playground "Delete" overlap:** I gated the slot override in `LowCodeEditor.vue` on `isReadOnly` (`v-if="isReadOnly && taskProps.execution && taskProps.taskRun"`). The editor/playground now correctly retains the standard `NodeMenu` and avoids mixing the destructive flow-authoring actions with the execution ones.
2. **`taskRuns[0]` iteration collapse:** To keep the scope of this PR focused on harmonizing the base action menus, I've left it deliberately targeting the first iteration for now. I will open a follow-up issue to design and implement a multi-iteration UI selector for the Topology dropdown.
3. **`nodeActions?: any[]` cleanup:**
   - **Typed Prop:** Exported `NodeAction` from `@kestra-io/topology/src/index.ts` and updated `TaskRunActions.vue` to strictly use `nodeActions?: NodeAction[]`.
   - **DRY Template:** Extracted the duplicate `<KsDropdownItem>` rendering block into a new `NodeMenuItem.vue` component and reused it in both `NodeMenu.vue` and `TaskRunActions.vue`.
   - **Consistent Casing:** Renamed `task-actions` to `taskActions` in `TaskNode.vue` so it matches the house style and aligns seamlessly with `Topology.vue` and `LowCodeEditor.vue`.
   - **Hoisted Filter Keys:** Moved the excluded literal array out of the template and into an `EXCLUDED_NODE_ACTIONS` constant inside `<script setup>` to make the intent clearer.

flcarre (commented):
Good follow-up on the two blocking issues from the first pass.

Resolved and verified. Gating the slot override on `isReadOnly` is the right fix for the playground "Delete" collision, and it is structurally sound rather than a surface `v-if`: I traced `LowCodeEditor.isReadOnly` to the `Topology` prop, into `generateGraph`, into `isReadOnlyTask`, to `TaskNode`'s `readOnly`, which is the same flag that suppresses `edit`, `delete` and `add-error` in the `actions` computed. In the one branch where `TaskRunActions` renders, those destructive node actions cannot exist at all. `nodeActions?: any[]` is now typed through the re-exported `NodeAction`, and the duplicated `KsDropdownItem` markup is extracted into `NodeMenuItem.vue`, reused by both `NodeMenu.vue` and `TaskRunActions.vue`. That is the single source of truth this PR is meant to deliver, not two lists edited to look alike.

Not resolved. The `taskRuns[0]` collapse for looping tasks is still there, and it is where the two menus diverge most: Gantt renders one row per iteration, while topology now silently targets iteration zero for outputs, restart, change-status and logs. Last review asked for either the plural slot prop or an explicit scope-out with a follow-up issue. The PR comment promises the issue but none exists. Please open it now and link it in the description, or fix it here. A promise in a comment thread is not discoverable once this merges.

`TaskRunActions.vue` also still carries its own `.node-action--danger` rule after the extraction. It is redundant, since the extracted component owns that class, so it is exactly the copy that drifts on the next colour change. Drop it.

Smaller: the merged dropdown puts the topology-only actions ahead of the execution actions, which was not in the issue's screenshots and is worth a quick design sanity check on ordering. And the previous review's non-blocking ask for a `TaskNode.test.ts` case covering slot-replaces-`NodeMenu` and the exclusion filter is still open.

None of this is a functional regression, so not a hard block, but I would like the follow-up issue actually opened and the dead CSS dropped before merge.

hemantmm:
Thanks for the follow-up!

- CSS: Removed the redundant `.node-action--danger` rule from `TaskRunActions.vue`.
- Ordering: I moved the topology node actions to the bottom of the dropdown in `TaskRunActions`, so the core execution actions take priority at the top.
- Tests: I've added a test case to `TaskNode.test.ts` that verifies `NodeMenu` is correctly replaced by the slot and that the filtered items don't leak through.

hemantmm:
**Iteration collapse:** I've created the follow-up issue here: https://github.com/kestra-io/kestra/issues/18987

ui/src/components/inputs/LowCodeEditor.vue
```diff
@@ -56,6 +56,18 @@
...
                 </slot>
             </template>
+            <template #taskActions="taskProps">
+                <TaskRunActions
+                    v-if="isReadOnly && taskProps.execution && taskProps.taskRun"
+                    class="node-action-button"
+                    :taskRun="taskProps.taskRun"
+                    :execution="taskProps.execution"
+                    :flow="flowStore.flow"
+                    :nodeActions="taskProps.actions.filter(a => !EXCLUDED_NODE_ACTIONS.includes(a.key))"
```
flcarre: Small nicety now that these two menus are merged: none of the topology actions kept here (`description`, `condition`, `logs`, `open`, `expand`, `show-details`) set `divided`, so they get appended right after the existing WorkerInfo/logs actions with no visual seam — a bit unclear that this is actually two menus. Worth marking the first appended one as `divided`?

```suggestion
                    :nodeActions="taskProps.actions
                        .filter(a => !EXCLUDED_NODE_ACTIONS.includes(a.key))
                        .map((a, i) => i === 0 ? {...a, divided: true} : a)"
```

Purely optional — skip it if the menu already reads fine visually in your recording.

ui/src/components/inputs/LowCodeEditor.vue
```diff
@@ -232,14 +244,17 @@
...
     import * as YAML_UTILS from "@kestra-io/topology/flow-yaml-utils"
+    import TaskRunActions from "../executions/TaskRunActions.vue"
     import {useEditorBindings} from "../../composables/useEditorBindings"
     import {loadTaskRunOutputs} from "../../composables/useTaskRunOutputs"
 
+    const EXCLUDED_NODE_ACTIONS = ["outputs", "replay", "edit"]
+
     import {TOPOLOGY_CLICK_INJECTION_KEY} from "../no-code/injectionKeys"
     import BlockTaskPicker from "../no-code/blocks/BlockTaskPicker.vue"
     import {useTaskPicker} from "../no-code/blocks/useTaskPicker"
```
flcarre, line 261: Nit: `EXCLUDED_NODE_ACTIONS` is declared in the middle of the import block, which breaks the "imports together at the top" convention used elsewhere in this file.

```suggestion
    import TaskRunActions from "../executions/TaskRunActions.vue"
    import {useEditorBindings} from "../../composables/useEditorBindings"
    import {loadTaskRunOutputs} from "../../composables/useTaskRunOutputs"
    import {TOPOLOGY_CLICK_INJECTION_KEY} from "../no-code/injectionKeys"
    import BlockTaskPicker from "../no-code/blocks/BlockTaskPicker.vue"
    import {useTaskPicker} from "../no-code/blocks/useTaskPicker"

    const EXCLUDED_NODE_ACTIONS = ["outputs", "replay", "edit"]
```

ui/src/components/executions/TaskRunActions.vue
```diff
@@ -116,6 +121,8 @@
     import TaskEdit from "../flows/TaskEdit.vue"
     import SubFlowLink from "../flows/SubFlowLink.vue"
     import AiIcon from "../ai/AiIcon.vue"
+    import {NodeMenuItem, type NodeAction} from "@kestra-io/topology"
+
```
flcarre: Nit: this leaves a double blank line before `const props`.

```suggestion
    import {NodeMenuItem, type NodeAction} from "@kestra-io/topology"

```

flcarre (commented):
Thanks for tackling this — harmonizing the Gantt and Topology action menus has been on the radar for a while, and the approach here is a nice, minimal way to do it. Two things I especially liked:

- Extracting `NodeMenuItem.vue` out of `NodeMenu.vue` and exporting it from `index.ts` so `TaskRunActions.vue` can reuse the exact same dropdown-item markup instead of re-implementing it — good DRY move that keeps the two menus from drifting apart again.
- The `TaskNode.vue` `taskActions` slot with the `NodeMenu` fallback is backward-compatible for every other `<Topology>` consumer, and the new Vitest test (`"should replace NodeMenu when the taskActions slot is provided..."`) actually exercises that fallback/override behavior rather than restating the template.

One thing worth a second look before merge, plus a few smaller suggestions below.

### Could you double-check the multi-taskRun case?

`TaskNode.vue`'s `taskActions` slot only exposes a single task run (`:taskRun="taskRuns[0]"`), but a graph node can have more than one `taskRun` (a looped/parallel task, or a task that was manually restarted) — `onReplayTask()` in this same file already branches on `taskRuns.length > 1` to open a run-picker, and the outputs drawer iterates `taskRuns` (plural) for the same reason. `EXCLUDED_NODE_ACTIONS` drops the topology-native `outputs`/`replay` actions (which were multi-run-aware) in favor of `TaskRunActions`'s `Outputs`/`Restart`, which both take a single required `taskRun` with no run-picker of their own.

Could you try this against a flow with a looped task (e.g. `ForEachItem`) in the Topology tab and confirm Outputs/Restart still do the right thing when a node has more than one run? If they don't, the fix is probably either keeping the native multi-run-aware actions, or teaching `TaskRunActions` to accept the full `taskRuns` array for those two entries.

### A couple of smaller suggestions

- `EXCLUDED_NODE_ACTIONS` is a plain `string[]` with no compile-time link to `NodeAction.key` (typed as a bare `string` in `NodeMenu.vue`). If a key gets renamed later, this filter silently stops excluding it and the item reappears duplicated. Worth exporting a typed key union or a `NODE_ACTION_KEYS` const from `NodeMenu.vue` at some point so both sides check against the same source of truth — not blocking for this PR.
- Neither `NodeMenuItem.vue` nor the new `nodeActions` prop on `TaskRunActions.vue` has a direct test — `TaskNode.test.ts`'s new test stubs `NodeMenu` out entirely, so it never exercises `NodeMenuItem`'s `divided`/`icon`/danger-class rendering, and `TaskRunActions.vue` has no test file at all. A small Vitest test (or story, per this repo's preference for Storybook over unit tests) mounting either component and asserting the passthrough would close that gap.
- Not asking you to fix scope here, but noticed the harmonization currently only goes one way: Gantt's `TaskRunActions` call sites don't pass `nodeActions`, so Topology now shows more than Gantt does (`description`/`condition`/`show-details`). If that's deliberate for this PR, no action needed — just flagging in case the linked issue expects both directions.

Ran this through a separate design-system-focused pass as well (colors, tokens, component reuse, i18n, `:deep()` usage) and it came back clean — `NodeMenuItem` is a legitimate, honest extraction with no element-plus/hex-color/`:deep()` issues. Nice work overall, looking forward to getting this in.

hemantmm:
Thanks for the review!

- **Multi-taskRun Case:** I've updated the filter logic. When `taskRuns.length > 1`, we now preserve the native multi-run-aware outputs and replay actions from the topology, and dynamically hide the single-run specific Outputs/Restart from TaskRunActions. This brings back the native multi-run behavior in the unified menu!
- **Divided Action:** Added the `divided: true` property to the first appended topology node action so there is a visual seam.
- **Nits:** Fixed the import block sorting and the double blank line.

flcarre (commented):
All three things from my last review are genuinely fixed — I traced the emit/handler chain rather than just reading the diff. Multi-run filtering, the divider, and the import-order nits are all correct on HEAD.

Two smaller things worth a note before merge, not blockers:
1. Fix-with-AI and the SubFlowLink action sit outside the new `taskRuns.length > 1` guard, so for a looping task with multiple runs they still silently evaluate against `taskRuns[0]` — the same class of bug the multi-run fix targeted, just not in scope for #18987. Worth adding to that follow-up.
2. Still no test for the actual harmonization logic — the one new test only proves the slot mechanism works, not the multi-run branch, the `divided` mapping, or `EXCLUDED_NODE_ACTIONS`.

Nit: the divider added before the first appended action can land with nothing above it in the multi-run branch when neither Fix-with-AI nor SubFlowLink render — a stray separator line.

hemantmm:
Thanks for the review!

1. I've updated the follow-up issue to explicitly include fixing `Fix-with-AI` and `SubFlowLink` for the multi-run branch.
2. I also pushed a commit to address the nit. The divider is now guarded by `!(taskProps.taskRuns?.length > 1)`, so it won't be added in the multi-run branch where there are no preceding actions.

flcarre (approved):
Re-reviewed at 64ab62f2 (post the `fix: nit divider` commit).

**Divider nit — verified fixed.** `LowCodeEditor.vue:69` now reads:
```
.map((a, i) => i === 0 && !(taskProps.taskRuns?.length > 1) ? {...a, divided: true} : a)
```
Confirmed: in the multi-run branch no item gets `divided: true` (correct, since there are no preceding single-run items to separate from there), and the single-run branch keeps marking the first appended action as divided. Matches what you described.

**Deferred item (Fix-with-AI / SubFlowLink for multi-run) — I checked, and I'm OK with this going to the follow-up issue.** I see #18987 now has your comment explicitly calling out that `Fix-with-AI` and `SubFlowLink` in `TaskRunActions.vue` still sit outside the `taskRuns.length > 1` guard and evaluate against `taskRuns[0]` only. That's consistent with the "iteration 0 only" limitation you already flagged for this PR, so bundling it with the multi-iteration selector work in #18987 rather than special-casing it here makes sense — no need to hold this PR for it.

**Design-system pass on the current head:** clean. No element-plus, no duplicated components (`NodeMenuItem` is a good dedup, not a reinvention), no `:deep()`/`.el-*` introduced by this PR, no hardcoded colors — actually the opposite: this PR fixes two dead/nonexistent CSS variables in `Topology.vue` (`--ks-border-primary` → `--ks-border-default`, `--ks-button-background-secondary-hover` → `--ks-btn-secondary-bg-hover`) to tokens that actually exist and are used elsewhere in the DS. Nice catch.

One very minor nit, non-blocking: no regression test covers the divider-guard ternary itself (`TaskNode.test.ts` covers the slot-replacement/filtering behavior but not this specific one-liner). Not worth holding up the PR for.

Approving.

PR #18406, opened 2026-08-21
fix: fix grammar issues in english UI text

aj-emerich:

PR #18920, opened 2026-09-02
fix(core): translate the hindi kill message instead of leaving it in english

MilosPaunovic:
Related to https://github.com/kestra-io/kestra/pull/18406.

## What

Fixes the failing `Translations - Key parity check` on `develop`, in both `kestra` and `kestra-ee`.

The Hindi bulk-kill toast held the untouched English source:

```json
"executions killed": "<code>{executionCount}</code> execution(s) killed",
```

## Why it happened

The translation model judges that string untranslatable into Hindi and returns it verbatim - reproduced 8/8 against the exact production prompt, while the sibling `executions deleted` translates cleanly 8/8. The generator wrote the English text back as the Hindi value *and* recorded its fingerprint, so the key looked settled: no later run re-picked it up, and the scheduled `auto-translate-ui-keys` workflow would never have healed it on its own. Only the gate's English-copy rule caught it.

The `EE` half of the gate passes; it is the `OSS` half that fails, which is why https://github.com/kestra-io/kestra-ee/pull/10252 showed a red translations check despite touching only `EE` files.

## The fix

The prompt already carries a German and a Polish rule for this exact verb, both present because kill and stop collapse into near-synonyms and produce nonsense. Hindi had no such rule, so the model declined the string rather than choosing a word. Adding the matching Hindi rule makes it translate stably (5/5 identical output across the three kill strings), keeps the reserved `execution(s)` in English like its siblings, and leaves every other message untouched.

The prompt lives only in `OSS` and `kestra-ee` imports it, so this one change covers both repos. `EE`'s own `hi.json` was already clean and needs no companion `PR`.

## Verification

The dependency-free CI gate, run locally at the `develop` tip of both repos:

| Scope | Before | After |
|-------|--------|-------|
| `oss` | fails on 1 key | `Translation check passed (scope: oss)` |
| `ee` | passes | passes |
| `all` (`OSS` + `EE` together) | fails | `Translation check passed (scope: all)` |

`fingerprints.json` is deliberately untouched - the English source did not change, so its hash is still current.

## Backport

Not needed. `releases/v2.0.x`, `releases/v1.3.x` and `releases/v1.0.x` all still hold a real Hindi value (`<code>{executionCount}</code> निष्पादन kill किए गए`) and the pre-https://github.com/kestra-io/kestra/pull/18406 English source, so none of them fails the check.


MilosPaunovic:
@aj-emerich Dumb problem, but a problem nonetheless. It's sorted now.

ui/src/translations/hi.json
```diff
@@ -819,7 +819,7 @@
     "executions unqueue": "<code>{executionCount}</code> execution(s) अनक्यूड",
     "executions force run": "<code>{executionCount}</code> execution(s) जबरन चलाए गए",
     "executions restarted": "<code>{executionCount}</code> execution(s) पुनः आरंभ की गईं",
-    "executions killed": "<code>{executionCount}</code> execution(s) killed",
+    "executions killed": "<code>{executionCount}</code> execution(s) समाप्त",
```
Piyush-r-bhaskar: समाप्त 😁....well translated

MilosPaunovic:
Backported to:

- `releases/v2.0.x` branch via https://github.com/kestra-io/kestra/pull/19126.

PR #18921, opened 2026-09-02
refactor(core): add a @TenantId constraint for the tenant-id format

fhussonnois:
### 🔗 Related Issue

No issue — found while fixing https://github.com/kestra-io/kestra-ee/issues/10592, which was caused
by an id whose length nobody had bounded.

### ✨ Description

The tenant-id format was written out as a literal `@Pattern(regexp = "^[a-z0-9][a-z0-9_-]*")` on 27
models across OSS and EE. This declares it once as `@TenantId`, next to the existing
`@Rfc1123Label`, and applies it to the 14 OSS sites.

Consolidating the copies surfaced two things the duplication was hiding:

- **`TriggerContext` carried `"^[a-z0-9][a-z0-9_-]"` — no quantifier.** Bean Validation matches the
  whole value, so that pattern accepted a two-character tenant id and rejected every other one.
- **The format had no upper bound at all.** `@TenantId` caps it at 100 characters, so an over-long
  tenant id is rejected as invalid input instead of reaching an insert that cannot succeed.

100 is the narrowest `tenant_id` column in the schema (EE's `invitations`), which makes the bound
provably non-breaking: nothing longer than that can be stored today. A tenant id is also a URL path
segment, which is its own reason to keep it short.

`McpServer.id` and `ApiMcpServer.id` keep their own `@Pattern`: they share the regex today but are
server ids, not tenant ids.

### 🛠️ Backend Checklist

### 📝 Additional Notes

`TenantIdTest` mirrors `Rfc1123LabelTest`: accepted and rejected forms through `ModelValidator` and
through the static `isValid` helper, the Bean Validation null contract, and an explicit case for the
two-character bug above so that regex cannot come back unnoticed. `:core:test --tests TenantIdTest`
passes; `:core` and `:webserver` compile.

**Companion PR:** https://github.com/kestra-io/kestra-ee/pull/10656, which applies `@TenantId` to
the 13 EE models. That PR does not compile without this one, so this should merge first.

**Follow-up, not fixed here.** The survey behind the 100-character bound also found that four EE
tables declare `tenant_id` with a *different* width per backend — `security_integrations` has three
different widths for one table. Filed as
https://github.com/kestra-io/kestra-ee/issues/10654. The cap added here is what currently keeps any
of them from being reachable.

### DIFF
--- a/core/src/main/java/io/kestra/core/mcp/models/McpServer.java
+++ b/core/src/main/java/io/kestra/core/mcp/models/McpServer.java
@@ -12,6 +12,7 @@ import io.kestra.core.models.SoftDeletable;
 import io.kestra.core.queues.event.BroadcastEvent;
 import io.kestra.core.utils.Enums;
 import io.kestra.core.utils.IdUtils;
+import io.kestra.core.validations.TenantId;
 
 import io.swagger.v3.oas.annotations.Hidden;
 import jakarta.validation.constraints.NotBlank;
@@ -23,7 +24,7 @@ import jakarta.validation.constraints.Pattern;
  */
 public record McpServer(
     @Hidden
-    @Pattern(regexp = "^[a-z0-9][a-z0-9_-]*") String tenantId,
+    @TenantId String tenantId,
 
     @NotNull
     @NotBlank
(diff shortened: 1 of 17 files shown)


brian-mulier-p (approved):
LGTM

PR #18924, opened 2026-09-02
fix(flows): exclude drafts from the flow export endpoint

### LINKED ISSUE #311: Ensure that Git Sync doesn't push Flows saved as Drafts to Git
We need to validate if this is happening and fix if drafts are pushed to git because they should not. We should only push to git flows that are fully saved, excluding drafts.

Malaydewangan09:
`/flows/export/by-query` returns a ZIP of YAML sources, and `draft` is revision metadata that never appears in the source. Consumers cannot tell a draft apart, which is why the draft filters in `plugin-git` and `plugin-ee-git` never worked.

closes https://github.com/kestra-io/plugin-git/issues/311


fdelbrayelle (approved):
You'll need to do a new Java SDK version + upgrade the plugin using it right @Malaydewangan09 ?

Malaydewangan09:
No SDK release or plugin bump needed.

PR #18938, opened 2026-09-02
fix(core): keep explicit nulls nested in task outputs

### LINKED ISSUE #110: [Bash2.0] outputType RECORDS drops nulls and ignores dropNulls false
### Describe the issue

`outputType: RECORDS` drops null values and ignores `dropNulls: false`. `outputType: STORE` on the same input honours it.

With `onError: "NULL"` and `dropNulls: false`, only `outputType` differing:

- `STORE` writes `{a:1,b:null}`, correct
- `RECORDS` returns `[{"a":1}]`, the `b` key is gone

So there is no way to get an explicit null out of `RECORDS`.

Image

Run the flow and read the `show` logs, which print the RECORDS output and the STORE file side by side.

```yaml
id: v_records_strips_nulls
namespace: company.team

tasks:
  - id: w
    type: io.kestra.plugin.core.storage.Write
    extension: .json
    content: '[{"a":1,"b":null}]'

  - id: as_records
    type: io.kestra.plugin.transform.Map
    from: "{{ outputs.w.uri }}"
    outputType: RECORDS
    onError: "NULL"
    dropNulls: false
    fields:
      a: a
      b: b

  - id: as_store
    type: io.kestra.plugin.transform.Map
    from: "{{ outputs.w.uri }}"
    outputType: STORE
    onError: "NULL"
    dropNulls: false
    fields:
      a: a
      b: b

  - id: show
    taskRunner:
    inputFiles:
      store.ion: "{{ outputs.as_store.uri }}"
    commands:
      - 'echo "RECORDS:"; echo ''{{ outputs.as_records.records }}'''
      - 'echo "STORE:"; cat store.ion; echo'
```

Note: `onError: NULL` must be quoted, otherwise YAML parses it as a null and the setting silently falls back to `FAIL`.

### Environment

Kestra Version: 2.0.0-rc10, commit 9900dcf
Setup: Standalone with Postgres, Kestra EE

jymaire:
Closes https://github.com/kestra-io/plugin-transform/issues/110.

## Problem

An explicit `null` nested inside a task output — a null value in a `Map` or `List`, not a top-level bean property — was turned into an **absent key**. A downstream expression like `{{ outputs.produce.records[0].b }}` then failed with `Unable to find 'b'` instead of rendering an empty value. `outputType: RECORDS` in plugin-transform was one victim, but the loss happened in core, so it was a whole class of bugs affecting every output producer.

## Fix

Two independent hops dropped the null; each is fixed narrowly so nothing else changes.

**1. Output storage.** `Output.toMap()` used the mapper configured with the single-argument `setDefaultPropertyInclusion(NON_NULL)`, which drops null map entries and collection elements at every depth. A dedicated mapper separates the two inclusion settings — `NON_NULL` for bean properties, `ALWAYS` for map content — and is used only by `Output.toMap()`. Because it sits on the `Output` interface, it covers every producer (task outputs, trigger variables, subflow, schedule) at once; flow YAML, API payloads and revision diffs keep using the existing mappers. Null *properties* of an output stay opt-in per-property via `@JsonInclude(ALWAYS)`, the pre-existing pattern.

**2. Rendering (opt-in).** Even with the null stored, serializing a whole collection back to a string dropped it again: `{{ records | toJson }}` used the same NON_NULL mapper. Flipping the shared render mapper globally would change every rendered JSON payload in Kestra — HTTP request bodies (missing ≠ null for many APIs), files written from templates, cache/dedup keys — so null inclusion stays opt-in: `{{ records | toJson(includeNulls=true) }}` keeps null map content, while the argument-less filter and bare interpolation are unchanged. Direct access (`records[0].b`) and data flow into downstream tasks already work from fix (1); this only covers the case of re-serializing a collection to a JSON string.

## Known limitation

`MapUtils.merge` cannot distinguish a present-null from an absent key (`valueB == null` is treated as "no value"). `TaskOutputService#computeOutputs` uses `merge` only for batched-iteration tasks whose outputs land flat at the same level, so a stale earlier-iteration value can survive there. That is a different failure from the one reported here (a present-null with no prior value, which works), and fixing it would change `merge` semantics for every variable merge in Kestra, so it is out of scope.

## Evidence

- `JacksonMapperTest` — both new overloads, plus a characterization test pinning that plain `toMap` still drops null map entries, so the two mappers' difference is asserted rather than assumed.
- `NullOutputTest` — the two null-inclusion cases (bean property vs map content) side by side on the memory runner.
- `AbstractRunnerTest#nullContentOutput` — the full round-trip through the output store, run on **H2, MySQL and Postgres**.
- `ToJsonFilterTest` — asserts `toJson` still drops nulls while `toJson(includeNulls=true)` keeps them.
- Verified live against a standalone build: without the fix the reported error reproduces verbatim (`Unable to find 'b'`); with it, `records[0].b` renders empty and `{{ records | toJson(includeNulls=true) }}` renders `[{"a":1,"b":null}]`.


jymaire:
# QA

## Flow

``` yaml
id: null_content_repro
namespace: company.team
tasks:
  - id: w
    type: io.kestra.plugin.core.storage.Write
    extension: .json
    content: '[{"a":1,"b":null}]'

  - id: produce
    type: io.kestra.plugin.transform.Map
    from: "{{ outputs.w.uri }}"
    outputType: RECORDS
    onError: "NULL"
    dropNulls: false
    fields:
      a: a
      b: b

  - id: render_null
    format: "[{{ outputs.produce.records[0].b }}]"

  - id: render_sibling
    format: "[{{ outputs.produce.records[0].a }}]"

```

## Resultats

### Develop
image

### fix
image
image
image

jymaire:
# QA of second commit

## Flow 

```yaml
id: v_records_strips_nulls
namespace: company.team

tasks:
  - id: w
    type: io.kestra.plugin.core.storage.Write
    extension: .json
    content: '[{"a":1,"b":null}]'

  - id: as_records
    type: io.kestra.plugin.transform.Map
    from: "{{ outputs.w.uri }}"
    outputType: RECORDS
    onError: "NULL"
    dropNulls: false
    fields:
      a: a
      b: b

  - id: as_store
    type: io.kestra.plugin.transform.Map
    from: "{{ outputs.w.uri }}"
    outputType: STORE
    onError: "NULL"
    dropNulls: false
    fields:
      a: a
      b: b

  - id: show
    taskRunner:
    inputFiles:
      store.ion: "{{ outputs.as_store.uri }}"
    commands:
      - 'echo "RECORDS:"; echo ''{{ outputs.as_records.records | toJson(includeNulls=true) }}'''
      - 'echo "STORE:"; cat store.ion; echo'

```

## Screenshot

image

core/src/main/java/io/kestra/core/serializers/JacksonMapper.java
```diff
@@ -102,6 +102,32 @@ public static Map<String, Object> toMap(Object object) {
         return MAPPER.convertValue(object, MAP_TYPE_REFERENCE);
     }
 
+    /**
+     * Keeps null map entries and collection elements ({@code ALWAYS} content inclusion) while still omitting null
+     * bean properties ({@code NON_NULL} value inclusion). The single-argument {@code setDefaultPropertyInclusion}
+     * used by the other mappers sets both, which is why they drop null map entries.
+     */
+    private static final ObjectMapper NULL_CONTENT_MAPPER = MAPPER
```
loicmathieu:
```suggestion
    private static final ObjectMapper MAPPER_WITH_NULL = MAPPER
```

core/src/main/java/io/kestra/core/serializers/JacksonMapper.java
```diff
@@ -102,6 +102,32 @@ public static Map<String, Object> toMap(Object object) {
...
+        .copy()
+        .setDefaultPropertyInclusion(JsonInclude.Value.construct(JsonInclude.Include.NON_NULL, JsonInclude.Include.ALWAYS));
+
+    public static ObjectMapper ofJsonKeepingNullValues() {
```
loicmathieu:
```suggestion
    public static ObjectMapper ofJsonWithNullValues() {
```

core/src/main/java/io/kestra/core/serializers/JacksonMapper.java
```diff
@@ -102,6 +102,32 @@ public static Map<String, Object> toMap(Object object) {
...
+    }
+
+    /** @see #NULL_CONTENT_MAPPER */
+    public static Map<String, Object> toMapKeepingNullValues(Object object, ZoneId zoneId) {
+        return NULL_CONTENT_MAPPER
+            .copy()
+            .setTimeZone(TimeZone.getTimeZone(zoneId.getId()))
+            .convertValue(object, MAP_TYPE_REFERENCE);
+    }
+
```
loicmathieu: Remove them, they are niche, and do that at use site

loicmathieu (changes requested):

core/src/main/java/io/kestra/core/models/tasks/Output.java
```diff
@@ -3,20 +3,33 @@
...
 import java.util.Map;
 import java.util.Optional;
+import java.util.TimeZone;
 
+import com.fasterxml.jackson.core.type.TypeReference;
 import io.kestra.core.models.flows.State;
 import io.kestra.core.serializers.JacksonMapper;
 
 public interface Output {
+    TypeReference<Map<String, Object>> MAP_TYPE_REFERENCE = new TypeReference<>() {};
```
loicmathieu: Optional, can be removed and its usage replaced by `JacksonMapper.MAP_TYPE_REFERENCE`

loicmathieu (approved):
LGTM

Issue #18925, opened 2026-09-02
Topology artifacts show raw Pebble expressions: resolve them host-side via a bound slot prop


elevatebart:
## Problem

Plugin UI artifacts in the topology display raw Pebble expressions instead of values — `{{ vars.region }}` where the user wants `us-east-1`. This is the problem #16874 described, with the same `taskRunner` example.

#16874 was closed as completed, but only its backend half shipped:

| #16874 required | State on `develop` |
|---|---|
| Backend `resolveForDisplay(template, flow, execution?)` | ✅ `DisplayExpressionRenderer` + `POST /api/v1/{tenant}/expressions/render` |
| A `displayContext` shipped alongside topology/artifact API responses | ❌ no occurrence of `displayContext` in `core`/`webserver` |
| A client-sdk `resolveExpression(value, context)` helper, "reusable by all plugin artifact/topology components" | ❌ no occurrence of `resolveExpression`/`DisplayContext` in `ui/src`, `design-system`, `kestra-sdk` |
| Anything calling the endpoint | ❌ zero non-generated callers in OSS or EE |

So `/expressions/render` exists, is hardened, and is dead code, while the artifacts it was built for render raw expressions.

Separately, the GCP BigQuery artifact's own call to `renderExpressions` was removed for a **structural** reason: plugin UI code should not be able to trigger server-side code. That constraint is right and should be kept — but it removed the display behaviour without the replacement being in place.

## Why not the shape #16874 specified

#16874 proposes shipping a `displayContext` to the plugin plus an exported `resolveExpression` helper for plugins to call. The slot contract already solves this problem twice, in a better shape.

Two props in `ui/packages/slot-contracts/src/topology-details.ts` are host-implemented, pre-bound closures rather than data. `fetchOutputs(query?)` resolves to this task run's outputs, and `fetchMetrics(query?)` to a page of its metric entries — each with the execution, the tenant and the task already bound by the host. Their own comments say it:

> Outputs of this task's current task run, fetched on call so an artifact that doesn't render them costs no request. […] Resolves to `{}` outside an execution.

> This task's metrics in the current execution — searchByExecution with the execution, the tenant and this task already bound.

`LowCodeEditor.vue` binds them per task at the call site (`:fetchOutputs="fetchTaskOutputs(...)"`). The plugin cannot choose what to ask for — it calls a function and gets data back. That is exactly the structural property the removal was protecting.

## Proposal

[...]

PR #18941, opened 2026-09-02
feat(topology): host-bound resolveForDisplay for artifact expressions

elevatebart:
### 🔗 Related Issue

Part of https://github.com/kestra-io/kestra/issues/18925 — PR 1 of the split proposed there, so no closing keyword.

### ✨ Description

Plugin artifacts in the topology show `{{ vars.region }}` where the user wants `us-east-1`. The backend renderer built for this (`/expressions/render`) has no caller, and letting each plugin resolve expressions itself would pin a resolver copy in every bundle.

`resolveForDisplay(value)` joins `fetchOutputs` / `fetchMetrics` in the topology slot contracts: host-implemented, host-bound, so a plugin calls a function and gets a display string back without deciding what the host does.

- Substitutes plain dotted paths **per occurrence** — `"prefix-{{ vars.x }}/{{ vars.y }}"` resolves both.
- `secret('K')` is always `[secret: K]`, filter or no filter.
- Everything else comes back raw: filters, function calls, indexing, arithmetic, `env`, `kv`, a `{% %}` block anywhere in the value, and a looped task's outputs (no single iteration is *the* value). A confidently wrong value in a topology node is worse than a visible `{{ }}`, because nobody re-checks one that looks resolved.
- The context mirrors the executor's own map: `vars` / `flow` always, `execution` / `inputs` / `outputs` / `trigger` / `labels` only during a run — so the pre- vs post-execution rule falls out of the map rather than needing a rule of its own.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit` — 240 files, 2213 tests)
- [ ] Screenshots or video recordings attached showing the `UI` changes — **nothing to show yet**: no OSS component renders these artifacts, so the pixels only change once the plugin-side PR lands (PR 4: plugin-gcp, plugin-aws).

### 📝 Additional Notes

Deliberately left out:

- **PR 2** (batched `/expressions/render` for `env` / `globals`) — the prop is synchronous, so the host has to pre-resolve the graph into a cache and re-render when it fills. Same contract, but its own design discussion.
- **PR 5** (`source` removal) — can't go before the plugins stop needing it.
- **PR 3** is not actionable: `ui-ee/src/utils/contextExpression.ts` doesn't exist on EE `develop`, so there's no second resolver to fold in yet.

`LowCodeEditor` now parses the flow source once and feeds both the task index and the display context, instead of parsing per consumer.

Why did the plugin bundle refuse to resolve its own expressions? It had no idea what it was talking about — it just kept saying *"{{ meow }}"*. 🐱

https://claude.ai/code/session_01J1i7kHvNkSr4wEWf8TWerv

---
_Generated by [Claude Code](https://claude.ai/code/session_01J1i7kHvNkSr4wEWf8TWerv)_

### DIFF
--- a/ui/packages/slot-contracts/src/topology-details.ts
+++ b/ui/packages/slot-contracts/src/topology-details.ts
@@ -28,6 +28,9 @@ export const propsSchema = z.object({
     fetchOutputs: z.custom<(query?: {
         taskRunId?: string
     }) => Promise<Record<string, unknown>>>().optional(),
+    // Resolves a value's expressions against the flow — and, in an execution, that run — for display,
+    // host-side: `secret('K')` is masked as `[secret: K]`, and what it cannot resolve stays raw.
+    resolveForDisplay: z.custom<(value: string) => string>().optional(),
     // This task's metrics in the current execution — searchByExecution with the execution, the
     // tenant and this task already bound. A looped task has one entry set per iteration; pass
     // `taskRunId` to narrow to one. Resolves to an empty page outside an execution.
--- a/ui/src/components/inputs/LowCodeEditor.vue
+++ b/ui/src/components/inputs/LowCodeEditor.vue
@@ -53,6 +53,7 @@
                         :progress="taskProgress(taskProps.data.node?.task?.id)"
                         :fetchOutputs="fetchTaskOutputs(taskProps.data.node?.task?.id)"
                         :fetchMetrics="fetchTaskMetrics(taskProps.data.node?.task?.id)"
+                        :resolveForDisplay="resolveForDisplay"
                     />
                 </slot>
             </template>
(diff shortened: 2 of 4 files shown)


elevatebart:
`kestra-ee/backend-tests` is red on this PR, and it isn't this PR's.

The `trigger-ee` job never got a verdict — the EE webhook answered `HTTP 500` with an empty body:

```
Raw webhook response (HTTP 500)
##[error]EE compile check inconclusive for ref claude/github-issue-18925-jqk5h3 (no compile verdict; HTTP 500). Likely a transient runner error — re-run the job.
```

The same status is failing right now on unrelated PRs — #18902 (frontend-only) and #18938 (core) — and this diff is four frontend files with no Java in it, so there is no EE compilation for it to break and no fix to port.

Everything else on this head is green: Frontend Tests, Build Artifacts, spelling, codecov bundles. I'll re-run the `trigger-ee` job once the workflow run finishes (GitHub refuses a re-run while it's still going), and will report rather than churn if it comes back red.

---

fdelbrayelle (approved):

elevatebart:
too complicated to be sustainable

PR #18944, opened 2026-09-02
refactor(iam): name no-ACL repository reads *WithoutAcl and split findLatestForStates

nkwiatkowski:
## Problem

Several repository reads bypass namespace ACL by design — the executor's `ExecutionStateStore`, the scheduler's `TriggerStateStore`, and trigger/audit detail lookups all run with no user context — but they were named like the ACL-aware reads sitting next to them, so calling the wrong overload silently skips ACL. Separately, `findLatestForStates` enforced ACL inconsistently across backends: JDBC applied it incidentally through the inherited method, Elasticsearch applied none.

## Fix

- `ExecutionStateStore.findById(String)` → `findByIdWithoutAcl(String)`.
- `TriggerRepositoryInterface` / `TriggerStateStore` `findById(TriggerId)` → `findByIdWithoutAcl(TriggerId)`.
- Add `ExecutionRepositoryInterface.findLatestForStatesWithoutAcl`, and make `findLatestForStates` ACL-aware and consistent across backends.

These are behaviour-preserving renames plus one new no-ACL variant; the no-ACL reads now announce themselves in their names.

## Evidence

Existing executor, scheduler and repository test suites updated to the new names and green.

Companion PR (Elasticsearch asset-purge fix, public-App fix, EE renames, ACL tests): kestra-io/kestra-ee#10692, same branch `fix/ee-repository-acl-holes`.

### DIFF
--- a/core/src/main/java/io/kestra/core/repositories/ExecutionRepositoryInterface.java
+++ b/core/src/main/java/io/kestra/core/repositories/ExecutionRepositoryInterface.java
@@ -65,6 +65,15 @@ public interface ExecutionRepositoryInterface extends QueryBuilderInterface<Exec
      */
     Optional<Execution> findLatestForStates(String tenantId, String namespace, String flowId, List<State.Type> states);
 
+    /**
+     * Same as {@link #findLatestForStates(String, String, String, List)} but without applying the caller's
+     * namespace ACL, for trusted callers that gate access themselves (e.g. a public App rendered anonymously).
+     * Defaults to the ACL-aware variant for backends that enforce no ACL of their own.
+     */
+    default Optional<Execution> findLatestForStatesWithoutAcl(String tenantId, String namespace, String flowId, List<State.Type> states) {
+        return findLatestForStates(tenantId, namespace, flowId, states);
+    }
+
     ArrayListTotal<Execution> find(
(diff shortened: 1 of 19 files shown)


core/src/main/java/io/kestra/core/repositories/ExecutionRepositoryInterface.java
```diff
@@ -65,6 +65,15 @@ default Optional<Execution> findById(String tenantId, String id) {
      */
     Optional<Execution> findLatestForStates(String tenantId, String namespace, String flowId, List<State.Type> states);
 
+    /**
+     * Same as {@link #findLatestForStates(String, String, String, List)} but without applying the caller's
+     * namespace ACL, for trusted callers that gate access themselves (e.g. a public App rendered anonymously).
+     * Defaults to the ACL-aware variant for backends that enforce no ACL of their own.
+     */
+    default Optional<Execution> findLatestForStatesWithoutAcl(String tenantId, String namespace, String flowId, List<State.Type> states) {
```
loicmathieu: Why a default? If all implementors override it I would prefer to avoid any default implementation

loicmathieu (changes requested):

PR #18948, opened 2026-09-02
fix(ci): pin third-party GitHub Actions to commit SHAs

josemimbre:
### 🔗 Related Issue

No related issue — found by running `zizmor` locally against `.github/workflows/` while auditing the CI setup. Opening as a proactive hardening PR.

### ✨ Description

**Problem:** every third-party action in the workflows was referenced by a floating major-version tag (`@v7`, `@v4`, `@v9`...). A tag can be moved by the action's maintainer — or by an attacker who compromises their account — to point at different code without this repository changing at all. This is the exact GitHub Actions supply-chain vector `zizmor`'s `unpinned-uses` rule flags.

**Fix:** resolved each tag to the commit SHA it currently points at and pinned every occurrence to that SHA, keeping the version visible as a trailing comment (`@<sha> # v7.0.1`) so the file stays readable and diffable on the next bump. Covers every third-party action across `.github/workflows/`: `actions/checkout`, `actions/setup-node`, `actions/setup-java`, `actions/download-artifact`, `actions/upload-artifact`, `actions/create-github-app-token`, `actions/github-script`, `github/codeql-action/*`, `docker/*`, `codespell-project/*`, `dataaxiom/ghcr-cleanup-action`, `dorny/paths-filter` and `gradle/actions/*`.

Deliberately left `kestra-io/actions/...@main` unpinned: those are the org's own shared reusable workflows and composite actions, not an external dependency. Pinning them would freeze this repo on one commit of the platform team's shared CI tooling instead of tracking `main`, trading the convenience of automatically picking up their fixes for a supply-chain risk that doesn't really apply within the same org. Happy to revisit if the team prefers otherwise.

**Evidence:** `zizmor --persona=pedantic .github/workflows/` no longer reports any `unpinned-uses` finding, other than the intentional `kestra-io/actions@main` references noted above.

### 📝 Additional Notes

Pure YAML change, no runtime behavior touched: same actions, same versions, just addressed by SHA instead of by a mutable tag.

### DIFF
--- a/.github/workflows/auto-translate-ui-keys.yml
+++ b/.github/workflows/auto-translate-ui-keys.yml
@@ -66,7 +66,7 @@ jobs:
             fetch-depth: 0
 
         - name: Set up Node
-          uses: actions/setup-node@v7
+          uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
           with:
             node-version: "24.x"
 
--- a/.github/workflows/auto-translate-ui-keys.yml
+++ b/.github/workflows/auto-translate-ui-keys.yml
@@ -59,7 +59,7 @@ jobs:
       matrix:
         branch: ${{ fromJSON(needs.plan.outputs.branches) }}
     steps:
-        - uses: actions/checkout@v7
+        - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
           name: Checkout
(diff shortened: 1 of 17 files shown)


AcevedoR:
thanks for bringing this up, I'm looking right now at integrating a static check for this maybe with zizmor in our CI

AcevedoR (approved):

PR #18950, opened 2026-09-02
fix(kv): prevent namespace and key UID collisions

### LINKED ISSUE #18506: [Bash2.0] Distinct KV entries silently overwrite each other on a namespace/key collision
## Describe the issue

Two KV entries whose namespace and key concatenate to the same string collide: creating the second silently overwrites the first.

## Steps to reproduce

1. In namespace `company.team`, set KV key `x_y` to `valueA`.
2. In namespace `company.team_x`, set KV key `y` to `valueB`.

## Actual behaviour

The `company.team` / `x_y` entry is gone. Only one KV entry remains, holding `valueB`.

## Visuals

https://github.com/user-attachments/assets/c8012a90-f434-4269-8fb7-3ab3ad57fd88

## Environment

- Kestra Version: 2.0.0-rc10

AradhyaSharma31:
**Related Issue**

Closes https://github.com/kestra-io/kestra/issues/18506.

**Description**

Fixes a collision in KV metadata UIDs when a namespace and key combination can produce the same underscore-separated UID as another namespace and key combination.

The KV metadata UID now uses a pipe separator, ensuring that distinct (namespace, key) pairs produce distinct UIDs. A migration is also added to update existing KV metadata records to the new UID format.

**Demo Video**

Watch Here If embedded video doesn't work: []https://jam.dev/v/0a774d8f-c564-4c90-b66c-671965664efd

[jam-video.webm](https://github.com/user-attachments/assets/b57c9404-7461-4658-b4af-bfa77257f94e)

### DIFF
--- a/core/src/main/java/io/kestra/core/models/kv/PersistedKvMetadata.java
+++ b/core/src/main/java/io/kestra/core/models/kv/PersistedKvMetadata.java
@@ -98,6 +98,6 @@ public class PersistedKvMetadata implements SoftDeletable<PersistedKvMetadata>, 
 
     @Override
     public String uid() {
-        return IdUtils.fromParts(getTenantId(), getNamespace(), getName(), String.valueOf(getRevision()));
+        return IdUtils.fromPartsPipeDelimited(getTenantId(), getNamespace(), getName(), String.valueOf(getRevision()));
     }
 }
--- a/core/src/main/java/io/kestra/core/utils/IdUtils.java
+++ b/core/src/main/java/io/kestra/core/utils/IdUtils.java
@@ -12,6 +12,7 @@ import com.google.common.hash.Hashing;
 abstract public class IdUtils {
     private static final HashFunction HASH_FUNCTION = Hashing.md5();
     private static final char ID_SEPARATOR = '_';
+    private static final char ID_SEPARATOR_PIPE = '|';
 
     public static String create() {
         return FriendlyId.createFriendlyId();
@@ -27,6 +28,10 @@ abstract public class IdUtils {
 
     public static String fromParts(String... parts) {
         return fromPartsAndSeparator(ID_SEPARATOR, parts);
+    }
+
+    public static String fromPartsPipeDelimited(String... parts) {
+        return fromPartsAndSeparator(ID_SEPARATOR_PIPE, parts);
     }
 
     public static String fromPartsAndSeparator(char separator, String... parts) {
--- a/core/src/test/java/io/kestra/core/repositories/AbstractKvMetadataRepositoryTest.java
+++ b/core/src/test/java/io/kestra/core/repositories/AbstractKvMetadataRepositoryTest.java
@@ -281,6 +281,37 @@ public abstract class AbstractKvMetadataRepositoryTest {
         assertThat(namespaces).doesNotContain(deletedNamespace);
     }
 
+    @Test
+    void shouldNotOverwriteWhenNamespaceAndKeyConcatenateToSameString() throws IOException {
+        String tenantId = TestsUtils.randomTenant();
+        PersistedKvMetadata first = PersistedKvMetadata.builder()
+            .tenantId(tenantId)
+            .namespace("company.team")
+            .name("x_y")
+            .revision(1)
+            .build();
+        PersistedKvMetadata second = PersistedKvMetadata.builder()
+            .tenantId(tenantId)
+            .namespace("company.team_x")
+            .name("y")
+            .revision(1)
+            .build();
+
+        kvMetadataRepositoryInterface.save(first);
+        kvMetadataRepositoryInterface.save(second);
+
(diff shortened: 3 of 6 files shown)


flcarre (changes requested):
The core fix is correct: switching `PersistedKvMetadata.uid()` to a `|` delimiter resolves the reported collision, since `|` cannot appear in a tenantId, a namespace or a key given their existing validation patterns. Two things should be addressed before merge.

First, `V2_0_13KvMetadataUidMigration` rewrites `kv_metadata.key` for every existing row and has no test. This module already has the pattern for exactly this shape of change: `V2_0_11PluginAutoInstallMigrationTest` spins up an in-memory H2 database, seeds rows, runs `migrate()` and asserts the resulting state, including idempotency (a second run must be a no-op) and the corrupt-JSON failure path. A one-way rewrite of a persisted column is the case where an untested regression is both expensive and hard to notice, which is precisely what the repo's testing guidance asks for a test on.

Second, this patches the KV case only. `IdUtils.fromParts` is the default underscore joiner and it backs several other `uid()` implementations that combine namespace and path-like values the same way. `NamespaceFileMetadata.uid()` (tenantId + namespace + path + revision) is structurally identical to the bug being fixed and is still exploitable today: a file at path `x_y` in namespace `team` can silently collide with path `y` in namespace `team_x`. The same joiner also backs `FlowId`, `TriggerId`, `Dashboard`, `ExecutionKilledTrigger`, `MultipleConditionWindow` and `AssetUser`. At minimum, open a follow-up issue and link it here. Shipping this as "KV collisions are fixed" without noting the sibling exposure will come back as a duplicate bug report.

Minor: the migration loads the whole table and updates row by row rather than using jOOQ's `.batch()`. Worth a look for large multi-tenant installs, not blocking.

AradhyaSharma31:
@flcarre Thanks for the quick review, I'll start working on the suggestions right away.

flcarre (changes requested):
Thanks for addressing both review points — the migration test and batched updates look good, and the follow-up issue for the sibling `fromParts` exposure is exactly right.

One thing still needs fixing before this is mergeable: `shouldNotOverwriteWhenNamespaceAndKeyConcatenateToSameString` in `AbstractKvMetadataRepositoryTest.java` builds both `PersistedKvMetadata` objects without `.revision(...)`. Since `version` is a NOT NULL generated column derived from the JSON, and Jackson's `NON_NULL` inclusion drops a null `revision` entirely, `save()` will throw before the test ever reaches its assertions — this can't currently pass on any backend. Add `.revision(1)` to both builders, matching `buildTestKvDescription`.

Two smaller things, non-blocking:
- `V2_0_13KvMetadataUidMigration`: when `updates` is empty you log "nothing to migrate" but still call `batch(updates).execute()` — return early instead.
- The migration trusts stored JSON as-is for the no-`|`-in-parts invariant, without re-running the `@Pattern` validation. Not exploitable today, but worth a note if this is a risk you want tracked.

Re-request once the test is fixed.

AradhyaSharma31:
@flcarre Thanks for pointing out, I've made the necessary changes. Please review again when you get a chance.

flcarre (approved):
The revision fix landed exactly as requested: both `PersistedKvMetadata` builders in `shouldNotOverwriteWhenNamespaceAndKeyConcatenateToSameString` now set `.revision(1)`, so the test actually reaches its assertions on every backend instead of throwing on `save()`. The `V2_0_13KvMetadataUidMigration` also picked up the early-return on an empty `updates` list — good catch on your end too.

Verified against the current head (`45a18869bd`):
- `PersistedKvMetadata.uid()` now delegates to `IdUtils.fromPartsPipeDelimited`, resolving the reported collision.
- `V2_0_13KvMetadataUidMigrationTest` covers migrate, idempotency, already-correct rows, the ambiguous-boundary collision case, empty values, and corrupt JSON — matches the `V2_0_11PluginAutoInstallMigrationTest` pattern this module already uses.
- The migration batches updates via jOOQ `.batch()` and returns early when there's nothing to do.
- Follow-up issue kestra-io/kestra#19015 tracks the sibling `IdUtils.fromParts` exposure (`NamespaceFileMetadata`, `FlowId`, `TriggerId`, etc.) as agreed, so this PR's scope stays limited to KV.

Two purely cosmetic nits, not blocking:
- `V2_0_13KvMetadataUidMigration.java` line 3: import ordering puts a local `io.kestra.*` import ahead of `java.io.IOException`, inconsistent with sibling migrations.
- `V2_0_13KvMetadataUidMigrationTest.java` line 168: `private java.util.List<String> fetchKeys()` — inline fully-qualified name instead of an import.

Approving.

loicmathieu:
Hi,
In fact, the scope is broader than just namespace/KV as all entities are affected, this is more tricky that what it appears so we'll keep this issue internal until we know what we want to do with it.

For the record https://github.com/kestra-io/kestra/pull/19081 proposes a global solution for it.

loicmathieu:
Closing it as if we do something about it, we'll do it broadly.

PR #18951, opened 2026-09-03
fix(core): reset pagination when returning to a list page

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra-ee/issues/7241.

Issue: after paginating a list, say Executions page 10, navigating elsewhere and coming back through the menu restored page 10 instead of starting over.
Fix: the URL restore now skips page and size, so only filters and sort are brought back when a page is reopened with a bare URL.
Now: returning to any list lands on page 1 with the default size, while restored filters and shared links with explicit pagination keep working as before.

### DIFF
--- a/ui/src/composables/useRestoreUrl.ts
+++ b/ui/src/composables/useRestoreUrl.ts
@@ -1,6 +1,7 @@
 /**
- * Persists a page's URL query (filters, sort, pagination, etc.) to sessionStorage
- * so it can be restored when the user navigates back to that page with no query.
+ * Persists a page's URL query (filters, sort, etc.) to sessionStorage so it can be
+ * restored when the user navigates back to that page with no query. Pagination
+ * (page/size) is excluded (PY-1529) from restoration so returning lands on the default page.
  *
  * Auto-saves on every same-path query change and auto-restores on mount when the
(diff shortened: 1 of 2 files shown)


MilosPaunovic (approved):

karelkryda:
@Piyush-r-bhaskar, the original issue of mine was meant to fix the problem with remembering the page **number**, **NOT** the page **size**. Remembering the page **size** is proper design and currently broken in our upgraded `1.3.39`.

Could you, please, restore the functionality to remember the page **SIZE**, and drop only the page **NUMBER** when navigating between dashboard pages?

**Intended now:** returning to any list lands on page 1 with the **previously defined** size.

Thanks!

**Same applies to backported version to `1.3.38+` under #18953.**

Piyush-r-bhaskar:
> drop only the page NUMBER when navigating between dashboard pages?

@karelkryda what do you mean by Dashboard pages ? is it pages in general or Dashboard ?

karelkryda:
@Piyush-r-bhaskar _"Dashboard pages"_ like whole Kestra web UI :)

Piyush-r-bhaskar:
Thank You 🙏

PR #18958, opened 2026-09-03
feat(scripts): write the execution context to a file for script tasks

fdelbrayelle:
Backend half of kestra-io/libs#4. Pairs with kestra-io/libs#40, which is the Python reader for the file this PR writes.

## What

New `executionContext` property on script tasks. When enabled, the execution metadata is serialized to `.kestra-execution-context.json` in the working directory, right where input files are already materialized:

```yaml
  - id: transform
    executionContext: true
    script: |
      from kestra import context

      csv_path = context.outputs.extract.uri
      run_label = context.labels.env
```

instead of today's only option:

```yaml
    script: |
      csv_path = "{{ outputs.extract.uri }}"
```

Because it is on `AbstractExecScript`, all 18 language tasks inherit it — Python, Node, Shell, Powershell, R, Julia, Ruby, Go, Groovy, PHP, Perl, Lua, Deno, Bun, .NET, JBang, Jython, Nashorn.

## Why

Three problems with interpolating Pebble into script source, from [the customer report on the issue](https://github.com/kestra-io/libs/issues/4#issuecomment-3426701818):

1. Scripts synced as namespace files and edited in an IDE are no longer valid, lintable code.
2. Orchestration concerns get mixed into business logic.
3. Any value holding a `"` or a newline breaks the generated script — [the point acknowledged as valid](https://github.com/kestra-io/libs/issues/4#issuecomment-3426799742).

Point 3 is the one that cannot be worked around. Verified against `kestra/kestra:v2.0.0-rc13`: with the file, an input whose value is `a "quoted" and\nmultiline value\n` arrives in Python byte-exact.

## Design decisions, and what is deliberately left open

**Disabled by default.** The file holds the same values the task could already render — secret-typed inputs included, since `DefaultRunContext.decryptSecretSubtrees` decrypts `inputs`, `outputs` and `trigger` before `getVariables()` returns them. Writing that to the working directory on every script task would be a silent change to what lands on disk, so it stays an explicit choice. If the KIP prefers on-by-default, that needs a secret-stripping pass first, which needs access to the still-encrypted raw variables — a larger change than this one.

**`envs` and `globals` excluded.** Both already reach the script as environment variables; writing them again only widens what is on disk for no gain.

**No allowlist of the remaining roots.** Everything else in `runContext.getVariables()` is written as-is, so EE's extra `namespace` root (added by `io.kestra.ee.runners.RunContextFactory`) flows through with no EE-side change. An allowlist built from `RunVariables.EXECUTION_CONTEXT_PATHS` would have silently dropped it.

Two things I did not decide, and that the KIP should:

- `outputFiles: ["*"]` will collect the context file along with everything else. Should it be filtered out?
- The property name. `executionContext: true` reads well but is a bare boolean where `namespaceFiles` uses a nested `enabled:`. Happy to switch to the nested shape for consistency.

## Tests

New `CommandsWrapperExecutionContextTest`:

- with `executionContext: true`, the script copies the file out; the parsed JSON contains `flow`, `execution`, `task`, `taskrun` and excludes `envs`, `globals`
- by default the file is absent, asserted by the script itself with `test ! -f`

`./gradlew :script:test --tests '…runners.*'` → **13 passed**, so the two existing `CommandsWrapper` suites still pass after the added constructor argument.

## No `kestra-ee` PR needed

`CommandsWrapper` lives in this repo's `script` module and no EE main source references it — only `core-ee` tests do, and those cover output encryption, not the working directory. EE's `RunContextFactory` builds on `RunVariables.DefaultBuilder`, so its `namespace` root is picked up here for free.

### DIFF
--- a/script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/CommandsWrapper.java
+++ b/script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/CommandsWrapper.java
@@ -2,6 +2,7 @@ package io.kestra.plugin.scripts.exec.scripts.runners;
 
 import java.io.IOException;
 import java.net.URI;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Duration;
 import java.util.*;
@@ -16,6 +17,7 @@ import io.kestra.core.models.tasks.runners.*;
 import io.kestra.core.models.tasks.runners.DefaultLogConsumer;
 import io.kestra.core.runners.FilesService;
 import io.kestra.core.runners.RunContext;
+import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.utils.IdUtils;
 import io.kestra.core.utils.NamespaceFilesUtils;
 import io.kestra.plugin.core.runner.Process;
@@ -32,6 +34,10 @@ import static io.kestra.core.utils.Rethrow.throwFunction;
 @AllArgsConstructor
 @Getter
 public class CommandsWrapper implements TaskCommands {
+    static final String EXECUTION_CONTEXT_FILE_NAME = ".kestra-execution-context.json";
+
+    private static final Set<String> EXECUTION_CONTEXT_EXCLUDED_VARIABLES = Set.of("envs", "globals");
+
     private RunContext runContext;
 
     private Path workingDirectory;
@@ -98,6 +104,9 @@ public class CommandsWrapper implements TaskCommands {
     @With
     private KotlpOptions kotlp;
 
+    @With
+    private boolean executionContext;
+
     public CommandsWrapper(RunContext runContext) {
         this.runContext = runContext;
         this.workingDirectory = runContext.workingDir().path();
@@ -130,7 +139,8 @@ public class CommandsWrapper implements TaskCommands {
             enableOutputDirectory,
             timeout,
             targetOS,
-            kotlp
+            kotlp,
+            executionContext
         );
     }
 
@@ -150,6 +160,23 @@ public class CommandsWrapper implements TaskCommands {
         this.env.putAll(envs);
 
         return this;
+    }
+
+    /**
+     * Writes the execution context as JSON into the working directory, so a script can read its
+     * metadata as data instead of interpolating Pebble expressions into its own source.
+     *
+     * `envs` and `globals` are left out: both are already exposed to the script as environment
+     * variables, so writing them again would only widen what ends up on disk.
+     */
+    private void writeExecutionContextFile() throws IOException {
+        Map<String, Object> variables = new TreeMap<>(runContext.getVariables());
+        EXECUTION_CONTEXT_EXCLUDED_VARIABLES.forEach(variables::remove);
+
+        Files.write(
+            this.workingDirectory.resolve(EXECUTION_CONTEXT_FILE_NAME),
+            JacksonMapper.ofJson().writeValueAsBytes(variables)
+        );
     }
 
     public <T extends TaskRunnerDetailResult> ScriptOutput run() throws Exception {
@@ -176,6 +203,10 @@ public class CommandsWrapper implements TaskCommands {
 
         if (this.inputFiles != null) {
             FilesService.inputFiles(runContext, runnerVars, this.inputFiles);
+        }
+
+        if (this.executionContext) {
+            this.writeExecutionContextFile();
         }
 
         if (this.isKotlpEnabled()) {
--- a/script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/CommandsWrapper.java
+++ b/script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/CommandsWrapper.java
@@ -17,6 +17,7 @@ import io.kestra.core.models.tasks.runners.*;
 import io.kestra.core.models.tasks.runners.DefaultLogConsumer;
 import io.kestra.core.runners.FilesService;
 import io.kestra.core.runners.RunContext;
+import io.kestra.core.runners.RunVariables;
 import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.utils.IdUtils;
 import io.kestra.core.utils.NamespaceFilesUtils;
@@ -36,7 +37,14 @@ import static io.kestra.core.utils.Rethrow.throwFunction;
 public class CommandsWrapper implements TaskCommands {
     static final String EXECUTION_CONTEXT_FILE_NAME = ".kestra-execution-context.json";
 
-    private static final Set<String> EXECUTION_CONTEXT_EXCLUDED_VARIABLES = Set.of("envs", "globals");
+    private static final Set<String> EXECUTION_CONTEXT_EXCLUDED_VARIABLES = Set.of(
+        // already handed to the script as environment variables
+        "envs",
+        "globals",
+        // internal plumbing, not execution metadata
+        RunVariables.SECRET_CONSUMER_VARIABLE_NAME,
+        RunVariables.FIXTURE_FILES_KEY
+    );
 
     private RunContext runContext;
 
--- a/script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/CommandsWrapper.java
+++ b/script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/CommandsWrapper.java
@@ -18,6 +18,7 @@ import io.kestra.core.models.tasks.runners.DefaultLogConsumer;
 import io.kestra.core.runners.FilesService;
 import io.kestra.core.runners.RunContext;
 import io.kestra.core.runners.RunVariables;
+import io.kestra.core.runners.WorkingDir;
 import io.kestra.core.serializers.JacksonMapper;
 import io.kestra.core.utils.IdUtils;
 import io.kestra.core.utils.NamespaceFilesUtils;
@@ -35,8 +36,6 @@ import static io.kestra.core.utils.Rethrow.throwFunction;
 @AllArgsConstructor
 @Getter
 public class CommandsWrapper implements TaskCommands {
-    static final String EXECUTION_CONTEXT_FILE_NAME = ".kestra-execution-context.json";
-
     private static final Set<String> EXECUTION_CONTEXT_EXCLUDED_VARIABLES = Set.of(
         // already handed to the script as environment variables
         "envs",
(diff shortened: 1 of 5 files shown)


fdelbrayelle:
## Verified the reader against the bytes this PR actually writes

The QA on kestra-io/libs#40 produced the context file from the flow itself (`inputFiles` + `{{ … | toJson }}`), so it validated the Python reader against **Pebble-serialized** JSON, not against the **Jackson-serialized** JSON this PR emits. Different serializers, so I dumped the real file from a run of this branch and fed those exact bytes to the reader.

That surfaced one defect, fixed in 230e260d10: **`addSecretConsumer` was being written into the file.**

```json
{
    "addSecretConsumer": {},
    "execution": { "id": "475TRMAwOlLfEQ13QC6P2I", "state": "RUNNING", "startDate": "2026-09-03T07:52:08.132772Z", … },
    "flow": { "id": "mockruncontext", "namespace": "io.kestra.core.utils.testsutils", "revision": 1, "tenantId": "main" },
    "kestra": { "url": "…" },
    "labels": {}, "outputs": {}, "parents": [],
    "task": { "id": "test", "type": "test" },
    "taskrun": { "id": "6SES9Q98OcdRcWDY9cC9Zd", "attemptsCount": 0, … }
}
```

`RunVariables.SECRET_CONSUMER_VARIABLE_NAME` is the secret-consumer callback that lives in the variable map — internal plumbing, not execution metadata. It now joins the exclusion set, along with `RunVariables.FIXTURE_FILES_KEY`, and the test asserts its absence.

With that fixed, the same file read back through the Python client:

```
flow: mockruncontext / io.kestra.core.utils.testsutils rev 1
execution: 4igdwjqPrqNlvvyyp5LzpS RUNNING 2026-09-03T07:52:43.983112Z
taskrun: 3mFHH0N9t3aJhQJzKFwkBl attempts 0
kestra url: https://mysuperhost.com/subpath
no plumbing: True
absent key -> 'inputs' is not available in the execution context (available: …)
```

Also worth recording: Jackson writes timestamps as ISO-8601 strings (`2026-09-03T07:52:08.132772Z`), matching what Pebble's `toJson` produced, so the reader needs no type coercion and the earlier QA results carry over.

`:script:test --tests '…runners.*'` → **13 passed**.

Malaydewangan09 (approved):

script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/CommandsWrapper.java
```diff
@@ -152,6 +170,23 @@ public CommandsWrapper addEnv(Map<String, String> envs) {
...
+     *
+     * `envs` and `globals` are left out: both are already exposed to the script as environment
+     * variables, so writing them again would only widen what ends up on disk.
+     */
+    private void writeExecutionContextFile() throws IOException {
+        Map<String, Object> variables = new TreeMap<>(runContext.getVariables());
+        EXECUTION_CONTEXT_EXCLUDED_VARIABLES.forEach(variables::remove);
+
+        Files.write(
+            this.workingDirectory.resolve(EXECUTION_CONTEXT_FILE_NAME),
```
jymaire: **Secrets can be persisted to internal storage through `outputFiles`.**

The context file is written into the working directory, which is then swept by the `outputFiles` globs. Java's `PathMatcher` matches leading dots (unlike a shell), so this file is collected by a plain `*`:

```
executionContext(true) + withOutputFiles(List.of("*"))
  -> outputFiles=[out.txt, .kestra-execution-context.json]
```

`runContext.getVariables()` returns the *decrypted* `inputs` / `outputs` / `trigger` subtrees (`DefaultRunContext.decryptSecretSubtrees`), so a flow combining `executionContext: true` with `outputFiles: ["*"]` — or `["*.json"]`, or a `WorkingDirectory` with `cache.patterns: ["**"]` — permanently persists plaintext secrets into internal storage, downloadable from the Executions UI and exposed as a task output URI.

The schema description acknowledges the file holds secret-typed inputs, and the PR leaves the collection path as an open question. It should be a hard exclusion in `FilesService.outputFiles` / `findAllFilesMatching` rather than a KIP decision — otherwise the feature ships with an exfiltration path that a user hits by writing the most natural glob there is.

script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/CommandsWrapper.java
```diff
@@ -178,6 +213,10 @@ public <T extends TaskRunnerDetailResult> ScriptOutput run() throws Exception {
             FilesService.inputFiles(runContext, runnerVars, this.inputFiles);
         }
 
+        if (this.executionContext) {
+            this.writeExecutionContextFile();
```
jymaire, line 216: **The file is never removed, so it leaks across tasks in a `WorkingDirectory`.**

Inside a `WorkingDirectory` flowable the working directory is shared between child tasks, and nothing deletes `.kestra-execution-context.json` after the runner returns. A second script task that never set `executionContext` still finds the file on disk, so `from kestra import context` silently hands it the *previous* task's context: wrong `task` / `taskrun`, and `outputs` missing everything produced since.

That staleness also makes the `outputFiles` exposure above reachable from a task that never opted in.

Either delete the file in a `finally` around `taskRunner.run`, or write it unconditionally so a disabled task overwrites rather than inherits.

fdelbrayelle:
## End-to-end QA on a server built from this branch

Everything before this was either unit-level (`:script:test`) or a reader test against a hand-written file. This run is the real path: a Kestra server built from `230e260d10`, a flow submitted over the API, the backend writing the file, and the Python client from kestra-io/libs#40 reading it inside the task.

Setup: local CLI from this branch (`2.0.0-SNAPSHOT`), memory queue/repository, local storage, Process task runner, `namespaceFiles` carrying `kestra.py`.

```yaml
  - id: transform
    executionContext: true
    script: |
      from kestra import context
      print("prev output uri:", context.outputs.extract.uri)
```

Result — **SUCCESS**:

```
"executionContext":"true"          ← property accepted on the deserialized flow
STATE=SUCCESS
INFO labels: prod
INFO input repr: 'a "quoted" and\nmultiline value\n'
INFO var: 42
INFO prev output uri: kestra:///qa/context-e2e/executions/…/tasks/extract/…/2440955939531304265.tmp
INFO flow: context_e2e qa rev 2
INFO execution: 1O7MardgoqVJTgedJZGePG RUNNING
INFO taskrun: 5fZAtkNturc9fjxoBLGjIh
INFO roots: ['execution', 'flow', 'inputs', 'kestra', 'labels', 'outputs', 'parents', 'task', 'taskrun', 'tasks', 'vars']
INFO no plumbing: True | no envs: True
INFO EXPECTED: no context file when executionContext is not set
```

What this confirms that the earlier passes did not:

- the property deserializes and reaches `CommandsWrapper` through `AbstractExecScript.commands()`
- the file lands in the working directory where the script actually runs, readable by the Python client with no flow-side boilerplate
- the exclusions hold on a real execution — no `addSecretConsumer`, no `envs`
- the second task in the same flow, without the property, sees no file, so the default really is off

`roots` has no `trigger` because the execution was API-triggered; a scheduled execution adds it.

**Not covered:** `Kestra.outputs()` from inside the same task reported no outputs on the execution record. That is pre-existing library behavior, untouched by this PR, and the minimal memory-repository setup is the likelier explanation, but I did not chase it down and am not claiming it verified.

script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/CommandsWrapper.java
fdelbrayelle: Confirmed and fixed in 661c5956bb — this was exploitable exactly as described.

I reproduced it before fixing: with `executionContext(true)` and `withOutputFiles(List.of("*"))`, `run.getOutputFiles()` contained `.kestra-execution-context.json` alongside `out.txt`, so the decrypted `inputs`/`outputs`/`trigger` subtrees were persisted to internal storage and exposed as a task output URI.

Agreed it belongs in the collection path rather than in the KIP. The name is now reserved as `WorkingDir.EXECUTION_CONTEXT_FILE_NAME` and filtered out in `LocalWorkingDir.findAllFilesMatching`, which is the choke point for `outputFiles`, `WorkingDirectory` `cache.patterns` and `UploadFiles` at once — so `["*"]`, `["*.json"]` and `["**"]` are all covered by one exclusion.

`shouldNeverCollectTheExecutionContextAsAnOutputFile` covers it, and I checked it is not vacuous: reverting just the filter makes it fail.

One limit worth stating — the exclusion guards collection that goes through `WorkingDir`. A task runner living in a plugin repo that expands `outputFiles` globs on its own remote side would need the same reserved-name check; the `finally` delete from the other thread does not help there, since remote collection happens inside `taskRunner.run`.

script/src/main/java/io/kestra/plugin/scripts/exec/scripts/runners/CommandsWrapper.java
fdelbrayelle: Confirmed and fixed in 661c5956bb.

Went with the `finally` delete rather than writing unconditionally — writing on every script task is the default-on behavior this PR deliberately avoids, and it would put the file on disk for tasks that never asked for it.

`shouldRemoveTheExecutionContextOnceTheCommandsAreDone` reuses one run context across two runs, the way a `WorkingDirectory` shares its directory: first task with `executionContext` asserts the file exists, second task without it asserts the file is gone. Reverting just the `finally` makes it fail, so the staleness was real.

The delete is best-effort — an `IOException` is logged at debug rather than failing the task, since the commands have already run by then and the reserved-name exclusion from the other thread already blocks the exposure path a leftover file would open.

jymaire (approved):

fdelbrayelle:
## Correction on the `plugin-scripts` claim

I stated in the PR description that `plugin-scripts` "no longer holds Java sources". That is wrong — it holds the 18 language tasks in per-language subprojects (`plugin-script-python/`, `plugin-script-node/`, …). My earlier check looked for sources at the repo root and found none.

The conclusion does not change, but the reason is different from what I wrote. What actually lives in this repo's `script` module is `AbstractExecScript` and `CommandsWrapper`; every one of those 18 tasks extends `AbstractExecScript`, so declaring the property there makes all of them inherit it with **no change needed in `plugin-scripts`**:

```
Bun, Deno, .NET, Go, Groovy, JBang, Julia, Jython, Lua, Nashorn,
Node, Perl, PHP, Powershell, Python, R, Ruby, Shell
```

each with both its `Script` and `Commands` variant. I verified this by grepping `extends AbstractExecScript` across the repo rather than assuming it this time.

So: one PR here, none in `plugin-scripts`, none in `kestra-ee`. Docs are now covered separately in kestra-io/docs#5489.

PR #18960, opened 2026-09-03
fix(design-system): separate the two halves of a split button with a visible divider

MilosPaunovic:
Closes https://github.com/kestra-io/kestra-ee/issues/10126.
Related to https://github.com/kestra-io/kestra/pull/18962.

## What

The two halves of a `KsDropdown` split button (the action label and the caret that opens the menu) are now separated by a visible divider, on every theme.

## Why

Element Plus paints the split-button divider as a 1px `::before` on the caret button, filled with the generic border colour at half opacity. Against `Kestra`'s button fills that line is one or two shades away from the background, so the label and the caret read as one plain button - the `Download` button in the `Usage` section of the `System Overview` page is the case reported, and the `Save` / `Publish` split button of the `flow` editor and the playground `Run task` button share the same styles.

## How

- `KsDropdown.vue` overrides the caret's `::before` to use `--ks-border-strong` at full opacity, which is visible against the secondary fill on the light, dark and dark-2 themes (`--ks-border-default` is only three units away from the secondary fill on the dark theme, so it does not work there).
- Primary buttons are filled with the primary colour, so they use Element Plus's own button-group divide colour (`--kel-button-divide-border-color`, a translucent white), the same line a primary `KsButtonGroup` already draws between its buttons. Disabled primary buttons fall back to the border token since their fill is grey.
- A `SplitButton` story documents the primary, secondary (`buttonProps: {plain: true}`) and disabled variants, with a play assertion that the divider stands out from the button fill.

No `EE` change is needed: `Stats.vue` already renders the split button through `KsDropdown`.

Backport to `releases/v2.0.x`: https://github.com/kestra-io/kestra/pull/18962.

## Screenshots

`System Overview` > `Usage` > `Download`, before / after:

| Theme | Before | After |
|-------|--------|-------|
| light | before light | after light |
| dark | before dark | after dark |
| dark-2 | before dark-2 | after dark-2 |

The new `SplitButton` story after the fix (primary, secondary, disabled):

story light
story dark
story dark-2

### DIFF
--- a/ui/packages/design-system/src/components/Navigation/KsDropdown/KsDropdown.vue
+++ b/ui/packages/design-system/src/components/Navigation/KsDropdown/KsDropdown.vue
@@ -46,6 +46,20 @@
     @use '../../../assets/styles/el-ns';
     @use 'element-plus/theme-chalk/src/dropdown';
 
+    .kel-dropdown {
+        // Element Plus draws the split-button divider as the border colour at half opacity,
+        // which vanishes against every Kestra button fill; use the strong border token instead
+        .kel-dropdown__caret-button.kel-button::before {
+            background: var(--ks-border-strong);
+            opacity: 1;
+        }
+
+        // a primary button is filled, so use the lighter button-group divide colour instead
+        .kel-dropdown__caret-button.kel-button--primary:not(.is-disabled)::before {
+            background: var(--kel-button-divide-border-color);
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


Pradumnasaraf (commented):
Hey @MilosPaunovic, This PR can't be reviewed because the issue was happening on Enterprise Edition and I can't verify on open source.

PR #17589, opened 2026-07-22
fix(flows): render flow validation error as a single coherent message

Siddhant-Kashyap:
All PRs submitted by external contributors that do not follow this template (including proper description, related issue, and checklist sections) **may be automatically closed**.

As a general practice, if you plan to work on a specific issue, comment on the issue first and wait to be assigned before starting any actual work. This avoids duplicated work and ensures a smooth contribution process - otherwise, the PR **may be automatically closed**.

---

### ✨ Description
Flow editor validation panel was splitting a single YAML parsing error into multiple separate rows, making the message unreadable. SnakeYAML formats its errors with , column N: patterns internally, which collided with the , ? regex used to split distinct constraint messages. Each fragment was then rendered in its own block with padding and border.

Fixed by treating the constraints string as a single error entry, joining multiple distinct constraints with \n instead of ,, and applying white-space: pre-wrap so embedded newlines render correctly in the tooltip.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/17483

### 🎨 Frontend Checklist

_If this PR does not include any frontend changes, delete this entire section._

- [x] Code builds without errors (`npm run build`)
- [ ] All existing E2E tests pass (`npm run test:e2e`)
- [x] Screenshots or video recordings attached showing the `UI` changes

image

### 📝 Additional Notes

Changes are limited to ui/src/stores/flow.ts (split/join logic) and ui/src/components/flows/ValidationError.vue (CSS). No backend changes. The fix also correctly handles the edge case where two distinct constraint sources (API error + namespace permission error) are both present — they now appear separated by a newline within the single error block rather than as separate entries.

PR #18966, opened 2026-09-03
fix(flows): validation errors split on the wrong separator

loicmathieu:
Closes https://github.com/kestra-io/kestra-ee/issues/8967.

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra-ee/issues/8967.

### ✨ Description

Adding `expiredOnly: true` to a `PurgeKV` task made the flow editor report **22 errors**, with a garbled multi-line tooltip, for what the backend correctly reports as a single validation error.

The editor's validation badge split the backend's `constraints` message on `/, ?/` (comma), while the backend only ever joins multiple violations with a newline. A message that legitimately contains commas — Jackson's `Unrecognized field "expiredOnly" (class ...PurgeKV), not marked as ignorable (21 known properties: "id", "type", ...)` — was shattered into one fake error per comma.

This PR:
- Introduces a shared `splitValidationErrors()` helper that splits on newlines instead, and applies it at the three places still splitting on commas (flow store, `TaskEdit`, no-code `Task` panel), reusing it in `flowableBlockOps`' `groupValidationIssuesByTask` too, which already expected newline-delimited input. This restores a fix from #17589 that was inadvertently reverted by #16054 — a regression test (`ui/src/stores/flow.spec.ts`) now guards it.
- Fixes `PurgeKV`'s own `@Example`/`@Schema` doc, which still advertised the `expiredOnly` property directly on the task even though it was removed in `3def65d714`. Copying that stale example is exactly what produced the issue's reproducer.

### 🎨 Frontend Checklist

- [x] Translations are complete if `en.json` changed (no translation keys touched)

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched modules (`./gradlew :core:test --tests "PurgeKVTest"`, `:core:compileJava`)

### 📝 Additional Notes

- No screenshot attached: I couldn't get a local standalone instance to boot in this environment in time (H2 lock from an earlier aborted attempt, then out of time) — happy to add one, or a reviewer can reproduce with the issue's YAML in under a minute.
- The `PurgeKV.java` change is docs-only (no behavior change), so no new backend test was added for it, per the "test earns its place" guidance.
- Why did I choose plain `console.log` for a cat? I didn't — but two engineers walk past a server rack and one says "why is the cat sleeping on the router?" The other replies: "It's checking the up-time."

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/core/src/main/java/io/kestra/plugin/core/kv/PurgeKV.java
+++ b/core/src/main/java/io/kestra/plugin/core/kv/PurgeKV.java
@@ -30,9 +30,7 @@ import lombok.extern.slf4j.Slf4j;
 @Schema(
     title = "Purge keys from the KV store.",
     description = """
-        Deletes keys across Namespaces using a purge `behavior` (default: expired-only). Filter by explicit `namespaces` or `namespacePattern`, optional `keyPattern`, and include/exclude child namespaces.
-
-        Deprecated `expiredOnly` overrides `behavior` if set."""
+        Deletes keys across Namespaces using a purge `behavior` (default: expired-only). Filter by explicit `namespaces` or `namespacePattern`, optional `keyPattern`, and include/exclude child namespaces."""
 )
 @Plugin(
     examples = {
@@ -46,7 +44,9 @@ import lombok.extern.slf4j.Slf4j;
                 tasks:
                   - id: purge_kv
                     type: io.kestra.plugin.core.kv.PurgeKV
-                    expiredOnly: true
+                    behavior:
+                      type: key
+                      expiredOnly: true
                     namespaces:
                       - company
                     includeChildNamespaces: true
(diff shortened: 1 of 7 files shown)


elevatebart (commented):
Why not use a completely different splitter? 
What if an error contains a newline?

Why not return the errors as a JSON object instead?
This splitting feels very dirty to me.

loicmathieu:
@elevatebart I agree this is ugly; this PR only fixed an issue introduced recently. I too think we should provides something better for validation issues as they can be multiple one in a single error. 

We now switch to the RFC 9457 Problem API which should allow to do better but I'll let that as a followup, this PR only fix a regression.

loicmathieu:
See https://github.com/kestra-io/kestra/issues/18972 as a followup

elevatebart (approved):

PR #18988, opened 2026-09-03
fix(dashboards): drill-down page filters, clicked time bucket, drawer reload

### LINKED ISSUE #18496: [Bash2.0] Clicking a Total Executions chart bar opens a Preview that ignores the applied filter
## Describe the issue

On the Executions page, with an advanced filter applied, the Total Executions chart shows the correct filtered count, but clicking a bar opens a Preview that ignores the filter and lists executions that do not match it.

## Steps to reproduce

1. Go to Executions.
2. Add an advanced filter: Namespace In `system`, AND Flow ID In `test`, AND State In `SUCCESS`. The chart and the list both show 2.
3. Click the bar in the Total Executions chart.

## Actual behaviour

The Preview panel lists executions from other namespaces and flows (for example `compa/sdads`, `system/dads`) and from other times, not limited to the applied filter or to the bar's time bucket. Only 2 of the 5 executions shown actually match the filter.

## Visuals

Image

Image

## Environment

- Kestra Version: 2.0.0-rc10
- Edition: Enterprise

RemiBarthe:
Closes https://github.com/kestra-io/kestra/issues/18496.

###  Description

Clicking a bar in a chart opened a preview that ignored the filters applied to the page and the bar's own time bucket: the query was built from the chart's `where` alone, then a default `chartDefaultDuration` was appended. With a filter applied you got rows that did not match it, and on a bar older than that default duration you got an empty list.

Three things now hold when a segment is clicked:

- The page's top-level filters seed the query, through the existing `keepTopLevelFilters`. They are resolved once per click rather than per dimension — resolved in the loop, the last dimension's copy landed back on top of the dimensions already applied and discarded their clicked values, which broke every Bar chart since those always pass two dimensions.
- The clicked bucket's `startDate`/`endDate` pair replaces the default duration instead of joining it, since the backend treats the two forms as mutually exclusive (`TimeLineSearch.extractFrom` throws otherwise). The row's own date is used as the boundary, the backend having already truncated it. With no bucket, the page's own range is carried over so the query is not half-inherited.
- The preview drawer reloads when the target changes and drops superseded responses. It only reloaded on a page change, so a second click showed the first segment's rows.

### Additional Notes

`chartSegmentDrillDown` is kept as a thin facade over the new `chartDrillDownTarget` so its ten existing tests in `charts.spec.ts` still exercise the same behaviour unchanged.

Not covered by tests: `bucketDateRange` is not exported from the `<script setup>` of `TimeSeries.vue`, and the drawer's reload and sequence guard are component-level. Both were checked in the browser instead.


flcarre (commented):
The pagination/filter/time-bucket carry-through and out-of-order response guard are good fixes, and none of the design-system non-negotiables are violated. Before merging I'd want two things looked at:

1. `chartDrillDownTarget` sets `timeWindow` from `context.dateRange` regardless of `descriptor.timeFiltered`, and `buildFullQuery` checks `timeWindow` before `timeFiltered` — so a bucket click on a non-time-filterable descriptor (e.g. Flows) could still push `startDate`/`endDate` onto a full-page query that isn't built to accept them. Gate that branch on `timeFiltered` too.
2. `DrillDownDrawer.spec.ts` is untouched by this PR (byte-identical to develop), so the two behaviors that are actually the point of this PR — the sequence-based stale-response guard and reload-on-reopen — ship with zero test coverage, even though the existing spec already has the mocking scaffolding to add one.

Also: `chartSegmentDrillDown` is now dead code (nothing calls it since `drillDown` was rewritten to call `chartDrillDownTarget` directly) — worth deleting along with its tests rather than carrying it forward.

RemiBarthe:
@flcarre All three addressed in f9ec9150f4.

1. Gated at construction rather than in `buildFullQuery`, so the target never carries a window that the query builder would have to know to ignore. Worth noting the reachability differs by descriptor: `Flows` has no date field (`ID`/`NAMESPACE`/`REVISION`), so `bucketDateRange` can't resolve a bucket there at all. EE `Assets` does (`CREATED`/`UPDATED`), and nothing stops a TimeSeries over it. `TimeSeriesChartValidator` only checks `graphStyle` uniquenes so that was the real path.

2. Added two tests to the existing spec: reload-on-retarget and a superseded fetch resolving last. Both fail against develop's component — verified by reverting. Also asserted `toHaveBeenCalledTimes(2)` so a double-reload regression can't slip through.

3. Deleted. I migrated its ten cases to `chartDrillDownTarget` rather than dropping them — they cover `whereToFilters`, the comparator mapping, nested key-value filters and `registerDrillDown`, all still live behind the new entry point. Only the wrapper was dead, not the behaviour.

flcarre (approved):
Re-reviewed f9ec9150f4 against my three points from the previous pass — all confirmed fixed in the code, not just in the description:

1. **Time window gating** — `chartDrillDownTarget` now decides `timeWindow` at construction based on the descriptor's own `timeFiltered` flag (`chartDrillDown.ts`), so `buildFullQuery` never has to guess. Checked both edge cases you called out: `Flows` (`timeFiltered: false`, no date field at all) and EE `Assets` (`timeFiltered: false` explicitly set in `registerAssetsDrillDown.ts`, despite having `CREATED`/`UPDATED`) — confirmed `TimeSeriesChartValidator` only checks `graphStyle` uniqueness, so nothing else was stopping a TimeSeries over Assets before this fix. Good catch keeping the gate on the descriptor rather than the data source.

2. **Drawer reload tests** — the two new specs (`reload-on-retarget`, `superseded fetch landing last`) correctly exercise the `sequence` guard and the `watch(target) → resetAndReload()` wiring, and the `toHaveBeenCalledTimes(2)` assertion will actually catch a double-reload regression. Traced the `resetAndReload` → `callLoad` path in `KsDataTable` to confirm it drives the component's own `loadData`, not just a UI no-op.

3. **Dead wrapper removal** — confirmed `chartSegmentDrillDown` has zero remaining references anywhere in the tree. The migrated cases in `charts.spec.ts` still cover `whereToFilters`, the comparator table, nested key/value (LABELS/METADATA) filters, and `registerDrillDown` through the new `chartDrillDownTarget` entry point — nothing lost.

Design-system pass on top: no element-plus, no hex/rgb/`:deep()`, `KsDataTable` pagination stays fully controlled (`:currentPage`/`:pageSize` bound, no parent-side `.reload()` call on page click), i18n untouched by this diff.

Two non-blocking nits, neither worth holding this up for:
- `TimeSeries.vue`'s `bucketDateRange` matches the clicked bucket to a row by formatted-label string equality — fine today, but would silently pick the wrong row if two rows ever shared a label. Fallback (page's own time range) degrades safely either way.
- The new fetch-sequence guard in `DrillDownDrawer.vue` lives in the component rather than a store, which is what `ui/AGENTS.md` asks for elsewhere — but this drawer never had a store backing its rows, so it's consistent with the file as it already was.

Approving.

PR #18995, opened 2026-09-03
fix(dashboards): include headers in empty CSV exports

### LINKED ISSUE #18577: [Dashboards] Chart CSV export of an empty chart returns 0 bytes instead of a header row
## Describe the issue

Exporting a dashboard chart with no rows returns `HTTP 200` with an empty body, so the download is a 0 byte file rather than a valid CSV with a header row.

`webserver/src/main/java/io/kestra/webserver/controllers/api/DashboardController.java:535` runs `CSVUtils.toCSV` over an empty result list, which emits nothing at all — no header, no rows.

## Why this is a follow-up

https://github.com/kestra-io/kestra/pull/18530 (closing #18334) handles the symptom in the frontend: `downloadHandler` now refuses to write an empty body and the card tells the user the chart has nothing to export. That stops the unusable file reaching the user, but the endpoint still answers a successful request with a payload that is not a valid CSV, which anyone calling the API directly still has to special-case:

```
POST /api/v1/{tenant}/dashboards/charts/export?format=CSV
no data   -> HTTP 200, 0 bytes
with data -> HTTP 200, "namespace,id,state,flow,start_date" + rows
```

## Expected behaviour

Emit the header row derived from the chart's declared columns, so an empty export is a valid CSV with zero data rows. That makes the response self-describing and lets the frontend drop its empty-body special case later.

Raised by @MilosPaunovic while reviewing https://github.com/kestra-io/kestra/pull/18530, who noted it is a few lines in that method.

## Environment

- Kestra Version: `develop`
- Edition: OSS (reproduces on EE, the endpoint is shared)

---

ToxicCalvin:
Closes https://github.com/kestra-io/kestra/issues/18577.

### 🔗 Related Issue

https://github.com/kestra-io/kestra/issues/18577

### ✨ Description

Dashboard CSV exports returned a 0-byte response when a chart had no results. Data charts now use their declared columns for both empty and populated exports, producing a header-only CSV when empty and keeping the column order consistent when rows are present. Values are written in that same column order.

The existing controller export test now exports the same chart before and after inserting a row, with declared columns in the opposite order to the JDBC result. It replaces the separate empty-result test and verifies both header consistency and value alignment. The unused import identified in review has also been removed.

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched modules (`./gradlew :webserver:test`)

### 📝 Additional Notes

The regression test failed with the previous header-selection logic and passes with the fix. All 1,299 webserver tests passed with no failures or skipped tests in 8m 23s. Spotless completed successfully with Java 25.

Rebased onto `develop`, preserving the upstream `StreamedFile` filename handling.

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/utils/CSVUtils.java
+++ b/webserver/src/main/java/io/kestra/webserver/utils/CSVUtils.java
@@ -14,7 +14,8 @@ import reactor.core.publisher.Flux;
 /**
  * Renders records to RFC 4180 compliant CSV.
  *
- * <p>The keys of the first record define the columns: every subsequent record is projected onto them, so a record
+ * <p>
+ * Columns are either provided explicitly or derived from the first record. Every subsequent record is projected onto them, so a record
  * holding extra or missing keys can never shift the remaining values into the wrong columns.
  */
 public final class CSVUtils {
@@ -28,12 +29,21 @@ public final class CSVUtils {
      * @param lines the records to render, nothing is written when empty
      */
     public static void toCSV(Writer outWriter, List<Map<String, Object>> lines) {
+        if (lines.isEmpty()) {
+            return;
+        }
+        toCSV(outWriter, lines, headers(lines.getFirst()));
+    }
 
+    /**
+     * Writes all records using the provided headers.
+     *
+     * @param outWriter the writer to render the CSV to
+     * @param lines the records to render
+     * @param headers the columns to render, including when there are no records
+     */
+    public static void toCSV(Writer outWriter, List<Map<String, Object>> lines, List<String> headers) {
         try (var csvWriter = CsvWriter.builder().build(outWriter)) {
-            if (lines.isEmpty()) {
-                return;
-            }
-            List<String> headers = headers(lines.getFirst());
             csvWriter.writeRecord(headers);
             for (Map<String, Object> record : lines) {
                 csvWriter.writeRecord(values(record, headers));
--- a/webserver/src/test/java/io/kestra/webserver/controllers/api/DashboardControllerTest.java
+++ b/webserver/src/test/java/io/kestra/webserver/controllers/api/DashboardControllerTest.java
@@ -12,6 +12,7 @@ import org.junit.jupiter.api.Test;
 import org.slf4j.event.Level;
 
 import io.kestra.core.junit.annotations.KestraTest;
+import io.kestra.core.junit.assertions.Problems;
 import io.kestra.core.models.Label;
 import io.kestra.core.models.QueryFilter;
 import io.kestra.core.models.QueryFilter.Op;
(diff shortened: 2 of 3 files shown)


webserver/src/test/java/io/kestra/webserver/controllers/api/DashboardControllerTest.java
```diff
@@ -12,6 +12,7 @@
 import org.slf4j.event.Level;
 
 import io.kestra.core.junit.annotations.KestraTest;
+import io.kestra.core.junit.assertions.Problems;
```
ammeek, line 15: This import statement seems to be redundant.

webserver/src/main/java/io/kestra/webserver/utils/CSVUtils.java
```diff
@@ -28,12 +29,21 @@ private CSVUtils() {
...
 
+    /**
+     * Writes all records using the provided headers.
+     *
+     * @param outWriter the writer to render the CSV to
+     * @param lines the records to render
+     * @param headers the columns to render, including when there are no records
+     */
+    public static void toCSV(Writer outWriter, List<Map<String, Object>> lines, List<String> headers) {
         try (var csvWriter = CsvWriter.builder().build(outWriter)) {
```
ammeek, line 46: These two methods do not guarantee the same order of headers. 

`toCSV(Writer outWriter, List<Map<String, Object>> lines)` uses the order provided in the first line of the csv which is not guaranteed  to be the name as the headers value passed into the `toCSV(Writer outWriter, List<Map<String, Object>> lines, List<String> headers)` method.

This means that empty vs none empty CSV file could have a different header order.

ammeek (changes requested):
Thank you for the PR @ToxicCalvin. A small issue around header ordering behaviour.

webserver/src/test/java/io/kestra/webserver/controllers/api/DashboardControllerTest.java
ToxicCalvin: Removed the unused `Problems` import in c328443816.

webserver/src/main/java/io/kestra/webserver/utils/CSVUtils.java
ToxicCalvin: @ammeek Thanks for pointing this out. I reproduced the mismatch: the chart declares chart_execution_id,chart_namespace, while the JDBC result returns the opposite order.
The controller now passes the declared columns to toCSV for both empty and populated results. This keeps the header order consistent and aligns the values with those headers.
I updated the controller test to export the same chart before and after inserting a row. It fails with the previous logic and passes with the fix. All 1,299 webserver tests also pass. Please let me know if there are more problems. Thanks.

MilosPaunovic: If you've addressed any of the requested changes from the PR review, @ToxicCalvin, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

ToxicCalvin: Good to know, thanks!

AGENTS.md
```diff
@@ -166,6 +166,7 @@ See [Comments](#comments) for when a Javadoc block is worth writing at all. Once
 - Unit tests must assert that a user can only access a given API if authorized to do so, and that access is denied otherwise
 - APIs must be documented with OpenAPI annotations
 - Use DTOs for requests/responses
+- Tabular exports must use the same column definition and order for empty and non-empty results.
```
ammeek: I am not sure that adding this to the AGENT.md file is the best way forward at the moment. Currently multiple parts of our codebase do not follow this rule for example ExecutionController:2770, TriggerController:428, etc. I would remove the edits to the AGENT.md from this PR to keep the scope clear and we can look at making this a follow up issue.

PR #18997, opened 2026-09-03
fix(flows): breadcrumb navigation from execution to flow edit no longer blanks the page

flcarre:
Closes https://github.com/kestra-io/kestra-ee/issues/10722

## Problem
Clicking the breadcrumb from any execution tab to that execution's flow edit page left the
main content area completely blank (sidebar rendered fine, no spinner, no error).

## Fix
Two independent races had to be fixed:
- `stores/flow.ts`: scope the "superseded load" drop-check to the flow being loaded
  (namespace/id) instead of a single global counter, so an unrelated concurrent flow load
  can no longer drop the response for the flow actually being navigated to.
- `FlowRoot.vue`: render a loading state (matching `ExecutionRoot.vue`'s existing pattern)
  instead of nothing while `!ready`, so a future race is a flicker, not a permanent blank
  page.
- `useExecutionRoot.ts`: the execution page's unmount cleanup unconditionally cleared
  `flowStore.flow`/`flowGraph`. Vue unmounts the outgoing execution view *after* the
  incoming flow view has already loaded and rendered, so this cleanup was firing late and
  wiping out the destination page's already-loaded data — turning the original blank page
  into a permanent stuck loading spinner instead. The cleanup now skips clearing when the
  store already holds the flow the app is navigating to.

## Evidence
- Regression tests: `tests/unit/stores/entityLoadRace.spec.ts` and
  `tests/unit/components/executions/useExecutionRoot.spec.ts`, both verified to fail on
  pre-fix code and pass after.
- `check:types`, `test:unit`, `test:lint` all pass. CI green.
- Manual browser verification, independently reproduced twice (once during initial QA,
  once after the follow-up fix): breadcrumb navigation from an execution to its flow now
  renders the editor and stays rendered — confirmed via DOM polling over several seconds,
  not just a single frame.

## Demo

https://github.com/user-attachments/assets/41a339a7-186c-4e1a-9828-e71827e3318e

### DIFF
--- a/ui/src/components/executions/composables/useExecutionRoot.ts
+++ b/ui/src/components/executions/composables/useExecutionRoot.ts
@@ -53,6 +53,11 @@ export function useExecutionRoot() {
         return executionsStore.execution !== undefined
     })
 
+    // By the time either cleanup below runs, router navigation has already updated `route` to the
+    // destination: if the store holds the flow being navigated to (e.g. breadcrumb -> flow edit,
+    // #10722), clearing it here would erase data the destination page already loaded and rendered.
+    const flowMatchesTarget = () => flowStore.flow?.namespace === route.params.namespace && flowStore.flow?.id === route.params.id
+
     const follow = () => {
         previousExecutionId.value = route.params.id as string
         executionsStore.followExecution(route.params as any, t)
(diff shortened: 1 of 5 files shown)


elevatebart (approved):
TYVM LGTM

PR #18998, opened 2026-09-03
fix(plugins): avoid uninitialized KestraContext during migration-time plugin auto-install

fdelbrayelle:
Closes https://github.com/kestra-io/kestra-ee/issues/10727.

## Problem

On a fresh slim-image start (`KESTRA_PLUGINS_AUTO_INSTALL_ENABLED=true`), every startup logs:

```
WARN  i.k.c.p.PluginAutoInstallService Plugin auto-install failed for artifacts [...].
java.lang.IllegalStateException: Kestra context not initialized
	at io.kestra.core.contexts.KestraContext.getContext(KestraContext.java:49)
	at io.kestra.core.plugins.PluginInstallJobRegistry.ensureStarted(PluginInstallJobRegistry.java:175)
	...
	at io.kestra.jdbc.migration.V2_0_11PluginAutoInstallMigration.migrate(...)
```

`PluginInstallJobRegistry.ensureStarted()` reads `KestraContext.getContext()` to size its thread pool. `KestraContext.Initializer` is a plain `@Context` bean with default init order, while `MigrationStartupRunner` is `@Context @Order(HIGHEST_PRECEDENCE)` and runs `V2_0_11PluginAutoInstallMigration`, which triggers plugin auto-install before `KestraContext` is set. The resulting `IllegalStateException` is caught by `PluginAutoInstallService.installArtifacts`'s best-effort catch block and logged as the WARN above — plugin auto-install then silently fails on every startup that hits this migration.

## Fix

`PluginInstallJobRegistry` already has `ExecutorsUtils` injected, which reads the allocated-CPU-cores configuration directly with no bean-init-order dependency. Swapped `KestraContext.getContext().getAllocatedCpuCores()` for `executorsUtils.getAllocatedCpuCores()`.

## Evidence

`./gradlew :core:compileJava` succeeds.

### DIFF
--- a/core/src/main/java/io/kestra/core/plugins/PluginInstallJobRegistry.java
+++ b/core/src/main/java/io/kestra/core/plugins/PluginInstallJobRegistry.java
@@ -18,7 +18,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
-import io.kestra.core.contexts.KestraContext;
 import io.kestra.core.docs.JsonSchemaCache;
 import io.kestra.core.exceptions.KestraRuntimeException;
 import io.kestra.core.utils.ExecutorsUtils;
@@ -172,7 +171,7 @@ public class PluginInstallJobRegistry {
                 // an instance that briefly needed a burst of installs.
                 int poolSize = concurrency > 0
                     ? concurrency
-                    : Math.max(8, 4 * KestraContext.getContext().getAllocatedCpuCores());
+                    : Math.max(8, 4 * executorsUtils.getAllocatedCpuCores());
                 installExecutor = executorsUtils.maxCachedThreadPool(poolSize, "plugin-install");
             }
         }


fdelbrayelle:
## QA: reproduced the real migration-time race, confirmed on rc14 too

The first pass of this QA (below, superseded) used fresh volumes and never actually exercised the race: `V2_0_11PluginAutoInstallMigration` only calls `installMissingTypes` synchronously when the DB **already has flows referencing missing plugin types at migration time** — i.e. an existing volume being upgraded, exactly Anna's setup (reused `kestra_data`/`kestra_db`). A fresh volume finds "nothing to install" during migration and any later installs go through the async per-flow-save path, which never hits the race regardless of the fix.

**Proper repro**, reusing a single H2 volume across image swaps:

1. Ran the full `kestra/kestra:v2.0.0-rc14` image once (fresh volume) — this seeds 9 built-in "Getting Started" tutorial flows, and migration `2.0.11-plugin-auto-install` runs and gets marked applied ("nothing to install", since the full image bundles every plugin).
2. Stopped it, then forced that migration back to "unapplied" directly in the H2 file (`DELETE FROM kestra_migration_history WHERE script_id = '2.0.11-plugin-auto-install'`), simulating an instance that had never run it before with pre-existing flows on disk.
3. Restarted against the **same volume** with the **slim** image, Anna's exact command:

```bash
docker run --rm --user=root --name kestra \
  -v kestra_repro_storage:/app/storage \
  -v kestra_repro_data:/app/data \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -v /tmp:/tmp \
  -e KESTRA_PLUGINS_AUTO_INSTALL_ENABLED=true \
  <image> server local
```

**On `kestra/kestra:v2.0.0-rc14-slim` (unpatched, released)**: reproduces exactly, same shape as the original report — migration detects 13 missing plugin types (serdes, slack, jdbc-duckdb, script-shell, script-python, terraform, ai, jdbc-sqlite, docker, ...) and throws:

```
WARN  i.k.c.p.PluginAutoInstallService Plugin auto-install failed for artifacts [...].
java.lang.IllegalStateException: Kestra context not initialized
	at io.kestra.jdbc.migration.V2_0_11PluginAutoInstallMigration.migrate(V2_0_11PluginAutoInstallMigration.java:91)
```

So the bug is present in rc14, not just rc13 — confirmed still live on latest release.

**On this branch** (same primed volume, `writeExecutableJar` build packaged into an equivalent slim image): migration `[2.0.11-plugin-auto-install]` detects the same 13 missing types, installs them **synchronously and successfully**, logs `Auto-installed plugin artifacts [...]` at `INFO`, no exception. `grep -c "IllegalStateException|Kestra context not initialized|Plugin auto-install failed"` on the full log → **0**.

No regressions in the rest of the startup sequence.

loicmathieu (approved):
LGTM

PR #19002, opened 2026-09-03
fix(flows): reject a past schedule date in the execute dialog

### LINKED ISSUE #17348: [Executions] Scheduled execution allows selecting a date and time in the past

AcevedoR:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/17348.

### ✨ Description

In the **Execute** dialog, the *Schedule date* picker accepted a date and time in the past. The executor only delays an execution whose schedule date is in the future, so the run started immediately and the value the user chose was silently dropped. On top of that, executing with **no** schedule date still sent a `scheduleDate` query parameter equal to the click time, because the conversion ran `moment(undefined)`.

Reproduced in a browser against `kestra/kestra:develop`: with `2026-09-01 10:00` selected, the request carried `scheduleDate=2026-09-01T10:00:00.000+02:00` and the execution was created and finished at click time.

This PR:
- disables past days in the picker (today stays selectable);
- blocks the submit with a validation message (*Schedule date must be in the future*) when the selected time has already passed, re-checking the clock at click time so a date that went stale while the dialog was open is caught too;
- omits the `scheduleDate` parameter entirely when no date is set.

The date logic lives in a small `ui/src/utils/scheduleDate.ts` module covered by unit tests with an injected clock.

Verified in Chrome through the Vite dev server:
- today's date with an earlier time → Execute disabled, message shown, no request sent;
- past days greyed out in the calendar;
- no date → request has no `scheduleDate` parameter;
- future date → request carries it and the execution is delayed.

### 🎨 Frontend Checklist

- [ ] Code builds without errors (`npm run build`) — not run locally, CI covers it
- [ ] Translations are complete if `en.json` changed — one English key added (`scheduleDateInPast`); the other twelve languages are left to the CI translations job
- [ ] Screenshots or video recordings attached showing the `UI` changes — verified with a scripted browser run, see above

### 📝 Additional Notes

The backend is unchanged: a past `scheduleDate` sent directly to the API is still ignored rather than rejected, to avoid breaking API clients on clock skew.

### 🤖 AI Authors

Why did the cat refuse to schedule an execution in the past? Because it already used up that one of its nine lives. 🐱

### DIFF
--- a/ui/src/components/flows/FlowRun.vue
+++ b/ui/src/components/flows/FlowRun.vue
@@ -54,6 +54,7 @@
                         <KsDatePicker
                             v-model="scheduleDate"
                             type="datetime"
+                            :disabledDate="disabledScheduleDate"
                         />
                     </KsFormItem>
                     <KsFormItem
@@ -117,6 +118,7 @@
     import {useRouter, useRoute} from "vue-router"
     import {useI18n} from "vue-i18n"
     import {useToast} from "../../utils/toast"
+    import {buildScheduleDateParam, isPastScheduleDate, isScheduleDayDisabled} from "../../utils/scheduleDate"
     import moment from "moment-timezone"
     import {useCoreStore} from "../../stores/core"
     import {useApiStore} from "../../stores/api"
@@ -261,6 +263,14 @@
         hasInvalidLabelKeys(executionLabels.value),
     )
 
+    const validationClock = ref(Date.now())
+
+    const hasPastScheduleDate = computed(() =>
+        isPastScheduleDate(scheduleDate.value, new Date(validationClock.value)),
+    )
+
+    const disabledScheduleDate = (day: Date) => isScheduleDayDisabled(day)
+
     const validationMessages = computed(() => {
         const messages: string[] = []
         if (haveBadLabels.value) {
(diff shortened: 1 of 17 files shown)


ui/src/components/flows/FlowRun.vue
```diff
@@ -261,6 +263,14 @@
         hasInvalidLabelKeys(executionLabels.value),
     )
 
+    const validationClock = ref(Date.now())
+
+    const hasPastScheduleDate = computed(() =>
+        isPastScheduleDate(scheduleDate.value, new Date(validationClock.value)),
+    )
+
+    const disabledScheduleDate = (day: Date) => isScheduleDayDisabled(day)
```
elevatebart: NIT: Why not skip this step, avoid creating this function and use `isScheduleDayDisabled` directly in the template?

elevatebart (approved):
One small NIT, great PR

ui/src/components/flows/FlowRun.vue
AcevedoR: should be good now :)

PR #19003, opened 2026-09-03
feat(design-system): file type symbols for storage URIs in KsJsonTree

### LINKED ISSUE #18785: Complex task outputs do not render FILE outputs correctly
### Describe the issue

Complex task outputs, such as those produced by the `io.kestra.plugin.scripts.shell.Commands` task, contain a mix of output data types. When one of them is a `FILE` output (e.g. the `outputFiles` of `Commands`), the rendered shows just its URI and not any strong hint the output actually being a file.

This is confusing for end users.

Image

---

Reproduction task example:

```yaml
  - id: run-metadata-extraction
    outputFiles:
      - abc
    commands:
      - >
        echo "test" > abc
```

### Environment

elevatebart:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18785.

### ✨ Description

A complex task output holding a `FILE` (the `outputFiles` of `scripts.shell.Commands`, say) rendered as its raw storage URI, so the only hint that the value was a file at all was that a very long string happened to fit in the row.

A leaf whose value is a storage URI now renders as a file: a symbol picked from the extension, and the output key as its name, with the full URI left to the tooltip. Two pieces make that up:

- `KsFileTag`, a new design-system component. `uri` drives the symbol and the tooltip, `name` is the label (defaulting to the URI's last segment).
- `fileUtils` (`isFileUri`, `fileName`, `fileExtension`, `fileIcon`), which moves the URI-scheme list into the design system. `Utils.isFile` now delegates to it instead of keeping a second copy of the prefixes.

`KsJsonTree` applies it to every leaf that passes `isFileUri`, so both places that mount a tree viewer, the execution variable explorer and the log lines, get it without a call-site change. Object keys name the file; array entries fall back to the URI's own segment, since an index is not a name.

### 🎨 Frontend Checklist

- [ ] Unit tests pass (`npm run test:unit`) — the new tests pass; the suite has four failures that are pre-existing on `develop`, see the note below
- [ ] Translations are complete if `en.json` changed (`npm run translations:check` reports no missing, extra, or stale keys) — not applicable, `en.json` is untouched (the tag's label comes from the caller, its tooltip is the URI)
- [ ] Screenshots or video recordings attached showing the `UI` changes — see the note below

### 📝 Additional Notes

Screenshots of the tree in light and dark mode, plus the full set of extension symbols, were captured from Storybook, but binaries cannot be uploaded through the API from where this PR was raised, so they need pasting into a comment by hand.

`npm run test:unit` has four failures on this branch, in `logsCursor`, `pluginsLookup` and `useBlueprintPlugins`. They reproduce identically with the branch stashed, so they are pre-existing on `develop` and unrelated to this change.

One deliberate redundancy: when the output key already carries an extension the row reads `"report.csv": [csv] report.csv`. Naming the file by its key is what the issue asks for, and generated URI segments are unreadable, so no de-duplication rule was invented for it.

Covered by a Storybook play test on `KsJsonTree` (file rows, and the URI staying out of the rendered text), one on `KsFileTag`, and unit tests for the extension and name parsing. No backend change.

---

https://claude.ai/code/session_01KyMrz5E7uiT1Ud12GG46gP

---
_Generated by [Claude Code](https://claude.ai/code/session_01KyMrz5E7uiT1Ud12GG46gP)_


flcarre:
## UI Design-System Review (`/kestra-ui-reviewing`)

Scope: 100% frontend/DS PR — `KsFileTag.vue` (new), `KsJsonTree.vue`, `utils/file.ts` (new), `ui/src/utils/utils.ts`, plus stories/tests/docs.

### 🔴 BLOCK — `KsFileTag` duplicates `KsTag`'s default/plain treatment instead of composing it

`ui/packages/design-system/src/components/Data/KsFileTag.vue`:

```scss
.ks-file-tag {
    background: var(--ks-bg-tag);
    color: var(--ks-text-primary);
    ...
}
```

`KsTag` (`Data/KsTag/KsTag.vue`) already renders an icon + label with exactly this look — `.kel-tag--default.kel-tag--plain` sets `--kel-tag-bg-color: var(--ks-bg-tag)` and `--kel-tag-text-color: var(--ks-text-primary)`, and `KsTag`'s default `effect` is already `"plain"`. `KsFileTag` re-derives the identical two tokens by hand instead of composing `<KsTooltip :content="uri"><KsTag :icon="icon" :label="label" /></KsTooltip>`.

Suggested fix: build `KsFileTag` on top of `KsTag` (adding a `truncate`/ellipsis prop to `KsTag` if that's the only real gap), rather than a parallel implementation — per `ui/AGENTS.md` rule 8, "Don't fork — extend."

### 🔴 BLOCK — icon rendered outside `KsIcon`

Same file, template:

```vue
<component :is="icon" aria-hidden="true" class="ks-file-tag__icon" />
```

and style:

```scss
.ks-file-tag__icon {
    width: var(--ks-icon-size-sm);
    height: var(--ks-icon-size-sm);
    color: var(--ks-icon-default);
}
```

`ui/AGENTS.md` → Icons: "All icons come from `vue-material-design-icons` via `<KsIcon>`." `KsIcon` already accepts a `size` token (`"sm"` → `var(--ks-icon-size-sm)`) — this hand-wires exactly that instead of using it:

```vue
<KsIcon size="sm" aria-hidden="true">
    <component :is="icon" />
</KsIcon>
```
(and drop `.ks-file-tag__icon`'s width/height/color rules).

### 🟡 NIT

`padding: 0.125rem var(--ks-spacing-2);` — no `--ks-spacing-*` token exists below `-1` (0.25rem), so this can't be tokenized as-is. Moot if the two BLOCK items above are addressed, since `KsTag`/`ElTag` padding would apply instead.

### ✅ Clean

- No element-plus, no hex/`--el-*`, no `:deep()`, no hardcoded user-facing strings (labels are data-driven, not UI copy).
- Storybook stories + unit tests present for both new pieces, `ui/AGENTS.md` component table updated in the same PR.
- `Utils.isFile` correctly delegates to the new `fileUtils.isFileUri` instead of keeping a second copy of the prefix list — nice dedup.
- Every `--ks-*` token referenced (`--ks-bg-tag`, `--ks-text-primary`, `--ks-radius-base`, `--ks-icon-size-sm`, `--ks-icon-default`, `--ks-spacing-1/2`) resolves in `ks-tokens.scss`/theme files — none are phantom.

**Verdict: CHANGES REQUESTED** (2 BLOCK, both pointing the same direction: compose `KsTag` + `KsIcon` rather than a hand-rolled equivalent).

---
🤖 Posted by an automated `/kestra-ui-reviewing` pass.

ui/packages/design-system/src/utils/file.ts
```diff
@@ -0,0 +1,66 @@
...
+    [FilePdfBox, ["pdf"]],
+    [LanguageMarkdownOutline, ["md", "markdown"]],
+    [CodeJson, ["json", "jsonl", "ndjson", "yaml", "yml", "xml", "toml", "ini", "properties", "conf", "env"]],
+    [FileCodeOutline, ["js", "mjs", "cjs", "ts", "jsx", "tsx", "vue", "html", "htm", "css", "scss", "py", "java", "kt", "rb", "go", "rs", "c", "cpp", "h", "cs", "php", "sh", "bash", "zsh", "ps1", "sql", "r", "scala", "swift", "pl"]],
+    [FileDocumentOutline, ["txt", "log", "out", "err"]],
+]
+
+const ICONS_BY_EXTENSION: Record<string, Component> = Object.fromEntries(
+    ICONS_BY_GROUP.flatMap(([icon, extensions]) => extensions.map((extension) => [extension, icon])),
+)
```
flcarre: 🟡 [Performance/Correctness] Plain-object lookup inherits `Object.prototype`
Problem: `ICONS_BY_EXTENSION` is a plain object from `Object.fromEntries`, so bracket access falls through to inherited members. `ICONS_BY_EXTENSION["constructor"]` returns the native `Object` function and `ICONS_BY_EXTENSION["__proto__"]` returns `Object.prototype` — both truthy, so `fileIcon`'s `?? FileOutline` fallback never kicks in for a file named e.g. `report.constructor` or `report.__proto__`, and `<component :is="icon">` in `KsFileTag` receives a non-Vue-component value instead of the generic file icon. Verified with a plain `Object.fromEntries` repro (`ICONS_BY_EXTENSION["constructor"]` → `[Function: Object]`, not `undefined`).
Fix: use a `Map`, which has no prototype-chain lookup surface.
```suggestion
const ICONS_BY_EXTENSION = new Map<string, Component>(
    ICONS_BY_GROUP.flatMap(([icon, extensions]) => extensions.map((extension) => [extension, icon] as const)),
)
```

ui/packages/design-system/src/utils/file.ts
```diff
@@ -0,0 +1,66 @@
...
+/** Lowercased extension of a URI or file name, empty when it carries none. */
+export function fileExtension(uri: string): string {
+    const name = fileName(uri)
+    const dot = name.lastIndexOf(".")
+    return dot > 0 ? name.slice(dot + 1).toLowerCase() : ""
+}
+
+/** Icon for a URI or file name, falling back to a generic file for unknown extensions. */
+export function fileIcon(uri: string): Component {
+    return ICONS_BY_EXTENSION[fileExtension(uri)] ?? FileOutline
```
flcarre: Companion fix to the `ICONS_BY_EXTENSION` comment above — switch the lookup to `.get()`:
```suggestion
export function fileIcon(uri: string): Component {
    return ICONS_BY_EXTENSION.get(fileExtension(uri)) ?? FileOutline
}
```

flcarre (commented):
## Kestra OSS Code Review

Issue #18785 asked for a visual hint when a JSON-tree leaf holds a `FILE` output instead of a raw storage URI. `KsJsonTree` now renders any leaf that passes `isFileUri` through the new `KsFileTag` (extension icon + readable name, full URI in tooltip), covering both mount points (`ExecutionVariableExplorer`, `LogLine`). Object keys name the file, array entries fall back to the URI segment since an index isn't a name — matches the issue's reproduction case (`outputFiles` of `scripts.shell.Commands`).

### Kestra Guidelines (frontend/design-system) — 0 blocking findings
✅ `Ks*`-only usage (no direct `element-plus` import in `KsFileTag.vue`, composes `KsTooltip`); colors/spacing/radii all via `--ks-*` tokens (verified `--ks-bg-tag`, `--ks-text-primary`, `--ks-icon-default`, `--ks-spacing-*`, `--ks-radius-base`, `--ks-icon-size-sm` are all defined in `ks-tokens.scss`/all three `ks-theme-*.scss` files); no `:deep()`, no `.el-*` overrides; no hardcoded user-facing strings (label comes from the caller, tooltip is the raw URI, so `en.json` is correctly untouched); Storybook story + unit test present for the new `KsFileTag` component per the non-negotiable; `Utils.isFile`'s delegation to `fileUtils.isFileUri` is behavior-preserving (identical prefix list, same signature/return type, all 12 existing call sites in `VarValue.vue`, `ExecutionVariableExplorer.vue`, `ExpressionDebugger.vue`, `Cascader.vue` unaffected).

### Security (OWASP Top 10) — 0 issues found
✅ No injection surface: label is rendered via `{{ }}` text interpolation (auto-escaped), tooltip content is a plain string prop on `KsTooltip` (no `rawContent`/`v-html`). Storage-URI prefix allowlist unchanged.

### Performance — 1 finding (🟡 suggestion, inline)
See the two inline comments on `ui/packages/design-system/src/utils/file.ts` — `ICONS_BY_EXTENSION` is a plain object built with `Object.fromEntries`, so a lookup key that collides with an inherited `Object.prototype` member (`constructor`, `__proto__`) returns that inherited value instead of `undefined`, silently bypassing the `?? FileOutline` fallback. Narrow trigger (a task output literally named `*.constructor` or `*.__proto__`), not a security issue, but a real and easily-avoided correctness gap — a `Map` sidesteps it entirely.

### Nit (🟢, non-blocking)
`KsFileTag.vue`'s scoped `<style>` (wrapper flex/gap/padding/radius/background, icon size/color, name ellipsis) is a near line-for-line duplicate of `KsEntityLink.vue`'s (`ui/packages/design-system/src/components/Data/KsEntityLink/KsEntityLink.vue`), down to the same `0.125rem` padding value. Not asking for a rework in this PR — matching the sibling component's established pattern is the right call over inventing something new — but worth a follow-up to extract the shared "icon + name + tooltip, background `--ks-bg-tag`" chrome into one place so the two don't drift.

**Verdict:** APPROVE — no blocking or guideline-violation findings; one non-blocking correctness suggestion inline, one style nit.

flcarre:
### ✏️ Correction to my two BLOCK findings above

Traced further after `kestra-core-code-reviewer`'s pass ([review](https://github.com/kestra-io/kestra/pull/19003#pullrequestreview-5110650580)) flagged the same spot but reached a different conclusion — checked `ui/packages/design-system/src/components/Data/KsEntityLink/KsEntityLink.vue` directly, and it's right: `KsEntityLink` is already-shipped and structurally near-identical to `KsFileTag` — same `KsTooltip` wrapper, same raw `<component :is="icon" aria-hidden="true" class="...__icon">` (bypassing `KsIcon`), same `background: var(--ks-bg-tag); color: var(--ks-text-primary); border-radius: var(--ks-radius-base); padding: 0.125rem var(--ks-spacing-2);`.

So `KsFileTag` isn't inventing a one-off deviation from `KsTag` — it's faithfully following an existing DS convention (`KsEntityLink`'s "icon + tooltip + truncated label, tag-colored pill" shape) that itself doesn't go through `KsTag` or `KsIcon`. That downgrades both items above from **BLOCK** to a **non-blocking observation**: per `ui/AGENTS.md` rule 12 ("copying an existing rule is not proof it's correct... treat as debt, don't add more"), this is arguably propagating existing debt rather than introducing new debt, since it matches the one directly-analogous component in the codebase. Worth a follow-up to extract a shared base (or route both through `KsTag`/`KsIcon`) at some point, but not a reason to block this PR — consistent with `kestra-core-code-reviewer`'s APPROVE.

Sorry for the noise — retracting the BLOCK severity on those two items.

elevatebart:
Both BLOCKs fixed in 3994c5b, and you were right that they pointed the same direction — composing `KsTag` resolves the second one for free, since `KsTag` already renders `icon` through `KsIcon`.

`KsFileTag` is now:

```vue
<KsTooltip :content="uri">
    <KsTag :icon="icon" :label="label" truncate />
</KsTooltip>
```

with no `<style>` block left at all.

On the NIT: the padding is gone with the rest of the local styles, but clipping turned out to be a real gap rather than a moot one. `ElTag`'s content is an `inline-flex` that will not shrink, so with the hand-rolled styles dropped a long name pushed the tag straight out of its container (verified in the `LongName` story before fixing). Rather than restyle at the call site, `truncate` went onto `KsTag`: the `label` is wrapped in a `kel-tag__label` span, and `.kel-tag--truncate` caps the tag at `max-width: 100%` and gives the content `min-width: 0` so the ellipsis can appear. Covered by a new `Truncated` story whose play function asserts `scrollWidth > clientWidth` on the long label and equality on a short one, and `ui/AGENTS.md` documents the prop on the `KsTag` row.

Checked before pushing: `check:types` clean, `lint` clean, the design-system storybook project green (23 tests across `KsTag`, `KsFileTag`, `KsJsonTree`), and the tree re-screenshotted in light and dark. Two failures in the design-system unit suite (`KsIconButton` router-link, `KsEmptyState` i18n key) reproduce with this branch stashed, so they are not from this change.

---

elevatebart:
Prototype-chain finding fixed in 9bf1e98 — good catch, and it reproduces exactly as described (`Object.fromEntries([["csv","X"]])["constructor"]` → `[Function: Object]`, so the `?? FileOutline` fallback is skipped). `ICONS_BY_EXTENSION` is a `Map` and `fileIcon` uses `.get()`; both inline threads resolved. Added a regression test alongside it, which fails on the previous commit with `expected [Function Object] to be { name: 'FileOutlineIcon' }` and passes now.

On the retraction: no noise caused, but it arrived after I had already pushed the rework in 3994c5b, so `KsFileTag` now composes `KsTag` (no `<style>` block left) and `truncate` lives on `KsTag`. I have kept it rather than reverting, because it is the follow-up your correction asks for ("route both through `KsTag`/`KsIcon`"), and undoing a validated change to restore the duplication would be churn. Worth knowing what that widened, though, since it is more than the original diff: `KsTag` now wraps its `label` in a `kel-tag__label` span for every caller in the app. The full design-system unit suite is unchanged by it (only the two `KsIconButton` / `KsEmptyState` failures that reproduce with the branch stashed), and the storybook project is green.

That leaves `KsEntityLink` as the remaining copy of the "icon + tooltip + tag-colored pill" chrome. Happy to move it onto `KsTag` in a separate PR — it is a one-component change now that `truncate` exists, but it is outside this PR's scope so I have not touched it.

---

elevatebart:
`trigger-ee` is red on 9bf1e98 ([job](https://github.com/kestra-io/kestra/actions/runs/33852741077/job/100959306831)), and it is not this PR's:

```
Webhook HTTP status: 200
##[error]EE compile check inconclusive for ref claude/output-file-type-symbols-4i0mh9 (no compile verdict; HTTP 200). Likely a transient runner error — re-run the job.
```

The EE side answered 200 with every verdict field empty, so the job failed itself as inconclusive rather than reporting a compile or test failure. The diff is OSS frontend only (design-system Vue/TS, no Java, no `webserver`), and this same job passed on 494ca71 earlier on this branch.

Re-run attempted and refused with `403 This workflow is already running` — the run's other jobs are still going. I will re-run the failed job once the run completes; no code change is warranted for this one. If it comes back inconclusive a second time, that is a real signal and I will treat it as such rather than re-running again.

---

elevatebart:
Updating my previous comment: `trigger-ee` has now failed a second time on 9bf1e98 ([attempt 2](https://github.com/kestra-io/kestra/actions/runs/33852741077/job/100967206385)), so it is not the transient hiccup its own error text suggests, and I am not re-running it again.

Both attempts are byte-identical in outcome:

| attempt | started | duration | response |
|---|---|---|---|
| 1 | 08:19:03 | 5m 09s | `HTTP 200` `{"testFailed":"","testReport":"","compileFailed":"","compileError":""}` |
| 2 | 08:50:07 | 5m 19s | `HTTP 200` `{"testFailed":"","testReport":"","compileFailed":"","compileError":""}` |

Two ~5-minute runs ending in an empty verdict reads as a timeout on the EE side rather than a compile result, since a real failure would populate `compileFailed`/`compileError`.

What I can rule out from the OSS side:

- **Not an EE-wide outage.** Other PR runs concluded `success` in an overlapping window — [#16172](https://github.com/kestra-io/kestra/actions/runs/33854242897) ran 08:36–08:53, straddling attempt 2.
- **Not introduced by the branch's content in an obvious way.** The same job passed on 494ca71 on this branch yesterday, and the diff since then is `ui/packages/design-system` plus `ui/src/utils/utils.ts` — no Java, no Gradle, no `webserver`, nothing an EE *compile* check reads.

What I cannot determine: why the EE service returns no verdict for this ref. The verdict comes from an EE-side pipeline, and I can find no corresponding run in `kestra-io/kestra-ee` for `claude/output-file-type-symbols-4i0mh9`, so the failure is only visible to someone with EE-side access.

This needs EE eyes rather than another re-run. Worth checking whether the absent paired EE branch matters here (there is no `claude/output-file-type-symbols-4i0mh9` in `kestra-io/kestra-ee`, since this change is OSS-only) — though that was equally true when the job passed on 494ca71, so I am flagging it as a question, not a diagnosis. Everything else on the head is green.

---

elevatebart:
@flcarre re-requesting your review — this is ready, and the two things you'd have been waiting on are both gone.

Your review landed as `COMMENTED` rather than `APPROVE` (the body says APPROVE, but GitHub recorded it as a comment), so the PR still shows as blocked on review. If you're happy with it, the approve click is the only thing left.

Since you looked at it, on 494ca71:

- Your `ICONS_BY_EXTENSION` finding is fixed — it's a `Map` with `.get()`, plus a regression test that fails on the previous commit with `expected [Function Object] to be { name: 'FileOutlineIcon' }`. Both inline threads resolved.
- `KsFileTag` now composes `KsTag` rather than re-deriving its tokens, which also covers the `KsEntityLink` duplication nit for the new component. The one real gap that surfaced was clipping, so `truncate` went onto `KsTag` — worth a look, since it wraps the label in a `kel-tag__label` span for every caller.
- Branch rebased onto `develop` (no merge commit), and CI is fully green on `aaeaa10`, `trigger-ee` included.

On that last one: the `trigger-ee` failures I flagged above needed no fix. It has now passed twice against a current base (6m01s on the rebase) and only ever returned an empty verdict against the stale one, so please ignore that thread — nothing in the diff was implicated.

---

ui/packages/design-system/src/components/Data/KsFileTag.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+    <KsTooltip :content="uri">
```
flcarre: 🟡 [Suggestion] Tooltip could reveal the full name too, not just the URI

**Problem:** The tooltip content is always the raw `uri` — never the `name` — so when `name` is long enough to get ellipsis-truncated (see the `LongName` story), hovering shows the internal storage path (`kestra:///company/team/6Yd2A/outputs/8f2c1d.parquet`) rather than the thing the user actually wants to read in full (the truncated label, e.g. `extremely-long-output-file-name-that-has-to-be-clipped`). Since the design intent (per the PR description) is "the full URI left to the tooltip", this may be exactly as intended — but worth double-checking against the mockup/issue, since a truncated long output key otherwise has no way to be read in full.

**Fix (optional):** combine both when a name is present, e.g.:
```suggestion
    <KsTooltip :content="name ? `${name} — ${uri}` : uri">
```

ui/packages/design-system/tests/storybook/Data/KsFileTag.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+    render: (args) => ({
+        components: {KsFileTag},
+        setup() { return {args} },
+        template: "<div style=\"width:180px;padding:24px\"><ks-file-tag v-bind=\"args\" /></div>",
+    }),
+    args: {
+        uri: "kestra:///company/team/6Yd2A/outputs/8f2c1d.parquet",
+        name: "extremely-long-output-file-name-that-has-to-be-clipped",
+    },
+}
```
flcarre, line 93: 🟢 [Nit] `LongName` story has no play test asserting the truncation, unlike `KsTag`'s `Truncated` story

Every other behavior in this PR is nicely covered by a play test (`NamedByKey`, `NameFromUri`, `KsJsonTree`'s `FileOutputs`, `KsTag`'s `Truncated`), but `LongName` — the one story that specifically exercises truncation on `KsFileTag` — doesn't assert it actually truncates. Mirroring the `KsTag.stories.ts` pattern would close the loop and catch a regression if the CSS ever breaks.

```suggestion
export const LongName: Story = {
    render: (args) => ({
        components: {KsFileTag},
        setup() { return {args} },
        template: "<div style=\"width:180px;padding:24px\"><ks-file-tag v-bind=\"args\" /></div>",
    }),
    args: {
        uri: "kestra:///company/team/6Yd2A/outputs/8f2c1d.parquet",
        name: "extremely-long-output-file-name-that-has-to-be-clipped",
    },
    play: async ({canvasElement}: {canvasElement: HTMLElement}) => {
        const label = await waitFor(() => within(canvasElement).getByText("extremely-long-output-file-name-that-has-to-be-clipped"))
        expect(label.scrollWidth).toBeGreaterThan(label.clientWidth)
    },
}
```

flcarre (approved):
## Kestra OSS Code Review

Thanks for this one — it's a clean, well-scoped fix for #18785. A few things stood out as done really well:

- The `Map`-based `ICONS_BY_EXTENSION` lookup with a unit test guarding against `constructor`/`__proto__` extension names is a nice bit of defensive coding that a plain-object lookup wouldn't have caught.
- `decodeURIComponent` in `fileName()` is wrapped in try/catch, avoiding an uncaught exception on a malformed URI segment (which, per this repo's own conventions, could otherwise blank the whole tree silently).
- Moving the storage-URI-scheme knowledge into `design-system/utils/file.ts` and having `ui/src/utils/utils.ts#isFile` delegate to it is a good DRY move, and it follows the existing precedent of `state.ts` (`STATES`/`LOG_LEVELS`) already living in the design-system despite being Kestra-domain concepts — this isn't a module-boundary violation.
- Storybook coverage is thorough (icon families, named-by-key, name-from-URI, `KsJsonTree` integration, `KsTag` truncation), and the tooltip correctly renders as plain text (no `rawContent`), so there's no injection risk there.
- Verified the click-to-select behavior in `ExecutionVariableExplorer`/`LogLine` still bubbles from the row through `KsFileTag`/`KsTooltip` unaffected, since neither intercepts or stops the click.
- No `en.json` changes needed (correct — no new static strings), CI is fully green (14/14), and the `Utils.isFile` refactor is behavior-preserving (same three prefixes, just relocated).

Addresses the issue directly: a `FILE`-typed leaf (e.g. `outputFiles` from `scripts.shell.Commands`) now renders as a recognizable file with an extension-based icon instead of a raw storage URI.

### Kestra Guidelines (backend) — N/A
No Java changes in this diff.

### Frontend / Design System — 2 findings
- 🟡 Suggestion: `KsFileTag`'s tooltip only ever shows the raw `uri`, never the resolved `name` — worth confirming this is deliberate for the long-name truncation case (see inline comment).
- 🟢 Nit: `LongName` story in `KsFileTag.stories.ts` has no `play` test asserting the truncation actually happens, unlike the sibling `KsTag` `Truncated` story (see inline comment).

Tooltip content renders as text (no `rawContent`), `Map`-based lookup avoids prototype-chain footguns, no new dependencies.

Icon lookup table built once at module scope; per-row `isFileUri` check is O(1) string-prefix matching.

**Verdict:** APPROVE

flcarre:
One more thing worth a quick sanity check before this lands: `ui/packages/design-system/src/utils/file.ts` introduces a second, independent extension → icon mapping (`ICONS_BY_GROUP`, using `vue-material-design-icons` — flat monochrome outlines) alongside the one already used by the namespace/flow file explorer at `ui/src/components/utils/icons/Type.vue` (via `material-file-icons` — colored, brand-styled icons).

Practical effect: the same extension (e.g. `.csv`) will now render as two visually different icon styles depending on the surface — a colored file-type icon in the file explorer vs. a flat monochrome MDI icon in `KsJsonTree`/`KsFileTag`.

This isn't a blocker — there's a reasonable case for a monochrome set in a small tag context (matches `KsIcon`'s recoloring, avoids clashing with brand colors at tiny sizes). Just want to confirm it's a deliberate choice rather than a missed reuse of the existing `material-file-icons` mapping. If it's intentional, might be worth a one-line note in `ui/AGENTS.md` or the story so future contributors don't try to "fix" the inconsistency by merging the two systems.

Everything else here looks solid — nice de-dup on `Utils.isFile()` delegating to the new util, and good story/unit-test coverage on `KsFileTag` and the new `truncate` prop on `KsTag`. 🙌

ui/packages/design-system/src/components/Data/KsFileTag.vue
elevatebart: Good catch, and not intended — fixed in 006e091. You're right that the clipped label had nowhere left to be read: the URI's last segment is a generated hash, so the tooltip was answering a question nobody asked.

Two deviations from your snippet:

- `(${uri})` rather than ` — ${uri}`, to keep the em dash out.
- The prefix is conditional on `name` being supplied, not on it being truncated. Without a `name` the label already *is* the URI's last segment, so prefixing there would read `report.csv (kestra:///…/report.csv)`.

```vue
<KsTooltip :content="tooltip">
```
```ts
// A caller-supplied name is the part that gets clipped, so the tooltip has to carry it too.
const tooltip = computed(() => (props.name ? `${label.value} (${props.uri})` : props.uri))
```

Kept it as a single plain-text string rather than two lines, since a line break would mean `rawContent` on `KsTooltip` — and both halves here are attacker-influenced (the output key and the storage path), so the text-only path you called out in the security section is worth more than the formatting.

Two unit tests cover it, and the one for the name case fails on the parent commit with `expected 'kestra:///…' to be 'a-very-long-output-name (kestra:///…)'`.

On the play test: you'd already pushed it in eed10c9 by the time I got here, so I dropped my duplicate and kept yours. I did confirm it bites — removing `truncate` from `KsFileTag` fails it with `expected 332 to be greater than 332`.

---

elevatebart:
Deliberate, and it turns out not to be reusable either way — documented in 063e459, as a fourth bullet under Icons in `ui/AGENTS.md` so nobody tries to merge the two.

Two hard constraints, not just a style preference:

1. `material-file-icons` is declared in `ui/package.json` only, not in `ui/packages/design-system/package.json`. A `Ks*` component importing it would mean adding a dependency to the design system.
2. `Type.vue` renders it as `` `data:image/svg+xml;base64,${btoa(SVG)}` `` in an ``. The colors are baked inside that payload, so no `--ks-icon-*` token can reach them — it would be fixed-color in both themes, and `KsTag` specifically tints its icon `--ks-icon-muted` on the `default`/`plain` variant this tag uses.

So the monochrome set isn't a smaller-is-better call, it's the only one that can recolor. Your reasoning for why it's right in a tag context (matches `KsIcon` recoloring, avoids brand colors clashing at 1em) is exactly the trade-off — the colored set stays right in the file explorer, where the brand styling is the product.

The note I added:

> **Two file-type icon sets coexist on purpose, so don't merge them.** `fileUtils.fileIcon()` (behind `KsFileTag`) maps an extension to a monochrome `vue-material-design-icons` component, which inherits `--ks-icon-*` and so recolors per theme and per tag variant. `ui/src/components/utils/icons/Type.vue` renders the colored `material-file-icons` SVGs for the namespace file explorer, where the brand colors are the point. That package is a `ui/` dependency the design system does not have, and it bakes its colors into a base64 `` that no token can reach, so it cannot be used from a `Ks*` component.

Unrelated, for anyone reading the checks: the red `Required checks` on eed10c9 is that commit's run being cancelled when 006e091 landed on top — every job reads `cancelled`, none `failure`. The run on the current head is green.

---

PR #19007, opened 2026-09-03
fix: disable MCP docs client health checks

### LINKED ISSUE #18981: AI Copilot docs MCP client floods the logs with a health-check stack trace every 30s
### Describe the issue

Running a local OSS server logs a `WARN` with a full 40-line stack trace **every 30 seconds** as soon as the AI Copilot beans are created, whenever the public docs MCP server is not reachable:

```
09:52:10.463 WARN  mcp-server-health-checker d.l.mcp.client.DefaultMcpClient MCP server health check (client key: f6801ed3-...) failed. Attempting to reconnect...
java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.RuntimeException: Unexpected status code: 502
    at dev.langchain4j.mcp.client.DefaultMcpClient.checkHealth(DefaultMcpClient.java:507)
    at dev.langchain4j.mcp.client.DefaultMcpClient.lambda$startAutoHealthCheck$0(DefaultMcpClient.java:725)
    ...
Caused by: java.lang.RuntimeException: Unexpected status code: 502
    at dev.langchain4j.mcp.client.transport.http.StreamableHttpMcpTransport.lambda$execute$0(StreamableHttpMcpTransport.java:232)
```

The 502 itself was a transient gateway error from `https://api.kestra.io/v1/mcp` (the endpoint answers `ping` with 200 again now), so nothing was actually broken. The problem is the polling and the logging around it.

**Where it comes from**

Two MCP clients are built against the same public URL `https://api.kestra.io/v1/mcp`:

- `webserver/src/main/java/io/kestra/webserver/services/ai/KestraDocsContextTool.java` — the client is built **in the constructor**, on a `@Singleton` guarded by `@Requires(property = "kestra.ai.enabled", value = "true", defaultValue = "true")`, so it is on by default
- `webserver/src/main/java/io/kestra/webserver/services/ai/agent/tool/DocsMcpToolProvider.java` — same URL via `AgentConfiguration.docsMcpUrl` (default `https://api.kestra.io/v1/mcp`), built lazily on the first Ask-mode turn

`DefaultMcpClient` starts its health-check scheduler inside its own constructor (`DefaultMcpClient.java:209-210`), with `autoHealthCheck` defaulting to `TRUE` and `autoHealthCheckInterval` to 30s (`DefaultMcpClient.java:136-137`). So the ping loop against a third-party endpoint starts the moment the bean is created, whether or not anyone ever opens the Copilot, and each failure is logged at `WARN` with the full stack trace before a reconnect is attempted.

**Impact**

Functionally harmless — every tool call is best-effort (`callMcpTool` catches and returns `null`) and the client reconnects on its own. But any instance that cannot reach `api.kestra.io` — offline, air-gapped, or behind a restrictive proxy — floods its logs with this stack trace every 30 seconds, forever, for an optional feature that is not being used.

**Suggested fix**

- Pass `.autoHealthCheck(false)` (or a much longer `autoHealthCheckInterval`) on both builders. Failures already surface at call time and `transport.onFailure` still drives reconnection, so the 30s ping buys nothing here.
- While in there: the two clients point at the same URL, so an instance currently keeps two MCP sessions and two `mcp-server-health-checker` threads. They could share a single bean.

**Workaround**

```
--logging.level.dev.langchain4j.mcp=ERROR
```

### Environment

- Kestra Version: develop (2.1.0-SNAPSHOT)

abhiramaab:
## Description

Fixes #18981.

The AI Copilot docs MCP clients were starting an automatic health-check scheduler every 30 seconds. When the public docs MCP endpoint was unreachable, this resulted in recurring WARN logs with a full stack trace even when the Copilot feature was not being actively used.

## Changes

- Disabled automatic MCP health checks in `KestraDocsContextTool`.
- Disabled automatic MCP health checks in `DocsMcpToolProvider`.
- Preserved the existing MCP transport failure/reconnection behavior.
- Added regression tests covering both MCP clients.
- Added coverage verifying that the automatic health-check scheduler is not created.
- Added coverage verifying graceful behavior when the docs MCP endpoint is unreachable.

## Testing

- `./gradlew :webserver:test --tests "io.kestra.webserver.services.ai.KestraDocsContextToolTest" --tests "io.kestra.webserver.services.ai.agent.tool.DocsMcpToolProviderTest"`
- `./gradlew :webserver:test --tests "io.kestra.webserver.services.ai.*"`

All tests passed.

### DIFF
--- a/webserver/src/test/java/io/kestra/webserver/services/ai/KestraDocsContextToolTest.java
+++ b/webserver/src/test/java/io/kestra/webserver/services/ai/KestraDocsContextToolTest.java
@@ -1,0 +1,70 @@
+package io.kestra.webserver.services.ai;
+
+import java.lang.reflect.Field;
+
+import org.junit.jupiter.api.Test;
+
+import io.kestra.core.utils.VersionProvider;
+
+import com.github.tomakehurst.wiremock.junit5.WireMockRuntimeInfo;
+import com.github.tomakehurst.wiremock.junit5.WireMockTest;
+import dev.langchain4j.mcp.client.DefaultMcpClient;
+import dev.langchain4j.mcp.client.McpClient;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.okJson;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlPathMatching;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@WireMockTest
+class KestraDocsContextToolTest {
+
+    @Test
+    void shouldDisableAutoHealthCheckOnMcpClient(WireMockRuntimeInfo wmRuntimeInfo) throws Exception {
+        // Given
+        stubFor(post(urlPathMatching("/v1/mcp.*"))
+            .willReturn(okJson("""
+                {"jsonrpc":"2.0","id":0,"result":{"protocolVersion":"2024-11-05","capabilities":{"tools":{}},"serverInfo":{"name":"mock-docs","version":"1.0"}}}
+            """)));
+
+        VersionProvider versionProvider = mock(VersionProvider.class);
+        when(versionProvider.getVersion()).thenReturn("2.1.0");
+
+        // When
+        try (KestraDocsContextTool tool = new KestraDocsContextTool(wmRuntimeInfo.getHttpBaseUrl(), versionProvider)) {
+            Field clientField = KestraDocsContextTool.class.getDeclaredField("mcpClient");
+            clientField.setAccessible(true);
+            McpClient client = (McpClient) clientField.get(tool);
+
+            // Then
+            assertThat(client).isNotNull();
+            assertThat(client).isInstanceOf(DefaultMcpClient.class);
+
+            Field autoHealthCheckField = DefaultMcpClient.class.getDeclaredField("autoHealthCheck");
+            autoHealthCheckField.setAccessible(true);
+            assertThat(autoHealthCheckField.get(client)).isEqualTo(Boolean.FALSE);
+
+            Field schedulerField = DefaultMcpClient.class.getDeclaredField("healthCheckScheduler");
+            schedulerField.setAccessible(true);
+            assertThat(schedulerField.get(client)).isNull();
+        }
+    }
+
+    @Test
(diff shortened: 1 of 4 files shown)


webserver/src/test/java/io/kestra/webserver/services/ai/KestraDocsContextToolTest.java
```diff
@@ -0,0 +1,65 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.urlPathMatching;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@WireMockTest
+class KestraDocsContextToolTest {
+
+    @Test
+    void shouldDisableAutoHealthCheckOnMcpClient(WireMockRuntimeInfo wmRuntimeInfo) throws Exception {
```
ammeek, line 26: Can you please add //Given/When/Then comments to match the style of our other tests?

webserver/src/test/java/io/kestra/webserver/services/ai/agent/tool/DocsMcpToolProviderTest.java
```diff
@@ -0,0 +1,82 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlPathMatching;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@WireMockTest
+class DocsMcpToolProviderTest {
+
+    @Test
+    void shouldDisableAutoHealthCheckOnMcpClient(WireMockRuntimeInfo wmRuntimeInfo) throws Exception {
```
ammeek, line 29: Can you please add //Given/When/Then comments to match the style of our other tests?

ammeek (changes requested):
Thanks for the PR @abhiramaab. Only requires a very small change around comments in the test.

webserver/src/test/java/io/kestra/webserver/services/ai/agent/tool/DocsMcpToolProviderTest.java
abhiramaab: Done. Added the Given/When/Then comments.

webserver/src/test/java/io/kestra/webserver/services/ai/KestraDocsContextToolTest.java
abhiramaab: Done. Added the Given/When/Then comments.

ammeek (approved):

PR #19017, opened 2026-09-03
fix(namespaces): prevent folder moves into descendants (#18466)

### LINKED ISSUE #18466: [Bash2.0] Moving a namespace files folder into its own subfolder makes all files unreachable
## Describe the issue

Moving a namespace files folder into its own subfolder succeeds, and every file in the namespace becomes unreachable: the Files tab goes empty and the files can no longer be read, by the API or by a flow. Nothing warns you. Kestra already refuses other bad moves (a `..` path, or a name that is taken), so this case is just missing.

Only reachable through the API: the UI has no rename or move for folders, and the tree blocks dragging a folder into its own subfolder.

## Steps to reproduce

```bash
B=http://localhost:8080/api/v1/main
A='your-user:your-password'
NS=demo.movebug
echo hello > /tmp/f.txt

curl -s -u "$A" -X POST "$B/namespaces/$NS/files?path=/work/one.txt" -F "fileContent=@/tmp/f.txt"

# folder is listed
curl -s -u "$A" -G "$B/namespaces/$NS/files/directory" --data-urlencode "path=/"

# move /work into its own subfolder
curl -s -u "$A" -X PUT -G "$B/namespaces/$NS/files" \
  --data-urlencode "from=/work" --data-urlencode "to=/work/nested"

# namespace now looks empty, and the file cannot be read
curl -s -u "$A" -G "$B/namespaces/$NS/files/directory" --data-urlencode "path=/"
curl -s -u "$A" -G "$B/namespaces/$NS/files" --data-urlencode "path=/work/nested/one.txt"
```

## Actual behaviour

The move returns `200`. After it, the root listing goes from `[{"fileName":"work","type":"Directory"}]` to `[]`, the Files tab is empty, and reading `/work/nested/one.txt` returns `404` "Revision 1 of file ... was not found" even though search still lists that path. A flow using `read('work/nested/one.txt')` fails the same way.

## Expected behaviour

- Reject a move whose destination is inside the folder being moved, and leave the files untouched.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise

jasperjonkhans:
# PR title

`fix(namespaces): reject moves into source descendants`

# PR body

Closes https://github.com/kestra-io/kestra/issues/18466.

### ✨ Description

Moving a namespace files directory beneath itself copied entries into the source tree and then purged that tree, leaving the files unreachable.

This rejects normalized descendant destinations before any metadata or storage mutation. The check uses path-segment semantics, so sibling names such as `/work2` remain valid, and reports the invalid move as a `409 Conflict`.

A regression test verifies that rejection preserves the source directory and readable file content, creates no target entries, and still permits a sibling-prefix move.

### 🛠️ Backend Checklist

### 📝 Additional Notes

Focused validation passed with `./gradlew :core:test --tests "io.kestra.core.storages.InternalNamespaceTest"`. The full `:core:test` task also passed. Existing same-path move behavior is unchanged.


ammeek (approved):
LGTM thank you @jasperjonkhans for raising this issue.

MilosPaunovic:
@ammeek Can this be merged?

PR #19019, opened 2026-09-04
fix(triggers): do not unschedule a flow by deleting a declared trigger

### LINKED ISSUE #18476: [Bash2.0] Deleting a trigger silently unschedules the flow with no way to restore it
## Describe the issue

Deleting a trigger from the Triggers admin page stops the flow's schedule for good, silently. The flow still lists the trigger, but it is gone from the Triggers page, nothing fires again, and there is no way to bring it back from the UI or API. Only restarting Kestra recreates it.

## Steps to reproduce

1. Create a flow with a Schedule trigger:

```yaml
id: trigflow
namespace: qa.trig
triggers:
  - id: sched
    cron: "*/5 * * * *"
    timezone: UTC
tasks:
  - id: a
    message: hello
```

2. Wait for it to appear on the **Triggers** admin page with a next execution date.
3. Select it and click **Delete triggers**.
4. Wait past the next cron slot and check the flow's executions.

## Actual behaviour

- The trigger disappears and is never recreated while Kestra runs (still gone after 12 minutes), while the flow definition still lists `sched`.
- Nothing fires. A second flow with the same `*/5` cron, created afterwards, ran at the next slot while this one produced no executions, so the scheduler is fine.
- No way to recover it:
  - `POST /triggers/{ns}/{flow}/{trigger}/restart` → `404` "Trigger ... not found"
  - `PUT /triggers/set-disabled` with `disabled: false` → `404` "Trigger disappeared after set-disabled"
  - saving the flow again does nothing
- Restarting Kestra recreates it and the schedule resumes.

## Expected behaviour

- Either refuse to delete a trigger whose flow still declares it, or recreate it right away so the schedule keeps running.
- If deletion is meant as a reset, make it recoverable from the Triggers page instead of needing a restart.

## Environment

- Kestra Version: rc-10
- Edition: OSS and Enterprise

rishi-jat:
## Summary

Deleting a trigger from the Triggers page removed its runtime state while the flow YAML still declared it. The scheduler only recreates that state on restart, so the flow stayed unscheduled with no UI/API recovery.

Delete is now refused (409) when the flow still declares the trigger. Orphan leftover state can still be deleted. Use Restart, or remove the trigger from the flow.

Fixes #18476


MilosPaunovic:
Can you resolve the merge conflicts here @rishi-jat?

rishi-jat:
> Can you resolve the merge conflicts here @rishi-jat?

Sure, will do asap.

rishi-jat:
Resolved the merge conflicts against `develop`.

Also folded in Copilot's review notes on the same pass:
- `validateToggleable()` is null-safe when a flow has no `triggers` list (no 500 on orphan delete)
- 409 docs cover both "still declared" and scheduler delete failure
- conflict message now says trigger *state*

Locally: `:webserver:test` for the declared vs orphan delete cases is green.

PR #19021, opened 2026-09-04
build(core): translation gate fails on keys used in code but never defined

MilosPaunovic:
## What

The translation gate compared the twelve locale files with each other, so a key that was missing from *all* of them - never added, or deleted in a cleanup while a component still used it - passed every check and the `UI` rendered its raw id. `credentials.delete_error` sat like that for months (https://github.com/kestra-io/kestra-ee/pull/10744), and this audit found seven more in `OSS` and three in `EE`.

Two commits:

1. **Restore and add the keys the code already uses.** The task run state-change hint in `ChangeStatus.vue` still read `change status hint`, a key renamed to `change state hint` in 2024, and the renamed key was then deleted as unused by the ghost-key cleanup, so the info box under the state picker has been empty since. The component now uses the renamed key and the block is restored verbatim from before its deletion in all thirteen locales, fingerprints included, so the generator keeps the recovered values. The global search empty state reuses the existing `no_results_found`. `toast confirm`, `no`, `flows not imported`, `plugin_card.tasks`, `plugin_card.blueprints` and the design-system `filter.titles.mcp_tools_filters` are added with their generated translations.
2. **The rule.** `check-translations.mjs` now also reads the `OSS`, design-system and topology sources (and `ui-ee/src` for the `EE` scope) and reports every literal key passed to `t()`, `$t()` or `<i18n-t keypath>` that no English file defines, as a `::error file=…,line=…::` annotation so it lands inline on the `PR`. Runtime-built keys (`t(e.message)`, `` t(`errors.${code}`) ``, `t("crud.type." + type)`, `:keypath="expr"`) and keys guarded by `te()` are skipped, so a failure is always a real raw-id render. Unused keys are deliberately not reported, since those same dynamic lookups make "unused" unprovable. `build-comment.mjs` and the README are updated, and the `OSS` translations check job now runs on any `ui/**` change, since the `PR` that adds a `t("new.key")` call touches no locale file.

## Merge order

`EE` companion https://github.com/kestra-io/kestra-ee/pull/10747 **first**: `EE`'s `Translation Tests` runs this checker from `OSS` `develop`, so once this lands, `EE` fails on its three missing keys until that `PR` is in.

## Evidence

Run on `develop` before the fixes, the rule reports exactly the seven `OSS` and six `EE` usages (the three credentials ones are since fixed), and nothing else. With both branches applied, `--scope all` passes.

```
::error file=ui-ee/src/components/login/Login.vue,line=299::[EE] Translation key "email required for passwordless" is used in ui-ee/src/components/login/Login.vue:299 but defined in no en.json, so it renders as its raw id - add it to en.json (or reuse an existing key) and run `npm run translations:generate`
```

No screenshots: the visible change is a raw key id turning into a sentence in each of the affected spots, and the gate output above is the more useful evidence.


MilosPaunovic:
Backported to:

- `releases/v2.0.x` branch via https://github.com/kestra-io/kestra/pull/19022.

PR #19053, opened 2026-09-04
fix(flows): plugin auto-install UX papercuts on the slim image

### LINKED ISSUE #19051: [2.0] Slim Image - Pasting new flow with plugins throws error because it hasn't installed it yet
### Describe the issue

The slim image installs plugins once the flow has been saved. However when pasting a task that is not installed, it throws an error. You can ignore it and save the flow, but it's not clear that saving will be possible.

https://github.com/user-attachments/assets/a2908f88-663c-4d2a-8d0c-009a1418c574

### Environment

- Kestra Version: 2.0 RC14 OSS Slim

### LINKED ISSUE #19052: [2.0] Slim Image - Install pop up has a scroll bar that isn't obvious for a notification alert
### Describe the issue

When installing a plugin, a pop up comes up to show it. But it has a scroll bar:

1. Seems odd to put a scroll bar into an alert/notification style thing. Maybe it's there to match the width of the save alert but the error one is wider
2. Scroll bar wasn't obvious to me, I didn't expect to see one there

We should either:

1. Make it wider
2. Make the alert more compact

https://github.com/user-attachments/assets/8c31ecd1-7576-45a5-bd43-35336e487127

Image

### Environment

- Kestra Version: 2.0 RC14 OSS Slim

fdelbrayelle:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/19051.
Closes https://github.com/kestra-io/kestra/issues/19052.

### ✨ Description

Two small UX papercuts in the plugin auto-install flow on the slim image, one commit each:

1. **#19051** — pasting a task from a plugin that isn't installed yet fired the global HTTP error toast, because `generateFlowGraphFromSource`'s 422 went through the axios error interceptor before the store's own `.catch` could handle it (it already falls back to the invalid-graph state gracefully). The flow was never actually blocked — auto-install fetches the missing plugin on save — so the toast was misleading. Fixed by passing `showMessageOnError: false` on that call, same pattern already used elsewhere in the store (e.g. `createFlow`).
2. **#19052** — the plugin-install progress notification used the design system's default 266px notification width, too narrow for its content, forcing an easy-to-miss vertical scrollbar. Reused the existing `kel-notification__large` variant (already used by the error toast) instead of inventing new sizing.

Also renamed `docs/architecture/PLUGIN_AUTO_INSTALL.md` to `PLUGIN_AUTO_DOWNLOAD.md` to match the feature's current name, per request.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run lint`)
- [ ] Translations are complete if `en.json` changed — n/a, no strings changed
- [ ] Screenshots or video recordings attached showing the `UI` changes — no running instance available in this session; both fixes reuse existing, already-screenshotted UI patterns (silent HTTP call, `kel-notification__large` variant)

### 📝 Additional Notes

Scoped the `showMessageOnError: false` fix to `loadGraphFromSource` only (used by the flow editor's live graph regeneration on every edit) — left `getGraphFromSourceResponse` (blueprint preview) untouched since that call site isn't part of either reported issue.

### 🤖 AI Authors

Why did the plugin refuse to install? It said the JAR was already purr-fect as a schema stub. 🐱


elevatebart (approved):

fdelbrayelle:
### QA — live-tested on a slim local instance (empty plugins dir, `server local`, auto-install on by default)

Both fixes verified with a real repro, not just static analysis: reverted each fix locally (Vite HMR), confirmed the bug reproduces, restored the fix, confirmed it's gone.

**#19051 — pasting a task from an uninstalled plugin**

Before (pre-fix): pasting a task referencing an uninstalled plugin (`io.kestra.plugin.aws.s3.Download`) fires an unrelated global "Validation failed / No plugin registered for the defined type" toast, even though the flow would auto-install and save fine.

before

After (fix applied): same paste, no toast — the flow just shows the calm inline validation note ("The plugin is not installed yet and will be installed automatically when the flow is saved"), no popup.

after

Root cause confirmed via network log: `POST /api/v1/main/flows/graph` 422s twice on paste (the editor's live graph regeneration). The fix passes `showMessageOnError: false` on that call so the store's own graceful fallback handles it instead of the global HTTP error interceptor.

**#19052 — install notification scrollbar**

Triggered a real install (Save on a flow with uninstalled plugin tasks, single and multi-artifact). Confirmed via DOM inspection the notification now carries the `kel-notification__large` class (the same width variant already used by the error toast) instead of the cramped default 266px:

notification

Renders cleanly, no scrollbar, for both a single artifact and multiple.

*(Screenshots hosted on throwaway branch `qa/19053-screenshots` — safe to delete after review.)*

wrussell1999 (approved):

PR #19058, opened 2026-09-04
fix(ui): prevent error panel collapse on copy

### LINKED ISSUE #18910: [Executions] Error message copy collapse the error panel
### Describe the issue

**Steps to reproduce:**
1. Create a flow with an error
2. Uncollapse the execution error
3. Try to copy it (the cursor pointer is also missing on the "Copy" icon)
4. It's copied but the error panel is then collapsed

https://github.com/user-attachments/assets/c51966fb-0b70-4457-a7ae-6142342aab34 

Example here: https://prod.kestra-tech.kestra.cloud/ui/main/executions/kestra.engineering.plugins/index_plugin_release/1of8s6jYlTM3KmwFSmQTrH/overview

Kestra 2.0.0

### Environment

- Kestra Version: develop

vasanth642:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18910

### ✨ Description

Fixes the execution error panel collapsing when clicking inside the alert.

- The `KsAlert` had an `@click` handler that toggled the expanded state, while the component already had a dedicated expand/collapse button using the same logic. This caused clicks on the error content, including the Copy action, to collapse the panel.

- This PR removes the redundant `KsAlert` click handler and the `cursor: pointer` styling from the alert, leaving the dedicated expand button as the control for expanding/collapsing the error panel.

### 🎨 Frontend Checklist

https://github.com/user-attachments/assets/74a48526-2d5d-4a75-a23c-88ef9e6d62ea
- [ ] Unit tests pass (`npm run test:unit`)

### 📝 Additional Notes

The issue was reproduced locally on the Execution Overview. I verified that the Copy action successfully copied the error but also collapsed the panel before the fix.

I also checked the other `KsAlert` usages in `ui/src` and searched for the `expanded = !expanded` pattern. The alert-level expansion logic was specific to `ErrorAlert.vue`; the other matching occurrence was an unrelated dashboard button.

The full unit test suite was run. 244 test files passed, with 1 unrelated test file failing due to a 5-second timeout in `tests/unit/components/logs/logsWrapperCursor.spec.ts`. This test is unrelated to the changes in this PR.

After the change:
- Clicking the error content no longer collapses the panel.
- Clicking Copy copies the error and keeps the panel expanded.
- The dedicated expand button still expands/collapses the panel.
- The error panel uses the normal cursor, while the interactive controls retain pointer behavior.


flcarre (approved):
Clean, minimal fix — removing the stray `@click` on `KsAlert` leaves the dedicated chevron button as the only way to toggle expand/collapse, which is exactly right. Approving.

Two pre-existing things I noticed but that aren't this PR's responsibility: the toggle button has no accessible name (icon-only, no aria-label), and there's a pre-existing `:deep()` override in the same file. Neither was touched or worsened here — could be a good follow-up.

PR #19065, opened 2026-09-04
fix(dashboards): avoid tenant defaults before config loads

### LINKED ISSUE #19055: [UI] Dashboards page returns 404 on local startup
### Description

Navigating to the default dashboard URL (`http://localhost:8080/ui/main/dashboards`) results in an in-app 404 "Page not found" error after a fresh local start.

### Steps to Reproduce

1. Start Kestra locally (version `v2.1`).
2. Open `http://localhost:8080` in the browser.
3. Click **Dashboards** in the sidebar, or navigate directly to `http://localhost:8080/ui/main/dashboards`.
4. Observe the rendered view.

### Expected Behavior

The **Dashboards** page should load successfully, or the application should redirect to a valid default landing page.

### Actual Behavior

An internal 404 screen is displayed:

> **Page not found**
> The requested URL was not found on this server. That's all we know.

Image

### Environment

- Kestra Version: develop

ayush-singh-0601:
Closes https://github.com/kestra-io/kestra/issues/19055.

---

### Related Issue

The dashboard can try to load tenant defaults before the custom-dashboard capability is known. On OSS, that endpoint does not exist, so the initial dashboard navigation can end up on the 404 page.

### Description

Only fetch tenant dashboard defaults when the instance explicitly reports that custom dashboards are enabled. Until then, use the bundled default dashboard.

This also adds a regression test for the startup state where the capability has not loaded yet.

### Frontend Checklist

- [x] Translations are complete if `en.json` changed (`en.json` is unchanged)
- [x] Screenshots attached showing the UI change

### Backend Checklist

- [x] Code compiles for the touched module (`:executor:compileJava`)
- [x] No backend behavior was added; the follow-up only removes a stale reference to an enum value deleted on `develop`

### Screenshots

Before:

Dashboard route showing the 404 page

After:

Bundled default dashboard loading successfully

### Additional Notes

The initial E2E job failed before running Playwright because the current `develop` head removed `OnChildFailure.UNKNOWN` but `ExecutorService` still referenced it. The second commit removes that stale comparison; unknown values already deserialize to `CONTINUE`. The exact failing target now compiles successfully under JDK 25.

The dashboard change was also verified with the focused unit suite, the app and test TypeScript projects, lint, and a production build.

### AI Authors

Cat joke: the dashboard stopped chasing a 404 and finally landed on its default paws.

### DIFF
--- a/ui/src/stores/dashboard.ts
+++ b/ui/src/stores/dashboard.ts
@@ -93,7 +93,7 @@ export const useDashboardStore = defineStore("dashboard", () => {
 
     /** Loads the tenant default dashboards; an instance that cannot store dashboards has none and does not serve the route. */
     async function loadDefaults() {
-        if (useMiscStore().configs?.isCustomDashboardsEnabled === false) {
+        if (useMiscStore().configs?.isCustomDashboardsEnabled !== true) {
             defaultDashboards.value = {}
             return defaultDashboards.value
         }
--- a/ui/tests/unit/stores/dashboardTenantDefaults.spec.ts
+++ b/ui/tests/unit/stores/dashboardTenantDefaults.spec.ts
@@ -69,6 +69,16 @@ describe("dashboard store tenant defaults", () => {
         expect(get).not.toHaveBeenCalled()
     })
 
+    it("does not ask for tenant defaults before the dashboard capability is known", {timeout: TEST_TIMEOUT_MS}, async () => {
+        isCustomDashboardsEnabled = undefined
+        get.mockResolvedValue({data: {defaultHomeDashboard: "tenant-default"}})
+
+        const {useDashboardStore} = await import("../../../src/stores/dashboard")
+
+        await expect(useDashboardStore().getDashboardId(route)).resolves.toBe("default")
+        expect(get).not.toHaveBeenCalled()
+    })
+
     it("asks for tenant defaults when the instance can store dashboards", {timeout: TEST_TIMEOUT_MS}, async () => {
         isCustomDashboardsEnabled = true
         get.mockResolvedValue({data: {defaultHomeDashboard: "tenant-default"}})


ayush-singh-0601:
Hi @MilosPaunovic, could you please review this when you have a chance? It fixes the startup dashboard 404 by avoiding the tenant-default endpoint until custom dashboards are explicitly enabled. Thanks.

ayush-singh-0601:
hey @MilosPaunovic I've fixed the issue and the actions pipeline is green, if there is any suggestion or additional fixes please do tell, i am happy to contribute

Piyush-r-bhaskar (commented):
Hey @ayush-singh-0601 

I can't reproduce this.. Can you share the steps to repro you used to move ahead with review ?

ayush-singh-0601:
Thanks for checking. I could not reproduce it once the current OSS /api/v1/configs response had loaded with isCustomDashboardsEnabled: false either. The deterministic failure case is when that capability is absent or still undefined during startup.\n\nTo reproduce the underlying request on develop before this patch:\n\n1. Run the OSS server and sign in.\n2. Use a browser local override or proxy so GET /api/v1/configs returns its normal payload without isCustomDashboardsEnabled.\n3. Open /ui/main/dashboards with no dashboard saved in local storage.\n4. The dashboard store requests GET /api/v1/main/dashboards/settings/default-dashboards. That route moved to EE in #18971, so OSS returns 404.\n\nWith this branch, the store only uses tenant defaults when the capability is explicitly true. Otherwise it uses the bundled default dashboard and does not call the missing route.\n\nThe focused case is in ui/tests/unit/stores/dashboardTenantDefaults.spec.ts. It leaves the capability undefined, expects the bundled default dashboard, and verifies that no request is made. On a fully initialized current checkout where the flag is already false, the issue will not reproduce because #18971 already handles that state.

Piyush-r-bhaskar:
Sorry I won't be able to review your PR then..Thanks for all the efforts 🎉

closing

PR #19068, opened 2026-09-04
fix(executions): keep preview parameters when redirecting to the owning execution

elevatebart:
closes [10687](https://github.com/kestra-io/kestra-ee/issues/10687)

### ✨ Description

When a FILE preview path belongs to a different execution than the one in the URL, `validateFile()` redirects to the owning execution. The redirect URI was built with `path` only, so `maxRows` and `encoding` were dropped on the way.

Two consequences:

- The preview silently fell back to the default row count and UTF-8, ignoring what the user picked in the preview panel.
- On EE the redirected request returned `HTTP 400 missing required QueryValue maxRows`, because the EE override of the endpoint does not declare `maxRows` as optional (fixed separately in kestra-io/kestra-ee#10771).

This is the path a subflow takes: the child execution's FILE input points at a file produced by the parent execution, so previewing it always goes through the redirect. Standalone executions own their files, never redirect, and worked fine.

This is a backport of the equivalent change already on `develop` (#16878), which never reached the 1.3 line.

Test: two executions of the same flow, preview the first one's file through the second one's id with `maxRows=1`, assert the content is truncated to one row. The response it now gets, which is also the proof the parameter survived:

```
{"content"="first\n", "extension"="txt", "maxLine"=1, "truncated"=true, "type"="TEXT"}
```

Before this change that request lands with no `maxRows` and returns 100 rows with `truncated=false`.

### 🔗 Related Issue

Fixes https://github.com/kestra-io/kestra-ee/issues/10687

Companion PRs: kestra-io/kestra-ee#10771 (EE, same 1.3 line) and kestra-io/kestra-ee#10772 (EE `develop`, now green).

### 🛠️ Backend Checklist

Ticked on CI evidence, not my own run: **4375 tests, 0 failed** (`webserver` 459 passed), plus Build Artifacts, `Check Openapi generated spec is up to date`, and E2E all green. This session's container has only JDK 21 against a required 25 toolchain, with downloads blocked by the network policy, so CI was always going to be the first real compile.

### 📝 Additional Notes

The redirect still interpolates the raw `path` without URL-encoding it, same as before this change. Left alone to keep the diff to the reported bug.

The first run flagged four failures that were not this diff's, all passing on re-run: `RequestRunnerTest.requestBasicAuth`/`requestBasicAuthDeprecated` hit a live third-party host, and `MysqlRunnerTest.restartFailedWithAfterExecution`/`restartFailedWithFinally` were `awaitExecution` races that h2 and postgres ran clean in the same job. Those two MySQL races look like a real timing gap in `TestRunnerUtils.awaitExecution` on the finally/afterExecution path for slower backends, worth its own issue rather than a widened await smuggled in here.

### 🤖 AI Authors

I asked the cat to help review the redirect logic. It walked across the keyboard, dropped two query parameters, and went to sleep. Reproduced the bug perfectly on the first try.

https://claude.ai/code/session_013dnyj1R2Sxoq5Qe8AWddsx

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/ExecutionController.java
@@ -2099,7 +2099,11 @@ public class ExecutionController {
             throw new NoSuchElementException("Unable to find execution id '" + executionId + "'");
         }
 
-        HttpResponse<?> validateResponse = this.validateFile(execution.get(), path, "/api/v1/" + this.getTenant() + "executions/{executionId}/file/preview?path=" + path);
+        // keep the preview parameters, otherwise a redirect to the owning execution would lose them
+        String redirect = "/api/v1/" + this.getTenant() + "executions/{executionId}/file/preview?path=" + path
+            + (maxRows != null ? "&maxRows=" + maxRows : "")
+            + "&encoding=" + encoding;
+        HttpResponse<?> validateResponse = this.validateFile(execution.get(), path, redirect);
         if (validateResponse != null) {
             return validateResponse;
         }
--- a/webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionControllerRunnerTest.java
+++ b/webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionControllerRunnerTest.java
@@ -976,8 +976,8 @@ class ExecutionControllerRunnerTest {
 
         // Then - maxRows survived the redirect
         assertThat(preview).isNotNull();
-        assertThat(preview).containsEntry("content", "first");
         assertThat(preview).containsEntry("truncated", true);
+        assertThat(preview.get("content").toString().lines()).containsExactly("first");
     }
 
     @Test
--- a/webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionControllerRunnerTest.java
+++ b/webserver/src/test/java/io/kestra/webserver/controllers/api/ExecutionControllerRunnerTest.java
@@ -939,6 +939,48 @@ class ExecutionControllerRunnerTest {
     }
 
     @Test
+    @LoadFlows(value = { "flows/valids/inputs.yaml" }, tenantId = "previewredirectedfilefromexecution")
+    void previewRedirectedFileFromExecutionKeepsQueryParameters() throws TimeoutException, QueueException {
+        String tenantId = "previewredirectedfilefromexecution";
+        when(tenantService.resolveTenant()).thenReturn(tenantId);
+
+        // Given - two executions of the same flow, the previewed file belongs to the first one
+        // (this is what happens for a subflow reading a file produced by its parent execution)
+        Map<String, Object> multilineInputs = new HashMap<>(inputs);
+        multilineInputs.put(
+            "file",
+            Objects.requireNonNull(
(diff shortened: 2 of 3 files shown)


elevatebart:
Backend tests came back with 3 failures on `2f9dd20`. One was mine, two are not.

**Mine, fixed in `833bed6`.** `previewRedirectedFileFromExecutionKeepsQueryParameters` asserted `content == "first"` but got `"first\n"`. The response body is the thing worth reading here:

```
{"content"="first\n", "extension"="txt", "maxLine"=1, "truncated"=true, "type"="TEXT"}
```

`maxLine=1` and `truncated=true` are the fix working: `maxRows=1` reached the handler after the redirect to the owning execution. Without this PR the redirect drops it and you get 100 rows with `truncated=false`. `DefaultFileRender.renderContent` appends the newline before it flags truncation, so a one-row preview of a three-line file carries the trailing `\n`. The assertion now compares lines rather than the raw string.

**Not mine:** `core > RequestRunnerTest.requestBasicAuth` and `requestBasicAuthDeprecated`, both `Expected size: 2 but was: 1` with the `request` task FAILED. Those sanity-check flows do a live `io.kestra.plugin.core.http.Request` against `https://testpages.eviltester.com/styled/auth/basic-auth-results.html`; the task failed before the `assert` task ran. This PR touches only `ExecutionController` and its test, nothing in the HTTP plugin, and an unreachable third-party host is not something the diff can affect. The push above re-runs them; if they stay red they are an external-endpoint problem for the 1.3 branch generally, not this change.

---

elevatebart:
`833bed6`: **`webserver` is green — 459 passed, 0 failed.** `previewRedirectedFileFromExecutionKeepsQueryParameters` passes, so the fix is verified: previewing a parent execution's file through a child execution's id now keeps `maxRows` across the redirect.

Four failures remain, none in this diff's reach.

**`RequestRunnerTest.requestBasicAuth` / `requestBasicAuthDeprecated`** — the same two as on `2f9dd20`, failing identically. Both sanity-check flows do a live `Request` against `https://testpages.eviltester.com/...`; the `request` task fails, so the flow never reaches its `assert` task. Unreachable third-party host.

**`MysqlRunnerTest.restartFailedWithAfterExecution` / `restartFailedWithFinally`** — new on this head; `jdbc-mysql` was green on `2f9dd20`. Both are `awaitExecution` races, not assertion failures:

```
Execution 4YrG39... is currently at the status SUCCESS which is not the awaited one
Execution 1gwCIi... is currently at the status RUNNING which is not the awaited one
```

Those cases come from `RestartCaseTest`, which `jdbc-h2` (491 passed) and `jdbc-postgres` (490 passed) both ran clean in this same run. Same test bodies, same flows, three backends, only MySQL raced — and the flows restart a task gated on `taskrun.attemptsCount == 0`, so the await is timing-sensitive by construction. This PR changes one webserver controller method and its test; there is no backend-specific code in it.

I'd rather not paper over these by widening the await inside a preview-fix PR. `restartFailedWithFinally` catching the execution mid-`RUNNING` after a RESTARTED transition looks like a genuine gap in `TestRunnerUtils.awaitExecution` for the finally/afterExecution path on slower backends, but that is its own change with its own reviewers.

The workflow is still running so `rerun-failed-jobs` is refused; I will re-run these once it finishes and report what comes back.

---

PR #19076, opened 2026-09-04
fix(executions): preserve working directory subtask results on failur…

### LINKED ISSUE #18917: fix(executions): a WorkingDirectory can terminate before any subtask result is joined
### Problem

`io.kestra.worker.processors.WorkerTaskProcessor.runWorkingDirectory()` publishes the *parent* `WorkingDirectory` taskrun's failure itself on its three failure paths (`preExecuteTasks`, rendering a subtask's `runIf`, `postExecuteTasks`), bypassing `WorkingDirectory.resolveState()` and its `hasInFlightSubtask` guard (added in #18281).

Because the parent's terminal `WorkerTaskResult` and a sibling subtask's `WorkerTaskResult` are two independent messages on the worker-task-result queue — which is a plain dispatch queue, not keyed by execution id like `WorkerJobEvent` is — the executor can join the parent's failure *before* an already-executed subtask's result. The execution is then published as terminal (FAILED) while the subtask's taskrun has never been joined.

Consumers that snapshot the terminal execution (notifications, subflow/`WaitFor` parents, the first UI render, `runnerUtils.runOne` in tests) then see an incomplete `taskRunList`.

The `hasInFlightSubtask` guard from #18281 doesn't help here: it only defers while an *already-joined* child taskrun is non-terminal. In this failure mode, zero children have been joined yet, so the guard's precondition is never true.

### Evidence

CI run https://github.com/kestra-io/kestra/actions/runs/33613076653 (on PR #18903, unrelated to that PR's diff), test `io.kestra.plugin.core.flow.WorkingDirectoryTest.invalidRunIf()`, execution `24QSQTyeqJubNzgRUnUjTt`:

1. Worker runs subtask `s1` (`Return`) to `SUCCESS`.
2. Worker fails to render subtask `s2`'s `runIf: "{{ outputs.failed }}"` and, in the catch block, emits a `WorkerTaskResult` failing the **parent** `workingDirectory` taskrun directly.
3. The executor joins the parent's `FAILED` result before `s1`'s `SUCCESS` result. All known taskruns (just the parent) are terminal, so the execution is published `FAILED` with a single taskrun (`attempts=null`, history `CREATED -> FAILED` in ~23ms, no `RUNNING`).

```
java.lang.AssertionError:
Expected size: 2 but was: 1 in:
[TaskRun(... taskId=workingDirectory, attempts=null, outputs=null,
  state=State(current=FAILED, histories=[CREATED@09:32:22.523032Z, FAILED@09:32:22.546213Z]))]
```

This is the same failure family as #13134 (`postExecuteTasks` race, still keeping `WorkingDirectoryTest.outputFiles()` disabled via `@FlakyTest`) and #18281 (which fixed the already-joined-child case).

### Candidate fixes

- **(a)** Report these worker-side failures on the *subtask* taskrun instead of the parent, and let `WorkingDirectory.resolveState()` own the parent's terminal state as it does on other paths. The `hasInFlightSubtask` guard would then need to also cover "no subtask result joined yet" (currently it only checks already-joined non-terminal children).
- **(b)** Make the worker-task-result queue keyed by execution id (`KeyedDispatchEvent`, as `WorkerJobEvent` already is) so that all results for one execution are joined in emission order. This would close the whole race class, including #13134.

### Repro

`core/src/test/java/io/kestra/plugin/core/flow/WorkingDirectoryTest.java#invalidRunIf()` against `core/src/test/resources/flows/valids/working-directory-invalid-runif.yaml` reproduces this intermittently (race window is a few ms, so it is rare — first observed occurrence found after scanning 700 recent CI runs).

jasperjonkhans:
## Title

`fix(executions): preserve working directory subtask results on failure`

## Body

Closes https://github.com/kestra-io/kestra/issues/18917.

### 🔗 Related Issue

A `WorkingDirectory` parent failure can overtake results from subtasks that already finished, exposing a terminal execution with an incomplete task-run list.

### ✨ Description

This change attaches preceding subtask results to applicable `WorkingDirectory` parent failures and joins them atomically before the parent result. It preserves subtask state and outputs when an invalid `runIf` or `postExecuteTasks` failure is delivered first, while keeping duplicate delivery idempotent.

Transport fallbacks continue to remove unsupported output payloads. If the original result arrives later with valid outputs, those outputs are restored without rejoining the task run, changing execution state, or notifying listeners twice.

### 🛠️ Backend Checklist

### 📝 Additional Notes

`preExecuteTasks` has no preceding subtask result to aggregate. The successful `postExecuteTasks` timing tracked by #13134 remains unchanged; this fix is limited to the parent-failure paths in #18917.

Validated with focused Core, Executor, and Worker tests, Worker Controller compilation, and the mandatory `H2RunnerTest` suite.

### DIFF
--- a/core/src/main/java/io/kestra/core/repositories/TaskOutputRepositoryInterface.java
+++ b/core/src/main/java/io/kestra/core/repositories/TaskOutputRepositoryInterface.java
@@ -18,6 +18,13 @@ public interface TaskOutputRepositoryInterface {
     Optional<TaskOutput> findById(String tenantId, String taskRunId);
 
     /**
+     * Returns whether a task output exists.
+     */
+    default boolean exists(String tenantId, String taskRunId) {
+        return findById(tenantId, taskRunId).isPresent();
+    }
+
+    /**
      * Save a task output.
      */
     TaskOutput save(TaskOutput taskOutput);
--- a/core/src/main/java/io/kestra/core/runners/WorkerTaskResult.java
+++ b/core/src/main/java/io/kestra/core/runners/WorkerTaskResult.java
@@ -4,7 +4,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
 
 import io.kestra.core.models.HasUID;
 import io.kestra.core.models.executions.TaskRun;
@@ -12,13 +14,11 @@ import io.kestra.core.queues.event.DispatchEvent;
 
 import jakarta.annotation.Nullable;
 import jakarta.validation.constraints.NotNull;
-import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Value;
 import lombok.With;
 
 @Value
-@AllArgsConstructor
 @Builder
 public class WorkerTaskResult implements DispatchEvent, HasUID {
     @NotNull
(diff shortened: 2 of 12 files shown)


loicmathieu (changes requested):
Hi,

Thanks for your contribution.
I don't think this is the good direction, there is no point to convey the previous outputs as task should send their outputs once.

PR #19081, opened 2026-09-07
fix(core): prevent UID collisions in IdUtils.fromParts

### LINKED ISSUE #19015: fix: prevent UID collisions in IdUtils.fromParts
## Summary
`IdUtils.fromParts()` constructs UIDs by joining multiple values with `_` without preserving the boundaries between those values.

When two adjacent parts can contain `_`, different logical identifiers can produce the same UID.

For example:
```text
namespace = team
path      = x_y
```
and:
```text
namespace = team_x
path      = y
```
both produce:
```text
team_x_y
```
when passed to `IdUtils.fromParts()`.

This can cause distinct resources to share the same UID. When the UID is used as a primary or unique key, one resource can overwrite another.

## Affected code
`IdUtils.fromParts()` is used by multiple UID implementations, including:

- `NamespaceFileMetadata`
- `FlowId`
- `FlowWithPath`
- `TriggerId`
- `Dashboard`
- `AssetUser`
- `ExecutionQueued`
- `ExecutionCommand`
- and other callers of `IdUtils.fromParts()`

The exact impact should be reviewed for each caller based on the validation and uniqueness constraints of its individual components.

## Expected behavior
UID generation should preserve the boundaries between the individual parts so that distinct logical identifiers cannot produce the same UID.

For example:
```text
(team, x_y) != (team_x, y)
```
should result in distinct UIDs.

## Context
This was identified while fixing the same class of collision in KV metadata in #18950. The KV metadata UID was changed to use a pipe separator instead of `_`.

The broader `IdUtils.fromParts()` usages are tracked separately here to avoid expanding the scope of the KV metadata fix.

### Environment

- Kestra Version: 2.0.0-rc10

Vinu2111:
Fixes #19015

## Problem
`IdUtils.fromParts()` joined parts with a plain `_` separator with no escaping. Since parts themselves can contain `_`, different logical inputs could produce the same UID e.g. `("team", "x_y")` and `("team_x", "y")` both collapsed to `team_x_y`.

## Fix
`fromParts()` now length-prefixes each part (`{length}:{content}`) before concatenating, instead of joining with a separator. Since the decoder always knows exactly how many characters belong to each part, no character inside a part can be mistaken for a boundary, two different part arrays can never produce the same output.

- Method signature is unchanged, so none of the ~30 existing call sites needed modification.
- `fromPartsAndSeparator()` is untouched `ConcurrencyLimit`, `ExecutionRunning`, and `AbstractJdbcConcurrencyLimitStateStore` continue using `|` exactly as before.
- Null-part-skipping behavior is preserved identically.

## Tests
Added to `IdUtilsTest`:
- `shouldNotCollidWhenPartsContainSeparator()` - asserts `fromParts("team", "x_y") != fromParts("team_x", "y")`
- `shouldSkipNullPartsIdentically()` - confirms null parts at any position are still skipped
- Updated the existing `fromParts()` test to reflect the new output format

All tests pass (`./gradlew :core:test --tests "io.kestra.core.utils.IdUtilsTest"` - 7/7 green).

## Note on scope
This changes `fromParts()`'s *output format*, which means any UID already persisted using the old `_`-joined format will no longer match on re-computation. As discussed on the issue, this PR intentionally keeps scope narrow (fixing the encoding itself) a data migration for already-persisted UIDs, if needed, would be a separate follow-up per domain.

### DIFF
--- a/core/src/main/java/io/kestra/core/utils/IdUtils.java
+++ b/core/src/main/java/io/kestra/core/utils/IdUtils.java
@@ -11,7 +11,6 @@ import com.google.common.hash.Hashing;
 @SuppressWarnings({ "deprecation" })
 abstract public class IdUtils {
     private static final HashFunction HASH_FUNCTION = Hashing.md5();
-    private static final char ID_SEPARATOR = '_';
 
     public static String create() {
         return FriendlyId.createFriendlyId();
@@ -25,8 +24,30 @@ abstract public class IdUtils {
         );
     }
 
+    /**
+     * Produces a collision-safe identifier from the given parts.
+     *
+     * <p>Each non-null part is length-prefixed ({@code length:content}) so that
+     * different logical part arrays always produce different output, regardless of
+     * what characters appear inside individual parts.  For example,
+     * {@code fromParts("team", "x_y")} and {@code fromParts("team_x", "y")}
+     * produce distinct strings — a guarantee that a plain separator-based join
+     * cannot provide when parts may contain the separator character.
+     *
+     * <p>Null parts are silently skipped, so
+     * {@code fromParts(null, "a", "b")} equals {@code fromParts("a", "b")}.
+     *
+     * @param parts the parts to encode
+     * @return a collision-safe identifier string
+     */
     public static String fromParts(String... parts) {
-        return fromPartsAndSeparator(ID_SEPARATOR, parts);
+        StringBuilder sb = new StringBuilder();
+        for (String str : parts) {
+            if (str != null) {
+                sb.append(str.length()).append(':').append(str);
+            }
+        }
+        return sb.toString();
     }
 
     public static String fromPartsAndSeparator(char separator, String... parts) {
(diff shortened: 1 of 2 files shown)


Vinu2111:
Good catch, this is the exact trade-off flagged in the original issue and in this PR's "Note on scope": rows persisted with the old separator-joined UID compute a different key after this change, so a delete-by-recomputed-uid() misses them. For ExecutionQueued specifically this is worse than a typical stale-key mismatch, since a missed delete on pop() leaves the row eligible for re-selection, risking duplicate execution.

Three ways I see to resolve it:

1. Add a migration that recomputes execution_queued keys to the new encoding on upgrade.
2. Add a fallback in AbstractJdbcExecutionQueuedStateStore to check the legacy-format key if the new-format delete finds nothing.
3. Keep ExecutionQueued.uid() on the old format explicitly (opt it out of this fix), and only change fromParts() for callers where this doesn't apply.

@MilosPaunovic, since this touches an active queue table rather than a static reference key, would you like me to add a migration/fallback for execution_queued in this PR, or opt it out (option 3) to keep scope narrow?

loicmathieu:
Freezing this PR as long as we didn't settle internaly of what to do about it.

Vinu2111:
Understood, thanks @loicmathieu, happy to wait on the internal decision. I'll keep this branch in sync with develop in the meantime and I'm around whenever there's a direction to move forward on.

PR #11590, opened 2025-09-30
fix(core): properly show flow doc in multipanel with the plugin doc

Piyush-r-bhaskar:

PR #19082, opened 2026-09-07
fix(plugins): flow description block in the editor Docs panel

elevatebart:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra-ee/issues/10740.

### ✨ Description

The editor's Docs panel opened on the flow's own `description:` instead of on plugin docs. `PluginListWrapper.vue` prepended a `.flow-doc` block above the plugin catalogue whenever no task was bound to the cursor, and its gate was `flowStore.flowParsed !== undefined` (read: "the YAML parses"), not "the flow has a description" — so every flow without one got an italic *"No description. Add a description to your flow YAML…"* placeholder taking up the top of the panel.

Introduced by #17096 (`feat(flows): add the no-code Blocks editor`). Verified by diffing the file across its history: #17048 has no such block, #17096 has it verbatim as it stands on develop, and #18223 later only swapped `t(...)` for `$t(...)`. The similarly-named #11590 is a red herring — it touched `PluginList.vue` and `EnhancedMarkdown.vue`.

Removed the block along with its now-dead `KsMarkdown`/`useFlowStore` imports and `.flow-doc*` styles, the `flow_description`/`flow_no_description` keys across the thirteen locales plus `fingerprints.json` (no other caller), and `tests/unit/components/plugins/PluginListWrapper.spec.ts` — four of its five tests asserted the removed block, and the fifth only asserted that `PluginList` renders.

Second commit cleans up dead code found while reading the same file: `isLoading` was a `ref(false)` that nothing has assigned since #17048, so `isLoading || !pluginsData` only ever evaluated `!pluginsData`. The store's `plugins` is `ref<Plugin[]>()` and stays `undefined` until the fetch lands, so that guard alone still drives the skeleton branch.

### 🎨 Frontend Checklist

All green on the current head (`eac1a49`):

- [x] Type checking passes — `Frontend - Tests`
- [x] Code builds without errors — `Build Artifacts`
- [x] Unit tests pass — `Frontend - Tests`; `E2E - Tests` green too
- [x] Translations complete — `Translations - Key parity, placeholders and drift`, and `check-translations.mjs` reports no missing, extra, or stale keys
- [ ] Screenshots — see notes

### 📝 Additional Notes

No screenshot, and that is a real gap in this PR rather than a formality I skipped: the container this was authored in cannot boot the UI (`ensure-sdk` shells out to `npm install`, and it has npm 10.9.7 / Node 22 against the repo's npm ≥ 11.7 / Node 24, so `@kestra-io/kestra-sdk` never gets generated), and it has no Docker daemon, so the PR image is not runnable either. The change is one deleted block; the before state is the screenshot on the linked issue. Happy for a reviewer to grab the after shot from `ghcr.io/kestra-io/kestra-pr:19082`, or say the word and I will get one another way.

~~The deleted styles referenced `--ks-font-size-sm`, which is not declared in any theme file.~~ Wrong, and thanks to @flcarre for catching it: `--ks-font-size-sm` is declared in `ui/packages/design-system/src/assets/styles/ks-tokens.scss` and used across ~30 design-system files. I inferred its absence from the token families listed in `ui/AGENTS.md` instead of grepping for it, which is exactly the mistake that file's rule 11 warns about. No code impact — the whole style block is deleted either way.

Why did the panel keep showing the flow description? It was just very attached to its owner. 🐱

https://claude.ai/code/session_01TAug4a76FHxVfXWtQ7UwzT


ui/tests/unit/components/plugins/PluginListWrapper.spec.ts
```diff
@@ -1,156 +0,0 @@
-import {afterEach, beforeEach, describe, expect, test, vi} from "vitest"
```
flcarre, line 1: 🟡 [Frontend] Coverage regression on the wrapper itself
Problem: All 5 removed tests targeted the flow-doc block (4 directly, 1 only checking PluginList renders). After the deletion, PluginListWrapper.vue has zero dedicated unit tests — nothing exercises the loading-skeleton branch or confirms PluginList still receives the `plugins` prop.
Fix (non-blocking): consider a small replacement spec covering just the two remaining branches, e.g. `shows loading skeleton while plugins are not yet loaded` and `renders PluginList with the plugins data once loaded`.
Not a blocker given how trivial the remaining component is, but worth a follow-up so this file does not go fully untested.

flcarre (commented):
## Kestra OSS Code Review — PR #19082

Thanks for the very well-documented fix, @elevatebart — the PR description's walk through the file's git history (#17048 -> #17096 -> #18223) made this easy to verify independently, and matches what `git log --follow` shows on `PluginListWrapper.vue`.

The linked issue (kestra-io/kestra-ee#10740) is exactly "why do we show the flow description in the Docs panel" — a regression report. The removed `.flow-doc` block was gated on `flowStore.flowParsed !== undefined` (i.e. "the YAML parses"), not on the flow actually having a `description`, so every flow without one showed the italic placeholder. This PR removes the block entirely, restoring the Docs panel to plugin docs only, which is what the issue asked for.

### Kestra Guidelines (backend) — N/A (no Java changed)

### Frontend / Design System — 1 finding
- See inline comment on the deleted spec file: the whole spec was removed rather than trimmed to the two branches (loading / PluginList render) that survive in the simplified component, so `PluginListWrapper.vue` currently ships with no dedicated unit test at all. Non-blocking — the remaining logic is a two-branch v-if/v-else — but worth a minimal follow-up test.

Verified safe removals (traced actual usages, not inferred from the diff alone):
- `flow_description` / `flow_no_description`: after this PR, zero references anywhere in the repo (component, tests, all 13 locale files, `fingerprints.json`). Confirmed via `git grep` on both the PR branch and `develop` — these keys had no other caller even before this change.
- `KsMarkdown` and `useFlowStore` are still imported/used extensively elsewhere (`useFlowStore` in ~64 other files, `KsMarkdown` in 20+ components) — removing them from this file only drops an unused local import; nothing else depends on this file's copy.
- `PluginListWrapper.vue` has exactly one caller (`ui/src/override/components/flows/panelDefinition.ts`, the editor's Docs panel), so no other surface relied on the removed block.
- Ran the actual translation checker (`node ui/scripts/translations/check-translations.mjs`) on the PR branch: "Translation check passed (scope: all)." — confirms the 13-locale + fingerprints removal is complete and consistent.

Nit (doesn't affect correctness, since the whole style block is deleted regardless): the PR description states the deleted `.flow-doc*` styles referenced `--ks-font-size-sm`, "which is not declared in any theme file." That token is in fact declared in `ui/packages/design-system/src/assets/styles/ks-tokens.scss` and used across ~10 other design-system components — just a minor inaccuracy in the writeup, not a code issue.

### Security (OWASP Top 10) — No issues found
Pure UI text/markup removal; no new data flow, endpoint, or permission surface.

### Performance — No issues found
Removes a computed/render branch; if anything, marginally less work per render.

**Verdict:** APPROVE

Nice, surgical fix backed by real historical/usage verification rather than a guess — appreciated on an external contribution. The one suggestion (spec coverage) is a good follow-up, not a blocker.

ui/tests/unit/components/plugins/PluginListWrapper.spec.ts
elevatebart: Taking the `--ks-font-size-sm` correction first: you're right, it's declared in `ks-tokens.scss:15` and used in ~30 design-system files. I inferred its absence from the token families listed in `ui/AGENTS.md` rather than grepping for it — precisely the mistake rule 11 in that file warns about. PR description fixed.

On the spec, I'd rather leave it deleted, and the reason is `AGENTS.md` rather than laziness. What's left of the component after this PR is:

```vue
<div v-if="isLoading || !pluginsData" class="loading-container">
    <KsSkeleton animated :rows="8" />
</div>
<PluginList v-else :plugins="pluginsData ?? []" :key="useMiscStore().theme" />
```

A test over those two branches would assert "the skeleton renders when the store has no plugins, `PluginList` renders otherwise" — which is `ui/AGENTS.md` → *Testing UI* almost verbatim on what not to write ("Don't test that a prop is passed through … or that a `Ks*` component renders: the template is only restated by those assertions"), and the root `AGENTS.md` → *When a test earns its place* ("Removing a feature should reduce the test count rather than raise it"). It couldn't fail for a reason a reviewer would care about; it would just break on the next refactor of this file. The deleted spec's fifth test was that exact assertion, which is why I dropped it with the other four rather than keeping it.

Something more useful did fall out of looking at it, though: `isLoading` is a `ref(false)` that nothing ever mutates — already true on `develop`, since the `isLoading.value = true/false` assignments went away in #17048 — so the guard is effectively just `!pluginsData`. Worth deleting, but it's a different scope from this fix, so I've left it alone rather than widening the PR. Happy to fold it in here if you'd prefer, or open it separately.

Leaving this thread open for you to close, since I'm pushing back rather than acting on it.

---

elevatebart: Folded the `isLoading` cleanup into this PR after all (`eac1a49`), rather than leaving it as a follow-up.

```diff
-<div v-if="isLoading || !pluginsData" class="loading-container">
+<div v-if="!pluginsData" class="loading-container">
-    import {onMounted, ref, computed} from "vue"
+    import {onMounted, computed} from "vue"
-    const isLoading = ref(false)
```

Checked the guard still means something before removing it: `plugins` in the store is `ref<Plugin[]>()` (`stores/plugins.ts:247`), so it's `undefined` until `listWithSubgroup` lands and `!pluginsData` genuinely drives the skeleton branch. `isLoading` was contributing nothing — no assignment to it has existed since #17048 — so the behaviour is identical either way.

Lint clean, and the plugins folder still passes 50/50 locally.

My position on the replacement spec is unchanged for the reasons above, but the component is now a one-condition `v-if`/`v-else`, which if anything makes a test over it thinner still. Your call on the thread.

---

flcarre (approved):

PR #19093, opened 2026-09-07
fix(ui): keep custom columns footer fixed

### LINKED ISSUE #19078: [UI] Customize table columns panel footer moves when scrolling the column list
### Describe the issue

In the **Customize table columns** panel, the column list is scrollable, but the **bottom section showing “8 of 12 columns visible” does not remain fixed at the bottom of the panel**.

When scrolling through the list, the bottom area moves along with the column items instead of remaining anchored to the bottom of the panel.

**Steps to reproduce**

1. Go to the **Executions** page.
2. Open **Customize table columns**.
3. Scroll up/down inside the columns list.
4. Observe the **“8 of 12 columns visible”** section at the bottom.

**Actual behavior**

* The column list scrolls.
* The bottom **“8 of 12 columns visible”** section moves with the content.
* The footer is therefore not consistently positioned at the bottom of the panel.

**Expected behavior**

* Only the **column list (the 12 columns)** should scroll.
* The **“8 of 12 columns visible”** footer should remain fixed/sticky at the bottom of the Customize table columns panel.

https://github.com/user-attachments/assets/05e5f660-9cb9-43bf-a34b-9bae7b7445fd

### Environment

- Kestra Version: develop

1328837476-hug:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/19078.

### ✨ Description

The custom columns panel already made the column list scrollable and marked its footer as sticky, but the panel itself had no height limit. As a result, the popover became the scroll container and moved the footer together with the column rows.

This adds the same 500px height limit used by the neighboring custom filters panel. The column list now takes the available space and scrolls independently, while the visible-column count stays at the bottom of the panel.

I also added a browser-level Storybook test with 12 columns. It checks that the list actually overflows and that the footer keeps the same position after scrolling to the bottom.

### 🎨 Frontend Checklist

- [x] Design system type checking passes (`npm run types:test --workspace=@kestra-io/design-system`)
- [x] Design system builds without errors (`npm run build:design`)
- [x] Existing `CustomColumns` unit tests pass (5/5)
- [x] New Chromium Storybook test passes (1/1)
- [x] No translation changes were needed

### 📝 Additional Notes

The full design system unit suite was also exercised locally, but its highly parallel default run hit unrelated 5-second timeouts under local resource contention. The component-specific unit and browser tests above pass consistently.

### DIFF
--- a/ui/packages/design-system/src/components/Data/KsDataTable/filter/segments/CustomColumns.vue
+++ b/ui/packages/design-system/src/components/Data/KsDataTable/filter/segments/CustomColumns.vue
@@ -64,6 +64,7 @@
 <style lang="scss" scoped>
 .customize-columns-panel {
     height: fit-content;
+    max-height: 500px;
     display: flex;
     flex-direction: column;
     border-radius: 0.5rem;
--- a/ui/packages/design-system/tests/storybook/Data/CustomColumns.stories.ts
+++ b/ui/packages/design-system/tests/storybook/Data/CustomColumns.stories.ts
@@ -1,0 +1,47 @@
+import type {Meta, StoryObj} from "@storybook/vue3-vite"
+import {expect, waitFor} from "storybook/test"
+import CustomColumns from "../../../src/components/Data/KsDataTable/filter/segments/CustomColumns.vue"
+
+const columns = Array.from({length: 12}, (_, index) => ({
+    label: `Column ${index + 1}`,
+    prop: `column-${index + 1}`,
+    default: index < 8,
+}))
+
+const meta: Meta<typeof CustomColumns> = {
+    title: "Components/Data/CustomColumns",
+    component: CustomColumns,
+    args: {
+        storageKey: "custom-columns-story",
+        columns,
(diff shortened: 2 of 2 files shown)


hemantmm:
@1328837476-hug, can you provide a video of the changes made?

1328837476-hug:
[page@121f9828b0e90d9ac98db3d653d5e4fb.webm](https://github.com/user-attachments/assets/60765452-9c61-4a28-9046-252076ea3d7c)

flcarre (approved):
LGTM — the max-height fix re-activates the existing scroll rule so the footer stays pinned, and the new story actually exercises the regression (asserts the footer's position is unchanged across a scroll).

One NIT for awareness, not blocking: the new `max-height: 500px` is hardcoded rather than a token, but that matches the existing convention in every other dropdown panel in this folder (CustomizeFilters, FilterMultiSelect, etc.) — there's no `--ks-*` token for this yet.

Approving.

PR #19096, opened 2026-09-07
feat(executions): widen empty default time range and reorder columns

### LINKED ISSUE #19069: [UI] Widen the Executions list default time range when it returns nothing
### Why

The Executions list applies a "Started Last 24 hours" filter on load. On a new instance holding 35 executions, all of them older than a day, the table opens empty and says "No Executions Found". The filter is a removable chip, so it can be found, but a new user has no reason to suspect a filter is the reason the page looks empty.

The column order is the second half. State now leads, after the recent column fix, but Flow and Namespace still sit behind Start date, End date and Duration. The flow name and the state are what an execution is usually looked for by.

Image
(note in the screenshot I used a 1-hour default instead of 24 hours)

### What - Proposal

- On load, if the default window returns nothing, widen it one step at a time and stop at the first window that has results: 24 hours, then 7 days, then 30 days.
- Show the window that was actually applied in the chip, so it stays visible and removable. The user should never be looking at a filtered list without seeing which filter is on.
- Keep every query date bounded. Do not fall back to no time filter at all. On a big instance an empty 24 hour window can also mean a quiet weekend, and an unbounded query with a row count over a large executions table.
- Take the starting window from kestra.ui.charts.default-duration instead of a hardcoded 24 hours, so an admin who already set that value gets it here too.
- Change the default column order to Id, State, Flow, Namespace (fields identifying what run) and then Start date, End date, Duration, etc (fields identifiying when it run).

### Acceptance criteria

- A new instance whose executions are all older than 24 hours opens the list with rows visible.
- The chip shows the window that was applied, and removing it still shows everything.
- Every query made while widening is date bounded.
- An instance whose 24 hour window is not empty makes exactly one query, same as today.
- The default column order is Id, State, Flow, Namespace, Start date, End date, Duration.

ayush-singh-0601:
Closes https://github.com/kestra-io/kestra/issues/19069.

### 🔗 Related Issue

The Executions list always opened with a 24-hour window. On a quiet instance whose runs are all older than that, the table came up empty with only a removable chip as a hint, which is easy to miss. Flow and Namespace also sat behind the date columns, even though those are usually how you look an execution up.

### ✨ Description

The first load now takes its window from `kestra.ui.charts.default-duration` instead of a hardcoded 24 hours. If that query returns nothing, we widen one step at a time — 7 days, then 30 days — and stop at the first window that has rows. Every one of those requests stays date-bounded; we never fall back to an unfiltered scan.

The chip is updated to the window that actually landed, so it stays visible and removable. Taking the chip off still shows everything. If the first window already has rows, we make a single query, same as today.

Default columns are now Id, State, Flow, Namespace, then Start date, End date, Duration. Saved column orders are left alone.

### 🎨 Frontend Checklist

- [x] Code builds without errors (`npx vue-tsc -p tsconfig.app.json`)
- [x] Unit tests pass (`npm run test:unit -- tests/unit/components/executions/timeRangeWiden.spec.ts`)
- [x] Translations are complete if `en.json` changed (`en.json` is unchanged)

### Screenshots

24 hours empty, list widened to 7 days, columns Id / State / Flow / Namespace / Start date / End date / Duration:

**Light**

Executions list after widening to Last 7 days, light theme

**Dark**

Executions list after widening to Last 7 days, dark theme

### 📝 Additional Notes

7-day / 30-day steps use `PT168H` and `P30D` so the chip can label them as Last 7 days / Last 30 days. A custom relative range that is not on that ladder (for example Last 5 minutes) is left as-is.

I could not compile the backend here — this machine has a Java 25 JRE without `javac` — so the screenshots are the Vite app against a stub API that returns no rows for 24 hours and one row for 7 days. The chrome, filters, and table are the real components.

### 🤖 AI Authors

Cat joke: I asked the executions list why it was hiding, and it said it was just waiting for someone to widen the cat flap. 🐱


ayush-singh-0601:
Hi @marco-comi — this is ready for a look whenever you have a moment.

The list now starts from the configured chart default duration and only widens when that first window is empty, stopping at 7 days or 30 days. The chip follows whatever window actually returned rows, and taking it off still shows everything. If the first window already has results we still make just the one query.

Default columns are Id, State, Flow, Namespace, then the date fields. Existing saved column orders are left alone.

Happy to tweak the ladder or the first-load gating if you'd rather it behave differently.

ayush-singh-0601:
The E2E failure was on Restart/Replay: after the bulk action the list is filtered to SUCCESS and stays empty until those runs finish, then the spec reloads until the pagination total shows up.

That empty SUCCESS window was being treated as “default 24 hours found nothing, widen to 7/30 days”, which raced the reload loop and left the table without a pagination total.

Widening now only runs when the empty result is from the default time window itself — not when state, labels, flow id, or other user filters are already on. The landing-page case from the issue is unchanged.

ayush-singh-0601:
The two red checks are the same failure: E2E Restart/Replay, and Required checks failing because of that.

Those specs reload every 2s while a SUCCESS filter is still empty. The first-load wait for the default time-range chip was holding the table long enough that pagination never showed up in that window.

Pushed a follow-up so the list searches immediately again, still widens only an empty default window, and only rewrites the URL when a widen actually happens.

ayush-singh-0601:
@elevatebart can you please suggest me some changes i need to make or is it fine?

ayush-singh-0601:
@elevatebart is this issue fixed or do i need to make some more changes

MilosPaunovic:
Hi @ayush-singh-0601,

docs/pr-assets/19069/executions-widened-dark.png
elevatebart, line 1: Remove those two screenshot files, they have nothing to do in the codebase

ayush-singh-0601:
Removed the two screenshot files and replaced the new `any` casts flagged by the frontend check. Typecheck, lint, and the focused tests pass locally. Thanks for the review!

elevatebart (approved):
Looks good

Issue #18758, opened 2026-08-27
[UI] Timestamp tooltip overlaps and blocks the theme toggle button (appears across all pages)


YashKushalvardhan:
## Describe the bug
On every page of the application , hovering over the timestamp text near the bottom-right corner triggers tooltip(s) that visually overlap and block the theme toggle button 
positioned just below it, making it hard to click the toggle directly.

## Steps to reproduce
1. Open Kestra UI (any page — e.g. Dashboard, Flows)
2. Move the mouse over the rotated timestamp text near the bottom-right 
   corner (above the theme toggle button)
3. Observe the tooltip(s) showing commit hash, full date, and relative 
   time ("2 days ago")
4. Try moving the mouse straight down to click the theme toggle button
5. Observe the tooltip overlaps the toggle button's position, blocking 
   direct interaction

## Expected behavior
The tooltip should not obstruct the clickable area of the theme toggle 
button, on any page. Users should be able to move directly to the toggle 
without routing the cursor around the tooltip.

## Actual behavior
The tooltip renders directly over the theme toggle across the entire 
application, requiring the user to approach the button from the side 
to avoid the tooltip blocking it.

## Environment
- Kestra version: 1.3.35 (OSS)
- Deployment: Docker Compose (local)
- OS: Windows
- Browser: Brave and Chrome

## Screen recording

https://github.com/user-attachments/assets/ee1f6e98-02d4-4bfc-8fc8-7a996c3a4aff

### Environment

- Kestra Version: 1.3.35 (OSS)

DavidZ79:
Hi, mind if I give this issue a shot?

MilosPaunovic:
Yup, go ahead @DavidZ79!

MilosPaunovic:
Hey @DavidZ79, are there any updates on this?

DavidZ79:
Hi @MilosPaunovic, sorry for the delay, I haven't gotten the chance to have a proper look yet. Looking to start implementing on Sunday 👍

DavidZ79:
Hi @MilosPaunovic, I can't reproduce this on develop since the tooltip and toggle aren't near each other any more.

I did reproduce it on 1.3.35 though, and I think it would be a very simple fix there. So should we close this as fixed in 2.0, or should I open a small PR on releases/v1.3.x?

On develop:
[...]

PR #19099, opened 2026-09-07
fix(system): stop the version tooltip covering the theme toggle

DavidZ79:
---

### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18758

### ✨ Description

Change tooltip opening direction from bottom (default) to top, unblocking the theme toggle button. See bottom right of screenshot.

### 🎨 Frontend Checklist

image

### DIFF
--- a/ui/src/components/ContextInfoBar.vue
+++ b/ui/src/components/ContextInfoBar.vue
@@ -20,6 +20,7 @@
 
         <el-tooltip
             effect="light"
+            placement="top"
             :persistent="false"
             transition=""
             :hideAfter="0"


MilosPaunovic (approved):
Thanks for digging into this one @DavidZ79, and for checking `develop` before opening the PR - that saved a round.

The fix is right: `.versionNumber` sits directly above `.theme-switcher` in the vertical context bar, so the default bottom placement dropped the popup straight onto the toggle. `placement="top"` moves it into the empty flex gap above instead.

Two follow-ups on our side, nothing for you to do here: `kestra-ee` replaces this component wholesale in `ui-ee/src/override/components/layout/ContextInfoBar.vue` (two occurrences of the same `el-tooltip`), and `releases/v1.0.x` carries the same code in both repos, so I am opening the `EE` companion for 1.3 plus both 1.0 backports.

PR #19100, opened 2026-09-07
fix(ui): don't collapse browser history on the tenant-redirect guard

### LINKED ISSUE #18994: [Redirection]Browser back navigation from an Execution (opened via Logs) redirects to Dashboard instead of Logs
## Describe the bug
When navigating to an Execution detail page by clicking a link 
(namespace, flowId, or executionId) from the Logs page, using the 
browser's back button does not return to the Logs page. Instead, it 
redirects to the Dashboard/Overview page.

## Steps to reproduce
1. Go to the Logs page
2. Click on a namespace, flowId, or executionId link in any log entry
3. This opens the corresponding Execution detail page
4. Click the browser's back button
5. Observe that instead of returning to the Logs page, the app 
   navigates to the Dashboard (Overview) page

## Expected behavior
Using the browser back button after navigating from Logs to an 
Execution detail page should return the user to the Logs page (with 
their previous filters/scroll position ideally preserved), matching 
standard browser navigation behavior.

## Actual behavior
The back button takes the user to the Dashboard/Overview page instead, 
losing their place in the Logs view.

## Environment
- Kestra version: 1.3.35 (OSS)
- Deployment: Docker Compose (local)
- OS: Windows
- Browser: Brave

## Screen recording

https://github.com/user-attachments/assets/b9377e82-3256-4910-8601-aa669c9d623c

### Environment

- Kestra Version: 1.3.35 (OSS)

Bryandero98:
## Summary
Fixes #18994 — the browser's back button, after opening an execution from the Logs page, lands on the Dashboard instead of Logs.

## Root cause
`setupTenantRouter`'s `beforeEach` guard (`ui/src/composables/useTenant.ts`) redirects any navigation whose target is missing an explicit `tenant` param, and hardcodes `replace: true` for that redirect unconditionally:

```ts
if (to.path !== "/" && !to.params.tenant) {
    const currentTenant = from.params?.tenant || "main"
    return {path: `/${currentTenant}${to.path}`, query: to.query, hash: to.hash, replace: true}
}
```

Any in-app navigation built without going through `$routeTo` (this component's own tenant-aware helper) hits this branch — including the execution/namespace/flowId `<router-link>`s in `LogLine.vue` that #18994 is about, since their `:to` objects don't set `tenant` explicitly. `replace: true` means the tenant-corrected navigation overwrites the *current* history entry instead of pushing a new one, so whatever page the user was already on silently drops out of the back stack.

I confirmed this mechanism with an isolated `vue-router` reproduction outside the app (same route shape: a parent route with a `redirect` to a default child tab, plus this exact tenant guard) — going from Dashboard → Logs → Execution, the browser's `history.length` stayed flat across the Dashboard → Logs hop instead of growing, and pressing Back from Execution then skipped straight over the collapsed entry.

## Fix
Only collapse (`replace: true`) when there is no real previous page to preserve — the very first navigation of a session (`from.matched.length === 0`, e.g. a bookmarked/typed URL missing the tenant segment). Any later in-app hop that also lacks an explicit tenant still gets its own, real history entry.

## Verification
- Added `ui/tests/unit/composables/useTenant.spec.ts`: spies on the router's underlying history `push`/`replace` to confirm a later navigation (after the session's first) uses `push`, not `replace`. Confirmed this test fails against the pre-fix code (asserts `push` was called; the original code calls `replace` instead).
- Ran `tests/unit/composables` — the new spec passes; the 4 pre-existing failures there are all `@kestra-io/kestra-sdk/plugins` module-resolution errors from the generated SDK not being built in this environment (needs a live backend), unrelated to this change.


ui/src/composables/useTenant.ts
```diff
@@ -31,7 +31,17 @@ export function setupTenantRouter(router: Router, app: App): void {
...
+            // real previous page to preserve - i.e. the very first navigation of
+            // the session (a bookmarked/typed URL missing the tenant segment).
+            // Forcing `replace: true` unconditionally overwrites whatever page the
+            // user was already on (e.g. the dashboard) with the tenant-corrected
+            // target, silently dropping it from browser history: a later
+            // navigation that also lacks an explicit tenant - any in-app link
+            // built without $routeTo - then hits this same branch again, and the
+            // "back" button skips straight over the missing entry.
+            const isInitialNavigation = from.matched.length === 0
+            return {path: `/${currentTenant}${to.path}`, query: to.query, hash: to.hash, replace: isInitialNavigation}
```
MilosPaunovic: Nice diagnosis, this is the right level to fix it at.

One thing on the `replace` flag: `vue-router` merges the guard-returned location **over** the outer navigation options, so `replace: false` does not just mean "do not collapse", it actively overrides a caller's `router.replace()` and turns it into a `push`:

```js
// vue-router 5.1.0, pushWithRedirect
return pushWithRedirect(assign({ replace }, locationAsObject(failure.to), { state, force }), ...)
```

And the `isInitialNavigation` branch is not reachable in a way you can observe, because `finalizeNavigation` already forces a replace on the session's first navigation whatever the guard returns:

```js
const isFirstNavigation = from === START_LOCATION_NORMALIZED;
if (isPush) if (replace || isFirstNavigation) routerHistory.replace(...)
```

So omitting the key entirely fixes the reported bug and keeps `router.replace()` working as a replace:

```suggestion
            return {path: `/${currentTenant}${to.path}`, query: to.query, hash: to.hash}
```

The comment just above needs a small reword to match, since the collapse then comes from `vue-router` itself rather than from this guard.

Nothing in `OSS`, `EE` or the design system relies on that today (every `router.replace()` call site spreads `route.params`, so `tenant` is present and this branch never fires for them), so it is not a live regression either way, but the key-less version is the one that cannot bite later.

With that in, the first test in the new spec can go, since as its own comment notes it pins nothing. The second one is the real regression guard and it does its job: I confirmed it fails on `develop` and passes here.

MilosPaunovic (commented):
One suggestion inline on `useTenant.ts`, otherwise the diagnosis and the fix look right to me.

Bryandero98:
@MilosPaunovic pushed \`b208d343\` — dropped the \`replace\` key entirely rather than tweaking its value, matching your suggestion exactly. Reworded the comment to explain the guard relies on vue-router's own first-navigation handling instead of trying to control it, and removed the first test since it wasn't pinning anything beyond that. The remaining test still passes.

Thanks for catching this — I hadn't considered how the guard's return value merges over the caller's own navigation options.

MilosPaunovic (approved):
Verified locally on `b208d343`: the `Logs` -> `execution` hop now lands on its own history entry, and a caller's `router.replace()` without an explicit `tenant` still replaces rather than being turned into a `push`. `CI` is green, and `EE` is unaffected since `ui-ee` never calls `setupTenantRouter`.

Thanks for the quick turnaround on the merge-order detail, and for the isolated `vue-router` repro in the description - that made the root cause easy to confirm.

I will raise the `releases/v2.0.x` backport, and follow up on `releases/v1.3.x`, where the guard still uses the `next({...})` form and needs the change by hand.

PR #19103, opened 2026-09-07
fix(ui): guard execution labels reload immediately after save

### LINKED ISSUE #18766: Modifying execution labels doesn't reload Overview
### Describe the issue

The execution overview view should be refreshed when modifying (adding, removing) the labels via local _Manage labels_ dialog.

1. Execute a flow
2. Navigate to Executions
3. Open the execution's detail Overview
4. Add a label via _Manage labels_
5. The added label does not show among existing labels until the Overview gets manually refreshed

### Environment

Bryandero98:
## Summary
Relates to #18766 — "Modifying execution labels doesn't reload Overview". Not using the `closes` keyword: I still haven't heard from the original reporter on whether this was actually their repro path, so that call belongs to them/a maintainer, not to my own inference. (The keyword was still in the body from an earlier revision despite an earlier comment here saying it was dropped - actually removed now.)

## The race
While an execution is being followed, `throttledExecutionUpdate` (`stores/executions.ts`) overwrites `execution.value` from SSE push events roughly every 500ms. A local write with fresher data - `SetLabels.vue`'s save response, a replay, a new execution being triggered - could get silently reverted by an SSE push that was already queued (or already in flight) when the local write landed.

## What changed since the last review
@flcarre's review found the previous revision **narrowed** this race instead of closing it:
1. `throttledExecutionUpdate`'s flow-reload branch (`loadFlowForExecutionByExecutionId(...).then(...)`) is a second write path `.cancel()` can never reach - once that promise chain starts, cancelling the throttle does nothing to stop its callback from landing later and clobbering a local write made in between.
2. The component test (`executionLabelsReactivity.spec.ts`) passed identically whether `SetLabels.vue` called the new guarded action or the old raw assignment - it never actually asserted which one ran.

Fixed both:
- Added a generation counter (`executionWriteGeneration`, `stores/executions.ts`) that every write bumps through one shared `applyExecution()`. The flow-reload branch captures the generation right after its own synchronous write and skips its follow-up write if anything newer landed while the flow was loading - the check `.cancel()` structurally cannot provide.
- Routed **every** direct `executionsStore.execution = ...` assignment in the app - 9 call sites across 8 files (`ChangeExecutionStatus.vue`, `ReplayWithInputs.vue`, `PlaygroundLog.vue`, `SubFlowLink.vue`, `useFlowEditorActions.ts`, `submitTask.ts`, `stores/playground.ts` x2, `useExecutionRoot.ts`) plus the store's own internal writes (`deleteExecution`, `followExecution`'s reset, `waitForStateChange`) - through `applyLocalExecutionUpdate` or a new `clearExecution()` counterpart, so the generation counter actually observes every write instead of only the ones that happened to go through an action. This was the "sweep or make it read-only" ask from the review; I did the sweep. I did **not** wrap the exposed `execution` ref in `readonly()` to enforce this at the type level going forward - didn't want to introduce that unverified without being able to run `tsc`/the suite locally this round (see Verification below).
- `executionLabelsReactivity.spec.ts` now spies on `applyLocalExecutionUpdate` and asserts `SetLabels.vue`'s save calls it - reverting to the raw assignment now fails this test, not just a coincidentally-identical render.
- Moved the race test out of `executions.spec.ts` (which otherwise stays pure-function-only, matching where `entityLoadRace.spec.ts`/`logsSearchRace.spec.ts` already live) into a new `executionLabelsRace.spec.ts` next to `executionsFollow.spec.ts`, and added a second case that specifically drives the flow-reload branch. Kept the hand-rolled controllable stream rather than `executionsFollow.spec.ts`'s `fakeFollowStream` - that helper yields a fixed array with no way to pause between two specific events for a mid-stream assertion, which both race cases need.

## Verification
I couldn't safely run a full local install this round (resource-constrained machine) and skipped `vitest`/`tsc` for this revision - traced every write path by hand instead (grepped for all remaining `executionsStore.execution =`/`store.execution =` assignments after the sweep; none left outside the store's own `applyExecution`). Flagging this explicitly rather than claiming a green run I didn't get. Happy to fix anything CI turns up.


flcarre (changes requested):
This is a test-only PR (no implementation change) opened against issue #18766, which is still open — and by your own comment on that issue, you couldn't reproduce the bug and think the fix might already be an incidental side effect of an unrelated refactor. I'd hold off on "Fixes #18766" auto-closing a live bug report on unproven grounds.

More importantly, the test doesn't cover the scenario most likely to actually reproduce this: the SSE live-follow path (`subscribeToExecution`/`throttledExecutionUpdate` in `executions.ts`) overwrites `store.execution` from server push events roughly every 500ms while an execution is being followed. If the original report came from a running execution (the linked EE version suggests so), a save's response could get raced and clobbered by the next throttled snapshot — and that's exactly the kind of gap this test doesn't touch. It only proves the static, non-following chain works.

Smaller: the test hand-reimplements a one-line stand-in for `Overview.vue` rather than rendering the real component, so a regression introduced directly there wouldn't be caught despite the commit message's claim.

The test itself isn't a rubber stamp — it drives the real `SetLabels.vue` save handler and would catch a revert of the actual fix line, so it does earn its place. I'd just extend it to cover the SSE race (or drop the auto-close keyword) before merging, given the issue is still open.

Bryandero98:
@flcarre pushed \`4e37328705\` — you were right, and it goes further than test coverage.

Traced \`throttledExecutionUpdate\`: \`loadExecution\` already guards against exactly this class of race (calls \`.cancel()\` before its own write, with a comment explaining why), but \`SetLabels.vue\`'s save handler never got the same treatment — it wrote \`executionsStore.execution\` directly. Wrote a store-level test that reproduces the race precisely (SSE push A applies via the throttle's leading edge, push B within the same window gets queued for the trailing edge, a save lands in between) — confirmed it fails without a fix and passes with one, by reverting just the \`.cancel()\` call and re-running both ways.

Fix: a new \`applyLocalExecutionUpdate\` store action mirroring \`loadExecution\`'s existing pattern, used by \`SetLabels.vue\` in place of the direct assignment.

Dropped the auto-close keyword from the PR body per your note — I still can't confirm this was the original reporter's exact repro path, so I'd rather leave closing #18766 to a maintainer's call than claim it myself. Also flagged in the PR body that a few other call sites (\`ChangeExecutionStatus.vue\`, \`ReplayWithInputs.vue\`, etc.) write to \`executionsStore.execution\` the same unguarded way — left those out of scope here but the new action is there if you want them swept too.

Bryandero98:
Rebased onto `develop`. Two conflicts, both resolved by keeping upstream's version and re-applying this PR's addition on top: `toExecutionSearchParams()` picked up a typed `ExecutionSearchOptions` parameter upstream (kept as-is; `applyLocalExecutionUpdate` simply sits above it), and `executions.spec.ts` gained a `FilePreview` test whose import is now merged with this PR's.

Re-verified after the rebase rather than just re-running: the regression test still goes red on the unfixed code (removed the `throttledExecutionUpdate.cancel()` line -> `AssertionError: expected [] to deeply equal [{key: 'env', value: 'prod'}]`) and green with it. Both affected spec files pass (4/4).

@flcarre this still carries the SSE-race fix you asked for in your review — `applyLocalExecutionUpdate` in the store, used by `SetLabels.vue` in place of the direct assignment, mirroring `loadExecution`'s existing `.cancel()` guard.

flcarre (changes requested):
Thanks for digging into this properly — the tracing is right. `loadExecution` (`stores/executions.ts:318`) does call `throttledExecutionUpdate.cancel()` before its own write, and `SetLabels.vue` genuinely never got the same treatment. The new store-level race test is a real test: I walked the throttle semantics and it does go red without the `.cancel()`. That's a solid step up from the first pass.

Still blocking on the following.

**1. The PR body still auto-closes the issue.** Line 2 reads `closes #18766`. Your comment says you dropped the keyword, and the last paragraph of the body says "Not marking this as auto-closing #18766" — but the keyword is still there and GitHub will act on it. #18766 is open, unreproduced, and the repro in it (execute a flow, then add a label on the Overview) describes a *terminal* execution, where the follow stream has already sent `end` and closed and the throttle isn't pushing anything. The race you fixed is real and worth fixing, but it is not demonstrably the reporter's bug. Drop the keyword.

**2. The line this PR exists to protect is untested.** The only behavioural change is `SetLabels.vue:129`. Revert it to `executionsStore.execution = response as any` and *both* new specs stay green: the store test calls `store.applyLocalExecutionUpdate(...)` directly and never touches `SetLabels.vue`, and the component test never constructs a follow stream, so both write paths produce the same render. For a PR whose title is "guard execution labels reload immediately after save", the guard doesn't cover the wiring. Either assert the component calls the action, or — better — drive the component test through a followed execution with a queued throttle push so it fails on a revert.

**3. `executionLabelsReactivity.spec.ts` still has no teeth.** All 148 lines pass on unmodified `develop`; it characterises a chain that already worked. This is the same objection as last time and it hasn't moved — it also still hand-rolls `OverviewLike` rather than mounting the real `Overview.vue`, so a regression introduced directly in Overview still wouldn't be caught. And it's a near-verbatim copy of the `Banner.spec.ts` sitting beside it. Give it the SSE scenario or delete it and fold the useful assertion into `Banner.spec.ts`.

**4. The guard doesn't actually close the race — it narrows it.** Two holes:

- `throttledExecutionUpdate` has a second write path that `.cancel()` cannot reach. When the flow doesn't match (`executions.ts:479-491`), it does `loadFlowForExecutionByExecutionId(...).then(() => { execution.value = parsedExecution })`. That's a plain promise callback, not a throttle timer; a save landing while that load is in flight is overwritten regardless. Your store test doesn't hit it because the first push resolves the mocked flow load and populates `flow.value`, so the branch looks safe when it isn't.
- lodash's `cancel()` sets `lastInvokeTime = 0` and clears `lastCallTime`, so `shouldInvoke` returns true for the very next call. The next SSE push after a save therefore applies *immediately* on the leading edge with no throttle delay — a snapshot the server emitted before the save but delivered after it now clobbers the labels faster than it did before this change. You remove one clobber vector and accelerate another.

If the SSE race is the real defect, the durable fix is on the receiving side: have the SSE handler reject a snapshot that is older than the last applied write, rather than asking every call site to cancel a timer.

**5. One action, ten call sites.** `stores/playground.ts`, `utils/submitTask.ts`, `ReplayWithInputs.vue`, `ChangeExecutionStatus.vue`, `useFlowEditorActions.ts`, `SubFlowLink.vue` and `PlaygroundLog.vue` all still assign `executionsStore.execution` directly, and the store still exports `execution` as a writable ref — so nothing stops the next one. I'd rather see either the sweep in this PR or the ref made read-only from outside than a new action that one of ten callers happens to use.

Smaller:
- Put the race test next to `ui/tests/unit/stores/executionsFollow.spec.ts` and reuse its `fakeFollowStream` instead of hand-rolling `createControlledStream`. Repo convention also puts race tests in their own file (`entityLoadRace.spec.ts`, `logsSearchRace.spec.ts`) — `executions.spec.ts` was a pure-function spec and now drags a `vue-router` mock and an SDK factory mock over the unrelated `normalizeFilePreview` tests.
- The test hard-codes the 500ms window. If that constant grows, `advanceTimersByTimeAsync(500)` never reaches the trailing edge and the test passes vacuously even on unfixed code.
- Title should be `fix(ui):`, not `test(ui):` — this ships a behavioural change and should show up in the changelog as one.
- `applyLocalExecutionUpdate` names a server save response "local". And it's a copy of `loadExecution`'s tail; consider one internal `applyExecution(data)` both use. The six-line comment above it can come down to the two lines `loadExecution` gets.

Design system: clean. No element-plus imports, no hex/rgb, no `--el-*`/`--bs-*`, no `:deep()` or `.el-*` overrides, no duplicated DS component, no new user-facing strings, and none of the DataTable/deep-watch/discard-guard traps are in scope here.

flcarre (changes requested):
Thanks — this is a real step forward and I want to be clear about what it fixes before what it doesn't. The `closes` keyword is gone, the title is `fix(ui):`, the race test is in its own file, the sweep of all nine direct `executionsStore.execution =` assignments is done (that was my point 5, and doing the sweep rather than the `readonly()` is the right call), and `executionLabelsReactivity.spec.ts` now spies on `applyLocalExecutionUpdate` so reverting `SetLabels.vue` actually goes red — that was point 2 and it's properly closed. The generation counter is a sound piece of design and it genuinely closes the flow-reload hole from point 4's first bullet; the second race test drives it. Not shipping a green run you didn't get, and saying so, is the right instinct.

Still blocking on the following.

**1. Two CI failures are yours.** `tests/unit/components/executions/useExecutionRoot.spec.ts` fails both its tests:
- `useExecutionRoot unmount cleanup > clears the flow store when navigating away from any flow`
- `useExecutionRoot unmount cleanup > keeps the flow store when navigating to that flow's edit page (breadcrumb, #10722)`

Both throw `TypeError: executionsStore.clearExecution is not a function` from `useExecutionRoot.ts:114` inside the `onUnmounted` hook, which then cascades into `TypeError: Cannot read properties of null (reading '$')` at spec line 84 because the wrapper never unmounts. The store does export `clearExecution`; the spec doesn't. Its `vi.mock("../../../../src/stores/executions")` stub returns only `execution`, `logs`, `resetLogs`, `closeSSE`, `followExecution`. You changed the production call site and left the mock behind — add `clearExecution: vi.fn()`.

**2. The rest of the red is not yours, but it's telling you to rebase.** The other 47 failing tests and all 6 failed suites are one root cause: `SyntaxError: Unexpected error` (`Serialized Error: { code: 32 }`) out of `createI18n` at `tests/unit/setup.ts:60`, hitting about 22 specs this PR never touches. You touch no translation file and no i18n code, and a UI PR that ran twelve minutes before yours the same afternoon was green. This is your base: 2026-09-15, 84 commits behind. Rebase and it should clear.

**3. Rebasing is not optional here — lodash is gone.** `executions.ts:6` is still `import throttle from "lodash/throttle"`. develop dropped lodash in #19439 on 2026-09-17, the day before this run, and `throttle` now comes from `@kestra-io/design-system`. Git will auto-merge that line away, but the whole mechanism and the timing assumptions in `executionLabelsRace.spec.ts` are reasoned against lodash's throttle. Rebase and re-verify against the DS implementation before we read this again.

**4. The guard still does not close the race.** This is point 4's second bullet and it hasn't moved. I checked the replacement: the DS `throttle.cancel()` in `packages/design-system/src/utils/lang.ts` sets `lastCall = 0`, exactly as lodash did. So after `applyLocalExecutionUpdate` calls `cancel()` (`executions.ts:355`, and `clearExecution` at 363), the next SSE push computes `remaining = wait - (Date.now() - 0)`, takes the leading-edge branch, and invokes immediately. It lands on `applyExecution(parsedExecution)` at `executions.ts:504` with no guard at all — the generation counter is only consulted inside the flow-reload callback. A snapshot the backend emitted before the save but delivered after it still reverts the labels, and it does so sooner than it did before this PR. You've closed one clobber vector and left the other accelerated.

And neither new test would catch it: both push their SSE events *before* the local write. Add the case that matters — save, then push a stale snapshot — and you'll see it go red on this revision.

The durable fix is the one I suggested last time and I'd still rather see it than more call-site bookkeeping: make the receiving side reject a snapshot older than the last applied write. You already have the state for it — compare on the execution's state history or a server timestamp inside `throttledExecutionUpdate` before `applyExecution`, and the generation counter, the `cancel()` calls, and the nine-call-site sweep all become belt-and-braces rather than the mechanism.

Smaller:
- `executionLabelsReactivity.spec.ts:106` still mounts `OverviewLike` instead of the real `Overview.vue`. The spy gives the file a reason to exist now, so I won't push again, but a regression written directly into Overview still slips through.
- `executionLabelsRace.spec.ts:104` hard-codes the 500ms window; if that constant grows the test passes vacuously.
- `SubFlowLink.vue:73`: the new `if (value)` wrapper quietly changes behaviour — a falsy `loadExecution` result used to clear the execution and now leaves the stale one on screen. Probably unreachable, but it wasn't part of the ask.

Design system: clean again. No element-plus, no hex/rgb, no `--el-*`/`--bs-*`, no `:deep()` or `.el-*` overrides, no duplicated DS component, no new user-facing strings, and none of the DataTable-pagination, deep-watch or discard-guard traps are in scope.

PR #18815, opened 2026-08-29
fix(ui): default topology panel to vertical, add orientation setting

harshit-seth:
✨ Description

Reverts the flow editor's Topology panel to default vertical (top-to-bottom) instead of horizontal, and adds a "Default Topology Orientation" setting so users who prefer horizontal can keep it.

Root cause: the editor's Topology panel and the execution Overview's Topology panel shared a single localStorage key (topology-orientation) through useStorage in LowCodeEditor.vue. Overview.vue passes a responsive horizontalDefault prop (true on viewports wider than the sm breakpoint). Because useStorage writes its default into storage whenever the key is absent, the first visit to an execution's Overview on a desktop-width window silently persisted topology-orientation: true. The editor's Topology panel never sets horizontalDefault itself, so it fell back to reading that same poisoned key. This was a re-regression of behaviour already fixed twice before (#1162, #1758).

What changed:

LowCodeEditor.vue — useStorage now uses {writeDefaults: false}, so a responsive layout default (Overview) can no longer poison the persisted preference. Only an explicit user toggle writes to storage now. When nothing is stored, the initial orientation falls back to the new Default Topology Orientation setting (vertical by default).
ui/packages/topology/src/Topology.vue — the package's own isHorizontal prop default was true; changed to false to match the intended default everywhere it's used without an explicit override.
BasicSettings.vue — added a "Default Topology Orientation" dropdown (Vertical / Horizontal) under Settings → Main Configuration, next to the other editor options.
constants.ts — added TOPOLOGY_ORIENTATION / DEFAULT_TOPOLOGY_ORIENTATION storage keys and a topologyOrientations enum.
en.json (+ other locales via npm run translations:generate) — new translation keys for the setting's label, description, and options.

A responsive layout choice (Overview adapting to viewport width) and a remembered user preference are conceptually different things and shouldn't share one written value — disabling writeDefaults keeps that separation without touching either component's public API, and an explicit toggle still persists exactly as before.

🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18742

🎨 Frontend Checklist
 Code builds without errors (npm run build)
 All existing E2E tests pass (npm run test:e2e)
 Screenshots or video recordings attached showing the UI changes.

📝 Additional Notes
*Cleared localStorage, opened a flow's Editor → Topology panel renders vertical.
*Opened an execution's Overview on a wide viewport → Topology renders horizontal (responsive behaviour unchanged), without overwriting the editor's stored preference.
[...]

PR #19104, opened 2026-09-07
fix(ui): drop stray target=_blank on the download anchor

### LINKED ISSUE #17322: [Flows] Exporting flow code navigates away from the flow details page
### Describe the issue

When exporting flow code from inside a flow, the UI takes me out of the flow details page and drops me on a different page.

1. Open a flow details page.
2. Use the flow code export action.
3. After exporting, the current flow details context is lost and there is no clear way back to the original flow page.

https://github.com/user-attachments/assets/e35ec526-69fc-4eeb-b988-958a157fafc0

### Expected Behaviour
Exporting flow code should keep the user in the current flow context, or provide a clear way to return to the flow details page.

### Environment

- Kestra Version: `1.3.27`
- Observed: `Wed, Jul 8, 2026 7:12 PM IST`
- Edition: Open Source

Bryandero98:
## Summary
closes #17322 — exporting a flow's YAML code from the flow details page can take the user away to a different page instead of just downloading a file. @ameyypawar's comment on the issue had already traced this to `exportYaml()` → `Utils.downloadUrl()`.

## Change
`downloadUrl()` (`ui/src/utils/utils.ts`) always set `target="_blank"` on the download anchor. `target` does nothing on a `download`-attributed anchor click in evergreen browsers, and is a documented footgun for `download` + `blob:` specifically in some browsers (the #17322 symptom).

An earlier revision of this PR made the guard conditional on origin instead of removing it outright, reasoning that two call sites (`FileExplorer.vue`, `useBaseNamespaces.ts`'s `exportFileDirectory`) might hit a cross-origin redirect. @flcarre's review traced all 9 call sites: 7 pass `blob:` URLs, and those two are same-origin in every current deployment — so the guard was protecting against a navigation nothing produces today. Removed it: `downloadUrl()` now always omits `target`, matching this PR's title.

If a cross-origin `getFileContent` path is worth defending later, that's a separate issue (the server-side fix would be adding `.attach()` there, which `exportNamespaceFiles` already has and `getFileContent` doesn't) - not something this anchor should special-case for.

## Verification
- `ui/tests/unit/utils/utils.spec.ts`: one test asserting `downloadUrl()` never sets `target`.
- The CI failure on the previous push (`Translations - Key parity...`) was a `develop`-side issue from #18815, already fixed in #19526 - unrelated to this change.


flcarre (changes requested):
Before merging this I'd want the actual root cause nailed down — the diagnosis behind this fix traces back to a GitHub comment on #17322 that GitHub itself has flagged and minimized as spam, from an account with a pattern of posting plausible-sounding "root cause" comments across many unrelated open issues. Neither the original reporter nor this PR reproduced the navigation in any browser. Per the WHATWG spec, a download-attributed anchor's activation behavior should already override `target` for blob/same-origin resources in every evergreen browser, so `target="_blank"` was likely already inert wherever the report came from — meaning this may not change anything for the actual bug.

More concretely, `downloadUrl()` is shared by several other call sites (`useBaseNamespaces.ts`, `FileExplorer.vue`, the flow/execution bulk-export helpers) that don't all pass `blob:` URLs — some pass plain API response URLs that could redirect cross-origin. `target="_blank"` was acting as a fallback there; removing it could shift the "navigates away" symptom onto a different caller instead of fixing it.

Can you confirm the reporter's actual browser/OS and check whether any `downloadUrl()` caller can hit a cross-origin redirect before we merge this? The new test is a reasonable attribute guard, but note jsdom can't actually exercise the navigation bug either way.

Bryandero98:
@flcarre pushed \`9be8262e47\` addressing the cross-origin concern - you were right to flag it.

Checked all 8 \`downloadUrl()\` call sites: 6 build \`blob:\` URLs (flow export, task run output, logs, dashboard, bulk export) where \`target\` really is dead weight. But \`FileExplorer.vue\` passes a plain API URL, and \`useBaseNamespaces.ts\`'s \`exportFileDirectory\` explicitly reads \`request.responseURL\` (post-redirect) - both can land on a different origin, where \`download\` isn't honored by the browser at all and the anchor just navigates there. Blanket-dropping \`target\` would've risked reproducing #17322's own symptom on those two.

Made \`downloadUrl()\` origin-aware instead: omits \`target\` only for \`blob:\`/\`data:\`/same-origin, keeps it for anything else. Test now covers all three shapes.

On root cause: still no reproduction on my end in any real browser, and I haven't heard back from the reporter on browser/OS yet (asked on the issue). I don't want to overstate confidence here - this fix removes a genuine footgun and closes the cross-origin regression risk you caught, but I can't yet confirm it's what the original report actually hit. Happy to leave this in your hands on whether that's enough to merge, or whether it should wait for reporter confirmation.

Bryandero98:
Rebased onto `develop` — the conflict was benign: upstream added `downloadText()` at the same spot in `utils.ts` where this PR adds `isDownloadableOrigin()`, so both are kept. They compose correctly: `downloadText()` hands `downloadUrl()` a `blob:` URL, which is exactly the case this PR stops setting `target="_blank"` on. `utils.spec.ts` is green (27/27).

@flcarre the cross-origin concern you raised is addressed in `9be8262e47` (now `90f05af824` after the rebase) — `downloadUrl()` decides per-call and keeps `target="_blank"` for anything not `blob:`/`data:`/same-origin. The root-cause question is still open on my side: no reproduction in any real browser, and no reply from the reporter on browser/OS. Still happy for you to judge whether that's enough to merge.

Bryandero98:
Fair to distrust that source — I don't take @ameyypawar's comment as evidence either, it's minimized as spam for a reason. I re-verified the mechanism independently by reading current `develop`, not by trusting that comment:

- `exportYaml()` in `useFlowEditorActions.ts` does build a `blob:` URL via `window.URL.createObjectURL(blob)` and hand it to `Utils.downloadUrl()`, which unconditionally sets `target="_blank"` — that part is just reading the code, no external claim involved.

On whether `target` is actually inert for `blob:` + `download` in every evergreen browser: it isn't, in practice. This is a real, currently-open compatibility gap, not settled spec purity:
- Firefox bugzilla [1756980](https://bugzilla.mozilla.org/show_bug.cgi?id=1756980) — an anchor with `download` and a `blob:`/same-origin href still opens in the same tab instead of downloading, for certain content types.
- Firefox bugzilla [1766420](https://bugzilla.mozilla.org/show_bug.cgi?id=1766420) — `download` + `blob:` href not honored, PDF specifically.

So the WHATWG "activation behavior overrides target" rule is the spec's intent, but Firefox doesn't fully implement it yet for every content type — `text/yaml` (what `exportYaml` blobs as) isn't the exact PDF case in those bugs, so I can't claim 100% certainty this is the same root cause the reporter hit. I still couldn't reproduce the actual navigation in headless Chromium, and I don't have the reporter's browser/OS.

On the cross-origin concern: confirmed too, independently — `useBaseNamespaces.ts`'s `exportFileDirectory()` calls `Utils.downloadUrl(request.request?.responseURL ?? "", name)` with a raw XHR `responseURL`, which for namespace file exports can be a redirect to cloud storage (cross-origin). That's exactly why the fix is conditional (`isDownloadableOrigin()`) rather than an unconditional removal — cross-origin callers keep `target="_blank"`, same-origin/blob/data callers (like `exportYaml`) don't. So the "shifts the bug elsewhere" risk should already be closed by that split, not just asserted.

Net: I'm confident the code-level claim (what calls what) is correct on my own reading, and that `target="_blank"` + `download` + `blob:` genuinely can misbehave in Firefox today — but I can't independently confirm this is *the* browser/repro the original reporter hit without their environment. Happy to leave this open pending that confirmation if you'd rather not merge on the current evidence, or to narrow the fix if you want more certainty before taking it.

flcarre (changes requested):
Thanks for the rework — the call-site audit was the right response to my second point, and 6-of-8 callers passing `createObjectURL` blobs is a useful thing to have written down. But the conclusion it reached is wrong, and the change has drifted away from what the PR says it does.

**The cross-origin branch rests on a premise that isn't true.**

Both `utils.ts` and the new test comment justify keeping `target="_blank"` with "a namespace files export that redirects to cloud storage (`useBaseNamespaces.ts`'s `exportFileDirectory`)". That endpoint does not redirect. `NamespaceFileController.exportNamespaceFiles` builds the ZIP in memory and returns it as a `StreamedFile` with `.attach(namespace + "_files.zip")` — i.e. `Content-Disposition: attachment`, same origin, no 3xx anywhere. `request.request?.responseURL` is simply the URL that was requested; that read is defensive coding, not evidence of a redirect.

Which means: for every current caller, the server already sends `Content-Disposition: attachment` or hands us a `blob:`, so the browser downloads and stays put regardless of `target`. The branch you've added doesn't protect against a navigation that can happen — and in the one shape where it fires, it trades "downloads, page stays" for "downloads, plus a stray blank tab in some browsers". That's a regression, not a guard.

There *is* a real cross-origin vector, but it's not the one described: `apiUrl()` derives from `VITE_APP_API_URL` (`ui/src/override/utils/route.ts`), which a deployment can point at a different host. That would affect `FileExplorer.vue`, whose endpoint (`GET {namespace}/files`) notably does *not* set `Content-Disposition`. If you want to keep the conditional, that's the case to name and defend. Please don't ship a comment asserting a redirect that the controller does not perform — it will be read as fact by the next person here.

**The root cause is still unconfirmed, and the change now says the opposite of its title.**

You're candid in the description that the navigation has never been reproduced in any browser. That was my first concern and it hasn't moved. Meanwhile the diff no longer drops `target="_blank"` — it keeps it conditionally — while the PR title and the lead commit still say "drop stray target=_blank on the download anchor". That headline lands in the changelog. Either the title follows the code, or the code goes back to being the one-line unconditional removal it was.

My suggestion: make this the two-line change it wants to be. Remove `target` unconditionally, drop `isDownloadableOrigin` entirely, and if the `VITE_APP_API_URL` cross-origin case is genuinely a worry, open it as its own issue with the `FileExplorer` missing-`Content-Disposition` angle attached. A speculative branch guarding a path that no current deployment exercises is worse than no branch.

**Tests.**

49 added lines that assert the attribute the function immediately above them sets. In jsdom, `link.click()` cannot navigate, so these can never observe the bug in the issue title — they lock in the branch conditions, which is a restatement of the implementation rather than a regression guard. If the conditional goes away, one short assertion that `target` is absent is the honest version of this.

One concrete problem regardless: `createElementSpy.mockRestore()` sits as the last statement of each `it` rather than in an `afterEach`. This suite runs under `isolate: false` (see `vitest.config.unit.js` and the `leakGuard.ts` setup file), so a single failing assertion leaves `document.createElement` spied for every subsequent spec in the worker — and `leakGuard` doesn't watch `document`, so you'd get cascading failures with no signal pointing back here. Move it to `afterEach(() => vi.restoreAllMocks())`.

**Comments.**

Nine prose lines in `utils.ts` and twelve in the spec, explaining WHATWG behaviour and rehearsing the reasoning. We don't ship narrative comments — extract the intent into the function name (`isDownloadableOrigin` already does most of that work) and let the rationale live in the commit message and this thread, where it can be corrected.

**CI — not your fault.**

The red check is `Npm - check ts-any`, failing on `ui/tests/unit/composables/useReadOnlyYamlKeys.spec.ts: 0 -> 1`. That file isn't in your diff — it arrived on `develop` without a matching baseline entry, and #19461 fixed it on `develop` at 11:42 on 2026-09-15, shortly after your run started. Rebase onto current `develop` and it clears. The frontend unit tests themselves passed.

No design-system concerns — this touches a util and a spec, no components, element-plus imports, tokens, styles, or user-facing strings.

flcarre (changes requested):
Thanks for the rebase — and to be clear up front, the red check is not yours (details at the bottom).

But the rebase is all that landed. Diffing `b37e855` (what I reviewed on the 16th) against the current head, the only changes are develop's `moment` → `dayjs` migration in `getDateGrouping`. The `downloadUrl` hunk and the `describe("downloadUrl()")` block are byte-for-byte what they were, and the PR description still ends at the response to my *first* review. So none of the five points from last time have been answered. Rather than repeat them at length, here they are compactly, with the two that block.

**1. The comment still asserts a redirect that does not happen.** `ui/src/utils/utils.ts:317` still says the namespace files export "redirects to cloud storage". `NamespaceFileController.exportNamespaceFiles` builds the ZIP into a `ByteArrayOutputStream` and returns `new StreamedFile(...).attach(namespace + "_files.zip")` — same origin, `Content-Disposition: attachment`, no 3xx. This will be read as fact by the next person in this file, so it cannot ship as written.

**2. The branch does not do what it is there to do.** I re-checked all nine call sites: seven pass `createObjectURL` blobs, and the other two (`FileExplorer.vue:1083`, `useBaseNamespaces.ts:258`) are same-origin in every current deployment. So the guard protects against a navigation no caller can produce today. And in the one shape where it *would* fire, it makes things worse rather than better: `getFileContent` — the endpoint `FileExplorer` hits — has no `.attach()`, so cross-origin the browser ignores `download` entirely and the file downloads named after the URL's last path segment (`files`, no extension) instead of `file.fileName`. You get a stray tab *and* a lost filename. If that cross-origin path is worth defending, the fix is `.attach()` on `getFileContent` server-side, which is its own PR.

One correction to my last review while I'm here: the `VITE_APP_API_URL` vector is a little wider than I said. `useBaseNamespaces.ts`'s `base()` goes through `apiUrl()` too, so `exportFileDirectory` is subject to it as well, not only `FileExplorer`. Doesn't change the conclusion — both are same-origin by default — but worth having right if you do open that follow-up issue.

**3. Title vs code.** The title and lead commit still say "drop stray target=_blank on the download anchor"; the code keeps it conditionally. That headline goes into the changelog.

**4. Test hygiene.** `createElementSpy.mockRestore()` is still the last statement of each `it` instead of an `afterEach`. This suite runs `isolate: false` and `tests/unit/setup.ts` has no global restore, so one failing assertion leaves `document.createElement` spied for the rest of the worker — and `leakGuard.ts` watches globals, `document.title`, body children and storage, never `document`. Cascading failures with nothing pointing back here.

**5. Narrative comments.** Nine prose lines in `utils.ts`, twelve in the spec. We don't ship those; the reasoning belongs in the commit message and this thread.

**What I'd like, concretely.** Make it the two-line change it wants to be:

- delete `isDownloadableOrigin` and remove `link.setAttribute("target", "_blank")` unconditionally;
- delete the comment blocks in both files;
- collapse the three tests to one assertion that `target` is absent, with `afterEach(() => vi.restoreAllMocks())`;
- realign the title and the description with that.

That's a change I can merge on its own merits — the attribute is dead weight on every current caller — without needing the original repro, which is the one thing that genuinely can't be settled from here. If the `VITE_APP_API_URL` case still worries you, please open it as its own issue with the `getFileContent` missing-`Content-Disposition` angle attached; I'd be glad to see that one filed.

**CI — not your fault, and nothing for you to generate.** The failing check is `Translations - Key parity, placeholders and drift`, on four `topology_orientation` keys. Those came from #18815 — which is the exact commit your branch is rebased onto — and `develop` fixed them in #19526 at 17:25 on 2026-09-16, about two hours after your run started. Rebase once more and it goes green; you do not need to run `translations:generate` yourself.

No design-system concerns: a util and a spec, no components, styles, tokens or user-facing strings.

flcarre (approved):
This is the change I asked for, and all five points are addressed.

I re-checked the load-bearing claims rather than taking them from the commit message. `isDownloadableOrigin` and both comment blocks are gone, so the redirect assertion no longer ships. The call-site audit holds: seven of nine callers pass `createObjectURL` blobs, and `useBaseNamespaces.ts:261` and `FileExplorer.vue:1083` both resolve through `apiUrl()`, same-origin in every current deployment. Server side, `NamespaceFileController.exportNamespaceFiles` still ends in `.attach(namespace + "_files.zip")` and `getFileContent` still has no `.attach()` — which is the follow-up, not this anchor. Nothing in `kestra-ee` calls `downloadUrl`, so there is no EE surface either.

The test does what a regression guard should. `vi.spyOn(document, "createElement")` calls through, the anchor is the first element created, and with the attribute back in place `getAttribute("target")` returns `"_blank"` and the assertion fails. It still cannot observe the navigation in jsdom, which is fine — it locks the attribute, which is all this change claims. And `afterEach(() => vi.restoreAllMocks())` closes the leak properly: `vitest.config.unit.js` runs `isolate: false`, and `leakGuard.ts` watches globals, `document.title`, body children and storage but never `document` itself, so the per-test `mockRestore()` really was the wrong place for it.

Title, description and code now say the same thing. Approving.

One thing before you merge, and it is not code: the description says `closes #17322`. Nobody has reproduced that navigation and the reporter never came back with a browser/OS, so merging this will silently close a report that may still be live. Please switch it to `refs #17322` and leave the issue open until someone confirms — the attribute was dead weight on every caller, which is why this merges on its own merits, but that is a different claim from having fixed what was reported. And if you are still willing, the `getFileContent` missing-`Content-Disposition` issue is worth filing; happy to review it.

PR #19106, opened 2026-09-08
fix(core): enforce 63-char max for RFC 1123 labels

### LINKED ISSUE #19105: Rfc1123LabelValidator allows 64 character labels
### Describe the issue

I was going through the validation code and noticed Rfc1123LabelValidator has MAX_LENGTH = 64, and there is already a test that accepts a 64 character label.

From what I understand RFC hostname labels are limited to 63 characters, so this looks like it might be an off by one issue.

I ran the existing test too and the 64 character case passes.

Not sure if 64 is intentional here. If not, I can pick this up and send a fix.

### Environment

- Kestra Version: develop

siddheshbandgar:
## What does this PR do?

Fixes #19105 — `Rfc1123LabelValidator` currently allows labels up to **64** characters, but [RFC 1123 §2.1](https://www.rfc-editor.org/rfc/rfc1123#section-2.1) limits hostname labels to a maximum of **63** characters. This off-by-one lets invalid 64-character identifiers through for label-like fields (Worker Queue ids, Worker Group ids, routing tags, ...).

## Changes

- `Rfc1123LabelValidator`: `MAX_LENGTH` 64 → 63, and the regex segment `[a-z0-9-]{0,62}` → `{0,61}` so the compiled pattern itself caps at 63 characters (kept consistent with `MAX_LENGTH`).
- `Rfc1123Label`: updated Javadoc and validation message (`max 64` → `max 63` chars).
- `Rfc1123LabelTest`: the previously-accepted **64-character** label (`"a" + "-".repeat(62) + "b"`) is now asserted to be **rejected**; the valid boundary case is updated to a **63-character** label (`"a" + "-".repeat(61) + "b"`).

## Why is this correct?

RFC 1123 §2.1 ("Host Names and Numbers") states hosts "must also be able to handle host names of up to 63 characters", so a hostname label (lowercase alphanumerics and hyphens, starting/ending with alphanumeric) is at most 63 characters. The previous 64 limit and the matching test contradicted that definition.

## Tests

Live in `Rfc1123LabelTest` (updated above). `shouldRejectInvalidLabels` now covers both `MAX_LENGTH + 1` `a`s and the 64-char label that previously slipped through.


loicmathieu (approved):
LGTM, thanks

loicmathieu:
Even if it's a fix, it's also a small breaking change so I'm not backporting it.

PR #19112, opened 2026-09-08
fix(flows): recipe panels and task picker stop mutating their props

Piyush-r-bhaskar:
The recipe wizard panels wrote straight into the `recipe` prop with `v-model="recipe.cron"` and friends, and `BlockTaskPicker` assigned into `picker.pickerSearchInput.value` from a template ref. Ten `vue/no-mutating-props` violations in five files.

Each edited field is now its own `defineModel`, so `FlowRecipe.vue` binds `v-model:cron`, `v-model:webhookKey` and so on, and the parent keeps ownership of `recipe`. `BlockTaskPicker` destructures the ref out of `picker` and assigns through a `setSearchInput` handler.

After review: `watchNamespace` and `timezone` are bound to `clearable` selects, which emit `undefined` on clear, so those two models are no longer `required` and the `Invalid prop: type check failed` warning on clear is gone. `ExecutionPanel` and `NotifyGrid` now take `states`, `notify` and `triggerType` instead of the whole `recipe`, so each value reaches them by one route only.

`vue/no-mutating-props` is enabled in `eslint.config.js`. A new violation fails `npm run test:lint` locally; in CI eslint runs through reviewdog, so it shows up as a review comment rather than a red check.

### DIFF
--- a/ui/eslint.config.js
+++ b/ui/eslint.config.js
@@ -38,6 +38,7 @@ export default defineConfig([
             // Semantic rules
             "vue/block-lang": ["error", {"script": {"lang": "ts"}}],
             "vue/component-api-style": ["error", ["script-setup"]],
+            "vue/no-mutating-props": "error",
             "vue/this-in-template": "error",
             "vue/block-order": ["error", {order: ["template", "script", "style"]}],
             "vue/enforce-style-attribute": ["warn", {"allow": ["scoped"]}],
--- a/ui/src/components/flows/recipe/FlowRecipe.vue
+++ b/ui/src/components/flows/recipe/FlowRecipe.vue
@@ -53,6 +53,8 @@
                     <div class="trigger-config">
                         <ExecutionPanel
                             v-if="recipe.triggerType === 'execution'"
+                            v-model:watchNamespace="recipe.watchNamespace"
+                            v-model:includeSub="recipe.includeSub"
                             :recipe="recipe"
                             :namespaceOptions="namespaceOptions"
                             :namespacesLoading="namespacesLoading"
@@ -60,11 +62,12 @@
                         />
                         <SchedulePanel
                             v-else-if="recipe.triggerType === 'schedule'"
-                            :recipe="recipe"
+                            v-model:cron="recipe.cron"
+                            v-model:timezone="recipe.timezone"
                         />
                         <WebhookPanel
                             v-else-if="recipe.triggerType === 'webhook'"
-                            :recipe="recipe"
+                            v-model:webhookKey="recipe.webhookKey"
                             :systemNamespace="systemNamespace"
                             :flowId="flowId"
                         />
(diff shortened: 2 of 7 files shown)


MilosPaunovic (approved):
Verified the claim end to end: on `develop`, `vue/no-mutating-props` reports exactly 10 violations and all of them are in the 5 files touched here, so nothing else in `ui` trips the new rule. At this head the two changed directories lint clean.

Cross-repo check: `kestra-ee/ui-ee/eslint.config.js` re-exports this config wholesale, so the rule lands on `EE` the moment this merges. Ran the same override across `ui-ee` on `develop`: 328 `.vue` files, zero violations, so no `EE` companion is needed. No `kestra-io/actions` change either, since the `EE` composite already checks out `ui/eslint.config.js`.

One nit, and two notes.

**`required: true` on the two models fed by a `clearable` select**

`watchNamespace` in `ExecutionPanel.vue` and `timezone` in `SchedulePanel.vue` are bound to `clearable` `KsSelect`s. Clicking the clear icon makes `element-plus` emit `valueOnClear`, which defaults to `undefined` (`useSelect.mjs` -> `use-empty-values`; neither `KsSelect` nor an `el-config-provider` overrides it). That `undefined` round-trips through `recipe` and comes back down as the prop.

`defineModel<string>("timezone", {required: true})` compiles to `"timezone": {type: String, ...{required: true}}`, and Vue's prop validation does not short-circuit on `undefined` when `required` is set, so it falls through to the type check:

    [Vue warn]: Invalid prop: type check failed for prop "timezone". Expected String with value "undefined", got Undefined
      at <SchedulePanel cron="0 9 * * *" timezone=undefined >

Dev-only and nothing breaks (the value still writes through), and the underlying type lie is pre-existing: `RecipeState.timezone` is `string` while a `clearable` select can null it. But the warning is new. Dropping `required: true` on just those two (`defineModel<string | undefined>("timezone")`), or setting `:valueOnClear="''"` on the selects, avoids it. The other six models are fed by inputs and checkboxes and are fine as they are.

**`ExecutionPanel` and `NotifyGrid` still take `:recipe` alongside the new models**

Both keep the whole `recipe` prop for reads (`recipe.states`, `recipe.notify`, `recipe.triggerType`) while writing through the models, so the same reactive object arrives by two routes and `recipe.watchNamespace` and `watchNamespace` are the same value under two names. Not a bug, and `SchedulePanel` and `WebhookPanel` did get the full treatment, but it leaves a trap for whoever next reaches for `recipe.slackChannel` inside `NotifyGrid`. Worth a follow-up to finish the migration, or a comment marking `recipe` read-only in those two.

**On "so this cannot come back"**

`eslint` runs through `reviewdog/action-eslint@v1` with `reporter: github-pr-review` and no `fail_on_error`, in both the `OSS` and `EE` frontend workflows. A future violation gets a review comment on the diff but does not turn the check red. Nothing to change here, just worth not treating it as a hard gate.

`BlockTaskPicker` looks good: the destructure matches how the rest of the file already reads `props.picker`, the handler signature matches the `ref` in `useTaskPicker.ts` exactly, and the stable function also stops the `ref` churn the inline arrow caused on every re-render.

PR #19137, opened 2026-09-08
fix(scheduler): an unparseable flow no longer stalls its scheduling loop or floods the logs

AcevedoR:
### 🔗 Related Issue

No issue yet: found by the 1.3 → 2.0 migration QA campaign (internal report `oss-postgres-standalone-unmigrated-schedule-atda`, finding F8).

### ✨ Description

**Problem.** After upgrading to 2.0, a flow that still uses a removed task type (for example `io.kestra.plugin.core.flow.ForEach` left unmigrated) is stored as a `FlowWithException` without trigger definitions. The first time its `Schedule` trigger comes due, `DefaultSchedulableTriggerFetcher` dereferences the null trigger list; the `NullPointerException` aborts the evaluation of every trigger of that loop's vNodes, and because `TriggerSchedulingLoop` only advances `nextScheduleTime` after `onSchedule` returns, the tick is re-run immediately. Measured on 2.0.0 (OSS, Postgres, standalone): ~360 iterations per second, two stack traces each, 2.4 GB of container logs in 20 minutes (≈ 8 GB/h), and every schedule sharing the thread (a quarter of them in standalone) silent until the flow is fixed. Nothing is visible in the UI: no execution, and the trigger row is not listed.

**Fix.**
- `TriggerFlowParser.parseForTrigger` returns `null` when the stored flow has no trigger definitions (there is nothing to degrade to) and logs the parse failure without the stack trace. Flows that still carry their triggers keep the existing degrade-to-stored-flow behaviour.
- `DefaultSchedulableTriggerFetcher` disables the trigger state once and writes the reason to the trigger's logs, the same way an `InvalidTriggerConfigurationException` is handled. Saving the fixed flow re-enables the trigger through the existing flow-update path (verified on 2.0.0: the rewritten flow fired at its next tick). The trigger-list dereference is also null-guarded as a safety net.
- Null guard on the remaining trigger-list dereference in `TriggerEventHandler.findTrigger`.

**Evidence.** The two new tests fail on `develop` and pass with this change: the fetcher test with `NullPointerException: Cannot invoke "java.util.List.stream()" because the return value of "FlowWithSource.getTriggers()" is null`, the parser test because the broken flow is returned instead of rejected.

End to end, on `v2.0.0` + this commit cherry-picked (OSS, Postgres, standalone, 1.3.37 auto-migrated with the same forgotten ForEach flow): 0 `Error in scheduling loop` in 33 minutes (161 353 in 7.5 minutes before), the every-minute schedule kept firing once a minute during an 11-minute soak (stalled before), and the invalid trigger was disabled once with `Disabled: the flow cannot be parsed on this version (… Invalid type: io.kestra.plugin.core.flow.ForEach). Fix the flow and save it to re-enable the trigger.` in its trigger log.

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched modules (`./gradlew :scheduler:test`, `./gradlew :jdbc-h2:test --tests "H2RunnerTest"`)

### 📝 Additional Notes

- Not in this PR (review): `TriggerSchedulingLoop` only advances `nextScheduleTime` after `onSchedule` returns, so any exception escaping `onSchedule` re-runs the tick immediately; that is what turned this NPE into ~360 iterations/s and ~8 GB/h of logs. Left to the ongoing loop rework.
- Trade-off: a transient parse failure (a plugin not installed yet) also disables the trigger until the flow is saved again or the trigger is re-enabled by hand. The WARN and the trigger log say so. This felt safer than re-parsing and re-logging the same flow every second forever.
- Deliberately not in this PR (different module): `TriggerController.toApiTriggerAndState` drops the trigger record when the flow has no trigger definitions, so the disabled trigger is not visible in the Triggers page while the flow is broken. Follow-up `fix(triggers)` PR.
- Candidate for the 2.0.x patch branch: any customer with one forgotten scheduled flow hits this at the first cron tick after the upgrade.

### 🤖 AI Authors

Drafted with Claude Code from the QA findings, reviewed by @AcevedoR. Why did the cat refuse to migrate its ForEach flow? It only had nine lives and the scheduler was burning through 360 of them per second. 🐱

### DIFF
--- a/scheduler/src/main/java/io/kestra/scheduler/TriggerScheduler.java
+++ b/scheduler/src/main/java/io/kestra/scheduler/TriggerScheduler.java
@@ -22,6 +22,7 @@ import org.slf4j.event.Level;
 import com.google.common.base.Throwables;
 
 import io.kestra.core.exceptions.FlowBlockedException;
+import io.kestra.core.exceptions.FlowProcessingException;
 import io.kestra.core.exceptions.InvalidTriggerConfigurationException;
 import io.kestra.core.metrics.MetricRegistry;
 import io.kestra.core.models.Label;
@@ -291,6 +292,8 @@ public class TriggerScheduler {
                 e.getMessage()
             );
             logBlockedByGovernance(flow, e);
+            return null;
+        } catch (FlowProcessingException e) {
             return null;
         }
     }
--- a/scheduler/src/main/java/io/kestra/scheduler/TriggerEventHandler.java
+++ b/scheduler/src/main/java/io/kestra/scheduler/TriggerEventHandler.java
@@ -52,6 +52,7 @@ import io.kestra.core.scheduler.model.TriggerType;
 import io.kestra.core.scheduler.service.TriggerExecutionPublisher;
 import io.kestra.core.scheduler.store.TriggerStateStore;
 import io.kestra.core.services.ConditionService;
+import io.kestra.core.utils.ListUtils;
 import io.kestra.core.utils.Logs;
 import io.kestra.scheduler.internals.NextEvaluationDate;
 import io.kestra.scheduler.stores.FlowMetaStore;
@@ -611,7 +612,7 @@ public class TriggerEventHandler {
             return Pair.of(null, null);
         }
 
-        AbstractTrigger trigger = flow.getTriggers().stream()
+        AbstractTrigger trigger = ListUtils.emptyOnNull(flow.getTriggers()).stream()
             .filter(it -> it.getId().equals(event.id().getTriggerId()))
             .findFirst()
             .orElse(null);
--- a/scheduler/src/main/java/io/kestra/scheduler/internals/DefaultSchedulableTriggerFetcher.java
+++ b/scheduler/src/main/java/io/kestra/scheduler/internals/DefaultSchedulableTriggerFetcher.java
@@ -16,6 +16,7 @@ import org.slf4j.LoggerFactory;
 import org.slf4j.event.Level;
 
 import io.kestra.core.exceptions.FlowBlockedException;
+import io.kestra.core.exceptions.FlowProcessingException;
 import io.kestra.core.exceptions.InvalidTriggerConfigurationException;
 import io.kestra.core.models.conditions.ConditionContext;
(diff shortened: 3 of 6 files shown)


scheduler/src/main/java/io/kestra/scheduler/TriggerScheduler.java
```diff
@@ -286,6 +287,8 @@ private FlowWithSource parseForTriggerOrSkip(final FlowWithSource flow) {
             );
             logBlockedByGovernance(flow, e);
             return null;
+        } catch (FlowProcessingException e) {
+            return null;
```
fhussonnois, line 297: Why you don't disable the trigger here ?

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
```diff
@@ -167,8 +167,8 @@ public void run() {
 
                     // Check whether triggers should be scheduled
                     if (now.isAfter(nextScheduleTime) || now.equals(nextScheduleTime)) {
-                        triggerScheduler.onSchedule(clock, now, assignments);
                         nextScheduleTime = nextScheduleTime.plusMillis(SCHEDULE_INTERVAL_MILLIS);
+                        triggerScheduler.onSchedule(clock, now, assignments);
```
fhussonnois: This change is not related to the current fix; in addition, I'm currently working on fixing some issues on the loop so best to not change the TriggerSchedulingLoop.

fhussonnois (changes requested):

scheduler/src/main/java/io/kestra/scheduler/TriggerSchedulingLoop.java
AcevedoR: ok

scheduler/src/main/java/io/kestra/scheduler/TriggerScheduler.java
AcevedoR: done

fhussonnois (approved):

PR #19145, opened 2026-09-08
refactor(iam): centralize repository namespace ACL into one policy bean

nkwiatkowski:
## Problem

Every namespace-scoped EE repository computed its own namespace ACL and ANDed it onto its tenant filter. The same clause was duplicated across each backend, and the JDBC and Elasticsearch implementations had drifted, so a caller could see different rows depending on the backend. The pattern also forced an EE subclass per repository whose only job was to add that filter.

## Fix

Introduce a single access-control collaborator, `NamespaceAccessControl`, that returns a backend-agnostic `AccessScope` (`GLOBAL` / `NAMESPACES` / `DENY_ALL`) for a `QueryFilter.Resource`. OSS ships a no-op `GlobalNamespaceAccessControl` (grants everything, since OSS has no namespace ACL); EE `@Replaces` it with a `CurrentUserContext`-based bean.

Each namespace-scoped JDBC repository consults the bean at its `defaultFilter` chokepoint, and `AbstractJdbcRepository.aclCondition(...)` translates the `AccessScope` into a jOOQ condition in one place. The bean is injected lazily (`BeanProvider`) because the EE implementation transitively depends on a repository of the same base, which would otherwise be a startup cycle. `AccessScope.namespaces(empty)` collapses to `DENY_ALL`, so an empty grant fails closed rather than degrading to allow-all.

This is the OSS half; the EE half (the `@Replaces` bean, deletion of the per-repository EE subclasses, and the matching Elasticsearch translator) is the companion PR.

## Evidence

`aclCondition` reproduces the previous EE clause exactly (global bypass → no restriction, otherwise the granted namespace or its dot-delimited subtree, prefix-sibling excluded). OSS behaviour is unchanged: under `GlobalNamespaceAccessControl` the clause is a no-op. JDBC repository tests pass on H2/Postgres/MySQL; the EE ACL suites (H2 + Elasticsearch) in the companion PR exercise the enforcing path.

Companion PR: kestra-io/kestra-ee#10853.

### DIFF
--- a/core/src/main/java/io/kestra/core/repositories/GlobalNamespaceAccessControl.java
+++ b/core/src/main/java/io/kestra/core/repositories/GlobalNamespaceAccessControl.java
@@ -7,7 +7,7 @@ import jakarta.inject.Singleton;
 
 /**
  * Default OSS {@link NamespaceAccessControl}: grants global access to every resource, since OSS has no
- * namespace ACL. EE replaces this bean with a {@code CurrentUserContext} implementation.
+ * namespace ACL.
  */
 @Singleton
 public class GlobalNamespaceAccessControl implements NamespaceAccessControl {
--- a/core/src/main/java/io/kestra/core/repositories/GlobalNamespaceAccessControl.java
+++ b/core/src/main/java/io/kestra/core/repositories/GlobalNamespaceAccessControl.java
@@ -1,0 +1,19 @@
+package io.kestra.core.repositories;
+
+import io.kestra.core.models.AccessScope;
+import io.kestra.core.models.QueryFilter;
+
+import jakarta.inject.Singleton;
+
+/**
+ * Default OSS {@link NamespaceAccessControl}: grants global access to every resource, since OSS has no
+ * namespace ACL. EE replaces this bean with a {@code CurrentUserContext} implementation.
+ */
+@Singleton
+public class GlobalNamespaceAccessControl implements NamespaceAccessControl {
+
+    @Override
+    public AccessScope namespaceScope(QueryFilter.Resource resource) {
+        return AccessScope.global();
+    }
+}
--- a/core/src/main/java/io/kestra/core/repositories/NamespaceAccessControl.java
+++ b/core/src/main/java/io/kestra/core/repositories/NamespaceAccessControl.java
@@ -7,11 +7,8 @@ import io.kestra.core.models.QueryFilter;
  * Access-control collaborator for namespace-scoped repositories. A repository consults it at its
  * {@code defaultFilter} chokepoint and translates the returned {@link AccessScope} into its own query
  * language (a jOOQ condition, an Elasticsearch query). This keeps the access-control <em>policy</em> in
- * one place (a single EE bean) while the <em>translation</em> stays in each dialect-aware backend, so
- * there is no per-repository ACL subclass.
- * <p>
- * OSS provides a {@link #GLOBAL} no-op default; EE replaces the bean with a {@code CurrentUserContext}
- * implementation that maps each resource to the permission gating its namespaces.
+ * one place while the <em>translation</em> stays in each dialect-aware backend, so there is no
+ * per-repository ACL subclass.
  */
 public interface NamespaceAccessControl {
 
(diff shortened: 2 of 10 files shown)


core/src/main/java/io/kestra/core/repositories/GlobalNamespaceAccessControl.java
```diff
@@ -0,0 +1,19 @@
+package io.kestra.core.repositories;
+
+import io.kestra.core.models.AccessScope;
+import io.kestra.core.models.QueryFilter;
+
+import jakarta.inject.Singleton;
+
+/**
+ * Default OSS {@link NamespaceAccessControl}: grants global access to every resource, since OSS has no
+ * namespace ACL. EE replaces this bean with a {@code CurrentUserContext} implementation.
```
loicmathieu:
```suggestion
 * namespace ACL.
```

Don't leak too much detail on EE implementation (maybe already too late)

core/src/main/java/io/kestra/core/repositories/NamespaceAccessControl.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Access-control collaborator for namespace-scoped repositories. A repository consults it at its
+ * {@code defaultFilter} chokepoint and translates the returned {@link AccessScope} into its own query
+ * language (a jOOQ condition, an Elasticsearch query). This keeps the access-control <em>policy</em> in
+ * one place (a single EE bean) while the <em>translation</em> stays in each dialect-aware backend, so
+ * there is no per-repository ACL subclass.
+ * <p>
+ * OSS provides a {@link #GLOBAL} no-op default; EE replaces the bean with a {@code CurrentUserContext}
+ * implementation that maps each resource to the permission gating its namespaces.
```
loicmathieu:
```suggestion
```

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
```diff
@@ -59,6 +62,17 @@ public abstract class AbstractJdbcRepository {
     // not bean-managed) set it in AbstractJdbcLogDataStore.initFrom — hence protected, not private.
     protected SystemFlowsConfiguration systemFlowsConfiguration;
 
+    @Inject
+    // Injected lazily as a provider: the EE bean depends (via CurrentUserContext → RBACService) on the
+    // role repository, which itself extends this base, so eager injection would be a circular dependency.
+    // Resolved only when a query is built, by which point the bean graph is complete; null for
+    // non-bean-managed instances (deserialized log-store plugins), which enforce ACL another way.
```
loicmathieu:
```suggestion
    // Injected lazily as a provider: the EE bean depends on the
    // role repository, which itself extends this base, so eager injection would be a circular dependency.
```

jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcRepository.java
```diff
@@ -89,6 +103,35 @@ protected Condition defaultFilterWithNoACL(String tenantId, boolean deleted) {
...
+     * imposes no restriction, {@code DENY_ALL} matches nothing, {@code NAMESPACES} matches a granted
+     * namespace or its dot-delimited subtree ({@code io.kestrax} is not inside {@code io.kestra}).
+     *
+     * @param namespaceColumn the column holding the namespace, which is not always named {@code namespace}.
+     */
+    protected Condition aclCondition(Resource resource, String namespaceColumn) {
+        AccessScope scope = namespaceAccessControl().namespaceScope(resource);
+        Field<String> column = field(namespaceColumn, String.class);
+        return switch (scope.kind()) {
+            case GLOBAL -> DSL.trueCondition();
```
loicmathieu:
```suggestion
            case GLOBAL -> DSL.noCondition();
```

`noCondition()`  emits no filters in ther SQL query, whereas `trueCondition` emits a `WHERE TRUE`, small query parsing improvements.

loicmathieu (changes requested):
LGTM, I send small improvements, please apply them before merging.

loicmathieu (approved):
LGTM

PR #19146, opened 2026-09-08
feat(ai): copilot stop button

### LINKED ISSUE #19072: [2.0] AI Copilot - Add a stop button to an ongoing chat
### Feature description

While the AI Copilot is thinking, there is no stop button. This is common with things like Claude Desktop or Gemini to prevent waste of tokens or if it's clearly going the wrong way. The send button is greyed out but instead we could add a stop button instead?

Image

Avirup-001:
Closes https://github.com/kestra-io/kestra/issues/19072

---

### ✨ Description

While Copilot is thinking, Send was greyed out with no way to halt the turn. The composer now shows **Stop** button instead. Clicking it aborts the in-flight SSE request, marks the turn cancelled, and returns the thread to IDLE so the user can send again. Once the provider has emitted a first partial (and langchain4j has supplied a `StreamingHandle`), Stop also cancels that HTTP stream so further tokens are not generated. An early Stop before that first partial still abandons the turn, but the upstream request may keep running.

Stop also clears the global page loaders and skips the thinking-mark “turn finished” animation (that beat is for completion, not cancel). There is no new cancel endpoint — aborting the existing chat/confirm SSE is the signal.

Verified locally from the UI against a backend built from this branch: Send → Stop logged `Client disconnected; aborting turn` and the composer returned to Send. The logs are attached below..

### 🎨 Frontend Checklist

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched modules (`./gradlew :webserver:test --tests "AgentOrchestratorTest"`)

### Recording

https://github.com/user-attachments/assets/75b65810-8f27-4d88-8f51-5f9f58dbb654

### Backend Logs
```
17:41:11.445 INFO  ai-agent-orchestrator_%d i.k.w.s.ai.agent.AgentOrchestrator Copilot thread 2jbQb01KRkHYGP3Cy4m5k0: model requested 1 tool call(s): [search_docs]
17:41:11.114 INFO  ai-agent-orchestrator_%d i.k.w.s.ai.agent.AgentOrchestrator Copilot thread 2jbQb01KRkHYGP3Cy4m5k0: calling tool 'search_docs' (kind=PLATFORM, family=READ)
17:41:12.859 INFO  ai-agent-orchestrator_%d i.k.w.s.ai.agent.AgentOrchestrator Copilot thread 2jbQb01KRkHYGP3Cy4m5k0: tool 'search_docs' returned ok (22682 chars)
17:41:12.169 DEBUG ai-agent-orchestrator_%d i.k.w.s.ai.agent.AgentOrchestrator Client disconnected; aborting turn for thread 2jbQb01KRkHYGP3Cy4m5k0
17:41:19.265 INFO  ai-agent-orchestrator_%d i.k.w.s.ai.agent.AgentOrchestrator Copilot thread 2jbQb01KRkHYGP3Cy4m5k0: model requested 1 tool call(s): [get_doc]
17:41:19.641 INFO  ai-agent-orchestrator_%d i.k.w.s.ai.agent.AgentOrchestrator Copilot thread 2jbQb01KRkHYGP3Cy4m5k0: calling tool 'get_doc' (kind=PLATFORM, family=READ)
17:41:20.486 INFO  ai-agent-orchestrator_%d i.k.w.s.ai.agent.AgentOrchestrator Copilot thread 2jbQb01KRkHYGP3Cy4m5k0: tool 'get_doc' returned ok (23898 chars)
17:41:22.072 INFO  default-eventLoopGroup-1-3 io.kestra.webserver.access 2026-09-08T17:41:18.089Z | POST /api/v1/main/ai/threads/2jbQb01KRkHYGP3Cy4m5k0/chat HTTP/1.1 | status: 200 | ip: 192.168.215.1 | length: 1524 | duration: 4474
```


MilosPaunovic (changes requested):
Thanks for picking this up. The shape of the change is right: no new cancel endpoint, aborting the existing `chat`/`confirm` `SSE` as the signal, and reusing the `CANCELLED` message type and `abortCancelled` plumbing that was already there. Reusing the existing `ai.copilot.cancelled` transcript line rather than inventing a new state was the right call.

I ran the following locally against the branch, all green: `./gradlew :webserver:test --tests "io.kestra.webserver.services.ai.agent.AgentOrchestratorTest"` (23/23), `vitest` over the touched and neighbouring unit specs (195/195), `vue-tsc` across the `design-system`, app and test projects, `eslint` on every changed `.ts`/`.vue`, and `node ui/scripts/translations/check-translations.mjs --scope oss`. The backend checklist item "Code compiles and tests pass for the touched modules" is unchecked but does pass, so you can tick it. I also ran `shouldAbortHungModelCallWhenClientDisconnects` five times in isolation and it was stable.

A few things to address before this goes in.

### 1. Send is re-enabled before the server has released the `thread`, so an immediate re-send `409`s

`ui/src/components/ai/copilot/useAiChat.ts:322-327`

The abort branch sets `status.value = "IDLE"` locally, so `canSend` flips true as soon as the `fetch` rejects. Server-side the `thread` row is only returned to `IDLE` by `abortCancelled` -> `finishTurn`, which is reached at the next `runLoop` checkpoint. Inside `callModel` that is bounded by the 50 ms poll, but a cancel that lands **while a tool is executing** waits for that tool to return, because the only check on that path is at the top of the next `runLoop` iteration. The log in your description shows `search_docs` at 1.7 s and `get_doc` at 0.8 s.

In that window `POST …/ai/threads/{id}/chat` reaches `tryMarkRunning(thread, mode, IDLE)`, gets an empty `Optional`, and throws `ConflictException` -> `409` -> `toErrorCode` -> `error.value = "turnInProgress"`.

Concretely: the user presses Stop while a tool is running (exactly the "it is clearly going the wrong way" case from https://github.com/kestra-io/kestra/issues/19072), types a corrected prompt within a second, and gets "a turn is already in flight" instead of a new turn.

The cheapest fix is a `sink.isCancelled()` check on the tool-dispatch path, so the window is bounded by the poll interval rather than by a tool's duration. Alternatively, do not flip to `IDLE` optimistically: re-`GET` the `thread` once after the abort and enable Send only when the server reports `IDLE`.

### 2. The new `instanceof Error` guard in `isAbortError` is a narrowing the tests cannot catch

`ui/src/components/ai/copilot/useAiChat.ts:448`

This changed from `(e as Error)?.name === "AbortError"` to `e instanceof Error && (…)`. A real abort surfaces as `DOMException("…", "AbortError")` out of `reader.read()`, and in current browsers `DOMException instanceof Error` is true, so production behaviour is fine today. But the guard is now realm-sensitive for no gain, since the only thing read off `e` is `.name`.

It is already false in this repo's own test environment. In `jsdom`, `new DOMException("Aborted", "AbortError") instanceof Error` evaluates to `false` (the prototype chain prints `DOMException -> Error -> Object`, but against a different `Error` intrinsic). If you swap the `useAiChat.spec.ts` mock's hand-built `Error` for the `DOMException` a browser actually throws, `cancel aborts a streaming turn, marks it cancelled, and returns to IDLE` fails with `expected 'generic' to be null`: Stop is reported as a generic error, no `CANCELLED` marker is pushed, and the end-gather animation plays after all. The new tests pass only because the mock patches `name` onto a plain `Error`.

Dropping `e instanceof Error &&` and keeping the name check removes the hazard and costs nothing.

While you are in there: `"CanceledError"` is axios's cancel name, and this path goes through the fetch-based `useClient()` facade, so that branch is dead.

### 3. Provider-stream cancellation is best-effort, and the timeout path never cancels at all

`webserver/src/main/java/io/kestra/webserver/services/ai/agent/AgentOrchestrator.java:493` and `:515`

`handle` is only populated from `PartialResponseContext`, `PartialThinkingContext` and `PartialToolCallContext`, so until the provider's first partial arrives it is null and `abortIfCancelled` cancels nothing. An early Stop, which is the common case, correctly abandons the turn but leaves the upstream request running.

For the `OSS` provider specifically: `langchain4j-google-ai-gemini` 1.18.1 only dispatches `onPartialResponse` and `onPartialThinking` with a handle, never `onPartialToolCall`. So a Gemini turn that goes straight to a tool call with no text or thinking preamble never yields a handle either. `shouldAbortHungModelCallWhenClientDisconnects` exercises precisely this no-handle path, since `hang()` never assigns `lastHandle`, so nothing in the suite covers the case.

Separately, the timeout branch does `future.cancel(true)` but never `handle.get().cancel()`, so a timed-out call keeps generating and billing tokens. Since https://github.com/kestra-io/kestra/issues/19072 is motivated by token waste, that is worth closing in the same change.

Could you also soften the claim in the description that Stop "cancels the provider HTTP stream so further tokens are not generated"? It holds only once a partial carrying a handle has arrived.

### 4. `en.json` picked up unrelated reformatting

`ui/src/translations/en.json`

The file lost its trailing newline (`develop` ends `0a`, the branch ends `}`) and the `state_history` block was re-indented from 8 spaces to 6, which is 14 diff lines unrelated to the Stop button. That signature is `JSON.stringify(obj, null, 2)` rather than a write through `ui/scripts/translations/files.ts:writeIfChanged`, which exists specifically to keep the tooling formatting-neutral (see https://github.com/kestra-io/kestra/issues/17823). No `CI` gate catches it: `Translations - Key parity, placeholders and drift` is green.

Please restore the newline and revert the `state_history` block so the diff is just the one new `ai.copilot.stop` key.

### Worth considering, not blocking

- **Push-based cancel instead of polling.** `FluxTurnEventSink` already receives a push signal via `emitter.onCancel` and narrows it to a volatile boolean. A `TurnEventSink.onCancel(Runnable)` hook would let `callModel` register `handle.cancel()` plus the future completion directly, making Stop instant and removing the loop, `MODEL_CANCEL_POLL_MS` and the 50 ms worst-case latency. To be clear, the current loop is not a busy-wait: `future.get(waitMs)` wakes on completion, so a normal turn pays no added latency. The cost is roughly 20 wakeups per second per in-flight turn.
- `client-facade.ts:225`: the comment says "without this the NProgress `requestsCompleted` bump never runs and the loading indicator stays stuck", but the sibling `kestraHttp.ts` change opts every `stream()` call out of `NProgress` entirely, so that bump is now deliberately skipped. The catch itself is correct (error-interceptor parity with `axiosLikeRequest`); only the stated reason is stale.
- The timeout `IllegalStateException` lost its `TimeoutException` cause.
- The `onPartialThinking` and `onPartialToolCall` overrides no longer delegate to their single-argument counterparts the way the interface defaults do. No behaviour change today, since neither is implemented, but a future `onPartialThinking(PartialThinking)` implementation would be silently dead.
- After Stop the composer re-enables but is not focused. `onReject` already does `footerComposer.value?.focus()` for the same "hand control back so the user can keep typing" beat, so wiring the same thing into the stop handler would match.
- `shouldAbortHungModelCallWhenClientDisconnects` races the class-wide `PT1S` `kestra.ai.agent.model-call-timeout`: the cancel has to cross a freshly started single-thread executor and be observed by a 50 ms poll before the deadline, otherwise the assertion flips to the timeout path. The margin is large and it was stable over five local runs, so this is only a note.

### Scope notes

No `EE` companion `PR` is needed. `EE` overrides only `CopilotThreadControls.vue`, `CopilotThreadList.vue`, `useAiThreads.ts` and `appDraftActions.ts` under `ui-ee/src/override/components/ai/copilot/`, and the `EE` `AiAgentController` adds only `resolveUserId` and thread listing. `ui-ee/src/main.ts` imports `setupKestraHttp` from the `OSS` source directly, so the `NProgress` change carries over. `ai.copilot.stop` is an `OSS` key, so no `ee_translations` entry is required. `ui-ee/tests/e2e/ai-copilot-tools.spec.ts` does drive `[data-test="copilot-send"]`, which now leaves the `DOM` while streaming rather than only being disabled, but it clicks Send only before the turn starts and its mocked `SSE` resolves immediately, so it is unaffected.

One thing to be aware of for `EE`: it ships ten providers, and for any whose `langchain4j` module supplies no `StreamingHandle`, Stop degrades to "turn abandoned, upstream request keeps running". The null checks handle that correctly, so it is graceful, just worth knowing.

On backporting: all six touched files exist on `releases/v2.0.x` and are currently identical to `develop`, so this cherry-picks cleanly, and the issue is tagged `[2.0]`. `releases/v1.3.x` has none of them, so 1.3 is out of scope.

@ammeek this is also worth a look from you on the backend side. The `StreamingHandle` capture in `abortIfCancelled` and the rework of `callModel`'s bounded wait into a polling loop are the parts I would most want a second pair of eyes on.

webserver/src/main/java/io/kestra/webserver/services/ai/agent/AgentOrchestrator.java
ammeek, line 330: We need to check if the sink is canceled between tool requests to prevent a chain of tool request running after the sink has been canceled.

ammeek (changes requested):
Thank you for the pull request @Avirup-001. All looks good to me other than a very small change around tool executions.

webserver/src/main/java/io/kestra/webserver/services/ai/agent/AgentOrchestrator.java
Avirup-001: the tool loop now checks sink.isCancelled() between requests (and after the batch) so a cancelled sink cannot run the rest of the chain.

Avirup-001:
@MilosPaunovic

• Tool-dispatch now checks sink.isCancelled() so Stop does not start the next tool. An in-flight catalog.dispatch() still has to return; that is the cheap-fix tradeoff.

• isAbortError is a .name === "AbortError" check again. The axios CanceledError branch is gone, and the spec now rejects with DOMException("Aborted", "AbortError").

• Timeout and interrupt call cancelProviderStream on the captured handle. Early Stop before the first partial remains best-effort. The description has been updated to match.

• en.json now differs from develop only by the ai.copilot.stop key. Trailing newline and state_history indent are restored.

• Also applied the nits: NProgress comment, TimeoutException cause, single-arg delegation for onPartialThinking / onPartialToolCall, and composer focus after Stop.

webserver/src/main/java/io/kestra/webserver/services/ai/agent/AgentOrchestrator.java
MilosPaunovic: If you've addressed any of the requested changes from the PR review, @Avirup-001, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

MilosPaunovic (changes requested):
Round 2 on `3bfc864`. Verified locally: `./gradlew :webserver:test --tests "io.kestra.webserver.services.ai.agent.AgentOrchestratorTest"` 24/24, the touched `vitest` specs 195/195, `eslint` clean, `node ui/scripts/translations/check-translations.mjs --scope oss` passing.

Four things still open, the first of which is new and is a data-corruption bug.

### 1. The cancel guard in `executeTool` leaves a dangling `TOOL_CALL` that breaks the `thread` permanently

`webserver/src/main/java/io/kestra/webserver/services/ai/agent/AgentOrchestrator.java:376`

`runLoop` persists the call row at `:344` (`threadManager.appendToolCall`) and only then calls `executeTool` at `:361`. The new guard at `:376` returns before `catalog.dispatch`, so no `TOOL_RESULT` row is ever written for that call. The race window is `parseArguments` + `catalog.byName` + that `DB` insert - milliseconds, but a real disconnect arrives on a Netty thread while the turn runs on a virtual thread, so it is genuinely reachable.

I reproduced it with a probe sink whose `isCancelled()` flips exactly between the batch-loop check at `:331` and the `executeTool` check at `:376`. The persisted transcript comes out as:

```
ROW TEXT       toolCallId=-
ROW TOOL_CALL  toolCallId=c1
ROW CANCELLED  toolCallId=-
```

No `TOOL_RESULT`. `ChatMessageAdaptor.project` has no pairing guard - it turns every `TOOL_CALL` row into an `AiMessage` carrying a `ToolExecutionRequest`, and only emits a `ToolExecutionResultMessage` for `TOOL_RESULT` rows. So the next turn on that `thread` projects to:

```
PROJECTED UserMessage
PROJECTED AiMessage REQUESTS=[c1]
```

an assistant turn whose tool call has no response. Gemini and OpenAI both reject that shape, and because the projection is rebuilt from the durable log on every turn, the `thread` stays broken for good - New chat is the only way out.

`resumeHeldAction:245` is the same class: the held action's `TOOL_CALL` is persisted at `:344` before `suspendForAction`, and the new guard aborts without writing a result for it.

The `:376` guard is also redundant with the batch-loop check at `:331`, which runs a few statements earlier, so the simplest fix is to drop it and let `:331` carry the intent (the tool-chain concern this was added for is already covered by `:331` and `:364`). If you would rather keep it, write a cancelled tool result before returning so the pair is closed, or move `appendToolCall` inside `executeTool` after the check. Either way `resumeHeldAction:245` needs the same treatment.

Worth adding regardless: a defensive pairing guard in `ChatMessageAdaptor.project` that drops a `TOOL_CALL` with no matching `TOOL_RESULT`. That turns any future occurrence of this into a lost tool call rather than a dead `thread`, and it heals threads that have already hit it - the same reasoning as the existing "never replay an empty assistant turn" guard a few lines above it.

### 2. The composer focus in `onStop` never lands

`ui/src/components/ai/copilot/CopilotChat.vue:405`

`cancel()` only calls `abort.abort()`. `streaming` and `canSend` flip later, when the abort rejection reaches `runStream`'s `finally`. So at `nextTick` the textarea still has `:disabled="disabled"` set, and `HTMLElement.focus()` on a disabled `<textarea>` is a no-op.

Mounting `CopilotChat` with `attachTo: document.body`, `streaming` true, and emitting `stop` leaves `document.activeElement` on `BODY`. Same result when `cancel` flips `streaming`/`canSend` in a microtask, which is the realistic ordering - `nextTick` still resolves first.

`onReject` works because it `await`s the whole `confirm()` turn before focusing. The equivalent here is to focus on the transition rather than immediately: watch `canSend` and focus once it goes true after a stop.

### 3. The frontend still returns to `IDLE` optimistically

`ui/src/components/ai/copilot/useAiChat.ts:322`

The backend window is smaller now that the rest of the chain is skipped, but `catalog.dispatch` for the tool already in flight still has to return before `abortCancelled` runs, and the composer re-enables the moment the `fetch` rejects. A prompt sent inside that window still reaches `tryMarkRunning(thread, mode, IDLE)`, gets an empty `Optional`, and comes back as `409` -> `turnInProgress`.

### 4. The provider stream still cannot be cancelled before the first partial

`handle` is populated only from `PartialResponseContext` / `PartialThinkingContext` / `PartialToolCallContext`, so a Stop pressed before the provider's first partial has nothing to cancel and the upstream request runs on. That is a `langchain4j` 1.18.1 limitation rather than something to fix here, but the description still states that Stop "cancels the provider HTTP stream so further tokens are not generated" without qualification - please soften that line.

### Minor

The backend checklist item "Code compiles and tests pass for the touched modules" is still unchecked. It passes - 24/24 locally on this head - so you can tick it.

@ammeek flagging you again since item 1 falls out of the tool-loop guard you asked for, and the fix is a backend call.

Avirup-001:
@MilosPaunovic

• Dropped the executeTool cancel guard. The batch-loop check still skips later tools; an in-flight catalog.dispatch() still has to return (same cheap-fix tradeoff), but the pair is always closed. resumeHeldAction writes a cancelled TOOL_RESULT before aborting, so a confirmed mutate does not run after Stop. ChatMessageAdaptor.project now drops a TOOL_CALL with no matching TOOL_RESULT (a pending PROPOSED_ACTION still counts as paired, so confirmation resume is intact).

• onStop no longer focuses on nextTick. It sets a flag; we watch canSend and focus once it goes true, after the textarea is re-enabled.

• Stop no longer flips IDLE locally. streaming clears (Stop goes away) but status stays RUNNING, so canSend stays false. We GET the thread and poll until the server reports it is no longer RUNNING, then re-enable Send. reset / loadThread cancel that wait so it cannot overwrite a newer conversation.

• Description softened: Stop cancels the provider HTTP stream only once langchain4j has given us a StreamingHandle (first partial). An early Stop still abandons the turn, but the upstream request may keep running.

Backend checklist item is ticked.

MilosPaunovic (changes requested):
Round 3 on `6f16e369`.

### The red `Frontend - Tests` is not yours

`tests/unit/override/components/flows/Actions.spec.ts` fails 11/11 on the shared-state guard (`2 teleported popper node(s) left behind`). I ran that spec three ways with identical tooling:

| Tree | Result |
|---|---|
| this `PR` head alone (`6f16e369`) | 11/11 pass |
| `origin/develop` alone | 11/11 fail |
| this `PR` head merged with `origin/develop` (what `CI` builds) | 11/11 fail |

Nothing on this branch touches flows or that spec, so it only arrives through the `CI` merge. It is a `develop`-side breakage and will clear on its own; no action needed here.

For the rest, verified locally at this head: `./gradlew :webserver:test` for `AgentOrchestratorTest` + `ChatMessageAdaptorTest` 31/31, the touched `vitest` specs 198/198, `eslint` clean, `node ui/scripts/translations/check-translations.mjs --scope oss` passing.

Two things still open on the new idle wait, plus one non-blocking note.

### 1. `awaitServerIdle` polls without a bound, and survives unmount

`ui/src/components/ai/copilot/useAiChat.ts`

The loop is `for(;;)` with a 100 ms `setTimeout` and no attempt cap or deadline. Its only exits are a generation bump (`reset`/`loadThread`), a thread-uid change, a non-`RUNNING` status, or a 404. Every other failure is swallowed by the `catch` and retried, so a backend restart, a 500, or a dropped connection during a Stop turns into a permanent 10 requests/second poll - and the composer never re-enables, because `status` stays `RUNNING` the whole time.

A thread that is genuinely stuck `RUNNING` behaves the same way: `AiThreadManager` has no scheduled reaper, so nothing ever moves that row back to `IDLE` and the loop runs for the life of the page.

`onBeforeUnmount(cancel)` does not stop it either. `cancel()` only calls `abort?.abort()`, and `abort` is already `null` by the time the wait starts - `runStream`'s `finally` clears it before the `if (waitForServerIdle)` block runs. So navigating away from the copilot mid-Stop leaves the loop polling.

Please bound it: an attempt cap or a deadline, then fall back to setting `IDLE` locally, and make unmount stop it (bumping `idleWaitGeneration` from an `onScopeDispose`/teardown would be enough). A small backoff between attempts would also be kinder than a flat 100 ms - a Stop during a 1.7 s `search_docs` already costs ~17 `GET`s.

### 2. Adopting the server status verbatim can wedge the composer

Same file. `if (data.status !== "RUNNING") { status.value = data.status; return }` will adopt `AWAITING_CONFIRMATION`, but `pendingConfirmation` is still `null` at that point. The result is `canSend` false, no `ProposedActionCard` rendered, and no Stop button - a dead composer whose only exit is New chat.

It is reachable when the abort lands between `markAwaiting` and the `proposed_action` / `done` emits in `suspendForAction`; both emits are dropped on a cancelled `FluxSink`, so the client never learns there is a proposal, while the row says `AWAITING_CONFIRMATION`.

`awaitServerIdle` already has the `ThreadDetail` in hand, and it carries `pendingConfirmationId` and `messages` - so it can rebuild the proposal the way `loadThread` does. Alternatively, only adopt `IDLE` and leave anything else to a reload.

### 3. One poisoning path still slips past the projection guard (non-blocking)

The `PROPOSED_ACTION` carve-out in `pairedToolCallIds` is correct and necessary - `resume` projects at `AgentOrchestrator.java:176`, before `resumeHeldAction` appends the result, so dropping the held call there would hand the provider a result with no preceding call, which is rejected just the same. No change wanted to that.

The residual is that `failTurn` calls `resetToIdleIfExists` without closing the pending call. So any throw inside `resume` before that append - `catalog.byName(held.name()).orElseThrow()` when a tool was removed between the proposal and the confirm, for example - leaves the `thread` `IDLE` with an unpaired `TOOL_CALL` that the carve-out lets through, and every later turn on it fails at the provider.

Narrow, and strictly better than before this `PR`, so I would not block on it. The fix is the same shape as the new cancel path at `:243`: append a failed result for the pending action before resetting to `IDLE`. Happy for it to be a follow-up if you would rather keep this `PR` focused.

Avirup-001:
@MilosPaunovic 

• awaitServerIdle is bounded: 10s deadline then local IDLE, backoff 100ms → 1s, and onScopeDispose bumps the generation so unmount stops the poll.

• AWAITING_CONFIRMATION from that GET no longer sets status alone. It applies the thread detail and rebuilds the proposal the same way loadThread does, so the confirm card is there.

• Item 3 (failTurn leaving an unpaired TOOL_CALL on resume) left as a follow-up, as you suggested.

MilosPaunovic (approved):
Round 4 on `b4fcff22`. Everything from the last round is handled, and I am happy with this.

Verified locally: `./gradlew :webserver:test` for `AgentOrchestratorTest` + `ChatMessageAdaptorTest` 31/31, the copilot and http `vitest` specs 252/252 (including `Actions.spec.ts`, now that https://github.com/kestra-io/kestra/pull/19459 has fixed it on `develop`), `eslint` clean, `node ui/scripts/translations/check-translations.mjs --scope oss` passing. I could not get a meaningful `vue-tsc` signal locally - plain `origin/develop` reports the same `TS7006` family in `kestraHttp.ts` and `main.ts` with my locally built `kestra-sdk` dist, so it is the dist rather than either tree.

One thing to note on the `PR` itself: only `Codespell` has reported on this head. The main `Pull Request Workflow` is still pending a maintainer approving the fork run, so please make sure that goes green before merging - my local runs are a stand-in, not a substitute.

Three small things I am deliberately not blocking on:

- The 10 s deadline in `awaitServerIdle` is only evaluated between requests, and `client.get` carries no timeout or abort signal. A `GET` that hangs open therefore blocks the loop past the deadline and leaves the composer disabled. Passing an `AbortSignal` with a timeout would close it, but it is a narrow case now that every other failure is bounded.
- Once the deadline passes, the frontend reports `IDLE` while the server may still be `RUNNING`, so a prompt sent in that window still comes back `409` / `turnInProgress`. I think that is the right trade - a visible error beats a composer that never re-enables - just flagging that it is the remaining shape of the original race.
- `failTurn` still calls `resetToIdleIfExists` without closing a pending `PROPOSED_ACTION`, so the carve-out in `pairedToolCallIds` lets that one unpaired `TOOL_CALL` through. Same cheap fix as the cancel path, and fine as a follow-up.

I also checked this against the copilot work that landed on `develop` in the meantime (https://github.com/kestra-io/kestra/pull/19330 and https://github.com/kestra-io/kestra/pull/19382): the composer's Stop/Send pair merges cleanly after the mic button, and `applyThreadDetail` rebuilding the transcript does not disturb the diff-preview state, since `dismissedDraftIds` and `appliedDraftIds` are `draftId`-keyed sets held outside `messages`.

On backporting: this wants a `releases/v2.0.x` companion, and `releases/v1.3.x` is not applicable (none of the touched files exist there). Heads up that the 2.0 cherry-pick will not be clean - `CopilotChat.vue` and `CopilotComposer.vue` have diverged between `develop` and 2.0, though `useAiChat.ts`, `kestraHttp.ts`, `client-facade.ts`, `AgentOrchestrator.java` and `ChatMessageAdaptor.java` are still identical on both.

PR #19155, opened 2026-09-09
refactor(core): use KsCheckbox and KsTag instead of local copies

Piyush-r-bhaskar:
The file explorer checkbox was hand-rolled with `input { display: none }`, so it could not be focused or announced. It uses `KsCheckbox` now and the local component is gone.

`Badge` rendered a `KsButton` for a static label, so it sat in the tab order. It renders `KsTag` now, so the four call sites are unchanged.


flcarre (approved):
Clean removal of duplicated design-system components — approving.

- `Badge.vue`: swaps a `KsButton` that was being repurposed as a badge (and patched with `--kel-button-border-color` to fight its internal styling) for the actually-correct `KsTag type="primary" round`. Removes an anti-pattern rather than adding one.
- `FileExplorer.vue` / `layout/Checkbox.vue`: replaces the hand-built "neon checkbox" with `KsCheckbox`, using the same one-way `:modelValue` + `@change` contract the design system already uses internally (see `KsBulkSelect.vue`). No other file referenced the deleted component, so nothing is orphaned.
- No element-plus creep, no hex/rgb, no `:deep()`, no hardcoded px — the surviving style rule uses `var(--ks-spacing-2)`.
- Lint and type-check pass clean on the touched files.

One tiny non-blocking note: `KsCheckbox`'s `change` payload is typed `any` in the design system itself, so `toggleCheckboxSelection(checked: boolean, ...)` isn't actually type-checked against it — pre-existing DS gap, not something to fix in this PR.

Nice cleanup.

PR #19215, opened 2026-09-09
fix(flows): exclude flow inputs and triggers from task-property autoc…

### LINKED ISSUE #18749: fix(flows): task-property autocomplete scoping also matches flow inputs and triggers
Follow-up to https://github.com/kestra-io/kestra/pull/18639, which scoped task-property autocomplete to the resolved task type. Two loose ends were deliberately left out of that PR.

### `isTaskLike` matches flow inputs and outputs

`findTaskLikeAtCursor` identifies a task by the presence of a string `id` and a string `type`. Flow `inputs:` and `outputs:` entries have exactly that shape:

```yaml
inputs:
  - id: myinput
    type: STRING
```

So placing the cursor inside an input body resolves the "task type" as `STRING`, and the completion provider then asks the backend for a plugin whose class name is `STRING`. That 404s (verified: `GET /api/v1/plugins/STRING` → 404). Nothing breaks, because the lookup fails open and the unscoped suggestions are returned, but it is a wasted request on every property-key completion while editing inputs or outputs.

Worth scoping the ancestry walk to a `tasks` / `errors` / `finally` context, or otherwise rejecting candidates whose `type` is not a plugin FQCN.

### Trigger property scoping is untested

`triggers:` entries also carry `{id, type}` with a real plugin FQCN, so the same code path scopes trigger property completion. That works (with `all: true`, `io.kestra.plugin.core.trigger.Schedule` reports 20 property keys rather than 9), but it was neither intended nor covered by tests in the original PR, and `AbstractTrigger` has its own set of inherited properties.

Either cover it deliberately with tests, or exclude triggers from the scoping until it is.

hemantmm:
…omplete scoping

closes: #18749

### ✨ Description

Updated the `isTaskLike` function to strictly scope valid candidates. It now requires the `type` property to include a dot (`.`), which guarantees that only plugin FQCNs are matched (safely ignoring uppercase types like `STRING` from inputs). Additionally, it explicitly excludes any candidate whose type contains `.trigger.`, safely falling back to unscoped suggestions for triggers.


flcarre (changes requested):
Thanks for tackling #18749 — the `inputs`/`outputs` half of the fix (requiring `type` to contain a `.`) looks correct. The trigger-exclusion half needs another pass before this can merge.

**1. The `.trigger.` substring check doesn't exclude real triggers (blocker)**

`isTaskLike` (taskCompletionScoping.ts:14) excludes a candidate only if its `type` contains the literal substring `.trigger.`. That's true for Kestra-core's own trigger classes (`io.kestra.plugin.core.trigger.Schedule`, `.Flow`, `.McpToolTrigger`, ...), because those happen to live in a package named `trigger`. It is **not** true for the vast majority of real plugin triggers, named `<provider>.<resource>.Trigger` with no `trigger` package segment — e.g. `io.kestra.plugin.aws.s3.Trigger`, `io.kestra.plugin.fs.local.Trigger`, even core's own `io.kestra.plugin.core.http.Trigger`. For all of these, task-property autocomplete stays scoped as if they were tasks. This needs a more robust signal than a package-name substring — e.g. resolving whether the type is a trigger via the plugin schema lookup already available (`pluginsStore`), rather than string-matching the FQCN.

**2. Regression in the "insert missing required properties" snippet (major)**

`isTaskLike` is shared with `findTaskLikeAtCursor` → `filterMissingRequiredTaskProperties`, an unrelated completion feature. For the core triggers the new exclusion does catch, that feature now returns the full `requiredProperties` list unfiltered — it will re-suggest properties already set on the trigger. Consider splitting the two predicates so this fix doesn't leak into an unrelated feature.

**3. Test coverage doesn't catch either issue above (major)**

Please add a case using a non-core-style trigger FQCN (e.g. `io.kestra.plugin.aws.s3.Trigger`) to prove the exclusion generalizes, and a case confirming `filterMissingRequiredTaskProperties`/`findTaskLikeAtCursor` still work for trigger bodies.

**4. Minor: outputs case untested** — the linked issue also calls out `outputs:` entries; worth a test for completeness.

**5. Nit: unrelated formatting churn** — most of the diff is a reflow unrelated to the fix; consider keeping it out of a bug-fix PR.

hemantmm:
Thanks for the detailed feedback @flcarre. I’ve addressed all the points:

* Removed the `.trigger.` string check and now use `pluginsStore.findPluginByCls(type)` to determine whether the resolved plugin is a trigger.
* Kept `isTaskLike` trigger-agnostic so `filterMissingRequiredTaskProperties` continues to work correctly for trigger bodies.
* Added regression coverage for a non-core trigger (`io.kestra.plugin.aws.s3.Trigger`) and verified trigger required-property filtering still works.
* Added coverage for both flow inputs and outputs with non-FQCN types such as `STRING`.
* Reverted the unrelated formatting changes in `isTaskLike`.

flcarre (changes requested):
Thanks for the rework — two of the five points are properly fixed, but the trigger half still does not work, for a new reason.

### Resolved

- **#2 (regression in the required-properties snippet)** — fixed correctly. `isTrigger` is applied only in `taskIdentityAtCursor`; `findTaskLikeAtCursor` is untouched, and `taskCompletionScoping.spec.ts:144` pins it.
- **#4 (outputs untested)** — covered at `taskCompletionScoping.spec.ts:112-123`.

### Still blocking

**1. The new `isTrigger` predicate is a no-op in the normal editing session**

`yamlLanguageConfigurator.ts:356-360` resolves triggerness through `pluginsStore.findPluginByCls(type)`. That function's first line is `if (!cls || !plugins.value) return null` — and `plugins.value` is only ever assigned by `list()` (`stores/plugins.ts:322`) or `listWithSubgroup()` (`:362`). On the Monaco completion path neither is called: repo-wide, `ensurePlugins()` comes from the no-code task picker (and only lazily, when the picker opens — `useTaskPicker.ts:149`), the plugin catalog pages, and `useFlowRecipe`; `list()` fires only from `PluginInstallToast` and after a successful plugin auto-install.

So: open a flow, stay in Source view, type inside a `triggers:` body -> `plugins.value` is `undefined` -> `findPluginByCls` returns `null` -> `isTrigger` returns `false` -> the exclusion never runs and the trigger is scoped as a task, exactly as before this PR. It only works if you happened to open the no-code picker or a plugin page first. We traded a naming-dependent bug for a race-dependent one, and both fail silently.

The surrounding provider is already `async`, so a minimal fix is to `await pluginsStore.ensurePlugins()` before calling `taskIdentityAtCursor` — but see point 3, which I think is the better answer.

**2. Test coverage does not exercise the predicate**

`taskCompletionScoping.spec.ts:126-140` passes its own `isTrigger: (type) => type === "io.kestra.plugin.aws.s3.Trigger"`. That asserts `taskIdentityAtCursor` calls the callback it was handed — which was never the risk. All the risk lives in the real predicate, and that has no test at all. My earlier ask was for a case proving the exclusion *generalizes*; a stub that hardcodes the answer cannot prove that, and it is precisely why the no-op above shipped green.

**3. Use the YAML path, not the type string — and it dissolves 1 and 2**

We are still asking "is this class registered as a trigger?" when the question is "is the cursor inside `triggers:`?", and the answer is already in hand. `localizeElementAtIndex` returns `parents` filtered to maps, so `parents[0]` is the root flow document. Checking whether the resolved node is reachable through the root's `triggers` (or `inputs`/`outputs`) rather than `tasks`/`errors`/`finally` is deterministic, synchronous, needs no store and no network, and is unit-testable from a YAML string with no stubbing.

That single signal replaces *both* halves of this PR: it supersedes the `type.includes(".")` proxy as well as the registry lookup. The linked issue offered it as the first option for the inputs/outputs half; it's the right one for triggers too.

**4. `findPluginByCls`'s subGroup shortcut is order-dependent**

Even with the store populated, it starts with `plugins.value.find(p => p.subGroup && cls.startsWith(p.subGroup + "."))` — first prefix match, not longest. Since `ensurePlugins()` populates via `listWithSubgroup()`, a broad subgroup entry can shadow a specific one and return a plugin whose `triggers` array does not contain the class. Pre-existing behaviour, but this PR newly depends on it for correctness.

### Non-blocking

- **`isTaskLike`'s dot gate leaks into the inputs path** (`taskCompletionScoping.ts:11`). Inside an `inputs:` body `findTaskLikeAtCursor` now returns `undefined`, so `filterMissingRequiredTaskProperties` hits `:199` and returns the *full* unfiltered required list — the same class of leak as my earlier point 2, moved from triggers to inputs. Latent today because the snippet path bails earlier, but it shouldn't depend on that.
- **Edge cases still uncovered**: worth adding ForEach/Parallel, Dag's `tasks: - task:` shape, `errors:`/`finally:`, a trigger carrying a nested `conditions:` sub-map, and a genuinely unparseable document to pin the `catch` path. The half-typed-key cases are well covered.
- **#5 (formatting churn) is only half reverted.** `isTaskLike` was restored, but the rest is still an 80-column reflow — it is what makes this +292/-112 for roughly fifteen lines of behaviour. It passes lint (I ran eslint against the head: clean), but `{ type: string; version?: string }` reads against the repo's `object-curly-spacing: never` — the base rule just doesn't cover TS type literals. Please reflow back so the behavioural change is the diff.
- **Nit**: `as {cls: string}[]` at `:358` re-declares `PluginElement`, already exported from `ui/src/utils/pluginUtils.ts`.

My suggestion: drop the `isTrigger` callback and the registry lookup entirely, decide task-vs-trigger-vs-input from the root map in `parents`, and test that against YAML fixtures. It is less code than what's here and it removes the store dependency that points 1, 2 and 4 all hang off.

flcarre:
Picking this back up — you pushed after my `CHANGES_REQUESTED` and GitHub didn't re-request me, so it dropped off my list. Re-reviewing the autocomplete scoping now.

hemantmm:
Thanks for the detailed feedback @flcarre! 

* Replaced the plugin registry lookup with deterministic YAML AST path checking using the root map and recursive ID-based reachability.
* Removed the `isTrigger` callback, `pluginsStore` dependency, and `type.includes(".")` check.
* Restored formatting in `taskCompletionScoping.ts` to minimize unrelated diff changes.
* Added comprehensive tests covering:

  * `triggers`, `inputs`, and `outputs` exclusion.
  * Tasks inside `errors`, `finally`, and nested DAG structures.
  * Nested trigger property maps.
  * Unparseable YAML documents.
  * Required-property filtering for trigger and input bodies.

flcarre (commented):
Thanks for the rework — this is a real step forward. You dropped the registry lookup and decided it from the parsed document instead, which is what I asked for, and the new tests use actual YAML fixtures rather than a stubbed predicate. One correctness gap left, plus a guard I'd rather you kept.

### Resolved from my last review

- **#1 (the `isTrigger` predicate was a no-op)** — gone. `taskIdentityAtCursor` no longer touches `pluginsStore`; it is synchronous, needs no network, and works on the first keystroke of a fresh session. This was the blocker and it is properly fixed.
- **#2 (tests stubbed the predicate)** — fixed. `taskCompletionScoping.spec.ts:89-124` drives the real code from YAML strings with no injected callback.
- **#4 (`findPluginByCls` subGroup ordering)** — moot, the dependency is gone.
- **Non-blocking: `isTaskLike`'s dot gate leaking into the inputs path** — fixed. `isTaskLike` is back to `id` + `type`, `findTaskLikeAtCursor` is untouched, and `taskCompletionScoping.spec.ts:169-198` pins `filterMissingRequiredTaskProperties` for both a trigger body and an input body.
- **Non-blocking: edge cases** — Dag's `tasks: - task:` shape, `errors:`, `finally:`, a trigger carrying a nested `conditions:` sub-map and an unparseable document are all covered now. ForEach/Parallel are structurally identical to the existing Sequential case, so I won't ask for those.
- **#5 (formatting churn)** — reverted. +141/-9 across three files is the behavioural change and nothing else.

### What still needs a fix

**1. The check is by `id`, not by path (`taskCompletionScoping.ts:130-139`)**

`isReachableById(root[section], task.id)` deep-searches for a node whose `id` equals the task's. That is an id-name proxy for the path, and the two diverge: `FlowValidator` checks duplicate ids *within* `tasks` (`FlowValidator.java:62`), *within* `inputs` (`:87`) and *within* `outputs` (`:114`), never across sections. So a flow with

```yaml
inputs:
  - id: data
    type: STRING
tasks:
  - id: data
```

is valid, and the `data` task's property completion silently stops being scoped — back to the union-of-every-plugin list this PR exists to remove. Same for an output named after the task that produces it, and for any nested `id` under a trigger (`conditions[].id`, `preconditions.id`) colliding with a task id. It fails open rather than hiding suggestions, so it's a silent degradation rather than a break, but it's the exact class of naming-dependent bug the first round already hit.

Worth knowing before you reach for object identity as the fix: it won't work. `localizeElementAtIndex` does `parents.filter(isMap).map((p) => p.toJS(yamlDoc))`, so every ancestor is an independent `toJS()` copy and `root.triggers[0] !== task`.

The deterministic version is to surface the ancestry *keys*. The `visit` callback inside `localizeElementAtIndex` already receives the `Pair` ancestors, so the key path is in hand there — returning it alongside `parents` lets this become `path[0] === "triggers" | "inputs" | "outputs"`, which no id can collide with and which is trivially testable from a YAML string.

**2. Don't drop the dotted-type gate (`yamlLanguageConfigurator.ts:337`)**

The removed `task.type.includes(".")` was orthogonal to the section check, not redundant with it. The flow root has other `id` + `type` lists that are task-shaped but carry enum types — `sla:` is the live one (`SLA.java:39-45`: `id`, `type: MAX_DURATION | EXECUTION_ASSERTION`, `behavior`). Those sit outside `inputs`/`outputs`/`triggers`, so the new check doesn't catch them, and typing inside an `sla:` entry now issues `pluginsStore.load({cls: "MAX_DURATION", all: true})`. The rejection is cached in `unscopableTaskTypes`, so it's one wasted 404 per type per session rather than per keystroke — but it's a request that was impossible before, and the comment you deleted is what documented why. Please keep both guards.

### Minor

- `taskCompletionScoping.ts:25` — the `Array.isArray` branch is unreachable: arrays already satisfy the `typeof container === "object"` test above it. It only works because `Object.values(array)` happens to yield the elements. Drop it, or hoist it above the object branch.
- `taskCompletionScoping.ts:129` — trailing whitespace on the blank line. The repo's eslint config has no `no-trailing-spaces` rule, so CI won't tell you.
- Once the path check is in, add a fixture where an input and a task share an id and assert the task still resolves — that's the one case the current predicate gets wrong.

### Confirmed fine

`{{ inputs.foo }}` completion inside a task property value is unaffected: the scoping block is gated on `!isTypeValueContext` and filters only `CompletionItemKind.Property`, while pebble suggestions come from the separate value provider and `registerPebbleAutocompletion`. And when the exclusion fires the code returns `undefined`, which fails open — nothing is ever removed from the list by the new check.

Design-system pass: nothing to report. No `.vue`, no styles, no user-facing strings, no element-plus, no DS components touched.

hemantmm:
Thanks for the detailed review! I've addressed all the remaining points:

1. **Path-based check:** I surfaced the ancestry keys by filtering the `parents` stack for `isPair` and extracting the `scalarKey` during the `visit` callback in `localizeElementAtIndex`. This gave me the exact `path` array, allowing me to completely delete the `isReachableById` function and its unreachable `Array.isArray` branch. The check in `taskIdentityAtCursor` is now a deterministic `path[0] === "triggers" | "inputs" | "outputs"`.
2. **Dotted-type gate:** Restored the `task?.type.includes(".")` gate in `yamlLanguageConfigurator.ts` so we don't fire 404s for enum types like `MAX_DURATION` under `sla:`.
3. **Tests & Formatting:** Fixed the trailing whitespace and added a new test fixture in `taskCompletionScoping.spec.ts` asserting that scoping still works perfectly when a flow input and a task share the same `id`.

flcarre (approved):
Both points from my last review are properly fixed, and I checked this one by running it rather than reading it.

### Resolved

- **#1 (the check was by `id`, not by path)** — done the way I suggested. `localizeElementAtIndex` now surfaces the `Pair` key ancestry as `path` (`positions.ts:176-179`), `taskIdentityAtCursor` decides on `path[0]` (`taskCompletionScoping.ts:117-123`), and `isReachableById` is gone. I built a worktree at `8ce1cf1` and probed it against the PR base (`da7198d`) to isolate the delta: on base, a trigger body resolves `io.kestra.plugin.aws.s3.Trigger`; on head it returns `undefined` — for a half-typed key, on a blank line inside the trigger, on a completed `key: value`, and behind a leading `---`. Same for `inputs:`/`outputs:`. And nothing else moved: `tasks:`, `errors:`, `finally:`, `afterExecution:`, the Dag `tasks: - task:` shape and nested flowables (`Sequential`, `ForEach`) all still resolve their type. The id-collision case your fixture pins (`:92-103`) behaves correctly. 22/22 in the spec, 104/104 in the topology yaml-util tests, 9/9 in `flowAutoCompletionProvider.spec.ts`, eslint clean on all four files.
- **#2 (the dotted-type gate)** — restored at `yamlLanguageConfigurator.ts:341-343`, with the comment explaining why. It is genuinely load-bearing, not belt-and-braces: I confirmed `taskIdentityAtCursor` still returns `{type: "MAX_DURATION"}` for an `sla:` body, so the path check alone would not have caught it. Good call keeping both.
- **Minors** — `isReachableById` is deleted so the unreachable `Array.isArray` branch is moot; no added line in the diff carries trailing whitespace; the shared-id fixture is in. Still +156/-14 with no formatting churn.

Nothing blocking left. The rest is optional.

### Worth adding (non-blocking)

**1. Pin the `sla:` case.** The dotted-type gate is now the only thing keeping `sla:` out of scoping, and nothing in the suite says so — which is how it got dropped last round. One line in `taskCompletionScoping.spec.ts` would document the contract:

```ts
it("resolves sla entries as a bare enum type, which the dotted-FQCN guard then rejects", () => {
    const flow = `id: myflow\nnamespace: my.ns\nsla:\n  - id: mysla\n    type: MAX_DURATION\n    beha`
    expect(taskTypeAtCursor({source: flow, cursorIndex: flow.length})).toBe("MAX_DURATION")
})
```

**2. Pin that the exclusion is anchored at the root.** `inputs` is also a *task* property — `io.kestra.plugin.core.flow.Subflow` has one. Your implementation reads `path[0]` so it is correct (I checked: `tasks: - id: sub / type: …Subflow / inpu` still resolves `Subflow`, and a key typed after a nested `inputs:` map is unaffected), but nothing pins it, and a later refactor to `path.includes("inputs")` would pass every test in the file. A Subflow fixture asserting the task still resolves closes that.

### Minor

- `taskCompletionScoping.ts:58` — `probeTaskOwningCursor` returns `{task, candidates, path}` but `taskIdentityAtCursor` only destructures `{task, path}`. `candidates` is dead; drop it from the shape.
- `taskCompletionScoping.ts:119` — the `["triggers", "inputs", "outputs"]` literal is built on every keystroke and is invisible to anyone adding a root section later. A module-level `const EXCLUDED_ROOT_SECTIONS` makes it the obvious edit point. (For what it's worth, I checked `Flow.java`: `afterExecution` is the other root task list and your deny-list handles it correctly, `checks` and `pluginDefaults` carry no `id` so they never reach here.)
- `positions.ts:122` — `path` on a public type reads like a file path; `keyPath` says what it is. And `path: path` can be shorthand.
- `positions.ts:8` — `isNode, isPair` on one line breaks the one-per-line import list around it.

Design-system pass: nothing to report. No `.vue`, no styles, no user-facing strings, no element-plus, no DS components touched.

Thanks for sticking with this through four rounds — the final shape is smaller and more deterministic than where it started, and it is the right one.

PR #19216, opened 2026-09-09
refactor(core): delete the components and utils nothing imports

### LINKED ISSUE #19218: Some UI components and utils are not imported by anything
Thirteen files under src in the two UI repos are not imported by anything: eight in OSS (a tag component, a demo button set, two input components, two no-code field components, an unsaved-changes composable and a query builder util) and five in EE (two composables, a model, an api-token message component and an apps onboarding component).

They cost review time and search time rather than bytes. One of them was edited in a recent PR, so a reviewer read a diff against a file nothing renders, and they answer greps first, so the next person looking for the unsaved-changes dialog finds the dead composable before the one the design system actually ships.

Deleting them needs care because a file can look dead and not be: several similar components are pulled in by an import.meta.glob rather than an import, one has a live namesake in the other repo, and one still shows up in stale build output from before the refactor that orphaned it.

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra/issues/19218.
EE PR: https://github.com/kestra-io/kestra-ee/pull/10908


MilosPaunovic (approved):
Approving.

Verified all eight deletions are genuinely unreferenced, and that the removal is safe across both repos:

- Unrestricted grep for every deleted name across `OSS` and `EE`, all file types, including `e2e-tests` and `ui/packages/*` - zero hits. No `EE` file reaches these through the `kestra` alias.
- Not glob-loaded: the only component glob is `getTaskComponent.ts`'s `./Task*.vue`, and `no-code/components/inputs/` keeps just `InputText.vue` afterwards.
- The exported symbols are dead too, not only the paths - `split`, `toLucene`, `toTextLucene` and `iso` from `queryBuilder.ts` are unused everywhere.
- `vue-tsc --build tsconfig.app.json --force` and `vue-tsc --project tsconfig.test.json` are both clean locally on this branch.
- `Secrets.vue` is the right loose end to pick up: only the orphan `.ee-tag-wrap` / `:deep(.enterprise-tag)` rule goes, and no template in either repo still carries that class.
- Each file is long dead rather than unwired work in flight. `queryBuilder.ts` lost its last consumer in 2022 when Lucene left the `UI`, `InputSwitch` mid-2025, `InputPair` in October 2025, and `useUnsavedChangesDialog` was superseded within days of landing. `AcceptDecline`, `SaveExecuteAnimation`, `EnterpriseTag` and `DemoButtons` died in the Copilot v2, onboarding-tour and secrets-page changes.
- `InputPair.vue` picking up the `rowKey` fix in https://github.com/kestra-io/kestra/pull/19154 a day ago is not a problem: that sweep also covered the live `TaskDict.vue`, `LabelInput.vue` and `SecretsTable.vue`, so no fix is lost, and `TaskDict.vue` is the live pair editor that superseded `InputPair`.

One non-blocking follow-up, and one warning about backporting.

**Orphaned translation keys.** `AcceptDecline.vue` was the only consumer of `draft_available`, `accept` and `reject`, so those three keys are now used nowhere in either repo while still shipping in `en.json` and every generated locale. `check-translations.mjs` only enforces used -> defined, so nothing will ever flag them and future translation runs keep carrying them. `draft_available` is unambiguous; `accept` and `reject` are generic single words, so they warrant the same caution the earlier ghost-key sweep applied to `dynamic` and `resource`, since a backend-emitted string rendered through `t(e.message)` could reach them. The other keys these files touched (`demos.get_a_demo_button`, `duplicate-pair`, `key`, `value`) all still have live consumers and must stay. Happy either way on folding this in here or filing it separately.

**Do not cherry-pick this to `releases/v1.3.x`.** The files are dead on `develop` and on `releases/v2.0.x`, but on `releases/v1.3.x` `EnterpriseTag` (2 files), `DemoButtons` (2), `AcceptDecline` (3) and `SaveExecuteAnimation` (1) are still live, because the commits that orphaned them landed after 1.3 branched. A blind backport breaks the 1.3 `UI` build.

PR #19217, opened 2026-09-09
fix(design-system): add spacing between file tags

### LINKED ISSUE #19179: [Executions] File tags in the Input/Output tree have no spacing and overlap each other
### Describe the issue

On an execution's Input/Output tab, a task output that holds files renders each file as a tag. When a task returns several files, the tags on consecutive rows sit too close together. Each tag's background touches the one above and below it, so they read as one stack rather than separate rows.

The plain text rows in the same tree, such as `exitCode` and `containerId`, keep normal line spacing. Only the rows carrying a file tag lose it.

Input/Output tree with overlapping file tags

1. Go to Flows and create the flow below.
2. Click Execute and wait for it to finish.
3. Open the execution and click the Input/Output tab.
4. Expand `outputFiles`. The five file tags touch each other with no gap between the rows.

```yaml
id: shell_commands_output_files
namespace: qa.outputs

tasks:
  - id: run_metadata_extraction
    outputFiles:
      - abc
      - report.csv
      - payload.json
      - notes.txt
      - archive.zip
    taskRunner:
    containerImage: alpine
    commands:
      - echo "test" > abc
      - printf 'a,b\n1,2\n' > report.csv
      - printf '{"k":"v"}\n' > payload.json
      - echo "plain text notes" > notes.txt
      - echo "not really a zip" > archive.zip
```

### Environment

Kestra Version: 2.1.0-SNAPSHOT (develop, commit 42056cb)
Kestra Edition: OSS

BharadwajKanneveti:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/19179.

### ✨ Description

- Adds vertical spacing between consecutive file tags in the execution Input/Output tree.
- Previously, when a task returned multiple files, the file tags were rendered directly against each other, causing their backgrounds to touch and making them appear as a single stack. File tags now have spacing between consecutive rows while leaving other tags unchanged.


flcarre (commented):
This adds vertical spacing between stacked file tags — good fix for the underlying bug (#19179). One thing worth addressing before merge since this touches the design-system source of truth itself:

**`ui/packages/design-system/src/components/Data/KsTag/KsTag.vue`**
```scss
&.kel-file-tag {
    margin-bottom: 8px;
}
```
- `8px` is a raw pixel value in the DS source file. `--ks-spacing-2` already equals `0.5rem` (8px) in `ks-tokens.scss` — please use `margin-bottom: var(--ks-spacing-2);` instead.
- Consider whether this spacing belongs here at all: baking a `margin-bottom` into `KsTag` keyed off a `KsFileTag`-specific class couples the shared leaf component to one consumer's vertical-stack layout, and adds unwanted trailing space if `KsFileTag` is ever used inline or as the last item in a list. A `gap: var(--ks-spacing-2)` on whichever container renders the list of file tags would be more composable.
- Minor: the new class doesn't follow this file's own `kel-tag--*` modifier naming convention, and there's some stray indentation/blank-line inconsistency around lines 160-164.

Not blocking, but please swap the literal `8px` for the token before merging a design-system change.

flcarre (approved):
Thanks for moving this — `ecf6a1c6` addresses everything I raised, and cleanly.

- The raw `8px` is now `var(--ks-spacing-2)`.
- The spacing is out of `KsTag.vue` entirely and lives in the consuming layout, so the shared leaf component is no longer coupled to one consumer's vertical stack.
- `&--file + &--file` is a better answer than the container `gap` I suggested: it targets only the second-and-subsequent file rows, so there is structurally no trailing space on the last tag, and non-file rows stay tight.
- `json-tree__row--file` matches this file's own `__row--selected` convention.

I checked the things that could have bitten this:

- **Adjacency is sound.** The rows are direct siblings, and the `rootLiteral` row is mutually exclusive with the `v-for` rows (`v-if="!rows.length"`) and never carries `--file`, so the `+` selector can't misfire.
- **No margin compounding.** `.json-tree` has no `gap` and preceding rows carry no `margin-bottom`.
- **No regression in the other consumer.** `LogLine.vue` also renders `KsJsonTree` for structured logs; consecutive file URIs there pick up the same spacing, which is the same bug being fixed rather than a regression.

On the missing test — I don't think one is needed. `KsJsonTree.stories.tsx` already has a story with four consecutive file URIs, which is exactly the case this fixes. A scoped-SCSS `margin-top` isn't meaningfully assertable in jsdom regardless.

Two optional nits, neither blocking:

1. The gap sits *outside* the row's background box, so the `:hover` highlight blinks off as the pointer crosses the strip between two file rows. `padding-block` on the `--file` row would keep the hover band contiguous.
2. That multi-file story's play function asserts only the icons — one assertion that consecutive rows carry `json-tree__row--file` would lock in the class contract this fix depends on.

No conflict with #19204, which touches `KsFileTag.vue`/`KsTooltip.vue` — different layer, zero file overlap, merges in either order.

LGTM.

PR #19221, opened 2026-09-09
refactor(topology): flowYamlUtils typing and module split

### LINKED ISSUE #19268: flowYamlUtils.ts: replace explicit `any` with real types
60 explicit `any` in `packages/topology/src/utils/flowYamlUtils.ts`, from `ui/scripts/explicit-any/baseline.json`.

Already in progress in https://github.com/kestra-io/kestra/pull/19221, which brings this file from 54 lines with `any` down to 15. Pick up whatever is left after it merges rather than starting over.

| any | file |
| --- | --- |
| 60 | `packages/topology/src/utils/flowYamlUtils.ts` |

elevatebart:
Companion PR: kestra-io/kestra-ee#10922. Same branch name on both, so CI pairs them. **That one merges first** — see Merge order at the bottom.

### 🔗 Related Issue

None — this started as a question about the number of `any` in `ui/packages/topology/src/utils/flowYamlUtils.ts`.

### ✨ Description

Three commits, each self-contained. Worth reviewing in order rather than as one diff, since the third is almost entirely code movement.

**1. `refactor(topology): typed yaml document access` (b58e913)**

54 `any`, of which nine caused about forty. Each of `updateMetadata`, `getMetadata`, `deleteMetadata`, `extractFieldFromMaps`, `extractTypedBlocksWithMeta`, `localizeElementAtIndex`, `getAllCharts`, `getChartAtPosition` and `getTasksLines` opened with `parseDocument(source) as any`, so every access after that line was unchecked and the callbacks below were annotated `any` to match. They now use the `parseDocumentTyped` helper already in the file, with `yaml`'s own guards. The `any` on the `visit` callbacks was actively worse than no annotation: `yaml` types `Map`, `Pair` and `Seq` visitors itself, so `Pair(_, pair: any)` discarded types that were already correct.

Two bugs it was hiding, both reproduced against the parent commit before being touched:

- `updateMetadata` threw `TypeError: Cannot create property 'spaceBefore' on string` for any flow with a comment before a non-first list item — `commentBefore` is a `string` and the code treated it as a node, which throws in ESM strict mode. Reachable from `stores/flow.ts:696` and `:738`. Setting `spaceBefore` on the item already puts the blank line before that item's comment, so the branch was broken *and* redundant. Regression test included.
- `deleteMetadata(source: any, ...)` was called as `deleteMetadata(flow.value?.source, "deleted")`, so a missing flow reached `parseDocument(undefined)`. Guarded at the call site. The same call exists in EE, which is what kestra-io/kestra-ee#10922 fixes.

Also: `YAMLMap<{value: string}, Node>` → `Scalar<string>` at 16 sites (the structural stand-in satisfied the compiler while discarding range, comment and quoting info, which is why `getIn` and `items.find` results kept needing casts); a callback parameter on `extractBlockWithPathFromDocument` deleted because nothing passed one; `localizeElementAtIndex` corrected to return `YamlElement | undefined`, which all four callers already optional-chained.

**2. `fix(topology): insert path typed its target sequence as a mapping` (def0b75)**

`insertBlockWithPath` and helpers declared the node at `parentPath` as a `YAMLMap`, but an insert target is always a sequence (`tasks`, `errors`, a flowable's `then`). The two remaining `as any` in that path existed only to push a node past the wrong type. Confirmed against the running library: `getIn` returns a `YAMLSeq`, `createNode` returns a real node. So `getNodeIndexInParent` now takes the sequence it receives, `getParentNode` returns the map-or-sequence union it returns, and its two branches merge since they computed the same parent path by different routes.

Output is byte-identical for every insert shape I could construct: into an existing sequence with and without a `refPath`, before and after, a missing top-level key, a nested path under an existing task, a quoted segment. Inserting at a path holding a mapping now throws a message naming the path — it already failed there with `yaml`'s internal "Map items must all be pairs", this just says which path and why.

**3. `refactor(topology): split flowYamlUtils into one module per concern` (2f21a1d)**

The file did three unrelated jobs: plain-object serialization through js-yaml, mutation of the yaml AST, and read-only cursor queries for the editor. Now seven modules under `src/utils/yaml/`:

| module | lines | holds |
|---|---|---|
| `document.ts` | 27 | `parseDocumentTyped`, `scalarKey`, `TOSTRING_OPTIONS` |
| `paths.ts` | 73 | dotted paths ↔ segment arrays, no yaml involved |
| `serialization.ts` | 123 | `parse`, `stringify`, `pairsToMap` |
| `fields.ts` | 103 | every map with a given field, every typed block |
| `metadata.ts` | 153 | the flow's root keys |
| `blocks.ts` | 453 | extract, insert, replace, delete, swap |
| `positions.ts` | 301 | what sits at a cursor or on a line |

`flowYamlUtils.ts` is an 11-line barrel, so the `@kestra-io/topology/flow-yaml-utils` subpath and every `import * as` consumer are untouched. Verified rather than assumed: the runtime export surface is identical — all 29 names, diffed against the parent commit. Dependencies run one way, no cycles.

The 1632-line test file is split the same way into six files, each importing the module it exercises rather than the barrel, so a test fails loudly if the export it covers moves.

**4. `fix(topology): insert parent resolution on a non-collection node` (dab090f)**

Review round. The `isCollection` guard in commit 2 changed behaviour for a parent path that exists but is not a collection: inserting at `foo.tasks` against `foo: bar` used to throw a TypeError downstream, and instead fell through to the create branch and emitted `foo` twice. Now throws, naming the path and what it holds; `undefined` and `null` still take the create path. Also unified the created parent's key, which came from the full path for quoted segments and the parent path for unquoted ones, and removed the last cast `scalarKey` was meant to replace. Both new `blocks` tests fail on 2f21a1d.

### 🎨 Frontend Checklist

- [x] Type checking passes (`npm run check:types`) — clean across the whole `ui/` workspace
- [x] Unit tests pass (`npm run test:unit`) — 139 in topology across 12 files (was 133 in one), plus 248 affected tests in `ui/`
- [ ] Screenshots or video recordings attached showing the `UI` changes — nothing user-visible changes beyond `updateMetadata` no longer throwing, so there is nothing to screenshot

### 📝 Additional Notes

**Why one PR.** The commits would ideally be separate PRs; this session had a single designated branch, so they are stacked here. Commit boundaries are clean if you would rather they were split.

**`any` count: 54 → 13.** What is left is deliberate. Eleven sit on the plain-object boundary (`parse`, `stringify`, `transform`, and the `Record<string, any>` public returns) where the shape genuinely is unknown and tightening reaches into `ui/src/`. One is a lie made visible rather than silent, with a two-line comment at the cast: `swapBlocks` passes a callback returning a `Document`, which lands in a sequence slot where a `Node` belongs, and works only because `Document` happens to stringify there.

Three of those thirteen are not boundary cases and would come out without needing an exemption, if someone wants them: the private `sort(Record<string, any>)` only reads `Object.keys`, `pairsToMap(pairs?: any[])` only adds its items to a map, and `load(item) as any` should be `as T` against the function's own type parameter. Left out to keep the scope.

**Pre-existing bugs found and deliberately not fixed**, since all three are functional rather than typing:

- `insertBlockWithPath` with a path like `tasks[0].then` against a document that has no `tasks` emits a key literally named `tasks[0]`, because `lastSegmentKey` splits on `.` and leaves the index in the key.
- A parent key that is present but null gets duplicated rather than filled, so an empty `tasks:` yields two `tasks:` keys. Same defect family as the one found in review, in its null form.
- Root keys are ordered by two different lists that disagree: `SORT_FIELDS` in `serialization.ts` (applied by `replaceBlockWithPath`) and `ORDERED_FLOW_ROOT_KEYS` in `metadata.ts` (applied by `updateMetadata`).

**On enforcement:** `@typescript-eslint/no-explicit-any` is already configured in `ui/.oxlintrc.json` and deliberately switched off there, so this PR does not touch it. (An earlier revision of this description claimed the rule was absent — that was wrong; `ui/` lints with oxlint first and eslint second, and the rule lives in the oxlint config.)

### 🔀 Merge order

kestra-io/kestra-ee#10922 merges **first**. That guard compiles against OSS `develop` as it stands today, where `deleteMetadata` still takes `any`, so landing it is a no-op in behaviour and leaves EE green. This PR then tightens the signature into an EE that already guards.

The reverse order is the one to avoid: it puts the tightened signature on `develop` while EE still passes `string | undefined`, so EE is red until the companion lands. `trigger-ee` on this PR went green without catching that, so CI will not stop it.

### 🤖 AI Authors

I asked our office cat to review the `any` types. She sat on the keyboard and typed `unknown`, which was more helpful than the original code. She then split the file into seven by walking across it.

https://claude.ai/code/session_01SHfMvHS2ZAHU8uc5xyCBYF

Relates to https://github.com/kestra-io/kestra/issues/19268 and #19286.


ui/packages/topology/src/utils/yaml/blocks.ts
```diff
@@ -0,0 +1,453 @@
...
+        : parentPath.substring(0, parentPath.lastIndexOf("."))
+
+    if (parentPathWithoutKey === "") {
+        if (!yamlDoc.contents) {
+            throw new Error(`Document is empty, cannot insert block with path ${parentPath}`)
+        }
+        return yamlDoc.contents
+    }
+
+    const parentNode = yamlDoc.getIn(parsePath(parentPathWithoutKey))
```
Piyush-r-bhaskar, line 312: `isCollection` no longer returns for a node that exists but isn't a collection, so it falls into the create-parent branch below. Inserting at `foo.tasks` against `foo: bar` used to throw a TypeError and now emits `foo` twice.

Throw here with `nodeKind`, keeping `undefined` and `null` on the create-parent path since an empty `tasks:` comes back as `null`. Test next to `blocks.test.ts:245`:

```js
expect(() => insertBlockWithPath({
    source: "id: t\nnamespace: n\nfoo: bar\n",
    parentPath: "foo.tasks",
    newBlock: "id: new\ntype: io.kestra.plugin.core.log.Log\n",
})).toThrow(/foo.*not a collection/)
```

ui/packages/topology/src/utils/yaml/blocks.ts
```diff
@@ -0,0 +1,453 @@
...
+    if (isCollection<Node>(parentNode)) {
+        return parentNode
+    }
+
+    const newParentKey = lastSegmentKey(
```
Piyush-r-bhaskar: Uses a different path for quoted vs unquoted input, so either explain it or use `parentPathWithoutKey` for both.

ui/packages/topology/src/utils/yaml/blocks.ts
```diff
@@ -0,0 +1,453 @@
...
+    const sectionNode = getSectionFromDocument({yamlDoc, section})
+    return {yamlDoc, sectionNode}
+}
+
+function getSectionFromDocument({yamlDoc, section}:
+    {
+        yamlDoc: Document<YAMLMap<Scalar<string>, Node>>,
+        section: string
+    }) {
+    const sectionNode = yamlDoc.contents?.items?.find(
```
Piyush-r-bhaskar: Commit 1 added `scalarKey` so key lookups stop reading `e.key.value` directly, and applied it at around sixteen sites. This one was missed.

It also leans on the cast for the value: nothing checks that the section is actually a sequence, so a source like `tasks: foo` gets through here and fails further down. `.find((e) => scalarKey(e) === section)` plus an `isSeq` check on the value would drop the cast entirely.

ui/packages/topology/src/utils/flowYamlUtils.ts
```diff
@@ -1,1176 +1,11 @@
...
-            }
-        }
-    }
-    return tasksLines
-}
\ No newline at end of file
+// Public surface of the flow YAML helpers, kept as one entry point so that
+// `@kestra-io/topology/flow-yaml-utils` consumers are unaffected by the split.
+
+export {parse, stringify, pairsToMap} from "./yaml/serialization.ts"
```
Piyush-r-bhaskar, line 4: With `flowYamlUtils.test.ts` gone, nothing covers the barrel's exports.

ui/packages/topology/src/utils/yaml/metadata.ts
```diff
@@ -0,0 +1,153 @@
...
+            isMap(item.value) || isSeq(item.value)
+                ? item.value.toJSON()
+                : isScalar(item.value)
+                    ? item.value.value
+                    : undefined
+    }
+    return metadata
+}
+
+export function deleteMetadata(source: string, metadata: string) {
```
Piyush-r-bhaskar, line 139: `deleteMetadata` now takes a `string`, but EE's own `flows/Actions.vue` still passes `flowStore.flow?.source`.

Reproducible with `npm run check:types` in `ui-ee`:

```
src/override/components/flows/Actions.vue(269,45): error TS2345:
  Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
```

Fix is the guard you already added on the OSS side.

Piyush-r-bhaskar (changes requested):
Nice cleanup ❤️, the module split makes this a lot easier to read.

Few remarks and then we are good

ui/packages/topology/src/utils/yaml/blocks.ts
elevatebart: Correct, and my "byte-identical for every insert shape" claim in the description simply didn't cover that shape. Fixed in dab090f: throws with `nodeKind`, `undefined` and `null` still take the create path. Your test is in as `rejects a parent path that holds a scalar rather than a collection`, and it fails on 2f21a1d.

One thing your `null` note led me to that is worth separating out. `null` on the create path is right, but the create path then pushes a *new* pair rather than filling the existing one, so a key that is present-but-null gets duplicated — the same defect you found, in its null form:

```
insertBlockWithPath({source: "id: t\nnamespace: n\ntasks:\n", parentPath: "tasks", newBlock: "id: new\ntype: T\n"})

id: t
namespace: n
tasks:
tasks:
  - id: new
    type: T
```

Same for a null nested parent (`cases:` with nothing under it). This is pre-existing, not from this PR — `if (!parentNode)` treated null as absent before the refactor too — and the fix is different in kind (fill the existing pair instead of pushing), so I have left it out rather than widening this PR on my own. Happy to do it here if you would rather it went in now.

I had written a test over that case asserting only `toContain("id: new")`, which passed against the corrupt output; I have dropped it rather than leave a test that hides this.

---

ui/packages/topology/src/utils/yaml/blocks.ts
elevatebart: Unified on `parentPathWithoutKey` in dab090f — the asymmetry was a bug, not a subtlety worth a comment.

Working through why, in case it is useful: `getParentNode` returns the container that should receive a pair named after `parentPath`'s last segment, so the key for a container it has to create is the last segment of `parentPathWithoutKey`. The unquoted branch did that; the quoted branch passed the full `parentPath` and so named the new container after the leaf. `tasks[0].cases["1.0"]` against a task with no `cases` produced a `1.0` key directly on the task instead of `cases`. Covered by `names a created parent after its own path segment, not the quoted leaf`, which also fails on 2f21a1d.

The existing dotted-case test never reached it, since `cases` exists there and the `isCollection` branch returns first.

Not fixed, and visible in that new test's output: the created container is a `YAMLSeq` even where a mapping is wanted, so the result is `cases:\n  - "1.0": [...]`. That is the same pre-existing shape problem as the `tasks[0]` garbage key noted in the description, and wants the same separate change.

---

ui/packages/topology/src/utils/yaml/blocks.ts
elevatebart: Missed that one. Done in dab090f:

```ts
const value = yamlDoc.contents?.items?.find((e) => scalarKey(e) === section)?.value
return isSeq<YAMLMap<Scalar<string>, Node>>(value) ? value : undefined
```

Behaviour note, since the cast was load-bearing in one direction: a section whose value is not a sequence (`tasks: foo`) was previously cast to `YAMLSeq` and handed back, and now reads as absent. Callers treat absent as "no section" and return the source unchanged or `false`, so malformed input fails quietly instead of part-way through.

---

ui/packages/topology/src/utils/flowYamlUtils.ts
elevatebart: Fair — I diffed the surface by hand while doing the split and then left nothing behind to hold it. Added `tests/units/utils/yaml/barrel.test.ts` in dab090f: asserts the exact set of 29 re-exported names, and that each is callable rather than merely present.

The name list is deliberately spelled out rather than derived, so dropping a re-export fails the test instead of quietly shrinking an expectation computed from the same source.

---

ui/packages/topology/src/utils/yaml/metadata.ts
elevatebart: Confirmed against a fresh EE clone (`c268f294`) — `ui-ee/src/override/components/flows/Actions.vue:269` is the same call with `flowStore.flow?.source`, so it breaks exactly as you describe. Thanks for catching it; I only grepped OSS for callers.

I have **not** pushed the EE side. This session was given one branch in this repo, so a branch and PR in `kestra-ee` is beyond what I should do without being asked. The patch, matching the OSS guard:

```diff
     const restoreFlow = () => {
+        if (!flowStore.flow?.source) return
         flowStore.createFlow({
-            flow: YAML_UTILS.deleteMetadata(flowStore.flow?.source, "deleted"),
+            flow: YAML_UTILS.deleteMetadata(flowStore.flow.source, "deleted"),
         }).then(() => {
```

Stated plainly: that is a proposal, not something I have compiled. I did not install `ui-ee` dependencies, so I am relying on your `check:types` reproduction rather than my own.

Two ways to land it, your call:

1. **Paired branch.** Same branch name on `kestra-ee` so CI treats it as a linked cross-repo build. Correct if EE's typecheck gates this merge — say the word and I will push it.
2. **Widen the signature back.** `deleteMetadata(source: string | undefined)` with an internal early return, so no caller in either repo needs a guard. Cheaper, but it puts back the nullability hole that tightening exposed, and your comment reads as preferring the guard.

I would go with 1. Leaving this thread open since it is unresolved either way.

---

Piyush-r-bhaskar (approved):
Now it's good 🚀 Thank You

ui/packages/topology/src/utils/yaml/metadata.ts
elevatebart: Pushed the companion branch: `elevate-bot/beautiful-darwin-y7ki0x` on `kestra-ee` (`c88a1fa`), same name as this one so CI pairs them.

I said earlier I was relying on your reproduction rather than my own. That is no longer the case — `ui-ee`'s `check:types` resolves OSS through `../../kestra/ui`, so I pointed it at this branch and ran it both ways.

Without the guard, your error exactly:

```
design-system compiled
oss compiled
src/override/components/flows/Actions.vue(269,45): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
```

With it, all four stages:

```
design-system compiled
oss compiled
app checked
test checked
```

The change is the one-line guard, nothing else — `npm install` had added a `node: ">=24"` line to `ui-ee/package-lock.json` and I reverted that rather than carry it along.

Resolving this one. Worth merging the two together, or EE first, since EE's typecheck goes red on this PR alone.

---

PR #19222, opened 2026-09-09
refactor(executions): typed executions store, no `any` left in it

### LINKED ISSUE #19269: Executions store: replace explicit `any` with real types
55 explicit `any` in `src/stores/executions.ts`, from `ui/scripts/explicit-any/baseline.json`.

| any | file |
| --- | --- |
| 55 | `src/stores/executions.ts` |

elevatebart:
### 🔗 Related Issue

None, this is a typing cleanup.

### ✨ Description

`ui/src/stores/executions.ts` had 51 uses of `any`. They are all gone, replaced by the types the SDK already generates from the OpenAPI spec rather than by a wider hand-written guess.

Two of them were hiding actual mismatches rather than just being lazy:

- **`logs` was declared `{total, results}` but has always been an array at runtime.** `loadLogs` assigned the endpoint's `LogEntry[]` through `data as any`, so consumers had grown `Array.isArray(raw) ? raw : raw?.results` sniffing and three `FIXME: any` casts to cope with both shapes. `logs` is now `LogEntry[]` and all of that is deleted. The `Array.isArray` check also doubled as "a fetch has come back" for the empty-state placeholder, so that is an explicit `logsLoaded` ref now instead of type-sniffing.
- **`replayExecutionWithInputs` leaked its axios envelope**, and `ReplayWithInputs.vue`'s in-place-replay branch did `response = await waitForState(...) as any` — assigning an execution over an axios response and then reading `.data` off it, which would be `undefined`. The store's `validate*`/`resume*`/`replayExecutionWithInputs` now resolve with the body, which fixes it.

The rest, briefly:

| Was | Now |
|---|---|
| `flow: any`, `flowsExecutable: any[]` | `FlowForExecution` |
| `flowGraph: any` | the topology package's `FlowGraph` (same as `stores/flow.ts`) |
| `subflowsExecutions: Record<string, any>` | `Record<string, Execution>` |
| `Record<string, any>` option bags (13×) | `FilterQuery` plus the action's declared options |
| `findExecutions(): Promise<any>` | overloaded on `onlyTotal` → `number` \| `PagedResultsApiLightExecution` |
| `labels: any`, `flowFilters: any` | `Label[]`, `ExecutionRepositoryInterfaceFlowFilter[]` |
| bare `axios.post`/`axios.get` | `axios.post<Execution>` / `<ValidationResponse>` / `<FilePreview>` / `<string>` |

`routeQueryToQueryFilters` now takes `Record<string, unknown>`, which is what callers were handing it all along: the bulk-action dispatcher in `Executions.vue` merges `latestRevision`/`newStatus` into the route query and passes one bag, and only `filters[…]` keys are ever read. That removed six rest-destructures whose sole job was hiding the extras.

`loadNamespaces` and `loadFlowsExecutable` move onto the generated SDK — the endpoints their comments claimed were missing are `listExecutableDistinctNamespaces` and `listFlowExecutionsByNamespace`, on the same paths.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit`) — 249 files, 2271 tests, plus the topology package's 133

No screenshots: nothing user-visible changes. The one behavioural difference is the "no logs" placeholder now keying off `logsLoaded` instead of off the type of `logs`, which renders the same in both states.

### 📝 Additional Notes

- **Consumers touched, and why.** Making the store honest breaks callers that were relying on `any`: `Logs.vue` and `useExecutionRoot.ts` for the logs shape, `LowCodeEditor.vue`'s `flowGraph` prop, `TriggerFlow.vue`'s local `ExecutableFlow` duck type (now `FlowForExecution`), `PrevNext.vue` (search hits are `ApiLightExecution`, which is what the endpoint returns), and `filterValidLabels` in `components/executions/utils.ts`, which declared `key: string | null` and so could never produce an SDK `Label` even though it filters the nulls out at runtime.
- **The topology package** needed `BranchType` exported so the graph types can be named from the app, and its `subflowsExecutions?: Record<string, any[]>` replaced — the values are executions, not arrays, and the code reads `.id` off them. The package has no SDK dependency, hence a minimal local `GraphExecution` rather than `Execution`.
- **Test churn** is the `{data: …}` unwrapping in the validate stubs across four spec files, plus a return-type annotation where a widened string literal no longer satisfies `InputType`.
- **Storybook tests** could not run where this was written (Playwright browser build 1194 available, 1228 pinned, download blocked), so the one edited story — `LowCodeEditor.stories.tsx`, a JSON fixture cast to `FlowGraph` — was typechecked but not executed locally. CI's `Frontend - Tests` has since run it green, so nothing is outstanding here.
- **Deliberately left alone**: the `as unknown as Promise<Execution>` casts on the single-execution SDK calls (the SDK returns its own `Execution`, the store extends it), and `waitForState`'s `Promise<unknown>` return in `executionUtils.ts`. Both predate this PR and widening them pulls in a different set of files.

### 🤖 AI Authors

Why did the cat refuse to review this PR? It had already typed everything itself, by walking across the keyboard, and `any` is what came out. 🐱

https://claude.ai/code/session_011vbti2ByZ1dbNrJEVeToR7

Closes https://github.com/kestra-io/kestra/issues/19269.
Relates to #19272, #19284, #19286, #19294, #19307 and #19308.


elevatebart:
The same pass on the flow store is stacked on top of this branch: kestra-io/kestra#19227. It needs the widened `routeQueryToQueryFilters` from here, so its base is this branch rather than `develop`; merging this one first lets it retarget to `develop` on its own.

---

Piyush-r-bhaskar (approved):

Issue #2326, opened 2023-10-18
Validate that a WorkingDirectory task didn't contains any task with a workerGroup


loicmathieu:
### Feature description

Validate that a WorkingDirectory task didn't contains any task with a workerGroup as otherwise the workerGroup will be ignored leading to possible confusion.

PR #19224, opened 2026-09-09
fix(executions): scope the concurrency-limit read to the flow

AcevedoR:
Backport of a1a31d5be9 (develop) to `releases/v1.3.x`.

### Problem

Reported in Pylon #2326 on **1.3.32**: a flow's **Concurrency** tab never opens for a flow that declares a `concurrency:` block — it renders `failed to load concurrency limit`.

`FlowConcurrency.vue` reads the running/limit ratio from `GET /concurrency-limit/search`. EE restricts that endpoint to superadmins (class-level `@IsSuperAdmin`), so every other user gets a **403**. No IAM role can grant it. The tab is pushed for every user (`FlowRoot.vue`), so a normal user always sees a broken tab. The same 403 also fires when opening a QUEUED execution.

Flows *without* a concurrency block appear to work only because the error branch lives inside `v-if="flow?.concurrency"` — the identical 403 happens and is never displayed.

### Fix

Add a flow-scoped `GET /api/v1/{tenant}/concurrency-limit/{namespace}/{flowId}` and read that instead of the instance-wide search. The companion EE PR gates it on `FLOW` `READ`, which the tab already requires to load the flow itself.

A flow declaring a concurrency block but never executed has no limit record, so the new endpoint legitimately 404s. On this branch a 404 sets `coreStore.error` and swaps in the global not-found page, so this also adds the `ignoreNotFound` request option (present on develop via `kestraHttp`, which does not exist here) to let a caller treat a missing record as a normal outcome.

### Not included

The **Force Run / Unqueue permission gating** from the original commit is deliberately left out. It depends on the granular `EXECUTION.FORCE_RUN` / `EXECUTION.UNQUEUE` actions, and `ui/src/models/action.js` on this branch defines only `READ/CREATE/UPDATE/DELETE`. Including it would mean backporting the granular action model and its EE permission counterpart — a much larger change than this fix warrants on a release branch.

### Conflicts resolved

The pick was **not** clean:

- `ConcurrencyLimitController.java` auto-merged but referenced `concurrencyLimitRepository`, which does not exist on this branch — retargeted to `concurrencyLimitService.findById(...)` (same signature, returns `Optional`).
- `FlowConcurrency.vue` — took the new fetch logic only; the surrounding template and the `runningCount`/`runningCountSet` refs this branch's template binds to are kept, and `error` becomes a boolean.
- `Executions.vue`, `ForceRun.vue`, `Unqueue.vue` — restored to the branch version (see *Not included*).
- `ui/tests/unit/components/FlowConcurrency.spec.ts` — develop's spec targets `@kestra-io/kestra-sdk` / `@kestra-io/design-system` and asserts the later `staleLimit` behaviour, none of which exist here. Rewritten against this branch's `useAxios` / `ui-libs` / `el-*` components, covering only what is backported.

### Verification (local)

- `./gradlew :webserver:test --tests "ConcurrencyLimitControllerTest"` → **3 passing**, including the flow-scoped GET and its 404 case.
- `npm run test:unit` → **222 passing (32 files)**, including the 3 new `FlowConcurrency` specs.
- `npx vue-tsc --noEmit` → clean.
- `npm run lint` → 0 errors (43 pre-existing warnings, all in untouched files).

Companion EE PR: kestra-io/kestra-ee (linked below once open).
Upstream issue: kestra-io/kestra-ee#10182.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -631,6 +631,34 @@ paths:
               schema:
                 $ref: "#/components/schemas/PagedResults_ConcurrencyLimit_"
   /api/v1/{tenant}/concurrency-limit/{namespace}/{flowId}:
+    get:
+      tags:
+      - Flows
+      summary: Get the concurrency limit of a flow
+      operationId: getConcurrencyLimit
+      parameters:
+      - name: namespace
+        in: path
+        required: true
+        schema:
+          type: string
+      - name: flowId
+        in: path
+        required: true
+        schema:
+          type: string
+      - name: tenant
+        in: path
+        required: true
+        schema:
+          type: string
+      responses:
+        "200":
+          description: getConcurrencyLimit 200 response
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ConcurrencyLimit"
     put:
       tags:
       - Flows
--- a/ui/src/axios.d.ts
+++ b/ui/src/axios.d.ts
@@ -3,6 +3,7 @@ import "axios";
 declare module "axios" {
   export interface AxiosRequestConfig {
     showMessageOnError?: boolean;
+    ignoreNotFound?: boolean;
     skipAuthErrorHandling?: boolean;
   }
 }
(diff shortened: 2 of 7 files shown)


AcevedoR:
Companion EE PR: kestra-io/kestra-ee#10911 (must merge after this one).

loicmathieu (approved):
LGTM

elevatebart (approved):

PR #19227, opened 2026-09-09
refactor(flows): typed flow store, no `any` left in it

### LINKED ISSUE #19277: Flow store: replace explicit `any` with real types
24 explicit `any` in `src/stores/flow.ts`, from `ui/scripts/explicit-any/baseline.json`.

| any | file |
| --- | --- |
| 24 | `src/stores/flow.ts` |

elevatebart:
### 🔗 Related Issue

None, this is a typing cleanup. Was stacked on kestra-io/kestra#19222 for the widened `routeQueryToQueryFilters` signature; that PR has merged, so this is now rebased straight onto `develop` and carries only the flow-store scope.

### ✨ Description

The same pass as #19222, on `ui/src/stores/flow.ts`. Its 22 uses of `any` are gone, replaced by the types the SDK generates.

Three of them were hiding real problems:

- **`loadGraph` returned `Promise<any>`**, and its `.catch` swallowed a failed graph load into `undefined`. `playground.ts` then did `graph.nodes.find(...)` and passed the result straight to `getNextTaskNodes`, which dereferences both the graph and the node. So a failed graph load, or a `taskId` not present in the graph, was a `TypeError`. Both guards are now in place.
- **`loadFlow`'s 404 branch read `e.body`**, a field nothing in the client or the interceptors ever sets, so the deleted-flow path always returned `{}`. It now says that. The one caller (subflow-input autocompletion) only reads `inputs` and falls back to `[]`, so this is a no-op at runtime.
- **`transformResponse` was handed `data as any` by both its callers** because the dependency endpoints declare `nodes`, `edges` and the ids inside them as optional, while the helper demanded all of them. The helper now takes the declared shape.

The rest:

| Was | Now |
|---|---|
| `search: any[]` | `SourceSearchResult[]` |
| `revisions: any[]` | `FlowRevision` — `FlowWithSource` with `revision` filled, which is what a revision list is keyed on |
| `metrics: any[]`, `tasksWithMetrics: any[]` | `string[]` (both endpoints return metric/task names), which also drops two `as string[]` casts in `FlowMetrics.vue` |
| `{[key: string]: any}` option bags | `FlowSearchOptions` and `SourceSearchOptions` |
| `findFlows(): Promise<any>` | overloaded on `onlyTotal` → `number` \| `PagedResultsFlow` |
| `catch (e: any)` ×2 | `KestraHttpError` |
| `Input.defaults: any`, `t(key, values: Record<string, any>)` | `unknown` |
| `body: options.task as any` | cast through the generated body type; `result.constraints` read off the typed result |

`kestraHttp.ts` declares `KestraHttpError` for exactly these catch sites, and its own comment notes that a `catch (e: any)` defeats the `unknown` it puts on `response.data` — so those two were working against a guard the codebase had already put up.

Also dropped a dead `validateStatus` from `fetchGraph`: it was passed inside a `config` bag that only `params.subflows` is ever read from, and never reached a client.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit`) — 256 files, 2281 tests, re-run after the rebase onto `develop`

No screenshots: nothing user-visible changes.

### 📝 Additional Notes

- **Two casts I kept, both narrowing the SDK rather than widening it.** `data as FlowRevision[]` in `loadRevisions`, because `FlowWithSource` is shared with the single-flow shape and leaves `revision` optional; and `{} as Flow` in the 404 branch described above, where the comment records why an empty flow is the right stand-in.
- **`SourceSearchOptions` is spelled out rather than derived** from `Parameters<typeof FlowsAPI.searchFlowsBySourceCode>`. Deriving it puts this module's own `@kestra-io/kestra-sdk/flows` instance into the store's exported types, and EE — which imports this store and has its own SDK build — then resolves its SDK to OSS's copy and loses every EE-only endpoint. Verified with `ui-ee`'s `check:types` against this branch.
- **`useFlowRoot.spec.ts`** stubbed `loadFlow` as resolving `undefined`, which only compiled while the return type was `any`. `useFlowRoot` ignores the result, so the stub keeps returning nothing with a cast that says so.
- **`transformResponse`'s other caller** (`useDependencies.ts`) still passes `data as any`. That is a different endpoint's response and outside this PR; the helper now accepts the honest shape, so removing that cast is a follow-up.
- **Not touched**: `useFlowRoot.ts` calls `loadFlow({...route.params, ...route.query} as any)`, spreading a route into the options bag. Typing that means deciding what the route contributes, which belongs with the route-params work rather than here.

### 🤖 AI Authors

The cat has reviewed this one too, and reports that `unknown` is much better: you have to check what's in the box before you get in it. 🐱

https://claude.ai/code/session_011vbti2ByZ1dbNrJEVeToR7

Closes https://github.com/kestra-io/kestra/issues/19277.
Relates to #19287 and #19307.


elevatebart:
`Generate PR docker image / Publish Docker` fails on this PR, and it is not the diff — it is the stacked base branch. [The job](https://github.com/kestra-io/kestra/actions/runs/34366108201/job/102517163326) ends at:

```
ERROR: failed to build: failed to solve:
failed to parse stage name "kestra/kestra:elevate-bot/ecstatic-allen-5un1pj": invalid reference format
```

The reusable workflow passes the PR's base ref as `KESTRA_DOCKER_BASE_VERSION`, and `Dockerfile.pr` does `FROM kestra/kestra:$KESTRA_DOCKER_BASE_VERSION`. A branch name containing `/` is not a legal Docker tag, so buildx rejects it before reading a single line of the change. Any PR based on a branch rather than `develop` fails this job the same way, whatever it contains — and even a slash-free feature branch would fail one step later, since no `kestra/kestra` image is published per branch.

**No fix belongs in this PR.** The build arg comes from `kestra-io/actions/.github/workflows/kestra-oss-pullrequest-publish-docker.yml`, another repository, so `Dockerfile.pr` cannot correct it from here. The fix there is to fall back to `develop` when the base ref is not a published tag, roughly:

```yaml
# in the reusable workflow, where KESTRA_DOCKER_BASE_VERSION is computed
KESTRA_DOCKER_BASE_VERSION=${{ (github.base_ref == 'develop' || startsWith(github.base_ref, 'releases/')) && github.base_ref || 'develop' }}
```

Not re-run: `invalid reference format` is a Dockerfile parse error, identical on every attempt, so a re-run would only spend another multi-arch build to print the same line.

It does not block the merge either — `required-checks`, the branch-protection gate, needs `[file-changes, frontend, design-system-frontend, backend, build-artifacts, e2e-tests]` and not this job. Everything in that list has passed or been legitimately skipped so far, with E2E still running. The practical consequence is only that this PR has no `ghcr.io/kestra-io/kestra-pr:19227` image to try; #19222's image covers the same branch content minus the flow-store commit. Once #19222 merges and this PR retargets to `develop`, the job will pass on its own.

---

elevatebart:
Pushed `1308f5e`: this PR broke the EE frontend build, and OSS CI cannot see it.

`SourceSearchOptions` was derived from `Parameters<typeof FlowsAPI.searchFlowsBySourceCode>[0]`, which put this module's own `@kestra-io/kestra-sdk/flows` instance into the flow store's exported types. EE imports this store (`kestra/src/stores/flow`, in `usePromoteFlow.ts` among others) and ships its own SDK build, so EE's `@kestra-io/kestra-sdk/flows` then resolved to OSS's copy and lost every EE-only endpoint:

```
ui-ee/src/stores/policies.ts(220,31): Property 'previewPolicies' does not exist on type
  'typeof import("…/kestra/ui/node_modules/@kestra-io/kestra-sdk/dist/flows")'
ui-ee/src/stores/promote.ts(290,37): Property 'flowHashesByIds' does not exist on type …
```

Writing the shape out fixes it. Established by running `ui-ee`'s own `check:types` — which compiles the design system, the whole OSS app, then the EE app — against three OSS checkouts: clean on `develop`, those errors with the derived type (reproduced three times), clean again with the shape spelled out.

Worth flagging for the repo rather than just this PR: **`trigger-ee` runs EE `compileJava` only**, so on a frontend-only PR its green tick says nothing about `ui-ee`. Nothing on the OSS side typechecks the EE frontend, which is why this got as far as a green PR. The same trap is open to anyone deriving a type from `typeof <SDK module>` in a store EE imports.

---

elevatebart:
`trigger-ee` also failed on `1308f5e`, and it never got as far as compiling anything:

```
curl: (28) Operation timed out after 1800002 milliseconds with 0 bytes received
Webhook HTTP status: 000000
Webhook did not return a JSON result (HTTP 000000). Raw response:
```

The EE compile check posts to a webhook on a Kestra instance and waits for a verdict; here the request hit its 30-minute timeout with **zero bytes received**. That is the action doing what it is designed to do — it treats a non-JSON or inconclusive response as a failure rather than a false green — so this is the CI service not answering, not an EE compile error. A real one prints `EE compilation failed` with the javac output.

Two independent reasons it isn't this PR's: the check compiles **Java**, and this branch changes TypeScript only; and the request never reached a compile step at all.

I re-ran the failed jobs once ([run 34372848372](https://github.com/kestra-io/kestra/actions/runs/34372848372)). If it times out again, the webhook endpoint needs looking at rather than the diff.

For the EE frontend, which `trigger-ee` does *not* cover: `ui-ee`'s own `check:types` is clean against this head — that is what the previous commit was about.

---

elevatebart:
Follow-up on the `trigger-ee` failure: the re-run failed too, differently, and the second failure narrows it down rather than repeating the first.

```
Webhook HTTP status: 500
Raw webhook response (HTTP 500)
EE compile check inconclusive for ref elevate-bot/typed-flow-store (no compile verdict; HTTP 500).
Likely a transient runner error — re-run the job.
```

So the two attempts were a 30-minute timeout with zero bytes, then a 500 whose verdict fields — `compileFailed` and `compileError` — are **empty**. Nothing compiled either time. It is the EE CI flow erroring, not a network blip on the way to it, and not a compile error in this branch (which changes TypeScript only; the check compiles Java).

That was my one re-run, so I am not spending another; the action's own hint says to re-run, but two inconclusive results in a row make that the wrong lever. This wants someone with access to the Kestra CI instance behind `KESTRA_CI_EEBUILD_WEBHOOK_URL` to look at why the flow 500s with an empty verdict.

It does not gate the merge: `trigger-ee` is not in `required-checks`'s `needs`, and `Required checks` has already passed on this head, along with `Frontend - Tests`, `E2E - Tests`, `Build Artifacts` and translations parity. The same job passed on kestra-io/kestra#19233 twenty minutes earlier, so it is intermittent rather than branch-specific.

---

elevatebart:
Rebased onto the new base and pushed (`b8fe880`). @Piyush-r-bhaskar force-pushed `elevate-bot/ecstatic-allen-5un1pj` to `7e852fc` (current `develop` + the executions store + its baseline), which left this PR's head hanging off the old `develop` and would have given it the same 121-file phantom diff #19233 just had. The two commits here replay cleanly onto `7e852fc`; nothing else changed in them.

One commit added, mirroring `7e852fc`'s for the executions store: `develop` gained the explicit-any baseline gate (#19261) while this was open, so the baseline needed lowering. `npm run check:ts-any` reports exactly what this PR does:

```
src/stores/flow.ts: 24 -> 0 (removed from the baseline)
src/stores/playground.ts: 5 -> 3
```

The `playground.ts` drop is the two guards this PR adds around the graph load, not a separate change.

Revalidated on the rebased head: `check:types`, `test:unit` (256 files, 2281 tests on the new `develop`), `lint` and `build` all clean.

**One thing for someone with kestra-ee access, unrelated to this PR.** `ui-ee`'s `check:types` now fails on current `develop`:

```
ui-ee/src/override/components/flows/Actions.vue(269,45): error TS2345:
Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
```

That is `YAML_UTILS.deleteMetadata(flowStore.flow?.source, "deleted")` meeting the stricter `flowYamlUtils` signatures from #19221. I verified it is not this branch: the identical error appears with the OSS checkout at `7e852fc`, i.e. with the executions store but **without** the flow-store change. EE needs a `?? ""` or a widened signature; nothing on the OSS side gates it, since `trigger-ee` compiles Java only.

---

Piyush-r-bhaskar (approved):

Issue #10, opened 2019-12-13
RunContext output is unreachable with Flowable test


tchiotludo:
Each flowable tasks at the `parentId` on the output making it unreachable with handlebar vars

tchiotludo:
too old

PR #19229, opened 2026-09-09
feat(topology): add iteration selector to node action menu

### LINKED ISSUE #18987: Topology: Support multi-iteration task selection in the node action menu
### Feature description

Currently, when interacting with a looping task (e.g. `ForEach`) in the Topology execution overview, the `TaskNode` silently targets iteration zero (`taskRuns[0]`) for actions like Outputs, Logs, Restart, and Change Status. 

While the Gantt chart renders one `TaskRunActions` row per iteration, the Topology overview harmonized action menu needs a way to let users select which specific iteration they want to target when clicking an action.

This issue tracks adding a plural slot prop or a UI selector to the Topology node menu to handle multiple task iterations gracefully.

hemantmm:
… menu

closes: #18987

### ✨ Description

Added an iteration selector (`KsSelect`) to the `TaskRunActions` dropdown that appears when multiple task iterations exist. Removed the `taskRuns.length > 1` guard to expose all standard actions, and updated all underlying callbacks and component bindings (including Fix-with-AI and SubFlowLinks) to dynamically evaluate against the user-selected iteration (`currentTaskRun`) rather than defaulting to the first iteration.


flcarre (changes requested):
Design-system usage here is solid (correct `KsSelect`/`KsOption`, token-friendly, i18n key parity across all 13 locales) — but two things should be fixed before merge:

**1. Stale iteration selection after `taskRuns` changes (major)**
`ui/src/components/executions/TaskRunActions.vue:174-180`

```js
const selectedTaskRunIndex = ref(0)
const currentTaskRun = computed(() => {
    if (props.taskRuns && props.taskRuns.length > 0) {
        return props.taskRuns[selectedTaskRunIndex.value] || props.taskRun
    }
    return props.taskRun
})
```

`selectedTaskRunIndex` is never reset or clamped when `props.taskRuns` changes. If the array shrinks or the previously-picked run disappears, `currentTaskRun` silently falls back to `props.taskRun` (iteration 0) while the `KsSelect` label keeps showing the old, now-wrong index. Since this menu drives Restart/ChangeStatus/download-copy-delete-logs, a user could act on a different iteration than the one displayed — the exact class of bug this PR means to fix. Suggest resetting or clamping `selectedTaskRunIndex` on a `watch(() => props.taskRuns, ...)`.

**2. Six locales ship the untranslated placeholder (minor)**
`es.json`, `fr.json`, `it.json`, `pl.json`, `pt.json`, `pt_BR.json` all add `"iteration": "Iteration"` — the literal English word — while the sibling key `"attempt"` is properly localized in every one of those files. Please translate: fr "Itération", es "Iteración", it "Iterazione", pl "Iteracja", pt/pt_BR "Iteração".

**3. No test coverage for the behavior change (minor)**
The `taskRuns.length > 1` guard that previously hid most actions for multi-iteration tasks was removed, and several props switched from `taskRun` to `currentTaskRun`, with no accompanying test/story change.

hemantmm:
Thanks for the review @flcarre! I've pushed an update addressing all three points:

1. Added a `watch` on `props.taskRuns` that explicitly clamps `selectedTaskRunIndex` to the new bounds (or resets it to `0` if the array shrinks to empty), ensuring users never silently interact with the wrong iteration.
2. Updated `fr`, `es`, `it`, `pl`, `pt`, and `pt_BR` with the correct localized strings for the `iteration` key. 
3. Added a Vitest unit test suite (`ui/tests/unit/components/executions/TaskRunActions.spec.ts`) that specifically verifies:

   - The index clamping/reset logic when the `taskRuns` array mutates.
   - The `KsSelect` iteration dropdown only displays for multi-iteration tasks, while actions remain correctly unguarded.
   - `SubFlowLink` dynamically evaluates against the selected iteration's `executionId`.
   - The **Fix-with-AI** action actively pulls logs and generates prompts against the properly selected `taskRunId` rather than defaulting to `taskRuns[0]`.

flcarre (changes requested):
The i18n and fingerprint half of this PR is clean (details at the bottom). The component half needs work before it can land — two of the problems are user-visible in exactly the multi-iteration scenario the PR targets.

### Blockers

**1. Duplicate menu entries for multi-iteration tasks — `ui/src/components/inputs/LowCodeEditor.vue:68`**

```js
.filter(a => taskProps.taskRuns?.length > 1 ? !['edit'].includes(a.key) : !EXCLUDED_NODE_ACTIONS.includes(a.key))
```

`EXCLUDED_NODE_ACTIONS = ["outputs", "replay", "edit"]` exists precisely because `TaskRunActions` renders its own `<Outputs>` and `<Restart isReplay>`. Narrowing to `['edit']` was only safe while `TaskRunActions` hid those behind `v-if="!(taskRuns && taskRuns.length > 1)"`. This PR removes that guard but leaves the filter alone, so a multi-iteration node now shows the topology `outputs` action *and* `<Outputs>`, and the topology `replay` action *and* `<Restart isReplay>` — two entries both labelled **Replay** in the same menu.

Apply `EXCLUDED_NODE_ACTIONS` in both branches, and drop the matching `taskRuns?.length > 1` condition on the `divided` marker on line 69 so the node-action block stays separated.

**2. The selector iterates the wrong list — `TaskRunActions.vue:8-23`**

The topology binds the slot prop as `:taskRuns="taskRunsWithDynamicChildren"` (`ui/packages/topology/src/nodes/TaskNode.vue:45`). That list is this task's runs **plus child taskRuns with a different `taskId`** — it exists so "show task logs" can surface e.g. Ansible play logs. Feeding it to an *iteration* picker means:

- a task with one real run but dynamic children trips `taskRuns.length > 1` and grows a selector with no iterations in it;
- child rows have no `.value`, so they render as `Iteration 2`, `Iteration 3` — numbering that corresponds to nothing;
- selecting one points `TaskEdit :taskId="currentTaskRun.taskId"` at a different task's source, and aims Restart / ChangeStatus / delete-logs at a different task's run.

Either pass the task's own `taskRuns` to the `taskActions` slot, or filter in `TaskRunActions` with `taskId === props.taskRun.taskId` before building the options.

### Selection correctness

**3. Selection is an index into a live-mutating array — `TaskRunActions.vue:174-186`.** `taskRuns` is `execution.taskRunList` filtered at render time. On a running execution it grows, and nothing guarantees stable ordering — so while the index stays at 2, *which iteration index 2 means* can change under the user. The watch only clamps shrinkage. Store `selectedTaskRunId` and resolve `currentTaskRun` by id, falling back to `props.taskRun` when that id is gone.

The arithmetic itself is fine: `index + 1` is consistent, and the clamp to `newRuns.length - 1` is correct. Single-iteration is handled — apart from blocker 2, which is what makes "single iteration" mis-detected in the first place.

**4. Deep watch — `TaskRunActions.vue:175-181`.** The handler reads only `newRuns.length`, but `deep: true` walks every taskRun's attempts, outputs and state histories on every execution poll. Watch `() => props.taskRuns?.length`.

**5. Stale remount keys — `TaskRunActions.vue:51,62`.** `attemptIndex` is always `0` from the topology (LowCodeEditor passes neither `attemptIndex` nor `forcedAttemptNumber`), so the key varies only by start date. Parallel `ForEach` iterations starting in the same millisecond produce an identical key, and the dialog is not remounted across an iteration switch. Add `currentTaskRun.id` to both keys.

**6. Destructive action without context — `TaskRunActions.vue:90-95`.** `deleteLogs` is now reachable for multi-iteration tasks and permanently deletes logs for *the selected iteration only*, behind the generic `"Are you sure you want to delete the log?"`. Right after a user has changed a selector, that confirmation needs to name the iteration.

### Design system

**7. `border-bottom` — `:8`.** Bootstrap's utility resolves to `--bs-border-color`, which nothing in the design system overrides, so it will not track the dark theme. `border-bottom` and `px-3` each appear **zero** other times in `ui/src`. Use `<KsDivider />` and `--ks-spacing-*`.

**8. Select nested in a menu — `:8-23`.** `KsDropdownMenu` renders element-plus' `<ul role="menu">` with a roving-focus keydown handler. A `<div>` is not a valid child of `ul[role=menu]`, the menu's arrow-key handler competes with the select's own option navigation, and `ElSelect` teleports its popper to `body` so an option click is an outside-click on the parent `ElDropdown`. There is no precedent for this pattern anywhere in `ui/src`. **Please verify in a browser** that the menu stays open when an iteration is picked and that arrow keys still work.

**9. A11y — `:9`.** No `aria-label` on the select; add `:aria-label="$t('iteration')"`.

**10. Label inconsistency — `:14-22`.** Options render `run.value || 'Iteration N'`, but the `#label` slot always renders `Iteration {index+1}` — pick `orders.csv` and the closed field reads "Iteration 2". `TaskRunLine.vue:88-90` is the pattern to mirror: same text in both places, guarded with `(value ?? 0) + 1` (yours renders `Iteration NaN` if the model is ever undefined).

**Four async states:** N/A — the selector is fed entirely from props and the PR introduces no new fetch.

### Latent

`:255` — `fixErrorWithAi` still prefers `props.attemptLogs`, which belongs to `props.taskRun`, not `currentTaskRun`. Harmless today only because `LowCodeEditor` never passes `attemptLogs` and `TaskRunLine` never passes `taskRuns` — but the props contract permits both, and the prompt would then quote a different iteration's errors. Gate the reuse on `currentTaskRun.id === props.taskRun.id`.

### Tests

Five tests, covering less than the line count suggests. All drive a **stubbed** `KsSelect` by emitting `update:modelValue`, then assert on `wrapper.vm` internals rather than rendered output. What's missing is the part that matters:

- **Nothing asserts what is sent to the backend.** `downloadContent`, `copyContent` and `deleteLogs` are never exercised, so no test pins `taskRunId === currentTaskRun.id` after an iteration switch — the core claim of the PR.
- No assertion that one `KsOption` per taskRun is rendered, nor on the labels (would have caught #10).
- Only `SubFlowLink` is checked for receiving the selected run; `Restart`, `ChangeStatus`, `TaskEdit`, `Metrics` and `Outputs` are not.
- No case with `attemptIndex > 0` — exactly where `selectedAttempt` and the remount keys (#5) break.
- Roughly 250 of 333 lines are the same `stubs` object copy-pasted five times; extract a `mountActions(props)` factory.

### i18n and fingerprints — clean

Recording this so it isn't re-litigated:

- All **13** locales received `iteration` at the same root-level path, so `$t('iteration')` resolves; no collision with the existing `iterations` key.
- Values are **real translations**, not copy-pasted English: `Iteración` / `Itération` / `Iterazione` / `Iteracja` / `Iteração` / `Итерация` / `반복` / `反復` / `迭代` / `पुनरावृत्ति`. `de: "Iteration"` matches English but is correct German.
- `fingerprints.json` is **consistent, not hand-faked**: `sha256("Iteration").slice(0,12) === "baf6d316dc75"`, inserted in sorted position — exactly what `writeFingerprints` emits.

One nit: `hi.json`, `ja.json` and `ru.json` gained a trailing newline they previously lacked — unrelated churn in an otherwise one-line-per-file diff.

### Naming

The title reads as multi-*select*. It's a single-select over the iterations of a multi-iteration task; every action targets one `currentTaskRun`. Worth rewording for the changelog.

hemantmm:
Thanks for the review, @flcarre! I’ve pushed updates addressing all the blockers and latent issues you mentioned.

 **Component logic**

 * Applied `EXCLUDED_NODE_ACTIONS` unconditionally in `LowCodeEditor` to prevent duplicate Outputs/Replay/Edit actions.
 * Filtered iteration candidates by `taskId` in `TaskRunActions` so dynamic child task runs aren’t treated as iterations.
 * Switched from index-based selection to `selectedTaskRunId`, with a fallback to `props.taskRun` if the selected run disappears.
 * Replaced the deep watcher with a length-based watcher.
 * Added `currentTaskRun.id` to the Restart/ChangeStatus remount keys.
 * Added the selected iteration to the delete-logs confirmation.
 * Fixed `Fix with AI` so `attemptLogs` are only reused for the original task run.

 **UI / accessibility**

 * Updated the dropdown structure to keep the menu DOM valid and replaced the Bootstrap spacing/border utilities with the design-system equivalents.
 * Added the iteration `aria-label`.
 * Normalized iteration labels using the native `KsOption` label to avoid inconsistencies/`NaN`.

 **Tests**

 * Reworked `TaskRunActions.spec.ts` around a shared `mountActions()` factory.
 * Added coverage for iteration rendering/counts, `attemptIndex > 0`, correct `taskRunId` propagation, and the selected `currentTaskRun` across Restart, Change Status, Metrics, Outputs, SubFlowLink, delete logs, and Fix with AI.
 * Also cleaned up the unrelated trailing-newline changes in the locale files.

flcarre (changes requested):
Thanks — this round genuinely moved. Going through my previous list, **all ten points are addressed**, and correctly:

- **1 (duplicate menu entries)** — `EXCLUDED_NODE_ACTIONS` is now applied in both branches of the `LowCodeEditor` filter and the `divided` marker is unconditional.
- **2 (wrong list)** — `currentTaskRuns` filters on `taskId === props.taskRun?.taskId`, so dynamic child taskRuns no longer masquerade as iterations. The test at `TaskRunActions.spec.ts:133` pins it.
- **3 (index into a live array)** — selection is `selectedTaskRunId` resolved by id with a `props.taskRun` fallback.
- **4 (deep watch)** — replaced with a length watch.
- **5 (remount keys)** — `currentTaskRun.id` added to both.
- **6 (destructive action without context)** — the iteration is now named in the confirmation.
- **7 (bootstrap `border-bottom`/`px-3`)** — gone, `--ks-spacing-2` used.
- **9 (a11y)** — `:aria-label="$t('iteration')"` reaches the combobox.
- **10 (label inconsistency / `NaN`)** — the `#label` slot is gone, `KsOption :label` is the single source.
- **Latent (`fixErrorWithAi`)** — `attemptLogs` reuse is gated on run identity.
- **Tests** — the `mountActions()` factory is extracted, option count and labels are asserted on rendered output, and `deleteLogs` pins `taskRunId` at the backend boundary. That is the review I asked for.

What is left is smaller but still blocking.

### Blockers

**CI is red, and it is this PR.** `vue-tsc` fails, so not a single test runs:

1. `ui/tests/unit/composables/useReadOnlyYamlKeys.spec.ts:277` — your head carries a duplicated key:

```ts
editor: ref(double.editor) as never,
editor: ref(double.editor) as unknown as ReadOnlyYamlKeysOptions["editor"],
```

`TS1117: An object literal cannot have multiple properties with the same name.` This came out of a bad merge with `develop` — `develop` has exactly one `editor:` line there (the typed cast, from #19461). Revert the file to the `develop` version so the diff stays scoped to this feature.

2. `ui/tests/unit/components/executions/TaskRunActions.spec.ts:44` — `TS6133: 'msg' is declared but its value is never read` in the `toast.confirm` mock. Use `_msg` or drop the parameter.

3. `ui/tests/unit/components/executions/TaskRunActions.spec.ts:98` — `TS2322`: `mountActions(propsData: Record<string, unknown>)` isn't assignable to the component's props. Type the parameter against the component's props instead.

### The divider is the wrong primitive

`TaskRunActions.vue:23`. `KsDivider` wraps `ElDivider`, which renders `<div class="el-divider" role="separator">`. `KsDropdownMenu` wraps `ElDropdownMenu`, which renders a `<ul role="menu">`. So this puts a bare `<div>` as a direct child of a `<ul>` — invalid HTML — and drags in ElDivider's `24px 0` margin, which is a large gap in a compact action menu. My fault for naming `KsDivider` without checking what it renders inside a menu. The right idiom is the one `NodeMenuItem` already uses: put `divided` on the first `KsDropdownItem` after the select and delete the `KsDivider`. The `<li role="presentation">` wrapper you added around the select is correct and should stay.

### i18n by concatenation

`TaskRunActions.vue:243-248` builds a sentence out of two translated fragments plus hardcoded English punctuation and word order. Add a second key instead, e.g. `delete_log_iteration: "Are you sure you want to delete the log for {iteration}?"`, and call `t("delete_log_iteration", {iteration})` — one key per rendered sentence, so translators can reorder it. That also means the 13 locale files get a real string rather than inheriting `(…)` from English.

### My point 8 is still open

`TaskRunActions.vue:9`. `KsSelect` doesn't declare `teleported`, so ElSelect's default `teleported: true` applies and the option popper mounts on `document.body` — outside the `ElDropdown` popper. `ElDropdown` closes on outside click, so clicking an iteration should dismiss the whole menu before you can use the selection. `KsSelect` forwards `$attrs` to `ElSelect`, so `:teleported="false"` is a one-line fix. There is still no precedent anywhere in `ui/src` for a select nested in a dropdown menu. **Please open a ForEach execution in the topology and confirm in a browser** that the menu stays open through a selection and that arrow keys still reach the options — a screenshot or short clip on the PR would settle it.

### Selection correctness, second order

- `:179` — the watch keys on `currentTaskRuns.value.length`. A poll that replaces the list with a same-length array in which the selected id is gone won't fire it, and `currentTaskRun` then falls back to `props.taskRun` while the select still shows the stale id — the display/target mismatch this PR exists to prevent. Watch the ids instead: `() => currentTaskRuns.value.map(r => r.id).join()`.
- `:19` — labels are positional (`index + 1`) in an array whose order isn't guaranteed on a running execution, so "Iteration 2" can change meaning between polls whenever `run.value` is absent. Prefer the run's own iteration value where one exists.
- `:199` — `attemptIndex` isn't reset on an iteration switch. If the newly selected run has fewer attempts, `selectedAttempt` goes `undefined`, which silently hides "Fix with AI" and hands an out-of-range index to `Restart`/`ChangeStatus`. Harmless from the topology today (it always passes `0`), but the props contract permits otherwise.

### Nits

- `:8` — the padding token is right (`--ks-spacing-2`) but it's an inline `style` attribute; the file already has a `<style scoped>` block, move it there.
- `:246` — trailing whitespace.
- `downloadContent` and `copyContent` are still untested; `deleteLogs` is. One more case each would close the loop on "every action targets the selected run".

**Four async states:** N/A — the selector is fed entirely from props, no new fetch.

Get CI green, swap the divider, turn the confirmation into a proper key, and post a browser check of the select-in-menu interaction, and I'll re-review.

flcarre (changes requested):
### Diff shape — blocking, read this first

GitHub reports +2858/-2578 across 66 files for what should be a focused topology change. I compared the PR's base (`808495e3`) against the current `develop` tip (`51875738`) and against the PR head (`8d0d5e49`) file by file. For a long list of files — `ui/src/components/ai/copilot/CopilotChat.vue`, `CopilotComposer.vue`, `DiffView.vue`, `useAiChat.ts`, `useApplyDraft.ts`, `webserver/.../ai/agent/AgentOrchestrator.java`, `ChatMessageAdaptor.java`, `ui/packages/kestra-sdk/src/client-facade.ts`, `ui/src/utils/kestraHttp.ts`, `ui/src/stores/flow.ts`, plus `en.json` — **base and current develop are byte-identical**, meaning develop has not touched them since this PR forked. Yet **head differs from both**, and the difference is a clean removal: gone are the Copilot draft dismiss/apply tracking, the flow-diff preview mirrored into the main editor, and the streaming/cancel (`StreamingHandle`, `abortIfCancelled`, `stop`) support — already-shipped work with no relationship to an iteration selector.

This is not "the PR is behind develop". It is content in head that neither base nor develop ever had, i.e. a genuine revert of merged work, almost certainly from resolving conflicts by keeping the wrong side across the branch's six-plus `Merge branch 'develop'` commits.

On top of that, `ui/scripts/translations/fingerprints.json` is no longer valid JSON: it contains two concatenated copies of the fingerprint map (duplicate `"yes"` key at line 2050 and again at line 4057) with a closing brace missing between them. That is the direct cause of the red **Translations - Key parity, placeholders and drift** job — `check-translations.mjs` throws `SyntaxError: Expected ',' or '}' ... (line 2051 column 3)` before any real check runs.

**What to do:** reset this branch onto current `develop` with a clean rebase (or a fresh branch cherry-picking only the topology commits) and force-push, so the diff shows only `TaskRunActions.vue`, `LowCodeEditor.vue`, the `iteration` / `delete_log_iteration` translation additions, and their tests. Please don't hand-patch the corrupted files in place — check the rebase produces a diff of a few hundred lines, not thousands.

### Previous round — all addressed

Going through my 2026-09-16 review, every one of the ten points is genuinely fixed: `EXCLUDED_NODE_ACTIONS` applied on both branches of the `LowCodeEditor` filter, `currentTaskRuns` filtered by `taskId`, selection resolved by id with a fallback, the deep watch replaced with a length/id-based watch, remount keys carrying `currentTaskRun.id`, the delete-log confirmation naming the iteration via `delete_log_iteration`, the a11y label added, and label rendering unified through a single `KsOption :label`. The `mountActions()` test factory and the rendered-output assertions are in place too. None of that needs redoing.

### Design system

`TaskRunActions.vue:22` — `.iteration-divider` uses `var(--ks-border-primary)`, which does not exist in `ks-tokens.scss`; the real tokens are `--ks-border-default`, `--ks-border-subtle`, `--ks-border-focus`. The border won't render or track the theme in either mode. This also sidesteps the earlier suggestion to use the `divided` prop on the first `KsDropdownItem` — the pattern `NodeMenuItem` already uses — rather than a hand-rolled `<li role="separator">`.

### Still outstanding

The browser confirmation that the select stays open and arrow keys still work with `:teleported="false"` inside the dropdown menu hasn't been posted, and with CI red there's no automated signal either.

Security and performance: nothing to flag in the topology change itself.

Blocked on the corrupted merge history and the invalid `fingerprints.json`; happy to re-review as soon as the diff is just the topology work.

hemantmm:
@flcarre just closing this PR as it is getting more bloated due to conflicts. In that case just opening a new PR to make the work faster to solve.

Closing in favor of #19568

PR #19234, opened 2026-09-09
docs: update unavailable AI Copilot message

aj-emerich:
Currently, this message is angled towards enterprise by saying "admin". A single OSS user would also come across this message, so it is better to be more general. A user who doesn't have access to the Kestra configuration file will know to ask whoever does.


wrussell1999 (approved):

MilosPaunovic (approved):

MilosPaunovic:
Can this be merged now @aj-emerich?

PR #19260, opened 2026-09-10
feat(topology): add run from task actions to flow editor task menu

### LINKED ISSUE #18887: Add a "Run from Here" option for individual tasks in the Flow Editor
## Feature description

Add a **"Run from Here"** action to the Kestra Flow Editor that allows users to start an execution from a selected task rather than always running the entire flow from the beginning.

### Motivation

When developing or debugging a large flow, users often need to repeatedly test a specific task or a downstream section after making changes. Running the entire flow from the beginning can be time-consuming, especially when earlier tasks perform expensive operations such as API calls, data processing, or external resource creation.

A "Run from Here" option would make flow development and debugging significantly faster.

### Proposed behavior

When a user selects a task in the Flow Editor, provide an action such as:

**Run from Here**

The action could:

* Start a new execution beginning at the selected task.
* Skip tasks that precede the selected task.
* Preserve the normal execution order for downstream tasks.
* Clearly indicate in the execution details that the flow was started from a specific task.
* Warn the user if the selected task depends on outputs from skipped upstream tasks.

For example:

```text
Task A
   ↓
Task B
   ↓
Task C  ← Run from Here
   ↓
Task D
```

Selecting **Run from Here** on Task C would execute C → D without requiring A and B to run again.

### Optional enhancement

The feature could provide a choice between:

* **Run only this task**
* **Run from this task**

This would allow users to either test a single task or execute the selected task and its downstream dependencies.

### Benefits

This would improve:

* Flow development speed
* Debugging and troubleshooting
* Iterative workflow development
* Testing of individual sections of large flows
* Developer productivity when working with complex workflows

The feature could initially be limited to manual executions from the UI, with more advanced dependency handling added later.

yuvaraj-2006:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/18887

### Description

Adds **"Run task & downstream"** and **"Run this task"** actions to the task node context menu in the Flow Topology view.
This allows users to execute a workflow starting from any selected task during workflow editing, skipping upstream tasks or testing only an isolated task with downstream breakpoints.

- Supports `disabled` state on node menu actions when an execution is already running.
- Reuses existing design system components and localized translation keys across all 13 supported languages.

### 🎨 Frontend Checklist

### DIFF
--- a/ui/packages/topology/src/nodes/NodeMenu.vue
+++ b/ui/packages/topology/src/nodes/NodeMenu.vue
@@ -42,6 +42,7 @@
         tooltip?: string;
         danger?: boolean;
         divided?: boolean;
+        disabled?: boolean;
         onClick: () => void;
     }
 
--- a/ui/packages/topology/src/nodes/NodeMenuItem.vue
+++ b/ui/packages/topology/src/nodes/NodeMenuItem.vue
@@ -1,6 +1,7 @@
 <template>
     <KsDropdownItem
         :divided="action.divided"
+        :disabled="action.disabled"
         :icon="action.icon"
         :class="{'node-action--danger': action.danger}"
         @click="action.onClick()"
--- a/ui/packages/topology/src/nodes/TaskNode.vue
+++ b/ui/packages/topology/src/nodes/TaskNode.vue
@@ -182,8 +182,8 @@
         (event: typeof EVENTS.DELETE, data: any) :void;
         (event: typeof EVENTS.ADD_TASK, data: any) :void;
         (event: typeof EVENTS.SHOW_CONDITION, data: any) :void;
-        (event: typeof EVENTS.SHOW_DESCRIPTION, data: any) :void;
-        (event: typeof EVENTS.RUN_TASK, data: { task: any }) :void;
+        (event: typeof EVENTS.SHOW_DESCRIPTION, data: any): void;
+        (event: typeof EVENTS.RUN_TASK, data: { task: any; runDownstreamTasks?: boolean }) :void;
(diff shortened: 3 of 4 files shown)


RemiBarthe (changes requested):
Overall the implementation makes sense, but I found two regressions that I think need fixing before merge.

**1. The actions are available when Playground is disabled.** In the flow editor's topology (`?playground=off`, the default), both "Run from here" and "Run this task" are visible and enabled. Clicking them goes through `runUntilTask()`, which does not check `playgroundStore.enabled`, so if the editor has unsaved changes it creates a new revision (`saveAll()`) and starts a `PLAYGROUND` execution, while the editor never switches to the execution graph. Before this PR there was no way to trigger that path from the topology when Playground was off.

**2. The actions disappear after the first Playground run.** Once an execution exists, every node gets an `executionId`, `taskExecution` becomes truthy, and `!taskExecution.value` removes both menu entries. That means the feature is only available before the first run, whereas the replay path (`runUntilTask(..., true)`) is precisely the interesting case after a previous execution.

The condition should match the existing floating play button instead: gate it on `playgroundEnabled && !readOnly`, but not on `!taskExecution`.

A few non-blocking notes while I was there:

* The floating play button already maps to **Run this task**, so that menu entry is currently duplicated.
* The PR description says translations were added, but the `playground.run_this_task` / `run_task_and_downstream` keys already existed (and so did the translated locales).
* `TaskNode.test.ts` already has an `actionKeys()` helper; adding a few cases for these visibility rules would lock down the regressions pretty cheaply.

PR #19310, opened 2026-09-10
fix(assets): accept expressions and free-form types on output assets in the flow editor

MilosPaunovic:
Closes https://github.com/kestra-io/kestra-ee/issues/6508.
Related to https://github.com/kestra-io/docs/pull/5581 (the docs example that used a non-existent `Table` type name).

## What

The flow editor underlined two valid things in an `assets.outputs` declaration:

- `namespace: "{{ flow.namespace }}"` - the flow schema copied the literal format of an asset `namespace` and `id` (`^[a-z0-9][a-z0-9._-]*`) as is, so a Pebble expression failed the pattern. The whole `assets.outputs` declaration is a `Property<List<Asset>>` rendered at runtime, and the docs recommend exactly this form.
- a free-form `type` - the schema only offered the asset types provided by plugins, each pinned to a `const`, so any other type was flagged. `Custom`, the class `AssetDeserializer` falls back to for such a type, is `@Hidden`, hence skipped by `PluginScanner` and absent from the schema, while the docs document free-form types as supported.

## How

`JsonSchemaGenerator` only, no `EE` change needed (the `EE` generator inherits the `Asset` branch):

- on `Asset` properties carrying a `pattern`, the pattern becomes an `anyOf` of the literal format and a Pebble-expression pattern: a literal is still held to its format, an expression is accepted. This is schema-side only, the `@Pattern` bean validation behind the assets `API` is untouched.
- `Custom` is added by hand to the `Asset` subtypes of the schema as the free-form branch, with its `type` left as an open string (with a title and a description) instead of a `const`. The plugin-provided types keep their `const`, so autocompletion is unchanged. A consequence worth knowing: a misspelled built-in type is no longer flagged, which follows from free-form types being a documented feature.

The regression test validates real flow `YAML` against the generated schema with the `json-schema-validator` already on the classpath, for both the expression and the free-form type, and checks that a bad literal is still rejected.

## Screenshots

Captured on an `OSS` instance, so the only asset type available is `External` and the free-form type stands in for `io.kestra.plugin.ee.assets.Table`. The `1 Error(s)` badge on both captures is the `OSS` server-side validation (`assets` are `EE` only), unrelated to the editor markers.

Before:

before

After:

after


fhussonnois (approved):
LGTM!

PR #19313, opened 2026-09-10
fix(topology): inserting into an empty section repeated its key

elevatebart:
Follow-up to kestra-io/kestra#19221, where @Piyush-r-bhaskar found this defect's sibling in review. This form was reported in that thread and deliberately left out, as it needs a different kind of change.

### 🔗 Related Issue

None.

### ✨ Description

Inserting a block into a section written with no value emitted the key twice:

```js
insertBlockWithPath({source: "id: t\nnamespace: n\ntasks:\n", parentPath: "tasks", newBlock: "id: new\ntype: T\n"})
```

```yaml
id: t
namespace: n
tasks:
tasks:
  - id: new
    type: T
```

Same for a nested empty section, such as a flowable's `cases:`:

```yaml
tasks:
  - id: sw
    cases:
    cases:
      - "1.0":
          - id: new
```

A key written with no value parses to a pair holding `null`, and `getIn` returns that `null`. Both the top-level branch of `insertBlockWithPath` and the create-parent branch of `getParentNode` tested it with `!parentNode`, read it as "key absent", and pushed a second pair alongside the first.

Both now go through one `attachSequence` helper, which fills the existing pair when the key is already present and pushes only when it is not. `createPairNode` had no other caller and is gone.

Reachable from the no-code editor: adding the first task to a flow whose `tasks:` key exists but is empty is an ordinary intermediate state while editing, and the result is a document with a duplicate key.

### Evidence

Four insert shapes, emitted YAML diffed before and after. The only change is the two duplicate lines disappearing — the shapes that already worked are byte-identical:

| shape | before | after |
|---|---|---|
| empty top-level `tasks:` | `tasks:` twice | fixed |
| empty nested `cases:` | `cases:` twice | fixed |
| absent top-level key | correct | unchanged |
| populated section with `refPath` | correct | unchanged |

Both new tests fail on 799a0a1, the commit this branches from.

### 🎨 Frontend Checklist

- [x] Type checking passes (`npm run check:types`) — clean across the whole `ui/` workspace
- [x] Unit tests pass (`npm run test:unit`) — 141 in topology, up from 139
- [ ] Screenshots or video recordings attached showing the `UI` changes — the change is in YAML emission; the table above is the observable difference

### 📝 Additional Notes

Two related defects from that PR's description remain untouched, both still in the same function and both wanting their own change:

- `insertBlockWithPath` with a path like `tasks[0].then` against a document with no `tasks` emits a key literally named `tasks[0]`, because `lastSegmentKey` splits on `.` and leaves the index in the key.
- The container created for a missing parent is always a `YAMLSeq`, even where a mapping is wanted, so a quoted nested create yields `cases:\n  - "1.0": [...]` rather than a mapping.

https://claude.ai/code/session_01SHfMvHS2ZAHU8uc5xyCBYF

---
_Generated by [Claude Code](https://claude.ai/code/session_01SHfMvHS2ZAHU8uc5xyCBYF)_


elevatebart:
`Frontend - Tests` failed on 2589598. Two sub-steps, neither caused by this PR's change, both now handled.

**`Npm - check ts-any` — real, pre-existing.** It fails on `develop` at 799a0a1 unmodified, with output byte-identical to this branch. Two merged PRs moved `any` between files without re-recording it:

- kestra-io/kestra#19221 split `flowYamlUtils.ts`, so its 60 entries and the 1 in its test file moved to the new `src/utils/yaml/` paths (14 across four modules, 1 in `blocks.test.ts`).
- kestra-io/kestra#19230 typed the two trigger tables, taking `TriggersManage.vue` 61 → 0 and `FlowTriggers.vue` 47 → 0.

Refreshed the baseline in 3c33004. `--write` rewrites the whole file, so the trigger reductions are locked in alongside the split; that is what the script asks for when a count improves, and it stops those two files regressing quietly. No `any` is added or removed — the total moves 1818 → 1649 purely by recording work that already landed.

The first of those is drift from my own merged PR, so it is mine to clean up either way.

**`Npm - lint (eslint)` — transient, not a lint finding.** eslint reported no errors. reviewdog then failed fetching the PR diff:

```
reviewdog: fail to get diff: GET https://api.github.com/repos/kestra-io/kestra/pulls/19313: 500 Server Error:
Sorry, this diff is temporarily unavailable due to heavy server load.
[{Resource:PullRequest Field:diff Code:not_available Message:}]
##[error]Process completed with exit code 1.
```

A GitHub API 500, nothing to do with the code. `npx eslint` exits 0 locally on this branch and on `develop`, both with the same 3 pre-existing `vue/enforce-style-attribute` warnings. The push above re-runs the job, which is the one retry this warrants.

Verified before pushing: `check:ts-any` clean, the baseline script's own 5 tests pass, 141 topology tests pass, `check:types` clean across `ui/`, eslint 0 errors.

---

Piyush-r-bhaskar (approved):

PR #19317, opened 2026-09-10
feat(executions): failure debug panel over the Gantt view

### LINKED ISSUE #19314: Add a Failure Debug Mode to the execution Gantt view
## Summary

Debugging a failed execution today means manually cross-referencing the Overview, Gantt, Topology, and Logs tabs, with no shared "focused task" between them. This adds a Failure Debug Mode: on a FAILED/KILLED execution, a focused panel opens over the Gantt view, pre-assembled around the failing task — a scoped timeline, structural impact, and the relevant logs side by side — with a restart action and on-demand AI help.

The panel anchors on **Gantt**, not Overview: real usage data shows the Gantt tab already carries roughly 10x Overview's traffic and roughly 90x its errors, so that is where users already are when a run fails.

## Scope / Design

- New component `ui/src/components/executions/gantt/FailureDebugPanel.vue`, mounted from `Gantt.vue`, rendered only when `execution.state.current` is `FAILED` or `KILLED`. No new route, no backend changes — reads/writes through the existing execution, task-run, log, restart, and Copilot endpoints.
- Reuses existing pieces rather than rebuilding: Gantt row rendering, Topology node rendering, the Logs query scoped by `taskId`, the existing restart action, the existing AI Copilot component (instead of a new rule-based "likely cause" engine), and the existing copy-to-clipboard utility.
- Handles multiple simultaneous failures: failed task-runs are listed earliest-first, auto-focusing the earliest; a failure switcher appears only when there is more than one.
- Drag-to-select a time range on the mini-timeline filters the log panel client-side (a plain click still focuses a task — click and drag are disambiguated).
- Fully keyboard/screen-reader accessible: the auto-opening panel announces itself (`aria-live`), moves focus to itself on open, and returns focus on close.
- Design-system only (`Ks*` + `--ks-*` tokens, light and dark) — the failure/elevation tokens used in the design mockup need verifying against the real `ks-theme-*.scss` files before implementation, since they were resolved by hand for the mockup.

## Acceptance Criteria

- [ ] `FailureDebugPanel` renders over the Gantt view only for a FAILED/KILLED execution, and never for a task that failed then succeeded on retry (checked against the task-run's latest attempt, not "any failed attempt exists").
- [ ] Single failure: panel auto-opens and focuses that task; no failure switcher shown.
- [ ] Multiple simultaneous failures: failure switcher shown, earliest-first auto-focus, switching a failure updates the mini-timeline, structural impact, and logs together.
- [ ] Drag-to-select a time range filters the log panel; `Escape` clears the selection; a plain click on a task row still focuses it.
- [ ] Restart action reuses the existing restart call and shows a non-blocking caption on what will and won't re-run.
- [ ] "Ask Copilot" is scoped to the focused failure's error text.
- [ ] Panel announces itself via `aria-live`, moves focus to its heading on open, returns focus to a reopen affordance on close; `Escape` closes it.
- [ ] Restart and "Ask Copilot" controls are hidden (not just disabled) without permission, reusing the existing `@HasAnyPermission` checks.
- [ ] Very long executions: rendering is scoped to failed task-runs and their immediate structural neighbors, not the full task list.
- [ ] Storybook stories: empty (no failure — doesn't render), single failure, multi-failure, light + dark.
- [ ] Vitest unit tests: focus management on open/close, `Escape` handling, earliest-first sort.
- [ ] New user-facing strings added to `en.json` and translated into all locale files.
- [ ] `npm run check:types && npm run test:unit && npm run lint` and `node ui/src/translations/check.js` pass.

## Context

Grounded in real customer-request issues already open on this repo: #13682 (Playground mode from the last execution), #16603 (drag-select a time range on the logs chart), #18271 (Gantt doesn't visually distinguish parent/child task blocks — informs the structural-impact view here).

## Out of Scope

Automated rule-based root-cause inference (replaced by on-demand Copilot); defaulting the execution route to Gantt on failure (needs a `execution.status=FAILED` analytics event to validate first — no funnel data exists yet); "Open in Playground" with pre-filled inputs (tracked separately in #13682); per-task side-effect metadata for the restart caption (ship the generic caption first); subflow-failure drill-down (needs a product decision before scoping).

### LINKED ISSUE #19390: Failure Debug panel: surface task inputs and full state history
## Summary

Review feedback on the Failure Debug panel prototype (#19317) is that it currently re-displays what the Gantt view already shows (a mini-timeline, structural impact) in a different layout, rather than surfacing genuinely new debugging information. This issue was opened to explore going further: surfacing the failing task run's inputs and its full state transition history, the two concrete pieces called out as missing.

**Update:** this went beyond design exploration — both pieces (and two more, once "task inputs" was pressed on further) are now fully implemented directly in #19317 rather than split into a separate follow-up. See below.

## Scope / Design

- Task run inputs: `TaskRun` does not carry the resolved/rendered values used for that specific run, and persisting one raised real questions (per-attempt vs per-execution, secret storage). Resolved instead by extending `ExpressionController`'s existing display-safe `/expressions/render` endpoint with an optional `taskRunId`, computed on demand from data already persisted (execution history + flow revision) — nothing new stored.
- Clarifying what "task inputs" meant surfaced two distinct things, both now shipped: the task's own resolved configuration (its YAML block with expressions resolved against that task run's context), and the flow's own submitted inputs (Kestra reserves "input(s)" for flow-level parameters, distinct from a task's own properties) — both resolved the same task-run-scoped way, and a third: the outputs of upstream tasks the failing task's config actually reads via `outputs.<id>`, fetched through the existing `/outputs/tasks` endpoint.
- State history: `TaskRun.state.histories` already carried every state transition with a timestamp; the panel only used it internally to position bars on its mini-timeline. Now surfaced as a readable list — state, timestamp, and elapsed time since the previous transition, down to the millisecond.
- State history, resolved configuration, and inputs & outputs are grouped as tabs of one card rather than four separate always-visible ones, after an initial pass left the panel feeling cluttered rather than informative.

## Acceptance Criteria

- [x] Design/approach agreed for surfacing a task run's resolved inputs (including how retries are handled) — resolved on demand per task run, not persisted, so retries naturally get the resolved values for their own attempt
- [x] Design/approach agreed for surfacing a task run's full state transition history in the debug panel
- [x] Follow-up implementation issue(s) scoped once the approach is settled — implemented directly in #19317 instead of a separate issue, since the resolve-on-demand approach turned out not to need new persistence or backend scoping work beyond one optional controller parameter

## Context

- Prototype / implementation: https://github.com/kestra-io/kestra/pull/19317
- Original issue: https://github.com/kestra-io/kestra/issues/19314

> Review feedback (paraphrased): this is a good idea but it didn't go as far as it should — currently it basically just displays what's already inside the Gantt in a different way, but to debug a failure we need more information, most specifically the task inputs and the state history.

## Out of Scope

~~Implementation itself — this issue is scoped to exploring the design, not building it.~~ Superseded — implementation shipped directly in #19317, which closes this issue on merge.

flcarre:
## Problem

Debugging a failed execution means cross-referencing the Overview, Gantt, Topology and Logs tabs, with no shared "focused task" between them. Worse, the Gantt only ever describes a failure at the *task* level, while the failure itself lives one level down: in a specific attempt, on a specific worker, often inside a child execution. Nothing in the UI answers "which attempt failed, against what resolved config, and where is the real error" without leaving the page.

## Fix

A Failure Debug Mode on a `FAILED`/`KILLED` execution's Gantt view. A "Debug this failure" button sits in the Gantt's own stats row, next to "Copy All Logs"; clicking it opens `FailureDebugPanel.vue` pre-assembled around the failing task run. It never opens itself — a failed execution is common enough that hijacking the Gantt on every visit would be disruptive.

The panel is built around the failing **attempt**, not the failing task, which is what keeps it from being a re-layout of the Gantt:

- **Identity** — the header names the focused task run, its `value` when it has one, and which attempt failed (`Attempt 2/2`). With several simultaneous failures a `role="tablist"` switcher lists them earliest-first, each carrying its own `value`, so three failures of the same task id are told apart.
- **The error, on its own** — the ERROR-level text for that task run, above the actions rather than left in the log stream. Fetched once per focused failure on first open and reused by Copy error and Ask Copilot, which are disabled (not silently inert) when nothing was recorded.
- **Attempts** — one row per attempt: state, start, that attempt's own duration, and the worker that ran it. `attempts[]` was previously read only for its length, to index the restart action. A retried task that failed twice on the same worker is now readable as such.
- **State history** — every transition of the focused task run, with elapsed time since the previous one, down to the millisecond since `CREATED` → `SUBMITTED` → `RUNNING` usually happens inside one tick.
- **Resolved configuration** — the failing task's YAML block with its Pebble expressions resolved against that task run's context, secrets masked. Computed on demand, nothing new persisted: `ExpressionController`'s existing display-safe `/expressions/render` endpoint gains an optional `taskRunId`, and the flow is fetched at the revision the execution actually ran on.
- **Inputs & outputs** — the execution's declared inputs (resolved in one batched request) and the outputs of the upstream tasks the failing task's own config reads via `outputs.<id>`, parsed out of its resolved block.
- **Subflow drill-down** — when the failing task produced a child execution, a link straight into it. The parent only ever says "Subflow execution … ends in FAILED state"; the real error is one level down.
- **Structural impact** — parent/siblings/children of the failing task (capped at 6) with the app's plugin-icon component, each offering its raw, as-authored YAML block. Raw, never Pebble-rendered: a neighbour may have produced no `TaskRun` at all, so there is no context to safely resolve its expressions against.
- **A scoped mini-timeline** positioning the focused task and those neighbours on a time axis, with drag-to-select filtering the log panel by timestamp (`Escape` clears; a plain click still focuses a row).
- **Actions** — restart via the existing `Restart.vue` (`isReplay`, scoped to the focused task run), an "Edit flow" deep link into the failing task's no-code edit tab (a new `editTask` query param), and "Ask Copilot" scoped to the error text through the existing `promptCopilot` entry point.

Accessibility: `role="region"` (not a modal), focus moves to the panel heading on open and back to the reopen affordance on close — never merely because the component mounted — `Escape` closes the panel (or clears an active selection first), and a separate `role="status" aria-live="polite"` announcer reports the failure whether or not the panel is open. The panel never renders for a task that failed then succeeded on retry (it filters on each task run's current state), and is suppressed for `PLAYGROUND` executions since the restart action only knows how to replay a persisted one.

### Two bugs found by actually running it

Recording the demo against a live instance surfaced two things unit tests could not:

- **`taskRun.outputs` is not serialized into the execution payload at all.** It is `@Deprecated(forRemoval)`, and `useTaskRunOutputs` says so in as many words; a real task run carries only `id`, `taskId`, `state` and `attempts`. The subflow drill-down originally read `taskRun.outputs.executionId` and so could never render. It now comes from the `/outputs/tasks/{executionId}/{taskRunId}` endpoint the panel already uses for upstream outputs. Note `TaskRunActions.vue` reads the same non-existent field for its own subflow link and is dead the same way — pre-existing, left alone here.
- **An attempt that started and ended in the same millisecond showed no duration**, because the value was gated on being truthy. It reports `0ms` now, matching the millisecond precision the state history already commits to.

The error text is also loaded only once the panel is opened. It used to load on mount, so every failed execution's Gantt view paid for a request whether or not anyone opened the panel.

## Evidence

- `FailureDebugPanel.spec.ts` — starts closed with the reopen affordance as the only entry point; focus management on open/close; `Escape`; earliest-first sort; single-vs-multi switcher; never renders on retry-success; announcement fires regardless of open state; identity by `value`; attempt counter; the error surfaced on its own and fetched once per focused failure; no fetch at all while closed and none again on reopen; error-dependent actions disabled when nothing was recorded while Edit flow stays enabled; subflow drill-down read from the outputs API and absent when there is no child execution; the flow fetched once per execution, not per switcher focus change.
- `FailureAttempts.spec.ts` — one row per attempt, numbered from 1 to match the log view's selector; the worker per attempt; omitted when none recorded; each attempt's own duration, including the sub-millisecond case.
- `FailureStateHistory.spec.ts`, `FailureResolvedConfig.spec.ts`, `FailureExecutionInputs.spec.ts`, `FailureUpstreamOutputs.spec.ts`, `FailureStructuralImpact.spec.ts`, `useTimeRangeSelection.spec.ts` — per-component behaviour, empty/loading/error states, drag-vs-click disambiguation.
- `ExpressionControllerTest.shouldRenderAgainstTaskRunWithTaskLocalBindings` — `taskrun.*` bindings resolve only once the render is scoped to a task run, proving the new `taskRunId` param does real work.
- Storybook stories (empty / single / multi-failure, light + dark) assert the identity badges, the isolated error, both attempt rows with their worker, tab switching, and a neighbour's raw YAML toggle — in a real browser.

## Test plan

- [x] `npm run check:types && npm run test:unit && npm run lint`
- [x] `npm run translations:check`
- [x] `npx vitest run --project=storybook tests/storybook/components/executions/gantt/FailureDebugPanel.stories.ts`

## Demo

Recorded against a running instance on a purpose-built failing execution: three simultaneous failures, one of them a task that retried twice on a worker, another a subflow.

Open the panel from the Gantt → switch to the failure that matters → read the error on its own → see which attempt failed and on which worker → the resolved configuration with its expressions rendered (`1482 rows for region eu-west-1`) → the inputs it ran on → then the subflow drill-down, which lands in the child execution on the real cause (`Ledger checksum mismatch`) where the parent only said "ends in FAILED state".

https://github.com/user-attachments/assets/dd38cb45-4592-4f7e-8e12-8b687b902ad7

Closes https://github.com/kestra-io/kestra/issues/19314
Closes https://github.com/kestra-io/kestra/issues/19390


ui/src/components/executions/gantt/FailureDebugPanel.vue
```diff
@@ -0,0 +1,562 @@
...
+        color: var(--ks-text-secondary);
+        font-size: var(--ks-font-size-xs);
+    }
+
+    // The mini-timeline gets its own full-width row rather than sharing a 3-column grid with the
+    // taller structural-impact/logs cards: it never grows to fill a stretched grid row (its own
+    // content doesn't expand), which used to leave a block of dead space under a couple of short
+    // bars. Full width also gives long task ids more room on the label side of the track.
+    .failure-debug-panel__timeline {
+        :deep(.kel-card__body) {
```
flcarre: 🟠 [design-system] New `:deep()` selector
Problem: `ui/AGENTS.md` golden rule #4 is "No `:deep()` selectors" — `KsCard` exposes no prop/slot for its body layout, so this reaches into `.kel-card__body` internals. It mirrors an existing pattern elsewhere in `Gantt.vue`, but that's pre-existing debt (rule #12: "copying an existing rule is not proof that it is correct"), not a reason to add more of it in a brand-new file.
Fix: wrap the card's own children in a local div and style that instead of the card internals, e.g.:
```suggestion
    .failure-debug-panel__timeline-body {
        display: flex;
        flex-direction: column;
        gap: var(--ks-spacing-2);
    }
```
(and change the template to `<KsCard ...><template #header>...</template><div class="failure-debug-panel__timeline-body">...content...</div></KsCard>`). Non-blocking given the existing precedent in the same file, but worth fixing since this is new code.

flcarre (commented):
Self-review pass before requesting team review. Ran `npm run check:types`, targeted `npm run test:unit`, `npm run lint`, `npm run translations:check`, and the new Storybook interaction tests locally — all green. See inline comments; summary below.

ui/src/components/executions/gantt/FailureStructuralImpact.vue
```diff
@@ -0,0 +1,120 @@
...
+    .structural-node__icon {
+        flex-shrink: 0;
+        display: inline-flex;
+        align-items: center;
+        justify-content: center;
+        width: 1.5rem;
+        height: 1.5rem;
+        padding: var(--ks-spacing-1);
+        border: 1px solid var(--ks-border-default);
+        border-radius: 0.5rem;
```
flcarre: 🟡 [design-system] Raw magic number instead of a radius token
Problem: `border-radius: 0.5rem` is a hardcoded value. `--ks-radius-base` already resolves to exactly `0.5rem` (`$border-radius: 0.5rem` in `variables.scss`), so this is a token that exists and isn't being used (golden rule #6).
Fix:
```suggestion
        border-radius: var(--ks-radius-base);
```

ui/src/composables/useTimeRangeSelection.ts
```diff
@@ -0,0 +1,81 @@
...
+    // Live pixel band while dragging, for the caller to render a preview before the gesture ends.
+    const dragBandPx = computed(() => {
+        if (dragStartX.value === undefined || dragCurrentX.value === undefined) return undefined
+        return {
+            from: Math.min(dragStartX.value, dragCurrentX.value),
+            to: Math.max(dragStartX.value, dragCurrentX.value),
+        }
+    })
+
+    function onPointerDown(event: {button: number; clientX: number}) {
```
flcarre, line 34: 🟠 [performance/correctness] Drag gesture has no pointer capture — can get stuck
Problem: `onPointerDown`/`onPointerMove`/`onPointerUp` are wired (in `FailureMiniTimeline.vue`) only to `pointerdown`/`pointermove`/`pointerup` on the track `<div>`, with no `setPointerCapture` and no window-level fallback. If the pointer leaves the track's bounding box mid-drag (plausible on a compact, multi-row timeline) and the button is released outside it, `onPointerUp`/`onTrackPointerUp` never fires: `dragStartX`/`dragCurrentX` stay set, so `dragBandPx` keeps returning a value and the blue selection-preview band in `FailureMiniTimeline.vue` stays rendered indefinitely, until the user starts a new drag (which silently overwrites the stuck state in `onPointerDown`). I traced this from the control flow rather than confirming it live in the browser — worth a quick manual check (start a drag inside the track, release the mouse button outside it) before merging.
Fix: capture the pointer on `pointerdown` (`trackRef.value?.setPointerCapture(event.pointerId)` in `FailureMiniTimeline.vue`, since the composable itself has no DOM reference) and add a `pointercancel`/`lostpointercapture` handler that clears `dragStartX`/`dragCurrentX` defensively. This touches both this composable and `FailureMiniTimeline.vue`'s template, so a suggestion block can't express it here.

ui/src/components/executions/Gantt.vue
```diff
@@ -9,173 +9,184 @@
...
+                                        </span>
+                                        <span class="separator">/</span>
+                                        <span class="item">
+                                            <span class="label">{{ $t("tasks") }}</span>
+                                            <span class="value">{{ tasksSummary }}</span>
+                                        </span>
+                                    </div>
+                                    <div class="actions">
+                                        <span v-if="debugShouldRender && !debugOpen" :ref="setReopenRef">
+                                            <KsButton :icon="BugOutline" @click="reopenDebugPanel">
```
flcarre: 🟢 [nit] Icon choice for "Debug this failure"
`BugOutline` isn't in the canonical icon map (`useLeftMenu.ts` / `panelDefinition.ts`) since "debug" has no existing concept there — flagging that this choice is inferred rather than reused, per the icon-reuse rule. Reasonable pick, not blocking.

ui/src/components/executions/gantt/FailureDebugPanel.vue
```diff
@@ -0,0 +1,562 @@
...
+        if (errors.length > 0) return errors.map((log) => log.message).join("\n")
+
+        return [...results].reverse().find((log) => (log.message ?? "").length > 0)?.message ?? ""
+    }
+
+    async function askCopilot() {
+        const taskRun = focusedTaskRun.value
+        if (!taskRun) return
+
+        const errorLines = await loadFocusedErrorText()
```
flcarre, line 501: 🟡 [suggestion] Silent empty-error path in "Ask Copilot"
Problem: `loadFocusedErrorText()` swallows the log-load failure (`.catch(() => [])`) and can resolve to `""`. `copyFocusedError` already guards on that (`if (!errorLines) return`), but `askCopilot` doesn't — it will happily open a Copilot thread with the prompt `"Fix the task X as it generated the following error:\n"` and no error text, with no feedback to the user that nothing was found.
Fix: short-circuit `askCopilot` the same way `copyFocusedError` does (e.g. a toast or just returning) when `errorLines` is empty.

flcarre:
## Kestra OSS Code Review

Scope: full cumulative diff of `feat/failure-debug-panel` against `develop` (26 files, UI-only). Backend/webserver/worker tracks are N/A — no Java files changed.

### Business Requirements — partial
The linked issue's acceptance criteria (#19314) state: *"Single failure: panel auto-opens and focuses that task; no failure switcher shown."* The implementation deliberately never auto-opens the panel — for single **or** multiple failures — and always starts closed behind the new "Debug this failure" button (`FailureDebugPanel.vue`'s `isOpen = ref(false)`, no auto-open branch). The PR body documents this as an intentional pivot with solid UX reasoning ("a failed execution is common enough that hijacking the Gantt view on every visit would be disruptive"), and the commit history (`rethink the failure debug panel's entry point`, `drop the reopen alert`) shows it was iterated on deliberately rather than overlooked.

That reasoning is convincing, but it's a direct, unacknowledged deviation from a specific, still-unchecked acceptance-criteria line in the issue, and there's no comment on #19314 recording the change of direction. Since this is a self-review before requesting team review: worth either (a) updating #19314's acceptance criteria to match the shipped behavior before requesting review, so reviewers aren't comparing the PR against a stale spec, or (b) explicitly flagging the deviation in the PR description's "Design decisions" section as a call-out rather than folding it into the general rationale (it already partly is — making it more prominent, e.g. "deviates from the issue's stated single-failure auto-open criterion" would remove any ambiguity for reviewers skimming the checklist).

All other acceptance criteria verified against the code: never renders for a task that failed-then-succeeded (checked against current state, not attempt history) ✅, switcher earliest-first / hidden for single failure ✅, drag-to-select + Escape-to-clear ✅, restart reuses `Restart.vue` (permission-gated via its own `hasPermission` check, not re-implemented) ✅, Ask Copilot scoped to focused task's error text ✅, `aria-live` announcement + focus management on open/close ✅, structural impact capped at 6 neighbors ✅, Storybook stories for empty/single/multi-failure with light+dark via the shared theme toolbar ✅, translations added to all 12 locales with a clean `translations:check` ✅.

### Kestra Guidelines (frontend, no `ui/` design-system violations found beyond noted) — 2 findings
- 🟠 New `:deep(.kel-card__body)` selector in `FailureDebugPanel.vue` (inline comment) — violates the "no `:deep()`" golden rule; `KsCard` has no body-styling prop/slot to reach for instead. Mirrors existing debt in `Gantt.vue` itself, but that's not a reason to add more in new code.
- 🟡 Raw `border-radius: 0.5rem` in `FailureStructuralImpact.vue` (inline comment) instead of the equivalent `var(--ks-radius-base)` token, which already resolves to the same value.

No `element-plus` imports, no hex/`rgb()`/`--el-*`/`--bs-*` colors, no raw px spacing, no i18n literal-`{{ }}` traps, no `any` added (baseline unchanged), types check clean, lint clean, `KsDataTable` pagination contract N/A (no data table here), no deep-watch-on-computed-spread pattern present.

### Frontend / Design System — 1 finding (see above, `:deep()`)
Everything else composes existing `Ks*` components (`KsCard`, `KsButton`, `KsIconButton`, `KsExecutionStatus`, `KsAlert`, `KsEmpty`) and `--ks-*` tokens correctly. Loading/empty/error/data states are all rendered in `FailureLogPanel.vue`. `useDiscardGuard` doesn't apply — the panel has no data-entry form to guard, just a read-only debug surface. Icons via `KsIcon`/`vue-material-design-icons`, no inline SVG. `BugOutline` for "Debug this failure" is an inferred icon choice (not in the canonical Flows/Apps/Tests/etc. concept map, since "debug" has no existing mapping) — noted inline as a nit, not blocking.

### PostHog analytics naming (surface.object.action) — N/A
No analytics events are emitted or renamed anywhere in this diff (checked `ui/src/components/executions/gantt/`, `Gantt.vue`, and the new composable for `posthog`/`capture`/`trackEvent` calls — none found). Nothing to check against the naming convention.

No new route, no new backend endpoint, no new data path: reuses the existing `loadLogs` store action (`taskRunId` filter, same as existing log affordances), the existing `Restart.vue` component (own permission gating), and the existing `promptCopilot` entry point (same mechanism as `Banner.vue`/`TaskRunActions.vue`'s "Fix with AI"). Ask Copilot / Edit flow are hidden (not just disabled) without `COPILOT`/`FLOW:UPDATE` permission via `v-if`, consistent with the "hidden, not disabled" acceptance criterion.

### Performance — 1 finding
- 🟠 Missing pointer capture in the mini-timeline's drag-to-select gesture (`useTimeRangeSelection.ts` + `FailureMiniTimeline.vue`, inline comment) — traced from the control flow (no `setPointerCapture`, no window-level fallback, no `pointercancel` handling); releasing the pointer outside the track div mid-drag leaves the selection-preview band stuck rendered until the next drag starts. **Needs live browser verification** — I did not have interactive access to confirm this in a running instance, flagging it as a traced-but-unconfirmed finding rather than an asserted bug.

One non-blocking suggestion (inline): `askCopilot` doesn't guard against an empty error string the way `copyFocusedError` does, so a failed/empty log load can open a Copilot thread with no actual error content.

**Verdict:** REQUEST CHANGES

Rationale: nothing here is a correctness/security blocker, and the build/type/lint/test/translation gates are all green — but the auto-open acceptance-criteria deviation should get an explicit decision (update the issue or make the call-out unmissable in the PR description) before this goes to team review, and the `:deep()` selector + missing pointer-capture handling are worth a quick fix first since both are new code in this PR rather than inherited debt.

ui/src/components/executions/gantt/FailureDebugPanel.vue
flcarre: Fixed in 42c50e3ea0. The card body no longer reaches into `.kel-card__body`; its children are wrapped in a local `.failure-debug-panel__timeline-body` div and that is what carries the layout. Verified on the current head: the cumulative diff against `develop` adds zero `:deep(` selectors across `ui/src`.

ui/src/components/executions/gantt/FailureStructuralImpact.vue
flcarre: Fixed. `border-radius` in this file now uses `var(--ks-radius-base)` at both call sites (lines 101 and 143), no raw `0.5rem` left.

ui/src/components/executions/gantt/FailureDebugPanel.vue
flcarre: Fixed, and taken a step further than the suggestion. `askCopilot` no longer swallows the empty case silently: the error text is now loaded once per focused failure into reactive state, and both error-dependent actions carry `:disabled="!focusedErrorText"`, so Ask Copilot and Copy error are visibly unavailable rather than inert when nothing was recorded. `Edit flow` deliberately stays enabled, since fixing a flow does not depend on an error message existing — there is a unit test pinning exactly that asymmetry.

ui/src/composables/useTimeRangeSelection.ts
flcarre: The code fix is in: `FailureMiniTimeline.vue` calls `trackRef.value?.setPointerCapture(event.pointerId)` on `pointerdown` and wires `@pointercancel` / `@lostpointercapture` to a `cancelDrag()` that clears `dragStartX` / `dragCurrentX` (42c50e3ea0).

The manual check this comment asked for is still not done, and I want to be straight about why rather than quietly resolve this.

I tried to automate it against a running instance: open the panel, press inside `.mini-timeline__track`, drag, then release the button far outside the track, and assert no preview band survives. It passed — but it also passed with `setPointerCapture` and both handlers stripped out, so it was not discriminating. Adding a positive control showed why: `.mini-timeline__band` never appears at all under synthetic Playwright mouse input, not even mid-drag inside the track. The harness is not exercising the gesture, so the run proves nothing in either direction.

So: fixed by construction and by code reading, not confirmed live. Worth one real mouse-and-hand check before merge.

PR #19320, opened 2026-09-10
feat(ui): inform users about potential breaking changes after an upgrade

### LINKED ISSUE #19214: [UI] Inform users about potential breaking changes after migration to a new version
### Why

Users upgrading to a 2.0 might not be aware of the breaking changes to their existing flows, even after a successful technical migration.

### What

After a minor or major version change, show a notification in the UI once per user with a single link to what changed.

- Trigger on a change of the major or minor version (1.3 to 2.0, 2.0 to 2.1). Nothing on patch bumps like 2.0.0 to 2.0.1.
- Show it once per user. Once marked as read, it stays dismissed for that user, and marking it as read does not hide it for other users on the same instance.
- Keep it light. It goes to everyone, including users with nothing to migrate, so it must not read as a marketing popup.
- One call to action and one link only, pointing to the migration guide for the new version. Use an indirect link with a redirect we control, so the target can move when the website structure changes.
- It has to work in OSS. This is where the problem is and where most of the affected users are.
- Wording, as a heading, one line of body and one CTA:

> **Kestra was upgraded to 2.0.0**
>
> This is a new release and it may include breaking changes that require updates to your flows.
>
> **[See what changed](https://kestra.io/docs/migration-guide/v2.0.0)**

Say "may include" rather than asserting breaking changes exist, because the same notification fires on a minor bump like 2.0 to 2.1 where usually there are none.

### Acceptance criteria

- Upgrading from 1.3.x to 2.0.x shows the notification to each user on their next visit.
- Upgrading from 2.0.0 to 2.0.1 shows nothing.
- A fresh install with no previous version recorded shows nothing.
- Dismissal is per user and persists.
- The link goes through a redirect, not a hardcoded docs URL.

AcevedoR:
Closes #19214

After a major or minor version change, a dismissible single-line banner links to the migration guide for the new version. `VersionService` already read the previous instance version just before overwriting it at startup; that transition is now kept as `instance.previous-version` and exposed on `/configs` as `versionUpgrade`.

Dismissal lives in localStorage, keyed by the account id the EE auth-store override provides, so one user dismissing it never hides it for another. OSS has a single user and collapses onto one key.

### What a user sees

| Situation | Result |
|---|---|
| Fresh install, first login | Nothing — no previous version was ever recorded |
| 1.3.4 → 2.0.x, first login after the upgrade | Banner: "Kestra was upgraded to 2.0.x…" |
| Same user, second login | Nothing — dismissal persists |
| Another user on the same instance, first login | Banner — dismissal is per user |
| Same user, different browser | Banner again (localStorage is per browser) |
| 2.0.0 → 2.0.1 (patch) | Nothing |
| 2.0.5 → 2.1.0 (minor) | Banner, linking to the v2.1.0 guide |
| First login more than 30 days after the upgrade | Nothing |
| Downgrade, or a dev build reporting `Snapshot` | Nothing |

The CTA takes the major and minor of the running version and pins the patch to `.0`, so every patch of a line points at that line's guide. Verified against real builds rather than only unit tests: a 2.0.5 build upgraded from 1.3.4 renders the banner and links to `v2.0.0`; a 2.1.0-SNAPSHOT build upgraded from 2.0.5 links to `v2.1.0`; a patch-only bump renders nothing.

### Deviations from the issue, for the record

- **Direct docs URL, not a redirect.** The issue asked for an indirect link we control so the target can move; this links to `kestra.io/docs/migration-guide/v<major>.<minor>.0`.
- **Single line, not heading + body + CTA.** The two-tier version read heavier than an outage alert, so it now matches the shape of the kill-switch and maintenance banners.
- **30-day relevance window**, so someone joining months later is not shown a migration that never concerned them.

Companion PR: kestra-io/kestra-ee#10967 (same branch name).


AcevedoR:
here is what it looks like 
image

nkwiatkowski (approved):

core/src/main/java/io/kestra/core/services/VersionService.java
```diff
@@ -21,6 +26,15 @@
 public class VersionService {
     private static final String MIN_VERSION = "1.0.0";
```
loicmathieu:
```suggestion
    private static final String MIN_VERSION = "1.3.0";
```

Please upgrade to 1.3.0, it was a mistake to keep 1.0.0

loicmathieu (approved):
LGTM, please change the MIN_VERSION before merge.

PR #19321, opened 2026-09-10
build(tests): count `any` in templates too

Piyush-r-bhaskar:
EE PR: https://github.com/kestra-io/kestra-ee/pull/10968

## What

oxlint only reads the `<script>` block, so an `any` written in a template was invisible to `check:ts-any`. The script now parses every `.vue` file with the Vue SFC compiler and counts the `any` in interpolations and directive values, which adds 121 across the two repos. Parsing rather than grepping is what keeps prose and static attributes out of the count.

The second commit closes the way around the check. `--write` used to raise a count silently, and the failure message told you to run it, so hiding an `any` was easier than fixing one. Now `--write` only ever lowers, raising takes `--accept-new-any` alongside it, and the message says to type the value. `--lock` is gone, since it had become the same command as `--write`, and both pre-commit hooks call `--write` instead.

Merge this first, then kestra-io/kestra-ee#10968 immediately after: until the EE baseline lands, every EE PR is red on `check:ts-any`.


MilosPaunovic (commented):
Nice one, using the real `SFC` parser instead of regexing the raw file keeps prose and static attributes safe, and both baselines reproduce byte-exactly against the new script (`1752` in `433` files for `OSS`, `1044` in `246` files for `EE`). I dumped every template expression the new regex matches across both repos (120 hits) and all of them are genuine `any` usages, so no false positives on the current tree.

Two edges of the detector are wrong in ways the spec does not cover, both a few lines plus a test case each:

1. **Multi-`any` generics are undercounted** (`baseline.mjs:9`). `,\s*any\s*>` only ever matches the last type argument and `<\s*any\s*>` needs `any` to be the only one, so `:x="v as Map<any, any>"` counts `1`. On a ratchet that is a silent loosening: a file baselined at `1` for `Record<string, any>` still passes after gaining a second `any`. Counting the `any` token itself (or adding `<\s*any\s*,` and `,\s*any\s*,` alternatives) fixes it.
2. **String literals inside expressions are scanned** (`baseline.mjs:42`). `:title="'Type: any'"` and `{{ $t('accepts: any value') }}` each count `1`. An innocent tooltip string would red-line a `PR`, and the failure message then recommends `--write`, which permanently inflates the baseline. Blanking out string and template literals before matching (same-length whitespace) avoids it. The existing spec only covers static attributes and a `$t` argument without a colon, so a case for this would be good.

Nits, take or leave:

- `node.type === 5` / `prop.type !== 7`: `@vue/compiler-sfc` re-exports `NodeTypes`, so `NodeTypes.INTERPOLATION` / `NodeTypes.DIRECTIVE` cost nothing.
- The dynamic directive argument branch (`prop.arg?.isStatic === false`) cannot fire for `any` detection: Vue rejects spaces inside `:[k as any]` as a parse error, and a well-formed `:[k]` never matches. Either drop it or comment it as defensive.
- `parse()` returns an `errors` array that is never read, so a template that fails to parse silently contributes a partial count and reads as "improved". Failing loudly like `lint()` does would keep the tooling honest.
- `@vue/compiler-sfc` is required by the script and the spec but not declared in `ui/package.json` or `ui-ee/package.json`; it resolves only through `vue` hoisting. Worth adding to `devDependencies` in both.
- `vueFiles` calls `readdirSync` unguarded, so a missing path throws a raw `ENOENT` instead of the friendly exit used for `oxlint` failures.
- `templateExpressions` is exported but only used internally and not tested. Either test it directly or keep it module-private.
- A short `## What` in both descriptions, with the merge order below, would save the next person the archaeology.

Merge order: this one first, then https://github.com/kestra-io/kestra-ee/pull/10968 immediately after. Until the `EE` baseline lands, every `EE` `PR` is red on `check:ts-any` with a message telling authors to lower the baseline back, so the two merges should be back to back. No `kestra-io/actions` change needed, the `EE` frontend job does a full `OSS` checkout.

MilosPaunovic (commented):
Round 2. Everything from the first pass is closed, and I verified each one against the head (`Map<any, any>` -> 2, `'Type: any'` -> 0, parse errors exit 2, `@vue/compiler-sfc` declared in both repos, both baselines reproduce exactly, spec 11/11, `oxlint` and `eslint` clean). Thanks for the thorough pass.

The second commit, making `--write` refuse a raise, introduces a few new things worth a look:

1. **A plain file rename now hard-fails the check and the pre-commit hook** (`baseline.mjs:170`). `compare()` is per-file, so a move is "old entry removed, new entry added", and the `added` branch fires before `--write` is ever reached. Reproduced with `git mv src/components/ErrorToast.vue src/components/ZzRenamed.vue` and zero code changes:

   ```
   New `any` in 1 file(s). Type the value, do not raise the baseline.
     src/components/ZzRenamed.vue: 0 -> 2
   ```

   Same output with `--write`, which is what `.github/.hooks/pre-commit:45` runs. The two obvious remedies (hand-editing `baseline.json`, running `--accept-new-any`) are both forbidden by the `AGENTS.md` change in the same commit, and the message tells the contributor to "type the value" in a file they only moved. Suggestion: let `--write` through when the total does not increase (`sum(current) <= sum(baseline)`), or match an added file's count against a file that vanished entirely. At minimum, document the rename case and the command that resolves it.

2. **`withoutStrings` mis-parses a string ending in an escaped backslash** (`baseline.mjs:55`). `expression[i - 1] !== "\\"` looks one character back, so `'a\\'` never closes and the rest of the expression is blanked: `:x="'a\\' + (v as any)"` counts 0, should be 1. Same silent-loosening class as round 1. One line at the top of the in-quote branch fixes it: on `\\`, blank two characters and `i++`. No occurrence in either tree today (checked against a reference lexer over all 848 `.vue` files), so latent rather than active.

3. **The new gate has no test.** It lives entirely in `main()`, which is not exported, so nothing covers `--write` refusing a raise, `--write --accept-new-any` accepting one, `--accept-new-any` alone, or the rename case above. The two new spec cases only cover the counting changes. Extracting a small `decide({added, removed, write, acceptNewAny})` helper would make this cheap to test.

Nits, take or leave:

- `${...}` scanning uses `indexOf("}")` (`baseline.mjs:48`), so a nested object or block inside the interpolation ends it early. A brace-depth counter fixes it.
- `Raised the baseline for ${added.length || removed.length} file(s)` (`baseline.mjs:191`) under-reports when both directions moved, while the list below prints both. `added.length + removed.length`.
- `exit 2` fires on any entry in `parse().errors` (`baseline.mjs:66`), including recoverable ones (`Duplicate attribute.`, `Element is missing end tag.`) where a usable `ast` is still returned and the component builds today. Gating on `!descriptor.template?.ast` would be tighter. Nothing in-tree trips it.
- `ui/AGENTS.md:251` only forbids `--accept-new-any`. One line with the maintainer command (`--write --accept-new-any`) would save the next person a trip into the script.
- `withoutStrings` is exported but only used inside the module and not imported by the spec. Same as the round-1 `templateExpressions` remark.

One consequence for the merge: in the window between this landing and https://github.com/kestra-io/kestra-ee/pull/10968 landing, `EE` `PRs` fail on the added side, which `--write` now refuses, so `EE` contributors have no local workaround at all until the `EE` baseline is in. The descriptions already say "immediately after", just flagging that it matters more now.

Piyush-r-bhaskar:
All four done @MilosPaunovic  .

Rename: did it differently than suggested. Going by the total would also let someone add an `any` in one file and remove one elsewhere, so `compare()` now matches a file that appeared with one that disappeared at the same count and calls it a move. A plain `git mv` just asks for `--write` now. A move that also adds an `any` still fails.

Escaped backslash fixed, and the gate is its own `decide()` with tests per flag combination. All nits in too.

One from round 1 I couldn't do: `@vue/compiler-sfc` doesn't export `NodeTypes` in 3.5.39, only `compiler-core` does and it isn't a declared dep. Named constants with a comment instead.

Baselines unchanged, 11/11, lint and types clean.

MilosPaunovic (commented):
Round 3. Everything from the previous two rounds is closed and verified, including the rename case: a plain move now reports `moved from …` and `--write` records it. The move detection it added has two gaps of its own.

1. **A rename that also removes an `any` is rejected, and the message says the opposite of what happened** (`baseline.mjs:107`). The pairing requires `was === arrival.now`, so a move that improves the file falls through to `added`. Renaming a component and typing one of its two `any` in the same commit:

   ```
   New `any` in 1 file(s). Type the value, do not raise the baseline.
     src/components/ErrorBanner.vue: 0 -> 1
   ```

   Exit 1, and `--write` refuses and leaves `baseline.json` untouched, so the pre-commit hook blocks the commit. The only sanctioned escape is `--write --accept-new-any`, which `AGENTS.md` forbids. Splitting it into two commits works but nothing says so. Matching a departure whose count is greater than or equal to the arrival's fixes it, and `Old: 2 -> New: 3` still lands in `added` where it belongs.

2. **The pairing is count-only, so fresh `any` in a brand-new file rides in on an unrelated deletion** (`baseline.mjs:107`). Deleting a component with 2 `any` and adding a new file with 2 fresh ones:

   ```
   Updated the baseline for 1 file(s), now 1645 explicit any in 429 files:
     src/components/ZzBrandNew.vue: moved from src/components/basicauth/BasicAuthLogin.vue, still 2
   ```

   `--write` alone records it, exit 0, no `--accept-new-any` needed. Counts here are small integers, so a collision is likely rather than exotic, and the line asserts a move that did not happen. Worth noting this pulls against the fix above: relaxing the count test widens the hole, so the pairing probably needs a second signal, for example an identical basename, with anything else left to the maintainer flag.

Nits:

- The annotation for a moved file reads `explicit any went from 2 to 2`, since the `moved` entry sets `was` to the new count. Reads oddly on the `PR`.
- `readFileSync(baselinePath, …)` (`baseline.mjs:205`) is unguarded, so a missing `baseline.json` throws a raw `ENOENT` stack rather than the friendly exit the `readdirSync` walk now gets.

On the red `Frontend - Tests`: all 2287 tests pass and the step exits 1 on six unhandled `EnvironmentTeardownError` rejections from a design-system locale import in `inputsFormBoolDefault.spec.ts`. I ran the same suite on `origin/develop` and it fails the same way, so it is not this `PR`. A re-run should do it.

Otherwise verified against this head: `OSS` baseline reproduces at 1645 in 429 files, `EE` at 1044 in 246 against the new script, spec 11/11, `oxlint` and `eslint` clean, and every counting case from the last two rounds is correct, escaped backslashes and nested `${}` braces included. Merge order is unchanged, this one first and https://github.com/kestra-io/kestra-ee/pull/10968 straight after, and note the `EE` green predates this force-push.

Piyush-r-bhaskar:
Both cases reproduced,Thank You 🙏. The count-based pairing was the problem, so it is gone: the script now asks git which files were renamed (`git status -M`, staged renames) and only pairs those. Rename and fix an `any` on the way: reported as moved, `2 -> 1`, `--write` follows. Delete a file and add an unrelated one with the same count: reported as new `any` and refused. A rename that gained one is still new `any`. An unstaged rename shows as removed plus new until it is staged, which the pre-commit hook always is.

MilosPaunovic (commented):
Round 4. Asking `git` for the renames instead of pairing on equal counts is a better answer than the one I suggested, and it closes both gaps from last round: a move that also cleans up an `any` is recorded as `2 -> 1`, a move that gains one is still refused, and an unrelated deletion no longer absorbs a brand-new file. `Frontend - Tests` is green on this head too, so the red last round was indeed the `develop`-side teardown flake.

One case is left, and it is the same dead end narrowed once more: the lookup only sees **staged** renames, so the moment the rename is committed it becomes invisible. Committing the rename and then running exactly what the failure message names:

```
$ git commit -m "chore: rename"        # no hook installed
$ npm run check:ts-any -- --write
New `any` in 1 file(s). Type the value, do not raise the baseline.
  src/components/ErrorBanner.vue: 0 -> 2
```

Exit 1, baseline untouched, and the only thing that gets past it is `--write --accept-new-any`, which `AGENTS.md` forbids. The hook makes this invisible, but the hook is opt-in through `setup_hooks.sh`, so it bites anyone who has not run it, anyone who used `--no-verify`, and anyone fixing up after a red `CI` run. Coding agents are the most likely to land here, and they are the audience that section of `AGENTS.md` is written for. Consulting the branch's committed renames as well, something like `git diff --name-status -M $(git merge-base HEAD origin/develop)...HEAD` behind the same try/catch, would close it; `CI` itself never needs it, since the committed baseline already reflects the move.

Nits:

- A `git`-reported rename that gains an `any` prints `src/components/ErrorBanner.vue: 0 -> 3` (`baseline.mjs:227`). It really went `2 -> 3`; the arrival keeps `was: 0` because only `moved` entries carry the departure's count. The instruction is right, the number is not.
- `stagedRenames` strips the surrounding quotes from a quoted path but leaves the C-style escapes inside it (`baseline.mjs:210`). Only reachable with a non-ASCII or space-bearing filename, of which this tree has none, so purely a note.

Verified on this head: `OSS` baseline reproduces at 1645 in 429 files, `EE` at 1044 in 246 against the same script, spec 12/12, `oxlint` and `eslint` clean, and the staged rename, rename-with-cleanup, rename-with-a-new-any and unrelated-deletion cases all behave. Merge order unchanged, this one first and https://github.com/kestra-io/kestra-ee/pull/10968 straight after.

MilosPaunovic (approved):
This is thorough, and it went past what I raised: `-z` instead of parsing `" -> "`, the arrival now carrying the departure's count so a move that gains an `any` reads `moved from …, 2 -> 3`, `AGENTS.md` updated, and a real `git` integration test with a spaced path and a `README.md` decoy for a walker that scans for `R`. Approving.

Two nits, neither worth holding this up:

- A rename chained across a commit boundary is not followed. `A -> B` committed, then `B -> C` staged, gives ``New `any` in 1 file(s)`` with `src/components/ErrorC.vue: 0 -> 2` and `--write` refuses: the committed walker knows `A -> B`, the staged one `B -> C`, and nothing maps `C` back to `A`. Only reachable without the hooks installed, so it is an edge of an edge. Worth mentioning mostly because one call replaces both walkers and handles the chain in one step, `git diff --name-status -M -z "$(git merge-base origin/develop HEAD)"`, which reports `ErrorToast.vue -> ErrorC.vue` directly. The three-dot form is what makes the two-walker split necessary, and the merge-base form avoids the artefacts a plain two-point diff against `origin/develop` would bring.
- `origin/develop` is hardcoded and the committed pass degrades to nothing when the ref is absent. That is the right call for `CI`, where the checkout is shallow and the committed baseline already reflects the move, but it also means a branch based on anything else silently loses the feature. A `@{upstream}` fallback, or just a line in the comment saying the ref is expected to be missing in `CI`, would make that intentional rather than incidental.

`Frontend - Tests` is red with the same signature as round 3: all 2298 tests pass, this `PR`'s own spec passes 13/13, and the step exits 1 on six unhandled `EnvironmentTeardownError` rejections from a design-system locale import in `inputsFormBoolDefault.spec.ts`. The base commit fails the suite the same way, so it is not this `PR`. A re-run or a rebase once `develop` settles should clear it.

Verified on this head: the committed rename is followed and recorded by `--write`, a committed rename that gains an `any` is refused with the real previous count, an unrelated deletion no longer absorbs a new file, `OSS` reproduces at 1645 in 429 files and `EE` at 1044 in 246 against the same script, `oxlint` and `eslint` clean. Merge order unchanged: this one first, https://github.com/kestra-io/kestra-ee/pull/10968 straight after.

Issue #19326, opened 2026-09-10
fix(executions): restart, resume and force-run skip the disabled check


loicmathieu:
## Description

`getFlowIfExecutableOrThrow` (the resolver behind create-execution and replay) rejects a disabled flow. The restart, resume, and force-run paths resolve the flow through a different route entirely and never check `isDisabled()`:

- `ExecutionService.restart` — `flowRepositoryInterface.findByExecutionWithoutAcl(execution)`
- resume (including `/resume/validate`, `/resume/by-ids`, `/resume/by-query`) — `findByExecutionWithoutAcl`
- `ExecutionService.forceRun` — `flowRepositoryInterface.findByExecution(execution)`

None of these guard on `isDisabled()`, so a flow an operator has disabled can still be restarted, resumed, or force-run. This is the same bug class as `GHSA-g25f-25m4-3hhw` ("Disabled webhook trigger remains remotely executable"), which Kestra rated **high**.

## Open question

Design intent is unclear for at least `restart`/`resume`: finishing work already in flight for a flow that was disabled *after* the execution started may be deliberate (that's plausibly why `findByExecutionWithoutAcl`'s javadoc explicitly says it exists "to restart/resume a flow"). `force-run`, which starts new task runs on demand, is harder to defend disregarding `disabled` for.

Needs a decision on which of the three should gain the check before implementing.

## Evidence

Found while reviewing and fixing a related soft-delete execution-bypass advisory (GHSA-52wv-cgfg-4j6x); this is a distinct, sibling gap on the same resolver family, not part of that fix.

hemantmm:
Hi @loicmathieu, I’d like to confirm the implementation approach for the disabled-flow execution issue before starting.

The current `force-run`, `restart`, and `resume` operations can bypass the flow’s `isDisabled()` check.

**My proposed approach:**

* Apply the `isDisabled()` check to all three operations: `force-run`, `restart`, and `resume`.
* Use `flowService.getFlowIfExecutableOrThrow()` or an equivalent latest-flow-state validation before dispatching execution commands.
* Add backend tests in `ExecutionControllerTest` covering disabled flows for each operation.
* Verify that execution commands are not dispatched when the flow is disabled.

The open question is whether `restart` and `resume` should be allowed to finish work that was already in progress when the flow was disabled, or whether all three operations should consistently respect the disabled state.

PR #19327, opened 2026-09-10
fix(flows): reject execution of soft-deleted flow revisions

loicmathieu:
### 🔗 Related Issue

No linked issue — found and fixed while reviewing an unpublished security advisory (GHSA-52wv-cgfg-4j6x). One follow-up filed separately, not addressed here: https://github.com/kestra-io/kestra/issues/19326.

Fixes https://github.com/kestra-io/kestra-ee/issues/10869

### ✨ Description

A soft-deleted flow (`deleted = true` on its latest revision) could still be executed in two ways:

- `POST /executions/{ns}/{id}?revision=<deleted>` resolved and ran the deleted revision directly: `getFlowIfExecutableOrThrow` checked disabled and invalid but never checked `isDeleted()`.
- Deleting a flow whose head was a **draft** left the tombstone itself flagged as a draft. The ranking used to resolve an execution without an explicit revision (`lastNonDraftRevision`) excludes draft revisions before applying the deleted filter, so the draft tombstone was invisible to it and the published revision underneath it resurfaced as executable — with **no revision parameter needed**.

Fix, in two commits:

1. `getFlowIfExecutableOrThrow` now rejects a resolved flow that `isDeleted()`, with the same "not found" message and status as a nonexistent flow — so an execute-only caller can't distinguish a deleted flow from one that never existed.
2. `Flow#toDeleted()` / `FlowWithSource#toDeleted()` now clear `draft` on the tombstone. A delete tombstone isn't a draft of anything: it's the flow's end state, and always the highest revision number, so once it's never draft-excluded from the ranking it correctly wins and gets filtered out — instead of the ranking falling through to the revision beneath it.

An adjacent existence leak is fixed alongside commit 1: a deleted flow whose tombstone had inherited `draft = true` (pre-fix behavior) made the no-revision execution path answer `422 "only has draft revisions"` instead of `404`, confirming the flow's existence to a caller who shouldn't be able to tell.

### 🛠️ Backend Checklist

- [x] Code compiles and tests pass for the touched modules (`:core:test`, `:jdbc-h2:test`, `:webserver:test`)

### 📝 Additional Notes

- 1.3.x is separately affected and needs its own backport — no `draft` column exists on that line, so this exact patch doesn't apply as-is.

### DIFF
--- a/core/src/main/java/io/kestra/core/models/flows/Flow.java
+++ b/core/src/main/java/io/kestra/core/models/flows/Flow.java
@@ -308,6 +308,7 @@ public class Flow extends AbstractFlow implements HasUID {
         return this.toBuilder()
             .revision(this.revision + 1)
             .deleted(true)
+            .draft(false) // switch to false to avoid resurrecting the previous revision
             .build();
     }
 
--- a/core/src/main/java/io/kestra/core/models/flows/FlowWithSource.java
+++ b/core/src/main/java/io/kestra/core/models/flows/FlowWithSource.java
@@ -54,6 +54,7 @@ public class FlowWithSource extends Flow {
         return this.toBuilder()
             .revision(this.revision + 1)
             .deleted(true)
+            .draft(false) // switch to false to avoid resurrecting the previous revision
             .build();
     }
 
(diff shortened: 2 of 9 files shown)


loicmathieu:
1.3.x backport: kestra-io/kestra#19328

nkwiatkowski (approved):

loicmathieu:
Backported to 2.0 and 1.3 (#19328)

PR #19330, opened 2026-09-10
feat(ai): show a diff before applying a Copilot proposed change

### LINKED ISSUE #19329: AI Copilot: show a diff before applying a proposed change
## Summary

The flow-editor AI Copilot lets users approve a proposed action ("Approve & execute") without ever seeing the YAML change it is about to make, and applying a drafted flow over an existing one goes through a generic confirm dialog with no diff either. Users currently approve a mutation on a shared flow file based on a text summary alone.

## Scope / Design

- `ui/src/components/ai/copilot/ProposedActionCard.vue`: `argEntries` only surfaces short scalar arguments (e.g. namespace/id) — the actual YAML diff of a mutating action is never rendered before approval.
- `ui/src/components/ai/copilot/useApplyDraft.ts`: applying a draft over an existing flow goes through a generic `KsMessageBox.confirm` (i18n key `ai.copilot.draft.applyConfirm`) with no diff shown.
- Add an inline unified diff view to `ProposedActionCard` (and replace the generic confirm modal for the draft-apply-on-existing-flow path) using `+`/`-` line markers — not color alone, for accessibility — plus a compact `+N/−N` summary badge.
- Design-system only: reuse existing `Ks*` components and `--ks-*` tokens, no new hex colors.

## Acceptance Criteria

- [ ] A proposed mutating action shows an inline diff of the change before the user can approve it
- [ ] Applying a draft over an existing flow shows the same diff instead of the generic confirm dialog
- [ ] Diff additions/removals are distinguishable without relying on color alone
- [ ] Tests added/updated (Vitest + Storybook story covering the diff states, light and dark)
- [ ] `./gradlew build` passes

## Context

Related to the flow-editor AI Copilot chat (`ui/src/components/ai/copilot/`).

## Out of Scope

- Error-recovery UX for failed generations (separate follow-up)
- Tool-call activity narration (separate follow-up)

flcarre:
Closes https://github.com/kestra-io/kestra-ee/issues/10992.

## Summary

- `ProposedActionCard` renders a diff for a mutating action that carries a YAML/source body argument (`body`/`source`/`yaml`), instead of silently dropping it as "too verbose to show inline". There is no fallback onto "the sole overlong argument": a tool can carry a long argument that isn't the proposal itself (e.g. a `currentFlowYaml`-style "before" argument alongside the proposed one), and guessing wrong there would render a confidently backwards diff — no diff is safer than a wrong one.
- `useApplyDraft`'s flow-apply confirm shows the same diff, inside a widened `KsMessageBox` (the default ~420px box was too narrow for a code editor), instead of the generic `KsMessageBox.confirm` text prompt.
- Diff rendering goes through `KsEditor`'s built-in Monaco diff mode — the same one `Revisions.vue` already uses to compare flow revisions — instead of the bespoke line-diff renderer this PR originally shipped with, which turned out too small/unreadable in the confirm modal. `DiffView.vue` is now a thin wrapper around `KsEditor` (`oldValue`/`newValue` → `original`/`modelValue`), and the hand-rolled `textDiff.ts` LCS diff is removed as dead code.
- A past (resolved) `PROPOSED_ACTION` message in the transcript no longer mounts its own diff editor — only the still-pending card does.

**Where the "before" YAML comes from** (unchanged from the original diff-preview logic):
- `ProposedActionCard`: `CopilotChat.vue` only passes a `currentFlowSource` prop when the pending action's own `namespace`/`flowId` (or `id`) arguments match the flow currently focused in the route — otherwise the diff renders the proposed content as a pure addition, rather than risk showing a diff against the wrong flow.
- `useApplyDraft`: the live editor buffer (`flowStore.flowYaml`) when the target flow is already open (reflecting unsaved edits), otherwise a `store: false` fetch of its persisted source. A fetch failure (or a brand-new flow, i.e. nothing persisted yet — the primary path, since applying a just-drafted flow is the common case) falls back to an empty "before", so the drafted content renders as an addition rather than blocking the apply; that fetch also no longer raises a global "Not Found" error toast for the brand-new-flow case, since it's an expected outcome the caller already handles.

## Follow-up: mirror the diff into the main editor, save applied flows as drafts

The diff above is still chat-only, which only helps when the flow you're asking about isn't the one you have open. When it *is* the currently open flow, the main "Flow Code" editor now mirrors the same diff live via `flowStore.previewSource` (already wired into `FlowFileEditorTab.vue`'s `KsEditor` diff mode from a past feature, just never populated until now) — so it reads as an in-IDE diff, not only a chat aside. The chat panel's own diff/Approve/Reject/Apply buttons are unchanged and remain the only actions; the editor just visually mirrors what's pending, and goes back to normal (editable, unsaved edits intact) once the preview clears.

Wired for both:
- the artefact-draft card (`CopilotArtefactDraft.vue` / `useApplyDraft.ts`'s `apply` path) — the most recent `FLOW`-kind draft targeting the open flow;
- the pending mutate-confirmation path (`ProposedActionCard`'s `currentFlowSource`).

Both reuse `useApplyDraft.ts`'s `isViewingFlow`/`parseArtefactYaml` (now exported) so there's a single "is this targeting the open flow" check instead of two that could drift apart. The preview clears on approve/reject, when a new turn starts, when superseded by a newer draft for the same flow, on "New chat", and on unmount (closing the copilot dock).

Approving/applying a Copilot flow proposal now saves it as a **draft revision** (`draft=true` on both the create and update calls) instead of going live immediately — since the primary interaction model stays chat-first, the user reviews and publishes the draft explicitly rather than a Copilot change going live unattended. `Revisions.vue`'s existing draft badge already surfaces this correctly; no new "you have a pending draft" UI was added. The confirm dialog copy (`ai.copilot.draft.applyConfirm`) was updated to say so, translated into all languages.

I checked `webserver/.../services/ai/agent/tool/` for a backend tool that persists a flow change through the `ProposedActionCard`/mutating-tool-call path (as opposed to `AuthorFlowTool`, which only ever produces a non-persisted `ArtefactDraft`): none exists today in OSS or EE — `RestartExecutionTool` is the only tool with `AgentWritePolicy.CONFIRM`, and its arguments are short ids, never a `body`/`source`/`yaml`. So the `draft=true` treatment only applies to the `useApplyDraft.ts` path for now; the mutate-confirmation path's diff/draft-saving is forward-looking, not yet exercisable by CI or manual testing until a confirmable, flow-mutating tool ships.

Also: the `.kel-message-box__message { flex: 1; min-width: 0 }` rule added earlier in this PR is a design-system-wide change (every `KsMessageBox` in OSS and EE), with no dedicated DS story/test. It's a fix for this PR's own widened confirm dialog and reads correct in what I've checked, but is worth a second look on other, busier confirm dialogs across the app.

## Demo

https://github.com/user-attachments/assets/68ff9f14-d4f1-4505-bf1d-759a295898c2

## Test plan

- [x] `npx vitest run tests/unit/components/ai/copilot` — all passing
- [x] `npm run check:types`, `npm run test:lint`, `npm run translations:check` — all clean
- [x] Manually verified in Storybook (real Monaco, real `KsMessageBox`) that the confirm modal renders a readable, scrollable, properly-sized diff in both light and dark mode
- [x] Verified end-to-end against a running backend that `POST/PUT .../flows?draft=true` (the exact call `useApplyDraft.ts` now makes) creates/updates a flow revision with `"draft": true`, and that fetching a not-yet-created flow 404s as expected (the case `persistedFlowSource`'s `ignoreNotFound` now handles silently)

closes: https://github.com/kestra-io/kestra/issues/19329


flcarre:
## QA report — PR #19330

**Verdict: PASS**

Tested the shipped code directly (backend `:cli:runLocal` + `ui/` Vite dev server booted from the `feat/copilot-diff-preview` worktree, embedded H2, deterministic local admin). No live AI provider was configured in this environment, so the AI Copilot's `/ai/threads/*` SSE endpoints were mocked at the `fetch` layer with realistic SSE frames (`token`/`tool_call`/`proposed_action`/`artefact_draft`/`error`/`done`) matching the real wire format in `types.ts`/`streamSse.ts` — every other layer (Pinia stores, `useAiChat`, `CopilotChat.vue`, `ProposedActionCard.vue`, `DiffView.vue`, `useApplyDraft.ts`) ran unmocked, real production code, against the real `tutorial/hello-world` flow.

### Summary

| Scenario | Result |
|---|---|
| Proposed mutating action shows inline diff before approval | ✅ SUCCESS |
| Draft-apply-on-existing-flow shows the same diff instead of the old text-only confirm | ✅ SUCCESS |
| Diff additions/removals distinguishable without color alone | ✅ SUCCESS (`+`/`−` glyphs + badge, not just red/green fill) |
| Non-regression: error banner (unrelated to this PR) unchanged | ✅ SUCCESS |
| Console errors across all scenarios | ✅ none (one unrelated pre-existing `[kestra-sdk] SDK looks out of date` dev warning) |

### Scenario 1: Proposed action → inline diff (✅ SUCCESS)

Prompt: "Add error handling to second_task and notify on failure" → mocked `proposed_action` event carrying a `body` argument with a modified `hello-world` flow (added `retry:` block).

**Observed**: `ProposedActionCard` renders a `DiffView` with a `+5 −8` badge, collapsed-context lines ("1 unchanged line hidden"), and `+`/`−` line markers on top of color — exactly the diff-first review the issue asked for. "Reject"/"Approve" actions present, `argEntries` no longer double-shows the diffed key.

### Scenario 2: Artefact draft → Apply → diff-embedded confirm (✅ SUCCESS)

Prompt: "add a draft test for this flow" → mocked `artefact_draft` (kind `FLOW`) → clicked **Apply** on the card.

**Observed**: the confirm dialog is still titled "Apply flow" with the exact original copy ("Apply the flow tutorial.hello-world? It will be created, or overwritten if it already exists.") but now embeds a `DiffView` (`+8 −23` badge) computed against the live editor buffer — replacing the old plain-text-only `KsMessageBox.confirm`.

### Non-regression: error banner (✅ SUCCESS, unrelated to this PR)

Forced an SSE `error` frame — the red error banner still shows no action buttons (Direction 2 "recoverable errors" is a separate, not-yet-filed issue; this PR correctly didn't touch it).

### Note (not a finding against this PR)

The `KsMessageBox`-based "Apply flow" dialog rendered with a light-mode background while the rest of the app was in dark mode. `useApplyDraft.ts` didn't touch the message-box wrapper/type/background token (only added a VNode diff), and the design-system SCSS for `.kel-message-box` already uses `--ks-bg-elevated` (theme-aware), so this looks pre-existing to `KsMessageBox` rather than introduced here — flagging for someone to double-check separately, not blocking this PR.

### Environment
- OSS, embedded H2, `feat/copilot-diff-preview` worktree, ports 8190 (API) / 5390 (Vite) — isolated from other local dev servers.

ui/src/utils/textDiff.ts
```diff
@@ -0,0 +1,133 @@
...
+            result.push({kind: "gap", count: run.length - leading - trailing})
+            if (trailing > 0) result.push(...run.slice(run.length - trailing))
+        }
+        i = j
+    }
+    return result
+}
+
+function splitLines(text: string): string[] {
+    return text.length ? text.split("\n") : []
```
flcarre: 🟡 [Performance/Correctness] Spurious diff line on trailing-newline-only differences
Problem: `splitLines` treats a trailing `\n` as an extra empty-string line, so two YAML sources that are identical except for a trailing newline (e.g. the live editor buffer, which typically has no trailing newline, vs. a persisted/AI-drafted source that does) render as a spurious `+`/`−` empty line even though nothing meaningfully changed. Not a correctness bug in the diff algorithm itself, just noise a reviewer would notice comparing the proposed change to what's already there.
Fix: normalize a single trailing newline before splitting, mirroring how unified diff tools treat EOF.
```suggestion
function splitLines(text: string): string[] {
    if (!text.length) return []
    return text.endsWith("\n") ? text.slice(0, -1).split("\n") : text.split("\n")
}
```

ui/src/components/ai/copilot/CopilotChat.vue
```diff
@@ -224,6 +225,19 @@
...
+    // namespace/id arguments match the flow currently focused, so a diff is never shown against the
+    // wrong flow's content. `id` is accepted alongside `flowId` since tool argument naming isn't fixed.
+    const pendingConfirmationFlowSource = computed<string | undefined>(() => {
+        const scope = routeInFocus.value
+        const args = pendingConfirmation.value?.arguments
+        if (!args || scope?.kind !== "FLOW" || !scope.namespace || !scope.flowId) return undefined
+        const namespace = typeof args.namespace === "string" ? args.namespace : undefined
+        const flowId = typeof args.flowId === "string" ? args.flowId : typeof args.id === "string" ? args.id : undefined
+        if (namespace !== scope.namespace || flowId !== scope.flowId) return undefined
+        return editorFlowSource.value
```
flcarre: 🟡 [Design] `pendingConfirmationFlowSource` inherits the context-chip dismissal state
Problem: this reuses `editorFlowSource`, which returns `undefined` once the user has dismissed the "flow" context pill (`dismissedParts.value.has("flowId")`). That dismissal exists to control what's sent to the AI as `additionalContext`, but it also silently switches the diff preview from "compare against the live buffer" to "render as a pure addition" for a proposal that still targets the exact same open flow (namespace/flowId already matched above) — which can misrepresent an update as if the flow didn't exist yet. Might be intentional (dismissed = "don't use this flow's state at all"), but worth confirming that's the desired UX rather than an accidental coupling between two unrelated concerns (AI context vs. diff preview).
Fix: if unintentional, read `flowStore.flowYaml` directly here instead of going through `editorFlowSource`, so the diff preview's "before" source doesn't depend on the context-chip dismissal.

ui/src/components/ai/copilot/DiffView.vue
```diff
@@ -0,0 +1,135 @@
...
+    import {diffLines, summarizeDiff, collapseContext, type TextDiffLine} from "../../../utils/textDiff"
+
+    const props = withDefaults(defineProps<{
+        /** The current content ("before"). Empty when there's nothing to diff against (e.g. a brand-new
+         *  flow) — every line of `newValue` then renders as an addition. */
+        oldValue: string
+        newValue: string
+        /** Unchanged lines kept around each change before the rest collapses into a gap marker. */
+        context?: number
+    }>(), {context: 3})
```
flcarre: 🟢 [Nit] Duplicated default context value
`textDiff.ts` already exports/uses `DEFAULT_CONTEXT_LINES = 3` internally (via `collapseContext`'s own default parameter), and this prop default re-hardcodes the same `3`. Not exported today, so there's nothing to import — either export `DEFAULT_CONTEXT_LINES` from `textDiff.ts` and reference it here, or drop this default and rely on `collapseContext`'s own default by not passing `context` at all when unset. Purely cosmetic, not blocking.

flcarre (commented):
## Kestra OSS Code Review

Closes #19329: a diff is now shown before applying a Copilot-proposed mutating action (`ProposedActionCard`) and before a draft-apply confirm (`useApplyDraft`), matching the issue's ask.

### Kestra Guidelines (backend) — N/A (no Java changed)

### Frontend / Design System — 0 blocking, 2 suggestions, 1 nit
- All colors resolve to existing `--ks-*` tokens (verified against `ks-theme-*.scss`); no hex/`rgb()`/`--el-*`/`:deep()`/direct `element-plus` imports.
- `+`/`−` glyphs plus `aria-label`s on the summary tags — state isn't color-only, matches the a11y rule.
- All 7 new `ai.copilot.diff.*` keys are present in `en.json` and correctly translated (with proper plural forms per language) in all 12 locales; `fingerprints.json` was regenerated, not hand-edited.
- 🟡 Trailing-newline diff noise in `textDiff.ts` (see inline).
- 🟡 `pendingConfirmationFlowSource`'s dismissal coupling in `CopilotChat.vue` (see inline, non-blocking judgment call).
- 🟢 Minor constant duplication in `DiffView.vue` (see inline).

### Security (OWASP Top 10) — 0 issues found
- AI-controlled content (proposed YAML/diff text) is rendered exclusively through Vue text interpolation (`{{ entry.text }}`) in `DiffView.vue` — no `v-html`/`dangerouslyUseHTMLString` anywhere, so it's XSS-safe despite being LLM-authored content.
- `useApplyDraft.ts`'s `KsMessageBox({..., message: () => h("div", ..., [h(DiffView, ...)]) })` uses the render-function form (VNode tree, not a raw HTML string) — same pattern already used in `stores/flow.ts`, so no injection surface there either.
- The "before" source resolution is correctly scoped: `ProposedActionCard`'s diff only compares against the currently-open flow when the action's own `namespace`+`flowId`/`id` arguments match the route's focused flow (`CopilotChat.vue`'s `pendingConfirmationFlowSource`), preventing a diff from leaking one flow's content into a proposal targeting a different flow.

### Performance — 0 issues found
- `lcsDiff` is O(n·m) time/space but is guarded by `MAX_LCS_CELLS` (4M cells) with a documented, bounded-input rationale (flow/dashboard YAML), falling back to a plain remove-all/add-all rendering above that — reasonable for a client-side UI preview.
- No repository/query calls in a loop, no deep-watch-on-computed-spread pattern, no unbounded allocation beyond the guarded LCS table.

### Tests
Genuinely behavioral, not tautological: `textDiff.spec.ts` exercises the diff algorithm's actual edge cases (identical/pure-addition/pure-removal/minimal-edit/context-collapse-boundaries); `ProposedActionCard.spec.ts` covers the body/source/yaml-key extraction, the single-overlong-argument fallback, and the no-changes state; `CopilotChat.spec.ts` adds the three cases that matter for the namespace/flowId matching guard (same flow, different flow, non-flow route); `useApplyDraft.spec.ts` asserts the live-buffer-vs-fetch-vs-failure-fallback resolution with concrete call-count/arg assertions on `loadFlow`.

**Verdict:** APPROVE — no blocking findings. Two non-blocking suggestions and a nit inline, optional to address before merge.

flcarre:
## QA follow-up — KsEditor rework (commit 8bbdcdb668)

**Verdict: PASS**

Re-tested after the follow-up fix that replaced the bespoke `textDiff.ts`/`DiffView.vue` renderer with `KsEditor`'s Monaco diff mode (same pattern as `Revisions.vue`), addressing the "diff too small / can't scroll" feedback on the original implementation.

Same test method as the initial QA: real running app (`:cli:runLocal` + this worktree's `ui/`, embedded H2), AI Copilot SSE endpoints fetch-mocked (no live provider locally), everything else real production code, against the real `tutorial/hello-world` flow.

### Results

| Check | Result |
|---|---|
| Proposed-action diff (Copilot dock) renders as a real Monaco diff | ✅ line numbers, syntax highlighting, overview ruler with change markers |
| Diff is actually scrollable | ✅ scrolled programmatically, revealed the added `retry:` block further down |
| "Apply flow" confirm modal width | ✅ now ~750px (`customStyle: {maxWidth: "min(750px, 90vw)"}`), was ~420px before |
| Apply-flow diff scrollable | ✅ same Monaco scroll behavior confirmed inside the `KsMessageBox` |
| Light mode | ✅ |
| Dark mode | ✅ — as a bonus, the modal background itself is now correctly dark-themed too (the first pass had flagged the modal chrome staying light in dark mode as a separate, likely-pre-existing `KsMessageBox` quirk; not reproducible now) |
| Console errors | ✅ none |

This resolves the size/scroll feedback. Good to merge from a QA standpoint.

ui/src/components/ai/copilot/DiffView.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+    <KsEditor
```
flcarre, line 2: 🔴 [Frontend/DS] KsEditor renders with the wrong theme in light mode
Problem: `theme` isn't bound here, so `KsEditor` falls back to its literal default `theme: "dark"` (`design-system/src/components/Form/KsEditor.vue:98`), which is wired straight into `monaco.editor.createDiffEditor({..., theme: props.theme})` (`useKsEditor.ts:312/321`) and into the `theme-dark`/`custom-dark-vs-theme` container classes (`useKsEditor.ts:162-166`). Every other `<KsEditor>` call site in `ui/src` (`Revisions.vue`, `TaskEditPanes.vue`, `RawPreview.vue`, `VarValue.vue`, `WebhookCurl.vue`, `AddTriggerModal.vue`, `McpConnect.vue`, `TestEventDialog.vue`) binds `v-bind="editorBindings"` (from `useEditorBindings()`), which supplies `theme: computed(() => getTheme())` so the editor tracks the app's actual light/dark setting. `DiffView.vue` — used both by `ProposedActionCard` and by the `useApplyDraft` confirm modal — skips that, so a user in light mode gets a dark Monaco diff editor inside an otherwise light modal/card. This is the exact problem `Revisions.vue` (the pattern this PR is reusing) already solved.
Fix: `import {useEditorBindings} from "../../composables/useEditorBindings"`, add `const editorBindings = useEditorBindings()` in `<script setup>`, and add `v-bind="editorBindings"` on the `<KsEditor>` tag (before the more specific props so `readOnly`/`lang`/etc. still take precedence). Spans both the template and the script block, so not expressible as a single-line suggestion.

flcarre (commented):
## Kestra OSS Code Review — re-review of the KsEditor rewrite (commit 8bbdcdb668)

The switch from the hand-rolled `textDiff.ts`/div-based renderer to a thin `KsEditor` wrapper is the right call and correctly follows `Revisions.vue`'s precedent for *what* API to use (`original`/`modelValue`, `readOnly`, diff mode). i18n cleanup is symmetric across all 13 locale files (verified: same 7-line add/remove in every locale + `en.json`, net-zero diff against `develop`), the CSS-specificity fix is real (see inline comment — the DS rule it needs to beat is actually *more* specific than the code comment claims, but Vue's `scoped` attribute selector still makes the fix reliable), tests were rewired against a real `KsEditor` stub rather than hollowed into tautologies, and the deleted `DiffView.stories.ts`/`textDiff.ts` leave no dangling references.

One correctness bug found by reading `KsEditor`'s actual defaults: `DiffView.vue` never binds the `theme` prop, and `KsEditor` defaults `theme` to the **literal string `"dark"`**, which is wired straight into `monaco.editor.createDiffEditor({..., theme: props.theme})`. Every other `<KsEditor>` call site in `ui/src` (`Revisions.vue`, `TaskEditPanes.vue`, `RawPreview.vue`, `VarValue.vue`, `WebhookCurl.vue`, `AddTriggerModal.vue`, `McpConnect.vue`, `TestEventDialog.vue`) passes `v-bind="editorBindings"`, which supplies `theme: computed(() => getTheme())` — the one thing that makes the editor track the app's actual light/dark setting. `DiffView.vue` doesn't, so in light mode both the `ProposedActionCard` diff and the `useApplyDraft` confirm modal will render a dark Monaco surface. See the inline comment for detail.

### Business Requirements — met, with one regression risk
Closes #19329 (renders a diff before applying/approving a Copilot change). The rewrite additionally fixes the "too small/unscrollable" complaint that triggered it. The `+N/−N` badge removal doesn't regress the "distinguishable without color alone" criterion — Monaco's diff editor defaults `renderIndicators: true` (gutter `+`/`-` glyphs, not color-only), and this matches the already-shipped `Revisions.vue` pattern.

### Kestra Guidelines (backend) — N/A, no Java changed

### Frontend / Design System — 1 blocking, 2 suggestions, 1 nit
- 🔴 **Block** — `DiffView.vue` omits `theme`, so the diff editor ignores the app's light/dark setting (see inline comment on `DiffView.vue`).
- 🟡 Suggestion — the CSS-specificity comment in `DiffView.vue` overstates its case ("same specificity") when the DS rule it beats is actually more specific; worth correcting so a future reader doesn't remove the compound selector thinking it's cosmetic (inline comment).
- 🟡 Suggestion — `min(750px, 90vw)` in `useApplyDraft.ts` duplicates `KsDialog`'s `large` width formula as a literal instead of a shared constant; a future change to that DS width silently drifts the two apart (inline comment).
- 🟢 Nit — the confirm message in `useApplyDraft.ts` builds raw `h("div")`/`h("p")` instead of `h(KsText, ...)`, so it doesn't route through the DS typography abstraction (inline comment, non-blocking, no visible issue reported).

Confirmed `KsEditor`/Monaco creates the diff via `monaco.editor.createModel`/`createDiffEditor`, a plain-text model — no `v-html`/`innerHTML` path anywhere in `KsEditor.vue`. AI-authored YAML rendered through `DiffView` has no code-execution or markup-injection surface beyond what already exists for every other read-only `KsEditor` usage in the app.

At most one `DiffView` (one Monaco diff editor) is mounted at a time in `CopilotChat` (`v-if="pendingConfirmation"` gates a single `ProposedActionCard`, not one per transcript turn); `KsEditor` already handles resize via `ResizeObserver` + `automaticLayout`, so mounting inside an animating `KsMessageBox` isn't a concern. No new loops, no repeated store/API calls introduced.

**Verdict:** REQUEST CHANGES — fix the theme binding (it's a real, reproducible light-mode regression, not theoretical) before merge; the two 🟡 suggestions and the 🟢 nit are worth a follow-up but not blocking.

ui/src/components/ai/copilot/DiffView.vue
```diff
@@ -0,0 +1,31 @@
...
+         *  flow) — Monaco's diff editor then renders every line of `newValue` as an addition. */
+        oldValue: string
+        newValue: string
+    }>()
+</script>
+
+<style scoped>
+    /* KsEditor's own stylesheet sets `.ks-editor { height: 100% }` at the same specificity as a bare
+       `.diff-view` class would, and wins the cascade tie outside a full-height layout (e.g. inside a
+       KsMessageBox, which sizes to its content) — target both classes together to win outright. */
```
flcarre: 🟡 [Nit] The specificity claim in this comment overstates its case
Problem: this says KsEditor's `height: 100%` rule is "at the same specificity as a bare `.diff-view` class" — it isn't. The actual rule is `:not(.namespace-defaults, .kel-drawer__body) > .ks-editor { height: 100% }` (two classes, since `:not()` contributes the specificity of its most specific argument), which is *more* specific than a bare one-class `.diff-view`, so `.diff-view` alone wouldn't tie, it would lose outright. `.ks-editor.diff-view` matches that two-class specificity — the fix still reliably wins because Vue's `scoped` style adds a `[data-v-...]` attribute selector to every rule in this block, pushing `.ks-editor.diff-view[data-v-...]` above the DS rule's two bare classes. The fix is real, just not for the stated reason — worth correcting so a future reader doesn't "simplify" it back to a bare `.diff-view` thinking the compound selector is cosmetic.
Fix:
```suggestion
    /* KsEditor's `height: 100%` rule is `:not(.namespace-defaults, .kel-drawer__body) > .ks-editor`
       (two classes) — more specific than a bare `.diff-view`, so that alone would lose outright, not
       tie. Matching it with `.ks-editor.diff-view` plus Vue's scoped `[data-v-...]` attribute wins. */
```

ui/src/components/ai/copilot/useApplyDraft.ts
```diff
@@ -124,6 +124,53 @@ export function useApplyDraft() {
...
+            message: () => h("div", null, [
+                h("p", null, t("ai.copilot.draft.applyConfirm", {namespace, id})),
+                h(DiffView, {oldValue: before, newValue: yaml}),
+            ]),
+            showCancelButton: true,
+            confirmButtonText: t("ai.copilot.draft.apply"),
+            cancelButtonText: t("cancel"),
+            // The default message box tops out at ~420px — too narrow for the diff editor below the
+            // confirm text. Match KsDialog's "large" variant width rather than inventing a new size.
+            customStyle: {maxWidth: "min(750px, 90vw)"},
```
flcarre, line 181: 🟡 [Suggestion] Duplicated magic width instead of a shared constant
Problem: `min(750px, 90vw)` is a literal copy of `KsDialog`'s `large` width formula (`resolvedWidth = props.width ?? (props.large ? "min(750px, 90vw)" : ...)`, `KsDialog.vue:72`). `KsMessageBox` has no equivalent `large` option to reuse, so this works today, but the two literals can silently drift apart the next time someone tunes `KsDialog`'s large width.
Fix: export the width as a small shared constant from the design system (e.g. next to `KsDialog`, something like `KEL_DIALOG_LARGE_WIDTH`) and import it here instead of re-typing the literal — or, if `KsMessageBox` callers commonly need this, add a `large`-equivalent option to `KsMessageBox` itself. Not blocking; a good follow-up.

ui/src/components/ai/copilot/useApplyDraft.ts
```diff
@@ -124,6 +124,53 @@ export function useApplyDraft() {
...
+        id: string,
+        yaml: string,
+        onThisFlow: boolean,
+        flowStore: ReturnType<typeof useFlowStore>,
+    ): Promise<boolean> {
+        const before = onThisFlow ? (flowStore.flowYaml || "") : await persistedFlowSource(namespace, id, flowStore)
+        return KsMessageBox({
+            type: "warning",
+            title: t("ai.copilot.draft.applyTitle"),
+            message: () => h("div", null, [
```
flcarre, line 172: 🟢 [Nit] Raw h() VNodes bypass KsText
Problem: `h("div", ...)` / `h("p", ...)` render the confirm text without going through `KsText`, so this paragraph doesn't carry the DS typography abstraction the rest of the app uses for body copy (`ui/AGENTS.md` — "Typography comes from KsText or typography tokens"). Low risk here since `ElMessageBox`'s own CSS likely gives it reasonable default styling, and QA confirmed it looks correct in both themes, but it's a precedent for future VNode-based `KsMessageBox` content.
Fix: `h(KsText, {size: ...}, () => t("ai.copilot.draft.applyConfirm", {namespace, id}))` instead of the raw `<p>`. Non-blocking.

flcarre (commented):
## Kestra OSS Code Review — re-review of the theme-binding fix (commit cebbee6110)

Re-verified the blocking finding from the prior review round (theme binding on `DiffView.vue`'s `KsEditor`) is resolved, plus the two non-blocking cleanups requested alongside it. Scoped to what changed in this commit on top of the already-reviewed `8bbdcdb668`.

ui/src/components/ai/copilot/DiffView.vue
```diff
@@ -0,0 +1,35 @@
...
+        newValue: string
+    }>()
+
+    const editorBindings = useEditorBindings()
+</script>
+
+<style scoped>
+    /* KsEditor's own `.ks-editor { height: 100% }` rule is less specific than the scoped `.diff-view`
+       selector below (Vue adds a `[data-v-...]` attribute to it), so it would otherwise be overridden —
+       target both classes together so this rule wins instead. */
```
flcarre: 🟡 [Nit] The corrected comment still misstates KsEditor's actual selector
Problem: this still describes KsEditor's rule as the bare `.ks-editor { height: 100% }` (one class). The actual selector that sets `height: 100%` is `:not(.namespace-defaults, .kel-drawer__body) > .ks-editor` (two classes, per KsEditor.vue:178) — the same inaccuracy flagged in the prior review round, just restated with "less specific" instead of "same specificity". The fix itself is still correct (`.ks-editor.diff-view[data-v-...]` does win), only the comment's premise is off.
Fix:
```suggestion
    /* KsEditor's `height: 100%` rule is `:not(.namespace-defaults, .kel-drawer__body) > .ks-editor`
       (two classes, KsEditor.vue) — more specific than a bare `.diff-view`. Matching it with the
       two-class `.ks-editor.diff-view` selector, plus Vue's scoped `[data-v-...]` attribute, wins. */
```
Non-blocking — purely a comment-accuracy nit, doesn't affect the CSS's actual behavior.

flcarre:
## Kestra OSS Code Review

Re-review of commit `cebbee6110` on top of `8bbdcdb668`, scoped to the fix for the blocking finding from the prior review round (https://github.com/kestra-io/kestra/pull/19330#pullrequestreview-5168477776).

No change in scope; this commit is a pure fix for the previously-flagged theme-binding bug plus two non-blocking cleanups.

### Kestra Guidelines (backend) — N/A, no Java changed

### Frontend / Design System — 1 finding
- ✅ **Blocking finding resolved**: `DiffView.vue` now imports `useEditorBindings`, calls it in `<script setup>`, and binds `v-bind="editorBindings"` on `<KsEditor>` — placed before the more specific props, exactly mirroring `Revisions.vue`'s call site (`ui/src/components/layout/Revisions.vue:100-108`). No remaining path renders the diff with the hardcoded `theme: "dark"` default; both `ProposedActionCard`'s embedded diff and `useApplyDraft`'s `KsMessageBox` confirm modal now go through the same `editorBindings.theme` as every other `KsEditor` call site.
- ✅ `useApplyDraft.ts` now builds the confirm message with `h(KsText, {tag: "p"}, () => t(...))` instead of raw `h("p")`, matching the DS typography rule. `KsText` accepts `tag` and a function default slot, so this renders correctly.
- 🟡 The CSS-specificity comment on `DiffView.vue`'s `<style>` block was corrected but still misstates KsEditor's actual selector as bare `.ks-editor { height: 100% }` (one class) — the real rule is `:not(.namespace-defaults, .kel-drawer__body) > .ks-editor` (two classes, `KsEditor.vue:178`), the same inaccuracy flagged last round. The CSS fix itself is unaffected (posted as an inline nit with a suggestion — non-blocking).

### Test changes — sound
`CopilotChat.spec.ts` and `ProposedActionCard.spec.ts` add `vi.mock(".../useEditorBindings", () => ({useEditorBindings: () => ({})}))`, matching the pre-existing pattern in `VarValue.spec.ts` / `FlowFileEditorTab.spec.ts` (stubbing to avoid pulling in three Pinia stores unrelated to what those specs assert). Both mocked call sites (`ProposedActionCard.vue`'s `<DiffView>`, and `CopilotChat.vue` → `ProposedActionCard` behind `pendingConfirmation`) were verified in source. The mock doesn't assert theme behavior — it doesn't need to, since it exists purely to keep unrelated tests from breaking on newly-required stores, not to cover the theme fix (which was verified independently by manual light/dark toggling against a running instance).

**Verdict:** APPROVE — the blocking finding is genuinely resolved and independently verified (source read + reported runtime measurement). One non-blocking nit posted inline for a future touch-up.

MilosPaunovic (changes requested):
## Findings

### 1. The new pre-confirm fetch raises a global error toast for a brand-new drafted `flow`

`useApplyDraft.ts:167` - `persistedFlowSource` calls `flowStore.loadFlow({namespace, id, store: false})` with no request options. `isReportedCentrally` (`ui/src/utils/kestraHttp.ts:71`) raises the global error toast for every 404 unless `ignoreNotFound: true` is passed.

Applying a `flow` the copilot just drafted is the primary path for this feature, and the function's own doc comment names the case ("a brand-new flow, i.e. nothing persisted yet"). So the user now sees a red Not Found toast, and only then the confirm dialog. Before this `PR` no fetch happened at all on that path.

The idiom already exists in the codebase (`FlowConcurrency.vue:93`, `stores/plugins.ts:560`), and this very file already declares a `silent` constant for the same reason:

```ts
const data = await flowStore.loadFlow({namespace, id, store: false}, {ignoreNotFound: true, showMessageOnError: false})
```

### 2. The `ProposedActionCard` half of the `PR` is unreachable in `OSS` and `EE` today

`RestartExecutionTool` is the only tool anywhere with `AgentWritePolicy.CONFIRM`, and nothing in either repo registers `AgentToolFamily.MUTATE`. Authoring tools (`AuthorFlowTool`, and `EE`'s `AuthorAppTool` / `EeAuthorDashboardTool`) are registered with `family = null` and `AgentWritePolicy.AUTO` at `ToolCatalog.java:111`, so they emit `artefact_draft` rather than a confirmation.

No proposed action can therefore carry a `body` / `source` / `yaml` argument, or a long argument at all - `RestartExecutionTool`'s arguments are all short ids. The two new stories invent `create-flow` and `update-flow` tools that do not exist. That is fine as forward-looking work, but it is worth saying so in the description, because nothing in `CI` or in manual testing can exercise it.

### 3. The "sole overlong argument is the source" fallback will pick the wrong side of the diff

`ProposedActionCard.vue:131`. `AuthorFlowTool` already declares a `currentFlowYaml` parameter carrying the **current** `flow`. Any future confirmable tool shaped like it, invoked with a short instruction, leaves `currentFlowYaml` as the only overlong argument - so the heuristic selects it and renders the existing `flow` as the proposed content, on the new side of the diff.

A diff that is confidently backwards is worse than no diff. I would require one of the known keys and drop the fallback entirely, rather than guess at a tool contract that has not been written yet.

### 4. Historical proposals mount the diff too

`CopilotMessage.vue:96` renders `<ProposedActionCard :action="historicalProposal" resolved />` for every past `PROPOSED_ACTION` message, and `proposedSource` has no `resolved` guard.

Once a mutating tool does ship, every past proposal in a thread mounts its own 20rem Monaco diff editor, with `currentFlowSource` undefined - so an update that was already applied renders as an all-green pure addition. Misleading to read, and expensive in a long thread. A `v-if="!resolved"` on the `DiffView` would cover it.

### 5. No busy state across the new fetch

`applying` only flips to `true` after `confirmApplyFlow` resolves, so the Apply button stays enabled for the whole round trip. A second click opens a second confirm dialog.

### 6. The `.kel-message-box__message` rule is a design-system-wide change in a `feat(ai)` `PR`

`flex: 1; min-width: 0` on `.kel-message-box__message` changes every `KsMessageBox` in `OSS` and `EE`, and ships with no design-system story or test. Probably benign or an improvement, but it deserves a line in the description and a look at the busier confirm dialogs.

## Nits

- **Two test hooks on one element.** `DiffView` sets `data-test="copilot-diff"` on its root while `ProposedActionCard` passes `data-test="copilot-proposed-diff"`. Fallthrough attributes win over the child's own, so the inner hook never appears in that usage. Worth picking one.
- **The scoped-style comment is still off, in a third direction.** `KsEditor`'s rule is `:not(.namespace-defaults, .kel-drawer__body) > .ks-editor`, which is `(0,2,0)` since `:not()` contributes the specificity of its most specific argument. A bare scoped `.diff-view` compiles to `.diff-view[data-v-...]`, also `(0,2,0)` - so it would **tie**, with source order deciding, rather than lose outright or win. `.ks-editor.diff-view[data-v-...]` is `(0,3,0)` and wins cleanly, which is why the fix is correct.
- Agreed on your own open note that `pendingConfirmationFlowSource` inherits the context-pill dismissal state. Dismissing the pill governs what is sent to the agent; silently turning an update into a pure addition is a separate concern.

## `CI`

Red, but not because of the code. The only failing step in `Frontend - Tests` is `Npm - lint (eslint)`, where reviewdog got a 500 from the GitHub diff `API` ("temporarily unavailable due to heavy server load"). `check:types` and `oxlint` both passed.

The consequence does matter though: `Run front-end unit tests` and `Run storybook component tests` were **skipped** when the job bailed, so the test suites have never run against `877c1ef525`. Worth a re-run before merging.

## Backport

`ProposedActionCard.vue`, `useApplyDraft.ts` and `element-plus.scss` all exist on `releases/v2.0.x`; none of the copilot files exist on `releases/v1.3.x`. So `releases/v2.0.x` is the only candidate. It is a feature rather than a fix, so my read is no backport, but your call.

flcarre:
@MilosPaunovic all 6 findings + the 3 nits from your review are addressed in `e64cd66d` (on top of `cf7f48c6`), CI green:

1. `persistedFlowSource` now passes `ignoreNotFound: true` — no more spurious "Not Found" toast on a brand-new drafted flow.
2. Documented explicitly in the PR description: no OSS/EE tool mutates a flow through `ProposedActionCard` today, so that half is forward-looking.
3. Dropped the "sole overlong argument" fallback in `ProposedActionCard.vue` — only `body`/`source`/`yaml` are recognized now, no more risk of a backwards diff.
4. `DiffView` in `ProposedActionCard.vue` is now gated with `!resolved`, so historical proposals no longer mount a diff editor.
5. `applying.value = true` is now set before `confirmApplyFlow` (the fetch), so a second click can't open a second confirm dialog.
6. Documented explicitly in the PR description: `.kel-message-box__message { flex: 1 }` is a DS-wide change, worth a second look on other confirm dialogs.

Nits: dropped the redundant `data-test="copilot-proposed-diff"` prop (DiffView's own `copilot-diff` is now the only hook); rewrote the CSS-specificity comment using your exact tie → source-order explanation.

Also folded in on top: the "everything happens in chat" → "so let's also mirror it in the main editor + save Approve as a draft" feedback from a separate design discussion, plus a route-matching bug I found in manual QA (`isViewingFlow` used `route.name === "flows/update"`, but the real route on the Edit tab is the nested `flows/update/edit` — fixed via the existing `routeFamily()` helper, `e64cd66d`).

Re-requested your review whenever you have a moment — thanks again for the thorough pass 🙏

MilosPaunovic (changes requested):
Round 2 on `a684f24e1b` (reviewed at `0b7fc74b09`; the two commits since are `develop` catching up and touch nothing in this `PR`). Everything from the previous round is closed, including the description notes on the unreachable mutate path and the design-system rule - thanks. `CI` is green now too.

The follow-up commit `3e80b5b7b3` is a new feature rather than review feedback, so the items below are all new.

## 1. The mirrored preview silently locks the main `flow` editor, with no way out

`FlowFileEditorTab.vue:34` makes the editor read-only whenever `previewSource !== undefined`, and `editorUpdate` (`:293`) discards edits on the same condition. So the moment a `FLOW` draft card targeting the open `flow` exists, the main Flow Code editor stops accepting input.

The problem is what clears it. `pendingDraftFlowSource` (`CopilotChat.vue:297`) picks the newest `ARTEFACT_DRAFT` in the transcript targeting the open `flow`, and a draft card lives in the transcript forever - `CopilotArtefactDraft.vue` offers only Open in editor and Apply, there is no dismiss. The clears are `onSubmit` (`:411`), `onNewChat` (`:424`), `onBeforeUnmount` (`:481`), and `loadFlow`.

None of those is "I read the suggestion and I would rather keep editing by hand". That user has to send another chat message, start a new chat, navigate away, or close the copilot dock, and nothing on screen says the editor is read-only or why. The description's "goes back to normal (editable, unsaved edits intact) once the preview clears" is accurate about the listed transitions, but declining a draft is not one of them.

A dismiss on the draft card, or an explicit banner over the editor with a Dismiss preview action, would close this.

## 2. The lock re-arms on reload, including after the draft was already applied

`loadThread` rehydrates `ARTEFACT_DRAFT` messages with their `draft` payload (`useAiChat.ts:207`), and the mirror watcher is `{immediate: true}` (`CopilotChat.vue:316`). So reopening the copilot on that `flow` later re-applies an old draft's preview and re-locks the editor.

It is worse after a successful apply. `loadFlow` clears `previewSource` at the store level (`stores/flow.ts:545`), but `activeFlowPreview` never changes, so the watcher does not re-fire to reflect that - it only re-fires on the next mount or route change, at which point it sets the preview straight back. Since `flowStore.flowYaml` now equals the applied draft, the result is an empty diff in an editor the user cannot type in. That is a hard dead end to reason about.

## 3. `draft: true` is a product behaviour change riding in a diff-preview `PR`

The mechanism is sound: the backend parameter exists and its own description matches the new confirm copy word for word (`FlowController.java:367`), `getFlow` returns the latest revision including drafts, and the `UI` already surfaces drafts with a badge (`FlowRootTopBar.vue:11`), a warning banner and a Publish action (`FlowRun.vue:7-12`). No complaints about the implementation.

But "applying a `Copilot` proposal no longer publishes it" is a product decision, not a review one, and it is arriving inside a `PR` whose title and linked issue (https://github.com/kestra-io/kestra/issues/19329) are about showing a diff. I would rather see it split out and signed off separately.

One inconsistency if it does stay: `applyDashboard` is untouched, so dashboards still apply live, and `applyConfirmDashboard` still says "overwritten if it already exists". The same card now behaves two different ways depending on the artefact kind.

## Nits

- **`pendingDraftFlowSource` recomputes on every streamed token.** `useAiChat.ts:345` mutates `activeAssistant.content` in place inside `messages.value`, so deep reactivity re-runs the computed for each token. It walks the transcript newest-first and `parseArtefactYaml` is a full YAML parse, so a thread carrying several drafts for other flows parses all of them per token. Cheap in the common case, but it is on the streaming hot path.
- **`{immediate: true}` writes `previewSource = undefined` on mount** even when nothing is pending. Nothing else owns that field today, so it is latent only.

## Backport

Unchanged from last round: `releases/v2.0.x` has all the touched files, `releases/v1.3.x` has none of them. Feature, so my read is still no backport.

flcarre:
@MilosPaunovic on #3: keeping `draft:true` in this PR is intentional, not scope creep on my part — it came directly from a stakeholder conversation about this exact PR (the diff-preview only helps if you can trust what you're approving; approving straight to a live revision without a chance to review felt like the wrong default once the diff review UX existed at all). So it's a deliberate product call tied to this feature, not an unrelated behavior change riding along — happy to be overruled by whoever owns that call, but I don't think it needs a separate PR/sign-off cycle to land here.

On the `applyDashboard` inconsistency: agreed it's real, and it's a deliberate scope boundary for now rather than an oversight — this PR only touches the flow path (issue #19329 is flow-scoped), and dashboards are EE-only in OSS anyway (the `useApplyDraft.ts` doc comment already notes dashboard drafts only actionable when the backend serves custom dashboards). Leaving `applyDashboard` on its current live-apply behavior rather than half-fixing draft semantics for an artefact kind this PR doesn't otherwise touch. Can file a follow-up if we want dashboards to match once there's appetite for it.

Fixing #1 and #2 now (both are real bugs, no disagreement there) plus the two nits.

MilosPaunovic (changes requested):
Round 3 on `046eeebbf0`. `CI` is green. Almost everything is closed, one item left.

On #3: accepted, thanks for the context. A stakeholder call tied to this feature is exactly the answer I was after - I raised it because nothing in the `PR` said where it came from, not because I disagreed with the default. Your `applyDashboard` scope boundary reads fine to me too; a follow-up when there is appetite is the right shape.

#1 is fully fixed. The Dismiss on the card plus the banner over the locked editor, both behind one `declineActivePreview`, is a better answer than either alone, and registering `flowStore.declinePreview` with an identity guard on unmount is the right way to hand it across. The `apply` signature change to `Promise<boolean>` so `markDraftApplied` only fires on a real write is a nice touch.

The `draftScanSignal` watch also does close the per-token nit properly: reading `messages.value[last]?.id` tracks `length`, the index and `id`, never `content`, so a streamed token no longer re-parses candidate YAML.

## Still open: the dismissed/applied tracking does not survive a remount

`dismissedDraftIds` and `appliedDraftIds` are component-local refs, and the `ARTEFACT_DRAFT` messages they filter are rehydrated from the persisted thread by `loadThread` (`useAiChat.ts:207`), `draftId` included. So both sets come back empty against a transcript that still has the draft.

Two ordinary actions lose them:

- **A page reload.** `restoreThread()` on mount brings the draft back and the scan picks it up again.
- **Closing the copilot dock.** `ContextDrawer.vue:37` is `<KeepAlive v-if="activeTab">`, so the cache survives switching between the `ai` / docs / news tabs but is destroyed when the drawer closes, taking the `CopilotChat` instance with it.

The applied case is the one that reads worst: apply a draft, reload, and the editor locks again on a diff where both sides are now identical. The banner means it is no longer a dead end, which is the important half, but the user has to dismiss the same resolved draft after every reload.

The applied half has a fix that needs no persistence at all - skip a draft whose YAML already matches what the editor is showing, since a zero-change diff should never lock anything:

```ts
if (yaml === flowStore.flowYaml) continue
```

in `findPendingDraft`, alongside the existing dismissed/applied checks. That is reload-proof by construction and also covers "the user hand-typed the same change themselves".

The dismissed half genuinely needs storage if you want it to persist. Keying a small set in `localStorage` by thread id would do it, but I am happy for that to be a follow-up if you would rather not grow the scope again - the empty-diff guard is the part I would like in this `PR`.

## Backport

Unchanged: `releases/v2.0.x` has the touched files, `releases/v1.3.x` has none. Feature, so still no backport from me.

flcarre:
Fixed in `46c0ee89f5` — added the `yaml === flowStore.flowYaml` skip in `findPendingDraft`, exactly as suggested. Left the dismissed-half persistence as a follow-up per your note. Test added simulating a remount (fresh tracking, matching YAML) → editor stays unlocked.

MilosPaunovic (approved):
Round 4 on `46c0ee89f5`. Nothing left open from my side.

The `yaml === flowStore.flowYaml` skip in `findPendingDraft` closes the last item, and I like that the comment says *why* it is there rather than just what it does. Worth noting for anyone reading it later that the guard is strictly stronger than the tracking it backs up: it also covers a draft the user happened to apply by hand, and a draft that was already live before the thread was restored, neither of which the id sets would ever have caught.

I checked the ordering it depends on and it holds. `draftScanSignal`'s watch is created before the `activeFlowPreview` mirror watch, so on a route change the rescan runs first and the mirror sees the already-resolved `pendingDraft`, not a stale one.

One observation, not a change request. `flowStore.flowYaml` is deliberately absent from `draftScanSignal`, which is right, since it changes on every keystroke and would put the scan back on a hot path. The cost is a narrow window on page load: if `restoreThread` resolves before the flow's own `loadFlow`, the scan compares against an empty buffer, picks the draft and mirrors it for a moment. It resolves itself, because `loadFlow` clears `previewSource` on the way in (`stores/flow.ts:545`) and the mirror watch does not re-fire for an unchanged value. So at worst there is a brief locked-editor flash on a reload, bounded by the flow fetch. Not worth trading the hot-path scan for, in my view. Flagging it only so it is a known quantity rather than a surprise if anyone ever reports it.

Approving. Thanks for the patience across four rounds, and for pushing back on #3 with the actual provenance rather than just changing it.

## Backport

To restate the answer for the record: `releases/v2.0.x` carries every file this touches, `releases/v1.3.x` carries none of them. This is a feature rather than a fix, so no backport from me. Say if you want a 2.0 `PR` anyway and I will raise the question with whoever owns the 2.0 scope.

ui/src/components/ai/copilot/useApplyDraft.ts
flcarre: Closing as a documented deferral rather than a code change.

Fixing this properly would mean exposing `KsDialog`'s width constant through `KsMessageBox`, which has no `width` or `large` prop at all today — so it is a design-system API change, not a local edit. This PR already carries one DS-wide change (the `.kel-message-box__message` flex rule, flagged in the description as wanting a second look), and adding a second one this late is the wrong trade.

The literal stays, with the comment above it naming `KsDialog`'s `large` formula as the source so the duplication is traceable if `KsMessageBox` ever grows a real size option.

PR #19332, opened 2026-09-10
refactor(tests): one shared i18n mount helper consistent across repos

### LINKED ISSUE #19331: Every unit spec builds its own i18n instance to mount a component
Almost every unit spec that mounts a component starts by building its own vue-i18n instance: import `createI18n`, call it with `legacy: false` and `locale: "en"`, decide whether to pass messages, then hand it to `mount` through `global.plugins`. Across `ui`, the design system, topology and `ui-ee` that is 164 specs doing the same setup with small variations, and none of the packages has a shared mount helper.

That has a cost beyond the noise. When a design-system component started using `$t`, the specs that mounted it without an i18n plugin failed with `_ctx.$t is not a function`, which reads like a component bug rather than missing test setup. A shared helper makes that mistake impossible.

What to do:

- Add one mount helper per package, next to the existing test setup: `i18nMount` and `i18nShallowMount`. It installs i18n in English, takes `messages` for the keys a spec cares about or `locales` for a whole translation file, and merges anything passed in `global` over the defaults. The design-system version also installs the design-system plugin, since every spec there needs it.
- Topology and `ui-ee` reuse the OSS helper, the way the translation tooling and the test setup already do, instead of getting a copy.
- Convert the specs to it and delete the local `createI18n` calls and duplicated plugin lists. No test assertion changes: a spec that asserted on raw keys keeps empty messages, a spec that loaded `en.json` keeps loading it.

Two config fixes belong with it. `ui-ee` has to pin `@vue/test-utils` and `vue-i18n` to its own `node_modules`, as it already does for `vue`, otherwise the shared helper mounts through a second copy of test-utils that never sees the global stubs and router set up in `setup.ts`. And the topology package's vitest config needs the same Node 26 `localStorage` line the other packages already have, which is why its `Duration` spec fails locally today.

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra/issues/19331.
EE PR: https://github.com/kestra-io/kestra-ee/pull/10973

Also enables jsdom's storage in `ui/packages/topology/vitest.unit.config.ts`, the same line `packages/design-system/vitest.unit.config.ts` already carries. Without it the topology suite does not run at all on Node 26 (`localStorage` is undefined, 13 failures), so the specs this PR moves there could not otherwise be verified.


flcarre (commented):
Mechanically this is a clean migration. I checked the whole diff programmatically: **zero** `it`/`test`/`describe` blocks and **zero** `expect(...)` lines were added or removed across all 93 files, and in the ~26 files I read hunk-by-hunk every custom `stubs`, `provide`, `slots`, `attachTo` and extra plugin was carried over verbatim, with specs that needed real translations correctly switching to `locales: en`. No test was dropped and none silently stopped asserting. The EE side importing `kestra/tests/unit/i18nMount` means the cross-repo claim genuinely holds for the app helper.

The issues are in the convention this PR establishes, not in the migration.

### 1. The design-system helper's `messages`/`locales` options cannot work

`ui/packages/design-system/tests/units/setup.ts` (pre-existing, untouched here) replaces `vue-i18n` wholesale:

```js
vi.mock("vue-i18n", () => ({
  useI18n: () => ({t: (key) => key}),
  createI18n: () => ({install(app) { app.config.globalProperties.$t = (key) => key }}),
}))
```

That stub discards `createI18n`'s options entirely. So in the design-system `i18nMount.ts` the `messages` and `locales` fields are inert, and the 32 arguments this PR adds to call sites — 22 `locales,` in `KsTopNavBar.test.ts`, 8 `messages,` in `KsEmptyState.test.ts`, 2 in `FilterMultiSelectGroups.test.ts` — are dead weight that reads as if it controls rendering. Nothing breaks today (those specs already asserted on raw keys), but the next person will write a design-system assertion against translated text and spend an afternoon on it.

Pick one: drop `messages`/`locales` from the design-system helper and its call sites, or drop the blanket mock from the design-system setup so the helper is real there too. The second is the better end state and would let the two helpers collapse into one.

### 2. AGENTS.md documents that inert behaviour as working

`ui/AGENTS.md:258` says to pass "`messages` for the keys the spec asserts on, or `locales: en` when it needs the real `en.json`". True under `ui/tests/unit/`, false under `ui/packages/design-system/tests/units/`, which the same sentence names. It also doesn't cover `ui/packages/topology`, whose specs now import the **app** helper via `../../../../../tests/unit/i18nMount` — a third arrangement. Since the rule is the deliverable here, it should say exactly what applies where.

### 3. The topology specs lost their missing-key warning suppression

`ui/packages/topology/tests/units/{misc/Duration,nodes/BasicNode,nodes/TaskNode}.test.ts` each dropped `missingWarn: false, fallbackWarn: false`. Under `ui/tests/unit/` that's fine — `setup.ts` mocks `createI18n` to default both false. Topology's own setup has no such mock, so those three now get the real `createI18n` with warnings on. `Duration.test.ts` is the worst case: its own comment says nearly every key is intentionally left unset so the spec can assert raw-key fallback, so essentially every render logs `[intlify] Not found '<key>' key`.

The clean fix is in the shared helper rather than per-spec:

```ts
const i18n = createI18n({legacy: false, locale: "en", missingWarn: false, fallbackWarn: false, messages: locales ?? {en: messages}})
```

That also removes the helper's hidden dependence on the app-level `vue-i18n` mock, which is what makes it behave differently in three packages today.

### Smaller

- **Package boundary.** `@kestra-io/topology` is standalone and publishable, but its tests now reach outside the package into `ui/tests/unit/`. The helper also carries `import type {Pinia} from "pinia"`, and pinia is in none of topology's dependency lists. Type-only, so nothing fails today, but it's latent coupling.
- **`pinia` option is dead.** Nothing here or in the EE companion passes it; the one spec needing a pinia instance uses `global.plugins`.
- **Leftover empty objects.** `const globalConfig = {}` remains in `KsEmpty.test.ts`, `KsSelect.test.ts` and `KsDurationPicker.test.ts`.
- **Scope creep.** The `NODE_OPTIONS` / `--no-experimental-webstorage` line in `ui/packages/topology/vitest.unit.config.ts` is an unrelated Node 26 jsdom fix and isn't mentioned in the description.

flcarre (approved):
Re-reviewed after the three commits. Every point from my earlier pass is addressed, and the first one is fixed the way I hoped rather than the cheap way.

### Previously raised, now resolved

1. **The design-system helper's `messages`/`locales` were inert** — fixed at the root. `9498dba3` deletes the blanket `vi.mock("vue-i18n")` from `ui/packages/design-system/tests/units/setup.ts`, so the design-system helper now builds a real `createI18n` like the app one. That this was real erosion and not a theoretical worry shows up in the diff: `KsEmptyState.test.ts` carries the **only** changed `expect` in the whole PR, and it goes from the raw key `"ks_empty_state.learn_more"` to `"Learn more"`. The 22 `locales` arguments in `KsTopNavBar.test.ts` now genuinely install `KsTopNavBar.locale`. Assertions got stronger, not weaker.

2. **`ui/AGENTS.md` documented that inert behaviour as working** — rewritten. It now names all three arrangements (`tests/unit/`, `packages/topology/tests/`, `packages/design-system/tests/units/`), says which helper each imports, and states that missing-key warnings are off in both. The documented API matches what the helpers actually accept.

3. **Topology specs lost their missing-key suppression** — fixed in the shared helper rather than per-spec, as suggested. Both copies now hardcode `missingWarn: false, fallbackWarn: false`, which also removes the helper's hidden dependence on an app-level mock that no longer exists.

4. **Dead `pinia` option and the `import type {Pinia}`** — both gone from `ui/tests/unit/i18nMount.ts`. Specs that need a store still pass `createPinia()` through `global.plugins`, and the helper's `[i18n, ...userPlugins]` ordering keeps i18n installed first.

5. **Leftover `const globalConfig = {}`** — removed from `KsEchart`, `KsDataTable`, `KsEmpty`, `KsSelect`, `KsDurationPicker` and the rest.

6. **Scope creep** — the `--no-experimental-webstorage` line is now explained in the description: `packages/design-system/vitest.unit.config.ts` already carries it, and without it the topology suite doesn't run at all on Node 26. Fine as stated.

### Re-verification of the migration itself

Re-ran the mechanical check across all 96 files at the new head: **zero** `it`/`test`/`describe` blocks added or removed, exactly **one** `expect` changed (the strengthening above), all 13 custom `stubs`, all 4 `attachTo: document.body` in `BlockEditor.test.ts`, and every `provide` / `slots` carried over verbatim.

I also checked the new risk the mock removal creates: with the stub gone, any design-system spec bare-mounting a component that calls `useI18n()` would now throw rather than silently get `{t: key => key}`. Nine `src` files use `useI18n`; every one of their specs is in the migrated set, and `KsEditor.test.ts` — whose `useKsEditor` composable also uses it — never mounts anything. `Frontend - Design System Tests` is green on `9498dba3`, which confirms it.

On the boundary question: `files` in the design-system `package.json` is `["dist"]`, so neither helper can leak into the published package.

### Nits, none blocking

- The two helper files are now identical apart from the `KestraDesignSystem` plugin. The design-system's own `setup.ts` already imports `../../../../tests/unit/monacoCssEscapePolyfill`, so reaching into `ui/tests/unit/` is established precedent there — the design-system copy could wrap the app helper rather than duplicate it, and AGENTS.md wouldn't have to promise that two copies "behave the same".
- `messages: locales ?? {en: messages}` silently drops `messages` if a call site passes both. No call site does today; a union type would keep it that way.
- Several call sites read `locales: messages`, where the local const named `messages` holds a whole locale map. Renaming those consts to `locales` would stop the option and the value disagreeing.
- Topology's `../../../../../tests/unit/i18nMount` is a five-level reach out of a publishable package. Documented now and no longer dragging a pinia type along, but an alias would be sturdier than the relative path.

Good refactor. LGTM.

PR #19333, opened 2026-09-10
fix(executions): bound every large-value render site in the outputs UI

### LINKED ISSUE #19316: BUG: Execution UI freezes for seconds when a task emits more than 1 MB of individual output values
### Version

Kestra 1.3.37 OSS (`kestra/kestra:v1.3.37-no-plugins`, `server local`, H2), Chromium 1440x900.

### What happens

When a task emits a large number of individual output values, opening the execution's **Outputs** tab or a taskrun's **Outputs** drawer blocks the main thread for seconds to almost a minute. The tab is not slow, it is frozen: no scrolling, no clicks, no tab switching.

1 MiB of output is enough to make it obvious. It is not a rendering-quantity cliff either, the cost grows faster than the payload.

### Reproduction flow

```yaml
id: large_output_ui
namespace: company.team

inputs:
  - id: values_count
    type: INT
    defaults: 15000

tasks:
  - id: many_individual_values
    values: |
      {%- if true -%}
      {
      {%- for i in range(1, inputs.values_count) -%}
      {% if i > 1 %},{% endif %}"item_{{ i }}": "value {{ i }} - 0123456789abcdefghijklmnopqrstuvwxyz0123456789abcdefghij"
      {%- endfor -%}
      }
      {%- endif -%}

  - id: size_report
    message: "emitted {{ inputs.values_count }} individual output values"
```

Run it with `values_count: 200` (fine), then `15000` (1.24 MiB), then `50000` (4.2 MiB). Open the **Outputs** tab, then Gantt > the taskrun's ⋮ menu > **Outputs**.

### Measurements

Headless Chromium, `PerformanceObserver` on `longtask`, on an M-series Mac with the server on localhost. Real users on slower machines will see worse.

| values | task output | Outputs tab render | longest single task | total blocking |
|---|---|---|---|---|
| 200 | 16 KB | 1.2 s | 226 ms | 246 ms |
| 15,000 | 1.24 MiB | 3.2 s | **2.4 s** | 3.8 s |
| 50,000 | 4.2 MiB | 15.6 s | **14.0 s** | 17.0 s |

| values | Outputs drawer render | longest single task | total blocking |
|---|---|---|---|
| 200 | 2.9 s | 129 ms | 148 ms |
| 15,000 | 16.1 s | **34.3 s** | 47.2 s |
| 50,000 | 48.1 s | **30.2 s** | 57.4 s |

At 15,000 values a `setTimeout(…, 0)` scheduled while the Outputs tab was building fired 1,118 ms later. At 50,000, Playwright's own click on the drawer menu item could not complete within 20 s.

3.3x the data costs 5.8x the blocking on the tab. Both paths are superlinear.

### Root cause

Two independent paths, both eager and neither virtualized.

**Outputs tab** — `ui/src/components/executions/outputs/Wrapper.vue`. `transform()` (L366) walks the whole outputs object and builds one node per value, and the result is handed to `el-cascader-panel` as `:options` (L6-L9), which is not virtualized. `onMounted` (L313) then auto-expands three levels, so a task with 15,000 output values paints a column of 15,000 `el-cascader-node` elements up front. Element Plus also instantiates its own `Node` for every option while building the store.

**Outputs drawer** — `ui/src/components/executions/Outputs.vue` renders `Vars` with the taskrun's outputs (L7-L10), which are already inlined in the execution payload, so there is no lazy load to soften the hit. `Vars` calls `Utils.executionVars` (`ui/src/utils/utils.ts` L61) and every resulting value goes through `VarValue.vue`, where a complex value is serialized into a Monaco model with `JSON.stringify(getDisplayValue(value), null, 2)` (L42) and serialized *a second time* in the same template to compute `customHeight`, including a `.split('\n')` over the whole document (L40). Both run again on every re-render.

**`Utils.flatten` is broken, which makes the drawer case much worse.** `ui/src/utils/utils.ts` L46:

```js
static flatten(object) {
    return Object.assign({}, function _flatten(child, path = []) {
        ...
        return Object.keys(child).map(key => typeof child[key] === "object" ?
            _flatten(child[key], path.concat([key])) :
            ({[path.concat([key]).join(".")]: child[key]}));
    }(object));
}
```

The recursion returns an **array**, and `Object.assign({}, [ … ])` copies index keys rather than merging the entries. For anything nested, the whole payload collapses into a single `"0"` row holding a nested array of `{"values.item_N": …}` objects, instead of one row per `a.b.c` path. So the drawer hands the entire 1.24 MiB to one Monaco editor under a blank Name column, rather than 15,000 small rows.

In the drawer screenshot attached below, that shows up as a single row with an empty **Name** column and one editor holding everything. Worth fixing on its own, independently of the performance work.

### Suggested direction

- Virtualize or paginate the cascader column in `Wrapper.vue`, or cap the children built per node with a "show more" affordance. Building the tree lazily per expanded level would also avoid the up-front `transform()` walk.
- In `VarValue.vue`, stringify once into a `computed` instead of twice in the template, and refuse to mount Monaco above a size threshold (offer a download or a truncated preview instead).
- Fix `Utils.flatten` so nested outputs become one row per leaf path. That alone turns the drawer's single multi-MB editor into ordinary rows, and it is a correctness fix regardless.
- Consider serving large taskrun outputs out of the execution payload (`develop` already lazy-loads them in `Outputs.vue`; 1.3.x does not).

elevatebart:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/19316.

### ✨ Description

A task emitting a few MiB of output values froze the execution UI for tens of seconds. Six render sites reached the DOM unbounded, and one of them could not render at all.

**The outputs list.** `Vars` rendered one `KsTable` row per flattened leaf path, and `KsTable` is not virtualized. It now uses a `DynamicScroller` — the same virtual list `Logs.vue`, `TaskRunDetails.vue` and `Gantt.vue` already use — so all rows stay in one continuous scroll and only the visible ones are built. Bounded by `max-height` rather than a fixed height, so a short list still renders at its own size and the trigger-variable, cascader and expression-debugger tables are unchanged.

**The value editor.** `VarValue` serialized the value twice per render, once for the editor model and once for a `.split("\n")` that only wanted a line count, then handed Monaco the whole document. It now serializes once and passes a bounded preview.

**What "bounded" means, and why it is not a text clip.** Clipping the serialized JSON cuts mid-token, so Monaco's JSON language service reported every truncated preview as a syntax error: unclosed objects and arrays at the tail, and unterminated strings wherever the per-line clip landed mid-document. Repairing the tail cannot fix the second kind. So `Utils.boundForDisplay` bounds the **parsed value** and serializes that, which is well-formed by construction. `Utils.capForDisplay` stays for text nothing parses.

Five bounds, because a preview can be large for five unrelated reasons and each of them was demonstrated by a real shape:

| bound | value | what it stops |
|---|---|---|
| `PREVIEW_MAX_ENTRIES` | 100 | a container with 15,000 entries |
| `PREVIEW_MAX_STRING_CHARS` | 500 | one long string — as a value **and as a key** |
| `PREVIEW_MAX_NODES` | 1000 | many tiny nodes, each of which is a line |
| `PREVIEW_MAX_CHARS` | 32 KiB | the total, whatever combination produced it |
| indent, charged per entry as `depth × 2` | — | deep nesting, where the leading spaces are the document |

The last two are worth spelling out. A node budget alone allows `PREVIEW_MAX_NODES × PREVIEW_MAX_STRING_CHARS`, about 500 KiB, so the total needs its own budget. And a content budget alone does not bound depth: `JSON.stringify(…, null, 2)` indents by nesting level, so a 1200-deep value is megabytes of leading spaces before a single leaf is reached. Charging each entry the indent its depth earns makes depth pay for itself.

Dropped entries are named by an `…` marker carrying how many went — `"… 5900"` at the end of an array, `"…": 5900` in an object — so the count is visible without the marker being a translatable string inside a JSON document.

Measured, against the same shapes:

| | before | after |
|---|---|---|
| Ansible, 6000 tasks (1.1 MiB) | 200 lines, longest 72, **unparseable** | 513 lines, longest 72, **parses** |
| Flat object, 15,000 keys | 200 lines, longest 26, **unparseable** | 103 lines, longest 24, **parses** |
| One 2.5 MB string leaf | 2 lines, longest 2000, **unparseable** | 3 lines, longest 516, **parses** |
| One 2 MB object **key** | 1 line of 2,000,013 chars, **no banner at all** | 3 lines, longest 514, banner |
| 1200-deep nesting | 3,003 lines, 3,030,212 chars, 89.8 ms | 525 lines, 96,234 chars, 0.9 ms |
| 1000 × 600-char strings | 103 lines, 51,704 chars | 67 lines, 33,092 chars |

The last three come from @MilosPaunovic's review of the first structural version, which bounded values but not keys and had no total budget. The 2 MB key was the serious one: it came back `truncated: false`, so the whole 2 MB went to Monaco on one line **and** the banner that carries **Copy** and **Download** never rendered, leaving the value both unrenderable and unreachable.

**The plain-string branch.** A non-JSON string fails `isComplexValue`, so it never reached that editor and landed in an uncapped text node instead. Both of `VarValue`'s paths are now bounded, the string one by character count since there is no structure to bound.

**Debug Expression.** `ExpressionDebugger` fed the whole evaluated result to Monaco. Now bounded, structurally when the result parses as a container and by character count otherwise, so a long scalar result is still readable in full up to the text cap.

**The explorer's raw-JSON and scalar views** did the same with the selected value. The raw view only ever holds an object, so it previews structurally; the scalar view keeps the character clip.

**The explorer's sidebar** kept every string value whole in each item's `preview` — a single line the CSS clips anyway. `previewText` is now bounded on every path, and its object branch slices the key list before joining it, so a 15,000-key object no longer builds a string nobody would see.

**`flatten` could not render large outputs at all.** `concat(...keys.map())` and `Object.assign({}, ...leaves)` each spread one argument per key, so past V8's argument limit it threw `RangeError` and the table rendered nothing. It gave way at 150,000 leaves, roughly an Ansible run of 25,000 tasks. Virtualization cannot help — the throw precedes any row. It now accumulates into one object. One behaviour change comes with that: a `path.length > 0` guard, so a top-level `{}` flattens to `{}` rather than to `{"": {}}` and no blank-key row is rendered. `releases/v2.0.x` already carries the identical guard, so this is `develop` catching up rather than a new decision.

**Truncation is never silent and never a dead end.** A `KsAlert` states the real size and offers the whole value two ways: **Copy**, and **Download** as a file. Copy alone meant putting a few MiB on the clipboard and then finding somewhere to put it. The file is named after the output it came from — `Vars` passes the flattened key down, the explorer uses the selected path — so pulling several values out of one execution does not produce a pile of `output(1).json`. The banner names no limit, since only one of the five applies to any given value.

### 🎨 Frontend Checklist

- [x] Type checking passes (`npm run check:types` — all three passes)
- [x] Unit tests pass (`npm run test:unit`) — 259 files, 2296 tests
- [x] Translations are complete if `en.json` changed — `npm run translations:check` reports no missing, extra, stale, untranslated or mismatched-placeholder keys

### 📝 Additional Notes

**Evidence.** Docker Hub and ghcr blob storage are both unreachable from the environment this was written in, so no live instance was possible. Numbers come from a component harness: the real `Vars`, `VarValue` and `SidebarList` mounted in a production Vite build with the app's own stylesheets, driven by headless Chromium, `PerformanceObserver` on `longtask` registered through `addInitScript` plus a `setTimeout(…, 0)` probe. Budget: no single task over 500 ms, total blocking under 1 s.

| Render site | Longest task before | after | Total blocking before | after |
|---|---|---|---|---|
| Outputs list, 15,000 values | 7,596 ms | 414 ms | 10,048 ms | 458 ms |
| Value editor, 2.4 MiB | 8,299 ms | 327 ms | 8,955 ms | 277 ms |
| Debug Expression, 2.5 MB | 6,586 ms | 352 ms | 7,425 ms | 302 ms |
| Sidebar preview, 50 × 1.3 MB | 13,484 ms | 57 ms | 26,023 ms | 7 ms |
| `VarValue` plain string, 2.5 MB | 1,335 ms | 0 ms | 1,285 ms | 0 ms |
| Outputs list + plain string | 1,086 ms | 66 ms | 1,426 ms | 16 ms |
| `flatten`, 150,000 leaves | `RangeError` | renders | — | — |

That table predates the switch from character clipping to structural bounding, and the harness is gone with the environment that built it, so it has not been re-run. What changed under it is the document the editor receives, measured in the table above. The list and editor rows also predate the `flatten` rewrite, which only lowers them: `flatten` of 50,000 leaves went from 229 ms to 28 ms, and a million leaves now flatten instead of throwing.

**A verification pass found four sites the first version of this PR missed**, three of them worse than the outputs table it started with — Debug Expression, the sidebar preview, the plain-string branch, and the absence of any way to retrieve a truncated value. It also found that **the cap itself was insufficient**: character and line caps together still let one pathological line through, and a 2.5 MiB string value pretty-prints to exactly that, so the capped editor still blocked for 994 ms. `capForDisplay` therefore also bounds line length, which is what took that case to 352 ms.

**@MilosPaunovic's reviews found three more**, each a case the round before it had not covered: the truncated plain string was a 2000-character unbroken line that the rewritten grid's value track clipped with no horizontal scroll, because it had no `min-width: 0` or `overflow-wrap`; the structural preview bounded values but not keys; and it had no total size bound. Fixed in the rounds above, along with the row surface `KsTable` used to paint, the accent colour `code` had inherited before `.vars-key` overrode it, a deprecated `sizeDependencies` prop that warned on every mount, and a filename derived with a bare `split(".")` that produced `b"].json` for a key containing a dot.

**The cost of virtualizing rather than paginating**, since both were built: pagination at 100 rows a page was faster (209 ms / 159 ms at 15,000) because it never builds a scroller entry per row, but it made a user page through 40 screens to read 4,000 values. Three costs come with virtualization, all measured or observed:

- Rows whose values mount Monaco are unmounted and remounted as the scroller recycles. Hard-scrolling 2,000 such rows costs 519 ms.
- Ctrl+F reaches less than it did before, and less than pagination would have left it: 20–40 rows are in the DOM against pagination's 100, and against the whole list on `develop`. Nothing in the list is unreachable, it is one continuous scroll, but the browser's own find is no longer a way to search it.
- `Vars` is no longer a `KsTable`: the two-column row is feature CSS on `--ks-*` tokens, because the design system has no virtualized table. If a `Ks*` virtual table lands later, this is the first caller that should move to it.

**Still over budget, deliberately left:**

- **Outputs list at 50,000 values** — 1,483 ms / 2,098 ms. Rendering is bounded; what remains is `DynamicScroller` building a size and view entry per item plus Vue reactivity over a 50,000-element array. The honest fix is server-side paging of outputs, a webserver and SDK change rather than a UI one.
- **One deeply nested value** flattening to ~48,000 rows — 1,666 ms / 2,720 ms before the `flatten` rewrite, which removes about 340 ms of it. Same root cause as above.
- **Explorer sidebar at 3,000 items** — 624 ms / 751 ms. A plain `v-for`, so a `ForEach` with thousands of iterations puts every task run in one section. Bounding it means virtualizing inside `KsCollapse`.

**Scoping against the release branches.** `releases/v1.3.x` has its own fix, f50564b081, which landed while this one was in review, so nothing is owed there. The two branches solve the same bug differently and their i18n keys diverge: 1.3 refuses to render anything over 10 KB and offers a JSON download (`large_outputs.value_too_large`, `download_json`, `load_more`, behind `largeValues.ts`); this one renders a bounded preview and offers Copy and Download (`large_outputs.value_truncated`, behind `Utils.boundForDisplay` and `Utils.capForDisplay`). The escape hatch is now the same on both, so what is left to settle is whether a large value should be previewed or refused outright — a product call rather than a merge blocker for either. Two thirds of the 1.3.x diff has no counterpart here:

- `Utils.flatten` already spread the recursion's array into `Object.assign` on `develop`. The `Object.assign({}, [ … ])` collapse that stuffed every nested output into a single row is `releases/v1.3.x` only. (The spread bug fixed here is a different one, present on both.)
- `outputs/Wrapper.vue` and its `el-cascader-panel` are gone, replaced by `ExecutionVariableExplorer.vue` + `SidebarList.vue`. That explorer is flat, so nothing builds a node per output value and there is no cascader column to cap. The load-more row, its scroll-offset restore and its node-count label do not exist on this branch.

`releases/v2.0.x` still has the unvirtualized `KsTable` `Vars` and no bounded preview, so the issue reproduces there and needs its own PR against that branch.

**Tests.** Six in `VarValue.spec.ts`, one in `Vars.spec.ts`, eight in `src/utils/utils.spec.ts`. Each fails on the commit before its fix, checked in a scratch worktree rather than by stashing, including the copy and download tests, which assert the whole value reaches the clipboard and the file rather than the preview, and the `flatten` test, which fails with the exact `RangeError`. The parse-validity test asserts both directions in one place: the bounded preview parses and the old clipped text does not. Two exceptions, stated rather than hidden: `flatten`'s structural test is a characterization test for a rewrite of a function that had none, and the small-value cases pin the boundary where nothing should be bounded. Two tests were deleted rather than added to: the `Vars` test replaces two pagination-shaped ones from the first round, and `VarValue`'s line-limit test went when structural bounding made the size and line caps one behaviour. The banner's layout and the value column's wrap are CSS with no layout engine behind them in jsdom, so neither has a unit test; @MilosPaunovic measured the wrap in a live harness.

**Not exercised.** With no live instance: the Outputs tab's auto-expand, clicking Debug Expression in situ, Gantt-drawer navigation, the file download buttons against the real API, the new blob download against a real browser's download manager, and the banner's own layout. The Storybook suite cannot run in this environment either (`chrome-headless-shell` is absent and cannot be downloaded), so CI's `Frontend - Tests` is the only signal for whichever stories render `Vars` and `VarValue`.

**Rebased on #19339.** An earlier round of this branch carried its own fix for the `check:ts-any` drift in `TaskAnyOf.stories.tsx`, which was red on `develop` for reasons unrelated to this PR. #19339 has since fixed it at the source, and more thoroughly: all thirteen casts typed and the file dropped from the baseline, against the six this branch had typed. Those two commits are dropped and `develop`'s version is taken, so nothing about that story is in this diff any more.

**Related.** #19395 fixes an unrelated design-system i18n teardown race that was failing `Frontend - Tests` on this PR and on `develop`.

A cat walks into a bar with 15,000 output values. The bartender says "we don't serve your kind here". The cat says "that's fine, you only ever render thirty". 🐱

---

https://claude.ai/code/session_01XmNZXEzKESw9E7cAXYHovT


elevatebart:
`Npm - check ts-any` (inside `Frontend - Tests`) is failing on this branch, and it is not this PR's:

```
New `any` in 1 file(s).
  tests/storybook/components/no-code/components/tasks/TaskAnyOf.stories.tsx: 7 -> 13
```

That file is not in this diff. It is red on `develop` itself — checking out `origin/develop` clean and running `node scripts/explicit-any/baseline.mjs src packages tests` reproduces the identical line, so every PR that merges current `develop` inherits it.

Origin: 4383997 (#18959, the Schedule timezone picker) added 77 lines to `TaskAnyOf.stories.tsx`, taking its `any` count from 7 to 13, without touching `ui/scripts/explicit-any/baseline.json`. `develop`'s baseline still records 7.

No fix exists to port yet — #19321 touches the same checker but changes what it counts rather than fixing this drift, and it is still open. So the proposed patch, for whoever owns #18959:

```diff
--- a/ui/scripts/explicit-any/baseline.json
+++ b/ui/scripts/explicit-any/baseline.json
-    "tests/storybook/components/no-code/components/tasks/TaskAnyOf.stories.tsx": 7,
+    "tests/storybook/components/no-code/components/tasks/TaskAnyOf.stories.tsx": 13,
```

or better, type the six new `any` in that story, which is what the rule is for. I have deliberately not made that change here: it would absorb another PR's debt into this diff and no-op the moment `develop` gets its own fix.

Re-running the job cannot help — the count is deterministic, and this branch's own `check:ts-any` is clean (`No change against the baseline: 1649 explicit any in 393 files`) with the failure appearing only once `develop` is merged in. `Npm - check types`, both lint steps, and the unit and storybook suites all pass on this head.

---

elevatebart:
Correcting my comment above: the `TaskAnyOf` drift **is** fixed in this PR now, so the "deliberately not made that change here" line no longer holds.

Rather than bumping the baseline, the six `any` are typed. `Schema` requires `type` and `$language`, and those `anyOf` branches carry neither by design — that is the backend shape the stories cover — so a local `asSchema(partial: Partial<Schema>)` holds the one unavoidable assertion and leaves every other key checked; a mistyped `enum` or `pattern` now fails to compile where `as any` waved it through. The seven older casts are untouched, so the file returns to its recorded count of 7 and `baseline.json` needs no edit.

The branch is also rebased onto current `develop` (`7928ce4`), so it carries #18959 and the fix applies to it directly. `check:ts-any` reports `No change against the baseline: 1556 explicit any in 391 files`, all three `check:types` passes are clean, and the unit suite is 258 files / 2286 tests.

One gap worth stating: I could not run the Storybook suite locally — it needs `chrome-headless-shell`, which is absent from this environment's preinstalled browsers and cannot be downloaded here. `asSchema` is an identity function at runtime, so the story args are unchanged and only their types moved, but the `Frontend - Tests` Storybook job is the real confirmation.

The original diagnosis in my previous comment still stands for anyone hitting this on another branch: `develop` was red because 4383997 (#18959) took that file from 7 to 13 without touching the baseline.

---

elevatebart:
`Frontend - Tests` is red on `6f2c5c2`, and it is not this PR's:

```
Error: Test timed out in 5000ms.
 ❯ tests/unit/stores/logsCursor.spec.ts:18:5
```

That file is not in this diff, and a clean `origin/develop` checkout reproduces it exactly — `captures OFFSET mode with total from the response`, 5009 ms, 1 failed / 9 passed. On this branch, identically: 5007 ms, same single case.

**Mechanism.** It is the first test in the file, and the only one that pays the `await import("../../../src/stores/logs")` cold. `beforeEach` calls `vi.resetModules()`, which clears the registry but not Vite's transform cache, so tests 2-10 re-import in milliseconds and pass. Warming the cache does not help — three consecutive local runs all failed at ~5007 ms — so this is the store's dependency graph exceeding the 5 s default on a first transform, not a flake in the usual sense.

**Why it is intermittent in CI rather than constant.** `vitest.config.unit.js` sets `isolate: false`, so one worker shares a module registry across files. When any other spec imports `src/stores/logs` first, this file's import is free and the suite is green — which is what happened on `fdc0986` and on `df4ad4e` attempt 2, both green here. File ordering decides it.

#19341 hit the same thing independently and records it in its description: *"two logsCursor.spec.ts tests exceeding their default 5-second timeout. The logs cursor suite passes 10/10 with a 15-second timeout."* It does not carry a fix, so there is nothing to port into this branch.

Proposed patch, for whoever owns the file:

```diff
-describe("logs store cursor pagination", () => {
+describe("logs store cursor pagination", {timeout: 15000}, () => {
```

or hoist the import into a `beforeAll` and drop the per-test dynamic import, which removes the cliff rather than widening it. I have not made either change here: the file is unrelated to this PR and the fix belongs with the logs store, not behind an outputs-rendering change.

Re-running the job, since ordering is what decides it.

---

MilosPaunovic (changes requested):
Reviewed the diff and the changed files directly, plus a live component harness (the real `Vars` and `VarValue` mounted in Storybook, driven headless) to check the layout rewrite. The core of this holds up: `flatten` is semantically equivalent to `develop`'s and the `RangeError` is gone, the list really is virtualized (403 `.vars-row` nodes in the `DOM` before, 21-25 after, recycling confirmed under a real wheel scroll), and the Monaco rows measure correctly once `ResizeObserver` settles them (exact offsets, zero overlap). All 9 new tests pass, `CI` is green.

Three things I would like fixed before this merges, then some smaller ones.

### 1. The value column does not wrap, so a truncated plain string is unreadable

`.vars-row`'s second grid track has no `min-width: 0` or `overflow-wrap`. `.vars-key` got `overflow-wrap: anywhere`; the value cell did not. A long unbroken string is laid out on one line and clipped, with no horizontal scroll to recover it (`scrollWidth === clientWidth === 1248`).

This lands on exactly what this `PR` produces: `capForDisplay` clips a long line at `DISPLAY_MAX_LINE_CHARS`, so **every** truncated plain string is a 2000-character unbroken line. Measured on a 2000-char value at 1000px wide:

| | before (`KsTable`) | after |
|---|---|---|
| characters readable | 2000, wrapped over 19 lines | ~110, the rest clipped |

**before**

before: the value wraps over 19 lines and is fully readable

**after**

after: the value is one unwrapped line, clipped at the container edge

### 2. The truncation message is wrong whenever the line-length cap is what fired

`capForDisplay` truncates for three different reasons (256 KiB total, 200 lines, 2000 chars per line) but `large_outputs.value_truncated` always says "Only the first 200 lines ... are shown". Rendering this `PR`'s own headline case, a 2.5 MiB single-line string value, gives:

> Only the first 200 lines of this 2.4 MiB value are shown.

above **one** line of 2000 characters. Same for a 50-line value holding one 5000-char line: all 50 lines are shown, one is clipped, banner still claims 200 lines. Either distinguish the reason or word it neutrally, e.g. `"Showing a truncated preview of this {size} value."`.

### 3. The `releases/v1.3.x` scoping in the description is out of date

The description says there is no `releases/v1.3.x` `PR` to cross-reference and that `large_outputs.value_truncated` is absent from its `en.json`, concluding that a later backport needs no rename. `f50564b081` (`fix(executions): bound what the outputs screens render`) is already on `releases/v1.3.x` with `largeValues.ts` and three keys under `large_outputs`: `value_too_large`, `download_json`, `load_more`. So the keys do diverge, and the two branches take different product decisions for the same bug: 1.3 refuses to render anything over 10 KB and offers a `JSON` download, this one truncates and offers **Copy**. Worth settling which is intended while both are still in flight.

### Smaller things

- **Output keys lose their accent colour.** `.vars-key { color: var(--ks-text-primary) }` overrides Bootstrap reboot's `code` colour (`--kbs-code-color: var(--ks-text-link)`) that `<code class="key-col">` inherited. Measured `#631bf3` to `#08090a`, both themes. Intentional?
- **The list loses the table's surface.** Probably fine inside `KsDrawer` and modals, but worth an eyeball at the four call sites: `Outputs.vue`, `FlowTriggers.vue`, `TriggersManage.vue`, `LowCodeEditor.vue`.
- **`sizeDependencies` is deprecated** in the installed `vue-virtual-scroller` and logs a console warning on every mount of the list ("`sizeDependencies` is deprecated ... Dynamic item sizing uses `ResizeObserver` in modern browsers"). No other `DynamicScroller` in the repo passes it, and `ResizeObserver` already does the job, so `:sizeDependencies="[item.value]"` can go.
- `Outputs.vue` still passes `class="table-unrounded mt-1"` to `Vars`; `table-unrounded` was a `KsTable` class and is now dead.
- **Description vs code, sidebar preview.** `previewText`'s object branch still does `keys.join(", ")` over every key before `preview()` slices to 200 chars, unchanged from `develop`. The slice does bound what is stored, which is the part that mattered, but "a 15,000-key object no longer builds a 200 KB key list either" is not what the code does.
- **Description vs code, Ctrl+F.** "left Ctrl+F searching only the current page" is listed as a cost of the pagination alternative, but virtualization has the same cost and more of it: 20-40 rows in the `DOM` against pagination's 100.
- **`flatten`'s new `path.length > 0` guard is not mentioned.** It changes `flatten({})` from `{"": {}}` to `{}`, so no more blank-key row. It already exists on `releases/v2.0.x` with the identical comment, so this restores `develop`/2.0 parity, which is good, but it is a behaviour change beyond the `RangeError` fix and the new characterization test only pins the nested `{}` case.
- `humanFileSize(text.length)` labels a character count as bytes; it understates multi-byte values.
- Missing blank line between `copyFullResult` and `resultLang` in `ExpressionDebugger.vue`.

### Backport

`origin/releases/v2.0.x` still has the unvirtualized `KsTable` `Vars` and no `capForDisplay`, so https://github.com/kestra-io/kestra/issues/19316 reproduces there and needs its own `PR` against `releases/v2.0.x`. `releases/v1.3.x` already has its own fix, so nothing is owed there beyond reconciling the key names.

elevatebart:
All confirmed and fixed in `920cc14a4`, `8cdc90e3a`, `e4aef733f`. The measurements were the useful part: the value-column clip is a regression I introduced and would not have found from the diff, since the cap and the layout rewrite only collide on values the cap produced.

**1. Value column does not wrap.** `min-width: 0` and `overflow-wrap: anywhere` on a `.vars-value` wrapper. The grid track was the problem, not the cell content: `minmax(10rem, 15rem) 1fr` gives the value track a default `min-width: auto`, so the 2000-character line set its own minimum and the row overflowed silently.

**2. Truncation message.** Reworded to your suggestion, `Showing a truncated preview of this {size} value.`, and the `lines` argument is gone from all four bindings. New `capForDisplay` test pins the case you named, a value under both other limits with one 5000-character line, which had no coverage before. Also took the `humanFileSize(text.length)` point: the size goes through `TextEncoder` now, so a multi-byte value is not understated.

**3. `releases/v1.3.x` scoping.** My check was wrong, and it was wrong because I ran it before `f50564b081` landed (it is 17 hours old, my check was a day older) and then never re-ran it. Description rewritten: 1.3 has its own fix, the keys do diverge, and the two branches take different product decisions (10 KB hard refusal plus JSON download against truncate plus Copy). That is a product call rather than mine to settle in this PR, so it is stated as open. `releases/v2.0.x` still reproduces and is named as needing its own PR.

**Smaller ones**

- **Key colour.** Not intentional. `.vars-key`'s `color` is gone, so `code` inherits the accent again as it did under `KsTable`.
- **The list's surface.** You were right to make me look: `KsTable` paints `--ks-bg-overlay` on rows and header, and the bare list painted nothing. Now painted, so the four call sites read as they did.
- **`sizeDependencies`.** Gone. The warning string is in the installed 3.0.4 dist, and nothing else in the repo passes the prop.
- **`table-unrounded`.** Gone, and it is defined nowhere in the tree, so it was already dead before this branch touched it.
- **`previewText`'s key join.** Bounded, since `preview()` slices to 200 characters and every key costs a character at least. The description's claim is now what the code does rather than what I meant it to do.
- **Ctrl+F.** Fair, and it was worse than a mis-attribution: I listed it as a cost of the alternative while shipping more of it. The description now says virtualization reaches less than pagination would have and less than `develop` does.
- **`flatten`'s `path.length > 0` guard.** Now in the description, including the `releases/v2.0.x` parity point.
- **Blank line** in `ExpressionDebugger.vue`.

No unit test for the wrap: jsdom has no layout engine, so any assertion would only restate the class name. Your harness is the evidence for that one.

`npm run check:types`, `check:ts-any`, `lint` and `test:unit` clean, bar the `logsCursor.spec.ts` timeout I reported [above](https://github.com/kestra-io/kestra/pull/19333#issuecomment-5627957487), which reproduces the same way on a clean `origin/develop`.

---

MilosPaunovic (approved):
Re-reviewed at `90219c77`. Everything from the last round is fixed, and I checked each one in the harness rather than on the diff alone: the value column now wraps (504px of wrapped lines, `scrollWidth === clientWidth`, nothing clipped), the banner reads "Showing a truncated preview of this 2.4 MiB value." with `lines` gone from every call site and all 13 locales regenerated, the key accent colour is back at `rgb(99, 27, 243)`, the row surface is `--ks-bg-overlay` (the same token `KsTable` paints), and `sizeDependencies` is gone with no deprecation warning left in the console. The Monaco rows still measure correctly after the wrapper change: offsets `1009 / 2018 / 3027`, zero overlap, no horizontal scroll. 10 unit tests pass and `node scripts/translations/check-translations.mjs --scope oss` reports `Translation check passed`.

Nothing left open. Two nits you can take or leave:

- `humanTextSize` does `new TextEncoder().encode(text).length`, an O(n) pass allocating a copy the size of the value: 14 ms / 2.5 MB at 2.5 MiB, 55 ms / 20 MB at 20 MiB. Well inside this `PR`'s own 500 ms budget and memoized by the `computed`, but it is a new unbounded pass added for a cosmetic accuracy gain.
- `DISPLAY_MAX_LINES` and `DISPLAY_MAX_CHARS` are no longer referenced outside `utils.ts` and could drop their `export`.

`LGTM`. The `releases/v2.0.x` backport still needs its own `PR` against that branch, as the description says.

elevatebart:
`Frontend - Tests` is red on `90219c77`, and it is not this PR's. Not fixing it here, but I have a verified patch.

**What is failing.** Every test passed (`259 passed`, `2291 passed`); the job exits 1 on `Errors 6 errors` — six unhandled rejections outside any test body. All six are the same:

```
EnvironmentTeardownError: Cannot load
'/node_modules/@kestra-io/design-system/src/components/Navigation/KsTopNavBar/KsTopNavBar.locale.ts'
imported from .../@kestra-io/design-system/src/i18n/index.ts after the environment was torn down.
```

All six originate in `tests/unit/components/inputs/inputsFormBoolDefault.spec.ts`, one per test. The failure reporter agrees no test failed: `No UI test failures — nothing to report`.

**Why it is not this PR's.** `packages/design-system/src/index.ts:511`, in the plugin's `install`:

```ts
if (i18n) void registerDesignSystemI18n(i18n)
```

`registerDesignSystemI18n` awaits an `import.meta.glob` loader per `*.locale.ts` in sequence. The `void` leaves that chain in flight, so a spec that installs the plugin (that one does, with a real `createI18n`) finishes while the loop is still importing, and teardown kills whichever module it had reached. Which module gets named is therefore arbitrary.

Reproduced on a clean checkout of this PR's own base, `37e4d3856`, with a shared `node_modules` and nothing else changed:

```
Test Files  1 failed | 255 passed (256)
Tests       2 failed | 2279 passed (2281)
Errors      6 errors
```

Identical six errors, differing only in the locale module named (`KsPassword.locale.ts` there, `KsTopNavBar.locale.ts` in CI), which is what a race looks like. The two test failures are the `logsCursor.spec.ts` timeouts already reported [above](https://github.com/kestra-io/kestra/pull/19333#issuecomment-5627957487), also present on the base. This PR touches no design-system file, no `inputs/` component and no i18n registration; its 24 changed files are the outputs UI, `utils.ts`, and generated translations.

It is intermittent rather than constantly red: `6f2c5c2`, on the same base, was green a few hours ago.

**Proposed patch**, not applied here since it widens this PR into the design system. Hand the promise back instead of dropping it:

```diff
 export {designSystemLocale, setDesignSystemLocale, registerDesignSystemI18n} from "./i18n"
+
+let i18nRegistration: Promise<void> = Promise.resolve()
+
+/** The registration `install` started, so a caller can await it rather than leave it in flight. */
+export const designSystemI18nReady = (): Promise<void> => i18nRegistration

@@ packages/design-system/src/index.ts  (install)
-        if (i18n) void registerDesignSystemI18n(i18n)
+        if (i18n) i18nRegistration = registerDesignSystemI18n(i18n)
```

```diff
@@ ui/tests/unit/setup.ts
 enableAutoUnmount(afterEach)
+
+// The plugin's install starts the locale glob without awaiting it; teardown mid-import throws.
+afterEach(() => designSystemI18nReady())
```

With that, the full unit suite is `259 passed / 2291 passed`, zero errors. `src/utils/init.ts:131` already awaits the same call, so this only closes the plugin-install path that no one was awaiting.

Happy to open it as its own PR against `develop` if that is wanted. Re-running the failed job once in the meantime.

---

elevatebart:
@MilosPaunovic new commit after your approval, `cd226c6bf` — sorry to pull you back. A bug you did not see because it only shows with Monaco's language service running: **every truncated preview was reported as a JSON syntax error**.

Clipping the serialized text cuts mid-token. Unclosed objects and arrays at the tail are the obvious half; the worse half is `clipLines`, which truncates *every* over-long line, so a long string value in the middle of the document ended mid-string. Repairing the tail cannot fix that one.

So the editor paths no longer clip text. `Utils.boundForDisplay` bounds the **parsed value** and serializes that, which is well-formed by construction: each container keeps its first 100 entries and names how many it dropped (`"… 5900"` in an array, `"…": 5900` in an object), leaf strings clip at 500 characters, and a 1000-node budget stops a deep structure expanding without limit. `Utils.capForDisplay` stays for text nothing parses — plain strings, and a non-container debugger result, which is still readable in full up to the text cap rather than clipped at 500.

Measured on three shapes:

| | before | after |
|---|---|---|
| Ansible shape, 6000 tasks (1.1 MiB) | 200 lines, longest 72 chars, **unparseable** | 513 lines, longest 72 chars, **parses** |
| Flat object, 15,000 keys | 200 lines, longest 26 chars, **unparseable** | 103 lines, longest 24 chars, **parses** |
| One 2.5 MB string leaf | 2 lines, longest 2000 chars, **unparseable** | 3 lines, longest 516 chars, **parses** |

Line counts go up on the nested shape (200 → 513) and down on the flat one, but the longest line falls everywhere, which is the dimension that actually hurt: the 2000-character line was both your unreadable-wrap finding and the thing that cost 994 ms before `DISPLAY_MAX_LINE_CHARS` existed. Hundreds of short lines is a document Monaco virtualizes. The harness that produced the timing table in the description is gone with the environment that built it, so I have not re-run it and the description says so rather than implying otherwise.

Also took your first nit: `DISPLAY_MAX_CHARS` and `DISPLAY_MAX_LINES` lost their `export`, now that nothing outside `utils.ts` references them. Left `humanTextSize` as it is, for the reason you gave — it is inside budget and memoised, and there is no cheaper exact byte count in JS.

Tests: three for `boundForDisplay`, one of which asserts both directions in one place (the bounded preview parses, the old clipped text throws). `VarValue`'s line-limit test is deleted rather than updated — under structural bounding the size cap and the line cap are one behaviour, and the large-value test covers it. All three new ones verified failing with the bound neutered.

`check:types`, `check:ts-any`, `lint` and `test:unit` clean (259 files, 2293 tests).

---

MilosPaunovic (changes requested):
Round 3's structural preview is the right idea - `capForDisplay` on serialized `JSON` really does cut mid-token and make Monaco report the preview as a syntax error. But the new path dropped every bound the text path had, and two cases got measurably worse than `90219c77`. Same harness, identical inputs, `PerformanceObserver` on `longtask`:

| case | round 2 longest / total | round 3 longest / total | editor lines r2 -> r3 | banner in round 3 |
|---|---|---|---|---|
| 1000 nodes x 600-char strings (600 KiB value) | 677 / 2831 ms | **1876 / 6179 ms** | 1180 -> 4973 | yes |
| one 2 MB object key | 429 / 1156 ms | **6524 / 12538 ms** | 17 -> 15387 | **none** |
| Ansible, 6000 tasks (614 KiB) | 426 / 1198 ms | 432 / 1727 ms | 200 -> 413 | yes |

The shape from the issue is unaffected, which is why it does not show up in the numbers in the description.

### 1. The bounded preview has no size bound

`boundForDisplay` replaces `capForDisplay` on the object path, so `DISPLAY_MAX_CHARS`, `DISPLAY_MAX_LINES` and `DISPLAY_MAX_LINE_CHARS` no longer apply there. What is left is `PREVIEW_MAX_NODES` x `PREVIEW_MAX_STRING_CHARS` = 1000 x 500, about 500 KiB, roughly twice the old character budget and with no line cap. A 600 KiB value of many medium strings measures 1876 ms in a single task against this `PR`'s own 500 ms budget.

### 2. Object keys are not bounded

`bound()` clips string values and recurses into containers, but `bounded[keys[index]] = ...` copies the key verbatim. One 2 MB key therefore comes back with `truncated: false`: no banner, no **Copy**, no **Download**, and the whole 2 MB handed to Monaco on a single line. That is the pathological single-long-line case `DISPLAY_MAX_LINE_CHARS` exists for, reintroduced on the object path with the escape hatch removed, since the banner is what carries **Copy** and **Download**.

Both want the same fix: clip keys the way values are clipped, and keep a total-character budget inside `boundForDisplay` (or fall back to `capForDisplay` when the serialized preview still exceeds `DISPLAY_MAX_CHARS` / `DISPLAY_MAX_LINES` - broken `JSON` beats a 6.5 s freeze, though bounding keys should avoid the choice in practice).

### Nit

`ExecutionVariableExplorer.downloadValue()` names the file from `expressionPath.value.split(".").pop()`, but `formatStep` emits bracket notation for keys that are not identifiers, so a key containing a dot produces a filename like `b"].json`.

Everything from the earlier rounds is still good, and the 13 unit tests pass locally. `CI` has only reported `submit-gradle` and `Check for spelling errors` so far.

elevatebart:
Both confirmed and fixed in `0fb361be1`. You were right that the structural path dropped every bound the text path had — I replaced `capForDisplay` on the object path without replacing what it was doing.

**2. Object keys.** Reproduced exactly: `{["k".repeat(2_000_000)]: "value"}` came back `truncated=false`, 2,000,013 characters on one line. The worst part is the one you named: with no banner there is no **Copy** and no **Download**, so the value was both unrenderable and unreachable. `clip()` now runs on keys as well as string values, and sets `truncated` either way.

**1. No size bound.** A character budget (`PREVIEW_MAX_CHARS`, 32 KiB) now bounds the whole projection rather than only the node count. Chasing it turned up a third shape neither of us had: **a content-only budget does not bound depth**, because `JSON.stringify(…, null, 2)` indents by nesting level, and a 1200-deep value is megabytes of leading spaces before a single leaf is reached. So each entry is charged `ENTRY_PREVIEW_CHARS + depth * INDENT_PREVIEW_CHARS`, which makes depth pay for itself.

Same shapes, serialized preview:

| | before | after |
|---|---|---|
| one 2 MB key | 2,000,013 chars, 1 line, **no banner** | 518 chars, 3 lines, longest 514, banner |
| 1200-deep nesting | 3,030,212 chars, 3,003 lines, 89.8 ms | 96,234 chars, 525 lines, longest 363, 0.9 ms |
| 1000 × 600-char strings | 51,704 chars, 103 lines | 33,092 chars, 67 lines |
| Ansible, 6000 tasks | 19,429 chars, 513 lines | unchanged |

Two tests, both verified failing against the code you reviewed: one asserts a 2 MB key clips and flags truncated, one asserts the deep-nesting preview stays under 256 KiB (the size the text path allowed). I did not take the `capForDisplay` fallback you offered as an alternative — with keys bounded and depth charged I could not construct a shape that still needed it, and it would put broken JSON back in Monaco.

**Nit.** Fixed. `expressionPath.split(".").pop()` on `a["b.c"]` really did produce `b"].json`. It now matches the bracket form first and sanitises what is left, so an awkward key degrades to underscores rather than quotes.

**Also, separately** (@elevatebart's review of the screenshot, not yours): the explorer's download button moved onto the banner's own row, right-aligned and vertically centred via the `title` slot, and the banner's inset is now padding on a wrapper rather than a margin on the alert — the alert is `width: 100%`, so the margin was pushing it 32px past the panel and making the whole panel scroll sideways.

`check:types`, `check:ts-any`, `lint` clean; 259 files, 2296 tests.

---

MilosPaunovic (approved):
Re-measured at `0fb361be` on the same harness and inputs as the last round. Both blockers are fixed:

| case | round 3 longest / total | round 4 | editor payload r3 -> r4 |
|---|---|---|---|
| 1000 nodes x 600-char strings | 1876 / 6179 ms | **585-684 / 893-1990 ms** | ~500 KiB -> 34 KiB |
| one 2 MB object key | 6524 / 12538 ms | **469-849 / 1253-2664 ms** | 2 MB -> 521 chars, banner + **Copy** + **Download** |

The Ansible row moved from 432 to 544-811 ms, but two back-to-back runs of the same build gave 811 and 544, so that is harness noise rather than a change. 15 unit tests pass locally.

Two nits, neither blocking:

- **The character budget undercounts at depth.** `entryCost(depth)` charges the opening line's indent, but pretty-printed `JSON` writes a closing `}` at the same indent, so the real cost is roughly double. Against `PREVIEW_MAX_CHARS = 32 KiB`, a flat wide value lands at 34 KiB (accurate) while the 1200-level value from your own new test lands at 99 KiB. Still far better than round 3's 1.13 MiB and the test's `< 256 KiB` assertion holds, so this is only worth knowing about: on that input the `JS` side is 56 ms (`boundForDisplay` 2 ms, preview stringify 2 ms, the full `JSON.stringify` behind the size banner 42 ms) and the remaining ~1.2 s is Monaco.
- **Clipped keys can collide.** Two keys longer than `PREVIEW_MAX_STRING_CHARS` that share their first 500 characters clip to the same string, so the second overwrites the first and the preview shows one entry where the value had two, with no `…` marker. `truncated` is set by `clip`, so the banner and **Download** are still offered; only the entry count is silently off.

`LGTM`. `CI` still has `Frontend - Tests`, `E2E - Tests`, `trigger-ee` and the docker publish in progress. The `releases/v2.0.x` backport still needs its own `PR` against that branch.

elevatebart:
Both nits taken in `6022b6f87`, since neither was more than a few lines.

**Closing indent.** You were right that `entryCost` only charged the opening line. A container now also pays for its own closing line at its own depth, which is the missing half. Your 1200-level input: **99 KiB → 49 KiB over 375 lines**, against the 32 KiB budget. Still over, because the budget is charged as the projection is built rather than measured after, but it is now the right order rather than triple.

**Key collisions.** A clipped key that matches one already in the preview is dropped instead of overwriting it, so the `…` count includes it. `{[p+"a"]: 1, [p+"b"]: 2, other: 3}` with a 600-character shared prefix now previews as the survivor, `other`, and `"…": 1`; before it was two entries with no marker. One test, which fails with `expected [ …(2) ] to have a length of 3 but got 2` when the guard is removed.

`check:types`, `check:ts-any`, `lint` clean; 259 files, 2297 tests. Merging once this round of CI is green.

---

Issue #10219, opened 2025-07-21
Live demo can't be visited


mogudian:
### Describe the issue

Link: https://demo.kestra.io/ui/login?auto

DNS_PROBE_FINISHED_NXDOMAIN

### Environment

- Kestra Version: develop

PR #19334, opened 2026-09-10
feat(executions): timeline view for the Executions page

flcarre:
Closes https://github.com/kestra-io/kestra-ee/issues/7231.

## Problem

The Executions page could only count runs, never place them. A bar chart per day answers "how many failed" but not "when did it start failing", "did these three runs overlap", or "what was running at 14:07". Picking a time window meant guessing a duration from a dropdown and checking the result afterwards.

## Fix

A `Counts | Timeline` segmented control in `KsFilter`'s existing `#extra` slot — a list-page tab means "filter", not "view", so this is a view switch, not a new tab.

The Timeline view lays executions on a real time axis, drilling namespace → flow → concurrent runs:

- **Drill-down** through `KsBreadcrumb`, with per-level run and failure counts, scoping the table below it as well as the chart.
- **Concurrency lanes** per flow, so overlapping runs stack instead of overprinting; a burst too dense to separate collapses into one bucketed lane rather than a wall of sub-pixel slivers.
- **Density bucketing** per row once a bar would render under ~3px, evaluated against real positions on the axis rather than rendered width.
- **Bar states** coloured from the existing `--ks-chart-*` tokens (not `--ks-status-*`, which is missing most state mappings). No design-system token was added: those tokens already carried the corrected KILLED=dark-red / KILLING=yellow / PAUSED=orange / QUEUED=grey mapping that the issue's own colour brief got wrong. KILLED gets a diagonal hatch, RUNNING/KILLING an animated stripe distinct from QUEUED's static one.
- **Hover tooltip and click popover** on a bar, with "Open execution" and "Show only this flow".
- **Legend as a state filter**, synced with the filter-bar chip and the table; a dimmed state keeps its entry and its count instead of vanishing from the legend.

The time control is a **density histogram** rather than a slider. The track draws execution density over twice the visible window, stacked by state with failures at the base, so the window is chosen by pointing at the runs instead of guessing a duration:

- The violet band is the current window. Drag inside it to move it keeping its duration, drag an edge to resize from that side, drag outside it to draw a new one.
- Hovering a bar reports its time, its runs and its failures.
- The track labels the time it spans, and stops at now — no execution can land in the future.
- A continuous baseline runs the full width, so an empty stretch reads as "nothing ran" rather than as a rendering gap.
- Everything drives the chart, the table and the URL through the same `startDate`/`endDate` execution filters the app already supports.

Two pieces of logic are extracted rather than duplicated: `useExecutionsQueryScope.ts` takes the namespace/flow/status filter scoping `Executions.vue` already had (removing 5 `any`s on the way), and `executionsTimeline.ts` holds the pure grouping, lane-packing and bucketing, unit tested with no component mounted. The histogram reuses that same bucketing through `densitySeries()`, a variant that keeps empty buckets so the track can draw its baseline.

**Note on the commit history:** a `KsRangeSlider` design-system component appears and is then removed. The scrubber replaced it, and shipping a component to one release only to retire it in the next runs against the deprecation contract in `ui/AGENTS.md`, so it was dropped rather than left in. It is absent from the diff against `develop`.

## Scope decisions

- **No backend change.** The bounded `/executions/search` endpoint the table already uses supplies everything: individual bars, and per-namespace/flow failure counters aggregated client-side from the same fetched list, per the issue's own "only if cheap" escape hatch. The histogram's wider context window costs one additional call on the same endpoint, mirroring what the legend already does when a state is dimmed.
- **Retention wall not implemented.** OSS has no execution/log retention concept in `miscStore.configs` to hang it off — it is an EE feature.
- **Scheduled-but-not-yet-run executions are not rendered.** A "Now" bound is drawn, but filling the axis with upcoming runs needs a different data source (trigger next-fire-time, not the execution list) and is the explicit scope of kestra-io/kestra#10429.
- Also addresses the "chart already eats the viewport" complaint (kestra-io/kestra#10219): the collapsed timeline stays chart-slot-sized so the table underneath stays visible; Expand takes the full page.

## Demo

https://github.com/user-attachments/assets/bc3e138c-339e-493c-bf2b-cc965dce8887

## Evidence

Recorded against a running instance on this branch, light and dark: the histogram and its labels, the hover readout, moving the window, resizing it by an edge, brushing a new one, widening back with zoom-out and Now, and the namespace → flow drill-down.

Browser QA on the same instance found and fixed, in order: a bar-click popover blocked by a duplicated tooltip trigger, drill-down that did not scope the table, a legend that did not sync the filter bar, a 422 from sending `timeRange` alongside `startDate`/`endDate`, a histogram whose domain reached into the future, buckets whose dominant state hid every failure behind its successes, and a drag that selected the axis labels.

`npm run check:types`, `npm run test:unit` and `npm run test:lint` pass. Timeline coverage lives in `executionsTimeline.spec.ts`, `useTimelineRange.spec.ts`, `TimelineScrubber.spec.ts`, `TimelineRow.spec.ts`, `TimelineBar.spec.ts`, `TimelineToolbar.spec.ts` and `ExecutionsTimeline.spec.ts`.


flcarre:
## QA Report — Executions Timeline

**Verdict: FAIL** — one scenario the PR explicitly targets does not work: clicking an execution bar does not open the detail popover.

### Summary

| # | Scenario | Result |
|---|---|---|
| 1 | Non-regression (Counts view) | ✅ PASS |
| 2 | Counts ⇄ Timeline switch | ✅ PASS |
| 3 | Collapsed default (chart-slot-sized, table visible) | ✅ PASS |
| 4 | Expand/collapse | ⚠️ NOT VERIFIED (see notes) |
| 5 | Time-range control (zoom/pan/Now, syncs URL) | ✅ PASS |
| 6 | Drill-down (namespace → flow → concurrent lanes) | ✅ PASS, with a gap (table doesn't scope) |
| 7 | Bar states / colors | ✅ PASS |
| 8 | Hover tooltip | ✅ PASS |
| 8b | Click bar → popover | ❌ **FAIL** |
| 9 | Legend-as-filter | ✅ PASS, with a gap (doesn't sync filter bar/table) |
| 10 | Loading / empty states | ✅ PASS (empty); loading not independently screenshotted |
| 11 | Light + dark mode | ✅ PASS |

### ❌ FAIL — Click a bar does not open the popover

**Evidence:** `document.elementFromPoint()` at a bar's own center returns a *different* element than the bar itself:

```
querySelector('.timeline-bar') → <button class="timeline-bar" aria-label="hourly-inventory-sync — SUCCESS" role="button">
elementFromPoint(center) → <div class="timeline-bar kel-tooltip__trigger kel-tooltip__trigger">   (different element, isSameElement: false)
```

A duplicate, overlapping `<div class="timeline-bar kel-tooltip__trigger">` sits on top of the real `<button>` and intercepts the pointer. This is the same root cause as a console warning repeated on every bar render:

```
[warning] Runtime directive used on component with non-element root node. The directives will not
function as intended. Proxy(Object) at <ElPopper> at <ElTooltip> at <KsTooltip> at <ElOnlyChild>
at <ElPopperTrigger> at <ElTooltipTrigger> at <ElPopper> at <ElTooltip> at <ElPopover> at <KsPopover>
at <TimelineBar> at <TimelineRow>
```

`ElOnlyChild` (used internally by `KsTooltip`'s trigger) is supposed to merge into its single child without adding a DOM node; something about how `TimelineBar` composes `KsTooltip` + `KsPopover` breaks that merge, so it renders an extra wrapper `<div>` instead. Consequence: hover tooltips work (the wrapper itself carries the tooltip), but a real click never reaches the underlying `<button>` — confirmed two ways: a synthetic `el.click()` on the button did nothing (no popover, no DOM node matching `[class*=popover]` ever appears), and a real Playwright click via ref times out on actionability (5000ms) because the target is covered.

### Gaps vs. the accepted spec (not blocking, but real)

- **Drill-down doesn't scope the table.** Clicking into a namespace correctly scopes the timeline (breadcrumb, per-flow rows) but the table below still shows rows from other namespaces. Spec: "Clicking a namespace or a flow drills the timeline **and the table** into it."
- **Legend-as-filter doesn't sync the filter bar/table.** Clicking "Failed" in the legend correctly dims FAILED bars (`opacity: 0.2`, confirmed via computed style) but the filter-bar's "State" chip and the table stay unfiltered. Spec: "synced with the state filter chip in the filter bar and with the table below."
- **Drill-down scope isn't in the URL.** Time range zoom/pan rewrite the URL correctly, but the namespace/flow drill scope is client-state only — a page reload loses it. Minor, but means a drilled-in view can't be shared/bookmarked.

### What worked well

- **Concurrency payoff is genuinely good.** Seeded 3 overlapping 8-second runs of the same flow (2s apart); drilling into that flow rendered them on 3 distinct, clearly staggered lanes — this is the core value proposition from the issue and it delivers.
- **Density-bucket LOD threshold works**: at 24h zoom the 15s cluster of activity rendered as a thin tinted bucket near "Now"; narrowing to the actual 2-minute window revealed individual, correctly-colored, correctly-positioned bars.
- **Empty state** is well done: "No executions in the selected range and filters" plus working "Reset filters" / "Widen to 7 days" actions.
- **Dark mode** looks correct, proper token usage throughout.
- **Non-regression confirmed**: existing Counts chart + table works unchanged after the toggle refactor.
- Bars carry real `aria-label`s (`"hourly-inventory-sync — SUCCESS"`), a genuine accessibility win.

### Environment notes (not app bugs)

- The Counts/Timeline radio inputs (`role=radio`) reliably fail Playwright's real-click actionability check (likely zero-size, visually hidden per common styled-radio patterns) — had to click via the wrapping label instead. Worth a quick look since it may affect Playwright E2E tests written directly against the radio role.
- Scenario 4 (expand/collapse) and the loading skeleton weren't independently re-verified due to shared browser-daemon contention with a concurrent unrelated session; not a defect in this PR.

### Test data used

Seeded via API (not committed): 3 flows across `qa.timeline.retail`/`qa.timeline.marketing`, 7 executions (5 SUCCESS incl. 3 deliberately overlapping, 2 FAILED).

flcarre:
## QA Re-verification — commit `98c6ff48d1`

**Verdict: PASS** — all three previously-reported issues confirmed fixed; no regressions found.

### Summary

| # | Item | Result |
|---|---|---|
| 1 | Bar click → popover (was FAIL) | ✅ **FIXED** |
| 2 | Hover tooltip (regression check) | ✅ PASS, unaffected |
| 3 | Drill-down scopes the table (was gap) | ✅ **FIXED** |
| 4 | Legend-as-filter syncs filter bar/table (was gap) | ✅ **FIXED** |
| 5 | Counts ⇄ Timeline switch | ✅ PASS |
| 6 | Expand/collapse | ✅ toggles without error |
| 7 | Zoom (URL sync) | ✅ PASS |
| 8 | Dark mode | ✅ PASS |

### 1. Bar click → popover: FIXED

Clicking a bar now opens the popover with execution id, namespace, flow, state, start/end dates, and both working action buttons ("Open execution", "Show only this flow"). The `ElOnlyChild` console warning from the original QA pass is gone entirely. Hover tooltip verified independently right after — still works, unaffected by the restructuring.

### 2. Drill-down scopes the table: FIXED

Drilled `All → qa.timeline.retail`: table narrowed from 7 rows to the 5 `qa.timeline.retail` rows, matching the timeline's own scope. Drilled further into `hourly-inventory-sync`: table narrowed to exactly its 3 executions. Breadcrumb, timeline rows, and table stayed in sync at both drill levels.

### 3. Legend-as-filter syncs filter bar/table: FIXED (slightly beyond the original gap description)

Clicking "Failed" in the legend at the top level added a real `State not in FAILED` chip to the filter bar, dropped the table to the 5 SUCCESS rows only, and updated the legend to show `Success (5)`. This applies an actual filter rather than just dimming bar opacity — a stronger fix than the reported gap required.

### Regression smoke

- Counts ⇄ Timeline switch: clean, no console errors.
- Zoom correctly rewrote the URL to absolute timestamps; landed on the empty state for an out-of-range window with working Reset filters / Widen to 7 days actions.
- Dark mode: correct `--ks-chart-*` token colors throughout, no light-mode bleed-through.
- No new console errors; only a pre-existing, unrelated local codegen staleness warning.

### Minor, non-blocking observation

`Escape` does not close the popover (click-elsewhere works). `KsPopover` is documented as managing Escape-to-close — worth a quick look, but not required by the accepted spec and not blocking.

flcarre:
## QA re-verification — polish batch (commit `0178efdfa4` and later)

**Verdict: CHANGES REQUESTED** — 2 real bugs found, everything else confirmed fixed.

### Summary

| # | Item | Result |
|---|---|---|
| 1 | Toolbar consolidation (range pill + Relative/Absolute popover) | ✅ **FIXED** |
| 2 | Time axis with tick labels + persistent "Now" label | ✅ **FIXED** |
| 3 | Drill-down scopes the table (namespace + flow levels) | ✅ **FIXED** |
| 4 | Concurrent-run lanes | ✅ PASS |
| 5 | Bar click → popover | ✅ **FIXED** |
| 6 | Escape-to-close on popover | ✅ **FIXED** |
| 7 | Legend-as-filter syncs filter bar + table | ✅ PASS (confirmed again) |
| 8 | Light + dark mode | ✅ PASS |
| 9 | Expand grows to real page height | ❌ **FAIL** (new bug) |
| 10 | Density bucketing at low zoom | ❌ **FAIL** (new bug) |

### ✅ Confirmed working

1. **Toolbar** — single range pill opening a `Relative`/`Absolute` popover, matching `DateFilter.vue`'s established pattern. Zoom/pan/custom-range all write clean `startDate`/`endDate` URL params with no leftover `timeRange` key.
2. **Time axis** — tick labels render across the lanes, ending in a persistent "Now" label in the axis strip (not hover-only).
3. **Drill-down** — namespace → flow breadcrumb works at both levels; the table correctly scopes to match (15 → 12 rows on namespace drill, all matching namespace).
4. **Concurrent lanes** — 3 overlapping runs of the same flow render on 3 distinct stacked lanes; the core payoff of the feature reads clearly.
5. **Bar click → popover** — opens reliably, full content (execution id, namespace, flow, `KsExecutionStatus` badge, timing, both action buttons).
6. **Escape-to-close** — confirmed closing the popover (the design-system-level `KsPopover` fix holds).
7. **Legend-as-filter** — clicking "Failed" adds a real `State not in FAILED` filter-bar chip and drops the table from 15→13 rows — a genuine filter, not just visual dimming.
8. **Dark mode** — correct token usage throughout, no light-mode bleed-through.

### ❌ Bug 1 — Expand doesn't push the table below the fold, it destroys it

The table's scroll wrapper (`.kel-scrollbar__wrap`) reports `scrollHeight: 710, clientHeight: 15` in the expanded state — squeezed to a **15px-tall sliver** with 710px of content trapped inside, and not scrollable (the page's own `scrollHeight` equals `innerHeight`, so there's no scrollbar to reveal it either). The table is present in the DOM (15 rows confirmed via query) but effectively invisible and unusable between the legend and the pagination footer. This is worse than the original "doesn't take the page" finding — growing the timeline now actively breaks the table's layout rather than just failing to grow.

### ❌ Bug 2 — Density bucketing never triggers, even in the textbook case

Zoomed out to a 32-day range with all 15 seed executions clustered within a single ~1-hour window — true bar width is many orders of magnitude under any reasonable "3px" threshold. Expected: these collapse into one tinted density bucket per row. Actual: zero `[data-bucket-state]` elements found, 15 individual `[data-state]` bars, each clamped to a `width: 0.3%` floor (~4px), overlapping at nearly identical positions with no visual blending. The `color-mix()` bucket styling from the polish batch appears unreachable — likely the bucketing decision runs against the post-floor rendered width (which never drops below the floor, so the per-row threshold check never fires) rather than the true pre-floor width. Since "resolution is a consequence, not a setting" is one of the feature's named design decisions, this is a functional gap, not cosmetic.

### Not independently re-verified this pass

KILLED/RUNNING non-color signals (hatch/stripe) — no KILLED or RUNNING executions exist in the current seed data; confirmed via code reading in the earlier DS-compliance review, not re-screenshotted live here.

ui/src/components/executions/timeline/ExecutionsTimeline.vue
```diff
@@ -0,0 +1,497 @@
...
+    import type {ApiLightExecution, PagedResultsApiLightExecution} from "@kestra-io/kestra-sdk"
+
+    import TimelineToolbar from "./TimelineToolbar.vue"
+    import TimelineRow from "./TimelineRow.vue"
+    import ChartLegend from "../../dashboard/sections/ChartLegend.vue"
+    import {useExecutionsStore} from "../../../stores/executions"
+    import {useExecutionsQueryScope} from "../../../composables/useExecutionsQueryScope"
+    import {useTimelineRange} from "../../../composables/useTimelineRange"
+    import {useExecutionFilter, useFlowExecutionFilter} from "../../filter/configurations"
+    import {groupByNamespace, countByState, buildAxisTicks, type TimelineExecution} from "../../../utils/executionsTimeline"
```
flcarre: 🔴 [Correctness] `scopedFailedCount` reimplements `isFailedLikeState` and disagrees with it
Problem: `groupByNamespace()` (used for every namespace/flow row) computes "failed" via the shared, unit-tested `isFailedLikeState()` in `executionsTimeline.ts`, which counts `FAILED`, `KILLED`, `KILLING`, `WARNING`. `scopedFailedCount` below (used for the breadcrumb total, and directly as `row.failed` at flow-view level) reimplements the same concept inline with `["FAILED", "KILLED", "WARNING"]` — missing `KILLING`. Any KILLING execution in the visible window makes the breadcrumb's "X failed" tag under-count relative to the sum of the per-row failed counts shown right below it.
Fix: import and reuse the shared helper here instead of a second hand-rolled list.
```suggestion
    import {groupByNamespace, countByState, buildAxisTicks, isFailedLikeState, type TimelineExecution} from "../../../utils/executionsTimeline"
```

ui/src/components/executions/timeline/ExecutionsTimeline.vue
```diff
@@ -0,0 +1,497 @@
...
+        const raw = route.query[STATE_EXCLUDE_KEY]
+        const value = Array.isArray(raw) ? raw[0] : raw
+        return new Set((value ?? "").split(",").filter(Boolean))
+    })
+
+    const showBreadcrumb = computed(() => props.flowId === undefined)
+
+    const scopedExecutions = computed(() => rawExecutions.value)
+    const scopedFailedCount = computed(() =>
+        scopedExecutions.value.filter(e => ["FAILED", "KILLED", "WARNING"].includes(e.state)).length,
```
flcarre: 🔴 [Correctness] Same finding as the import-line comment above — this is the line that needs to call the shared helper instead of a second, inconsistent literal list.
```suggestion
    const scopedFailedCount = computed(() =>
        scopedExecutions.value.filter(e => isFailedLikeState(e.state)).length,
    )
```

ui/src/components/executions/timeline/ExecutionsTimeline.vue
```diff
@@ -0,0 +1,497 @@
...
+
+    async function fetchExecutions() {
+        loading.value = true
+        error.value = undefined
+        try {
+            // Search executions returns the light DTO (ApiLightExecution), not the full Execution;
+            // the store's findExecutions is declared Promise<any>, so this cast is what vue-tsc would
+            // otherwise infer on its own.
+            const response = await executionsStore.findExecutions(loadQuery({
+                size: MAX_FETCHED_EXECUTIONS,
```
flcarre, line 374: 🟠 [Correctness/Data] Fetch is silently capped at 1000 executions with no truncation signal
Problem: `PagedResultsApiLightExecution` carries both `results` and `total`, but only `results` is used — `response.total` is discarded. For a busy namespace/tenant with more than `MAX_FETCHED_EXECUTIONS` (1000) executions in the selected range, the breadcrumb totals, the per-row/failed counts, and the legend counts all silently reflect only the most recent 1000 runs (sorted `state.startDate:desc`) with no indication to the user that the numbers are partial — an on-call engineer could read "3 failed" in a window that actually had many more.
Fix: compare `response.total` to `response.results.length` and surface a "showing latest 1000 of N — narrow the range" notice (e.g. next to the breadcrumb stats) when truncated.

ui/src/components/executions/timeline/ExecutionsTimeline.vue
```diff
@@ -0,0 +1,497 @@
...
+        }
+    }
+
+    function onShowOnlyFlow({namespace, flowId}: {namespace: string; flowId: string}) {
+        drillNamespace.value = namespace
+        drillFlowId.value = flowId
+    }
+
+    function onLegendToggle(state: string) {
+        const next = new Set(dimmedStates.value)
```
flcarre, line 334: 🟠 [Correctness] Legend "click-to-dim" toggle removes data instead of dimming it, and the toggled item then vanishes from the legend itself
Problem: clicking a legend entry writes `filters[state][NOT_IN]=X` to the route query. Since `"state"` is a supported field on `executionFilter`/`flowExecutionFilter`, that same key survives `dropUnsupportedFilters` in `useExecutionsQueryScope` and is sent to `/executions/search` by this component's own `fetchExecutions` (triggered by the `fetchKey` watcher, which includes `route.query`). That means state X's executions are excluded from `rawExecutions` entirely — not just flagged — so:
1. `dimmedStates`/`TimelineBar`'s `dimmed` opacity styling never actually applies (no bar of that state is ever fetched to dim).
2. `legendItems` is derived from `countByState(scopedExecutions.value)`, i.e. the same already-filtered list, so the entry the user just clicked disappears from the legend on the next fetch — leaving no way to re-enable that state from the legend itself (only via the separate filter-bar chip).
This contradicts the "click-to-dim" interaction described in the PR (a reversible visual toggle) with an actual hide-and-vanish. Please confirm live whether this was exercised in QA — a static screenshot wouldn't show the item disappearing, only interacting with it a second time would.
Fix: either keep the legend's own item list independent of the state-exclude filter (e.g. seed `legendItems` from an unfiltered fetch, or from the full historical set of states seen before the current exclude filter was applied) so the toggled entry stays visible and clickable, or rename/restyle the interaction to make clear it hides rather than dims.

ui/src/components/executions/timeline/ExecutionsTimeline.vue
```diff
@@ -0,0 +1,497 @@
...
+                .filter((execution): execution is ApiLightExecution & {state: {startDate: string}} => Boolean(execution.state?.startDate))
+                .map((execution) => ({
+                    id: execution.id,
+                    namespace: execution.namespace,
+                    flowId: execution.flowId,
+                    state: execution.state.current,
+                    startMs: new Date(execution.state.startDate).getTime(),
+                    endMs: execution.state.endDate ? new Date(execution.state.endDate).getTime() : Date.now(),
+                }))
+        } catch {
```
flcarre: 🟡 [A09 Logging] Error is swallowed with no diagnostic trail
Problem: `catch {}` discards the actual error entirely; only a generic i18n string reaches the user, and nothing is logged for debugging a real production failure (network error, 5xx, malformed response). Other components in this codebase (e.g. `SetLabels.vue`) `console.error` the caught error before showing a toast/message.
```suggestion
        } catch (fetchError) {
            console.error("Failed to load timeline executions", fetchError)
            error.value = t("executionsTimeline.error.description")
```

flcarre (commented):
Final cumulative review of the branch (all rounds). Frontend-only, confirmed — no backend/Java files in the diff. DS/UX and browser-QA passes already covered visuals and interaction fidelity, so this pass focuses on data-correctness, test coverage, and code quality. Full summary posted separately.

flcarre:
## Kestra OSS Code Review

Reviewed the full cumulative diff against `develop` (raw PR diff, all 31 files / 2906 additions), not just the latest commits. Confirmed **this PR is entirely frontend** — no Java/backend files anywhere in the diff (`ui/packages/design-system`, `ui/src`, translations only). DS-compliance and UX-fidelity have already been reviewed separately, so this pass focused on correctness, tests, security, and performance.

### Kestra Guidelines (backend) — N/A, no backend files changed

### Frontend / Design System — no new findings beyond what's already been reviewed
Spot-checked independently: `KsSegmented`/`#extra` slot both genuinely exist and are wired correctly (not a silent no-render), `Ks*` components used throughout (no direct `element-plus` imports), tokens used consistently (`--ks-chart-*`, `--ks-spacing-*`), all locale files translated with no literal `{{ }}` in any translation value, `KsPopover` Escape-key fix is a real, well-reasoned bug (verified the watch-ordering argument against Vue 3's immediate-watch timing — it's correct).

### Security (OWASP Top 10) — 1 finding
- 🟡 A09: `fetchExecutions`'s `catch {}` swallows the real error with no logging (see inline comment). No injection/XSS/access-control issues found — all interpolation goes through Vue's auto-escaping, no `v-html`, and the fetch is scoped through the existing, already-authorized `/executions/search` endpoint.

### Correctness — 3 findings (posted inline)
- 🔴 `ExecutionsTimeline.vue`'s `scopedFailedCount` reimplements the shared, unit-tested `isFailedLikeState()` with an inline list that's missing `KILLING`. This makes the breadcrumb's "X failed" tag (and the flow-level row's failed count, which reuses the same value) disagree with the sum of the per-row failed counts computed via `groupByNamespace()` just below it, whenever a KILLING execution is in the visible window. Verified `KILLING` is a real `State` value and that `groupByNamespace` does count it.
- 🟠 `MAX_FETCHED_EXECUTIONS = 1000` truncates silently: `response.total` (present on `PagedResultsApiLightExecution`) is never compared to `response.results.length`, so every count/legend/breadcrumb in the timeline can silently under-report for any namespace/tenant with more than 1000 executions in the visible range, with zero indication to the user.
- 🟠 The legend's "click-to-dim" toggle is wired through the real `filters[state][NOT_IN]` query key (by design, to sync with the table), which means the toggled state's executions are excluded from the fetch entirely rather than dimmed — and since the legend's own item list (`countByState`) is derived from that same filtered fetch, the just-toggled legend entry disappears with no way to re-enable it from the legend. This is a static-trace-verified structural consequence of the current wiring, not a guess — but I'd recommend a quick live re-check given how many browser QA passes already happened, in case something else compensates for it that isn't visible from the source alone.

### Test quality — 2 findings
- 🟠 `ExecutionsTimeline.vue` (497 lines — the orchestrator that owns fetch, drill-down, the breadcrumb, and the legend wiring, i.e. exactly where the `scopedFailedCount` bug above lives) has **no dedicated test file**, unlike every other component added in this PR (`TimelineBar`, `TimelineRow`, `TimelineToolbar`, `executionsTimeline.ts` utils all have specs). The one component most likely to harbor integration bugs is the one with zero coverage.
- 🟡 `TimelineBar` ends up with two separate spec files exercising it — `ui/src/components/executions/timeline/TimelineBar.spec.ts` (colocated, added first) and `ui/tests/unit/components/executions/timeline/TimelineBar.spec.ts` (added later, in the fix-commit for the popover-trigger bug). Both are picked up by the same Vitest project (`vitest.config.unit.js` has no `include` narrowing and doesn't exclude `src/**`), so both run in CI with independent mock/stub setups for the same component. `ui/AGENTS.md` states tests should be "colocated next to the component" — worth consolidating into one file. Same inconsistency for `useTimelineRange.spec.ts`, which landed in `ui/tests/unit/composables/` instead of colocated next to `useTimelineRange.ts` like every other composable/util added in this PR.

### Performance — no new findings
`assignLanes`/`bucketize`/`shouldBucketRow` are bounded by the same 1000-execution fetch cap, greedy-interval-scheduling is the right algorithm for lane packing, `ResizeObserver` is disconnected in `onUnmounted`, no repository/N+1 concerns (frontend-only).

**Verdict:** REQUEST CHANGES — one correctness bug (the failed-count mismatch) plus the missing truncation signal and the legend/dim behavior should be resolved or explicitly confirmed before merge; the missing `ExecutionsTimeline.vue` test coverage should be added given it's where the correctness bug lives.

ui/src/components/executions/timeline/ExecutionsTimeline.vue
flcarre: Fixed: `scopedFailedCount` now calls the shared `isFailedLikeState()` helper (imported from `executionsTimeline.ts`) instead of a second inline list, so it can never drift from the per-row counts again. Added a regression test in the new `ExecutionsTimeline.spec.ts` that seeds a KILLING execution and asserts the breadcrumb's failed tag agrees with the sum of the per-row failed counts (fails without the fix, confirmed locally before restoring it).

ui/src/components/executions/timeline/ExecutionsTimeline.vue
flcarre: Same fix as the sibling import-line comment - `scopedFailedCount` now delegates to `isFailedLikeState()`. Covered by the new regression test in ExecutionsTimeline.spec.ts.

ui/src/components/executions/timeline/ExecutionsTimeline.vue
flcarre: Fixed: fetches now compare `response.total` to `response.results.length` and, when truncated, show a KsAlert (type=warning) near the breadcrumb/count area with `executionsTimeline.truncated.description` ("Showing the latest {shown} of {total} executions..."), translated into all locales. Not attempting unbounded fetching in this PR, per the review's guidance - just making the cap visible. Covered by two tests in ExecutionsTimeline.spec.ts (notice shown when truncated, hidden when not).

ui/src/components/executions/timeline/ExecutionsTimeline.vue
flcarre: Fixed. Kept the real filters[state][NOT_IN] exclusion (confirmed intentional per the review), but the legend's item list is now derived from a second, unfiltered-by-state fetch (same namespace/flow/range scope, exclude key stripped) rather than from the already-filtered scopedExecutions - so a toggled-off state's entry stays visible with its real count and can be clicked again to re-include it. The main row/table data still comes from the filtered fetch, unaffected. Didn't need to touch ChartLegend.vue itself - its own toggledOff/opacity handling already does the right thing once the item stops disappearing from its items prop. Covered by a new test that toggles a state off then back on and asserts the legend entry and row counts at each step.

ui/src/components/executions/timeline/ExecutionsTimeline.vue
flcarre: Fixed: the catch now does console.error("Failed to load timeline executions", fetchError) before setting error.value, matching the SetLabels.vue convention. error.value already drives the existing KsAlert type="error" state in the template, so the catch feeds the same user-facing error state as before, just with a diagnostic trail now.

flcarre:
Addressed the remaining two findings from the cumulative code review that weren't posted as inline comments:

- **Missing `ExecutionsTimeline.vue` test coverage**: added `ui/src/components/executions/timeline/ExecutionsTimeline.spec.ts`, covering the failed-count/per-row agreement regression (KILLING execution present), the truncation-notice behavior (shown when `total > results.length`, hidden otherwise), and the legend-toggle-preserves-entry behavior (toggle a state off, confirm it stays visible with its real count and the table narrows, toggle back on, confirm it's restored).
- **Duplicate `TimelineBar.spec.ts` files**: merged the two popover-click regression tests from `tests/unit/components/executions/timeline/TimelineBar.spec.ts` into the colocated `src/components/executions/timeline/TimelineBar.spec.ts` (they weren't redundant with the colocated file's bucket/state-attribute tests, so kept both sets), then deleted the non-colocated duplicate. Also moved `useTimelineRange.spec.ts` from `tests/unit/composables/` to sit next to `useTimelineRange.ts`, matching every other composable/util added in this PR.

All four inline-commented bugs (failed-count/KILLING, truncation, legend-toggle vanishing, swallowed fetch error) are fixed and replied to individually on their threads. `npm run check:types`, `npm run test:unit` and `npm run test:lint` all pass locally.

flcarre (commented):
## Kestra OSS Code Review — Cycle 2 (re-review)

Re-verified all 5 findings from the previous REQUEST CHANGES round against the current diff (not just the reply text):

1. **Breadcrumb/per-row failed-count mismatch** — `scopedFailedCount` now imports and calls the shared `isFailedLikeState()` from `executionsTimeline.ts` instead of a hand-rolled `["FAILED","KILLED","WARNING"]` list. Confirmed the KILLING gap is closed and `groupByNamespace()`'s per-row counts and the breadcrumb tag now derive from the same helper. Covered by a new regression test seeding a KILLING execution.
2. **Silent 1000-execution truncation** — `fetchedTotal`/`isTruncated` now compare `response.total` to `rawExecutions.length` and render a `KsAlert type="warning"` with `executionsTimeline.truncated.description`. Verified the key is translated in all 13 locale files (not just `en.json`) and the fingerprints file was updated, so `translations:check` stays clean. Two tests cover the shown/hidden cases.
3. **Legend-toggle-vanishes bug** — this is the one I dug into most, per the task. `legendExecutions` is now populated from a second fetch that reuses the exact same `query` object built by `loadQuery` (so namespace/flowId/time-range scope match the primary fetch exactly) with only the `filters[state][NOT_IN]` key stripped. It's correctly gated behind `dimmedStates.value.size > 0` — when nothing is toggled off, `legendExecutions.value = rawExecutions.value` is reused with **no second network call**, so the common path isn't doubled. `STATE_EXCLUDE_KEY` matches the literal query key produced by `useExecutionsQueryScope`/`keepSupportedFilters` (confirmed by tracing `loadQuery`), so the strip actually works. New test toggles a state off then back on and asserts the legend entry (with its real count) survives and the row totals move correctly both ways. This resolves the finding.
4. **Duplicate spec files** — confirmed only one spec per component/util now (`ExecutionsTimeline.spec.ts`, `TimelineBar.spec.ts`, `TimelineRow.spec.ts`, `TimelineToolbar.spec.ts`, `executionsTimeline.spec.ts`, `useTimelineRange.spec.ts`), matching the file list from `gh pr view --json files`.
5. **Silent `catch {}`** — now `console.error("Failed to load timeline executions", fetchError)` before setting the existing `error.value`, matching the `SetLabels.vue` convention. No sensitive data logged.

No scope change in this batch; the fix batch only addresses the 5 flagged findings, consistent with the linked issue's spec (kestra-io/kestra-ee#7231).

### Kestra Guidelines (backend) — N/A (frontend-only diff, no Java files)

### Frontend / Design System — ✅ No violations
- No direct `element-plus` imports outside the DS package itself (`KsPopover`'s internal `ElPopover` is pre-existing/in-package).
- No hex/`rgb`/`--el-*`/`--bs-*`/SCSS `$color` in feature code; bars/legend use `--ks-chart-*` tokens via `color-mix()`.
- No `:deep()` or `.el-*` overrides introduced.
- All new strings route through `$t()`; `truncated.description` translated into all 12 non-English locales with matching fingerprints.

A09: the previously-swallowed error now has a diagnostic trail without leaking response bodies/secrets to the console.

### Performance — 2 non-blocking suggestions
🟡 **Sequential legend fetch** — `ExecutionsTimeline.vue#fetchExecutions` (~line 376-386): the primary and legend fetches are independent (both built from the same `query`/`legendQuery`, neither depends on the other's result) but are `await`ed sequentially. When a state is toggled off, this doubles the round-trip latency unnecessarily. Consider `Promise.all([...])` to run them concurrently — cuts latency roughly in half in that path with no behavior change.

🟡 **Legend fetch not covered by the truncation notice** — the legend's own unfiltered fetch is still capped at `MAX_FETCHED_EXECUTIONS` (1000) and isn't checked against its own `total`. If the *unfiltered* result set exceeds 1000 while the *filtered* primary result stays under it, the legend's per-state counts could themselves be silently truncated with no warning — a narrower recurrence of finding #2, scoped only to the legend. Edge case (needs >1000 executions in-range while a state filter is active), not blocking, but worth a follow-up if it comes up in practice.

Neither blocks merge. All 5 targeted fixes are correct, tested, and match what was asked.

**Verdict:** APPROVE (posted as COMMENT — GitHub blocks self-approval on your own PR)

PR #19341, opened 2026-09-10
fix(triggers): filter typing and relative date metadata

### LINKED ISSUE #19302: Filter components: replace explicit `any` with real types
7 explicit `any` across 4 files in `src/components/filter`, from `ui/scripts/explicit-any/baseline.json`:

| any | file |
| --- | --- |
| 2 | `configurations/executionFilter.ts` |
| 2 | `configurations/logFilter.ts` |
| 2 | `configurations/triggerFilter.ts` |
| 1 | `configurations/flowFilter.ts` |

ZHCHJ888:
Closes https://github.com/kestra-io/kestra/issues/19302.

### ✨ Description

Replaces the seven explicit `any` casts in the execution, flow, log, and trigger filter configurations with `FilterConfiguration["keys"]`, removing all four files from the explicit-any baseline.

Fixes the trigger time-range value provider to read `options.meta` according to the filter contract, so `Last triggered date` uses past relative dates. The regression test now lives in `triggerFilter.spec.ts`.

### 🎨 Frontend Checklist

- [x] Type checking passes (`vue-tsc` for design-system, app, and tests)
- [x] Code builds without errors (`vite build --emptyOutDir`)
- [x] Affected unit tests pass (3/3)
- [x] Lint passes (`npm run lint`)
- [x] Explicit-any baseline check passes

### 📝 Additional Notes

No translations or visual layout changed.

On Windows, the full unit suite passed 2451/2452 tests across 267/268 files. The only failure is an unrelated path-separator mismatch in `noLegacyErrorFields.spec.ts`, which reports `src\\utils\\kestraHttp.ts` instead of the POSIX path expected by the test.


ui/src/components/filter/configurations/executionFilter.ts
```diff
@@ -61,7 +61,7 @@ export const useExecutionFilter = (): ComputedRef<FilterConfiguration> => {
                         },
                         searchable: true,
                     },
-                ] : []) as any,
+                ] : []) as FilterConfiguration["keys"],
```
Piyush-r-bhaskar, line 64: Correction on my earlier comment: this is right as-is. `logExecutionsFilter.ts:102` already used `as FilterConfiguration["keys"]` on develop, so this makes all 8 sites consistent. Ignore the `satisfies` suggestion, that would be a separate cleanup across all of them.

ui/src/components/filter/configurations/triggerFilter.ts
```diff
@@ -67,16 +67,16 @@ export const useTriggerFilter = (): ComputedRef<FilterConfiguration> => {
...
                 {
                     key: "timeRange",
                     label: t("filter.timeRange_trigger.label"),
                     description: t("filter.timeRange_trigger.description"),
                     comparators: [Comparators.EQUALS],
                     valueType: "select",
                     groupable: false,
-                    valueProvider: async (meta?: FilterMeta) => {
-                        return meta?.dateFilter === "LAST_TRIGGERED_DATE"
+                    valueProvider: async (options?: {meta?: FilterMeta}) => {
```
Piyush-r-bhaskar, line 78: This is a bug fix, not alignment. The DS calls `valueProvider({search, meta})`, so the old code always got `undefined` and showed "Next 5 minutes" on the triggers page while filtering backwards.

Worth retitling, `refactor(executions)` hides a `fix(triggers)`.

Keep the param count at 1 though, `FilterEditPopper` enables server-side search via `valueProvider.length > 0`.

ui/tests/unit/components/filter/configurations/executionFilter.spec.ts
```diff
@@ -39,3 +41,14 @@ describe("execution filter configurations declare taskId", () => {
         expect(config.value.keys.map((k: {key: string}) => k.key)).toContain("taskId")
     })
 })
+
+describe("trigger time range filter", () => {
```
Piyush-r-bhaskar, line 45: Good test, it does fail without the fix. Belongs in `triggerFilter.spec.ts` though.

Piyush-r-bhaskar (dismissed):
Small and careful, CI is green.

Correction to what I wrote first: the casts are fine as-is, `logExecutionsFilter.ts` already used this exact form on develop so this converges all 8 sites. Disregard the `satisfies` ask.

What is left is minor: the trigger change is a real bug fix and the title hides it, and the new test belongs in `triggerFilter.spec.ts`. Not blocking.

Piyush-r-bhaskar (approved):
LGTM

PR #19342, opened 2026-09-10
feat(topology): No-code authoring (insert, edit, undo) with usage instrumentation

### LINKED ISSUE #19338: [UI] Topology: add No-code authoring (insert, edit, undo) and instrument both editors
## Summary

The flow editor exposes No-code and Topology as two separate panels. Topology already renders every flowable correctly (If/Else, Switch, Parallel, EachParallel, Sequential, Dag, WorkingDirectory) with a server-computed layout, and already owns subflow expansion. What it lacks is authoring: adding, editing and organising tasks the way the No-code editor does. This brings the No-code authoring affordances into Topology instead of rebuilding graph layout elsewhere.

Both surfaces stay available for now, so we can measure which one users actually author in before removing either.

## Scope / Design

- `ui/` only. No backend, no schema, no EE impact.
- Topology already renders every flowable and owns subflow expansion, and `NodeMenu.vue` already emits `edit` / `delete` / `on-add-flowable-error` / `expand-subflow`. `isReadOnly` and `isAllowedEdit` are already threaded down to the nodes.
- Correction to an earlier version of this issue: the edge `+` affordance did **not** exist. `Topology.vue` forwarded an `add-task` event and the graph computed `data.haveAdd`, but `EdgeNode.vue` never emitted it and never rendered a button — the wiring was dead on the producer side. Adding that button is part of this work.
- Today `add-task` and `edit` are converted into a `topologyClick` that `useTopologyPanels` turns into a **separate No-code panel**. Change them to act in place, exactly as `onAddFlowableError` in `LowCodeEditor.vue` already does — it calls `useTaskPicker().openTaskPickerAtPath(...)`, so the pattern and the import already exist in that file.
- Editing reuses the No-code `TaskEditModal` unchanged, so the edit surface is identical across both tabs.
- Wire `useYamlUndo` to graph-originated mutations (it operates on the YAML, not on the layout), and show per-node validation state via `groupValidationIssuesByTask` + `BlockErrorBadge` on `TaskNode`.
- Nothing is deleted: both tabs, `BlockSectionLane`, `BranchLane` and `useTopologyPanels` all stay while the two surfaces coexist.

## Instrumentation (required for the coexistence period)

`utils/tabTracking.ts` currently emits only `open` / `close` per tab type, which cannot distinguish "opened Topology to look at a flow" from "built a flow in Topology". Add authoring events through the same pipeline (`makeEvent` → PostHog, same `isAnonymousUsageEnabled` gate):

- `task_added`, `task_edited`, `task_deleted`
- each carrying the originating surface (`no_code` / `topology`), plus `task_type`, and `position` for an insertion
- emitted from both surfaces so the comparison is symmetric

## Acceptance Criteria

- [ ] The `+` on a Topology edge opens the No-code task picker in place, inserting at the correct position (before/after the target task)
- [ ] Editing a task from Topology opens the same `TaskEditModal` as the No-code editor
- [ ] Undo covers mutations made from Topology
- [ ] Nodes show their validation state
- [ ] `isReadOnly` / `isAllowedEdit` are respected by the picker and the modal
- [ ] Authoring events are emitted from both surfaces with the originating surface recorded
- [ ] Both editor tabs remain available and functional
- [ ] Light and dark mode verified
- [ ] Any new i18n keys translated into all locale files
- [ ] Tests added/updated (Vitest for path resolution and tracking, Storybook for the node badge)

## Context

- Keyboard navigation on the graph was deliberately deferred when this issue was written; it was brought in during implementation, see the scope change below. The No-code list keeps its full keyboard model (`useCanvasFocus.ts`) and the two surfaces now share one keymap.
- Adjacent issues on the same panel/tab model: #17881 (Improve the UX of the flow editor) and #18996 ([UI] Flow editor: switch the active panel by default, with side-by-side as an explicit choice).

## Scope change (recorded during implementation)

Four items this issue originally listed as out of scope were built as part of #19342, because the authoring affordances they gate turned out not to stand up on their own: a canvas you can insert into but not traverse, reorder or address by keyboard is only half a surface, and the picker needed a way in that was not a mouse.

Now in scope:

- the command menu on the graph
- a graph keyboard-navigation model, sharing the No-code keymap
- drag-reorder in the graph
- section-level add affordances beyond the existing `add-error`

Also added, beyond the original description:

- the flow header bar (flow identity, description, labels, and a way into the flow's own properties), replacing the flow node that used to sit in the graph
- trigger insertion from the canvas
- changes to nodes shared with the execution topology (`BasicNode` tooltips, `DotNode` opacity)

## Out of Scope

Removing the No-code editor (kept on purpose to gather usage data), and the analysis of the collected data.

flcarre:
Closes https://github.com/kestra-io/kestra/issues/19338

## Summary

- Topology already rendered every flowable and had the add/edit/delete event plumbing in place, but `onCreateNewTask`/`onEditTask` only redirected those events into a separate No-code tab, and `EdgeNode` never rendered the "+" button its own `haveAdd` computation already supported. This routes them in place instead: the edge `+` opens the No-code task picker at the right position, editing opens the same `TaskEditModal` used by the No-code canvas, and deletes/inserts/edits all go through `useYamlUndo` so they're undoable and the graph refreshes.
- Nodes now show their validation state (`groupValidationIssuesByTask`, reused from the No-code canvas) via a small badge, threaded through a new `validationIssuesByTask` prop/injection on the `@kestra-io/topology` package.
- Both editor tabs stay available on purpose, so usage can be measured before either is retired: `task_added`/`task_edited`/`task_deleted` events now carry the originating surface (`no_code`/`topology`) plus `task_type` and, for an insertion, `position`, through the existing tracking pipeline (`utils/tabTracking.ts` → backend + PostHog).
- Extracted the No-code modal stack (`pushModalTarget`/`popModalTo`/`closeModal`/`resolveCreatedTarget`) into a shared `useTaskEditModalStack` composable and the delete-undo toast into a shared `UndoToast` component, so both surfaces reuse the exact same mechanics instead of forking them.
- `isReadOnly`/`isAllowedEdit` are already threaded down to every node (`isReadOnly` already folds in `!isAllowedEdit` in the flow store), so the picker and the modal are unreachable wherever editing isn't allowed.

**Decision on collapsed flowables:** a collapsed flowable already hides every interactive affordance of its contents (no edge, no burger menu, no edit action reaches inside it), so none of the new authoring entry points can target invisible content — there is nothing to auto-expand, by construction of the existing collapse mechanics.

**Nothing removed:** both editor tabs, `BlockSectionLane`, `BranchLane` and `useTopologyPanels` all stay, so the two surfaces coexist while usage is measured.

## Demo

Same flow (`If`/then/else + an `errors` task + a `Schedule` trigger), same edge, on `develop` and on this branch.

**Before** — the Topology panel draws the whole flow but offers no way to change it: hovering an edge gives nothing, and editing a task from a node opens a fourth panel that squeezes the canvas.

https://github.com/user-attachments/assets/e825681e-ac9c-477d-b206-0b1af83f49d2

**After** — reaching the canvas shows a `+` on every edge that can take a task, so the places to insert are visible instead of guessed; clicking one opens the No-code task picker in place, and the task lands exactly there rather than at the end of the list. Clicking a task card opens the same `TaskEditModal` the No-code editor uses, so editing never starts with hunting through a menu; delete stays undoable, and the `errors` lane has its own `+` that inserts into `errors` rather than into `tasks`.

https://github.com/user-attachments/assets/3db66e80-a870-4a13-a91b-459c12569e90

**DAG, keyboard and drag** — a `Dag` orders its tasks through `dependsOn` rather than list position, so the edge `+` had to mean something different there: dropping a task on `fetch_orders -> join_data` rewires the chain instead of appending a disconnected root. The same clip shows the shared keymap on the canvas (arrows walk and step into a flowable, `Enter` opens, `a` inserts) and dragging an existing task onto an edge, which heals the chain it leaves and splices into the one it joins. It is the browser's own drag, the same one the No-code block cards use, with the drag image drawn by the graph instead of the OS so it can be styled and stays the same across platforms. The bar above the canvas is the flow itself.

https://github.com/user-attachments/assets/26e45eda-45b6-474f-915f-3bf535a18768

**No-code parity** — everything the No-code editor could do that the canvas could not. The flow's own properties (labels, inputs, variables, concurrency) open from the bar above the canvas — the flow is not a step, so it is not a node in the graph. A trigger is not a task, so no edge `+` can reach one: the Triggers box carries its own, and the trigger it inserts opens its editor like a task's. `d` duplicates the focused task, `Alt` + arrows reorder it among its siblings, `/` opens the same command palette, and `?` lists every binding the canvas answers.

https://github.com/user-attachments/assets/2ca6d36f-bccd-4931-bd45-1f38f2db25bb

## Test plan

- [x] `npm run test:unit` passes (2289 tests, including the topology package's own 141)
- [x] `npm run test:lint` passes with no new warnings
- [x] `npm run check:ts-any` — net improvement, baseline updated
- [x] `node scripts/translations/check-translations.mjs --scope oss` passes; the one new key (`topology-graph.add-task`) is translated into all 12 locales
- [x] New Vitest coverage: `resolveTaskInsertionTarget` path resolution (first/last task, inside a flowable, inside an errors lane, unknown id) and `trackAuthoringAction` (payload shape, no-op when anonymous usage is disabled)
- [x] New Storybook story (`Validation Issues`) on `Topology.stories.ts` showing 0/1/several validation issues on different nodes


ui/packages/topology/src/Topology.vue
```diff
@@ -1,157 +1,196 @@
...
+    <div class="topology-shell">
+        <FlowSummaryChip
+            v-if="flowId"
+            class="topology-flow-bar"
+            :flowId="flowId"
+            :namespace="namespace"
+            :description="flowDescription"
+            :labels="flowLabels"
+            @editFlow="emit(EVENTS.EDIT_FLOW)"
+        />
```
MilosPaunovic: The chip is the one authoring affordance that is not behind a node, so it escapes the `isReadOnly` / `isAllowedEdit` gating the rest of the canvas relies on. `executions/Topology.vue` and `BlueprintDetailView.vue` both pass `flowId` with `isReadOnly`, so today this renders a Configure button on an execution's topology that opens `FlowPropertiesModal` and writes back through `applyGraphYaml`. Both props default to the safe side, so guarding here is enough.

```suggestion
        <FlowSummaryChip
            v-if="flowId && !isReadOnly && isAllowedEdit"
            class="topology-flow-bar"
            :flowId="flowId"
            :namespace="namespace"
            :description="flowDescription"
            :labels="flowLabels"
            @editFlow="emit(EVENTS.EDIT_FLOW)"
        />
```

ui/src/components/inputs/LowCodeEditor.vue
```diff
@@ -710,44 +808,112 @@
                     }
                     return
                 }
-                const updatedYmlSource = YAML_UTILS.deleteBlock({
-                    source: props.source ?? "",
-                    section,
-                    key: event.id,
+                const taskType = flowParsed.tasks.find((e: any) => e.id === event.id)?.type as string | undefined
```
MilosPaunovic: This runs for every delete, including a trigger delete, where the `flowParsed.tasks.length` guard above is skipped. A flow mid-edit with no `tasks:` key throws here instead of deleting.

```suggestion
                const taskType = flowParsed?.tasks?.find((e: any) => e.id === event.id)?.type as string | undefined
```

ui/packages/topology/src/utils/vueFlowUtils.ts
```diff
@@ -362,18 +369,36 @@ export function haveAdd(
...
+    const withDag = (target: AddTaskTarget): AddTaskTarget => {
+        if (!isDagLane) return target
+        const fromId = dagTaskIdAt(edge.source)
+        const toId = dagTaskIdAt(edge.target)
+        return fromId || toId ? {...target, dagDependency: {fromId, toId}} : target
+    }
+
     if (targetNode.type.endsWith("GraphClusterRoot")) {
-        return [clusterRootTaskId, "before"]
+        return {refId: clusterRootTaskId ?? "", position: "before"}
```
MilosPaunovic: `?? ""` turns an unresolved id into a target that is still truthy, so `EdgeNode` renders the `+` and `onCreateNewTask` then bails out of `resolveTaskInsertionTargetInAnySection` with no feedback. Returning `undefined` means no button rather than a dead one. The same shape applies to the two other `?? ""` returns just below.

```suggestion
    if (targetNode.type.endsWith("GraphClusterRoot")) {
        return clusterRootTaskId ? {refId: clusterRootTaskId, position: "before"} : undefined
    }
```

MilosPaunovic (changes requested):
Solid work, and the `Dag` handling in particular (`healDagRemoval` / `rewireDagDependency`, plus the `dagDependency` threading from `haveAdd` down to the picker) is the right model. Two blockers and one behaviour bug below, then smaller things.

Before the findings, a scope note: https://github.com/kestra-io/kestra/issues/19338 puts the command menu, a graph keyboard-navigation model, drag-reorder and section-level add affordances beyond `add-error` under **Out of Scope**, and this `PR` ships all four, plus the flow header bar and trigger insertion. It also changes shared nodes (`BasicNode` tooltips, `DotNode` opacity) that render on the execution topology. That is a lot to land and revert as one unit. Either splitting the authoring core from the keyboard/command/drag layer, or updating the issue to record the new scope, would make this reviewable and bisectable.

## 1. The flow bar is not gated by `isReadOnly` / `isAllowedEdit`, so it renders on the execution topology and on blueprint previews with a working Configure button

`ui/packages/topology/src/Topology.vue` renders `FlowSummaryChip` on `v-if="flowId"` alone, and `LowCodeEditor.vue`'s `@editFlow="flowPropertiesOpen = true"` carries no guard either. `LowCodeEditor` has three consumers, and two of them are read-only but still pass `flowId`:

- `ui/src/components/executions/Topology.vue` (`:flowId="execution.flowId"`, `isReadOnly`)
- `ui/src/components/flows/blueprints/BlueprintDetailView.vue` (`:flowId="parsedFlow.id"`, `isReadOnly`)

I mounted `Topology` with `isReadOnly: true, isAllowedEdit: false, flowId: "my_flow"` and both `topology-flow-chip` and `topology-flow-configure` render, and both emit `editFlow`. From there `FlowPropertiesModal` opens `FlowPropertiesEdit`, which writes through `UPDATE_YAML_FUNCTION_INJECTION_KEY` (provided by the new `useBlockEditorProvides` call as `applyGraphYaml`), so saving flow properties from an execution page would set `flowStore.flowYaml` and call `flowStore.onEdit`. The `PR` description says the picker and the modal are unreachable where editing is not allowed, and that is true of everything that goes through a node, but the chip is not a node.

Suggestion inline on `Topology.vue`. Both props already default to the safe side (`isReadOnly: true`, `isAllowedEdit: false`), so the guard is enough.

## 2. `Ctrl`/`Cmd` + `Z` typed in the `Flow Code` panel is swallowed by the topology surface

Three things line up:

- `useBlockEditorKeyboard.ts` lists `undo` in `ALWAYS_GLOBAL_IDS`, which skips the `isTypingTarget` guard.
- `useAuthoringSurface.isActive()` returns `true` for the last-mounted surface, and `engage` only ever *sets* `engaged`; clicking outside every surface never clears it, so focus sitting in `Monaco` does not make the canvas inactive.
- `dispatchTopologyShortcut`'s `case "undo"` returns `undefined`, so `handled !== false` and `event.preventDefault()` runs even when the undo history is empty.

I reproduced it with a scratch `vitest` that wires `useAuthoringSurface` + `useBlockEditorKeyboard` exactly as `LowCodeEditor` does and dispatches `ctrl+z` from a `textarea` inside a `.monaco-editor` outside the surface root: `dispatched: ['undo'] defaultPrevented: true`.

With `Flow Code` and `Topology` open side by side (a layout users pick from the panel bar), that means `Monaco`'s own undo is dead, and once the canvas has history, `Ctrl+Z` rewinds the whole flow `YAML` to an earlier snapshot instead of undoing the characters just typed. This is pre-existing for `Flow Code` + `No-code`, since `BlockEditor` already registered the same keymap globally, but the `useAuthoringSurface` gate added here was meant to settle exactly this class of conflict and does not cover it, and the `Topology` panel is newly affected.

The narrow fix is to make a globally-handled binding still stand down when the event target is a typing surface outside the authoring surface's own root, for example by passing the root element into `useBlockEditorKeyboard` and checking `isTypingTarget(event.target) && !root.contains(event.target)` before dispatching. Dropping `undo` from `ALWAYS_GLOBAL_IDS` also fixes it, at the cost of `Ctrl+Z` inside the `No-code` editor's own inputs.

## 3. The focus order misses `else` branches, every `Switch` case after the first, and a flowable's own `errors` lane, and that silently breaks the node menu's `Duplicate`

`topologyFocus.ts`'s `laneOf` returns on the *first* non-empty lane it finds, so `walk` only ever recurses into one branch per flowable. Measured on the branch:

| flow | `buildTopologyFocusOrder` |
|---|---|
| `If` with `then` + `else` | `['gate', 'in_then']` |
| `Switch` with cases `a` + `b` | `['router', 'in_case_a']` |
| `Sequential` with `tasks` + `errors` | `['seq', 'child']` |

This is not only the keyboard model. `onDuplicate`, which is the node menu's `Duplicate` action, goes through `duplicateTaskById` -> `focusOrder.value.find(entry => entry.id === id)` -> `if (!node) return false`. So clicking `Duplicate` on a task in an `else`, in a second `Switch` case, or in a flowable's `errors` lane does nothing at all, with no error and no toast. `reorderFocusedTask`, `insertRelativeToFocused` and the command menu's `goToSection` have the same blind spot.

The new `topologyFocus.spec.ts` fixture uses an `If` with `then` only and a `Dag`, so nothing in the suite sees it. The fix is for `laneOf` to collect every lane rather than return the first, with `TopologyFocusNode.childrenPath` becoming a list and `firstChildOf` / `parentOf` reading it as one:

```ts
function lanesOf(item: Record<string, unknown>, itemPath: string): {key: string; path: string}[] {
    const task = displayTaskOf(item)
    const taskPath = task === item ? itemPath : `${itemPath}.task`
    const lanes: {key: string; path: string}[] = []
    for (const key of NESTED_BLOCK_KEYS) {
        const branch = task[key]
        if (Array.isArray(branch) && branch.length > 0) lanes.push({key, path: `${taskPath}.${key}`})
    }
    const cases = task.cases
    if (cases && typeof cases === "object" && !Array.isArray(cases)) {
        for (const caseKey of Object.keys(cases as Record<string, unknown>)) {
            lanes.push({key: "cases", path: `${taskPath}.cases.${caseKey}`})
        }
    }
    return lanes
}
```

Worth a test with an `If` carrying both branches and a `Switch` carrying two cases, plus one asserting `Duplicate` works on a task in an `else`.

## 4. Smaller things

- `useYamlUndo` is instantiated once per surface, so an undo history built in `No-code` is invisible from `Topology` and vice versa, even though both write the same `flowStore.flowYaml`. Given that the whole point is that the two surfaces coexist, deleting a task in one tab and pressing `Ctrl+Z` in the other doing nothing is worth a deliberate decision rather than a side effect of the split.
- `ui/packages/topology/tests/units/utils/vueFlowUtils.test.ts` "builds no node or edge for the flow itself" asserts that no element id or source contains `__flow__` and that no element type is `flow`. `__flow__` appears nowhere in production code on this branch, and none on `develop` either, so the test can never fail. Either assert on something the generator actually emits or drop it.
- `Topology.vue` adds the `dragover` listener in `onTaskDragStart` and removes it in `onTaskDragEnd` only, so a component unmounted mid-drag leaks it; `onUnmounted` removes `dragend`/`blur` but not this one.
- `eventNaming.ts` adds `task_duplicated` but `eventNaming.spec.ts` has fixtures for the other four only.
- `EdgeNode.vue`'s `+` is a native `<button>`, which already fires `click` for `Enter` and `Space`; the extra `@keydown.enter.stop.prevent` / `@keydown.space.stop.prevent` handlers only work out because `.prevent` suppresses the click the browser would synthesise. Harmless, but they can go.

ui/packages/topology/src/Topology.vue
flcarre: Applied, with the guard exactly as suggested — `v-if="flowId && !isReadOnly && isAllowedEdit"`.

I added `packages/topology/tests/units/Topology.test.ts` covering your repro: mounting `Topology` read-only with a `flowId` now renders neither `topology-flow-chip` nor `topology-flow-configure`, across `(true, false)`, `(true, true)` and `(false, false)`, with the editable case asserted the other way so the chip is not silently lost. Reverting the guard fails three of the four.

Fixed in 25e69474b4 (chip) — the same commit also releases the `dragover` listener on unmount, which you flagged further down.

ui/src/components/inputs/LowCodeEditor.vue
flcarre: Applied. I also gave the guard just above it the same treatment — `flowParsed?.tasks?.length === 1` — since it throws on the same flow shape whenever the section really is `tasks`; `undefined === 1` is false, which is the right answer when there are no tasks to protect.

Fixed in 5517923490.

ui/packages/topology/src/utils/vueFlowUtils.ts
flcarre: Applied to all four returns, not just this one — the `GraphClusterEnd` and `sourceIsEndOfCluster` branches and the trailing `before` return had the same shape:

```ts
if (targetNode.type.endsWith("GraphClusterRoot")) {
    return clusterRootTaskId ? {refId: clusterRootTaskId, position: "before"} : undefined
}
const sourceIsEndOfCluster = nodeByUid[edge.source].type.endsWith("GraphClusterEnd")
if (!sourceIsEndOfCluster && targetNode.type.endsWith("GraphClusterEnd")) {
    const refId = Utils.afterLastDot(edge.source)
    return refId ? withDag({refId, position: "after"}) : undefined
}
```

No button beats a dead one.

Fixed in 25e69474b4.

flcarre:
Thanks — this was a genuinely useful review: both blockers and the behaviour bug were real, and the `laneOf` one was reachable through the node menu, not just the keyboard, which I had not seen. All of it is fixed, each with a test that fails without its fix. Six commits, `eb875afe61..25e69474b4`.

## Scope

Fair, and the issue was the thing that was wrong, not the branch. I have recorded the change on #19338 rather than splitting: the four items you named have moved out of **Out of Scope** into an explicit **Scope change** section, together with the flow header bar, trigger insertion and the shared-node changes, with the reason they were pulled in — a canvas you can insert into but cannot traverse, reorder or address by keyboard is half a surface, and the picker needed a way in that was not a mouse. The **Context** bullet that still claimed graph keyboard navigation was deferred is updated too.

I went for recording over splitting because the keyboard/command/drag layer is what makes the authoring core usable rather than a separate feature on top of it, and because a split now would drop the review context you have already built here. If you would still rather see it as two PRs, say so and I will do the split.

## 1. Flow bar not gated — fixed, `eb875afe61`

Guard applied as suggested. New `packages/topology/tests/units/Topology.test.ts` mounts `Topology` the way you did and asserts neither `topology-flow-chip` nor `topology-flow-configure` renders across `(isReadOnly, isAllowedEdit)` of `(true, false)`, `(true, true)` and `(false, false)`, with the editable case asserted the other way so the chip cannot be silently lost. Reverting the guard fails three of the four.

I checked the feature path still holds: `LowCodeEditorWrapper` reads both from the store, and `isReadOnly` is computed as `!isAllowedEdit || ...`, so an editable flow is `(false, true)` and the chip renders as before.

## 2. `Ctrl`/`Cmd` + `Z` swallowed — fixed, `f2c2ddd471` and `46ea3cb7e4`

Took your narrow fix. `useBlockEditorKeyboard` now accepts the surface root, and a globally-handled binding stands down when the event target is a typing surface outside it:

```ts
const typingOutsideRoot =
    typing && options.root !== undefined && !options.root.value?.contains(event.target as Node)

if (event.key !== "Escape" && typingOutsideRoot) return
```

`LowCodeEditor` passes `vueFlow`, `BlockEditor` passes `editorEl`. That keeps `Ctrl+Z` working inside the No-code editor's own inputs, which dropping `undo` from `ALWAYS_GLOBAL_IDS` would have cost.

Your third point is fixed separately: `performUndo` returns a boolean and both `dispatchTopologyShortcut` and `dispatchBlockEditorAction` return it, so an empty history no longer reaches `event.preventDefault()`.

Three tests in `useBlockEditorKeyboard.spec.ts` — a `ctrl+z` from a `textarea` inside a `.monaco-editor` outside the root dispatches nothing and leaves `defaultPrevented` false; the same chord inside the surface's own input still dispatches `undo`; and an empty history does not preventDefault. The first fails without the fix.

## 3. Focus order misses branches — fixed, `4626ae41bf`

`laneOf` is now `lanesOf`, collecting every lane, and `childrenPath` became `childrenPaths: string[]`, with `firstChildOf` / `parentOf` reading it as a list. I kept your shape but skip a case whose branch is empty, to match how the `NESTED_BLOCK_KEYS` loop already treats an empty lane.

Measured on the new fixture:

| flow | `buildTopologyFocusOrder` |
|---|---|
| `If` with `then` + `else` | `['gate', 'in_then', 'in_else']` |
| `Switch` with cases `a` + `b` | `['router', 'in_case_a', 'in_case_b']` |
| `Sequential` with `tasks` + `errors` | `['seq', 'child', 'recover']` |

Four new tests: the order above, each branch as its own sibling lane so `moveWithinSiblings` cannot cross between them, `parentOf` from inside an `else` / a second case / an `errors` lane, and the one you asked for — `Duplicate` resolving on a task in each. All four fail on the old `laneOf`.

## 4. Smaller things

- **`useYamlUndo` per surface** — decided rather than left as a side effect: the history is now shared, since both surfaces write the same `flowStore.flowYaml` and are two views of one document. It is scoped to the flow and dropped when that changes, so a shared history cannot rewind into another flow's YAML — which a plain module-level history would have made possible and the per-instance one did not. Three tests in the new `useYamlUndo.spec.ts`, all three failing before. `46ea3cb7e4`.
- **Vacuous `__flow__` test** — replaced. `FLOW_NODE_ID` was deleted on this branch, so the assertion named a constant that no longer exists. It now asserts every emitted element traces back to a uid the backend graph declares, which catches any synthetic node appended again whatever it ends up being called. `25e69474b4`.
- **`dragover` leak** — `onUnmounted` now removes it alongside `dragend` and `blur`. `eb875afe61`.
- **`task_duplicated` fixture** — added to `eventNaming.spec.ts`. `25e69474b4`.
- **`EdgeNode`'s redundant keydown handlers** — dropped; the native `<button>` already covers `Enter` and `Space`. `25e69474b4`.

## Verification

`vue-tsc` clean on both `tsconfig.app.json` and `tsconfig.test.json`, `oxlint` and `eslint` clean. Unit suite 2659 passing. One file fails locally, `tests/unit/tokens/ksTokens.spec.ts`, and it is environmental — it shells out to `node_modules/stylelint/bin/stylelint.mjs`, which is absent from my checkout, so the subprocess exits `MODULE_NOT_FOUND` and the JSON parse fails. Nothing in this diff touches SCSS or tokens; CI is the check that counts for it.

ui/src/components/no-code/blocks/useYamlUndo.ts
```diff
@@ -6,12 +6,15 @@ const EDIT_DEBOUNCE = 1000
...
 }
 
+const undoHistory = ref<string[]>([])
+const historyScope = ref<string | undefined>(undefined)
+
 export function useYamlUndo(flowStore: FlowStoreLike, deletedLabel: (name: string) => string) {
     const onEditTimeout = ref<ReturnType<typeof setTimeout>>()
-    const undoHistory = ref<string[]>([])
     const undoState = ref<{label: string} | null>(null)
     let undoTimer: ReturnType<typeof setTimeout> | undefined
```
MilosPaunovic: These two are still per-instance while `undoHistory` above them is module-level, and `applyYaml` dismisses only its own badge - so a delete's toast can outlive the edit it describes and undo a different one (details in the review body).

Hoisting them next to `undoHistory` and clearing them from `enterCurrentScope` keeps the badge and the history telling the same story:

```ts
const undoHistory = ref<string[]>([])
const historyScope = ref<string | undefined>(undefined)
const undoState = ref<{label: string} | null>(null)
let undoTimer: ReturnType<typeof setTimeout> | undefined
```

with `enterCurrentScope()` calling `dismissDeleteBadge()` alongside `undoHistory.value = []`.

ui/src/components/no-code/blocks/topologyFocus.ts
```diff
@@ -0,0 +1,114 @@
...
+        const lanes = lanesOf(item, path)
+        out.push({id: String(id), path, parentPath, depth, childrenPaths: lanes.map(lane => lane.path)})
+        for (const lane of lanes) {
+            walk(getIn(item, lane.path, path), lane.path, depth + 1, out)
+        }
+    })
+}
+
+function getIn(item: Record<string, unknown>, lanePath: string, itemPath: string): unknown {
+    const relative = lanePath.slice(itemPath.length).replace(/^\./, "")
```
MilosPaunovic: Splitting the relative path on `.` is what makes a `Switch` case key containing a dot unreachable: `cases: {"1.0": [...]}` resolves to `item.cases["1"]["0"]`, which is `undefined`, so `walk` never descends and the recorded `path` would not resolve either.

`lanesOf` already has the lane's value in hand, so handing it back with the path removes the need to re-parse the string at all:

```ts
function lanesOf(item, itemPath): {key: string; path: string; items: unknown}[]
```

and `walk` then calls `walk(lane.items, lane.path, depth + 1, out)`, dropping `getIn` entirely.

ui/src/components/no-code/blocks/useBlockEditorKeyboard.ts
```diff
@@ -1,4 +1,4 @@
-import {onActivated, onDeactivated, onMounted, onUnmounted} from "vue"
+import {onActivated, onDeactivated, onMounted, onUnmounted, type Ref} from "vue"
 
 const ALWAYS_GLOBAL_IDS = new Set(["save", "undo", "command-menu", "clear"])
```
MilosPaunovic: `undo` being global is what lets `Ctrl`/`Cmd` + `Z` reach the surface from inside its own text fields, since only `typingOutsideRoot` stands in the way and that check passes for anything inside the root. Dropping it from the set lets the plain `typing` guard apply, which is what the commit title describes:

```suggestion
const ALWAYS_GLOBAL_IDS = new Set(["save", "command-menu", "clear"])
```

`typingOutsideRoot` then only still matters for `save`, `command-menu` and `clear`, which is the right set for it.

ui/src/utils/flowableBlockOps.ts
```diff
@@ -441,15 +547,29 @@ export function groupValidationIssuesByTask(
...
+        const field = rawField?.replace(/^task\./, "")
+        const entry = field ? `${field}: ${message.trim()}` : message.trim()
+
+        // A task constraint violation comes back id-keyed (`tasks[publish].message`), so the last
+        // bracket already names the task; only a numeric path has to be resolved against the flow.
+        const lastBracket = /\[["']?([^"'\]]+)["']?\]$/.exec(taskPath)?.[1]
+        if (lastBracket && !/^\d+$/.test(lastBracket)) {
+            add(lastBracket, entry)
+            continue
+        }
```
MilosPaunovic: `taskPath` was matched greedily up to the last `]` on the line, so a message carrying its own bracket gets mined for the id. `tasks[0].commands: invalid value [x]: not allowed` files the issue under a task called `x`, and the badge lands on nothing. Not a regression - the numeric-path branch below dropped that line too - but the shortcut can check that the id it found is real:

```suggestion
        // A task constraint violation comes back id-keyed (`tasks[publish].message`), so the last
        // bracket already names the task; only a numeric path has to be resolved against the flow.
        const lastBracket = /\[["']?([^"'\]]+)["']?\]$/.exec(taskPath)?.[1]
        if (lastBracket && !/^\d+$/.test(lastBracket) && ids.has(lastBracket)) {
            add(lastBracket, entry)
            continue
        }
```

where `ids` is the set of task ids already reachable from `flow`. If building that set is more than this is worth, anchoring the path match non-greedily would also do it.

MilosPaunovic (changes requested):
Round 2, re-reviewed at `25e69474b4`. Four things still open, the first two of them user-visible.

## 1. `useYamlUndo` now shares the history but not the toast, so the toast can undo the wrong thing

`undoHistory` and `historyScope` moved to module scope, but `undoState` and `undoTimer` are still per-instance, and `applyYaml` calls `dismissDeleteBadge()` on **its own** instance only. With both panels open:

1. Delete a task in `No-code` - its "log_task deleted" toast appears (6s).
2. Insert a task from the `Topology` `+` inside that window. `applyYaml` pushes to the shared history and dismisses only `Topology`'s (absent) badge.
3. Click Undo on the toast that still reads "log_task deleted".

Verified against the branch: `flowYaml` comes back as `tasks: [a]`, the state before the *insert*, not `tasks: []`, the state before the delete. The toast names one change and performs another.

The shared history is the right call; the badge has to follow it. Hoisting `undoState`/`undoTimer` alongside `undoHistory`, and clearing them in `enterCurrentScope` the same way, is the smallest fix - and it also puts the toast on whichever surface the user is looking at rather than only the one that ran the delete. The new `useYamlUndo.spec.ts` covers the shared history but not the badge.

## 2. `Switch` cases whose key contains a dot are still unreachable

`lanesOf` now enumerates every case, but `getIn` resolves the lane by splitting the relative path on `.`, so a case key that contains one - a version string like `cases: {"1.0": [...], "2.0": [...]}`, which is what a `Switch` on a version value looks like - resolves to `item.cases["1"]["0"]`, which is `undefined`.

Measured on the branch: `buildTopologyFocusOrder` on such a flow returns `['router']`, with both case bodies missing. Same user-visible consequence as round 1 - arrow navigation cannot reach those tasks, and `Duplicate` from the node menu silently no-ops on them, because `duplicateTaskById` looks the id up in `focusOrder`. The recorded `path` would be unusable by `duplicateBlockAtPath` even if the walk did reach them.

`lanesOf` already holds the lane's value, so returning it alongside the path (and dropping `getIn` entirely) closes this without any string re-parsing.

## 3. `Ctrl`/`Cmd` + `Z` inside the surface's own text fields still rewinds the whole `YAML`

The `root` guard is right and settles the `Flow Code` case. It only covers fields *outside* the root, though: an `<input>` inside the authoring surface passes `typingOutsideRoot`, and then `undo` being in `ALWAYS_GLOBAL_IDS` skips the plain `typing` guard. Verified on the branch by dispatching `ctrl+z` from an input inside the root: `dispatched: ['undo']`, `defaultPrevented: true`. So typing a task id or a message in a `No-code` field and pressing `Ctrl+Z` rolls the flow back a whole edit instead of undoing the characters just typed.

Pre-existing rather than introduced here, so your call whether it belongs in this `PR` - raising it because the commit is titled "leave `Ctrl+Z` to whoever owns the field being typed in" and it currently does that for half the fields. Suggestion inline.

## 4. Nit

`groupValidationIssuesByTask`'s `lastBracket` shortcut trusts the last bracketed token in a path that was matched greedily up to the last `]` on the line. Suggestion inline.

flcarre:
Follow-up on two things after the push above.

**I broke something with my own fix for #2, and it is now fixed properly — `9314c5c578`.**

The guard I wrote stood down on *any* typing target outside the surface root. `TaskEditModal` renders through `KsDialog appendToBody`, so element-plus teleports it to `document.body` and it is never a DOM descendant of the canvas. Modal editing is the default mode, and on Topology it is the only edit path — so `Cmd+S`, `Ctrl+Z` and the command menu went dead the moment a field in the task editor had focus. `BlockEditor` swapping its canvas out for the inline `TaskEdit` form left `editorEl` unset, with the same effect.

The guard now names the thing it is actually protecting — a rival code editor, which is what carries its own undo and save stack:

```ts
const rivalCodeEditor =
    options.root !== undefined &&
    isCodeEditorTarget(event.target) &&
    !options.root.value?.contains(event.target as Node)
```

A plain field in a teleported overlay belongs to the surface and keeps its shortcuts; Monaco keeps its own, whether it is the Flow Code panel or a script field inside the modal — which is the behaviour you wanted in the first place. Two tests cover it (teleported-modal input, and an unmounted root), both failing before the fix, and your original Monaco case still passes.

I also took the DRY nit in passing: `lanesOf` now uses `taskEditPathFor` instead of repeating the wrapped-item path rule.

**CI: `Frontend - Tests` is red, and it is not this branch.**

The unit suite passes in full — 291 files, 2665 tests. The one failure is `tests/storybook/components/executions/Logs.stories.tsx > Compact Fullscreen Virtualized Tasks`, timing out at 15s. That file is not in this PR's diff at all, and the same story fails identically on #19784, which touches nothing related. Looks repo-wide rather than mine — worth someone picking up separately, since it currently blocks **Required checks** on every PR.

ui/src/components/no-code/blocks/useBlockEditorKeyboard.ts
```diff
@@ -54,7 +60,12 @@ export function useBlockEditorKeyboard(options: UseBlockEditorKeyboardOptions) {
         const typing = isTypingTarget(event.target)
         const isGlobal = ALWAYS_GLOBAL_IDS.has(binding.id)
         const ignoresOverlayGuard = IGNORES_OVERLAY_GUARD_IDS.has(binding.id)
+        const rivalCodeEditor =
+            options.root !== undefined &&
+            isCodeEditorTarget(event.target) &&
+            !options.root.value?.contains(event.target as Node)
 
+        if (event.key !== "Escape" && rivalCodeEditor) return
```
MilosPaunovic: Restricting this to `.monaco-editor` means every other typing surface outside the root falls through to the global-binding path. Verified at this head with a plain `<input>` outside the root and no `Monaco` in the page: `dispatched: ['undo']`, `defaultPrevented: true` - so `Ctrl`/`Cmd` + `Z` in a sibling panel's search box rewinds the flow `YAML` instead of the text being typed.

The reason DOM containment had to be loosened is that the surface's own dialogs are `appendToBody`, so they sit outside `root`. Claiming them explicitly keeps the original rule and fixes the modal case too, rather than trading one rival for all of them:

```ts
// Any typing surface outside this one belongs to whoever owns it: another editor pane, or an
// unrelated panel's search box. The surface's own dialogs are appended to the body, so they are
// claimed by a marker rather than by DOM containment.
const foreignTypingTarget =
    typing &&
    options.root !== undefined &&
    !options.root.value?.contains(event.target as Node) &&
    !(event.target as HTMLElement | null)?.closest?.("[data-authoring-overlay]")

if (event.key !== "Escape" && foreignTypingTarget) return
```

Deliberately not a one-click `suggestion`: it only works once `TaskEditModal`, `BlockShortcutsDialog`, `BlockCommandMenu` and `FlowPropertiesModal` carry that attribute, so applying it alone would put the modal back where it was. If marking four dialogs is more than this is worth, the other end of the same rule is dropping `undo` from `ALWAYS_GLOBAL_IDS` and letting the plain `typing` guard cover every field, inside the surface and out.

MilosPaunovic (changes requested):
Round 3, re-reviewed at `9314c5c578`. One new item from the narrowed keyboard guard, and three carried over.

## New: narrowing the guard to `Monaco` leaves every other outside input handing `Ctrl`/`Cmd` + `Z` to the canvas

`rivalCodeEditor` only stands down for a target inside a `.monaco-editor`. Any other typing surface outside the authoring root - a sibling panel's search box, the `Blueprints` or `Files` panel's filter, the top-nav search - is not a code editor, so the guard passes, and `undo` / `save` / `command-menu` being in `ALWAYS_GLOBAL_IDS` then skips the plain `typing` guard too.

Verified at this head: a plain `<input>` outside the surface root, no `Monaco` anywhere, gives `dispatched: ['undo']` and `defaultPrevented: true`. So typing in that field and pressing `Ctrl+Z` rewinds the whole flow `YAML` and the field's own undo never runs.

I see why the previous shape had to go: `TaskEditModal` and the shortcuts dialog are `appendToBody`, so DOM containment calls the surface's own modal foreign. That argues for marking those overlays rather than for narrowing the test to one rival. Sketch inline.

## Still open from round 2

- **The undo toast and the shared history still disagree.** Re-checked at this head: delete in `No-code`, insert from the `Topology` `+` inside the 6s window, then Undo from the toast - `flowYaml` goes to `tasks: [a]` (the insert) while the toast still reads `log_task deleted`. Thread: `useYamlUndo.ts`.
- **A `Switch` case key containing a dot is still unreachable.** `lanesOf` now routes through `taskEditPathFor`, but `getIn` still splits the relative path on `.`, so `cases: {"1.0": [...]}` gives a focus order of `['router']` at this head and `Duplicate` on those tasks is still a silent no-op. Thread: `topologyFocus.ts`.
- **The `lastBracket` validation-path nit.** Thread: `flowableBlockOps.ts`.

Separately, I exercised `moveTaskOntoEdge` at this head across the same-lane reorder cases (`a` onto `b -> c`, `c` onto `a -> b`, `a` after `c`, a no-op onto its own neighbour) and the `Dag` splice, and it is correct in all of them, orphan-free. No action needed there; noting it so it does not get re-litigated.

flcarre:
More on the red `Frontend - Tests` I mentioned above — I traced it, and it is neither this branch nor a flake.

**`develop` itself is red on it**, run [35963705000](https://github.com/kestra-io/kestra/actions/runs/35963705000), same story, same 15s timeout.

The bisect lands on a single commit:

| | head | storybook |
|---|---|---|
| last green run | `4dcc296aa3` — 23/09 12:59 | pass |
| first red run | `c56fe4a10d` — 24/09 06:16 | fail |

`git log 4dcc296aa3..c56fe4a10d` holds **exactly one commit**: `c56fe4a10d`, *feat: Add full screen button for task logs* (#18062). That commit is also the one that **added the failing story** — it introduces `compactFullscreenPlay()`, `CompactFullscreenVirtualizedTasks`, and the `waitFor` that throws `"recycled first task log scroller not ready"`, alongside a rewrite of `executions/Logs.vue` (+333 lines). The story and the code it exercises landed together, and that wait never resolves in CI.

No open issue for it, and no open PR touching `Logs.stories.tsx` or `executions/Logs.vue`, so nobody has picked it up yet. It blocks *Required checks* on every open PR, not just this one, so I am taking it to its own ticket rather than here.

Nothing changes for this PR: the unit suite passes in full in CI (291 files, 2665 tests), and the other 70 storybook files pass.

ui/src/components/no-code/blocks/useYamlUndo.ts
flcarre: Fixed as suggested — `2f161e8f35`. `undoState` and `undoTimer` now sit next to `undoHistory`, and `enterCurrentScope` calls `dismissDeleteBadge()` alongside clearing the history.

Two tests cover the pair, both failing before: your exact sequence (delete in one surface, edit from the other inside the window, toast cleared rather than left to undo the wrong change), and the side effect you pointed out — the toast now shows on whichever surface the user is looking at, not only the one that ran the delete.

ui/src/components/no-code/blocks/topologyFocus.ts
flcarre: Took the shape you sketched — `7308bf1397`. `lanesOf` returns `{key, path, items}`, `walk` recurses on `lane.items`, and `getIn` is deleted.

One thing beyond the sketch: dropping `getIn` fixes the descent, but the recorded `path` was still unusable downstream, since `tasks[0].cases.1.0[0]` is ambiguous however it is later parsed. The lane path now goes through `appendKeyToPath`, which already quotes a key matching `keyNeedsQuoting`, giving `tasks[0].cases["1.0"][0]` — which `parsePath` reads back correctly, so `duplicateBlockAtPath` resolves it.

Tests assert both halves on a `Switch` with `"1.0"` / `"2.0"` cases: the focus order is `["router", "legacy_path", "current_path"]`, and `duplicateBlockAtPath` resolves each recorded path. Both fail without the fix — the second one still failed after the `getIn` removal alone, which is what surfaced the quoting half.

ui/src/components/no-code/blocks/useBlockEditorKeyboard.ts
flcarre: Took the second end of your rule rather than the marker — `d4586a9b32`.

`ALWAYS_GLOBAL_IDS` is now `["save", "command-menu", "clear"]`. That covers both cases in one move: a field inside the surface (round 2 #3) and any field outside it, Monaco or not (this one), because the plain `typing` guard applies to all of them.

With `undo` out of the set, I then removed the `root` option rather than marking the five overlays, and I want to be explicit about why, since it walks back something from round 2 you had called right:

- The only bindings left global are `save`, `command-menu` and `clear`. `clear` is `Escape`, already exempt everywhere.
- For `save` and `command-menu`, standing down in a foreign field is worse than dispatching: `Cmd+S` in a sibling panel's search box would fall through to the browser's save-page dialog. Those two are app-level, and firing them from anywhere is both the pre-existing behaviour and the one I think users want.
- Which of the two surfaces receives a shortcut is already arbitrated by `useAuthoringSurface.isActive()`, not by DOM containment — so `root` was not doing that job either.

That leaves the marker with nothing to protect, and five components spared an attribute whose only purpose would be to undo a guard.

Four tests, the first two failing before: `Ctrl+Z` in a field inside the surface is not dispatched and not `preventDefault`ed; same for a `Monaco` textarea; `Ctrl+Z` from the canvas itself still dispatches `undo`; and `Cmd+S` from a field in a teleported overlay still dispatches `save`.

If you would rather keep `root` for `save`/`command-menu` with the `data-authoring-overlay` marker, say so — it is additive on top of this and I will put it back.

ui/src/utils/flowableBlockOps.ts
flcarre: Fixed — `6005700135` — but with your fallback rather than the id set, because the set breaks existing coverage.

I tried the sketch first. Two tests in the suite pass a `flow` alongside an id that is deliberately not in it (`resolves the id-keyed path the validate endpoint returns` with `tasks[publish]`, and the Dag one with `tasks[pipeline].tasks[publish_report]`, both against the `probe` fixture whose ids are `if_task` / `nested_then` / `on_error` / `on_finally` / `on_after`). Gating on `ids.has(...)` drops both. They are testing the id-keyed shortcut in isolation, and rewriting their expectations to make my change pass seemed like the wrong direction.

So: anchoring the path match non-greedily, which is the actual cause — one character.

```diff
-const pathMatch = /^(.+\])(?:\.([A-Za-z0-9_.]+))?\s*:\s*(.+)$/.exec(cleaned)
+const pathMatch = /^(.+?\])(?:\.([A-Za-z0-9_.]+))?\s*:\s*(.+)$/.exec(cleaned)
```

The path now ends at the first `]` that still lets the rest of the line parse. Your example files under `if_task` with `commands: invalid value [x]: not allowed` as the message, and nothing is filed under `x`. Backtracking still reaches the longer path where it is the real one, so `tasks[0].then[0].message` and the Dag wrapper case are unaffected — the whole file passes, 123 tests.

flcarre:
Round 3 addressed — `2f161e8f35..6005700135`, four commits, each with tests that fail without their fix. Replies are on the four threads; the summary and the one place I went a different way:

| | fix | commit |
|---|---|---|
| toast vs shared history | `undoState`/`undoTimer` hoisted next to `undoHistory`, cleared in `enterCurrentScope` | `2f161e8f35` |
| dotted `Switch` case key | `lanesOf` returns the lane's items, `getIn` deleted; lane path via `appendKeyToPath` | `7308bf1397` |
| `Ctrl`/`Cmd` + `Z` | `undo` out of `ALWAYS_GLOBAL_IDS`; `root` guard removed | `d4586a9b32` |
| `lastBracket` | path match anchored non-greedily | `6005700135` |

**Where I diverged, and why.**

On the keyboard I took the second end of your rule — dropping `undo` from `ALWAYS_GLOBAL_IDS` — and then removed the `root` option rather than marking the overlays. That walks back something you called right in round 2, so the reasoning: once `undo` is not global, the only bindings left are `save`, `command-menu` and `clear`. `clear` is `Escape`, already exempt. For the other two, standing down in a foreign field is worse than dispatching — `Cmd+S` in a sibling panel's search box would fall through to the browser's save-page dialog. And arbitration between the two surfaces is already `useAuthoringSurface.isActive()`, never DOM containment. So `root` had nothing left to protect, and five components are spared an attribute whose only job would be to cancel a guard. It is additive if you disagree — say so and I will put the marker back.

On `lastBracket` I used your fallback rather than the id set: two existing tests deliberately pass a `flow` alongside an id that is not in it, and gating on `ids.has(...)` drops both. Anchoring non-greedily fixes the cause in one character without touching their expectations.

Also noted your `moveTaskOntoEdge` check — thanks for writing it down, that is exactly the kind of thing that otherwise gets re-litigated.

**CI.** `Frontend - Tests` is still red on `Logs.stories.tsx > Compact Fullscreen Virtualized Tasks`, which I traced in the comment above to #18062 — `develop` itself is red on it and the bisect is a single commit. Unrelated to this branch; the unit suite passes in full.

ui/src/components/no-code/blocks/useBlockEditorKeyboard.ts
```diff
@@ -1,6 +1,6 @@
 import {onActivated, onDeactivated, onMounted, onUnmounted} from "vue"
 
-const ALWAYS_GLOBAL_IDS = new Set(["save", "undo", "command-menu", "clear"])
+const ALWAYS_GLOBAL_IDS = new Set(["save", "command-menu", "clear"])
```
MilosPaunovic: `save` being global with no root check is what lets a `Ctrl+S` typed in the `Flow Code` editor reach this surface. `FlowFileEditorTab.vue` already has its own window listener for that chord (`handleGlobalSave`, registered at `window.addEventListener("keydown", handleGlobalSave)`), and it does not `stopPropagation`, so with both panels open the one keystroke runs `useFlowEditorActions().save` and `LowCodeEditor.saveFlow()` concurrently against the same flow.

Whichever way it goes, only one owner should answer the chord. If the surface is meant to own it, the code editor's listener should stand down while an authoring surface is mounted; if the editor owns it, the simplest version is the same treatment `undo` just got:

```suggestion
const ALWAYS_GLOBAL_IDS = new Set(["command-menu", "clear"])
```

That leaves `Ctrl+S` on the canvas working (the body is not a typing target) while a `Ctrl+S` typed in any field goes to whoever owns that field, which is the rule the previous commit settled on.

MilosPaunovic (commented):
Round 4, re-reviewed at `6005700135`. Nothing from the previous rounds is left open. One item found this round.

## `Ctrl`/`Cmd` + `S` now runs two save paths at once

`save` stays in `ALWAYS_GLOBAL_IDS` and the root check is gone, so a `Ctrl+S` typed anywhere - including inside the `Flow Code` editor - reaches this surface and runs `saveFlow()` -> `flowStore.save()`.

`FlowFileEditorTab.vue` registers its own window `keydown` for the same chord (`handleGlobalSave` -> `useFlowEditorActions().save`), and it calls `preventDefault()` without `stopPropagation()`. Two window listeners, neither yielding to the other: with the `Flow Code` and `Topology` panels open side by side, one `Ctrl+S` starts both saves concurrently against the same flow, each running its own `onEdit` and revision check.

This is pre-existing in shape - `BlockEditor` already had `save` global, so `Flow Code` + `No-code` has had it - but the topology wiring makes it reachable in a layout where it was not before, and this `PR` is what puts the keymap on that surface. Suggestion inline.

Two notes, neither a defect:

- Now that the toast is shared, it renders in both panels at once when both are open. That follows directly from what I asked for, so flagging it rather than objecting - worth a glance to confirm two toasts read as intentional.
- The red `Frontend - Tests` you may have seen at `9314c5c578` was not this branch. `Logs.stories.tsx > Compact Fullscreen Virtualized Tasks` does not exist here (absent at the base `4dcc296aa3` and at the branch tip, present on `develop`); it arrived with `c56fe4a10d` "feat: Add full screen button for task logs (#18062)" at 06:16 `UTC`, and the `pull_request` build at 06:47 `UTC` tested the merge into that `develop`, where it timed out at 15s. 19767 and 19768 are green only because their runs predate the story.

ui/src/components/no-code/blocks/useBlockEditorKeyboard.ts
flcarre: Verified, fixed, and it turned out to need more than the one line — `c5bb6268a3`.

Confirmed the conflict first: `handleGlobalSave` is a window listener that calls `preventDefault()` with no `stopPropagation()`, and the `Flow Code` panel is `FlowFileEditorTab` with `flow: true` (`override/components/flows/panelDefinition.ts`), so `props.flow` is truthy and it runs `saveFlowYaml()` on the chord.

Where I ended up somewhere else: dropping `save` from `ALWAYS_GLOBAL_IDS` only settles the half where the chord is typed in a field. `handleGlobalSave` tests neither the target nor the phase — it fires on **any** `Ctrl+S` while mounted — so a chord typed on the canvas, which is not a typing target, still reaches both listeners. The one-liner leaves that case double-saving.

So I took your first option instead, which also means restoring the `root` guard I removed in `d4586a9b32` — you were right in round 3 and I was wrong to drop it. It is back with the marker, since containment is what could not see the surface's own `appendToBody` dialogs:

| chord typed in | `Flow Code` | authoring surface |
|---|---|---|
| the `Flow Code` editor | saves | stands down (foreign typing target) |
| the surface's own overlay | stands down (surface active, outside its root) | saves (claimed by marker) |
| the canvas | stands down | saves (not a typing target) |
| anywhere, `Flow Code` alone | saves | not mounted |

`data-authoring-overlay` is on `TaskEditModal`, `FlowPropertiesModal`, `BlockShortcutsDialog`, `BlockCommandMenu` and `BlockTaskPicker` — five, not four; the task picker is `Teleport to="body"` as well. I checked rather than assumed that a plain attribute survives `KsDialog` -> `ElDialog` `$attrs` onto the teleported element and is reachable by `closest()` from a field inside it.

`FlowFileEditorTab` stands down through a new `hasActiveAuthoringSurface()` plus containment on its own root, so the `Flow Code`-only layout is untouched.

`undo` stays out of the global set — that part of the previous commit was right and is unchanged.

Four tests, the foreign-editor one failing without the guard.

flcarre:
Round 4 addressed — `c5bb6268a3` and `de255066de`.

**The double save was real, and bigger than the one-liner.** Details on the thread; the short version is that `handleGlobalSave` checks neither the target nor the phase, so dropping `save` from `ALWAYS_GLOBAL_IDS` only fixes the chord typed *in a field* — typed on the canvas, which is not a typing target, it still reached both listeners.

So I took your first option, which means **restoring the `root` guard I removed in `d4586a9b32`**. You were right in round 3 and I was wrong to drop it: I argued it had nothing left to protect once `undo` was non-global, and this is exactly the thing it was protecting. It is back with the marker you sketched, on five overlays rather than four — `BlockTaskPicker` is `Teleport to="body"` too. Ownership now resolves to exactly one handler in every layout, table on the thread.

**The shared toast.** Thanks for flagging rather than waving through — two identical toasts with two Undo buttons that do the same thing is not what I wanted. Fixed in `de255066de`: the badge follows the surface the user is on, which `useAuthoringSurface.isActive()` already resolves to exactly one. So the history is shared, the toast is singular, and it appears wherever you are looking rather than only where the delete happened.

**On the CI note** — thank you for pinning the timeline, and your version is more accurate than mine. I had told Florian it was failing consistently; it is not, which is why 19767 and 19768 look green. Your framing is the right one: the story arrived on `develop` with #18062 at 06:16 UTC and the `pull_request` build tests the merge into that `develop`, so every branch inherits it from then on. That has its own ticket now rather than riding along here.

Nothing else from rounds 1–3 is open. Unit suite 2670 passing; the only local failure is `ksTokens.spec.ts`, which shells out to a `stylelint` binary absent from my checkout and passes in CI.

MilosPaunovic (commented):
Round 5, re-reviewed at `de255066de`. One item open, plus a test-coverage note.

## The `Ctrl`/`Cmd` + `S` handshake leaves a field with nobody to answer it

The editor yields on `hasActiveAuthoringSurface() && !tabRootEl.value?.contains(event.target)`, which is a broader set than the one the surface actually answers for. The surface answers for its own canvas and its marked overlays; it stands down for any other typing target. Measured at this head, with a surface root set and the four targets dispatched at the real composable:

| `Ctrl+S` typed in | surface | `FlowFileEditorTab` | result |
|---|---|---|---|
| `Monaco` (`Flow Code`) | stands down | saves (target is inside its root) | one save |
| an authoring overlay | answers, `preventDefault` | stands down | one save |
| the canvas | answers, `preventDefault` | stands down | one save |
| **a third panel's text field** | **stands down** | **stands down** | **nobody answers, nothing prevented** |

So with `Flow Code`, `Topology` and any third panel carrying a text field open at once - the `Blueprints` search, the `Files` rename box - `Ctrl+S` typed there now falls through to the browser's own save-page dialog. Before this `PR` the editor's listener answered it and saved the flow.

The editor cannot work out on its own whether the surface will answer, because the surface's root and its overlay marker are not visible from there. Asking rather than inferring keeps the two in step: a module-level `answersKeyFor(target)` on `useAuthoringSurface`, checking the registered roots plus `[data-authoring-overlay]`, used in place of that condition. Suggestion inline.

## The marker's delivery through `KsDialog` is untested

`useBlockEditorKeyboard.spec.ts` exercises the overlay path against a hand-made `div[data-authoring-overlay]`, so it would keep passing if the attribute never reached a real dialog. It does reach one - I mounted a `KsDialog` with it and the attribute lands on `DIV.el-dialog`, with `closest()` from an input inside finding it - but that rests on `KsDialog` spreading `$attrs` onto `ElDialog` and `ElDialog` forwarding them to the dialog box, neither of which this `PR` owns. A single test that mounts `TaskEditModal` (or any `KsDialog`) and asserts the marker is reachable from a field inside would pin the part that can actually break under a design-system or `element-plus` bump.

ui/src/components/inputs/FlowFileEditorTab.vue
```diff
@@ -406,14 +407,18 @@
...
-            event.preventDefault()
-            if (props.flow) {
-                saveFlowYaml()
-            } else if (isDirty.value) {
-                saveFileContent()
-            }
+        if (!((event.ctrlKey || event.metaKey) && event.key === "s")) return
+        // An authoring surface answers this chord for everything outside this editor, so without
+        // standing down both listeners would run their own save against the same flow.
+        if (hasActiveAuthoringSurface() && !tabRootEl.value?.contains(event.target as Node)) return
```
MilosPaunovic: This yields for every target outside the editor once any surface is mounted, but the surface only answers for its own canvas and its `[data-authoring-overlay]` overlays - it stands down for any other typing target, exactly as this editor does. A field in a third panel therefore falls between them and `Ctrl+S` reaches the browser instead of the flow.

The editor cannot see the surface's root or its overlay marker, so the condition has to come from the surface itself:

```ts
// useAuthoringSurface.ts - roots are already tracked per instance; expose the question.
export function authoringSurfaceAnswersKeyFor(target: Node | null): boolean {
    if (!target) return active.value.length > 0
    const el = target as HTMLElement
    if (el.closest?.("[data-authoring-overlay]")) return true
    return rootsOf(active.value).some(root => root?.contains(target))
}
```

and here:

```ts
if (authoringSurfaceAnswersKeyFor(event.target as Node)) return
```

That keeps the three working cases as they are and hands the fourth back to this listener.

flcarre: Right, and the gap is one I talked myself into: I reasoned through exactly this case in round 3, decided the browser dialog was an acceptable cost, and never checked what the behaviour had been before. It saved the flow. That is a regression, not a trade.

Took your sketch as written — `01f4df9d96`. `useAuthoringSurface` now tracks each active surface's root alongside its symbol, and exports:

```ts
export function authoringSurfaceAnswersKeyFor(target: Node | null): boolean {
    if (!target) return active.value.length > 0
    if ((target as HTMLElement).closest?.(`[${AUTHORING_OVERLAY_ATTRIBUTE}]`)) return true
    return active.value.some(id => roots.get(id)?.value?.contains(target))
}
```

`FlowFileEditorTab` is one line, and `hasActiveAuthoringSurface` plus the `tabRootEl` ref it needed are both gone — the editor no longer reasons about its own containment at all, which is what made the two conditions drift apart in the first place.

Your fourth row now lands on the editor:

| `Ctrl+S` typed in | surface | `FlowFileEditorTab` |
|---|---|---|
| `Monaco` (`Flow Code`) | stands down | saves |
| an authoring overlay | answers | stands down |
| the canvas | answers | stands down |
| a third panel's text field | stands down | **saves** |
| anywhere, `Flow Code` alone | not mounted | saves |

Four tests on `authoringSurfaceAnswersKeyFor`, one per row, including the one that was falling through.

flcarre:
Round 5 addressed — `01f4df9d96`.

**The handshake gap.** Fixed with your sketch; details on the thread. Worth naming how it got there: I reasoned through this exact case in round 3, decided a `Ctrl+S` in an unrelated field falling through to the browser was an acceptable cost, and wrote it down as acceptable — without checking what it did before. It saved the flow. So that was a regression I argued myself into, twice, and the asymmetry you spotted is what made it visible. `authoringSurfaceAnswersKeyFor` also removes the shape of the mistake: there is one condition now instead of two kept in step by hand.

**The marker test.** Fair, and the gap was real: my overlay test built its own `div[data-authoring-overlay]`, so it proved the guard reads a marker, not that a marker ever gets there. `TaskEditModal.spec.ts` now mounts the modal with the real `KsDialog` — only the heavy children stubbed — and asserts the marker is on the rendered dialog and reachable by `closest()` from a field inside it, plus that the element is `.el-dialog`, so the test cannot quietly start passing against a stub. Removing the attribute from `TaskEditModal` fails it.

Unit suite 2674 passing, and `Frontend - Tests` was green on the previous head.

For what it is worth: five rounds and every one of them found something real, including two things I had convinced myself were fine. Thank you for the persistence — particularly for measuring rather than describing, which is what made the last three unarguable.

ui/src/components/no-code/blocks/useAuthoringSurface.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Whether an authoring surface will answer a shortcut raised on this target. Another window
+ * listener for the same chord has to ask rather than infer: a surface's root and its
+ * `[data-authoring-overlay]` dialogs are not visible from outside, and a target belonging to
+ * neither is one the surface stands down for.
+ */
+export function authoringSurfaceAnswersKeyFor(target: Node | null): boolean {
+    if (!target) return active.value.length > 0
+    if ((target as HTMLElement).closest?.(`[${AUTHORING_OVERLAY_ATTRIBUTE}]`)) return true
+    return active.value.some(id => roots.get(id)?.value?.contains(target))
+}
```
MilosPaunovic: `contains` is the wrong question for a target that is nobody's descendant. With nothing focused the keydown targets `body`, which is an ancestor of every surface root, so this returns `false`, the editor keeps the chord, and the surface takes it too - two saves against the same flow from one keystroke.

The surface's actual rule is in `useBlockEditorKeyboard`: it stands down only for a *typing* target that is neither inside its root nor inside a marked overlay. Mirroring that here removes the guesswork, and exporting `isTypingTarget` (module-private today) keeps one definition rather than two that can drift - which is where both of these holes came from:

```ts
export function authoringSurfaceAnswersKeyFor(target: Node | null): boolean {
    if (active.value.length === 0) return false
    if (!target) return true
    if ((target as HTMLElement).closest?.(`[${AUTHORING_OVERLAY_ATTRIBUTE}]`)) return true
    if (active.value.some(id => roots.get(id)?.value?.contains(target))) return true
    // A chord raised on nothing in particular belongs to the canvas, not to a field.
    return !isTypingTarget(target)
}
```

Left as a plain block rather than a one-click `suggestion` because it does not stand alone: `isTypingTarget` has to be exported from `useBlockEditorKeyboard.ts` first, and applying this on its own would not compile.

Worth a test for the `body` target specifically - the suite covers monaco, the overlay and the canvas child, and the case that regressed is the one none of them reach.

MilosPaunovic (commented):
Round 6, re-reviewed at `01f4df9d96`. One item open.

## The handshake now double-saves when nothing is focused

Re-running the matrix at this head with both real window listeners in play:

| `Ctrl+S` typed in | surface | editor | result |
|---|---|---|---|
| `Monaco` | stands down | saves | one save |
| an authoring overlay | saves | stands down | one save |
| a third panel's text field | stands down | saves | one save |
| **nothing focused (`body`)** | **saves** | **saves** | **two saves** |

`authoringSurfaceAnswersKeyFor` asks `roots.get(id)?.value?.contains(target)`, and `body` is an *ancestor* of the canvas rather than a descendant, so `contains` is `false` and the editor no longer stands down. The previous `!tabRootEl.value?.contains(event.target)` covered that case as a side effect of being broader.

It is the ordinary state, not an exotic one: a keydown with no focused element targets `body`, and the topology canvas is not focusable - there is no `tabindex` anywhere in `@vue-flow/core` - so clicking the canvas and pressing `Ctrl+S`, which is how the shortcut is normally reached on that surface, lands exactly there. Measured: `surface: ["save"]` and `editorSaved: true` off one keystroke.

Suggestion inline. The shape that closes it for good is having the helper mirror the surface's own rule instead of approximating it, which also means the two sides stop being able to drift - both holes so far have come from the same drift.

## Note on the red build

The only failing check is `trigger-ee`: `curl: (28) Operation timed out after 2700002 milliseconds`, `HTTP 000000`, so the `EE` compile webhook returned no verdict at all. That is the inconclusive infra path, not a result - it wants a re-run rather than a change. Everything else, `Frontend - Tests` included, is green.

ui/src/components/no-code/blocks/useAuthoringSurface.ts
flcarre: Right on both counts — the hole and where it came from. Fixed in `34091f80a8`, taking your block as written, with `isTypingTarget` exported so there is one definition rather than two.

The `body` case is the ordinary one, not an edge: I checked `@vue-flow/core` for a `tabindex` and there is none, so clicking the canvas leaves focus on `body` and that is the normal path to the chord on that surface.

You are right that both holes came from the same drift, so I added the thing that makes drift fail instead of shipping: `saveChordOwnership.spec.ts` mounts a real surface, runs **both** real window listeners over every kind of target — `body`, a canvas child, a field inside the surface, a field in a teleported marked overlay, a `Monaco` textarea outside, a third panel's plain field — and asserts `surface !== editor` for each, plus the no-surface-mounted case.

I checked it catches what it is meant to catch rather than assuming: reverting the helper to the round-6 shape fails the `body` row, and swapping the editor's decision back to the round-5 containment shape fails three rows. Had that file existed at round 4 neither of us would have spent a round on this.

Also added the `body` unit test on the helper itself, as you suggested.

flcarre:
Round 6 addressed — `34091f80a8`. Details on the thread.

You have now found the same class of bug three rounds running, and each time I fixed the instance you named rather than the thing generating them. That is on me, so this round also closes the generator.

**What changed structurally.** `authoringSurfaceAnswersKeyFor` no longer approximates the surface's rule, it mirrors it, sharing one exported `isTypingTarget` rather than a second copy of the same idea. And `saveChordOwnership.spec.ts` mounts a real surface and runs **both** real window listeners over every kind of target — `body`, a canvas child, a field inside the surface, a field in a teleported marked overlay, a `Monaco` textarea outside, a third panel's plain field — asserting exactly one side answers each. I verified it fails on both shapes it is meant to reject: the round-5 editor condition fails three rows, the round-6 helper fails the `body` row.

**I also swept my own changes adversarially before pushing this time**, instead of reasoning about them and letting you measure. Three of those probes found nothing but were worth keeping, so they are in the suite now: a `Switch` case keyed `'a.b[0]'` and `'with space'` (the path grammar quotes them and `duplicateBlockAtPath` resolves them), an `If` with an empty `then` and a populated `else`, and a validation message carrying two brackets. I also checked the undo history's scope key against flow creation — `FlowCreate` sets `flowStore.flow` with a `getRandomID()` before the editor mounts, and the editor is gated on that being present, so the empty scope I was worried about is unreachable and two successive new flows get distinct ones.

**On `trigger-ee`** — agreed, and thank you for reading the actual failure rather than the red. `curl: (28)` after 2700s with `HTTP 000000` is no verdict, not a negative one. It re-ran with this push.

Unit suite 2686 passing.

MilosPaunovic (approved):
Approving at `34091f80a8`.

Everything raised over the previous rounds is closed. I reproduced each failure before the fix and re-ran it after, so this is not a read-through: the `Ctrl+S` ownership matrix is correct in all six cases at this head, the focus walk reaches every branch and case, and the undo history and its toast now tell one story.

Two things this approval deliberately does not cover:

- **`CI` has not reported on this head.** `Frontend - Tests` and `E2E - Tests` are both still in progress at `34091f80a8`, and the last commit touches `useBlockEditorKeyboard`, which several suites exercise. `trigger-ee` also needs a re-run after its earlier webhook timeout (`curl: (28)` after 2700s, `HTTP 000000`, no compile verdict). Please hold the merge until that run is back.
- **The scope question from the first round is still unanswered.** https://github.com/kestra-io/kestra/issues/19338 puts the command menu, a graph keyboard-navigation model, drag-reorder and section-level add affordances beyond `add-error` under **Out of Scope**, and this `PR` ships all four, plus the flow header bar, trigger insertion, and shared `BasicNode` / `DotNode` changes that reach the execution topology. That is a product call rather than a code one, so it is not blocking the review - but it deserves an answer from someone rather than merging by default.

One more thing worth saying plainly: I have not driven this in a browser. The drag-and-drop, the keyboard walk, the command palette and the picker positioning I verified by reading the code and by unit test, not by using them. For a `PR` that adds a whole authoring surface, someone should exercise it for real before it lands.

flcarre:
## QA — browser pass on a running instance

**Verdict: PASS** for every scenario exercised. Milos noted in his approval that nobody had driven this in a browser; this closes that. OSS from source at `488b26bd10` (branch + develop merged), embedded H2, isolated ports, Flow Code and Topology open side by side — the layout the keyboard conflicts needed.

Not a read-through: each row below was measured in the page, and where a negative is the result (no chip, one save) I checked the positive control first, because a selector that matches nothing looks identical to a feature that is correctly absent.

| # | Scenario | Result |
|---|---|---|
| 1 | Edge `+` opens the task picker in place | ✅ 15 `+` on the canvas, click opens the picker |
| 2 | Picker → `TaskEditModal` with the chosen type | ✅ modal opens, type prefilled, required-field error shown |
| 3 | **`Ctrl+Z` in a text field does not rewind the flow** | ✅ typed in the modal's `id` field, `Ctrl+Z` left the graph at 9 nodes and the modal open |
| 4 | **`Cmd+S` in the modal → one save** | ✅ exactly one `PUT /flows/...` |
| 5 | **`Cmd+S` on the canvas (focus on `body`) → one save** | ✅ exactly one `PUT /flows/...` |
| 6 | `Duplicate` on a task inside `else` | ✅ `in_else_copy` inserted |
| 7 | Flow chip on the **editor** topology | ✅ chip + Configure render |
| 8 | Flow chip on the **execution** topology | ✅ absent, with the graph proven rendered (3 nodes) |
| 9 | `data-authoring-overlay` on the real teleported dialog | ✅ lands on `DIV.kel-dialog`, and on the picker |

Details worth having:

- **4 and 5** are the six-round question, so I counted requests rather than trusting the absence of a visible glitch: network cleared, one chord, then every `PUT`/`POST` to `/flows` that is not `validate`. One each. The `PUT` answers `422` because the task I inserted has no `message` yet — that is the validation working, not the save misfiring.
- **3** proves the canvas does not rewind. It does not prove the field's own undo ran: synthetic typing does not populate the browser's native undo stack, so the field kept its text. The half that regressed is the half that is covered.
- **8** first looked green against a URL that turned out to be `Page not found`, which would have made the assertion meaningless. The row above is from `/overview`, with `vue-flow` present and three nodes on screen.
- **9** is the part no jsdom test can settle, since it runs through `KsDialog` → `ElDialog` `$attrs` forwarding. The attribute is on `DIV.kel-dialog` in the live page.

**Not exercised**, so not claimed: drag-reorder on the canvas, the command palette, and the full arrow-key walk. The focus order behind the last two is covered by unit tests that fail without their fix, including `duplicateBlockAtPath` resolving every recorded path, and row 6 exercises that order end-to-end through the node menu.

<details><summary>Flow used</summary>

```yaml
id: qa_topology_authoring
namespace: company.team
tasks:
  - id: extract
    message: extract
  - id: gate
    condition: "{{ true }}"
    then:
      - id: in_then
        message: then
    else:
      - id: in_else
        message: else
  - id: router
    value: "{{ 'a' }}"
    cases:
      a:
        - id: in_case_a
          message: a
      b:
        - id: in_case_b
          message: b
  - id: publish
    message: publish
errors:
  - id: notify
    message: notify
```
</details>

PR #19343, opened 2026-09-10
fix(worker): honour terminationGracePeriod when draining jobs on shutdown

Skraye:
Fixes the worker shutdown drain, which gave up after 10 seconds instead of the configured kestra.server.terminationGracePeriod, so a scaled-down or restarted worker abandoned its in-flight tasks and skipped the Worker IO flush entirely

### DIFF
--- a/worker/src/main/java/io/kestra/worker/AbstractWorker.java
+++ b/worker/src/main/java/io/kestra/worker/AbstractWorker.java
@@ -36,6 +36,7 @@ import io.kestra.worker.senders.WorkerIOSender;
 
 import io.micronaut.context.event.ApplicationEventPublisher;
 import lombok.extern.slf4j.Slf4j;
+import org.awaitility.core.ConditionTimeoutException;
 
 import static io.kestra.core.server.Service.ServiceState.TERMINATED_FORCED;
 import static io.kestra.core.server.Service.ServiceState.TERMINATED_GRACEFULLY;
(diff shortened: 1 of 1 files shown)


worker/src/main/java/io/kestra/worker/AbstractWorker.java
```diff
@@ -330,25 +331,37 @@ private boolean waitForJobsCompletion(final Duration timeout) {
...
-                    log.info("All worker threads are terminated");
-                } else {
-                    log.warn("Waiting for all worker job to terminate (remaining: {}).", runningJobs);
-                }
-                return false;
-            });
+        try {
+            Await.await()
+                // Awaitility caps at 10s by default, and honouring the grace period is the whole point here.
+                .atMost(timeout)
```
fhussonnois: You have to add a margin here, otherwise the Awaitility timeout will happen at the same time that the VirtualThread waiting for the executor shutdown leading to a potential race condition on `shutdownState` . 

`private static final Duration FORCED_SHUTDOWN_MARGIN = Duration.ofSeconds(5);`

`.atMost(timeout.plus(FORCED_SHUTDOWN_MARGIN)`

worker/src/main/java/io/kestra/worker/AbstractWorker.java
```diff
@@ -330,25 +331,37 @@ private boolean waitForJobsCompletion(final Duration timeout) {
...
+                    } else {
+                        log.warn("Waiting for all worker job to terminate (remaining: {}).", runningJobs);
+                    }
+                    return false;
+                });
+        } catch (ConditionTimeoutException e) {
+            // Returning instead of propagating keeps doStop() on its normal path, which still has the
+            // Worker IO senders to flush.
+            log.warn(
+                "Grace period of {} elapsed with {} job(s) still running, forcing termination.",
```
fhussonnois: Then, I would change this by: 

`    log.warn(
        "Worker shutdown did not settle within {} ({} job(s) still running), forcing termination.",
        timeout.plus(FORCED_SHUTDOWN_MARGIN), this.workerJobExecutor.getRunningJobCount()
    );`

fhussonnois (approved):
Approved, minus some changes :)

PR #19344, opened 2026-09-10
refactor(ui): replace explicit any in KV components

### LINKED ISSUE #19293: KV components: replace explicit `any` with real types
22 explicit `any` across 2 files in `src/components/kv`, from `ui/scripts/explicit-any/baseline.json`:

| any | file |
| --- | --- |
| 18 | `KVTable.vue` |
| 4 | `kvValue.ts` |

LouisDeconinck:
Closes #19293.

The KV table and value helpers contained 20 explicit `any` usages that bypassed the frontend compiler.

This replaces them with the generated KV/API types and existing design-system form types, adds typed adapters for the form controls' value shapes, and removes both files from the explicit-any baseline. The API request shape is typed from the generated SDK, including the required sort-array representation.

Validation:
- `npm run check:types`
- `npm run test:unit -- tests/unit/components/kv/kvValue.spec.ts` (31 tests)
- `npm run test:lint` (passes; existing warnings remain outside this change)
- `npm run check:ts-any`


ui/src/components/kv/kvValue.ts
```diff
@@ -1,34 +1,39 @@
...
+export function formatKvValueForDisplay(type: string, value: unknown, timezone?: string): string {
     if (type === "JSON") {
-        return JSON.stringify(value, null, 2)
+        return JSON.stringify(value, null, 2) ?? ""
     }
     if (type === "DATETIME") {
         // Follow Timezone from Settings to display KV of type DATETIME (issue #9428)
         const tz = timezone || moment.tz.guess()
-        return moment(value).tz(tz).format()
+        return moment(value as string | number | Date).tz(tz).format()
```
Piyush-r-bhaskar: Swaps `any` for `unknown` then casts right back, so nothing is actually checked. Could we narrow it instead? Line 57 already does it properly.

ui/src/components/kv/kvValue.ts
```diff
@@ -1,34 +1,39 @@
...
     }
     return String(value)
 }
 
 /**
  * Converts a KV value returned by the API into what the edit form's value control expects.
  */
-export function hydrateKvValueForForm(type: string, value: any, timezone?: string): any {
+export function hydrateKvValueForForm(type: "DATETIME", value: unknown, timezone?: string): Date
+export function hydrateKvValueForForm(type: "BOOLEAN", value: unknown, timezone?: string): boolean
```
Piyush-r-bhaskar: Nothing calls this overload, the component passes a `KvType` union. Drop it? The DATETIME one is used by a test.

ui/src/components/kv/kvValue.ts
```diff
@@ -1,34 +1,39 @@
...
+export function hydrateKvValueForForm(type: string, value: unknown, timezone?: string): KvFormValue
+export function hydrateKvValueForForm(type: string, value: unknown, timezone?: string): KvFormValue {
     if (type === "JSON") {
-        return JSON.stringify(value)
+        return JSON.stringify(value) ?? ""
     }
     if (type === "BOOLEAN") {
-        return value
+        return value as boolean
```
Piyush-r-bhaskar: Says boolean without checking. `value === true` would be honest.

ui/src/components/kv/kvValue.ts
```diff
@@ -37,19 +42,20 @@ export function hydrateKvValueForForm(type: string, value: any, timezone?: strin
  * Serializes a form value into the ION payload the API infers the KV type back from,
  * so that saving then reopening an entry yields the same type and value.
  */
-export function serializeKvValueForSave(type: string, value: any): string {
+export function serializeKvValueForSave(type: string, value: KvFormValue): string {
     if (type === "STRING") {
         // Quoted, so a string that looks like a number or a boolean stays a string.
         return JSON.stringify(value)
```
Piyush-r-bhaskar: Lines 11 and 29 got a `?? ""` guard, this one can be `undefined` too. Same treatment?

ui/src/components/kv/KVTable.vue
```diff
@@ -8,12 +8,12 @@
         :pageSize="urlSize"
         :defaultSort="{prop: 'key', order: 'ascending'}"
         @page-changed="({page, size}: {page: number; size: number}) => router.push({query: {...route.query, page: String(page), size: String(size)}})"
-        @sort-change="({prop, order}: {column: any; prop: string | null; order: string | null}) => router.push({query: {...route.query, sort: `${prop}:${order === 'ascending' ? 'asc' : 'desc'}`}})"
+        @sort-change="({prop, order}: {column: unknown; prop: string | null; order: string | null}) => router.push({query: {...route.query, sort: `${prop}:${order === 'ascending' ? 'asc' : 'desc'}`}})"
```
Piyush-r-bhaskar: `column` is unused, we only take `prop` and `order`. Drop it from the type.

ui/src/components/kv/KVTable.vue
```diff
@@ -580,25 +601,25 @@
...
         }
     }
 
-    async function updateKvModal(entry: any) {
+    async function updateKvModal(entry: KvEntry) {
         kv.value.namespace = entry.namespace
         kv.value.key = entry.key
-        const {type, value} = await namespacesStore.kv({namespace: entry.namespace, key: entry.key}) as {type: string, value: any}
-        kv.value.type = type
+        const {type, value}: KvControllerKvDetail = await namespacesStore.kv({namespace: entry.namespace!, key: entry.key!})
```
Piyush-r-bhaskar: `KvControllerKvDetail` says the value is an object map (from `Object value` in `KVController`), but for STRING/NUMBER/BOOLEAN it is a scalar. Nothing breaks now, but `value.something` would compile later.

The two `!` also promise without checking. Both callers get the entry from a rendered row, so one guard covers it.

ui/src/components/kv/KVTable.vue
```diff
@@ -395,7 +397,9 @@
         total.value = kvsResponse.total ?? 0
     }
 
-    const loadQuery = (base: any) => {
+    type KvQuery = NonNullable<ListAllKeysData["query"]>
```
Piyush-r-bhaskar, line 400: Minor: says `KvQuery`, but `_merge` still folds in leftover route query keys, so the object carries more than the type shows.

ui/src/components/kv/KVTable.vue
```diff
@@ -700,7 +723,14 @@
                 : undefined
             const ttl = preservedTtl ?? kv.value.ttl
 
-            const payload = {
+            const payload: {
```
Piyush-r-bhaskar: Same shape is already on `createKv` in `useBaseNamespaces`. `Parameters<typeof namespacesStore.createKv>[0]`, or just `...(ttl ? {ttl} : {})` with no annotation.

ui/src/components/kv/KVTable.vue
```diff
@@ -757,7 +787,7 @@
         }
     })
 
-    const formRef = ref()
+    const formRef = ref<FormInstance>()
```
Piyush-r-bhaskar, line 787: Fine to leave, matches SecretsTable and McpEdit. Just noting the ref holds the `KsForm` wrapper, which exposes 5 methods, not a full Element Plus form.

Piyush-r-bhaskar (changes requested):
Nice cleanup, the component half reads much better.

Ran locally, all green: `check:types`, `check:ts-any`, the 31 `kvValue` tests, oxlint. The `sort` change is safe, a one element array serialises the same as the old string.

Main thing: `kvValue.ts` does not really remove the `any`, it writes `unknown` and casts back on the next line. Small fixes inline.

Could you also run `tests/e2e/kv/kv-types.spec.ts`? It covers all 7 types through the drawer and the e2e job is skipped on forks.

LouisDeconinck:
Done — the casts are gone rather than renamed:

- `kvValue.ts`: a `toDateInput` guard narrows `unknown` to `string | number | Date | undefined`, so the `moment(...)`/`new Date(...)` calls no longer need `as`. BOOLEAN is `value === true`, the dead BOOLEAN overload is dropped (DATETIME stays — the spec relies on it), and the STRING serialize path got the same `?? ""` guard.
- `KVTable.vue`: sort-change types only `prop`/`order`; a `KvDetail` type reflects that the detail `value` is a scalar for non-JSON types, so `value.something` no longer compiles; `entry.namespace`/`entry.key` are guarded in `updateKvModal`/`viewKvModal` instead of asserted; `loadQuery` returns `KvQuery & Record<string, unknown>` since `_merge` keeps leftover route keys; and the save payload reuses `Parameters<typeof namespacesStore.createKv>[0]` with a conditional `ttl` spread.

Checks: `check:types`, `check:ts-any` (unchanged vs baseline), the 31 `kvValue` unit tests and oxlint all pass. Also ran `tests/e2e/kv/kv-types.spec.ts` locally against the vite dev server + the e2e docker backend — all 13 pass. Note for running it on a fork: the dev-server index.html never gets the `csrf-token` meta the backend injects, so cookie-authenticated writes hit `CsrfTokenFilter` — I set `micronaut.security.csrf.enabled: false` on the test backend to run the suite (reverted after). CI serves the built UI so it is unaffected.

LouisDeconinck:
All inline comments addressed in fa3343df1 (pushed earlier today), plus a follow-up test commit pinning the new guards:

- `kvValue.ts`: `toDateInput` now narrows instead of casting back; dropped the unused `BOOLEAN` overload; `value === true` for the switch; `?? ""` on the serialize STRING branch.
- `KVTable.vue`: `column` dropped from the sort-change type; `KvDetail` corrects the SDK's object-map `value` type to scalar-or-map and both `!` are replaced by one guard shared by `updateKvModal`/`viewKvModal`; `loadQuery` returns `KvQuery & Record<string, unknown>` so leftover route keys are honest; payload uses `Parameters<typeof namespacesStore.createKv>[0]`; `formRef` left as `FormInstance` per the SecretsTable/McpEdit convention.

`tests/e2e/kv/kv-types.spec.ts` — ran it locally against this branch's production build (the e2e job skips forks): **13/13 pass**. Note the spec flakes at 10 workers on a shared store; I used `--workers=2`.

Also verified: `check:types`, `check:ts-any` (baseline unchanged), `vitest` kvValue spec (32 tests).

ui/src/components/kv/KVTable.vue
LouisDeconinck: Addressed in fa3343df1 — `loadQuery` now returns `KvQuery & Record<string, unknown>`, so the leftover route keys folded in by `_merge` are explicit in the type rather than hidden behind `KvQuery`.

MilosPaunovic: If you've addressed any of the requested changes from the PR review, @LouisDeconinck, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

Piyush-r-bhaskar:
@LouisDeconinck could you fix the conflicts please ?

LouisDeconinck:
Conflicts resolved — rebased onto the latest `develop`, merging the upstream moment → dayjs migration with the narrowed types. `check:types`, `check:ts-any` and the kv unit tests all pass locally.

Piyush-r-bhaskar (approved):
I fixed some types f5d7f86844d5ae8ab6f04427d1c7b68577817efc to avoid the usage of `unknown` just to bypass `any` now LGTM 🚀 Thanks for the contribution

PR #19349, opened 2026-09-10
feat(core): add keyset findAfterWithoutAcl to the log data store

nkwiatkowski:
## What

Adds a demand-driven, keyset-paginated read to the log data store:
`LogDataStoreInterface.findAfterWithoutAcl(tenantId, filters, afterTimestamp, afterKey, pageSize)`
returning one bounded page ordered by `(timestamp, key)`, strictly after the given
position, holding no cursor or transaction between calls. JDBC implements it as a
jOOQ row-value seek `(timestamp, key) > (?, ?)`.

This is the OSS half of the EE `LogShipper` OOM fix (companion EE PR below).
`findAsync` is left unchanged (still used elsewhere).

## Why

The EE log shipper reads the whole backlog through the eager, unbounded `findAsync`
(`Flux.create(..., BUFFER)`), which OOMs on large exports when the destination upload
lags and the reactive buffer grows without bound. Keyset pagination lets the shipper
export one bounded page at a time.

## Tests

- `AbstractLogDataStoreTest.findAfterWithoutAcl_paginatesAcrossEqualTimestampsWithoutLossOrDuplicates`
  pages across a same-timestamp group asserting exactly-once. Green on H2, Postgres, MySQL.

## Notes

- `WithoutAcl`: no repository-layer ACL (matches `findAsync`); the caller (the shipper, which has
  no user context during execution) checks authorization before reading.
- A `(tenant_id, timestamp, key)` composite index is a planned follow-up; correctness holds without
  it (Postgres incremental sort).

Part of https://github.com/kestra-io/plugin-ee-aws/issues/198

Companion PR (EE, consumes this): https://github.com/kestra-io/kestra-ee/pull/10981

### DIFF
--- a/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLogDataStore.java
+++ b/jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLogDataStore.java
@@ -1,5 +1,8 @@
 package io.kestra.jdbc.repository;
 
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.*;
 import java.util.function.BiFunction;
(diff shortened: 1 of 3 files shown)


jdbc/src/main/java/io/kestra/jdbc/repository/AbstractJdbcLogDataStore.java
```diff
@@ -209,6 +212,42 @@ public Flux<LogEntry> findAsync(@Nullable String tenantId, List<QueryFilter> fil
...
+        // Same default-kind rule as findAsync: NORMAL only unless a KIND or executionId filter is present.
+        var condition = this.filter(filters, DATE_COLUMN, Resource.LOG);
+        if (!QueryFilter.hasField(filters, QueryFilter.Field.KIND) && !QueryFilter.hasField(filters, QueryFilter.Field.EXECUTION_ID)) {
+            condition = NORMAL_KIND_CONDITION.and(condition);
+        }
+
+        // With a key, seek strictly after the (timestamp, key) row; without one (first run or a legacy date-only
+        // offset), fall back to a strict timestamp lower bound (which matches the previous START_DATE behaviour).
+        Field<OffsetDateTime> dateField = field(DATE_COLUMN, OffsetDateTime.class);
+        OffsetDateTime after = afterTimestamp.atOffset(ZoneOffset.UTC);
```
loicmathieu, line 231: Are you sure it's correct also in MySQL? Offset dates are usually handled separatly there.

nkwiatkowski: It's not an issue because it's persisted with it's offset:  
 `timestamp` DATETIME(6) GENERATED ALWAYS AS
    (STR_TO_DATE(value ->> '$.timestamp', '%Y-%m-%dT%H:%i:%s.%fZ')) STORED NOT NULL

loicmathieu (approved):
LGTM

PR #19374, opened 2026-09-10
fix(webserver): make bulk flow delete by-ids atomic and validate ids upfront

### LINKED ISSUE #18511: [Bash2.0] Bulk flow delete by-ids deletes valid flows then returns 404 when any id is missing
## Describe the issue

Deleting flows in bulk by ids permanently deletes the valid flows but returns HTTP 404 when any id in the list does not exist. The deletion is not atomic, and the error response makes it look like nothing was deleted.

## Steps to reproduce

1. Have an existing flow, e.g. `company.team/my_flow`.
2. Bulk-delete by ids with the existing flow followed by a non-existent id:
   ```
   curl -X DELETE -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
     -d '[{"namespace":"company.team","id":"my_flow"},{"namespace":"company.team","id":"does-not-exist"}]' \
     http://localhost:8080/api/v1/a/flows/delete/by-ids
   ```

## Actual behaviour

The response is HTTP 404, but `my_flow` is already deleted. Valid flows before the missing id are removed, while the request reports failure.

## Environment

- Kestra Version: 2.0.0-rc10

nancysangani:
## What
Fixes #18511 — bulk-deleting flows via `DELETE /api/v1/{tenant}/flows/delete/by-ids` deleted valid flows before checking later ids in the list, then returned a 404 as soon as it hit a missing id. This made the operation non-atomic and the error response misleading (it looked like nothing was deleted, when in fact prior flows in the list were already gone).

## Why it happened
The endpoint used a single Java stream that mapped each id to a lookup (`.orElseThrow()`) and deleted it via `.peek()` in the same pass. Streams evaluate lazily per element, so flows before a bad id were deleted before the bad id was even reached.

## Fix
Validate every id upfront, collect any missing ones, and only delete if the whole batch is valid — mirroring the existing pattern already used in `ExecutionController#deleteExecutionsByIds`. If any id is missing, the request fails with `400 Bulk Validation Failed` (one `ProblemError` per missing flow) and nothing is deleted.

## Testing
- Added `deleteFlowsByIdsShouldDeleteAllGivenFlows` — happy path, all ids valid.
- Added `deleteFlowsByIdsShouldNotDeleteAnyFlowWhenOneIdIsMissing` — regression test for the reported bug: asserts a `400`/`bulk-validation-failed` problem is returned and the valid flow in the batch is *not* deleted.

### DIFF
--- a/webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
+++ b/webserver/src/main/java/io/kestra/webserver/controllers/api/FlowController.java
@@ -53,6 +53,11 @@ import io.kestra.webserver.responses.PagedResults;
 import io.kestra.webserver.services.SourceSearchService;
 import io.kestra.webserver.utils.CSVUtils;
 import io.kestra.webserver.utils.PageableUtils;
+import io.kestra.webserver.errors.ProblemDetail;
+import io.kestra.webserver.errors.ProblemError;
+import io.kestra.webserver.errors.ProblemType;
+import io.kestra.webserver.errors.ProblemTypes;
+import io.kestra.webserver.exceptions.BulkValidationException;
 
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.data.model.Pageable;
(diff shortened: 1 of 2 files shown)


loicmathieu:
@nancysangani I refactored this part of the code last week, can you check if the issue still exist and in this case rebase your changes?

nancysangani:
Thanks for looking out @loicmathieu! I checked develop - `deleteFlowsByIds` in `FlowController.java` is unchanged, still has the same stream `.peek()`-before-validation bug. My fix still applies cleanly. I've already merged the latest `develop` into my branch, so it's up to date.

PR #19383, opened 2026-09-11
fix(executions): duration without queued time, queued duration field and column

### LINKED ISSUE #7989: Queued execution time
### Describe the issue

When a task is queued, its execution time is counted from the moment it is queued. It would be more interesting to display its real execution time first. Ex: I launch a ForEach of 10000 flows with a concurrencyLimit of 32. Each task only takes a few tens of seconds but the execution time displayed for the last tasks indicates several hours because the start time is that of queued and not that of running

### Environment

- Kestra Version: 0.21.7

gasper94:
Closes https://github.com/kestra-io/kestra/issues/7989.

### ✨ Description

**Problem.** `state.duration` was the time between an execution's first and last state transition, so the time spent `QUEUED` behind a concurrency limit counted as if the execution had been running. With a flow limited to a few concurrent executions, the executions list shows queued rows with a growing duration, and the last executions to run end up with a duration of minutes or hours for a few seconds of work. The same value feeds the `state_duration` column (sorting, dashboards, statistics), the executor duration metric, notification templates and the execution progress estimate.

**Fix.** `State.getDuration()` now counts only the spans opened by a running or paused state, which is the split the execution overview already shows in its duration popover (`computeDurationBreakdown`: queued / running / paused):

- `duration` is the elapsed time minus the waiting time. Waiting time is every span opened by a state that is neither running (`RUNNING`, `KILLING`) nor paused (`PAUSED`, `BREAKPOINT`): before the first `RUNNING` transition, while `QUEUED`, and between retry attempts. Paused time stays in `duration`, as before.
- A new read-only `state.queuedDuration` exposes that waiting time on executions, task runs and attempts. It is omitted while the execution is still waiting, since a persisted value would go stale; the UI computes it live from the histories, as it already does for `duration`.
- Executions list: the Duration column shows the running time, live and without the queue wait for non-terminated executions. A new "Queued duration" column, hidden by default and not sortable (no database column backs it), shows the wait.
- Execution overview and Gantt: the duration popover headlines the running time instead of the total elapsed time; the split bar, the shares and the state history keep the total. The breakdown now treats `QUEUED`, `RETRYING` and `RESTARTED` as in progress, so a queued execution's wait counts live instead of freezing at the transition.
- Execution progress bar: the elapsed time excludes the queue wait, like the average duration it is compared with.

`duration + queuedDuration` is always the elapsed time between the first and last transition.

**Evidence.** `StateTest` covers the new definition (queued wait excluded from the persisted and the live duration, `queuedDuration` empty while queued, retry gaps billed as queued, paused time kept); `durationBreakdown.test.ts` and `Duration.test.ts` cover the UI side. The repository sort tests keep passing with realistic histories that go through `RUNNING`.

Against a local `server local` instance, with a flow limited to 2 concurrent executions (`behavior: QUEUE`) running a 20 s `Sleep` task and 6 executions triggered at once, the search API returns:

| execution | state | duration | queuedDuration | histories |
|---|---|---|---|---|
| 1, 2 | SUCCESS | PT20.2S | PT0.03S | CREATED > RUNNING > SUCCESS |
| 3, 4 | SUCCESS | PT20.2S | PT20.1S | CREATED > QUEUED > RUNNING > SUCCESS |
| 5, 6 | SUCCESS | PT20.3S | PT40.2S | CREATED > QUEUED > RUNNING > SUCCESS |
| still queued | QUEUED | absent | absent | CREATED > QUEUED |

Before this change, executions 5 and 6 reported a duration of about a minute.

Executions list with the "Queued duration" column enabled, 26 s after triggering (queued rows show a running time of 0 s and a live wait; running rows show their live running time and the wait they went through; terminated rows show 20 s and 0 s / 40 s):

executions list

Execution overview of an execution that waited 20 s in the queue, with the duration popover open:

execution overview

### 🎨 Frontend Checklist

### 🛠️ Backend Checklist

### 📝 Additional Notes

- Issue #7989 was closed as not planned; this PR proposes the fix anyway, since the compounded duration is still what users see on a concurrency-limited flow. Happy to split or scale it down if you prefer only the `QUEUED` state to be excluded.
- The definition is deliberately the one of the existing UI breakdown rather than "exclude the `QUEUED` state only", so that the persisted `duration`, the executions list and the overview popover agree. The consequence is that the retry delay between attempts and the time a `FAILED` execution sat before a restart no longer count as duration either.
- Existing rows keep the `state.duration` they were stored with (the `state_duration` generated column is computed from the JSON at write time); only executions written after the upgrade use the new definition.
- The `kestra-sdk` types are not regenerated here (CI does it on `develop`): the list reads `state.queuedDuration` from the untyped row and derives the live value from `state.histories`.
- The column description for the new column was added to the 13 locales of `KsFilter.locale.ts` by hand, with its fingerprint computed through `fingerprintRules.mjs`, since generating it needs a Gemini key.
- Not changed on purpose: the dashboards table `DURATION` column still falls back to `start_date` for non-terminated rows, which carry no histories.

### 🤖 AI Authors

Why did the cat refuse to wait in the execution queue? It had already used up eight of its nine retries.

https://claude.ai/code/session_013uQ5MAqXDFcaDuGqCa3qf5


loicmathieu:
Hi, thanks for your contribution but we close the issue as not planned.

PR #19385, opened 2026-09-11
refactor(core): drop the i18n-t component, with a test that keeps it out

### LINKED ISSUE #19384: Drop the last uses of `<i18n-t>`, and keep it out
Ten components use `<i18n-t>` to put a link, a code tag or an id inside a translated sentence. We do not use that component anywhere else, and the same thing is done in two simpler ways in the rest of the codebase: pass the markup as a parameter to `$t()`, or split the sentence around the slot and render the component in the middle.

Nothing is broken today, so this is tidying rather than a fix. The reason to do it is that one component doing it its own way is how the next person learns the wrong pattern, and `<i18n-t>` needs its own import, its own named slots and a `keypath` instead of a plain `$t()` call.

Seven are in OSS: the task edit modal, the trigger enable dialog, the two copilot components, and the three cross-resource search ones. Three are in EE: the case activity card, the worker delete dialog, and the reusable inputs panel.

While doing it, add a unit test that reads every `.vue` file and fails when one breaks the i18n conventions, so this does not come back:

- no `<i18n-t>`
- templates call `$t()`, not the `t()` from `useI18n()`
- `<script>` calls `t()`, not `$t()`
- no `v-t` directive
- no `<i18n>` block inside a component, keys live in the translation files

The test only reads code, so a comment or a piece of text that happens to contain one of these does not trip it.

Piyush-r-bhaskar:
Closes https://github.com/kestra-io/kestra/issues/19384.
EE PR: https://github.com/kestra-io/kestra-ee/pull/10988

## What

Ten components used `<i18n-t>` to put a link, a code tag or an id inside a translated sentence. Nothing else in the codebase does that, so they now use the two patterns the rest of the code already uses: the markup passed as a parameter to `$t()`, or the sentence split around the slot with the component rendered in the middle. Seven components here, three in EE.

A unit test reads every `.vue` file under `src` and fails when one breaks an i18n convention, so this does not come back:

| Rule | Where it looks | Why |
| --- | --- | --- |
| `<i18n-t>` or `<I18nT>` | template | use `$t()` with named arguments |
| `t(` | template | templates call `$t()`, the `useI18n` `t` belongs in `<script>` |
| `v-t=` | template | the legacy directive, call `$t()` in the expression |
| `$t(` | script | scripts call `t()` from `useI18n()` |
| `<i18n>` block | whole file | keys live in the translation files, not in a component |

It reads code only: comments are blanked first, so a comment mentioning one of these does not trip it. The failure names the file, the line and the fix.

Checked each rewrite: three in the running app, six through specs that assert the exact lines, one through a throwaway mount.


MilosPaunovic (commented):
Reviewed at `1d715ec8f1`. Scope matches https://github.com/kestra-io/kestra/issues/19384 exactly: all seven `OSS` `<i18n-t>` sites are gone and the guard implements all five rules from the issue.

Verified locally: the new guard spec, the four existing specs covering the touched components, and `vue-tsc --project tsconfig.test.json` (only the pre-existing stale `kestra-sdk` dist errors, none in the changed files). I also pointed `findI18nViolations` at `ui-ee/src` and at `ui/packages/*/src` - both clean. The rewrites themselves are correct: I checked each split against its `en.json` message, confirmed `escapeParameter` is not set in `setupI18n` so the `v-html` parameter route really does render `<code>`, and confirmed `readOnlyExcludedCount` and `selectionSummary` in `SourceSearchPreview.vue` are genuinely props, so `props.*` in the new computed is not reading `undefined`.

Four things I would like addressed, then some smaller notes.

### 1. Two `AGENTS.md` files still mandate the component this `PR` bans

- `AGENTS.md:465`: "**MANDATORY** ... `<i18n-t keypath="...">` with named slots when the string embeds markup or a component (e.g. a `<code>` fragment)"
- `ui/AGENTS.md:100`: "Use `<i18n-t>` for plurals and interpolation - never string-concatenate"

`kestra-ee/AGENTS.md:47` points at the first for the `EE` rules, so `EE` inherits it too. The guard's own failure message is `Fix the i18n usage, see ui/AGENTS.md`, which sends the reader to a document telling them to do the banned thing - and whose next clause forbids string concatenation, which is what `splitTranslation` is. The issue's rationale is "one component doing it its own way is how the next person learns the wrong pattern"; right now the docs are that component.

Both sentences need replacing with the two replacement patterns, ideally with a line on which to pick: `splitTranslation` when a Vue component goes in the gap (`router-link`, `KsId`), `$t()` plus `v-html` when it is static markup.

### 2. The rendered output now depends on the template staying on one physical line

Mounting the two shapes:

```
one line  : <span>before <code>X</code> after</span>
multi-line: <span>before  <code>X</code>  after</span>
```

Vue's `condense` whitespace handling only drops a whitespace-only node between two *elements*; between an interpolation and an element it collapses to a single space instead. So the `splitTranslation` sites are correct only because they are written as one long line. `<i18n-t>` was immune to this. There is no `max-len` in `eslint.config.js` or `.oxlintrc.json`, so tooling will not reformat them, but a person wrapping one of those lines for readability silently introduces double spaces, and (see the next point) almost nothing would fail. A one-line comment at each site, or a shared note next to `splitTranslation`, would cost nothing.

### 3. Only one of the rewrites is locked by a test

The description says "six through specs that assert the exact lines". The only assertion of that kind is `tests/unit/components/ai/copilot/CopilotMessage.spec.ts:60`:

```ts
expect(notice.text()).toBe("Removed flow good-morning from context.")
```

The others are `toContain(<the value>)`: `SourceSearchResults.spec.ts:207` asserts `toContain("company.platform")`, `CopilotContextChip.spec.ts` asserts `ids(w)` and `toContain("Flow:")`, `SourceSearchPreview.spec.ts` asserts `toContain("company.data")`. All of those pass whatever the surrounding sentence looks like, wherever the slot lands, and with any amount of whitespace around it. `TaskEditModal.vue`, `TriggerEnableDialog.vue` and `FlowsSearch.vue` have no rendered-text assertion at all.

Separately, `splitTranslation` is a new shared util with real edge cases - a missing placeholder appends the component at the end of the sentence, and a placeholder present twice makes `const [before = "", after = ""] = ...split()` drop the third part and lose text - with no unit test. Neither case is reachable from today's translation files, because `placeholderProblems` in `scripts/translations/translationRules.mjs` compares the ordered placeholder list against English, so this is about documenting intent rather than a live bug. Still, in a `PR` whose point is "a test keeps it out", the new util having no direct coverage stands out.

A handful of `toBe(...)` assertions on the full sentence would lock both this and point 2.

### 4. The guard reports the wrong line for a file with more than one `<script>` block

```ts
const offset = block === "source" ? found : source.indexOf(blocks[block]) + found
```

`blocks.script` is the script bodies joined with `"\n"`, so for a file with two `<script>` blocks that joined string does not occur in `source` and `indexOf` returns `-1`. Reproduced with a fixture: a `$t(` on line 10 of a file with `<script>` plus `<script setup>` is reported as `Two.vue:5`, pointing into the wrong block. Four files under `ui/src` have two script blocks today: `FlowFileEditorTab.vue`, `FileExplorer.vue`, `ThemeWindow.vue`, `ChildTableOfContents.vue`. Only the message is affected, not pass or fail, but the message is the whole value of the check. Searching each script body separately and carrying its own start offset fixes it.

### Smaller notes

- **Only the first violation per rule per file is reported.** `blocks[block].search(pattern)` returns one index, so a file reintroducing three `<i18n-t>` blocks reports one; you fix it, re-run, get the next. Confirmed on a fixture holding `<i18n-t>` and `<I18nT>` in the same template: one line out. `matchAll` would list them all in one run.
- **Scope is `src` only.** `ui/packages/design-system/src` and `ui/packages/topology/src` hold `.vue` files and use `vue-i18n` through `registerDesignSystemI18n`, and `test:lint` already runs `oxlint src packages tests`. I ran the guard over both: zero violations, so extending `SRC` to cover them is free today and closes the obvious way back in.
- **Worth considering: this is a lint rule wearing a unit test.** `ui/lint-rules/` already hosts custom `oxlint` `jsPlugins` wired through `.oxlintrc.json`. A rule there would give per-occurrence reporting with real line and column, feedback in the editor, and would let `ui-ee` use its own `.oxlintrc.json` instead of importing an `OSS` *test* file across repos. Not a blocker, the current form works and `CI` proves it, but the infrastructure is already there.

Merge order in the description is right, `OSS` first. I checked the `kestra-io/actions` side too: `kestra-ee-frontend-tests.yml` uses the `kestra-ee-checkout` composite, which is a full `actions/checkout` of `kestra-io/kestra` rather than a sparse one, so no companion `actions` change is needed here.

MilosPaunovic (approved):
Re-reviewed at `fcf798a086`. Ran the new guard spec (both cases), `splitTranslation.spec.ts`, the two new component specs and the four tightened ones locally: 65 tests, all passing. I also swept `ui/src` and `ui/packages/*/src` for a root `<template>` tag the segment regex would not match - the only file without a bare `<template>` is `ChildTableOfContents.vue`, a render-function component with no template block at all, so nothing is silently unscanned.

Approving. Four small things left, none of them blocking:

### 1. `FlowsSearch.vue` is the one rewrite with no assertion

`summaryCross` is the most involved of the three `v-html` sites - three nested pluralized `t()` calls through `_escape` - and `ui/tests` has no hit for `summary_cross`, `source-search__summary` or `summaryCross`. Every other one of the ten now pins its sentence.

### 2. The translations used-key scanner still advertises `<i18n-t>`

`scripts/translations/usageRules.mjs:40` documents the `KEYPATH` extractor as `<i18n-t keypath="...">`, and `usageRules.spec.js:16` and `:41` assert it works. With the guard in place that rule can no longer fire on `develop` sources, so a reader landing there would reasonably conclude `<i18n-t>` is still a supported pattern. It is genuinely still needed on `releases/v2.0.x` and `releases/v1.3.x`, where `<i18n-t>` still exists and the scripts run per branch, so this is a one-line comment saying so rather than a deletion.

### 3. `i18nGuard.spec.ts` leaks its temp directory when the guard throws

`mkdtempSync`, then `findI18nViolations(dir)`, then `rmSync(dir)` with no `try`/`finally`, so a throw inside the guard leaves the directory behind in `os.tmpdir()`. Minor, but this tree has explicit test-hygiene lint rules in `ui/lint-rules/` for exactly this shape.

### 4. The fixture assertion drops the rule text

`offenders.map((offender) => offender.split(" ")[0])` pins the three lines but not which rule produced each, and the expected order `["Two.vue:3", "Two.vue:4", "Two.vue:13"]` follows the `RULES` array order, so reordering `RULES` fails the test for no behavioural reason. Asserting the full strings pins both.

Issue #19266, opened 2026-09-10
[EPIC] Replace explicit `any` with real types across the frontend


Piyush-r-bhaskar:
oxlint's `typescript/no-explicit-any` rule reports 1803 explicit `any` in `ui/`, across 392 files. Each one is a spot where the compiler stops checking, and most of them cover data that already has a type in `@kestra-io/kestra-sdk`.

The count per file is committed in `ui/scripts/explicit-any/baseline.json`. Run `npm run check:ts-any` from `ui/` to check it. On a clean checkout it prints one line and exits 0:

```
No change against the baseline: 1803 explicit any in 392 files.
```

That is the confirmation: oxlint just recounted every file and the result matches what is committed.

If the code has drifted from the file, it prints the files that moved and exits 1. A new `any`:

```
New `any` in 1 file(s). Type it, or if it is genuinely unavoidable say why in the PR and run `npm run check:ts-any -- --write`.
  src/utils/anyProbe.ts: 0 -> 1
```

Or after someone removed some:

```
1 file(s) improved. Lock it in with `npm run check:ts-any -- --write` so the baseline keeps meaning something.
  src/stores/executions.ts: 55 -> 52
```

The same check runs in CI, so the number can only go down. Each sub-issue below takes one hotspot file or one feature area and closes when its files hit zero.

**Where the types come from**

1. `@kestra-io/kestra-sdk`, generated from the OpenAPI spec, for anything that comes from the backend. Most of these do.
2. The library's own types for anything that comes from a dependency: `vue-router`, `element-plus`, `monaco-editor`, `echarts`, the DOM. A package that ships no types gets a `declare module` shim next to `src/material-icons.d.ts`, not an `any`.
3. An interface the app already has.
4. A new interface, only when none of the above fits, named for the domain.

For the design system, the packages and the tests, which never touch the API, skip step 1. `unknown` narrowed once at the boundary is fine; `unknown` that every reader casts back is not.

When the SDK type and the code disagree, the SDK is right: fix the code or narrow with a guard, never re-add a cast.

**What the typing will surface**

Once a file is typed, the compiler starts checking code it never checked before, and it will flag places that read a field that might be missing or pass the wrong shape, both in that file and in files that use it. Fix every one of them, wherever it lives, so the PR leaves the app working end to end. A typed file with broken callers is a regression, not progress.

**When you open a PR for a sub-issue**

[...]

PR #19386, opened 2026-09-11
refactor(core): replace explicit any in the override directory

### LINKED ISSUE #19289: Override layer: replace explicit `any` with real types
41 explicit `any` across 9 files in `src/override`, from `ui/scripts/explicit-any/baseline.json`:

| any | file |
| --- | --- |
| 15 | `stores/auth.ts` |
| 10 | `services/flowAutoCompletionProvider.ts` |
| 8 | `components/flows/blueprints/BlueprintDetail.vue` |
| 2 | `components/namespaces/Namespaces.vue` |
| 2 | `components/useLeftMenu.ts` |
| 1 | `components/flows/Actions.vue` |
| 1 | `components/flows/blueprints/Blueprints.vue` |
| 1 | `components/namespaces/Actions.vue` |
| 1 | `stores/misc.ts` |

Fayupable:
Closes https://github.com/kestra-io/kestra/issues/19289.

### ✨ Description

Replaces all 40 explicit `any` types in the `ui/src/override` directory with real types, as part of the `[EPIC] Replace explicit any with real types across the frontend` (#19266) initiative.

- `stores/auth.ts` (15): the OSS no-op permission stubs now type their `permission`/`action`/`namespace` parameters as `string`, matching how `resource.ts`/`action.ts` are already typed elsewhere in the codebase (confirmed against the merged `fix(triggers)` PR #19230, which uses the same `string` convention for action parameters).
- `services/flowAutoCompletionProvider.ts` (10): introduced a `ParsedFlow` interface for the parsed-YAML shape passed around the autocompletion methods, typed the YAML map helpers against `YAMLMap` from the `yaml` package, and dropped two casts (`schema as any`, `.catch() as any`) that were unnecessary — the underlying types (`JSONSchema`, `PluginComponent`) already covered the accessed fields.
- `components/useLeftMenu.ts` (2): menu icon `element`/`class` typed as `Component`/`string`.
- `stores/misc.ts` (1): `configs` typed with the generated SDK type `MiscControllerConfiguration` instead of `Record<string, any>`.
- `components/flows/blueprints/BlueprintDetail.vue` (8): reuses `BlueprintType`/`BlueprintKind`/`BlueprintTag` from the blueprints store instead of casting to `any` (exported `BlueprintKind`, which existed but wasn't exported).
- `components/flows/Actions.vue` and `components/namespaces/Actions.vue` (1 each): `onSelectDashboard` value typed as `string`.
- `components/flows/blueprints/Blueprints.vue` (1): the `loaded` event never actually carries a payload (the underlying `BlueprintsBrowser` emits it with no argument), so the emit signature was corrected to `[]` instead of forwarding an untyped `$event`.
- `components/namespaces/Namespaces.vue` (1 real fix + 1 dead-code removal): removed a `:props="{class: 'tree'} as any"` binding on `KsTree` that never matched the component's actual `props` shape and had no corresponding CSS rule (confirmed dead), and typed the namespace-hierarchy builder's intermediate map as `Record<string, Node>` instead of lying about it being `Node[]`.

Every change was verified against real types read from the SDK / design-system / dependency sources (not guessed), and cross-checked against how the epic's already-merged PRs (#19230, #19227) typed equivalent patterns.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit`) — 256 files / 2281 tests passed
- [ ] Translations are complete if `en.json` changed — not applicable, no translation file touched
- [ ] Screenshots or video recordings attached — not applicable, this is a types-only change with no UI/behavior difference

### 📝 Additional Notes

- No runtime behavior changes anywhere in this PR — every fix is a type annotation, a cast replaced by a narrower cast/type, or the removal of confirmed dead code (the `KsTree` `:props` binding, the unused `Blueprints.vue` event payload).
- `stores/blueprints.ts` has one line changed outside the issue's file list: `BlueprintKind` was made `export`ed (it existed already, just wasn't exported) so `BlueprintDetail.vue` could reuse it instead of duplicating the same union type.
- `npm run check:types` (the stricter `--build` project-references check) caught 5 real type errors in `flowAutoCompletionProvider.ts` that plain `vue-tsc --noEmit` had missed — all fixed and verified (undefined-filtering on mapped arrays, an explicit string cast where a truthy-narrowed `unknown` collapsed to `{}`, and a `this`-context cast for a Pinia store method whose wrapped/unwrapped `Ref` types don't structurally match even though they're the same object at runtime).
- `check:types` on this branch also reports 4 pre-existing errors (`FilePreview.vue`, `FilePreviewForm.vue`, `TaskRunActions.vue`, `Secrets.vue`) that are unrelated to this PR — confirmed via `git diff develop...fix/override-any-types --stat` on those files, which shows zero diff. They already exist on `develop` and are out of scope here.


Fayupable:
Saw the CI failure. Looking into it now, will find the issues and push a fix shortly.

Fayupable:
Re: the `check:types` failure on `TaskRunActions.vue(175,63)` — I looked into where it comes from: this file uses the generic `useRoute()`, whose `params` type always resolves every key to `string | string[]` (vue-router's default typing when typed routes aren't used, regardless of the actual route pattern). That value is passed straight into `authStore.user?.isAllowed(resource.FLOW, action.VIEW, route.params.namespace)`, whose third parameter expects a plain `string | undefined`.

I confirmed this is pre-existing on `develop` and unrelated to this branch (checked out `develop` with none of this PR's commits, error is already there).

I wasn't sure of the right fix here — whether `namespace` should be normalized to a single string before this call (e.g. `Array.isArray(...) ? ...[0] : ...`), whether this component should use a typed route instead of generic `useRoute()`, or something else entirely — so I didn't want to guess and touch code outside this issue's scope. Happy to fix it here if someone can point me at the intended approach.

ui/src/override/stores/auth.ts
```diff
@@ -1,35 +1,35 @@
...
 
 
-    hasAnyAction(_permission: any, _action: any, _namespace?: any) {
+    hasAnyAction(_permission: string, _action: string, _namespace?: string) {
         return true
     }
 
 
-    isAllowed(_permission: any, _action: any, _namespace: any) {
+    isAllowed(_permission: string, _action: string, _namespace?: string) {
```
Piyush-r-bhaskar, line 14: This narrowing is the right call, it matches EE's `Me` exactly.

It does break one caller though: `TaskRunActions.vue:175` passes `route.params.namespace`, which is `string | string[]`. That's the CI failure. Fixing the call site is the way to go, not reverting this.

ui/src/override/services/flowAutoCompletionProvider.ts
```diff
@@ -370,7 +382,8 @@ export class FlowAutoCompletion extends YamlAutoCompletion {
                 if (namespace === undefined) {
                     return Promise.resolve([])
                 }
-                return Array.from(new Set<string>((await (this.namespacesStore as any).usableSecrets(namespace)).map((secret: string) => QUOTE + secret + QUOTE)))
+                const namespacesStore = this.namespacesStore as unknown as ReturnType<typeof useBaseNamespacesStore>
```
Piyush-r-bhaskar: `as unknown as` is the strongest cast there is, and we don't need it here. The real cause is the `this:` type on `usableSecrets` in `useBaseNamespaces.ts:139`, which asks for the whole store when it only uses two methods.

I tried it: narrow that `this` to `{loadInheritedSecrets; listSecrets}` and call `this.namespacesStore.usableSecrets(...)` directly. Compiles clean, no cast, and the extra import goes away too. That import is worth losing anyway, `override/` shouldn't reach into the OSS base store since EE swaps it.

ui/src/override/services/flowAutoCompletionProvider.ts
```diff
@@ -105,9 +117,9 @@ export class FlowAutoCompletion extends YamlAutoCompletion {
         }
     }
 
-    private tasks(source: string): any[] {
+    private tasks(source: string): YAMLMap[] {
         const tasksFromTasksProp = YAML_UTILS.extractFieldFromMaps(source, "tasks")
-            .flatMap(allTasks => allTasks.tasks)
+            .flatMap(allTasks => allTasks.tasks as YAMLMap[] ?? [])
```
Piyush-r-bhaskar: These two contradict each other: the cast says it's definitely an array, the `??` says it might be null. `extractFieldFromMaps` doesn't check the field value, so the `??` is the honest half.

ui/src/override/components/namespaces/Namespaces.vue
```diff
@@ -153,17 +152,17 @@
                         description: isLeaf ? item.description : undefined,
                         children: [],
                     }
-                currentLevel = currentLevel[label].children
+                currentLevel = currentLevel[label].children as unknown as Record<string, Node>
```
Piyush-r-bhaskar: Two `any` became two `as unknown as`, which is a step back, a double cast hides more than `any` did.

The actual problem is that `children` is a keyed map while building and an array afterwards, and one `Node` type can't be both. A small separate type for the accumulator with `children: Record<string, ...>`, turned into `Node[]` in `build()`, drops both casts.

ui/src/override/components/namespaces/Namespaces.vue
```diff
@@ -153,17 +152,17 @@
...
 
-        const build = (nodes: Node[]): Node[] => {
+        const build = (nodes: Record<string, Node>): Node[] => {
             return Object.values(nodes).map((node) => {
                 const result: Node = {
                     id: node.id,
                     label: node.label,
                     description: node.description,
-                    children: node.children ? build(node.children) : undefined,
+                    children: node.children ? build(node.children as unknown as Record<string, Node>) : undefined,
```
Piyush-r-bhaskar: Same cast as above, the accumulator type would remove this one too.

Piyush-r-bhaskar (changes requested):
Good work overall, most of this is correct and a few bits are genuinely better than just silencing the compiler.

Blocker: the branch doesn't compile. `check:types` fails on `TaskRunActions.vue(175,63)`, `route.params.namespace` is `string | string[]` and this PR narrows `isAllowed` to `string | undefined`. It's the only call site passing route params. One line fixes it: `route.params.namespace as string` (like `Dashboard.vue:115`) or `?.toString()` (like `Revisions.vue:328`).

Could you also fix the description? It says those 4 errors already exist on develop and are out of scope. They aren't: `TaskRunActions.vue` is identical on develop and compiles there, and the other 3 are fixed inside this PR, which is the giveaway that they came from the `misc.ts` change here.

Things I checked and they're fine, no need to touch them: the auth signatures now match EE's `Me` exactly, the `KsTree` `:props` removal is safe (nothing styles `.tree`), and the `Blueprints` emit change is right since `BlueprintsBrowser` emits it with no payload.

Piyush-r-bhaskar (changes requested):
Hey @Fayupable 

Please fix the test.

Fayupable:
I will fix the problem as soon as I am available

Piyush-r-bhaskar (approved):
LGTM

PR #19403, opened 2026-09-11
refactor(core): colours come from `--ks-` tokens instead of hex

Piyush-r-bhaskar:
EE PR: https://github.com/kestra-io/kestra-ee/pull/11014

Issue: colours were set in feature code as hex, `rgb()`, `--el-*`, `--bs-*` and SCSS palette variables, so a surface was only correct in the theme its author happened to check. Nothing caught it.

Fix: every colour in `ui/src` and `packages/topology/src` now comes from a `--ks-*` token or a `Ks*` prop. Alongside the swaps: `State.colorClass()` is gone (a state's colour is read with `State.getStateColor()`), `FlowRun`'s `.onboarding-glow` and `logs.ts`'s `graphColors()` are deleted as unused, `LowCodeEditor`'s `:deep(.vue-flow__node.rounded-3)` block is deleted as dead, and `vue-nprogress.scss` goes with it.

Guard: `tests/unit/designSystem/colorGuard.ts` fails the build on a hex, `rgb()`/`hsl()`/`oklch()`/`lab()`/`lch()`/`hwb()`, `--el-*`, `--bs-*`, a SCSS colour variable or a bare keyword reaching a colour property, across `.vue`, `.scss`, `.css`, `.ts` and `.js` in both roots. It reports every offending line, and the only ways to mute one are `design-system-disable-next-line` or a balanced `-start`/`-end`, each with a reason; an unterminated `-start` is reported rather than silently muting the rest of the file. `KsThemePicker.test.ts` asserts every `--tp-*` resolves from the palette.

Evidence: guard tests and unit suite pass. The three overlays converted to `--ks-bg-scrim` were checked in light and dark against a running instance; the light value measures as a no-op over a white page, raised as a design-system question in the review thread rather than patched here.


flcarre (changes requested):
The direction is right and the mechanical parts are done well: `cssVar()` is used correctly in `ksTheme.ts` (it really does read `getComputedStyle`, so the token resolves), and the two places that pass a literal `var(--ks-...)` string into JS — `KsProgress`'s `color` prop in `FlowConcurrency.vue` and `inputStyle.boxShadow` in `InputText.vue` — both end up in an **inline style**, where `var()` resolves normally. Neither silently renders transparent. Stripping the `var(--ks-x, #hex)` fallbacks is the right call per rule 11, and `KsThemePicker` arrives with a story and a test.

What needs fixing is the half of the sweep where the replacement is hex-equivalent in light mode but not semantically the same token.

### Contrast / affordance regressions

**`SurveyDialog.vue` (~224, ~305, ~313)** — `#918BA9` -> `--ks-border-strong` for unchecked radio and checkbox outlines. That token is `$base-gray-cool-400` (#404559) in dark on a `--ks-bg-surface` of #20232d: **~1.6:1**. In light it is #d0d0da on white: **~1.3:1**. `#918BA9` cleared 3:1 in both. WCAG 1.4.11 wants 3:1 for a control boundary, and an unchecked control you cannot see is the whole point of the control. Reach for a text/icon-family token (`--ks-icon-inactive`, `--ks-text-secondary`) — the border family has no visible grey in dark.

**`MultiPanelTabs.vue` (~1063, ~1074)** — `rgba(30, 30, 30, 0.5)` -> `color-mix(in srgb, var(--ks-bg-active) 50%, transparent)`. `--ks-bg-active` is `#f7f7f8` in light, so the drop-zone scrim becomes a near-white veil over a light background and the target stops reading as a target. This was deliberately theme-independent. A scrim is `--ks-bg-scrim`, not a surface-state token.

**`ThemeWindow.vue` (41, 49)** — this one undoes its own feature. `--ks-Artwork-fill` and `--ks-shadow-element` are theme-reactive, but this file is the miniature inside `KsThemePicker`, whose new Storybook text says the previews keep their colours whatever theme the page is in, and whose new AGENTS.md row says the same. `#9869f7` equals `$base-primary-300` in dark and dark-2, so the swap looks lossless — under the light theme every preview's dot flips to #631bf3. The `--tp-*` block three files over already does the right thing with `palette.$base-*`; do the same here, since the component now lives inside the design system where palette variables are allowed.

### Left behind in a file this PR edits

**`ui/src/styles/app.scss` (~145, ~151)** — the `.kel-button.status-*` block is edited (the `html.dark` branch dropped, `--ks-bg-surface` -> `--ks-btn-secondary-bg-default`, both mapping exactly onto the old values — nice), but `border: 0.5px solid $value`, `.material-design-icon { color: $value }` and `.bg-#{$key} { background-color: $value }` still draw from the SCSS `design.$colors` map in feature code. That is rule 2 verbatim and rule 12 ("clean up the ones in the component you are already editing"). The new guard cannot see it because `bareWords()` strips `$vars` — which is itself worth knowing.

### Scope creep with a public-API break

**`state.ts`** — dropping `StateModel.color`/`.colorClass` is in scope. Dropping the exported `SUBMITTED`, `QUEUED`, `RETRYING`, `colorClass()` and `icon()` is not: those are part of `export * as State`, i.e. the published surface of `@kestra-io/design-system`. I grepped OSS and EE and found no consumer, so it is safe *today*, but it is a breaking change riding in a colour refactor, and keeping `CREATED`/`RESTARTED` while deleting `SUBMITTED` is arbitrary. Either split it out or say so in the description — and it locks this to kestra-ee#11014 landing together.

### Guard and documentation

- `colorGuard` scans **only `ui/src`**. `ui/packages/design-system` and `ui/packages/topology` (both touched here) and the EE `ui/` are unguarded, and it matches hex/rgb/keyword only — not `--el-*`, `--bs-*` or SCSS `$colour`, which rule 2 also bans. Worth stating that limit in a comment or widening it.
- `ui/AGENTS.md:434` still advertises `State` as exposing "execution state constants, **icons**, and **colors**" after this PR removes exactly those.
- The PR introduces `design-system-disable` / `design-system-disable-next-line` and an automated check, and documents **neither**; §284 still presents the manual grep as the colour gate. A new escape hatch nobody can find is one everyone re-invents.
- The whole-file `design-system-disable` on `CopilotMarkAnimation.vue` is broader than its justification; per-line disables keep the rest of the file covered.

### Smaller notes

- `DocIdDisplay.vue:46` — `color: var(--ks-bg-elevated)` on a `--ks-text-link` fill is ~2.1:1 in dark. Text on a primary fill is `--ks-btn-primary-text`.
- `--ks-border-focus` is doing duty as a generic brand purple in three places (`KsEditor.vue:329`, `ContextDocsSearch.vue:187`, `SupportLinks.vue:79`). A focus token standing in for selection, hover and decoration is what makes the next rebrand wrong in three different ways.
- `MultiPanelTabs.vue:890` — a `--ks-text-primary` scrim inverts black->white between themes; `--ks-bg-scrim` is the family, even if a `--ks-bg-scrim-subtle` needs adding for a 10% veil.
- `ksTheme.ts:115` — correct mechanism, but `--ks-shadow-elevated` is 8% alpha in light versus the old `rgba(0,0,0,0.3)`, so chart hover emphasis mostly vanishes there.
- `KsEditor.vue:336` — `--ks-editor-pabble` is the right token, but it is misspelled in all three theme files and resolves to `$base-green-200` under dark-2, so the pebble goes green. Pre-existing; worth a follow-up issue.
- `BigChildCards.vue:63` drops a gradient and `border-image-source` for a flat `--ks-bg-surface`. That is a redesign, not a migration, and §287 asks for light **and** dark screenshots.
- `FlowRun.vue`'s `.onboarding-glow` and `logs.ts`'s `graphColors()` are both genuinely unreferenced (checked OSS and EE), so deleting them is fine — just call it out rather than letting it hide in a 39-file diff.

CI is green, including the new guard, so none of this is caught by tooling; all of it needs eyes on light and dark.

Piyush-r-bhaskar:
Thanks, this was a genuinely useful review. All of it is addressed in 77a63c0 except one item where the measurement came out the other way round, detail below.

I measured every pairing in the running app rather than from the theme files, by resolving the tokens through `getComputedStyle` in all three themes and compositing the translucent ones over the real page background.

### Contrast / affordance

**`SurveyDialog.vue` 224 / 276 / 313** (the third one is at 276, not ~305) is now `--ks-icon-muted`. It was the only candidate in the text/icon families that clears 3:1 in all three themes:

| | light | dark | dark-2 |
|---|---|---|---|
| `--ks-icon-muted` | 3.93:1 | 5.44:1 | 7.71:1 |
| `--ks-border-strong` (before) | 1.53:1 | 1.65:1 | 1.74:1 |

`--ks-icon-inactive` was the other suggestion but it is `$base-gray-neutral-300` in light, 2.0:1 on white, so it fails the same way.

**`MultiPanelTabs.vue`** all four sites are on the scrim family now. The percentages are picked so the composited alpha lands on the original dark-theme value: the editor overlay is `--ks-bg-scrim` at 15% and 40% (was 0.1 and 0.3), the drop zone at 65% and then the bare token (was 0.5 and 0.8). No `--ks-bg-scrim-subtle` needed as a result.

**`ThemeWindow.vue`** now uses `palette.$base-primary-300` for the dot and a fixed `rgba(palette.$base-gray-neutral-950, 0.1)` drop shadow, following the `--tp-*` block. Screenshot below shows the previews holding their own colours with the page in light and in dark.

**`DocIdDisplay.vue:46` I left as `--ks-bg-elevated`, because the numbers run the other way.** `--ks-text-link` is `$base-primary-200` in dark, so the fill is light lavender `rgb(205,182,251)` and white text on it is the low-contrast case:

| | light | dark | dark-2 |
|---|---|---|---|
| `--ks-bg-elevated` (current) | 7.04:1 | 9.02:1 | 9.47:1 |
| `--ks-btn-primary-text` | 7.04:1 | 1.80:1 | 1.80:1 |

Happy to switch if you were measuring something I am not seeing.

### Left behind in edited files

The `@each $key, $value in design.$colors` block in `app.scss` is gone rather than migrated. It generated `.bg-primary`, `.bg-success`, … and `.kel-button.status-*`, and nothing in OSS or EE references any of them, so the block only produced dead selectors. The `design` namespace import went with it. The `:root` block below it no longer reads `$box-shadow` either: the three `--kel-box-shadow*` variables are composed from `--ks-shadow-*` now.

### Public API

`SUBMITTED`, `QUEUED`, `RETRYING` and `icon()` are restored. You were right that dropping them was arbitrary and unrelated to colour.

`colorClass()` stays deleted, deliberately. It returned `"blue-500"`-style names whose only consumer was the `design.$colors` map removed above, so restoring it would mean restoring a `colorClass` field of dead class names on all sixteen states. That is the one intentional public-API removal in this PR and it is now called out here rather than hidden in the diff.

### Guard and documentation

The guard reads `--el-*`, `--bs-*`, `--kel-*` and SCSS `$variables` now, scoped to colour properties so `font-size: var(--kel-font-size-small)` stays quiet. An SCSS variable that the file declares as `$x: var(--ks-…)` is treated as the token it aliases. It found seven more real violations, all fixed in this commit:

- `LowCodeEditor.vue` `--bs-gray-900` on a node border
- `Restart.vue` two `--kel-text-color-regular`
- `GlobalSearch.vue` and `TourOverlay.vue` `--kel-overlay-color-lighter` on a scrim
- `vue-nprogress.scss` `$indigo`, which was declared `"" !default` and never assigned, so the loading bar had no colour at all
- `app.scss` `$box-shadow` / `-sm` / `-lg`

Scope is still `ui/src` and that is now stated in the file: `packages/design-system` composes the tokens from the raw palette, so scanning it would report the palette itself.

`design-system-disable` is documented in `ui/AGENTS.md`, replacing the manual grep as the colour gate, and there is a third form. `design-system-disable-start` / `-end` covers a run of lines, which is what the copilot mark needed: the whole-file opt-out is gone and the markers now wrap only the SVG artwork, so the style block stays guarded. `icons.ts` uses two per-line disables on the two artwork strings.

`ui/AGENTS.md:434` no longer advertises colours on `State`.

### Smaller notes

`--ks-border-focus` is down to one site and it is a real focus indicator: `ContextDocsSearch`'s `.selected` is the arrow-key cursor that Enter activates, so the rail stays. Its tinted background is `--ks-bg-active` now, which is the surface state it actually wanted. `SupportLinks` hover is `--ks-border-strong`, a step up from its resting `--ks-border-default` rather than a brand accent. `KsEditor`'s highlight glow is `--ks-shadow-elevated`.

`ksTheme.ts:115` passes the alpha through `cssVar`: `cssVar("--ks-shadow-elevated", 0.3)`, which strips the token's own alpha and applies 0.3, so the chart hover emphasis matches the old `rgba(0,0,0,0.3)` while the hue still comes from the token.

`BigChildCards.vue` keeps a gradient instead of flattening: `--ks-bg-elevated` to `--ks-bg-base`, and hover `--ks-bg-hover-elevated` to `--ks-bg-elevated`. The `border-image-source` is dropped, since a border gradient needs a `border-image-slice` this rule never set and the flat token border is what was actually painted.

`--ks-editor-pabble` I left alone. The misspelling is consistent across the three theme files so it resolves, but the dark-2 value really is `$base-green-200`. That is a token value bug for design rather than something to fix inside a migration, and it predates this PR.

`FlowRun.vue`'s `.onboarding-glow` and `logs.ts`'s `graphColors()` are deleted and unreferenced in both repos, as you found. Calling it out here explicitly.

Checks are green: 2127 unit tests, `check:types`, `check:ts-any` unchanged at 1644, and eslint clean on every touched file.

flcarre (changes requested):
Thanks for the second pass — most of round 1 is genuinely resolved, and the parts that were fiddly are right.

**Fixed:** `SurveyDialog` now clears 3:1 in both themes (`--ks-icon-muted` is #7e7e9b on white ~3.9:1, #9797a6 on #20232d ~5.4:1). `ThemeWindow` is theme-independent again via `palette.$base-primary-300` (= the old #9869f7) and a palette-based drop shadow, consistent with the `--tp-*` block. The `design.$colors` `@each` in `app.scss` is gone and I found no consumer of `.bg-{primary,success,...}` or `.kel-button.status-*` in OSS or EE — and as a bonus, `--kel-box-shadow: $box-shadow` was a broken Sass custom-property interpolation (SassScript is not evaluated in a custom property without `#{}`), so those shadows were dead and now actually render; worth a screenshot. `SUBMITTED`/`QUEUED`/`RETRYING`/`icon()` are back. The disable hatches are documented and narrowed to `-start`/`-end` and `-next-line`, and the guard now catches `--el-*`/`--kel-*`/`--bs-*` and SCSS `$vars`. `index.ts` is purely additive, so exporting `KsThemePicker` is not a public-API break. And `cssVar("--ks-shadow-elevated", 0.3)` is safe: the token is an 8-digit hex (`#3a1d6d14` / `#0d061826`), so the hex parser yields `rgba(58, 29, 109, 0.3)` rather than `rgba(NaN, ...)` — worth knowing the next time `cssVar`'s second argument is reached for, since it would not have been safe for an `rgba()`-valued token.

Three things still need changing.

### `--ks-bg-scrim` makes the drop zones disappear

I pointed you at the scrim family and that was right for the modal overlays, but not for these, and the token's actual values make it worse than what it replaced.

`--ks-bg-scrim` is `rgba(#ffffff, 0.7)` in light and `rgba(#14181f, 0.8)` in dark — it is white in light and near-black in dark, i.e. it always matches the background.

- **`MultiPanelTabs.vue:890`** `.editor-content-overlay` — `color-mix(... var(--ks-bg-scrim) 15%, transparent)` gives ~10.5% white over a white/near-white panel in light, and ~12% #14181f over a dark panel in dark. Both are imperceptible. The previous `--ks-text-primary` at least inverted (black in light, white in dark) and so read in both.
- **`MultiPanelTabs.vue:1063, 1074`** `.new-panel-drop-zone` — 65% scrim (100% on hover) is a ~45-70% white plate over a light background, so this is still the round-1 regression, just arrived at from the other side. Only the 2px dashed border survives.

These two are *highlights*, not scrims. Use something that contrasts with the surface in both themes — `color-mix(in srgb, var(--ks-border-focus) N%, transparent)` reads as a drop affordance and matches the focus border you already put on `:hover` — or add a `--ks-bg-dropzone` to the design system.

The four genuine overlays (`GlobalSearch.vue:334`, `TourOverlay.vue:827`, `DocsLayout.vue:241`, and the existing `BlockCommandMenu`/`BlockTaskPicker`) are correct with `--ks-bg-scrim`, but they do flip from a dark dim to a 70% white veil in light mode. That is a real user-visible change across four surfaces, and the PR still has no light/dark screenshots.

### `--ks-editor-pabble` regresses Dark 2.0

**`KsEditor.vue:336`** — I called the misspelling pre-existing and a follow-up, which undersold it. The token is `$base-orange-500` in light, `#ffca16` in dark, and **`$base-green-200` (#80fac5) in `ks-theme-dark-2.scss`**. The old rule was `color: #977100` with an `html.dark &` branch for `#ffca16`, and `html.dark &` does not match `html.dark-2` — so Dark 2.0 users got #977100 before and get bright green after. Dark 2.0 is a shipped option in the picker this PR moves into the design system. Either get the token fixed (it is generated from Figma, so a designer has to touch `ks-theme-dark-2.scss`) before adopting it here, or leave this one hardcoded behind a `design-system-disable-next-line` with a link to that follow-up.

### `LowCodeEditor` activates a rule that was never live

**`LowCodeEditor.vue:1023`** — `--bs-gray-900` is defined nowhere: `bootstrap.scss` sets `$prefix: 'kbs-'`, so the variables are `--kbs-*` and `var(--bs-gray-900)` resolved to nothing, making the whole `border-color: ... !important` declaration invalid and the override inert. Swapping in `--ks-border-strong` turns it on for the first time, and #d0d0da in light is too weak for a hover highlight whose stated job (in the comment right above) is to *replace* the success/failure border. Pick a hover/focus-family token and look at it in VueFlow before merging — this is the one change in the PR whose behaviour is new rather than recoloured.

### Guard coverage vs. what AGENTS.md now promises

`ui/AGENTS.md:280` says colour is "gated automatically" and drops the hex/rgb/`--el-`/`--bs-` terms from the manual grep, leaving only `:deep(` and px. But `colorGuard.spec.ts:8` resolves `SRC` to `ui/src` and that is the only root ever scanned. The docstring's justification covers `ui/packages/design-system` (it composes tokens from the palette), but not `ui/packages/topology` — a package this PR edits, and which the deleted grep used to cover. Net effect: the gate got narrower for the packages while the doc reads as though it got wider. Either add `ui/packages/topology` as a second root in the spec, or keep the colour clause in the manual grep and scope it to `ui/packages/**`.

### Smaller

- **`state.ts:189`** — `colorClass()` and `StateModel.color`/`.colorClass` are still removed from the `export * as State` surface. Still safe (no consumer in OSS or EE), still unmentioned: one line in the description covering this, the `FlowRun` `.onboarding-glow` deletion and `logs.ts` `graphColors()` would close it.
- **`vue-nprogress.scss:5`** — nothing imports this file. `vendor.scss` pulls `nprogress/nprogress.css` and the design system's `index.scss` already styles `#nprogress` with `palette.$base-primary-500`. `$indigo` was `""` so the rules were invalid anyway. Delete the file rather than re-colour it; if it were ever wired up, `--ks-text-link` is `$base-primary-200` in dark and would fight the design system's `$base-primary-500`.
- **`KsThemePicker.test.ts`** — the story says "the previews keep their colours whatever theme the page itself is in", which is exactly what regressed last round, and no test asserts it. Even a shallow assertion that `.theme-picker__window--light` resolves `--tp-sidebar` from the palette and not from a `--ks-*` token would stop the next person re-tokenising it.
- **`ContextDocsSearch.vue:188`** — the background is fixed, but `border-left: 3px solid var(--ks-border-focus)` on `.selected` is the last of the three focus-token-as-brand-purple spots.
- **`Labels.vue`** and **`BigChildCards.vue`** are both correct token-wise but are visual redesigns (label chips change colour and text colour in both themes; the card goes from theme-independent dark to theme-following). Fine by me, just screenshot them.
- **`DocIdDisplay.vue:46`** — withdrawing my round-1 note: dark `--ks-text-link` is #cdb6fb and `--ks-bg-elevated` is #1e202a, so contrast is fine. `--ks-btn-primary-text` is still the semantically right token for text on a primary fill.
- `SupportLinks.vue` has no trailing newline, and `FlowConcurrency.vue` loses an unrelated comment about stale concurrency limits.

CI is green on all of this, so as before it needs eyes on light, dark **and** dark-2.

flcarre (changes requested):
Third pass. Most of round 2 is properly fixed, and three of them better than I asked for.

**Fixed:** all four `MultiPanelTabs` sites are off the scrim family and onto `color-mix(… var(--ks-border-focus) N%, transparent)` — 10/30% on the editor overlay, 15/35% on the new-panel zone, with the focus border on hover — which reads in light, dark and dark-2. `--ks-editor-pabble` is reverted to the develop bytes with a comment. `vue-nprogress.scss` is gone. `SupportLinks` gets its newline, `FlowConcurrency` its comment, and `AGENTS.md:460` no longer advertises colours on `State`. In topology, `.text-danger` finally gets a dark value via `--ks-text-error`, and `ClusterNode`'s `color: white` was dead anyway because `badgeStyle` sets `color` inline — right call to drop it rather than tokenise it.

Three are better than what I proposed. `LowCodeEditor`'s `:deep(.vue-flow__node.rounded-3)` block is deleted rather than re-tokenised, and that is the correct read: `rounded-3` appears nowhere in the topology source, and `topology.scss:113` on develop already carries the live `.topology-node-hovered` outline/border pair. The whole block was dead, not just the `--bs-gray-900` line — I had that wrong. The guard picked up `packages/topology/src` as a second root *and* `hsl/hwb/lab/lch/oklch`, `$var: #hex`, script colour keys, per-line reporting, and an unterminated-`-start` error, with the whole-file opt-out removed. And `KsThemePicker.test.ts` asserts every `--tp-*` resolves from `#{palette.$…}`, which is exactly the regression that bit us in round 1.

I also re-checked `cssVar("--ks-shadow-elevated", 0.3)`: `#3a1d6d14` / `#0d061826` / `#000000a6`, all 8-digit hex, so it is safe in all three themes.

Two of the fixes over-corrected, though, and both are one-liners.

### `DocIdDisplay.vue:46` is now 1.80:1 in both dark themes

`color` moved to `--ks-btn-primary-text` (`#ffffff` everywhere) but `background-color` is still `var(--ks-text-link)`, which is `$base-primary-200` = `#cdb6fb` in dark and dark-2. White on that is **1.80:1** — the exact number in your own round-1 table, which is why you kept `--ks-bg-elevated`, and I did not mean to talk you out of it.

The defect is the fill, not the text: `--ks-text-link` is a text token being used as a background. `:hover` is already `--ks-btn-primary-bg-hover`, so make the resting state `--ks-btn-primary-bg-default` (`#631bf3` in all three themes). White on that is 7.0:1 everywhere and the resting/hover pair finally comes from one family.

### `ContextDocsSearch.vue:187` — my nit was wrong, please put it back

`.selected` is the arrow-key cursor: `selectedIndex`, `@keydown.up/down` at lines 8-10, and Enter clicks it at 77-79. Your round-1 defence was right and I should have accepted it. `--ks-border-strong` is `#d0d0da` in light, so a 3px pale-grey rail on a `--ks-bg-active` row leaves no visible cursor. Revert to `--ks-border-focus`.

### Screenshots, third time

`GlobalSearch.vue:334` and `TourOverlay.vue:853` were `--kel-overlay-color-lighter` and `DocsLayout.vue:241` was `rgba(0, 0, 0, 0.6)`. All three are now `--ks-bg-scrim`, which is `rgba(#ffffff, 0.7)` in light — three modal overlays go from dimming the page to bleaching it. The token family is right, so if it reads badly that is a design-system issue worth raising rather than something to patch here, but nobody has looked. Same for `Labels.vue` and `BigChildCards.vue`. The round-1 comment says "Screenshot below" and has no image in it.

### Smaller

- `topology.scss:3` — the `design-system-disable-start` reason says the block composes the tokens "from the palette", but `$node-default: #7081b9` is not in `_color-palette.scss`. Either promote it to a palette entry or narrow the reason. The span is also 26 lines, which is the whole-file opt-out this PR just removed, arrived at from the other side.
- `KsEditor.vue:343` — the comment explains the hold but names no issue, and `packages/design-system` is unguarded, so nothing will ever surface this again. A `TODO` with the issue number costs one line.
- The description is still `EE PR: …`. The `State.colorClass()` removal, `FlowRun`'s `.onboarding-glow` and `logs.ts`'s `graphColors()` are covered in a comment, but the description is what survives into the commit log.

CI is green again, so as before none of this is caught by tooling.

Piyush-r-bhaskar:
Round 3 addressed in b3910f5.

**`DocIdDisplay.vue`** — you were right and I should not have moved off `--ks-bg-elevated` without re-checking the fill. The resting background is `--ks-btn-primary-bg-default` now, so it is `#631bf3` in all three themes and pairs with the `--ks-btn-primary-bg-hover` that was already on `:hover`.

**`ContextDocsSearch.vue`** — reverted to `--ks-border-focus`. Agreed, `.selected` is the arrow-key cursor, and a `#d0d0da` rail on `--ks-bg-active` is not a cursor.

**`topology.scss`** — the 26-line span is gone. I removed the `-start`/`-end` and ran the guard against the unmuted file: it reports exactly one line, `$node-default: #7081b9`, and nothing else in that block, since the rest is already `$base-*`. So it is now a single `design-system-disable-next-line`. I did not promote the colour to the palette because `_color-palette.scss` carries a generated-from-Figma header and is rewritten by `generate-palette.mjs`, so a hand-added entry would not survive. The reason now says that instead of claiming the block composes from the palette.

**The screenshots, with numbers.** Checked against a running instance at 1440x900, sampling the page pixels behind the open global-search modal:

| | page behind the modal |
|---|---|
| light, no overlay | `rgb(255,255,255)` |
| light, `--ks-bg-scrim` | `rgb(255,255,255)` |
| light, what develop rendered | `rgb(127,127,127)` |
| dark, no overlay | `rgb(32,35,45)` |
| dark, `--ks-bg-scrim` | `rgb(22,26,34)` |
| dark, what develop rendered | `rgb(16,17,22)` |

Dark is fine. Light is worse than "reads badly": `rgba(#ffffff, 0.7)` over a white page is a no-op, so the modal has no backdrop at all and the page behind it is untouched. That is not something my three conversions introduced in isolation, though. `BlockCommandMenu.vue:159` and `BlockTaskPicker.vue:183` already use `--ks-bg-scrim` on `develop` and already render with no backdrop in light today.

So this is the design-system issue you predicted, and the fix is one line in `ks-theme-light.scss`: a scrim has to darken, so the light value wants to be a dark translucent, `rgba($base-gray-cool-950, 0.5)` in the shape of the dark and dark-2 values, not white. That repairs the two surfaces already shipping with it as well as these three. I have not made that change here since it is design's call and it touches surfaces outside this PR, but I am happy to take it in this PR or a follow-up, whichever you prefer.

`Labels.vue` and `BigChildCards.vue` were checked in both themes on the same instance. The docs cards previously used a hardcoded dark gradient in **both** themes; they now follow the theme, so light mode gets light cards instead of dark navy ones on a white page, and dark mode is unchanged in character. Screenshots are local, say the word and I will attach them.

Piyush-r-bhaskar:
Rebased on develop (now 1af61c5). Three conflicts, all where #19534's z-index tokens landed on the same lines as a colour swap, so both sides are kept:

- `MultiPanelTabs.vue`, `DocsLayout.vue`, `ContextDocsSearch.vue`: develop's `--ks-z-*` token plus this branch's colour token.
- `GlobalSearch.vue`: the `z-index: 10000` this branch had added is now `var(--ks-z-overlay)`, to match what #19534 did everywhere else rather than reintroduce a raw value.
- `vue-nprogress.scss`: develop had re-tokenised its z-index, this branch deletes the file, so the deletion wins.

Guard tests, unit suite, types and lint all pass on the rebased tree.

flcarre (approved):
Fourth pass, and this one lands.

**Verified fixed.** `DocIdDisplay.vue:45` is on `--ks-btn-primary-bg-default`, which is `$base-primary-500` `#631bf3` in light, dark *and* dark-2, so white text is 7.0:1 everywhere and the resting/hover pair finally comes from one family. `ContextDocsSearch.vue:188` is back on `--ks-border-focus` (`$base-primary-500` in light and dark, `$base-primary-300` in dark-2 — visible in all three). `topology.scss` is a single `design-system-disable-next-line` instead of a 26-line span, and I checked the guard's behaviour rather than taking the claim: only `PATTERNS[1]` can fire in that block, because `DECLARATION`'s value capture is `[^;{}]+` and every other palette reference in the file sits inside a `#{…}` interpolation that truncates the value before the `$var` check, and the `$node-colors` map entries are not declarations at all. One line, correctly. `vue-nprogress.scss` was genuinely orphaned — nothing `@use`s `components/vue-nprogress`, `app.scss` imports only `components/markdown`, `vendor.scss` pulls `nprogress/nprogress.css`, and the design system's `index.scss:47` already styles `#nprogress`. The three rebase resolutions keep both sides as described.

### The scrim: you are right, and I am taking it

Your numbers hold. `--ks-bg-scrim` is `rgba($base-gray-cool-white, 0.7)` in light over a `--ks-bg-base` of `$base-gray-neutral-50` `#f7f7f8`, which composites to about `#fdfdfe`. Two corrections to the framing, neither of which changes the conclusion.

It is not "no backdrop". Your sample was one pixel of page *background*, which is white either way — but every piece of content behind the modal drops to 30% toward white. The light scrim bleaches rather than dims. It does separate the modal, just the wrong way round.

And the fix is not one line in `ks-theme-light.scss`. `generate-palette.mjs` writes `ks-theme-<mode>.scss` with the same `fs.writeFileSync` it uses for `_color-palette.scss`, and those files carry the same Figma header. So a hand-edited scrim would be overwritten for exactly the reason you correctly refused to hand-add `#7081b9` to the palette — the argument you made for `$node-default` applies to the token you then proposed to patch by hand.

Which settles the question you asked me. There is no token today that inverts correctly for a scrim: the `--ks-text-primary` trick we agreed on for the `MultiPanelTabs` highlights gives a white veil over a dark page, so it is wrong here. The only alternatives are three `design-system-disable-next-line`s carrying the old `rgba()`, or shipping the token and fixing it centrally. Five surfaces wrong in one place beats three escape hatches plus two already-broken surfaces, so keep your conversions. I will get `--ks-bg-scrim` corrected in Figma; open the issue and link it from the PR so the trail exists.

### `GlobalSearch.vue:337` — the z-index you added is dead

At the merge base `.search-overlay` has no `z-index` at all; the template sets it at runtime: `:style="{zIndex: topLayer}"` from `useTopLayer()` (imported at 76, called at 103). An inline style beats a scoped stylesheet rule, so `z-index: var(--ks-z-overlay)` never applies. It is also the wrong value to leave lying around: `ks-tokens.scss:47` says in as many words that the scale sits "below Element Plus, whose overlays start at 2000 and climb. To sit above one, use `useTopLayer()`" — which this component already does. If someone later drops `useTopLayer()` trusting this rule, the search overlay silently falls under every EP dialog. Delete the line; nothing in this PR needs it.

### `AGENTS.md` and the guard disagree about the hatch

Two small mismatches in the documentation this PR ships:

- §298 says the hatches exist for "artwork whose colours **are** the asset (a brand mark, a third-party logo) rather than a themed surface" and that "nothing else qualifies". `topology.scss:3` is a themed node colour, so the diff breaks its own rule on arrival. Widen the sentence — a colour the Figma-generated palette does not carry yet is a legitimate second case, and it is the one you are actually in.
- It says each hatch takes a reason. The guard does not require one: `DISABLE_NEXT_LINE` and `DISABLE_START` match the bare marker, and `colorGuard.spec.ts` pins that behaviour with a reasonless `/* design-system-disable-next-line */`. Either require the reason in the regex or stop promising it.

### Smaller

- **`GlobalSearch.vue:352`** — `0 8px 24px rgba(0,0,0,0.35)` becomes `--ks-shadow-elevated`, `#3a1d6d14`, 8% alpha in light. This is the one element that gets both conversions, so in light the modal is a white surface on a bleached white field with an 8% shadow, held by a 1px `--ks-border-default`. Worth a look once the token is fixed.
- **`KsEditor.vue:335`** — `.highlight-text`'s `0 19px 44px rgba(157, 29, 236, 0.31)` was a purple emphasis glow; `--ks-shadow-elevated` at 44px blur and 8% is nothing. `packages/design-system` is unguarded, so this one was voluntary: either keep the emphasis with a `color-mix` off `--ks-border-focus`, or drop the `box-shadow` and say the glow is gone.
- **`BigChildCards.vue:83`** — in light the hover gradient is `#f7f7f8 → #ffffff` and the resting gradient is `#ffffff → #f7f7f8`: the same two colours reversed, so only `border-color` carries the hover. Dark is fine (`cool-600 → cool-900`).
- **`topology.scss:3`** and **`KsEditor.vue:343`** still name no issue in their reasons. Third time; one line each.
- **`BasicSettings.vue`** now imports from `@kestra-io/design-system` on three separate lines (312, 335, 337) — fold `KsThemePicker` into the `KsButton` import.
- **`markdown.scss:57`** — `#8997bd` to `--ks-border-strong` is `$base-gray-neutral-200` in light, a distinctly paler blockquote rule than before. Fine if intended.

None of the above blocks. Fix the dead `z-index`, the two `AGENTS.md` sentences and the issue links on the way in.

PR #19404, opened 2026-09-11
ci(2.0): run E2E on release-branch pushes and drop the inert scheduling workflow

AcevedoR:
Companion PR to kestra-io/kestra-ee#11016 (same branch name), which does the same for EE.

Pushes to `releases/v2.0.x` ran no E2E at all — the only thing a push triggers on this branch is Main Workflow, which had no E2E job.

## `e2e-scheduling.yml` deleted

Both of its triggers are inert **on this branch**:

- `push: branches: [develop]` — a push to `develop` runs `develop`'s copy of the file, never this one.
- `schedule: cron` — GitHub only runs schedules from the default branch, so the daily cron never fires from here.

`workflow_dispatch` was the one live entry point, and it is preserved: Main Workflow is dispatchable on this branch and now carries the E2E job.

`develop` keeps its copy untouched — this deletion only affects `releases/v2.0.x`.

## E2E added to Main Workflow

Gated to `push` and `workflow_dispatch`, under the same `skip-test` input as the backend and frontend jobs.

`e2e-tests` is added to the `end` job's `needs`, so a failure turns the run red and reaches Slack like the other suites. Revert that one line if E2E turns out too noisy to gate on. It is also added to `otel-export-trace`'s `needs` so the trace export waits for it rather than closing early.

**Pre-release is untouched.** Nothing in `.github/workflows/` calls `main-build.yml` via `workflow_call`, so nothing added here reaches a tag build.

## PR E2E already reuses the prebuilt exe

`pull-request.yml` on this branch already consumes the `exe` artifact from `Build Artifacts`, so no change was needed there. This PR is the push half only.

## Verification

- Java 25 throughout, matching every other job on the branch.
- The E2E fixtures this branch needs are present: `build-and-start-e2e-tests.sh` and `ui/tests/e2e/`.
- Both edited workflows parse as YAML and the job graph resolves.

The push half lands on the first merge to the branch; this PR's own checks do not exercise it.


fhussonnois (approved):
LGTM!

PR #19407, opened 2026-09-11
fix(triggers): recover stuck polling evaluation locks

rishi-jat:
## Problem

A polling trigger on the JDBC backend can stay locked forever when evaluation produces **no result**. Thrown exceptions already released the lock; the leak came from a missing result.

Two cases:

1. **Hanging evaluation.** `evaluate()` ran with no timeout. A blocking call that ignores `Thread.interrupt()` (for example a stuck `KafkaConsumer.poll()`) wedged the worker thread. No `WorkerTriggerResult` was emitted, so `TriggerState.locked` stayed `true` until a manual unlock.
2. **No dispatch.** If the worker job was never sent, taking the lock first left the trigger locked with nothing running. The 2.x scheduler already locks only after a successful dispatch; this change adds a polling-specific regression test for that path.

PR #17446 added `Timeout.withInterrupt()` but called `kill()` only **after** `evaluate()` returned. Interrupt alone does not unblock `poll()`, so the worker stayed occupied and the lock was never released.

## Fix

- Bound polling-trigger evaluation with `kestra.worker.polling-trigger-timeout` (default `10m`).
- When the timeout fires, a watchdog thread calls `kill(false)` immediately (`trigger.kill()` / `task.kill()` plus interrupt), matching `WorkerJobLifecycle`. That is what actually unblocks plugin I/O (e.g. `consumer.wakeup()`).
- `onTimeout` runs in `try/finally` so `kill(false)` still runs if the metric callback throws.
- Task timeouts share the same `callWithTimeout` path (replacing Failsafe). Timeout still reports `FAILED` + `TimeoutExceededException`, not `KILLED`.
- Realtime triggers are unchanged.

A hang that ignores interrupt **and** whose plugin `kill()` is a no-op still occupies the worker. That is the existing plugin contract.

## Tests

- `WorkerTriggerCallableTest`: interruptible hang, interrupt-ignoring hang, in-time success
- `WorkerTaskCallableTest`: timed-out task is `FAILED` with `TimeoutExceededException` and `task.kill()` is invoked
- `TriggerSchedulerTest.shouldNotLockPollingTriggerWhenWorkerJobIsNotDispatched`

## Config

```yaml
kestra:
  worker:
    polling-trigger-timeout: 10m
```

Lower this if faster hang recovery is desired. The default is generous so legitimate long evaluations are not cut.

FIxses #17445

### DIFF
--- a/worker/src/main/java/io/kestra/worker/processors/internals/AbstractWorkerCallable.java
+++ b/worker/src/main/java/io/kestra/worker/processors/internals/AbstractWorkerCallable.java
@@ -3,7 +3,11 @@ package io.kestra.worker.processors.internals;
 import java.time.Duration;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.slf4j.Logger;
 
@@ -13,8 +17,6 @@ import io.kestra.core.models.flows.State;
 import io.kestra.core.runners.RunContext;
 import io.kestra.core.utils.Exceptions;
 
-import dev.failsafe.Failsafe;
-import dev.failsafe.Timeout;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.StatusCode;
 import lombok.Getter;
@@ -25,6 +27,12 @@ import static io.kestra.core.models.flows.State.Type.*;
 
 @SuppressWarnings("this-escape")
 public abstract class AbstractWorkerCallable implements Callable<State.Type> {
+    private static final ScheduledExecutorService TIMEOUT_SCHEDULER = Executors.newSingleThreadScheduledExecutor(runnable -> {
+        Thread thread = new Thread(runnable, "worker-job-timeout");
+        thread.setDaemon(true);
+        return thread;
+    });
+
     /** The state to report once interrupted, or {@code null} if not interrupted (or interrupted without marking, e.g. on timeout). */
     volatile State.Type killedState = null;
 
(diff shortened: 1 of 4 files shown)


worker/src/main/java/io/kestra/worker/processors/internals/AbstractWorkerCallable.java
```diff
@@ -162,28 +167,36 @@ protected State.Type callWithTimeout(Duration timeout, Evaluation work, Runnable
...
-            .build();
+        AtomicBoolean timedOut = new AtomicBoolean(false);
+        ScheduledFuture<?> timeoutTask = TIMEOUT_SCHEDULER.schedule(() -> {
+            if (timedOut.compareAndSet(false, true)) {
+                try {
+                    if (onTimeout != null) {
+                        onTimeout.run();
+                    }
+                } finally {
+                    kill((State.Type) null);
```
rishi-jat: **[suggestion]** `TIMEOUT_SCHEDULER` is a single thread, and the timeout task calls `kill(null)`, which runs plugin `kill()` *before* `interrupt()` (`WorkerTaskCallable` / `AbstractWorkerTriggerCallable`). A plugin `kill()` that blocks or hangs (Docker API, remote HTTP with no deadline) stalls every other job timeout on this worker and also delays interrupt of the job that just timed out. Failsafe only interrupted on its scheduler; `kill()` ran later on the worker thread after `TimeoutExceededException`.

**Suggestion:** Keep the scheduled task to flipping `timedOut` and interrupting (or submitting `kill` to another executor). Do not run plugin `kill()` on the shared scheduler thread. If `kill(null)` stays on this thread, interrupt first so interruptible work still unblocks when `kill()` is slow.

fhussonnois:
Hey @rishi-jat,  Thanks for looking into this.

You found something real about Failsafe: in the sync path it only throws the timeout after the work returns. So the `kill(null)` from #17446 never runs while the evaluation is stuck. That part is correct and good to know.

But I don't think this fixes the issue.

The fix calls the plugin's `kill()` when the timeout hits. The Kafka trigger in the report has no `kill() `method, it uses the empty default from `WorkerJobLifecycle`. In all our plugin repos only a few implement that method, and most of them are `RealtimeTrigger`. So for this issue nothing changes.

The rest works the same as before. `work.run()` still runs on the worker thread, and we still wait for it to return before sending the result. If the plugin never returns, we are back to the same problem.

In addition, the tests can't catch that. `HangUntilKilledTrigger.eval()` loops on the same flag kill() sets, so it stops because the test made it stop. A real stuck `poll()` does not behave like that.

Also, your PR changes how `timeout`: works on tasks. `callWithTimeout` is shared with `WorkerTaskCallable`, so every task with a timeout now gets `task.kill()` at the deadline, while it is still running. Before, it was only interrupted. 

Then: `TIMEOUT_SCHEDULER` is a single static thread, and `kill()` runs on it. `kill()` is plugin code. If it blocks, no other timeout in the worker fires anymore. Failsafe's timer thread only submits to a pool, for that reason.

So I'm closing this PR.

PR #19410, opened 2026-09-11
feat(version): restyle the upgrade notice with a button CTA, left aligned

elevatebart:
Targets `feat/version-upgrade-notice`, not `develop`. It is a design change on top of https://github.com/kestra-io/kestra/pull/19320, so merge that one first (or merge this into it).

### 🔗 Related Issue

Follow-up on https://github.com/kestra-io/kestra/pull/19320, which closes https://github.com/kestra-io/kestra/issues/19214.

### ✨ Description

The banner was easy to read past, and the "See what changed" link rendered as body text inside a sentence, so the one actionable part of the notice was the least visible.

This keeps the full-width bar at the top of the app shell, which is where transient system alerts live in GitHub, GitLab, Grafana, Sentry and Vercel, and fixes the readability instead:

- The CTA is a real button (`KsButton type="primary"`, rendered as an anchor via `tag="a"`) rather than an inline link.
- The message is `--ks-text-primary` at `--ks-font-size-base`, instead of the low-contrast info tint the alert applies by default.
- The row is left aligned rather than centred.

Everything else is unchanged: same message, same info colors around it, same per-user dismissal and the same 30-day window. The backend, the composable, the storage key and the i18n keys are untouched, so no translation regeneration.

The whole change is contained in `VersionUpgradeNotice.vue`. This branch briefly moved the notice into the side menu instead; that commit is still in the history, and the follow-up commit puts `App.vue` and `SideBar.vue` back exactly as #19320 had them.

### 🎨 Frontend Checklist

Screenshots are in the thread. @AcevedoR posted one of the side-menu version in https://github.com/kestra-io/kestra/pull/19410#issuecomment-5660972096, and @bledoux is posting the current left-aligned bar in light and dark.

### 📝 Additional Notes

Design decisions came out of the spike discussed with @mcomi and @AcevedoR: position from the original banner, styling from the sidebar card, left alignment on top.

The CTA is primary violet rather than blue, so it is deliberately the loudest thing in the bar. If it should stay in the info family, `type="info"` on the `KsButton` gives a surface background with blue text and border instead. One prop.

Also, the two existing unit tests were kept and only their stubs updated, since what they assert (the guide URL pins the patch to `.0`) is unchanged by the redesign.

The cat from the previous revision has reviewed this one too and walked away without comment, which I am choosing to read as approval. 🐱

https://claude.ai/code/session_01941d6RC4Mh86wFhxX23B9e

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -2,7 +2,6 @@
     <DocIdDisplay />
     <ErrorToast v-if="coreStore.message" :noAutoHide="true" :message="coreStore.message" />
     <component :is="SdkDriftBanner" v-if="SdkDriftBanner" />
-    <VersionUpgradeNotice v-if="loaded && route?.name && !route.meta?.anonymous" />
     <div id="app-shell">
         <AppTopNavBar  v-if="loaded && route?.name && !route.meta?.anonymous"  />
         <div id="app-body">
@@ -39,7 +38,6 @@
     import AppTopNavBar from "./components/layout/AppTopNavBar.vue"
     import DocIdDisplay from "./components/DocIdDisplay.vue"
     import UnsavedChangesDialog from "./components/UnsavedChangesDialog.vue"
-    import VersionUpgradeNotice from "./components/VersionUpgradeNotice.vue"
     import DrillDownDrawer from "./components/dashboard/DrillDownDrawer.vue"
     import PwaInstallPrompt from "./components/PwaInstallPrompt.vue"
     import {useThemeCycle} from "./composables/useThemeCycle"
--- a/ui/src/components/VersionUpgradeNotice.vue
+++ b/ui/src/components/VersionUpgradeNotice.vue
@@ -1,23 +1,33 @@
 <template>
-    <KsAlert
-        v-if="visible"
-        type="info"
-        center
-        closable
-        class="banner"
-        @close="dismiss"
-    >
-        <template #title>
+    <aside v-if="visible" class="version-upgrade-notice">
+        <div class="notice-header">
+            <InformationOutline :size="16" />
+            <KsIconButton size="xs" :tooltip="$t('close')" placement="right" @click="dismiss">
+                <Close :size="14" />
+            </KsIconButton>
+        </div>
+        <p class="notice-message">
             {{ $t("versionUpgradeNotice.message", {version: notice!.to}) }}
-            <KsLink type="primary" :href="migrationGuideUrl" target="_blank">
-                {{ $t("versionUpgradeNotice.cta") }}
-            </KsLink>
-        </template>
-    </KsAlert>
+        </p>
+        <KsButton
+            class="notice-cta"
+            type="primary"
+            size="small"
+            tag="a"
+            :href="migrationGuideUrl"
+            target="_blank"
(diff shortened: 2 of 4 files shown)


elevatebart:
`Generate PR docker image / Publish Docker` is red on `fe8fd31`, and it is not this PR's failure. Everything else on this head is green: frontend tests, translations parity, build artifacts, file-change detection and the EE trigger. E2E was still running when I wrote this.

The buildx error is:

```
ERROR: failed to build: failed to solve: failed to parse stage name
"kestra/kestra:feat/version-upgrade-notice": invalid reference format
```

The `Dockerfile` takes the base image as a build arg (`ARG BASE_IMAGE`, then `FROM ${BASE_IMAGE}`), and `kestra-io/actions/.github/workflows/kestra-oss-pullrequest-publish-docker.yml` composes that value from the PR's target branch. A Docker tag cannot contain a slash, so a base ref of `feat/version-upgrade-notice` produces a reference Docker refuses to parse. This PR stacks on https://github.com/kestra-io/kestra/pull/19320 rather than on `develop`, which is the whole cause. The same job would fail on any PR targeting a branch with a `/` in its name, and it would pass on this exact diff if it were retargeted to `develop`.

No fix exists to port. The failing string is built in `kestra-io/actions`, which is a different repository, so the patch cannot live in this PR, and the change is a one-liner there: sanitize the ref before using it as a tag, e.g. `${BASE_REF//\//-}`, or fall back to `develop` when the base ref is not a valid tag. Happy to open that upstream if a maintainer wants it, though it only ever bites stacked PRs like this one.

Two notes for whoever looks at the red cross:

- `generate-pull-request-docker-image` is not among the `needs` of the `required-checks` gate in `pull-request.yml`, so this does not block merging.
- I did not spend a re-run on it. The failure is deterministic by construction, the base branch name is right there in the error, so a second run would produce the same line.

The practical consequence is only that no `ghcr.io/kestra-io/kestra-pr:19410` image is published for manual testing. Pulling the PR branch and running it locally is the workaround, or reviewing this on top of #19320 once that merges to `develop`.

---

AcevedoR:
here whats it looks like 
image

elevatebart:
Correcting my earlier comment: the PR image does exist now.

`Generate PR docker image / Publish Docker` passed on `0c81b12` and published `ghcr.io/kestra-io/kestra-pr:19410`, so ignore the part of https://github.com/kestra-io/kestra/pull/19410#issuecomment-5636905001 that says no image would be published and to pull the branch by hand.

```bash
docker run --pull=always --rm -it -p 8080:8080 --user=root -v /var/run/docker.sock:/var/run/docker.sock -v /tmp:/tmp ghcr.io/kestra-io/kestra-pr:19410 server local
```

The base branch still has a slash in its name, and the same job failed twice on the previous head with `invalid reference format`, so something changed outside this repository between Friday and this morning. The reusable workflow is referenced `@main` from `kestra-io/actions`, which I cannot read from here, so I am not going to guess at which change fixed it. The diagnosis in the earlier comment was right about the cause and wrong about it still being true today.

CI is otherwise green on this head: required checks, E2E, frontend tests, translations parity, build artifacts and the EE trigger.

---

PR #19412, opened 2026-09-11
feat(ui) :version badge inside the Kestra menu

### LINKED ISSUE #19198: [UI] Move the version badge inside the Kestra menu
### Why

The bottom-left corner of the sidebar is cluttered: `Instance Owner`, `Settings` + version badge, the environment badge, the tenant switcher and the user card are all stacked in the same small area. `Settings` is the redundant one — it is already reachable from the user menu, which is where users expect it (Notion, Claude, and most apps do the same).

### Proposal

Drop the `Settings` row from the sidebar footer, keeping Settings in the user menu only.
If we want to preserve one-click access, replace the row with a bare gear icon (no "Settings" label) placed somewhere more space-efficient.

Open question for design/frontend: the version badge (`v.2.1`) currently rides on that row — it needs a new home (user menu, or next to the logo).

rk192324217:
Closes https://github.com/kestra-io/kestra/issues/19198

### Description

Moves the version badge from the Admin row into the Kestra user menu, as proposed in the issue discussion.

The version entry preserves the existing version information popover and displays:

1. Version
2. Commit ID
3. Commit date

The version badge is now displayed inside the user menu as `v2.1.0`.

### Expected Behaviour

1. The version badge is displayed inside the Kestra user menu.
2. Hovering over the version `v2.1.0` opens the version information popover.
3. The popover displays the version, commit ID, and commit date.

https://github.com/user-attachments/assets/93c03a52-cfaa-4b5f-b5ef-91f70041be43

### Frontend Checklist

- [x] Screenshots or video recordings attached showing the UI changes


MilosPaunovic (approved):
Approved 🚀 Thanks for the contribution!

I pushed one commit on top of yours instead of sending it back for another round:

- The row moved into a shared `ui/src/components/layout/VersionMenuItem.vue`. `AdminItem.vue` is imported by `EE` as well, while `Auth.vue` is replaced there, so the version info would have disappeared from the `EE` `UI`. The `EE` companion `PR` renders the same component.
- Restored the `v-if` guards on `commitId` and `commitDate`. Both are null when `git.properties` is not on the classpath, and `dateUtils.dateFilter(undefined)` renders the current date, so the popover would have shown today as the commit date.
- `onCommand` now ends with `else if (command === "logout")`, so the empty `version` branch is no longer the only thing keeping a click on the row from logging the user out.
- Reverted the incidental re-indentation of `Auth.vue` so the diff carries only the new row, switched the icon to `InformationOutline`, and put the badge back at `--ks-font-size-2xs` to match the mockup.

PR #19413, opened 2026-09-12
fix(ui): clarify execution dashboard titles(#19009)

### LINKED ISSUE #19009: [Dashboards] Two panels share the title "Total Executions" on the default dashboard
### Describe the issue

On the Dashboards overview, two panels are both titled "Total Executions". One is the
time series, subtitled "Executions duration and count per date", the other is the donut,
subtitled "By status". Only the subtitle tells them apart.

The same pair sits on the default namespace and default flow dashboards, and there the
donut has no subtitle, so both titles read exactly the same.

### Steps to reproduce

1. Open Dashboards and look at the two panels under the KPI row.
2. Open a namespace, then a flow, and look at the same two panels there.

### Actual behaviour

Two panels titled "Total Executions". On the overview the subtitle is the only difference.
On the namespace and flow dashboards there is no difference at all.

### Expected behaviour

Each panel gets a title that says what it shows, for example "Executions over time" for
the time series and "Executions by status" for the donut. Applied to the main, namespace
and flow default dashboards.

Image

### Environment

Version: 2.0.0-rc13
Commit id: 7823c4b
Date: Tue, Sep 1, 2026 11:04 AM

yunuschoudhurywork-lang:
### 🔗 Related Issue

Closes #19009.

### ✨ Description

The default dashboards currently use the same `Total Executions` title for two different execution charts, making them difficult to distinguish when the subtitles are unavailable.

This PR gives each panel a descriptive title:

- `Executions over time` for the execution time-series chart.
- `Executions by status` for the execution status donut chart.

The titles are updated consistently across the default main, namespace, and flow dashboards.

### 🛠️ Backend Checklist

- [ ] Code compiles and tests pass for the touched modules (`./gradlew :module-name:test`, or `./gradlew build` for cross-module changes)

### 📝 Additional Notes

The change is limited to the three default dashboard definition YAML files. No changes were made to the generic `TimeSeries` or `Pie` chart implementations.

`git diff --check` passes successfully, and the final diff contains only the six intended `displayName` changes.

### 🤖 AI Authors

N/A.

### DIFF
--- a/core/src/main/resources/dashboards/default_flow_definition.yaml
+++ b/core/src/main/resources/dashboards/default_flow_definition.yaml
@@ -92,7 +92,7 @@ charts:
   - id: total_executions_pie
     type: io.kestra.plugin.core.dashboard.chart.Pie
     chartOptions:
-      displayName: Total Executions
+      displayName: Executions by status
       graphStyle: DONUT
       tooltip: ALL
       width: 6
@@ -110,7 +110,7 @@ charts:
   - id: total_executions_timeseries
     type: io.kestra.plugin.core.dashboard.chart.TimeSeries
     chartOptions:
-      displayName: Total Executions
+      displayName: Executions over time
       description: Executions duration and count per date
       legend:
         enabled: true
--- a/core/src/main/resources/dashboards/default_main_definition.yaml
+++ b/core/src/main/resources/dashboards/default_main_definition.yaml
@@ -83,7 +83,7 @@ charts:
   - id: total_executions_timeseries
     type: io.kestra.plugin.core.dashboard.chart.TimeSeries
     chartOptions:
-      displayName: Total Executions
+      displayName: Executions over time
       description: Executions duration and count per date
       legend:
         enabled: true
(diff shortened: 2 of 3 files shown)


flcarre (approved):
Verified: correct, minimal and well-scoped.

- In all three files the Pie chart (grouped by `STATE`) is now **Executions by status**, and the TimeSeries chart (date column, COUNT + duration SUM per date) is now **Executions over time**. Both match the underlying `data`/`columns` aggregation in the same file, and the wording issue #19009 asked for.
- These `chartOptions.displayName` values are rendered as-is by `getChartTitle()` (`ui/src/components/dashboard/composables/useDashboards.ts`) — backend-supplied config, not i18n keys — so there is nothing to add to `en.json` and the translations check is unaffected.
- Nothing asserts the literal string 'Total Executions' (checked `DashboardControllerTest` and the other default-dashboard consumers on develop), so no test breakage.

One nit for a follow-up, not blocking: the same time-series definition is duplicated with the still-ambiguous 'Total Executions' in `ui/src/components/dashboard/assets/executions_timeseries_chart.yaml`, `logs_timeseries_chart.yaml`, and the inline `CHART_DEFINITION` in `Flows.vue`. Out of scope for #19009 (which named only the three default dashboards), and the one in `Flows.vue` renders in `short` mode where the title is hidden.

LGTM.

PR #19418, opened 2026-09-14
feat(system): pwa manifest shortcuts for create flow, flows, executions and copilot

elevatebart:
### ✨ Description

Installing Kestra as a PWA gave you an app icon whose jump list (right-click on desktop, long-press on Android) was empty: the manifest declared no `shortcuts`.

It now declares four:

| # | Shortcut | URL | Resolves to |
|---|---|---|---|
| 1 | Create Flow | `flows/new` | `flows/create` |
| 2 | Flows | `flows` | `flows/list` |
| 3 | Executions | `executions` | `executions/list` |
| 4 | AI Copilot | `ai` | `ai` |

Creating a flow leads because it is the only entry the launcher genuinely adds — the rest are one sidebar click away once the app is open. Four keeps the whole list inside the range platforms actually render: Chrome for Android shows three (four before Chrome 92), desktop Chrome and Edge show ten. Logs was dropped for the same reason, and Dashboards is absent because it is already `start_url`.

The URLs are **relative**, matching the existing `start_url` / `scope: "./"` and `base: ""`. The manifest is served from `<contextPath>/ui/manifest.webmanifest`, so `flows` resolves to `<contextPath>/ui/flows`. An absolute `/ui/flows` would have broken every deployment with a non-empty `micronaut.server.context-path`.

The array lives in `ui/plugins/pwaShortcuts.ts` rather than inline in `vite.config.js` so a unit test can import it, and because `ui/plugins/` is already the directory `ui-ee` reaches into by relative path.

### 🎨 Frontend Checklist

Verified end to end on macOS: the shortcuts render in the dock context menu of a PWA installed from this PR's image (`ghcr.io/kestra-io/kestra-pr:19418`). The screenshot below predates the reorder and shows the earlier five-entry list, so treat it as evidence that the menu works rather than as the final labels.

From the emitted `webserver/src/main/resources/ui/manifest.webmanifest` after `npm run build`:

```json
"start_url": "./", "scope": "./",
"shortcuts": [
  {"name": "Create Flow", "url": "flows/new"},
  {"name": "Flows",       "url": "flows"},
  {"name": "Executions",  "url": "executions"},
  {"name": "AI Copilot",  "url": "ai"}
]
```

Resolution under a sub-path deployment: with `context-path: /kestra`, every entry lands inside `scope` at `https://host/kestra/ui/…`.

**Gotcha worth knowing before anyone re-debugs this.** Chrome caches the manifest and only re-checks it on browser restart or after 24 hours, and on macOS the dock menu is built from the generated `.app` bundle written at install time. A Kestra PWA installed before this change therefore keeps showing an empty menu. Uninstall, restart Chrome, reinstall. `chrome://web-app-internals` shows the manifest as Chrome actually parsed it, which is the quickest way to tell "not ingested" from "not rendered".

image-1789373962843

### 📝 Additional Notes

**Labels come from `en.json`, but the manifest stays English.** The four names are read from the keys the app itself uses (`create_flow`, `flows`, `executions`, `ai.flow.title`), so they cannot drift from the product's own wording — TypeScript infers the file's literal shape, which makes a renamed or removed key a compile error rather than an `undefined` label. That is as far as i18n can go here: the manifest is a single build-time artifact with no locale negotiation, so translating it would mean emitting one manifest per language and serving it by `Accept-Language`. Noting that so nobody later mistakes the English-only names for an oversight.

**No shortcut icons.** No 96×96 assets exist in the repo, `icons` is optional per the manifest spec, and four identical copies of the app icon would read worse than none. Text-only entries now; artwork is a follow-up for design, ideally covering EE at the same time.

**EE is not covered, and cannot be yet.** `ui-ee` has no PWA at all: no `vite-plugin-pwa` dependency, no manifest, no `registerServiceWorker()` in `ui-ee/src/main.ts`, no `PwaInstallPrompt` in its `App.vue`. So it is not that the shortcuts miss EE — EE is not installable. Tracked in kestra-io/kestra-ee#11025 rather than smuggled in here.

**One test, and it can fail.** `ui/tests/unit/pwaShortcuts.spec.ts` pins each shortcut to the route it resolves to, and a second assertion fails if a shortcut is added without a pinned destination. Nothing in the app ever navigates through the manifest, so a renamed path would otherwise break the installed jump list in total silence. Verified by pointing `flows` at `flows/search`: 1 failed, 4 passed.

Why did the flow refuse to schedule itself? It was feline a bit under the weather. 🐱

https://claude.ai/code/session_01FGVQHnxx4Xy1n3uK8ph2Et


flcarre (commented):
Nice, small and well-targeted. I checked the five shortcut paths against the real route table: `flows` -> `flows/list`, `executions` -> `executions/list`, `ai` -> `ai`, `logs` -> `logs/list`, `flows/new` -> `flows/create`. Relative URLs are the right call given `base: ""` and `start_url: "./"`, and `setupTenantRouter` rewrites the tenantless `/flows` to `/main/flows`, so the jump list lands where you expect. The test also fails correctly if someone writes an absolute `"/flows"` (it resolves to `//flows` -> wildcard), which is a nice accidental guard on the load-bearing invariant.

A few things worth tightening before merge:

**1. The test can go vacuously green — `ui/tests/unit/pwaShortcuts.spec.ts:13`**

```ts
expect(router.resolve(`/${url}`).name).not.toBe("errors/404-wildcard")
```

The guard is a string literal duplicating a route name defined elsewhere. Rename that record in `src/routes/routes.ts` and this suite passes forever without asserting anything — exactly the silent failure the test exists to prevent. Either derive it (`routes.find(r => r.path.includes(":pathMatch"))?.name`, with an `expect(...).toBeDefined()`), or better, assert the exact destination per shortcut: `{url: "flows", route: "flows/list"}`, `{url: "flows/new", route: "flows/create"}`. That pins the contract instead of just "not 404", and documents where each entry goes.

Also the comment above it describes route *names*, but the shortcuts are paths and the test matches on paths — a renamed path is the real risk.

**2. Ordering and icons — `ui/plugins/pwaShortcuts.ts`**

Five entries, none with `icons`. Platforms truncate the jump list (Android surfaces roughly the first three or four) and without per-shortcut icons render them with a fallback glyph — five text-only rows that look alike, with the last likely never shown. That last one is "New flow", which is arguably the only entry that earns its place: Flows / Executions / Logs / AI Copilot are all one sidebar click away once the app is open, whereas creating a flow straight from the launcher is a genuinely new affordance. I'd put it first, and consider dropping Logs to stay inside the guaranteed-visible range. Adding `icons` matching the sidebar glyphs would be a good follow-up.

**3. Wording — `ui/plugins/pwaShortcuts.ts:8`**

`"New flow"` isn't what the product calls this. `en.json` has `create_flow: "Create Flow"`, the dashboard header CTA for exactly this action. Match it. The other four do match their sidebar labels — which raises the related point that they're hand-copied from `en.json` with nothing catching drift. `vite.config.js` can import `src/translations/en.json` at build time; English-only is unavoidable for a static manifest, but the values don't have to be a second copy.

**One to confirm (not blocking):** `ui-ee/vite.config.js` doesn't configure `VitePWA` at all, so the manifest — and these shortcuts — never reach EE users. Flagging it in case that's unexpected.

Design-system checklist is N/A here: no Vue components, no styles, no tokens, no async states. Reviewed as ordinary frontend/build code.

flcarre (commented):
Nice, small and well-targeted. I checked the five shortcut paths against the real route table: `flows` -> `flows/list`, `executions` -> `executions/list`, `ai` -> `ai`, `logs` -> `logs/list`, `flows/new` -> `flows/create`. Relative URLs are the right call given `base: ""` and `start_url: "./"`, and `setupTenantRouter` rewrites the tenantless `/flows` to `/main/flows`, so the jump list lands where you expect. The test also fails correctly if someone writes an absolute `"/flows"` (it resolves to `//flows` -> wildcard), which is a nice accidental guard on the load-bearing invariant.

A few things worth tightening before merge:

**1. The test can go vacuously green — `ui/tests/unit/pwaShortcuts.spec.ts:13`**

```ts
expect(router.resolve(`/${url}`).name).not.toBe("errors/404-wildcard")
```

The guard is a string literal duplicating a route name defined elsewhere. Rename that record in `src/routes/routes.ts` and this suite passes forever without asserting anything — exactly the silent failure the test exists to prevent. Either derive it (`routes.find(r => r.path.includes(":pathMatch"))?.name`, with an `expect(...).toBeDefined()`), or better, assert the exact destination per shortcut: `{url: "flows", route: "flows/list"}`, `{url: "flows/new", route: "flows/create"}`. That pins the contract instead of just "not 404", and documents where each entry goes.

Also the comment above it describes route *names*, but the shortcuts are paths and the test matches on paths — a renamed path is the real risk.

**2. Ordering and icons — `ui/plugins/pwaShortcuts.ts`**

Five entries, none with `icons`. Platforms truncate the jump list (Android surfaces roughly the first three or four) and without per-shortcut icons render them with a fallback glyph — five text-only rows that look alike, with the last likely never shown. That last one is "New flow", which is arguably the only entry that earns its place: Flows / Executions / Logs / AI Copilot are all one sidebar click away once the app is open, whereas creating a flow straight from the launcher is a genuinely new affordance. I'd put it first, and consider dropping Logs to stay inside the guaranteed-visible range. Adding `icons` matching the sidebar glyphs would be a good follow-up.

**3. Wording — `ui/plugins/pwaShortcuts.ts:8`**

`"New flow"` isn't what the product calls this. `en.json` has `create_flow: "Create Flow"`, the dashboard header CTA for exactly this action. Match it. The other four do match their sidebar labels — which raises the related point that they're hand-copied from `en.json` with nothing catching drift. `vite.config.js` can import `src/translations/en.json` at build time; English-only is unavoidable for a static manifest, but the values don't have to be a second copy.

**One to confirm (not blocking):** `ui-ee/vite.config.js` doesn't configure `VitePWA` at all, so the manifest — and these shortcuts — never reach EE users. Flagging it in case that's unexpected.

Design-system checklist is N/A here: no Vue components, no styles, no tokens, no async states. Reviewed as ordinary frontend/build code.

elevatebart:
Thanks — 1 and 3 were right and are pushed in 5a34184.

**1. Vacuously-green test.** Fixed, taking your second option. Each shortcut now pins its destination rather than asserting "not 404":

```ts
const EXPECTED_ROUTE_BY_URL: Record<string, string> = {
    "flows": "flows/list",
    "executions": "executions/list",
    "ai": "ai",
    "logs": "logs/list",
    "flows/new": "flows/create",
}
```

There's a second test asserting the shortcut URLs and the pinned keys are the same set, so a shortcut added without a destination fails rather than sliding through. Order-insensitive on purpose, since the ordering is still open (below). Verified it fails for the right reason by pointing `flows` at `flows/search`: 1 failed, 5 passed. The absolute-URL guard you spotted survives — `"/flows"` still resolves to the wildcard, which is now a mismatch instead of a near-miss.

Comment reworded to say *path*, which is what the test matches on.

**3. Wording.** Also right: `en.create_flow` is `"Create Flow"` and it's the dashboard header CTA (`src/components/dashboard/components/Header.vue:47`). Renamed. (Note there are two `create_flow` keys — `recipe.create_flow` is `"Create flow"`, lowercase f; I took the top-level one.)

On importing `en.json` into `vite.config.js` to stop drift: I'd rather not, and I don't think it's only cost. It makes every manifest label a function of a translation key, so restructuring or renaming a key silently changes what the OS jump list says, with no test in the way — the same silent-failure shape as the bug you just caught, moved somewhere harder to see. The four matching sidebar labels are matching by intent, not by derivation; a shortcut label and a nav label are free to diverge. Happy to be overruled if you'd rather have the coupling.

**2. Ordering and icons — leaving to @elevatebart.** Not ducking it, but the order is a product decision he made explicitly when scoping this (four destinations first, create action fifth), so I'm not reversing it on my own. Two things that bear on it:

- Empirically, all five render. Verified on macOS from this PR's image, screenshot in the description. Desktop Chrome and Edge render up to 10; the truncation is Android, which shows 3 (it was 4 before Chrome 92). So "likely never shown" applies to the mobile jump list only.
- Your argument for `flows/new` leading is a good one and I don't have a counter beyond "Executions is what people open twenty times a day". If he wants it first, or Logs dropped, it's a one-line change and I'll push it.

Icons: agreed as a follow-up. Deliberately skipped here because no 96×96 assets exist in the repo and five copies of the app icon would read worse than none — it needs real artwork from design, ideally covering EE at the same time.

**EE.** Confirmed, and expected rather than an oversight: `ui-ee` has no PWA at all — no `vite-plugin-pwa` dependency, no manifest, no `registerServiceWorker()` in `ui-ee/src/main.ts`, no `PwaInstallPrompt` in its `App.vue`. So it isn't that the shortcuts miss EE, it's that EE isn't installable. Tracked with the full scope and the agreed EE shortcut order in kestra-io/kestra-ee#11025.

---

elevatebart:
@flcarre point 2 taken in full — @elevatebart agreed. Pushed in 1c2e207:

```js
export const pwaShortcuts = [
    {name: "Create Flow", url: "flows/new"},
    {name: "Flows", url: "flows"},
    {name: "Executions", url: "executions"},
    {name: "AI Copilot", url: "ai"},
]
```

Create Flow leads, Logs dropped, four entries so the whole list stays inside what Android renders. Tests, types, lint and build all clean; the emitted manifest carries exactly those four. Title and description updated to match.

The screenshot in the description predates this and still shows the five-entry list — it stands as evidence the dock menu works, not as the final labels. I've said so inline rather than re-shooting it, since the machine that produced it isn't this one.

Icons remain the open follow-up, for design, covering OSS and EE together.

---

elevatebart:
@flcarre I was wrong to push back on the `en.json` derivation — done in 8500e43, and it's better than the hand-copied version for a reason I'd missed:

```ts
import en from "../src/translations/en.json"

export const pwaShortcuts = [
    {name: en.en.create_flow, url: "flows/new"},
    {name: en.en.flows, url: "flows"},
    {name: en.en.executions, url: "executions"},
    {name: en.en.ai.flow.title, url: "ai"},
]
```

My objection was that a renamed key would silently change what the OS jump list says. It doesn't — TypeScript infers the literal shape of `en.json`, so a removed or renamed key is a **compile error**, not an `undefined` label:

```
plugins/pwaShortcuts.ts(6,18): error TS2339: Property 'create_flow_renamed' does not exist on type '{ loading: string; ... }'
```

Verified by temporarily renaming the reference. That is strictly stronger than the drift protection I was arguing to keep, so the concern I raised was backwards.

Two things I checked before taking it, since they were the other half of my hesitation:

- No `tsconfig` change needed — it type-checks as is, no `resolveJsonModule` to add to the shared base config.
- No type-checking slowdown from inferring a literal type over a file that size: `check:types` runs in ~8.8s.

All four keys resolved to exactly the strings that were hardcoded (`create_flow` → "Create Flow", `flows` → "Flows", `executions` → "Executions", `ai.flow.title` → "AI Copilot"), so the emitted manifest is unchanged — same four entries, same order. Tests, types, lint and build all clean.

That's every point in your review addressed: pinned destinations (5a34184), ordering and wording (5a34184, 1c2e207), labels derived (8500e43), EE confirmed as kestra-io/kestra-ee#11025. Icons remain the open follow-up for design, OSS and EE together.

---

elevatebart:
**CI: `Npm - check ts-any` is failing, and it is not this PR's.** Flagging rather than fixing, since the fix belongs in someone else's file.

```
New `any` in 1 file(s). Type the value, do not raise the baseline.
  tests/unit/composables/useReadOnlyYamlKeys.spec.ts: 0 -> 1
##[error]explicit any went from 0 to 1; type the value instead of raising the baseline
```

That file is not in this diff — this PR touches `ui/plugins/pwaShortcuts.ts`, `ui/tests/unit/pwaShortcuts.spec.ts` and `ui/vite.config.js`. It arrived on `develop` in 9bcdc34 (#18103, merged ~40 minutes before this run), which added `useReadOnlyYamlKeys.spec.ts` carrying

```ts
editor: ref(double.editor) as any,
```

at line 276, without a matching entry in `scripts/explicit-any/baseline.json`. CI builds the PR merge ref, so every open PR against `develop` picks this up; it is not specific to this branch.

**Evidence, rather than assertion.** `npm run check:ts-any` is clean on this branch alone (`No change against the baseline: 1644 explicit any in 429 files`). Dropping `develop`'s copy of that spec and `develop`'s own `baseline.json` into the tree reproduces the failure exactly — both files from `develop`, so the reproduction owes nothing to this branch.

**Proposed patch** (not applied here — it is #18103's file, and pushing it would widen this PR into unrelated code):

```diff
-            editor: ref(double.editor) as any,
+            editor: ref(double.editor) as unknown as Parameters<typeof useReadOnlyYamlKeys>[0]["editor"],
```

`unknown` plus the composable's own parameter type keeps exactly the intent of the cast — a hand-rolled double standing in for a Monaco editor — with no explicit `any`, so it follows AGENTS.md's "fix it by typing the value, not by raising the baseline". Verified: with that one line changed, `useReadOnlyYamlKeys.spec.ts` drops out of the check's output.

I have not re-run the job. A re-run would fail identically — the cause is in the merged base, not a flake — so it would burn a cycle to tell us nothing the local reproduction has not already established.

cc @ameyypawar @flcarre — happy to open a one-line PR against `develop` for this if you would rather not touch it, just say.

---

elevatebart:
Fix for the `check:ts-any` failure is up as #19461, against `develop`. This PR stays red until that merges — the failure is inherited from the base branch, so there is nothing to change here.

Landed on the exported options type rather than the `Parameters<...>` expression I sketched earlier, which reads better:

```diff
+    type ReadOnlyYamlKeysOptions,
 } from "../../../src/composables/useReadOnlyYamlKeys"

-            editor: ref(double.editor) as any,
+            editor: ref(double.editor) as unknown as ReadOnlyYamlKeysOptions["editor"],
```

Verified on `develop` with the composable actually present, rather than on a spliced tree: `check:ts-any` clean (1595 in 417 files), `check:types` clean, the spec's 54 cases all pass, lint 0 errors.

Separately, `trigger-ee` also failed on this head, and it is not the same thing: the job returned an empty verdict and diagnosed itself.

```
EE compile check inconclusive for ref elevate-bot/admiring-davinci-49xrzj
(no compile verdict; HTTP 200). Likely a transient runner error — re-run the job.
```

It passed on the two earlier commits of this branch and produced no verdict at all here, so it is the died-before-it-started case rather than a real failure. I tried to re-run it and GitHub refused with `403 This workflow is already running` (E2E and Publish Docker still in flight). I will re-run it once the workflow finishes; that is the one re-run this PR gets.

---

flcarre (approved):
All three points landed, and the `en.json` derivation is the right call — you're correct that the literal inference turns a renamed key into a compile error rather than an `undefined` label, which is strictly stronger than what I was asking for. Good catch on my behalf.

The pinned-destination test is what I wanted: asserting the mapping rather than "not 404", with the set-equality check so a shortcut added without a destination fails instead of sliding through. Verifying it fails for the right reason by pointing `flows` at `flows/search` is the part people usually skip — thank you.

Four entries with Create Flow leading is right for the Android render budget. No objection to leaving the screenshot as-is with the inline note; it's evidence the dock menu works, not a spec for the labels.

On CI: #19461 merged at 11:42 on 2026-09-15, so the inherited `check:ts-any` failure is gone — the head here is green. Your diagnosis was right and the fix belonged where you put it.

Icons as a follow-up for design, covering OSS and EE together, works for me. Approving.

PR #19419, opened 2026-09-14
fix(executions): new-tab clicks and menu dismissal on nav bar action links

elevatebart:
### 🔗 Related Issue

None — reported directly, from the execution page's "Edit flow" item.

### ✨ Description

Three things were wrong with the links inside the nav bar actions menu (`NavBarAction`, so this covers Executions, Flows, Dashboards and Copilot):

- **Cmd/Ctrl+Click opened the new tab and navigated the current one away.** The dropdown item pushed the route itself on every click, on top of the `RouterLink` it wraps. `RouterLink.guardEvent` deliberately skips `preventDefault` for modifier and middle clicks so the browser can open a new tab, so the imperative push still ran. A plain click on the link navigated twice for the same reason.
- **A plain click left the menu open** until the route change unmounted it. Element Plus skips its own dismissal when the click was `preventDefault`-ed, which is exactly what `RouterLink` does on the clicks it handles.
- **Right-click → "Open link in new tab" left the menu open for good**, since no `click` event is fired at all.

The fallback push now only runs for clicks the link did not handle, the dropdown hands its items a closer built on `KsDropdown`'s exposed `handleClose`, and an item holding a link calls it on both the click and the `contextmenu` path. The popper is `persistent`, so closing only hides it and the anchor stays in the DOM for the native context menu to act on. The link also stretches across the dropdown row, shrinking the padding-only area where the JS fallback is still what navigates.

### 🎨 Frontend Checklist

### 📝 Additional Notes

No screenshot: nothing visible changes, and the three behaviours are all about which tab ends up where. `NavBarAction.spec.ts` covers them instead — the five cases assert on `aria-expanded` and the router push, not on `.kel-*` selectors, and the new ones fail on the commit before this one. The jsdom probe that pinned down the Element Plus behaviour is in the test file's fourth case.

Worth one real click before merge: browsers capture the context-menu target when the menu opens, so hiding the link underneath it is fine in Chrome and Firefox, but it is the one bit I could not verify without a running instance.

### 🤖 AI Authors

Cmd+Click used to leave the execution behind, which is unfortunate, because unlike a cat, a tab does not always land on its feet.

https://claude.ai/code/session_01TREbDc5CLuzmrW9mPzzAmF

---
_Generated by [Claude Code](https://claude.ai/code/session_01TREbDc5CLuzmrW9mPzzAmF)_


RemiBarthe (approved):
Overall this looks good 👍 

**One thing I'd fix before merge:** the `RouterLink` mock currently calls `preventDefault()` on every click, which is the opposite of `guardEvent` for Cmd/Ctrl/Shift/Alt clicks. Since this PR is about modifier-click behavior, the main scenario isn't actually exercised by the current tests. I'd mirror the real `guardEvent` behavior in the mock so the tests match the production flow.

**Non-blocking:** `event.button !== 0` in `isLinkOwnedClick` is unreachable from a `click` listener on modern browsers (`auxclick` handles middle-click), so the comment reads as though that case is covered when it isn't. Either tighten the comment or handle `auxclick` while you're in the file.

PR #19421, opened 2026-09-14
fix(design-system): splitter dragger hit zone over neighbouring scrollbars

### LINKED ISSUE #19420: Flow editor: splitter drag zone covers the code scrollbar, dragging it resizes the panel instead of scrolling
### Describe the issue

In the flow editor, with a second panel open next to the code (Docs, Topology, No-code, Files...), the vertical scrollbar of the code editor can barely be grabbed. Most attempts to drag the scrollbar thumb start resizing the panel instead (the splitter line turns purple and the Docs panel moves). Reaching the top or bottom of a long flow means scrolling with the mouse wheel for a long time.

Recording, Kestra Cloud 2.0, Flow Code panel with the Docs panel open. Repeated attempts to grab the scrollbar between the two panels, each one lands on the splitter:

https://github.com/user-attachments/assets/ff8188d8-7084-4cf8-b662-093788ddaa55

### Cause

Element Plus `ElSplitter` renders a splitter bar of width 0 with an inlined 16px-wide dragger centred on it, so the dragger overlays the 8px of each neighbouring panel that run along the bar. Monaco's vertical scrollbar in the flow editor is 10px wide and sits in that strip. The dragger is also above the scrollbar in stacking order: `.ks-editor` has `z-index: 1` as a grid item, which creates a stacking context, so Monaco's `z-index: 11` on the scrollbar never competes with `.kel-splitter-bar { z-index: 1 }`.

Measured with `document.elementFromPoint` across the 10px of the scrollbar, on Kestra Cloud (2.0) and reproduced against `develop` (same code in `packages/design-system`):

```
D = splitter dragger receives the click, s = Monaco scrollbar receives it

scrollbar visible:   ssDDDDDDDD   (2 of 10 px usable)
scrollbar faded out: ..DDDDDDDD   (0 of 10 px usable, the mouse never reaches Monaco so it never fades in)
```

Coming from the Docs side, the pointer enters the dragger zone before it enters the editor, so Monaco never shows its scrollbar at all. Coming from the code side, only the 2 leftmost pixels of the scrollbar respond.

### Environment

- Kestra Cloud 2.0 (prod), also present on `develop` at b1751a8
- Chrome on macOS, any window width
- Flow Code panel with the Docs panel open on its right

### Expected behavior

Grabbing the scrollbar thumb scrolls the code. The splitter is still grabbable on its own line.

florian-create:
### 🔗 Related Issue

Closes https://github.com/kestra-io/kestra/issues/19420.

### ✨ Description

**Problem.** In the flow editor, with the Docs (or any other) panel open next to the code, grabbing the code editor's vertical scrollbar resizes the panel instead of scrolling. Element Plus `ElSplitter` inlines a 16px-wide dragger centred on a 0px bar, so the dragger covers the 8px of each neighbouring panel that run along the bar. Monaco's vertical scrollbar in the flow editor is 10px wide and sits exactly in that strip, and the dragger wins the hit test because `.ks-editor { z-index: 1 }` on a grid item traps Monaco's own `z-index: 11` inside a stacking context.

**Fix.** Halve the dragger's hit zone to 8px in `KsSplitter.vue`, so it only overlaps 4px of each neighbour. Under `(pointer: coarse)` the original 16px is kept. The overlap is not mouse-specific — Monaco's scrollbar is driven by `pointerdown` and the dragger binds `touchstart` with `touch-action: none`, so touch hits it the same way — but the editor is panned directly on touch, so conceding that strip costs far less there than an 8px handle does under a finger. Coarse pointers are therefore unchanged from `develop`; fine pointers take a trade rather than a strict win, since the rule is design-system wide and every `KsSplitter` consumer gives up half its drag handle to buy back the scrollbar. Nine views are affected (`BlockEditor`, `Dependencies`, `NamespaceFilesEditorView`, `ExecutionVariableExplorer`, `ContextDrawer`, `MultiPanelGenericEditorView`, `MultiPanelTabs`, `BlueprintDetailView`, `FlowsSearch`); the collision is not Monaco-specific, any scrollable region abutting a splitter has it.

Sizes stay in px rather than the `--ks-spacing-*` scale on purpose: they are calibrated against Monaco's `verticalScrollbarSize`, which is device-independent px and does not follow root font size. At a 20px root, `--ks-spacing-2` (0.5rem) would resolve to 10px and cover the scrollbar completely. The width is inlined by Element Plus (`draggerStyles` in `split-bar.vue`, no prop for it), hence the `!important`. The visible 2px line, hover colour and collapse buttons are untouched. 8px is still twice the width of VS Code's default sash.

**Evidence.** `document.elementFromPoint` sampled across the 10px of the Monaco scrollbar, on Kestra Cloud with the same CSS injected and on a local repro (`ElSplitter` + Monaco with the same options):

```
D = splitter dragger receives the click, s = Monaco scrollbar receives it

before   ssDDDDDDDD   (2 of 10 px reach the scrollbar)
after    ssssssDDDD   (6 of 10 px reach the scrollbar)
```

Those figures are for a 0px bar. In the flow editor the bar is not 0px: `MultiPanelGenericEditorView.vue` carries `default-theme`, which overrides the bar to `width: 2px` (horizontal) and `height: 4px` (vertical). The dragger is centred on it, so there the overhang per side is 7px before the fix and 3px after, and the fix is marginally better than the sampling above suggests.

Recording of the original problem (Kestra Cloud 2.0, Flow Code + Docs panels):

https://github.com/user-attachments/assets/ff8188d8-7084-4cf8-b662-093788ddaa55

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit`). `KsSplitter.test.ts` still covers rendering only — jsdom does not evaluate CSS — but the hit zone is now pinned in the storybook browser project (`npm run test:storybook`): `getComputedStyle` assertions on both orientations in `KsSplitter.stories.ts`. Verified to have teeth: dropping the override turns both red at `16px`. Chromium reports `pointer: fine`, so the `(pointer: coarse)` branch stays uncovered.
- [ ] ~~Translations~~ not applicable, no `en.json` change
- [x] Recording of the problem above; the change itself is a hit zone, not a visible one

### 📝 Additional Notes

- 4px of the scrollbar stay under the dragger. Going below 8px felt too thin for a resize handle; if reviewers prefer zero overlap, the alternative is to give the bar a real width, which changes the panel size maths in `ElSplitter`.
- The `z-index: 1` on `.ks-editor` is what makes the dragger win over Monaco's `z-index: 11`. I left it alone: removing it would only help when the pointer comes from the code side, since coming from the Docs side the pointer enters the dragger before Monaco ever shows its scrollbar.
- The scrollbar thumb in `KsEditor.vue` is styled `13px` with a `4px` border on a 10px scrollbar, which leaves a 5px visible core. Worth a separate look, out of scope here.

### 🤖 AI Authors

Drafted by Claude for Florian, who found the bug. Cat joke as requested: our cat also grabs the wrong bar every time, but at least when it resizes the panel it lands on its feet. 🐱


ui/packages/design-system/src/components/Basic/KsSplitter/KsSplitter.vue
```diff
@@ -19,4 +19,16 @@
 <style lang="scss">
     @use '../../../assets/styles/el-ns';
     @use 'element-plus/theme-chalk/src/splitter';
+
+    // Element Plus inlines a 16px hit zone on the dragger, centred on a 0px bar,
```
flcarre: 🟡 [Guidelines] Narrative comment block exceeds the one-sentence convention
Problem: AGENTS.md is explicit that a *why* comment (and this genuinely is one — an undocumented upstream inline-style quirk is a good reason for one) should be kept to one sentence; "when a paragraph seems necessary, make the code clearer instead." This is 4 lines of narration restating the issue's investigation.
Fix: condense to one line and link the issue instead of re-explaining the whole investigation inline.
```suggestion
    // Element Plus centers a 16px inline hit-zone on the dragger with no prop/CSS var to configure it; halved here so it stops swallowing clicks on the 10px Monaco scrollbar next to it (kestra-io/kestra#19420).
    .kel-splitter-bar__dragger-horizontal {
        width: 8px !important;
    }

    .kel-splitter-bar__dragger-vertical {
        height: 8px !important;
    }
```

flcarre (commented):
Reviewed the CSS fix against the Element Plus splitter source (`split-bar.vue`) and the app's existing `KsSplitter` overrides. The core mechanism checks out: the dragger's 16px width/height is set via an inline `style` binding (`draggerStyles` in `split-bar.vue`) with no prop or CSS var exposed for it, so `!important` on the class is genuinely the only lever available here, not a shortcut around a supported API. The unscoped `<style>` block is also the established (and necessary) pattern in this file: `ElSplitterBar`'s internals are rendered by a different component, so a `scoped` style in `KsSplitter.vue` couldn't reach them regardless. The collapse buttons are confirmed untouched — `.el-splitter-bar__collapse-icon` and its `*-collapse-icon-start/end` variants are separate elements sized by their own theme-chalk rules (16×24 / 24×16), independent of `.dragger-*`.

Two things worth addressing before merge, left as inline comments below: the touch/coarse-pointer hit-zone regression, and the narrative comment block (project convention keeps a *why* comment to one sentence, not a paragraph).

Two more for the PR description rather than the code: the checklist says unit tests pass, which is true, but `KsSplitter.test.ts` doesn't actually exercise this fix (it only asserts render + slot content) — worth being explicit that the fix is verified manually (`elementFromPoint` sampling) rather than by an automated regression test. And since the change is in the design system, it applies globally to all 9 `KsSplitter` consumers at once (`BlockEditor`, `Dependencies`, `NamespaceFilesEditorView`, `ExecutionVariableExplorer`, `ContextDrawer`, `MultiPanelGenericEditorView`, `MultiPanelTabs`, `BlueprintDetailView`, `FlowsSearch`) — the PR's evidence and recording only cover the flow-editor/Docs-panel case from the linked issue, worth a quick manual pass on the others too.

## Kestra OSS Code Review

### Kestra Guidelines (frontend/design-system) — 2 findings
🟡 Narrative comment exceeds the one-sentence convention (inline)
🟢 No regression test/story added for this fix — acceptable given a real hit-test regression needs Playwright/E2E rather than jsdom, but the PR checklist's "unit tests pass" line should not be read as coverage for this specific change

Pure CSS change, no data flow, no injection/access-control surface.

Negligible: two additional CSS rules.

### Accessibility / Regression — 1 finding
🟠 Dragger hit-zone halved for touch/coarse-pointer input with no carve-out (inline)

**Verdict:** REQUEST CHANGES — the fix itself is sound and minimal, but the touch/coarse-pointer hit-zone regression should be addressed (or explicitly accepted with a note) before merge, and the comment block should be trimmed to match project convention.

ui/packages/design-system/src/components/Basic/KsSplitter/KsSplitter.vue
```diff
@@ -19,4 +19,16 @@
...
+    // so it overlays the 8px of each neighbouring panel that sit along the bar.
+    // Monaco's vertical scrollbar is 10px wide and lives exactly there: grabbing
+    // it resized the panel instead of scrolling. Halve the zone so 4px overlap.
+    .kel-splitter-bar__dragger-horizontal {
+        width: 8px !important;
+    }
+
+    .kel-splitter-bar__dragger-vertical {
+        height: 8px !important;
+    }
```
flcarre, line 44: 🟠 [Accessibility/Regression] Dragger hit-zone halved uniformly, including for touch/coarse-pointer input
Problem: `split-bar.vue` wires `onTouchStart`/`onTouchMove` and `touchAction: none` on this exact dragger, so touch dragging is a supported interaction, not an edge case. This rule halves its hit zone from 16px to 8px for every pointer type, with no carve-out. The app already recognizes this need: `ExecutionVariableExplorer.vue` has an `@media (max-width: 768px)` override that widens `.kel-splitter-bar` specifically so it stays grabbable on small/touch screens (line ~630). This PR's global rule cuts against that precedent for every `KsSplitter` consumer (9 in the app) at once, with no touch-specific exception and no mention of touch impact in the PR's evidence (only desktop `elementFromPoint` sampling with a mouse).
Fix: scope the reduction to fine pointers and keep the original 16px for coarse/touch input, e.g.:
```suggestion
    .kel-splitter-bar__dragger-horizontal {
        width: 8px !important;
    }

    .kel-splitter-bar__dragger-vertical {
        height: 8px !important;
    }

    @media (pointer: coarse) {
        .kel-splitter-bar__dragger-horizontal {
            width: 16px !important;
        }

        .kel-splitter-bar__dragger-vertical {
            height: 16px !important;
        }
    }
```

ui/packages/design-system/src/components/Basic/KsSplitter/KsSplitter.vue
flcarre: Addressed in ddf72e5. The block is gone; `$dragger-hit-zone-fine` / `$dragger-hit-zone-coarse` carry the meaning now, and one line marks the `!important` as the Element Plus inline-style workaround. The full rationale stays in the PR description where it belongs.

ui/packages/design-system/src/components/Basic/KsSplitter/KsSplitter.vue
flcarre: Addressed in ddf72e5 — added a `@media (pointer: coarse)` block restoring the original 16px. Touch never grabs the Monaco scrollbar by pointer (you pan the content directly), so there is nothing to steal on coarse input and the narrowed target was pure loss there. Fine pointers keep the 8px fix; the coarse block is later in source order, so it wins at equal specificity.

Verified by compiling the style block with `sass` against the real `element-plus` load path: output is correct, and it confirms the visible 2px line comes from `::before`/`::after` rules that dragger sizing never touches. `KsSplitter.test.ts` still passes (4/4).

flcarre: Correction on my own earlier reply here — the rationale I gave was wrong, and a second review round caught it.

I claimed touch never grabs the Monaco scrollbar by pointer. That is false, verified in the installed sources: Monaco's slider listens on `POINTER_DOWN` and `_onPointerDown` explicitly branches `if (isLeftClick || !isMouse)` (`abstractScrollbar.js`), and the Element Plus dragger binds `onTouchstart` with an inline `touchAction: "none"` (`split-bar.vue`). Touch hits the overlap exactly like a mouse does.

The 16px coarse-pointer restoration still stands, but for a different reason: the editor is panned directly on touch, so conceding the scrollbar strip costs far less there than an 8px target does under a finger. Relative to `develop` (16px everywhere) no pointer type ends up worse off — fine pointers gain the fix, coarse pointers are unchanged.

Amended in fba80d3, with the commit message and the PR description corrected too.

On the spacing tokens: I did not switch `8px`/`16px` to `--ks-spacing-2`/`--ks-spacing-4`. They match only at a 16px root font size. The values are calibrated against Monaco's `verticalScrollbarSize` (`useKsEditor.ts:221`), a device-independent px constant that does not scale with text size — at a 20px root, `--ks-spacing-2` resolves to 10px and covers the scrollbar completely, reintroducing the bug for users who enlarge their text. The bespoke SCSS variables are gone, so the one-off pattern is resolved without the rem hazard.

ui/packages/design-system/src/components/Basic/KsSplitter/KsSplitter.vue
flcarre: Condensed to a single line in c3b2d17 — the paragraph is gone and the rationale stays in the PR description.

ui/packages/design-system/src/components/Basic/KsSplitter/KsSplitter.vue
flcarre: Rebased on develop in c3b2d17; the `(pointer: coarse)` carve-out is unchanged and still restores the original 16px, with the corrected rationale above.

ui/packages/design-system/src/components/Basic/KsSplitter/KsSplitter.vue
```diff
@@ -19,4 +19,23 @@
 <style lang="scss">
     @use '../../../assets/styles/el-ns';
     @use 'element-plus/theme-chalk/src/splitter';
+
+    // Element Plus inlines the dragger size; px, not rem, to stay matched to Monaco's verticalScrollbarSize.
```
MilosPaunovic: Optional nit: `verticalScrollbarSize` lives in the same package, and nothing but this comment ties the two numbers together, so naming the file makes the coupling navigable from here.

```suggestion
    // Element Plus inlines the dragger size; px, not rem, to stay matched to Monaco's verticalScrollbarSize (useKsEditor.ts).
```

MilosPaunovic (approved):
Checked the fix against the installed `element-plus` sources and against every `KsSplitter` consumer. The mechanism holds up:

- `split-bar.vue` inlines the dragger size through `draggerStyles` (`width: 16px` / `height: 100%` horizontal, `width: 100%` / `height: 16px` vertical) with no prop or `CSS` variable to configure it, so an `!important` class rule is the only lever available, and an author `!important` does beat a non-important inline declaration.
- The dragger is `position: absolute; top: 50%; left: 50%; transform: translate(-50%, -50%)` in `theme-chalk/src/splitter.scss`, so halving the width shrinks the overlap symmetrically instead of shifting it to one side. The visible line is the dragger's `::before` (`2px`, centred) and the collapse icons are separate absolutely positioned elements at `left: ±12px`; neither is affected.
- Both blocks carry `!important` at equal specificity, so the later `(pointer: coarse)` block wins on coarse input as intended. `kel` is the right namespace (`_el-ns.scss` sets `$namespace: 'kel'`), and nothing in `ui-ee` uses `KsSplitter`, so there is no `EE` counterpart to keep in sync.

Three things below: one worth doing before merge, one question, one nit.

### 1. The hit zone is testable in this package

The checklist explains the missing test with jsdom not evaluating `CSS`, which is true of the `unit` project, but `packages/design-system/vitest.config.ts` also defines a `storybook:*` project running in headless Chromium through `@vitest/browser-playwright`, and `getComputedStyle` assertions in `play` functions are already the house pattern there (`KsButton.stories.ts:249` asserts `marginLeft` is `8px`, `KsDropdown.stories.ts:198` reads a `::before`). The whole fix is one computed value, so two lines in `KsSplitter.stories.ts` pin it, and would catch the realistic regression: an `element-plus` upgrade renaming the dragger classes, or the override silently ceasing to apply.

In `Default`'s existing `play`:

```ts
const dragger = canvasElement.querySelector(".kel-splitter-bar__dragger-horizontal")!
await expect(getComputedStyle(dragger).width).toBe("8px")
```

and for `Vertical`, which has no `play` yet:

```ts
async play({canvasElement}) {
    const dragger = canvasElement.querySelector(".kel-splitter-bar__dragger-vertical")!
    await expect(getComputedStyle(dragger).height).toBe("8px")
},
```

Chromium reports `pointer: fine`, so the fine-pointer value is the stable one to assert; leaving the `(pointer: coarse)` branch uncovered is fine.

### 2. The other eight consumers

The change is design-system wide, so it halves the drag target in `BlockEditor`, `Dependencies`, `NamespaceFilesEditorView`, `ExecutionVariableExplorer`, `ContextDrawer`, `MultiPanelGenericEditorView`, `MultiPanelTabs`, `BlueprintDetailView` and `FlowsSearch`, while the evidence and the recording cover only the `flow` editor with the Docs panel. That is a defensible bargain, since any scrollable region abutting a splitter has the same collision and not just Monaco, but it is a bargain rather than a strict improvement, so "no pointer type ends up worse off than `develop`" reads stronger than it is: fine pointers everywhere trade half the handle for scrollbar access. Worth stating that way in the description, and worth confirming the drag still feels right in a couple of the others.

`ContextDrawer` is the one I would actually try. It sets `.kel-splitter-bar { background-color: transparent }` and puts its visible affordance in a `:deep(.kel-splitter__splitter)` block, but that class is never rendered by the installed `element-plus` (the splitter emits only `kel-splitter`, `kel-splitter__horizontal|vertical` and `kel-splitter__mask*`), so the drawer's hover highlight is dead `CSS` today and the dragger is its entire affordance. Pre-existing and not yours to fix here, just the consumer where the reduction shows most.

### 3. The flow editor's bar is 2px, not 0px

`MultiPanelGenericEditorView.vue:16` puts `default-theme` on the `KsSplitter`, and lines 183-190 override the bar to `width: 2px !important` (horizontal) and `height: 4px !important` (vertical). So in the view the issue is about, the dragger overhangs 7px per side before the fix and 3px after, not 8px and 4px; the sampled figures come from a 0px bar. It makes the fix marginally better than advertised, and it is worth a line in the description so the next reader does not re-derive it from the `ElSplitter` defaults.

**Verdict:** approve. The fix is correct and minimal; the story assertion in point 1 is the only thing I would like to see land with it.

ui/packages/design-system/src/components/Basic/KsSplitter/KsSplitter.vue
flcarre: Done in da7a0c7 — the comment now names `useKsEditor.ts`.

Also took point 1: `KsSplitter.stories.ts` asserts the computed hit zone on both orientations, and I checked the assertions have teeth — dropping the override turns both red at `16px`.

Point 2 and point 3 are folded into the description: the fine-pointer claim is restated as the trade it actually is (nine consumers give up half the handle), and the `default-theme` 2px/4px bar is spelled out, so the flow editor overhang is 7px before / 3px after rather than the 8px/4px the 0px sampling suggests. `ContextDrawer` 's dead `.kel-splitter__splitter` rule I left alone as pre-existing.

PR #19422, opened 2026-09-14
fix(logs): bound dynamic taskrun task ids at their source

nkwiatkowski:
## What

Dynamic taskruns (only dbt produces them) carry a runtime-generated task id that can exceed the fixed-width `task_id` DB column. Persisting such an id aborts the batch insert and **shuts down the `LogEntry` queue consumer** (reads as an app crash), and would also fail when the taskrun itself is persisted in the execution.

## Why not just truncate the log/metric rows

Truncating only on the `LogEntry`/`MetricEntry` side hides the problem: the over-long id is still carried on the taskrun and persisted elsewhere, which fails too. The fix has to bound the id **at the source**.

## Fix

Bound the task id at the single point where plugins hand dynamic taskruns to the framework — `RunContext.dynamicWorkerResult(...)` in `DefaultRunContext` — upstream of **both** the logs (emitted synchronously at registration) and the persisted taskrun. An execution holds many task ids, so a stable short hash of the full id is appended to a 250-char prefix (`250 + "-" + 5-digit hash` = 256) to keep distinct taskruns distinct; a single WARN names the full id.

> `WorkerTaskProcessor` (taskrun finalization) is too late — the crashing logs are emitted earlier, at registration. `RunContext.dynamicWorkerResult` covers both.

## Tests

`DefaultRunContextTest`: an over-long dynamic taskrun id is bounded to <= 256 at registration; two ids sharing a 250-char prefix stay distinct.

Closes https://github.com/kestra-io/kestra-ee/issues/11019


core/src/main/java/io/kestra/core/runners/DefaultRunContext.java
```diff
@@ -559,13 +558,42 @@ public List<WorkerTaskResult> dynamicWorkerResults() {
...
+        TaskRun taskRun = workerTaskResult.getTaskRun();
+        if (taskRun == null || taskRun.getTaskId() == null || taskRun.getTaskId().length() <= Task.ID_MAX_LENGTH) {
+            return workerTaskResult;
+        }
+
+        String fullTaskId = taskRun.getTaskId();
+        String suffix = String.format("%05d", Math.floorMod(fullTaskId.hashCode(), 100_000));
+        String boundedTaskId = fullTaskId.substring(0, TASK_ID_TRUNCATED_PREFIX_LENGTH) + "-" + suffix;
+
+        this.logger.logger().warn(
```
loicmathieu: As we didn't own the task generation (it's DBT node ID for ex), and the generated suffix is reproducible, I would recommand to log at DEBUG or don't log at all

loicmathieu (approved):
LGTM

PR #19424, opened 2026-09-14
fix(tenants): tenant guard missed by the router's first navigation

elevatebart:
### 🔗 Related Issue

None filed; found while debugging a blank dashboard in dev mode.

### ✨ Description

Opening `/ui/` while already logged in landed on `/ui/dashboards`, with no tenant segment, and the dashboard rendered empty: no filter bar, no charts, only the top nav. `Dashboard.vue` gates its load on `route.params.tenant`, so nothing was ever fetched.

The tenant path rewrite lived inside `setupTenantRouter`, which `main.ts` calls in `initApp`'s `.then()`. `vue-router` reads `beforeGuards.list()` after the leave-guard queue, so a guard registered synchronously after `app.use(router)` is still picked up; what loses it is registering after an `await`, and `initApp` has several (`registerDesignSystemI18n`, `loadLocaleMessages`, the `moment` locale import) between installing the router and returning. Logged out the bug hid behind the login redirect, which navigates again once the guard exists.

- `tenantGuard` is now exported and passed through `initApp`, next to the auth guard, so it is registered before `app.use(router)`.
- `setupTenantRouter` keeps the `router.resolve` patch and `$routeTo`; the rewrite logic itself is unchanged.
- No new test. A test can reproduce the gap, by awaiting a few microtasks before registering, but it would pin `vue-router`'s timing rather than this app's wiring: it would still pass if someone dropped `beforeEach: tenantGuard` from `main.ts`. The test that would earn its keep is an `initApp` contract test, passing a `beforeEach` through `guards` and asserting the first navigation honours it. Happy to add that here or as a follow-up.

### 🎨 Frontend Checklist

### 📝 Additional Notes

No screenshot attached. What a screenshot would show here is a URL and a populated dashboard, and the e2e suite green on this head asserts that against a real browser and backend, which the unit environment cannot: it exercises the first navigation the fix is about. Manual check if you want it anyway: load `/ui/` while logged in and confirm it lands on `/ui/main/dashboards` with charts.

Rebased on `develop` at `7a7e9b867`, so it carries #19426 and #19429.

Why don't cats play poker in the jungle? Too many cheetahs.

### Related PRs

- Companion: https://github.com/kestra-io/kestra-ee/pull/11027 — same symptom in EE, where the rewrite depends on a remembered tenant that a new browser does not have.

https://claude.ai/code/session_018uaa7WdWh538GESjnntTL6


elevatebart:
@MilosPaunovic this one is ready for your review when you have a minute. It is out of draft, rebased on `develop` at `7a7e9b867` (so it carries #19426 and #19429), and CI is green on `f81556fd6`, e2e included.

The diff is three files: `tenantGuard` moves out of `setupTenantRouter` and goes through `initApp`, so it is registered before `app.use(router)` installs the router and starts the first navigation. The rewrite logic itself is unchanged.

Nothing is blocking except the review; merging unblocks nothing else on my side, so no rush beyond your queue.

---

MilosPaunovic (approved):
Correct fix, and the diagnosis holds. I reproduced both the bug and the fix locally with throwaway specs on the branch, so the three findings below are all about the prose rather than the code, and none of them block.

### What I verified

| Check | Result |
|---|---|
| `npm run check:types` on the branch | passes (`design-system compiled` / `app checked` / `test checked`) |
| `tests/unit/composables/useTenant.spec.ts` | passes |
| `gh pr checks 19424` | 13 passed, 0 failed |
| Guard registered through `initApp` | first navigation to `/logs` lands on `/main/logs` |
| Old wiring (register after awaited microtasks) | first navigation stays on `/logs`, bug reproduces |
| `root` redirect to a tenant-less target | lands on `/main/dashboards`, `params.tenant = "main"` |
| `meta.anonymous` routes on first navigation | left alone, `/login` stays `/login` |
| History semantics of the first-nav rewrite | `history.replace` x1, `history.push` x0, so no back-button trap |

`EE` is unaffected and the two `PRs` are independent: `kestra-ee` never calls `setupTenantRouter`. It patches `router.resolve` itself in its own `main.ts` and does its tenant rewrite inside `authGuard`, which already goes through `initApp`'s `beforeEach` slot. So removing the `beforeEach` from `setupTenantRouter` breaks nothing on the `EE` side, and there is no merge-order constraint against https://github.com/kestra-io/kestra-ee/pull/11027.

### Findings

**1. The doc comment states the wrong mechanism** (`ui/src/composables/useTenant.ts:4`)

> a guard added after the router installs is missed by the first navigation

That is not quite what happens. `vue-router`'s `navigate()` reads `beforeGuards.list()` inside the `.then()` that follows the `beforeRouteLeave` queue, so a guard registered *synchronously* after `app.use(router)` is still picked up, which is why the old spec passed. What actually loses the guard is registering after an `await`, and `initApp` has several (`registerDesignSystemI18n`, `loadLocaleMessages`, the `moment` locale import) between `app.use(router)` and its own return. I confirmed all three cases: before install gets rewritten, after five awaited microtasks does not, synchronously after install does. Since this comment is the only record of a non-obvious constraint, it is worth saying "after any `await`" rather than "after the router installs". The same wording is repeated in `ui/tests/unit/composables/useTenant.spec.ts:7`, where it is false in that file's own environment.

**2. The "no new test" justification in the description does not hold**

> the unit environment registers the guard in the same microtask as the install, so it cannot reproduce the gap that `initApp`'s awaits create

It can. Awaiting a handful of microtasks between `app.use(router)` and the registration reproduces it exactly, which is what I did. The true statement is a different one: such a test pins `vue-router`'s timing rather than the app's wiring, so it would still pass if someone dropped `beforeEach: tenantGuard` from `main.ts`. The test that would earn its keep is an `initApp` contract test, passing a `beforeEach` through `guards` and asserting the first navigation honours it. Not a blocker, but worth correcting the reason so the next reader does not take "cannot be tested" at face value.

**3. The trimmed comment drops the sentence that is now load-bearing** (`ui/src/composables/useTenant.ts:13`)

The old comment explained that omitting `replace` lets `vue-router`'s own first-navigation handling force a replace. That was hypothetical before this `PR`, because the guard never ran on the first navigation. It is now the actual path being enabled, and it is what keeps `/ui/dashboards` out of history (measured above: `replace` x1, `push` x0). Trimming the rationale right as it becomes relevant seems backwards, so I would keep a trace of it.

### Notes, not change requests

- Same latent trap one line over: `createUnsavedChanged(app, router)` registers a `beforeEach` inside `initApp` after `app.use(router)` *and* after the i18n awaits, so it is missed by the first navigation for the identical reason. Harmless today, since nothing is unsaved on a cold start, but it is the same footgun and nothing flags it.
- Arity footgun in `initApp`'s convention: `guards.beforeEach.bind(null, router)` leaves `tenantGuard` with a bound `length` of 2, keeping `vue-router` in return-value mode. A fourth parameter would push it to 3, flipping `vue-router` to `next()`-callback mode where the returned redirect is silently ignored and the navigation hangs. Pre-existing, but this guard is now one parameter away from it.
- `initApp` has a single `beforeEach` slot. `OSS` could take it only because its auth guard is a `beforeResolve`, and `EE` already spends its slot on `authGuard`. Anything `OSS` adds later has to compose by hand.

### Backport

- `releases/v2.0.x` has the same bug and takes a clean backport: byte-identical `setupTenantRouter` and `init.ts` (`guards.beforeEach` support, `app.use(router)` after registration), and its `main.ts` passes only `beforeResolve`, so the guard is missed on the first navigation there too.
- `releases/v1.3.x` has the same bug in substance but is not a cherry-pick: `ui/src/utils/init.js` there has no `guards` parameter at all, and the guard is `next()`-style with an explicit `replace: true`. It would need re-implementing, so that one is your call.

elevatebart:
Thanks, all three findings are right and all three are fixed in `78c8d43e9`.

**1. Wrong mechanism in the doc comment.** Confirmed independently: when I wrote the spec I registered the guard after `app.use(router)` and it still passed, which is exactly your synchronous case. I read that as "the unit environment can't reproduce it" instead of "synchronous registration is fine, awaits are not". The comment now says a guard registered after one of `initApp`'s awaits is missed by the first navigation, and `useTenant.spec.ts:7` says the same.

**2. The "no new test" reason.** Corrected in the description. The honest reason is not that it cannot be reproduced but that reproducing it pins `vue-router`'s timing rather than this app's wiring, so it would stay green if someone dropped `beforeEach: tenantGuard` from `main.ts`. Your `initApp` contract test is the one that would catch that; say the word and I will add it here rather than as a follow-up.

**3. The trimmed rationale.** Restored, and pointed at what it now guards: `vue-router` forcing a replace on the first navigation is what keeps the tenant-less URL out of history, which your `replace` x1 / `push` x0 measurement confirms.

On the notes: `createUnsavedChanged` is the same footgun one line over and I would rather not widen this PR into it, so I will raise it separately unless you would rather it rode along here. The `bind(null, router)` arity trap and `initApp`'s single `beforeEach` slot are both worth writing down somewhere that is not a PR comment; `ui/AGENTS.md` seems right.

On the backports: `releases/v2.0.x` taking this cleanly matches what I saw while backporting #19426 and #19429 there earlier today. `releases/v1.3.x` needing a re-implementation against the `next()`-style guard in `init.js` is a call for the maintainers rather than something I will do unasked.

---

PR #19431, opened 2026-09-14
fix(e2e): address the Gantt row label instead of racing its tooltip

AcevedoR:
`flow.spec.ts` fails intermittently on `Flow Page › should create and execute the example Flow` and `… a Flow with input`, with a strict-mode violation:

```
strict mode violation: getByText('hello', { exact: true }) resolved to 2 elements:
    1) <code>hello</code> aka getByText('hello').first()
    2) <code>hello</code> aka getByRole('code')
  - locator resolved to <span>hello</span>
  - element is not visible … element was detached from the DOM, retrying
```

The Gantt row renders the task name twice — once in the visible row label, and once inside the tooltip content, which is mounted hidden and attaches a moment later (`Gantt.vue`, the `<code>{{ item.name }}</code>` in the `KsTooltip` `#content` slot and the one in `span.task-name`). The unscoped text lookup races that: it either resolves to the hidden copy and never becomes clickable, or trips strict mode once both are up.

Addressing `span.task-name` directly picks the copy the test always meant to click, and the tooltip no longer participates.

Same flake, same fix, in EE: companion PR kestra-io/kestra-ee#11039.


elevatebart (approved):
Get rid of that comment (or make it shorter)

PR #19435, opened 2026-09-14
refactor(deps): replace moment with dayjs behind the design-system date utils

elevatebart:
### 🔗 Related Issue

None — raised from a direct request to get `moment` out of the bundle.

### ✨ Description

`moment` and `moment-timezone` were 34 import sites across three packages, and the packed 1970-2030 zone table sat in the vendor chunk that every page loads. Element Plus already depends on `dayjs`, so we were shipping two date libraries to render the same timestamps.

The design system now owns the date library outright. `packages/design-system/src/date/dayjs.ts` is the single configured instance (utc, timezone, duration, calendar, isoWeek, localizedFormat, minMax, relativeTime, weekOfYear, isSameOrBefore), and `dateUtils` / `durationUtils` are the only import path for the app and for `@kestra-io/topology`. No package outside the design system depends on a date library any more. dayjs's timezone plugin goes through `Intl`, so no zone table ships at all.

That let the injection seam collapse rather than just move: `setMomentInstance`, `setDateFormatter`, their getters and the `$moment` global property are gone. `KsDateAgo` now calls dayjs directly instead of waiting for the app to hand it a formatter at startup, which also means its story and unit test no longer have to hand-roll a fake `fromNow()`. `submitTask`'s `{$moment}` adapter parameter is gone from all three exports.

Two dead dependencies went with it: `moment-range` (`extendMoment()` was called, `.range()` never was — it also carried a patch file whose only job was making its types compile) and `@js-joda/core`, declared with zero imports anywhere.

**Bundle impact**, measured over two full production builds, this branch vs `develop`:

| | before | after | delta |
|---|---:|---:|---:|
| `vendor` chunk, gzipped — loaded on every page | 310,761 B | 252,157 B | **−58,604 B (−19%)** |
| 12 lazy locale chunks combined, gzipped | 11,271 B | 9,642 B | −1,629 B |
| all JS, gzipped | 8,629,824 B | 8,567,437 B | −62,387 B |

Effectively all of the win is the vendor chunk, which is the one that matters since every page pays it. The locale chunks are a wash: moment's per-locale files were already small (0.7–1.6 KB gzipped) and dayjs's are comparable.

Also 12 fewer explicit `any`, and a net −161 lines.

### 🎨 Frontend Checklist

- [x] Unit tests pass (`npm run test:unit` — 262 files, 2332 tests; `npm run test:storybook` — 69 files, 260 tests)
- [x] Translations are complete if `en.json` changed (`en.json` is untouched)

### 📝 Additional Notes

**No screenshots**, which is the one box left unticked: this was done in a headless environment with no running instance. The surfaces worth a manual pass before merge are the Settings timezone dropdown, the executions date-range shortcuts, the dashboard time-series buckets, and any page under a non-English locale.

**Three behaviour differences**, none of them cosmetic:

- `moment(x, ISO_8601, true)` has no dayjs equivalent, and bare `dayjs(5)` reads an epoch number as a date — that would have turned chart category labels into 1970. Replaced with `dateUtils.parseIso()`, regex-gated and tested.
- `moment.duration(x).isValid()` returns true for literally any string, so `TaskString`'s "this must be a Pebble expression" detection on duration fields never actually fired. `durationUtils.isValidDuration()` is a real ISO 8601 check, so it fires now. This is an unrequested fix riding along; say the word and I'll restore the no-op.
- dayjs has no lowercase year token, so `getDateGrouping` moves to `YYYY-MM-DD`. The week bucket went from `yyyy-'W'ww` — which moment rendered as `2026-'35'35`, quotes and all, since `W` is its ISO-week token — to `YYYY-[W]ww` → `2026-W35`.

**One cost I could not erase**: building the Settings timezone list now costs ~100 ms of `Intl.DateTimeFormat` construction across 418 zones, where moment-timezone's precomputed tables were free. `dateUtils.timezonesWithOffset()` memoises it for the session, so it's paid once per page load rather than once per visit to Settings. This is the known trade-off of the Intl-backed timezone plugin and applies anywhere we convert timezones in bulk.

`ui/AGENTS.md` is updated in the same PR with the new utility surface and the two dayjs foot-guns (immutability, and `dayjs()` coercing non-dates).

### 🤖 AI Authors

I asked the cat whether we should keep shipping two date libraries. She knocked one off the table. Vendor chunk down 19%.

https://claude.ai/code/session_01PjfLtebHKZUSY3coERiES1


Piyush-r-bhaskar (changes requested):
Good migration and the right shape: one configured instance in the design system, and the injection seam collapsed rather than ported. `parseIso` and `isValidDuration` are real fixes. OSS is green on my run: `check:types` exits 0, 2203 unit tests across 242 files pass, and no `moment` reference survives in `ui/src` or `ui/packages`.

What blocks it is EE. Three removed seams have live consumers there and no companion PR exists.

### 1. `$moment` removal breaks EE at runtime, invisibly

`ui-ee/src/components/cluster/ServiceEvents.vue:53` calls `$moment.parseZone(timestamp).format()` from the template, once per event row. With `init.ts` no longer setting the global property it throws. EE declares `$moment: any` in `vue.d.ts:22`, so `check:types` stays green and nothing catches it until someone opens Cluster service events.

### 2. `submitTask` signature change breaks EE compilation

`normalizeInputValues` and `inputsToFormData` lost their first parameter; EE still passes it at `CaseActionsCard.vue:107` and `BlockForm.vue:115`. I ran EE `check:types` with this branch checked out in the sibling OSS repo: exactly two errors, both `TS2554: Expected 2 arguments, but got 3`. EE develop goes red on merge.

Both need a companion EE PR landing in lockstep.

### 3. The timezone dropdown silently loses 179 of 597 entries

`Intl.supportedValuesOf("timeZone")` returns 418 canonical zones; `moment.tz.names()` returned 597 including link names. Dropped: `US/Eastern`, `US/Pacific`, `Australia/Canberra`, `America/Montreal`, `America/Argentina/Buenos_Aires`, `America/Indiana/Indianapolis`, `Etc/GMT+5`.

Formatting still works, `Intl` resolves link names fine. The damage is the picker at `BasicSettings.vue:271`: a user whose stored zone is a link name has no matching option and loses the setting if they touch the field. Worth appending the stored zone when it is missing from the list, and worth a line in the description.

### Minor

- `durationUtils.duration("-PT5S")` returns `5`; moment returned `-5`. `isValidDuration` accepts the sign and `duration.test.ts:6` asserts it, so the test blesses input `duration()` then mishandles.
- `packages/design-system/README.md:14` still lists `moment` and `moment-timezone` as peer deps.
- `Intl.supportedValuesOf` is unguarded. Widely available, but without it the whole Settings page throws during setup rather than degrading.
- `timezonesWithOffset()` memoises for the session, so a tab open across a DST transition shows stale offsets. Previously recomputed per mount. Cosmetic.

### Two that look like bugs but are not

- **`yyyy` to `YYYY` is a no-op.** Moment does support lowercase `yyyy`, so nothing was broken before.
- **The week bucket is fine.** `YYYY-[W]ww` renders `2026-W38`. I measured otherwise in a standalone Node script first, but that was CJS interop handing `dayjs.extend` a namespace object. Through the real module `fillTimeBucketLabels` gives distinct weekly labels.

I also confirmed `dayjs.duration(ms).toISOString()` matches moment's shape, `subtract(duration)` works, and locale switching drives both `llll` and relative time.

One suggestion: `utils.spec.ts:135` asserts the week format *string*, never what it renders. An assertion that a weekly grouping yields distinct labels would catch a future plugin-registration slip in `date/dayjs.ts`.

elevatebart:
Thanks — that's a genuinely useful review, and every finding reproduced. Pushed as `66d1fd4`.

### Blockers

**1 & 2 — EE.** Both confirmed and fixed in **kestra-io/kestra-ee#11102**, same branch name so CI pairs them. I reproduced your exact result: without the change, `check:types` in `ui-ee/` gives precisely two errors, `BlockForm.vue(115,78)` and `CaseActionsCard.vue(107,110)`, both `TS2554: Expected 2 arguments, but got 3`; with it, all four stages pass and it exits 0.

`ServiceEvents` now imports `moment` directly rather than migrating to dayjs — EE declares its own `moment` dependency in `ui-ee/package.json`, so the other twelve call sites there are self-sufficient and untouched. Migrating EE is separate work that shouldn't ride on this.

That EE run also caught something neither of us had: EE's **test** tsconfig has a narrower `lib` than OSS's, so `Intl.supportedValuesOf` failed to resolve there. I'd papered over that in OSS by widening `tsconfig.base.json`, which only worked because it was our own config. It's now typed locally in `date.ts`, the `tsconfig.base.json` change is reverted, and no consumer has to widen `lib` to compile the design system.

**3 — timezone link names.** Confirmed: 418 canonical zones against moment-timezone's 597, and every name you listed is absent. `timezonesWithOffset()` takes the stored zone and appends it when the list omits it; `BasicSettings` passes `settings.timezone`. Tests cover the link-name case and the no-duplicate case. You're right that formatting was never affected — `Intl` resolves `US/Eastern` fine, it just isn't enumerated.

### Minor

- **Negative duration** — real, and mine. `dayjs.duration("-PT5S").asMilliseconds()` is `+5000` where moment gave `-5000`. Sign reapplied, test added. Good catch that the validator blessed input the parser then mishandled.
- **README** — fixed.
- **`Intl.supportedValuesOf` unguarded** — now `intl.supportedValuesOf?.("timeZone") ?? []`, so a missing implementation yields an empty list plus the stored zone rather than throwing during setup.
- **DST staleness** — the memo is now keyed on the machine's UTC offset and re-derives when it moves. Cheap enough that "cosmetic" didn't justify leaving it.
- **Your test suggestion** — taken, and it's the better assertion. `fillTimeBucketLabels` over a fortnight now asserts `["2026-W36", "2026-W37", "2026-W38"]`, which fails if the `weekOfYear` registration ever goes missing; the old format-string assertion could not.

### On the two you flagged as non-bugs

You're right about `yyyy`, and my description overclaimed. I verified against real moment: `format("yyyy-MM-DD")` gives `2026-09-16`, identical to `YYYY`. The change was forced by dayjs having no lowercase year token, not by anything being broken. I'll correct the wording.

The quoted week format is the one place I'd push back slightly: real moment renders `yyyy-'W'ww` as `2026-'38'38`, quotes included, since `'` isn't an escape character for it and `W` is the ISO-week token. So the old string was malformed — but as you say, the new one is fine, and it was never the bug I implied in the surrounding sentence. I read your note as correcting your own earlier week measurement rather than that claim; say if you meant otherwise.

### Verification, with one honest gap

`check:types`, `check:ts-any` and lint are clean. Design-system utils (34 tests, including all the new ones) and the touched app specs (131 tests across 5 files) pass. I could **not** complete a full local `test:unit` run on this head — the sandbox kept restarting under it — so the 2408-test figure in the description is from the previous head. CI is running it now; I'll follow up if anything falls out.

Screenshots are still the one unticked box, for the same reason.

---

Piyush-r-bhaskar (approved):
Now it's good 🎉

elevatebart:
`Frontend - Tests` is red for a reason that has nothing to do with this diff: the GitHub App installation's hourly API quota is exhausted, so the steps that call the API fail while every step that runs actual code passes.

Attempt 1 (`d21dc62`): unit tests, storybook, types, ts-any, lint and build all green; the job failed on the post-test step `Report failing UI tests on the PR` with `403 API rate limit exceeded for installation` (`x-ratelimit-used: 5000`).

Attempt 2 (same commit, re-run): the reporting step succeeded (`No UI test failures — nothing to report.`), and the job failed one step earlier instead:

```
Found 67 warnings and 0 errors.
reviewdog: fail to get diff: GET https://api.github.com/repos/kestra-io/kestra/pulls/19435:
403 API rate limit exceeded for installation
##[error]Process completed with exit code 1.
```

eslint itself reported **0 errors**. `reviewdog/action-eslint` runs with `fail_on_error: false` and `fail_level: none`, so the non-zero exit comes from reviewdog being unable to fetch the PR diff, not from a lint violation. `npm run test:lint` on this exact tree locally: 0 errors, 2 pre-existing `vue/enforce-style-attribute` warnings in `ChangeStatus.vue` and `Restart.vue`, neither touched here.

Nothing to port and nothing to fix in the branch. Waiting for the quota window to roll over and re-running the job.

---

